PR #744, opened 2016-04-12
Fix problem when swap memory unsupported

hqhq:
When swap memory is unsupported, Docker will set
cgroup.Resources.MemorySwap as -1.

Fixes: https://github.com/docker/docker/pull/21937


crosbymichael:
LGTM

PR #749, opened 2016-04-13
Fix OCI reference in README

jonboulle:
Signed-off-by: Jonathan Boulle <jonathanboulle@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## runc
 
-`runc` is a CLI tool for spawning and running containers according to the OCF specification.
+`runc` is a CLI tool for spawning and running containers according to the OCI specification.
 
 ## State of the project
 


mrunalp:
@crosbymichael @LK4D4 jenkins is broken. Can someone look into it?

crosbymichael:
@mrunalp it does not look like a jenkins issue, it looks like a packaging issue:

```
The following packages have unmet dependencies:
 curl : Depends: libcurl3-gnutls (= 7.47.0-1) but 7.38.0-4+deb8u3 is to be installed
E: Unable to correct problems, you have held broken packages.
```

mrunalp:
@crosbymichael Hmm, thanks!

mrunalp:
I can reproduce it locally. @tianon Any ideas on how we can fix this?

tianon:
I think this is caused by mixing and matching `jessie` and `stretch` (that line adding `testing` to `sources.list` is really fishy).

@LK4D4 it looks like it's been there since you first added the file -- do you happen to recall what it was necessary for?  Just getting a new enough `criu` (which we now download and install directly) or something more than that?

tianon:
(Worth adding that by removing that line, I can build the image successfully locally, although I'm not positive about whether it will function properly as I haven't run the tests on it yet.)

tianon:
seccomp tests fail, so it looks like we might need to add `jessie-backports` instead; testing further

tianon:
Opened a PR at https://github.com/opencontainers/runc/pull/750 where we can continue discussing. :+1:

crosbymichael:
LGTM

mrunalp:
LGTM

PR #751, opened 2016-04-13
Add -q to list to print only container IDs

mrunalp:

### DIFF
--- a/list.go
+++ b/list.go
@@ -47,11 +47,22 @@ in json format:
 
     # runc list -f json`,
 		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "display only container IDs",
+		},
 	},
 	Action: func(context *cli.Context) {
 		s, err := getContainers(context)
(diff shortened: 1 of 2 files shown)


hqhq:
Maybe update man page as well?

laijs:
Since it has the `--format` (`-f), I think we can make it as one of the format
like this: `--format quiet`  or `--format container-id-only`

mikebrow:
Looks good to me.

mrunalp:
Updated man page.

mrunalp:
@laijs I think typically commands have a separate `quiet` option so I like it better than adding this to quiet. Also format with specific headers is different since it includes the header in the output whereas output of quiet excludes that and is convenient for processing.

mrunalp:
@crosbymichael @LK4D4 PTAL

LK4D4:
LGTM

crosbymichael:
LGTM

PR #758, opened 2016-04-17
Update the comment for container pause

rajasec:
As per current implementation running  container can be paused. Updated the comment to remove pausing state


crosbymichael:
LGTM

PR #759, opened 2016-04-18
Add cause to error messages

crosbymichael:
This is the inital port of the libcontainer.Error to added a cause to
all the existing error messages.  Going forward, when an error can be
wrapped because it is not being checked at the higher levels for
something like `os.IsNotExist` we can add more information to the error
message like cause and stack file/line information.  This will help
higher level tools to know what cause a container start or operation to
fail.


mrunalp:
LGTM

cyphar:
LGTM. Merging.

PR #765, opened 2016-04-21
Add infomation about ocitools in runc spec

hqhq:


hqhq:
So we don't need to do any `runc spec` expansion as https://github.com/opencontainers/runc/pull/764 tried to do.

cyphar:
I'd like it if we added the information to `runc spec --help` as well. Something like "if you want more control of the output `config.json` please use `ocitools`".

hqhq:
@cyphar It is in `runc spec --help`.

cyphar:
Oops. I misread the filename. ;)

LGTM.

mrunalp:
LGTM

PR #767, opened 2016-04-21
Add ps command

hqhq:

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -1,0 +1,76 @@
+// +build linux
+
+package main
+
+import (
+	"fmt"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/codegangsta/cli"
+)
+
+var psCommand = cli.Command{
+	Name:      "ps",
+	Usage:     "ps displays the processes running inside a container",
+	ArgsUsage: `<container-id> <ps options>`,
+	Action: func(context *cli.Context) {
+		container, err := getContainer(context)
+		if err != nil {
+			fatal(err)
+		}
+
+		psArgs := context.Args().Get(1)
+		if psArgs == "" {
+			psArgs = "-ef"
+		}
+
+		output, err := exec.Command("ps", strings.Split(psArgs, " ")...).Output()
+		if err != nil {
+			fatal(err)
+		}
+
+		lines := strings.Split(string(output), "\n")
+		pidIndex, err := getPidIndex(lines[0])
+		if err != nil {
+			fatal(err)
+		}
+
+		pids, err := container.Processes()
+		if err != nil {
+			fatal(err)
+		}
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

ps.go
```diff
@@ -0,0 +1,76 @@
...
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/codegangsta/cli"
+)
+
+var psCommand = cli.Command{
+	Name:      "ps",
+	Usage:     "ps displays the running processes of a container",
```
mrunalp: Maybe reword this a bit "ps displays the processes running inside a container" ?

hqhq:
@mrunalp Fixed, thanks.

cyphar:
I'm not really convinced why we need this. It feels weird to me to have a special case command that does `runc exec` with `ps`...

cyphar:
Can you make it clear (preferably with some docs) that this shows the processes from the perspective of the host? It took me a couple of minutes to figure that out. Also, I feel like there's a nicer way of doing this than doing the equivalent of a `grep` of the output of `ps -ef`.

hqhq:
@cyphar Yeah, the intention is to get pids of the host, if we want pids in the container, we can just use `runc exec ps`. I want to do this because I want `runc kill` to send a signal to any processes in the container, and I think the pid should be id of the host (I don't know how to do it with pid in the container was passed), so we need a way to get these pid information.

What other nicer way of doing this do you have in mind? I think current implementation is pretty neat already. And I think it's pretty the same as `docker top`.

cyphar:
@hqhq You can always make `runc kill` join the PID namespace of the container and kill the processes that way. Then you wouldn't have to worry about the mapping of host-to-container PIDs, since I think that the UX `runc kill <container> <host PID>` looks weird.

To be honest, I can't think of a nice way at the moment that doesn't require iterating over `/proc` but I personally am averse to shelling out to binaries when you don't absolutely have to.

hqhq:
Yeah, it's doable, thanks, but we can talk about `runc kill` later.
Since more and more people are using runc binary, I don't see any reasons why we should prevent useful commands from being shelled out? If `runc ps` shows host pids looks weird to some people, we can add options like `--inner` to show container pids. And I remember that `exec` is never encouraged, neither for docker or runc. So we can do it by `runc exec` should not be a solid reason I think.

cyphar:
Yeah, I think you're right. I'm not sure about how we discourage `runc exec` (it's invaluable in many situations), but I can see that `runc ps` could be useful for some people (especially when the rootfs doesn't have a `ps` binary). I'd be interested in adding the "join the container PID namespace" code, since that also would be quite useful for the `runc kill` case as well as showing a "from the inside" view of what's happening in the container.

PR #768, opened 2016-04-21
Not showing up the events for destroyed container

rajasec:
When the container is destroyed state, there is no process associated with that. Looks like there is no process activity associated with cgroup ( cgroup.procs is empty). Do not show the event statistics for destroyed container.

### DIFF
--- a/events.go
+++ b/events.go
@@ -37,13 +37,12 @@ information is displayed once every 5 seconds.`,
 		if err != nil {
 			fatal(err)
 		}
-		id := context.Args().First()
 		status, err := container.Status()
 		if err != nil {
 			fatal(err)
 		}
 		if status == libcontainer.Destroyed {
-			fatalf("container with id %s is not running",id)
+			fatalf("container with id %s is not running", container.ID())
 		}
 		var (
 			stats  = make(chan *libcontainer.Stats, 1)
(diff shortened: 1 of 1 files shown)


events.go
```diff
@@ -98,3 +99,13 @@ information is displayed once every 5 seconds.`,
 		group.Wait()
 	},
 }
+
+func containerStatus(container libcontainer.Container) {
```
crosbymichael: the name of this function does not tell or hint at what it does.  It maybe better to just move this code into the caller or find a better name that explains its checking if the container is not destroyed, etc....

cyphar: I don't think it makes sense to have a function that calls `fatal()`, since if you just read the calling code it's not clear the function might kill the whole process. Do it in the caller.

rajasec:
@crosbymichael @cyphar 
Updated as per the comments by moving to caller.

events.go
```diff
@@ -37,6 +37,14 @@ information is displayed once every 5 seconds.`,
 		if err != nil {
 			fatal(err)
 		}
+		id := context.Args().First()
```
cyphar: You can just use `container.ID()` here since it's already been loaded by factory.

rajasec:
@cyphar 
Updated.

cyphar:
LGTM

crosbymichael:
LGTM

PR #770, opened 2016-04-22
Add signal mapping to man page

hqhq:

### DIFF
--- a/man/runc-kill.8.md
+++ b/man/runc-kill.8.md
@@ -7,9 +7,48 @@
 Where "<container-id>" is the name for the instance of the container and
 "<signal>" is the signal to be sent to the init process.
 
+The mapping between "<signal>" and the real golang signal is:
+
+    "ABRT":   syscall.SIGABRT,
+    "ALRM":   syscall.SIGALRM,
+    "BUS":    syscall.SIGBUS,
+    "CHLD":   syscall.SIGCHLD,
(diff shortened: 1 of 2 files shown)


man/runc-kill.8.md
```diff
@@ -7,9 +7,48 @@
 Where "<container-id>" is the name for the instance of the container and
 "<signal>" is the signal to be sent to the init process.
 
+The mapping between "<signal>" and the real golang signal is:
+
+    "ABRT":   syscall.SIGABRT,
```
cyphar: I'm fairly sure all of these mappings are the same as in `signal(7)`. Maybe we should just reference that?

hqhq: Do you mean the mapping values? I don't know, they are defined by golang package, why would that be directly related to `signal(7)`?
As for the mapping keys, I think they are defined by runc.

cyphar: The values have to be related to `signal(7)`, because that defines the values that the language uses (it's written by the authors of the kernel). As for the mappings, I'm fairly sure we follow the same convention as `kill(1)` and I don't like copy-pasting Go code into our man pages.

hqhq:
ping @crosbymichael Do you have some reference sources of these signal mappings when you wrote them? I suppose they are just from consts in syscall package?

crosbymichael:
@hqhq ya, no real reference

PR #772, opened 2016-04-22
Updating kcore in validator test

rajasec:
Changing to /proc/kcore instead of kcores


crosbymichael:
LGTM

Issue #38, opened 2015-06-26
Support running runc as non/less privileged user


discordianfish:
Right now runc requires to be run as root where technically it should be possible to run containers as unprivileged user (at least if user namespaces are used)

cgwalters:
See also https://git.gnome.org/browse/linux-user-chroot/tree/README
which supports use by unprivileged users even without user namespaces.

See discussion about use of `PR_SET_NO_NEW_PRIVS` (which is enforced by seccomp) and additional note about local DoS attacks.

LK4D4:
We just need to reconsider some default mounts for this to work I think, maybe drop `/proc` readings in init.

chrisfilo:
This allow containers to be used in shared computing environments such as HPCs. Very exciting!

zeneofa:
I would be very interested to find out how/when this is implemented, especially as it may help me create a transferable environment to use in HPC environments, as there I have no sudo and no chance to install docker.

discordianfish:
So science is interested. Now we need enterprise so somebody will actually start working on this ;)

davidlt:
+1 from HEP (High Energy Physics) community. You can have your hundreds of thousands of cores even with a common operating system like RHEL/CentOS/Scientific Linux, but you still end with Android-like fragmentation because all computing centres do updates on their schedules. When you send your job to various computing centres you also want to provide your container as environment. __Preferably that runs  as unprivileged container.__ The container protects you from the fragmentation and you don't get magic differences due for example due to update of _libm_.

HTCondor bash system already have some support for Docker: https://research.cs.wisc.edu/htcondor/HTCondorWeek2015/presentations/ThainG_Docker.pdf

wking:
On Wed, Oct 07, 2015 at 12:27:24PM -0700, davidlt wrote:
> HTCondor bash system already have some support for Docker:
> https://research.cs.wisc.edu/htcondor/HTCondorWeek2015/presentations/ThainG_Docker.pdf

Page 29 of those slides shows the host's sysadmin starting a Docker
service.

More generally, I'm not sure how this is going to work for
unprivileged users.  namespaces(7) [1] has:

  Creation of new namespaces using clone(2) and unshare(2) in most
  cases requires the CAP_SYS_ADMIN capability.  User namespaces are
  the exception: since Linux 3.8, no privilege is required to create a
  user namespace.

So an unprivileged user should be able to create a user namespace,
and have some flexiblity inside it.  However, you don't have complete
[...]

PR #774, opened 2016-04-23
Rootless Containers

cyphar:
This enables the support for "rootless container mode". There are
certain restrictions on what non-root users can do, resulting in several
runC features not being available. ~~There are no checks in place at
the moment to make this clear to users.~~ I've implemented the config
validation.
- All cgroup operations require having write access to your current
  cgroup directory. By default, the directories are owned by `root` and have
  the mode `0755`. This means that we cannot set up _any_ cgroups, or join
  cgroups. Therefore new cgroup namespace doesn't fix this for us either, but
  hopefully we can get a patch upstream to fix this. **But** we should still
  improve cgroup handling so that we apply any cgroups we can if we have
  write access to the directory.
- setgroups(2) cannot be used in a non-privileged user namespace setup.
  We also have to set `/proc/self/setgroups` to "deny".
- We cannot map any user other than ourselves in a rootless container,
  which means that any user-related directives won't work. You can only be
  "root".

If you want to use this, you have to make sure you remove the `gid=5` entry from the `/dev/pts` mount, and only map **your own user** in the namespace.

Here's `runc start` working in both `root` and `rootless` setup:
it works

And here's `runc exec` working in both `root` and `rootless` setup:
it also works
### `TODO`
- [x] Provide a meaningful error message if the user specifies a configuration which maps more than one user (or doesn't map the effective user).
- [x] Provide a meaningful error message (don't just ignore it) if the user tries to use the `user` directive to run as a different user.
- [x] Provide a meaningful error message (don't just ignore it) if the user tries to specify any cgroup settings. Currently we can't provide cgroup settings (though it should be possible if the user just so happens to own the cgroup they currently reside in).
- [x] `runc exec` doesn't work, and we should be able to implement it. This actually complicates the code in `nsenter.c` which checks whether the container is unprivileged. `setgroup` needs special treatment.
- [x] `runc exec` doesnt' work with `root` running the exec and a rootless container (ironically). This is because we autodetect the `rootless` parameter on run, which isn't accurate. This can be fixed by storing the `rootless` flag in the `state.json`.
- [x] `rootless` should be passed to the init through netlink. Currently we are doing the `rootless` check in two places and it doesn't make sense to do the check in `nsenter.c` -- we might actually have to do it with capability checks in the future.
- [x] `runc events` doesn't work because the `rootless.Manager` doesn't appear to manage the paths properly, so it can't get any data from the cgroups.
- [x] Add `runc spec --rootless`.
- [x] `loadContainer` doesn't properly load the cgroup manager for the container, because the API forces that to happen (`libcontainer.New()` takes the cgroup manager as an argument). We can probably fix this by making it load the cgroup manager from the container state (but it might be ugly). Without this, we can't even hope to have `runc pause` and `runc resume` working.
- :x: Currently the cgroup setup is binary (either we use cgroups or we don't). But since cgroupv1 lets you have different permissions on different hierarchies, we should check for each subsystem that we have access to create a subtree in that cgroup. Then we can support several setups (as well as the kernel patch I've proposed). This would require many changes in config validation. In addition, we would have to store what cgroups we are using in a bitmask (like in the kernel).
  - :x: This would require adding a bunch of tests to `libcontainer/cgroup/rootless`. Luckily we already all of the mock stuff we need in `cgroupfs`.
- [x] Add unit tests to:
  - [x] `libcontainer/config/validate/rootless`
  - [x] `libcontainer/specconv/spec_linux.go` with `Rootless == true`.
  - :x: ~~`libcontainer/cgroup/rootless`~~ (not necessary at the moment)
- [x] ~~Detaching doesn't work due to a bug in runC with `--console` and user namespaces. #814 and #883~~
- [x] Add a setup for testing the rootless containers so we can make sure there's no regressions in this set up. We can try running the whole test suite (minus cgroups and a few other things). ~~All of the sniff tests should work.~~ The sniff tests are no more. ~~**THIS IS CURRENTLY BLOCKED ON FIXING THE `--console` BUG**.~~ The console bug has been fixed as part of #1018.
  - [x] We have to refactor all of the tests to use a wrapper around `runc` (so we can mess around with arguments in rootless mode).
  - [x] Also, we need to skip certain tests in rootless mode, since they are not supported.
- [x] We currently cannot use network namespaces in a rootless container setup (and still connect to the network). A fix for this is to use the host network, but that currently doesn't work in runC (#799). It should also be noted that things like `ping` don't work in rootless containers (user namespace issues). #807 fixes the validation issue.
  - [x] Figure out why we still can't talk to the internet.
- :x: Switch `container.Processes()` to join the container PID namespace, enumerate the list of PIDs and then send them over a UNIX socket (this causes the PIDs to be translated). The end result is to not require cgroups to enumerate PIDs (which actually isn't a good idea since processes may join sub-cgroups). This is necessary for `runc ps` and similar things to work properly.
  - :x: There are some unanswered questions about sending the PIDs though, because you need `CAP_SYS_ADMIN` in order to send a different PID. And there's also a valid question about atomicity (enumeration is not atomic, reading from `cgroup.procs` is).
### Open Questions
- [x] Make sure that the cgroup namespace actually allows us to set cgroups. **It doesn't**. I've sent [an email to upstream proposing a potential solution](http://article.gmane.org/gmane.linux.kernel.cgroups/15948). You can also [take a look at the state of my (probably broken) kernel patchset](https://github.com/cyphar/linux/tree/cgroupns-rootless-containers).
- [x] Should we move `runc` back to `/usr/bin`, since it's no longer an admin piece of software? This would also mean moving the `man` pages to `man1`.
### What works?
- As unprivileged user:
  - :x: ~~`runc checkpoint`~~ (while potentially possible, not implemented)
  - [x] `runc create` (#814)
  - [x] `runc create --console` (#814)
  - [x] `runc delete`
  - :x: ~~`runc events`~~ (not _really_ useful -- cgroups)
  - [x] `runc exec`
  - [x] `runc exec --console` (#814)
  - [x] `runc kill`
  - [x] `runc list`
    - [x] with containers not readable by us.
  - :x: ~~`runc pause`~~ (cgroups)
  - :x: ~~`runc ps`~~ (cgroups)
  - :x: ~~`runc restore`~~ (while potentially possible, not implemented)
  - :x: ~~`runc resume`~~ (cgroups)
  - [x] `runc run`
  - [x] `runc run -d --console` (#814)
  - [x] `runc spec`
  - [x] `runc start` (`create` doesn't work -- #814)
  - [x] `runc state`
  - :x: ~~`runc update`~~ (cgroups)
- As `root`:
  - :x: ~~`runc checkpoint`~~ (while potentially possible, not implemented)
  - [x] `runc delete`
  - :x: ~~`runc events`~~ (not _really_ useful -- cgroups)
  - [x] `runc exec`
  - [x] `runc exec --console` (#814)
  - [x] `runc kill`
  - [x] `runc list`
  - :x: ~~`runc pause`~~ (cgroups)
  - :x: `runc ps` (cgroups)
  - :x: ~~`runc restore`~~ (while potentially possible, not implemented)
  - :x: ~~`runc resume`~~ (cgroups)
  - [x] `runc run`
  - [x] `runc run -d --console` (#814)
  - [x] `runc spec`
  - [x] `runc start` (`create` doesn't work -- #814)
  - [x] `runc state`
  - :x: ~~`runc update`~~ (cgroups)
### Kernel Patches
- `CLONE_NEWCGROUP` fix to allow unprivileged processes to allow us to create subtrees.
  - [x] v1: https://lkml.org/lkml/2016/5/1/77
  - [x] v2: https://lkml.org/lkml/2016/5/1/87
  - [x] v3: https://lkml.org/lkml/2016/5/2/280
  - :x: v4: https://lkml.org/lkml/2016/5/13/576

Implements #38.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
 PREFIX := $(DESTDIR)/usr/local
-BINDIR := $(PREFIX)/sbin
+BINDIR := $(PREFIX)/bin
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
--- a/Makefile
+++ b/Makefile
@@ -79,10 +79,10 @@ runcimage:
 	docker build -t $(RUNC_IMAGE) .
 
 test:
-	make unittest integration
+	make unittest integration rootlessintegration
 
 localtest:
-	make localunittest localintegration
+	make localunittest localintegration localrootlessintegration
 
 unittest: runcimage
 	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localunittest
@@ -95,6 +95,13 @@ integration: runcimage
 
 localintegration: all
 	bats -t tests/integration${TESTFLAGS}
+
+rootlessintegration: runcimage
+	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) --cap-drop=ALL -u rootless $(RUNC_IMAGE) make localintegration
+
+# FIXME: This should not be separate from rootlessintegration's method of running.
+localrootlessintegration: all
+	sudo -u rootless -H PATH="${PATH}" bats -t tests/integration${TESTFLAGS}
 
 shell: all
 	docker run -e TESTFLAGS -ti --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) bash
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -109,7 +109,7 @@ type Mount struct {
 	Subsystems []string
 }
 
-func (m Mount) GetThisCgroupDir(cgroups map[string]string) (string, error) {
+func (m Mount) GetOwnCgroup(cgroups map[string]string) (string, error) {
 	if len(m.Subsystems) == 0 {
 		return "", fmt.Errorf("no subsystem for mount")
 	}
@@ -203,8 +203,8 @@ func GetAllSubsystems() ([]string, error) {
 	return subsystems, nil
 }
 
-// GetThisCgroupDir returns the relative path to the cgroup docker is running in.
-func GetThisCgroupDir(subsystem string) (string, error) {
+// GetOwnCgroup returns the relative path to the cgroup docker is running in.
+func GetOwnCgroup(subsystem string) (string, error) {
 	cgroups, err := ParseCgroupFile("/proc/self/cgroup")
 	if err != nil {
 		return "", err
@@ -213,14 +213,47 @@ func GetThisCgroupDir(subsystem string) (string, error) {
 	return getControllerPath(subsystem, cgroups)
 }
 
-func GetInitCgroupDir(subsystem string) (string, error) {
+func GetOwnCgroupPath(subsystem string) (string, error) {
+	cgroup, err := GetOwnCgroup(subsystem)
+	if err != nil {
+		return "", err
+	}
 
+	return getCgroupPathHelper(subsystem, cgroup)
+}
+
+func GetInitCgroup(subsystem string) (string, error) {
 	cgroups, err := ParseCgroupFile("/proc/1/cgroup")
 	if err != nil {
 		return "", err
 	}
 
 	return getControllerPath(subsystem, cgroups)
+}
+
+func GetInitCgroupPath(subsystem string) (string, error) {
+	cgroup, err := GetInitCgroup(subsystem)
+	if err != nil {
+		return "", err
+	}
+
+	return getCgroupPathHelper(subsystem, cgroup)
+}
+
+func getCgroupPathHelper(subsystem, cgroup string) (string, error) {
+	mnt, root, err := FindCgroupMountpointAndRoot(subsystem)
+	if err != nil {
+		return "", err
+	}
+
+	// This is needed for nested containers, because in /proc/self/cgroup we
+	// see pathes from host, which don't exist in container.
+	relCgroup, err := filepath.Rel(root, cgroup)
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(mnt, relCgroup), nil
 }
 
 func readProcsFile(dir string) ([]int, error) {
--- a/libcontainer/configs/validate/rootless.go
+++ b/libcontainer/configs/validate/rootless.go
@@ -1,0 +1,117 @@
+package validate
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+var (
+	geteuid = os.Geteuid
+	getegid = os.Getegid
+)
+
+func (v *ConfigValidator) rootless(config *configs.Config) error {
+	if err := rootlessMappings(config); err != nil {
+		return err
+	}
+	if err := rootlessMount(config); err != nil {
+		return err
+	}
+	// Currently, cgroups cannot effectively be used in rootless containers.
+	// The new cgroup namespace doesn't really help us either because it doesn't
+	// have nice interactions with the user namespace (we're working with upstream
+	// to fix this).
+	if err := rootlessCgroup(config); err != nil {
+		return err
+	}
+
+	// XXX: We currently can't verify the user config at all, because
+	//      configs.Config doesn't store the user-related configs. So this
+	//      has to be verified by setupUser() in init_linux.go.
+
+	return nil
+}
+
+func rootlessMappings(config *configs.Config) error {
+	rootuid, err := config.HostUID()
+	if err != nil {
+		return fmt.Errorf("failed to get root uid from uidMappings: %v", err)
+	}
+	if euid := geteuid(); euid != 0 {
+		if !config.Namespaces.Contains(configs.NEWUSER) {
+			return fmt.Errorf("rootless containers require user namespaces")
+		}
+		if rootuid != euid {
+			return fmt.Errorf("rootless containers cannot map container root to a different host user")
+		}
+	}
+
+	rootgid, err := config.HostGID()
+	if err != nil {
+		return fmt.Errorf("failed to get root gid from gidMappings: %v", err)
+	}
+
+	// Similar to the above test, we need to make sure that we aren't trying to
+	// map to a group ID that we don't have the right to be.
+	if rootgid != getegid() {
+		return fmt.Errorf("rootless containers cannot map container root to a different host group")
+	}
+
+	// We can only map one user and group inside a container (our own).
+	if len(config.UidMappings) != 1 || config.UidMappings[0].Size != 1 {
+		return fmt.Errorf("rootless containers cannot map more than one user")
+	}
+	if len(config.GidMappings) != 1 || config.GidMappings[0].Size != 1 {
+		return fmt.Errorf("rootless containers cannot map more than one group")
+	}
+
+	return nil
+}
+
+// cgroup verifies that the user isn't trying to set any cgroup limits or paths.
+func rootlessCgroup(config *configs.Config) error {
+	// Nothing set at all.
+	if config.Cgroups == nil || config.Cgroups.Resources == nil {
+		return nil
+	}
+
+	// Used for comparing to the zero value.
+	left := reflect.ValueOf(*config.Cgroups.Resources)
+	right := reflect.Zero(left.Type())
+
+	// This is all we need to do, since specconv won't add cgroup options in
+	// rootless mode.
+	if !reflect.DeepEqual(left.Interface(), right.Interface()) {
+		return fmt.Errorf("cannot specify resource limits in rootless container")
+	}
+
+	return nil
+}
+
+// mount verifies that the user isn't trying to set up any mounts they don't have
+// the rights to do. In addition, it makes sure that no mount has a `uid=` or
+// `gid=` option that doesn't resolve to root.
+func rootlessMount(config *configs.Config) error {
+	// XXX: We could whitelist allowed devices at this point, but I'm not
+	//      convinced that's a good idea. The kernel is the best arbiter of
+	//      access control.
+
+	for _, mount := range config.Mounts {
+		// Check that the options list doesn't contain any uid= or gid= entries
+		// that don't resolve to root.
+		for _, opt := range strings.Split(mount.Data, ",") {
+			if strings.HasPrefix(opt, "uid=") && opt != "uid=0" {
+				return fmt.Errorf("cannot specify uid= mount options in rootless containers where argument isn't 0")
+			}
+			if strings.HasPrefix(opt, "gid=") && opt != "gid=0" {
+				return fmt.Errorf("cannot specify gid= mount options in rootless containers where argument isn't 0")
+			}
+		}
+	}
+
+	return nil
+}
--- a/libcontainer/configs/validate/rootless.go
+++ b/libcontainer/configs/validate/rootless.go
@@ -37,7 +37,7 @@ func (v *ConfigValidator) rootless(config *configs.Config) error {
 }
 
 func rootlessMappings(config *configs.Config) error {
-	rootuid, err := config.HostUID()
+	rootuid, err := config.HostRootUID()
 	if err != nil {
 		return fmt.Errorf("failed to get root uid from uidMappings: %v", err)
 	}
@@ -50,7 +50,7 @@ func rootlessMappings(config *configs.Config) error {
 		}
 	}
 
-	rootgid, err := config.HostGID()
+	rootgid, err := config.HostRootGID()
 	if err != nil {
 		return fmt.Errorf("failed to get root gid from gidMappings: %v", err)
 	}
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -72,6 +72,7 @@ struct nlconfig_t {
 	char *namespaces;
 	size_t namespaces_len;
 	uint8_t is_setgroup;
+	uint8_t is_rootless;
 	char *oom_score_adj;
 	size_t oom_score_adj_len;
 };
@@ -87,6 +88,7 @@ struct nlconfig_t {
 #define GIDMAP_ATTR			27284
 #define SETGROUP_ATTR		27285
 #define OOM_SCORE_ADJ_ATTR	27286
+#define ROOTLESS_ATTR	    27287
 
 /*
  * Use the raw syscall for versions of glibc which don't include a function for
@@ -175,6 +177,7 @@ static void update_setgroups(int pid, enum policy_t setgroup)
 			policy = "deny";
 			break;
 		case SETGROUPS_DEFAULT:
+		default:
 			/* Nothing to do. */
 			return;
 	}
@@ -328,6 +331,9 @@ static void nl_parse(int fd, struct nlconfig_t *config)
 		switch (nlattr->nla_type) {
 		case CLONE_FLAGS_ATTR:
 			config->cloneflags = readint32(current);
+			break;
+		case ROOTLESS_ATTR:
+			config->is_rootless = readint8(current);
 			break;
 		case OOM_SCORE_ADJ_ATTR:
 			config->oom_score_adj = current;
@@ -574,9 +580,21 @@ void nsexec(void)
 
 					exit(ret);
 				case SYNC_USERMAP_PLS:
-					/* Enable setgroups(2) if we've been asked to. */
+					/*
+					 * Enable setgroups(2) if we've been asked to. But we also
+					 * have to explicitly disable setgroups(2) if we're
+					 * creating a rootless container (this is required since
+					 * Linux 3.19).
+					 */
+					if (config.is_rootless && config.is_setgroup) {
+						kill(child, SIGKILL);
+						bail("cannot allow setgroup in an unprivileged user namespace setup");
+					}
+
 					if (config.is_setgroup)
 						update_setgroups(child, SETGROUPS_ALLOW);
+					if (config.is_rootless)
+						update_setgroups(child, SETGROUPS_DENY);
 
 					/* Set up mappings. */
 					update_uidmap(child, config.uidmap, config.uidmap_len);
@@ -818,8 +836,10 @@ void nsexec(void)
 			if (setgid(0) < 0)
 				bail("setgid failed");
 
-			if (setgroups(0, NULL) < 0)
-				bail("setgroups failed");
+			if (!config.is_rootless && config.is_setgroup) {
+				if (setgroups(0, NULL) < 0)
+					bail("setgroups failed");
+			}
 
 			s = SYNC_CHILD_READY;
 			if (write(syncfd, &s, sizeof(s)) != sizeof(s))
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -72,18 +72,21 @@ struct nlconfig_t {
 	char *namespaces;
 	size_t namespaces_len;
 	uint8_t is_setgroup;
+	char *oom_score_adj;
+	size_t oom_score_adj_len;
 };
 
 /*
  * List of netlink message types sent to us as part of bootstrapping the init.
  * These constants are defined in libcontainer/message_linux.go.
  */
-#define INIT_MSG		62000
+#define INIT_MSG			62000
 #define CLONE_FLAGS_ATTR	27281
 #define NS_PATHS_ATTR		27282
-#define UIDMAP_ATTR		27283
-#define GIDMAP_ATTR		27284
+#define UIDMAP_ATTR			27283
+#define GIDMAP_ATTR			27284
 #define SETGROUP_ATTR		27285
+#define OOM_SCORE_ADJ_ATTR	27286
 
 /*
  * Use the raw syscall for versions of glibc which don't include a function for
@@ -186,7 +189,7 @@ static void update_setgroups(int pid, enum policy_t setgroup)
 	}
 }
 
-static void update_uidmap(int pid, char *map, int map_len)
+static void update_uidmap(int pid, char *map, size_t map_len)
 {
 	if (map == NULL || map_len <= 0)
 		return;
@@ -195,13 +198,22 @@ static void update_uidmap(int pid, char *map, int map_len)
 		bail("failed to update /proc/%d/uid_map", pid);
 }
 
-static void update_gidmap(int pid, char *map, int map_len)
+static void update_gidmap(int pid, char *map, size_t map_len)
 {
 	if (map == NULL || map_len <= 0)
 		return;
 
 	if (write_file(map, map_len, "/proc/%d/gid_map", pid) < 0)
 		bail("failed to update /proc/%d/gid_map", pid);
+}
+
+static void update_oom_score_adj(char *data, size_t len)
+{
+	if (data == NULL || len <= 0)
+		return;
+
+	if (write_file(data, len, "/proc/self/oom_score_adj") < 0)
+		bail("failed to update /proc/self/oom_score_adj");
 }
 
 /* A dummy function that just jumps to the given jumpval. */
@@ -317,6 +329,10 @@ static void nl_parse(int fd, struct nlconfig_t *config)
 		case CLONE_FLAGS_ATTR:
 			config->cloneflags = readint32(current);
 			break;
+		case OOM_SCORE_ADJ_ATTR:
+			config->oom_score_adj = current;
+			config->oom_score_adj_len = payload_len;
+			break;
 		case NS_PATHS_ATTR:
 			config->namespaces = current;
 			config->namespaces_len = payload_len;
@@ -425,13 +441,31 @@ void nsexec(void)
 	if (pipenum == -1)
 		return;
 
-	/* make the process non-dumpable */
-	if (prctl(PR_SET_DUMPABLE, 0, 0, 0, 0) != 0) {
-		bail("failed to set process as non-dumpable");
-	}
-
 	/* Parse all of the netlink configuration. */
 	nl_parse(pipenum, &config);
+
+	/* Set oom_score_adj. This has to be done before !dumpable because
+	 * /proc/self/oom_score_adj is not writeable unless you're an privileged
+	 * user (if !dumpable is set). All children inherit their parent's
+	 * oom_score_adj value on fork(2) so this will always be propagated
+	 * properly.
+	 */
+	update_oom_score_adj(config.oom_score_adj, config.oom_score_adj_len);
+
+	/*
+	 * Make the process non-dumpable, to avoid various race conditions that
+	 * could cause processes in namespaces we're joining to access host
+	 * resources (or potentially execute code).
+	 *
+	 * However, if the number of namespaces we are joining is 0, we are not
+	 * going to be switching to a different security context. Thus setting
+	 * ourselves to be non-dumpable only breaks things (like rootless
+	 * containers), which is the recommendation from the kernel folks.
+	 */
+	if (config.namespaces) {
+		if (prctl(PR_SET_DUMPABLE, 0, 0, 0, 0) < 0)
+			bail("failed to set process as non-dumpable");
+	}
 
 	/* Pipe so we can tell the child when we've finished setting up. */
 	if (socketpair(AF_LOCAL, SOCK_STREAM, 0, sync_child_pipe) < 0)
@@ -681,6 +715,11 @@ void nsexec(void)
 				 * clone_parent rant). So signal our parent to hook us up.
 				 */
 
+				/* Switching is only necessary if we joined namespaces. */
+				if (config.namespaces) {
+					if (prctl(PR_SET_DUMPABLE, 1, 0, 0, 0) < 0)
+						bail("failed to set process as dumpable");
+				}
 				s = SYNC_USERMAP_PLS;
 				if (write(syncfd, &s, sizeof(s)) != sizeof(s))
 					bail("failed to sync with parent: write(SYNC_USERMAP_PLS)");
@@ -691,6 +730,11 @@ void nsexec(void)
 					bail("failed to sync with parent: read(SYNC_USERMAP_ACK)");
 				if (s != SYNC_USERMAP_ACK)
 					bail("failed to sync with parent: SYNC_USERMAP_ACK: got %u", s);
+				/* Switching is only necessary if we joined namespaces. */
+				if (config.namespaces) {
+					if (prctl(PR_SET_DUMPABLE, 0, 0, 0, 0) < 0)
+						bail("failed to set process as dumpable");
+				}
 			}
 
 			/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
 RUN apt-get update && apt-get install -y \
     build-essential \
     curl \
+    sudo \
     gawk \
     iptables \
     jq \
@@ -21,6 +22,12 @@ RUN apt-get update && apt-get install -y \
     python-minimal \
     --no-install-recommends \
     && apt-get clean
+
+# Add a dummy user for the rootless integration tests. While runC does
+# not require an entry in /etc/passwd to operate, one of the tests uses
+# `git clone` -- and `git clone` does not allow you to clone a
+# repository if the current uid does not have an entry in /etc/passwd.
+RUN useradd -u1000 -m -d/home/rootless -s/bin/bash rootless
 
 # install bats
 RUN cd /tmp \
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -39,6 +39,11 @@ checkpointed.`,
 		if err := checkArgs(context, 1, exactArgs); err != nil {
 			return err
 		}
+		// XXX: Currently this is untested with rootless containers.
+		if isRootless() {
+			return fmt.Errorf("runc checkpoint requires root")
+		}
+
 		container, err := getContainer(context)
 		if err != nil {
 			return err
--- a/exec.go
+++ b/exec.go
@@ -90,9 +90,6 @@ following will output a list of processes running in the container:
 		if err := checkArgs(context, 1, minArgs); err != nil {
 			return err
 		}
-		if os.Geteuid() != 0 {
-			return fmt.Errorf("runc should be run as root")
-		}
 		if err := revisePidFile(context); err != nil {
 			return err
 		}
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -267,25 +267,8 @@ func getCgroupData(c *configs.Cgroup, pid int) (*cgroupData, error) {
 	}, nil
 }
 
-func (raw *cgroupData) parentPath(subsystem, mountpoint, root string) (string, error) {
-	// Use GetThisCgroupDir instead of GetInitCgroupDir, because the creating
-	// process could in container and shared pid namespace with host, and
-	// /proc/1/cgroup could point to whole other world of cgroups.
-	initPath, err := cgroups.GetThisCgroupDir(subsystem)
-	if err != nil {
-		return "", err
-	}
-	// This is needed for nested containers, because in /proc/self/cgroup we
-	// see pathes from host, which don't exist in container.
-	relDir, err := filepath.Rel(root, initPath)
-	if err != nil {
-		return "", err
-	}
-	return filepath.Join(mountpoint, relDir), nil
-}
-
 func (raw *cgroupData) path(subsystem string) (string, error) {
-	mnt, root, err := cgroups.FindCgroupMountpointAndRoot(subsystem)
+	mnt, err := cgroups.FindCgroupMountpoint(subsystem)
 	// If we didn't mount the subsystem, there is no point we make the path.
 	if err != nil {
 		return "", err
@@ -297,7 +280,10 @@ func (raw *cgroupData) path(subsystem string) (string, error) {
 		return filepath.Join(raw.root, filepath.Base(mnt), raw.innerPath), nil
 	}
 
-	parentPath, err := raw.parentPath(subsystem, mnt, root)
+	// Use GetOwnCgroupPath instead of GetInitCgroupPath, because the creating
+	// process could in container and shared pid namespace with host, and
+	// /proc/1/cgroup could point to whole other world of cgroups.
+	parentPath, err := cgroups.GetOwnCgroupPath(subsystem)
 	if err != nil {
 		return "", err
 	}
--- a/libcontainer/cgroups/rootless/rootless.go
+++ b/libcontainer/cgroups/rootless/rootless.go
@@ -1,0 +1,128 @@
+// +build linux
+
+package rootless
+
+import (
+	"fmt"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fs"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/configs/validate"
+)
+
+// TODO: This is copied from libcontainer/cgroups/fs, which duplicates this code
+//       needlessly. We should probably export this list.
+
+var subsystems = []subsystem{
+	&fs.CpusetGroup{},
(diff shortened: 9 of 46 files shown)


mrunalp:
For cgroups, we can skip doing any setup if cgroupsPath == "" and Resources == nil in the config.
We can either skip or introduce a NoOp cgroups manager.

cyphar:
@mrunalp I'm going to go with a rootless cgroup manager so we might expand it later (there are some upcoming kernel features that might make cgroups in rootless containers usable).

mrunalp:
Sounds good.

libcontainer/configs/validate/rootless.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// cgroup verifies that the user isn't trying to set any cgroup limits or paths.
+func (v *RootlessValidator) cgroup(config *configs.Config) error {
+	// Nothing set at all.
+	if config.Cgroups == nil || config.Cgroups.Resources == nil {
+		return nil
+	}
+
+	// Used for comparison.
```
crosbymichael: This is some pretty complex validation.  So some cgroups are ok but others are not?

mrunalp: We could just fail the check if any values are set at all. No need to go check for defaults. User can create the config without setting resources or path which is simple to check.

cyphar: The /actual/ check is that all cgroups have no non-default settings. Unfortunately, `specconv` adds device cgroup settings that get merged with the config the user specified -- so a simple "is this equal to the zero value" check doesn't cut it. I haven't figured out a nice way of dealing with that (`specconv` runs long before we get to this part, and we need it to run before we can do anything with the `config` (like figuring out if we're rootless)).

crosbymichael: Ya, that sounds better.  We should be looking for this in runc not in libcontainer.

cyphar: I'm not really convinced that we shouldn't be doing any checking in `libcontainer`. The same question can be asked about `libcontainer/configs/validate` -- why do we do any config verification inside `libcontainer`? There's also a question of whether or not `libcontainer` should autodetect rootless mode or whether it should be passed as an option (you can't use rootless containers with root as far as I can tell -- and it's definitely a **bad** idea).

crosbymichael: runc should populate the correct config that libcontainer gets and it should not be modified inside libcontainer.  All the changes that need to be made should happen while we generate the config not after it is made.

cyphar: The current state of this patchset doesn't modify the config inside `libcontainer`. The issue is that `specconv` adds device options to `[]Device` even if the user doesn't specify anything. So we can either do the verification of the cgroups in runC (which means that if someone uses libcontainer directly they probably won't immediately realise they can't set cgroup settings) or we do the verification in libcontainer and make `specconv` not generate any cgroup options in rootless mode.

jfrazelle:
i think the rootless cgroups manager is good, then it can be expanded when the cgroups ns is added :) just my opinion but ianam, thanks for this

jfrazelle:
also wrt the features that might not be possible or are hard for the time being, they could be disabled, and then slowly turned back on as implementations evolve, kinda like how we did userns in docker, and then slowly added more features back in wrt sharing namespaces
it's easier to make a smaller change then iterate on it, then one huge one

cyphar:
@jfrazelle AFAICS all of the core features work. But some of them either just require `root` (`criu` IIRC) or currently can't be done under user namespaces (cgroups -- which I'm working on a patch for). But all of the others should still work, and in principle I want to try to get all of the features working for `root` operating on a `rootless` container.

jfrazelle:
Nice :)

On Tuesday, April 26, 2016, Aleksa Sarai <notifications@github.com> wrote:

> @jfrazelle <https://github.com/jfrazelle> AFAICS all of the core features
> work. But some of them either just require root (criu IIRC) or currently
> can't be done under user namespaces (cgroups -- which I'm working on a
> patch for). But all of the others should still work, and in principle I
> want to try to get all of the features working for root operating on a
> rootless container.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/opencontainers/runc/pull/774#issuecomment-214914690>
>


-- 


Jessie Frazelle
4096R / D4C4 DD60 0D66 F65A 8EFC  511E 18F3 685C 0022 BFF3
pgp.mit.edu <http://pgp.mit.edu/pks/lookup?op=get&search=0x18F3685C0022BFF3>

cyphar:
Note: as far as I can see, the only thing left to do before we can clear most of the checkpoints is for me to fix up the rootless cgroup manager so that it stores the process's actual cgroup path. Then `runc events` should work, as well as all of the freezer code (for root). Checkpoint and restore might also work just by doing that.

cyphar:
Whoops wrong button. ;)

cyphar:
Whoops wrong button. ;)

cyphar:
@avagin Do you know if it's possible to run `criu` as an unprivileged user (specifically from the kernel side when interacting with unprivileged user namespaces)?

davidlt:
Looks like this is moving forward, very interesting. Thus I started testing it. Built on Fedora 24 (updated on May 2nd). Looks like mount-bind works fine without root permissions.  I wasn't lucky with internet connectivity without _sudo_.  Later on I need to check if there is a way not to be a root user within container (our software will not using root account). Nice progress!
```
[davidlt@pccms205 magic_dir]$ pwd
/home/davidlt/magic_dir
[davidlt@pccms205 magic_dir]$ ls
[davidlt@pccms205 magic_dir]$ touch HOST_FILE
[davidlt@pccms205 magic_dir]$ ls
HOST_FILE
[davidlt@pccms205 magic_dir]$ cat /etc/os-release | grep PRETTY_NAME >> HOST_FILE
[davidlt@pccms205 magic_dir]$ cat HOST_FILE
PRETTY_NAME="Fedora 24 (Workstation Edition)"
[davidlt@pccms205 test]$ runc --root $PWD start test_container
sh-4.2# cat /etc/os-release | grep PRETTY_NAME
PRETTY_NAME="CentOS Linux 7 (Core)"
sh-4.2# cd /some
sh-4.2# cat /etc/os-release | grep PRETTY_NAME >> HOST_FILE
sh-4.2# touch NEW_FILE
sh-4.2# exit
[davidlt@pccms205 test]$ cat ~/magic_dir/HOST_FILE 
PRETTY_NAME="Fedora 24 (Workstation Edition)"
PRETTY_NAME="CentOS Linux 7 (Core)"
[davidlt@pccms205 test]$ file ~/magic_dir/NEW_FILE 
/home/davidlt/magic_dir/NEW_FILE: empty
```
For anyone else who wants to test this I am also sharing diff between original and modified `config.json`:
```
--- config.json	2016-05-02 13:25:24.468181348 +0200
+++ config.json.correct	2016-05-02 13:25:14.661496040 +0200
@@ -6,7 +6,7 @@
 	},
 	"process": {
 		"terminal": true,
-		"user": {},
+		"user": { "uid": 0, "gid": 0, "additionalGids": null },
 		"args": [
 			"sh"
 		],
@@ -35,6 +35,12 @@
 	},
 	"hostname": "runc",
 	"mounts": [
+                {
+                        "destination": "/some",
+                        "type": "bind",
+                        "source": "/home/davidlt/magic_dir",
+                        "options": [ "rbind" ]
+                },
 		{
 			"destination": "/proc",
 			"type": "proc",
@@ -60,8 +66,7 @@
 				"noexec",
 				"newinstance",
 				"ptmxmode=0666",
-				"mode=0620",
-				"gid=5"
+				"mode=0620"
 			]
 		},
 		{
@@ -112,14 +117,20 @@
 	],
 	"hooks": {},
 	"linux": {
-		"resources": {
-			"devices": [
-				{
-					"allow": false,
-					"access": "rwm"
-				}
-			]
-		},
+                "uidMappings": [
+                        {
+                                "hostID": 1000,
+                                "containerID": 0,
+                                "size": 1
+                        }
+                ],
+                "gidMappings": [
+                        {
+                                "hostID": 1000,
+                                "containerID": 0,
+                                "size": 1
+                        }
+                ],
 		"namespaces": [
 			{
 				"type": "pid"
@@ -135,7 +146,10 @@
 			},
 			{
 				"type": "mount"
-			}
+			},
+                        {
+                                "type": "user"
+                        }
 		],
 		"maskedPaths": [
 			"/proc/kcore",
@@ -152,4 +166,4 @@
 			"/proc/sysrq-trigger"
 		]
 	}
-}
\ No newline at end of file
+}

```

cyphar:
@davidlt 

> Later on I need to check if there is a way not to be a root user within container (our software will not using root account). Nice progress!

Unfortunately this is not possible, due to restrictions within the kernel. Essentially this is a logcal result of these two restrictions on user namespaces:

1. All user namespaces must provide a mapping for root inside a container.
2. Unprivileged user namespaces can only provide a mapping for one user, the user which created the namespace.

As a result, the only user that is mapped inside the container is your user (as root). You can discuss with the kernel upstream about restriction number 1, because it's the only restriction which it might be possible to fix. But at the moment, there isn't a way to do what you want.

davidlt:
Yeah, I tried a few thing. I even added a user within a container using Docker, which I can see while running with runC, but cannot launch anything under that user.

What about internet connectivity?

cyphar:
Unfortunately, creating bridges between a container's network namespace and the hosts's network namespace requires creating a virtual interface in the host's network namespace. AFAIK that requires root (but I may be wrong). One potential solution would be not to create a network namespace (this currently doesn't work due to some bindmount issues, but that's nothing we can't fix).

davidlt:
Okay. Looks like internet connectivity will arrive at some point. This is needed in my case because majority of data is not local (i.e. not available on some shared file system mounted via bind/slave mount to the container). It most cases it has to be streamed (network IO). I think, at this point just having internet connectivity is good step forward.

Do you know what was the reasoning for namespaces to provide root inside the container? Quick googling didn't reveal too much documentation around this.

cyphar:
To be honest, I just quickly read through the kernel code and I'm not sure this is a restriction imposed by the kernel. It's possible it's just how we've set up rootless containers to work. Currently my runC build is failing with the error:

```
process_linux.go:247: getting pipe fds for pid 16189 caused "readlink /proc/16189/fd/0: permission denied"
```

Which tells me there's some permission issues with the `/proc` setup we have (where we read the pipe file descriptors over stdin -- but for some reason it looks like the process can't open its own stdin?).

mrunalp:
@davidlt Best bet for networking would be to use the host network stack (i.e. don't add it to the config).
The way typically networking is setup requires moving network devices from host network namespace to the container's network namespace. With privileged user namespaces, the runc hooks can do that work but that isn't possible with unprivileged containers. It would need discussion with upstream as @cyphar suggested.

avagin:
> All cgroup operations require having CAP_SYS_ADMIN in the root user
> namespace. This means that we cannot set up any cgroups, or join
> cgroups. The new cgroup namespace doesn't fix this for us either, but
> hopefully we can get a patch upstream to fix this.

I don't understand this passage. I tried and cgroups works for unprivileged users by the same way as other file systems. I've read the kernel code and haven't found places in cgroup code which are protected by CAP_SYS_ADMIN.
```
[avagin@laptop ~]$ whoami 
avagin
[avagin@laptop ~]$ sudo mkdir /sys/fs/cgroup/cpu/test
[avagin@laptop ~]$ sudo chown avagin:avagin /sys/fs/cgroup/cpu/test
[avagin@laptop ~]$ echo $$ > /sys/fs/cgroup/cpu/test/tasks 
bash: /sys/fs/cgroup/cpu/test/tasks: Permission denied
[avagin@laptop ~]$ mkdir /sys/fs/cgroup/cpu/test/sub
[avagin@laptop ~]$ echo $$ > /sys/fs/cgroup/cpu/test/sub/tasks 
[avagin@laptop ~]$ cat /sys/fs/cgroup/cpu/test/sub/cpu.shares 
1024
[avagin@laptop ~]$ echo 2014 > /sys/fs/cgroup/cpu/test/sub/cpu.shares
[avagin@laptop ~]$ echo 512 > /sys/fs/cgroup/cpu/test/sub/cpu.shares
[avagin@laptop ~]$ ls -l /sys/fs/cgroup/cpu/test/
total 0
-rw-r--r--. 1 root   root   0 May  3 14:26 cgroup.clone_children
-rw-r--r--. 1 root   root   0 May  3 14:26 cgroup.procs
-r--r--r--. 1 root   root   0 May  3 14:26 cpuacct.stat
-rw-r--r--. 1 root   root   0 May  3 14:26 cpuacct.usage
-r--r--r--. 1 root   root   0 May  3 14:26 cpuacct.usage_percpu
-rw-r--r--. 1 root   root   0 May  3 14:26 cpu.cfs_period_us
-rw-r--r--. 1 root   root   0 May  3 14:26 cpu.cfs_quota_us
-rw-r--r--. 1 root   root   0 May  3 14:26 cpu.shares
-r--r--r--. 1 root   root   0 May  3 14:26 cpu.stat
-rw-r--r--. 1 root   root   0 May  3 14:26 notify_on_release
drwxrwxr-x. 2 avagin avagin 0 May  3 14:25 sub
-rw-r--r--. 1 root   root   0 May  3 14:25 tasks
[avagin@laptop ~]$ ls -l /sys/fs/cgroup/cpu/test/sub/
total 0
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 cgroup.clone_children
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 cgroup.procs
-r--r--r--. 1 avagin avagin 0 May  3 14:25 cpuacct.stat
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpuacct.usage
-r--r--r--. 1 avagin avagin 0 May  3 14:25 cpuacct.usage_percpu
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpu.cfs_period_us
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpu.cfs_quota_us
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpu.shares
-r--r--r--. 1 avagin avagin 0 May  3 14:25 cpu.stat
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 notify_on_release
-rw-r--r--. 1 avagin avagin 0 May  3 14:25 tasks
```

jfrazelle:
Just device cgroups will fail

On Tue, May 3, 2016 at 2:29 PM, Andrew Vagin <notifications@github.com>
wrote:

> All cgroup operations require having CAP_SYS_ADMIN in the root user
> namespace. This means that we cannot set up any cgroups, or join
> cgroups. The new cgroup namespace doesn't fix this for us either, but
> hopefully we can get a patch upstream to fix this.
>
> I don't understand this passage. I tried and cgroups works for
> unprivileged users by the same way as other file systems. I've read the
> kernel code and haven't found places in cgroup code which are protected by
> CAP_SYS_ADMIN.
>
> [avagin@laptop ~]$ whoami
> avagin
> [avagin@laptop ~]$ sudo mkdir /sys/fs/cgroup/cpu/test
> [avagin@laptop ~]$ sudo chown avagin:avagin /sys/fs/cgroup/cpu/test
> [avagin@laptop ~]$ echo $$ > /sys/fs/cgroup/cpu/test/tasks
> bash: /sys/fs/cgroup/cpu/test/tasks: Permission denied
> [avagin@laptop ~]$ mkdir /sys/fs/cgroup/cpu/test/sub
> [avagin@laptop ~]$ echo $$ > /sys/fs/cgroup/cpu/test/sub/tasks
> [avagin@laptop ~]$ cat /sys/fs/cgroup/cpu/test/sub/cpu.shares
> 1024
> [avagin@laptop ~]$ echo 2014 > /sys/fs/cgroup/cpu/test/sub/cpu.shares
> [avagin@laptop ~]$ echo 512 > /sys/fs/cgroup/cpu/test/sub/cpu.shares
> [avagin@laptop ~]$ ls -l /sys/fs/cgroup/cpu/test/
> total 0
> -rw-r--r--. 1 root   root   0 May  3 14:26 cgroup.clone_children
> -rw-r--r--. 1 root   root   0 May  3 14:26 cgroup.procs
> -r--r--r--. 1 root   root   0 May  3 14:26 cpuacct.stat
> -rw-r--r--. 1 root   root   0 May  3 14:26 cpuacct.usage
> -r--r--r--. 1 root   root   0 May  3 14:26 cpuacct.usage_percpu
> -rw-r--r--. 1 root   root   0 May  3 14:26 cpu.cfs_period_us
> -rw-r--r--. 1 root   root   0 May  3 14:26 cpu.cfs_quota_us
> -rw-r--r--. 1 root   root   0 May  3 14:26 cpu.shares
> -r--r--r--. 1 root   root   0 May  3 14:26 cpu.stat
> -rw-r--r--. 1 root   root   0 May  3 14:26 notify_on_release
> drwxrwxr-x. 2 avagin avagin 0 May  3 14:25 sub
> -rw-r--r--. 1 root   root   0 May  3 14:25 tasks
> [avagin@laptop ~]$ ls -l /sys/fs/cgroup/cpu/test/sub/
> total 0
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 cgroup.clone_children
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 cgroup.procs
> -r--r--r--. 1 avagin avagin 0 May  3 14:25 cpuacct.stat
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpuacct.usage
> -r--r--r--. 1 avagin avagin 0 May  3 14:25 cpuacct.usage_percpu
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpu.cfs_period_us
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpu.cfs_quota_us
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 cpu.shares
> -r--r--r--. 1 avagin avagin 0 May  3 14:25 cpu.stat
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 notify_on_release
> -rw-r--r--. 1 avagin avagin 0 May  3 14:25 tasks
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/opencontainers/runc/pull/774#issuecomment-216670442>
>



-- 


Jessie Frazelle
4096R / D4C4 DD60 0D66 F65A 8EFC  511E 18F3 685C 0022 BFF3
pgp.mit.edu <http://pgp.mit.edu/pks/lookup?op=get&search=0x18F3685C0022BFF3>

cyphar:
@avagin Sorry, I need to update the first paragraph. But if you look at your session log:

```
$ sudo chown avagin:avagin /sys/fs/cgroup/cpu/test
```

If you have to use root to enable using cgroups, it's not useful for some usecases of rootless containers. I've been working upstream to allow an unprivileged cgroup namespace to create their own subtrees, which is what is necessary to make rootless containers mostly feature-complete.

But yeah, `CAP_SYS_ADMIN` is definitely the wrong thing and I'll fix up what I wrote.

Do you know anything about whether you can use `criu` as an unprivileged user? My guess is that it probably isn't, since it messes around with saving and restoring kernel state.

avagin:
@cyphar 
> If you have to use root to enable using cgroups, it's not useful for some usecases of rootless containers.

Are you sure that this should be fixed in a kernel space? Maybe we need to fix this in systemd? How does LXC handles this problem.

> Do you know anything about whether you can use criu as an unprivileged user? My guess is that you probably can't, since it messes around with saving and restoring kernel state.

We announced "Unprivileged dump" in CRIU 2.0 and now we are working on "Unprivileged restore". I don't know how good it will work for root-less containers, but I think it isn't unsolvable task.

davidlt:
I have been testing DMTCP (transparently checkpoints a single-host or distributed computation in user-space) and that worked in user-land for checkpointing and restoring complex applications. Thus CRIU in user-land should technologically possible.

avagin:
On Tue, May 3, 2016 at 2:31 PM, Jess Frazelle <notifications@github.com>
wrote:

> Just device cgroups will fail
>
> Yes, you are right. It isn't only one problem with devices. mknod is
protected by CAP_MKNOD too.

cyphar:
@avagin 

> > If you have to use root to enable using cgroups, it's not useful for some usecases of rootless containers.
>
> Are you sure that this should be fixed in a kernel space? Maybe we need to fix this in systemd? How does LXC handles this problem.

Reliance on systemd is a bad idea, since not all systems use systemd, systemd doesn't support all of the cgroups we want to use, systemd is a daemon running as root, and there's definitely a more general solution. LXC requires a daemon (`cgmanager`) that provides an authenticated way of requesting cgroup setup, which requires installing software that runs as root.

The reason I think this should be solved in the kernel is actually not because of containers, it's because cgroups are meant to be a general resource limiting system -- why can't an unprivileged process set up resource limiting for its own subprocesses? And it looks like upstream agrees with me on this point (the details are still a bit murky, but cgroup namespaces will probably be part of the solution).

avagin:
> The reason I think this should be solved in the kernel is actually not because
> of containers, it's because cgroups are meant to be a general resource limiting
> system -- why can't an unprivileged process set up resource limiting for its
> own subprocesses? And it looks like upstream agrees with me on this point (the
> details are still a bit murky, but cgroup namespaces will probably be part of
> the solution).
> 

This looks reasonable. Thank you for the explanation.

cyphar:
I've opened #836 and #837 to take some of the more general cleanup patches an apply them to runC as a whole.

dqminh:
WDYT of adding another cgroup driver that interfacing with https://linuxcontainers.org/cgmanager/introduction ? Since cgmanager's quite established, it potentially provides us the shortest path to rootless container while we work on the kernel bits ?

cyphar:
@dqminh Currently we're working on a lazy cgroup management implementation (where you only join the cgroups that you can and have set limits for). Once we have that, we could add hooks for things like CGManager (though I don't like the model for it, because it requires getting an admin to install a daemon that runs as root).

On the kernel side, it looks like the maintainer (Tejun) doesn't really agree about the logic behind my changes. Mainly he doesn't like the fact that it requires changing the VFS semantics of cgroups. If anyone has any ideas on how to change cgroups so you get subtree delegation without breaking the VFS semantics I'd be very happy to hear it. :D

cyphar:
I've rebased this patchset to be based on #977, #975 and #1018. The integration tests now pass (as well as the rootless version of the integration tests). But we'll have to wait until all of the other code is merged before this. :cry:

richfelker:
I have a rough outline of a solution to the network namespace problem, albeit a bit heavy: instead of bridging the container to the host with actual network interfaces, use a tunnel interface inside the container, and a userspace ip stack with high-level nat-like functionality that translates ip traffic on the container side to socket calls on the host side. It would not allow arbitrary network traffic (although ping could even work with modern Linux's icmp dgram sockets) but certainly provides strong isolation of the container from the host's network.

cyphar:
@richfelker Yeah, I've discussed that idea (amongst others) with some people and in general the real solution is kernel support (which just doesn't exist at the moment -- and if you look at FreeBSD jails it's just not there). However, all of this can be handled transparently outside of runC (by creating that forwarding server and then just creating a child that forks itself into the network namespace of the containers).

So runC won't implement network hacks like that, but it would be a cool project to work on. Maybe there's a way to modify CNI to get it to work in a rootless mode.

cyphar:
#1018 has (finally) been merged. That means that this PR is now testable after being rebased on master -- it's no longer blocked on any other PRs.

Currently I still need to work on the following things:

* [ ] cgroup handling. @brauner helped me with this, but I'm going to have to rebase his patches and then see if I need to modify them to get everything running nicely. I would prefer to not need to have a separate `rootless` cgroup manager -- but then again explicit is better than implicit.

* [ ] Figure out if I'm going to solve the `runc ps` problem. On the one hand, it would be _nice_ to solve it, on the other I don't think reimplementing `runc ps` to not use cgroups is something that I would ever voluntarily do. If someone wants to buy me beer at a conference, I will consider it.

That's it. In addition, I've started working on some new `cgroupns` patches that *might* be accepted (it revolves around using extended attributes to say that a particular subtree is owned by the cgroup namespace).

cyphar:
Oh, and with the existence of [`skopeo`](https://github.com/projectatomic/skopeo) and [`umoci`](https://github.com/cyphar/umoci)` you can now (as an unprivileged user) pull an image from the internet and extract it (and also create new images):

```
% skopeo copy docker://opensuse/amd64:42.2 oci:opensuse:latest 
% umoci unpack --rootless --image opensuse:latest bundle
% runc run -b bundle ctr
sh-4.4# <inside a rootless container>
% umoci repack --image opensuse:new bundle
% skopeo copy oci:opensuse:new docker://opensuse/amd64:42.3
```

cyphar:
**Dammit**. It looks like the fix for CVE-2016-9962 (5d93fed3d27f1e2bab58bad13b180a7a81d0b378) actually breaks rootless containers -- you cannot access `/proc/<pid>/setgroups` or any other files. We will have to figure out a way to handle it but for now I've just **disabled the security patch so DO NOT MERGE IT.**

cyphar:
Looks like a kernel issue? The following simple program will `abort` if you enable the first line but will work if you don't. And as far as I can tell, dumpability isn't checked in the `open("/proc/self/setgroups")` path...

```
#define _GNU_SOURCE
#include <sched.h>
#include <stdio.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdlib.h>
#include <sys/prctl.h>

int main(void)
{
	if (prctl(PR_SET_DUMPABLE, 0, 0, 0, 0) != 0)
		abort();

	if (unshare(CLONE_NEWUSER) < 0)
		abort();

	int fd = open("/proc/self/setgroups", O_WRONLY);
	if (fd < 0)
		abort();

	return 0;
}
```

cyphar:
Alright, it's because of how owners of `/proc/[pid]/...` files work. Here's a kernel patch I've sent upstream which fixes this issue for me.

```diff
From 908ffbb30e39ddfa40bde1145331a91c5084e1f5 Mon Sep 17 00:00:00 2001
From: Aleksa Sarai <asarai@suse.de>
Date: Wed, 18 Jan 2017 13:43:40 +1100
Subject: [PATCH] procfs: change the owner of non-dumpable and writeable files

In order to protect against ptrace(2) and similar attacks on container
runtimes when they join namespaces, many runtimes set mm->dumpable to
SUID_DUMP_DISABLE. However, doing this means that attempting to set up
an unprivileged user namespace will fail because an unprivileged process
can no longer access /proc/self/{setgroups,{uid,gid}_map} for the
container process (which is the same uid as the runtime process).

Fix this by changing pid_getattr to *also* change the owner of regular
files that have a mode of 0644 (when the process is not dumpable). This
ensures that the important /proc/[pid]/... files mentioned above are
properly accessible by a container runtime in a rootless container
context.

The most blantant issue is that a non-dumpable process in a rootless
container context is unable to open /proc/self/setgroups, because it
doesn't own the file.

int main(void)
{
	prctl(PR_SET_DUMPABLE, 0, 0, 0, 0);
	unshare(CLONE_NEWUSER);

	/* This will fail. */
	int fd = open("/proc/self/setgroups", O_WRONLY);
	if (fd < 0)
		abort();

	return 0;
}

Cc: dev@opencontainers.org
Cc: containers@lists.linux-foundation.org
Signed-off-by: Aleksa Sarai <asarai@suse.de>
---
 fs/proc/base.c | 6 ++++--
 1 file changed, 4 insertions(+), 2 deletions(-)

diff --git a/fs/proc/base.c b/fs/proc/base.c
index ca651ac00660..ebabb12f4536 100644
--- a/fs/proc/base.c
+++ b/fs/proc/base.c
@@ -1729,6 +1729,7 @@ int pid_getattr(struct vfsmount *mnt, struct dentry *dentry, struct kstat *stat)
 			return -ENOENT;
 		}
 		if ((inode->i_mode == (S_IFDIR|S_IRUGO|S_IXUGO)) ||
+		    (inode->i_mode == (S_IFREG|S_IRUGO|S_IWUSR)) ||
 		    task_dumpable(task)) {
 			cred = __task_cred(task);
 			stat->uid = cred->euid;
@@ -1770,6 +1771,7 @@ int pid_revalidate(struct dentry *dentry, unsigned int flags)
 
 	if (task) {
 		if ((inode->i_mode == (S_IFDIR|S_IRUGO|S_IXUGO)) ||
+		    (inode->i_mode == (S_IFREG|S_IRUGO|S_IWUSR)) ||
 		    task_dumpable(task)) {
 			rcu_read_lock();
 			cred = __task_cred(task);
@@ -2394,7 +2396,7 @@ static int proc_pident_instantiate(struct inode *dir,
 	return -ENOENT;
 }
 
-static struct dentry *proc_pident_lookup(struct inode *dir, 
+static struct dentry *proc_pident_lookup(struct inode *dir,
 					 struct dentry *dentry,
 					 const struct pid_entry *ents,
 					 unsigned int nents)
@@ -2536,7 +2538,7 @@ static const struct pid_entry attr_dir_stuff[] = {
 
 static int proc_attr_dir_readdir(struct file *file, struct dir_context *ctx)
 {
-	return proc_pident_readdir(file, ctx, 
+	return proc_pident_readdir(file, ctx,
 				   attr_dir_stuff, ARRAY_SIZE(attr_dir_stuff));
 }
 
-- 
2.11.0
```

mrunalp:
@cyphar Nice!

teddyking:
Is rootless runc currently expected to work for a create/exec workflow?
We're seeing the following error when running a create:

```
$ whoami
max
$ runc create test
operation not permitted
```

We ran an strace on the process which revealed:

```
[pid 12290] <... futex resumed> )       = 0
[pid 12288] futex(0xb196b0, FUTEX_WAIT, 0, NULL <unfinished ...>
[pid 12293] fchown(0, 4294967294, 4294967294 <unfinished ...>
[pid 12290] select(0, NULL, NULL, NULL, {0, 3} <unfinished ...>
[pid 12293] <... fchown resumed> )      = -1 EPERM (Operation not permitted)
```

We tracked it down to [here](https://github.com/cyphar/runc/blob/d9684792f5427ab993ac9e68890222b64c3ee84e/tty.go#L74-L76), which is Fchowning std{in,out,err} to a non-root user.

cyphar:
@teddyking Yes, this works for both `create+start` as well as `run`. It also works with `exec` and practically all other commands. Have you created a spec file with `runc spec --rootless`? If your spec file doesn't include all of the necessary namespace configuration, then you'll get errors like that.

cyphar:
Note however that this is currently broken (working on fixing it now -- it's more broken stuff because of the `PRCTL_SET_DUMPABLE` change):

```
% runc create ctr
% runc exec ctr sh
nsenter: failed to open /proc/27820/ns/user: Permission denied
exec failed: container_linux.go:262: starting container process caused "process_linux.go:81: executing setns process caused \"exit status 14\""
```

teddyking:
@cyphar in the example I pasted above we were running as a user that had a uid/gid of 4294967294, but I get the same error when running as a "normal" user as well:

```
$ id
uid=1001(teddyking) gid=1001(teddyking) groups=1001(teddyking)
$ runc spec --rootless
$ sed -i 's/"terminal": true/"terminal": false/' config.json
$ runc create test
operation not permitted
```

The config.json was created with `runc spec --rootless` and was only modified to set terminal to false (as otherwise we get the `cannot allocate tty...` error). I've uploaded it [here](https://gist.github.com/teddyking/5780767fbe81bf199dc42e447d692c81) for reference.

teddyking:
Ah sorry my mistake, I should've actually read the error message properly!
If I run the create passing `--console-socket` then everything works as expected:

```
$ runc spec --rootless
$ runc create --console-socket /tmp/socket test
$ runc list
ID          PID         STATUS      BUNDLE                     CREATED                         OWNER
test        19917       created     /tmp/test-teddy-rootless   2017-02-24T08:04:48.26473722Z   teddyking
```

teddyking:
But actually thinking about it, is there any way for a rootless create to work without needing to pass a socket? For context, Garden doesn't ever call a start (we exec processes into the container) so it doesn't really make sense for us to need to pass a socket here.

cyphar:
@teddyking It should not be necessary to use `--console-socket` (I use it on my local machine without it all the time). `--console-socket` is required for our tests, but is not required to just use it. What happens if you run with a more sane UID/GID (like `1000`) -- I'm worried this might be the underlying issue.

`fchown(<thing I already own>, <my uid>, <my gid>)` should never fail, so I'm quite confused why you're hitting that error.

teddyking:
@cyphar Apologies, we had some confusion over our ssh session. We were going in as root and then `su` to unprivileged so our shell's stdio fds were pointing to a `/dev/pts` that was still owned by root. When we called `runc create` the fchown then failed at https://github.com/cyphar/runc/blob/d9684792f5427ab993ac9e68890222b64c3ee84e/tty.go#L74-L76

We just had to ensure that our shell's stdio was actually owned by us and then `runc create` worked successfully, who'd have thought!

cyphar:
Cool @teddyking. This is actually one of the many reasons I'm against the current way we handle `"terminal": false` containers, and is one of the main reasons why `--console-socket` came about -- but at least it works for you now. :wink:

cyphar:
@teddyking Can you try #1349 and see if it helps out? Terminal handling for `terminal=false` is something we really need to sort out...

teddyking:
@cyphar #1349 helps a lot! With that change we are able to `runc create test` a rootless container with `terminal=false` and without passing `--console-socket`, which is exactly what we wanted. Thanks!

buchanae:
Is it feasible that the linux kernel could support a many (container) to one (host) user/group ID mapping, so that rootless containers could allow setuid/setgid inside the container while mapping to a single user on the host?

I can ask on a linux kernel forum if you prefer, but thought maybe I'd get a quick sanity check here first.

In the academic/HPC setting, rootless containers would be a huge win, but it's still going to be difficult to convince people to write containers that have only a single user, IMO.

cyphar:
@buchanae The short answer is "unlikely if not impossible", for two reasons:

1. It would incur a performance penalty in every single `kuid <-> uid` translation, because it would no longer be a simple loop and addition.

2. The `kuid` translation would become lossy. When I `chown` a file to a different UID how will the kernel be able to remember what the owner of the file should be inside my user namespace? If the owner is different then user namespace would no longer implement POSIX.

> but it's still going to be difficult to convince people to write containers that have only a single user, IMO.

crosbymichael:
@cyphar i tested this when working on the fchown issue.  Its looking really good.  Should we look into merging most of the feature so that you don't have big rebases or is it still too risky at the moment as it touches other code?

crosbymichael:
One bit of feedback is when you create the spec for `runc specs --rootless` set the hostID to the current users id so that they don't have to go in and change that.  They can just generate the spec and `runc run ...` the container

cyphar:
@crosbymichael 

> Should we look into merging most of the feature so that you don't have big rebases or is it still too risky at the moment as it touches other code?

As I mentioned there's only one truly blocking feature, which is that `!dumpable` currently breaks that `create+exec` (without `start`) workflow -- which is something I'm still working on cleaning up. Unfortunately because of the `!dumpable` changes certain features like `oom_score_adj` don't really work anymore. So maybe the best solution is to just not set `!dumpable` if we're rootless (though this sounds like a bad idea, especially since I know that @brauner (LXC) wants to be able to run runC underneath an unprivileged LXC container).

> One bit of feedback is when you create the spec for runc specs --rootless set the hostID to the current users id so that they don't have to go in and change that.

I believe it already does that, see the below output:

```
% ./runc spec --rootless
% grep hostID config.json
                                "hostID": 1000,
                                "hostID": 100,
% id -a
uid=1000(cyphar) gid=100(users)
```

cyphar:
`!dumpable` is nearly fixed, the only thing left to fix is to make `oom_score_adj` be set inside `nsexec.c`.

cyphar:
@crosbymichael Alright. Everything works now, and tests pass. Did you want me to squash the rootless cgroup manager changes into the "add support for rootless containers" commit?

crosbymichael:
@cyphar if you want since the commit has `[wip]` in the title

crosbymichael:
LGTM

cyphar:
@crosbymichael I rebased and removed `[wip]` from the cgroup manager patch, which I'm going to keep separate as in the future I intend to extend it and it'll be nice to know what exact parts had to be changed to make the rootless cgroup manager work.

jessfraz:
its-happening

crosbymichael:
LGTM

avagin:
LGTM

mrunalp:
I will also review this. Please wait before merge :)

discordianfish:
Awesome :)

libcontainer/nsenter/nsexec.c
```diff
@@ -681,6 +715,11 @@ void nsexec(void)
 				 * clone_parent rant). So signal our parent to hook us up.
 				 */
 
+				/* Switching is only necessary if we joined namespaces. */
+				if (config.namespaces) {
+					if (prctl(PR_SET_DUMPABLE, 1, 0, 0, 0) < 0)
+						bail("failed to set process as dumpable");
+				}
```
hqhq: Why do we need this?

cyphar: `/proc/self/uid_map` and `/proc/self/gid_map` become root-owned if you're not dumpable. So the process doing mapping doesn't have privileges to do the mapping.

hqhq: I mean, shouldn't this unset and reset dumpable happen in parent process?

hqhq: OK, I see, misunderstood this.

libcontainer/cgroups/utils.go
```diff
@@ -223,6 +231,29 @@ func GetInitCgroupDir(subsystem string) (string, error) {
...
+func GetInitCgroupPath(subsystem string) (string, error) {
+	cgroup, err := GetInitCgroup(subsystem)
+	if err != nil {
+		return "", err
+	}
+
+	return getCgroupPathHelper(subsystem, cgroup)
+}
+
+func getCgroupPathHelper(subsystem, cgroup string) (string, error) {
```
hqhq: This function is kind of subtle, can you keep the comments?

cyphar: The only comment that makes sense now is on the `filepath.Rel` is that the one you want me to keep?

hqhq: In this function yes, and maybe leave the other comment to `raw.path()` to specify why we use `GetOwnCgroupPath` instead of `GetInitCgroupPath`, I used to get a lot of people ask me about these cgroup path functions...

cyphar: Fair enough. Not sure why I removed them in the first place.

Makefile
```diff
@@ -91,11 +91,18 @@ localunittest: all
 	go test -timeout 3m -tags "$(BUILDTAGS)" ${TESTFLAGS} -v $(allpackages)
 
 integration: runcimage
-	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localintegration
+	docker run -e TESTFLAGS -ti --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localintegration
```
hqhq: You don't really need `-i` here, do you? We removed this to fix CI in non-interactive mode, see https://github.com/opencontainers/runc/pull/1252 .

libcontainer/cgroups/utils.go
```diff
@@ -223,6 +231,29 @@ func GetInitCgroupDir(subsystem string) (string, error) {
 	return getControllerPath(subsystem, cgroups)
 }
 
+func GetInitCgroupPath(subsystem string) (string, error) {
```
hqhq: Do we need this since it's not used? And I doubt it'll be any usage.

Makefile
cyphar: No I don't think we need it.

libcontainer/cgroups/utils.go
cyphar: We don't need it now, but the systemd cgroup manager does use `GetInitCgroup`. I can drop it if you prefer, this is more for the benefit of users of `libcontainer`.

hqhq: That's probably erroneous, maybe just nobody using systemd cgroup inside a container, I'm OK we keep it, so we'll say no to subsequent PR which'll try to remove this unused functions :)

hqhq:
LGTM

cyphar:
Hang on, lemme squash it first. :wink:

cyphar:
@hqhq Squashed and rebased.

hqhq:
LGTM

mrunalp:
Should we drop groups that are unmapped?

```
[mrunal@acme busybox]$ ./runc --root ~/runc/state run 1234
/ # id
uid=0(root) gid=0(root) groups=65534,0(root)
```

cyphar:
@mrunalp We don't have privileges to do that. In fact, it's a security feature of the kernel to not allow unprivileged users to drop supplementary groups because of paths with modes such as `0707`. Such ACLs make it easy to blacklist a group from accessing something.

crosbymichael:
ping @mrunalp

mrunalp:
LGTM

cyphar:
:tada:

davidlt:
Looks like it's party time! 11 months in development. Someone should post this on Hacker News.

marcosnils:
image
 
:D

muayyad-alsadi:
Any link to updated docs. Blog post?

cyphar:
@muayyad-alsadi No doc updates, I'll follow up with those. Here's [a blog post from last year](https://www.cyphar.com/blog/post/rootless-containers-with-runc) and [my talk at Linux.conf.au from earlier this year](https://www.youtube.com/watch?v=r6EcUyamu94).

PR #780, opened 2016-04-25
Improve stats output format for stability 

crosbymichael:
This adds specific types and improves the json format for the marshaled
structure so that it is inline with the output that the spec produce,
camelCase not snake_case.

This should be the last change needed for people to really depend on the
output of this command and ensure that it does not change with any
internal changes instead of just marshaling the libcontainer structure.


hqhq:
LGTM

cyphar:
LGTM.

PR #784, opened 2016-04-26
Add integration test for ps command

hqhq:

### DIFF
--- a/tests/integration/ps.bats
+++ b/tests/integration/ps.bats
@@ -11,7 +11,7 @@ function teardown() {
   teardown_busybox
 }
 
-@test "ps -eaf" {
+@test "ps" {
   # start busybox detached
   runc run -d --console /dev/pts/ptmx test_busybox
   [ "$status" -eq 0 ]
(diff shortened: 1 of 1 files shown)


cyphar:
I think we should seriously consider condensing the man page for `runc` into a single man page. The subcommands aren't complicated enough to warrant forcing the user to switch between man pages.

hqhq:
I don't think condensing all man pages into a single one is a good idea. We shouldn't do this only because they are not complicated enough, according to this logic, we should split them one day when they are complicated. I don't see why we bother doing this in the first place.

Regarding what's better for users. I don't think there are many people who want to know every detail about all runc commands in one time. If they want an overview, they can use `runc --help`, then use man page for more detailed information about one specific command.

For now, man pages don't give more information than `runc command --help`, we can elaborate them for good, and that is what I think we should take effort on. WDYT?

cyphar:
I don't think that making the documentation more verbose should be a solution to the "our subcommand man pages are too small" problem. Personally, I only like the `git` style of man pages (one for each subcommand) if the commands are as complicated as `git rebase`. We should not aim for `git rebase` levels of complexity.

In my mind, the `osc` man pages are much closer to how our man pages should be. They're about the same level of complexity as our subcommands -- you don't need any more than two paragraphs to describe every command.

hqhq:
@cyphar Man page for ps command is already existed, I updated this PR to add integration test only.

crosbymichael:
LGTM

hqhq:
ping @cyphar PTAL, #1065 would depend on this to add more tests.

hqhq:
Rebased.

cyphar:
LGTM

tests/integration/ps.bats
```diff
@@ -21,8 +21,39 @@ function teardown() {
 
   testcontainer test_busybox running
 
-  runc ps test_busybox -eaf
+  runc ps test_busybox
```
cyphar: Can we keep `-eaf` here? I'm planning on updating the busybox image in the console PR, and the default `ps` format changed between 2014 and 2016.

cyphar: Actually never mind, I get why you're testing this.

crosbymichael:
LGTM

PR #554, opened 2016-02-11
Add the most basic sniff tests of runc

duglin:
just so that we're not merging code into master w/o any tests at all.
I expect this to be removed once we have a real testing infrastructure.

Signed-off-by: Doug Davis <dug@us.ibm.com>

PR #785, opened 2016-04-26
Remove sniffTest

hqhq:
We have integration test now, not ideal though, but it
surely can replace sniffTest.


cyphar:
> I don't think snifftest is needed anymore, it's all covered by integration test, so I tried to remove it in #785 WDYT?

I don't really have a strong opinion, but it does appear redundant to have a sniff test that is then replicated in an integration test.

/cc @duglin -- do you have any reason why we should keep it in? In #554 you mention that the sniff tests were just meant to be a stopgap until we had a proper testing framework -- I think bats qualifies.

duglin:
I agree and if there is some tests in sniff that's missing from integration then we should add it.
so +1 from me

cyphar:
In that case, LGTM.

mrunalp:
LGTM

PR #790, opened 2016-04-28
Runc update cgroup kmem limit

mlaventure:
This PR does 2 things:

- Re-allow to update `memory.kmem.limit_in_bytes` once a container is running (provided that the value was set before the any task joined the cgroup)
- Add a new `update` command, allowing an user to update some of the cgroup constraints for a given container

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -5,6 +5,7 @@ package fs
 import (
 	"bufio"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -12,6 +13,7 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/system"
 )
 
 type MemoryGroup struct {
@@ -33,7 +35,7 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 			}
 		}
 		// We have to set kernel memory here, as we can't change it once
-		// processes have been attached.
+		// processes have been attached to the cgroup.
 		if err := s.SetKernelMemory(path, d.config); err != nil {
 			return err
 		}
@@ -55,9 +57,44 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 }
 
 func (s *MemoryGroup) SetKernelMemory(path string, cgroup *configs.Cgroup) error {
-	// This has to be done separately because it has special constraints (it
-	// can't be done after there are processes attached to the cgroup).
-	if cgroup.Resources.KernelMemory > 0 {
+	// This has to be done separately because it has special
+	// constraints (it can only be initialized before setting up a
+	// hierarchy or adding a task to the cgroups. However, if
+	// sucessfully initialized, it can be updated anytime afterwards)
+	if cgroup.Resources.KernelMemory != 0 {
+		kmemInitialized := false
+		// Is kmem.limit_in_bytes already set?
+		kmemValue, err := getCgroupParamUint(path, "memory.kmem.limit_in_bytes")
+		if err != nil {
+			return err
+		}
+		switch system.GetLongBit() {
+		case 32:
+			kmemInitialized = uint32(kmemValue) != uint32(math.MaxUint32)
+		case 64:
+			kmemInitialized = kmemValue != uint64(math.MaxUint64)
+		}
+
+		if !kmemInitialized {
+			// If hierarchy is set, we can't change the limit
+			usesHierarchy, err := getCgroupParamUint(path, "memory.use_hierarchy")
+			if err != nil {
+				return err
+			}
+			if usesHierarchy != 0 {
+				return fmt.Errorf("cannot initialize kmem.limit_in_bytes if use_hierarchy is already set")
+			}
+
+			// If there's already tasks in the cgroup, we can't change the limit either
+			tasks, err := getCgroupParamString(path, "tasks")
+			if err != nil {
+				return err
+			}
+			if tasks != "" {
+				return fmt.Errorf("cannot initialize kmem.limit_in_bytes after task have joined this cgroup")
+			}
+		}
+
 		if err := writeFile(path, "memory.kmem.limit_in_bytes", strconv.FormatInt(cgroup.Resources.KernelMemory, 10)); err != nil {
 			return err
 		}
@@ -110,6 +147,10 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 
 func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 	if err := setMemoryAndSwap(path, cgroup); err != nil {
+		return err
+	}
+
+	if err := s.SetKernelMemory(path, cgroup); err != nil {
 		return err
 	}
 
--- a/libcontainer/system/sysconfig.go
+++ b/libcontainer/system/sysconfig.go
@@ -10,3 +10,7 @@ import "C"
 func GetClockTicks() int {
 	return int(C.sysconf(C._SC_CLK_TCK))
 }
+
+func GetLongBit() int {
+	return int(C.sysconf(C._SC_LONG_BIT))
+}
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -1,0 +1,77 @@
+#!/usr/bin/env bats
+
+load helpers
+
+UPDATE_TEST_RUNC_ROOT="$BATS_TMPDIR/runc-cgroups-integration-test"
+
+CGROUP_MEMORY=""
+
+TEST_CGROUP_NAME="runc-cgroups-integration-test"
+
+function init_cgroup_path() {
+   base_path=$(grep "rw,"  /proc/self/mountinfo | grep -i -m 1 'MEMORY$' | cut -d ' ' -f 5)
+   CGROUP_MEMORY="${base_path}/${TEST_CGROUP_NAME}"
+}
+
+function teardown() {
+    rm -f $BATS_TMPDIR/runc-update-integration-test.json
+    teardown_running_container_inroot test_cgroups_kmem $UPDATE_TEST_RUNC_ROOT
+    teardown_busybox
+}
+
+function setup() {
+    teardown
+    setup_busybox
+
+    init_cgroup_path
+}
+
+function check_cgroup_value() {
+    cgroup=$1
+    source=$2
+    expected=$3
+
+    current=$(cat $cgroup/$source)
+    echo  $cgroup/$source
+    echo "current" $current "!?" "$expected"
+    [ "$current" -eq "$expected" ]
+}
+
+@test "cgroups-kernel-memory-initialized" {
+    # Add cgroup path
+    sed -i 's/\("linux": {\)/\1\n    "cgroupsPath": "runc-cgroups-integration-test",/'  ${BUSYBOX_BUNDLE}/config.json
+
+    # Set some initial known values
+    DATA=$(cat <<-EOF
+    "memory": {
+        "kernel": 16777216
+    },
+EOF
+    )
+    DATA=$(echo ${DATA} | sed 's/\n/\\n/g')
+    sed -i "s/\(\"resources\": {\)/\1\n${DATA}/" ${BUSYBOX_BUNDLE}/config.json
+
+    # start a detached busybox to work with
+    "$RUNC" --root $UPDATE_TEST_RUNC_ROOT start -d --console /dev/pts/ptmx test_cgroups_kmem
+    [ "$status" -eq 0 ]
+    wait_for_container_inroot 15 1 test_cgroups_kmem $UPDATE_TEST_RUNC_ROOT
+
+    # update kernel memory limit
+    "$RUNC" --root $UPDATE_TEST_RUNC_ROOT update test_cgroups_kmem --kernel-memory 50331648
+    [ "$status" -eq 0 ]
+    check_cgroup_value $CGROUP_MEMORY "memory.kmem.limit_in_bytes" 50331648
+}
+
+@test "cgroups-kernel-memory-uninitialized" {
+    # Add cgroup path
+    sed -i 's/\("linux": {\)/\1\n    "cgroupsPath": "runc-cgroups-integration-test",/'  ${BUSYBOX_BUNDLE}/config.json
+
+    # start a detached busybox to work with
+    run "$RUNC" --root $UPDATE_TEST_RUNC_ROOT start -d --console /dev/pts/ptmx test_cgroups_kmem
+    [ "$status" -eq 0 ]
+    wait_for_container_inroot 15 1 test_cgroups_kmem $UPDATE_TEST_RUNC_ROOT
+
+    # update kernel memory limit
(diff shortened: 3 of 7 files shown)


update.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/codegangsta/cli"
+)
+
+type updateRessources struct {
```
crosbymichael: Can we use the `specs.Resources` struct here?

mlaventure: That struct contains a lot more fields (e.g. Devices array, OOM, Pids, Network, etc..)

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+	// This has to be done separately because it has special
+	// constraints (it can only be initialized before setting up a
+	// hierarchy or adding a task to the cgroups. However, if
+	// sucessfully initialized, it can be updated anytime afterwards)
+	if cgroup.Resources.KernelMemory != 0 {
+		kmemInitialized := false
+		// Is kmem.limit_in_bytes already set?
+		if b, err := ioutil.ReadFile(filepath.Join(path, "memory.kmem.limit_in_bytes")); err != nil {
+			return err
+		} else {
```
cyphar: This doesn't need to be an `else`.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+			case 64:
+				kmemInitialized = kmemValue != uint64(math.MaxUint64)
+			}
+		}
+
+		if !kmemInitialized {
+			// If hierarchy is set, we can't change the limit
+			if b, err := ioutil.ReadFile(filepath.Join(path, "memory.use_hierarchy")); err != nil {
+				return err
+			} else {
```
cyphar: As above this doesn't need to be an else.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+				val := strings.TrimSpace(string(b))
+				if val != "0" {
+					return fmt.Errorf("Cannot initialize kmem.limit_in_bytes if use_hierarchy is already set")
+				}
+			}
+
+			// If there's already tasks in the cgroup, we can't change the limit either
+			if b, err := ioutil.ReadFile(filepath.Join(path, "tasks")); err != nil {
+				return err
+			} else {
```
cyphar: ... nor does this.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+			}
+			switch system.GetLongBit() {
+			case 32:
+				kmemInitialized = uint32(kmemValue) != uint32(math.MaxUint32)
+			case 64:
+				kmemInitialized = kmemValue != uint64(math.MaxUint64)
+			}
+		}
+
+		if !kmemInitialized {
```
cyphar: My main concern with this is when will a runC container have this setup? Currently, whenever we start a container, it will have this all set up anyway.

libcontainer/cgroups/fs/memory.go
mlaventure: Sorry, old habits die hard, different coding convention :)

libcontainer/cgroups/fs/memory.go
mlaventure: On my system the memory cgroups have `use_hierachy = 0` by default (ubuntu 14.04). Also a `runc spec` doesn't specify a value for kernel memory which mean it'll be uninitialized by default.

cyphar: > On my system the memory cgroups have use_hierachy = 0 by default (ubuntu 14.04). Also a runc spec doesn't specify a value for kernel memory which mean it'll be uninitialized by default.

Fair enough, I didn't know that (it doesn't happen on any of my systems).

> And like in the bats test I wrote, when specifying a new cgroupath for the container, they may not have any tasks inside if it's brand new.

Right, but the issue isn't when creating a container (the old code would work in that case). My question is when would we be updating a container's limits without it having processes inside it (isn't that the point of the PR)? The reason the code is written the way it is now is so that Docker could update containers (before the `containerd` split) -- and that worked perfectly fine.

update.go
cyphar: I think it's a better idea to use the OCI spec fields, since runC is meant to be the example OCI runtime implementation. We could check that the user hasn't set any other fields in the config (and error out in that case), but then we wouldn't add a slightly incompatible (and possibly out of date in the future) JSON payload for updating resources.

Also, you typo'd the word `Resources` in `updateRessources`.

libcontainer/cgroups/fs/memory.go
mlaventure: It's perfectly  possible to update the kernel memory limit when there is processes inside. What's prohibited is to set that limit the **first** time if there is already processes assigned to thst cgroup.

libcontainer/cgroups/fs/memory.go
mlaventure: Actually, the else is needed to make use of the `b` variable. I could move the b in the outer scope, but I kind of prefer this style. is that ok with you?

mlaventure:
Updated to use `specs.Resources` struct instead.

libcontainer/cgroups/fs/memory.go
LK4D4: It's sort of "official" code style :) https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

mlaventure: I guess I should read that page, thanks @LK4D4 !

mlaventure:
Updated with proper go coding style.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,51 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+		case 64:
+			kmemInitialized = kmemValue != uint64(math.MaxUint64)
+		}
+
+		if !kmemInitialized {
+			// If hierarchy is set, we can't change the limit
+			b, err = ioutil.ReadFile(filepath.Join(path, "memory.use_hierarchy"))
+			if err != nil {
+				return err
+			} else {
```
LK4D4: haha, preferred style is to remove else here :)

mlaventure: oups

mlaventure:
ping @crosbymichael @cyphar

libcontainer/cgroups/fs/memory.go
crosbymichael: ping

LK4D4:
LGTM

crosbymichael:
LGTM

update.go
```diff
@@ -0,0 +1,186 @@
...
+func u64Ptr(i uint64) *uint64 { return &i }
+func u16Ptr(i uint16) *uint16 { return &i }
+
+var updateCommand = cli.Command{
+	Name:      "update",
+	Usage:     "update container resource constraints",
+	ArgsUsage: `<container-id> <ps options>`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "ressources, r",
```
crosbymichael: `s/ressources/resources`

update.go
```diff
@@ -0,0 +1,186 @@
...
+				Mems:   sPtr(""),
+			},
+			BlockIO: &specs.BlockIO{
+				Weight: u16Ptr(0),
+			},
+		}
+
+		config := container.Config()
+
+		if in := context.String("ressources"); in != "" {
```
crosbymichael: `resources`

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+
+		if !kmemInitialized {
+			// If hierarchy is set, we can't change the limit
+			b, err = ioutil.ReadFile(filepath.Join(path, "memory.use_hierarchy"))
+			if err != nil {
+				return err
+			}
+			val := strings.TrimSpace(string(b))
+			if val != "0" {
+				return fmt.Errorf("Cannot initialize kmem.limit_in_bytes if use_hierarchy is already set")
```
cyphar: These errors should match the Go style (all lower case).

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+			}
+
+			// If there's already tasks in the cgroup, we can't change the limit either
+			b, err = ioutil.ReadFile(filepath.Join(path, "tasks"))
+			if err != nil {
+				return err
+			}
+			val = strings.TrimSpace(string(b))
+			if val != "" {
+				return fmt.Errorf("Cannot initialize kmem.limit_in_bytes after task have joined this cgroup")
```
cyphar: Same here.

update.go
```diff
@@ -0,0 +1,186 @@
...
+				if val := context.String(opt); val != "" {
+					var err error
+					*dest, err = strconv.ParseUint(val, 10, 64)
+					if err != nil {
+						fatal(fmt.Errorf("Invalid value for %s: %s", opt, err))
+					}
+				}
+			}
+		}
+
```
cyphar: We should probably check here that the user didn't specify any extra config values. You could do it by comparing the value of `r` to the zero value of the config (after setting all the values we care about to `0`).

mlaventure: I think I'm missing something. Are you referring on the case where we get the new value via json?

In that case I only assign the supported values to the config instance ultimately passed down to runc. 

Isn't that enough?

libcontainer/cgroups/fs/memory.go
hqhq: @mlaventure The reason I added this restriction is because I found it's restriction from kernel https://github.com/torvalds/linux/blob/master/Documentation/cgroup-v1/memory.txt#L284-L286

But after more tests and code review, turns out it's a false alarm, the restriction is no longer in kernel for a long time, I already sent a patch to fix this doc in kernel. We can remove this restriction on Docker side once this PR merged, thanks.

hqhq:
@mlaventure Since I found the restriction for kernel memory before is totally wrong, we can refactor codes in `libcontainer/cgroups/fs/memory.go`, function `SetKernelMemory` is not needed, we can merge kmem setting into the normal path. Do you mind doing this?

And I suggest you add kmem setting in this PR, and separate the `update` command part to another, because we have certain spec considerations for `update` containers.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,9 +58,50 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+		kmemInitialized := false
+		// Is kmem.limit_in_bytes already set?
+		b, err := ioutil.ReadFile(filepath.Join(path, "memory.kmem.limit_in_bytes"))
+		if err != nil {
+			return err
+		}
+		kmemValue, err := strconv.ParseUint(strings.TrimSpace(string(b)), 10, 64)
+		if err != nil {
+			return err
+		}
```
hqhq: You can just use `getCgroupParamUint`.

update.go
cyphar: My concern is that a user specifies a config update that serialises to the `spec` JSON but we can't update. We should error out in that case rather than silently ignore the invalid option.

mlaventure: If one of the value is invalid, the 'Set(' call will error out.

If the invalid value is from one of the unsupported options (e.g. DisableOOMKiller), it will just be ignored.

Just to clarify, you want us to error out instead of ignoring the value in case 2?

libcontainer/cgroups/fs/memory.go
mlaventure: Oh, thanks!

crosbymichael:
Why do you have weird commits in your pr? like this one.
https://github.com/opencontainers/runc/pull/790/commits/806470e2aa29a939421534fde0867cf1b106e86a

mlaventure:
@crosbymichael bad rebase, it'll be fixed on my next push

crosbymichael:
LGTM

hqhq:
LGTM
@cyphar I think your concern can be addressed in following up PRs, so I'm merging this.

cyphar:
@hqhq Yup, it's just a UX issues. (late) LGTM.

libcontainer/system/sysconfig.go
```diff
@@ -10,3 +10,7 @@ import "C"
 func GetClockTicks() int {
 	return int(C.sysconf(C._SC_CLK_TCK))
 }
+
+func GetLongBit() int {
+	return int(C.sysconf(C._SC_LONG_BIT))
```
tianon: FYI, `_SC_LONG_BIT` doesn't exist in musl, so Alpine's going to need a workaround/patch here

cc @ncopa

tianon: It might be sufficient to add something like this: (untested)
```go
/*
#ifndef _SC_LONG_BIT
#define _SC_LONG_BIT (8*sizeof(long))
#endif
*/
import "C"
```

cyphar: This might be nicer:

```
/*
#include <limits.h>
#ifndef _SC_LONG_BIT
#define _SC_LONG_BIT (CHAR_BIT * sizeof(long))
#endif
*/
import "C"
```

ncopa: You should not set `_SC_LONG_BIT` to `(CHAR_BIT * sizeof(long))`.

The `GetLongBit` function should return `(CHAR_BIT * sizeof(long))` if `_SC_LONG_BIT` is undefined or if the `syconf(_SC_LONG_BIT)` returns error.

similar issue we previously had: http://permalink.gmane.org/gmane.linux.distributions.alpine.devel/2269

cyphar: Sorry, I meant to write that it should be conditional based on if `_SC_LONG_BIT` is set (not actually setting that value).

tests/integration/cgroups.bats
```diff
@@ -0,0 +1,77 @@
...
+    sed -i 's/\("linux": {\)/\1\n    "cgroupsPath": "runc-cgroups-integration-test",/'  ${BUSYBOX_BUNDLE}/config.json
+
+    # start a detached busybox to work with
+    run "$RUNC" --root $UPDATE_TEST_RUNC_ROOT start -d --console /dev/pts/ptmx test_cgroups_kmem
+    [ "$status" -eq 0 ]
+    wait_for_container_inroot 15 1 test_cgroups_kmem $UPDATE_TEST_RUNC_ROOT
+
+    # update kernel memory limit
+    run "$RUNC" --root $UPDATE_TEST_RUNC_ROOT update test_cgroups_kmem --kernel-memory 50331648
+    [ ! "$status" -eq 0 ]
```
cyphar: This doesn't appear to fail on Linux 4.6, anymore. Causing this test to fail on newer kernels.

hqhq: That's right, I opened a PR to fix this: https://github.com/opencontainers/runc/pull/828

PR #793, opened 2016-04-29
Use '=' instead of ':' separator on labels

bboreham:
Because Docker has deprecated ':' in https://github.com/docker/docker/pull/21232

Triggered by https://github.com/docker/docker/issues/22413

### DIFF
--- a/libcontainer/selinux/selinux.go
+++ b/libcontainer/selinux/selinux.go
@@ -486,14 +486,14 @@ func DupSecOpt(src string) []string {
 		con["level"] == "" {
 		return nil
 	}
-	return []string{"label:user:" + con["user"],
-		"label:role:" + con["role"],
-		"label:type:" + con["type"],
-		"label:level:" + con["level"]}
+	return []string{"label=user:" + con["user"],
+		"label=role:" + con["role"],
+		"label=type:" + con["type"],
+		"label=level:" + con["level"]}
 }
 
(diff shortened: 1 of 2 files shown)


mrunalp:
@rhatdan PTAL

rhatdan:
LGTM

mrunalp:
LGTM

crosbymichael:
LGTM

runcom:
@thaJeztah this must be reverted as well, as in, `DupSecOpt` should not add `label=` at all based on what `InitLabels` does.

libcontainer/selinux/selinux.go
```diff
@@ -486,14 +486,14 @@ func DupSecOpt(src string) []string {
...
+		"label=role:" + con["role"],
+		"label=type:" + con["type"],
+		"label=level:" + con["level"]}
 }
 
 // DisableSecOpt returns a security opt that can be used to disabling SELinux
 // labeling support for future container processes
 func DisableSecOpt() []string {
-	return []string{"label:disable"}
+	return []string{"label=disable"}
```
runcom: this `label=` is docker specific, while here in libcontainer there shouldn't be any mention to docker. `DisableSecOpt` and `DupSecOpt` should just deal with `disable,role,type,level`. Both CRI-O and docker should just pass `disable,role,type.level` stuff and not `label=...`.

PR #796, opened 2016-04-30
Need to make sure labels applied to /dev

rhatdan:

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -69,6 +69,9 @@ func setupRootfs(config *configs.Config, console *linuxConsole, pipe io.ReadWrit
 		if err := setupDevSymlinks(config.Rootfs); err != nil {
 			return newSystemErrorWithCause(err, "setting up /dev symlinks")
 		}
+		if err := label.Relabel(filepath.Join(config.Rootfs, "dev"), config.MountLabel, false); err != nil {
+			return err
+		}
 	}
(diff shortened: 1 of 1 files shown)


libcontainer/rootfs_linux.go
```diff
@@ -69,6 +69,9 @@ func setupRootfs(config *configs.Config, console *linuxConsole, pipe io.ReadWrit
 		if err := setupDevSymlinks(config.Rootfs); err != nil {
 			return newSystemErrorWithCause(err, "setting up /dev symlinks")
 		}
+		if err := label.Relabel(filepath.Join(config.Rootfs+"/dev"), config.MountLabel, false); err != nil {
```
cyphar: Did you mean to do `filepath.Join(config.Rootfs, "dev")`?

rhatdan: Does it matter?

rhatdan: Oops doing a little too much python

cyphar:
LGTM.

mrunalp:
LGTM

rhatdan:
@mrunalp @crosbymichael This patch should be reverted.  /dev is mounted on a tmpfs with context="system_u:object_r:svirt_sandbox_file_t:s0:c1,c2".  Having this label will cause this relabel to fail with operation not supported.

I am not sure how to get the other error to happen with /dev not being on a tmpfs. 

Please revert this patch, or should I open another pull to remove it.

PR #804, opened 2016-05-05
Updating error condition in applying apparmor profile

rajasec:
When applying profiles, when the profile is not there, currently runc throws error saying
no such file or directory.
From user perspective, little confusion to identify where the error occured from libcontainer, so updated the message accordingly


crosbymichael:
LGTM

mrunalp:
LGTM

PR #807, opened 2016-05-08
validator: ensure user doesn't try to mount /sys without userns

### LINKED ISSUE #799: Cannot create user namespaced container without network namespaces
I discovered this while working on rootless containers. It looks like there's some issues using a non-network namespaced setup. This is also blocking rootless containers from having networking (since we need to just use host networking).

```
% sudo runc start test
rootfs_linux.go:53: mounting "/sys" to rootfs "/home/cyphar/src/runc/rootfs" caused "operation not permitted"
```

Here's the config, but the important thing to note is that I've added some dummy user namespace setup and removed the network section from `namespaces`.

```
{
	"ociVersion": "0.6.0-dev",
	"platform": {
		"os": "linux",
		"arch": "amd64"
	},
	"process": {
		"terminal": true,
		"user": {},
		"args": [
			"sh"
		],
		"env": [
			"TERM=xterm"
		],
		"cwd": "/",
		"capabilities": [
			"CAP_AUDIT_WRITE",
			"CAP_KILL",
			"CAP_NET_BIND_SERVICE"
		],
		"rlimits": [
			{
				"type": "RLIMIT_NOFILE",
				"hard": 1024,
				"soft": 1024
			}
		],
		"noNewPrivileges": true
	},
	"root": {
		"path": "rootfs",
		"readonly": true
	},
	"hostname": "runc",
	"mounts": [
		{
			"destination": "/proc",
			"type": "proc",
			"source": "proc"
		},
		{
			"destination": "/dev",
			"type": "tmpfs",
			"source": "tmpfs",
			"options": [
				"nosuid",
				"strictatime",
				"mode=755",
				"size=65536k"
			]
		},
		{
			"destination": "/dev/pts",
			"type": "devpts",
			"source": "devpts",
			"options": [
				"nosuid",
				"noexec",
				"newinstance",
				"ptmxmode=0666",
				"mode=0620",
				"gid=5"
			]
		},
		{
			"destination": "/dev/shm",
			"type": "tmpfs",
			"source": "shm",
			"options": [
				"nosuid",
				"noexec",
				"nodev",
				"mode=1777",
				"size=65536k"
			]
		},
		{
			"destination": "/dev/mqueue",
			"type": "mqueue",
			"source": "mqueue",
			"options": [
				"nosuid",
				"noexec",
				"nodev"
			]
		},
		{
			"destination": "/sys",
			"type": "sysfs",
			"source": "sysfs",
			"options": [
				"nosuid",
				"noexec",
				"nodev",
				"ro"
			]
		},
		{
			"destination": "/sys/fs/cgroup",
			"type": "cgroup",
			"source": "cgroup",
			"options": [
				"nosuid",
				"noexec",
				"nodev",
				"relatime",
				"ro"
			]
		}
	],
	"hooks": {},
	"linux": {
		"resources": {
			"devices": [
				{
					"allow": false,
					"access": "rwm"
				}
			]
		},
		"uidMappings": [
			{
				"hostID": 1000,
				"containerID": 0,
				"size": 100
			}
		],
		"gidMappings": [
			{
				"hostID": 1000,
				"containerID": 0,
				"size": 100
			}
		],
		"namespaces": [
			{
				"type": "user"
			},
			{
				"type": "pid"
			},
			{
				"type": "ipc"
			},
			{
				"type": "uts"
			},
			{
				"type": "mount"
			}
		],
		"maskedPaths": [
			"/proc/kcore",
			"/proc/latency_stats",
			"/proc/timer_stats",
			"/proc/sched_debug"
		],
		"readonlyPaths": [
			"/proc/asound",
			"/proc/bus",
			"/proc/fs",
			"/proc/irq",
			"/proc/sys",
			"/proc/sysrq-trigger"
		]
	}
}
```

Blocking #774.

cyphar:
Inside a user namespace, mount permissions become more nuanced. sysfs
requires you to have either CAP_SYS_ADMIN in the root user namespace
(for us this means that we're not using a user namespace) or that you
have CAP_SYS_ADMIN in the user namespace the network namespace was
created in. This means that having just a user namespace and no private
network namespace will result in errors when mounting sysfs[1]. Warn the
user about this in the validator.

[1]: http://lists.linuxfoundation.org/pipermail/containers/2013-August/033388.html

Closes #799.

/cc @dqminh @mrunalp

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -64,7 +64,16 @@ func (v *ConfigValidator) network(config *configs.Config) error {
 		if len(config.Networks) > 0 || len(config.Routes) > 0 {
 			return fmt.Errorf("unable to apply network settings without a private NET namespace")
 		}
+
+		if config.Namespaces.Contains(configs.NEWUSER) {
+			for _, mount := range config.Mounts {
+				if mount.Device == "sysfs" {
+					return fmt.Errorf("unable to mount sysfs in a USER namespace without a private NET namespace")
+				}
+			}
+		}
 	}
+
 	return nil
 }
 
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -36,6 +36,28 @@ func TestValidateWithInvalidRootfs(t *testing.T) {
 	}
 }
 
+func TestValidateSysfsWithoutNETNamespace(t *testing.T) {
+	config := &configs.Config{
+		Rootfs:      "/var",
+		Namespaces:  []configs.Namespace{{Type: configs.NEWUSER}},
+		UidMappings: []configs.IDMap{{ContainerID: 0, HostID: 1000, Size: 1}},
+		GidMappings: []configs.IDMap{{ContainerID: 0, HostID: 1000, Size: 1}},
+		Mounts: []*configs.Mount{
+			{
+				Destination: "/sys",
+				Device:      "sysfs",
+				Source:      "sysfs",
+			},
+		},
+	}
+
+	validator := validate.New()
+	err := validator.Validate(config)
+	if err == nil {
+		t.Error("Expected error to occur but it was nil")
+	}
+}
+
 func TestValidateNetworkWithoutNETNamespace(t *testing.T) {
 	network := &configs.Network{Type: "loopback"}
 	config := &configs.Config{


cyphar:
PTAL.

/ping @mrunalp @dqminh @hqhq

dqminh:
mmm, I think it's still possible that the user have cap_sys_admin on both the user namespace and the root ? ( though i dont know what's the purpose for that ). I dont think there's a lightweight method to check if an user possess cap_sys_admin either.

So i think this LGTM. I guess we can tune it if user complaints :p

cyphar:
@dqminh I'm not sure how that works actually. Because you lose all capabilities in your previous user namespace if you `setns` into a namespace (or `unshare` it). Maybe there's some odd setup where this happens, but it shouldn't happen within runC -- but we can change it if someone complains (as you said).

crosbymichael:
I'm not really a fan of these checks that prohibit runc from trying and letting the system return the error.  The point being is that when/if this is fixed in future kernels we have to make a change and push an update for runc instead of not having this type of kernel validation and letting the kernel return the errors.

cyphar:
My main justification when it comes to things like this is that the errors people get from the operating system *don't make sense* (you get an error when trying to mount `/sys`, but there's no other information). And if the error happens in the new process we create, we don't even send the error text to the parent so the user only gets `exit status 1` as an error message. I noticed this particularly when working with trying to implement rootless containers, I had to keep recompiling runC with debugging statements and messing around with process setup so I could _actually get the error_.

crosbymichael:
idk, pros and cons on both sides.  i don't know what is best

cyphar:
At the very least, we should probably fix the current issues with not actually outputting the error you get inside the init process. Unfortunately, we still have to do a lot if we want to return errors from inside `nsenter` (we'll have to serialise the JSON manually in C).

mrunalp:
I agree with @crosbymichael. The configuration isn't 100% foolproof and we can provide better examples on how to make certain features work.

hqhq:
I also think we should not prohibit runc from trying because in theory the config could be valid and kernel rules might change. And the poor error message in this case is also pain, maybe some kinds of warning can be a compromise?

crosbymichael:
Ya, if we cannot get a useful error message out of the C code then we should probably do a warning and not abort start for something like this.

cyphar:
The nsenter rewrites have improved the logging part of the nsenter code (though there still is some left to be desired).

PR #808, opened 2016-05-09
Add support to send signal to any processes of the container

hqhq:

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -14,10 +14,10 @@ import (
 )
 
 func killContainer(container libcontainer.Container) error {
-	_ = container.Signal(syscall.SIGKILL, false)
+	_ = container.Signal(syscall.SIGKILL, 0, false)
 	for i := 0; i < 100; i++ {
 		time.Sleep(100 * time.Millisecond)
-		if err := container.Signal(syscall.Signal(0), false); err != nil {
+		if err := container.Signal(syscall.Signal(0), 0, false); err != nil {
 			destroy(container)
 			return nil
 		}
--- a/kill.go
+++ b/kill.go
@@ -51,11 +51,11 @@ var signalMap = map[string]syscall.Signal{
 
 var killCommand = cli.Command{
 	Name:  "kill",
-	Usage: "kill sends the specified signal (default: SIGTERM) to the container's init process",
+	Usage: "kill sends the specified signal (default: SIGTERM) to any of the container's processes (default: init process)",
 	ArgsUsage: `<container-id> [signal]
 
 Where "<container-id>" is the name for the instance of the container and
-"[signal]" is the signal to be sent to the init process.
+"[signal]" is the signal to be sent to the process of the container.
 
 EXAMPLE:
 For example, if the container id is "ubuntu01" the following will send a "KILL"
@@ -67,6 +67,10 @@ signal to the init process of the "ubuntu01" container:
 			Name:  "all, a",
 			Usage: "send the specified signal to all processes inside the container",
 		},
+		cli.IntFlag{
+			Name:  "pid, p",
+			Usage: "specify the pid to which process the signal would be sent (default: init process)",
+		},
 	},
 	Action: func(context *cli.Context) error {
 		if err := checkArgs(context, 1, minArgs); err != nil {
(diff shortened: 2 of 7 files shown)


cyphar:
It should be clearer (especially given the fact that `runc ps` exists) that the `--pid` flag takes the PID of the process inside the container's PID namespace.

crosbymichael:
Kinda like what @cyphar said, I'm not sure that this is any easier than just getting the pid and running a `kill` on it.  

However, I'm not sure how to translate a pid from inside the container to outside like @cyphar said other than joining the pidns and doing the kill there i'm not sure you can translate between the two.

hqhq:
Yeah, I'll think about implement inner pids instead.

crosbymichael:
@hqhq whats the status on this one?  Are you still wanting to implement it?

hqhq:
@crosbymichael I'll let my colleague keep working on this, this PR can be closed for now, thanks for reminding.

xiekeyang:
I found the way to get nested ns PID is described by http://man7.org/linux/man-pages/man5/proc.5.html . 
In chapter **/proc/[pid]/status**, it present that,


```
* NStgid Thread group ID (i.e., PID) in each of the PID namespaces of which [pid] is a member.  The leftmost entry shows the value with respect to the PID namespace of the reading process, followed by the value in successively nested inner namespaces.  (Since Linux 4.1.)
* NSpid Thread ID in each of the PID namespaces of which [pid] is a member.  The fields are ordered as for NStgid.  (Since Linux 4.1.)
* NSpgid Process group ID in each of the PID namespaces of which [pid] is a member.  The fields are ordered as for NStgid.  (Since Linux 4.1.)
* NSsid descendant namespace session ID hierarchy Session ID in each of the PID namespaces of which [pid] is a member. The fields are ordered as for NStgid.  (Since Linux 4.1.)
```

We notice that it must run on the kernel over version 4.1. So, do you consider that we do these execution through global PID? It seems that global PID is as well. We can get global PID by `container.Processes()`.
Thanks a lot!

hqhq:
@crosbymichael @cyphar Is it OK that we add a newly kernel version depended feature? Or is it enough that we implement it with global pid, since it's basically an admin behavior.

cyphar:
I would personally try to avoid adding a kernel dependency for a kernel as new as 4.1. Though, if possible it would be nice to use the feature if available and otherwise fall back to the old method. However, I'm a bit fuzzy on what the benefits of using `/proc/<pid>/status` are over just using the global PID method (which is actually currently the cgroup method, which translates PIDs to the _current_ namespace).

xiekeyang:
@cyphar ,agree! actually I think the global PID is nice.

hqhq:
@cyphar @crosbymichael I've rebased this PR, PTAL.

cyphar:
From what I can see, this still doesn't have much benefit over getting the PID from `runc ps` and doing a normal `kill` on it. I could see this being useful if it supported PIDs from **inside** the container, which would mean that if you had a `rootfs` which didn't have `kill` you could still kill processes. But even then I'm not convinced. Can you explain in what case this would be better than just using an old-fashioned `kill` command?

hqhq:
@cyphar Nothing special, mostly just for convenience and consistency, three reasons that I can think of:

- We have upper level tools call runc api to send signals to init, and want to send signals to other processes in container, it'll be better if we can use a consistent api;
- As the issue pinpointed in #1126 , container init process only receive `SIGKILL` or `SIGSTOP` by default, so it doesn't make much sense we actually support all other signals, especially the default one is `SIGTERM`, it'll look reasonable if we can use these signals to other processes by default;
- `kill` command didn't do anything special in the first place, we can also use plain unix kill command to send signals to container init process, so I think what `kill` command does is supposed to be a wrap of unix kill command, we just expand the target scope a little bit.

hqhq:
ping @crosbymichael @cyphar @mrunalp

cyphar:
I'm still not really seeing what benefit this gives.

> As the issue pinpointed in #1126 , container init process only receive SIGKILL or SIGSTOP by default, so it doesn't make much sense we actually support all other signals, especially the default one is SIGTERM, it'll look reasonable if we can use these signals to other processes by default;

But that's not true if you use a container init (like Docker is doing now-a-days with `tini`), or if you have a process that explicitly has signal handlers set up.

> kill command didn't do anything special in the first place, we can also use plain unix kill command to send signals to container init process, so I think what kill command does is supposed to be a wrap of unix kill command, we just expand the target scope a little bit.

I see the "kill" command as meaning that you are semantically killing *the container* -- internally that happens to only require that you send the signal to the init.

AkihiroSuda:
@hqhq Are you still interested in this?

hqhq:
> @hqhq Are you still interested in this?

It has been too long, closing it...

PR #811, opened 2016-05-10
Adding kernel mem tcp for update command

rajasec:
Added kernel mem TCP  limit for update command
Flag name of cpu-share changed to cpu-shares

### DIFF
--- a/update.go
+++ b/update.go
@@ -64,7 +64,7 @@ other options are ignored.
 			Usage: "CPU hardcap limit (in usecs). Allowed cpu time in a given period.",
 		},
 		cli.StringFlag{
-			Name:  "cpu-share",
+			Name:  "cpu-shares",
 			Usage: "CPU shares (relative weight vs. other containers)",
 		},
 		cli.StringFlag{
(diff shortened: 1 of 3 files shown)


update.go
```diff
@@ -151,8 +156,9 @@ other options are ignored.
 			for opt, dest := range map[string]*uint64{
 				"cpu-period":         r.CPU.Period,
 				"cpu-quota":          r.CPU.Quota,
-				"cpu-share":          r.CPU.Shares,
+				"cpu-shares":         r.CPU.Shares,
```
runcom: Was this a bug?

rajasec: No, this was not an bug, I have changed the name to cpu-shares for consistency.

runcom: alright, I think this change needs its own commit (or PR better) since you're changing a flag, aren't you?

rajasec: The main intent of PR is to add the kernelTCP limits, along with that I had modified this flag too
Yes flag name for cpu-share changed. I have updated the update.bats with cpu.shares too.

runcom: > The main intent of PR is to add the kernelTCP limits

so that flag change would need its own commit, I'll let other maintainers chime in though..

hqhq:
Yes commits should be edited, option change should be in a separate commit, and kmem tcp support and bats change should be combined, because we should make every commit compileable and pass all tests.

PR #812, opened 2016-05-10
integration: fix cgroup parsing

cyphar:
On some systems, the cgroup hierarchies are grouped together
(cpu,cpuacct). In order to avoid fake failures, update the cgroup
parsing to just check whether or not the mountinfo options *contain* the
cgroup type.

/cc @hqhq

### DIFF
--- a/script/test_Dockerfile
+++ b/script/test_Dockerfile
@@ -6,6 +6,7 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
 RUN apt-get update && apt-get install -y \
     build-essential \
     curl \
+    gawk \
     iptables \
     libaio-dev \
     libcap-dev \
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -9,8 +9,8 @@ CGROUP_MEMORY=""
 TEST_CGROUP_NAME="runc-cgroups-integration-test"
 
 function init_cgroup_path() {
-   base_path=$(grep "rw,"  /proc/self/mountinfo | grep -i -m 1 'MEMORY$' | cut -d ' ' -f 5)
-   CGROUP_MEMORY="${base_path}/${TEST_CGROUP_NAME}"
+	base_path=$(grep "cgroup"  /proc/self/mountinfo | gawk 'toupper($NF) ~ /\<MEMORY\>/ { print $5; exit }')
+	CGROUP_MEMORY="${base_path}/${TEST_CGROUP_NAME}"
 }
 
 function teardown() {
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -2,10 +2,7 @@
 
 load helpers
 
-UPDATE_TEST_RUNC_ROOT="$BATS_TMPDIR/runc-cgroups-integration-test"
-
 CGROUP_MEMORY=""
-
 TEST_CGROUP_NAME="runc-cgroups-integration-test"
 
 function init_cgroup_path() {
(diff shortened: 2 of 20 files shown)


script/test_Dockerfile
```diff
@@ -6,6 +6,7 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
 RUN apt-get update && apt-get install -y \
     build-essential \
     curl \
+	gawk \
```
hqhq: I think here should be spaces instead of tab.

cyphar: Fix'd. PTAL.

tests/integration/update.bats
```diff
@@ -11,7 +11,7 @@ CGROUP_BLKIO=""
 
 function init_cgroup_path() {
     for g in MEMORY CPUSET CPU BLKIO; do
-        base_path=$(grep "rw,"  /proc/self/mountinfo | grep -i -m 1 "$g\$" | cut -d ' ' -f 5)
+        base_path=$(grep "rw,"  /proc/self/mountinfo | gawk 'toupper($10) ~ /\<'${g}'\>/ { print $5 }')
```
mrunalp: On my fedora system, the cgroup options are $11 not $10.

mrunalp: We could probably do
```
base_path=$(grep "rw,"  /proc/self/mountinfo | gawk 'toupper($NF) ~ /\<'${g}'\>/ { print $5 }')
```

cyphar: Even more weirdly, the number changes between being in a container and being on the host. I'll use `$NF` instead then.

tests/integration/cgroups.bats
```diff
@@ -2,20 +2,17 @@
...
-UPDATE_TEST_RUNC_ROOT="$BATS_TMPDIR/runc-cgroups-integration-test"
-
 CGROUP_MEMORY=""
-
 TEST_CGROUP_NAME="runc-cgroups-integration-test"
 
 function init_cgroup_path() {
-   base_path=$(grep "rw,"  /proc/self/mountinfo | grep -i -m 1 'MEMORY$' | cut -d ' ' -f 5)
-   CGROUP_MEMORY="${base_path}/${TEST_CGROUP_NAME}"
+	base_path=$(grep "rw,"  /proc/self/mountinfo | gawk 'toupper($NF) ~ /\<MEMORY\>/ { print $5 }')
```
mrunalp: I think we could probably just grep for `cgroup` instead of `rw,`. WDYT?

mlaventure:
Thanks for fixing this @cyphar :)

Just realized that on ubuntu LTS 16.04 that will return 2 lines though:
```
~# grep "cgroup" /proc/self/mountinfo | gawk 'toupper($NF) ~ /\<CPUSET\>/ { print $5 }'
/sys/fs/cgroup/cpuset
/run/lxcfs/controllers/cpuset
```

mrunalp:
@mlaventure Hmm, Are those mounts only present when you install lxc?
This is getting messy :(

mlaventure:
@mrunalp I think it comes pre-installed o 16.04 now :/. This was on a DigitalOcean droplet, but I doubt they would have taken the extra steps of adding lxcfs if it wasn't there by default.

tests/integration/cgroups.bats
cyphar: Sure, `grep cgroup` works.

cyphar:
We could just do `| head -n1`, since each mountpoint should work properly.

cyphar:
@mlaventure @mrunalp I've fixed up both concerns. Since all cgroup mounts are (basically) equal, we can just pick the first one. Which is what I've done.

mlaventure:
@cyphar 

This is the cause of the janky failure on your pr:
`run "$RUNC" --root $UPDATE_TEST_RUNC_ROOT update test_update --kernel-memory-tcp 41943040`

`--root ...` wasn't removed on this one.

cyphar:
@mlaventure Fixed. :D

crosbymichael:
LGTM

mrunalp:
LGTM

PR #816, opened 2016-05-11
Revert "Need to make sure labels applied to /dev"

cyphar:
Reverts opencontainers/runc#796

As per @rhatdan's comment in #796:

> @mrunalp @crosbymichael This patch should be reverted.  /dev is mounted on a tmpfs with context="system_u:object_r:svirt_sandbox_file_t:s0:c1,c2".  Having this label will cause this relabel to fail with operation not supported.

> I am not sure how to get the other error to happen with /dev not being on a tmpfs. 

> Please revert this patch, or should I open another pull to remove it.


mrunalp:
LGTM
@rhatdan Do we need another patch with `context=` for /dev?

rhatdan:
No that seems to be already there.  The tmpfs is getting mounted with the correct label if you set mount-label in your config.json file.

crosbymichael:
LGTM

PR #820, opened 2016-05-12
Handling destroy of container

rajasec:
After killing the detached container, it goes back to destroyed state
When container is in destroyed state, if we start the container via exec, container started. After exiting the container, it goes back to destroyed state ( still retains the state.json).

Added the logic both start and exec to depend on container Processes interface for destroying the container rather than shouldDestroy variable.
Tested the following conditions
       runc start test
        runc exec test
	exit the container
state.json is retained

       runc start test -d
        runc kill test SIGKILL
        runc exec test
              exit the container
state.json is cleared properly

runc start test -d
runc kill test SIGKILL
runc delete test

runc delete is not impacted by this change.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -115,7 +115,6 @@ func execProcess(context *cli.Context) (int, error) {
 	}
 	r := &runner{
 		enableSubreaper: !context.Bool("no-subreaper"),
-		shouldDestroy:   false,
 		container:       container,
 		console:         context.String("console"),
 		detach:          detach,
--- a/start.go
+++ b/start.go
@@ -127,7 +127,6 @@ func startContainer(context *cli.Context, spec *specs.Spec) (int, error) {
 	}
 	r := &runner{
 		enableSubreaper: !context.Bool("no-subreaper"),
-		shouldDestroy:   true,
 		container:       container,
 		listenFDs:       listenFDs,
 		console:         context.String("console"),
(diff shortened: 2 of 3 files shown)


hqhq:
Integration tests can be added in this case.
And add useful PR descriptions to commit message would help, and please always do that. Thanks.

hqhq:
```
# runc start test -d
# runc kill test KILL
(container test is in destroyed status)
# runc exec test bash
        exit the container
(container test is in destroyed status)
```

Why is this a problem? I think it is reasonable that if a container goes to destroyed status, it should be cleaned up by `runc delete`.

rajasec:
@hqhq 
Thanks for your review, will add  the commit message in PRs.

I was thinking from state perspective for the above scenario.
destroyed status -> running status  -> ( upon exit) destroyed status

Instead
destroyed status -> running status ->( upon exit) clean up the container, do not fall back to destroyed state again.

For this scenario
Moreover as there is no process running inside, my view was there is no point is moving to destroyed state back if we exit the container.
I'm depending on cgroup.procs( container.Processes interface) for any container clean up, in this case as there is no process running so clean up the container

crosbymichael:
This state handling should be solved via the create/start split now. I'll close this for now but let me know if you find any inconsistencies with the states in master.

PR #825, opened 2016-05-16
Add comments for error cases in status functions

hqhq:

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1044,6 +1044,9 @@ func (c *linuxContainer) isRunning() (bool, error) {
 	// return Running if the init process is alive
 	if err := syscall.Kill(c.initProcess.pid(), 0); err != nil {
 		if err == syscall.ESRCH {
+			// It means the process does not exist anymore, could happen when the
+			// process exited just when we call the function, we should not return
+			// error in this case.
 			return false, nil
 		}
 		return false, newSystemErrorWithCausef(err, "sending signal 0 to pid %d", c.initProcess.pid())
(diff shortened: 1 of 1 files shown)


libcontainer/container_linux.go
```diff
@@ -1043,9 +1043,6 @@ func (c *linuxContainer) isRunning() (bool, error) {
 	}
 	// return Running if the init process is alive
 	if err := syscall.Kill(c.initProcess.pid(), 0); err != nil {
-		if err == syscall.ESRCH {
-			return false, nil
-		}
```
dqminh: this is a known error from kill, means that the process does not exist anymore (in waiting state) for example, so i think the original code makes sense.

libcontainer/container_linux.go
```diff
@@ -1054,9 +1051,6 @@ func (c *linuxContainer) isRunning() (bool, error) {
 func (c *linuxContainer) isPaused() (bool, error) {
 	data, err := ioutil.ReadFile(filepath.Join(c.cgroupManager.GetPaths()["freezer"], "freezer.state"))
 	if err != nil {
-		if os.IsNotExist(err) {
-			return false, nil
-		}
```
dqminh: iirc i think this is where we check if the host has freezer cgroup, if not the application obviously cant be frozen, hence we can resolved the pause state here.

cyphar:
I agree with @dqminh, it looks like they are there because "errors" from the OS are really just "this already happened" states. We could add comments though.

hqhq:
@dqminh Thanks, I think you are right, I've updated the PR to add comments instead.

crosbymichael:
LGTM

cyphar:
LGTM

PR #506, opened 2016-01-22
Split start/create

duglin:
First pass at splitting create and start.
`runc create` will create all of the namespaces, except PID.
`runc start` will then run the main proc.
There's `runc delete` too for when you're done.
Also added `runc batch <list of commands file>` to show how you could do create+a series of cmds.
And added `runc run <cmd>` which is like `runc exec` but uses a main proc - again just for demo purposes. Not sure if we want to add batch/run to runc proper but it was an interesting exercise.

Any way, I'm not convinced all of this is correct - i think the new 'mount' property probably needs to go away but I wanted to let people play with it to see how it felt to use create and start separately.

PR #827, opened 2016-05-16
Implement create and start

crosbymichael:
This implements create and start in runc without the need for a unix socket or other complexity.  It implements it by blocking the init process waiting on a `SIGCONT` before the users process is started.  

This does not remove hooks, that can be done separately.  This also updates the libcontainer stats to correctly report if the container is created vs running(user code).  

It does not bind mount namespaces, if you want namespaces to be bind mounted then you can write code to bind them after `create` returns and before calling `start`.  

It retains the current functionality of `start` today by adding a `runc run` command that does the same workflow as today. 

Closes #506

### DIFF
--- a/create.go
+++ b/create.go
@@ -43,30 +43,18 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Usage: "do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk",
 		},
 	},
-	Action: func(context *cli.Context) {
-		bundle := context.String("bundle")
-		if bundle != "" {
-			if err := os.Chdir(bundle); err != nil {
-				fatal(err)
-			}
-		}
-		spec, err := loadSpec(specConfig)
+	Action: func(context *cli.Context) error {
+		spec, err := setupSpec(context)
 		if err != nil {
-			fatal(err)
-		}
-		notifySocket := os.Getenv("NOTIFY_SOCKET")
-		if notifySocket != "" {
-			setupSdNotify(spec, notifySocket)
-		}
-		if os.Geteuid() != 0 {
-			fatalf("runc should be run as root")
+			return err
 		}
 		status, err := startContainer(context, spec, true)
 		if err != nil {
-			fatal(err)
+			return err
 		}
 		// exit with the container's exit status so any external supervisor is
 		// notified of the exit with the correct exit status.
 		os.Exit(status)
+		return nil
 	},
 }
--- a/create.go
+++ b/create.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"os"
+
+	"github.com/codegangsta/cli"
+)
+
+var createCommand = cli.Command{
+	Name:  "create",
+	Usage: "create a container",
+	ArgsUsage: `<container-id>
+
+Where "<container-id>" is your name for the instance of the container that you
+are starting. The name you provide for the container instance must be unique on
+your host.`,
+	Description: `The create command creates an instance of a container for a bundle. The bundle
+is a directory with a specification file named "` + specConfig + `" and a root
+filesystem.
+
+The specification file includes an args parameter. The args parameter is used
+to specify command(s) that get run when the container is started. To change the
+command(s) that get executed on start, edit the args parameter of the spec. See
+"runc spec --help" for more explanation.`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "bundle, b",
+			Value: "",
+			Usage: `path to the root of the bundle directory, defaults to the current directory`,
+		},
+		cli.StringFlag{
+			Name:  "console",
+			Value: "",
+			Usage: "specify the pty slave path for use with the container",
+		},
+		cli.StringFlag{
+			Name:  "pid-file",
+			Value: "",
+			Usage: "specify the file to write the process id to",
+		},
+		cli.BoolFlag{
+			Name:  "no-pivot",
+			Usage: "do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk",
+		},
+	},
+	Action: func(context *cli.Context) {
+		bundle := context.String("bundle")
+		if bundle != "" {
+			if err := os.Chdir(bundle); err != nil {
+				fatal(err)
+			}
+		}
+		spec, err := loadSpec(specConfig)
+		if err != nil {
+			fatal(err)
+		}
+		notifySocket := os.Getenv("NOTIFY_SOCKET")
+		if notifySocket != "" {
+			setupSdNotify(spec, notifySocket)
+		}
+		if os.Geteuid() != 0 {
+			fatalf("runc should be run as root")
+		}
+		status, err := startContainer(context, spec, true)
+		if err != nil {
+			fatal(err)
+		}
+		// exit with the container's exit status so any external supervisor is
+		// notified of the exit with the correct exit status.
+		os.Exit(status)
+	},
+}
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -17,16 +17,12 @@ type Status int
 const (
 	// Created is the status that denotes the container exists but has not been run yet.
 	Created Status = iota
-
 	// Running is the status that denotes the container exists and is running.
 	Running
-
 	// Pausing is the status that denotes the container exists, it is in the process of being paused.
 	Pausing
-
 	// Paused is the status that denotes the container exists, but all its processes are paused.
 	Paused
-
 	// Stopped is the status that denotes the container does not have a created or running process.
 	Stopped
 )
@@ -126,6 +122,17 @@ type BaseContainer interface {
 	// ContainerPaused - Container is paused,
 	// Systemerror - System error.
 	Start(process *Process) (err error)
+
+	// StartI immediatly starts the process inside the conatiner.  Returns error if process
+	// fails to start.  It does not block waiting for a SIGCONT after start returns but
+	// sends the signal when the process has completed.
+	//
+	// errors:
+	// ContainerDestroyed - Container no longer exists,
+	// ConfigInvalid - config is invalid,
+	// ContainerPaused - Container is paused,
+	// Systemerror - System error.
+	StartI(process *Process) (err error)
 
 	// Destroys the container after killing all running processes.
 	//
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -123,7 +123,7 @@ type BaseContainer interface {
 	// SystemError - System error.
 	Start(process *Process) (err error)
 
-	// StartI immediatly starts the process inside the conatiner.  Returns error if process
+	// Run immediatly starts the process inside the conatiner.  Returns error if process
 	// fails to start.  It does not block waiting for a SIGCONT after start returns but
 	// sends the signal when the process has completed.
 	//
@@ -132,7 +132,7 @@ type BaseContainer interface {
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
 	// SystemError - System error.
-	StartI(process *Process) (err error)
+	Run(process *Process) (err error)
 
 	// Destroys the container after killing all running processes.
 	//
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -15,7 +15,7 @@ import (
 type Status int
 
 const (
-	// Created is the status that denotes the container exists but has not been run yet
+	// Created is the status that denotes the container exists but has not been run yet.
 	Created Status = iota
 
 	// Running is the status that denotes the container exists and is running.
@@ -27,15 +27,8 @@ const (
 	// Paused is the status that denotes the container exists, but all its processes are paused.
 	Paused
 
-	// Destroyed is the status that denotes the container does not exist.
-	Destroyed
-
 	// Stopped is the status that denotes the container does not have a created or running process.
 	Stopped
-
-	// Initialized is the status where the container has all the namespaces created but the user
-	// process has not been start.
-	Initialized
 )
 
 func (s Status) String() string {
@@ -48,12 +41,8 @@ func (s Status) String() string {
 		return "pausing"
 	case Paused:
 		return "paused"
-	case Destroyed:
-		return "destroyed"
 	case Stopped:
 		return "stopped"
-	case Initialized:
-		return "initialized"
 	default:
 		return "unknown"
 	}
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -76,13 +76,13 @@ type BaseContainer interface {
 	//
 	// errors:
 	// ContainerDestroyed - Container no longer exists,
-	// Systemerror - System error.
+	// SystemError - System error.
 	Status() (Status, error)
 
 	// State returns the current container's state information.
 	//
 	// errors:
-	// Systemerror - System error.
+	// SystemError - System error.
 	State() (*State, error)
 
 	// Returns the current config of the container.
@@ -92,7 +92,7 @@ type BaseContainer interface {
 	//
 	// errors:
 	// ContainerDestroyed - Container no longer exists,
-	// Systemerror - System error.
+	// SystemError - System error.
 	//
 	// Some of the returned PIDs may no longer refer to processes in the Container, unless
 	// the Container state is PAUSED in which case every PID in the slice is valid.
@@ -102,7 +102,7 @@ type BaseContainer interface {
 	//
 	// errors:
 	// ContainerDestroyed - Container no longer exists,
-	// Systemerror - System error.
+	// SystemError - System error.
 	Stats() (*Stats, error)
 
 	// Set resources of container as configured
@@ -110,7 +110,7 @@ type BaseContainer interface {
 	// We can use this to change resources when containers are running.
 	//
 	// errors:
-	// Systemerror - System error.
+	// SystemError - System error.
 	Set(config configs.Config) error
 
 	// Start a process inside the container. Returns error if process fails to
@@ -120,7 +120,7 @@ type BaseContainer interface {
 	// ContainerDestroyed - Container no longer exists,
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
-	// Systemerror - System error.
+	// SystemError - System error.
 	Start(process *Process) (err error)
 
 	// StartI immediatly starts the process inside the conatiner.  Returns error if process
@@ -131,7 +131,7 @@ type BaseContainer interface {
 	// ContainerDestroyed - Container no longer exists,
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
-	// Systemerror - System error.
+	// SystemError - System error.
 	StartI(process *Process) (err error)
 
 	// Destroys the container after killing all running processes.
@@ -140,12 +140,12 @@ type BaseContainer interface {
 	// No error is returned if the container is already destroyed.
 	//
 	// errors:
-	// Systemerror - System error.
+	// SystemError - System error.
 	Destroy() error
 
 	// Signal sends the provided signal code to the container's initial process.
 	//
 	// errors:
-	// Systemerror - System error.
+	// SystemError - System error.
 	Signal(s os.Signal) error
 }
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -29,6 +29,13 @@ const (
 
 	// Destroyed is the status that denotes the container does not exist.
 	Destroyed
+
+	// Stopped is the status that denotes the container does not have a created or running process.
+	Stopped
+
+	// Initialized is the status where the container has all the namespaces created but the user
+	// process has not been start.
+	Initialized
 )
 
 func (s Status) String() string {
@@ -43,6 +50,10 @@ func (s Status) String() string {
 		return "paused"
 	case Destroyed:
 		return "destroyed"
+	case Stopped:
+		return "stopped"
+	case Initialized:
+		return "initialized"
 	default:
 		return "unknown"
 	}
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -181,7 +181,7 @@ func (c *linuxContainer) Start(process *Process) error {
 	if err != nil {
 		return err
 	}
-	doInit := status == Destroyed
+	doInit := status == Stopped
 	parent, err := c.newParentProcess(process, doInit)
 	if err != nil {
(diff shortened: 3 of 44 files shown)


julz:
Nice.

crosbymichael:
@julz thanks, plz take the time to play with it, runc is fully operational.  lets make sure that it will fit and work with all of our needs but I think this implementation is simple and clean and will work fine.

start.go
```diff
@@ -1,138 +1,27 @@
...
-			setupSdNotify(spec, notifySocket)
-		}
-
-		if os.Geteuid() != 0 {
-			fatalf("runc should be run as root")
-		}
-
-		status, err := startContainer(context, spec)
-		if err != nil {
+		if err := container.Signal(syscall.SIGCONT); err != nil {
```
wking: This probably needs at least a `container.Status()` guard like you added to `delete` in 572055d, since you only want to send SIGCONT to runtime code (and not send it after user-specified code has been executed).  That's not quite enough though, since “[the pending signal set is preserved across an execve(2)][1]”.  So “check the state and send SIGCONT if it was ‘created’” is going to be racy, and that race may result in user code seeing the extra SIGCONTs.  A more robust solution would lock a resource (setting a flag in the state registry?  Hold a Unix socket open?) when triggering code execution to avoid racing between two ‘start’ calls.

[1]: http://man7.org/linux/man-pages/man7/signal.7.html

mlaventure: We can also simply reset the signal handler before doing the execve.

wking: On Mon, May 16, 2016 at 09:00:22PM -0700, Kenfe-Mickaël Laventure wrote:
> We can also simply reset the signal handler before doing the execve.

That happens automatically, no?  Also from signal(7):

  “During an execve(2), the dispositions of handled signals are reset
  to the default; the dispositions of ignored signals are left
  unchanged.”

What needs to happen is that after a SIGCONT is received, you block
(somehow) ‘create’ from sending further SIGCONTs, then consume any
SIGCONTs from the pending queue, and then execve the user code.

libcontainer/standard_init_linux.go
```diff
@@ -150,6 +152,18 @@ func (l *linuxStandardInit) Init() error {
 			return err
 		}
 	}
-
-	return system.Execv(l.config.Args[0], l.config.Args[0:], os.Environ())
+	// check for the arg before waiting to make sure it exists and it is returned
+	// as a create time error
+	name, err := exec.LookPath(l.config.Args[0])
```
wking: If folks intend to never call `start` (which seems fine to me for some use-cases), then we probably don't want to require a `process` config at all.  If they do intend to call `start`, whether the command exists at `execve`-time seems different enough from whether the command exists at `create` time for folks to want this early check to at least be optional (e.g. “don't worry, by the time I call `start` I'll have created that command”).

utils_linux.go
```diff
@@ -215,7 +216,7 @@ func (r *runner) run(config *specs.Process) (int, error) {
 		r.destroy()
 		return -1, err
 	}
-	tty, err := setupIO(process, rootuid, r.console, config.Terminal, r.detach)
+	tty, err := setupIO(process, rootuid, r.console, config.Terminal, r.detach || r.create)
```
wking: I don't think the create/start split should impact I/O detachment.  The new `create` case is just like the old handling (`run` in this PR), so you should be able to do `create --detach` or not as you see fit.  Folks calling this from the command line, etc., can always background their `create` process, do their setup, call `start`, and then foreground their `create` process.

utils_linux.go
```diff
@@ -240,7 +241,15 @@ func (r *runner) run(config *specs.Process) (int, error) {
...
-	if r.detach {
+	if !r.create {
+		if err := process.Signal(syscall.SIGCONT); err != nil {
+			r.terminate(process)
+			r.destroy()
+			tty.Close()
+			return -1, err
+		}
+	}
+	if r.detach || r.create {
```
wking: My [earlier][1] comment applies here too.

[1]: https://github.com/opencontainers/runc/pull/827/files#r63464781

utils_linux.go
crosbymichael: No, it is correct as is.

utils_linux.go
crosbymichael: negative

wking: On Tue, May 17, 2016 at 11:11:16AM -0700, Michael Crosby wrote:
> > +	if r.detach || r.create {
> 
> negative 

Can you unpack that a bit?  Why do you need to close the TTY in the
‘create’ case?  Why aren't you waiting on the container process and
returning it's error code?

crosbymichael: When create returns without an error then the container is created and ready to be started.  You cannot attach because create cannot return.

wking: On Tue, May 17, 2016 at 11:25:06AM -0700, Michael Crosby wrote:
> > +	if r.detach || r.create {
>
> When create returns without an error then the container is created
> and ready to be started.  You cannot attach because create cannot
> return.

But somebody needs to stay around to collect the container exit code,
tear down cgroups, de-register the state, etc., right?  See waitpid
notes in [1].  In the ccon implementation the host-side ‘create’
process stays around to reap the container process [2], and folks who
want to get their terminal back to run ‘start’ can background the
‘create’ process [3].

[1]: https://github.com/opencontainers/runtime-spec/pull/384#issuecomment-213673145
[2]: https://github.com/wking/ccon/tree/create-start#lifecycle
[3]: https://github.com/opencontainers/runc/pull/827/#discussion_r63464781

wking: On Tue, May 17, 2016 at 11:32:28AM -0700, W. Trevor King wrote:
> But somebody needs to stay around to collect the container exit
> code, tear down cgroups, de-register the state, etc., right?

Rephrased in Go, this is mostly “But somebody needs to call
container.Destroy()”, and that's your separate ‘delete’ call.  The
only thing I'm missing is a way to collect the container exit code.

crosbymichael: Are you asking a question or what are you trying to get at? 

No, the create process does not have to stick around to reap, the proc calling create can handle that and orchestrate the entire thing from create to destroy.

wking: On Tue, May 17, 2016 at 01:05:45PM -0700, Michael Crosby wrote:
> Are you asking a question or what are you trying to get at? 

I'm trying to decide how you figure out the exit code for the
container process.  It sounds like you're saying:

  “The runtime will not track or provide access to this information.
  The process that called ‘create’ should setup its own wait to
  collect that information and call ‘delete’.”

wait(1p) is a fairly convenient way to do that from the shell, so I'm
fine with this approach, but it is shifting that reap/cleanup
functionality from the runtime (which used to handle it) to the
caller, so we probably want clear docs to this effect.

crosbymichael:
```go
package main

import (
	"io/ioutil"
	"log"
	"os"
	"os/exec"
	"strconv"
	"sync"
	"syscall"

	"github.com/docker/containerd/osutils"
)

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

func run() error {
	if err := osutils.SetSubreaper(1); err != nil {
		return err
	}
	if err := os.Chdir("/containers/redis"); err != nil {
		return err
	}
	if err := runc("create", "--pid-file", "/tmp/mypid", "test"); err != nil {
		return err
	}
	data, err := ioutil.ReadFile("/tmp/mypid")
	if err != nil {
		return err
	}
	pid, err := strconv.Atoi(string(data))
	if err != nil {
		return err
	}
	p, err := os.FindProcess(pid)
	if err != nil {
		return err
	}

	wg := &sync.WaitGroup{}
	wg.Add(1)
	go func() {
		defer wg.Done()
		s, err := p.Wait()
		if err != nil {
			log.Fatal(err)
		}
		log.Printf("process exited with %d\n", s.Sys().(syscall.WaitStatus).ExitStatus())
		runc("delete", "test")
	}()
	if err := runc("start", "test"); err != nil {
		return err
	}
	wg.Wait()
	return nil
}

func runc(args ...string) error {
	cmd := exec.Command("runc", args...)
	cmd.Stdin = os.Stdin
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	return cmd.Run()
}
```

wking: On Tue, May 17, 2016 at 01:30:08PM -0700, W. Trevor King wrote:
> wait(1p) is a fairly convenient way to do that from the shell…

Actually, this doesn't work:

  $ (sh -c 'sleep 60 &') &
  $ ps -u wking | grep sleep | grep -v grep
  22881 pts/13   00:00:00 sleep
  wking@ullr ~ $ wait 22881
  bash: wait: pid 22881 is not a child of this shell

And waitpid(2) is talking about child processes, not ancestor
processes.  So I'm back to wondering how you collect the container
exit status when the ‘create’ process exits early.

wking: On Tue, May 17, 2016 at 01:36:38PM -0700, Michael Crosby wrote:
> 	if err := osutils.SetSubreaper(1); err != nil {

This is fairly low-level for “I'd like to collect the container's exit
code”.  I think callers should at least have the option of having the
‘create’ process stay attached through container death.

That also allows the create process to do psuedo-terminal copying [1],
although the OCI spec around process.terminal is currently not very
clear to me [2].

[1]: https://github.com/wking/ccon/tree/create-start#terminal
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/hcbwL4T9lGs
     Subject: More detail for process.terminal?
     Date: Thu, 14 Jan 2016 14:44:08 -0800
     Message-ID: <20160114224408.GM6362@odin.tremily.us>

libcontainer/standard_init_linux.go
cyphar: I agree, we shouldn't error out during `create`.

hqhq:
So if host rebooted, all created containers would be gone? Is that acceptable?

wking:
On Fri, May 20, 2016 at 01:54:26AM -0700, Qiang Huang wrote:
> So if host rebooted, all created containers would be gone? Is that
> acceptable?

That's fine for me (it's how all my other processes work ;).  If you
want to restore a container, have an init system set it up on boot
(possibly using checkpoint/restore, depending on how much you want to
preserve).  But that all seems like it's out of scope for the runtime
layer.

mrunalp:
@hqhq I think that is okay as this isn't the same as docker create. As long as we are clear in the spec, it should be fine.

vishh:
One of the use-cases for hooks was to customize mounts on demand. Would it be possible to not pivot root as part of create and switch root only on start?

wking:
On Fri, May 20, 2016 at 03:31:07PM -0700, Vish Kannan wrote:
> Would it be possible to not pivot root as part of create and switch
> root only on start?

I don't think that's a good idea, because you may be using the
container process to hold open a namespace.  In that case, there's no
reason to call ‘start’, but you still want the pivoted root in the
mount namespace.

I'm still not clear on why the pre-pivot mounts need to be dynamic,
instead of setting them up in config.json's mounts during pre-create
processing.

crosbymichael:
FYI

This change makes us require go 1.6 because previous versions of go would not let you handle SIGCONT, it just ignores it and blocks forever.

libcontainer/integration/template_test.go
```diff
@@ -89,12 +89,14 @@ func newTemplateConfig(rootfs string) *configs.Config {
 				Data:        "mode=1777,size=65536k",
 				Flags:       defaultMountFlags,
 			},
-			{
-				Source:      "mqueue",
-				Destination: "/dev/mqueue",
-				Device:      "mqueue",
-				Flags:       defaultMountFlags,
-			},
+			/*
```
mrunalp: Was this temporarily commented out?

crosbymichael: bug

libcontainer/container.go
```diff
@@ -127,6 +123,17 @@ type BaseContainer interface {
...
 
+	// StartI immediatly starts the process inside the conatiner.  Returns error if process
+	// fails to start.  It does not block waiting for a SIGCONT after start returns but
+	// sends the signal when the process has completed.
+	//
+	// errors:
+	// ContainerDestroyed - Container no longer exists,
+	// ConfigInvalid - config is invalid,
+	// ContainerPaused - Container is paused,
+	// Systemerror - System error.
```
LK4D4: SystemError

libcontainer/container_linux.go
```diff
@@ -1027,31 +1054,47 @@ func (c *linuxContainer) refreshState() error {
...
+		return Stopped, newSystemErrorWithCausef(err, "sending signal 0 to pid %d", pid)
+	}
+	// check if the process that is running is the init process or the user's process.
+	// this is the difference between the container Running and Created.
+	environ, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/environ", pid))
+	if err != nil {
+		return Stopped, newSystemErrorWithCausef(err, "reading /proc/%d/environ", pid)
+	}
+	check := []byte("_LIBCONTAINER")
+	for _, v := range bytes.Split(environ, []byte("\x00")) {
```
LK4D4: maybe I don't see something, but I think it can be just `if bytes.Contains(environ, check)`

libcontainer/container.go
LK4D4: Probably comment for Start should be changed to describe why it's waiting on signal.

LK4D4:
I see that all test were changed so they use old version. Maybe couple of tests for create/start are needed.

libcontainer/container_linux.go
crosbymichael: done

duglin:
would it be bad if we had the `console` flag default to `/dev/pts/ptmx` for `runc create`?  From a UX perspective it would be easier to do that than to force all users to remember to supply it. W/o some value for that flag I think runc fails anyway.

crosbymichael:
@duglin No, because not all containers use a TTY.

duglin:
@crosbymichael how do I do a `runc create` w/o a `--console` flag?  I can't seem to get it to not error out.

crosbymichael:
@duglin does your spec have `terminal` true?

duglin:
ah that was it - thanks

duglin:
:-) we could still have it default to that value when terminal is true

LK4D4:
@crosbymichael I see timeout on CI and somehow it hangs now :/

crosbymichael:
@LK4D4 @mrunalp hum, it looks like having mqueue in test template causes it to lockup for some reason

mrunalp:
@crosbymichael Wonder if it is distro specific. I'll try running on Fedora and report back.

crosbymichael:
@mrunalp could be, i'm having trouble running on my ubuntu system also, but i'm getting a compile error inthe docker container:

```
go build -i -ldflags "-X main.gitCommit=1f7eef6c5b414823c98125ec2b911dfd1466f337" -tags "seccomp" -o runc .
# _/go/src/github.com/opencontainers/runc
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-165825633/000000.o: unrecognized relocation (0x2a) in section `.text'
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status

Makefile:22: recipe for target 'all' failed
make: *** [all] Error 2
Makefile:53: recipe for target 'unittest' failed
make[1]: *** [unittest] Error 2
make[1]: Leaving directory '/home/michael/development/gocode/src/github.com/opencontainers/runc'
Makefile:47: recipe for target 'test' failed
make: *** [test] Error 2
```

WTH?????

mrunalp:
I see these errors on Fedora with or without mqueue. 
```
# (in test file tests/integration/exec.bats, line 22)
not ok 14 runc exec --pid-file
# (in test file tests/integration/exec.bats, line 35)
```

mrunalp:
Also, the tests/compile did work for me on Ubuntu :/ However, one thing to note is that I have a custom kernel running that has a fix for mqueue/userns issue. Which test is hanging? Also, are we using 16.04 on jenkins?

libcontainer/container.go
```diff
@@ -124,21 +120,32 @@ type BaseContainer interface {
 	// ContainerDestroyed - Container no longer exists,
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
-	// Systemerror - System error.
+	// SystemError - System error.
 	Start(process *Process) (err error)
 
+	// StartI immediatly starts the process inside the conatiner.  Returns error if process
+	// fails to start.  It does not block waiting for a SIGCONT after start returns but
```
wking: The `InitContinueSignal` transition in 962febf2 missed this SIGCONT instance.

start.go
```diff
@@ -1,139 +1,32 @@
...
-		if err := factory.StartInitialization(); err != nil {
-			// as the error is sent back to the parent there is no need to log
-			// or write it to stderr because the parent process will handle this
-			os.Exit(1)
+		if status == libcontainer.Created {
+			if err := container.Signal(libcontainer.InitContinueSignal); err != nil {
+				fatal(err)
+			}
 		}
-		panic("libcontainer: container init failed to exec")
```
wking: `panic` is probably too strong, but I expect we want an error here to cover the "that container is in {state}, but you can only start 'Created' containers” case (where opencontainers/runtime-spec#384 is currently [calling for an error][1]).

[1]: https://github.com/opencontainers/runtime-spec/pull/384/files#diff-b84a8d65d8ed53f4794cd2db7e8ea731R86

script/test_Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.5.3
+FROM golang:1.6.2
```
wking: You can role this back with 1f7eef6c5.

crosbymichael: its fine to keep

libcontainer/standard_init_linux.go
crosbymichael: At this time it is required and if we want to change that its a different discussion

wking: On Thu, May 26, 2016 at 01:20:02PM -0700, Michael Crosby wrote:
> > +	// check for the arg before waiting to make sure it exists and it is returned
> > +	// as a create time error
> > +	name, err := exec.LookPath(l.config.Args[0])
>
> At this time it is required and if we want to change that its a
> different discussion

Required as in “that's what runC will do right now” or required as in
“opencontainers/runtime-spec#384 needs to be updated to list this
requirement”.  The distinction there is mostly in how compliance is
tested in ocitools.

crosbymichael:
I updated this with latest changes that were merged into the spec.

create.go
```diff
@@ -0,0 +1,60 @@
...
+			return err
+		}
+		status, err := startContainer(context, spec, true)
+		if err != nil {
+			return err
+		}
+		// exit with the container's exit status so any external supervisor is
+		// notified of the exit with the correct exit status.
+		os.Exit(status)
+		return nil
```
wking: Does a linter need this line?  It looks like dead code to me.

create.go
```diff
@@ -0,0 +1,60 @@
...
+		if err != nil {
+			return err
+		}
+		status, err := startContainer(context, spec, true)
+		if err != nil {
+			return err
+		}
+		// exit with the container's exit status so any external supervisor is
+		// notified of the exit with the correct exit status.
+		os.Exit(status)
```
wking: With the required-detach starting [here][1], I think the comment is a bit misleading.  The exit status is probably better described as “the container-setup exit status” or some such.

[1]: https://github.com/opencontainers/runc/pull/827/files#diff-b483819de81a70256d432f5245ca26feR258

libcontainer/integration/template_test.go
crosbymichael: This is from the issue of mqueue not working on debian kernels on the CI in userns

create.go
crosbymichael: compile err

crosbymichael:
rebased, PTAL

duglin:
@crosbymichael what do you think about having `--console /dev/pts/ptmx` be the default when the config asks for a terminal?  From a UX perspective its kind of annoying having to add that each time

crosbymichael:
Talking with @duglin on IRC, basically create/start/delete is something that you have to write code around and is for higher level tools to orchestrate.  `runc run` is for cmd line use where we can keep the STDIO or TTY open and have a better experience.  There is really no way around this. 

if you want flexibility you have to give up some usability.

duglin:
additionally, as @crosbymichael mentioned on irc, you can always do:
```
runc create test
runc exec test ...    # repeat....
```
to get *almost* the same results as doing `runc start` and connect up the tty to it.

cyphar:
@duglin `--console /dev/pts/ptmx` is not the same and this breaks for user namespaces (I'm working on fixing this in #814). This is why we currently can't run our integration tests for user namespaced containers.

libcontainer/container.go
```diff
@@ -124,21 +120,32 @@ type BaseContainer interface {
...
+	// sends the signal when the process has completed.
+	//
+	// errors:
+	// ContainerDestroyed - Container no longer exists,
+	// ConfigInvalid - config is invalid,
+	// ContainerPaused - Container is paused,
+	// SystemError - System error.
+	StartI(process *Process) (err error)
```
LK4D4: Maybe we should rename it to Run. It also semi-maps to `exec.Cmd` methods.

LK4D4:
I think it needs at least one integration test, which also will be usage example.

crosbymichael:
@LK4D4 there are tests

https://github.com/opencontainers/runc/pull/827/files#diff-ca657af2fedf0135188c8e3924a17c00R1

crosbymichael:
@LK4D4 changed StartI to Run

LK4D4:
LGTM

cyphar:
I'm going to test out this PR later today, please hold off on merging. Thanks.

crosbymichael:
@cyphar thanks!

mrunalp:
Tested. LGTM.

hqhq:
LGTM

cyphar:
One of the biggest issues I can see right off the bat is that `runc create` doesn't work for user namespaced containers because of #814 (you can't set any console path). While I'm *okay* with that being the case (for the moment while I work on fixing #814), we should definitely make it an explicit error rather than "here's some cryptic error message".

/cc @crosbymichael

crosbymichael:
@cyphar that is unrelated to this PR.  

I really don't think you all understand what `--console` does or is for if you are trying to pass it dev/null.  I'll show you want to do in another PR.

cyphar:
@crosbymichael explained on IRC that `--console` doesn't take a master, it takes a slave path and does no setup (which makes sense if you read the code). So the complaints about `--console` are not accurate, and our integration tests usage of `--console` is incorrect.

PR #828, opened 2016-05-17
Fix update kernel memory test

hqhq:
Since kernel 4.6, we can update kernel memory without
initialization, because it's accounted by default.

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -70,5 +70,12 @@ EOF
 
     # update kernel memory limit
     run "$RUNC" update test_cgroups_kmem --kernel-memory 50331648
-    [ ! "$status" -eq 0 ]
+    # Since kernel 4.6, we can update kernel memory without initialization
+    # because it's accounted by default.
+    if [ "$KERNEL_MAJOR" -lt 4 ] || [ "$KERNEL_MAJOR" -eq 4 -a "$KERNEL_MINOR" -le 5 ]; then
+        [ ! "$status" -eq 0 ]
+    else
+        [ "$status" -eq 0 ]
+        check_cgroup_value $CGROUP_MEMORY "memory.kmem.limit_in_bytes" 50331648
+    fi
 }
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -19,6 +19,12 @@ HELLO_BUNDLE="$BATS_TMPDIR/hello-world"
 # CRIU PATH
 CRIU="/usr/local/sbin/criu"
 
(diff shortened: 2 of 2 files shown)


tests/integration/cgroups.bats
```diff
@@ -70,5 +70,11 @@ EOF
 
     # update kernel memory limit
     run "$RUNC" update test_cgroups_kmem --kernel-memory 50331648
-    [ ! "$status" -eq 0 ]
+    # Since kernel 4.6, we can update kernel memory without initialization
+    # because it's accounted by default.
+    if [ "$KERNEL_MAJOR" -lt 4 ] || [ "$KERNEL_MAJOR" -eq 4 -a "$KERNEL_MINOR" -le 5 ]; then
+        [ ! "$status" -eq 0 ]
+    else
+        [ "$status" -eq 0 ]
```
cyphar: Can we also check that the limit was set properly? And personally I'd prefer to make it have two pass cases:

1. It fails. (we can check that the kernel version is right here).
2. It succeeds and the limit is set properly.

Also, the kernel version check looks wrong. `4.3.3` would fail to be considered as an "old" kernel.

hqhq: Isn't `4.3.3` failed as expected?

In my research, the behavior change because of b313aeee2509 ("mm: memcontrol: enable kmem accounting for all cgroups in the legacy hierarchy"), it is in `v4.5-2559-gb313aee`, and I tested on 4.5 and 4.4.3, they all failed as expected. So kernel version older than 4.5 (including) would fail.

(here `fail` I mean update kernel memory without initialization, or do you mean the test case failed on `4.3.3` in your box?)

hqhq: Added value check.

cyphar: I mean that the case:

```
if [ "$KERNEL_MAJOR" -lt 4 ] || [ "$KERNEL_MAJOR" -eq 4 -a "$KERNEL_MINOR" -le 5 ]; then
```

Would be considered `false` with the kernel version `4.3.3`, even though it is older than `4.4.5`. Is that intentional (it doesn't appear to be).

hqhq: This if statement will return **true** with kernel version `4.3.3`, no?

cyphar: Oh, sorry, I thought the second part read `[ "$KERNEL_MINOR" -eq 4 -a "$KERNEL_PATCH" -le 5 ]`. :P

cyphar:
LGTM.

mrunalp:
LGTM

PR #829, opened 2016-05-18
Memory swap comment update

rajasec:
Currently when we are trying to update the running container with swap as -1
invalid value for memory-swap: strconv.ParseUint: parsing "-1": invalid syntax

Moreover in memory.go check is added to have
cgroup.Resources.MemorySwap > 0.

Eliminated the comment for swap to have -1

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -132,6 +132,20 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 			}
 		}
 	}
+	if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap == -1 {
+		if err := writeFile(path, "memory.limit_in_bytes", strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
+			return err
+		}
+		if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
+			return err
+		}
+	} else {
+		if cgroup.Resources.MemorySwap == -1 {
+			if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
+				return err
+			}
+		}
+	}
 
 	return nil
 }
--- a/update.go
+++ b/update.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 
 	"github.com/codegangsta/cli"
-	"github.com/docker/go-units"
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
@@ -125,7 +124,7 @@ other options are ignored.
 		}
 
 		config := container.Config()
-
+		configPtr := &config
 		if in := context.String("resources"); in != "" {
 			var (
 				f   *os.File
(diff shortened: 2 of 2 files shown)


hqhq:
I think we should implement `runc update --memory-swap -1 <container-id>` (as well as others which can be unlimited by `echo -1` to cgroups limit files) other than change comments.

cyphar:
@hqhq I'd prefer if you could use a string argument like `unlimited` or `max`, rather than `-1` (which isn't as descriptive as one would like).

rajasec:
@hqhq @cyphar 
it is not only for update, even if you start the container ( -1 as argument), we will get unmarshal error

cyphar:
@rajasec I think we should also enable setting `unlimited` and similar flags for `start` as well.

hqhq:
`-1` is reflection of cgroups api, we should not use it as `unlimited` for all entries, only those actually take `-1` as a limitation reset.

cyphar:
@hqhq I know that _some_ of the cgroup API uses `-1`, but I think the UX should look like:

```
% runc update --memory unlimited
```

hqhq:
I'm OK with either.

rajasec:
I'll do the required changes to support unlimited swap and re-commit

rajasec:
@hqhq @cyphar 
With this PR, I have made the changes to support unlimited swap via update command.

Update command supports via json and stdin support
currently it support only for stdin, because if we want to support via json we need to change the spec uint64 for memory-swap type. So for any running container with this we can change the unlimited memory swap with this change.
Please let me know your comments.

rajasec:
@hqhq @cyphar 
With this change , we can have the update command to -1 as option for memory-swap
runc update --memory-swap -1 test

update.go
```diff
@@ -184,8 +183,15 @@ other options are ignored.
...
-		config.Cgroups.Resources.MemorySwap = int64(*r.Memory.Swap)
-
+		if val := context.String("memory-swap"); val != "" {
+			swap, err := strconv.ParseInt(val, 10, 64)
+			if err != nil {
+				return err
+			}
+			config.Cgroups.Resources.MemorySwap = swap
+		} else {
+			config.Cgroups.Resources.MemorySwap = int64(*r.Memory.Swap)
```
LK4D4: Seems like you need to add default value to `r` initialization above

LK4D4:
@rajasec maybe you forgot to push some changes because I don't see support for unlimited value :/

rajasec:
@LK4D4 
The unlimited changes ( -1) were in memory.go, If the user specifies the value as -1, it will take the unlimited value where it is highest memory value and sets in memory_limit_in_bytes.

rajasec:
@hqhq @cyphar @LK4D4 
Do I need to make any further changes ?

update.go
```diff
@@ -184,8 +183,15 @@ other options are ignored.
 		config.Cgroups.Resources.KernelMemoryTCP = int64(*r.Memory.KernelTCP)
 		config.Cgroups.Resources.Memory = int64(*r.Memory.Limit)
 		config.Cgroups.Resources.MemoryReservation = int64(*r.Memory.Reservation)
-		config.Cgroups.Resources.MemorySwap = int64(*r.Memory.Swap)
```
cyphar: Since we're using `int64` here anyway, why don't we just switch the parsing stage to `ParseInt` as well? Or make it the backup?

libcontainer/cgroups/fs/memory.go
```diff
@@ -141,6 +141,20 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
...
+	if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap == -1 {
+		if err := writeFile(path, "memory.limit_in_bytes", strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
+			return err
+		}
+		if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
+			return err
+		}
+	} else {
+		if cgroup.Resources.MemorySwap == -1 {
+			if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
```
cyphar: Why are you doing this twice? Just take this out of both if branches and add it after checking that `cgroup.Resources.Memory != 0`.

update.go
rajasec: @cyphar @hqhq 
I have made the changes to int64 and made it consistent for all memory operations. Let me know if you are OK with the changes.

cyphar:
Needs a rebase. Is this still necessary?

rajasec:
I will rebase

rajasec:
with the latest runc changes, intent of this changes were addressed. We can close this PR.
Thanks @cyphar @hqhq for the review.

PR #835, opened 2016-05-20
Fix outdated comment for loadSpec

zhaoleidd:
loadSpec() is not support blank path now, remove relative comments.


hqhq:
LGTM

PR #836, opened 2016-05-22
*: correctly chown() consoles

cyphar:
In user namespaces, we need to make sure we don't chown() the console to
unmapped users. This means we need to get both the UID and GID of the
root user in the container when changing the owner.

This is part of the cleanup I did for rootless containers, but benefits the
project as a whole so we might as well apply it now. Obviously we still
need to fix the user namespace interaction (#814).

/cc @crosbymichael


LK4D4:
LGTM

hqhq:
LGTM

PR #843, opened 2016-05-25
Update manuals

zhaoleidd:
This patchset updated manual and code for --help, mainly to:
1: To make manual and --help output same information.
2: To make manual in same format.
3: Add losted arguments into manual.


update.go
```diff
@@ -33,6 +35,7 @@ The accepted format is as follow (unchanged values can be omitted):
     "reservation": 0,
     "swap": 0,
     "kernel": 0
```
cyphar: Please add a trailing comma here.

man/runc-update.8.md
```diff
@@ -5,42 +5,41 @@
...
+       "cpu": {
+           "shares": 0,
+           "quota": 0,
+           "period": 0,
+           "cpus": "",
+           "mems": ""
+       },
+       "blockIO": {
+           "blkioWeight": 0
+       },
```
cyphar: I'm not sure why you moved this to options. I preferred it to be in description TBH.

update.go
zhaoleidd: done, patch updated.

man/runc-update.8.md
zhaoleidd: done, patch updated.
Thanks for review.

man/runc-list.8.md
```diff
@@ -4,12 +4,9 @@
...
-   The default format is table.  The following will output the list of containers
-in json format:
-
-    # runc list -f json
-
 # OPTIONS
    --format, -f         select one of: table or json.
+                        The default format is table. The following will output
+                        the list of containers in json format:
+                        # runc list -f json
```
cyphar: This and `runc-ps.8.md` are inconsistent. I think that we should remove the "the following will ..." part because it's not necessary (the information is already there). Just make it clear that `table` is the default. And do the same for `runc ps`.

zhaoleidd: And maybe we also need to remove the example in code's --help option, as in corrent code:

# ./runc list --help
...
OPTIONS:
   --format, -f         select one of: table or json.

The default format is table.  The following will output the list of containers
in json format:

    # runc list -f json
   --quiet, -q  display only container IDs
#  

What is your opinion?

cyphar: I don't personally like it, but let's see what @hqhq thinks.

hqhq: I think we should keep help message simple and clear (one line per option would be good), elaborations can landed in man pages.

zhaoleidd: Got it.
Patch updated for list and ps command, as:
```
# ./runc list --help
NAME:
   runc list - lists containers started by runc with the given root

USAGE:
   runc list [command options] [arguments...]

OPTIONS:
   --format value, -f value     select one of: table(default) or json
   --quiet, -q                  display only container IDs
# 
```

LK4D4:
LGTM
ping @cyphar @hqhq

hqhq:
LGTM

PR #707, opened 2016-03-29
Add support for copying up directories into tmpfs when a tmpfs is mounted over them

mrunalp:
This helps in supporting read-only containers.
For e.g. if one specifies a tmpfs at /etc then the contents of /etc are copied up into the tmpfs and discarded on container exit.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

PR #845, opened 2016-05-26
Add support for copying up directories into tmpfs when a tmpfs is mounted over them

mrunalp:
This replaces #707 by simply copying over a directory instead of using tar.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/docker/docker/pkg/mount"
 	"github.com/docker/docker/pkg/symlink"
+	"github.com/mrunalp/fileutils"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/label"
@@ -152,14 +153,40 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 		}
 		return nil
 	case "tmpfs":
+		copyUp := m.Extensions&configs.EXT_COPYUP == configs.EXT_COPYUP
+		tmpDir := ""
 		stat, err := os.Stat(dest)
 		if err != nil {
 			if err := os.MkdirAll(dest, 0755); err != nil {
 				return err
 			}
 		}
+		if copyUp {
+			tmpDir, err = ioutil.TempDir("/tmp", "runctmpdir")
+			if err != nil {
+				return newSystemErrorWithCause(err, "tmpcopyup: failed to create tmpdir")
+			}
+			defer os.RemoveAll(tmpDir)
+			m.Destination = tmpDir
+		}
 		if err := mountPropagate(m, rootfs, mountLabel); err != nil {
 			return err
+		}
+		if copyUp {
+			if err := fileutils.CopyDirectory(dest, tmpDir); err != nil {
+				errMsg := fmt.Errorf("tmpcopyup: failed to copy %s to %s: %v", dest, tmpDir, err)
+				if err1 := syscall.Unmount(tmpDir, syscall.MNT_DETACH); err1 != nil {
+					return newSystemErrorWithCausef(err1, "tmpcopyup: %v: failed to unmount", errMsg)
+				}
+				return errMsg
+			}
+			if err := syscall.Mount(tmpDir, dest, "", syscall.MS_MOVE, ""); err != nil {
+				errMsg := fmt.Errorf("tmpcopyup: failed to move mount %s to %s: %v", tmpDir, dest, err)
+				if err1 := syscall.Unmount(tmpDir, syscall.MNT_DETACH); err1 != nil {
+					return newSystemErrorWithCausef(err1, "tmpcopyup: %v: failed to unmount", errMsg)
+				}
+				return errMsg
+			}
 		}
 		if stat != nil {
 			if err = os.Chmod(dest, stat.Mode()); err != nil {
@@ -711,7 +738,9 @@ func mountPropagate(m *configs.Mount, rootfs string, mountLabel string) error {
 	if libcontainerUtils.CleanPath(dest) == "/dev" {
 		flags &= ^syscall.MS_RDONLY
 	}
-	if !strings.HasPrefix(dest, rootfs) {
+
+	copyUp := m.Extensions&configs.EXT_COPYUP == configs.EXT_COPYUP
+	if !(copyUp || strings.HasPrefix(dest, rootfs)) {
 		dest = filepath.Join(rootfs, dest)
 	}
 
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -234,7 +234,7 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 }
 
 func createLibcontainerMount(cwd string, m specs.Mount) *configs.Mount {
-	flags, pgflags, data := parseMountOptions(m.Options)
+	flags, pgflags, data, ext := parseMountOptions(m.Options)
 	source := m.Source
 	if m.Type == "bind" {
 		if !filepath.IsAbs(source) {
@@ -248,6 +248,7 @@ func createLibcontainerMount(cwd string, m specs.Mount) *configs.Mount {
 		Data:             data,
 		Flags:            flags,
 		PropagationFlags: pgflags,
+		Extensions:       ext,
 	}
 }
 
@@ -592,11 +593,12 @@ func setupUserNamespace(spec *specs.Spec, config *configs.Config) error {
 
 // parseMountOptions parses the string and returns the flags, propagation
 // flags and any mount data that it contains.
-func parseMountOptions(options []string) (int, []int, string) {
+func parseMountOptions(options []string) (int, []int, string, int) {
 	var (
-		flag   int
-		pgflag []int
-		data   []string
+		flag     int
+		pgflag   []int
+		data     []string
+		extFlags int
 	)
 	flags := map[string]struct {
 		clear bool
@@ -638,6 +640,12 @@ func parseMountOptions(options []string) (int, []int, string) {
 		"rslave":      syscall.MS_SLAVE | syscall.MS_REC,
 		"runbindable": syscall.MS_UNBINDABLE | syscall.MS_REC,
 	}
+	extensionFlags := map[string]struct {
+		clear bool
+		flag  int
+	}{
+		"tmpcopyup": {false, configs.EXT_COPYUP},
+	}
 	for _, o := range options {
 		// If the option does not exist in the flags table or the flag
 		// is not supported on the platform,
@@ -650,11 +658,17 @@ func parseMountOptions(options []string) (int, []int, string) {
 			}
 		} else if f, exists := propagationFlags[o]; exists && f != 0 {
 			pgflag = append(pgflag, f)
+		} else if f, exists := extensionFlags[o]; exists && f.flag != 0 {
+			if f.clear {
+				extFlags &= ^f.flag
+			} else {
+				extFlags |= f.flag
+			}
 		} else {
 			data = append(data, o)
 		}
 	}
-	return flag, pgflag, strings.Join(data, ",")
+	return flag, pgflag, strings.Join(data, ","), extFlags
 }
 
 func setupSeccomp(config *specs.Seccomp) (*configs.Seccomp, error) {
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -73,6 +73,10 @@
 		{
 			"ImportPath": "github.com/vishvananda/netlink",
 			"Rev": "1e2e08e8a2dcdacaae3f14ac44c5cfa31361f270"
+		},
+		{
+			"ImportPath": "github.com/mrunalp/fileutils",
+			"Rev": "ed869b029674c0e9ce4c0dfa781405c2d9946d08"
 		}
 	]
 }
--- a/Godeps/_workspace/src/github.com/mrunalp/fileutils/LICENSE
+++ b/Godeps/_workspace/src/github.com/mrunalp/fileutils/LICENSE
@@ -1,0 +1,191 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+   4. Redistribution. You may reproduce and distribute copies of the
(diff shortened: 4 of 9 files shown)


libcontainer/specconv/spec_linux.go
```diff
@@ -635,11 +643,17 @@ func parseMountOptions(options []string) (int, []int, string) {
...
 		} else if f, exists := propagationFlags[o]; exists && f.flag != 0 {
 			pgflag = append(pgflag, f.flag)
+		} else if f, exists := extensionFlags[o]; exists && f.flag != 0 {
+			if f.clear {
+				extFlags &= ^f.flag
+			} else {
+				extFlags |= f.flag
+			}
 		} else {
 			data = append(data, o)
```
crosbymichael: Are we sure there are no mount data or fstypes that take a `copyup` option?

mrunalp: I found one here http://linux.die.net/man/1/funionfs.
WDYT about calling it runc_copyup?

crosbymichael: is there something shorter?  Maybe `runccopyup` `rcopyup` `rccopyup`

mrunalp: I like `tmpcopyup`. I'll push updated code.

mrunalp:
Pushed update to parse it as `tmpcopyup`

libcontainer/rootfs_linux.go
```diff
@@ -152,12 +153,24 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
...
+		copyUp := m.Extensions&configs.EXT_COPYUP == configs.EXT_COPYUP
+		tmpDir := ""
 		stat, err := os.Stat(dest)
 		if err != nil {
 			if err := os.MkdirAll(dest, 0755); err != nil {
 				return err
 			}
 		}
+		if copyUp {
+			tmpDir, err = ioutil.TempDir("/run", "runctmpdir")
```
crosbymichael: Why are you using /run here? Shouldn't this be under the machines configured temp dir and not fillup space on /run?

cyphar: Why don't we do it like this:

1. Create a tmpfs somewhere (`/tmp/runc.XXXXX`).
2. Copy the contents of the directory we're going to dump the `tmpfs` on top of.
3. Bind-mount it over the directory.
4. Lazy unmount and remove the initial temporary directory.

That way, we don't have to do any double-copying.

cyphar:
@mrunalp 

Why don't we do it like this:

1. Create a tmpfs somewhere (`/tmp/runc.XXXXX`).
2. Copy the contents of the directory into that `tmpfs`.
3. Bind-mount the `tmpfs` over the directory.
4. Lazy unmount and remove the initial temporary directory.

That way, we don't have to do any double-copying.

libcontainer/specconv/spec_linux.go
cyphar: > There is no implementation of -o delete and -o copyup for now. 

:wink:

But I think we should have a nicer way of doing this than adding to a flag list that might eventually clash with the kernel. Maybe we should add an `extension` field to `mounts`? Or would that be overkill?

cyphar:
Also (and I know this sounds like a trivial thing), but please put your `fileutils` project under a free software license.

rhatdan:
The tmpdir should probably look for environment variables  Something like checking for the existence of
$XDG_RUNTIME_DIR, and then fall back to /tmp.  I always worry about creating directories in /tmp since other users could potentially read/write the content in this directory.  But since we want runc to  be able to run without being UID=0, we don't know for sure if it can write to /run.

cyphar:
@rhatdan 

> I always worry about creating directories in /tmp since other users could potentially read/write the content in this directory.

This isn't true if we decide to change the mode to something like `0700` (which I would actually recommend).

rhatdan:
Yes, also if this is in a different mount namespace it could be hidden from other users.

libcontainer/specconv/spec_linux.go
mrunalp: @cyphar I think we can revisit when we have more extensions :wink:

libcontainer/rootfs_linux.go
mrunalp: I'll move it to /tmp

mrunalp:
@cyphar With your suggested approach, I suspect that tmpfs won't be charged to cgroup of the container if bind mounted (I'll take a look anyway). Though the double copy take more time, it doesn't change the semantics of how --tmpfs works today.

cyphar:
@mrunalp We could use `MS_MOVE` instead if you want. But since the `tmpfs` was created after cgroup join, I'm fairly sure `kmemcg` will track it properly.

mrunalp:
@cyphar Yeah, MS_MOVE should work. I'll try that.

mrunalp:
I pushed a new commit (I'll clean up the PR after review) that uses MS_MOVE. Also, added code for using XDG_RUNTIME_DIR as prefix if it is set in the env. PTAL.

mrunalp:
btw, pullapprove is broke ` Invalid .pullapprove.yml, check pullapprove.com...`

crosbymichael:
@mrunalp do you need to rebase?

mrunalp:
Rebased.

libcontainer/rootfs_linux.go
```diff
@@ -152,14 +153,46 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
...
 		stat, err := os.Stat(dest)
 		if err != nil {
 			if err := os.MkdirAll(dest, 0755); err != nil {
 				return err
 			}
 		}
-		if err := mountPropagate(m, rootfs, mountLabel); err != nil {
-			return err
+		if copyUp {
+			tmpDirPrefix := os.Getenv("XDG_RUNTIME_DIR")
```
crosbymichael: I don't see how this would ever be set inside the container's init.

mrunalp: Yeah we can't leak it from outside with this code so was thinking whether setting in the config might be an acceptable compromise. Else we can special case leak it.

crosbymichael: Just use the container's /tmp dir or some hidden dir inside the rootfs to make this.  Since its temporary and is only there to make the initial tmpfs destination it can be anywhere or even back in your container's /run.

mrunalp: Want to make this work with read only rootfs that we can't write to at all.

crosbymichael: We don't set it to readonly until after the mounts and after we pivot.  Also you can just use `/dev/` or `/run` to do this inside the container.

mrunalp: I meant read only even before we make it read only and I know we will need more changes to get there. 

Could possibly use some other tmpfs mount if specified in the config before this mount.

rhatdan: Probably safe enough to just use /tmp with 0700 and a random directory.

mrunalp: I removed `XDG_RUNTIME_DIR` lookup and switched to using /tmp which works well.

mrunalp:
@cyphar @crosbymichael PTAL

cyphar:
@mrunalp It looks like everything is evaluated in the **host** user namespace, so the file permissions look wrong and will cause issues with certain namespace setups (I've set `/etc` to be `tmpcopyup`):

```
/ # ls /etc/ -la
total 72
drwxr-xr-x    6 root     root           480 Jun 10 08:44 .
drwxr-xr-x    1 1000     users          206 Jun 10 08:44 ..
-rw-r--r--    1 1000     users          466 Jun 10 08:44 fstab
-rw-r--r--    1 1000     users          298 Jun 10 08:44 group
-rw-r--r--    1 1000     users           10 Jun 10 08:44 hostname
-rw-r--r--    1 1000     users           40 Jun 10 08:44 hosts
drwxr-xr-x    2 root     root           140 Jun 10 08:44 init.d
-rw-r--r--    1 1000     users         1053 Jun 10 08:44 inittab
-rw-r--r--    1 1000     users         1180 Jun 10 08:44 inputrc
drwxr-xr-x    2 root     root           180 Jun 10 08:44 iproute2
-rw-r--r--    1 1000     users           21 Jun 10 08:44 issue
-rw-r--r--    1 1000     users            0 Jun 10 08:44 ld.so.conf
drwxr-xr-x    2 root     root            40 Jun 10 08:44 ld.so.conf.d
lrwxrwxrwx    1 1000     users           12 Jun 10 08:44 mtab -> /proc/mounts
drwxr-xr-x    8 root     root           180 Jun 10 08:44 network
-rw-r--r--    1 1000     users           95 Jun 10 08:44 os-release
-rw-r--r--    1 1000     users          371 Jun 10 08:44 passwd
-rw-r--r--    1 1000     users         1388 Jun 10 08:44 profile
-rw-r--r--    1 1000     users         2744 Jun 10 08:44 protocols
-rw-------    1 1000     users          512 Jun 10 08:44 random-seed
-rw-r--r--    1 1000     users          833 Jun 10 08:44 resolv.conf
-rw-r--r--    1 1000     users          386 Jun 10 08:44 securetty
-rw-r--r--    1 1000     users        10873 Jun 10 08:44 services
-rw-------    1 1000     users          239 Jun 10 08:44 shadow
```

cyphar:
@mrunalp There's a bug in `fileutils`, with the owner of directories (I've set `/etc/` to be `tmpcopyup` and everything is owned by my user `1000:1000`):

```
/ # ls /etc/ -la
total 72
drwxr-xr-x    6 root     root           480 Jun 10 08:44 .
drwxr-xr-x    1 1000     users          206 Jun 10 08:44 ..
-rw-r--r--    1 1000     users          466 Jun 10 08:44 fstab
-rw-r--r--    1 1000     users          298 Jun 10 08:44 group
-rw-r--r--    1 1000     users           10 Jun 10 08:44 hostname
-rw-r--r--    1 1000     users           40 Jun 10 08:44 hosts
drwxr-xr-x    2 root     root           140 Jun 10 08:44 init.d
-rw-r--r--    1 1000     users         1053 Jun 10 08:44 inittab
-rw-r--r--    1 1000     users         1180 Jun 10 08:44 inputrc
drwxr-xr-x    2 root     root           180 Jun 10 08:44 iproute2
-rw-r--r--    1 1000     users           21 Jun 10 08:44 issue
-rw-r--r--    1 1000     users            0 Jun 10 08:44 ld.so.conf
drwxr-xr-x    2 root     root            40 Jun 10 08:44 ld.so.conf.d
lrwxrwxrwx    1 1000     users           12 Jun 10 08:44 mtab -> /proc/mounts
drwxr-xr-x    8 root     root           180 Jun 10 08:44 network
-rw-r--r--    1 1000     users           95 Jun 10 08:44 os-release
-rw-r--r--    1 1000     users          371 Jun 10 08:44 passwd
-rw-r--r--    1 1000     users         1388 Jun 10 08:44 profile
-rw-r--r--    1 1000     users         2744 Jun 10 08:44 protocols
-rw-------    1 1000     users          512 Jun 10 08:44 random-seed
-rw-r--r--    1 1000     users          833 Jun 10 08:44 resolv.conf
-rw-r--r--    1 1000     users          386 Jun 10 08:44 securetty
-rw-r--r--    1 1000     users        10873 Jun 10 08:44 services
-rw-------    1 1000     users          239 Jun 10 08:44 shadow
```

The directories are owned by `root` because `CopyDirectory` doesn't `Lchown` directories. I'll open a PR against `fileutils` to fix this.

cyphar:
@mrunalp Since mrunalp/fileutils#1 was merged can you rebase with the vendor updated?

mrunalp:
@cyphar Updated.

mrunalp:
@opencontainers/runc-maintainers PTAL. All comments have been addressed.

rhatdan:
I really want to start pushing the concept of readonly containers, and this patch is critical to this.  Being able to mount tmpfs over certain directories allows us to preserve  the layout of a container image and still run the container in read-only mode.

cyphar:
I will take a look at this next week.

rhatdan:
Any movement on this.  I think we have been asking for this for over a year now...

libcontainer/rootfs_linux.go
```diff
@@ -152,14 +153,46 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
...
+				errMsg := fmt.Errorf("failed to move mount %s to %s: %v", tmpDir, dest, err)
+				if err1 := syscall.Unmount(tmpDir, syscall.MNT_DETACH); err1 != nil {
+					return fmt.Errorf("%v: failed to unmount: %v", errMsg, err1)
+				}
+				return errMsg
+			}
+		} else {
+			if err := mountPropagate(m, rootfs, mountLabel); err != nil {
+				return err
+			}
```
cyphar: I'd prefer we did something like this, because it would make it more clear what the difference with and without `copyUp` is:

```
if copyUp {
    // change m.Destination to a tmpdir
}

if err := mountPropagate(m, rootfs, mountLabel); err != nil {
    return err
}

if copyUp {
    // copy everything
}
```

libcontainer/rootfs_linux.go
LK4D4: @mrunalp I don't quite understand where we are here? You wrote that you removed lookup, but it's still here.

cyphar:
@rhatdan 

> I really want to start pushing the concept of readonly containers, and this patch is critical to this. Being able to mount tmpfs over certain directories allows us to preserve the layout of a container image and still run the container in read-only mode. 

To be clear, this is *possible* using `overlayfs` (with `upperdir` set to a `tmpfs` mount). It's not pretty but there are other ways than this method (and are arguably better than this because this does an expensive copy-up for **everything** (even files you don't ever modify)).

libcontainer/rootfs_linux.go
```diff
@@ -152,14 +153,46 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 		}
 		return nil
 	case "tmpfs":
+		copyUp := m.Extensions&configs.EXT_COPYUP == configs.EXT_COPYUP
+		tmpDir := ""
```
cyphar: Not sure why `tmpDir` is global? is it because of the `defer`? I'd personally prefer if we put it in a closure, but that's just my 2c.

LK4D4:
@mrunalp @rhatdan yeah, I agree with @cyphar about overlay
However, this patch looks ok to me.
@mrunalp On an un-relevant note - I believe that your package violates Apache 2 license of docker. Look for example this paragraph https://github.com/docker/docker/blob/master/LICENSE#L107
Also, your package uses part of `pkg/idtools` which should be specified, too.

rhatdan:
Not sure what Overlay has to do with this.  The basic idea is to run a read-only container with 

docker run -d --tmpfs /run --tmpfs /var/tmp --tmpfs /tmp --read-only httpd

httpd expects to be able to write a pid file to /run/httpd, which is usually on the underlying /run.  Without this patch, --read-only flag is useless to a large section of containers on docker.io.

OverlayFS is for writing read/writable containers.  For a security point of view,  containers should be blocked from writing to /usr when they are in production, or to anywhere on the base image, unless volumes are mounted into the container.

mrunalp:
I will add that to the list. Thanks!

cyphar:
@rhatdan 

> OverlayFS is for writing read/writable containers.

I'm not talking about the overlayfs driver (I'm not actually talking about Docker here, I'm talking about runC specifically). I'm saying that you store the rootfs you want on a readonly partition, then you set up the rootfs **that you mount** to have `lowerdir=/your/rootfs,upperdir=/tmp/container`. All writes will go to `tmpfs`. This is functionally **identical** to this PR (though not as nicely integrated since you have to mess with the other mounts) and doesn't incur the overhead of doing a copy-up of **everything**. Inside Docker this will also be a slight pain, but manageable IMO.

My reason for mentioning this is that it is not correct to say that this patch is "vital" for read-only containers. It just makes the setup much simpler (which **is** a benefit, and IMO this patch should be merged).

rhatdan:
Ok thanks for the explanation.  I did not understand where OverlayFS was coming into this.

libcontainer/rootfs_linux.go
mrunalp: @cyphar Maybe I am missing a trick. Could you paste a snippet to show what you mean? I need to access tmpDir in both the `copyUp` blocks below now.

mrunalp:
Pushed update @LK4D4 @cyphar PTAL.

libcontainer/rootfs_linux.go
cyphar: Well, you need it **now**. But you can use `m.Destination` in the second `copyUp` block. Doesn't really matter though it  was just a thought.

cyphar:
Dammit. GitHub doesn't let me retract the review with an "abstain". I'll LGTM this explicitly in the comments because clearly GitLab is still beating GitHub in the "reviewing code changes" space.

cyphar:
My main concern with this mode is the foot-gunning is quite easy. For example, if you put a mount for `/` with `tmpcopyup` after you've mounted `/proc` you get a lovely error like this:

```
container_linux.go:247: starting container process caused "process_linux.go:359: container init caused \"rootfs_linux.go:54: mounting \\\"tmpfs\\\" to rootfs \\\"/home/cyphar/src/runc/rootfs\\\" at \\\"/tmp/runctmpdir171994909\\\" caused \\\"failed to copy /home/cyphar/src/runc/rootfs to /tmp/runctmpdir171994909: read /home/cyphar/src/runc/rootfs/proc/1/attr/exec: invalid argument\\\"\""
```

Now, while it's true that you shouldn't be foot-gunning like this -- is there a nicer way we can handle these errors? Just asking. I'm also currently testing user namespace interactions and will probably test with rootless containers.

cyphar:
As expected, this doesn't play nicely with user namespace setups where there are unmapped UIDs in the thing you're copying over. Unfortunately, the only /nice/ way of handling this would be to do something exceptionally dodgy like: create all of the mounts until you hit a `tmpcopyup` then request the calling runC process to run a privileged process inside the mount namespace that does the copyup operation which is then `MS_MOVE`d. Since that is not going to pretty code by **any** stretch of the imagination, we should come up with some nice error message for things like this so that users don't spam us with "why doesn't this work".

This issue will be come quite pertinent with rootless containers.

mrunalp:
Added an integration test.

cyphar:
+1 on the integration test. While I'm concerned about the fact that this would break in *some* user namespaced container setups, there's not much we can do without some pretty dodgy modify-the-mount-namespace-from-another-process-that-is-actually-root shennanigans. So I'm fine with the current limitations, but would like that the error message we get is slightly nicer -- maybe just use a `genericError` with a cause mentioning `tmpcopyup`?

mrunalp:
@cyphar Updated error messages.

mrunalp:
@opencontainers/runc-maintainers PTAL. All comments have been addressed.

cyphar:
LGTM

LK4D4:
LGTM

rhatdan:
WooHoo.  Took a long time to finally get this functionality into Docker, but it is finally there.

bchallenor:
Is this functionality accessible from Docker? @rhatdan had a [Moby PR](https://github.com/moby/moby/pull/20501) but it was not merged, and the comments lead here...

At the moment, `docker --tmpfs` seems to hide the underlying contents:

```
$ docker --version
Docker version 18.01.0-ce, build 03596f51b1

$ docker run --tmpfs /run debian:stable find /run
/run

$ docker run debian:stable find /run
/run
/run/lock
/run/mount
/run/mount/utab
/run/utmp
```

cyphar:
@bchallenor I haven't tried this, but if you can specify `tmpcopyup` as one of the mount options with `--tmpfs` then it should trigger this functionality (as Docker just passes down the mount options to runc).

bchallenor:
Any idea what the syntax would be for that? The only options that the [Docker docs](https://docs.docker.com/storage/tmpfs/#specify-tmpfs-options) mention are `tmpfs-size` and `tmpfs-mode`.

rhatdan:
I believe it should be the default.  I am saddened to see that this behaviour does not work with k8s yet either.  We are missing an opportunity to run containres as Read-Only mode.  And this feature makes it easier.  I believe the default for Kubernetes should be to run container in read-only mode.  But this issue is containers tend to need to write to /run, /tmp, and /var/tmp.  If we could setup container runtimes to mount tmpfs on these directories if the image is running in readonly mode by default, then most containers would run, except for the fact that sometimes the underlying directories contain content that is needed for the container.  httpd for example usually ships with a /run/httpd directory and expects to write to this directory.  If you just mount a tmpfs over /run and don't copy up, the httpd will fail, because it does not create the /run/httpd directory if it does not exist.

bchallenor:
By "I believe it should be the default" do you mean that you think it should be the default, or that it already is? If the former, I agree, as I am also motivated by wanting a readonly rootfs with a few read/write holes punched in it.

You mentioned above "getting this functionality into Docker" - can it actually be used via the Docker UI, or is it confined to runc right now?

rhatdan:
I am not sure if this ever got fully merged into upstream docker.  Simple enough to check

docker run -ti --tmpfs /etc fedora ls /etc

cyphar:
@bchallenor If you use the `--mount` flag you can specify the mount options explicitly.

rhatdan:
I believe it should be the default.  I am saddened to see that this behaviour does not work with k8s yet either.  We are missing an opportunity to run containres as Read-Only mode.  And this feature makes it easier.  I believe the default for Kubernetes should be to run container in read-only mode.  But this issue is containers tend to need to write to /run, /tmp, and /var/tmp.  If we could setup container runtimes to mount tmpfs on these directories if the image is running in readonly mode by default, then most containers would run, except for the fact that sometimes the underlying directories contain content that is needed for the container.  httpd for example usually ships with a /run/httpd directory and expects to write to this directory.  If you just mount a tmpfs over /run and don't copy up, the httpd will fail, because it does not create the /run/httpd directory if it does not exist.

joshuaks:
So far I haven't been able to get this working. Does anyone currently have this working or know what options I need to give the mount declaration?

== DETAILS ==


```
# docker info
Docker version 20.10.3, build b455053
Docker Compose version v2.3.3
```


```
# container / docker-compose info
  radarr:
    image: lscr.io/linuxserver/radarr:latest
    read_only: true
    cap_drop:
      - all
    cap_add:
      ...
    environment:
      - S6_READ_ONLY_ROOT=1
      ...
    volumes:
      ...
    tmpfs:
      ...
      - /app
      ...
    expose:
      - 7878
```


I tried each of the following options **separately** with no luck. I scraped these mount options from comments in the post and through review the code.

```
# docker-compose tmpfs modes
      tmpfs:
        - /app:tmpcopyup
        - /app:runctmpdir
        - /app:runccopyup 
        - /app:runc_copyup
        - /app:rcopyup 
        - /app:rccopyup
```


Regardless of which option I tried I receive the following error:
```
# error
./run: line 3: cd: /app/radarr/bin: No such file or directory
```


[Docker documentation](https://docs.docker.com/storage/tmpfs/#specify-tmpfs-options) on volumes only includes mount options `tmps-size` and `tmps-mode`.

kolyshkin:
@joshuaks I suggest you ask Docker support about it.

PR #847, opened 2016-05-26
Add PullApprove support

caniszczyk:
Enforce 2 LGTMs for pull requests.

PR #848, opened 2016-05-27
pullapprove: use the right team

cyphar:
On GitHub the maintainers of runC are *not* the same as the maintainers
of runtime-spec. Fix this, and use the right team.

Fixes #847.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -4,6 +4,6 @@ reject_regex: ^Rejected
 reset_on_push: true
 reviewers:
   teams:
-  - runtime-spec-maintainers
+    - runc-maintainers
   name: default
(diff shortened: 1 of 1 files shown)


caniszczyk:
Sorry folks, I created a new team call 'runc-mantainers' you can use
https://github.com/orgs/opencontainers/teams/runc-maintainers

cyphar:
Actually, it looks like we need to add a new team just for `runC` maintainence. In addition, we should probably clean up all of the teams we have in `opencontainers` -- it doesn't make sense to group them because each project has its own maintainence.

hqhq:
LGTM

dqminh:
LGTM

PR #851, opened 2016-05-28
Update man pages to refect the latest cli change

hqhq:
The major change is the description of options, change
it as the latest cli help message shows, which specify
a "value" after an option if it takes value, and add
(default: xxx) if the option has a default value.

This also includes some other minor consistency fixes.


crosbymichael:
LGTM

mrunalp:
LGTM

PR #853, opened 2016-05-28
Remove use_hierarchy check when set kernel memory

hqhq:
Kernel memory cannot be set in these circumstances (before kernel 4.6):
1. kernel memory is not initialized, and there are tasks in cgroup
2. kernel memory is not initialized, and use_hierarchy is enabled,
   and there are sub-cgroups

While we don't need to cover case 2 because when we set kernel
memory in runC, it's either:
- in Apply phase when we create the container, and in this case,
  set kernel memory would definitely be valid;
- or in update operation, and in this case, there would be tasks
  in cgroup, we only need to check if kernel memory is initialized
  or not.

Even if we want to check use_hierarchy, we need to check sub-cgroups
as well, but for here, we can just leave it aside.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -76,22 +76,13 @@ func (s *MemoryGroup) SetKernelMemory(path string, cgroup *configs.Cgroup) error
 		}
 
 		if !kmemInitialized {
-			// If hierarchy is set, we can't change the limit
-			usesHierarchy, err := getCgroupParamUint(path, "memory.use_hierarchy")
-			if err != nil {
-				return err
-			}
-			if usesHierarchy != 0 {
-				return fmt.Errorf("cannot initialize kmem.limit_in_bytes if use_hierarchy is already set")
-			}
-
 			// If there's already tasks in the cgroup, we can't change the limit either
 			tasks, err := getCgroupParamString(path, "tasks")
 			if err != nil {
 				return err
 			}
 			if tasks != "" {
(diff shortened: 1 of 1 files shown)


hqhq:
ping @crosbymichael @LK4D4 @mrunalp @cyphar PTAL, this is a bug already impact docker, we can't start containers with kernel memory limit on lots of hosts which have `use_hierarchy` enabled by default.

mlaventure:
In apply phase, we could be joining an existing cgroup with sub-cgroup, and then it'd fail, no?

hqhq:
@mlaventure Do you mean start a container which will in a sub-cgroup of an exist container? Why would it fail?

cyphar:
LGTM. Probably a good example of why we shouldn't be overzealous with preemptively providing errors when the kernel may change when and why errors occur.

mrunalp:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/853/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

mlaventure:
@hqhq I meant if I specify a cgroup-path in the config that has already tasks and sub-cgroup in it :)

hqhq:
@mlaventure Yes in this case, if you also assigned kernel memory option, start the container would fail, which is already covered in our code. 

The scenario we didn't cover is when you tried to join a cgroup whose kernel memory is not initialized and `use_hierarchy=1` and there are no tasks in this cgroup but has sub-cgroups in it. It's a rare case, and since kernel would return error anyway, we can see if people will hit this case and if we need add specific error message in runc.

PR #854, opened 2016-05-29
bash completion step for update command

rajasec:
Adding the bash completion step for update command

### DIFF
--- a/contrib/completions/bash/runc
+++ b/contrib/completions/bash/runc
@@ -635,6 +635,41 @@ _runc_state() {
 		    ;;
 	esac
 }
+_runc_update() {
+	local boolean_options="
+	   --help
+	"
+
+	local options_with_args="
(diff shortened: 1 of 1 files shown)


rajasec:
@mrunalp 
Looks like janky not started the build. I could not get the reason out.

cyphar:
I've restarted the build.

rajasec:
Thanks @cyphar

cyphar:
As for the actual PR, I don't use bash so I can't really test it.

marcosnils:
> I don't use bash so I can't really test it.

curious about what you use? 

Just tested it and it works as described.

cyphar:
@marcosnils The one true shell ZSH. ;)

crosbymichael:
LGTM

mrunalp:
LGTM

PR #856, opened 2016-05-30
Add VERSION file to contain the version info

hqhq:
We need this because we need to get version info out of
go code, eg. build rpm package.


cyphar:
I'm going to be honest. The way we autogenerate this at SUSE is that we have a [few shell scripts](https://github.com/SUSE/package_sync_github_build_service) that parse out the version from the `.go` files. ;)

This is a welcome improvement.

cyphar:
LGTM

mrunalp:
LGTM
@crosbymichael PTAL

crosbymichael:
LGTM

Issue #806, opened 2016-05-06
Executing runc as a SCHED_RR process fails


bgray:
On Linux when trying to start a container if runc itself is in SCHED_RR mode it fails, even it you set a realtime bandwidth in the config.json.

For example my config.json has the following

```
        "resources": {
            "cpu": {
                "realtimeRuntime": 10000,
                "realtimePeriod": 1000000
            }
            ....
        }
```

And I try and execute runc like this

```
chrt -r 6 runc start blah
```

Then I get the following error

```
failed to write 4531 to cgroup.procs: write /sys/fs/cgroup/cpu/blah/cgroup.procs: invalid argument
```

I believe the issue is that the `cpu.rt_runtime_us` cgroup value is not inherited and instead always defaults to 0.  And the code in `runc/libcontainer/cgroups/fs/cpu.go` moves the process into the container before setting the `cpu.rt_runtime_us` value, and therefore because the process has inherited the SCHED_RR policy from runc it can't be moved into the cgroup due to the cgroup having 0 rt bandwidth.

The solution that worked for me was to set the RT scheduling values in the cpu cgroup before moving in the process (like how the `memory.kmem.limit_in_bytes` is set in the memory cgroup before moving in the process).

I've attached a patch of my changes.
[sched-rr-cpu-cgroup-diff.txt](https://github.com/opencontainers/runc/files/252246/sched-rr-cpu-cgroup-diff.txt)

So is there any appetite for this change? or do people think you shouldn't be using SCHED_RR anyway. Happy to create a pull request if wanted.

cyphar:
You can definitely open a PR for this change, if it means `runc` can work in even more setups.

hqhq:
Fixed by #860 .

PR #860, opened 2016-05-31
Set the cpu cgroup RT sched params before joining.

bgray:
This is a fix for issue #806, it applies the `cpu.rt_runtime_us` and `cpu.rt_period_us`
values (if not zero) before joining the cgroup.

This is required if runc itself is running in SCHED_RR mode, as it is not
possible to add a process in SCHED_RR mode to a cgroup which hasn't been
assigned any RT bandwidth. And RT bandwidth is not inherited, each new
cgroup starts with 0 bandwidth.

Signed-off-by: Ben Gray <ben.r.gray@gmail.com>

### DIFF
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -22,9 +22,47 @@ func (s *CpuGroup) Name() string {
 func (s *CpuGroup) Apply(d *cgroupData) error {
 	// We always want to join the cpu group, to allow fair cpu scheduling
 	// on a container basis
-	_, err := d.join("cpu")
+	path, err := d.path("cpu")
 	if err != nil && !cgroups.IsNotFound(err) {
 		return err
+	}
+	return s.ApplyDir(path, d.config, d.pid)
+}
+
+func (s *CpuGroup) ApplyDir(path string, cgroup *configs.Cgroup, pid int) error {
+	// This might happen if we have no cpu cgroup mounted.
+	// Just do nothing and don't fail.
+	if path == "" {
+		return nil
+	}
+	if err := os.MkdirAll(path, 0755); err != nil {
+		return err
+	}
+	// We should set the real-Time group scheduling settings before moving
+	// in the process because if the process is already in SCHED_RR mode
+	// and no RT bandwidth is set, adding it will fail.
+	if err := s.SetRtSched(path, cgroup); err != nil {
+		return err
+	}
+	// because we are not using d.join we need to place the pid into the procs file
+	// unlike the other subsystems
+	if err := writeFile(path, "cgroup.procs", strconv.Itoa(pid)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *CpuGroup) SetRtSched(path string, cgroup *configs.Cgroup) error {
+	if cgroup.Resources.CpuRtPeriod != 0 {
+		if err := writeFile(path, "cpu.rt_period_us", strconv.FormatInt(cgroup.Resources.CpuRtPeriod, 10)); err != nil {
+			return err
+		}
+	}
+	if cgroup.Resources.CpuRtRuntime != 0 {
+		if err := writeFile(path, "cpu.rt_runtime_us", strconv.FormatInt(cgroup.Resources.CpuRtRuntime, 10)); err != nil {
+			return err
+		}
 	}
 	return nil
 }
@@ -45,15 +83,8 @@ func (s *CpuGroup) Set(path string, cgroup *configs.Cgroup) error {
 			return err
 		}
 	}
-	if cgroup.Resources.CpuRtPeriod != 0 {
-		if err := writeFile(path, "cpu.rt_period_us", strconv.FormatInt(cgroup.Resources.CpuRtPeriod, 10)); err != nil {
-			return err
-		}
-	}
-	if cgroup.Resources.CpuRtRuntime != 0 {
-		if err := writeFile(path, "cpu.rt_runtime_us", strconv.FormatInt(cgroup.Resources.CpuRtRuntime, 10)); err != nil {
-			return err
-		}
+	if err := s.SetRtSched(path, cgroup); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


libcontainer/cgroups/fs/cpu.go
```diff
@@ -20,15 +20,55 @@ func (s *CpuGroup) Name() string {
...
+		if err := s.SetRtSched(path, d.config); err != nil {
+			return err
+		}
+	}
+
+	defer func() {
+		if err != nil {
+			os.RemoveAll(path)
+		}
+	}()
```
dqminh: we should not need the defer here as the path will be removed if there's any errors.

dqminh:
Can we also have a test for this ?

Also I think its cleaner to follow the pattern in `cpuset.go` where we have an ensureParent call to prepare the directory.

bgray:
Regarding adding a test, I couldn't see an easy way to implement it without changing the cgroup api (sorry new to the GO lang). The `Apply` and `join` calls always try and get the cgroup path from the mount point rather than `NewCgroupTestUtil`.  But I'll spend a bit more effort and create something like the `ApplyDir` in `cpuset.go`.

On the second point of `ensureParent`, I think the problem here is that we can't just inherit the parent's settings (unless they're -1) as the RT BW/quota is divided among the children.

dqminh:
> Regarding adding a test, I couldn't see an easy way to implement it without changing the cgroup api (sorry new to the GO lang)

I think we should be able to add an integration test that similar to how you invoke runc with `chrt` (https://github.com/opencontainers/runc/tree/master/tests/integration)

> I think the problem here is that we can't just inherit the parent's settings

Sure, i think we can skip the part where we get settings from the parent, that's only applicable to cpuset as we try to mimic clone_children iirc.

bgray:
Hi,  I've re-worked the code to make it unit-testable and added a new unittest.  I'd be interested to see if you think this is better, or not.

I spent a fair amount of time trying to get an integration test to work, but failed.  The problem is that docker doesn't set any `cpu.rt_runtime_us` value, therefore running the integration tests within a docker container doesn't work (even simply calling `chrt` fails).  It can be fixed by manually writing to the various cgroup docker files as root, but that seems a bit out of scope of the test.  And the same problem exists if you run the integration tests locally, you need to modify the cgroups tree to give each some bandwidth, something my ubuntu distro doesn't do automatically.

hqhq:
@bgray Thanks for the PR, janky is failed, can you check if it's related? Codes look good to me.

bgray:
Thanks @hqhq , I've looked at the janky log and I can't see how the failures could be related.  FWIW I've run the test locally and they all pass, so um, not sure.  Is it possible to re-trigger the jenkins job to try running the tests again ?

cyphar:
I've restarted the Jenkins build.

hqhq:
LGTM

@cyphar May I ask how to re-trigger Jenkins? Do all maintainers have access to re-trigger Jenkins? I tried to re-trigger, but didn't figure out how. :(

cyphar:
@hqhq Go to `#docker-maintainers` and say `!rebuild runc#<issue>`.

hqhq:
@cyphar Thanks, I'll try next time :smile:

libcontainer/cgroups/fs/cpu.go
```diff
@@ -45,15 +83,8 @@ func (s *CpuGroup) Set(path string, cgroup *configs.Cgroup) error {
...
-	if cgroup.Resources.CpuRtPeriod != 0 {
-		if err := writeFile(path, "cpu.rt_period_us", strconv.FormatInt(cgroup.Resources.CpuRtPeriod, 10)); err != nil {
-			return err
-		}
-	}
-	if cgroup.Resources.CpuRtRuntime != 0 {
-		if err := writeFile(path, "cpu.rt_runtime_us", strconv.FormatInt(cgroup.Resources.CpuRtRuntime, 10)); err != nil {
-			return err
-		}
+	if err := s.SetRtSched(path, cgroup); err != nil {
```
cyphar: Is this meant to be here? In the comment for `ApplyDir` (which I think shouldn't be a new function) you say that:

> ```
> // We should set the real-Time group scheduling settings before moving
> // in the process because if the process is already in SCHED_RR mode
> // adding it will fail.
> ```

Surely setting it afterward will cause it to fail? If not, can you please update the comment to explain when it's okay to use `SetRtSched`?

hqhq: I think it's here for updating settings, thought it means we'll double setting rt configs when start a container, but we already did this for kernel memory.

bgray: Yes - the issue is that you can't move a process that is already running SCHED_RR into a cpu cgroup that has no RT bandwidth.   It can still be changed after the process is moved into the cgroup (as long as it's not changed to 0).

Would the following suffice to make it a bit more clear ?
```
// We should set the real-Time group scheduling settings before moving
// in the process because if the process is already in SCHED_RR mode
// and no RT bandwidth is set adding it will fail.
```

cyphar: Yes, that's better.

cyphar:
Can you please squash your commits? I'd also prefer that the test addition be a separate commit.

libcontainer/cgroups/fs/cpu.go
```diff
@@ -22,10 +22,48 @@ func (s *CpuGroup) Name() string {
...
+	// because we are not using d.join we need to place the pid into the procs file
+	// unlike the other subsystems
+	if err := writeFile(path, "cgroup.procs", strconv.Itoa(pid)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *CpuGroup) SetRtSched(path string, cgroup *configs.Cgroup) error {
```
cyphar: I don't know how all of the other maintainers feel about this ( @opencontainers/runc-maintainers ), but maybe we should tell the user that they have to specify `CpuRt*` options if they are available and runC is `SCHED_RR`?

hqhq: I think this should be done by high level tools like Docker, runc should just take what's been given and apply. But runc does check some validations, I think we still don't have a clear bound yet.

hqhq:
LGTM
ping @cyphar @dqminh

cyphar:
LGTM.

PR #863, opened 2016-06-01
Disallow self-LGTMs

hqhq:
As https://github.com/opencontainers/project-template/pull/13
is merged, change pullapprove accordingly.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -2,6 +2,7 @@ approve_by_comment: true
 approve_regex: ^LGTM
 reject_regex: ^Rejected
 reset_on_push: true
+author_approval: ignored
 reviewers:
   teams:
     - runc-maintainers


cyphar:
LGTM

cyphar:
Can you also include the `MAINTAINERS_GUIDE.md` changes from that PR as well?

hqhq:
@cyphar I think the plan should not be that we sync these docs to every projects of OCI when they changed, if we want to maintain these docs in one place, we might need some action like we did in runtime-spec for [code-of-conduct.md](https://github.com/opencontainers/runtime-spec/blob/master/Makefile#L56), that's something we need to figure out.

cyphar:
If we're going to do that, we should link the files IMO. But in that case LGTM.

dqminh:
LGTM

PR #865, opened 2016-06-01
systemd cgroup: check for Delegate property

### LINKED ISSUE #801: runc and systemd slice (--systemd-cgroup flag)
OS : debian jessie / systemd 215

runc : 

```
VERSION:
   0.1.1
commit: 89c3c97a8482f3a57cd4bb683df1a7b2c61405d8
spec: 0.6.0-dev
```

bundle : 

```
test/
├── config.json
└── rootfs

```

rootfs is created with `debootsrap` and config.json generated using : `runc spec`

Basic start of the container works well using `runc start test` command.

Adding the --systemd-cgroup allow users to specify a systemd slice. Without a specific configuration, when the --systemd-cgroup flag is used, runc tries to attach the processus to the systemd system slice (defined in createCgroupConfig function https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L255)

but `runc --systemd-cgroup start test` gives : 

`process_linux.go:253: applying cgroup configuration for process caused "Cannot set property Delegate, or unknown property."`

Adding a specific slice config according to the OCI (https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#control-groups) : 

```
        "linux": {
                "resources": {
                        "devices": [
                                {
                                        "allow": false,
                                        "access": "rwm"
                                }
                        ],
                        "cgroupsPath": "system.slice:runc:434234"
                },
                "namespaces": [

```

Gives the same error.

dqminh:
Fix #801 

Delegate is only available in systemd >218, applying it for older systemd will
result in an error. Therefore we should check for it when testing systemd
properties.

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -74,6 +74,7 @@ var (
 	theConn                         *systemdDbus.Conn
 	hasStartTransientUnit           bool
 	hasTransientDefaultDependencies bool
+	hasDelegate                     bool
 )
 
 func newProp(name string, units interface{}) systemdDbus.Property {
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -144,6 +145,16 @@ func UseSystemd() bool {
 			}
 		}
 
+		hasDelegate = true
+		dl := newProp("Delegate", true)
+		if _, err := theConn.StartTransientUnit(scope, "replace", []systemdDbus.Property{dl}, nil); err != nil {
```
cyphar: Don't you have to destroy this unit first? Or is this dealt with by the "stop unit logic"?

dqminh: oops i must have done something wrong when checking it in. Fixed now.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #867, opened 2016-06-02
Removing the nil check for process label

rajasec:
Function SetProcessLabel is checking for nil, hence removed the redundant one from the caller.

func SetProcessLabel(processLabel string) error {
        if processLabel == "" {
                return nil
        }


hqhq:
LGTM

mrunalp:
LGTM

PR #872, opened 2016-06-03
runc events hang for zero duration

rajasec:
Tick function in go returns nil when the duration is lesser than or equal to 0. When we run the events command with interval --0s, runc hangs. So added the condition to throw out the error when the duration is lesser than or equal to.


crosbymichael:
LGTM

mrunalp:
LGTM

PR #875, opened 2016-06-03
Return error for GetLongBit

### LINKED ISSUE #841: sysconfig_notcgo does not implement GetLongBit
When compiling with `CGO_ENABLED=0` this results in a build error. This occurs for us because this package is a dependency of a dependency, and we're building our package using `CGO_ENABLED=0`. The error is that `system.GetLongBit` is not defined.

This is where `GetLongBit` defined for most build: https://github.com/opencontainers/runc/blob/master/libcontainer/system/sysconfig.go#L29

This file is used when CGO is not enabled: https://github.com/opencontainers/runc/blob/master/libcontainer/system/sysconfig_notcgo.go

`GetLongBit` is only used in `libcontainer/cgroups/fs` so you can reproduce it as follows:

```
$ cd $GOPATH/src/github.com/opencontainers/runc/libcontainer/cgroups/fs
$ CGO_ENABLED=0 go build
# github.com/opencontainers/runc/libcontainer/cgroups/fs
./memory.go:71: undefined: system.GetLongBit
```

crosbymichael:
For systems that are not compiled with CGO GetLongBit function is
unreliable and should return an error.

Fixes #841

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -68,7 +68,11 @@ func (s *MemoryGroup) SetKernelMemory(path string, cgroup *configs.Cgroup) error
 		if err != nil {
 			return err
 		}
-		switch system.GetLongBit() {
+		lb, err := system.GetLongBit()
+		if err != nil {
+			return err
+		}
+		switch lb {
 		case 32:
 			kmemInitialized = uint32(kmemValue) != uint32(math.MaxUint32)
 		case 64:
--- a/libcontainer/system/sysconfig.go
+++ b/libcontainer/system/sysconfig.go
@@ -26,6 +26,6 @@ func GetClockTicks() int {
 	return int(C.sysconf(C._SC_CLK_TCK))
 }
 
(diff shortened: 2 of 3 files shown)


mrunalp:
LGTM

mlaventure:
LGTM (IANAM)

cyphar:
NACK. This means that you disable kernel memory limit setting on systems that don't have `gc` (which is s390 and ppc64le was very recent too).

crosbymichael:
@cyphar we could assume 64bit but will give wrong answers on everything else compiling without CGO.  Most systems should have cgo, this is just a tag for people explicitly turning it off

dqminh:
Yah i'm also a bit confused with the long bit code. Will `strconv.IntSize` suffice for this purpose, which is essentially `32 << (^uint(0) >> 63)` ?

cyphar:
@dqminh I think the problem is that it's a runtime thing, so if you are running a 32-bit binary on top of a 64-bit host then you'll be writing the wrong values.

PR #878, opened 2016-06-04
Adjust man pages for create start split

hqhq:

### DIFF
--- a/man/runc.8.md
+++ b/man/runc.8.md
@@ -39,8 +39,9 @@ value for "bundle" is the current directory.
    ps           displays the processes running inside a container
    restore      restore a container from a previous checkpoint
    resume       resumes all processes that have been previously paused
+   run          create and run a container
    spec         create a new specification file
-   start        create and run a container
+   start        executes the user defined process in a created container
    state        output the state of a container
    update       update container resource constraints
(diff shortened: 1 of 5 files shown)


man/runc.8.md
```diff
@@ -39,8 +39,9 @@ value for "bundle" is the current directory.
    ps           displays the processes running inside a container
    restore      restore a container from a previous checkpoint
    resume       resumes all processes that have been previously paused
+   run          create and run a container
    spec         create a new specification file
-   start        create and run a container
+   start        start signals a created container to execute the user defined process
```
dqminh: might be better phrased as `start executes the user defined process in a created container`

hqhq: The initial phrase seems more accurate about what start does, but I'm ok with both, since it's Michael's words, @crosbymichael WDYT?

cyphar: As an aside, we shouldn't repeat the command name as the first word. So "signals a created container to start execution".

crosbymichael: whatever is more clear is good with me if you want to change both instances of the sentence

hqhq:
Updated. PTAL.

crosbymichael:
LGTM

cyphar:
LGTM

PR #883, opened 2016-06-06
tests: remove --console usage

cyphar:
The previous usage of --console was very misleading, and is not how the
flag should be used by users. In addition, due to other bugs and design
issues, --console does not work in user namespaces. By removing
--console usage in the tests, we can now (in principle) run all of the
tests in user namespaces.


cyphar:
The failures are because it looks like `terminal: false` and `-d` don't play properly together. The process isn't actually detached AFAICS.

/cc @crosbymichael

crosbymichael:
looking

cyphar:
Superceded by #1018.

PR #886, opened 2016-06-06
Use fifo for create / start instead of signal handling 

### LINKED ISSUE #871: state's status detection (based on PID lookups) is imprecise
Since a few days ago, the runtime spec has a `status` property in the state (opencontainers/runtime-spec#462).  I was curious about how that worked with detached containers, since there's no host-side monitor around to notice the container die.  Testing with 1d61abea (the current tip of #827, although I expect this applies to all recent runC code), it looks like the `state` call works around the lack of a monitor (who could waitid(2) or some such on the container process) by [looking up the container PID in /proc dynamically for each `state` call][1].  That's racy though, since “has the same PID in /proc” doesn't mean “is the same process”.  For example:

    $ ./runc spec
    $ JSON=$(jq 'del(.process.terminal)' config.json)
    $ echo "${JSON}" >config.json
    $ mkdir -p rootfs/bin
    $ cp $(command -v busybox) rootfs/bin/sh
    $ STATE=$(sudo ./runc state abc)
    $ echo "${STATE}"
    {
      "ociVersion": "0.6.0-dev",
      "id": "abc",
      "pid": 26373,
      "bundlePath": "/…/runc",
      "rootfsPath": "/…/runc/rootfs",
      "status": "created",
      "created": "2016-06-03T03:25:28.641226726Z"
    }
    $ PID=$(echo "${STATE}" | jq -r .pid)
    $ sudo kill -9 "${PID}"
    $ sudo ./runc state abc
    {
      "ociVersion": "0.6.0-dev",
      "id": "abc",
      "pid": 26373,
      "bundlePath": "/…/runc",
      "rootfsPath": "/…/runc/rootfs",
      "status": "stopped",
      "created": "2016-06-03T03:25:28.641226726Z"
    }

In a separate terminal, spawn processes until we match the old container PID:

    $ while true; do sh -c 'if test $$ -eq 26373; then echo muahaha; sleep 60; fi'; done
    muahaha
    ^C

And then runC *thinks* (wrongly) that the container is alive again:

    $ sudo ./runc state abc
    Password:
    {
      "ociVersion": "0.6.0-dev",
      "id": "abc",
      "pid": 26373,
      "bundlePath": "/…/runc",
      "rootfsPath": "/…/runc/rootfs",
      "status": "running",
      "created": "2016-06-03T03:25:28.641226726Z"
    }

One solution is to have the parent process (or an ancestor with `PR_SET_CHILD_SUBREAPER` after the parent dies) waitid(2) on the container process.

Another solution is to store more information about the container process (we probably already do) and compare that additional data when deciding whether the process running with the container-process PID was still the container process.  For example, the PID and start-time (field 22 in `/proc/[pid]/stat`) form a tuple that is almost certainly unique.

[1]: https://github.com/opencontainers/runc/blob/1d61abea46acf375008c983df29b1dcf03d9d127/libcontainer/container_linux.go#L1077-L1103

crosbymichael:
This removes the use of a signal handler and SIGCONT to signal the init
process to exec the users process.

Closes #871

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -29,10 +29,6 @@ import (
 
 const stdioFdCount = 3
 
-// InitContinueSignal is used to signal the container init process to
-// start the users specified process after the container create has finished.
-const InitContinueSignal = syscall.SIGCONT
-
 type linuxContainer struct {
 	id            string
 	root          string
@@ -195,14 +191,37 @@ func (c *linuxContainer) Run(process *Process) error {
 	if err != nil {
 		return err
 	}
-	isInit := status == Stopped
-	if err := c.start(process, isInit); err != nil {
+	if err := c.start(process, status == Stopped); err != nil {
 		return err
 	}
-	if isInit {
-		return process.ops.signal(InitContinueSignal)
+	if status == Stopped {
+		return c.exec()
 	}
 	return nil
+}
+
+func (c *linuxContainer) Exec() error {
+	c.m.Lock()
+	defer c.m.Unlock()
+	return c.exec()
+}
+
+func (c *linuxContainer) exec() error {
+	path := filepath.Join(c.root, execFifoFilename)
+	f, err := os.OpenFile(path, os.O_RDONLY, 0)
+	if err != nil {
+		return newSystemErrorWithCause(err, "open exec fifo for reading")
+	}
+	defer f.Close()
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		return err
+	}
+	if len(data) > 0 {
+		os.Remove(path)
+		return nil
+	}
+	return fmt.Errorf("cannot start an already running container")
 }
 
 func (c *linuxContainer) start(process *Process, isInit bool) error {
@@ -262,17 +281,21 @@ func (c *linuxContainer) newParentProcess(p *Process, doInit bool) (parentProces
 	if err != nil {
 		return nil, newSystemErrorWithCause(err, "creating new init pipe")
 	}
-	cmd, err := c.commandTemplate(p, childPipe)
+	rootDir, err := os.Open(c.root)
+	if err != nil {
+		return nil, err
+	}
+	cmd, err := c.commandTemplate(p, childPipe, rootDir)
 	if err != nil {
 		return nil, newSystemErrorWithCause(err, "creating new command template")
 	}
 	if !doInit {
-		return c.newSetnsProcess(p, cmd, parentPipe, childPipe)
+		return c.newSetnsProcess(p, cmd, parentPipe, childPipe, rootDir)
 	}
-	return c.newInitProcess(p, cmd, parentPipe, childPipe)
+	return c.newInitProcess(p, cmd, parentPipe, childPipe, rootDir)
 }
 
-func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.Cmd, error) {
+func (c *linuxContainer) commandTemplate(p *Process, childPipe, rootDir *os.File) (*exec.Cmd, error) {
 	cmd := &exec.Cmd{
 		Path: c.initPath,
 		Args: c.initArgs,
@@ -284,8 +307,10 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
-	cmd.ExtraFiles = append(p.ExtraFiles, childPipe)
-	cmd.Env = append(cmd.Env, fmt.Sprintf("_LIBCONTAINER_INITPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
+	cmd.ExtraFiles = append(p.ExtraFiles, childPipe, rootDir)
+	cmd.Env = append(cmd.Env,
+		fmt.Sprintf("_LIBCONTAINER_INITPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-2),
+		fmt.Sprintf("_LIBCONTAINER_STATEDIR=%d", stdioFdCount+len(cmd.ExtraFiles)-1))
 	// NOTE: when running a container with no PID namespace and the parent process spawning the container is
 	// PID1 the pdeathsig is being delivered to the container's init process by the kernel for some reason
 	// even with the parent still running.
@@ -295,7 +320,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	return cmd, nil
 }
 
-func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) (*initProcess, error) {
+func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe, rootDir *os.File) (*initProcess, error) {
 	cmd.Env = append(cmd.Env, "_LIBCONTAINER_INITTYPE="+string(initStandard))
 	nsMaps := make(map[configs.NamespaceType]string)
 	for _, ns := range c.config.Namespaces {
@@ -318,10 +343,11 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
 		process:       p,
 		bootstrapData: data,
 		sharePidns:    sharePidns,
+		rootDir:       rootDir,
 	}, nil
 }
 
-func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) (*setnsProcess, error) {
+func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe, rootDir *os.File) (*setnsProcess, error) {
 	cmd.Env = append(cmd.Env, "_LIBCONTAINER_INITTYPE="+string(initSetns))
 	state, err := c.currentState()
 	if err != nil {
@@ -342,6 +368,7 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, 
 		config:        c.newInitConfig(p),
 		process:       p,
 		bootstrapData: data,
+		rootDir:       rootDir,
 	}, nil
 }
 
@@ -360,6 +387,7 @@ func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
 		AppArmorProfile:  c.config.AppArmorProfile,
 		ProcessLabel:     c.config.ProcessLabel,
 		Rlimits:          c.config.Rlimits,
+		ExecFifoPath:     filepath.Join(c.root, execFifoFilename),
 	}
 	if process.NoNewPrivileges != nil {
 		cfg.NoNewPrivileges = *process.NoNewPrivileges
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/signal"
 	"path/filepath"
 	"regexp"
 	"runtime/debug"
@@ -24,7 +23,8 @@ import (
 )
 
 const (
-	stateFilename = "state.json"
+	stateFilename    = "state.json"
+	execFifoFilename = "exec.fifo"
 )
 
 var (
@@ -168,6 +168,9 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := os.MkdirAll(containerRoot, 0700); err != nil {
 		return nil, newGenericError(err, SystemError)
 	}
+	if err := syscall.Mkfifo(filepath.Join(containerRoot, execFifoFilename), 0666); err != nil {
+		return nil, newGenericError(err, SystemError)
+	}
 	c := &linuxContainer{
 		id:            id,
 		root:          containerRoot,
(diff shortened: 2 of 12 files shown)


libcontainer/factory_linux.go
```diff
@@ -168,6 +168,9 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := os.MkdirAll(containerRoot, 0700); err != nil {
 		return nil, newGenericError(err, SystemError)
 	}
+	if err := syscall.Mkfifo(filepath.Join(containerRoot, execFifoFilename), 0655); err != nil {
```
wking: `0655` is a pretty strange permission set.  I'm not clear on the intended auth here, but the signal-based approach was limited by kill(2)'s `CAP_KILL` and/or UID requirements (although SIGCONT had a same-session special case).  I expect you want `0600` here to match that.

libcontainer/standard_init_linux.go
```diff
@@ -44,7 +43,22 @@ func (l *linuxStandardInit) getSessionRingParams() (string, uint32, uint32) {
...
-func (l *linuxStandardInit) Init(s chan os.Signal) error {
+func (l *linuxStandardInit) Init() error {
+	var (
+		goStarted   = make(chan struct{}, 1)
+		waitForExec = make(chan struct{}, 1)
+	)
+	// use channel 'c' to ensure that the goroutine was scheduled so that the open is called
+	// before we pivot into the rootfs or else we won't have access to the path on the host.
+	go func() {
+		close(goStarted)
```
LK4D4: Is this really help? Because scheduler still can reschedule on OpenFile call

crosbymichael: It gets us closer than without

libcontainer/container_linux.go
```diff
@@ -195,12 +191,23 @@ func (c *linuxContainer) Run(process *Process) error {
...
+	return nil
+}
+
+func (c *linuxContainer) Exec() error {
+	f, err := os.OpenFile(filepath.Join(c.root, execFifoFilename), os.O_RDONLY, 0)
+	if err != nil {
+		return newSystemErrorWithCause(err, "open exec fifo for reading")
+	}
+	defer f.Close()
+	if _, err := ioutil.ReadAll(f); err != nil {
```
LK4D4: Why we can't use a pipe like in user namespace writing trick from @mrunalp in go syscall package? It seems like FIFO require strange assumptions about go scheduler behavior.

crosbymichael: Because we need it to persist on disk because the processes go away

start.go
```diff
@@ -27,7 +27,7 @@ your host.`,
 		}
 		switch status {
 		case libcontainer.Created:
-			return container.Signal(libcontainer.InitContinueSignal)
+			return container.Exec()
```
wking: I'd rather have the container process remove the FIFO after it unblocks.  Then `start` can always `Exec()`, and you can catch the “FIFO does not exist” error and translate it to a prettier “someone must have already started the container”.

libcontainer/standard_init_linux.go
LK4D4: Yeah, it's like 30% less race chance, but still... It's really bad.

libcontainer/container_linux.go
```diff
@@ -195,12 +191,23 @@ func (c *linuxContainer) Run(process *Process) error {
...
 	}
-	if isInit {
-		return process.ops.signal(InitContinueSignal)
+	if status == Stopped {
+		return c.Exec()
+	}
+	return nil
+}
+
+func (c *linuxContainer) Exec() error {
```
wking: You should probably call this `Start` to match the spec's operation.

crosbymichael: this is an internal api, we can call it what we want

wking: On Mon, Jun 06, 2016 at 05:04:59PM -0700, Michael Crosby wrote:
> > +func (c *linuxContainer) Exec() error {
> 
> this is an internal api, we can call it what we want

Can != should ;).  But you're the maintainer, so it's not my call.

crosbymichael: its not changing because we already have a start internally that is used in other places

wking: On Mon, Jun 06, 2016 at 05:06:57PM -0700, Michael Crosby wrote:
> > +func (c *linuxContainer) Exec() error {
> 
> its not changing because we already have a start internally that is
> used in other places

Ah, in that case the missmatch with the spec is probably unavoidable.

libcontainer/standard_init_linux.go
LK4D4: So, this code is definitely racy :( Maybe we can create fifo in rootfs?

crosbymichael: could we just call the raw syscall or still an issue?

crosbymichael: i don't think putting it in the rootfs will help because of readonly containers and such. we would want to delete it after we are done and could'nt

mrunalp: If there is no race free way to make this work, we could just use either SIGUSR1/SIGUSR2 signals that are reserved for user code.

libcontainer/standard_init_linux.go
```diff
@@ -44,7 +44,15 @@ func (l *linuxStandardInit) getSessionRingParams() (string, uint32, uint32) {
 // the kernel
 const PR_SET_NO_NEW_PRIVS = 0x26
 
-func (l *linuxStandardInit) Init(s chan os.Signal) error {
+func (l *linuxStandardInit) Init() error {
+	waitForExec := make(chan struct{}, 1)
+	go func() {
+		if _, err := syscall.Openat(l.rootfd, execFifoFilename, os.O_WRONLY|syscall.O_CLOEXEC, 0); err != nil {
```
cyphar: We should close this `fd` before `exec`ing the user's process. No need to give them access to somewhere in `/`.

libcontainer/factory_linux.go
```diff
@@ -220,13 +223,13 @@ func (l *LinuxFactory) Type() string {
...
 // This is a low level implementation detail of the reexec and should not be consumed externally
 func (l *LinuxFactory) StartInitialization() (err error) {
-	// start the signal handler as soon as we can
-	s := make(chan os.Signal, 1)
-	signal.Notify(s, InitContinueSignal)
-	fdStr := os.Getenv("_LIBCONTAINER_INITPIPE")
-	pipefd, err := strconv.Atoi(fdStr)
+	pipefd, err := strconv.Atoi(os.Getenv("_LIBCONTAINER_INITPIPE"))
 	if err != nil {
-		return fmt.Errorf("error converting env var _LIBCONTAINER_INITPIPE(%q) to an int: %s", fdStr, err)
```
cyphar: Why did you drop the error message? We should probably keep this one (and the new one).

cyphar:
Personally, I liked the signal approach to create-start. What is the reason for the change? While `SIGCONT` has issues with Go 1.6, I'm not sure why we aren't just switching to `SIGUSR1` -- IMO signals are the *right* Unix interface here (not FIFOs).

libcontainer/standard_init_linux.go
```diff
@@ -153,17 +161,15 @@ func (l *linuxStandardInit) Init(s chan os.Signal) error {
...
 	}
-	// close the pipe to signal that we have completed our init
+	// close the pipe to signal that we have completed our init.
 	l.pipe.Close()
-	// wait for the signal to exec the users process
-	<-s
-	// clean up the signal handler
-	signal.Stop(s)
-	close(s)
+	// wait for the fifo be be opened on the other side before
```
hqhq: typo "be be"

wking:
On Mon, Jun 06, 2016 at 10:05:32PM -0700, Aleksa Sarai wrote:
> Personally, I liked the signal approach to create-start. What is the
> reason for the change? While `SIGCONT` has issues with Go 1.6, I'm
> not sure why we aren't just switching to `SIGUSR1` -- IMO signals
> are the *right* Unix interface here (not FIFOs).

The problem with signals is that you're just launching them, and don't
know:

* If/when they are recieved and acted upon, or
* Which code will be running when the process recieves the signal [1].

I'm not entirely sure how much of that you fix with a FIFO, but the
ccon implementation [2] (which uses a two-way Unix socket for this)
can (but doesn't actually implement yet) have the container process
send a “message received, I'll go exec that code” message back to the
‘start’ process.  That way:

* The ‘start’ process that actually triggers the exec knows it, and
  can exit success.  Any parallel ‘start’ requests who are queued up
  waiting for the socket will know that they didn't trigger the exec,
  and can exit failure (as required by the spec [3]).

* You know you'll never racily send a signal that is processed by
  user-specified code, because the container process should be closing
  (and in ccon, the host-process will remove post-post start, once I
  have time to code that up, the filesystem portion of the
  socket/FIFO).

[1]: https://github.com/opencontainers/runc/pull/827#discussion_r63461012
[2]: https://github.com/wking/ccon/tree/create-start#socket-communication
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#start
     “Attempting to start a container that does not exist MUST
     generate an error. Attempting to start an already started
     container MUST have no effect on the container and MUST generate
     an error.”

libcontainer/standard_init_linux.go
mlaventure: `syscall.O_CLOEXEC` should take care of closing the fd before the new process takes over.

LK4D4: Can't we just do it at the end without goroutine?

crosbymichael:
@LK4D4 @cyphar i updated and now its "race free' ;)

wking:
On Mon, Jun 06, 2016 at 10:05:32PM -0700, Aleksa Sarai wrote:
> While I understand @wking's concerns with PID namespaces, it feels
> like we still have several other problems that we can't solve this
> way (if we used sockets) -- why should runc start work if runc ps or
> runc exec won't (since the PID is hardcoded into the
> namespace_paths)?

I don't see “we have out-of-spec operations that break under these
conditions” as a reason to avoid having more reliable support for
in-spec operations (although in this case, my main concern with
signals is the race which may lead to user-specified code handling a
signal intended for the runtime's idler code [1]).

And while using a filesystem object (in this cases a named pipe) gets
you out of the “I'm in a different PID namespace” issue [2], it still
has “I'm in a different mount namespace and can't access runC's state
registry”.  I don't see a way around that, although I do wish we had
better wording about this issue than the spec's current [3]
(e.g. [4]).

[1]: https://github.com/opencontainers/runc/pull/886#issuecomment-224266864
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/ujtABQoCmgk
     Subject: Linux: Adding the PID namespace inode to state JSON
     Date: Wed, 30 Dec 2015 21:34:57 -0800
     Message-ID: <20151231053457.GG3680@odin.tremily.us>
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#scope-of-a-container
[4]: https://github.com/opencontainers/runtime-spec/pull/225#issuecomment-175790463

crosbymichael:
ping,

I've been running stress tests and everything looks good with this change

dqminh:
LGTM

LK4D4:
@crosbymichael LGTM, but need rebase. Feel free to merge yourself after it.

mrunalp:
I am reviewing/testing this.

crosbymichael:
rebased

libcontainer/container_linux.go
```diff
@@ -195,16 +191,33 @@ func (c *linuxContainer) Run(process *Process) error {
...
-		return process.ops.signal(InitContinueSignal)
+	if status == Stopped {
+		return c.Exec()
 	}
 	return nil
 }
 
+func (c *linuxContainer) Exec() error {
+	path := filepath.Join(c.root, execFifoFilename)
+	f, err := os.OpenFile(path, os.O_RDONLY, 0)
```
mrunalp: If two processes end up opening this file at the same time, then one of them will be stuck waiting on the ReadAll below.

libcontainer/standard_init_linux.go
```diff
@@ -21,6 +20,7 @@ import (
 type linuxStandardInit struct {
 	pipe      io.ReadWriteCloser
 	parentPid int
+	rootfd    int
```
mrunalp: nit: Can we call this `statedirfd`? `rootfd` seems to imply that it is a fd to the rootfs.

libcontainer/container_linux.go
crosbymichael: True but we don't support multiple processes handling the container.  I can try to protect the libcontainer call with a lock but its not going to help out of process calls.

wking: On Mon, Jun 13, 2016 at 11:12:02AM -0700, Michael Crosby wrote:
> >  	}
> >  	return nil
> >  }
> >  
> > +func (c *linuxContainer) Exec() error {
> > +	path := filepath.Join(c.root, execFifoFilename)
> > +	f, err := os.OpenFile(path, os.O_RDONLY, 0)
>
> True but we don't support multiple processes handling the container.
> I can try to protect the libcontainer call with a lock but its not
> going to help out of process calls.

I thought putting a byte in the pipe would help with this [1].  Could
we make the reads unblocking?  If you got the byte, success.  If you
didn't get the byte, exit 1.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-06-07.log.html#t2016-06-07T17:05:15

cyphar:
I still want to take some time to review this. Unfortunately I have exams until Friday.

crosbymichael:
@cyphar we need this to fix the racyness and 1.6 issues that we currently have with using signals

cyphar:
Alright. I'll take a look at this after merging, and make follow-up PRs if necessary.

libcontainer/container_linux.go
mrunalp: @crosbymichael I agree that with current usage this should be fine. We could add flocks in the future if necessary.

mrunalp:
LGTM

crosbymichael:
@mrunalp updating your nits and added a lock to that function

mrunalp:
@crosbymichael Thanks! I'll wait for that before merging.

crosbymichael:
updated

mrunalp:
We have some more references to root instead of state but some of it predates this PR so we can fixup in a follow on.

mrunalp:
LGTM

mrunalp:
@LK4D4 this needs another LGTM as the PR was updated :)

LK4D4:
@mrunalp CI is not ready too :)

LK4D4:
LGTM

PR #888, opened 2016-06-07
start: use SIGUSR1 for start signal

cyphar:
Since we don't use SIGSTOP, the choice of wait signal is arbitrary.
However, Go versions before 1.6 did not support waiting for SIGCONT --
this is a problem for platforms which do not have the gc compiler (as
gcc-go implements Go 1.4).

This is an alternative to #886.


crosbymichael:
This can be closed now that we are using a pipe for this functionality.

PR #890, opened 2016-06-07
restore: add the empty-ns option

avagin:
The same option was added for the "checkpoint" command in https://github.com/opencontainers/runc/commit/3ad3d937d280f58ead3fc7814968d86c1724d6f9

For example:
./runc restore --empty-ns network CTID

In this case criu creates a network namespace, but doesn't restore it.

We are going to use this option to restore docker containers and
Docker sets a hook to restore a network namespace.

https://github.com/xemul/criu/issues/165
Cc: @boucher 
Signed-off-by: Andrew Vagin <avagin@virtuozzo.com>


crosbymichael:
LGTM

dqminh:
LGTM

PR #898, opened 2016-06-08
vendor: use https://github.com/govend/govend for vendoring

LK4D4:
Revendor should be as simple as changing vendor.yml and running
govend -v --prune.
It creates 6 Mb vendor instead of 2.1 Mb Godeps, but I think it vendors more tags, like "solaris", "windows", etc.

It's not like real PR, just PoC.
ping @cyphar


dqminh:
govend does look simpler to use.
The build failed, but the error looks expected (because the checked in vendor doesnt pass gofmt). Maybe we want to ignore the vendor from the validation ?

cyphar:
We're missing some changes in the `Makefile` (`grep -i godeps`). But also, I want to have a `hack/vendor.sh` script that we can easily update. One of the things that annoys me about `godep` is that there isn't an easy way to make it update a vendor to a specific commit.

I know @monstermunchkin wanted to make a PR.

monstermunchkin:
@cyphar I'm still having some trouble with failing tests after moving from Godeps to govend.

LK4D4:
@cyphar @monstermunchkin I also see a problem with our make commands, before we used `Godeps/_workspace` as `GOPATH`.

LK4D4:
Ok, I figured everything out, not sure why test fails though.

cyphar:
I really don't like that Go decided to make `vendor` not contain a `src` directory. It's really dumb, because it flies in the face of every tool that had already been written (and it breaks things like setting `GOPATH=vendor`).

But maybe there's something like `ln -sf . vendor/src` that lets us fake the `GOPATH=vendor` stuff.

LK4D4:
@cyphar I used just tmp gopath. Symlinks don't work :/ I think it's because of go tooling.

cyphar:
`govend` isn't as clever as I like, namely it doesn't correctly deal with `_test.go` files (it vendors them even though they aren't needed). What if we just use the `hack/vendor.sh` code from Docker?

cyphar:
@LK4D4 The reason the tests fail is that you haven't maintained the same commit IDs on the vendor. Please use a shell script to convert from `godeps.json` to `vendor.yml` -- some of our dependencies have presumably broken recently.

LK4D4:
@cyphar sadly `hack/vendor.sh` works only because of @tianon bash-skills :( It's very fragile and depends on tricky magic of `go list` which works differently with new vendoring.

LK4D4:
@cyphar govend actually deletes test files.

LK4D4:
@cyphar yup, with proper versions it pass. I think `govend` is good middle-ground between godep and glide. At least we can use it for now.

cyphar:
> @cyphar govend actually deletes test files.

Oh, you're right. I misread the `[no test files]` lines. :wink:

LGTM

LK4D4:
@cyphar Actually, I cleaned autogenerated vendor.yml and now vendor only 1.7M, less than godeps.

cyphar:
Wait, so what have we lost?

cyphar:
Have you tried this instead of doing the temporary directory stuff:

```
VENDOR_DIR := $(CURDIR)/vendor

gopath:
	ln -srfn . $(VENDOR_DIR)/src
	mkdir -p $(VENDOR_DIR)/src/github.com/opencontainers
	ln -sfn $(CURDIR) $(VENDOR_DIR)/src/github.com/opencontainers/runc
```

LK4D4:
@cyphar I think it doesn't work for stuff like go vet and go test. Also, I don't like modifying repo in Makefile. It was source of confusion before.

crosbymichael:
Now that go 1.7 is out, what is the best way to vendor our deps with a standard go vendor path?

dqminh:
@crosbymichael still need to use one of the tools, Godep, Govend etc. Nicer thing is that we dont have to do the symlink dance though.

WDYT about just moving our deps towards standard vendor location first ( with Godeps ), then we can look at other tools if we want ?

LK4D4:
@dqminh I agree, we should do that.
I've also created another PR, which uses my vndr tool: #998

crosbymichael:
Closing this in favor of #998

PR #900, opened 2016-06-09
bats: Fix spec validation test

mrunalp:
Use the commit that we have in Godeps.json rather than relying on
runtime-spec HEAD

### DIFF
--- a/tests/integration/spec.bats
+++ b/tests/integration/spec.bats
@@ -66,12 +66,15 @@ function teardown() {
 }
 
 @test "spec validator" {
+  TESTDIR=$(pwd)
   cd "$HELLO_BUNDLE"
-  # note this test runs from the temporary bundle directory not the integration root
-  # note this test is brittle against specs changes that lead runc's spec command
-  # todo get the validate program, gojsonschema, and schema/*s.json from godeps?
 
   run git clone https://github.com/opencontainers/runtime-spec.git src/runtime-spec
+  [ "$status" -eq 0 ]
+
+  SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
+  run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"
+  [ "$status" -eq 0 ]
   [ -e src/runtime-spec/schema/schema.json ]
 
   run bash -c "GOPATH='$GOPATH' go get github.com/xeipuuv/gojsonschema"


mrunalp:
https://github.com/opencontainers/runtime-spec/pull/481/files broke spec validation. This PR should make the test more reliable.

mrunalp:
@crosbymichael @LK4D4 PTAL. Tests are broken right now without fixing this.

tests/integration/spec.bats
```diff
@@ -66,12 +66,18 @@ function teardown() {
...
 
   run git clone https://github.com/opencontainers/runtime-spec.git src/runtime-spec
+  [ "$status" -eq 0 ]
+
+  SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
+  cd src/runtime-spec
+  run git reset --hard $SPEC_COMMIT
+  [ "$status" -eq 0 ]
+
+  cd "$HELLO_BUNDLE"
```
wking: Instead of `cd`ing in and back, you can use something like `run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"`.

mrunalp: done

tests/integration/spec.bats
```diff
@@ -66,12 +66,14 @@ function teardown() {
...
   cd "$HELLO_BUNDLE"
-  # note this test runs from the temporary bundle directory not the integration root
-  # note this test is brittle against specs changes that lead runc's spec command
-  # todo get the validate program, gojsonschema, and schema/*s.json from godeps?
 
   run git clone https://github.com/opencontainers/runtime-spec.git src/runtime-spec
+  [ "$status" -eq 0 ]
+
+  SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
+  run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"
```
wking: I'm not familiar with bats, but I think you want to check the exit status of the Git command with `[ "$status -eq 0]`.  The existence of `schema.json` isn't sufficient to tell if the `reset` succeeded.

mrunalp: done.

wking:
I'm not familiar with bats, and I haven't checked the Godeps grep.
But other than that, 10d7cfb5 looks good to me.

crosbymichael:
LGTM

tests/integration/spec.bats
```diff
@@ -66,12 +66,15 @@ function teardown() {
...
+  TESTDIR=$(pwd)
   cd "$HELLO_BUNDLE"
-  # note this test runs from the temporary bundle directory not the integration root
-  # note this test is brittle against specs changes that lead runc's spec command
-  # todo get the validate program, gojsonschema, and schema/*s.json from godeps?
 
   run git clone https://github.com/opencontainers/runtime-spec.git src/runtime-spec
+  [ "$status" -eq 0 ]
+
+  SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
```
cyphar: I'm guessing we can't get these files from `Godep/*` because we either clean them or it's in a directory not easily accessible by `bats`?

cyphar:
LGTM, but we need to fix this in a more nice way (`-A 4` seems quite brittle to me).

tests/integration/spec.bats
mrunalp: Yeah, those files aren't in Godeps. We can clean up -A 4 with either python or jq. I'll create a follow up. Wanted to fix the build fast.

PR #904, opened 2016-06-10
cgroups: Fix issue if cgroup path contains :

euank:
Prior to this change, a cgroup with a `:` character in its path was not parsed correctly.

See https://github.com/kubernetes/kubernetes/issues/26903 for discussion of an issue where this actually caused a problem.

Happy to add more test cases if desired. 
cc @philips

### DIFF
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -4,6 +4,8 @@ package cgroups
 
 import (
 	"bytes"
+	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


philips:
@mjg59 @crawford we may need to backport this into docker in CoreOS as well.

cyphar:
LGTM

libcontainer/cgroups/utils_test.go
```diff
@@ -190,3 +192,56 @@ func BenchmarkGetCgroupMounts(b *testing.B) {
...
+			// Taken from a CoreOS instance running systemd 225 with CPU/Mem
+			// accounting enabled in systemd
+			input: `9:blkio:/
+8:freezer:/
+7:perf_event:/
+6:devices:/system.slice/system-sshd.slice
+5:cpuset:/
+4:cpu,cpuacct:/system.slice/system-sshd.slice/sshd@126-10.240.0.15:22-xxx.yyy.zzz.aaa:33678.service
+3:net_cls,net_prio:/
+2:memory:/system.slice/system-sshd.slice/sshd@126-10.240.0.15:22-209.251.245.232:33678.service
```
derekwaynecarr: Not that it matters for this test, but I assume memory was equal to CPU and name=systemd and you doctored it for the test?

cyphar: Heh, it looks like someone left their IP in the tests. The actual values don't need to be "real" -- it's a unit test after all we just have to check that the edge cases are properly dealt with.

euank: Indeed they were equal and it was doctored to remove my public ip (not that it really matters). Edited the remaining instance of it for consistency.

I prefer using close-to-real data for unit tests here since it's easy to do so.

cyphar:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/904/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

hqhq:
LGTM

PR #906, opened 2016-06-11
Add "--" exec cli support for command arguments

TristanCacqueray:
This flag allows using parameters starting with a - such as:
  ps --args -afx

### DIFF
--- a/man/runc-exec.8.md
+++ b/man/runc-exec.8.md
@@ -2,7 +2,7 @@
    runc exec - execute new process inside the container
 
 # SYNOPSIS
-   runc exec [command options] <container-id> <container command>
+   runc exec [command options] <container-id> -- <container command> [args...]
 
 Where "<container-id>" is the name for the instance of the container and
 "<container command>" is the command to be executed in the container.
--- a/exec.go
+++ b/exec.go
@@ -17,7 +17,7 @@ import (
 var execCommand = cli.Command{
 	Name:  "exec",
 	Usage: "execute new process inside the container",
-	ArgsUsage: `<container-id> <container command>
+	ArgsUsage: `<container-id> -- <container command> [command options]
 
 Where "<container-id>" is the name for the instance of the container and
 "<container command>" is the command to be executed in the container.
@@ -149,6 +149,9 @@ func getProcess(context *cli.Context, bundle string) (*specs.Process, error) {
 	}
 	p := spec.Process
 	p.Args = context.Args()[1:]
+	if len(p.Args) > 1 && p.Args[0] == "--" {
+		p.Args = p.Args[1:]
+	}
 	// override the cwd, if passed
 	if context.String("cwd") != "" {
 		p.Cwd = context.String("cwd")


cyphar:
I'd prefer if we fix the flag library we're using to support standard Unix `--` semantics, so usage of the library would allow (by default):

```
% runc exec test -- ls -la
```

TristanCacqueray:
I agree this is sub-optimal compared to "--" support. Though this PR enables use-cases when you need to exec command with parameters prefixed by "-".

crosbymichael:
@TristanCacqueray can you add the same logic that is happening here https://github.com/opencontainers/runc/pull/933 to support this without a `--args` flag?

TristanCacqueray:
@crosbymichael sure thing, it's looking better now. Thank you for the review.

mrunalp:
I agree with @cyphar that we should implement standard unix semantics for `--` or switch to another library that supports it.

TristanCacqueray:
@mrunalp it seems like the actual cli library do support "--" and doesn't interpret dashed argument placed after. However it's up to exec.go to remove the extra Args from the list, see 619a523

crosbymichael:
LGTM

@mrunalp it does support it, we just have to get the subslice because we use arg 1 as the container id

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/906/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

man/runc-exec.8.md
```diff
@@ -2,7 +2,7 @@
    runc exec - execute new process inside the container
 
 # SYNOPSIS
-   runc exec [command options] <container-id> <container command>
+   runc exec [command options] <container-id> <container command> [-- command options]
```
dqminh: this usage is  abit weird for `--`. I would expect `runc exec <container-id> -- command options

mrunalp:
To be consistent with other Linux utilities, I think it should support this:
```
[root@dhcp-16-129 ~]# runc exec 2345 -- ls -la
exec failed: exec: "--": executable file not found in $PATH
```
But instead it supports this:
```
[root@dhcp-16-129 ~]# runc exec 2345  ls -- -la
total 124
drwxrwxr-x   21 1000     1000          4096 Aug 25 16:43 .
drwxrwxr-x   21 1000     1000          4096 Aug 25 16:43 ..
-rw-------    1 root     root           156 Jun 10 22:28 .ash_history
drwxrwxr-x    2 1000     1000          4096 Feb 26  2015 bin
drwxr-xr-x    2 1000     1000          4096 Apr 19 17:02 data
drwxr-xr-x    5 root     root           360 Aug 25 16:43 dev
drwxr-xr-x    6 1000     1000          4096 Jul 12 21:57 etc
drwxr-xr-x    2 root     root          4096 Apr 14 18:33 foo
drwxrwxr-x    3 1000     1000          4096 Dec  1  2014 home
drwxrwxr-x    2 1000     1000          4096 Feb 26  2015 lib
lrwxrwxrwx    1 1000     1000             3 Feb 26  2015 lib64 -> lib
lrwxrwxrwx    1 1000     1000            11 Feb 26  2015 linuxrc -> bin/busybox
drwxrwxr-x    2 1000     1000          4096 Dec  1  2014 media
drwxrwxr-x    2 1000     1000          4096 Dec  1  2014 mnt
drwxr-xr-x    2 root     root          4096 Jun 24 23:18 mydata
-rw-------    1 root     root             0 Aug  5 17:17 nohup.out
drwxrwxr-x    2 1000     1000          4096 Dec  1  2014 opt
dr-xr-xr-x  371 root     root             0 Aug 25 16:43 proc
drwx------    2 1000     1000          4096 Mar 18 20:01 root
lrwxrwxrwx    1 1000     1000             3 Dec  1  2014 run -> tmp
drwxr-xr-x    2 1000     1000          4096 Feb 26  2015 sbin
drwxr-xr-x    2 root     root          4096 Jul 12 21:49 srcdata
-rw-r--r--    1 root     root         47858 May 13 19:43 strace.log
dr-xr-xr-x   13 root     root             0 Aug 22 17:08 sys
drwxrwxrwt    3 1000     1000          4096 Feb 26  2015 tmp
drwxrwxr-x    6 1000     1000          4096 Feb 26  2015 usr
drwxrwxr-x    4 1000     1000          4096 Feb 26  2015 var
```

man/runc-exec.8.md
crosbymichael: I think this help string is just messed up, it should work the way you described.  @TristanCacqueray can you fix the usage strings?

dqminh: no it doesnt because `p.Args = append(p.Args[:1], p.Args[2:]...)`. I think @mrunalp also pasted an example :D

crosbymichael: ops, i missed that

TristanCacqueray:
@crosbymichael @mrunalp makes sense, I misread the logic of #933...
Thank you for the review, the last patch now support "runc exec 2345 -- ls -la" instead.
Note that the patch doesn't enforce usage of the "--" separator and it checks for its presence before shifting the process args.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #908, opened 2016-06-13
Removing unused variable for cgroup subsystem

rajasec:
Removed the unused variable, as we are returning nil for subsystems( freezer & devices) when there is no statistics.


dqminh:
LGTM

mrunalp:
LGTM

PR #912, opened 2016-06-14
Fix fifo usage with userns and not root users

crosbymichael:

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -159,16 +159,34 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := l.Validator.Validate(config); err != nil {
 		return nil, newGenericError(err, ConfigInvalid)
 	}
+	uid, err := config.HostUID()
+	if err != nil {
+		return nil, newGenericError(err, SystemError)
+	}
+	gid, err := config.HostGID()
+	if err != nil {
+		return nil, newGenericError(err, SystemError)
+	}
 	containerRoot := filepath.Join(l.Root, id)
 	if _, err := os.Stat(containerRoot); err == nil {
 		return nil, newGenericError(fmt.Errorf("container with id exists: %v", id), IdInUse)
 	} else if !os.IsNotExist(err) {
 		return nil, newGenericError(err, SystemError)
 	}
-	if err := os.MkdirAll(containerRoot, 0700); err != nil {
+	if err := os.MkdirAll(containerRoot, 0711); err != nil {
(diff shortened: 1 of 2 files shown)


libcontainer/factory_linux.go
```diff
@@ -159,16 +159,34 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
...
-	if err := os.MkdirAll(containerRoot, 0700); err != nil {
+	if err := os.MkdirAll(containerRoot, 0711); err != nil {
+		return nil, newGenericError(err, SystemError)
+	}
+	if err := os.Chown(containerRoot, uid, gid); err != nil {
+		return nil, newGenericError(err, SystemError)
+	}
+	fifoName := filepath.Join(containerRoot, execFifoFilename)
+	oldMask := syscall.Umask(0000)
+	if err := syscall.Mkfifo(fifoName, 0666); err != nil {
```
cyphar: Doesn't this mean that any user can start a container created by root? This doesn't make sense to me IMO -- what's the usecase? With the rootless container setup, my hope was that each user's containers could only be controlled by them (with root being a special case) -- this means that any user will be able to start any other users' containers. Why is this necessary? That's why I liked the signal setup, because it retained that access control.

crosbymichael: I'll change this to `0622` so that only root can read and unblock the container's process

mlaventure: Isn't the uid at that point basically 0,0 ?

Meaning that, in a userns context, the user at the time of writing to the fifo would be matched against "other".

cyphar: No, because we chown the FIFO when creating the container (it's the next line after this). If we didn't chown it, then we wouldn't have any guarantee that the UID would be mapped (and thus nobody in the container could read from the socket).

crosbymichael:
@LK4D4 @mrunalp can you please look at this PR.  It fixes permissions errors with the fifo under userns and non root users in regular containers on some machines with umask issues

mrunalp:
@crosbymichael Yeah, will review it shortly.

LK4D4:
LGTM

cyphar:
I still had some open concerns about the permissions:

> Since we `chown`ing the `FIFO`, surely we can just make it `0600`? Why do other users need to be able to write to the `FIFO`?

> No, because we chown the FIFO when creating the container (it's the next line after this). If we didn't chown it, then we wouldn't have any guarantee that the UID would be mapped (and thus nobody in the container could read from the socket).

PR #914, opened 2016-06-14
Introduce make release

### LINKED ISSUE #899: binaries released should be static?
Hey hey!
Cool new releases, was wondering if the binaries you upload could have the BUILDTAGS="apparmor seccomp" and be static :D

zhaoleidd:
So we can make all types of release binary with combination
of following flags:
 seccomp
 selinux
 apparmor
 static

All binary files are put in release/ dir, like:
 [root@zlosvm1 runc]# ls -l release
 total 53488
 -rwxr-xr-x. 1 root root 9517965 Jun 14 18:11 runc
 -rwxr-xr-x. 1 root root 9673533 Jun 14 18:10 runc.seccomp
 -rwxr-xr-x. 1 root root 9705839 Jun 14 18:09 runc.seccomp.selinux
 -rwxr-xr-x. 1 root root 9546175 Jun 14 18:10 runc.selinux
 -rwxr-xr-x. 1 root root 8169190 Jun 14 18:10 runc.selinux.static
 -rwxr-xr-x. 1 root root 8150060 Jun 14 18:11 runc.static
 ...

Closes #899

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,13 +20,43 @@ MAN_PAGES = $(shell ls $(MAN_DIR)/*.8)
 MAN_PAGES_BASE = $(notdir $(MAN_PAGES))
 MAN_INSTALL_PATH := ${PREFIX}/share/man/man8/
 
+RELEASE_DIR := $(CURDIR)/release
+
 VERSION := ${shell cat ./VERSION}
+
+SHELL ?= $(shell command -v bash 2>/dev/null)
 
 all: $(RUNC_LINK)
 	go build -i -ldflags "-X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -tags "$(BUILDTAGS)" -o runc .
 
 static: $(RUNC_LINK)
 	CGO_ENABLED=1 go build -i -tags "$(BUILDTAGS) cgo static_build" -ldflags "-w -extldflags -static -X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -o runc .
+
+release: $(RUNC_LINK)
+	@flag_list=(seccomp selinux apparmor static); \
+	unset expression; \
+	for flag in "$${flag_list[@]}"; do \
+		expression+="' '{'',$${flag}}"; \
+	done; \
+	eval profile_list=("$$expression"); \
+	for profile in "$${profile_list[@]}"; do \
+		output=${RELEASE_DIR}/runc; \
+		for flag in $$profile; do \
+			output+=."$$flag"; \
+		done; \
+		tags="$$profile"; \
+		ldflags="-X main.gitCommit=${COMMIT} -X main.version=${VERSION}"; \
+		CGO_ENABLED=; \
+		[[ "$$profile" =~ static ]] && { \
+			tags="$${tags/static/static_build}"; \
+			tags+=" cgo"; \
+			ldflags+=" -w -extldflags -static"; \
+			CGO_ENABLED=1; \
+		}; \
+		echo "Building target: $$output"; \
+		rm -rf "${GOPATH}/pkg"; \
+		go build -i -ldflags "$$ldflags" -tags "$$tags" -o "$$output" .; \
+	done
 
 $(RUNC_LINK):
 	ln -sfn $(CURDIR) $(RUNC_LINK)
@@ -88,6 +118,7 @@ clean:
 	rm -f runc
 	rm -f $(RUNC_LINK)
 	rm -rf $(GOPATH)/pkg
+	rm -rf $(RELEASE_DIR)
 
 validate:
 	script/validate-gofmt
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ vendor/pkg
 /runc
 Godeps/_workspace/src/github.com/opencontainers/runc
 man/man8
+release


zhaoleidd:
ping

cyphar:
I'm not _really_ sure why we should add something like this to our Makefile? Maybe it's a distribution-specific thing but in openSUSE we don't really do things like this (though, you could in principle create a new `rpm` package for each of the binaries we generate).

hqhq:
I don't object, it SGTM, @mrunalp @crosbymichael @dqminh WDYT?

cyphar:
I don't really like having to add a new iteration layer each time we add a new buildtag. Maybe the combinatorial stuff should be defined as a single list that we then iterate over. Is there a nicer way of doing this?

I just re-read #899 and I agree with the idea behind this PR, we just need to make sure that it doesn't mean we have to keep adding more layers of ifs.

crosbymichael:
@zhaoleidd is there a way to simplify the code a little more like some of @cyphar 's suggestions?

zhaoleidd:
Hi, cyphar and crosbymichael,
Patch is updated to remove multi-for-loop based on @cyphar 's suggestions.

crosbymichael:
LGTM

hqhq:
Is this really a cleaner approach? With the former approach, we can only add one line if we add a build tag and now we have to add a lot more.

cyphar:
Yeah, I agree with @hqhq that this isn't a cleaner approach. What I was hoping for was the shell equivalent of doing a chained cartesian product of `(enable, disable)` for each feature. But you probably can't do it nicely in `sh` so I prefer the old approach over this one (in the absence of alternatives).

zhaoleidd:
Actually, shell script can do chained cartesian product.
I'll update patch based on @hqhq and @cyphar's suggestion.

zhaoleidd:
Hi, @hqhq and @cyphar 
I updated the patch to use cartesian product instead of array.

Makefile
```diff
@@ -28,6 +28,32 @@ all: $(RUNC_LINK)
 static: $(RUNC_LINK)
 	CGO_ENABLED=1 go build -i -tags "$(BUILDTAGS) cgo static_build" -ldflags "-w -extldflags -static -X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -o runc .
 
+release: $(RUNC_LINK)
+	@flag_list=(seccomp selinux apparmor static); \
```
cyphar: This looks quite nice IMO. But maybe we should move this to an actual `.sh` script in `script/...` and call it from the Makefile (so we can avoid the `; \`, `$$` and other ugliness)?

zhaoleidd: Put them into a separate script will make code clean, but the script need to access some variable in Makefile($COMMIT, $VERSION), we have some way to do it, but each way will make code complex:
1: Pass these variable as script's argument, or export them for script
    we need update the argument when we add/modify variable.
2: Let the script parse these variable in Makefile
    complex
3: clone these variable into script
    hard to maintenance
So, maybe put the ugly script into Makefile seems better than other selection...

cyphar: Yeah, I see your point. Alright, this is fine for me as-is.

cyphar:
Can you please remove the changes you made to the two `Godep/...` files (since they don't appear to be related to this PR)?

zhaoleidd:
Hi, @cyphar 

Patch updated to keep Godep/... unmodified, thanks!

cyphar:
LGTM.

/cc @hqhq @crosbymichael

hqhq:
@zhaoleidd I'm getting error:
```
$ make release
/bin/sh: 1: Syntax error: "(" unexpected
make: *** [release] Error 2
```

I have to add `SHELL := /bin/bash` to fix this error.

hqhq:
@zhaoleidd Also we should add `release` to `.gitignore` and maybe remove it in `make clean`.

zhaoleidd:
Hi, @hqhq

Thanks for your good suggestion, patch updated.

Makefile
```diff
@@ -20,14 +20,44 @@ MAN_PAGES = $(shell ls $(MAN_DIR)/*.8)
 MAN_PAGES_BASE = $(notdir $(MAN_PAGES))
 MAN_INSTALL_PATH := ${PREFIX}/share/man/man8/
 
+RELEASE_DIR := $(CURDIR)/release
+
 VERSION := ${shell cat ./VERSION}
 
+SHELL := /bin/bash
```
hqhq: @cyphar doesn't use bash, as well as some other users, so I'm not sure if this is the right approach, I would rather we figure out why the command needs bash, maybe we should write some POSIX-shell-compatible commands.

Anyway if it runs well on machines without bash, I'm ok we merge it for now, @cyphar can you confirm it?

BTW, I would use `SHELL  ?= $(shell command -v bash 2>/dev/null)` if we want to define shell.

zhaoleidd:
In result of "grep -w bash * -R | grep -v 'Binary file'",
we can see bash is used in scripts in following dir:
script/
tests/integration/
man/

Current runc already have some relationship with bash, or we can say, bash is already necessary component for runc dev.

Makefile
zhaoleidd: I think maybe because she script is using array.

In result of "grep -w bash * -R | grep -v 'Binary file'",
we can see bash is used in scripts in following dir:
script/
tests/integration/
man/

Current runc already have some relationship with bash,
or we can say, bash is already necessary component for runc dev,
maybe we can use bash freely in current Makefile.

What is your opinion?

zhaoleidd:
Hi, Patch updated.

Changelog:
  - SHELL ?= $(shell command -v bash 2>/dev/null)
  + SHELL ?= $(shell command -v bash 2>/dev/null)

based on @hqhq 's suggestion.

hqhq:
LGTM ping @cyphar

cyphar:
Looking ...

cyphar:
LGTM

PR #918, opened 2016-06-15
cli: Workaround for parse -- in argument

### LINKED ISSUE #752: runc exec command parsing
Using runc exec fails when "container command" need an option.

For example : 

```
cmd="ls -l"
runc exec ${containerid} ${cmd}
```

gives

```
Incorrect Usage.

NAME:

USAGE:
   runc exec [command options] <container-id> <container command>

"<container command>" is the command to be executed in the container.

For example, if the container is configured to run the linux ps command the
following will output a list of processes running in the container:

       # runc exec <container-id> ps

OPTIONS:
   --console                    specify the pty slave path for use with the container
   --cwd                    current working directory in the container
   --env, -e [--env option --env option]    set environment variables
   --tty, -t                    allocate a pseudo-TTY
   --user, -u                   UID (format: <uid>[:<gid>])
   --process, -p                path to the process.json
   --detach, -d                 detach from the container's process
   --pid-file                   specify the file to write the process id to
   --process-label              set the asm process label for the process commonly used with selinux
   --apparmor                   set the apparmor profile for the process
   --no-new-privs               set the no new privileges value for the process
   --cap, -c [--cap option --cap option]    add a capability to the bounding set for the process
   --no-subreaper               disable the use of the subreaper used to reap reparented processes

flag provided but not defined: -l
```

zhaoleidd:
CLI library add '--' into argument list, and make following
result in current code:
 #  ./runc exec 123 -- ls
 exec failed: exec: "--": executable file not found in $PATH
 #  ./runc exec 123 -- ls -l
 exec failed: exec: "--": executable file not found in $PATH
 #  ./runc exec 123 ls
 bin   dev   etc   home  proc  root  sys   tmp   usr   var
 #  ./runc exec 123 ls -l
 Incorrect Usage.
 NAME:
 ...
 flag provided but not defined: -l
 #

Before CLI library got fixed, we can use a workaround to make user
feeling better:
 # ./runc exec 123 -- ls
 bin   dev   etc   home  proc  root  sys   tmp   usr   var
 # ./runc exec 123 -- ls -l
 total 16
 drwxr-xr-x    2 root     root          8192 Mar 18 16:39 bin
 drwxr-xr-x    5 root     root           360 Jun 15 10:06 dev
 ...
 #
 # ./runc exec 123 ls
 bin   dev   etc   home  proc  root  sys   tmp   usr   var
 # ./runc exec 123 ls -l
 Incorrect Usage.
 ...
 flag provided but not defined: -l
 #

And the old argument type ("--" before container-id) is also
supported after this patch.

Closes #752

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -382,7 +382,6 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, 
 func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
 	cfg := &initConfig{
 		Config:           c.config,
-		Args:             process.Args,
 		Env:              process.Env,
 		User:             process.User,
 		Cwd:              process.Cwd,
@@ -395,6 +394,11 @@ func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
 		ProcessLabel:     c.config.ProcessLabel,
 		Rlimits:          c.config.Rlimits,
 		ExecFifoPath:     filepath.Join(c.root, execFifoFilename),
+	}
+	if process.Args[0] == "--" {
+		cfg.Args = process.Args[1:]
+	} else {
+		cfg.Args = process.Args
 	}
 	if process.NoNewPrivileges != nil {
 		cfg.NoNewPrivileges = *process.NoNewPrivileges


zhaoleidd:
For the same reason why #933 is merged, could you consider review this patch? @hqhq

cyphar:
We're probably going to end up merging #1007 instead.

cyphar:
I believe this has been fixed by some other changes recently, as `runc exec test sh -c 'echo haha'` works now. Can you confirm whether this patch is still necessary can close this issue as appropriate?

hqhq:
I think we can close this now.

PR #919, opened 2016-06-17
Fix ps argument manual

zhaoleidd:
Argument of "ps options" for ps command is a optional parameter.
Should use [] instead of <> in manual.

### DIFF
--- a/man/runc-ps.8.md
+++ b/man/runc-ps.8.md
@@ -2,7 +2,7 @@
    runc ps - ps displays the processes running inside a container
 
 # SYNOPSIS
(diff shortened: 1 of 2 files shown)


cyphar:
@hqhq does this cause issues with our cli library?

hqhq:
We'll assign default options if they are absent, so it is optional now.

LGTM

mrunalp:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/919/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

PR #923, opened 2016-06-19
Fix FreeBSD build

0xffea:
Only makes "go build" compile without error on FreeBSD.

### DIFF
--- a/libcontainer/container_freebsd.go
+++ b/libcontainer/container_freebsd.go
@@ -1,0 +1,81 @@
+package libcontainer
+
+import (
+	"os"
+
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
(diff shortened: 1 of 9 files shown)


justincormack:
I am curious why? Shouldn't FreeBSD have its own OCI runtime, as it is completely different internally?

cyphar:
@justincormack We have separate builds for Solaris. In principle, I have no problem with runC supporting other platforms (though @crosbymichael might disagree).

crosbymichael:
I don't have a problem with it but this PR does not add any functionality for freebsd so it does make it build but you end up with a busted binary so I don't see a point in merging this PR as is without having full support.

@0xffea let me know if you have alternate reasons for making the build work without the functionality. 

Thanks!

PR #925, opened 2016-06-21
rootfs: clean up

cyphar:
Comparisons with paths aren't really a good idea unless you're
guaranteed that the comparison will work will all paths that resolve to
the same lexical path as the compared path.

There's no point in changing directory here. Syscalls are resolved local
to the linkpath, not to the current directory that the process was in
when creating the hard link. Changing directories just confuses people
who are trying to debug things.


mrunalp:
LGTM

LK4D4:
LGTM

PR #927, opened 2016-06-25
add ability to force removal of running container via runc delete -f

WPH95:
Add a` --force` flag in `runc delete`
Add ability to remove a running container in a single command, e.g.
```
# runc delete $container_id
Impossible to remove a running container, please stop it first or use -f

# runc delete -f $container_id
```
Signed-off-by: wph95 <wph657856467@gmail.com>

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -26,6 +26,12 @@ status of "ubuntu01" as "stopped" the following will delete resources held for
 "ubuntu01" removing "ubuntu01" from the runc list of containers:  
 	 
        # runc delete ubuntu01`,
+	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name:  "force, f",
+			Usage: "Force the removal of a running container (uses SIGKILL)",
+		},
+	},
 	Action: func(context *cli.Context) error {
 		container, err := getContainer(context)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


delete.go
```diff
@@ -56,8 +62,18 @@ status of "ubuntu01" as "stopped" the following will delete resources held for
 				}
 			}
 			return fmt.Errorf("container init still running")
+		case libcontainer.Running:
+			force := context.Bool("force")
+			if force {
+				if err := container.Signal(syscall.SIGKILL); err != nil {
+					return fmt.Errorf("Could not stop running container, cannot remove - %v", err)
+				}
+				destroy(container)
```
wking: The reason why the spec does not require “delete a running container” to work is that “I sent it a kill signal” doesn't mean that it's dead (it could be in an uninterruptible sleep disk sleep, see proc(5)), and the delete process has no way to block on the container process's actual exit.  See [here][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/384#discussion_r64801501

mrunalp:
Can we close this now that #928 is merged?

dqminh:
i think so. Thanks @WPH95 for your changes !

PR #928, opened 2016-06-28
Add force to delete

mlaventure:
Reasons:
- I'm lazy and I don't want to type 2 different commands :). Typing `--force` actually make it explicit that I know what I'm doing
- When using an higher level tool to control runc, it's cheaper to do only 1 exec instead of 2 (kill + delete)

P.S: This also update the `Makefile` to tag the images generated via `make` with the current branch name instead of always overriding `latest`. I can move that commit in a separate PR if needed.

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -13,6 +13,18 @@ import (
 	"github.com/urfave/cli"
 )
 
+func killContainer(container libcontainer.Container) error {
+	container.Signal(syscall.SIGKILL)
+	for i := 0; i < 100; i++ {
+		time.Sleep(100 * time.Millisecond)
+		if err := container.Signal(syscall.Signal(0)); err != nil {
+			destroy(container)
+			return nil
+		}
+	}
+	return fmt.Errorf("container init still running")
+}
+
 var deleteCommand = cli.Command{
 	Name:  "delete",
 	Usage: "delete any resources held by the container often used with detached containers",
@@ -23,9 +35,15 @@ Where "<container-id>" is the name for the instance of the container.
 EXAMPLE:
 For example, if the container id is "ubuntu01" and runc list currently shows the
 status of "ubuntu01" as "stopped" the following will delete resources held for
-"ubuntu01" removing "ubuntu01" from the runc list of containers:  
-	 
+"ubuntu01" removing "ubuntu01" from the runc list of containers:
+
        # runc delete ubuntu01`,
+	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name:  "force, f",
+			Usage: "Forcibly kills the container if it is still running",
+		},
+	},
 	Action: func(context *cli.Context) error {
 		container, err := getContainer(context)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


wking:
Duplicates #927?

mlaventure:
It is, will just move the image tag in a different PR then.

hqhq:
Yeah, would be better to separate them.

delete.go
```diff
@@ -43,21 +61,17 @@ status of "ubuntu01" as "stopped" the following will delete resources held for
...
-		case libcontainer.Created:
-			container.Signal(syscall.SIGKILL)
-			for i := 0; i < 100; i++ {
-				time.Sleep(100 * time.Millisecond)
-				if err := container.Signal(syscall.Signal(0)); err != nil {
-					destroy(container)
-					return nil
-				}
+		if context.Bool("force") {
+			killContainer(container)
```
hqhq: Looks weird to send signal to a stopped container, though it won't cause any errors.

wking: On Tue, Jun 28, 2016 at 07:51:47PM -0700, Qiang Huang wrote:
> > +		if context.Bool("force") {
> > +			killContainer(container)
>
> Looks weird to send signal to a stopped container, though it won't
> cause any errors.

Unless you hit #871 (which I think still needs to be fixed [1]).

[1]: https://github.com/opencontainers/runc/issues/871#issuecomment-225686558

hqhq: Right, I've reopened #871 , thanks.

delete.go
```diff
@@ -43,21 +61,21 @@ status of "ubuntu01" as "stopped" the following will delete resources held for
...
+			}
+		} else {
+			switch s {
+			case libcontainer.Stopped:
+				destroy(container)
+			case libcontainer.Created:
+				return killContainer(container)
+			default:
+				return fmt.Errorf("cannot delete container that is not stopped: %s", s)
 			}
```
hqhq: What about:
```
switch s {
case libcontainer.Stopped:
        return destroy(container)
case libcontainer.Created:
        return killContainer(container)
case libcontainer.Running:
        if context.Bool("force") {
                return killContainer(container)
        }
        return fmt.Errorf("cannot delete container that is not stopped: %s", s)
default:
        return fmt.Errorf("cannot delete container that is not stopped: %s", s)
}
```

mlaventure: What about a paused or pausing container?

mlaventure: maybe have:
```
switch s {
case libcontainer.Stopped:
        return destroy(container)
default:
        if context.Bool("force") {
                return killContainer(container)
        }
        return fmt.Errorf("cannot delete container that is not stopped: %s", s)
}
```
?

hqhq: Looks good.

hqhq: Should keep created state, it doesn't acquire force delete.

mlaventure: @hqhq it does, the container is up and running after `runc create` and waiting for a `runc start` to exec the user process

hqhq:
LGTM

dqminh:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/928/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

PR #930, opened 2016-06-30
Make state detection precise

hqhq:
Fixes: https://github.com/opencontainers/runc/issues/871

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -22,6 +22,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/criurpc"
+	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/syndtr/gocapability/capability"
 	"github.com/vishvananda/netlink/nl"
@@ -30,18 +31,19 @@ import (
 const stdioFdCount = 3
 
 type linuxContainer struct {
-	id            string
-	root          string
-	config        *configs.Config
-	cgroupManager cgroups.Manager
-	initPath      string
-	initArgs      []string
-	initProcess   parentProcess
-	criuPath      string
-	m             sync.Mutex
-	criuVersion   int
-	state         containerState
-	created       time.Time
+	id                   string
+	root                 string
+	config               *configs.Config
+	cgroupManager        cgroups.Manager
+	initPath             string
+	initArgs             []string
+	initProcess          parentProcess
+	initProcessStartTime string
+	criuPath             string
+	m                    sync.Mutex
+	criuVersion          int
+	state                containerState
+	created              time.Time
 }
 
 // State represents a running container's state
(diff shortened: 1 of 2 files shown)


libcontainer/container_linux.go
```diff
@@ -1124,6 +1133,16 @@ func (c *linuxContainer) runType() (Status, error) {
...
 	}
+	// check if the init process is still the same process as the initial one,
+	// it could happen that the process is exited and a new process created with
+	// the same pid, in this case, the container would already be stopped.
+	startTime, err := system.GetProcessStartTime(pid)
+	if err != nil {
+		return Stopped, newSystemErrorWithCausef(err, "getting init process %d start time", pid)
+	}
+	if c.initProcessStartTime != startTime {
+		return Stopped, nil
```
wking: This was my suggestion, but @cyphar suggested [stat'ing `/proc/{pid}`][1] (and using the inode?).  In either case, it may be better to abstract this extra identifier out into a `GetProcessUniquifier` helper (silly name, but whatever ;), since we're only using it for an equality comparison, and don't really care what the value actually means.

[1]: https://github.com/opencontainers/runc/issues/871#issuecomment-223636697

cyphar: I looked into it, and doing a `stat` is not guaranteed to give you a unique inode number (as far as I can tell, inode numbers in procfs are taken from a pool of unused inode numbers). So I think start time (while ugly) is better. But I agree that we should abstract the actual value.

libcontainer/container_linux.go
```diff
@@ -1124,6 +1133,16 @@ func (c *linuxContainer) runType() (Status, error) {
 		}
 		return Stopped, newSystemErrorWithCausef(err, "sending signal 0 to pid %d", pid)
 	}
+	// check if the init process is still the same process as the initial one,
+	// it could happen that the process is exited and a new process created with
```
mrunalp: the original process has exited and a new process has been created with..

hqhq:
@wking @cyphar @mrunalp Updated.

libcontainer/container_linux.go
```diff
@@ -1109,6 +1118,17 @@ func (c *linuxContainer) refreshState() error {
 	return c.state.transition(&stoppedState{c: c})
 }
 
+func (c *linuxContainer) isProcessUnique(pid int) (bool, error) {
```
wking: `doesProcessExist` is probably a better name for this function.  Processes are unique, and with this function we're just doing a better (but still not perfect) job of identifying whether a given /proc entry is the container process we're interested in.

libcontainer/container_linux.go
```diff
@@ -1124,6 +1144,17 @@ func (c *linuxContainer) runType() (Status, error) {
 		}
 		return Stopped, newSystemErrorWithCausef(err, "sending signal 0 to pid %d", pid)
 	}
+	// check if the init process is still the same process as the initial one,
+	// it could happen that the original process has exited and a new process
+	// has been created with the same pid, in this case, the container would
+	// already be stopped.
```
wking: I'd move this comment to before `isProcessUnique` (or whatever we want to call that function).  No need to interrupt the flow here.

libcontainer/container_linux.go
```diff
@@ -1124,6 +1144,17 @@ func (c *linuxContainer) runType() (Status, error) {
...
+	u, err := c.isProcessUnique(pid)
+	if err != nil {
+		return Stopped, err
+	}
+	if !u {
+		return Stopped, nil
```
wking: This is pretty much the same as the previous conditional, so you can collapse to:

    if !u || err != nil {
      return Stopped, err
    }

although if we rename `isProcessUnique` to `doesProcessExist` we probably want to rename `u` to `exists`.

hqhq:
@wking Updated, I've changed the function name to `doesInitProcessExist`, I think it's more accurate about what this function does.

wking:
On Mon, Jul 04, 2016 at 06:30:14PM -0700, Qiang Huang wrote:
> … I've changed the function name to `doesInitProcessExist`…

14e95b2 looks good to me.

dqminh:
LGTM

mrunalp:
LGTM

PR #933, opened 2016-07-06
cli: Workaround for ps's argument

### LINKED ISSUE #788: ps command with UNIX format
Not of bigger issue, minor issue while listing containers.

With current ps command in runc, it displays the processes running inside the container ( default option is -ef)
when option are specified, it gives error
runc ps containerID -eaf
flag not defined.

Currently by default runc supports ("-ef") UNIX format preceded by dash
when we give the options, we need to give as runc ps containerID eaf which is of BSD format ( without preceding dash)

zhaoleidd:
Currently, ps command can not support argument:
(But following usage is in manual)
 | # ./runc ps 123 -ef
 | Incorrect Usage.
 |
 | NAME:
 |    runc ps - ps displays the processes running inside a container
 |
 | USAGE:
 |    runc ps [command options] <container-id> [ps options]
 |
 | OPTIONS:
 |    --format value, -f value  select one of: table or json
 |
 | flag provided but not defined: -ef
 | #

Instead of using odd command like:
 | # ./runc ps -- 123 -ef

We can make it seems little better:
 | # ./runc ps 123 -- -ef
 | UID        PID  PPID  C STIME TTY          TIME CMD
 | root     29046 29038  0 11:18 pts/2    00:00:00 sh
 | #

This patch also fixed manual which can not working in current
code.

Closes #788

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -16,7 +16,7 @@ import (
 var psCommand = cli.Command{
 	Name:      "ps",
 	Usage:     "ps displays the processes running inside a container",
-	ArgsUsage: `<container-id> [ps options]`,
+	ArgsUsage: `<container-id> [-- ps options]`,
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:  "format, f",
@@ -43,6 +43,9 @@ var psCommand = cli.Command{
 		}
 
 		psArgs := context.Args().Get(1)
+		if psArgs == "--" {
+			psArgs = context.Args().Get(2)
+		}
 		if psArgs == "" {
 			psArgs = "-ef"
 		}
--- a/ps.go
+++ b/ps.go
@@ -42,15 +42,21 @@ var psCommand = cli.Command{
 			return nil
 		}
 
-		psArgs := context.Args().Get(1)
-		if psArgs == "--" {
-			psArgs = context.Args().Get(2)
-		}
-		if psArgs == "" {
-			psArgs = "-ef"
+		// [1:] is to remove command name, ex:
+		// context.Args(): [containet_id ps_arg1 ps_arg2 ...]
+		// psArgs:         [ps_arg1 ps_arg2 ...]
+		//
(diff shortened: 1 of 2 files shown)


ps.go
```diff
@@ -43,6 +43,9 @@ var psCommand = cli.Command{
 		}
 
 		psArgs := context.Args().Get(1)
+		if psArgs == "--" {
+			psArgs = context.Args().Get(2)
+		}
```
hqhq: What if people want to use `runc ps <id> -- -e -f`?

zhaoleidd: Yes, it is another thing need to be fixed, -- to let "runc ps" support multiple ps arguments.
I'll fix it on top of this patch.

zhaoleidd:
Hi, hqhq

I uploaded v2, and add support for multi-argument for ps command.
Any suggestion?

ps.go
```diff
@@ -42,12 +42,21 @@ var psCommand = cli.Command{
...
-		psArgs := context.Args().Get(1)
-		if psArgs == "" {
-			psArgs = "-ef"
+		// [1:] is to remove command name, ex:
+		// context.Args(): [containet_id ps_arg1 ps_arg2 ...]
+		// psArgs:         [ps_arg1 ps_arg2 ...]
+		//
+		psArgs := context.Args()[1:]
+
+		if len(psArgs) > 0 && psArgs[0] == "--" {
```
cyphar: Why is this not implemented in the library itself? Currently Docker is moving to Cobra, so maybe we should do the same if it means we don't have to embed these hacks into our client code.

crosbymichael:
LGTM

ps.go
hqhq: I have colleagues working on refactoring cli with Cobra, we can see how that works out. But for now, I think we can merge this PR.

hqhq:
LGTM

PR #934, opened 2016-07-06
Fix and refactor init args

macrosheep:
1. According to docs of Cmd.Path and Cmd.Args from package "os/exec":
   Path is the path of the command to run. Args holds command line
   arguments, including the command as Args[0]. We have mixed usage
   of args. In InitPath(), InitArgs only take arguments, in InitArgs(),
   InitArgs including the command as Args[0]. This is confusing.
2. InitArgs() already have the ability to configure a LinuxFactory
   with the provided absolute path to the init binary and arguements as
   InitPath() does.
3. exec.Command() will take care of serching executable path.
4. The default "/proc/self/exe" instead of os.Args[0] is passed to
   InitArgs in order to allow relative path for the runC binary.

Signed-off-by: Yang Hongyang <imhy.yang@gmail.com>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -35,7 +35,6 @@ type linuxContainer struct {
 	root                 string
 	config               *configs.Config
 	cgroupManager        cgroups.Manager
-	initPath             string
 	initArgs             []string
 	initProcess          parentProcess
 	initProcessStartTime string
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

dqminh:
@crosbymichael wasnt this done so that we can have reexec-style for those who embeded libcontainer ? Given that the embedding style is not supported anymore (?), i guess this change is simpler.

crosbymichael:
@dqminh maybe.  I think its still possible to do with the functions and providing an abs path to InitArgs

cyphar:
I don't think we should drop `InitPath`, since `exec.Command` resolves things using `$PATH` this might cause confusion to people who just do something silly like trying to set `$runtime` as the first argument. Not to mention that some runtimes might want to make `argv[0]` be how they communicate to themselves what mode they should run in (think `busybox`).

macrosheep:
@cyphar The InitPath in the codebase actually takes the full runc binary path, for example /home/xxx/runc, this usage is not compatible with "os/exec" package and is confusion. We could use InitArgs() function to do what InitPath() could do. I don't see why we need keep this. Ppl use libcontainer should be aware of what InitArgs() could do, I've also updated the comments of the InitArgs() function in this patch, it should be enough though IMO.

crosbymichael:
@cyphar PTAL

dqminh:
LGTM

PR #935, opened 2016-07-06
Use update time to detect if kmem limits have been set

### LINKED ISSUE #841: sysconfig_notcgo does not implement GetLongBit

vishh:
This PR should also fix #841 

xref: https://github.com/kubernetes/kubernetes/issues/27204

#841 needs to be fixed to make it possible to update libcontainer version in kubernetes.

cc @cyphar @crosbymichael @dubstack

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -5,15 +5,14 @@ package fs
 import (
 	"bufio"
 	"fmt"
-	"math"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
-	"github.com/opencontainers/runc/libcontainer/system"
 )
 
 type MemoryGroup struct {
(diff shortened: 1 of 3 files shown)


vishh:
cc @mrunalp

libcontainer/cgroups/fs/memory.go
```diff
@@ -28,53 +27,65 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+			return err
+		}
+	}
 	// We need to join memory cgroup after set memory limits, because
 	// kmem.limit_in_bytes can only be set when the cgroup is empty.
-	_, err = d.join("memory")
-	if err != nil && !cgroups.IsNotFound(err) {
+	if _, jerr := d.join("memory"); jerr != nil && !cgroups.IsNotFound(jerr) {
+		err = jerr
 		return err
```
mrunalp: could just return jerr

mrunalp:
Looks fine except for one nit.

libcontainer/cgroups/fs/memory.go
vishh: `err` needs to be set for the defer function to execute.

vishh:
ping @LK4D4 @cyphar @crosbymichael @dqminh

mrunalp:
LGTM

libcontainer/cgroups/fs/memory.go
```diff
@@ -28,53 +27,65 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
-		// We have to set kernel memory here, as we can't change it once
-		// processes have been attached to the cgroup.
-		if err := s.SetKernelMemory(path, d.config); err != nil {
-			return err
-		}
+	// reset error.
+	err = nil
+	if path == "" {
+		// Invalid input.
+		return fmt.Errorf("invalid path for memory cgroups: %+v", d)
```
hqhq: We don't do this in other subsystems because the only case that `path == ""` is when `err != nil` which will return in the above, so maybe remove this for consistency?

hqhq:
LGTM

PR #936, opened 2016-07-07
libcontainer: Add a helper func to set CriuPath

macrosheep:
Added a helper func to set CriuPath for LinuxFactory.

Signed-off-by: Yang Hongyang <imhy.yang@gmail.com>

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -103,6 +103,15 @@ func TmpfsRoot(l *LinuxFactory) error {
 	return nil
 }
 
+// CriuPath returns an option func to configure a LinuxFactory with the
+// provided criupath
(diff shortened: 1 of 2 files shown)


hqhq:
Sorry but I don't see the benefit, why would we do this?

macrosheep:
@hqhq thanks for your attention. I think CriuPath is like InitArgs and other members of linuxContainer struct, we'd better provide an option func within libcontainer package for users to set CriuPath like what we've done for InitArgs, cgroupManagers etc. Instead of let every libcontainer consumers to implement their own option func to directly access and set linuxContainer.CriuPath.

hqhq:
Sorry for the late. OK, although I think the option function for setting criuPath is simple enough, I don't mind adding this either.

LGTM

cyphar:
LGTM

PR #937, opened 2016-07-09
fix setting net_cls classid

hushan:
Setting classid of net_cls cgroup failed:

ERRO[0000] process_linux.go:291: setting cgroup config for ready process caused "failed to write 𐀁 to net_cls.classid: write /sys/fs/cgroup/net_cls,net_prio/user.slice/abc/net_cls.classid: invalid argument"
process_linux.go:291: setting cgroup config for ready process caused "failed to write 𐀁 to net_cls.classid: write /sys/fs/cgroup/net_cls,net_prio/user.slice/abc/net_cls.classid: invalid argument"

The classid is a *uint32, use strconv to convert to string.

Signed-off-by: Hushan Jia <hushan.jia@gmail.com>

### DIFF
--- a/libcontainer/configs/cgroup_unix.go
+++ b/libcontainer/configs/cgroup_unix.go
@@ -120,5 +120,5 @@ type Resources struct {
 	NetPrioIfpriomap []*IfPrioMap `json:"net_prio_ifpriomap"`
 
 	// Set class identifier for container's network packets
-	NetClsClassid string `json:"net_cls_classid"`
+	NetClsClassid uint32 `json:"net_cls_classid"`
 }
--- a/libcontainer/cgroups/fs/net_cls.go
+++ b/libcontainer/cgroups/fs/net_cls.go
@@ -3,6 +3,8 @@
 package fs
 
 import (
+	"strconv"
+
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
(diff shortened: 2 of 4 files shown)


cyphar:
I'd prefer if you fixed this in `libcontainer/cgroups/fs/net_cls.go` (by changing `writeFile` to `writeUint` or a similar util function).

hushan:
@cyphar `Cgroup.Resources.NetClsClassid` is of type `string`, `libcontainer/cgroups/fs/net_cls.go` won't have the correct data without this conversion.

cyphar:
Okay, then we should change `cgroup.Resources.NetClsClassid` to be a `uint32` (to match the spec). Looking through the git logs, the important commit is 9cd641486a5e46ac2dc98d5229856ea1205dfa57 (we changed the runtime-spec type but kept the runC one the same). IMO this looks like a mistake rather than an intended change (though we should probably see what @mrunalp thinks).

In any case, I'd prefer if we kept the internal meaning of "what a net_cls class ID is meant to be" to the cgroup code rather than touching it within `specconv`.

hushan:
Yeah, the type of `cgroup.Resources.NetClsClassid` does not match specs, I agree it's better to change to the same type.

libcontainer/configs/cgroup_unix.go
```diff
@@ -120,5 +120,5 @@ type Resources struct {
 	NetPrioIfpriomap []*IfPrioMap `json:"net_prio_ifpriomap"`
 
 	// Set class identifier for container's network packets
-	NetClsClassid string `json:"net_cls_classid"`
+	NetClsClassid *uint32 `json:"net_cls_classid"`
```
cyphar: We generally don't use pointers for the internal libcontainer configs. Though, I'm not sure what a `NetClsClassid` of `0` means so this might be fine. WDYT @mrunalp

cyphar: Okay, looking at the kernel docs the default `classid` is `0`. So IMO we can just use `!= 0` for `Set()` to decide whether or not to write the value.

hushan: Ok, I will update the patch.

mrunalp:
@cyphar The reason I kept it as string in that commit was because we couldn't break libcontainer which was used directly by docker back then.

cyphar:
Okay, thanks @mrunalp.

LGTM.

hushan:
PING anyone could do one more review?

mrunalp:
LGTM

Issue #24, opened 2015-06-24
The name of this project is unfortunate.


haf:
We live in an international world.

In the part of the world called Sweden, your project is named *masturbation*. C on its own in Swedish is pronounced k, and:

![screen shot 2015-06-27 at 20 55 27](https://cloud.githubusercontent.com/assets/193115/8393431/f56cfde2-1d0e-11e5-8737-3572488061f2.png)

The sells in the enterprises might not come. And then it would be over prematurely.

Just thought you wanted to know. ;)

shykes:
Ha ;) Well that's unfortunate. I think it could help if you spell it run-C with a dash maybe?

Also, it's ok if it's not an "enterprise-ready" name, it's meant to be low-level plumbing to power the internals of Docker. Enterprises care about Docker - and they will leave the plumbers alone to do the low-level coding here :)

keis:
if it's "run-see" that would get around the problem and anyway we learned to live with pull(a)

badbug0101:
@haf lol

rymohr:
Haha, some follow up commentary in an HN thread: https://news.ycombinator.com/item?id=9788058

errordeveloper:
What about a shorter reverse version `cr`?

omeid:
+1 for cr.

CameronNemo:
cr is too short, likely to conflict with something else.

errordeveloper:
I've checked, and it doesn't look like there's a command that is commonly found on Linux installations. There used to be one, but looks like it's [long gone](http://www.unix.com/man-page/debian/1/cr/), as the man page is dated to 1996 and is not like a POSIX command or anything.

errordeveloper:
It may be worth noting that `docker` command has conflicted with another from the start and we all got over it.

senlog:
how about "stevedore" or "wharfie" ?

crosbymichael:
I think we can close this issue now ;)

PR #940, opened 2016-07-11
version: Add a version command

wking:
And adjust `runc --version` to print exactly the same content.

The

```
$ COMMAND --version
```

interface is a popular one for this information and makes a lot of sense for single-action commands.  For example, [printf(1)] is only ever going to do one meaningful thing, so `--version`, `--help`, and other command-like actions work well as options.

Umbrella commands with multiple sub-commands, on the other hand, have a more consistent interface if “... and exit” actions get their own sub-commands.  That allows you to declare an interface like:

```
$ COMMAND [global-options] SUB-COMMAND [sub-command-specific-options] [sub-command-specific-arguments]
```

without having to say “except if you use a sub-command-like global option, in which case SUB-COMMAND is not allowed”.

With this commit, we add support for `version` while keeping `--version`, because while `version` makes more sense for umbrella commands, a user may not know if `runc` is an umbrella command when they ask for the version.

Existing umbrella commands are not particularly consistent:
- git(1) supports both `--version` and `version`, `--help` and `help`.
- btrfs(8) supports both `--version` and `version`, `--help` and `help`.
- ip(1) supports `-V` / `-Version` and `help`.
- npm supports `version` and `help`.
- pip supports `--version`, `--help` and `help`.

See [related discussion](https://github.com/opencontainers/runtime-spec/pull/513#discussion_r70104934) in opencontainers/runtime-spec#513.

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/urfave/cli"
+)
+
+// version will be populated by the Makefile, read from
+// VERSION file of the source code.
+var version = ""
+
+// gitCommit will be the hash that the binary was built from
+// and will be populated by the Makefile
+var gitCommit = ""
+
+var versionCommand = cli.Command{
+	Name:        "version",
+	Usage:       "output the runtime version",
+	Description: `The version command outputs the runtime version.`,
+	Action:      printVersion,
+}
+
+func printVersion(context *cli.Context) (err error) {
+	if version == "" {
+		_, err = fmt.Print("runc unknown\n")
+	} else {
+		_, err = fmt.Printf("runc %s\n", version)
+	}
+	if err != nil {
+		return err
+	}
+	if gitCommit != "" {
+		_, err = fmt.Printf("commit: %s\n", gitCommit)
+		if err != nil {
+			return err
+		}
+	}
+	_, err = fmt.Printf("spec: %s\n", specs.Version)
+	return err
+}
--- a/contrib/completions/bash/runc
+++ b/contrib/completions/bash/runc
@@ -760,6 +760,7 @@ _runc() {
 		start
 		state
 		update
+		version
 		help
 		h
 	)
--- a/main.go
+++ b/main.go
@@ -4,20 +4,10 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strings"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/urfave/cli"
 )
-
-// version will be populated by the Makefile, read from
-// VERSION file of the source code.
-var version = ""
-
-// gitCommit will be the hash that the binary was built from
-// and will be populated by the Makefile
-var gitCommit = ""
 
 const (
 	specConfig = "config.json"
@@ -50,16 +40,11 @@ func main() {
 	app := cli.NewApp()
 	app.Name = "runc"
 	app.Usage = usage
+	app.Version = ""
 
-	var v []string
-	if version != "" {
-		v = append(v, version)
+	cli.VersionPrinter = func(context *cli.Context) {
+		printVersion(context)
 	}
-	if gitCommit != "" {
-		v = append(v, fmt.Sprintf("commit: %s", gitCommit))
-	}
-	v = append(v, fmt.Sprintf("spec: %s", specs.Version))
-	app.Version = strings.Join(v, "\n")
 	app.Flags = []cli.Flag{
 		cli.BoolFlag{
 			Name:  "debug",
@@ -108,6 +93,7 @@ func main() {
 		startCommand,
 		stateCommand,
 		updateCommand,
+		versionCommand,
 	}
 	app.Before = func(context *cli.Context) error {
 		if context.GlobalBool("debug") {
--- a/man/runc-version.8.md
+++ b/man/runc-version.8.md
@@ -1,0 +1,14 @@
+# NAME
+   runc version - output the runtime version
+
+# SYNOPSIS
+   runc version
+
+Write version information to stdout and exit.
+
+# DESCRIPTION
+   The version command writes version information for runc and the
+supported spec release.  This information is also exposed through
+`runc --version`, and callers can use whichever is more convenient for
+them.
+
--- a/man/runc.8.md
+++ b/man/runc.8.md
@@ -44,6 +44,7 @@ value for "bundle" is the current directory.
    start        executes the user defined process in a created container
    state        output the state of a container
    update       update container resource constraints
+   version      output the runtime version
    help, h      Shows a list of commands or help for one command
    
 # GLOBAL OPTIONS
--- a/tests/integration/help.bats
+++ b/tests/integration/help.bats
@@ -7,11 +7,13 @@ load helpers
   [ "$status" -eq 0 ]
   [[ ${lines[0]} =~ NAME:+ ]]
   [[ ${lines[1]} =~ runc\ '-'\ Open\ Container\ Initiative\ runtime+ ]]
+  [ "${output}" = "${output/VERSION/}" ]
 
   runc --help
   [ "$status" -eq 0 ]
   [[ ${lines[0]} =~ NAME:+ ]]
   [[ ${lines[1]} =~ runc\ '-'\ Open\ Container\ Initiative\ runtime+ ]]
+  [ "${output}" = "${output/VERSION/}" ]
 }
 
 @test "runc command -h" {
--- a/tests/integration/version.bats
+++ b/tests/integration/version.bats
@@ -2,10 +2,18 @@
 
 load helpers
 
-@test "runc version" {
+@test "runc --version" {
   runc -v
   [ "$status" -eq 0 ]
-  [[ ${lines[0]} =~ runc\ version\ [0-9]+\.[0-9]+\.[0-9]+ ]]
+  [[ ${lines[0]} =~ runc\ [0-9]+\.[0-9]+\.[0-9]+ ]]
+  [[ ${lines[1]} =~ commit:+ ]]
+  [[ ${lines[2]} =~ spec:\ [0-9]+\.[0-9]+\.[0-9]+ ]]
+}
+
+@test "runc version" {
+  runc version
+  [ "$status" -eq 0 ]
+  [[ ${lines[0]} =~ runc\ [0-9]+\.[0-9]+\.[0-9]+ ]]
   [[ ${lines[1]} =~ commit:+ ]]
   [[ ${lines[2]} =~ spec:\ [0-9]+\.[0-9]+\.[0-9]+ ]]
 }


Makefile
```diff
@@ -25,10 +25,10 @@ MAN_INSTALL_PATH := ${PREFIX}/share/man/man8/
 VERSION := ${shell cat ./VERSION}
 
 all: $(RUNC_LINK)
-	go build -i -ldflags "-X main.gitCommit=${COMMIT} -X main.version=${VERSION}" -tags "$(BUILDTAGS)" -o runc .
+	go build -i -ldflags "-X version.gitCommit=${COMMIT} -X version.version=${VERSION}" -tags "$(BUILDTAGS)" -o runc .
```
mlaventure: Revert this and the bottom change.

the fist part refers to the package, not the filename.

mikebrow:
Suggest adding the following diff to update the integration test bucket:

```
--- a/tests/integration/version.bats
+++ b/tests/integration/version.bats
@@ -2,10 +2,18 @@
 
 load helpers
 
-@test "runc version" {
+@test "runc --version" {
   runc -v
   [ "$status" -eq 0 ]
   [[ ${lines[0]} =~ runc\ version\ [0-9]+\.[0-9]+\.[0-9]+ ]]
   [[ ${lines[1]} =~ commit:+ ]]
   [[ ${lines[2]} =~ spec:\ [0-9]+\.[0-9]+\.[0-9]+ ]]
 }
+
+@test "runc version" {
+  runc version
+  [ "$status" -eq 0 ]
+  [[ ${lines[0]} =~ runc\ version\ [0-9]+\.[0-9]+\.[0-9]+ ]]
+  [[ ${lines[1]} =~ commit:+ ]]
+  [[ ${lines[2]} =~ spec:\ [0-9]+\.[0-9]+\.[0-9]+ ]]
+}
```

wking:
Pushed fbc7490 → b8ad6a4 which:

* Updates the test suite [1].
* Reverts the broken ‘-X version.…’ in the Makefile [2].

[1]: https://github.com/opencontainers/runc/pull/940#issuecomment-231824069
[2]: https://github.com/opencontainers/runc/pull/940#discussion_r70309013

hqhq:
Looks good.

Jenkins failure seems unrelated, and I don't have authorization to retrigger yet. @wking Can you re-push to trigger Jenkins again?

wking:
On Fri, Jul 15, 2016 at 01:48:00AM -0700, Qiang Huang wrote:
> Can you re-push to trigger Jenkins again?

Rebased onto master with b8ad6a4 → 8ab786f.  Jenkins is still failing,
but the error message [1]:

  bats -t tests/integration
  /tmp/bats.1894.src: line 8: syntax error in conditional expression
  /usr/local/libexec/bats-exec-suite: line 20: let: count+=: syntax error: operand expected (error token is "+=")
  Makefile:68: recipe for target 'localintegration' failed

looks like sstephenson/bats#140.  I suspect an old Bash or otherwise
misconfigured shell on the test box.

[1]: https://jenkins.dockerproject.org/job/runc-PRs/1905/console

version.go
```diff
@@ -0,0 +1,43 @@
...
+
+var versionCommand = cli.Command{
+	Name:        "version",
+	Usage:       "output the runtime version",
+	Description: `The version command outputs the runtime version.`,
+	Action:      printVersion,
+}
+
+func printVersion(context *cli.Context) (err error) {
+	if version == "" {
```
crosbymichael: Why don't you just do the same logic with the slice and strings.Join so we don't have all these verbose if else

crosbymichael:
The build is still broken and needs to be fixed, its related to this change

version.go
wking: On Tue, Aug 23, 2016 at 10:46:25AM -0700, Michael Crosby wrote:
> > +// VERSION file of the source code.
> > +var version = ""
> > +
> > +// gitCommit will be the hash that the binary was built from
> > +// and will be populated by the Makefile
> > +var gitCommit = ""
> > +
> > +var versionCommand = cli.Command{
> > +	Name:        "version",
> > +	Usage:       "output the runtime version",
> > +	Description: `The version command outputs the runtime version.`,
> > +	Action:      printVersion,
> > +}
> > +
> > +func printVersion(context *cli.Context) (err error) {
> > +	if version == "" {
>
> Why don't you just do the same logic with the slice and strings.Join
> so we don't have all these verbose if else

The same logic as what?  This check is for “someone build this
withought ‘-X main.version=…’ like we have in the Makefile”, and I
think we want to keep a check for that case.

crosbymichael: Like we had in the `main.go` before this change.

wking: On Tue, Aug 23, 2016 at 11:09:30AM -0700, Michael Crosby wrote:
> > +// VERSION file of the source code.
> > +var version = ""
> > +
> > +// gitCommit will be the hash that the binary was built from
> > +// and will be populated by the Makefile
> > +var gitCommit = ""
> > +
> > +var versionCommand = cli.Command{
> > +	Name:        "version",
> > +	Usage:       "output the runtime version",
> > +	Description: `The version command outputs the runtime version.`,
> > +	Action:      printVersion,
> > +}
> > +
> > +func printVersion(context *cli.Context) (err error) {
> > +	if version == "" {
> 
> Like we had in the `main.go` before this change.

The code I removed from main.go was:

  - if version != "" {
  -   v = append(v, version)
  - }
  - if gitCommit != "" {
  -   v = append(v, fmt.Sprintf("commit: %s", gitCommit))
  - }
  - v = append(v, fmt.Sprintf("spec: %s", specs.Version))
  - app.Version = strings.Join(v, "\n")

That's building the output string in memory and using Join to store it
(which you need when you're using app.Version).  But with
cli.VersionPrinter, I don't think we need that single-string form
anymore, and using os.Stdout.WriteString directly avoids building the
the string in memory.  Do you want me to build the string in memory to
save some WriteString err checks?

wking:
On Tue, Aug 23, 2016 at 10:46:47AM -0700, Michael Crosby wrote:
> The build is still broken and needs to be fixed, its related to this
> change

Ah, looks like I was missing an escape for a space in the expected
lines[0] comparison.  Fixed with d99746a.

crosbymichael:
The version is broken when just running `runc` and seeing the binary's main output
```
   
VERSION:
   0.0.0
```

wking:
On Tue, Aug 23, 2016 at 11:51:55AM -0700, Michael Crosby wrote:
> The version is broken when just running `runc` and seeing the
> binary's main output
> ```
> VERSION:
>    0.0.0
> ```

Hmm.  I'm not sure why they use .Version directly there instead of
using VersionPrinter.  I see a few possible approaches:

a. Go back to the in-memory string and assign it to app.Version.

   i. Don't bother with whitespace and leave the current tip's
      inconsistent intend:

        VERSION:
           1.0.0-rc1
        commit: 0c6733d669165b53c0117ab522866025c70b3039
        spec: 1.0.0-rc1

   ii. Prefix every line but the first with matching whitespace:

        VERSION:
           1.0.0-rc1
           commit: 0c6733d669165b53c0117ab522866025c70b3039
           spec: 1.0.0-rc1

   iii. Prefix the whole Version string with a newline (although this
        will leave some dangling whitespace in the “blank” line):

        VERSION:
           
        1.0.0-rc1
        commit: 0c6733d669165b53c0117ab522866025c70b3039
        spec: 1.0.0-rc1

b. Adjust the help template [1].

   i. In combination with a.i, use:

        VERSION:

        {{.Version}}

      which would give a result like a.iii, but without the dangling
      whitespace.

   ii. Drop the VERSION section entirely (folks can run ‘runc version’
       or ‘runc --version’ if they want it).

   iii. Adjust the template so we can call cli.VersionPrinter instead
        of using {{.Version}} [2].  I'm not familiar enough with Go's
        template language and the cli package to know if this is
        possible.

Preferences?

[1]: https://github.com/urfave/cli/blob/v1.18.0/README.md#customization-1
[2]: https://golang.org/pkg/text/template/#hdr-Functions

mikebrow:
My preference is to drop the VERSION section from the default help.

wking:
On Tue, Aug 23, 2016 at 02:44:03PM -0700, Mike Brown wrote:
> My preference is to drop the VERSION section from the default help.

Done with tests in d99746a → e09dd4f, which also rebases onto master.
It turned out that I didn't have to touch the template, and just had
to set .Version to an empty string.

version.go
mikebrow: I believe that's what he's asking for, build a string in memory and output it once.

version.go
```diff
@@ -0,0 +1,43 @@
...
+		_, err = os.Stdout.WriteString("runC unknown\n")
```
crosbymichael: its runc not runC

version.go
```diff
@@ -0,0 +1,43 @@
...
+	} else {
+		_, err = os.Stdout.WriteString(fmt.Sprintf("runC %s\n", version))
```
crosbymichael: everything is printing with a `:` between the name and the version

version.go
```diff
@@ -0,0 +1,43 @@
...
+var gitCommit = ""
+
+var versionCommand = cli.Command{
+	Name:        "version",
+	Usage:       "output the runtime version",
+	Description: `The version command outputs the runtime version.`,
+	Action:      printVersion,
+}
+
+func printVersion(context *cli.Context) (err error) {
```
crosbymichael: this entire function's contents are poorly written and is bad quality.  No where else in our code do we use these functions on `os.Stdout` and all the branches with if/else are just bad.  It needs to be improved

version.go
wking: It's runC on http://runc.io/.  And runC sort of helps avoid #24.  But whatever, I can change it ;).

version.go
wking: > everything is printing with a : between the name and the version

So you want one here too?  I'm just matching the current master approach:

```
runc version 1.0.0-rc3
commit: 6b574d57594aedca4bb45b679788af73ae0d65f9
spec: 1.0.0-rc5
```

but I can go either way (with or without colons) if you prefer.

version.go
wking: > No where else in our code do we use these functions on os.Stdout…

I'm replacing the `os.Stdout.WriteString` calls with `fmt.Print*` calls.

> … and all the branches with if/else are just bad.

Most of the `if` blocks are for error checking, and I don't know how to avoid that in Go.  The only `else` is to cover “have/don't-have a version?”, and I think we want to cover both sides of that.  Besides error checking and version presence, the only other conditional is for “have a Git commit?”, and I think we want to cover that too.  Do you have any conditionals you'd like to see dropped?

wking:
@crosbymichael [wrote][1]
> Overall I don't see what all we gain by adding this as there is no standard saying you have to have a version command. Most people know to add -v or --version and that works fine so I don't know why we need this…

As I [pointed out in the initial comment][2], some other tools support `--version`, some support `version`, and some support both.  I think the best place to be from a usability standpoint is supporting both.

I've pushed e09dd4f → 0865185 rebasing onto master and addressing your [runC → runc][3] and [`os.Stdout.WriteString` → something we already use][4] requests.  Outstanding requests from your earlier review:

* [Fewer conditionals in `version.go`][4], but [I'm not sure which to drop][5]
* [Consistency around colon separation][6], but [I'm not sure if/how you'd like me to depart from master][7].

Let me know how you want those handled and I'll get them cleaned up too.

[1]: https://github.com/opencontainers/runc/pull/940#pullrequestreview-6501157
[2]: https://github.com/opencontainers/runc/pull/940#issue-164884756
[3]: https://github.com/opencontainers/runc/pull/940#discussion_r85807103
[4]: https://github.com/opencontainers/runc/pull/940#discussion_r85807466
[5]: https://github.com/opencontainers/runc/pull/940#discussion_r107298990
[6]: https://github.com/opencontainers/runc/pull/940#discussion_r85807233
[7]: https://github.com/opencontainers/runc/pull/940#discussion_r107297516

cyphar:
@wking

> I think the best place to be from a usability standpoint is supporting both.

AkihiroSuda:
Can we close?

mikebrow:
There is no requirement for a runtime version command in the runtime spec, the discussion linked above around adding one was closed. 

The output for runc version information wrt. prefixes used is slightly different between -v and -help/usage output, but such differences are common.

For lack of a treatise on the subject the discussion in the spec led to a suggestion for leaning on GNU's standard:
https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html
https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html#g_t_002d_002dversion

But, as stated, no formal requirements were made in the runtime spec. So, IMO this should be closed and if there is interest in normalizing "version" usage/output we could/should take it up again at the spec level.

PR #943, opened 2016-07-14
Change git -C reset to git reset

johnbieren:
git -C <path> reset --hard <spec commit> is not supported with older versions of git.  For tests/integration/spec.bats I changed from git -C to cd to the directory, run the git reset, then cd back where you were.  Very simple change, but it fixes problems with older versions of git.

### DIFF
--- a/tests/integration/spec.bats
+++ b/tests/integration/spec.bats
@@ -73,7 +73,10 @@ function teardown() {
   [ "$status" -eq 0 ]
 
   SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
-  run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"
+  (
+    cd src/runtime-spec &&
+    run git reset --hard "${SPEC_COMMIT}"
+  )
   [ "$status" -eq 0 ]
   [ -e src/runtime-spec/schema/schema.json ]
 


wking:
On Thu, Jul 14, 2016 at 09:08:09AM -0700, Johnny Bieren wrote:
> git -C <path> reset --hard <spec commit> is not supported with older
> versions of git.

Oops, sorry I suggested it [1] without checking :p.  Instead of cd'ing
in and out, you can use a subshell:

  (
    cd src/runtime-spec &&
    run git reset --hard "${SPEC_COMMIT}"
  )
  [ "$status" -eq 0 ]

which keeps a check for successfully resetting and avoids needing to
‘cd ../..’ and check that for success.

[1]: https://github.com/opencontainers/runc/pull/900#discussion_r66503452

hmeng-19:
Here is some details about the problem:
The git version on RHEL7/centos7 is 1.8.3.1, which does not support -C option. Currently, `runc/tests/integration/spec.bats` uses the `-C` option of git.

mrunalp:
@johnbieren Could you update the PR with @wking suggestion?

johnbieren:
@mrunalp done

wking:
c4b688b9 looks good to me, although the indent seems inconsistent.

I'm also not familiar enough with bats to know what ‘run’ is for;
maybe it should go outside the subshell?  You could test this by
putting a ‘rm -rf src/runtime-spec’ in front of the subshell and
ensuring that the test failed in that case.

johnbieren:
@wking I have fixed the indentation and can confirm that the test passes as written and putting a rm -rf src/runtime-spec in front of the subshell forces the test to fail

wking:
On Mon, Jul 25, 2016 at 12:07:23PM -0700, Johnny Bieren wrote:
> @wking I have fixed the indentation and can confirm that the test
> passes as written and putting a rm -rf src/runtime-spec in front of
> the subshell forces the test to fail

Thanks for testing :).  6c7d2bb looks good to me, although I'd have
indented the contents of the subshell further:

-  run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"
+  (
+    cd src/runtime-spec &&
+    run git reset --hard "${SPEC_COMMIT}"
+  )

tests/integration/spec.bats
```diff
@@ -73,7 +73,10 @@ function teardown() {
   [ "$status" -eq 0 ]
 
   SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
-  run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"
+  (
+  cd src/runtime-spec &&
```
mrunalp: Yeah, let's indent this more.

tests/integration/spec.bats
```diff
@@ -73,7 +73,10 @@ function teardown() {
   [ "$status" -eq 0 ]
 
   SPEC_COMMIT=$(grep runtime-spec ${TESTDIR}/../../Godeps/Godeps.json -A 4 | grep Rev | cut -d":" -f 2 | tr -d ' "')
-  run git -C src/runtime-spec reset --hard "${SPEC_COMMIT}"
+  (
+   cd src/runtime-spec &&
+   run git reset --hard "${SPEC_COMMIT}"
```
hqhq: There should be two spaces for indents, like this:
```
+  (
+    cd src/runtime-spec &&
+    run git reset --hard "${SPEC_COMMIT}"
+  )
```

cyphar: *ahem* Or we should use tabs.

mrunalp:
LGTM

dqminh:
LGTM

PR #947, opened 2016-07-15
Fixed typo in build constraint.

hencrice:
Just a typo that I found when browsing the code.


cyphar:
LGTM. Note that because you can specify build constraints in the filename, this is actually a noop.

mrunalp:
LGTM

PR #949, opened 2016-07-18
configs: fix json tags for CpuRt* options

### LINKED ISSUE #948: CpuRtRuntime and CpuQuota are defined with the same json key "cpu_quota".
`CpuRtRuntime` and `CpuQuota` have the same json field.
`CpuRtPeriod` and `CpuPeriod` have the same json field.
It makes libcontainer get wrong value of `CpuQuota` from docker's `state.json`

https://github.com/opencontainers/runc/blob/master/libcontainer/configs/cgroup_unix.go#L72

```
	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
	CpuQuota int64 `json:"cpu_quota"`

	// CPU period to be used for hardcapping (in usecs). 0 to use system default.
	CpuPeriod int64 `json:"cpu_period"`

	// How many time CPU will use in realtime scheduling (in usecs).
	CpuRtRuntime int64 `json:"cpu_quota"`

	// CPU period to be used for realtime scheduling (in usecs).
	CpuRtPeriod int64 `json:"cpu_period"`
```

cyphar:
Previously we used the same JSON tag name for the regular and realtime
versions of the CpuRt* fields, which causes issues when you want to use
two different values for the fields.

Fixes #948

### DIFF
--- a/libcontainer/configs/cgroup_unix.go
+++ b/libcontainer/configs/cgroup_unix.go
@@ -69,10 +69,10 @@ type Resources struct {
 	CpuPeriod int64 `json:"cpu_period"`
 
 	// How many time CPU will use in realtime scheduling (in usecs).
-	CpuRtRuntime int64 `json:"cpu_quota"`
+	CpuRtRuntime int64 `json:"cpu_rt_quota"`
 
 	// CPU period to be used for realtime scheduling (in usecs).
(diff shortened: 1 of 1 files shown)


dqminh:
LGTM

mrunalp:
LGTM

Issue #915, opened 2016-06-14
TestInitJoinNetworkAndUser failed on RHEL 7


haiyanmeng:
I tested `make test` of runc after enabling user namespace on my RHEL 7 system, all the bats tests succeeded.
However, there is one failure of the unit tests, and it ends up with a dead docker container.

```
=== RUN   TestInitJoinNetworkAndUser
time="2016-06-14T14:31:04Z" level=warning msg="os: process already finished"
--- FAIL: TestInitJoinNetworkAndUser (0.31s)
    utils_test.go:51: exec_test.go:1597: unexpected error: process_linux.go:245: running exec setns process for init caused "exit status 1"                                                  
```

Here is the configuration of my RHEL 7 VM, which runs via `virt-manager`.

```
[root@localhost runc]# uname -a
Linux localhost.localdomain 3.10.0-327.18.2.el7.x86_64 #1 SMP Fri Apr 8 05:09:53 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux
[root@localhost runc]# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.2 (Maipo)
[root@localhost runc]# go version
go version go1.6.2 linux/amd64  
[root@localhost runc]# git branch
* master
[root@localhost runc]$ git log
commit 42dfd606437b538ffde4f0640d433916bee928e3
Merge: c046127 394610a
Author: Qiang Huang <h.huangqiang@huawei.com>
Date:   Tue Jun 14 14:19:20 2016 +0800

    Merge pull request #904 from euank/fix-cgroup-parsing-err
```

hmeng-19:
@mrunalp , @rhatdan , PTAL.

mrunalp:
Could you paste errors using strace -f ?

hmeng-19:
Here is the log of `strace -f -o test.log make test`:
https://raw.githubusercontent.com/hmeng-19/logs/master/test.log

mrunalp:
Better would be to recreate the test using a rub config and just getting a strace on that.

hmeng-19:
Okay. Let me give it a try.

hmeng-19:
@mrunalp , I was not sure how to do the rug config you mentioned.
 
Here is what I did:
I modified the `go test` command for the `localunittest` target in the Makefile:
 
    localunittest: all
[...]

PR #950, opened 2016-07-18
nsenter: major cleanups

cyphar:
Just to make the code much easier to read, also removing redundant code. I also added the functionality that `setns(2)` will fail if we passed the wrong path to the bootstrap process (to avoid bugs in runC's namespace sending being non-fatal).

In addition, I switch `pr_perror` to `bail`, where the process will exit with a **distinct** error code. Currently we aren't handling `syncT` properly within the bootstrap process. But this is a good stopgap.

There are also lots of other fixes that make the code more robust, easier to read and much nicer to maintain. Some of this code comes from the rootless containers PR #774.

`TODO`:
* [x] Make errors distinctive (still not _perfect_ but good enough for now).
* [x] Simplify netlink parsing code.
* :x: Refactor pipes within `nsenter` and with bootstrap process. (not relevant here, there's only one pipe for bootstrap and one for the inter-process stuff).

There are other cleanup parts in #975 and #976 and #977.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -37,7 +37,8 @@ struct clone_arg {
 	jmp_buf *env;
 };
 
-struct nsenter_config {
+struct nlconfig_t {
+	char *data;
 	uint32_t cloneflags;
 	char *uidmap;
 	int uidmap_len;
@@ -226,7 +227,7 @@ static void update_gidmap(int pid, char *map, int map_len)
 
 #define JSON_MAX 4096
 
-static void start_child(int pipenum, jmp_buf *env, int syncpipe[2], struct nsenter_config *config)
+static void start_child(int pipenum, jmp_buf *env, int syncpipe[2], struct nlconfig_t *config)
 {
 	int len, childpid;
 	char buf[JSON_MAX];
@@ -264,13 +265,37 @@ static void start_child(int pipenum, jmp_buf *env, int syncpipe[2], struct nsent
 	exit(0);
 }
 
-static struct nsenter_config process_nl_attributes(int pipenum, char *data, int data_size)
+static void nl_parse(int fd, struct nlconfig_t *config)
 {
-	char *current = data;
-	struct nsenter_config config = { 0 };
+	size_t len, size;
+	struct nlmsghdr hdr;
+	char *data, *current;
 
-	config.consolefd = -1;
-	while (current < data + data_size) {
+	/* Retrieve the netlink header. */
+	len = read(fd, &hdr, NLMSG_HDRLEN);
+	if (len != NLMSG_HDRLEN)
+		bail("invalid netlink header length %lu", len);
+
+	if (hdr.nlmsg_type == NLMSG_ERROR)
+		bail("failed to read netlink message");
+
+	if (hdr.nlmsg_type != INIT_MSG)
+		bail("unexpected msg type %d", hdr.nlmsg_type);
+
+	/* Retrieve data. */
+	size = NLMSG_PAYLOAD(&hdr, 0);
+	current = data = malloc(size);
+	if (!data)
+		bail("failed to allocate %d bytes of memory for nl_payload", size);
+
+	len = read(fd, data, size);
+	if (len != size)
+		bail("failed to read netlink payload, %lu != %lu", len, size);
+
+	/* Parse the netlink payload. */
+	config->data = data;
+	config->consolefd = -1;
+	while (current < data + size) {
 		struct nlattr *nlattr = (struct nlattr *)current;
 		size_t payload_len = nlattr->nla_len - NLA_HDRLEN;
 
@@ -280,7 +305,7 @@ static struct nsenter_config process_nl_attributes(int pipenum, char *data, int 
 		/* Handle payload. */
 		switch (nlattr->nla_type) {
 		case CLONE_FLAGS_ATTR:
-			config.cloneflags = readint32(current);
+			config->cloneflags = readint32(current);
 			break;
 		case CONSOLE_PATH_ATTR:
 			/*
@@ -290,8 +315,8 @@ static struct nsenter_config process_nl_attributes(int pipenum, char *data, int 
 			 * decided here, but rather in container_linux.go. We just
 			 * follow the order given by the netlink message.
 			 */
-			config.consolefd = open(current, O_RDWR);
-			if (config.consolefd < 0)
+			config->consolefd = open(current, O_RDWR);
+			if (config->consolefd < 0)
 				bail("failed to open console %s", current);
 			break;
 		case NS_PATHS_ATTR:{
@@ -340,18 +365,20 @@ static struct nsenter_config process_nl_attributes(int pipenum, char *data, int 
 					close(fd);
 				}
 
+				free(fds);
+				free(paths);
 				break;
 			}
 		case UIDMAP_ATTR:
-			config.uidmap = current;
-			config.uidmap_len = payload_len;
+			config->uidmap = current;
+			config->uidmap_len = payload_len;
 			break;
 		case GIDMAP_ATTR:
-			config.gidmap = current;
-			config.gidmap_len = payload_len;
+			config->gidmap = current;
+			config->gidmap_len = payload_len;
 			break;
 		case SETGROUP_ATTR:
-			config.is_setgroup = readint8(current);
+			config->is_setgroup = readint8(current);
 			break;
 		default:
 			bail("unknown netlink message type %d", nlattr->nla_type);
@@ -359,13 +386,19 @@ static struct nsenter_config process_nl_attributes(int pipenum, char *data, int 
 
 		current += NLA_ALIGN(payload_len);
 	}
+}
 
-	return config;
+void nl_free(struct nlconfig_t *config)
+{
+	free(config->data);
 }
 
 void nsexec(void)
 {
 	int pipenum;
+	jmp_buf env;
+	int syncpipe[2];
+	struct nlconfig_t config = {0};
 
 	/*
 	 * If we don't have an init pipe, just return to the go routine.
@@ -375,31 +408,8 @@ void nsexec(void)
 	if (pipenum == -1)
 		return;
 
-	/* Retrieve the netlink header. */
-	int len;
-	struct nlmsghdr hdr;
-
-	len = read(pipenum, &hdr, NLMSG_HDRLEN);
-	if (len != NLMSG_HDRLEN)
-		bail("invalid netlink header length %d", len);
-
-	if (hdr.nlmsg_type == NLMSG_ERROR)
-		bail("failed to read netlink message");
-
-	if (hdr.nlmsg_type != INIT_MSG)
-		bail("unexpected msg type %d", hdr.nlmsg_type);
-
-	/* Retrieve data. */
-	size_t size = NLMSG_PAYLOAD(&hdr, 0);
-	char *data = alloca(size);
-
-	len = read(pipenum, data, size);
-	if (len != size)
-		bail("failed to read netlink payload, %d != %lu", len, size);
-
-	jmp_buf env;
-	int syncpipe[2];
-	struct nsenter_config config = process_nl_attributes(pipenum, data, size);
+	/* Parse all of the netlink configuration. */
+	nl_parse(pipenum, &config);
 
 	/* clone(2) flags are mandatory. */
 	if (config.cloneflags == -1)
@@ -447,6 +457,9 @@ void nsexec(void)
 			if (dup3(consolefd, STDERR_FILENO, 0) != STDERR_FILENO)
 				bail("failed to dup stderr");
 		}
+
+		/* Free netlink data. */
+		nl_free(&config);
 
 		/* Finish executing, let the Go runtime take over. */
 		return;
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -2,269 +2,263 @@
 #include <endian.h>
 #include <errno.h>
 #include <fcntl.h>
-#include <linux/limits.h>
-#include <sys/socket.h>
-#include <linux/netlink.h>
+#include <grp.h>
 #include <sched.h>
 #include <setjmp.h>
 #include <signal.h>
+#include <stdarg.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/ioctl.h>
-#include <sys/types.h>
-#include <sys/prctl.h>
 #include <unistd.h>
-#include <grp.h>
 
+#include <sys/ioctl.h>
+#include <sys/prctl.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#include <linux/limits.h>
+#include <linux/netlink.h>
 #include <linux/types.h>
 
-// All arguments should be above the stack because it grows down
+#define SYNC_VAL 0x42
+#define JUMP_VAL 0x43
+
+/* Assume the stack grows down, so arguments should be above it. */
 struct clone_arg {
 	/*
 	 * Reserve some space for clone() to locate arguments
 	 * and retcode in this place
 	 */
-	char     stack[4096] __attribute__((aligned(16)));
-	char     stack_ptr[0];
+	char stack[4096] __attribute__ ((aligned(16)));
+	char stack_ptr[0];
 	jmp_buf *env;
 };
 
 struct nsenter_config {
 	uint32_t cloneflags;
-	char     *uidmap;
-	int      uidmap_len;
-	char     *gidmap;
-	int      gidmap_len;
-	uint8_t  is_setgroup;
-	int      consolefd;
+	char *uidmap;
+	int uidmap_len;
+	char *gidmap;
+	int gidmap_len;
+	uint8_t is_setgroup;
+	int consolefd;
 };
 
-// list of known message types we want to send to bootstrap program
-// These are defined in libcontainer/message_linux.go
-#define INIT_MSG	    62000
-#define CLONE_FLAGS_ATTR    27281
-#define CONSOLE_PATH_ATTR   27282
-#define NS_PATHS_ATTR	    27283
-#define UIDMAP_ATTR	    27284
-#define GIDMAP_ATTR	    27285
-#define SETGROUP_ATTR	    27286
+/*
+ * List of netlink message types sent to us as part of bootstrapping the init.
+ * These constants are defined in libcontainer/message_linux.go.
+ */
+#define INIT_MSG		62000
+#define CLONE_FLAGS_ATTR	27281
+#define CONSOLE_PATH_ATTR	27282
+#define NS_PATHS_ATTR		27283
+#define UIDMAP_ATTR		27284
+#define GIDMAP_ATTR		27285
+#define SETGROUP_ATTR		27286
 
-// Use raw setns syscall for versions of glibc that don't include it
-// (namely glibc-2.12)
+/*
+ * Use the raw syscall for versions of glibc which don't include a function for
+ * it, namely (glibc 2.12).
+ */
 #if __GLIBC__ == 2 && __GLIBC_MINOR__ < 14
-    #define _GNU_SOURCE
-    #include "syscall.h"
-    #if defined(__NR_setns) && !defined(SYS_setns)
-	#define SYS_setns __NR_setns
-    #endif
+#	define _GNU_SOURCE
+#	include "syscall.h"
+#	if !defined(SYS_setns) && defined(__NR_setns)
+#		define SYS_setns __NR_setns
+#	endif
 
-    #ifdef SYS_setns
-	int setns(int fd, int nstype)
-	{
-	    return syscall(SYS_setns, fd, nstype);
-	}
-    #endif
+#ifndef SYS_setns
+#	error "setns(2) syscall not supported by glibc version"
 #endif
 
-#define pr_perror(fmt, ...)                                                    \
-	fprintf(stderr, "nsenter: " fmt ": %m\n", ##__VA_ARGS__)
-
-static int child_func(void *_arg)
+int setns(int fd, int nstype)
 {
-    struct clone_arg *arg = (struct clone_arg *)_arg;
-    longjmp(*arg->env, 1);
+	return syscall(SYS_setns, fd, nstype);
+}
+#endif
+
+/* TODO(cyphar): Fix this so it correctly deals with syncT. */
+#define bail(fmt, ...)							\
+	do {								\
+		fprintf(stderr, "nsenter: " fmt ": %m\n", ##__VA_ARGS__); \
+		exit(__COUNTER__ + 1);					\
+	} while(0)
+
+static int child_func(void *arg)
+{
+	struct clone_arg *ca = (struct clone_arg *)arg;
+	longjmp(*ca->env, JUMP_VAL);
 }
 
-static int clone_parent(jmp_buf *env, int flags) __attribute__((noinline));
+static int clone_parent(jmp_buf *env, int flags) __attribute__ ((noinline));
 static int clone_parent(jmp_buf *env, int flags)
 {
-	struct clone_arg ca;
-	int		 child;
+	int child;
+	struct clone_arg ca = {
+		.env = env,
+	};
 
-	ca.env = env;
-	child  = clone(child_func, ca.stack_ptr, CLONE_PARENT | SIGCHLD | flags,
-		      &ca);
-	// On old kernels, CLONE_PARENT cannot work with CLONE_NEWPID,
-	// unshare before clone to workaround this.
-	if (child == -1 && errno == EINVAL) {
-		if (unshare(flags)) {
-			pr_perror("Unable to unshare namespaces");
-			return -1;
-		}
-		child  = clone(child_func, ca.stack_ptr, SIGCHLD | CLONE_PARENT,
-			      &ca);
+	child = clone(child_func, ca.stack_ptr, CLONE_PARENT | SIGCHLD | flags, &ca);
+
+	/*
+	 * On old kernels, CLONE_PARENT didn't work with CLONE_NEWPID, so we have
+	 * to unshare(2) before clone(2) in order to do this. This was fixed in
+	 * upstream commit 1f7f4dde5c945f41a7abc2285be43d918029ecc5, and was
+	 * introduced by by 40a0d32d1eaffe6aac7324ca92604b6b3977eb0e.
+	 *
+	 * As far as we're aware, the last mainline kernel which had this bug was
+	 * Linux 3.12. However, we cannot comment on which kernels the broken patch
+	 * was backported to.
+	 */
+	if (errno == EINVAL) {
+		if (unshare(flags) < 0)
+			bail("unable to unshare namespaces");
+		child = clone(child_func, ca.stack_ptr, SIGCHLD | CLONE_PARENT, &ca);
 	}
+
 	return child;
 }
 
-// get init pipe from the parent. It's used to read bootstrap data, and to
-// write pid to after nsexec finishes setting up the environment.
-static int get_init_pipe()
+/*
+ * Gets the init pipe fd from the environment, which is used to read the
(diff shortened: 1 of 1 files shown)


libcontainer/nsenter/nsexec.c
```diff
@@ -32,112 +36,104 @@ struct clone_arg {
...
-#define GIDMAP_ATTR	    27285
-#define SETGROUP_ATTR	    27286
-
-// Use raw setns syscall for versions of glibc that don't include it
-// (namely glibc-2.12)
+/*
+ * List of netlink message types sent to us as part of bootstrapping the init.
+ * These constants are defined in libcontainer/message_linux.go.
+ */
+#define INIT_MSG			62000
```
mlaventure: nit: some alignment issues here

libcontainer/nsenter/nsexec.c
```diff
@@ -150,76 +146,60 @@ static uint8_t readint8(char *buf)
...
-	if (fd == -1) {
-		pr_perror("failed to open %s", buf);
-		exit(1);
-	}
+	len = snprintf(path, PATH_MAX, pathfmt, pid);
+	if (len < 0)
+		bail("failed to construct '%s' for %d", pathfmt, pid);
+
+	fd = open(path, O_RDWR);
+	if (fd < 0)
```
mlaventure: Can we keep the extra `{}` here (and in the other one liner `if`)?

libcontainer/nsenter/nsexec.c
```diff
@@ -150,76 +146,60 @@ static uint8_t readint8(char *buf)
...
-		if (write(fd, "allow", 5) != 5) {
-			// If the kernel is too old to support
-			// /proc/PID/setgroups, write will return
-			// ENOENT; this is OK.
+		/* XXX: What about old kernels. */
+		fd = open(path, O_RDWR);
+		if (fd < 0)
+			bail("failed to open %s", path);
+
+		if (write(fd, allow, strlen(allow)) != strlen(allow)) {
```
mlaventure: `sizeof(allow) - 1` to avoid calling `strlen`?

cyphar: Why is a call to `strlen` bad? `sizeof(allow) - 1` wouldn't be completely accurate, you would need to do `sizeof(allow) / sizeof(char) - 1` (even though it's not completely necessary).

libcontainer/nsenter/nsexec.c
cyphar: What do you mean? I'm just matching the kernel style (where single-line statements don't need wrapping braces).

libcontainer/nsenter/nsexec.c
cyphar: Fixed. I set my editor to a tabsize of 4 by accident.

crosbymichael:
For all your formatting needs...

https://github.com/crosbymichael/vim-cfmt

libcontainer/nsenter/nsexec.c
```diff
@@ -356,116 +397,95 @@ void nsexec(void)
...
+		/* Close the writing side of pipe. */
 		close(syncpipe[1]);
 
-		// sync with parent
-		if ((read(syncpipe[0], &s, 1) != 1) || (s != 1)) {
-			pr_perror("Failed to read sync byte from parent");
-			exit(1);
-		}
+		/* Sync with parent. */
+		if (read(syncpipe[0], &s, sizeof(s)) != sizeof(s) || s != SYNC_VAL)
```
mlaventure: can we add back the extras `()` for easier reading?

cyphar:
@crosbymichael Cheers, I've added that to my `dotfiles`. Rebased. :P

libcontainer/nsenter/nsexec.c
cyphar: IMO it isn't easier to read with more parenthesis.

libcontainer/nsenter/nsexec.c
mlaventure: I know, but I still find it easier to read. Also, it matches the style of Go which use we everywhere  else.

It's just a nit though.

libcontainer/nsenter/nsexec.c
mlaventure: It's not bad, just not really necessary :).

It's also just a nit.

libcontainer/nsenter/nsexec.c
cyphar: IMO the kernel style is much nicer to read if you're reading/writing C.

libcontainer/nsenter/nsexec.c
```diff
@@ -20,206 +21,190 @@
...
 {
-    struct clone_arg *arg = (struct clone_arg *)_arg;
-    longjmp(*arg->env, 1);
+	struct clone_arg *ca = (struct clone_arg *)arg;
+	longjmp(*ca->env, JUMP_VAL);
 }
 
-static int clone_parent(jmp_buf *env, int flags) __attribute__((noinline));
-static int clone_parent(jmp_buf *env, int flags)
+static int clone_parent(jmp_buf * env, int flags) __attribute__ ((noinline));
```
mrunalp: Space between `*` and `env` :/

cyphar: It looks like GNU indent doesn't know what the kernel style is.

libcontainer/nsenter/nsexec.c
```diff
@@ -228,122 +208,181 @@ static void update_process_gidmap(int pid, uint8_t is_setgroup, char *map, int m
...
+	 * (the bootstrap process). Also so we don't need to forward the
+	 * child's exit code or resend its death signal.
+	 */
 	childpid = clone_parent(env, config->cloneflags);
-	if (childpid < 0) {
-		pr_perror("Unable to fork");
-		exit(1);
-	}
+	if (childpid < 0)
+		bail("Unable to fork");
```
mrunalp: nit: Unable -> unable

libcontainer/nsenter/nsexec.c
```diff
@@ -356,116 +395,96 @@ void nsexec(void)
...
+				bail("failed to dup stderr");
 		}
 
-		// Finish executing, let the Go runtime take over.
+		/* Finish executing, let the Go runtime take over. */
 		return;
 	}
 
-	// Parent
+	/* Run the parent code. */
```
dqminh: i prefer `//` for comment. It's also consistent with go. Any reason you dislike it ?

libcontainer/nsenter/nsexec.c
dqminh: I also prefer to add braces :p

libcontainer/nsenter/nsexec.c
dqminh: is it just me who prefer `write(fd, "allow", 5) != 5` :p, its  simpler ( to me ) that way :D ?

libcontainer/container_linux.go
```diff
@@ -1249,7 +1254,7 @@ func (c *linuxContainer) orderNamespacePaths(namespaces map[configs.NamespaceTyp
 			if strings.ContainsRune(p, ',') {
 				return nil, newSystemError(fmt.Errorf("invalid path %s", p))
 			}
-			paths = append(paths, p)
+			paths = append(paths, fmt.Sprintf("%s:%s", configs.NsName(nsType), p))
```
dqminh: can you explain why do we need this change ?

libcontainer/nsenter/nsexec.c
cyphar: It's not classic C, it's against the Linux kernel style and it reminds me of C++.  :P

libcontainer/container_linux.go
cyphar: It's so that we have type safety with `setns`. I've also updated the `setns` code so that it will make sure we're attaching to the right type of namespace. While this isn't a functional change, it does make sure that we don't accidentally join a namespace we didn't intend to.

libcontainer/nsenter/nsexec.c
mrunalp: I am with @cyphar on adopting the Linux C style :+1:

crosbymichael: Yes, this is what we have always done.  -linux with the indent program.  Its like gofmt, we maybe not all agree according to taste, but atleast we are all consistent and that matters more.

mrunalp: @crosbymichael Agree on using some tool with the same flags. Maybe we could even add a small script that runs on the C code.

cyphar: Note that I ran with `-linux -l120` because the 80 character restriction isn't something we enforce anywhere else (and it makes the code less readable).

mlaventure:
I like the new `write_file` take :+1: 

Although, the commit message will need to be changed before merging :)

cyphar:
Of course, it's still a wip because I still don't like bits of how the management are done. There's also some more code that I want to pull out of the rootless containers patchset to put it here. :P

libcontainer/nsenter/nsexec.c
cyphar: Oh, and I think I've found several bugs in GNU indent that I should probably set about fixing.

libcontainer/container_linux.go
dqminh: i feel like this doesnt gain us anything since setns already knows the namespace type of the fd. Does this fix any bugs ?

libcontainer/nsenter/nsexec.c
```diff
@@ -2,351 +2,350 @@
...
-		exit(1);
-	} else if (len != map_len) {
-		pr_perror("Failed to write data to %s (%d/%d)",
-			buf, len, map_len);
-		close(fd);
-		exit(1);
-	}
+	len = write(fd, data, data_len);
+	if (len != data_len)
+		return -1;
```
hqhq: Shall we close fd before return?

libcontainer/nsenter/nsexec.c
```diff
@@ -2,351 +2,350 @@
...
+					int fd;
+
+					/* Resize fds. */
+					num++;
+					fds = realloc(fds, num * sizeof(int));
+					paths = realloc(paths, num * sizeof(char *));
+
+					fd = open(ns, O_RDONLY);
+					if (fd < 0)
+						bail("failed to open %s", ns);
```
hqhq: Same here, should close opened fds.

libcontainer/nsenter/nsexec.c
```diff
@@ -2,351 +2,350 @@
...
+				/*
+				 * Open each namespace path and setns it in the
+				 * order provided to us. We currently don't have
+				 * any context for what kind of namespace we're
+				 * joining, so just blindly do it.
+				 */
+				char *saveptr = NULL;
+				char *ns = strtok_r(current, ",", &saveptr);
+				int *fds = NULL, num = 0, i;
+				char **paths = NULL;
```
hqhq: A nit, can we define these variables outside and remove the braces? It'll make the format better.

libcontainer/nsenter/nsexec.c
cyphar: I'm currently thinking about that. In all of the error cases, we are going to exit so I don't think it _really_ matters that we close all of the file descriptors. I'm currently trying to figure out how `alloca` interacts with `clone` (hint: the answer is "not nicely") but I might also end up just adding a memory leak because we're going to `execve` anyway (which clears all of the allocated memory from the old process anyway).

libcontainer/nsenter/nsexec.c
cyphar: I prefer them here because they're local to this case. Also, I'm planning on rewriting most of this because we need to do all of this handling in the first child (due to the fact that the PID namespace applies on children and the first child has to send the PID of the second).

libcontainer/container_linux.go
cyphar: It doesn't fix any current bugs, but it does allow us to make sure that we don't accidentally add any bugs with regards to the ordering of namespace paths (which would cause very, very odd issues that probably wouldn't be easy to diagnose).

cyphar:
Phew. Okay, so now the tests pass after my huge rewrite of how we fork processes.

libcontainer/nsenter/nsexec.c
mlaventure: otherwise you could `return -(len != data_len)` that way `close(fd)` will always be called

cyphar:
Alright. I think I've included all of the refactorings and improvements that I wanted to. That was not as much fun as it looked. PTAL.

/cc @crosbymichael @hqhq @mrunalp @dqminh

libcontainer/nsenter/nsexec.c
```diff
@@ -1,471 +1,750 @@
...
+					if (config.is_setgroup)
+						update_setgroups(child, SETGROUPS_ALLOW);
+
+					/* Set up mappings. */
+					update_uidmap(child, config.uidmap, config.uidmap_len);
+					update_gidmap(child, config.gidmap, config.gidmap_len);
+
+					s = SYNC_USERMAP_ACK;
+					if (write(syncfd, &s, sizeof(s)) != sizeof(s)) {
+						kill(child, SIGKILL);
```
mlaventure: `waitpid()`?

cyphar: What do you mean? We didn't `waitpid` before, and I don't think we can because the processes are cloned with `CLONE_PARENT | SIGCHLD`.

mlaventure: right, overlooked the `CLONE_PARENT`, sorry!

cyphar:
From this weeks meeting, @crosbymichael suggested that we include a refactoring of the communication methods between the bootstrap process and the init. This would involve presumably figuring out how we should deal with both the netlink magic as well as the other bits and pieces.

Also, this is probably going to be the base for the console handling rewrite. So there's that. :wink:

libcontainer/nsenter/nsexec.c
```diff
@@ -1,471 +1,753 @@
...
+
+	if (hdr.nlmsg_type == NLMSG_ERROR)
+		bail("failed to read netlink message");
+
+	if (hdr.nlmsg_type != INIT_MSG)
+		bail("unexpected msg type %d", hdr.nlmsg_type);
+
+	/* Retrieve data. */
+	size = NLMSG_PAYLOAD(&hdr, 0);
+	current = data = malloc(size);
```
avagin: Need to check that malloc doesn't return NULL here

libcontainer/nsenter/nsexec.c
```diff
@@ -1,471 +1,753 @@
...
-			config.gidmap     = data + start;
-			config.gidmap_len = payload_len;
-		} else if (nlattr->nla_type == SETGROUP_ATTR) {
-			config.is_setgroup = readint8(data + start);
-		} else {
-			pr_perror("Unknown netlink message type %d",
-				  nlattr->nla_type);
-			exit(1);
-		}
+		if (setns(ns.fd, ns.ns) < 0)
```
avagin: Offtopic: We need to set a correct uid before entering into userns. I can't find where we do this.
https://access.redhat.com/security/cve/cve-2015-8709

cyphar: It's in the very last part (after all of the clones). Are you suggesting that we do this:

```
setns(CLONE_NEWUSER);
setresgid(...);
setresuid(...);
setns(the rest);
```

I'm not sure whether that will work fine in all cases.

avagin: You need to set uid, gid, groups from a target user namespace before switching into it, otherwise you will have a security issue.
setresgid(...);
setresuid(...);
setns(CLONE_NEWUSER);

https://github.com/xemul/criu/commit/6e1726f8fb7c5a22538c61ef3828c375cc44d0d1 is an example how the same problem was fixed in CRIU.

cyphar: Okay, I'll include that in this patchset. Thanks for the heads up. @avagin

hmeng-19:
@cyphar , @mrunalp , I tested this PR on Fodora 23, and `make test` finished successfully.

I also tested this PR on RHEL7, and the same failure happened as described in #915. However, I do not think the failure is due to this PR. 
It seems that, on RHEL7, once a process joins another process's user namespace, it can not
run `clone` with the `CLONE_NEWNS` flag.

mrunalp:
@cyphar I tested the PR. It doesn't fix #959.

cyphar:
@mrunalp I believe this is related to @avagin's points about the changing of `uid` and `gid`s. We need to set them to the root in the user namespace *before* `unshare(USER)` and then set them again after the `unshare(USER)` (before anything else). Working on it now.

mrunalp:
My fix relies on being root in the user namespace so sounds like that should fix it. Thanks!

cyphar:
@mrunalp Okay, I've pushed a fix that handles `setresuid`. PTAL. Note that I'm not really sure what security policy stuff we have to do with `setns` -- do we need to `setresuid()` before we do a `setns` and can we do it before the `open` in that case?

@avagin Can you comment on `setns()` and the relative safety of that?

avagin:
@cyphar you call setresuid after join_namespaces(), but it should be called before switching an user namespace.
The idea is that a process can switch an userns and before it sets proper uid-s someone can catch it by ptraces and execute arbitrary code in its context.

cyphar:
@avagin Okay, and we probably need to open `/proc/*/ns/*` before we  do that as well. I'll split `join_namespace` to that end (sucks because it's actually fairly clean right now).

cyphar:
@avagin Alright, I've made it so that we switch user to the owner of the `/proc/*/ns/user` file for the namespace we're joining. We don't have any other context in the bootstrap process about the process itself, so this is the least invasive way of doing it (is it fine?).

Now we have three different points where we change user, are there any we are missing (/cc @opencontainers/runc-maintainers)?

1. Before `setns(2)`.
2. Before `unshare(CLONE_NEWUSER)`.
3. Before `unshare(CLONE_*)` (the rest).

libcontainer/nsenter/nsexec.c
cyphar: Fix'd.

dqminh:
> Now we have four different points where we change user, are there any we are missing

1. Before setns(2).
2. Before unshare(CLONE_NEWUSER).
3. Before unshare(CLONE_*) (the rest).
4. Right before we return to the Go runtime.

1, 2, 3 seems a bit complex and logically aren't they 3 identical set of euid/guid being set ? Can we just set them once before we do any clone/setns ? We can derive the euid/egid from:
- if no userns, then 0,0
- if create new userns, derive from id config
- if join existing userns, get from the userns fd

libcontainer/nsenter/nsexec.c
```diff
@@ -1,471 +1,878 @@
...
+		.jmpval = jmpval,
+	};
+
+	/*
+	 * Okay, so this is quite annoying.
+	 *
+	 * In order to make sure that deal with older kernels (when
+	 * CLONE_NEWUSER wasn't guaranteed to be done first if you specify
+	 * multiple namespaces in a clone(2) invocation) as well as with
+	 * certain usecases like rootless containers, we cannot just dump all
```
dqminh: can you describe the failure of rootless container usecase if we dont do CLONE_NEWUSER first ?

do we even want to support older kernel with malfunctioned clone 😢 ?

libcontainer/nsenter/nsexec.c
```diff
@@ -1,471 +1,878 @@
...
+			return;
+		}
+	default:
+		bail("unexpected jump value");
+		break;
 	}
 
-	// Parent
-	start_child(pipenum, &env, syncpipe, &config);
+	/* Should never be reached. */
```
dqminh: lets remove this comment, `bail("should never be reached")` should be enough warning

libcontainer/nsenter/nsexec.c
crosbymichael: how old are we talking?  Pre 3.10? or something newer?

mrunalp:
@cyphar It works for #959 :+1:  I see the code changes that make it work i.e. setuid(0) after unshare CLONE_NEWUSER and before unshare CLONE_NEWIPC. Reviewing the rest of it.

cyphar:
> 1, 2, 3 seems a bit complex and logically aren't they 3 identical set of euid/guid being set ? Can we just set them once before we do any clone/setns ? We can derive the euid/egid from:

If we're not using user namespaces, they're all equivalent. If we are using user namespaces, they are not equivalent.

The first one changes us to have `euid = root in the namespace we're joining`. WIthout this you get security vulnerabilities (as @avagin mentioned) because you have `euid = (kuid 0)` and a racing process could `ptrace` it and start executing code as a root process.

The second one does a similar thing, but for a new mapping that we're creating (it's the same sort of argument). And the third one is required to `unshare` the other namespaces (your user **must** be mapped if you're using SELinux or some security policies).

In fact, I think the fourth one isn't necessary.

libcontainer/nsenter/nsexec.c
cyphar: There's also the SELinux case (/ping @mrunalp) that essentially means you have to split `CLONE_NEWUSER` and the other clones (because you need to set your UID).

Also, IMO there's no real reason for us to not do this. It makes no real difference for newer kernels, and pushes our kernel support back quite a bit (including Frankenkernels).

cyphar:
@opencontainers/runc-maintainers :: Okay, @dqminh and I had a discussion about the mutual exclusivity of `setns(2)` and `unshare(2)`. The current way `nsexec` is written, they are not enforced to be exclusive. Personally I think changing `join_namespaces` to make them mutually exclusive would make the code quite ugly (you'd need to parse the namespace string in order to figure out the right UID to change to, which would require more allocations).

Would everyone be fine if we enforce this in `nl_parse` (like we do `cloneflags`). Or what if we didn't enforce it in C at all (since we can enforce it in Go)? I peronally prefer enforcing it in `nl_parse` as a policy decision rather than changing the way we join namespaces just to avoid this particular case (which may be useful in the feature).

libcontainer/nsenter/nsexec.c
```diff
@@ -1,471 +1,920 @@
...
+			break;
+		case CONSOLE_PATH_ATTR:
+			/*
+			 * We open the console here because we currently evaluate console
+			 * paths from the *host* namespaces.
+			 */
+			config->consolefd = open(current, O_RDWR);
+			if (config->consolefd < 0)
+				bail("failed to open console %s", current);
+			break;
```
hqhq: Not introduced in this PR, I have a question about consolePath, is there any specific reasons that we have to handle consolePath in `nsexec.c`? It's only used by exec process, why can't we handle it in go code like we do for init process?

If possible, I think we should do so to make c code as simple as possible. @crosbymichael

hqhq:
@cyphar I think you can split this PR to speed up the process, the clean up part is a big change but can be merged rapidly, it'll also make other people easier to review and give other maintainers more confidence to merge :)

cyphar:
@hqhq Fair point. :P I will break up this PR into three parts:

1. Cleanups.
2. Fixing namespace guarantees.
3. Fixes for the orphaning and similar issues.

This PR blew up quite quickly. Hopefully the commit will rebase nicely. :P

cyphar:
Okay, I think this PR is done (in terms of development). The rest of the changes will happen in the separate split PRs based on this one (see #975, #976, #977). PTAL so we can merge this finally.

hqhq:
LGTM, ping @mrunalp @crosbymichael @avagin @dqminh

libcontainer/nsenter/nsexec.c
```diff
@@ -2,470 +2,463 @@
...
-			      &ca);
+	int child;
+	struct clone_arg ca = {
+		.env = env,
+	};
+
+	child = clone(child_func, ca.stack_ptr, CLONE_PARENT | SIGCHLD | flags, &ca);
+
+	/*
+	 * On old kernels, CLONE_PARENT didn't work with CLONE_NEWPID, so we have
```
dqminh: nit: can we note an example of the last known kernel that exhibits this problem so maybe eventually we can remove this if possible ?
If we want to support those kernels, it might also make sense to setup CI for it ?

libcontainer/nsenter/nsexec.c
```diff
@@ -2,470 +2,463 @@
...
-	}
 
-	pipenum = atoi(initpipe);
-	snprintf(buf, sizeof(buf), "%d", pipenum);
-	if (strcmp(initpipe, buf)) {
-		pr_perror("Unable to parse _LIBCONTAINER_INITPIPE");
-		exit(1);
-	}
+	pipenum = strtol(initpipe, &endptr, 10);
+	if (*endptr != '\0')
```
dqminh: i think we should also check for errno here in case it over/underflows

libcontainer/nsenter/nsexec.c
cyphar: This is the commit that fixed this: torvalds/linux@1f7f4dde5c945f41a7abc2285be43d918029ecc5. And this is the commit which caused the issue: torvalds/linux@40a0d32d1eaffe6aac7324ca92604b6b3977eb0e.

The only mainline kernel which this affected is 3.12, but I can't really comment on kernels that this change might've been backported to.

cyphar: Fix'd.

libcontainer/nsenter/nsexec.c
cyphar: Fix'd.

libcontainer/nsenter/nsexec.c
```diff
@@ -2,470 +2,469 @@
...
-	}
 
-	pipenum = atoi(initpipe);
-	snprintf(buf, sizeof(buf), "%d", pipenum);
-	if (strcmp(initpipe, buf)) {
-		pr_perror("Unable to parse _LIBCONTAINER_INITPIPE");
-		exit(1);
-	}
+	pipenum = strtol(initpipe, &endptr, 10);
+	if (*endptr != '\0' || errno != 0)
```
dqminh: nit: the man page also suggests that we reset errno to 0 before the call

dqminh:
LGTM except a small nit

cyphar:
Fix'd.

/cc @dqminh @hqhq

libcontainer/nsenter/nsexec.c
cyphar: Fixed.

cyphar:
Alright. Time for the final round of LGTMs.

dqminh:
LGTM

mrunalp:
@cyphar Where did the fix for #959 end up?

libcontainer/nsenter/nsexec.c
mrunalp: NULL check still missing?

cyphar:
@mrunalp It's in #975. The changes required to implement it require #977 (which is a pretty big change in of itself).

libcontainer/nsenter/nsexec.c
cyphar: I'm planning on cleaning up the entire `--console` thing semantics in a separate PR. But working on that PR requires some of the groundwork in this (and related) PRs. I'm not sure why we deal with it in C, but I'll see if I can remove it as part of #814.

cyphar: @hqhq It's because we `setns` inside the C code, and we can't be sure of where the slave path will be inside the container's mount namespace (or even if there will be a path to it) we need to open it before we join the container's namespaces. This isn't a problem for when we create a container, because nothing changes after an `unshare`.

All of this will be reworked as part of the `--console` rewrite in #814. But first we need to merge some of this code that will lay the groundwork for that other stuff. :P

libcontainer/nsenter/nsexec.c
mrunalp: @cyphar I don't see a NULL check here..

mrunalp:
LGTM besides one comment about null check.

libcontainer/nsenter/nsexec.c
cyphar: Dammit, I must've accidentally put the fix in a different patch.

cyphar: There, actually fixed this time.

cyphar:
@mrunalp Fixed the NULL check.

avagin:
LGTM

mrunalp:
LGTM

cyphar:
:tada: Now for the other PRs. :D

PR #956, opened 2016-07-19
Allow cgroup creation without attaching a pid

dubstack:
I am working on introducing pod level cgroups in Kubernetes (kubernetes/kubernetes#27204). As a part I would like to use libcontainer for creating and managing cgroups in the system. I would like to just create a cgroup with no pid attached and if need be apply a pid to the cgroup later on. But currently libcontainer doesn't support cgroup creation without attaching a pid. 

This would allow us to skip attaching a pid to a cgroup during creation.
cc @mrunalp  @vishh @derekwaynecarr

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strconv"
 	"sync"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -33,7 +32,6 @@ var (
 		&FreezerGroup{},
 		&NameGroup{GroupName: "name=systemd", Join: true},
 	}
-	CgroupProcesses  = "cgroup.procs"
 	HugePageSizes, _ = cgroups.GetHugePageSize()
 )
 
@@ -341,7 +339,7 @@ func (raw *cgroupData) join(subsystem string) (string, error) {
 	if err := os.MkdirAll(path, 0755); err != nil {
 		return "", err
 	}
-	if err := writeFile(path, CgroupProcesses, strconv.Itoa(raw.pid)); err != nil {
+	if err := cgroups.WriteCgroupProc(path, raw.pid); err != nil {
 		return "", err
 	}
 	return path, nil
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -16,7 +16,10 @@ import (
 	"github.com/docker/go-units"
 )
 
-const cgroupNamePrefix = "name="
+const (
+	cgroupNamePrefix = "name="
+	CgroupProcesses  = "cgroup.procs"
+)
 
 // https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
 func FindCgroupMountpoint(subsystem string) (string, error) {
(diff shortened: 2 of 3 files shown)


libcontainer/cgroups/utils.go
```diff
@@ -326,9 +326,12 @@ func PathExists(path string) bool {
 func EnterPid(cgroupPaths map[string]string, pid int) error {
 	for _, path := range cgroupPaths {
 		if PathExists(path) {
-			if err := ioutil.WriteFile(filepath.Join(path, "cgroup.procs"),
-				[]byte(strconv.Itoa(pid)), 0700); err != nil {
-				return err
+			// Dont attach any pid to the cgroup if -1 is specified as a pid
+			if pid != -1 {
+				if err := ioutil.WriteFile(filepath.Join(path, "cgroup.procs"),
```
cyphar: While replacing this, maybe we should be using `writeFile` here instead of `ioutil`.

cyphar:
Just to be clear, this is kubernetes using `libcontainer` as a library for its own internal management of things -- not as part of the OCI integration being worked on (which would mean making changes to the runtime specification and a few other things)?

libcontainer/cgroups/utils.go
dubstack: Sounds Good.

dubstack:
Yes this is needed for Kubernete's internal implementation.

libcontainer/cgroups/utils.go
dubstack: Re above, Now that I looked into it more carefully this was done to avoid import cycles.
I would suggest that we move the WriteFile method to cgroups.utils.go.

Or we can also not make any such changes and continue to use ioutil in the utils.go file.

mrunalp:
LGTM

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -341,8 +341,11 @@ func (raw *cgroupData) join(subsystem string) (string, error) {
 	if err := os.MkdirAll(path, 0755); err != nil {
 		return "", err
 	}
-	if err := writeFile(path, CgroupProcesses, strconv.Itoa(raw.pid)); err != nil {
-		return "", err
+	// Dont attach any pid to the cgroup if -1 is specified as a pid
+	if raw.pid != -1 {
```
derekwaynecarr: i worry the -1 is kind of undocumented unless we have it that all writeFile calls to cgroup.procs share a common function call.  a lot of this code is not unit tested, so would prefer that we at least consolidate the -1 checking into a common function like `writeCgroupProc(pid int)` and just keep the -1 check local to there..

dubstack: Yeah that sounds reasonable.

dubstack:
@derekwaynecarr @mrunalp @cyphar PTAL

derekwaynecarr:
LGTM, thanks for the update.

cyphar:
LGTM. Sucks that `writeFile` causes an import cycle.

crosbymichael:
@cyphar write file is not complicated, we can fix it or remove it

dubstack:
Should we consider moving a bunch of "utility like" functions in apply_raw to utils.go ??

cyphar:
Yeah, we should probably move it into `libcontainer/utils` because it's legitimately useful elsewhere.

dubstack:
I will send out a PR for the same.

dubstack:
@cyphar Can you LGTM this PR as it's blocking other PR's.

I will move the writeFile to libcontainer/utils in a  different PR.

cyphar:
LGTM, sure. But you still need one more maintainer to LGTM this version of the PR.

dubstack:
Oh, right! cc @mrunalp

mrunalp:
LGTM

Issue #946, opened 2016-07-14
`TestRunWithKernelMemory` failed


hmeng-19:
It seems that the commit https://github.com/opencontainers/runc/commit/8dd3d634556427c22769125ef2b0aa2ce613a2a7 broke the `TestRunWithKernelMemory` test somehow. Here is the error in the output of `make test`:

    --- FAIL: TestRunWithKernelMemory (0.13s)^M
        exec_test.go:671: runContainer failed with kernel memory limit: process_linux.go:291: setting cgroup config for ready process caused "cannot set kmem.limit_in_bytes after task have joined this cgroup"^M

Its previou commit https://github.com/opencontainers/runc/commit/9d7831e41d3ef428b67685eeb27f2b4a22a92391 works well for me.

hqhq:
I can reproduce this problem on some machines, I opened a PR to revert that commit.
https://github.com/opencontainers/runc/pull/961

cyphar:
@hqhq I was about to do the same, it's been bothering me for a few days.

PR #961, opened 2016-07-21
Revert "Use update time to detect if kmem limits have been set"

hqhq:
Revert: #935
Fixes: #946

I can reproduce #946 on some machines, the problem is on
some machines, it could be very fast that modify time
of `memory.kmem.limit_in_bytes` could be the same as
before it's modified.

And now we'll call `SetKernelMemory` twice on container
creation which cause the second time fail.

Revert this before we find a better solution.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -5,14 +5,15 @@ package fs
 import (
 	"bufio"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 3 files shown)


hqhq:
We should reopen #841 if we merge this.

hqhq:
As the problem that we'll call `SetKernelMemory` twice on container creation, I can file a PR to fix this, but it didn't fix the problem, if modify time of `memory.kmem.limit_in_bytes` is still the same as `memory.kmem.max_usage_in_bytes` we'll fail if we want to update kernel memory of the container.

cyphar:
LGTM. I see this alot, and was one of my concerns with the original PR (using the modified time for attach logic). We should figure out a proper fix for this problem.

dqminh:
LGTM

dqminh:
@hqhq @cyphar @vishh  i've reopened #841

PR #962, opened 2016-07-23
Remove kmem Initialization check while setting memory configuration

dubstack:
I suggest that we modify the logic for setting the kmem.limit_on_bytes to be more deterministic and
remove the kmemInitialized check.

If kernel memory is enabled we limit the cgroup during cgroup creation ie. in apply(). This will enable kernel memory accounting on the cgroup ( http://lxr.free-electrons.com/source/Documentation/cgroups/memory.txt?v=3.9#L284 ). 
We limit the cgroup by writing some arbitrary limiting value(say 1) so that the kernel memory is accounted, and then we remove the limitation by writing -1. 

Once accounting is enabled we should be able to set the kernel memory limit to configuration value in the Set method.

The only scenario in which this would fail is if someone straight out calls Set() on a cgroup which was not created by libcontainer, in which case, if there are already process attached to the cgroup setting kernel memory limit would return an Ebusy signal and we can error out.

cc @mrunalp @derekwaynecarr @cyphar  @vishh @hqhq PTAL

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -5,15 +5,19 @@ package fs
 import (
 	"bufio"
 	"fmt"
-	"math"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
-	"github.com/opencontainers/runc/libcontainer/system"
+)
+
+const (
+	cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
 )
 
 type MemoryGroup struct {
@@ -34,9 +38,7 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 				return err
 			}
 		}
-		// We have to set kernel memory here, as we can't change it once
-		// processes have been attached to the cgroup.
-		if err := s.SetKernelMemory(path, d.config); err != nil {
+		if err := EnableKernelMemoryAccounting(path); err != nil {
 			return err
 		}
 	}
@@ -55,38 +57,43 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 	return nil
 }
 
-func (s *MemoryGroup) SetKernelMemory(path string, cgroup *configs.Cgroup) error {
-	// This has to be done separately because it has special
-	// constraints (it can only be initialized before setting up a
-	// hierarchy or adding a task to the cgroups. However, if
-	// sucessfully initialized, it can be updated anytime afterwards)
-	if cgroup.Resources.KernelMemory != 0 {
-		kmemInitialized := false
-		// Is kmem.limit_in_bytes already set?
-		kmemValue, err := getCgroupParamUint(path, "memory.kmem.limit_in_bytes")
-		if err != nil {
-			return err
-		}
-		switch system.GetLongBit() {
-		case 32:
-			kmemInitialized = uint32(kmemValue) != uint32(math.MaxUint32)
-		case 64:
-			kmemInitialized = kmemValue != uint64(math.MaxUint64)
-		}
-		if !kmemInitialized {
-			// If there's already tasks in the cgroup, we can't change the limit either
-			tasks, err := getCgroupParamString(path, "tasks")
-			if err != nil {
-				return err
-			}
-			if tasks != "" {
-				return fmt.Errorf("cannot set kmem.limit_in_bytes after task have joined this cgroup")
-			}
-		}
+func EnableKernelMemoryAccounting(path string) error {
+	// Check if kernel memory is enabled
+	// We have to limit the kernel memory here as it won't be accounted at all
+	// until a limit is set on the cgroup and limit cannot be set once the
+	// cgroup has children, or if there are already tasks in the cgroup.
+	kernelMemoryLimit := int64(1)
+	if err := setKernelMemory(path, kernelMemoryLimit); err != nil {
+		return err
+	}
+	kernelMemoryLimit = int64(-1)
+	if err := setKernelMemory(path, kernelMemoryLimit); err != nil {
+		return err
+	}
+	return nil
+}
 
-		if err := writeFile(path, "memory.kmem.limit_in_bytes", strconv.FormatInt(cgroup.Resources.KernelMemory, 10)); err != nil {
-			return err
+func setKernelMemory(path string, kernelMemoryLimit int64) error {
+	if path == "" {
+		return fmt.Errorf("no such directory for %s", cgroupKernelMemoryLimit)
+	}
+	if !cgroups.PathExists(filepath.Join(path, cgroupKernelMemoryLimit)) {
+		// kernel memory is not enabled on the system so we should do nothing
+		return nil
+	}
+	if err := ioutil.WriteFile(filepath.Join(path, cgroupKernelMemoryLimit), []byte(strconv.FormatInt(kernelMemoryLimit, 10)), 0700); err != nil {
+		// Check if the error number returned by the syscall is "EBUSY"
+		// The EBUSY signal is returned on attempts to write to the
+		// memory.kmem.limit_in_bytes file if the cgroup has children or
+		// once tasks have been attached to the cgroup
+		if pathErr, ok := err.(*os.PathError); ok {
+			if errNo, ok := pathErr.Err.(syscall.Errno); ok {
+				if errNo == syscall.EBUSY {
+					return fmt.Errorf("failed to set %s, because either tasks have already joined this cgroup or it has children", cgroupKernelMemoryLimit)
+				}
+			}
 		}
+		return fmt.Errorf("failed to write %v to %v: %v", kernelMemoryLimit, cgroupKernelMemoryLimit, err)
 	}
 	return nil
 }
@@ -139,8 +146,10 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 		return err
 	}
 
-	if err := s.SetKernelMemory(path, cgroup); err != nil {
-		return err
+	if cgroup.Resources.KernelMemory != 0 {
+		if err := setKernelMemory(path, cgroup.Resources.KernelMemory); err != nil {
+			return err
+		}
 	}
 
 	if cgroup.Resources.MemoryReservation != 0 {
@@ -148,6 +157,7 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 			return err
 		}
 	}
+
 	if cgroup.Resources.KernelMemoryTCP != 0 {
 		if err := writeFile(path, "memory.kmem.tcp.limit_in_bytes", strconv.FormatInt(cgroup.Resources.KernelMemoryTCP, 10)); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


mlaventure:
Are we happy for it to fail if joining an existing cgroup with a limit already sets? On this PR, the error message is not very explicit from a user perspective as it will just be:

```
process_linux.go:258: applying cgroup configuration for process caused "failed to write 1 to memory.kmem.limit_in_bytes: write /sys/fs/cgroup/memory/xxx/memory.kmem.limit_in_bytes: device or resource busy"
```

I think adding a more explanatory message to the returned error on `EBUSY` may be a good idea.

WDYT?

hqhq:
Yeah, that's what I concerned in https://github.com/opencontainers/runc/issues/841#issuecomment-234821328

dubstack:
@mlaventure @hqhq Please have a look at my comment here: https://github.com/opencontainers/runc/issues/841#issuecomment-235044904

libcontainer/cgroups/fs/memory.go
```diff
@@ -34,9 +32,15 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 				return err
 			}
 		}
-		// We have to set kernel memory here, as we can't change it once
-		// processes have been attached to the cgroup.
-		if err := s.SetKernelMemory(path, d.config); err != nil {
+		// We have to limit the kernel memory here as it won't be accounted at all
```
dubstack: We need to have a check here for determining if kmem is enabled on the system.
@hqhq @vishh Suggestions on how to check for the same?

vishh: Check if `memory.kmem.limit_in_bytes` file is there?

dubstack:
Updated the patch to add a more specific error message when kernel memory limit write fails with an EBUSY.
@mlaventure @hqhq @cyphar  @vishh PTAL.

mlaventure:
LGTM (IANAM).

libcontainer/cgroups/fs/memory.go
```diff
@@ -34,10 +38,20 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 				return err
 			}
 		}
-		// We have to set kernel memory here, as we can't change it once
-		// processes have been attached to the cgroup.
-		if err := s.SetKernelMemory(path, d.config); err != nil {
-			return err
+
+		// Check if kernel memory is enabled
+		if _, err := os.Stat(filepath.Join(path, cgroupKernelMemoryLimit)); err == nil {
```
vishh: nit use cgroups.PathExists(...)

vishh: Why not move this check to `SetKernelMemory(...)` method?

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,38 +69,22 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
-		}
-		if !kmemInitialized {
-			// If there's already tasks in the cgroup, we can't change the limit either
-			tasks, err := getCgroupParamString(path, "tasks")
-			if err != nil {
-				return err
-			}
-			if tasks != "" {
-				return fmt.Errorf("cannot set kmem.limit_in_bytes after task have joined this cgroup")
+func (s *MemoryGroup) SetKernelMemory(path string, kernelMemoryLimit int64) error {
```
vishh: Should this method be exported?

dubstack: Yeah I think this function should not be exposed. I just kept the original definition.

cyphar: I exported it because I use it in the systemd cgroup driver. AFAICS you haven't updated that at all (please make sure your changes haven't broken the systemd driver).

dubstack: https://github.com/opencontainers/runc/blob/ff88baa42fa5b2a1568a3a14665142fb4bdb3a2a/libcontainer/cgroups/systemd/apply_systemd.go#L487

Systemd has its own setkernelMemory method defined. I checked once again SetKernelMemory is not being used anywhere else in the code.

libcontainer/cgroups/fs/memory.go
dubstack: The idea was we SetKernelMemory() only if kernel memory is enabled. So have the check and then call the method.

But having the check inside the method would be more cleaner. I will make that change.

dubstack:
@cyphar @hqhq @mrunalp @vishh Addressed all comments. PTAL. This is blocking work in Kubernetes. Can we get this quickly reviewed and merged?

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,38 +67,26 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+		return nil
+	}
+	if err := ioutil.WriteFile(filepath.Join(path, cgroupKernelMemoryLimit), []byte(strconv.FormatInt(kernelMemoryLimit, 10)), 0700); err != nil {
+		// Check if the error number returned by the syscall is "EBUSY"
+		// The EBUSY signal is returned on attempts to write to the
+		// memory.kmem.limit_in_bytes file once tasks have been attached to the cgroup
+		if pathErr, ok := err.(*os.PathError); ok {
+			if errNo, ok := pathErr.Err.(syscall.Errno); ok {
+				if errNo == syscall.EBUSY {
+					return fmt.Errorf("Cannot set %s after tasks have joined this cgroup", cgroupKernelMemoryLimit)
```
hqhq: EBUSY could also be returned if there are child cgroups, we need to change error message a bit to cover that.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,38 +67,26 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
-	// hierarchy or adding a task to the cgroups. However, if
-	// sucessfully initialized, it can be updated anytime afterwards)
-	if cgroup.Resources.KernelMemory != 0 {
-		kmemInitialized := false
-		// Is kmem.limit_in_bytes already set?
-		kmemValue, err := getCgroupParamUint(path, "memory.kmem.limit_in_bytes")
-		if err != nil {
-			return err
-		}
-		switch system.GetLongBit() {
```
hqhq: Here is the only consumer of `GetLongBit`, so you can remove this function in `libcontainer/system/sysconfig.go` in this PR.

libcontainer/cgroups/fs/memory.go
hqhq: Don't forget to change the above comments.

hqhq: And error messages start with lower case in runc.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,38 +67,26 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 	return nil
 }
 
-func (s *MemoryGroup) SetKernelMemory(path string, cgroup *configs.Cgroup) error {
```
cyphar: I say again, should we update the systemd manager as well. Right now the systemd manager just does this to "set the kernel memory":

```
func setKernelMemory(c *configs.Cgroup) error {
	path, err := getSubsystemPath(c, "memory")
	if err != nil && !cgroups.IsNotFound(err) {
		return err
	}

	return os.MkdirAll(path, 0755)
}
```

This code **used** to call this function (`apply_raw.go:SetKernelMemory`) directly. It no longer does this, meaning that we're now maintaining two versions of different hacks to enable accounting for cgroups. Can we please fix this (combine the two), or at least verify that it works as intended?

mrunalp: Yep, we should fix up systemd cgroup manager as well.

libcontainer/cgroups/fs/memory.go
dubstack: SG. Thanks.

libcontainer/cgroups/fs/memory.go
dubstack: Right! Thanks

libcontainer/cgroups/fs/memory.go
dubstack: @cyphar @mrunalp IIUC, currently on systemd we are just creating the memory cgroup and not setting any kmem limit on them during cgroup creation. And it doesn't look like the intended behaviour. 

I will update my patch to have the same logic as cgroup fs in the systemd implementation.

The call to SetKernelMemory was removed in the following commit : #790. Maybe @mlaventure would have more idea as to why that change was made.

dubstack:
@cyphar @hqhq @mrunalp @mlaventure I have updated the logic in systemd to be same as that of cgroup fs. 

@mrunalp @derekwaynecarr Can you please have a look and maybe test it on a systemd system.

libcontainer/cgroups/fs/memory.go
```diff
@@ -55,38 +57,43 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
...
+	}
+	if err := ioutil.WriteFile(filepath.Join(path, cgroupKernelMemoryLimit), []byte(strconv.FormatInt(kernelMemoryLimit, 10)), 0700); err != nil {
+		// Check if the error number returned by the syscall is "EBUSY"
+		// The EBUSY signal is returned on attempts to write to the
+		// memory.kmem.limit_in_bytes file if the cgroup has children or
+		// once tasks have been attached to the cgroup
+		if pathErr, ok := err.(*os.PathError); ok {
+			if errNo, ok := pathErr.Err.(syscall.Errno); ok {
+				if errNo == syscall.EBUSY {
+					return fmt.Errorf("failed to set %s, because either tasks have already joined this cgroupe or it has chilren", cgroupKernelMemoryLimit)
```
hqhq: s/cgroupe/cgroup
s/chilren/children

dubstack: Thanks! I will make the change!

dubstack:
@cyphar @mrunalp @mlaventure friendly ping!

vishh:
This LGTM. Thanks for the cleanup @dubstack !!

@opencontainers/runc-maintainers Can we get a quick review on this? This PR is important.

mrunalp:
@hmeng-19 Could you test this PR on RHEL 7?

hmeng-19:
@mrunalp , I tested this on RHEL7. 
All the bats tests finished successfully. 
All the unit tests except for `TestInitJoinNetworkAndUser` (as mentioned in #915) also finished successfully.

mrunalp:
@hmeng-19 Thanks for testing.

mrunalp:
LGTM

hmeng-19:
@mrunalp , no problem.
To make it easy for others to test this PR, here is the testing command I used on RHEL7:

First, I created a basic config.json:
```
ocitools generate --tty --output=config.json \
--uidmappings=0:0:1024 --gidmappings=0:0:1024 \
--cgroups-path user.slice:hmeng:test2222 --mount-cgroups ro
```

Then add kernel memory limit into config.json:

            "memory": {
                "kernel": 50000000
            },

Finally, start the container and check the kernel memory limit:
```
runc --systemd-cgroup run 1234
runc run 1234
```

hqhq:
LGTM

hqhq:
@dubstack Sorry for not being specific, we should do more cleanups for GetLongBit, I've opened a follow up PR #968 for not blocking this.

PR #963, opened 2016-07-24
libcontainer: rename keyctl package to keys

guilhermebr:
This avoid the goimports tool from remove the libcontainer/keys import line due the package name is diferent from folder name

### DIFF
--- a/libcontainer/keys/keyctl.go
+++ b/libcontainer/keys/keyctl.go
@@ -1,6 +1,6 @@
 // +build linux
 
-package keyctl
+package keys
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


mrunalp:
Shouldn't this be fixed in `goimports`? :)

guilhermebr:
I think it's not a goimports bug because the Go convention says to use package name the same name of the last folder...

Maybe rename package to keys?

I just opened this PR because I added a log line in standard_init_linux.go for debugging and editor removed the import line. (I use vim-go with auto-goimports on save, many others editor I tested also come with this)

crosbymichael:
Ya, the package name in `libcontainer/keys/keyctl.go` should change from `package keyctl` to `package keys`.

guilhermebr:
Nice! @crosbymichael  so if you prefer I can update this PR changing the package name

crosbymichael:
@guilhermebr yes, that would be great! thanks

hqhq:
LGTM

cyphar:
@hqhq Still needs DCO.

cyphar:
LGTM

mrunalp:
LGTM

PR #964, opened 2016-07-25
UNITTEST: Bypass userns test on platform without userns support

zhaoleidd:
We should bypass userns test instead of show fail in platform without userns support.


mrunalp:
LGTM

hqhq:
LGTM

PR #965, opened 2016-07-26
Use root kmem to detect if kmem limit have been initialized

hqhq:
Alternative of #962
Fixes: #841

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -5,7 +5,6 @@ package fs
 import (
 	"bufio"
 	"fmt"
-	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -13,7 +12,6 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
-	"github.com/opencontainers/runc/libcontainer/system"
 )
 
 type MemoryGroup struct {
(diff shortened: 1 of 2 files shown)


hqhq:
ping @vishh @mlaventure @cyphar @mrunalp @dubstack

cyphar:
Does this work properly for cases where `memory.use_hierarchy = 0`? Also, does this properly handle cases where the limit is updated in a parent after we've created the child?

libcontainer/cgroups/fs/memory.go
```diff
@@ -67,12 +65,15 @@ func (s *MemoryGroup) SetKernelMemory(path string, cgroup *configs.Cgroup) error
 		if err != nil {
 			return err
 		}
-		switch system.GetLongBit() {
-		case 32:
-			kmemInitialized = uint32(kmemValue) != uint32(math.MaxUint32)
-		case 64:
-			kmemInitialized = kmemValue != uint64(math.MaxUint64)
+		memcg, err := cgroups.FindCgroupMountpoint("memory")
```
vishh: Will this work when runc is used inside of a container?

hqhq:
@cyphar Yes I think it will work properly for cases where `memory.use_hierarchy = 0`, but for cases that kernel memory will be updated before, if kernel memory limit is set to a number, than updated to `-1` which will reset kernel memory to unlimited, logic in this PR will consider kernel memory as uninitialized. This issue is also existed on the current version.

@vishh Yeah there will be issue when runc is used inside of a container.

This PR is not workable, I'll close, thank you.

PR #966, opened 2016-07-26
fix init.scope in cgroup paths

### LINKED ISSUE #931: systemd-cgroup uses incorrect subsystem path in recent systemd versions
This issue was discussed in a comment a while ago: https://github.com/docker/docker/issues/16256#issuecomment-218935000

Now that systemd 226+ is more common, this issue has further ramifications.

Under systemd versions of at least 226, `systemd` is under `init.scope`. The code [here](https://github.com/opencontainers/runc/blob/629e35666d31f743090452416c8df207d9fdbd34/libcontainer/cgroups/systemd/apply_systemd.go#L387) behaves incorrectly in this configuration.

The following is an example of what goes wrong in this configuration:

```
$ runc --systemd-cgroup run foo
$ cat /proc/$(pidof foo)/cgroups
9:net_cls:/system.slice/runc-foo.scope
8:memory:/init.scope/system.slice/runc-foo.scope
7:blkio:/init.scope/system.slice/runc-foo.scope
6:devices:/init.scope/system.slice/runc-foo.scope
5:cpuset:/system.slice/runc-foo.scope
4:perf_event:/system.slice/runc-foo.scope
3:cpu,cpuacct:/init.scope/system.slice/runc-foo.scope
2:freezer:/system.slice/runc-foo.scope
1:name=systemd:/system.slice/runc-foo.scope
$ # Note that some subsystems have `init.scope` as part of their path
$ head /sys/fs/cgroup/cpu,cpuacct/init.scope/system.slice/runc-foo.scope/tasks /sys/fs/cgroup/cpu,cpuacct/system.slice/runc-foo.scope/tasks
==> /sys/fs/cgroup/cpu,cpuacct/init.scope/system.slice/runc-foo.scope/tasks <==
1085
1090
1091
1092
==> /sys/fs/cgroup/cpu,cpuacct/system.slice/runc-foo.scope/tasks <==
$ 
```

Essentially, the problem is as described in the above comment: runc thinks some cgroup subsystems should be under `init.scope` under the mistaken assumption that the cgroup of pid 1 is always the root of the cgroup tree. In reality, everything should just be under the real root, `/`.

I've begun putting together a patch (currently [WIP](https://github.com/euank/runc/commits/fix-systemd-init-scope)), but wanted an issue to point at.

This impacts projects, such as cadvisor, which read container's cgroups.

cc @vishh @marineam

sjenning:
Fixes #931 

There was talk in the issue about using dbus to determine which cgroup contains systemd, but I'm not seeing how that is relevant.  Seems like we just want to always start at the cgroup root.

@derekwaynecarr @vishh @euank @mrunalp

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -388,6 +388,8 @@ func getSubsystemPath(c *configs.Cgroup, subsystem string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	// if pid 1 is systemd 226 or later, it will be in init.scope, not the root
+	initPath = strings.TrimSuffix(filepath.Clean(initPath), "init.scope")
 
 	slice := "system.slice"
 	if c.Parent != "" {


euank:
The hardcoded string `system.slice` is what you'd use a systemd dbus call to replace iiuc.

All the same, LGTM as a definite improvement over the existing state of things. Thanks!

cyphar:
The reason why we do this is because of the usecase where someone is running a container inside a container (currently we don't have support for the cgroup namespace, and there's no requirement that a user will use that namespace even if we did support it -- see #781). If you start joining cgroups at the root, things start to not go very well (systemd can't track the process anymore and it doesn't like that very much).

Now, for the `apply_raw` cgroup handler we check against `/proc/1/cgroups` but if you tell runC to specifically use systemd for cgroup handling that's what we will do instead.

/cc @crosbymichael

sjenning:
@cyphar thanks for explaining that.  I have pushed a different possible fix.

The idea is that if "init.scope" is in the initPath, we should remove it, and it will only every appear once at the end of the initPath.  If pid 1 is not systemd, nothing changes.  If pid 1 is systemd < 226, nothing changes.  If pid 1 is systemd >= 226, "init.scope" is removed from the initPath.

Does this work?

mrunalp:
@sjenning This sounds good to me. IMO, this is the least intrusive fix for now.

mrunalp:
LGTM

mrunalp:
@cyphar @crosbymichael @hqhq PTAL

cyphar:
I'm wondering if we should use `strings.TrimSuffix`. This will replace the first occurrence of init.scope in the path, but we actually want to remove the last one -- right?

sjenning:
@cyphar I can do that.  There should only ever be one occurrence of init.scope in the path but I can change it just to make it more self-documenting.

sjenning:
@cyphar updated PTAL.  Also I think Jenkins flaked, might need someone retry that.

mrunalp:
I've kicked off the tests again.

cyphar:
LGTM.

mrunalp:
LGTM

PR #967, opened 2016-08-01
Change cgroup path creation for new systemd

johnbieren:
In the tests/integration directory, cgroups.bats and update.bats finds the path to the cgroup through 
base_path=$(grep "cgroup"  /proc/self/mountinfo | gawk 'toupper($NF) ~ /\<'${g}'\>/ { print $5; exit }')
In newer versions of systemd, this fails because it does not capture the entire path, including the splice part of the path.  
For example, the current code looks in /sys/fs/cgroup/memory for {test_name}/memory.kmem_limit_in_bytes, whereas with newer versions of systemd, it should be looking in /sys/fs/cgroup/user.slice/user-X.slice/session-X.scope/

This fix checks to make sure that the file is there before proceeding with the test, and if it is not there, it searches again, which fixes the issue with newer systemd versions.  Because of the if statement, older versions of systemd are not broken, nothing changes in that case.

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -52,6 +52,11 @@ EOF
     [ "$status" -eq 0 ]
     wait_for_container 15 1 test_cgroups_kmem
 
+    # Make sure $CGROUP_MEMORY dir was found correctly
+    if [ ! -e $CGROUP_MEMORY ] ; then
+      CGROUP_MEMORY=$(find $base_path -name "${TEST_CGROUP_NAME}")
+    fi
+
     # update kernel memory limit
     runc update test_cgroups_kmem --kernel-memory 50331648
     [ "$status" -eq 0 ]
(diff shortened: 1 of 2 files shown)


hqhq:
I think we should fix the implementation about how we get the cgroup path, it is wrong that we assume current process is in root cgroup, we should combine paths from `/proc/self/mountinfo` and `/proc/self/cgroup`.

hqhq:
After a second thought,  combine paths from `/proc/self/mountinfo` and `/proc/self/cgroup` may not right if we running test in container, maybe we should use absolute path for `cgroupsPath` when running the container.

hqhq:
@johnbieren I pushed #974 , can you confirm if it fixes your problem?

johnbieren:
@hqhq Yes, I can confirm that https://github.com/opencontainers/runc/pull/974 fixes my problem, and it also does not break the tests for older versions of systemd.  Great fix, much cleaner than mine =)

PR #972, opened 2016-08-05
tests: add requires cgroups_kmem

brauner:
On older kernels or kernels were CONFIG_MEMCG_KMEM is not set some cgroup tests
cannot be run. We simply test for the existence of the file
"/sys/fs/cgroup/memory/memory.kmem.limit_in_bytes" which sould be sufficient to
conclude that CONFIG_MEMCG_KMEM is not set.

Signed-off-by: Christian Brauner <cbrauner@suse.de>


tests/integration/update.bats
```diff
@@ -48,6 +48,7 @@ function check_cgroup_value() {
 }
 
 @test "update" {
+    requires cgroups_kmem
```
cyphar: Maybe we should split this test into a version that just tests kernel memory updating? I don't like skipping all of the `update` code if you're on an older kernel.

johnbieren:
The problem of the path to memory.kmem.limit_in_bytes is the problem I was addressing in https://github.com/opencontainers/runc/pull/967 as well, but I proposed finding the memory.kmem.limit_in_bytes file if it was not where expected instead of skipping testing

cyphar:
@johnbieren Your PR is _related_ to this one but is quite different. This PR is meant to fix the problem when you have an older kernel that doesn't support kernel memory accounting. Your PR fixes the problem that systemd changes what cgroup you are associated with.

tests/integration/update.bats
hqhq: Yeah, maybe split into several functions like:
```
update_blkio()
update_cpu()
update_cpuset()
update_memory()
```

But I think we can do it in a separate PR.

tests/integration/helpers.bash
```diff
@@ -19,6 +19,9 @@ HELLO_BUNDLE="$BATS_TMPDIR/hello-world"
 # CRIU PATH
 CRIU="/usr/local/sbin/criu"
 
+# CONFIG_MEMCG_KMEM support
+KMEM="/sys/fs/cgroup/memory/memory.kmem.limit_in_bytes"
```
hqhq: This is not always right, we should parse it from `/proc/self/mountinfo` like we did in `cgroups.bats`.

hqhq: To be clear, current implementation in `cgroups.bats` is not right, see my comments in https://github.com/opencontainers/runc/pull/967#issuecomment-238060361

brauner: > This is not always right, we should parse it from /proc/self/mountinfo like we did in cgroups.bats.

We just want to determine whether `CONFIG_MEMCG_KMEM` is enabled. If it is, then we can e.g. check in any cgroup in the memory hierarchy whether the `memory.kmem.limit_in_bytes` file exists. We don't need to know the containers cgroup path for that.

brauner: @cyphar just pointed out that you might be implicating that cgroups might not be mounted on `/sys/fs/cgroup` but somewhere else and your idea is to parse `/proc/self/mountinfo` to find the cgroup mountpoint.

brauner:
Fwiw, I updated the PR and to use the cgroup information available via  `/proc/self/mountinfo`.

hqhq:
LGTM

cyphar:
LGTM

PR #974, opened 2016-08-07
Use absolute cgroup path for integration test

hqhq:
So we can pass the test in container os local or systemd
environment.

Also fixes: #967


cyphar:
LGTM.

mrunalp:
LGTM

PR #975, opened 2016-08-08
nsenter: set {uid,gid} explicitly around namespace creation

cyphar:
In general, it is a bad idea to be unmapped inside a user namespace at
any point (especially when euid=[kuid 0]) as it can lead to security
vulnerabilities. Also, in certain SELinux setups you must also be mapped
in your user namespace when unsharing other namespaces.

Deal with all of this by parsing the {uid,gid}maps and then setresuid(2)
to the right user before and after the critical unshare(CLONE_NEWUSER)
(as well as dealing with setns(2) by changing user to the owner of the
namespace file we're joining).

Fixes: CVE-2015-8709
Reported-by: Andrey Vagin avagin@virtuozzo.com
Reported-by: Mrunal Patel mpatel@redhat.com

Fixes #959.

Note: This is based on ~~#950~~ and #977.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -1,4 +1,5 @@
 #define _GNU_SOURCE
+#include <ctype.h>
 #include <endian.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -19,6 +20,7 @@
 #include <sys/prctl.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <sys/stat.h>
 
 #include <linux/limits.h>
 #include <linux/netlink.h>
@@ -63,7 +65,13 @@ struct clone_t {
 };
 
 struct nlconfig_t {
+	/*
+	 * Stores a pointer to the netlink data payload. All other pointers
+	 * are inside this block. Free it with nl_free().
+	 */
 	char *data;
+
+	/* Options sent from the bootstrap process. */
 	uint32_t cloneflags;
 	char *uidmap;
 	size_t uidmap_len;
@@ -72,6 +80,17 @@ struct nlconfig_t {
 	char *namespaces;
 	size_t namespaces_len;
 	uint8_t is_setgroup;
+
+	/*
+	 * Namespace uids and gids. If cloneflags doesn't contain
+	 * CLONE_NEWUSER then these will be 0. host{uid,gid} are from the
+	 * host's perspective and root{uid,gid} are from the container's
+	 * perspective.
+	 */
+	uid_t hostuid;
+	gid_t hostgid;
+	uid_t rootuid;
+	gid_t rootgid;
 };
 
 /*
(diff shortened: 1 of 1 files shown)


cyphar:
NOTE: `map_field` is slightly broken at the moment (it currently only can parse the first line of the map data. I'll fix this later, but right now don't merge this.

libcontainer/nsenter/nsexec.c
```diff
@@ -399,9 +472,40 @@ void join_namespaces(char *nslist)
 		ns->fd = fd;
 		ns->ns = nsflag(namespace);
 		strncpy(ns->path, path, PATH_MAX);
+
+		/* Cache the user namespace entry. */
+		if (ns->ns == CLONE_NEWUSER || !strcmp(namespace, "user"))
+			userns = ns;
```
cyphar: IIRC I wrote it like this to avoid issues where `CLONE_NEWUSER` is not defined by glibc (so `nsflags` would return 0). Though I might have fixed that in the meantime, since `CLONE_NEWUSER` is required by a bunch of other things.

cyphar:
This has been rebased. @avagin can you double-check that this fix still is fine, since I had to un-split the `unshare(CLONE_NEWUSER)` I'm worried that we haven't actually fixed the issue.

/cc @opencontainers/runtime-spec-maintainers

crosbymichael:
LGTM

mrunalp:
unshare fails with EPERM on RHEL 7.

libcontainer/nsenter/nsexec.c
```diff
@@ -619,7 +724,20 @@ void nsexec(void)
...
+			 * If we already joined a user namespace this is also fine (we do a
+			 * similar operation in join_namespaces). There are some security
+			 * issues with having an euid=(kuid 0) inside a user namespace.
+			 */
+
+			/* Switch groups first, the order is important. */
+			if (setresgid(config.hostgid, config.hostgid, config.hostgid) < 0)
+				bail("failed to set gid to host");
+			if (setresuid(config.hostuid, config.hostuid, config.hostuid) < 0)
+				bail("failed to set uid to host");
```
cyphar: @mrunalp I guess this is what's causing the failure. The problem is that this **is** the fix for the security issue (a racing ptrace can cause root privilege escalation inside a container). I'm starting to think that the issues here are on the SELinux side and not on the runC side...

mrunalp: @cyphar I'll look closer but it looked like the split of unshare or NEWUSER and the other cloneflags.

cyphar: Recently, I read an email on a kernel mailing list about "safely creating container processes" and the process for doing it safely is actually quite a bit more complicated than what we're doing now. Essentially the method requires creating a "sacrifical" user namespace before we create all of the actual container namespaces (you `unshare(CLONE_NEWUSER)` twice). The problem is that because of SELinux we simply can't implement that -- which IMO is a serious problem.

I'll see if I can find the email. My point is that I feel like the issues are on the kernel side of things, because I'm trying to make this process safer and SELinux is having problems with it. If worse comes to worst we can have SELinux-specific code here that changes how we create containers -- it won't be pretty but at least container creation will be safer without breaking SELinux support.

cyphar:
Rebased.

@mrunalp If we get this merged then the split code for #959 will look much better in my opinion (it would be a few lines apart as opposed to being in separate processes).

dqminh:
@cyphar can you rebase this change if still relevant ?

cyphar:
@dqminh Yes, and I'm going to carry this and #959 to see if we can finally resolve these issues in a more sane way.

cyphar:
#1562 is my attempt to carry this and #975 together.

crosbymichael:
Closing since #1562 takes care of this.  Just cleaning up the milestone some.

PR #976, opened 2016-08-08
nsenter: correctly handle pidns orphaning

### LINKED ISSUE #971: Joining existing pid ns not reparenting process
We must be missing the double fork, like in the exec case, to reparent the process to the container's pid 1.

See https://github.com/docker/docker/issues/25348 for more details.

cyphar:
The main change here is to make sure that new processes that join a container are correctly reparented. Read [this LWN article](https://lwn.net/Articles/532748/) for more information about what the issue is and how the solution needs to be structured.

`TODO`:
- [x] Double-fork inside stage 2 to get reparented to the container's init (if we're an exec process).
  - [x] This is broken because we currently send the PID of the wrong process in stage 1.
- [ ] `runc exec` doesn't exit after the subprocess exits because of this feature. We'll need to think up a new way of handling this...

Fixes #971 

Note: This is based on ~~#950~~, #977 and #975.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -24,27 +24,51 @@
 #include <linux/netlink.h>
 #include <linux/types.h>
 
-#define SYNC_VAL 0x42
-#define JUMP_VAL 0x43
+/* Get all of the CLONE_NEW* flags. */
+#include "namespace.h"
+
+/* Synchronisation values. */
+enum sync_t {
+	SYNC_USERMAP_PLS = 0x40, /* Request parent to map our users. */
+	SYNC_USERMAP_ACK = 0x41, /* Mapping finished by the parent. */
+	SYNC_RECVPID_PLS = 0x42, /* Tell parent we're sending the PID. */
+	SYNC_RECVPID_ACK = 0x43, /* PID was correctly received by parent. */
+
+	/* XXX: This doesn't help with segfaults and other such issues. */
+	SYNC_ERR = 0xFF, /* Fatal error, no turning back. The error code follows. */
+};
+
+/* longjmp() arguments. */
+#define JUMP_PARENT 0x00
+#define JUMP_CHILD  0xA0
+#define JUMP_INIT   0xA1
+
+/* JSON buffer. */
+#define JSON_MAX 4096
 
 /* Assume the stack grows down, so arguments should be above it. */
-struct clone_arg {
+struct clone_t {
 	/*
 	 * Reserve some space for clone() to locate arguments
 	 * and retcode in this place
 	 */
 	char stack[4096] __attribute__ ((aligned(16)));
 	char stack_ptr[0];
+
+	/* There's two children. This is used to execute the different code. */
 	jmp_buf *env;
+	int jmpval;
 };
 
 struct nlconfig_t {
 	char *data;
 	uint32_t cloneflags;
 	char *uidmap;
-	int uidmap_len;
+	size_t uidmap_len;
 	char *gidmap;
-	int gidmap_len;
+	size_t gidmap_len;
+	char *namespaces;
+	size_t namespaces_len;
 	uint8_t is_setgroup;
 	int consolefd;
 };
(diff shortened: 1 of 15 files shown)


cyphar:
@opencontainers/runc-maintainers This sort of works. However, because of the new feature (that `exec` processes are no longer children of the runC process) there's a lot of issues with `waitProcess` in the libcontainer integration tests that I can't quite crack. I've managed to sort-of fix the issues with `runC` (but it requires some not-very nice hacks where I call `os.Exit` from a goroutine).

Is there a nicer way of handling the exit of a child exec'd process other than just getting an `EIO` from the `stdio` of the process and then exiting blindly?

cyphar:
Alright, I've fixed all of the issues through some fairly nasty code in 247d3550407912aa7ce46e628f9b79a2580ad5a7 (`[wip] libcontainer: handle exit of exec'd processes`). Unfortunately, I don't think there's a clean way of fixing it -- we won't ever get a `SIGCHLD` so we just have to parse everything and poll.

libcontainer/system/proc.go
```diff
@@ -25,3 +30,30 @@ func GetProcessStartTime(pid int) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	parts := strings.Split(string(data), " ")
+	// the state field is located in pos 3
+	return parts[3-1], nil // starts at 1
+}
+
+// GetProcessParent reads reads /proc/<pid>/stat to determine the parent of a
```
datawolf: two `reads` words, should delete one.

libcontainer/system/proc.go
```diff
@@ -25,3 +30,30 @@ func GetProcessStartTime(pid int) (string, error) {
...
+// process.
+func GetProcessParent(pid int) (int, error) {
+	data, err := ioutil.ReadFile(filepath.Join("/proc", strconv.Itoa(pid), "stat"))
+	if err != nil {
+		return -1, err
+	}
+
+	parts := strings.Split(string(data), " ")
+	// the state field is located in pos 4
```
cyphar: Thanks, but note that this PR is in quite a lot of flux (it doesn't really work at the moment and I'm trying to figure out why the hell processes are so hard on Linux).

libcontainer/nsenter/nsexec.c
```diff
@@ -548,8 +584,13 @@ void nsexec(void)
 		bail("missing cloneflags");
 
 	/* Pipe so we can tell the child when we've finished setting up. */
-	if (socketpair(AF_LOCAL, SOCK_STREAM, 0, syncpipe) < 0)
+	if (socketpair(AF_LOCAL, SOCK_STREAM, 0, parentpipe) < 0)
```
datawolf: The comment for function `sendpid` and `recvpid` using `AF_UNIX`.  Although `AF_LOCAL` and `AF_UNIX` are equal, I think it should be consistent.

libcontainer/nsenter/nsexec.c
```diff
@@ -61,7 +63,13 @@ struct clone_t {
 };
 
 struct nlconfig_t {
+	/*
+	 * Stores a pointer to the netlink data payload. All other pointers
+	 * are inside this block. Free it with nl_free().
+	 */
```
datawolf: @cyphar I want to know that why we need to store the pointer to the netlink data payload.  After we call `nl_parse` function, we will never use the payload data. Whether can we free the pointer(allocated by malloc in the function `nl_parse`) at the end `nl_parse` and delete the funciton `nl_free`?

sandyskies:
@cyphar  I merge ur fixed with runc of docker 1.12.5 but it does't work . An error was reported:
 docker run -ti 427deb629ee5 bash 
docker: Error response from daemon: containerd: container not started.

cyphar:
I'll be honest, I'm not sure that this code can ever be completely sane. I'll need to think about this some more, and I've been _very_ busy with other things recently. The big issue is that the polling system for `/proc/[pid]/stat` is just not a very good solution for tracking the death of a process, and it's just not fit-for-purpose if we want to get exit codes...

dqminh:
@cyphar is this still relevant ?

cyphar:
@dqminh I believe it's still relevant (we still have pidns orphaning problems) but I think that the exit-code problem is not really resolvable (and polling `/proc/$pid/stat` is a horrible way of not completely solving the problem as well).

cyphar:
This might be doable with `cn_proc` but this requires more kernel work to be possible as an unprivileged user. http://netsplit.com/the-proc-connector-and-socket-filters

cyphar:
We need more kernel work for this.

AkihiroSuda:
What's current status?

PR #977, opened 2016-08-08
nsenter: guarantee correct user namespace ordering

cyphar:
Depending on your SELinux setup, the order in which you join namespaces
can be important. In general, user namespaces should *always* be joined
and unshared first because then the other namespaces are correctly
pinned and you have the right priviliges within them. This also is very
useful for rootless containers, as well as older kernels that had
essentially broken unshare(2) and clone(2) implementations.

This also includes huge refactorings in how we spawn processes for
complicated reasons that I don't want to get into because it will make
me spiral into a cloud of rage. The reasoning is in the giant comment in
clone_parent. Have fun.

In addition, because we now create multiple children with CLONE_PARENT,
we cannot wait for them to SIGCHLD us in the case of a death. Thus, we
have to resort to having a child kindly send us their exit code before
they die. Hopefully this all works okay, but at this point there's not
much more than we can do.

`TODO`:
* [x] Tag each namespace we want to join, to avoid future bugs.
* [x] Guarantee correct namespace ordering.

~~This is based on #950.~~ Base has been merged.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1223,16 +1223,21 @@ func (c *linuxContainer) currentState() (*State, error) {
 // can setns in order.
 func (c *linuxContainer) orderNamespacePaths(namespaces map[configs.NamespaceType]string) ([]string, error) {
 	paths := []string{}
-	nsTypes := []configs.NamespaceType{
+	order := []configs.NamespaceType{
 		configs.NEWIPC,
 		configs.NEWUTS,
 		configs.NEWNET,
 		configs.NEWPID,
 		configs.NEWNS,
+		configs.NEWUSER,
 	}
-	// join userns if the init process explicitly requires NEWUSER
-	if c.config.Namespaces.Contains(configs.NEWUSER) {
-		nsTypes = append(nsTypes, configs.NEWUSER)
+
+	// Remove namespaces that we don't need to join.
+	var nsTypes []configs.NamespaceType
+	for _, ns := range order {
+		if c.config.Namespaces.Contains(ns) {
+			nsTypes = append(nsTypes, ns)
+		}
 	}
 	for _, nsType := range nsTypes {
 		if p, ok := namespaces[nsType]; ok && p != "" {
@@ -1249,7 +1254,7 @@ func (c *linuxContainer) orderNamespacePaths(namespaces map[configs.NamespaceTyp
 			if strings.ContainsRune(p, ',') {
 				return nil, newSystemError(fmt.Errorf("invalid path %s", p))
 			}
-			paths = append(paths, p)
+			paths = append(paths, fmt.Sprintf("%s:%s", configs.NsName(nsType), p))
 		}
 	}
 	return paths, nil
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1224,12 +1224,13 @@ func (c *linuxContainer) currentState() (*State, error) {
 func (c *linuxContainer) orderNamespacePaths(namespaces map[configs.NamespaceType]string) ([]string, error) {
 	paths := []string{}
 	order := []configs.NamespaceType{
+		// The user namespace *must* be done first.
+		configs.NEWUSER,
 		configs.NEWIPC,
 		configs.NEWUTS,
 		configs.NEWNET,
 		configs.NEWPID,
 		configs.NEWNS,
-		configs.NEWUSER,
 	}
 
 	// Remove namespaces that we don't need to join.
--- a/libcontainer/nsenter/namespace.h
+++ b/libcontainer/nsenter/namespace.h
@@ -1,0 +1,32 @@
+#ifndef NSENTER_NAMESPACE_H
+#define NSENTER_NAMESPACE_H
+
+#ifndef _GNU_SOURCE
+#	define _GNU_SOURCE
+#endif
+#include <sched.h>
+
+/* All of these are taken from include/uapi/linux/sched.h */
+#ifndef CLONE_NEWNS
+#	define CLONE_NEWNS 0x00020000 /* New mount namespace group */
+#endif
+#ifndef CLONE_NEWCGROUP
+#	define CLONE_NEWCGROUP 0x02000000 /* New cgroup namespace */
+#endif
+#ifndef CLONE_NEWUTS
+#	define CLONE_NEWUTS 0x04000000 /* New utsname namespace */
+#endif
+#ifndef CLONE_NEWIPC
+#	define CLONE_NEWIPC 0x08000000 /* New ipc namespace */
+#endif
+#ifndef CLONE_NEWUSER
+#	define CLONE_NEWUSER 0x10000000 /* New user namespace */
+#endif
+#ifndef CLONE_NEWPID
+#	define CLONE_NEWPID 0x20000000 /* New pid namespace */
+#endif
+#ifndef CLONE_NEWNET
+#	define CLONE_NEWNET 0x40000000 /* New network namespace */
+#endif
+
+#endif /* NSENTER_NAMESPACE_H */
--- a/libcontainer/nsenter/nsenter_test.go
+++ b/libcontainer/nsenter/nsenter_test.go
@@ -29,7 +29,7 @@ func TestNsenterValidPaths(t *testing.T) {
 
 	namespaces := []string{
 		// join pid ns of the current process
-		fmt.Sprintf("/proc/%d/ns/pid", os.Getpid()),
+		fmt.Sprintf("pid:/proc/%d/ns/pid", os.Getpid()),
 	}
 	cmd := &exec.Cmd{
 		Path:       os.Args[0],
@@ -87,7 +87,47 @@ func TestNsenterInvalidPaths(t *testing.T) {
 
 	namespaces := []string{
 		// join pid ns of the current process
-		fmt.Sprintf("/proc/%d/ns/pid", -1),
+		fmt.Sprintf("pid:/proc/%d/ns/pid", -1),
+	}
+	cmd := &exec.Cmd{
+		Path:       os.Args[0],
+		Args:       args,
+		ExtraFiles: []*os.File{child},
+		Env:        []string{"_LIBCONTAINER_INITPIPE=3"},
+	}
+
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+	// write cloneFlags
+	r := nl.NewNetlinkRequest(int(libcontainer.InitMsg), 0)
+	r.AddData(&libcontainer.Int32msg{
+		Type:  libcontainer.CloneFlagsAttr,
+		Value: uint32(syscall.CLONE_NEWNET),
+	})
+	r.AddData(&libcontainer.Bytemsg{
+		Type:  libcontainer.NsPathsAttr,
+		Value: []byte(strings.Join(namespaces, ",")),
+	})
+	if _, err := io.Copy(parent, bytes.NewReader(r.Serialize())); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := cmd.Wait(); err == nil {
+		t.Fatalf("nsenter exits with a zero exit status")
+	}
+}
+
+func TestNsenterIncorrectPathType(t *testing.T) {
+	args := []string{"nsenter-exec"}
+	parent, child, err := newPipe()
+	if err != nil {
+		t.Fatalf("failed to create pipe %v", err)
+	}
+
+	namespaces := []string{
+		// join pid ns of the current process
+		fmt.Sprintf("net:/proc/%d/ns/pid", os.Getpid()),
 	}
 	cmd := &exec.Cmd{
 		Path:       os.Args[0],
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -24,27 +24,51 @@
 #include <linux/netlink.h>
 #include <linux/types.h>
 
-#define SYNC_VAL 0x42
-#define JUMP_VAL 0x43
+/* Get all of the CLONE_NEW* flags. */
+#include "namespace.h"
+
+/* Synchronisation values. */
+enum sync_t {
+	SYNC_USERMAP_PLS = 0x40, /* Request parent to map our users. */
+	SYNC_USERMAP_ACK = 0x41, /* Mapping finished by the parent. */
+	SYNC_RECVPID_PLS = 0x42, /* Tell parent we're sending the PID. */
+	SYNC_RECVPID_ACK = 0x43, /* PID was correctly received by parent. */
+
+	/* XXX: This doesn't help with segfaults and other such issues. */
+	SYNC_ERR = 0xFF, /* Fatal error, no turning back. The error code follows. */
+};
+
+/* longjmp() arguments. */
+#define JUMP_PARENT 0x00
+#define JUMP_CHILD  0xA0
+#define JUMP_INIT   0xA1
+
+/* JSON buffer. */
+#define JSON_MAX 4096
 
 /* Assume the stack grows down, so arguments should be above it. */
-struct clone_arg {
+struct clone_t {
 	/*
 	 * Reserve some space for clone() to locate arguments
 	 * and retcode in this place
 	 */
 	char stack[4096] __attribute__ ((aligned(16)));
 	char stack_ptr[0];
+
+	/* There's two children. This is used to execute the different code. */
 	jmp_buf *env;
+	int jmpval;
 };
 
 struct nlconfig_t {
 	char *data;
 	uint32_t cloneflags;
 	char *uidmap;
-	int uidmap_len;
+	size_t uidmap_len;
 	char *gidmap;
-	int gidmap_len;
+	size_t gidmap_len;
+	char *namespaces;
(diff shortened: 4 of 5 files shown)


cyphar:
Alright @opencontainers/runc-maintainers this is the next PR in the "rewriting nsenter" series of patches. PTAL, the changes here are quite a bit more intrusive than #950.

libcontainer/nsenter/nsexec.c
```diff
@@ -184,18 +168,28 @@ static int write_file(char *data, size_t data_len, char *pathfmt, ...)
...
-#define SETGROUPS_DENY  "deny"
+enum policy_t {
+	SETGROUPS_DEFAULT = 0,
+	SETGROUPS_ALLOW,
+	SETGROUPS_DENY,
+};
 
 /* This *must* be called before we touch gid_map. */
-static void update_setgroups(int pid, bool setgroup)
+static void update_setgroups(int pid, enum policy_t setgroup)
```
dqminh: i dont think we ever set this to deny or noop when we go into here ?, seems like something like `allow_setgroup(pid)` is simpler.

libcontainer/container_linux.go
```diff
@@ -1249,7 +1255,7 @@ func (c *linuxContainer) orderNamespacePaths(namespaces map[configs.NamespaceTyp
 			if strings.ContainsRune(p, ',') {
 				return nil, newSystemError(fmt.Errorf("invalid path %s", p))
 			}
-			paths = append(paths, p)
+			paths = append(paths, fmt.Sprintf("%s:%s", configs.NsName(nsType), p))
```
dqminh: i'm still not sure if setting namespace type here is necessary. Seems like something we do with no gain since the namespace type is already "known" (implicitly) ( and result in more code )

libcontainer/nsenter/nsexec.c
```diff
@@ -413,60 +455,311 @@ void nsexec(void)
...
 
-		/* Close the writing side of pipe. */
-		close(syncpipe[1]);
+	/*
+	 * Okay, so this is quite annoying.
+	 *
+	 * In order to make sure that deal with older kernels (when CLONE_NEWUSER
+	 * wasn't guaranteed to be done first if you specify multiple namespaces in
+	 * a clone(2) invocation) as well as with certain usecases like rootless
+	 * containers, we cannot just dump all of the cloneflags into clone(2).
```
dqminh: i dont think this affect rootless containers, maybe i missed something. Do you have any reference on the failure cases ?

libcontainer/container_linux.go
cyphar: > since the namespace type is already "known" (implicitly)

It isn't though. All we have is a path from the perspective of `nsenter`. And we need this for the code in #976 (we only can double fork if we've joined a PID namespace).

libcontainer/nsenter/nsexec.c
cyphar: For rootless containers, we need to set deny. This patch is based on the rootless containers patch, and I don't see why we have to restrict it (it'll be useful once I rebase the rootless containers PR on top of the nsenter cleanup -- something I'm not looking forward to).

libcontainer/nsenter/nsexec.c
cyphar: We need to map the `{uid, gid}` as well as do the `setgroup` stuff before we can properly unshare the other namespaces with a rootless container setup(IIRC). There were some issues that I ran into when implementing rootless containers which were solved by correctly separating the `unshare` and `clone` components.

cyphar:
Rebase'd.

crosbymichael:
getting a ton of issues building this patch.

```
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:200:4: note: in expansion of macro ‘bail’
    bail("failed to write '%s' to /proc/%d/setgroups", policy, pid);
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:137:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    write(syncfd, &ret, sizeof(ret));  \
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:200:4: note: in expansion of macro ‘bail’
    bail("failed to write '%s' to /proc/%d/setgroups", policy, pid);
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c: In function ‘update_uidmap’:
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:136:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    write(syncfd, &s, sizeof(s));   \
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:210:3: note: in expansion of macro ‘bail’
   bail("failed to update /proc/%d/uid_map", pid);
   ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:137:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    write(syncfd, &ret, sizeof(ret));  \
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:210:3: note: in expansion of macro ‘bail’
   bail("failed to update /proc/%d/uid_map", pid);
   ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c: In function ‘update_gidmap’:
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:136:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    write(syncfd, &s, sizeof(s));   \
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:219:3: note: in expansion of macro ‘bail’
   bail("failed to update /proc/%d/gid_map", pid);
   ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:137:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    write(syncfd, &ret, sizeof(ret));  \
    ^
../development/gocode/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/nsenter/nsexec.c:219:3: note: in expansion of macro ‘bail’
   bail("failed to update /proc/%d/gid_map", pid);
```

cyphar:
None of that is meant to happen. I'll fix those up today. :P

cyphar:
Alright, I've fixed them up. Really the warnings aren't an issue (they're happening in an error path and we can't do anything if the writes fail).

hqhq:
@cyphar Janky failed seems related.

cyphar:
I remember fixing the same issue a week ago. It might be a dodgy rebase. :/

cyphar:
@hqhq Alright, I fixed that issue. It was because I had misplaced brackets in the `bail` macro. Dammit. :unamused:

libcontainer/nsenter/namespace.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2014, 2015 Docker Inc.
+ * Copyright 2015, 2016 The Linux Foundation
```
crosbymichael: I don't think this is correct, the original copyright in the license stays as it is, it doesn't just stop because the repo changed.

crosbymichael: i'm also not a fan of starting to add these on the files, we have a overall license and it applies to all files in the repo, we don't need them per file

cyphar: Fair enough, it's a matter of taste I guess. I'll drop the headers.

crosbymichael: @cyphar thanks

libcontainer/nsenter/nsenter_test.go
```diff
@@ -87,7 +87,7 @@ func TestNsenterInvalidPaths(t *testing.T) {
 
 	namespaces := []string{
 		// join pid ns of the current process
-		fmt.Sprintf("/proc/%d/ns/pid", -1),
+		fmt.Sprintf("pid:/proc/%d/ns/pid", -1),
```
x1022as: should try "pid:/proc/%d/ns/net" here to test the ns path validation?

cyphar: Yes, good idea.

cyphar:
Pinging this since it's needed for #975 (which is blocking -rc3).

cyphar:
Pinging this since it's needed for #975 (which is blocking -rc3).

crosbymichael:
LGTM

mrunalp:
I tested on RHEL 7.2. It is failing :/

```
 oci-runtime-tool generate --tty --output=config.json --uidmappings 1000:0:32000 --gidmappings 1000:0:32000

strace -f -o strace.log runc run 1234
unshare(CLONE_NEWNS|CLONE_NEWUTS|CLONE_NEWIPC|CLONE_NEWPID|CLONE_NEWNET) = -1 EPERM (Operation not permitted)
```

cyphar:
@mrunalp Wasn't that fixed by #975? It was split out of this PR.

mrunalp:
Pretty sure I had run the tests with 975 on RHEL once earlier hence this surprised me. I will debug further.

cyphar:
@mrunalp Any update?

mrunalp:
@cyphar I see the same issue with #975 as well. I have reached out to RHEL kernel team and will get back once I hear their recommendations.

mrunalp:
I played around a bit and I can get past this issue with this diff. I don't think this fixes #959.  It will most likely need changes similar to what I posted in #959. I will test that and get back in a bit.
```
diff --git a/libcontainer/nsenter/nsexec.c b/libcontainer/nsenter/nsexec.c
index d3a50b0..a28c202 100644
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -626,10 +626,9 @@ void nsexec(void)
                         * affect our ability to unshare other namespaces and are used as
                         * context for privilege checks.
                         */
+                       if (unshare(config.cloneflags) < 0)
+                               bail("failed to unshare namespaces");
                        if (config.cloneflags & CLONE_NEWUSER) {
-                               /* Create a new user namespace. */
-                               if (unshare(CLONE_NEWUSER) < 0)
-                                       bail("failed to unshare user namespace");
 
                                /*
                                 * We don't have the privileges to do any mapping here (see the
@@ -647,17 +646,8 @@ void nsexec(void)
                                if (s != SYNC_USERMAP_ACK)
                                        bail("failed to sync with parent: SYNC_USERMAP_ACK: got %u", s);
 
-                               config.cloneflags &= ~CLONE_NEWUSER;
                        }
 
-                       /*
-                        * Now we can unshare the rest of the namespaces. We can't be sure if the
-                        * current kernel supports clone(CLONE_PARENT | CLONE_NEWPID), so we'll
-                        * just do it the long way anyway.
-                        */
-                       if (unshare(config.cloneflags) < 0)
-                               bail("failed to unshare namespaces");
-
                        /* TODO: What about non-namespace clone flags that we're dropping here? */
                        child = clone_parent(&env, JUMP_INIT);
                        if (child < 0)
```

mrunalp:
I confirm that #959 isn't fixed with my diff above. The changes have to be ported exactly from #959. If you want I can create a PR for that.

cyphar:
:/ I really don't like the code in #959, because it breaks up unsharing of `IPC`. Unfortunately I don't really have a system to test that on...

cyphar:
@mrunalp Since the SELinux issue still happens on `master`, is it possible for us to punt on fixing the IPC issue for now (there are other PRs that depend on this one). I'll take some time to take a look at another way of implementing your delay IPC logic that doesn't require special-casing IPC. Is there a nice way of testing an SELinux setup without access to a RHEL machine (a Fedora VM or something?) -- I don't have a SUSE setup with SELinux atm.

mrunalp:
@cyphar I am fine with fixing #959 in a separate PR. I fear there is really no other way besides what I did :). You can use a Fedora 24 VM to reproduce the issue. Just need to compile runc with selinux build tag and use the config in #959.

mrunalp:
I will run other tests on RHEL7 and report back here to check that there are no regressions vs master.

mrunalp:
@cyphar I ran the tests on master vs this PR + my diff and they are the same (There are 4 failures) unrelated to this PR. If you can update the PR to include my suggested changes, we can merge this.

cyphar:
@mrunalp Can this issue be reproduced on a stock Fedora / CentOS VM? I'd like to play around with it a bit to see if I can get another solution to work.

mrunalp:
I meant the diff I posted for combining the unshare of all namespaces without which we don't get test parity with master on RHEL 7. For 959 which we can handle separately you can use a Fedora VM to debug.

cyphar:
@mrunalp AFAICS that patch will break rootless containers (#774), but I'll apply it and I can look at what we can do to fix that issue later.

cyphar:
@mrunalp I've added that patch as a separate commit so that we can track the old way of doing it in the future. PTAL, it should be good to merge now.

mrunalp:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/977/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

mrunalp:
@crosbymichael This needs LGTM again if you have no more comments.

libcontainer/nsenter/nsexec.c
```diff
@@ -413,60 +440,310 @@ void nsexec(void)
...
-		 * start_child() code after forking in the parent.
-		 */
-		uint8_t s = 0;
-		int consolefd = config.consolefd;
+	/* TODO: Currently we aren't dealing with child deaths properly. */
+
+	/*
+	 * Okay, so this is quite annoying.
+	 *
+	 * In order to make sure that deal with older kernels (when CLONE_NEWUSER
```
cyphar: It's mentioned later in the comment.

> This was fixed in upstream commit 1f7f4dde5c945f41a7abc2285be43d918029ecc5, and was introduced by 40a0d32d1eaffe6aac7324ca92604b6b3977eb0e.

hqhq: I thought that's for `CLONE_PATENT` and `CLONE_NEWPID` thing, is that the same thing as `CLONE_NEWUSER` not be handled first?

I was thinking if it's not a too new version, maybe we can claim it as a limitation, making the logic much more complicated to handle such a legacy issue seems not a worthy tradeoff to me.

cyphar: Yeah, you're right. The actual reason for the split is the various issues described underneath the section. This split is the **whole reason** for this PR (so removing the split would make this PR kinda pointless), as there are several legitimate non-legacy issues fixed by this split such as #975.

I can fix the comment, but I don't really want to wait another two weeks for two LGTMs because people are quite busy. I can fix the comment in a later PR if it's not a huge issue for you right now.

hqhq:
Over all it looks good to me, nothing big worth to be a blocker, I found some issues which can all be fixed in following up PRs, it's been quite a while and block lots of other things. Let's move on.

LGTM

cyphar:
@hqhq :heart: I can open a PR with some of your nits later.

libcontainer/nsenter/nsexec.c
```diff
@@ -622,15 +622,24 @@ void nsexec(void)
...
+			 * ordering might break in the future (especially with rootless
+			 * containers). But for now, it's not possible to split this into
+			 * CLONE_NEWUSER + [the rest] because of some RHEL SELinux issues.
+			 *
+			 * We also  can't be sure if the current kernel supports
+			 * clone(CLONE_PARENT | CLONE_NEWPID), so we'll just do it the long
+			 * way anyway.
+			 */
+			if (unshare(config.cloneflags) < 0)
+				bail("failed to unshare namespaces");
```
hqhq: @cyphar I didn't notice you append this commit before, does this mean we actually still can't guarantee correct user namespace ordering for old kernels which don't create user namespace first when it's specified along with other CLONE_NEW* flags?

hqhq: Yeah I see that after revisiting comments, so looks like we don't have any better ideas to fix the issue (user namespace joining ordering on old kernels) for now. Can you open a PR to update comments accordingly?

cyphar: Yup, I've opened #1165 accordingly. PTAL.

PR #978, opened 2016-08-08
Fix the err info of mount failure

hmeng-19:
Signed-off-by: Haiyan Meng <haiyanalady@gmail.com>


hmeng-19:
@mrunalp , PTAL.

cyphar:
Can you include *both* `Destination` and `Source` (`mounting {source} to rootfs {rootfs} at {destination}`)? `Source` is usually some garbage like `cgroup` which doesn't tell you anything about where the user was trying to mount it.

hmeng-19:
@cyphar , Fixed it. PTAL.

mrunalp:
LGTM

LK4D4:
LGTM

PR #979, opened 2016-08-08
Fix the err info of chdir(cwd) failure

hmeng-19:
Currently, if the `cwd` set in the config.json does not exist under `rootfs`, `runc create ...` only fails with the following error info: `no such file or directory`, which is not helpful enough to allow the user to diagnose the problem.

This PR changes the error info to be:
```
chdir to cwd ("/testdir") set in config.json failed: no such file or directory
```
Signed-off-by: Haiyan Meng <haiyanalady@gmail.com>


hmeng-19:
@cyphar , @mrunalp , PTAL.

mrunalp:
LGTM

LK4D4:
janky panic is unrelated
LGTM

PR #980, opened 2016-08-08
libcontainer/configs: make hooks run safer

LK4D4:
It's possible that `cmd.Process` is still nil when we reach timeout.
Start creates `Process` field synchronously, and there is no way to such
race.

Encountered in https://jenkins.dockerproject.org/job/runc-PRs/2004/console


cyphar:
Thanks for this. This should fix the really annoying Janky panics that have been happening for the past few months. :P

dqminh:
LGTM

cyphar:
LGTM

PR #982, opened 2016-08-12
add rpmbuild spec for runc(centos env)

moypray:
Signed-off-by: Wentao Zhang zhangwentao234@huawei.com

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -120,8 +120,8 @@ func finalizeRootfs(config *configs.Config) (err error) {
 	// remount dev as ro if specified
 	for _, m := range config.Mounts {
 		if libcontainerUtils.CleanPath(m.Destination) == "/dev" {
-			if m.Flags&syscall.MS_RDONLY != 0 {
-				if err := remountReadonly(m.Destination); err != nil {
+			if m.Flags&syscall.MS_RDONLY == syscall.MS_RDONLY {
+				if err := remountReadonly(m); err != nil {
 					return newSystemErrorWithCausef(err, "remounting %q as readonly", m.Destination)
 				}
 			}
@@ -700,17 +700,26 @@ func createIfNotExists(path string, isDir bool) error {
 	return nil
 }
 
-// remountReadonly will bind over the top of an existing path and ensure that it is read-only.
-func remountReadonly(path string) error {
+// readonlyPath will make a path read only.
+func readonlyPath(path string) error {
+	if err := syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REC, ""); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+	return syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REMOUNT|syscall.MS_RDONLY|syscall.MS_REC, "")
+}
+
+// remountReadonly will remount an existing mount point and ensure that it is read-only.
+func remountReadonly(m *configs.Mount) error {
+	var (
+		dest  = m.Destination
+		flags = m.Flags
+	)
 	for i := 0; i < 5; i++ {
-		if err := syscall.Mount("", path, "", syscall.MS_REMOUNT|syscall.MS_RDONLY, ""); err != nil && !os.IsNotExist(err) {
+		if err := syscall.Mount("", dest, "", uintptr(flags|syscall.MS_REMOUNT|syscall.MS_RDONLY), ""); err != nil {
 			switch err {
-			case syscall.EINVAL:
-				// Probably not a mountpoint, use bind-mount
-				if err := syscall.Mount(path, path, "", syscall.MS_BIND, ""); err != nil {
-					return err
-				}
-				return syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REMOUNT|syscall.MS_RDONLY|syscall.MS_REC|defaultMountFlags, "")
 			case syscall.EBUSY:
 				time.Sleep(100 * time.Millisecond)
 				continue
@@ -720,7 +729,7 @@ func remountReadonly(path string) error {
 		}
 		return nil
 	}
-	return fmt.Errorf("unable to mount %s as readonly max retries reached", path)
+	return fmt.Errorf("unable to mount %s as readonly max retries reached", dest)
 }
 
 // maskPath masks the top of the specified path inside a container to avoid
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -111,7 +111,7 @@ func (l *linuxStandardInit) Init() error {
 		}
 	}
 	for _, path := range l.config.Config.ReadonlyPaths {
-		if err := remountReadonly(path); err != nil {
+		if err := readonlyPath(path); err != nil {
 			return err
 		}
 	}


script/runc-centos.spec
```diff
@@ -0,0 +1,55 @@
+##  Build docker runc RPM.
+##   1. Copy runc project tarball into ~/rpmbuild/BUILD/docker-runc.tar.gz
```
coolljt0725: `docker-runc.tar.gz` -> `run.tar.gz`?

script/runc-centos.spec
```diff
@@ -0,0 +1,55 @@
...
+
+License: ASL 2.0
+Source: %{name}.tar.gz
+
+URL: https://www.opencontainers.org/
+Vendor: OCI
+Packager: OCI 
+
+BuildRequires: glibc
+#BuildRequires: glibc-static
```
coolljt0725: Commented line should be removed?

script/runc-centos.spec
```diff
@@ -0,0 +1,55 @@
...
+%description
+runc is a CLI tool for spawning and running containers according to the OCF specification
+
+%prep
+%setup -c -n runc
+
+%install
+git checkout -q %{_tag}
+make BUILDTAGS="seccomp selinux"
+#make BUILDTAGS="seccomp selinux" static
```
coolljt0725: same here

script/runc-centos.spec
```diff
@@ -0,0 +1,55 @@
...
+##   2. this spec will build runc in dynamic link not static (if static is need, just uncomment build rules)
+##   3. need to update the TAG of runc project if release a rpm package.
+##
+%global _bindir /usr/local/bin
+%global _tag   v1.0.0-rc1
+
+Name: docker-runc
+Version: 1.0.0
+Release: 1%{?dist}
+Summary: runc is a CLI tool for spawning and running containers according to the OCF specification
```
coolljt0725: `OCF` should be `OCI`?

script/runc-centos.spec
moypray: OK

script/runc-centos.spec
```diff
@@ -0,0 +1,55 @@
...
+##
+##  Note:
+##   1. tarball must be a git project and will checkout the version while building rpms.
+##   2. this spec will build runc in dynamic link not static (if static is need, just uncomment build rules)
+##   3. need to update the TAG of runc project if release a rpm package.
+##
+%global _bindir /usr/local/bin
+%global _tag   v1.0.0-rc1
+
+Name: docker-runc
```
coolljt0725: `docker-runc` -> `runc`?

script/runc-centos.spec
moypray: Commented line is the way to build static, if someone need this, I think he/she just uncomment it will be OK.

cyphar:
Can we also include some of the things I've added to the openSUSE spec file (so that both projects can use the same spec file)? https://build.opensuse.org/package/show/Virtualization:containers/runc

script/runc-centos.spec
```diff
@@ -0,0 +1,55 @@
...
+%prep
+%setup -c -n runc
+
+%install
+git checkout -q %{_tag}
+make BUILDTAGS="seccomp selinux"
+#make BUILDTAGS="seccomp selinux" static
+
+install -d $RPM_BUILD_ROOT/%{_bindir}
+install -p -m 755 runc $RPM_BUILD_ROOT/%{_bindir}/runc
```
hqhq: Man pages and completion files should be added to rpm file.

moypray:
@cyphar  i will try to merge your work on this one.  and will add man, completion files to spec file just like hqhq said.  
What does the patch for golang used for? and the newest Makefile in runc supports to build it directly.

hqhq:
@moypray That's for libseccomp-golang, runC upstream already vendered that change, it's in seccomp's upstream, I think openSUSE is using an old version of runC, so you can just ignore it.

cyphar:
@hqhq Yes, we (openSUSE) are using the last published version of runC (which reminds me that we need to tag an -rc2 at some point soon).

moypray:
Hi I have reworked my PR:
1. build man page to gz files (origin is not gz).
2. use "make rpm" to build rpm pacakges.
     but need we put the rpm in somewhere?  like $RUNC_DIR/release ??

@hqhq

cyphar:
In openSUSE (and SLE), for `rc` releases we generally use `~` notation for the rc part of the version -- otherwise you will end up with weird version comparisons (something like `1.0.0-2.1` is evaluated as older than `1.0.0-rc1.2.1` -- though I can't remember the specifics). So I think the rpm versions should be `runc-1.0.0~rc1` for release candidates.

moypray:
Hi, I have updated the PR , @cyphar  @hqhq , Is that OK?
for release version 1.0.0    rpm will be runc-1.0.0.rpm
for rc version 1.0.0-rc1       rpm will be runc-1.0.0~rc1.rpm

Makefile
```diff
@@ -18,11 +18,14 @@ RUNC_LINK := $(CURDIR)/Godeps/_workspace/src/github.com/opencontainers/runc
 export GOPATH := $(CURDIR)/Godeps/_workspace
 
 MAN_DIR := $(CURDIR)/man/man8
-MAN_PAGES = $(shell ls $(MAN_DIR)/*.8)
+MAN_PAGES = $(shell ls $(MAN_DIR)/*.8.gz)
 MAN_PAGES_BASE = $(notdir $(MAN_PAGES))
 MAN_INSTALL_PATH := ${PREFIX}/share/man/man8/
 
 VERSION := ${shell cat ./VERSION}
+RPM_VERSION := $(shell echo $(VERSION) | sed "s/-[^-]*$$//g")
```
cyphar: Can we use single quotes for these regular expressions -- `$$` makes me think of "get the PID of the shell" which is super-confusing as this is a makefile.

script/runc.spec
```diff
@@ -0,0 +1,60 @@
...
+
+URL: https://www.opencontainers.org/
+Vendor: OCI
+Packager: OCI 
+
+BuildRequires: glibc
+BuildRequires: make
+BuildRequires: libseccomp-devel 
+BuildRequires: libselinux-devel
+
```
cyphar: There's no `Requires: criu` here. We should probably include it (even as a recommends or suggests).

script/runc.spec
```diff
@@ -0,0 +1,60 @@
...
+%description
+runc is a CLI tool for spawning and running containers according to the OCI specification
+
+%prep
+%setup -c -n %{name}-%{version}
+
+%build
+echo $PWD
+make BUILDTAGS="seccomp selinux"
```
cyphar: What about apparmor? ;)

script/runc.spec
```diff
@@ -0,0 +1,60 @@
...
+man/md2man-all.sh
+
+%install
+%{__install} -d $RPM_BUILD_ROOT/%{_bindir}
+%{__install} -p -m 755 runc $RPM_BUILD_ROOT/%{_bindir}/runc
```
cyphar: I believe we use `/usr/local/sbin` for the `make install` so this should be `%{_sbindir}`.

script/runc.spec
```diff
@@ -0,0 +1,60 @@
...
+%description
+runc is a CLI tool for spawning and running containers according to the OCI specification
+
+%prep
+%setup -c -n %{name}-%{version}
+
+%build
+echo $PWD
+make BUILDTAGS="seccomp selinux"
+man/md2man-all.sh
```
cyphar: There's no `BuildRequires` for go-md2man. Is this fine? In openSUSE we have to include it (and then there's a whole thing about the fact that gcc-go and go-md2man don't really work well together so you can only build man pages for certain architectures and then you just wanna scream because it's such a dumb thing to get hung up over).

script/runc.spec
```diff
@@ -0,0 +1,60 @@
...
+##   2. this spec will build runc in dynamic link not static (if static is need, just uncomment build rules)
+##   3. need to update the TAG of runc project if release a rpm package.
+##
+%global _bindir /usr/local/bin
+%global _mandir /usr/share/man
+
+Name: runc
+Version: %{_version}
+Release: %{_release}
+Summary: runc is a CLI tool for spawning and running containers according to the OCI specification
```
cyphar: nit: This summary is too long IMO. It should be something like "A tool for managing OCI containers".

script/runc.spec
```diff
@@ -0,0 +1,60 @@
...
+Packager: OCI 
+
+BuildRequires: glibc
+BuildRequires: make
+BuildRequires: libseccomp-devel 
+BuildRequires: libselinux-devel
+
+
+%description
+runc is a CLI tool for spawning and running containers according to the OCI specification
```
cyphar: And this description is too short. :P

moypray:
hello, update done, thanks for your comments.

script/runc.spec
```diff
@@ -0,0 +1,64 @@
...
+##   2. run rpmbuild -ba script/runc-centos.spec  -> will build source and binary package both.
+##
+##  Note:
+##   1. tarball must be a git project and will checkout the version while building rpms.
+##   2. this spec will build runc in dynamic link not static (if static is need, just uncomment build rules)
+##   3. need to update the TAG of runc project if release a rpm package.
+##
+%global _sbindir /usr/local/sbin
+%global _mandir  /usr/local/share/man
+%global _docdir  /usr/local/share/doc
```
cyphar: Are these not defined on Fedora? In any case, because this is a package this should all be in `/usr/sbin`, `/usr/share/*` -- the `make install` semantics are just for people who are **not** installing things as part of the distribution.

script/runc.spec
```diff
@@ -0,0 +1,64 @@
...
+BuildRequires: gcc-go
+BuildRequires: glibc
+BuildRequires: make
+BuildRequires: libseccomp-devel 
+BuildRequires: libselinux-devel
+BuildRequires: libapparmor-devel
+BuildRequires: go-md2man
+
+%description
+Runc is a CLI tool for spawning and running containers according to the OCI
```
cyphar: `runC`* :wink:

script/runc.spec
```diff
@@ -0,0 +1,64 @@
...
+Vendor: OCI
+Packager: OCI 
+
+BuildRequires: gcc-go
+BuildRequires: glibc
+BuildRequires: make
+BuildRequires: libseccomp-devel 
+BuildRequires: libselinux-devel
+BuildRequires: libapparmor-devel
+BuildRequires: go-md2man
```
cyphar: Does this work on CentOS? In openSUSE we have some horrible [`go-github.com-...` package names](https://build.opensuse.org/package/show/openSUSE:Factory/golang-github-cpuguy83-go-md2man).

script/runc.spec
moypray: but by default these value are `/usr/sbin`, `/usr/share` not `/usr/local/sbin`, `/usr/local/share`, should we change the variable name ?

script/runc.spec
moypray: yes, I have a yum source except libapparmor-devel (find in a web.), and could search all the packages.  how to compatible with suse?

cyphar: I'm not sure, you'd have to do it with a mix of `%{_*_version}` statements. If it works for CentOS I'm fine with merging it and we can make it more cross-usable later.

cyphar:
LGTM

script/runc.spec
cyphar: We shouldn't change the variables -- we **should** be installing to `/usr/sbin` and `/usr/share`, etc. The `/usr/local/*` stuff is only for the Makefile, not for the RPM.

script/runc.spec
moypray: Redhat don't support "Recommends. And not sure if it supports mix of %{_*_version}, could you tell me how to write this statement? I can do some test based on your change.

cyphar: I'm saying that if this works for CentOS then I'm fine with merging it as-is and then we can improve it later. The way we name go packages needs to be handled on a per-distribution bases I feel (at least until Go figures out how to deal with that problem).

moypray: Yes, I think we could add a script to set the names which may be different of the buildrequire packages.

mrunalp:
@lsm5 could you review this?

script/runc.spec
lsm5: Do we really need a CentOS rpm spec file here? CentOS, RHEL and Fedora rpms are already being handled in the distros itself. Or is this supposed to be a generic guideline specfile thing?

In CentOS and RHEL we use the upstream bundled sources itself, so we don't have to deal with golang-* deps.

moypray: Hi, @lsm5 , Removed golang-* deps.
I think it is necessary to add redhat os rpms spec for runc so that we can release it anywhere by rpms.

lsm5: @moypray sorry, please help me understand "release it anywhere by rpms". Do you mean runc upstream will host its own rpms like what docker upstream does?

Also, do you want to perhaps grab the fedora rpm spec itself and then add conditionals for CentOS, SUSE etc if you want a one spec fits all approach? See: http://pkgs.fedoraproject.org/cgit/rpms/runc.git/tree/runc.spec

As far as CentOS Virt SIG goes (for example: https://wiki.centos.org/Cloud/Docker), things will be changing quite a bit, in that we might be moving soon toward using unbundled golang deps for Fedora rebuilds, but we'll continue to use bundled deps for RHEL rebuilds. So, we might wanna add golang-deps in distro-based conditionals, see the Fedora spec file for example.

Also, I don't find libapparmor-devel anywhere on RHEL or Fedora. Does SUSE have it @cyphar ?

script/runc.spec
```diff
@@ -0,0 +1,63 @@
...
+##   2. run rpmbuild -ba script/runc-centos.spec  -> will build source and binary package both.
+##
+##  Note:
+##   1. tarball must be a git project and will checkout the version while building rpms.
+##   2. this spec will build runc in dynamic link not static (if static is need, just uncomment build rules)
+##   3. need to update the TAG of runc project if release a rpm package.
+##
+%global _runc_sbindir /usr/local/sbin
+%global _runc_mandir  /usr/local/share/man
+%global _docdir  /usr/local/share/doc
```
lsm5: Are you using /usr/local to keep files installed separate from rest of the distro?

Fedora rpm convention would be to just use %{_sbindir}, %{_mandir} and so on

moypray: By default, the value of %{_sbindir} and %{_man_dir} is '/usr/sbin'  '/usr/share' on Redhat.  but the Makefile of runC install runc binary into  /usr/local/sbin/ ,  I think we should have same behavior with Makefile. So I defined new variables.

script/runc.spec
moypray: hi @lsm5  ,  I will release runC rpms based on redhat, if rpm spec was integrated into runc project, it would be  much more convenient.  

For  libapparmor-devel pacakge,  I found it in [https://www.rpmfind.net/](url)

script/runc.spec
cyphar: As I said before, since this is an RPM, you should **not** do what the Makefile does. The `make install` script is not for packagers -- it's for people who want to install runC from source.

lsm5: @moypray installing files into /usr/local can end up confusing users. For example:

1. user installs distro-supplied runc
2. Then user installs your (possibly newer) runc (everything goes into /usr/local) and doesn't conflict with the distro's runc. Also /usr/local probably won't be in the user's $PATH so they would still end up using the distro's runc.

script/runc.spec
lsm5: hmm, so looks like SUSE does provide apparmor, please enclose `BuildRequires: libapparmor-devel` into SUSE specific conditionals. I think you'd also need to use conditional BUILDTAGS similarly in %build.

lsm5: @moypray quick note which would sound like a disagreement about this PR (I should've mentioned this in the beginning itself, my apologies), I feel the right place of an rpm specfile is in the distro's rpm git repos, as the rpmbuild process expects the spec file outside of the source tarball so including it within the source feels a little weird to me. That implies CentOS/RHEL/Fedora won't be using this specfile but rather what exists in their dist-git.

You're more than welcome to help us out in co-maintaining Fedora/CentOS runc packages if you're interested.


@cyphar I don't know if above relates to SUSE as well.

But nonetheless, if people still feel the need to include the spec within the source, I'm ok with it.

hqhq: @lsm5 The intention is to support a general rpm file guideline, like what we have in Docker and some other projects. The ideal would be we adopt all fedora, centos, rhel, suse, oraclelinux, etc.. Add different conditions to cover all the differences between distros and different versions. So we can use it on any distros for a rpm building. This PR would be a first move.

script/runc.spec
moypray: Hi, I will update my PR. Thanks for your comments.

script/runc.spec
cyphar: This spec file should be moved to a `contrib` directory, and not be made a Makefile target IMO. Since it's just a guideline for distributions to follow essentially.

> @cyphar I don't know if above relates to SUSE as well.

Yeah, all of our spec files are stored inside OBS.

lsm5: @hqhq @moypray if it's meant to be a guideline file, you might just wanna grab the fedora spec file and adapt it to include other distros. There's a lot of stuff in there already as per fedora's golang packaging draft (including -devel files etc.) which could be useful elsewhere too. See: http://pkgs.fedoraproject.org/cgit/rpms/runc.git/tree/runc.spec

RE: Makefile `rpm` target, I feel it goes in a loop. Meaning, `make rpm` calls `rpmbuild -ba` which then uses the same Makefile but this time inside the untarred sources in the build dir. I guess you could get straight to `rpmbuild -ba` without needing a `make rpm`.

RE: Source0, perhaps you might want to list it as mentioned in  http://pkgs.fedoraproject.org/cgit/rpms/runc.git/tree/runc.spec#n40 , so the user could just `spectool -g runc.spec` to conveniently fetch the tarball.

Anyway, just my 2 cents. If people want to move ahead with this, please don't let me slow you down.

PR #983, opened 2016-08-12
Do not create /dev/fuse by default

justincormack:
This device is not required by the OCI spec.

The rationale for this was linked to https://github.com/docker/docker/issues/2393

So a non functional /dev/fuse was created, and actual fuse use still is
required to add the device explicitly. However even old versions of the JVM
on Ubuntu 12.04 no longer require the fuse package, and this is all not
needed.


justincormack:
See also https://github.com/docker/docker/pull/25660

dqminh:
LGTM

crosbymichael:
LGTM

PR #988, opened 2016-08-16
Support 32 bit UID on i386

chlunde:
The original SETUID takes a 16 bit UID.  Linux 2.4 introduced  a new
syscall, SETUID32, with support for 32 bit UIDs.  The setgid wrapper
already uses SETGID32.

Signed-off-by: Carl Henrik Lunde <chlunde@ifi.uio.no>

### DIFF
--- a/libcontainer/system/syscall_linux_386.go
+++ b/libcontainer/system/syscall_linux_386.go
@@ -8,7 +8,7 @@ import (
 
 // Setuid sets the uid of the calling thread to the specified uid.
 func Setuid(uid int) (err error) {
-	_, _, e1 := syscall.RawSyscall(syscall.SYS_SETUID, uintptr(uid), 0, 0)
+	_, _, e1 := syscall.RawSyscall(syscall.SYS_SETUID32, uintptr(uid), 0, 0)
 	if e1 != 0 {
 		err = e1
 	}


chlunde:
FYI, I don't really have any interest in 32 bit.  A coworker noticed that the i386 build of https://github.com/tianon/gosu did not handle 32 bit uids.

I have only tested gosu.

/cc @tianon

tianon:
I'm not familiar with the details, but should this be changed in `libcontainer/system/syscall_linux_64.go` as well?  Or do 64bit platforms already default to a higher value?

chlunde:
I think they removed the legacy syscalls when designing the x86-64 ABI.  Other platforms might have included them:

https://github.com/torvalds/linux/blob/0803e04011c2e107b9611660301edde94d7010cc/arch/x86/entry/syscalls/syscall_32.tbl

    23	i386	setuid			sys_setuid16
    213	i386	setuid32		sys_setuid

https://github.com/torvalds/linux/blob/0803e04011c2e107b9611660301edde94d7010cc/arch/x86/entry/syscalls/syscall_64.tbl

    105	common	setuid			sys_setuid

tianon:
Neat! :+1:

Seems sane to me!

crosbymichael:
LGTM

mrunalp:
LGTM

PR #992, opened 2016-08-19
move util function

xiekeyang:
`newTestRoot` is an util function which is called in some files. So it seems had better to be located in `utils_test.go`.
Signed-off-by: xiekeyang <xiekeyang@huawei.com>


cyphar:
LGTM

dqminh:
LGTM

PR #993, opened 2016-08-19
Disable the subreaper on exec

crosbymichael:
This keeps the flag but makes it hidden so that existing clients do not
encounter an error if we were to have removed the flag.

Linked to #971

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -79,8 +79,9 @@ following will output a list of processes running in the container:
 			Usage: "add a capability to the bounding set for the process",
 		},
 		cli.BoolFlag{
-			Name:  "no-subreaper",
-			Usage: "disable the use of the subreaper used to reap reparented processes",
+			Name:   "no-subreaper",
+			Usage:  "disable the use of the subreaper used to reap reparented processes",
+			Hidden: true,
 		},
 	},
 	Action: func(context *cli.Context) error {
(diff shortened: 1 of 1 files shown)


exec.go
```diff
@@ -115,7 +116,7 @@ func execProcess(context *cli.Context) (int, error) {
 		return -1, err
 	}
 	r := &runner{
-		enableSubreaper: !context.Bool("no-subreaper"),
+		enableSubreaper: false,
```
dqminh: remove this line as it because `false` is already the default value ?

crosbymichael: ya, just thought i would make it explicit. it was not omitted but set that way for a reason

cyphar:
LGTM. This still doesn't solve the whole problem though. I'm going to clean up #976 (after #975) over the weekend.

dqminh:
LGTM

crosbymichael:
@cyphar ya, that is why I didn't put Fixes ;)

sandyskies:
@crosbymichael  This doesn't fix the exec reparent problem， still child process of the exec process，PPID becomes 0 ， after  exec process exits.

cyphar:
@sandyskies Yup, we know. Which is why #976 exists (as we discussed above).

PR #995, opened 2016-08-22
Restored-from-checkpoint containers should have a start time

estesp:
Set the start time similar to a brand new container.

Docker-DCO-1.1-Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com> (github: estesp)

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1049,6 +1049,8 @@ func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Proc
 		}); err != nil {
 			return err
 		}
+		// create a timestamp indicating when the restored checkpoint was started
+		c.created = time.Now().UTC()
 		if _, err := c.updateState(r); err != nil {
 			return err
 		}


estesp:
@boucher make sure this makes sense--just happened to notice while playing with containerd+runc checkpoint/restore that restoring from checkpoint doesn't fill in a created time, leaving it at 0 epoch seconds as the default.

cyphar:
I feel like there's a valid semantic debate to be had here. Sure, if you checkpoint a container and then restore it the process "never died" but at the same time it was "started" in the strict sense when you restored it. Maybe there's something in the runtime-spec that can clarify this (I can't remember if there's anything in it relating to start times).

/cc @avagin

mrunalp:
We punted on checkpoint/restore for 1.0 in the runtime-spec. The PR seems sane to me but would defer to @avagin on it.

boucher:
I would also defer to @avagin and @crosbymichael here. Personally I think this makes sense.

crosbymichael:
LGTM

avagin:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/995/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

PR #998, opened 2016-08-24
Use vndr tool for vendoring

LK4D4:
To install vndr:

```
go get github.com/LK4D4/vndr
```

To update deps: change `vndr.cfg` and

```
vndr
```

It's possible to update only one dep:

```
vndr <import_path> <revision>
```

However, it does not update config. Config always must be modified by hands and sole source of truth for dependencies.

I've tried to reconstruct a behavior of `hack/vendor.sh` from docker without all its problem. This tool has a very small set of features and does not mess with your GOPATH.


vndr.cfg
```diff
@@ -0,0 +1,11 @@
+github.com/Sirupsen/logrus 26709e2714106fb8ad40b773b711ebce25b78914
+github.com/coreos/go-systemd v4
+github.com/docker/docker 0f5c9d301b9b1cca66b3ea0f9dec3b5317d3686d
+github.com/docker/go-units v0.1.0
+github.com/godbus/dbus v3
+github.com/golang/protobuf f7137ae6b19afbfd61a94b746fda3b3fe0491874
+github.com/opencontainers/runtime-spec v1.0.0-rc1
+github.com/seccomp/libseccomp-golang 32f571b70023028bd57d9288c20efbcb237f3ce0
+github.com/syndtr/gocapability 2c00daeb6c3b45114c80ac44119e7b8801fdd852
+github.com/urfave/cli edb24d02aa3cea2319c33f2836d4a5133907fe4c
```
tianon: Think we could upgrade this to `v1.18.0`? :pray: :innocent: (https://github.com/urfave/cli/commit/edb24d02aa3cea2319c33f2836d4a5133907fe4c)

vndr.cfg
```diff
@@ -0,0 +1,11 @@
+github.com/Sirupsen/logrus 26709e2714106fb8ad40b773b711ebce25b78914
+github.com/coreos/go-systemd v4
+github.com/docker/docker 0f5c9d301b9b1cca66b3ea0f9dec3b5317d3686d
```
tianon: Do you think it'd make sense to move this to a tag too? (the commit referenced here has been in since `v1.8.0-rc1`; https://github.com/docker/docker/commit/0f5c9d301b9b1cca66b3ea0f9dec3b5317d3686d)

vndr.cfg
```diff
@@ -0,0 +1,11 @@
+github.com/Sirupsen/logrus 26709e2714106fb8ad40b773b711ebce25b78914
```
tianon: This could be `v0.8.0` or probably even `v0.10.0`. :smile: (https://github.com/Sirupsen/logrus/commit/26709e2714106fb8ad40b773b711ebce25b78914)

tianon:
Looks simple and easy to use! :heart_eyes:

vndr.cfg
dqminh: think we probably want to do this later just to minimize the change ?

dqminh:
Yet another vendor tool yay 😝 

Looks ok so far though ( i skimmed though the file changes ). One plus for me is the removal of some unused nested vendor folders (like docker/docker/vendor)

spec_validator tests need fixing with this change too.

vndr.cfg
tianon: Yeah, I suppose that's fair. :innocent:

jfrazelle:
ah crap im 5 hours too slow, ill close #1000

LK4D4:
@dqminh Yup, sorry about that.
I encourage you at least to try this tool :) I'd like to use something like this - config-driven.

dqminh:
LGTM

I think this is great. I havent actually tried `vndr` yet (sorry @LK4D4 😛), but just moving to standard vendor location already let us experiment with vendor tools easier since the actual vendor source should not change.

Makefile
```diff
@@ -4,18 +4,16 @@
...
 PROJECT := github.com/opencontainers/runc
 TEST_DOCKERFILE := script/test_Dockerfile
 BUILDTAGS := seccomp
 RUNC_BUILD_PATH := /go/src/github.com/opencontainers/runc/runc
 RUNC_INSTANCE := runc_dev
+PKG_NAME := github.com/opencontainers/runc
 COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
-RUNC_LINK := $(CURDIR)/Godeps/_workspace/src/github.com/opencontainers/runc
-export GOPATH := $(CURDIR)/Godeps/_workspace
+TEMP_GOPATH := /tmp/runc-gopath
```
dqminh: nit: i dont like this TEMP_GOPATH, but i guess it has values when user doesnt check out to the standard gopath.

LK4D4: Don't like that either. But we had requirement to be able to install runc after just git clone.

imikushin:
Whoa! Can I suggest using [trash](https://github.com/rancher/trash) instead? ;)

LK4D4:
@imikushin btw I tried trash for docker and it doesn't support hg :)

imikushin:
@LK4D4 report an issue :)

Makefile
cyphar: [Note: I don't like how Go doesn't let you check out repos where you want to, which is why the `GOPATH` hacks are quite useful.]

cyphar:
@LK4D4 Do you want to update this yourself or should I carry this PR?

cyphar:
Closing in favour of #1340. Thanks for starting this off @LK4D4!

PR #1001, opened 2016-08-25
add pre-dump and parent-path to checkpoint

x1022as:
CRIU gets pre-dump subcmd to complete iterative migration.
pre-dump saves process memory info only. And it need parent-path
to specify the former memory files.

This patch add pre-dump and parent-path arguments to runc checkpoint

Signed-off-by: Deng Guangxing dengguangxing@huawei.com

ps. pre-dump is a subcmd in criu, and it is so much like criu dump(checkpoint in runc);
so just add pre-dump as an argument to runc checkpoint here :)

with this patch , one can get iterative migration through these commands:

```
//perform pre-dump and specify './parent' as it image-path
# ./runc checkpoint --pre-dump --image-path ./parent test
//criu dump. './image' as image-path and relative path '../parent' as the prev-image-dir
# ./runc checkpoint --image-path ./image/ --parent-path ../parent test
```


checkpoint.go
```diff
@@ -40,13 +42,14 @@ checkpointed.`,
 		}
 		defer destroy(container)
 		options := criuOptions(context)
+		preDump := context.Bool("pre-dump")
```
cyphar: Why don't we add that as part of `criuOptions`? I don't like that the interface for checkpointing now takes two arguments which are both technically "criu options".

tests/integration/checkpoint.bats
```diff
@@ -33,8 +33,18 @@ function teardown() {
...
+  [ "$status" -eq 0 ]
+
+  # busybox should still be running
+  runc state test_busybox
+  [ "$status" -eq 0 ]
+  [[ "${output}" == *"running"* ]]
+
   # checkpoint the running container
-  runc --criu "$CRIU" checkpoint test_busybox
+  runc --criu "$CRIU" checkpoint --parent-path ./parent-dir test_busybox
```
cyphar: Please make a new test for `--pre-dump`. We shouldn't be replacing old tests with new tests that test different features.

checkpoint.go
x1022as: technically, pre-dump is not a criu-option, it is a subcmd like dump(checkpoint), so I take this as an outer arg.
But, I can change this if people want checkpoint clear : )

tests/integration/checkpoint.bats
x1022as: ok, I will add a new one

adrianreber:
Having recently implement pre-dump for LXC I also tried this pre-dump PR for runc.

I needed to extent the existing patch with:

```
diff --git a/libcontainer/container_linux.go b/libcontainer/container_linux.go
index b51d435..3e12f04 100644
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -627,6 +627,7 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts, preDump bool) error {
        //pre-dump may need parentImage param to complete iterative migration
        if criuOpts.ParentImage != "" {
                rpcOpts.ParentImg = proto.String(criuOpts.ParentImage)
+               rpcOpts.TrackMem = proto.Bool(true)
        }
 
        // append optional manage cgroups mode
```

The same was necessary in LXC. In older versions it was possible to use memory tracking by just using ```--prev-images-dir```... nowadays it is also necessary to provide ```--track-mem```.

With above patch this PR works as expected.

For reference the LXC commit: https://github.com/lxc/lxc/commit/9f99a33fa97a81d395b6522fe7bd51e0ef6af454

x1022as:
@adrianreber tks for trying this pr, I thought pre-dump subcmd include --track-mem option implicitly, and final dump do not need it either (in [Incremental_dumps](https://criu.org/Incremental_dumps), it says `No need in memory tracking option`). ping @avagin to make sure :smiley:

avagin:
@x1022as you need to set --track-mem each time when we want to track memory changes. The pre-dump command doesn't enable it.

libcontainer/container_linux.go
```diff
@@ -960,12 +974,15 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
...
 	// Here we want to wait only the CRIU process.
-	st, err := cmd.Process.Wait()
-	if err != nil {
-		return err
-	}
-	if !st.Success() {
-		return fmt.Errorf("criu failed: %s\nlog file: %s", st.String(), logPath)
+	reqType := req.GetType()
+	if reqType != criurpc.CriuReqType_PRE_DUMP {
+		st, err := cmd.Process.Wait()
```
avagin: Can you explain why we don't need to wait a pre-dump process?

checkpoint.go
avagin: I would agree with @cyphar here

marcosnils:
@x1022as @avagin @cyphar we've done this in the past for a project that required multiple step migration with runc. Even though the commit is a bit outdated it might help for the proper implementation.

https://github.com/marcosnils/runc/commit/01ed07533dff1c8974513a6ea36992760710c895

libcontainer/container_linux.go
x1022as: checkpoint --pre-dump would block here waiting for something, I don't know the exact reason.
maybe need a  deep dig. for now do not call wait() here when pre-dump works fine for me.

cyphar: `cat /proc/[pid]/stack` might help you out. Note that doing an `strace` of a multithreaded process is not recommended (when you do `execve` there are some Linux kernel bugs which mean that you sometimes lock up the process because the other threads in the process become zombies and the `execve` is blocked on them dying).

adrianreber:
I rebased and adapted this PR in #1318

x1022as:
@adrianreber seems you have figured out the reason why pre-dump block.

adrianreber:
@x1022as Do you plan to rebase this PR so that it applies again? As far as I see there are four points not solved: It does not apply, you need to add track-mem, pre-dump should be part of ```criuOptions``` and the missing wait needs to be documented.

I would like to see this merged as I have additional patches on top of this to also support post-copy migration.

x1022as:
code updated.  fix criu blocking with criuClient.Close() and complete pre-dump bats test according to @adrianreber 's PR ,  add Adrian to SOB.

ping @avagin and @cyphar to review. let's move on

x1022as:
CI with go 1.7.X passed while go 1.6.X failed with this error:

```
FAIL: TestNotifyMemoryPressure (0.00s)
	notify_linux_test.go:98: expected event control to be closed
```

any idea about this problem?

avagin:
LGTM

adrianreber:
@x1022as thanks for the changes. Can an additional reviewer please have a look at this PR?

x1022as:
ping @cyphar  please take a look :smiley:

hqhq:
LGTM

hqhq:
CI failure seems unrelevant and we are dropping go 1.6 support, so I think we can move on for now.

ashwani29:
@adrianreber  @avagin If I do several predumps for a container  without a last dump/ checkpoint and then try to restore it on a destination. Why it doesn't work?
is there any example/demo for iterative migration. In my case a parent file is created in each pre-dump  and it says parent link is broken. For every new pre-dump, i've created a new directory under the same parent as of previous pre-dump.  I want to know how the parent-path should be for this case?

adrianreber:
> @adrianreber @avagin If I do several predumps for a container without a last dump/ checkpoint and then try to restore it on a destination. Why it doesn't work?

Without a final dump it does not contain all the necessary information for restore.

> is there any example/demo for iterative migration. In my case a parent file is created in each pre-dump and it says parent link is broken. For every new pre-dump, i've created a new directory under the same parent as of previous pre-dump. I want to know how the parent-path should be for this case?

Have a look at https://people.redhat.com/areber/criu/migrate which I have been using to migrate runc containers using pre-dumps.

PR #1002, opened 2016-08-25
Return 0 for pid if container is stopped

crosbymichael:
When a container is in the stopped state it no longer has a pid so the list and state commands should return 0


mrunalp:
LGTM

dqminh:
LGTM

PR #1003, opened 2016-08-26
Error handling when container not exists

rajasec:
When the container not exists, user gets the error saying
open /run/runc/container1/state.json: no such file or directory
 From end user perspective, if we do any operation on container we can show whether the container ID is valid one, instead of giving the state.json error.

Hence the change in this PR , upon executing runc on invalid ID, it will show as below.
"container not exist"

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -188,7 +188,7 @@ func (l *LinuxFactory) Load(id string) (Container, error) {
 		return nil, newGenericError(fmt.Errorf("invalid root"), ConfigInvalid)
 	}
 	containerRoot := filepath.Join(l.Root, id)
-	state, err := l.loadState(containerRoot)
+	state, err := l.loadState(containerRoot, id)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


libcontainer/factory_linux.go
```diff
@@ -277,7 +277,7 @@ func (l *LinuxFactory) loadState(root string) (*State, error) {
 	f, err := os.Open(filepath.Join(root, stateFilename))
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, newGenericError(err, ContainerNotExists)
+			return nil, newGenericError(fmt.Errorf("container not exists"), ContainerNotExists)
```
crosbymichael: Can you add the container's id to the error message? 

`"container %q does not exist"`

rajasec:
Added the container ID as per your comment, since root was passed ( it contains the full path: /run/runc/<containerID) so instead of printing the full path or trimming down this path, I passed the id as argument via loadState to get the ID.

libcontainer/factory_linux.go
```diff
@@ -273,11 +273,11 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 	return i.Init()
 }
 
-func (l *LinuxFactory) loadState(root string) (*State, error) {
+func (l *LinuxFactory) loadState(root, id string) (*State, error) {
 	f, err := os.Open(filepath.Join(root, stateFilename))
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, newGenericError(err, ContainerNotExists)
+			return nil, newGenericError(fmt.Errorf("container id %q not exists", id), ContainerNotExists)
```
crosbymichael: Can you remove the "id" part.  It should just read `container %q does not exist`  To people they associate the id as the container.

rajasec:
Updated as per your comment.

crosbymichael:
LGTM

hqhq:
LGTM

PR #1004, opened 2016-08-27
add more parameters of blkio subsystem to command line.

hopkings2008:
Signed-off-by: Yu Zou zouyu7@huawei.com

### DIFF
--- a/man/runc-update.8.md
+++ b/man/runc-update.8.md
@@ -24,7 +24,13 @@ accepted format is as follow (unchanged values can be omitted):
        "mems": ""
      },
      "blockIO": {
-       "blkioWeight": 0
+       "blkioWeight": 0,
+       "blkioLeafWeight": 0,
+       "blkioWeightDevice": "",
+       "blkioThrottleReadBpsDevice": "",
+       "blkioThrottleWriteBpsDevice": "",
+       "blkioThrottleReadIOPSDevice": "",
+       "blkioThrottleWriteIOPSDevice": ""
      },
    }
 
@@ -32,15 +38,21 @@ Note: if data is to be read from a file or the standard input, all
 other options are ignored.
 
 # OPTIONS
-   --resources value, -r value  path to the file containing the resources to update or '-' to read from the standard input
-   --blkio-weight value         Specifies per cgroup weight, range is from 10 to 1000 (default: 0)
-   --cpu-period value           CPU period to be used for hardcapping (in usecs). 0 to use system default
-   --cpu-quota value            CPU hardcap limit (in usecs). Allowed cpu time in a given period
-   --cpu-share value            CPU shares (relative weight vs. other containers)
-   --cpuset-cpus value          CPU(s) to use
-   --cpuset-mems value          Memory node(s) to use
-   --kernel-memory value        Kernel memory limit (in bytes)
-   --kernel-memory-tcp value    Kernel memory limit (in bytes) for tcp buffer
-   --memory value               Memory limit (in bytes)
-   --memory-reservation value   Memory reservation or soft_limit (in bytes)
-   --memory-swap value          Total memory usage (memory + swap); set '-1' to enable unlimited swap
+   --resources value, -r value                 path to the file containing the resources to update or '-' to read from the standard input
+   --blkio-weight value                        Specifies per cgroup weight
+   --blkio-leaf-weight value                   Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, cfq scheduler only 
+   --blkio-weight-device value                 Weight per cgroup per device, can override blkio-weight. Argument must be of the form "<MAJOR>:<MINOR> <WEIGHT> [LEAF_WEIGHT]"
+   --blkio-throttle-readbps-device value       IO read rate limit per cgroup per device, bytes per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --blkio-throttle-writebps-device value      IO write rate limit per cgroup per divice, bytes per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --blkio-throttle-readiops-device value      IO read rate limit per cgroup per device, IO per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --blkio-throttle-writeiops-device value     IO write rate limit per cgroup per device, IO per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --cpu-period value                          CPU period to be used for hardcapping (in usecs). 0 to use system default
+   --cpu-quota value                           CPU hardcap limit (in usecs). Allowed cpu time in a given period
+   --cpu-share value                           CPU shares (relative weight vs. other containers)
+   --cpuset-cpus value                         CPU(s) to use
+   --cpuset-mems value                         Memory node(s) to use
+   --kernel-memory value                       Kernel memory limit (in bytes)
+   --kernel-memory-tcp value                   Kernel memory limit (in bytes) for tcp buffer
+   --memory value                              Memory limit (in bytes)
+   --memory-reservation value                  Memory reservation or soft_limit (in bytes)
+   --memory-swap value                         Total memory usage (memory + swap); set '-1' to enable unlimited swap
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -30,7 +30,8 @@ function setup() {
         "cpus": "0"
     },
     "blockio": {
-        "blkioWeight": 1000
+        "blkioWeight": 1000,
+        "blkioLeafWeight": 1000
     },
 EOF
     )
@@ -62,6 +63,7 @@ function check_cgroup_value() {
 
     # check that initial values were properly set
     check_cgroup_value $CGROUP_BLKIO "blkio.weight" 1000
+    check_cgroup_value $CGROUP_BLKIO "blkio.leaf_weight" 1000
     check_cgroup_value $CGROUP_CPU "cpu.cfs_period_us" 1000000
     check_cgroup_value $CGROUP_CPU "cpu.cfs_quota_us" 500000
     check_cgroup_value $CGROUP_CPU "cpu.shares" 100
@@ -75,6 +77,10 @@ function check_cgroup_value() {
     runc update test_update --blkio-weight 500
     [ "$status" -eq 0 ]
     check_cgroup_value $CGROUP_BLKIO "blkio.weight" 500
+
+    # update blkio-leaf-weight
+    runc update test_update --blkio-leaf-weight 500
+    check_cgroup_value $CGROUP_BLKIO "blkio.leaf_weight" 500
 
     # update cpu-period
     runc update test_update --cpu-period 900000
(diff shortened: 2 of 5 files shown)


hopkings2008:
Dear maintainers,
Please help to review this PR, and this PR is to add more blkio subsystem parameters according to oci spec.
And seems that the test cannot be started due to "java.io.IOException: remote file operation failed: /home/ubuntu/workspace/runc-PRs at hudson.remoting.Channel@3bdba927:ubuntu (aufs) (i-7ca51ce9): java.io.IOException: Remote call on ubuntu (aufs) (i-7ca51ce9) failed"

hopkings2008:
Dear maintainers,
I triggered the jenkins test again, this time, all the tests are passed. So, please help to review this PR, thanks a lot.

hqhq:
Code looks good, can you add some test cases in `tests/integration/cgroups.bats` and `tests/integration/update.bats`?

update.go
```diff
@@ -57,6 +67,30 @@ other options are ignored.
 			Name:  "blkio-weight",
 			Usage: "Specifies per cgroup weight, range is from 10 to 1000",
 		},
+		cli.IntFlag{
+			Name:  "blkio-leaf-weight",
+			Usage: "Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only",
+		},
+		cli.StringSliceFlag{
+			Name:  "blkio-weight-device",
+			Usage: "Weight per cgroup per device, can override BlkioWeight.",
```
mlaventure: Can you add the expected input format in there (and the one below)?

hqhq:
@hopkings2008 Then I think you can test it on your local machine make sure they all passed, and add check for io scheduler and skip the test case if it's not cfq.

mlaventure:
@hopkings2008 I think in the flag description something more generic should be used, for instance:

`<MAJOR>:<MINOR> <WEIGHT> [LEAF_WEIGHT]` instead of `8:0 800 [200]`

wdyt?

hopkings2008:
@mlaventure thanks for your good suggestion. I have amended the description of them, please check it again.
@hqhq i have added unit tests for this PR, and due to the lack of cfq device in CI machine, the tests cannot be added into integration currently. I will added them later.

update.go
```diff
@@ -205,3 +267,85 @@ other options are ignored.
...
+	var ctds []*configs.ThrottleDevice
+	for _, d := range devices {
+		td := &configs.ThrottleDevice{}
+		td.Major = d.Major
+		td.Minor = d.Minor
+		td.Rate = *d.Rate
+		ctds = append(ctds, td)
+	}
+	return ctds
+}
```
hqhq: You can use `configs.NewWeightDevice` and `configs.NewThrottleDevice` instead of these.

hopkings2008: @hqhq Good suggestion, I will amend this. Thanks.

hopkings2008:
@hqhq i have amended the code according to your suggestion, please check it again. thanks.

update.go
```diff
@@ -205,3 +267,78 @@ other options are ignored.
...
+		dev.Rate = u64Ptr(rate)
+		devs = append(devs, dev)
+	}
+	return devs, nil
+}
+
+func convertSpecWeightDevices(devices []specs.WeightDevice) []*configs.WeightDevice {
+	var cwds []*configs.WeightDevice
+	for _, d := range devices {
+		wd := configs.NewWeightDevice(d.Major, d.Minor, *d.Weight, *d.LeafWeight)
```
hqhq: According to `getBlkioWeightDeviceFromString()`, `d.LeafWeight` could be null, you have to check null condition here.

hopkings2008: @hqhq thanks for finding this problem. I have fixed it by checking the nil pointer value. By default, the leafweight will be 0 which does nothing. Please help to check this PR again.

hqhq:
Sorry I should've mentioned before, `man/runc-update.8.md` and `contrib/completions/bash/runc` should also be altered.

hopkings2008:
@hqhq i missed those two related things, and thanks for your reminder. And i have added the missed changes. Please check it again. :)

hqhq:
LGTM, thanks.

hopkings2008:
@mlaventure could you please help to check this PR again? and welcome your suggestions, thanks a lot.

hopkings2008:
Dear maintainers, Please help to check this PR and after this can be merged, i will submit the PR for related change of containerd.

man/runc-update.8.md
```diff
@@ -24,23 +24,35 @@ accepted format is as follow (unchanged values can be omitted):
...
-   --kernel-memory value        Kernel memory limit (in bytes)
-   --kernel-memory-tcp value    Kernel memory limit (in bytes) for tcp buffer
-   --memory value               Memory limit (in bytes)
-   --memory-reservation value   Memory reservation or soft_limit (in bytes)
-   --memory-swap value          Total memory usage (memory + swap); set '-1' to enable unlimited swap
+   --resources value, -r value                 path to the file containing the resources to update or '-' to read from the standard input
+   --blkio-weight value                        Specifies per cgroup weight, range is from 10 to 1000 (default: 0)
+   --blkio-leaf-weight value                   Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only 
+   --blkio-weight-device value                 Weight per cgroup per device, can override blkio-weight. For example, "<MAJOR>:<MINOR> <WEIGHT> [LEAF_WEIGHT]"
+   --blkio-throttle-readbps-device value       IO read rate limit per cgroup per device, bytes per second. For example, "<MAJOR>:<MINOR> <RATE>"
```
mlaventure: nitpick: `s/For example/Argument must be of the form/` or something similar

hopkings2008: @mlaventure thanks very much for your good suggestion. I have amended that description and please check it again. thanks

hopkings2008:
@mlaventure could you please help to review this PR again? i have amended the help description message according to your suggestion. Thanks a lot.

hopkings2008:
Dear maintainers, Could you please check this PR  so that i can send my PR for related changes for containerd, and welcome your suggestions. thanks in advance.

hopkings2008:
@crosbymichael  , could you please help to review this PR? and weclome your suggestions. :)

hqhq:
LGTM

hopkings2008:
@mlaventure , could you please take some time to check this pr again? it has taken some long time. :)

man/runc-update.8.md
```diff
@@ -24,23 +24,35 @@ accepted format is as follow (unchanged values can be omitted):
...
+   --blkio-weight value                        Specifies per cgroup weight, range is from 10 to 1000 (default: 0)
+   --blkio-leaf-weight value                   Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only 
+   --blkio-weight-device value                 Weight per cgroup per device, can override blkio-weight. Argument must be of the form "<MAJOR>:<MINOR> <WEIGHT> [LEAF_WEIGHT]"
+   --blkio-throttle-readbps-device value       IO read rate limit per cgroup per device, bytes per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --blkio-throttle-writebps-device value      IO write rate limit per cgroup per divice, bytes per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --blkio-throttle-readiops-device value      IO read rate limit per cgroup per device, IO per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
+   --blkio-throttle-writeiops-device value     IO write rate limit per cgroup per device, IO per second. Argument must be of the form "<MAJOR>:<MINOR> <RATE>"
```
cyphar: You've added all of these flags ...

tests/integration/update.bats
```diff
@@ -76,6 +78,10 @@ function check_cgroup_value() {
     [ "$status" -eq 0 ]
     check_cgroup_value $CGROUP_BLKIO "blkio.weight" 500
 
+    # update blkio-leaf-weight
+    runc update test_update --blkio-leaf-weight 500
+    check_cgroup_value $CGROUP_BLKIO "blkio.leaf_weight" 500
```
cyphar: ... but only added a test for leaf weight. Is this intentional (meaning, is it possible for us to test the other options as well here)?

update.go
```diff
@@ -57,6 +68,30 @@ other options are ignored.
 			Name:  "blkio-weight",
 			Usage: "Specifies per cgroup weight, range is from 10 to 1000",
 		},
+		cli.IntFlag{
+			Name:  "blkio-leaf-weight",
+			Usage: "Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only",
```
cyphar: IMO we should be specifying "ranges" as part of the help docs. Shouldn't these be all listed inside the runtime-spec?

update.go
```diff
@@ -6,16 +6,21 @@ import (
...
+	"github.com/opencontainers/runc/libcontainer/configs"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/urfave/cli"
 )
 
 func u64Ptr(i uint64) *uint64 { return &i }
 func u16Ptr(i uint16) *uint16 { return &i }
 
+var regBlkioWeightDevice = regexp.MustCompile(`([0-9]+):([0-9]+) ([0-9]+)(?: ([0-9]+))?`)
+var regBlkioThrottleDevice = regexp.MustCompile(`([0-9]+):([0-9]+) ([0-9]+)`)
```
cyphar: Is this format defined in the spec (my guess is yes), and if so please at least comment that it comes from the spec or something so it's a bit more clear. In addition, if it comes from the spec maybe all of these conversion functions should go inside `libcontainer/specconv` (@hqhq what do you think?).

tests/integration/update.bats
hopkings2008: @cyphar  thanks for checking this PR. For other parameters of blkio, we cannot test them in current CI machine, because it is lack of cfq device in CI machine. What is your suggestion about the device limitation of CI machine?

cyphar: I would add the tests as "optional" if possible (at SUSE we run the test suite on SUSE machines, so the tests are useful outside of the CI setup for this repo). But if we're not testing it in the CI then it's not as important (and could wait for a follow-up PR).

update.go
hopkings2008: @cyphar , thanks for your checking. I have amended this and uses 'ranges from' instead of "range is from". Please check it again. Thanks.

tests/integration/update.bats
hopkings2008: @cyphar , currently, the CI tests are wrote with shell framework. If i add the new flags, some tests are not passed due to the "device" problem. I have tried to add the new flags to CI tests by using 'conditional' judgement, but if do so, the code seems very ugly. For example, if i add new flags into 'DATA; section, i will double the 'old flags' twice. I think this is not acceptable. Can I add those new flags into tests later in another PR if the test framework become more flexible? 
Welcome your suggestions.

update.go
cyphar: That's not what I meant. What I meant was that the *actual range* should not be in the help docs. The ranges are defined by the `runtime-spec`, not by us.

tests/integration/update.bats
cyphar: Yeah, we can handle this later I guess. We need to do the test framework improvements for the memory cgroup anyway.

AkihiroSuda:
Needs rebase

kolyshkin:
Per-device IOPS and weights are supported now, see PR #4775. We did not want to introduce more CLI flags, this can be handled via JSON.

PR #1006, opened 2016-08-29
fix make dbuild error

datawolf:
when type "make dbuild" command, it report the following error:

Pulling repository docker.io/library/runc_test
Error: image library/runc_test:latest not found
Makefile:37: recipe for target 'dbuild' failed
make: *** [dbuild] Error 1

just add "docker tag" command on runctestimage to fix it.


hqhq:
@datawolf Your PR is a work around, I think this should be fixed in another way, I've pulled #1008 to handle this.

PR #1007, opened 2016-08-29
cli: migrate cli framework to "cobra" and "pflag"

### LINKED ISSUE #752: runc exec command parsing

forever043:
This PR migrate cli framework to [cobra](https://github.com/spf13/cobra) and [pflag](https://github.com/spf13/pflag), which can fix #752 and #788 more appropriately.

All codes related to the old framework are removed.


hqhq:
@opencontainers/runc-maintainers Is this cobra cli idea sounds good to you? If yes, do you think we should do this step by step as this PR or do all substitution in one PR?

cyphar:
I am +1 on using a different CLI library as long as it solves our problems with `--` handling (not to mention fixing `runc ps` so that it's actually useful). However, I think I would prefer doing the substitution in a single PR otherwise you end up with `master` having different CLI semantics for different subcommands (which is icky). But then again, the cost of doing a big PR is that nobody can track what's going on.

I do like the idea of just piggybacking on top of the existing framework and then removing the framework boilerplate once all of the conversions are done.

crosbymichael:
+1 to what @cyphar said.  

@forever043 if you are interested in doing this and the rest of the maintainers think it is a good idea, we would want to make this change in one PR so that we are not using two cli packages at the same time.  

I think if you can split the commands into separate commits and have the ground work done in the fist commit it would be much easier for us to review when we make that change.  

Let me know what you think.

forever043:
@crosbymichael got it! And I will do this :)

hqhq:
This is a significant change we should include in 1.0, I'll add the milestone.

cyphar:
We do want to release 1.0 at some point, right @hqhq? :wink:

hqhq:
@cyphar Sure we do :wink: , let's focus on that and speed it up, I see the major block is your rootless container PRs, I'll review them once I can.

cyphar:
I think we can delay rootless for 1.1.0, but we need to get console handling and nsenter fixed (from my view). Once those are fixed then rootless becomes a much smaller patchset.

forever043:
hi all, @hqhq @cyphar @crosbymichael,
All commands are converted to cobra.
And I have to modify the 'runc foo -h' tests (		a2db5cf) because the error message is hard-coded in cobra.
Hope everything is ok :)

hqhq:
@forever043 For a patchset, you must make sure it won't break bisect, which means it should be compileable and pass all test cases.

cyphar:
Basically what @hqhq said. If your PR works in its current state, please rebase it to make sure that every commit builds and passes the tests (in other words, is bisectable).

forever043:
@hqhq @cyphar, rebase finished, and every commit can build and pass the tests :)

crosbymichael:
With the updated `cli` package fixing the `--` issues and arg parsing, is there a reason to write the entire cli on pflags and cobra?

hqhq:
I don't know, but IIRC it's mentioned that there are bunch of advantages for cobra when docker choose it to replace the old cli package. @forever043 can you collect the advantages of cobra so we can decide if it's worth doing the replacement? Anyway it can be decided post-1.0.

crosbymichael:
@hqhq well comparing docker's old cli package to cobra is an unfair comparison.  I think that gap is much closer when you compare cli vs cobra after this last update.

crosbymichael:
Sorry that you did all this work but I think making the change does not benefit us that much and will just add code churn when its not needed.  The cli package was updated to handle `--` correctly and meets all our requirement.s

PR #1010, opened 2016-08-30
Add command itself to the result of runc ps

keloyang:
currently ps itself is not in the result of runc ps 
```
root@ubuntu:~/# ./runc ps  test
root      6129  6119  0 20:11 pts/12   00:00:00 sh
```
this pr will add it(the result is similar to runc exec test ps)
```
root@ubuntu:~/# ./runc ps  test
root      6129  6119  0 20:11 pts/12   00:00:00 sh
root      6206  6201  0 20:17 pts/6    00:00:00 ps -ef
```

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -56,10 +56,12 @@ var psCommand = cli.Command{
 			psArgs = []string{"-ef"}
 		}
 
-		output, err := exec.Command("ps", psArgs...).Output()
+		cmd := exec.Command("ps", psArgs...)
+		output, err := cmd.Output()
 		if err != nil {
 			return err
 		}
+		pids = append(pids, cmd.Process.Pid)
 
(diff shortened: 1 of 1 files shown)


duglin:
Is the 'ps' run in the same PID namespace as the 'sh' ?  Not 100% sure but I don't think it is, and as an end-user I would expect to only see processes that are running within the container's PID namespace.

cyphar:
With `runc ps`, the `ps` process is not run inside the container. Rather, a `ps` process running in the same context as `runc ps` will get a list of all the processes on the system and then runC filters the output based on what PIDs are inside the container cgroups. So, I don't understand why you would want to mislead someone by adding `ps -ef` to the output of things inside the container, when in fact the process is **not** inside the container.

keloyang:
thank you,I got it.

PR #1009, opened 2016-08-30
Fix default cgroup path

hqhq:
Alternative of #895 , part of #892

The intension of current behavior if to create cgroup in
parent cgroup of current process, but we did this in a
wrong way, we used devices cgroup path of current process
as the default parent path for all subsystems, this is
wrong because we don't always have the same cgroup path
for all subsystems.

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>

For example, the cgroup path of current process is:
```
qhuang@ubuntu-25:~/tools/runc$ cat /proc/self/cgroup
11:hugetlb:/user/1000.user/64.session
10:perf_event:/user/1000.user/64.session
9:blkio:/user/1000.user/64.session
8:freezer:/user/1000.user/64.session
7:devices:/user/1000.user/64.session
6:memory:/user
5:name=systemd:/user/1000.user/64.session
4:cpuacct:/user/1000.user/64.session
3:cpu:/user/1000.user/64.session
2:cpuset:/user/1000.user/64.session
```

Before this patch, we create contianer cgroup in:
```
qhuang@ubuntu-25:~/tools/runc$ sudo runc run hq
root@runc:/# cat /proc/self/cgroup
11:hugetlb:/user/1000.user/64.session/hq
10:perf_event:/user/1000.user/64.session/hq
9:blkio:/user/1000.user/64.session/hq
8:freezer:/user/1000.user/64.session/hq
7:devices:/user/1000.user/64.session/hq
6:memory:/user/1000.user/64.session/hq       <<<<---- the difference
5:name=systemd:/user/1000.user/64.session/hq
4:cpuacct:/user/1000.user/64.session/hq
3:cpu:/user/1000.user/64.session/hq
2:cpuset:/user/1000.user/64.session/hq
```

After this patch:
```
qhuang@ubuntu-25:~/tools/runc$ sudo runc run hq
root@runc:/# cat /proc/self/cgroup
11:hugetlb:/user/1000.user/64.session/hq
10:perf_event:/user/1000.user/64.session/hq
9:blkio:/user/1000.user/64.session/hq
8:freezer:/user/1000.user/64.session/hq
7:devices:/user/1000.user/64.session/hq
6:memory:/user/hq                                    <<<<---- the difference
5:name=systemd:/user/1000.user/64.session/hq
4:cpuacct:/user/1000.user/64.session/hq
3:cpu:/user/1000.user/64.session/hq
2:cpuset:/user/1000.user/64.session/hq
```

PR #1013, opened 2016-09-01
Fix runc ps issue

hqhq:
After #1009, we don't always set `cgroup.Paths`, so
`getCgroupPath()` will return wrong cgroup path because
it'll take current process's cgroup as the parent, which
would be wrong when we try to find the cgroup path in
`runc ps`.

Fix it by using `m.GetPath()` to get the true cgroup
paths.

Also Fixes: #1034
Fixes: #1031 

Reported-by: Yang Shukui <yangshukui@huawei.com>

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -196,17 +196,8 @@ func (m *Manager) Set(container *configs.Config) error {
 		return nil
 	}
 	for _, sys := range subsystems {
-		// Generate fake cgroup data.
-		d, err := getCgroupData(container.Cgroups, -1)
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

hqhq:
I added a commit to fix update and pause issues, PTAL @crosbymichael @cyphar 

BTW janky didn't cache this because we use `runc run -d` and it's the same process that execute all other runc commands, so it always get the right cgroup path. I didn't think of an easy way to enhance our test to cover this scenario. Thoughts?

cyphar:
LGTM. Are there any other manager functions that we need to replace?

hqhq:
@cyphar Not as I can find, now `getCgroupData` is only called by `Apply`, and this doesn't impact systemd because it use different way to get cgroup path.

crosbymichael:
LGTM

PR #1014, opened 2016-09-01
Send signal to a container process inside the container's PID

keloyang:
As https://github.com/opencontainers/runc/pull/808 mentioned, I try another way to send signal to container, it send signal to a container process inside the container's PID.
this pr dependent upon https://github.com/opencontainers/runc/pull/1013
test code:
```
root@ubuntu:~/workspace/gocode/src/own/test/bin# cat sig.go 
package main

import (
        "fmt"
        "os"
        "os/signal"
        "syscall"
)

func main() {
        fmt.Println(os.Getpid())
        c := make(chan os.Signal, 0)
        signal.Notify(c)
        for {
                s := <-c
                fmt.Println("Got signal:", s)
                if s == syscall.SIGINT {
                        signal.Stop(c)
                        break
                }
        }
        fmt.Println("recv SIGINT, signal done")
}
go build sig.go
```
test:
```
root@ubuntu:~/workspace/runc-test# ./runc run test
/ # ./sig 
7
Got signal: user defined signal 1
Got signal: user defined signal 1
...
root@ubuntu:~/workspace/runc-test# ./runc kill --pid 7 test SIGUSR1
```

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -4,10 +4,14 @@ package main
 
 import (
 	"fmt"
+	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"syscall"
 
+	"github.com/opencontainers/runc/libcontainer"
+	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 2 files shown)


hqhq:
This is the idea that me and @xiekeyang would like to try after he failed finding an easy way to map the inner and outer pids. But then we realized we can't find a scenario that needs to send signal to a inner pid with runc command. So I'm not sure if we really need this.

crosbymichael:
I agree with @hqhq .  lets keep using the outer/global pids for this

PR #1015, opened 2016-09-01
make ps command to print the pid in container, not the pid in host

keloyang:
make ps command to print the pid in container, not the pid in host
this pr depends upon https://github.com/opencontainers/runc/pull/1013
before this ptch:
```
root@ubuntu:~/workspace/runc-test# ./runc ps test 
root     18185 18176  0 20:43 pts/12   00:00:00 sh
root     18192 18185  0 20:43 pts/12   00:00:00 sleep 9000
```
after this patch:
```
root@ubuntu:~/workspace/runc-test# ./runc ps test    
root         1     0  0 20:20 pts/12   00:00:00 sh
root        14     1  0 20:22 pts/12   00:00:00 sleep 100000
root       130   125  0 20:35 pts/5    00:00:00 ps -ef
```

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -7,9 +7,11 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"strconv"
 	"strings"
+	"syscall"
 
+	"github.com/opencontainers/runc/libcontainer"
+	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/urfave/cli"
 )
 
@@ -29,25 +31,33 @@ var psCommand = cli.Command{
 		if err != nil {
 			return err
 		}
-
 		pids, err := container.Processes()
-		if err != nil {
+		if err != nil || len(pids) == 0 {
 			return err
 		}
-
 		if context.String("format") == "json" {
 			if err := json.NewEncoder(os.Stdout).Encode(pids); err != nil {
 				return err
 			}
 			return nil
 		}
+		return ps(container, pids, context.Args()[1:])
+	},
+}
+
+func ps(container libcontainer.Container, pids []int, psArgs []string) error {
+	cmdPs := "enter_pid_ns_ps"
+	if os.Args[0] == cmdPs {
+		err := syscall.Mount("ps", "/proc", "proc", 0, "")
+		if err != nil {
+			return err
+		}
+		defer syscall.Unmount("/proc", 0)
 
 		// [1:] is to remove command name, ex:
 		// context.Args(): [containet_id ps_arg1 ps_arg2 ...]
 		// psArgs:         [ps_arg1 ps_arg2 ...]
 		//
-		psArgs := context.Args()[1:]
-
 		if len(psArgs) > 0 && psArgs[0] == "--" {
 			psArgs = psArgs[1:]
 		}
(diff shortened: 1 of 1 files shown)


ps.go
```diff
@@ -73,28 +82,45 @@ var psCommand = cli.Command{
...
-	},
+	}
+
+	ns := fmt.Sprintf("/proc/%d/ns/pid", pids[0])
+	fd, err := syscall.Open(ns, syscall.O_RDONLY, 0644)
+	if err != nil {
+		return fmt.Errorf("open /proc/%d/ns/pid %v", pids[0], err)
+	}
+	defer syscall.Close(fd)
+	err = system.Setns(uintptr(fd), 0)
```
cyphar: Ah, okay. This is how I would like this to be done, unfortunately there are several potential problems with trying to `Setns` in Go:

1. Since Go programs are multithreaded (with no way of disabling it), this syscall can fail arbitrarily (the interface says that you should only use `setns` for single-threaded programs).

2. In SELinux (and rootless container) contexts, you'll also need to join the user namespace as well.

I think this would work better if we made `libcontainer.Container.Processes()` do all of this internally, where it goes through the full `runc exec` process (but only joins the User and PID namespaces). Then we send the PIDs to the main runC process over an AF_UNIX socket.

cyphar:
I don't remember the usecase that @hqhq gave, but with this patch `runc ps` isn't much different to `runc exec test ps`, which makes it a bit useless IMO. I'm not really a _huge_ fan of the current `runc ps` interface overall, so @hqhq can probably give you some more feedback than I can. :P

ps.go
hqhq: I still don't think we need to show inner pids for `runc ps`, the global pids would be more useful, and we can do `runc exec <id> ps -ef` if we really need to know the inner pids. Plus, `docker top` shows global pids all the time and it works well, no one complained about it, why we think runc users would like this?

Even if someone really like it, we should add option like `runc ps --inner <id>` instead of replacing the default.

hqhq:
@cyphar I don't have any usecases, I only mentioned we can give it a try following your comments in https://github.com/opencontainers/runc/pull/808 :smile: And after a bit more think, I don't think it it useful.

keloyang:
@cyphar http://man7.org/linux/man-pages/man2/setns.2.html
 A process may not be reassociated with a new mount namespace if it is multithreaded and  A multithreaded process may not change user namespace with setns(),so I think setns(...pid) is ok.However,I have not a test for selinux and user namespace ,I will have a check.
@hqhq I do think we can add --inner, How do others think ?

ps.go
cyphar: The "send PIDs over an AF_UNIX socket" means you'll get PIDs in your current PID namespace. This was one of the extensions to the current set up I was considering (it means you could use `libcontainer.Container.Processes()` inside a rootless container -- since we don't have cgroups).

crosbymichael:
I think the global pids are more useful for people making the call outside the container.  Thanks for the PR @keloyang but I think we shouldn't do this for the other reasons @hqhq said.

Sorry if there was confusion about this feature in that issue.

PR #1017, opened 2016-09-04
Append string "-dirty" to version if git repo is unclean

WeiZhang555:
If runc binary is compiled from codes with modified but not commited
files, git commits should report "COMMIT-dirty" instead of single
"COMMIT" to warn users.


hqhq:
LGTM

mrunalp:
LGTM

PR #1030, opened 2016-09-09
Add backwards compat for net classid 

### LINKED ISSUE #999: runc list gives JSON unmarshaling error
While running the runc list
gave the following error
./runc list
json: cannot unmarshal string into Go value of type uint32

Though /run/runc had state.json for the container.
This came only once. I'll try to reproduce again and report back

crosbymichael:
Fixes #999

This makes the list command handle errors better and still list containers that could be loaded.  

It also handles the unmarshaling issues for the backwards incompatable change with the net class id which changed from a string to a uint32


mrunalp:
Test failure:
```
libcontainer/cgroups/fs/util_test.go:36: cannot use configs.Resources literal (type *configs.Resources) as type configs.Resources in assignment
```

PR #1033, opened 2016-09-09
Change netclassid json tag

### LINKED ISSUE #999: runc list gives JSON unmarshaling error

crosbymichael:
Fixes #999

This allows older state files to be loaded without the unmarshal error
of the string to int conversion.

### DIFF
--- a/libcontainer/configs/cgroup_unix.go
+++ b/libcontainer/configs/cgroup_unix.go
@@ -120,5 +120,5 @@ type Resources struct {
 	NetPrioIfpriomap []*IfPrioMap `json:"net_prio_ifpriomap"`
 
 	// Set class identifier for container's network packets
-	NetClsClassid uint32 `json:"net_cls_classid"`
+	NetClsClassid uint32 `json:"net_cls_classid_u"`
 }


cyphar:
LGTM. As an aside, we need to make `runc list` more resilient to issues (it will always exit on an error rather than printing every entry it can). I've done some of that work in #774 and probably can pull it out and put it into a separate PR.

libcontainer/configs/cgroup_unix.go
```diff
@@ -120,5 +120,5 @@ type Resources struct {
 	NetPrioIfpriomap []*IfPrioMap `json:"net_prio_ifpriomap"`
 
 	// Set class identifier for container's network packets
-	NetClsClassid uint32 `json:"net_cls_classid"`
+	NetClsClassid uint32 `json:"net_cls_classidu"`
```
mrunalp: nit: Maybe name it net_cls_classid_u?

crosbymichael:
@mrunalp updated

cyphar:
LGTM.

mrunalp:
LGTM

PR #1038, opened 2016-09-12
Fix make release error

keloyang:
currently make release may have a error in some distribution(e.g. Ubuntu 14.04.3)
/bin/sh: 1: Syntax error: "(" unexpected
make: *** [release] Error 2
this pr will fix this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ RELEASE_DIR := $(CURDIR)/release
 
 VERSION := ${shell cat ./VERSION}
 
-SHELL ?= $(shell command -v bash 2>/dev/null)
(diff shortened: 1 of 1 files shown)


hqhq:
We wanted this because ppl might want to use a specific shell to interpret commands, so you can use `make SHELL=/bin/bash release` on Ubuntu because ubuntu will use `/bin/sh` for Makefile by default which will not work because we used some uncompatible commands in `make release`.

hqhq:
I'm OK we force use bash since we already have bash dependence in runc somewhere else, and it'll make most of people's life easier, so

LGTM

crosbymichael:
LGTM

PR #1039, opened 2016-09-12
Continue for list on errors

crosbymichael:
This will print out the error on stderr when loading a container but
still list everything that was sucessful.

### DIFF
--- a/list.go
+++ b/list.go
@@ -121,15 +121,18 @@ func getContainers(context *cli.Context) ([]containerState, error) {
 		if item.IsDir() {
 			container, err := factory.Load(item.Name())
 			if err != nil {
-				return nil, err
+				fmt.Fprintf(os.Stderr, "load container %s: %v\n", item.Name(), err)
+				continue
 			}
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

cyphar:
LGTM

list.go
```diff
@@ -121,15 +121,18 @@ func getContainers(context *cli.Context) ([]containerState, error) {
 		if item.IsDir() {
 			container, err := factory.Load(item.Name())
 			if err != nil {
-				return nil, err
+				fmt.Fprintf(os.Stderr, "load container %s: %v", item.Name(), err)
```
cyphar: These need to have trailing `\n` otherwise it messes up the header. Also, personally I would prefer if we printed them in the order we would've output them (but that's a stylistic thing and doesn't really matter).

crosbymichael:
updated

cyphar:
LGTM.

mrunalp:
LGTM

PR #1041, opened 2016-09-13
just fix a typo

datawolf:


cyphar:
That isn't a typo. "X sped up by 10x" is valid English.

PR #1042, opened 2016-09-13
move m.GetPaths out of the loop

datawolf:
only call m.GetPaths once is ok. os move it out of the loop.


hqhq:
LGTM

mrunalp:
LGTM

PR #1045, opened 2016-09-14
Allow recrusive generic error

hqhq:
Error sent from child process is already genericError, if
we don't allow recrusive generic error, we won't get any
cause infomation from parent process.

Before, we got:
WARN[0000] exit status 1
ERRO[0000] operation not permitted

After, we got:
WARN[0000] exit status 1
ERRO[0000] container_linux.go:247: starting container process caused "process_linux.go:359: container init caused \"operation not permitted\""

it's not pretty but useful for detecting root causes.

### DIFF
--- a/libcontainer/generic_error.go
+++ b/libcontainer/generic_error.go
@@ -67,9 +67,6 @@ func newSystemErrorWithCause(err error, cause string) Error {
 // stack frames skipped. This is only to be called by the other functions for
 // formatting the error.
 func createSystemError(err error, cause string) Error {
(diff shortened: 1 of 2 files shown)


cyphar:
LGTM. There's quite a few place we need to improve how debuggable things are. Right now, the `nsenter` fixes that I've merged only really give you "which bail invocation caused this" without the output information. Which is better than it used to be, but we should be sending the JSON payload from `nsenter.c`. I'll work on this once all of the other PRs are merged.

crosbymichael:
LGTM

Issue #1050, opened 2016-09-16
Integration test failure when running single cpu Test: update --cpuset-cpus 1


mikebrow:
I get a failure running the integration tests on a single cpu box.

Here's my environment:

DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
4.4.0-36-generic 

cpuinfo:
```
mike@mike-VirtualBox:~/go/src/github.com/opencontainers/runc$ cat /proc/cpuinfo
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 94
model name	: Intel(R) Core(TM) i5-6600 CPU @ 3.30GHz
stepping	: 3
cpu MHz		: 3312.000
cache size	: 6144 KB
physical id	: 0
siblings	: 1
core id		: 0
cpu cores	: 1
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 22
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 syscall nx rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc pni pclmulqdq monitor ssse3 cx16 sse4_1 sse4_2 movbe popcnt aes xsave avx rdrand hypervisor lahf_lm abm 3dnowprefetch rdseed clflushopt
bugs		:
bogomips	: 6624.00
clflush size	: 64
cache_alignment	: 64
address sizes	: 39 bits physical, 48 bits virtual
power management:
```
cpu_sets should be supported:
```
mike@mike-VirtualBox:~/go/src/github.com/opencontainers/runc$ grep 'cpuset' /proc/filesystems
nodes	cpuset

```
Here's the error:

```
# runc update test_update --cpuset-cpus 1 (status=1):
# failed to write 1 to cpuset.cpus: write /sys/fs/cgroup/cpuset/runc-update-integration-test/cpuset.cpus: numerical result out of range
# runc list (status=0):
# ID            PID         STATUS      BUNDLE             CREATED
# test_update   20766       running     /tmp/busyboxtest   2016-09-16T17:33:18.949026787Z

```

Here's the broken test case code:

  ```
  # update cpuset if supported (i.e. we're running on a multicore cpu)
[...]

PR #1052, opened 2016-09-18
Fix update cpuset on single processor box

hqhq:
Fixes: #1050


cyphar:
LGTM

cyphar:
Why not just set the `cpuset` to `0`, which would work on a single processor?

hqhq:
@cyphar There is no need to test it on single processor because you can't verify it by checking cgroup values anyway. And this is basically the reason we set to `1`.

cyphar:
LGTM. Fair enough.

mikebrow:
Works for me as well!

mrunalp:
LGTM

PR #1058, opened 2016-09-20
update the comment for container.Pause() method on linux

datawolf:
if a container state is `running` or `created`, the `container.Pause()`
method can set the state to pausing, and then paused.

this patch update the comment, so it can be consistent with the code.


hqhq:
Does pause a created container make sense to you? @opencontainers/runc-maintainers

crosbymichael:
LGTM

@hqhq ya, if people are using it to run execs in then pausing makes sense to me

mrunalp:
LGTM

PR #1059, opened 2016-09-21
cgroup: using WriteCgroupProc to write the specified pid into the cgroup's cgroup.procs file

datawolf:
cgroupData.join method using `WriteCgroupProc` to place the pid into
the proc file, it can avoid attach any pid to the cgroup if -1 is
specified as a pid.

so, replace `writeFile` with `WriteCgroupProc` like `cpuset.go`'s
ApplyDir method.

### DIFF
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -46,7 +46,7 @@ func (s *CpuGroup) ApplyDir(path string, cgroup *configs.Cgroup, pid int) error 
 	}
 	// because we are not using d.join we need to place the pid into the procs file
 	// unlike the other subsystems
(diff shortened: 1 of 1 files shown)


hqhq:
We should change this for all cgroup subsystems, https://github.com/opencontainers/runc/pull/956 added this but only used in cpuset.

datawolf:
@hqhq ,https://github.com/opencontainers/runc/pull/956 add `WriteCgroupProc`, and it used by `cgroupData.join ` method. other cgroup subsystems are used `join` method directly.

So, with this patch, all cgroup subsystems are all use `WriteCgroupProc`.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1066, opened 2016-09-23
Allow systemd cgroup create cgroups without applying pid

hqhq:
To be consistent with fs cgroup.


crosbymichael:
@hqhq but this will still not work for the attributes set in the unit right?

hqhq:
@crosbymichael You're right, and looks like systemd unit doesn't support to create an empty cgroup, so we'll have to leave this only for fs cgroup.

PR #1067, opened 2016-09-23
make `runc list` print the command

datawolf:
this patch add `COMMAND` column when we type `runc list` command,

``` bash
root@ubuntu:~# runc list
ID          PID         COMMAND          STATUS      BUNDLE                        CREATED
3           124705      "sleep 20"       created     /mycontainer                  2016-09-21T12:21:57.299298318Z
4           124742      "sleep 20"       created     /mycontainer                  2016-09-21T12:22:29.311509961Z
5           124937      "sleep 20"       created     /mycontainer                  2016-09-21T12:29:03.171447434Z
zzz         0           "/runtimetest"   stopped     /tmp/tmp.37nKu55OYq/busybox   2016-09-22T10:59:30.475271129Z

```

### DIFF
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -61,6 +61,9 @@ type BaseState struct {
 
 	// Config is the container's configuration.
 	Config configs.Config `json:"config"`
+
+	// Process
+	Process Process `json:"process"`
 }
 
 // BaseContainer is a libcontainer container object.
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -248,6 +248,7 @@ func (c *linuxContainer) start(process *Process, isInit bool) error {
 	}
 	// generate a timestamp indicating when the container was started
 	c.created = time.Now().UTC()
+
 	c.state = &runningState{
 		c: c,
 	}
@@ -1183,12 +1184,17 @@ func (c *linuxContainer) currentState() (*State, error) {
 	var (
 		startTime           string
 		externalDescriptors []string
-		pid                 = -1
+		pid                          = -1
+		process             *Process = &Process{}
 	)
 	if c.initProcess != nil {
 		pid = c.initProcess.pid()
 		startTime, _ = c.initProcess.startTime()
 		externalDescriptors = c.initProcess.externalDescriptors()
+		p := c.initProcess.processInfo()
+		if p != nil {
+			process = p
+		}
 	}
 	state := &State{
 		BaseState: BaseState{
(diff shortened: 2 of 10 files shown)


hqhq:
Commits should be squashed.

cyphar:
Also, we need to make it so that you can select what columns should be output with `runc list`.

datawolf:
@hqhq Squashed.

hqhq:
After a second thought, I think we should store process info into `state.json`, spec config should only be input for initialization, it's not a reliable source to get runtime information from. @cyphar WDYT?

cyphar:
Yeah, we should be getting all of the information from the `state.json` IMO.

datawolf:
@hqhq @cyphar I Agree that we should getting all info from the `state.json`.  there are two solutions:

1. only save `spec.Process.Args` to `state.json`, and `runc list` get the the comnand from `state.json`.

2. save the `Process Info` to `state.json`, and `runc list` get the the comnand from `state.json`. there are problems with this sulution,**Which Process Info should we stored, spec.Process or libcontainer.Process?**
if we save `spec.Process` , it can not reflect the real state of the Process.

hqhq:
I think it should be `libcontainer.Process` but there would not be much difference regarding the fields we need to store.

datawolf:
@hqhq @cyphar I have update the patch.
1.  the first patch save the process info to `state.json`
2.  the second patch add command column to `runc list`

datawolf:
ping @hqhq @cyphar @crosbymichael

hqhq:
Let's hold this and see how #1094 goes.

crosbymichael:
Well I didn't want to comment because I have conflicting views from @cyphar on this.  I think we should store as little as possible and defer to the system for information like this.  It could be a small performance hit for recreating the state but the tradeoff is that it is correct.  

If we take the case of the container's command being a shell script that does an `exec`, what we store in the state.json would be incorrect, if we get this info from `/proc` then we would display the correct command running.  However, it could not be what the user expects either, they may expect to the the shell script as the command here.  

Anyway it goes I would not want to add *more* to the `state.json` and defer back to the bundle if we go down the route of just displaying the static command.

cyphar:
> if we get this info from /proc then we would display the correct command running.

Certain processes (like [postgres][1]) change what `/proc/self/cmdline` contains to a value that is not the actual command that was executed. This is again a matter of personal taste (I personally would prefer to know what  command was used to start a container rather than what process is init in the container _right now_ -- I can get that information from `runc ps`), but you're replacing the exec problem with a different problem -- not to mention that `argv[0]` can be set to any random value by `execve`.

[1]: https://github.com/postgres/postgres/blob/71fc49dfe1d99bd83cd99c2e7a39f93e07d19310/src/backend/utils/misc/ps_status.c

crosbymichael:
ok, defering to proc is probably not ideal so we know for sure that we want to display the command that was specified, not what is actually running.

cyphar:
Out of interest @crosbymichael are you already working on this? I might be able to work on it, but OCID is taking up quite a bit of my time recently.

crosbymichael:
@cyphar no i'm not working on this.  just reviewed the PR

cyphar:
Sorry, I thought I was commenting on #1094.

PR #1068, opened 2016-09-23
MaskPaths: support directory

AkihiroSuda:
For example, the /sys/firmware directory should be masked because it can contain some sensitive files (https://github.com/docker/docker/pull/26618):
  - /sys/firmware/acpi/tables/{SLIC,MSDM}: Windows license information:
  - /sys/firmware/ibft/target0/chap-secret: iSCSI CHAP secret

IMO the issue should be fixed but no need to get this PR in v1.0

~~~runtime-spec PR: https://github.com/opencontainers/runtime-spec/pull/582~~~


### DIFF
--- a/libcontainer/README.md
+++ b/libcontainer/README.md
@@ -83,6 +83,7 @@ config := &configs.Config{
 	},
 	MaskPaths: []string{
 		"/proc/kcore",
+		"/sys/firmware",
 	},
 	ReadonlyPaths: []string{
 		"/proc/sys", "/proc/sysrq-trigger", "/proc/irq", "/proc/bus",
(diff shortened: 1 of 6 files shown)


crosbymichael:
We should just use the maskedPaths for this.  You can handle the not a dir error and change the logic

AkihiroSuda:
OK, updated PR

crosbymichael:
LGTM

Is there any better alternative to using tmpfs mounts for this?

mrunalp:
LGTM

AkihiroSuda:
@crosbymichael Alternative idea is bind-mounting an empty directory, but it requires managing the empty directory. So I chose tmpfs here.

PR #1053, opened 2016-09-18
enhance runc delete command

datawolf:
This patch enhance the `runc delete` command as following

1) when `runc delete` without one container-id

```
# runc delete
runc: "delete" requires a minimum of 1 argument
```
2) we can delete more containers at one command

for example:

```
# runc list
ID          PID         STATUS      BUNDLE         CREATED
a           8490        created     /mycontainer   2016-09-18T03:49:32.259760434Z
b           8520        running     /mycontainer   2016-09-18T03:49:36.999299944Z
c           8535        created     /mycontainer   2016-09-18T03:49:40.975277538Z
d           8549        created     /mycontainer   2016-09-18T03:49:42.675282602Z
e           8562        running     /mycontainer   2016-09-18T03:49:44.175400931Z
# runc delete a b cc
cannot delete container b that is not stopped: running
container cc is not exist
# runc list
ID          PID         STATUS      BUNDLE         CREATED
b           8520        running     /mycontainer   2016-09-18T03:49:36.999299944Z
c           8535        created     /mycontainer   2016-09-18T03:49:40.975277538Z
d           8549        created     /mycontainer   2016-09-18T03:49:42.675282602Z
e           8562        running     /mycontainer   2016-09-18T03:49:44.175400931Z
# runc delete -f b c d e
# runc list
ID          PID         STATUS      BUNDLE      CREATED

Signed-off-by: Wang Long <long.wanglong@huawei.com>

PR #1069, opened 2016-09-23
[integration] add testcases for `runc delete` command

datawolf:
this patch add two testcases for the `runc delete` with
multi-containers. see : https://github.com/opencontainers/runc/pull/1053


crosbymichael:
LGTM

hqhq:
LGTM

PR #1018, opened 2016-09-04
Consoles, consoles, consoles.

cyphar:
This is the big console rewrite. For some design information, see [this](https://gist.github.com/cyphar/8c6b9db84fc1f2cc2d037ef07942ca83). TODO list:
- [x] Implement the console masterfd passing.
- [x] Rewrite parts of `tty.go` to work with the new setup.
- [x] Remove `--console` and add `--console-socket`.
  - [x] Need to figure out a sane protocol for sending things to `--console-socket`.
  - [x] Write a small implementation of a `--console-socket` consumer as a sanity check [can be done in a separate PR].
- [x] Figure out how to make the test suite behave.
  - [x] `TestExecInTTY`.
  - [x] bats `integration`.
- [x] Need to figure out why runC is blocking on `recvfd()`.

Fixes #814, #910, docker/docker#11462, docker/docker#8755 and countless other issues.
~~Depends on #975, #977.~~
docker/docker#9212 and similar issues are **not** fixed by this patchset, as we still need to consider how to handle `/dev/console` in certain cases.

Signed-off-by: Aleksa Sarai asarai@suse.de

PR #1071, opened 2016-09-23
Setup default stdio for child process

hqhq:
Otherwise if we enable process.terminal, we won't see
any output in child process before it dup stdio from
pty slave.


cyphar:
This will break the current implementation of #1018 -- though I probably should implement the "autodetection" differently. IMO this is part of a larger issue (providing a way for duping over different fds as I discussed with @mrunalp in [this ocid PR](https://github.com/kubernetes-incubator/ocid/issues/24#issuecomment-247787282)) -- but maybe we could implement this after #1018 (so we don't have to redo all of the review done by @mikebrow and @mrunalp).

crosbymichael:
I think with the console changes and the fact that we don't want error output going through the container's stdio, we can close this.  Errors are returned via the pipes and IO is dupped in time for us not to miss any of the user's processes data

PR #1072, opened 2016-09-24
Bug fix for make dbuild

keloyang:
If we modify the code of runc without a new branch of runc, the runc will be compiled with the code in the old runc-dev:** image. This pr will make the compiling use right code like make unittest and make integration.


keloyang:
@hqhq Added.

hqhq:
LGTM

crosbymichael:
LGTM

PR #1073, opened 2016-09-24
Add Dockerfile.dev to improve development efficiency

keloyang:
If we have a new branch and then we make test, it will spend a long time to build the runc_dev:**(branch name) image, but most of the time it is not  necessary. This pr will modify it to use a runc-dev:latest image,if we  need ot update runc_dev:latest image,just make image-update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,11 @@ man:
 	man/md2man-all.sh
 
 runcimage:
+ifneq ($(FORCE_RUNC_IMAGE), )
+	docker tag $(FORCE_RUNC_IMAGE) $(RUNC_IMAGE)
+else
 	docker build -t $(RUNC_IMAGE) .
+endif
 
 test:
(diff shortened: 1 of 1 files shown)


cyphar:
It shouldn't take long to build, since Docker caches previous layers. True, it'll create a new tag but does that _really_ matter? The issue with doing it like this is that I'm almost certainly going to forget to run `make image-update`. So IMO NACK.

keloyang:
@cyphar Build runc_dev is slow or not depends on the quality of the network.
thank for remind me about ' forget to run make image-update', I modify another way and it will not change the default action.

hqhq:
@keloyang It's not about network quality, docker build will use cache so you don't really *rebuild* a image, only for the last step in Dockerfile that you have to copy the new codes again, so I don't think we need any improvements here.

PR #1074, opened 2016-09-24
start multi-containers with `runc start` command

datawolf:
With this patch, `runc start` command can start multi-containers
at one command,  this patch also checks the argument of the `start`
command.
```
root@ubuntu:# runc list
ID          PID         STATUS      BUNDLE         CREATED
a           0           stopped     /mycontainer   2016-09-23T08:56:42.754026567Z
b           62979       created     /mycontainer   2016-09-23T09:01:36.421976458Z
c           62993       running     /mycontainer   2016-09-23T09:01:38.105940389Z
d           63006       created     /mycontainer   2016-09-23T09:01:39.65441942Z
e           63020       created     /mycontainer   2016-09-23T09:01:40.989995515Z
root@ubuntu:# runc start
runc: "start" requires a minimum of 1 argument
root@ubuntu:# runc start a b c d e f
cannot start a container that has run and stopped
cannot start an already running container
container f is not exist
root@ubuntu:# runc list
ID          PID         STATUS      BUNDLE         CREATED
a           0           stopped     /mycontainer   2016-09-23T08:56:42.754026567Z
b           62979       running     /mycontainer   2016-09-23T09:01:36.421976458Z
c           62993       running     /mycontainer   2016-09-23T09:01:38.105940389Z
d           63006       running     /mycontainer   2016-09-23T09:01:39.65441942Z
e           63020       running     /mycontainer   2016-09-23T09:01:40.989995515Z
```
Signed-off-by: Wang Long <long.wanglong@huawei.com>

PR #1075, opened 2016-09-24
pause and resume multi-containers

datawolf:
With this patch, `runc pasue` and `runc resume` can
pause and resume multi-containers.

### DIFF
--- a/tests/integration/pause.bats
+++ b/tests/integration/pause.bats
@@ -32,3 +32,81 @@ function teardown() {
   # test state of busybox is back to running
   testcontainer test_busybox running
 }
+
+@test "runc pause and resume with multi-container" {
+  # run test_busybox1 detached
+  runc run -d --console /dev/pts/ptmx test_busybox1
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox1
+
+  # run test_busybox2 detached
+  runc run -d --console /dev/pts/ptmx test_busybox2
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox2
+
+  # pause test_busybox1 and test_busybox2
+  runc pause test_busybox1 test_busybox2
+  [ "$status" -eq 0 ]
+
+  # test state of test_busybox1 and test_busybox2 is paused
+  testcontainer test_busybox1 paused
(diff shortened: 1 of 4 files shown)


crosbymichael:
For commands that handle multiple args/containers, should we still exit 0 if one of the containers encountered an error?  For programmatic use, if `runc` returns 0 but prints the errors, most people are not going to check stderr

datawolf:
@crosbymichael yes, we should exit with non zero if all or part of containers encounters an error.

I have update the patch, and the same with https://github.com/opencontainers/runc/pull/1074
thanks

mrunalp:
I think we should return a non-zero exit code if any container encounters an error. That would signal the caller to maybe parse the stderr if it want to.

datawolf:
@mrunalp thanks, now with the patch,` runc pause/resume` exit with `1` if any container encounters an error.

hqhq:
man pages should be updated.

datawolf:
@hqhq update man pages done.

datawolf:
ping @crosbymichael @hqhq @cyphar

datawolf:
ping

tests/integration/pause.bats
```diff
@@ -32,3 +32,42 @@ function teardown() {
...
+  # pause test_busybox1 and test_busybox2
+  runc pause test_busybox1 test_busybox2
+  [ "$status" -eq 0 ]
+
+  # test state of test_busybox1 and test_busybox2 is paused
+  testcontainer test_busybox1 paused
+  testcontainer test_busybox2 paused
+
+  # resume test_busybox1 and test_busybox2
+  runc resume test_busybox1 test_busybox2
```
cyphar: Can you include a test to make sure that if you try to pause/resume containers that don't exist the other containers still get changed (with `$? != 0`)? Just copy this test and add "nonexistant" to each `runc pause` and `runc resume` command.

hqhq:
LGTM

tests/integration/pause.bats
datawolf: have done.

datawolf:
Also rebased

cyphar:
LGTM

hqhq:
LGTM

PR #1076, opened 2016-09-26
Container must not checkpoint in created state

rajasec:
Currently when we checkpoint an container in created state ( not an ideal case) the created container also got destroyed.
Since the container fails to checkpoint because of created state, it must not delete the created container without user awareness.
With this PR, container can not checkpoint in created state.
 
### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -38,6 +38,13 @@ checkpointed.`,
 		if err != nil {
 			return err
 		}
+		status, err := container.Status()
+		if err != nil {
+			return(err)
+		}
+		if status == libcontainer.Created {
+			fatalf("Container cannot be checkpointed in created state")
+		}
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1076/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

dqminh:
LGTM

cyphar:
This PR broke `gofmt`. I'll open a PR to fix it but why didn't the tests fail?

cyphar:
@caniszczyk @crosbymichael The reason the tests didn't fail is because the Jenkins job doesn't merge PRs onto master before running the tests. Should we change the job to do this?

PR #1079, opened 2016-09-27
configs: add json tags for some config

datawolf:
This patch add json tags for Cgroup.Paths and Config.Hooks.
so the content of state.json are all lowercases words.

Also this patch change `rootPropagation` to `root_propagation`

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -98,7 +98,7 @@ type Config struct {
 	Readonlyfs bool `json:"readonlyfs"`
 
 	// Specifies the mount propagation flags to be applied to /.
-	RootPropagation int `json:"rootPropagation"`
+	RootPropagation int `json:"root_propagation"`
 
 	// Mounts specify additional source and destination paths that will be mounted inside the container's
 	// rootfs and mount namespace if specified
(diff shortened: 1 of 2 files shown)


hqhq:
LGTM

libcontainer/configs/config.go
```diff
@@ -98,7 +98,7 @@ type Config struct {
 	Readonlyfs bool `json:"readonlyfs"`
 
 	// Specifies the mount propagation flags to be applied to /.
-	RootPropagation int `json:"rootPropagation"`
+	RootPropagation int `json:"root_propagation"`
```
mrunalp: will have to consider backwards compatibility with docker. cc: @crosbymichael

crosbymichael: This is why we need https://github.com/opencontainers/runc/issues/1094 ;)

crosbymichael:
Overall, I think we should wait on this and fix the state issues by working on #1094..  Since this will break backwards compat I would rather not merge this and we can spend our time with the long term state fix of only saving the pid and bundle path if everyone agrees on that approach.

mrunalp:
I agree.

crosbymichael:
@datawolf thanks but lets close this one for now so we don't have the backwards compat issues with the state fields changing .

PR #1084, opened 2016-09-27
systemd cgroup driver supports slice management

derekwaynecarr:
This patch adds support for the systemd cgroup driver to manage slices.

The ability to manage slices via the dbus api was not available until systemd v229, but some distros have back-ported it to earlier levels.  The driver verifies the ability to start a slice via a transient unit is supported prior to using the feature.

If the supplied cgroup name ends with a .slice suffix, we assume the user wanted a slice created instead of a scope.  Creation of a slice does not require a PID, so support from the cgroupfs driver for pid=-1 is carried over into systemd driver.

Finally, `ExpandSlice` is made public for use by downstream consumers.  In addition, it fixes a logic error associated with `-.slice` needing to be `/` when expanded.

Signed-off-by: derekwaynecarr <decarr@redhat.com>

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -67,12 +66,14 @@ var subsystems = subsystemSet{
 
 const (
 	testScopeWait = 4
+	testSliceWait = 4
 )
 
 var (
 	connLock                        sync.Mutex
 	theConn                         *systemdDbus.Conn
 	hasStartTransientUnit           bool
+	hasStartTransientSliceUnit      bool
 	hasTransientDefaultDependencies bool
 	hasDelegate                     bool
 )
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -159,8 +160,36 @@ func UseSystemd() bool {
 			}
 		}
 
+		// Assume we have the ability to start a transient unit as a slice
+		// This was broken until systemd v229, but has been back-ported on RHEL environments >= 219
+		// For details, see: https://bugzilla.redhat.com/show_bug.cgi?id=1370299
```
vishh: Any details on other distros?

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -159,8 +160,36 @@ func UseSystemd() bool {
...
+		hasStartTransientSliceUnit = true
+
+		// To ensure simple clean-up, we create a slice off the root with no hierarchy
```
vishh: What if runc were to be running underneath a container? Is that a supported use-case @crosbymichael ?

libcontainer/cgroups/systemd/apply_systemd.go
derekwaynecarr: I would strongly suspect that other distros will not work with systemd <= 229, we can verify the other distros when we consume this in kubernetes.  i doubt others have patched their systemd.

libcontainer/cgroups/systemd/apply_systemd.go
derekwaynecarr: this follows the same convention as the scope check, so i think it should be fine...

derekwaynecarr:
poke @mrunalp

mrunalp:
LGTM

libcontainer/cgroups/systemd/apply_systemd.go
cyphar: Can I just clarify here what you mean by "broken"? Do you mean that the interface didn't exist / always returned an error? Or do you mean that there was some issues with how systemd treated such slices (in which case, is the code you've added sufficient to protect against that)? I'm just asking because we've had some horrible issues with systemd's DBus interface and I generally don't trust that project to handle things sanely anymore.

libcontainer/cgroups/systemd/apply_systemd.go
cyphar: runC inside a container is a supported config, however in general you aren't running systemd inside that container (so this is a moot point). Though, some people work on running systemd inside containers and might have issues with this. But since it would've been broken before I'm not worried about breaking it twice.

cyphar:
I'm holding off LGTM-ing this until I can better understand what you mean by "broken".

libcontainer/cgroups/systemd/apply_systemd.go
derekwaynecarr: @cyphar - by broken, I mean you would literally get an error back from systemd when you called start transient unit where the unit was a `.slice`.  

The error response had the text: `Unit type slice does not support transient units`. 

This was fixed in https://github.com/systemd/systemd/commit/17f62e9bd00f5fefd486475861b06d3ec6b7ee10

As part of https://github.com/systemd/systemd/pull/1886

The code addresses this by:

1. attempting to start a slice, and if successful, it sets hasStartTransientUnitSlice to true.
1. if it detects you cannot, and you later attempt to start a transient unit of type slice, it will error early with the message that your systemd level did not support the feature yet.

Does that help?

cyphar: Alright, in that case I'm okay with this.

cyphar:
LGTM. Merging.

PR #1086, opened 2016-09-28
Set ambient capabilities where supported

justincormack:
Since Linux 4.3 ambient capabilities are available. If set these allow unprivileged child
processes to inherit capabilities, while at present there is no means to set capabilities
on non root processes, other than via filesystem capabilities which are not usually
supported in image formats.

With ambient capabilities non root processes can be given capabilities as well, and so
the main reason to use root in containers goes away, and capabilities work as expected.

The code falls back to the existing behaviour if ambient capabilities are not supported.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -67,7 +67,7 @@
 		},
 		{
 			"ImportPath": "github.com/syndtr/gocapability/capability",
-			"Rev": "2c00daeb6c3b45114c80ac44119e7b8801fdd852"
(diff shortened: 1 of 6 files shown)


justincormack:
cc @mrunalp @crosbymichael

justincormack:
Docker PR https://github.com/docker/docker/pull/26979 but I don't think any decisions about whether to reduce capabilities for non root users affect `runc` these are just policy.

mrunalp:
Looks fine. I'll test it out and then ack it :)

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1090, opened 2016-09-29
Remove check for binding to /

crosbymichael:
In order to mount root filesystems inside the container's mount
namespace as part of the spec we need to have the ability to do a bind
mount to / as the destination.

I cannot remember why this is here specifically, I know the reasons for `/proc` and other paths and didn't know if we just added `/` just to be safe or if we had a specific reason. 

### DIFF
--- a/libcontainer/rootfs_linux_test.go
+++ b/libcontainer/rootfs_linux_test.go
@@ -32,14 +32,6 @@ func TestCheckMountDestFalsePositive(t *testing.T) {
 	}
 }
 
-func TestCheckMountRoot(t *testing.T) {
-	dest := "/rootfs"
-	err := checkMountDestination("/rootfs", dest)
-	if err == nil {
-		t.Fatal(err)
-	}
-}
-
 func TestNeedsSetupDev(t *testing.T) {
 	config := &configs.Config{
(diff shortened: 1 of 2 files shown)


mrunalp:
LGTM

libcontainer/rootfs_linux_test.go
```diff
@@ -32,14 +32,6 @@ func TestCheckMountDestFalsePositive(t *testing.T) {
 	}
 }
 
-func TestCheckMountRoot(t *testing.T) {
-	dest := "/rootfs"
-	err := checkMountDestination("/rootfs", dest)
-	if err == nil {
-		t.Fatal(err)
-	}
-}
```
cyphar: Can we keep this test anyway, as it's part of the spec?

crosbymichael: What do you mean?

cyphar: I mean to leave the test (just change `err == nil` to `err != nil`). In fact, we could add an integration test for mounting over `/` because I get the feeling there's probably quite a few places that we aren't testing with those semantics.

cyphar:
LGTM.

PR #1092, opened 2016-09-30
simplify ps command

datawolf:
the `-p pidlist` flag of `ps` command selects the process whose process
ID numbers apper in `pidlist`([man ps])

This patch use `-p pidlist` to filter process which we want.

[man ps]: http://man7.org/linux/man-pages/man1/ps.1.html

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


hqhq:
Nice, thanks.

LGTM

crosbymichael:
LGTM

PR #1093, opened 2016-09-30
user ns json mappings in configs

rajasec:
Updated the json vars in configs as per run time specs.


cyphar:
Rejected. These are the **internal** configuration values, and thus follow our own style (which is `snake_case`).

PR #1095, opened 2016-10-07
Don't enable kernel mem if not set

### LINKED ISSUE #1083: kmem.limit error when trying to run container
`container_linux.go:247: starting container process caused "process_linux.go:258: applying cgroup configuration for process caused \"failed to set memory.kmem.limit_in_bytes, because either tasks have already joined this cgroup or it has children\""`

I don't have a kmem.limit set in my config

crosbymichael:
Don't enable the kmem limit if it is not specified in the config.

Fixes #1083


mrunalp:
LGTM
@hqhq PTAL

hqhq:
LGTM

PR #1100, opened 2016-10-09
remove O_SYNC flag to improve logrus' efficiency

keloyang:
* Why remove O_SYNC
as we know, OpenFile with O_SYNC flag, sync operation come with write every time, this can lead to low efficiency.

* Necessity of O_SYNC
O_SYNC make sure the data in buffer can be flushed to disk timely, and the log data will not lost。
but the data in buffer will not lost actually except kernel is panic, so O_SYNC is not necessary for logrus. On the other hand,as practice shows, O_SYNC is not necessary for log, and docker'log (json log driver) also have no use O_SYNC.

### DIFF
--- a/main.go
+++ b/main.go
@@ -114,7 +114,7 @@ func main() {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
 		if path := context.GlobalString("log"); path != "" {
-			f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_APPEND|os.O_SYNC, 0666)
+			f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
 			if err != nil {
 				return err
 			}


cyphar:
I personally would prefer we leave it. It's not a particularly hot path, and if a user specifically asked us to log to a file then we should make sure they get the full logs that we intended to write. I'm also fairly sure that there are a few other cases aside from kernel panics (OOM, intermediate buffers) that require `O_SYNC`.

keloyang:
@cyphar thanks for your opinion.
I have a case bellow, I name it o_sync.go (no-o_sync.go is similar o_sync.go except the removal of os.O_SYNC)
```
package main

import (
        "os"
)

func main() {
        f, err := os.OpenFile("test", os.O_CREATE|os.O_WRONLY|os.O_APPEND|os.O_SYNC, 0666)
        if err != nil {
                return
        }
        for i := 0; i < 30; i++ {
                f.Write([]byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"))
        }
}
``` 

the test result:
```
root@ubuntu:~# go build o_sync.go       
root@ubuntu:~# go build no-o_sync.go         
root@ubuntu:~# time ./o_sync 

real    0m0.755s
user    0m0.000s
sys     0m0.049s
root@ubuntu:~# time ./no-o_sync 

real    0m0.003s
user    0m0.000s
sys     0m0.003s
```
The gap between the o_sync and no-o_sync is too big.
 if we have the cases which can lead to the lost of log ,such as kernel panic or oom or other, usually the system go wrong.  In this case , it is more important to check the system error  than runc.
On the other hand, if system go wrong, it's no use to add O_SYNC.

@cyphar WDYT

cyphar:
But we aren't spamming `logrus` _that_ heavily are we? In fact, we barely use `logrus` (except in debugging) so as I said it's not a "hot path" (we're not writing hundreds of lines every few seconds). But I'm not really that fussed about it (if someone else LGTMs it I'll be okay with it).

mlaventure:
I personally agree with @cyphar first comment.  `logrus` is (almost?) only used when reporting errors after which `runc` exits. So I'd rather have `O_SYNC` be left there.

Just my 2 cents :)

dqminh:
I think we introduced O_SYNC explicitly to make sure that error log is flushed immediately, there's no reason to not do it actually.

Rejected.

keloyang:
ok ,I got it, thank you.

PR #1101, opened 2016-10-09
log improve,with --log flag, cli will log to special place where user set

keloyang:
https://github.com/opencontainers/runc/pull/1089 only deal with cli.ErrWriter, this pr will add cli.App.Writer too. And with --log flag, cli will log to special places where user set.

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"strings"
 
@@ -119,6 +118,12 @@ func main() {
 				return err
 			}
 			logrus.SetOutput(f)
+
+			if path != "/dev/null" {
+				app.Writer = &LogWriter{f}
+				app.ErrWriter = app.Writer
+				cli.ErrWriter = app.Writer
+			}
 		}
 		switch context.GlobalString("log-format") {
 		case "text":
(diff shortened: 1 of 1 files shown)


cyphar:
For future reference, you don't need to open a new PR each time you change a patchset. Just use `git push --force` and GitHub will display the new set of commits. As for the actual patchset, can you explain what problem it actually fixes (or rather, what was broken before this patch) -- I'm not sure that we use `cli.App.Writer` anywhere inside runC. Thanks.

keloyang:
@cyphar thanks for your reply.
```
default cli.App:
func NewApp() *App {
	return &App{
                                ...
				Writer:       os.Stdout,
	}
}

cli print log:  fmt.Fprintf(a.Writer, ...
```
cli.App.Writer use os.Stdout by default, so cli print common log to stdout,with this pr ,if user set --log flag,
cli.App.Writer can be changed and cli can log to  the special place where user set.

mlaventure:
@keloyang this doesn't seem to work:

```
$ sudo ../runc --log /tmp/my-log.json --log-format json run top .
container_linux.go:247: starting container process caused "exec: \"foo\": executable file not found in $PATH"
$ cat /tmp/my-log.json 
{"level":"warning","msg":"os: process already finished","time":"2016-10-10T07:24:02-07:00"}
```

keloyang:
@mlaventure, thanks, I reproduce it and update done
```
root@ubuntu:~/workspace/runc-test# ./runc --log test.json --log-format json run aa     
root@ubuntu:~/workspace/runc-test# cat test.json 
{"level":"warning","msg":"os: process already finished","time":"2016-10-11T10:47:09+08:00"}
container_linux.go:247: starting container process caused "exec: \"bash\": executable file not found in $PATH"
```

mlaventure:
@keloyang your example is still invalid, the last line of the log file is not in json format.

keloyang:
yes, I ignore it , I think the pr can be closed. thank you.

Issue #1096, opened 2016-10-08
`runc ps <id> -ef` generate wrong output after apply #1092


forever043:
After apply #1092, using "-aux" or "-ef" with "runc ps" command will generate output for all the processes, not only processes in container.

> procps-ng version 3.3.9 / Ubuntu 14.04-LTS

forever043:
@datawolf It seems "-p xxx" will be ignored when both "-aux" and "-p xxx" is used for ps util.

datawolf:
@forever043 It is indeed a problem, the `ps` command has some flags to select process[1].
`-A`, `-a`, `-e`, `T`, `-N` etc, will select all processes and ignore the flags `-p` which select process by list.
Do you know how to make it workaround?


[1] https://linux.die.net/man/1/ps

datawolf:
The new version of procps-ng(>= v3.3.10), the `ps` command introduce a `-q` option[1], The option does a similar job to the `p/-p/--pid` option, but it can select process correctly as we expect, even with `a`' `-a`, `-A`, `-e`, `d` options.

[1] https://gitlab.com/procps-ng/procps/commit/e751606fccdeed7ab79fc4324938b3aa47989842

forever043:
Is that mean we need a dependency of "procps-ng >= 3.3.10"? @datawolf

cyphar:
No, because we filter the output of `ps` using the pids inside the container. I'm not sure why it's not working in this case (and to be honest, I'm still a bit iffy about the whole `runc ps` interface) but it shouldn't have anything to do with the version of `ps`. What could be happening is that we aren't correctly selecting the `PID` column.

/cc @hqhq Any ideas?

cyphar:
[As an aside, this is why I wanted us to implement `runc` ps without calling `ps` and then parsing the output. To be fair the other solution would be a pain to implement nicely with all of the features people would want.]

cyphar:
Ah, never mind I just read through #1092 (I didn't notice it got merged). Personally I'd prefer we revert #1092 rather than add an arbitrary dependency on `procps-ng`. But that's just me.

datawolf:
@cyphar sorry for that. before we find other solution for `runc ps`, I agree to revert #1092 .

The revert pr is https://github.com/opencontainers/runc/pull/1102

PR #1102, opened 2016-10-10
Revert "simplify ps command"

datawolf:
This reverts commit 067ce21f7aaf9b2fc4a8af890e00eb0966b22ef1.

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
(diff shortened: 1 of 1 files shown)


datawolf:
fix https://github.com/opencontainers/runc/issues/1096

cyphar:
LGTM. We can apply a fixed version of the patch once we've found a nicer way of filtering the PIDs that doesn't depend on version.

hqhq:
LGTM

PR #1109, opened 2016-10-12
DupSecOpt needs to match InitLabels

rhatdan:
At some point InitLabels was changed to look for SecuritOptions
separated by a ":" rather then an "=", but DupSecOpt was never
changed to match this default.

### DIFF
--- a/libcontainer/label/label_selinux.go
+++ b/libcontainer/label/label_selinux.go
@@ -33,15 +33,19 @@ func InitLabels(options []string) (string, string, error) {
 		pcon := selinux.NewContext(processLabel)
 		mcon := selinux.NewContext(mountLabel)
 		for _, opt := range options {
-			if opt == "disable" {
+			val := strings.SplitN(opt, "=", 2)
+			if val[0] != "label" {
+				continue
+			}
+			if len(val) < 2 {
+				return "", "", fmt.Errorf("bad label option %q, valid options 'disable' or \n'user, role, level, type' followed by ':' and a value", opt)
+			}
+			if val[1] == "disable" {
 				return "", "", nil
 			}
-			if i := strings.Index(opt, ":"); i == -1 {
-				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type' followed by ':' and a value", opt)
-			}
-			con := strings.SplitN(opt, ":", 2)
-			if !validOptions[con[0]] {
-				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable, user, role, level, type'", con[0])
+			con := strings.SplitN(val[1], ":", 2)
+			if len(con) < 2 || !validOptions[con[0]] {
+				return "", "", fmt.Errorf("bad label option %q, valid options 'disable, user, role, level, type'", con[0])
 
 			}
 			pcon[con[0]] = con[1]
--- a/libcontainer/label/label_selinux_test.go
+++ b/libcontainer/label/label_selinux_test.go
@@ -18,7 +18,7 @@ func TestInit(t *testing.T) {
 			t.Log("InitLabels Failed")
 			t.Fatal(err)
 		}
-		testDisabled := []string{"disable"}
+		testDisabled := []string{"label=disable"}
 		plabel, mlabel, err = InitLabels(testDisabled)
 		if err != nil {
 			t.Log("InitLabels Disabled Failed")
@@ -28,7 +28,7 @@ func TestInit(t *testing.T) {
 			t.Log("InitLabels Disabled Failed")
 			t.FailNow()
 		}
-		testUser := []string{"user:user_u", "role:user_r", "type:user_t", "level:s0:c1,c15"}
+		testUser := []string{"label=user:user_u", "label=role:user_r", "label=type:user_t", "label=level:s0:c1,c15"}
 		plabel, mlabel, err = InitLabels(testUser)
 		if err != nil {
 			t.Log("InitLabels User Failed")
@@ -40,7 +40,7 @@ func TestInit(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		testBadData := []string{"user", "role:user_r", "type:user_t", "level:s0:c1,c15"}
+		testBadData := []string{"label=user", "label=role:user_r", "label=type:user_t", "label=level:s0:c1,c15"}
 		if _, _, err = InitLabels(testBadData); err == nil {
 			t.Log("InitLabels Bad Failed")
 			t.Fatal(err)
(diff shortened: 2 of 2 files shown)


rhatdan:
@mrunalp PTAL

runcom:
InitLabels needs to be updated as well:
```diff
diff --git a/vendor/src/github.com/opencontainers/runc/libcontainer/label/label_selinux.go b/vendor/src/github.com/opencontainers/runc/libcontainer/label/label_selinux.go
index 4493bda..01c1bb0 100644
--- a/vendor/src/github.com/opencontainers/runc/libcontainer/label/label_selinux.go
+++ b/vendor/src/github.com/opencontainers/runc/libcontainer/label/label_selinux.go
@@ -39,14 +39,14 @@ func InitLabels(options []string) (string, string, error) {
 			if i := strings.Index(opt, ":"); i == -1 {
 				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type' followed by ':' and a value", opt)
 			}
-			con := strings.SplitN(opt, ":", 2)
-			if !validOptions[con[0]] {
-				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable, user, role, level, type'", con[0])
+			con := strings.SplitN(opt, ":", 3)
+			if !validOptions[con[1]] {
+				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable, user, role, level, type'", con[1])
 
 			}
-			pcon[con[0]] = con[1]
-			if con[0] == "level" || con[0] == "user" {
-				mcon[con[0]] = con[1]
+			pcon[con[1]] = con[2]
+			if con[1] == "level" || con[1] == "user" {
+				mcon[con[2]] = con[2]
 			}
 		}
 		processLabel = pcon.Get()
diff --git a/vendor/src/github.com/opencontainers/runc/libcontainer/selinux/selinux.go b/vendor/src/github.com/opencontainers/runc/libcontainer/selinux/selinux.go
index 2a18e2a..1c971c6 100644
--- a/vendor/src/github.com/opencontainers/runc/libcontainer/selinux/selinux.go
+++ b/vendor/src/github.com/opencontainers/runc/libcontainer/selinux/selinux.go
@@ -486,10 +486,10 @@ func DupSecOpt(src string) []string {
 		con["level"] == "" {
 		return nil
 	}
-	return []string{"label=user:" + con["user"],
-		"label=role:" + con["role"],
-		"label=type:" + con["type"],
-		"label=level:" + con["level"]}
+	return []string{"label:user:" + con["user"],
+		"label:role:" + con["role"],
+		"label:type:" + con["type"],
+		"label:level:" + con["level"]}
 }
 
 // DisableSecOpt returns a security opt that can be used to disabling SELinux
```

rhatdan:
@runcom Thanks I added your changes and pushed a new version.  Also ran localunittest which should run the SELinux tests.

libcontainer/label/label_selinux.go
```diff
@@ -33,13 +33,20 @@ func InitLabels(options []string) (string, string, error) {
 		pcon := selinux.NewContext(processLabel)
 		mcon := selinux.NewContext(mountLabel)
 		for _, opt := range options {
-			if opt == "disable" {
+			val := strings.SplitN(opt, "=", 2)
+			if val[0] != "label" {
+				continue
+			}
+			if len(val) < 2 {
+				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type' followed by ':' and a value", opt)
```
mrunalp: nit: Bad -> bad

mrunalp:
I am not seeing the change to DupSecOpt in the changes that @runcom posted in his comment.

runcom:
> I am not seeing the change to DupSecOpt in the changes that @runcom posted in his comment.

@mrunalp Dan reworked that cause even if that fixed the issue in cri-o, it wasn't _correct_, we should test with this version of the patch

mrunalp:
@runcom Ack.

mrunalp:
LGTM

rhatdan:
Yes it took many passes to figure this out.  It was surprising this was working in docker at all.  But I guess it was.  Also the test suite:

```
 BUILDTAGS="seccomp selinux" localunittest
```

Was breaking.  So I fixed those up. 

BTW Other stuff breaks in the localunittest.  Not sure if there is a way to run this safely via jenkins, but I think it would keep the code in a better state.

rhatdan:
@crosbymichael PTAL

runcom:
:+1: works for me

cyphar:
LGTM

mrunalp:
LGTM

thaJeztah:
This change broke SELinux in docker; the change in this PR assumes that `label=` or `label:` is passed on to `InitLabels()` but that's not the case; see the discussion on https://github.com/docker/docker/pull/30532#issuecomment-275864816

The change from `:` to `=` (both are still supported) does not affect containerd, as the split takes place before it's passed to `InitLabels()`

Only  the _value_ to libcontainer (everything after `label=` / `label:`); see the original commit that added `--security-opt label`; https://github.com/docker/docker/commit/87e732a0f3503517d7a66804bb9a7f74977347e5

It looks like this PR is fixing up a bug in cri-o, which didn't properly pass these values. As a result SELinux labeling broke completely in docker.

for reference; this is the original code adding SELinux labeling in docker; https://github.com/docker/docker/blob/87e732a0f3503517d7a66804bb9a7f74977347e5/daemon/daemon.go#L562-L569

only `con[1]` is added to `label_opts`, and passed to `label.InitLabels()`;
```go
	for _, opt := range config.SecurityOpt {
		con := strings.SplitN(opt, ":", 2)
		if con[0] == "label" {
			label_opts = append(label_opts, con[1])
		}
	}

	if container.ProcessLabel, container.MountLabel, err = label.InitLabels(label_opts); err != nil {
```

And the code in docker 1.13; https://github.com/docker/docker/blob/v1.13.0/daemon/daemon_unix.go#L174-L198

```go
for _, opt := range config.SecurityOpt {
		if opt == "no-new-privileges" {
			container.NoNewPrivileges = true
			continue
		}

		var con []string
		if strings.Contains(opt, "=") {
			con = strings.SplitN(opt, "=", 2)
		} else if strings.Contains(opt, ":") {
			con = strings.SplitN(opt, ":", 2)
			logrus.Warn("Security options with `:` as a separator are deprecated and will be completely unsupported in 1.14, use `=` instead.")
		}

		if len(con) != 2 {
			return fmt.Errorf("invalid --security-opt 1: %q", opt)
		}

		switch con[0] {
		case "label":
			labelOpts = append(labelOpts, con[1])
		case "apparmor":
			container.AppArmorProfile = con[1]
		case "seccomp":
			container.SeccompProfile = con[1]
		default:
			return fmt.Errorf("invalid --security-opt 2: %q", opt)
		}
	}

	container.ProcessLabel, container.MountLabel, err = label.InitLabels(labelOpts)
```

No changes should be needed in libcontainer; I think this patch should be reverted

rhatdan:
I don't believe this change was for CRI-O.  But it was the old way it was done.   So there does seem to be a mixup here.

PR #1110, opened 2016-10-13
checkpoint: handle config.Devices and config.MaskPaths

avagin:
In user namespaces devices are bind-mounted from the host, so
we need to add them as external mounts for CRIU.

Reported-by: @boucher

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -546,6 +546,29 @@ func (c *linuxContainer) addCriuDumpMount(req *criurpc.CriuReq, m *configs.Mount
 	req.Opts.ExtMnt = append(req.Opts.ExtMnt, extMnt)
 }
 
+func (c *linuxContainer) addMaskPaths(req *criurpc.CriuReq) error {
+	for _, path := range c.config.MaskPaths {
+		fi, err := os.Stat(fmt.Sprintf("/proc/%d/root/%s", c.initProcess.pid(), path))
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+			return err
+		}
+		if fi.IsDir() {
+			continue
+		}
+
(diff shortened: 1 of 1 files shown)


libcontainer/container_linux.go
```diff
@@ -778,6 +790,18 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 		}
 	}
 
+	for _, path := range c.config.MaskPaths {
+		m := &configs.Mount{Destination: path, Source: "/dev/null"}
+		c.addCriuRestoreMount(req, m)
+	}
+
+	if c.config.Namespaces.Contains(configs.NEWUSER) {
```
boucher: I don't really know enough about this to know the answer, but is it definitely the case that we only want to add mounts for c.config.Devices if there's a `newuser` namespace?

crosbymichael:
Should these only be added when userns is enabled?

boucher:
@crosbymichael heh, it was originally implemented that way and I asked the opposite question. I don't know the answer.

crosbymichael:
Well when userns is not enabled we don't create bind mounts for devices, we just mknod the device so I don't think it applies.

avagin:
```
        useBindMount := system.RunningInUserNS() || config.Namespaces.Contains(configs.NEWUSER)
```
@crosbymichael We create bind-mounts if userns is enabled for a container and if a container is started from a non-root userns. I suggest to always set these options for CRIU, because if devices are not mounted, CRIU will do nothing with them.

crosbymichael:
@avagin so even though you add all these mounts for the devices in criu it won't do anything with them if they were not originally mounted?

avagin:
@crosbymichael No, it won't.

crosbymichael:
LGTM

hqhq:
LGTM

PR #1113, opened 2016-10-14
Add --no-subreaper flag for create command

WeiZhang555:
`create` command should own a `--no-subreaper` flag for consistency with
`run` command.

Also remove one unnecessary temporary variable.

### DIFF
--- a/create.go
+++ b/create.go
@@ -39,6 +39,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Usage: "specify the file to write the process id to",
 		},
 		cli.BoolFlag{
+			Name:  "no-subreaper",
+			Usage: "disable the use of the subreaper used to reap reparented processes",
+		},
+		cli.BoolFlag{
 			Name:  "no-pivot",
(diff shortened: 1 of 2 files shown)


cyphar:
This doesn't make sense, since `runc create` will exit. Implicitly, `runc` **cannot** be a subreaper because the process doesn't stick around after the container has been set up. :wink:

WeiZhang555:
@cyphar 
>Also, it looks like your change is no-op anyway (you don't modify the runner config?).

 I don't think I have to modify runner config, it will fetch `no-subreaper` flag from context directly and it's already there 
```
 r := &runner{
        enableSubreaper: !context.Bool("no-subreaper"),
...
}
```

>This doesn't make sense, since runc create will exit.

Just read more about the description of `PR_SET_CHILD_SUBREAPER`, seems you are right, I misunderstood the usage of this. 

I'll close this now, thanks for your feedback! :smile:

cyphar:
> I don't think I have to modify runner config, it will fetch no-subreaper flag from context directly and it's already there 

Yup, you're right I forgot about that. :wink:

PR #1115, opened 2016-10-14
Fixing runc panic for missing file mode

rajasec:
If the file mode is missing in devices configuration ( config.json), runc panics.

goroutine 1 [running]:
panic(0x7ca4e0, 0xc8200100d0)
        /home/raj/go/src/runtime/panic.go:481 +0x3e6
github.com/urfave/cli.HandleAction.func1(0xc8200bf2f8)
        /home/raj/go/pkg/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/urfave/cli/app.go:478 +0x38e
panic(0x7ca4e0, 0xc8200100d0)
        /home/raj/go/src/runtime/panic.go:443 +0x4e9
github.com/opencontainers/runc/libcontainer/specconv.createDevices(0xc820112000, 0xc82010c200, 0x0, 0x0)
        /home/raj/go/pkg/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go:552 +0x59b
github.com/opencontainers/runc/libcontainer/specconv.CreateLibcontainerConfig(0xc8200beb88, 0x855bc0, 0x0, 0x0)
        /home/raj/go/pkg/src/github.com/opencontainers/runc/Godeps/_workspace/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go:202 +0xfcd
main.createContainer(0xc8200d43c0, 0x7ffd99456557, 0x3, 0xc820112000, 0x0, 0x0, 0x0, 0x0)
        /home/raj/go/pkg/src/github.com/opencontainers/runc/utils_linux.go:177 +0x144
main.startContainer(0xc8200d43c0, 0xc820112000, 0x0, 0x0, 0x0, 0x0)
        /home/raj/go/pkg/src/github.com/opencontainers/runc/utils_linux.go:294 +0x189
main.glob.func13(0xc8200d43c0, 0x0, 0x0)

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -537,6 +537,8 @@ func createDevices(spec *specs.Spec, config *configs.Config) error {
 	// merge in additional devices from the spec
 	for _, d := range spec.Linux.Devices {
 		var uid, gid uint32
+		var filemode os.FileMode = 0666
+
 		if d.UID != nil {
 			uid = *d.UID
 		}
@@ -547,12 +549,15 @@ func createDevices(spec *specs.Spec, config *configs.Config) error {
 		if err != nil {
 			return err
 		}
+		if d.FileMode != nil {
+			filemode = *d.FileMode
+		}
 		device := &configs.Device{
 			Type:     dt,
 			Path:     d.Path,
 			Major:    d.Major,
 			Minor:    d.Minor,
-			FileMode: *d.FileMode,
+			FileMode: filemode,
 			Uid:      uid,
 			Gid:      gid,
 		}


hqhq:
`FileMode` is optional in runtime-spec, we should not make it mandatory in runc.

cyphar:
Also, @rajasec if you're planning on making a bunch more of these PRs I would personally prefer if you did them in a single PR (though the other maintainers might not agree).

rajasec:
@hqhq 
But if we miss the option, runc panics, do you want to ignore here instead of error out

rajasec:
@cyphar 
Initially I thought of making single PR.. not sure whether maintainers will agree, hence created separate PRs

hqhq:
@rajasec Then we should change runc to adapt this, make a default for example, a field is OPTIONAL means it's not crucial, runtime can choose the behavior when it's absent, like you don't have to set filemode when create a device in Linux.

For other cases, we can take about them separately, I don't think we should error out for the absent of OPTIONAL fields without good reasons. Otherwise we should revisit runtime-spec for these fields.

cyphar:
I agree with @hqhq -- we must **not** error out for optional fields missing. If an optional field is missing, we should be setting it to the "platform default" (which is why all of the options are pointers in the first place).

rajasec:
@hqhq @cyphar 
Agree with your comments. I would like to mention one more point on same devices structure.
UID & gid are optional as per spec, currently if it is not, it will not take platform default as per the current code.

 var uid, gid uint32
                if d.UID != nil {
                        uid = *d.UID
                }
                if d.GID != nil {
                        gid = *d.GID
                }
If we have not declared the variables, uid & gid might get  value of 0.
I dont know 0 is platform default for uid & gid.

cyphar:
@rajasec The "platform default" is not defined by the spec -- we get to pick it as an implementation. Personally I'd just make it `0:0` (owned by root) because that seems like a better choice than any other IMO.

rajasec:
@hqhq @cyphar 
I have modified not to error out, handled the panic situation. Added the default file mode.
Please have an review.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1120, opened 2016-10-18
checkpoint: fix gofmt

cyphar:
Fixes: a60040c62d03 ("Container must not checkpoint in created state")
Fixes: #1076
Signed-off-by: Aleksa Sarai <asarai@suse.de>

PR #1121, opened 2016-10-18
tests: mask: use test paths rather than /sys

cyphar:
In certain circumstances (such as the rootless containers patchset), it
is not possible to test things using /sys/firmware. In addition, we
should be testing our own functionality rather than testing protection
against /sys attacks (for which the system might already have extra
protections).

Instead, just make some fake paths in the rootfs that we then mask.
Oddly I noticed that one of the errors changed when doing this (because
before we tested removing a file from /sys/firmware which is -EPERM). So
the old test was broken.

Fixes: 53179559a15f5 ("MaskPaths: support directory")
Fixes: #1068

### DIFF
--- a/tests/integration/mask.bats
+++ b/tests/integration/mask.bats
@@ -3,54 +3,61 @@
 load helpers
 
 function setup() {
-  teardown_busybox
-  setup_busybox
+	teardown_busybox
+	setup_busybox
+
+	# Create fake rootfs.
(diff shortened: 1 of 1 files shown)


cyphar:
Tests will fail because of #1120.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #1124, opened 2016-10-18
Ignore error when starting transient unit that already exists

derekwaynecarr:
If a unit already exists with the specified name, ignore the error.

This allows the call to be idempotent when doing slice management on systemd 231+.

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -282,7 +282,7 @@ func (m *Manager) Apply(pid int) error {
 		}
 	}
 
-	if _, err := theConn.StartTransientUnit(unitName, "replace", properties, nil); err != nil {
+	if _, err := theConn.StartTransientUnit(unitName, "replace", properties, nil); err != nil && !isUnitExists(err) {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


euank:
LGTM; applying this patch worked for me

vishh:
lgtm

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -283,7 +283,13 @@ func (m *Manager) Apply(pid int) error {
 	}
 
 	if _, err := theConn.StartTransientUnit(unitName, "replace", properties, nil); err != nil {
-		return err
+		if dbusError, ok := err.(dbus.Error); ok {
```
crosbymichael: The logic is a little confusing.  It would be better to have this as a function and reduce a few ifs.

```go
if _, err := theConn.StartTransientUnit(unitName, "replace", properties, nil); err != nil && !isUnitExist(err) {
    return err
}
```

derekwaynecarr: @crosbymichael that works for me, will update.

derekwaynecarr:
@crosbymichael - updated per request.

mrunalp:
LGTM

PR #1125, opened 2016-10-18
rootfs: make pivot_root not use a temporary directory

### LINKED ISSUE #1122: Provide an option to specify pivot root directory
Right now runc seems to assume that container root file system is writable and *always* tries to create a directory needed for pivot_root().

Slowly people are moving towards moving container rootfs read-only and don't expect runc to be scribbling on container rootfs. 

May be provide an option say --pivot-root-dir which specifies the directory to be used as pivot_root dir inside container rootfs. That way runc does not have to create/remove it and image builder/caller is responsible to make sure it is present.

cyphar:
Namely, use an undocumented feature of pivot_root(2) where
pivot_root(".", ".") is actually a feature and allows you to make the
old_root be tied to your /proc/self/cwd in a way that makes unmounting
easy. Thanks a lot to the LXC developers which came up with this idea
first.

This is the first step of many to allowing runC to work with a
completely read-only rootfs.

Fixes #1122.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -83,7 +83,7 @@ func setupRootfs(config *configs.Config, console *linuxConsole, pipe io.ReadWrit
 	if config.NoPivotRoot {
 		err = msMoveRoot(config.Rootfs)
 	} else {
-		err = pivotRoot(config.Rootfs, config.PivotDir)
+		err = pivotRoot(config.Rootfs)
 	}
 	if err != nil {
 		return newSystemErrorWithCause(err, "jailing process inside rootfs")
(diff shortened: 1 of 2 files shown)


mrunalp:
cool, testing on RHEL.

mrunalp:
@rhvgoyal could you also try this out?

mrunalp:
okay, this messed up the mounts on my host while testing.

mrunalp:
After running the tests on the host, this is all that remained. 
```
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime,seclabel)
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
devtmpfs on /dev type devtmpfs (rw,nosuid,seclabel,size=922032k,nr_inodes=230508,mode=755)
tmpfs on /run type tmpfs (rw,nosuid,nodev,seclabel,mode=755)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,seclabel,mode=755)
/dev/vda1 on / type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
rpc_pipefs on /var/lib/nfs/rpc_pipefs type rpc_pipefs (rw,relatime)
```

cyphar:
Ah, fun. Yeah, this is kinda what I was worried about. I'll need to play around with this more -- the code for `sys_pivot_root` is not very fun code. My guess is that there's something very wrong with how `umount(.)` is acting.

cyphar:
@mrunalp Does this patch help?

```diff
diff --git a/libcontainer/rootfs_linux.go b/libcontainer/rootfs_linux.go
index 4f21416ce233..0014cb82e6e6 100644
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -10,6 +10,7 @@ import (
        "os/exec"
        "path"
        "path/filepath"
+       "runtime"
        "strings"
        "syscall"
        "time"
@@ -566,6 +567,13 @@ func setupPtmx(config *configs.Config, console *linuxConsole) error {
 // pivotRoot will call pivot_root such that rootfs becomes the new root
 // filesystem, and everything else is cleaned up.
 func pivotRoot(rootfs string) error {
+       // We change the cwd during this function. Because Go is multithreaded and
+       // it doesn't appear to correctly implement POSIX thread semantics, we have
+       // to ensure we don't switch threads here. Otherwise we start napalming the
+       // host's filesystem.
+       runtime.LockOSThread()
+       defer runtime.UnlockOSThread()
+
        // While the documentation may claim otherwise, pivot_root(".", ".") is
        // actually valid. What this results in is / being the new root but
        // /proc/self/cwd being the old root. Since we can play around with the cwd
```

mrunalp:
@cyphar Nope, that patch doesn't help :/

libcontainer/rootfs_linux.go
```diff
@@ -563,48 +563,60 @@ func setupPtmx(config *configs.Config, console *linuxConsole) error {
...
+	if err := syscall.Fchdir(oldroot); err != nil {
+		return err
 	}
-	// path to pivot dir now changed, update
-	pivotDir = filepath.Join(pivotBaseDir, filepath.Base(pivotDir))
 
-	// Make pivotDir rprivate to make sure any of the unmounts don't
-	// propagate to parent.
-	if err := syscall.Mount("", pivotDir, "", syscall.MS_PRIVATE|syscall.MS_REC, ""); err != nil {
+	if err := syscall.Unmount(".", syscall.MNT_DETACH); err != nil {
```
avagin: You have to mark all old mounts as private before umounting them
syscall.Mount("", ".", "", syscall.MS_PRIVATE|syscall.MS_REC, "")

avagin:
@cyphar @mrunalp mount(NULL, ".", NULL, MS_REC | MS_PRIVATE, NULL) before umount(".") should solve this problem

rhvgoyal:
This looks like a cool idea. Will try tomorrow morning.

libcontainer/rootfs_linux.go
cyphar: Heh, whoops. I'll fix this.

cyphar: I've applied this fix. PTAL.

libcontainer/rootfs_linux.go
```diff
@@ -563,48 +563,66 @@ func setupPtmx(config *configs.Config, console *linuxConsole) error {
...
+	// Make oldroot rprivate to make sure our unmounts don't propogate to the
+	// host (and thus bork the machine).
+	if err := syscall.Mount("", ".", "", syscall.MS_PRIVATE|syscall.MS_REC, ""); err != nil {
+		return err
+	}
+	// Preform the unmount.
+	if err := syscall.Unmount(".", syscall.MNT_DETACH); err != nil {
 		return err
 	}
+	syscall.Close(oldroot)
```
rhvgoyal: You already have a `defer syscall.Close(oldroot)`. I am wondering why this call is required.

libcontainer/rootfs_linux.go
```diff
@@ -563,48 +563,66 @@ func setupPtmx(config *configs.Config, console *linuxConsole) error {
...
+		if err := syscall.PivotRoot(".", "."); err != nil {
 			return fmt.Errorf("pivot_root %s", err)
 		}
 	}
-	if err := syscall.Chdir("/"); err != nil {
-		return fmt.Errorf("chdir / %s", err)
+
+	// We cannot umount(".") because currently our . is newroot. So we need to
+	// switch back to oldroot before doing a MNT_DETACH (since we still have an
+	// open fd to it).
```
rhvgoyal: So looks like you are changing `cwd` of the process and root of the process continues to be `rootfs`. IOW, `.` being new root is not the problem. Probably problem is that `.` is cwd of the process and it will keep mounts busy. Hence you are changing cwd to oldroot so that unmount succeeds. 

If this is correct, it might be a good idea to modify comment a bit to reflect that.

crosbymichael:
code looks good.

Is it safe to test on my machine? i'm scared....

rhvgoyal:
I tested it on my machine and it worked for me.

libcontainer/rootfs_linux.go
cyphar: `MNT_DETACH` should allow us to unmount it without changing directories, from my reading of `pivot_root` it does a lot of tomfoolery that I believe might change the cwd. But I'll double check.

cyphar: I'm confused -- I just removed this line and the code still works. So I think you're right (that `.` is still `oldroot`) but you're wrong that `MNT_DETACH` will fail if we unmount without changing directories.

libcontainer/rootfs_linux.go
cyphar: My thinking was to make it clear that we didn't need that fd after that point, but I've removed it.

cyphar:
@rhvgoyal While the `fchdir(oldroot)` isn't _necessary_ (from what I can see) I decided to play it safe and keep it anyway (updating the comment to note that it isn't really necessary) -- since `pivot_root` doesn't really guarantee what your `cwd` will be after you call it.

hqhq:
I think it's safer to keep `fchdir(oldroot)` in case kernel `pivot_root` change behavior.

LGTM

libcontainer/rootfs_linux.go
```diff
@@ -563,48 +563,66 @@ func setupPtmx(config *configs.Config, console *linuxConsole) error {
...
 		// Try again
-		if err := syscall.PivotRoot(rootfs, pivotDir); err != nil {
+		if err := syscall.PivotRoot(".", "."); err != nil {
 			return fmt.Errorf("pivot_root %s", err)
 		}
 	}
-	if err := syscall.Chdir("/"); err != nil {
-		return fmt.Errorf("chdir / %s", err)
+
+	// Currently our "." is oldroot (according to the current kernel code).
```
rhvgoyal: I am not sure what does this comment mean. I suspect you are saying that cwd of process has not changed and oldroot is continuing to be cwd of the process?

If yes, old root is "/" and that's was not necessarily cwd of the process at the time of call to pivot_root(). 

May be say something like. `pivot_root() does not guarantee what will happen to cwd of the calling process. It is a possibility that pivot_root() changed cwd to new
root and in that case following umount() will fail. So to be safe, change dir to oldroot`.

cyphar: I think I dropped an "in". It should've read "Currently our "." is **in** oldroot ...".

mrunalp:
LGTM

libcontainer/rootfs_linux.go
cyphar: I don't think you're right (I thought about it some more). `pivot_root(a, b)` will make `a` the new root and mount the old root at `b`. What we're doing here is unmounting the old root, because the old root is also on `b`. Since `b` is `.` it matters whether or not `pivot_root` has changed your `cwd`. It turns out that this does change your `cwd` in the current Linux implementation, but I don't want to depend on that (the mount logic is very dodgy in the kernel).

PR #1127, opened 2016-10-19
Fixes set memory to unlimited

boynux:
Should set Swap to '-1' (unlimited) if memory is set to unlimited. Otherwise kernel validation fails.
these changes required to fix this issue in Docker project: https://github.com/docker/docker/pull/22578

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -16,7 +16,11 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
-const cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
+const (
+	cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
+	cgroupMemorySwapLimit   = "memory.memsw.limit_in_bytes"
+	cgroupMemoryLimit       = "memory.limit_in_bytes"
+)
 
 type MemoryGroup struct {
 }
@@ -96,9 +100,18 @@ func setKernelMemory(path string, kernelMemoryLimit int64) error {
 }
 
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
+	// If the memory update is set to -1 we should also set swap to -1
+	// -1 means unlimited memory
+	if cgroup.Resources.Memory == -1 {
+		// Only set swap if it's enbled in kernel
+		if cgroups.PathExists(filepath.Join(path, cgroupMemorySwapLimit)) {
+			cgroup.Resources.MemorySwap = -1
+		}
+	}
+
 	// When memory and swap memory are both set, we need to handle the cases
 	// for updating container.
-	if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap > 0 {
+	if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap != 0 {
 		memoryUsage, err := getMemoryData(path, "")
 		if err != nil {
 			return err
@@ -107,29 +120,29 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 		// When update memory limit, we should adapt the write sequence
 		// for memory and swap memory, so it won't fail because the new
 		// value and the old value don't fit kernel's validation.
-		if memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap) {
-			if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
+		if cgroup.Resources.MemorySwap == -1 || memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap) {
+			if err := writeFile(path, cgroupMemorySwapLimit, strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
 				return err
 			}
-			if err := writeFile(path, "memory.limit_in_bytes", strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
+			if err := writeFile(path, cgroupMemoryLimit, strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
 				return err
 			}
 		} else {
-			if err := writeFile(path, "memory.limit_in_bytes", strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
+			if err := writeFile(path, cgroupMemoryLimit, strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
 				return err
 			}
-			if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
+			if err := writeFile(path, cgroupMemorySwapLimit, strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
 				return err
 			}
 		}
 	} else {
 		if cgroup.Resources.Memory != 0 {
-			if err := writeFile(path, "memory.limit_in_bytes", strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
+			if err := writeFile(path, cgroupMemoryLimit, strconv.FormatInt(cgroup.Resources.Memory, 10)); err != nil {
 				return err
 			}
 		}
-		if cgroup.Resources.MemorySwap > 0 {
-			if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
+		if cgroup.Resources.MemorySwap != 0 {
+			if err := writeFile(path, cgroupMemorySwapLimit, strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
 				return err
 			}
 		}
--- a/libcontainer/cgroups/fs/memory_test.go
+++ b/libcontainer/cgroups/fs/memory_test.go
@@ -86,47 +86,6 @@ func TestMemorySetMemoryswap(t *testing.T) {
 	}
 }
 
-func TestMemorySetNegativeMemoryswap(t *testing.T) {
-	helper := NewCgroupTestUtil("memory", t)
-	defer helper.cleanup()
-
-	const (
-		memoryBefore     = 314572800 // 300M
-		memoryAfter      = 524288000 // 500M
-		memoryswapBefore = 629145600 // 600M
-		memoryswapAfter  = 629145600 // 600M
-	)
-
-	helper.writeFileContents(map[string]string{
-		"memory.limit_in_bytes":       strconv.Itoa(memoryBefore),
-		"memory.memsw.limit_in_bytes": strconv.Itoa(memoryswapBefore),
-	})
-
-	helper.CgroupData.config.Resources.Memory = memoryAfter
-	// Negative value means not change
-	helper.CgroupData.config.Resources.MemorySwap = -1
-	memory := &MemoryGroup{}
-	if err := memory.Set(helper.CgroupPath, helper.CgroupData.config); err != nil {
-		t.Fatal(err)
-	}
-
-	value, err := getCgroupParamUint(helper.CgroupPath, "memory.limit_in_bytes")
-	if err != nil {
-		t.Fatalf("Failed to parse memory.limit_in_bytes - %s", err)
-	}
-	if value != memoryAfter {
-		t.Fatal("Got the wrong value, set memory.limit_in_bytes failed.")
-	}
-
-	value, err = getCgroupParamUint(helper.CgroupPath, "memory.memsw.limit_in_bytes")
-	if err != nil {
-		t.Fatalf("Failed to parse memory.memsw.limit_in_bytes - %s", err)
-	}
-	if value != memoryswapAfter {
-		t.Fatal("Got the wrong value, set memory.memsw.limit_in_bytes failed.")
-	}
-}
-
 func TestMemorySetMemoryLargerThanSwap(t *testing.T) {
 	helper := NewCgroupTestUtil("memory", t)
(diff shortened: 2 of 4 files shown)


libcontainer/cgroups/fs/memory.go
```diff
@@ -96,9 +96,18 @@ func setKernelMemory(path string, kernelMemoryLimit int64) error {
 }
 
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
-	// When memory and swap memory are both set, we need to handle the cases
-	// for updating container.
-	if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap > 0 {
+	// Memeory set to '-1' which is unlimited. In this situation we need to update swap to -1
+	// to avoid kernel validation failures.
+	if cgroup.Resources.Memory == -1 {
+		if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(-1, 10)); err != nil {
```
hqhq: libcontainer as a low level library usually don't do this kind of thing, it only does what it's told to do.

So if docker want to set memory as -1, it should pass memorySwap as -1 as well, libcontainer only adapt the writing sequence to make the logic right.

boynux: Good point, I just followed the convention (see the exiting code), I can add checked if you think it's required. However the `getMemoryData` is suspicious to fail too.

cyphar: It's not "unnecessary overhead". `libcontainer` is *meant* to be stupid (though clever enough to know when to be clever about it), specifically this means that we in general don't take initiative past what the user has told us to do -- it's not `libcontainer`'s job to change the settings that a user asked for.

If you wanted to idiot-proof this then you could put it inside `libcontainer/specconv` so that runC makes this easier for users. But then again, I'm not even sure if we should be doing *that* to be honest.

hqhq: This is also not libcontainer's concern, a value is assigned, libcontainer set it, simple as that. Otherwise we'll end up with  another huge validation done by Docker and other high level tools.

boynux:
@hqhq  @cyphar  @coolljt0725 Shall I update the PR or these changes are sufficient?

hqhq:
@boynux You should update your PR.

I think a change like this would be sufficient,

```
diff --git a/libcontainer/cgroups/fs/memory.go b/libcontainer/cgroups/fs/memory.go
index 2a3ccf0..d3ba108 100644
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -98,7 +98,7 @@ func setKernelMemory(path string, kernelMemoryLimit int64) error {
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
        // When memory and swap memory are both set, we need to handle the cases
        // for updating container.
-       if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap > 0 {
+       if cgroup.Resources.Memory != 0 && cgroup.Resources.MemorySwap != 0 {
                memoryUsage, err := getMemoryData(path, "")
                if err != nil {
                        return err
@@ -108,6 +108,7 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
                if memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap) {
+                       // Here includes the case that `Limit == -1 && MemorySwap == -1`
                        if err := writeFile(path, "memory.memsw.limit_in_bytes", strconv.FormatInt(cgroup.Resources.MemorySwap, 10)); err != nil {
                                return err
                        }
```

boynux:
@hqhq That's actually the first thing I've tried, the RunC assumption is that `SwapMemory == -1` means skip Swap update! And the proposed changes will break the existing tests. (Unless we want to change the behaviour with this PR) Since it's a known implementation right now I'm not sure about the impact on other depending projects.

The other thing is should I actually check for `SwapMemory == -1` at all. Since if user requests for unlimited memory we can implicitly set `SwapMemory` to '-1' as it will be rejected by Kernel otherwise.

hqhq:
@boynux There's no such assumption in RunC AFAIK and there shouldn't be. If only test case assumes that, you can change it. That might break Docker and that's something you should fix in docker/docker#22578.

Check for `SwapMemory == -1` or not doesn't matter in this case, I'm ok with either. Usually RunC shouldn't do much validations if kernel can do the same thing, unless error message from kernel is vague and we want it to be clear. If `SwapMemory` is negative and it's not `-1`, I don't think there are much valuable information we can give than `Invalid argument` which is given by kernel.

boynux:
@hqhq  Now it should work as expected! 

* If memory and swap both set to '-1' will be set both to unlimited. 
* If wrong configuration provided (for example unlimited memory and limited swap) will fail with kernel error. 
* '0' means no update (was '-1' for swap before)

For Docker there should not be any issue, but I'm not sure about other clients that may rely on previous behavior!

LK4D4:
@hqhq is this okay for you?

libcontainer/cgroups/fs/memory_test.go
```diff
@@ -492,3 +497,49 @@ func TestMemorySetOomControl(t *testing.T) {
...
 }
+
+func TestMemorySetMemorNegativeToReset(t *testing.T) {
+	helper := NewCgroupTestUtil("memory", t)
+	defer helper.cleanup()
+
+	const (
+		memoryBefore     = 314572800 // 300M
+		memoryswapBefore = 629145600 // 600M
+		memoryAfter      = 0         // Unlimited
```
hqhq: Why 0 presents unlimited? Doesn't sound reasonable. Note that unit test didn't use real cgroup fs for test, so you better add an integration test to cover this case.

hqhq: You can check cgroup values of your container like we did in: https://github.com/docker/runc/blob/master/tests/integration/cgroups.bats#L55

boynux: Thanks for your reply.
Correct, my problem was the host machine total amount of RAM :)
I found a clean solution. Will push my changes when I did some clean ups.

boynux: @hqhq I'm done with integration tests.

boynux: Seems memory swap is not enabled in build server kernel. I'm going to put a check there.

boynux:
@hqhq Any updates here?

hqhq:
Super busy these days, will take a look next week, sorry for the late.

update.go
```diff
@@ -204,6 +202,28 @@ other options are ignored.
 					*pair.dest = uint64(v)
 				}
 			}
+
+			for _, pair := range []struct {
+				opt  string
+				dest *uint64
+			}{
+				{"memory", r.Memory.Limit},
+				{"memory-swap", r.Memory.Swap},
```
hqhq: The rule that '-1' means reset applies to all `*.limit_in_bytes`, so you can do this for all memory related limitations.

tests/integration/unlimited-memory.bats
```diff
@@ -0,0 +1,73 @@
...
+    "memory": {
+        "limit": 16777216
+    },
+EOF
+    )
+    DATA=$(echo ${DATA} | sed 's/\n/\\n/g')
+    sed -i "s/\(\"resources\": {\)/\1\n${DATA}/" ${BUSYBOX_BUNDLE}/config.json
+
+    # run a detached busybox to work with
+    runc run -d --console /dev/pts/ptmx test_cgroups_unlimited_mem
```
hqhq: Needs update to use `--console-socket`.

libcontainer/cgroups/fs/memory_test.go
```diff
@@ -94,12 +94,17 @@ func TestMemorySetNegativeMemoryswap(t *testing.T) {
 		memoryBefore     = 314572800 // 300M
 		memoryAfter      = 524288000 // 500M
 		memoryswapBefore = 629145600 // 600M
-		memoryswapAfter  = 629145600 // 600M
+		memoryswapAfter  = 0         // Unlimited
```
hqhq: We took negative value means not change in unit test, and that's the point of this unit test, since that's not true after your PR, you can remove that unit test, and no need to add another.

update.go
hqhq: Yep.

boynux:
@hqhq I updated the PR. Please check again.
Jenkins is failing on another test a rerun might fix the problem.

libcontainer/cgroups/fs/memory_test.go
hqhq: @boynux Can you check my comment here and amend your unit test changes?

tests/integration/update.bats
```diff
@@ -203,6 +204,47 @@ EOF
...
+    check_cgroup_value $CGROUP_MEMORY "memory.limit_in_bytes" 33554432
+
+    # try to remove memory limit
+    runc update test_update --memory -1
+    [ "$status" -eq 0 ]
+
+    # Get System memory limit
+    SYSTEM_MEMORY=$(cat "${CGROUP_SYSTEM_MEMORY}/memory.limit_in_bytes")
+	# check memory limited is gone
+    check_cgroup_value $CGROUP_MEMORY "memory.limit_in_bytes" ${SYSTEM_MEMORY}
```
hqhq: You can do the test at line 125 so you don't have to reset to initial test value again.

boynux: @hqhq I updated the tests please have a look if everything if fine I'll merge the commits and do a master rebase. Thank you

libcontainer/cgroups/fs/memory.go
```diff
@@ -96,9 +96,18 @@ func setKernelMemory(path string, kernelMemoryLimit int64) error {
 }
 
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
+	// If the memory update is set to -1 we should also set swap to -1
+	// -1 means unlimited memory
+	if cgroup.Resources.Memory == -1 {
+		// Only set swap if it's enbled in kernel
+		if _, err := readFile(path, "memory.memsw.limit_in_bytes"); err == nil {
```
hqhq: You can use `os.Stat`, otherwise looks good to me.

hqhq:
Please squash your commits.

boynux:
Done!

boynux:
@hqhq Should I rebase my changes with master?

Note: I just did it on my machine and managed to pass all the tests. If you want I can push it now.

hqhq:
@boynux You don't have to as long as there's no conflict with master.

LGTM

Thanks for your patience.

boynux:
How should we get the second approval?

libcontainer/cgroups/fs/memory.go
```diff
@@ -16,7 +16,11 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
-const cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
+const (
+	cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
+	cgroupMemorySwap        = "memory.memsw.limit_in_bytes"
```
dqminh: this should also be cgroupMemorySwapLimit just for consistency ?

libcontainer/cgroups/fs/memory.go
```diff
@@ -108,28 +121,28 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 		// for memory and swap memory, so it won't fail because the new
 		// value and the old value don't fit kernel's validation.
 		if memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap) {
```
dqminh: when reset memory swap to -1 while the container still have memory.limit and memory.memsw.limit, this will fail and we will set memory limit to -1 while memsw is still in place. I think that will cause an error.
We should check if we want to reset memory / memoryswap here and change the order accordingly.

boynux: If we set memsw to -1 without changing memlimit I don't think it fails, since memory will be '0' (means no change) and it just sets memsw to -1 (which means unlimited)

Did I miss something here?

boynux:
```
runc update test_update --memory-swap -1
 [ "$status" -eq 0 ]
```
 This test should cover what you just said, doesn't it? I just added it the test suite and it passed!

boynux: @dqminh Change const name to be more "consistent" and also added that test just in case ;)

boynux:
@dqminh ping!

libcontainer/cgroups/fs/memory.go
dqminh: ah i see what's happening here. `currentmemLimit < uint64(memSwap)` with memSwap = `-1` is practically always true because uint64(-1) is basically MaxUint64.

That being said, do you think its clearer to just add `if cgroup.Resources.MemorySwap == -1 || memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap)` so we dont depend on this uint64(-1) casting. I dont really like it.

tests/integration/update.bats
```diff
@@ -123,6 +124,34 @@ function check_cgroup_value() {
         check_cgroup_value $CGROUP_MEMORY "memory.memsw.limit_in_bytes" 96468992
     fi
 
+    if [ -f "$CGROUP_MEMORY/memory.memsw.limit_in_bytes" ]; then
```
dqminh: this test can be merge into the conditional above right, make it clearer to see that we have an existing memory-swap limit too.

boynux: I'll merge them.

boynux: For casting -1 to uint64 I think that's a defined behavior and pretty explanatory, but since you missed it on the first try it says that can be more intuitive. being said that I'll also add a check explicitly for -1.

boynux:
Ping @hqhq @Deming

Changes are done, please check again.

dqminh:
LGTM

boynux:
@hqhq  sorry for the trouble but we need your approval again.

hqhq:
LGTM

PR #1128, opened 2016-10-20
fix the pid-file option for runc exec/run/create command

datawolf:
When we use `--pid-file` option of the `runc exec`  command with an relative path. it will generate file in the bundle's directory.

This patch fix that. `runc exec --pid-file pid.txt` will generate file relative to the directory which we type the `runc exec` command.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -89,6 +89,9 @@ following will output a list of processes running in the container:
 		if os.Geteuid() != 0 {
 			return fmt.Errorf("runc should be run as root")
 		}
+		if err := revisePidFile(context); err != nil {
+			return err
+		}
 		status, err := execProcess(context)
 		if err == nil {
 			os.Exit(status)
--- a/utils.go
+++ b/utils.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -38,4 +39,23 @@ func setupSpec(context *cli.Context) (*specs.Spec, error) {
 		return nil, fmt.Errorf("runc should be run as root")
 	}
 	return spec, nil
+}
+
+func revisePidFile(context *cli.Context) error {
+	pidFile := context.String("pid-file")
+	if pidFile == "" {
+		return nil
+	}
+
+	// convert pid-file to an absolute path so we can write to the right
+	// file after chdir to bundle
+	pidFile, err := filepath.Abs(pidFile)
+	if err != nil {
+		return err
+	}
+	err = context.Set("pid-file", pidFile)
+	if err != nil {
+		return err
+	}
+	return nil
 }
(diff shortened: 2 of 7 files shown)


exec.go
```diff
@@ -119,6 +119,13 @@ func execProcess(context *cli.Context) (int, error) {
 	if err != nil {
 		return -1, err
 	}
+
+	// getPidFile must be called before getProcess function
```
datawolf: ok, update done. 
thanks.

utils.go
```diff
@@ -39,3 +41,20 @@ func setupSpec(context *cli.Context) (*specs.Spec, error) {
...
 	return spec, nil
 }
+
+func getPidFile(context *cli.Context) (string, error) {
+	pidFile := context.String("pid-file")
+	if pidFile == "" {
+		return pidFile, nil
+	}
+
+	if !path.IsAbs(pidFile) {
```
cyphar: You can just use `filepath.IsAbs` here. In general we avoid using `path`.

utils.go
```diff
@@ -39,3 +40,20 @@ func setupSpec(context *cli.Context) (*specs.Spec, error) {
 	}
 	return spec, nil
 }
+
+func getPidFile(context *cli.Context) (string, error) {
+	pidFile := context.String("pid-file")
+	if pidFile == "" {
+		return pidFile, nil
+	}
```
laijs: it seams the following code reinvent the filepath.Abs();

datawolf: @laijs @cyphar Thanks.

Fixed.

datawolf:
Also rebased.

crosbymichael:
Should we use this same function for getting the pid file path on run/create too?

datawolf:
@crosbymichael Yes, we should.
The `run/create` command chdir to bundle directory very earlier at `setupSpec` function. So for these two command. we can revise the `context.String("pid-file")` before `setupSpec`.

Also rebased.

exec.go
```diff
@@ -119,6 +119,14 @@ func execProcess(context *cli.Context) (int, error) {
 	if err != nil {
 		return -1, err
 	}
+
+	// convert pid-file to an absolute path so we can write to the right
+	// file after chdir to bundle
+	pidFile, err := getPidFile(context)
+	if err != nil {
+		return -1, err
+	}
```
cyphar: Why do the other two use `ctx.Set` while this one just uses the `pidFile` return value?

datawolf: @cyphar Thanks, I think that all of the three command can use `ctx.Set` method to revise `pid-fle`

I have updated the code and make some code refactoring.

cyphar:
LGTM.

hqhq:
Can you add a test case for this?

datawolf:
@hqhq The current integration tests have already used `pid-file` with a relative path.
because the working directory is the same as bundle directory in tests, so the tests can not find this issue.

I think it is not necessary to add new tests, we can change the current working directory to an directory different with bundle path on exist tests.

hqhq:
Anyway, needs to cover the scenario you're trying to fix here.

hqhq:
It doesn't worth a separate bats file for doing this, you can add test cases in run.bats, exec.bats etc.. 
And usually we don't split test cases in a separate commit, you can squash.

datawolf:
@hqhq Fixed.

hqhq:
LGTM

mrunalp:
LGTM

PR #1132, opened 2016-10-20
add test cases for create command

datawolf:
This patch add test `--pid-file` option for create command.

### DIFF
--- a/tests/integration/create.bats
+++ b/tests/integration/create.bats
@@ -39,3 +39,23 @@ function teardown() {
 
   testcontainer test_busybox running
 }
+
+@test "runc create --pid-file" {
+  runc create --pid-file pid.txt --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  testcontainer test_busybox created
+
+  # check pid.txt was generated
+  [ -e pid.txt ]
+
+  run cat pid.txt
+  [ "$status" -eq 0 ]
+  [[ ${lines[0]} =~ [0-9]+ ]]
+
+  # start the command
+  runc start test_busybox
+  [ "$status" -eq 0 ]
+
+  testcontainer test_busybox running
+}


tests/integration/create.bats
```diff
@@ -39,3 +39,19 @@ function teardown() {
...
 }
+
+@test "runc create --pid-file" {
+  runc create --pid-file pid.txt --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  testcontainer test_busybox created
+
+  # check pid.txt was generated
+  [ -e pid.txt ]
```
cyphar: Maybe we should also check that `pid` matches the pid shown in `runc list`? Or is that not really worth the effort?

datawolf: It is not necessary to check that  `pid` matches the pid shown in `runc list`.  but we can check the content of the file as following:

```
run cat pid.txt
[ "status" -eq 0 ]
[[ ${lines[0]} =~ [0-9]+ ]]
```

cyphar: Sure, that'd work. Thanks.

cyphar: I'm not sure that it's "not necessary". I don't think we have a test that makes sure that `pid-file` contains the right information? That's why I was suggesting it. But yeah, that change is also fine.

datawolf: I thought for a moment.  the `pid` is sent by C program to Go program and write to file by Go. It is necessary to check it matches. I update the patch with

datawolf: I thought for a moment. the pid is sent by C program to Go program and write to file by Go. It is necessary to check it matches. I update the patch with
```
run cat pid.txt
[ "status" -eq 0 ]
[[ ${lines[0]} =~ [0-9]+ ]]
```

I will open another pr to achieve this, because on other integration  tests also need to check, eg: `start_hello.bats`,`start_detached.bats`.

thanks

cyphar:
LGTM.

hqhq:
LGTM

PR #1133, opened 2016-10-20
add test cases for exec command

datawolf:
This patch add test `--cwd`, `--env`, `--user` option for exec command.

### DIFF
--- a/tests/integration/exec.bats
+++ b/tests/integration/exec.bats
@@ -58,3 +58,41 @@ function teardown() {
   [[ ${lines[1]} == *"."* ]]
   [[ ${lines[2]} == *".."* ]]
 }
+
+@test "runc exec ls -la with --cwd" {
+  # run busybox detached
+  runc run -d --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox
+
+  runc exec --cwd /bin test_busybox pwd
+  [ "$status" -eq 0 ]
+  [[ ${output} == "/bin" ]]
+}
+
+@test "runc exec --env" {
+  # run busybox detached
+  runc run -d --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox
+
+  runc exec --env RUNC_EXEC_TEST=true test_busybox env
+  [ "$status" -eq 0 ]
+
+  [[ ${output} == *"RUNC_EXEC_TEST=true"* ]]
+}
+
+@test "runc exec --user" {
+  # run busybox detached
+  runc run -d --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox
+
+  runc exec --user 1000:1000 test_busybox id
+  [ "$status" -eq 0 ]
+
+  [[ ${output} == "uid=1000 gid=1000" ]]
+}


tests/integration/exec.bats
```diff
@@ -57,3 +57,43 @@ function teardown() {
...
 }
+
+@test "runc exec ls -la with --cwd" {
+  # run busybox detached
+  runc run -d --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox
+
+  runc exec --cwd /bin test_busybox ls -la
```
cyphar: Why not make this test use `/bin/pwd` instead -- so we can actually check that `--cwd` does what it says?

tests/integration/exec.bats
```diff
@@ -57,3 +57,43 @@ function teardown() {
...
+}
+
+@test "runc exec --user" {
+  # run busybox detached
+  runc run -d --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox
+
+  runc exec --user 1000:1000 test_busybox id
```
cyphar: I kinda want to also add a test for `--user username:groupname`. Busybox contains a `passwd` file:

```
root:x:0:0:root:/root:/bin/ash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
news:x:9:13:news:/usr/lib/news:/sbin/nologin
uucp:x:10:14:uucp:/var/spool/uucppublic:/sbin/nologin
operator:x:11:0:operator:/root:/bin/sh
man:x:13:15:man:/usr/man:/sbin/nologin
postmaster:x:14:12:postmaster:/var/spool/mail:/sbin/nologin
cron:x:16:16:cron:/var/spool/cron:/sbin/nologin
ftp:x:21:21::/var/lib/ftp:/sbin/nologin
sshd:x:22:22:sshd:/dev/null:/sbin/nologin
at:x:25:25:at:/var/spool/cron/atjobs:/sbin/nologin
squid:x:31:31:Squid:/var/cache/squid:/sbin/nologin
xfs:x:33:33:X Font Server:/etc/X11/fs:/sbin/nologin
games:x:35:35:games:/usr/games:/sbin/nologin
postgres:x:70:70::/var/lib/postgresql:/bin/sh
nut:x:84:84:nut:/var/state/nut:/sbin/nologin
cyrus:x:85:12::/usr/cyrus:/sbin/nologin
vpopmail:x:89:89::/var/vpopmail:/sbin/nologin
ntp:x:123:123:NTP:/var/empty:/sbin/nologin
smmsp:x:209:209:smmsp:/var/spool/mqueue:/sbin/nologin
guest:x:405:100:guest:/dev/null:/sbin/nologin
nobody:x:65534:65534:nobody:/:/sbin/nologin
```

cyphar:
In future, if you're going to create a series of "add test case" patch sets or something like that, feel free to group them in a single PR. Please take a look at my nits.

tests/integration/exec.bats
datawolf: ok, good idea. 

update done.

tests/integration/exec.bats
datawolf: but the current implementation of `runc exec` command only support `--user value, -u value UID(format:<uid>[:gid])`

datawolf:
> In future, if you're going to create a series of "add test case" patch sets or something like that, feel free to group them in a single PR. Please take a look at my nits.

@cyphar , got it, thanks.

tests/integration/exec.bats
```diff
@@ -57,3 +57,42 @@ function teardown() {
...
+  # run busybox detached
+  runc run -d --console /dev/pts/ptmx test_busybox
+  [ "$status" -eq 0 ]
+
+  wait_for_container 15 1 test_busybox
+
+  runc exec --cwd /bin test_busybox pwd
+  [ "$status" -eq 0 ]
+
+  [[ ${output} == *"/bin"* ]]
```
cyphar: My main concern with this is that

```
% [[ "a" == *"a"* ]]; echo $?
0
% [[ "ab" == *"a"* ]]; echo $?
0
```

Can't we do something like:

```
[[ $(echo "${output}" | tr -d ' \n') == "/bin" ]]
```

cyphar: Actually you can just do `[[ ${line[0]} == "/bin" ]]`

datawolf: thanks for your review.

but `[[ $(echo "${output}" | tr -d ' \n') == "/bin" ]]` and `[[ ${lines[0]} == "/bin" ]]` do not works. the same as `[[  ${output} == "/bin" ]]`.

and I write a simple test `bats` scripts, it test ok

```
#!/usr/bin/env bats

@test "pwd test" {
   run runc exec --cwd /bin test_busybox pwd
   [[ ${output} == "/bin" ]]
}
```
So I don't know why they all do not work in `tests/integration/*.bats`?

hqhq: @datawolf `[[ $(echo "${output}" | tr -d '\r') == "/bin" ]]` will work.

hqhq: This is interesting since we should use `\n` on unix. @cyphar Any idea how this comes?

hqhq: Well, looks like it comes from `bats` itself, and they denial to say it is an issue, https://github.com/sstephenson/bats/issues/78 , so we would live with that.

cyphar: It's not a bats issue. For some reason, runc acts weirdly with `pwd`. Here's an strace log for runC just doing `runc exec test pwd`.

```
[pid 22107] getcwd("/", 192)            = 2
[pid 22107] fstat(1, {st_mode=S_IFCHR|0600, st_rdev=makedev(136, 7), ...}) = 0
[pid 22107] mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7ffbaf0b6000
[pid 22107] write(1, "/\n", 2)          = 2
[pid 22099] <... read resumed> "/\r\n", 32768) = 3
[pid 22107] exit_group(0)               = ?
[pid 22099] futex(0xc589d0, FUTEX_WAKE, 1) = 1
[pid 22098] <... futex resumed> )       = 0
[pid 22099] write(1, "/\r\n", 3/
 <unfinished ...>
[pid 22098] select(0, NULL, NULL, NULL, {0, 20} <unfinished ...>
[pid 22099] <... write resumed> )       = 3
```

datawolf: thanks @hqhq . why the output has a '\r' at the end? If we have no way to delete this character. A simpler way of doing it may be:
```
[[ ${output} == $(echo -e "/bin\r") ]]
```

Update patch done.

cyphar: It also happens with other commands! What!

```
% sudo ./runc exec test echo lol | cat -v
lol^M
```

cyphar: @datawolf 

>  If we have no way to delete this character. A simpler way of doing it may be:

I prefer using `tr -d '\r'. Or just using the variable regex replacement used in the linked bats issue.

datawolf: update done. @cyphar

cyphar: I've opened #1145 to track this. I have no clue what's going on, but I have a suspicion it's related to creating ptys...

cyphar:
My testing shows that #1146 fixes the issue with `\r`.

datawolf:
OK, after #1146 merged, I will update this pr.

datawolf:
As the #1146 has been merged.  I update this pr and rebased.

hqhq:
I don't see the updates, you are still `tr -d '\r'`.

datawolf:
@hqhq, Sorry for that. I just forgot to `git add` the update. Now is updated

hqhq:
LGTM

cyphar:
LGTM

PR #1140, opened 2016-10-23
Updating bash completion for ps command

rajasec:
Updating bash completion for ps command.

### DIFF
--- a/contrib/completions/bash/runc
+++ b/contrib/completions/bash/runc
@@ -289,6 +289,22 @@ _runc_pause() {
 	esac
 }
 
+_runc_ps() {
+	local boolean_options="
+	   --help
+	   -h
+	"
+
+	case "$cur" in
+		-*)
+			COMPREPLY=( $( compgen -W "$boolean_options $options_with_args" -- "$cur" ) )
+			;;
+		*)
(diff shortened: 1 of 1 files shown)


cyphar:
Can we add the bash completion for the host machine's `ps` command? Or is that just not going to work because it'd be managed by a separate package? :P

rajasec:
it will work for the runc ps command. since we are putting in .bashrc I hope it will work with specific executable. I dont think it will work with host ps.

cyphar:
No, I mean that it would be nice if `runc ps <tab>` would allow you to tab-complete the command-line options for `ps`. But it's probably not nicely doable in a distro-independent way.

rajasec:
This PR will do runc ps <tab> allow to complete the container names, not the options of ps.

cyphar:
I understand that, I was saying that it **would** be nice **if we could** complete `ps` options. And I'm wondering whether you know of a way we could do it without redefining all of the ps options ourselves (namely -- can we source the bash-completion for ps). My guess is that we probably can't but I thought I'd ask because you might know more than me about it.

LGTM. Doesn't matter.

mrunalp:
LGTM

PR #1142, opened 2016-10-23
Updating blkio default value in man page

rajasec:
0 is invalid value for blkio value and will be rejected by kernel. Internally in runc implementation also we did not explicitly set 0 value. So it will take from parent cgroup.
updated the man page accordingly.

### DIFF
--- a/man/runc-update.8.md
+++ b/man/runc-update.8.md
@@ -33,7 +33,7 @@ other options are ignored.
 
 # OPTIONS
    --resources value, -r value  path to the file containing the resources to update or '-' to read from the standard input
-   --blkio-weight value         Specifies per cgroup weight, range is from 10 to 1000 (default: 0)
+   --blkio-weight value         Specifies per cgroup weight, range is from 10 to 1000 (0 to use system default)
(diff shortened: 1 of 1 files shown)


man/runc-update.8.md
```diff
@@ -33,7 +33,7 @@ other options are ignored.
 
 # OPTIONS
    --resources value, -r value  path to the file containing the resources to update or '-' to read from the standard input
-   --blkio-weight value         Specifies per cgroup weight, range is from 10 to 1000 (default: 0)
+   --blkio-weight value         Specifies per cgroup weight, range is from 10 to 1000 (0 to use system default)
```
cyphar: This should match the output of `runc --help` (so you have to change that as well). Also I'd make the default text match for all of the options (the line below has a different format).

hqhq:
@rajasec Can you update and rebase?

AkihiroSuda:
needs rebase

kolyshkin:
This text was changed in PR #3032 (commit dfc0f069) to remove these details (now it only says "Set a new io weight"). I believe if someone is using `runc update` to change io weight for a running container, they know what they do.

PR #1143, opened 2016-10-23
Restoring container not in created state

rajasec:
Two use cases where the container is destroyed upon restore.

1) I have created the container using runc create.
post that when I tried to call runc restore, since the container was not in checkpoint
it failed and destroyed the created container.
Not an valid use case though :) 

Ideally created container should not get destroyed (  unless asked by delete  command or through proper container life cycle operations)

2) runc create test
   runc exec -p ~/process.json test 
 I  logged to terminal and application was running
 runc restore test

Since the restore failed, it went and deleted the container.

For both the scenarios, container was in created state.

Added the condition to prevent restore when the container in created state.

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -144,7 +144,9 @@ func restoreContainer(context *cli.Context, spec *specs.Spec, config *configs.Co
 	if status == libcontainer.Running {
 		fatalf("Container with id %s already running", id)
 	}
-
+	if status == libcontainer.Created {
+		fatalf("Container cannot be restored from created state")
+	}
 	setManageCgroupsMode(context, options)
 
 	if err := setEmptyNsMask(context, options); err != nil {


cyphar:
/cc @avagin 

I'm confused by your description (and the error message). You're checking whether or not the container you've loaded is `Created`, so you're actually checking whether or not you're going to clobber a container that already exists. (While I don't use c/r that much) one of the possible paths here is that you create the container (if it doesn't already exist) and then you try to restore into it -- which won't work with your patch. Maybe you should only do that if the `factory.Create` logic ran right before we do the check? I'm not sure, @avagin is the local expert on this code.

rajasec:
@cyphar 
If the container is in created state, if we tries to call runc restore command, created container is destroyed.
Though it throws the error correctly. I was expecting container which was created should have existed.
instead of modifying libcontainer I have added pre-check

cyphar:
@hqhq Why have you merged `master` into two different PRs and pushed them?

hqhq:
@cyphar It's a mistaken when I work https://github.com/opencontainers/runc/issues/1152 out with Chris, I clicked that `Update branch` button and appears that's a mistake now, I didn't find a way to revert unfortunately.

@rajasec Looks like need you to rebase, sorry about that.

rajasec:
@hqhq 
I have rebased.

crosbymichael:
We cannot do this because if you look at this line we create the container:

https://github.com/opencontainers/runc/pull/1143/files#diff-3c0547477fac6e623af9d08f7b12a4f3L133

rajasec:
Got it.

PR #1146, opened 2016-10-24
libcontainer: io: stop screwing with \n in console output

### LINKED ISSUE #1145: io: container output becomes windows-ised
For some reason, all container output gets a `\n` translated to `\r\n` on output. This is not caused directly by runC (we're not adding it), the `read` syscall is literally giving us a windows version of what the container process ran. This happens for both `runc run` and `runc exec`.

```
% runc run test | cat -v
lol^M
% runc exec test echo lol | cat -v
lol^M
```

Here's a snippet of the strace output for `runc exec test pwd`:

```
[pid 22107] getcwd("/", 192)            = 2
[pid 22107] fstat(1, {st_mode=S_IFCHR|0600, st_rdev=makedev(136, 7), ...}) = 0
[pid 22107] mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7ffbaf0b6000
[pid 22107] write(1, "/\n", 2)          = 2
[pid 22099] <... read resumed> "/\r\n", 32768) = 3
[pid 22107] exit_group(0)               = ?
[pid 22099] futex(0xc589d0, FUTEX_WAKE, 1) = 1
[pid 22098] <... futex resumed> )       = 0
[pid 22099] write(1, "/\r\n", 3/
 <unfinished ...>
[pid 22098] select(0, NULL, NULL, NULL, {0, 20} <unfinished ...>
[pid 22099] <... write resumed> )       = 3
```

I've tried playing around with `stty` and nothing seems to help.

This was discovered as part of #1133 as it affects how we write tests.

cyphar:
The default terminal setting for a new pty on Linux (unix98) has +ONLCR,
resulting in '\n' writes by a container process to be converted to
'\r\n' reads by the managing process. This is quite unexpected, and
causes multiple issues with things like bats testing. To fix it, make
the terminal sane after opening it by setting -ONLCR.

This patch might need to be rewritten after the console rewrite patchset
is merged.

Fixes #1145.

### DIFF
--- a/libcontainer/console_linux.go
+++ b/libcontainer/console_linux.go
@@ -17,6 +17,9 @@ func NewConsole(uid, gid int) (Console, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err := saneTerminal(master); err != nil {
+		return nil, err
+	}
 	console, err := ptsname(master)
 	if err != nil {
 		return nil, err
@@ -142,4 +145,27 @@ func ptsname(f *os.File) (string, error) {
 		return "", err
 	}
 	return fmt.Sprintf("/dev/pts/%d", n), nil
+}
+
+// saneTerminal sets the necessary tty_ioctl(4)s to ensure that a pty pair
+// created by us acts normally. In particular, a not-very-well-known default of
+// Linux unix98 ptys is that they have +onlcr by default. While this isn't a
+// problem for terminal emulators, because we relay data from the terminal we
+// also relay that funky line discipline.
+func saneTerminal(terminal *os.File) error {
+	// Go doesn't have a wrapper for any of the termios ioctls.
+	var termios syscall.Termios
+
+	if err := ioctl(terminal.Fd(), syscall.TCGETS, uintptr(unsafe.Pointer(&termios))); err != nil {
+		return fmt.Errorf("ioctl(tty, tcgets): %s", err.Error())
+	}
+
+	// Set -onlcr so we don't have to deal with \r.
+	termios.Oflag &^= syscall.ONLCR
+
+	if err := ioctl(terminal.Fd(), syscall.TCSETS, uintptr(unsafe.Pointer(&termios))); err != nil {
+		return fmt.Errorf("ioctl(tty, tcsets): %s", err.Error())
+	}
+
+	return nil
 }
--- a/libcontainer/integration/execin_test.go
+++ b/libcontainer/integration/execin_test.go
@@ -61,6 +61,9 @@ func TestExecIn(t *testing.T) {
 	if !strings.Contains(out, "cat") || !strings.Contains(out, "ps") {
 		t.Fatalf("unexpected running process, output %q", out)
 	}
+	if strings.Contains(out, "\r") {
+		t.Fatalf("unexpected carriage-return in output")
+	}
 }
 
 func TestExecInUsernsRlimit(t *testing.T) {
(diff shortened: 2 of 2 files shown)


hqhq:
My local experimental change looks like:
```
diff --git a/Godeps/_workspace/src/github.com/docker/docker/pkg/term/tc_linux_cgo.go b/Godeps/_workspace/src/github.com/docker/docker/pkg/term/tc_linux_cgo.go
index d47cf59..2804089 100644
--- a/Godeps/_workspace/src/github.com/docker/docker/pkg/term/tc_linux_cgo.go
+++ b/Godeps/_workspace/src/github.com/docker/docker/pkg/term/tc_linux_cgo.go
@@ -25,6 +25,8 @@ func MakeRaw(fd uintptr) (*State, error) {

        C.cfmakeraw((*C.struct_termios)(unsafe.Pointer(&newState)))
        newState.Oflag = newState.Oflag | C.OPOST
+       newState.Oflag = newState.Oflag ^ C.ONLCR
+
        if err := tcset(fd, &newState); err != 0 {
                return nil, err
        }
diff --git a/tty.go b/tty.go
index 5a76ebe..415d2e2 100644
--- a/tty.go
+++ b/tty.go
@@ -66,7 +66,8 @@ func createTty(p *libcontainer.Process, rootuid, rootgid int, consolePath string
        go io.Copy(console, os.Stdin)
        go io.Copy(os.Stdout, console)

-       state, err := term.SetRawTerminal(os.Stdin.Fd())
+       //state, err := term.SetRawTerminal(os.Stdin.Fd())
+       state, err := term.SetRawTerminal(console.Fd())
        if err != nil {
                return nil, fmt.Errorf("failed to set the terminal from the stdin: %v", err)
        }
```
I'm not sure if we should set stdin as raw mode.
BTW this default +onlcr is really suspicious.

hqhq:
```
$ stty -a
speed 38400 baud; rows 52; columns 211; line = 0;
intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W; lnext = ^V; flush = ^O; min = 1; time = 0;
-parenb -parodd cs8 -hupcl -cstopb cread -clocal -crtscts
-ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl ixon -ixoff -iuclc -ixany -imaxbel -iutf8
opost -olcuc -ocrnl onlcr -onocr -onlret -ofill -ofdel nl0 cr0 tab0 bs0 vt0 ff0
isig icanon iexten echo echoe echok -echonl -noflsh -xcase -tostop -echoprt echoctl echoke
```
Looks like linux default tty setting enables onlcr? Do we still miss something?

cyphar:
We shouldn't set stdin in raw mode. The default `+ONCLR` is a **kernel** default (you can take a look at `drivers/pty/n_tty.c` if you want to hate yourself). The reason you don't notice it normally is that a `pty` is only created by your **terminal emulator** for your shell. You never see the weird stuff because all subprocess output is done with non-magical pipes.

So this isn't weird, it's just that we need to make it appear as though the output came from a pipe. We can work on this later, but it's just not nice atm.

cyphar:
> Do we still miss something?

I don't think so. And if we end up missing something we can always update `SaneTerminal` to make the change.

libcontainer/utils/utils_unix.go
```diff
@@ -31,3 +34,26 @@ func CloseExecFrom(minFd int) error {
 	}
 	return nil
 }
+
+// SaneTerminal sets the necessary tty_ioctl(4)s to ensure that a pty pair
+// created by us acts normally. In particular, a not-very-well-known default of
+// Linux unix98 ptys is that they have +onlcr by default. While this isn't a
+// problem for terminal emulators, because we relay data from the terminal we
+// also relay that funky line discipline.
+func SaneTerminal(terminal *os.File) error {
```
cyphar: I'm actually not sure (I think I wanted to make it so other people could use it but I'm not sure why that's necessary). Fixed.

mrunalp:
LGTM

crosbymichael:
I'm guessing this fixes https://github.com/docker/docker/pull/22450 right?

mrunalp:
Looks like it should.

crosbymichael:
LGTM

PR #1147, opened 2016-10-25
Check pid file

datawolf:
At https://github.com/opencontainers/runc/pull/1132#pullrequestreview-5017081, It is necessary to make sure that pid-file contains the right information.

This patch employ `jq` and `runc state` command to achieve it.

### DIFF
--- a/tests/integration/create.bats
+++ b/tests/integration/create.bats
@@ -51,7 +51,7 @@ function teardown() {
 
   run cat pid.txt
   [ "$status" -eq 0 ]
-  [[ ${lines[0]} =~ [0-9]+ ]]
+  [[ ${lines[0]} == $(__runc state test_busybox | jq '.pid') ]]
 
   # start the command
   runc start test_busybox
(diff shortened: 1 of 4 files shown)


cyphar:
LGTM. Thanks. :smile_cat:

tests/integration/create.bats
```diff
@@ -51,7 +51,7 @@ function teardown() {
 
   run cat pid.txt
   [ "$status" -eq 0 ]
-  [[ ${lines[0]} =~ [0-9]+ ]]
```
mikebrow: Why delete the digit test here.... but not below.

tests/integration/exec.bats
```diff
@@ -42,6 +42,7 @@ function teardown() {
   run cat pid.txt
   [ "$status" -eq 0 ]
   [[ ${lines[0]} =~ [0-9]+ ]]
```
mikebrow: here?

mikebrow: thx missed the !=

LGTM

PR #1150, opened 2016-10-26
Detect and forbid duplicated namespace in spec

WeiZhang555:
When spec file contains duplicated namespaces, e.g.
```
specs: specs.Spec{
        Linux: &specs.Linux{
            Namespaces: []specs.Namespace{
                {
                    Type: "pid",
                },
                {
                    Type: "pid",
                    Path: "/proc/1/ns/pid",
                },
            },
        },
    }
```
runc should report malformed spec instead of using latest one by
default, because this spec could be quite confusing.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -187,6 +187,9 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 		if !exists {
 			return nil, fmt.Errorf("namespace %q does not exist", ns)
 		}
+		if config.Namespaces.Contains(t) {
+			return nil, fmt.Errorf("malformed spec file: duplicated ns %q", ns)
+		}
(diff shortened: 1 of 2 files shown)


wking:
Related spec work in opencontainers/runtime-spec#583 and
opencontainers/runtime-spec#584, although I haven't filed a similar PR
for namespaces yet.

cyphar:
While I'm fine with this, I'd prefer to get a runtime-spec PR landed first.

mrunalp:
LGTM

WeiZhang555:
@cyphar I can do that at the same time :) 
If runc is OK for this, I think runtime-spec won't reject this either :stuck_out_tongue_closed_eyes:

WeiZhang555:
@wking So if that spec change was accepted, I think this could make more sense.
I'm glad to help making the PR if you don't mind. :smile:

hqhq:
LGTM

PR #1151, opened 2016-10-26
integration: Make sure pivot_root() works even if parent mount is shared

rhvgoyal:
We recently had regression where parent mount of rootfs is "shared" and
rootfsPropataion=private. Write an integration test to test this
configuration, to make avoid future regressions.

Signed-off-by: Vivek Goyal vgoyal@redhat.com

### DIFF
--- a/tests/integration/mount_propagation.bats
+++ b/tests/integration/mount_propagation.bats
@@ -1,0 +1,44 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+  teardown_busybox
+  setup_busybox
+
+  # Create a "shared" parent mount point for container
+  # rootfs and make sure pivot_root() still works. This
+  cd ${BUSYBOX_BUNDLE}
+  mount --bind . .
+  mount --make-private .
+  mount --make-shared .
+
+  # Also use rootPropagation=private (and not rprivate) so that shared volumes
+  # can possibly work.
+  sed -i '/"linux": {/ a "rootfsPropagation":"private"', config.json
+}
+
+function teardown() {
+  cd ${BUSYBOX_BUNDLE}/../
+  umount -R ${BUSYBOX_BUNDLE}
+  teardown_busybox
+}
+
+@test "mount propagation parent mount shared" {
+  runc create --console /dev/pts/ptmx test_busybox1
+  [ "$status" -eq 0 ]
+
+  testcontainer test_busybox1 created
+
+  # start conatiner test_busybox1
+  runc start test_busybox1
+  [ "$status" -eq 0 ]
+
+  testcontainer test_busybox1 running
+
+  # delete test_busybox1
+  runc delete --force test_busybox1
+
+  runc state test_busybox1
+  [ "$status" -ne 0 ]
+}
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ integration: runcimage
 	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localintegration
 
 localintegration: all
-	bats -t tests/integration${TESTFLAGS}
+	unshare -m bats -t tests/integration${TESTFLAGS}
 
 install:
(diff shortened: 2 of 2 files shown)


mrunalp:
@rhvgoyal Did you run this on RHEL 7?

tests/integration/mount_propagation.bats
```diff
@@ -0,0 +1,44 @@
...
+load helpers
+
+function setup() {
+  teardown_busybox
+  setup_busybox
+
+  # Create a "shared" parent mount point for container
+  # rootfs and make sure pivot_root() still works. This
+  cd ${BUSYBOX_BUNDLE}
+  mount --bind . .
```
cyphar: If you put this inside `@test` and you use `unshare -m` to run a shell command, that might work a bit nicer since it means we won't be messing with host mountpoints while testing.

hqhq:
@rhvgoyal Sorry I messed the commits, can you please rebase?

rhvgoyal:
@mrunalp no, I have not run this on rhel7. Only on fedora24.

tests/integration/mount_propagation.bats
rhvgoyal: @cyphar, not sure how to use `unshare -m` in this script. IIUC, I will have to have another script to run actual test and do `unshare -m <new-script>`. If I just call `unshare -m` it will just launch another bash in new mount namespace but that does not run my rest of the test code.

rhvgoyal:
@mrunalp I tested on rhel7 just now. This particular individual test passes.

`bats tests/integration/mount_propagation.bats`

But, if I try to run all the integration tests locally, it hangs. And this happens on master branch too. So that seems to be a separate problem, independent of this PR.

cyphar:
@rhvgoyal Have you tried rebooting your machine and then running all the tests of master -- your current tests will modify the **host's** mountpoints. Which is why I want to use `unshare -m` because the current tests are dangerous to say the least.

tests/integration/mount_propagation.bats
cyphar: Yes, what I'm saying is to remove this from the `setup` code (where it doesn't belong because it's changing how other tests are running) and modify your `@test` so that it does something like:

```
% tmpscript=$(mktemp)
% cat >$tmpscript <<EOF
#!/bin/sh

cd ${BUSYBOX_BUNDLE}
mount --bind . .
mount --make-private .
mount --make-shared .
runc ...
EOF
% unshare -m sh $tmpscript
% rm $tmpscript
```

Which _might_ work (though it means we couldn't use the `runc` aliases I have in `helpers.bash`.

rhvgoyal:
@cyphar, I understand that why do you want to use separate mount namespace. It makes it harder to break host setup.  If running test in separate mount namespace makes sense for this test, then I think it makes sense for all the test? Any test can decide to create a mount and we probably would want to make sure it does not spoil host environment.

How about, running all integration tests in a separate mount namespace? I have modifed Makefile to call bats in a separate mount namespace. `unshare -m bats -t tests/integration`. Will this work?

cyphar:
@rhvgoyal The problem is that I'd like a new mount namespace for *each* test. I guarantee there's a way to hack this into `bats` (I'll take a look at this later), but the main issue is that I don't want to have explicit changes of the mount namespace in one test affect other tests.

crosbymichael:
Did you two find any solution for running the tests in a separate mountns?

cyphar:
@crosbymichael Not really, though I was looking at the `bats` source code recently and it might be possible to pull it off if you mess around with how the functions are generated. Almost certainly not worth the effort though. The only really viable method is this:

```
@test "some test" {
    script="$(mktemp --tmpdir=$BATS_TMPDIR)"
    cat >$script <<EOF
        the test function
    EOF

    # Then run bats inside unshare. Not sure how to get the script arguments from a function though.
    unshare -m "$@"
}
```

kolyshkin:
This is a corner case requiring very specific and hard to implement setup, which is probably not worth the hassle of trying to implement it. Feel free to reopen though if you don't think so.

PR #1153, opened 2016-10-27
Remove unnecessary cloneflag validation

hqhq:
config.cloneflag is not mandatory, when using `runc exec`,
config.cloneflag can be empty, and even then it won't be
`-1` but `0`.

So this validation is totally wrong and unneeded.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -438,10 +438,6 @@ void nsexec(void)
 	/* Parse all of the netlink configuration. */
 	nl_parse(pipenum, &config);
 
-	/* clone(2) flags are mandatory. */
-	if (config.cloneflags == -1)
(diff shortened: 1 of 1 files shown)


cyphar:
`cloneflags` is set to `-1` by `nl_parse` if the payload didn't contain a `cloneflags` payload.

hqhq:
But this is still not mandatory, is it?

cyphar:
Actually, you're right because in the `exec` case we **want** `cloneflags` to be `0`.

cyphar:
LGTM.

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1153/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

crosbymichael:
LGTM

PR #1155, opened 2016-10-28
Remove add command in Dockerfile

datawolf:
Now, all the `docker run` command in Makefile use `-v $(CURDIR):/go/src/github.com/
opencontainers/runc`. So we do not need to add the repo in RUNC_IMAGE.


hqhq:
Image built from Dockerfile is not supposed to use only with Makefile. It makes sense that we copy source code into runc image. So ppl can know what source code is this image built from when they get an image.

Rejected.

crosbymichael:
Yes, we need to copy this in as the standard build of the dockerfile

PR #1163, opened 2016-10-30
man page update for delete command

rajasec:
Changed the status of container from destroyed to stopped in man page.


hqhq:
LGTM

crosbymichael:
LGTM

crosbymichael:
@caniszczyk did we change something on this repo with the merge settings because its currently requiring you to approve via pullapprove and the new github review system before it can be merged?  

I cannot find the setting to change this for the repo so I don't think I made the change.

crosbymichael:
@caniszczyk I actually found the setting under protected branches and removed it so that the repo is the same as the runtime spec relying on pullapprove only.

PR #1172, opened 2016-11-02
Move ambient capabilties behind build tag

crosbymichael:
This moves the ambient capability support behind an `ambient` build tag
so that it is only compiled upon request.

ping @justincormack 


cyphar:
You need to also update `make release` (and the default `BUILDTAGS`).

crosbymichael:
@cyphar i didn't want to add it to the default.  What do you think?

cyphar:
I don't _really_ mind, but AFAIK most distributions ship a new enough kernel for it to not be an issue. Also, I've actually had some issues building with `seccomp` before so I don't think that our defaults are really taken _that_ seriously by packagers (at *SUSE we enable all of the tags).

hqhq:
LGTM

mrunalp:
LGTM

cyphar:
We still need to update `make release`... I'll make a PR for that in a bit.

PR #1173, opened 2016-11-04
Allow update rt_period_us and rt_runtime_us

WeiZhang555:
Currently runc already supports setting realtime runtime and period
before container processes start, this commit will add update support
for realtime scheduler resources.

PS: I searched github issues and PRs and didn't find why `runc` didn't support updating real time scheduler resources, as I can see, setting the rt_period_us and rt_runtime_us when container is booting already works, so `updating` should be a reasonable enhancement.

Let me know if I miss something. :smile:

### DIFF
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -47,6 +47,7 @@ function check_cgroup_value() {
     [ "$current" -eq "$expected" ]
 }
 
+# TODO: test rt cgroup updating
 @test "update" {
     requires cgroups_kmem
     # run a few busyboxes detached
--- a/update.go
+++ b/update.go
@@ -40,6 +40,8 @@ The accepted format is as follow (unchanged values can be omitted):
     "shares": 0,
     "quota": 0,
     "period": 0,
+    "realtimeRuntime": 0,
+    "realtimePeriod": 0,
     "cpus": "",
     "mems": ""
   },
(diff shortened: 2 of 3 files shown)


WeiZhang555:
Ooops, seems our `janky` server doesn't CONFIG_RT_GROUP_SCHED and can't set `cpu.rt_period_us` and `cpu.rt_runtime_us`

update.go
```diff
@@ -59,17 +61,25 @@ other options are ignored.
...
 		},
 		cli.StringFlag{
+			Name:  "cpu-rt-period",
+			Usage: "CPU realtime period to be used for hardcapping (in usecs). 0 to use system default",
+		},
+		cli.StringFlag{
+			Name:  "cpu-rt-runtime",
+			Usage: "CPU realtime hardcap limit (in usecs). Allowed cpu time in a given period",
+		},
+		cli.StringFlag{
```
cyphar: Please update the documentation as well as the code when adding / changing flags.

tests/integration/update.bats
```diff
@@ -86,6 +90,16 @@ function check_cgroup_value() {
...
 
+    # update cpu-rt-period
+    runc update test_update --cpu-rt-period 900000
+    [ "$status" -eq 0 ]
+    check_cgroup_value $CGROUP_CPU "cpu.rt_period_us" 900000
+
+    # update cpu-quota
+    runc update test_update --cpu-rt-runtime 600000
+    [ "$status" -eq 0 ]
+    check_cgroup_value $CGROUP_CPU "cpu.rt_runtime_us" 600000
```
cyphar: Yeah, we can't add this to the test because we don't have `rt` enabled on our janky server kernel. One of the things that we're planning on doing is splitting this test so we can have conditional cgroup testing (using `requires`), but that'll happen once someone has some spare time :wink:. In the meanwhile, just remove this from the test.

WeiZhang555:
@cyphar I added some test cases with condition check, after your "conditional cgroup testing" comes out, we can refactor this. What do you think, is this acceptable?

WeiZhang555:
@cyphar I added some test cases with condition check, after your "conditional cgroup testing" comes out, we can refactor this. What do you think, is this acceptable?

cyphar:
@WeiZhang555 Just drop the changed test cases and add a `# TODO: test rt cgroup updating`. I don't really like splitting test cases like this (and prefer a separate test case rather than creating several versions of the same case inside one function).

WeiZhang555:
@cyphar Test cases removed and docs added, PTAL :wink:

cyphar:
LGTM

hqhq:
LGTM

PR #1176, opened 2016-11-06
libcontainer: config: validate default filesystems

cyphar:
[According to the runtime-spec](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config-linux.md#default-filesystems), some filesystems must be mounted in all
containers'. So, add verification to make sure someone doesn't ask us to
create a rootfs that is missing key features.

### DIFF
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -106,6 +106,13 @@ func TestValidateSecurityWithMaskPaths(t *testing.T) {
 				{Type: configs.NEWNS},
 			},
 		),
+		// Required to pass mount validation.
+		Mounts: []*configs.Mount{
+			{Destination: "/proc"},
+			{Destination: "/sys"},
+			{Destination: "/dev/shm"},
+			{Destination: "/dev/pts"},
+		},
 	}
 
 	validator := validate.New()
@@ -124,6 +131,13 @@ func TestValidateSecurityWithROPaths(t *testing.T) {
 				{Type: configs.NEWNS},
 			},
 		),
+		// Required to pass mount validation.
+		Mounts: []*configs.Mount{
+			{Destination: "/proc"},
+			{Destination: "/sys"},
+			{Destination: "/dev/shm"},
+			{Destination: "/dev/pts"},
+		},
 	}
 
 	validator := validate.New()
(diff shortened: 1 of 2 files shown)


cyphar:
Also, apparently `runc` will not complain if you have `mounts` set but don't have the right namespace setup. We should probably error out validation in that case too.

libcontainer/configs/validate/validator_test.go
```diff
@@ -124,6 +131,13 @@ func TestValidateSecurityWithROPaths(t *testing.T) {
 				{Type: configs.NEWNS},
 			},
 		),
+		// Required to pass mount validation.
+		Mounts: []*configs.Mount{
+			{Destination: "/proc"},
+			{Destination: "/sys"},
+			{Destination: "/dev/shm"},
+			{Destination: "/dev/pts"},
+		},
```
hqhq: One more test to make it failed with invalid mount info?

dqminh:
do we want to swap this validator package with https://github.com/opencontainers/runtime-tools. They are mostly the same .

cyphar:
@dqminh We validate `configs.Config`, while the `runtime-tools` validator validates spec configs. Not to mention that AFAIK we do more comprehensive testing of edge cases than `runtime-tools`. We should eventually upstream our validation, but we have to switch away from the legacy `configs.Config` first.

dqminh:
@cyphar yah i supposed that's fair.

mrunalp:
LGTM

crosbymichael:
I don't know who added a `MUST` to the spec but these are not required, they are just things that must applications expect.

Things like `/sys` are not required and you can leave it out for added security.  Also you don't need `/dev/shm` at all for most applications and PTS is only required when you want a PTY.  

This is also weird because if we say that they are required then why even give the user the option to mess up.  

I would rather leave this as is and change the spec to a `SHOULD`.

crosbymichael:
@philips Could you take a look at this and let me know what you think since you added the section for these required filesystems?

justincormack:
Yes I have containers that really don't need or want these filesystems. They are required by (for example) most `libc` implementations, but eg only if you use `shm` do you need `/dev/shm`. `runc` is a low level tool and should allow for weird use cases. In particular if you want to totally lock down ability to write to files by having no writeable mounts you want to remove `/dev/shm`.

cyphar:
In which case, can you please make a PR against the runtime specification to switch things to `SHOULD`?

crosbymichael:
We changed this in the spec to SHOULDs so we don't need this PR, it would have created many issues if we did validate this in the first place

PR #1177, opened 2016-11-07
Add bash completions for new flags of `update`

WeiZhang555:
Add bash completions for flags "--cpu-rt-period" and "--cpu-rt-runtime"
for `update` command.

This is follow up of https://github.com/opencontainers/runc/pull/1173, sorry for forgetting the bash completions, now it's here.

### DIFF
--- a/contrib/completions/bash/runc
+++ b/contrib/completions/bash/runc
@@ -719,6 +719,8 @@ _runc_update() {
 	   --blkio-weight
 	   --cpu-period
 	   --cpu-quota
+	   --cpu-rt-period
+	   --cpu-rt-runtime
 	   --cpu-share
 	   --cpuset-cpus
 	   --cpuset-mems


cyphar:
I just found out the other day that `urfave/cli` has an automated tool to generate bash completion. Maybe we should be using that (as in automatically generating it after every change)?

cyphar:
LGTM though.

WeiZhang555:
@cyphar 

>  Maybe we should be using that (as in automatically generating it after every change)?

I've never tried this, it looks like not very user friendly：https://github.com/urfave/cli/tree/master/autocomplete#enabling , maybe I'm wrong :smile: 

By the way, I was just thinking that if we can find a way to translate `specs.Resources` into `config.Cgroups.Resources` easily, then it should be possible to allow update any Cgroup resources. So why did we provide limited flags and didn't open the door of updating all the resources? Is there any concern?

cyphar:
> By the way, I was just thinking that if we can find a way to translate specs.Resources into config.Cgroups.Resources easily, then it should be possible to allow update any Cgroup resources. So why did we provide limited flags and didn't open the door of updating all the resources? Is there any concern?

IIRC the concern in #790 was that you can't update every `spec.Resources` option (I think OOMKiller is one of them).

WeiZhang555:
@cyphar Hmmm, I read through that PR, but only find this: https://github.com/opencontainers/runc/pull/790#r62195016,  it still didn't give me the answer.

Whatever, it's another topic unrelated to this PR. I'll do more research and figure out why. Thanks :-)

dqminh:
LGTM

PR #1180, opened 2016-11-07
Add --all flag to kill

crosbymichael:
This allows a user to send a signal to all the processes in the
container within a single atomic action to avoid new processes being
forked off before the signal can be sent.

This is basically taking functionality that we already use being
`delete` and exposing it ok the `kill` command by adding a flag.

### DIFF
--- a/contrib/completions/bash/runc
+++ b/contrib/completions/bash/runc
@@ -325,6 +325,8 @@ _runc_kill() {
 	local boolean_options="
 	   --help
 	   -h
+          --all
+          -a
 	"
 
 	case "$prev" in
--- a/delete.go
+++ b/delete.go
@@ -14,10 +14,10 @@ import (
 )
 
 func killContainer(container libcontainer.Container) error {
-	container.Signal(syscall.SIGKILL)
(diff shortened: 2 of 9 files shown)


dqminh:
LGTM

datawolf:
I think it will be better to update the manual for the new flag at the same time.

WeiZhang555:
Also the `bash completions` :stuck_out_tongue_winking_eye:

cyphar:
I may be confused, but doesn't killing the init of a container guarantee that the container will die? How can a pidns survive without an init?

rhatdan:
If  you are sharing the host pid namespace or the pid namespace of another container, it will not kill all processes.

crosbymichael:
@cyphar ya, what @rhatdan said.  This expose the logic we already use on delete to signal as well so you can make atomic operations on containers that are sharing the pidns. 

I'll update the man pages and completions for this today.

crosbymichael:
I updated the completions and man pages for this new flag.

dqminh:
LGTM

mrunalp:
LGTM

PR #1184, opened 2016-11-09
Carry #781: Add support for cgroup namespace

yummypeng:
Carrying [#781](https://github.com/opencontainers/runc/pull/781). Thanks for @cyphar 's work. :smile:

Cgroup namespace can be configured in `config.json` as other
namespaces. Here is an example:

```
"namespaces": [
	{
		"type": "pid"
	},
	{
		"type": "network"
	},
	{
		"type": "ipc"
	},
	{
		"type": "uts"
	},
	{
		"type": "mount"
	},
	{
		"type": "cgroup"
	}
],
```

Note that if you want to run a container which has shared cgroup ns with
another container, then it's strongly recommended that you set
proper `CgroupsPath` of both containers(the second container's cgroup
path must be the subdirectory of the first one). Or there might be
some unexpected results.

### DIFF
--- a/libcontainer/configs/namespaces_syscall.go
+++ b/libcontainer/configs/namespaces_syscall.go
@@ -9,7 +9,7 @@ func (n *Namespace) Syscall() int {
 }
 
 // This is not yet in the Go stdlib.
-const syscall_CLONE_NEWCGROUP = (1 << 29)
+const syscall_CLONE_NEWCGROUP = (1 << 25)
 
 var namespaceInfo = map[NamespaceType]int{
 	NEWNET:    syscall.CLONE_NEWNET,
--- a/libcontainer/configs/namespaces_syscall.go
+++ b/libcontainer/configs/namespaces_syscall.go
@@ -8,13 +8,17 @@ func (n *Namespace) Syscall() int {
 	return namespaceInfo[n.Type]
 }
 
+// This is not yet in the Go stdlib.
+const syscall_CLONE_NEWCGROUP = (1 << 29)
+
 var namespaceInfo = map[NamespaceType]int{
-	NEWNET:  syscall.CLONE_NEWNET,
-	NEWNS:   syscall.CLONE_NEWNS,
-	NEWUSER: syscall.CLONE_NEWUSER,
-	NEWIPC:  syscall.CLONE_NEWIPC,
-	NEWUTS:  syscall.CLONE_NEWUTS,
-	NEWPID:  syscall.CLONE_NEWPID,
+	NEWNET:    syscall.CLONE_NEWNET,
+	NEWNS:     syscall.CLONE_NEWNS,
+	NEWUSER:   syscall.CLONE_NEWUSER,
+	NEWIPC:    syscall.CLONE_NEWIPC,
+	NEWUTS:    syscall.CLONE_NEWUTS,
+	NEWPID:    syscall.CLONE_NEWPID,
+	NEWCGROUP: syscall_CLONE_NEWCGROUP,
 }
 
 // CloneFlags parses the container's Namespaces options to set the correct
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -1694,3 +1694,60 @@ func TestTmpfsCopyUp(t *testing.T) {
 		t.Fatalf("/etc/passwd not copied up as expected: %v", outputLs)
 	}
 }
+
+func TestCGROUPPrivate(t *testing.T) {
+	if _, err := os.Stat("/proc/self/ns/cgroup"); os.IsNotExist(err) {
+		t.Skip("cgroupns is unsupported")
+	}
+	if testing.Short() {
+		return
+	}
+
+	rootfs, err := newRootfs()
+	ok(t, err)
+	defer remove(rootfs)
+
+	l, err := os.Readlink("/proc/1/ns/cgroup")
+	ok(t, err)
+
+	config := newTemplateConfig(rootfs)
+	config.Namespaces.Add(configs.NEWCGROUP, "")
+	buffers, exitCode, err := runContainer(config, "", "readlink", "/proc/self/ns/cgroup")
+	ok(t, err)
+
+	if exitCode != 0 {
+		t.Fatalf("exit code not 0. code %d stderr %q", exitCode, buffers.Stderr)
+	}
+
+	if actual := strings.Trim(buffers.Stdout.String(), "\n"); actual == l {
+		t.Fatalf("cgroup link should be private to the container but equals host %q %q", actual, l)
+	}
+}
+
+func TestCGROUPHost(t *testing.T) {
+	if _, err := os.Stat("/proc/self/ns/cgroup"); os.IsNotExist(err) {
+		t.Skip("cgroupns is unsupported")
+	}
+	if testing.Short() {
+		return
+	}
+
+	rootfs, err := newRootfs()
+	ok(t, err)
+	defer remove(rootfs)
+
+	l, err := os.Readlink("/proc/1/ns/cgroup")
+	ok(t, err)
+
+	config := newTemplateConfig(rootfs)
+	buffers, exitCode, err := runContainer(config, "", "readlink", "/proc/self/ns/cgroup")
+	ok(t, err)
+
+	if exitCode != 0 {
+		t.Fatalf("exit code not 0. code %d stderr %q", exitCode, buffers.Stderr)
+	}
+
+	if actual := strings.Trim(buffers.Stdout.String(), "\n"); actual != l {
+		t.Fatalf("cgroup link not equal to host link %q %q", actual, l)
+	}
+}
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -40,6 +40,12 @@ enum sync_t {
 	SYNC_ERR = 0xFF, /* Fatal error, no turning back. The error code follows. */
 };
 
+/*
+ * Synchronisation value for cgroup namespace setup.
+ * The same constant is defined in process_linux.go as "createCgroupns".
+ */
+#define CREATECGROUPNS 0x80
+
 /* longjmp() arguments. */
 #define JUMP_PARENT 0x00
 #define JUMP_CHILD  0xA0
@@ -570,6 +576,17 @@ void nsexec(void)
 							kill(child, SIGKILL);
 							bail("failed to sync with child: write(SYNC_RECVPID_ACK)");
 						}
+
+						/* Send the init_func pid back to our parent. */
+						len = snprintf(buf, JSON_MAX, "{\"pid\": %d}\n", child);
+						if (len < 0) {
+							kill(child, SIGKILL);
+							bail("unable to generate JSON for child pid");
+						}
+						if (write(pipenum, buf, len) != len) {
+							kill(child, SIGKILL);
+							bail("unable to send child pid to bootstrapper");
+						}
 					}
 					break;
 				case SYNC_CHILD_READY:
@@ -614,17 +631,6 @@ void nsexec(void)
 				}
 			}
 
-			/* Send the init_func pid back to our parent. */
-			len = snprintf(buf, JSON_MAX, "{\"pid\": %d}\n", child);
-			if (len < 0) {
-				kill(child, SIGKILL);
-				bail("unable to generate JSON for child pid");
-			}
-			if (write(pipenum, buf, len) != len) {
-				kill(child, SIGKILL);
-				bail("unable to send child pid to bootstrapper");
-			}
-
 			exit(0);
 		}
 
@@ -640,6 +646,7 @@ void nsexec(void)
 	case JUMP_CHILD: {
 			pid_t child;
 			enum sync_t s;
+			uint32_t actual_flags = config.cloneflags;
 
 			/* We're in a child and thus need to tell the parent if we die. */
 			syncfd = sync_child_pipe[0];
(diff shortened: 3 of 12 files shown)


libcontainer/standard_init_linux.go
```diff
@@ -79,6 +79,14 @@ func (l *linuxStandardInit) Init() error {
 	}
 
 	label.Init()
+
+	// Use Unshare() to enter a new cgroup_namespace before InitializeMountNamespace()
+	if l.config.Config.Namespaces.Contains(configs.NEWCGROUP) && l.config.Config.Namespaces.PathOf(configs.NEWCGROUP) == "" {
+		if err := syscall.Unshare(configs.Syscall_CLONE_NEWCGROUP); err != nil {
```
cyphar: This is seriously not safe. Do not **ever** call `Unshare` from inside Go code. The `Unshare` interfaces do not work in a multithreaded process (and Go programs are multithreaded). This is why `nsexec.c` exists in the first place.

libcontainer/specconv/spec_linux.go
```diff
@@ -190,6 +195,17 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 		if config.Namespaces.Contains(t) {
 			return nil, fmt.Errorf("malformed spec file: duplicated ns %q", ns)
 		}
+		// if this container is going to have shared cgroup namespace with host,
+		// we just skip the following .Add()
```
cyphar: Why?

libcontainer/standard_init_linux.go
YummyPeng: OK, I also have considered the way of using pipe to block `nsexec.c` code but got stuck someway, so I just sent this implementation for discussion. Though I would try to work it out.

cyphar: I'd be happy to give you a hand if you have any questions, I didn't mean to scare you off. :smile_cat:

libcontainer/specconv/spec_linux.go
YummyPeng: I have found that `docker` just uses `delNamespace()` to delete a namespace shared with host. If a user runs a container without adding cgroup ns in `config.json`, then this container will share cgroup ns with host and its cgroups are bind-mounted. I think it is the same with the scenario that we setup cgroup ns in `config.json` but set cgroup path as `/proc/1/ns/cgroup`. So I chose not to add cgroup ns into `config.Namespaces`. Then its cgroups will be bind-mounted without adding more codes.

libcontainer/standard_init_linux.go
YummyPeng: Haha, yeah, I know. I'm going to play around with it myself first. I would turn to you for help without hesitation if I have any questions. Thanks in advance :stuck_out_tongue_winking_eye:

YummyPeng:
@cyphar updated! :wink:

YummyPeng:
Oops, it seems that our `janky` server hasn't supported `cgroup namespace` yet.

cyphar:
@YummyPeng Yeah, I would make the tests conditional (do `t.Skip()` if `/proc/self/ns/cgroup` doesn't exist). We do a similar thing for the PIDs cgroup controller.

libcontainer/configs/namespaces_syscall.go
```diff
@@ -8,13 +8,17 @@ func (n *Namespace) Syscall() int {
 	return namespaceInfo[n.Type]
 }
 
+// This is not yet in the Go stdlib.
+const Syscall_CLONE_NEWCGROUP = (1 << 25)
```
cyphar: This shouldn't be exported.

libcontainer/integration/exec_test.go
```diff
@@ -1660,3 +1660,54 @@ func TestTmpfsCopyUp(t *testing.T) {
...
+	if testing.Short() {
+		return
+	}
+
+	rootfs, err := newRootfs()
+	ok(t, err)
+	defer remove(rootfs)
+
+	l, err := os.Readlink("/proc/1/ns/cgroup")
+	ok(t, err)
```
cyphar: As I mentioned above, this should be a `t.Skip` if `os.IsNotExist(err)`.

libcontainer/nsenter/nsexec.c
```diff
@@ -38,6 +38,9 @@ enum sync_t {
 	SYNC_ERR = 0xFF, /* Fatal error, no turning back. The error code follows. */
 };
 
+/* Synchronisation value for cgroup namespace setup */
+#define SYNC_PRIVATE_CGNS 0x80
```
cyphar: Since this synchronisation is different to the previous ones, we should consider naming it something distinctive. I actually didn't catch that this was a different type of synchronisation to the previous one on my first reading of this.

libcontainer/nsenter/nsexec.c
```diff
@@ -46,6 +49,9 @@ enum sync_t {
 /* JSON buffer. */
 #define JSON_MAX 4096
 
+/* buffer for synchronisation value */
+#define SYNC_BUF_LEN 10
```
cyphar: Since `SYNC_PRIVATE_CGNS` is a single byte, can you just follow the example of all of the other synchronisation code and not use a buffer (use a single character local variable).

libcontainer/nsenter/nsexec.c
```diff
@@ -741,6 +754,22 @@ void nsexec(void)
 					bail("failed to dup stderr");
 			}
 
+			/* ... wait until our topmost parent has finished cgroup setup in p.manager.Apply() ... */
+			if (config.cloneflags & CLONE_NEWCGROUP) {
+				len = read(pipenum, buf, SYNC_BUF_LEN);
+				if (len < 0)
+					bail("read synchronisation value failed");
+				value = strtol(buf, &endptr, 10);
```
cyphar: Yeah, this shouldn't be necessary if you just write a single byte. Also, why is value not local to this block?

libcontainer/process_linux.go
```diff
@@ -19,6 +19,9 @@ import (
 	"github.com/opencontainers/runc/libcontainer/utils"
 )
 
+// Synchronisation value for cgroup namespace setup.
+const privateCgroupns int = (1 << 7)
```
cyphar: I'm a bit mixed on this. On the one hand I think this should be implemented with an `#import` using `cgo` because I don't really want to duplicate this (especially since it's not really possible to `grep` for both usages of this constant). But then `cgo` is bad to use for something like that. :wink:

If we have to leave it here then please name it something similar to `SYNC_PRIVATE_CGNS` (after you rename that as per my comments above). Then at least it will be possible to `grep` for this constant.

libcontainer/nsenter/nsexec.c
```diff
@@ -741,6 +754,22 @@ void nsexec(void)
...
+				if (*endptr != '\0')
+					bail("unable to parse synchronisation value");
+				if (value == SYNC_PRIVATE_CGNS) {
+					if (unshare(CLONE_NEWCGROUP) < 0)
```
cyphar: I think that RHEL might have issues with this (WDYT @mrunalp). Also this is a _bit_ ugly but I'm guessing you can't combine this with the rest of the `unshare` code because we're currently sending the PID of `JUMP_INIT` not `JUMP_CHILD` to the `.Apply` code

libcontainer/process_linux.go
```diff
@@ -257,6 +260,13 @@ func (p *initProcess) start() error {
 	if err := p.manager.Apply(p.pid()); err != nil {
 		return newSystemErrorWithCause(err, "applying cgroup configuration for process")
 	}
+	// Now it's time to setup cgroup namesapce
+	if p.config.Config.Namespaces.Contains(configs.NEWCGROUP) && p.config.Config.Namespaces.PathOf(configs.NEWCGROUP) == "" {
+		if _, err := p.parentPipe.WriteString(strconv.Itoa(privateCgroupns)); err != nil {
```
cyphar: As I mentioned this should be a single byte write.

libcontainer/specconv/spec_linux.go
cyphar: My point is that you're adding **more** code to special case something that I don't understand why you need to special case it. If you do a `setns(...)` for a namespace that you've already joined then it doesn't matter anyway.

But also I don't understand why you're comparing the link to `/proc/1/ns/cgroup` -- how do you know that runC is inside the same cgroup namespace as the init in it's current PID namespace (that doesn't need to be true)? From what I can see you're going to skip joining cgroups people explicitly asked you to.

YummyPeng: Originally, I added this "skipping" logic because I thought a container's cgroups should be bind-mounted if it shares cgroup ns with host, but now it seems that I have went the wrong way. :disappointed_relieved:

So, you mean that it makes no difference whether a container shares cgroup ns with host or not? And, we also need to mount new cgroups (not bind-mount) in a container which shares cgroup ns with host?

libcontainer/nsenter/nsexec.c
YummyPeng: Yeah, as far as I can see, `unshare(CLONE_NEWCGROUP)` must be done in the final child process. Or if you have any better suggestions? :grin:

cyphar: Well, you /could/ do it by passing two different PIDs to the parent. But that's just a bad idea.

This is fine, don't worry about my gripes.

cyphar:
>> @@ -190,6 +195,17 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
>  		if config.Namespaces.Contains(t) {
>  			return nil, fmt.Errorf("malformed spec file: duplicated ns %q", ns)
>  		}
> +		// if this container is going to have shared cgroup namespace with host,
> +		// we just skip the following .Add()
>
> Originally, I added this "skipping" logic because I thought a
> container's cgroups should be bind-mounted if it shares cgroup ns with
> host, but now it seems that I have went the wrong way. 😥
>
> So, you mean that it makes no difference whether a container shares
> cgroup ns with host or not? And, we also need to mount new cgroups (not
> bind-mount) in a container which shares cgroup ns with host?

When you do a setns(2) to a cgroup namespace, you're going to share 
their namespace (meaning that the scoping and mounting rules will be 
shared with you). In order to setns(2) to a namespace you need to have 
certain privileges (ns_capable(CAP_SYS_ADMIN) in the user namespace of 
the cgroup namespace) -- which is the same privilege requirement as 
mounting the cgroupfs filesystem.

So if setns(2) doesn't fail then mount -t cgroup shouldn't fail either 
AFAICS. Thus this logic isn't necessary.

So the only time we should be bindmounting is when we haven't created a 
cgroupns and we aren't joining a cgroupns. Which is already being 
handled by the other code you wrote (.Contains is all you need).

I could actually check the kernel code if it made you feel better (the 
above is from memory the last time I checked how the code worked).

P.S.: GitHub is horrible, I can't load your comment in the web-view so I 
can't comment on it that way (and replying to notification emails in a 
review doesn't add the email to the right discussion -- EUGH!).

-- 
Aleksa Sarai
Software Engineer (Containers)
SUSE Linux GmbH
https://www.cyphar.com/

YummyPeng:
> So the only time we should be bindmounting is when we haven't created a
cgroupns and we aren't joining a cgroupns. Which is already being
handled by the other code you wrote (.Contains is all you need).

If I run a container with cgroup ns path set as `/proc/1/ns/cgroup`(shares cgroup ns with host) and mount new cgroups in it. Then its `/proc/self/cgroup` file would be like this: (as expected)
```console
root@ubuntu:~/tools# runc run test
root@runc:/# cat /proc/self/cgroup
13:name=systemd:/user/0.user/9.session/test
12:pids:/test
11:hugetlb:/user/0.user/9.session/test
10:net_prio:/user/0.user/9.session/test
9:perf_event:/user/0.user/9.session/test
8:net_cls:/user/0.user/9.session/test
7:freezer:/user/0.user/9.session/test
6:devices:/user/0.user/9.session/test
5:memory:/user/0.user/9.session/test
4:blkio:/user/0.user/9.session/test
3:cpuacct:/user/0.user/9.session/test
2:cpu:/user/0.user/9.session/test
1:cpuset:/test
```

And since its cgroup scoping is shared with host, so its `/sys/fs/cgroup` directory is the same as host too.
For example, we can see that our `/sys/fs/cgroup/pids` directory contains these files and subdirectories:
```console
root@runc:~# ls /sys/fs/cgroup/pids/
cgroup.clone_children  cgroup.procs  cgroup.sane_behavior  docker  notify_on_release  release_agent  tasks  test
```

Here, `docker` is a subdirectory owned by host, but process inside this container can see it as well. Will it cause any problems related with security?

cyphar:
@YummyPeng

I feel like if someone intentionally decides to `setns` to the host's cgroup namespace then they sort of deserve what they get. Though the other @opencontainers/runc-maintainers might have a different opinion.

But even if we decide that we want to always make `/sys/fs/cgroup` the cgroup the container is in, then there's a bunch of issues with `setns(...)` that aren't really easy to solve -- it's not just an issue with joining PID 1's cgroup namespace. Your `cgroup` mount could be completely different from your actual cgroup association. And we can't move the process because `cgroupPath` might be set to something outside of the `cgroup` root for the namespace we're joining.

I feel like the cgroup namespace `setns(...)` semantics probably require a bit more thought overall. The decision has to be consistent between all of the different cases you can have when you're joining a namespace.

YummyPeng:
@cyphar  Once again updated! I just removed the whole "skipping" logic this time.

hqhq:
@yummypeng If it's ready for review, you can remove `[WIP]` prefix.

yummypeng:
Sorry for late update. 

ping @opencontainers/runc-maintainers    Any comments? Thank you : )

cyphar:
Yikes, this has been around for a while. I can re-carry this PR for @yummypeng.

libcontainer/configs/namespaces_syscall.go
```diff
@@ -8,13 +8,17 @@ func (n *Namespace) Syscall() int {
 	return namespaceInfo[n.Type]
 }
 
+// This is not yet in the Go stdlib.
+const syscall_CLONE_NEWCGROUP = (1 << 25)
```
sargun: This has been added to the unix library.

yummypeng:
Thanks @cyphar :stuck_out_tongue_winking_eye:

crosbymichael:
I'm rebasing and caring this :)

cyphar:
Heh, thanks. I started rebasing it and it was pretty ugly from memory.

PR #1185, opened 2016-11-10
Allow update any resources in `spes.Resources`

WeiZhang555:
The command `runc update -r file/-` accepts input from file or standard
input in JSON format, before this commit, runc can only update limited
fields in `specs.Resources`, this commit will enhance the update command
to make updating any field of `specs.Resources` possible.

The list of Cgroups resources newly supported to update includes:
* devices
* pids
* blockIO
* hugepageLimits
* network

In one word, this commit enables updating all the fields in
`spec.Resources` except `disableOOMKiller` and `oomScoreAdj`.

Note:
1. the enhanced fields can only be passed to runc via `update -r`
flag, this doesn't add extra flags for each resource, because it's a
long list and `-r` suffices and is easier for maintainance than endless
new flags.
2. This commit removes example in command help message because it takes
large space, instead this added a richer example in man page.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -208,7 +208,7 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	if err := setupUserNamespace(spec, config); err != nil {
 		return nil, err
 	}
-	c, err := createCgroupConfig(opts.CgroupName, opts.UseSystemdCgroup, spec)
+	c, err := CreateCgroupConfig(opts.CgroupName, opts.UseSystemdCgroup, spec)
 	if err != nil {
 		return nil, err
 	}
@@ -255,7 +255,8 @@ func createLibcontainerMount(cwd string, m specs.Mount) *configs.Mount {
 	}
 }
 
-func createCgroupConfig(name string, useSystemdCgroup bool, spec *specs.Spec) (*configs.Cgroup, error) {
+// CreateCgroupConfig convert specs.Spec to config.Cgroups
+func CreateCgroupConfig(name string, useSystemdCgroup bool, spec *specs.Spec) (*configs.Cgroup, error) {
 	var myCgroupPath string
 
 	c := &configs.Cgroup{
--- a/libcontainer/specconv/spec_linux_test.go
+++ b/libcontainer/specconv/spec_linux_test.go
@@ -16,7 +16,7 @@ func TestLinuxCgroupsPathSpecified(t *testing.T) {
 		CgroupsPath: &cgroupsPath,
 	}
 
-	cgroup, err := createCgroupConfig("ContainerID", false, spec)
+	cgroup, err := CreateCgroupConfig("ContainerID", false, spec)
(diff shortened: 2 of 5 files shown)


WeiZhang555:
@cyphar I still can't understand why we do have the limit that `runc update` can only update some of the `specs.Resources` fields but not all, my thought is:

1. if we already support setting them while starting container, it should be reasonable to update them.
2. though `runc` is mainly used by `docker` currently, and docker doesn't have the needs for updating more fields, that doesn't mean other runtime(say `crio`?)  don't need them. As a very low level container runtime, it should provide enough capability, and higher level tools can decides whether to use it or not.
3. everything underneath for updating is ready, it takes us little effort to make this work. If user specify invalid value, `cgroupsManager.Set()` will report error; if user doesn't specify any or set a `zero` value for some field, then don't update it; If there's anything in `spec.Resources` that `runc update` don't recognize(`disableOOMKiller` and `oomScoreAdj`), ignore it.

WeiZhang555:
ping @crosbymichael @cyphar @hqhq @mrunalp What do you think of this ?

crosbymichael:
I think we did it this way because some cgroups you cannot update after processes are running in them.  There were also some that would update without error but would not do what you wanted if I remember right.

I'm fine with adding everything and letting cgroups error if the update wont work but there is still the risk of an update without error but the actual values won't reflect what you wanted.

WeiZhang555:
@crosbymichael To be honest, I'm not familiar with all details of Cgroups, my first though was, when you do an `runc update`, you are actually writing data directly into Cgroup files. So there are some situations:

1. first, user write some number into a Cgroup file but the number is invalid, then kernel and runc will report "write error: Invalid argument", it's clear and straightforward.
2. second, user write some number into a Cgroup file successfully, the data in file is updated correctly. Though user could get an unexpected result due to his poor understanding of Cgroup meanings, but runc can't do anything, right? `runc` is only responsible for writing specified data into Cgroup file without understanding the intention of user. If so, we should regard this as correct, and user should take care of his data themselves.
3.  next, user write some number(e.g. 1000) into a Cgroup file with an original value of 0, there's no error reported, but the value in this file is still 0 without proper update. I agree that this could bring some problem, because things are not operating as user expected. But is there really such situation? And even there is, I think this should be taken as a kernel bug, because it should report error.

So that's why I think updating all should be fine. 

I can understand your worries, but do you have a specific use case to illustrate the worst case?

crosbymichael:
@cyphar @hqhq what do you think about this?  There is still the issue of the write succeeding but the value not being set?

hqhq:
@crosbymichael Yes there is such issue, `runc update` calls `container.Set()` to update cgroup resources, while in `LinuxResources` there is a property (`OOMScoreAdj`) that doesn't belong to cgroup, so if user set that property, it won't update `OOMScoreAdj` and user won't get any error, that might be considered as a bug.

And the intention for this limitation was that we can keep consistency between `-r` and other single-field-options, I'm ok we loosen the limitation and give `-r` more options for updating resources, as long as we fix that above confusion it might cause.

Issue #487, opened 2016-01-20
Consistent GetStats behaviour for deleted cgroups


jimmidyson:
Some subsystems return `nil` error e.g. blkio, some return the `ErrNotExist` (or equivalent) e.g. cpuacct, & some return a reformatted error e.g. hugetlb.

What do you think the consistent behaviour should be?

I would prefer to return the `ErrNotExist` or similar error, as this is easier to handle with library consumers.

mrunalp:
@jimmidyson That sounds reasonable.

crosbymichael:
SGTM

PR #1190, opened 2016-11-17
return code changed to ErrNotExist

kunalkushwaha:
getBlkioStat was returning nil instead of ErrNotExist,
if any file in /sys/fs/cgroup/blkio/ was not found.

Correcting this error also, force to fix test cases,
which were not verifying if the files were really missing.

Signed-off-by: Kunal Kushwaha <kushwaha_kunal_v7@lab.ntt.co.jp>

### DIFF
--- a/libcontainer/cgroups/fs/blkio.go
+++ b/libcontainer/cgroups/fs/blkio.go
@@ -112,13 +112,18 @@ func splitBlkioStatLine(r rune) bool {
 	return r == ' ' || r == ':'
 }
 
+func isCFQEnabled(path string) bool {
+	// blkio.*_recursive are avaible only if kernel haver CFQ enabled and CFQ is io scheduler.
+	if _, err := getBlkioStat(filepath.Join(path, "blkio.io_serviced_recursive")); err == nil {
+		return true
+	}
+	return false
+}
+
 func getBlkioStat(path string) ([]cgroups.BlkioStatEntry, error) {
 	var blkioStats []cgroups.BlkioStatEntry
 	f, err := os.Open(path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return blkioStats, nil
-		}
 		return nil, err
 	}
 	defer f.Close()
@@ -160,13 +165,12 @@ func getBlkioStat(path string) ([]cgroups.BlkioStatEntry, error) {
 		}
 		blkioStats = append(blkioStats, cgroups.BlkioStatEntry{Major: major, Minor: minor, Op: op, Value: v})
 	}
-
 	return blkioStats, nil
 }
 
 func (s *BlkioGroup) GetStats(path string, stats *cgroups.Stats) error {
 	// Try to read CFQ stats available on all CFQ enabled kernels first
-	if blkioStats, err := getBlkioStat(filepath.Join(path, "blkio.io_serviced_recursive")); err == nil && blkioStats != nil {
+	if isCFQEnabled(path) {
 		return getCFQStats(path, stats)
 	}
 	return getStats(path, stats) // Use generic stats as fallback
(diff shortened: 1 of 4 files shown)


kunalkushwaha:
This PR is part of fix for https://github.com/opencontainers/runc/issues/487

hqhq:
Rejected.

This is not right, some files' absence is totally legal, such as different kernel configurations, for blkio, some files could only be seem on CFQ enabled kernels, so you can't just return error for not finding some files. Some subsystems don't have this issue, so we are having this inconsistency issue.

This PR doesn't look good to me, but I don't deny there can be some improvements to soften this issue, so #486 could still be valid. You'll be welcome to keep digging on this, and thanks for your contribution :) .

kunalkushwaha:
> some files' absence is totally legal, such as different kernel configurations, for blkio, some
files could only be seem on CFQ enabled kernels, so you can't just return error for not finding some files. Some subsystems don't have this issue, so we are having this inconsistency issue.

I Agree. The [GetStats()](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/blkio.go#L167) function already take cares about CFQ enabled/disabled configuration.

My fix simply checks, if CFQ enabled, then in case some files are missing, it reports ``ErrNotExist`` . Modification is done in ``getBlkioStat()`` which is not exposed outside this package.

hqhq:
> I Agree. The GetStats() function already take cares about CFQ enabled/disabled configuration.

Your PR will break CFQ enabled/disabled check. Fix that could make this PR reasonable.

kunalkushwaha:
@hqhq I have added a new function ``isCFQEnabled()`` to check if *_recursive attributes are present.
Please let me know, if you find this solution better?

Also, I found, in [``BlkioStats``](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/stats.go#L72) there were no entries for ``blkio.throttle.io_service_bytes`` and ``blkio.throttle.io_serviced`` which were making two test invalid. I also fixed that in this PR.

libcontainer/cgroups/stats.go
```diff
@@ -79,6 +79,8 @@ type BlkioStats struct {
 	IoMergedRecursive       []BlkioStatEntry `json:"io_merged_recursive,omitempty"`
 	IoTimeRecursive         []BlkioStatEntry `json:"io_time_recursive,omitempty"`
 	SectorsRecursive        []BlkioStatEntry `json:"sectors_recursive,omitempty"`
+	ThrottleServiceBytes    []BlkioStatEntry `json:"io_service_bytes,omitempty"`
```
hqhq: I'd prefer the name to be `ThrottleIoServiceBytes` (and I'd expect we have a follow up PR to s/Io/IO for all blkio entry names to keep consistency with other places) and json name to be `throttle_io_service_bytes`, and I think adding these two new fields should be in a separated commit.

Same with `ThrottleServiced`.

libcontainer/cgroups/fs/blkio.go
```diff
@@ -226,12 +230,15 @@ func getStats(path string, stats *cgroups.Stats) error {
 	if blkioStats, err = getBlkioStat(filepath.Join(path, "blkio.throttle.io_service_bytes")); err != nil {
 		return err
 	}
-	stats.BlkioStats.IoServiceBytesRecursive = blkioStats
+	//	stats.BlkioStats.IoServiceBytesRecursive = blkioStats
```
hqhq: You can just delete this.

libcontainer/cgroups/fs/blkio.go
```diff
@@ -226,12 +230,15 @@ func getStats(path string, stats *cgroups.Stats) error {
...
+	stats.BlkioStats.ThrottleServiceBytes = blkioStats
 
 	if blkioStats, err = getBlkioStat(filepath.Join(path, "blkio.throttle.io_serviced")); err != nil {
 		return err
 	}
-	stats.BlkioStats.IoServicedRecursive = blkioStats
+	//	stats.BlkioStats.IoServicedRecursive = blkioStats
```
hqhq: And this.

libcontainer/cgroups/fs/blkio.go
hqhq: After a second thought, I think we might reuse `IoServiceBytesRecursive` for some reasons, it's been there since the old libcontainer days, maybe upper level monitor tools don't have to care if bandwidth or throttle is used by blkio. And your change will break these tools.

hqhq: I digged a bit and found the PR added this code: https://github.com/docker/libcontainer/pull/167

Looks like it's intentional.

kunalkushwaha: Actually, while testing my code I realized testcase [``TestNonCFQBlkioStats``](https://github.com/docker/libcontainer/pull/167/files#diff-f734435c91f3ff99263df9f4577a3e15R208) doesn’t executes for non-CFQ condition as ``blkio.io_serviced_recursive`` attribute is present. As result, It always executes CFQ case only.

If some tools depends on ``IoServiceBytesRecursive`` and do not honour ``ThrottleIoServiceBytes`` , I can re-test the results with them.

libcontainer/cgroups/stats.go
kunalkushwaha: Thanks for review comments. Will Incorporate this.

crosbymichael:
Feel free to update and reopen if you still think this PR should get in and you have time. 

Thanks!

PR #1192, opened 2016-11-17
Add shell formatting via shfmt

mvdan:


Dockerfile
```diff
@@ -36,6 +36,10 @@ RUN mkdir -p /usr/src/criu \
     && cd /usr/src/criu \
     && make install-criu
 
+# install shfmt
+RUN go get -u github.com/mvdan/sh/cmd/shfmt \
```
cyphar: I personally don't like `go get`. Can you instead clone a specific commit of your repo, build and install it? Otherwise we can't be sure what version of `shfmt` we end up using.

mvdan: Sure. I was going to do that but I thought this would be simpler.

mvdan: Done. Note that I also publish binaries, but I assume you prefer source.

cyphar: You assumed correctly. :wink:

cyphar:
LGTM.

hqhq:
LGTM

@mvdan Thanks.

cyphar:
Now time to rebase all of the things. :wink:

PR #1196, opened 2016-11-19
Fix leftover cgroup directory issue

hqhq:
In the cases that we got failure on a subsystem's Apply,
we'll get some subsystems' cgroup directories leftover.

On Docker's point of view, start a container failed, use
`docker rm` to remove the container, but some cgroup files
are leftover.

Sometimes we don't want to clean everyting up when something
went wrong, because we need these inter situation
information to debug what's going on, but cgroup directories
are not useful information we want to keep.

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -114,8 +114,8 @@ func (m *Manager) Apply(pid int) (err error) {
 		return err
 	}
 
+	m.Paths = make(map[string]string)
 	if c.Paths != nil {
-		paths := make(map[string]string)
 		for name, path := range c.Paths {
 			_, err := d.path(name)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -148,6 +145,11 @@ func (m *Manager) Apply(pid int) (err error) {
 			return err
 		}
 		paths[sys.Name()] = p
+
+		if err := sys.Apply(d); err != nil {
+			m.Paths = paths
```
cyphar: If we want to do it this way, where we change m.Paths gradually, why not just change the above `paths[sys.Name()]` to `m.Paths[sys.Name()]`? Or has that map been used by callers so we have to replace the reference?

hqhq: Because `m.Paths` was initialized as `nil` when we create the container, but yes we can initialize `m.Paths` and change it directly, I've mended the commit.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -124,17 +124,13 @@ func (m *Manager) Apply(pid int) (err error) {
...
 			}
-			paths[name] = path
+			m.Paths[name] = path
 		}
-		m.Paths = paths
 		return cgroups.EnterPid(m.Paths, pid)
 	}
 
-	paths := make(map[string]string)
+	m.Paths = make(map[string]string)
```
YummyPeng: Why not just put this line of code before the above `if` so that the duplicated line inside the above `if` can be omitted?

hqhq: Done, thanks.

hqhq:
ping @cyphar @mrunalp @crosbymichael @dqminh

cyphar:
LGTM. This actually will make some of the rootless cgroup manager changes in #774 easier to do.

cyphar:
/me will test this right now. This actually will make some of the rootless cgroup manager changes in #774 easier to do.

cyphar:
LGTM.

crosbymichael:
LGTM

Issue #433, opened 2015-12-11
Proposal: Intel RDT/CAT support for OCI/runc and Docker


xiaochenshen:
```
The descriptions of Intel RDT/CAT features, user cases and Linux kernel interface are
heavily based on the Intel RDT documentation of the Linux kernel:

https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt

Thanks to the authors of the kernel patches:
* Vikas Shivappa <vikas.shivappa@linux.intel.com>
* Fenghua Yu <fenghua.yu@intel.com>
* Tony Luck <tony.luck@intel.com>
```
#### 1. Intel RDT/CAT support in OCI/runtime-spec
https://github.com/opencontainers/runtime-spec/pull/630
https://github.com/opencontainers/runtime-spec/pull/787
https://github.com/opencontainers/runtime-spec/pull/889

#### 2. Intel RDT/CAT support in OCI/runc
https://github.com/opencontainers/runc/pull/1279
https://github.com/opencontainers/runc/pull/1589
https://github.com/opencontainers/runc/pull/1590
https://github.com/opencontainers/runc/pull/1615

#### 5. Intel RDT/CAT support in Docker CLI
---

## What is Intel RDT and CAT:

Intel Cache Allocation Technology (CAT) is a sub-feature of Resource Director Technology (RDT). Currently L3 Cache is the only resource that is supported in RDT.

Cache Allocation Technology offers the capability of L3 cache Quality of Service (QoS). It provides a way for the Software (OS/VMM/Container) to restrict cache allocation to a defined 'subset' of cache which may be overlapping with other 'subsets'. This feature is used when allocating a line in cache i.e. when pulling new data into the cache. The programming of the h/w is done via PQR MSRs.

The different cache subsets are identified by CLOS (class of service) identifier and each CLOS has a CBM (cache bit mask). The CBM is a contiguous set of bits which defines the amount of cache resource that is available for each 'subset'.

More information can be found in the section 17.17 of [Intel Software Developer Manual] and
[Intel RDT/CAT introduction].

[Intel Software Developer Manual]: http://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-manual-325462.pdf

[Intel RDT/CAT introduction]: https://software.intel.com/en-us/articles/introduction-to-cache-allocation-technology

## Supported Intel Xeon CPU SKUs:
- Intel(R) Xeon(R) processor E5 v4 and newer generations
[...]

PR #447, opened 2015-12-24
libcontainer: cgroups: add intel_rdt support in runc

xiaochenshen:
This PR fixes issue #433 (Proposal: Intel RDT/CAT cgroup support in runc/libcontainer)

**Patch v2:**
Rebased the code for "unified config file" according to #284.
Commit 273cbbbd083d6cfb4b73568bd1232f2e7d7a7043: is to update specs for compiling, it is **_NOT_** necessary for this pull request
when https://github.com/opencontainers/specs/pull/267 is merged.

Commit 9808367f2473687c5210c196226c3e3a712bd3d9:
This PR fixes issue #433

**Patch v1:**
This version is for **code review only** because kernel patch is not upstream yet and the
specs change is under discussion in https://github.com/opencontainers/specs/pull/267

Commit b3282581e7252b2b08cb0305bbdd7e03b936f40b: is to update specs for compiling, it is **_NOT_** necessary for this pull request
when https://github.com/opencontainers/specs/pull/267 is merged. I am not sure if this will break Jenkins building.

Commit febaf8225b49019248c849bb0a16b07bd31cd478:
This PR fixes issue #433 

**About Intel RDT/CAT feature:**
Intel platforms with new Xeon CPU support Resource Director Technology (RDT).
Intel Cache Allocation Technology (CAT) is a sub-feature of RDT. Currently L3
Cache is the only resource that is supported in RDT.

This feature provides a way for the software to restrict cache allocation to a
defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
The different subsets are identified by class of service (CLOS) and each CLOS
has a capacity bitmask (CBM).

More information can be found in the section 17.16 of Intel Software Developer
Manual.

**About intel_rdt cgroup:**
Linux kernel 4.6 (or later) will introduce new cgroup subsystem 'intel_rdt'
with kernel config CONFIG_INTEL_RDT.

The 'intel_rdt' cgroup manages L3 cache allocation. It has a file 'l3_cbm'
which represents the L3 cache capacity bitmask (CBM). The CBM needs to have
only _contiguous bits set_ and number of bits that can be set is less than the
max bits. The max bits in the CBM is varied among supported Intel platforms.
The tasks belonging to a cgroup get to fill in the L3 cache represented by
the CBM. For example, if the max bits in the CBM is 10 and the L3 cache size
is 10MB, each bit represents 1MB of the L3 cache capacity.

Root cgroup always has all the bits set in the l3_cbm. User can create more
[...]

PR #1198, opened 2016-11-21
libcontainer: add support for Intel RDT/CAT in runc

xiaochenshen:
This PR fixes issue #433 

About Intel RDT/CAT feature:
Intel platforms with new Xeon CPU support Intel Resource Director Technology
(RDT). Cache Allocation Technology (CAT) is a sub-feature of RDT, which
currently supports L3 cache resource allocation.

This feature provides a way for the software to restrict cache allocation to a
defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
The different subsets are identified by class of service (CLOS) and each CLOS
has a capacity bitmask (CBM).

For more information about Intel RDT/CAT can be found in the section 17.17
of Intel Software Developer Manual.

About Intel RDT/CAT kernel interface:
In Linux kernel, the interface is defined and exposed via "resource control"
filesystem, which is a "cgroup-like" interface.

Comparing with cgroups, it has similar process management lifecycle and
interfaces in a container. But unlike cgroups' hierarchy, it has single level
filesystem layout.

Intel RDT "resource control" filesystem hierarchy:
```
mount -t resctrl resctrl /sys/fs/resctrl
tree /sys/fs/resctrl
/sys/fs/resctrl/
|-- info
|   |-- L3
|       |-- cbm_mask
|       |-- num_closids
|-- cpus
|-- schemata
|-- tasks
|-- <container_id>
    |-- cpus
    |-- schemata
    |-- tasks
```
For runc, we can make use of `tasks` and `schemata` configuration for L3 cache
resource constraints.

The file `tasks` has a list of tasks that belongs to this group (e.g.,
<container_id>" group). Tasks can be added to a group by writing the task ID
to the "tasks" file  (which will automatically remove them from the previous
group to which they belonged). New tasks created by fork(2) and clone(2) are
added to the same group as their parent. If a pid is not in any sub group, it
Is in root group.

The file `schemata` has allocation bitmasks/values for L3 cache on each socket,
which contains L3 cache id and capacity bitmask (CBM).
```
	Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
```
For example, on a two-socket machine, L3's schema line could be `L3:0=ff;1=c0`
which means L3 cache id 0's CBM is 0xff, and L3 cache id 1's CBM is 0xc0.

The valid L3 cache CBM is a *contiguous bits set* and number of bits that can
be set is less than the max bit. The max bits in the CBM is varied among
supported Intel Xeon platforms. In Intel RDT "resource control" filesystem
layout, the CBM in a group should be a subset of the CBM in root. Kernel will
check if it is valid when writing. e.g., 0xfffff in root indicates the max bits
of CBM is 20 bits, which mapping to entire L3 cache capacity. Some valid CBM
values to set in a group: 0xf, 0xf0, 0x3ff, 0x1f00 and etc.

For more information about Intel RDT/CAT kernel interface:  
https://git.kernel.org/cgit/linux/kernel/git/tip/tip.git/commit/?h=x86/cache&id=f20e57892806ad244eaec7a7ae365e78fee53377

An example for runc:
```
There are two L3 caches in the two-socket machine, the default CBM is 0xfffff
and the max CBM length is 20 bits. This configuration assigns 4/5 of L3 cache
id 0 and the whole L3 cache id 1 for the container:

"linux": {
	"resources": {
		"intelRdt": {
			"l3CacheSchema": "L3:0=ffff0;1=fffff"
		}
	}
}
```

### DIFF
--- a/Godeps/_workspace/src/github.com/opencontainers/runtime-spec/specs-go/config.go
+++ b/Godeps/_workspace/src/github.com/opencontainers/runtime-spec/specs-go/config.go
@@ -311,6 +311,13 @@ type Network struct {
 	Priorities []InterfacePriority `json:"priorities,omitempty"`
 }
 
+// IntelRdt for Linux Intel RDT/CAT resource management (Linux 4.10)
+type IntelRdt struct {
+	// The schema for L3 cache id and capacity bitmask (CBM)
+	// Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
+	L3CacheSchema *string `json:"l3CacheSchema,omitempty"`
+}
+
 // Resources has container runtime resource constraints
 type Resources struct {
 	// Devices configures the device whitelist.
@@ -331,6 +338,8 @@ type Resources struct {
 	HugepageLimits []HugepageLimit `json:"hugepageLimits,omitempty"`
 	// Network restriction configuration
 	Network *Network `json:"network,omitempty"`
+	// IntelRdt restriction configuration
+	IntelRdt *IntelRdt `json:"intelRdt,omitempty"`
 }
 
 // Device represents the mknod information for a Linux special device file
--- a/events.go
+++ b/events.go
@@ -24,11 +24,12 @@ type event struct {
 
 // stats is the runc specific stats structure for stability when encoding and decoding stats.
 type stats struct {
-	Cpu     cpu                `json:"cpu"`
-	Memory  memory             `json:"memory"`
-	Pids    pids               `json:"pids"`
-	Blkio   blkio              `json:"blkio"`
-	Hugetlb map[string]hugetlb `json:"hugetlb"`
+	Cpu      cpu                `json:"cpu"`
+	Memory   memory             `json:"memory"`
+	Pids     pids               `json:"pids"`
+	Blkio    blkio              `json:"blkio"`
+	Hugetlb  map[string]hugetlb `json:"hugetlb"`
+	IntelRdt intelRdt           `json:"intelRdt"`
 }
 
 type hugetlb struct {
@@ -93,6 +94,12 @@ type memory struct {
 	Kernel    memoryEntry       `json:"kernel,omitempty"`
 	KernelTCP memoryEntry       `json:"kernelTCP,omitempty"`
 	Raw       map[string]uint64 `json:"raw,omitempty"`
+}
+
+type intelRdt struct {
+	// The read-only default "schemas" in root, for reference
+	L3CacheSchemaRoot string `json:"l3CacheSchemaRoot,omitempty"`
+	L3CacheSchema     string `json:"l3CacheSchema,omitempty"`
 }
 
 var eventsCommand = cli.Command{
@@ -223,6 +230,10 @@ func convertLibcontainerStats(ls *libcontainer.Stats) *stats {
 	for k, v := range cg.HugetlbStats {
 		s.Hugetlb[k] = convertHugtlb(v)
 	}
+
+	is := cg.IntelRdtStats
+	s.IntelRdt.L3CacheSchemaRoot = is.IntelRdtRootStats.L3CacheSchema
+	s.IntelRdt.L3CacheSchema = is.IntelRdtGroupStats.L3CacheSchema
 	return &s
 }
 
--- a/libcontainer/SPEC.md
+++ b/libcontainer/SPEC.md
@@ -154,6 +154,93 @@ that no processes or threads escape the cgroups.  This sync is
 done via a pipe ( specified in the runtime section below ) that the container's
 init process will block waiting for the parent to finish setup.
 
+**intelRdt**:  
+Intel platforms with new Xeon CPU support Intel Resource Director Technology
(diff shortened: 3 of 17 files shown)


xiaochenshen:
This PR will obsolete #447  for the change of Intel RDT/CAT kernel interface.
Please find the design in https://github.com/opencontainers/runc/issues/433

crosbymichael:
If this is not a cgroup why are you modeling it after that?  Wouldn't it be much simpler to not try to hide this behind the cgroups interface and just have a simple function call to enable this and get the stats?

xiaochenshen:
@crosbymichael 
There are several reasons. Hope these can address your concerns. Thank you!

1. By nature, Intel RDT/CAT feature is really a "hardware control group" for L3 cache in Intel platform. Can you remember that we had a face-to-face talk in DockerCon Seattle? :relaxed: The original patch for Linux kernel was intel_rdt cgroup interface (you can find the history story in https://github.com/opencontainers/runc/issues/433), but the cgroup patch is rejected by kernel cgroup maintainer for some reasons, like incompatibility and corner cases. The new kernel interface - "resource control" filesystem, which is a "cgroup-like" interface, bypasses kernel cgroup path, but implements the very similar functionality (e.g., process management lifecycle, mkdir/rmdir group filesystem operations) in kernel. And what's more, the kernel issues are not existed in libcontainer cgroup framework.

   From my understanding, the resource constraints in libcontainer is managed by single cgroupManager. We can use clean and simple interfaces like Apply(), Set(), GetStats() for processes operations. In my opinion, the resource constraints framework is still applicable for Intel RDT/CAT.

   You can find more information (e.g., kernel patch, user interface, history story, old/new designs) about Intel RDT/CAT in https://github.com/opencontainers/runc/issues/433

2. Design and implementation consideration:
   Actually, I have implemented two solutions for https://github.com/opencontainers/runc/issues/433 prior to submitting this PR:

   (1) Adding package intelrdt and new resource manager (intelRdtManager) as a new infrastructure in libcontainer. It implements Manager interface which is similar to cgroup manager framework.

   This is the previous version of design which I haven't submitted PR, here are the workable patches in another branch for your reference: https://github.com/xiaochenshen/runc/commits/rdt-cat-resctrl-v1

   In https://github.com/opencontainers/runc/issues/433#issuecomment-241212914, @cyphar doesn't like the design ("Please don't make it so that we have to track an intel-specific feature within the generic code for libcontainer."). I agree with @cyphar that creating a new resource manager for Intel RDT/CAT means some Intel-specific codes scattering in libcontainer. It is not graceful even though I am working for Intel :innocent:

   From my understanding, @cyphar doesn't object to consolidate Intel RDT/CAT feature with cgroupManager ("If we can't consolidate it with the cgroup code..."). @cyphar also suggested a solution to have generic Manager to consolidate cgroup and other resource management including Intel RDT/CAT. I have tried, but I found the effort is huge and risky, it is hard to refactor the current state of affairs because cgroup codes are very tight-coupling.

   (2) (i.e. This PR) Making use of cgroupManager to consolidate Intel RDT "resource control" filesystem.

   In this design, we implement Intel RDT/CAT as a cgroup subsystem - IntelRdtGroup. Both in the functionality and unit tests change, this introduces minimal impact to libcontainer generic codes, and small impact to other libcontainer/cgroup codes. In addition, we can dynamically add IntelRdtGroup in cgroup subsystemSet only when Intel RDT/CAT is enabled by hardware and kernel.

   At present, personally I prefer the second design even though it is not a real "cgroup" in kernel. With this, we can keep libcontainer resource constraints codes clean and simple.

cyphar:
@xiaochenshen 

> Design and implementation consideration:

Understand that when I made those comments they were in response to a "design proposal". Personally, it's quite hard for me to reason about a patch if you're going to talk about how you're going to write it. I'm reading through this PR at the moment and I'm thinking that maybe this might be uglier than if it was handled outside of the cgroup code (all of the special casing around `== "intel_rdt"` is an example of what I was talking about). It probably will be nicer to have another manager for the resource control filesystem.

xiaochenshen:
@cyphar 
> I'm reading through this PR at the moment and I'm thinking that maybe this might be uglier than if it was handled outside of the cgroup code (all of the special casing around == "intel_rdt" is an example of what I was talking about). It probably will be nicer to have another manager for the resource control filesystem.

I have opened a new PR https://github.com/opencontainers/runc/pull/1279 to address this.

As you suggested in https://github.com/opencontainers/runc/issues/433#issuecomment-241212914:

It adds a new "ResourceManager" structure as the base interface for all resource managers, including cgroups manager and incoming IntelRdt manager.

All registered resource managers are consolidated in linuxContainer structure. We can apply to unified operations (e.g., Apply(), Set(), Destroy()) using all of the registered resource managers.

justincormack:
@xiaochenshen do you want to close this one now the other is open?

xiaochenshen:
@justincormack #1279 obsolete this PR. We can close this one. Thank you.

caniszczyk:
closing in favor of https://github.com/opencontainers/runc/pull/1279

PR #1200, opened 2016-11-22
Don't format string if it's necessary

likel:
If the format of ```fmt.Errorf``` is unnecessary, replace it with ```errors.New``` because ```fmt.Errorf``` calls ```errors.New``` internally after the string been formated.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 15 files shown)


cyphar:
I prefer using `fmt.Errorf` because it means we use the same thing to return errors everywhere. In particular I don't like the `errors.New` API (IMO it should've always been a part of `fmt` or `Errorf` should've been part of `errors.*`). But I'm not /really/ against this, I just don't really see the benefit.

hqhq:
I don't like this either, don't see the benefit and it breaks consistency. I'll close this for now, feel free to comments afterwards.

PR #724, opened 2016-04-06
HookState adhears to OCI

glestaris:
* Adds `BundlePath` field in `HookState`
* Renames JSON field for `HookState.Version` to `ociVersion`
* Promotes private `searchLabels` function to `libcontainer/utils`

Signed-off-by: George Lestaris <glestaris@pivotal.io>
Signed-off-by: Ed King <eking@pivotal.io>

/cc @teddyking @cloudfoundry/cf-garden

PR #1201, opened 2016-11-28
Sync HookState struct with OCI spec

WeiZhang555:
<del>* modify json name of `ociVersion` to `version`
<del>* Add missing `Status` and `Annotations` field

Edit:

`HookState` struct should follow definition of `State` in runtime-spec:
    
* Remove redundant `Rootfs` field as rootfs can be retrived from
 `bundlePath/config.json`

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
 type Rlimit struct {
@@ -243,13 +244,7 @@ func (hooks Hooks) MarshalJSON() ([]byte, error) {
 }
 
 // HookState is the payload provided to a hook on execution.
-type HookState struct {
-	Version    string `json:"ociVersion"`
-	ID         string `json:"id"`
-	Pid        int    `json:"pid"`
-	Root       string `json:"root"`
-	BundlePath string `json:"bundlePath"`
-}
+type HookState specs.State
 
 type Hook interface {
 	// Run executes the hook with the provided state.
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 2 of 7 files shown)


WeiZhang555:
I noticed that the json name of `Version` is changed from `version` to `ociVersion` in https://github.com/opencontainers/runc/pull/724 , according to current spec( [*State*](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/state.go) ), this `ociVersion` is wrong, and I didn't find any occurrence of `ociVersion` in git history of runtime spec.

This modification is a breakdown to backward compatibility but we have to do it, and the earlier the better. We should follow the name of the spec, this is important.

libcontainer/configs/config.go
```diff
@@ -244,11 +244,13 @@ func (hooks Hooks) MarshalJSON() ([]byte, error) {
...
-	Pid        int    `json:"pid"`
-	Root       string `json:"root"`
-	BundlePath string `json:"bundlePath"`
+	Version     string            `json:"version"`
+	ID          string            `json:"id"`
+	Status      string            `json:"status"`
+	Pid         int               `json:"pid"`
+	Root        string            `json:"root"`
+	BundlePath  string            `json:"bundlePath"`
+	Annotations map[string]string `json:"annotations"`
```
cyphar: Since you've added fields, can you please update the code to actually fill them? And also add some tests.

WeiZhang555: Yep, I'm already doing this now, was sending this for a first glance in case someone is against this.

I was just thinking that the `Status` field is not very useful, because `prestart`,`poststart`,`poststop` happens at special time point and we don't need the status to identify container status, so I am also modifying the spec and trying to sync up HookState struct between spec and runc. :-)

WeiZhang555:
I‘ve sent another PR to `runtime-spec`, for adding `Root` and removing `Status`. I'll update this once that's completed. :wink:

WeiZhang555:
This is ready for review, Janky is also good, ping @cyphar @hqhq @mrunalp @crosbymichael

hqhq:
LGTM

libcontainer/integration/exec_test.go
```diff
@@ -1048,25 +1049,44 @@ func TestHook(t *testing.T) {
...
 	rootfs, err := newRootfs()
 	ok(t, err)
 	defer remove(rootfs)
 
 	config := newTemplateConfig(rootfs)
-	expectedBundlePath := "/path/to/bundle/path"
+	expectedBundlePath := bundle
 	config.Labels = append(config.Labels, fmt.Sprintf("bundle=%s", expectedBundlePath))
+
+	getRootfsFromBundle := func(bundle string) (string, error) {
```
cyphar: Why not declare this as a function in `utils_test.go`. Seems a bit odd to define it like this.

WeiZhang555: Because this is a temporary function which is only used by `TestHook`, so I thought it's not worth a standalone function, and `utils_test.go` should contain common functions.  :-)

cyphar: Eh, okay. I don't really like it but it's just a test so w/e.

cyphar:
LGTM

Issue #1188, opened 2016-11-14
man pages missing arguments


rajasec:
when we use man pages for runc, options with angle brackets are coming up in man pages
NAME()                                                                                                                                             NAME()
       runc pause - pause suspends all processes inside the container
SYNOPSIS
       runc pause  [container-id...]
       Where "" is the name for the instance of the container to be paused.

DESCRIPTION
       The  pause  command  suspends  all processes in the instance of the container.  Use runc list to identiy instances of containers and their current
       status.

If you notice below help command, where the container ID is mentioned, it is missing in man pages.
For example:
runc pause help -h
NAME:
   runc pause - pause suspends all processes inside the container
USAGE:
   runc pause <container-id> [container-id...]
Where "<container-id>" is the name for the instance of the container to be
paused.
DESCRIPTION:
   The pause command suspends all processes in the instance of the container.

The issue applicable to all man pages for runc whenever we use angle brackets

hqhq:
It works normal in my box:
```
NAME()                                                                                                                                                                                                 NAME()

       runc pause - pause suspends all processes inside the container

SYNOPSIS
       runc pause <container-id> [container-id...]

       Where "<container-id>" is the name for the instance of the container to be paused.

DESCRIPTION
       The pause command suspends all processes in the instance of the container.  Use runc list to identiy instances of containers and their current status.

                                                                                                                                                                                                       NAME()
```

mrunalp:
Works for me, too.

cyphar:
In #1214 it turned out they were looking at the man pages in the markdown viewer, not the actual man pages.

PR #1214, opened 2016-12-04
fix typos for man page

sak0:
Hi, there is some missed escape sequence on man page.

Signed-off-by: CuiHaozhi <61755280@qq.com>

### DIFF
--- a/man/runc-create.8.md
+++ b/man/runc-create.8.md
@@ -4,7 +4,7 @@
 # SYNOPSIS
    runc create [command options] <container-id>
 
-Where "<container-id>" is your name for the instance of the container that you
+Where "\<container-id\>" is your name for the instance of the container that you
 are starting. The name you provide for the container instance must be unique on
 your host.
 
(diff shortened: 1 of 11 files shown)


man/runc-create.8.md
```diff
@@ -4,7 +4,7 @@
 # SYNOPSIS
    runc create [command options] <container-id>
 
-Where "<container-id>" is your name for the instance of the container that you
+Where "\<container-id\>" is your name for the instance of the container that you
```
sak0: Hi
when i access this page: 
https://github.com/opencontainers/runc/blob/master/man/runc-checkpoint.8.md

the page displayed:
runc checkpoint [command options]
Where "" is the name for the instance of the container to be checkpointed.

and not
runc checkpoint [command options]
Where "\<container-id\>" is the name for the instance of the container to be checkpointed.

so i think for Markdown format, '<' have to escaped to display.

rajasec:
@hqhq 
       Reported issue similar #1188

man/runc-create.8.md
sak0: ok, thanks. please close it :)

PR #1216, opened 2016-12-06
Fix thread safety of SelinuxEnabled and getSelinuxMountPoint

eparis:
Both suffered from different race conditions.

SelinuxEnabled assigned selinuxEnabledChecked before selinuxEnabled.
Thus racing callers could see the wrong selinuxEnabled.

getSelinuxMountPoint assigned selinuxfs to "" before it know the right
value. Thus racing could see "" improperly.

The gate selinuxfs, enabled, and mclist all on the same lock

### DIFF
--- a/libcontainer/selinux/selinux.go
+++ b/libcontainer/selinux/selinux.go
@@ -32,33 +32,73 @@ const (
 	stRdOnly         = 0x01
 )
 
+type selinuxState struct {
+	enabledSet   bool
+	enabled      bool
+	selinuxfsSet bool
+	selinuxfs    string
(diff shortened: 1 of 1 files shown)


eparis:
@rhatdan @pmorie @mrunalp

eparis:
https://play.golang.org/p/Ipe7Iql9EA is an example of the bug in `selinuxEnabled()`

mrunalp:
LGTM

rhatdan:
LGTM

cyphar:
LGTM

PR #1217, opened 2016-12-07
Don't add device to list if it doesn't exist anymore

mrunalp:
Fix a race where a device is first seen while walking /dev but has disappeared by the time we stat it. This is seen in environments where lots of privileged containers are launched/deleted rapidly with devicemapper as storage driver as lots of device mapper devices are created/deleted.

### DIFF
--- a/libcontainer/devices/devices_unix.go
+++ b/libcontainer/devices/devices_unix.go
@@ -94,6 +94,9 @@ func getDevices(path string) ([]*configs.Device, error) {
 			if err == ErrNotADevice {
 				continue
 			}
+			if os.IsNotExist(err) {
+				continue
+			}
 			return nil, err
 		}
 		out = append(out, device)


mrunalp:
@rhatdan @rhvgoyal ping

rhvgoyal:
@mrunalp In general this fix should solve the problem we are seeing where privileged container launch fails with "lstat /dev/foo: no such file or directory".

But in general, whatever list you are returning, there is no guarantee that device is present by the time list reaches back to the caller. So caller might have to do something similar and not assume that all the devices in the list are present.

mrunalp:
@rhvgoyal Agree that caller also needs to check. However we shouldn't error out here if we can.

libcontainer/devices/devices_unix.go
```diff
@@ -94,6 +94,9 @@ func getDevices(path string) ([]*configs.Device, error) {
 			if err == ErrNotADevice {
 				continue
 			}
+			if os.IsNotExist(err) {
```
rhatdan: This is fine, but I was writing a patch where 

```
func DeviceFromPath(path, permissions string) (*configs.Device, error) {
	fileInfo, err := osLstat(path)
	if err != nil {
                if os.IsNotExist(err) {
                     return nil, ErrNotADevice
               }
		return nil, err
}
```

rhvgoyal: @rhatdan, If you return error if device is not present, not sure what will caller do.

I think this whole function is racy. Caller passes in a  list of devices, and caller has not opened the devices in there. That means devices can disappear any time.

I think we should put a comment that this function goes through a list of devices and if some devices are not present, it ignores them.

Returning error if some device is not present, will not help.

rhatdan: If the path does not exist, it is clearly not a device.  As you say this is a racy situation, and it should anticpate the fact that devices can dissapear.  I believe the goal of the function is to list define the devices that are still there.  But of course the device could disappear even after it was defined as a device.

Either way I am fine with it. But version prevents other uses potentially hitting this error and not checking for ENOEXIST

cyphar: @rhvgoyal 

> I think we should put a comment that this function goes through a list of devices and if some devices are not present, it ignores them.

Ignoring is worse than returning `ErrNotADevice`. It is true that a path that doesn't exist is not a device. :P

rhvgoyal: Ok, I read it little closely and I am fine with your patch. getDevices() already ignores the ErrNotADevice error and continues and that should work.

mrunalp: I'll make a case for my patch. A lower level function can return an err as it sees it but a higher level function can translate/ignore it as it sees fit. `DeviceFromPath` could be used in other contexts where it might be useful to know what exact err it actually saw.

rhatdan: I am fine with your patch, I just wanted to suggest an alternative.  Either way we need to fix this race.

hqhq:
LGTM

mrunalp:

cyphar:
LGTM.

PR #1219, opened 2016-12-08
Windows: Back to compiling

jhowardmsft:
Signed-off-by: John Howard <jhoward@microsoft.com>

Get runc.exe compiling on Windows again. It remains totally non-functional though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ contrib/cmd/recvtty/recvtty
 Godeps/_workspace/src/github.com/opencontainers/runc
 man/man8
 release
+*.exe
+*.exe~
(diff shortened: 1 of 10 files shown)


hqhq:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1219/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

mrunalp:
LGTM

crosbymichael:
@jhowardmsft can you rebase this then we can merge, sorry for the delay

justincormack:
ping @jhowardmsft

crosbymichael:
Unless windows i going to use runc then we shouldn't have to make sure that it compiles on a platform that will never be used.  From what I understand, this is not the case and windows is not going to use a runtime like runc so this should be be needed.

PR #1220, opened 2016-12-09
Add recvtty to install path

rajasec:
install & uninsnstall recvtty from /usr/local/sbin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,9 @@ shell: all
 install:
 	install -D -m0755 runc $(BINDIR)/runc
 
+install-recvtty:
+	install -D -m0755 contrib/cmd/recvtty/recvtty $(BINDIR)/recvtty
+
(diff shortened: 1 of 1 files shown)


cyphar:
Rejected. We discussed this in #1018. `recvtty` is more of a testing tool (and example tool) than something that users are actually meant to use.

rajasec:
@cyphar 
Though it is used to create the console sockets, can we use other options to create the sockets instead of recvtty ?

cyphar:
@rajasec Yes. The plan is to make the console socket API part of the spec, and that other people will implement more "enterprise" implementations -- `recvtty` is just used for our own integration tests (and also for messing around inside a shell).

rajasec:
@cyphar 
Thanks for clarification, I'm closing the PR now.

PR #1221, opened 2016-12-10
validate: Check that the given namespace path is a symlink

sameo:
When checking if the provided networking namespace is the host
one or not, we should first check if it's a symbolic link or not
as in some cases we can use persistent networking namespace under
e.g. /var/run/netns/.

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -148,6 +148,15 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 	return nil
 }
 
+func isSymbolicLink(path string) (bool, error) {
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return false, err
+	}
+
+	return fi.Mode()&os.ModeSymlink == os.ModeSymlink, nil
+}
+
 // checkHostNs checks whether network sysctl is used in host namespace.
 func checkHostNs(sysctlConfig string, path string) error {
 	var currentProcessNetns = "/proc/self/ns/net"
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM

libcontainer/configs/validate/validator.go
```diff
@@ -148,6 +148,19 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
...
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return false, err
+	}
+
+	if (fi.Mode() & os.ModeSymlink) == 0 {
+		return false, nil
+	}
+
+	return true, nil
```
cyphar: This should be written as all one-line (no if statement):

```go
return fi.Mode()&os.ModeSymlink == os.ModeSymlink, nil
```

libcontainer/configs/validate/validator.go
```diff
@@ -156,6 +169,19 @@ func checkHostNs(sysctlConfig string, path string) error {
...
+	}
+
+	if symLink == false {
+		// The provided namespace is not a symbolic link,
+		// it is not the host namespace.
+		return nil
+	}
+
 	// readlink on the path provided in the struct
 	destOfContainer, err := os.Readlink(path)
```
cyphar: I'm confused what benefit this new code has -- this will always error out if `path` is not a symlink.

sameo: @cyphar If `path` is not symlink, i.e. a persistent namespace, it will *not* error out.
It will only error out if `path` is a symlink and point to the same namespace as `/proc/self/ns/net`, which is considered to be the host namespace.

In other words, this routine will error out if `path` is not pointing at a persistent namespace and if it's a symbolic link pointing at the host namespace.

cyphar: My point is that `readlink` will always error out if you try to read a non-symlink:

```
% touch /tmp/uts-ns
% sudo unshare --uts=/tmp/uts-ns hostname FOO
% sudo readlink /tmp/uts-ns
% echo $?
1
```

cyphar:
LGTM. Sorry about that confusion. :D

mrunalp:
LGTM

PR #1222, opened 2016-12-11
Split the code for remounting mount points and mounting paths.

### LINKED ISSUE #1223: mount flags on /dev are lost when remounting read only
If `/dev` is specified to be read only, all other specified mount flags are dropped.

Fix in #1222

justincormack:
fix #1223 

A remount of a mount point must include all the current flags or
these will be cleared:

```
The mountflags and data arguments should match the values used in the
original mount() call, except for those parameters that are being
deliberately changed.
```

The current code does not do this; the bug manifests in the specified
flags for `/dev` being lost on remount read only at present. As we
need to specify flags, split the code path for this from remounting
paths which are not mount points, as these can only inherit the
existing flags of the path, and these cannot be changed.

In the bind case, remove extra flags from the bind remount. A bind
mount can only be remounted read only, no other flags can be set,
all other flags are inherited from the parent. From the man page:

```
Since Linux 2.6.26, this flag can also be used to make an existing
bind mount read-only by specifying mountflags as:

MS_REMOUNT | MS_BIND | MS_RDONLY

Note that only the MS_RDONLY setting of the bind mount can be changed
in this manner.
```

MS_REC can only be set on the original bind, so move this. See note
in man page on bind mounts:

```
The remaining bits in the mountflags argument are also ignored, with
the exception of MS_REC.
```

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -120,8 +120,8 @@ func finalizeRootfs(config *configs.Config) (err error) {
 	// remount dev as ro if specified
 	for _, m := range config.Mounts {
 		if libcontainerUtils.CleanPath(m.Destination) == "/dev" {
-			if m.Flags&syscall.MS_RDONLY != 0 {
-				if err := remountReadonly(m.Destination); err != nil {
+			if m.Flags&syscall.MS_RDONLY == syscall.MS_RDONLY {
+				if err := remountReadonly(m); err != nil {
 					return newSystemErrorWithCausef(err, "remounting %q as readonly", m.Destination)
 				}
 			}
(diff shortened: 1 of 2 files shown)


justincormack:
Fixed the CI issue, sorry about that.

libcontainer/rootfs_linux.go
```diff
@@ -700,17 +700,26 @@ func createIfNotExists(path string, isDir bool) error {
...
-func remountReadonly(path string) error {
+// readonlyPath will make a path read only.
+func readonlyPath(path string) error {
+	if err := syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REC, ""); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+	return syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REMOUNT|syscall.MS_RDONLY|syscall.MS_REC, "")
```
cyphar: I actually figured this out today in umoci. So yeah, we need to include this. :P

cyphar:
Sorry for not looking into this earlier, I will take a look later today.

justincormack:
@cyphar no problem, thanks.

libcontainer/rootfs_linux.go
cyphar: I'm not sure we should've dropped the `|defaultMountFlags` from here.

rhatdan:
@rhvgoyal PTAL

libcontainer/rootfs_linux.go
justincormack: I specifically mention that in the commit message, bind mounts do not accept arbitrary mount flags, you can only set `RDONLY`. All other mount flags are inherited from the underlying mount and cannot be changed (because it is actually the same mount and hence same kernel datastructure).

libcontainer/rootfs_linux.go
```diff
@@ -121,7 +121,7 @@ func finalizeRootfs(config *configs.Config) (err error) {
 	for _, m := range config.Mounts {
 		if libcontainerUtils.CleanPath(m.Destination) == "/dev" {
 			if m.Flags&syscall.MS_RDONLY != 0 {
```
justincormack: Done

libcontainer/rootfs_linux.go
mrunalp: Yeah, good point. From my quick testing, it looks like doing it one shot should work.

cyphar: @justincormack What I mean is why not make the first `Mount` line like this:

```go
if err := syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REC|defaultMountFlags, ""); err != nil {
```

I accidentally commented on the wrong line.

justincormack: As I said, you cannot change the mount flags (other than to read only) on a bind mount, only a remount, so putting `defaultMountFlags` there is just ignored.

cyphar: Yeah, sorry you're right. I must've misread your original commit message.

cyphar: Okay, but I just tried to add `nosuid` to a bind mount. You're right that it doesn't work with the initial mount, but it does work with the remount:

```

```

The reason I'm harping on this is because `defaultMountFlags` contains security-related flags and I want to make sure that there's a good reason that we aren't going to be applying them for bindmounts anymore (though it looks like we never applied them).

justincormack: Hmm. But in this case if you wanted this path `nosuid` you would have mounted it like that in the config? Having the read-only masked part only being `nosuid` but not the read-write part seems weird.

cyphar: > But in this case if you wanted this path nosuid you would have mounted it like that in the config?

Yeah, by my concern is with paths marked as `readonlyPaths`, which don't have an entry in the config. The spec just mentions that the path needs to be readonly, so this is probably fine. I was just concerned about suddenly allowing something that wasn't intended to be allowed previously.

justincormack: But there will always be an entry for the mountpoint above, where the flags can be set. I don;t think anyone should rely on read only also setting other flags, and if they are set on the writeable part they are far more dangerous.

cyphar: :+1:

cyphar:
LGTM

hqhq:
LGTM

PR #1226, opened 2016-12-14
Clean apt archives and source directories in Dockerfile

nhlfr:
Signed-off-by: Michal Rostecki <michal@kinvolk.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,8 @@ RUN apt-get update && apt-get install -y \
     protobuf-c-compiler \
     protobuf-compiler \
     python-minimal \
-    --no-install-recommends
+    --no-install-recommends \
+    && apt-get clean
 
 # install bats
 RUN cd /tmp \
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

Dockerfile
```diff
@@ -34,7 +35,8 @@ ENV CRIU_VERSION 1.7
 RUN mkdir -p /usr/src/criu \
     && curl -sSL https://github.com/xemul/criu/archive/v${CRIU_VERSION}.tar.gz | tar -v -C /usr/src/criu/ -xz --strip-components=1 \
     && cd /usr/src/criu \
-    && make install-criu
+    && make install-criu \
+    && make clean
```
cyphar: Silly question, why don't we just do an `rm -rf /usr/src/criu`? Once it's installed do we need the source code lying  around?

nhlfr: Good idea, done.

mrunalp:
LGTM

hqhq:
LGTM

PR #1233, opened 2016-12-17
Bump runtime-spec to v1.0.0-rc3

WeiZhang555:
* Bump underlying runtime-spec to version 1.0.0-rc3
* Fix related changed struct names in config.go

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -59,8 +59,8 @@
 		},
 		{
 			"ImportPath": "github.com/opencontainers/runtime-spec/specs-go",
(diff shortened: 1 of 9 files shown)


WeiZhang555:
Latest runtime-spec is needed in my another PR: https://github.com/opencontainers/runc/pull/1201 , after this is merged, I'll try to make `HookState` consistent with runtime spec `State`

cyphar:
LGTM. Though this reminds me that we should probably switch away from `godep` -- for obvious reasons. I'll open a separate issue about that.

hqhq:
LGTM

Issue #1236, opened 2016-12-19
sync race issue in nsexec.c


hqhq:
When I use the latest runc with docker, got a lot of failure in integration tests, they all failed with `exit status 25`, full output looks like:
```
FAIL: docker_cli_run_test.go:3631: DockerSuite.TestRunAbsoluteInvalidCgroupParent

docker_cli_run_test.go:3642:
    // XXX: This may include a daemon crash.
    c.Fatalf("unexpected failure when running container with --cgroup-parent option - %s\n%v", string(out), err)
... Error: unexpected failure when running container with --cgroup-parent option - /go/src/github.com/docker/docker/bundles/1.11.2/test-integration-cli/../dynbinary/docker: Error response from daemon: rpc error: code = 2 desc = "oci runtime error: container_linux.go:247: starting container process caused \"process_linux.go:272: running exec setns process for init caused \\\"exit status 25\\\"\"\n".

exit status 125
```

To be explicit, we failed at: https://github.com/opencontainers/runc/blob/master/libcontainer/nsenter/nsexec.c#L579-L581

The code was added by: https://github.com/opencontainers/runc/pull/1154
But I don't think it introduced this issue, it's probably because we are using the same socketpair to sync with child and grandchild, so there is race condition that parent read data in the wrong sequence.

hqhq:
ping @cyphar , I'm working on a fix now.

PR #1237, opened 2016-12-19
Fix race condition when sync with child and grandchild

hqhq:
Fixes: #1236


hqhq:

teddyking:
fwiw it looks very likely that this PR also fixes #1281.

libcontainer/nsenter/nsexec.c
```diff
@@ -695,6 +723,10 @@ void nsexec(void)
 				bail("failed to sync with parent: SYNC_RECVPID_ACK: got %u", s);
 			}
 
+			s = SYNC_CHILD_READY;
+			if (write(syncfd, &s, sizeof(s)) != sizeof(s))
+				bail("failed to sync with patent: write(SYNC_CHILD_READY)");
```
runcom: s/patent/parent/g

hqhq: Updated, thanks.

crosbymichael:
@cyphar can you take a look at this one?

stevenh:
I'm assuming this is in a final state now, if so then would be good to get it in ASAP as its causing pretty much all other PR's to have failed tests :(

cyphar:
/me is looking at it. While it does fix the issue, there's some parts I'd like to get cleaned up before merging.

stevenh:
Out of interest does that include changing the way the children are created and reaping the initial child which would make https://github.com/opencontainers/runc/pull/1301 unneeded?

crosbymichael:
LGTM

hqhq:
ping @cyphar

libcontainer/nsenter/nsexec.c
```diff
@@ -570,20 +571,46 @@ void nsexec(void)
...
-					kill(child, SIGKILL);
-					bail("failed to sync with child: unexpected SYNC_RECVPID_ACK");
 					break;
 				case SYNC_CHILD_READY:
-					ready++;
+					ready = true;
 					break;
 				default:
-					bail("unexpected sync value");
+					bail("unexpected sync value: %u", s);
```
cyphar: While I understand this makes the code nicer, remember that the _only_ debugging information we normally get is the return code. Maybe we don't lose too much information, but something to keep in mind.

libcontainer/nsenter/nsexec.c
```diff
@@ -570,20 +571,46 @@ void nsexec(void)
...
+				}
+			}
+
+			/* Now sync with grandchild. */
+
+			ready = false;
+			while (!ready) {
```
cyphar: What happens if the grandchild is ready before the child -- or if the child dies above but the grandchild is still around?

libcontainer/nsenter/nsexec.c
```diff
@@ -413,7 +414,7 @@ void nsexec(void)
 {
 	int pipenum;
 	jmp_buf env;
-	int syncpipe[2];
+	int syncChildPipe[2], syncGrandchildPipe[2];
```
cyphar: Nit: Please use `snake_case` in C code.

libcontainer/nsenter/nsexec.c
hqhq: > What happens if the grandchild is ready before the child

It won't, grandchild won't do anything before father finishes syncing with child and send SYNC_GRANDCHILD to grandchild.

> or if the child dies above but the grandchild is still around

All cases that child dies by error after creating grandchild, child will kill grandchild before exit.

libcontainer/nsenter/nsexec.c
hqhq: Yeah, I don't think we lose any information here.

hqhq:
Updated to use snake case.

@crosbymichael @cyphar PTAL.

cyphar:
Failure is because `shfmt` is broken on Go 1.6.

LGTM

cyphar:
I've made an issue (mvdan/sh#66) to track it.

hqhq:
ping @crosbymichael

crosbymichael:
LGTM

Issue #1044, opened 2016-09-14
CGroup device configuration in prestart hook


3XX0:
Currently the container process joins its cgroup before the hooks are triggered ([source](https://github.com/opencontainers/runc/blob/master/libcontainer/process_linux.go#L257)) but the configuration is actually applied later when the `procReady` signal is received ([source](https://github.com/opencontainers/runc/blob/master/libcontainer/process_linux.go#L290)).

Is that the intended behavior? I'm asking because it makes it impossible to configure the device cgroup from within a prestart hook if the author of the hook doesn't control the runc config. The first device rule will most likely be `{Major: Wildcard, Minor: Wildcard, Allow: false}` or set `AllowAllDevices: false` thus overriding any cgroup configuration made by the hook itself.

cyphar:
This is intended behaviour. The reason for this change is because of the fact that the Go runtime loves to consume resources (especially PIDs), and setting the limits earlier would result in the author of a `config.json` having to worry about the overhead of `runc`. However, I'm open to a proposal which can allow for that problem to be dealt with as well as making hooks more functional.

It should be noted that if you use the `create` and `start` operations separately then you should be able to tweak the cgroups as well. Currently there is a lot of discussion in the specification repo whether we should remove hooks (since the `create`-`start` operations should be sufficient to handle all of the hooks features).

3XX0:
I see it makes sense. Yes, I'm aware of the `create`/`start` debate but I stick with hooks simply because in my case I can't control the invocation of runc (e.g. Docker).

Can't we just move the cgroup assignment only for the device subsystem? This shouldn't impact runc apart from the [/dev/null reopen](https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L92). I'm not sure about the other subsystems but the device wildcard override looks like an edge case that could be handled this way.

3XX0:
Any insights on this? This is the only thing stopping us from adding GPU support for runc/Docker.

3XX0:
Fixed by #1239

PR #553, opened 2016-02-11
libcontainer: integration: fix flaky pids limit tests

cyphar:
Because we are implemented in Go, the number of pids present in a
container is not very well-defined (other than it not being /much/
bigger than the limit you'd want to set). As a result, we need to make
the tests a bit less flaky in this regard.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

/cc @LK4D4 @crosbymichael @vmarmol @mrunalp @dqminh

PR #1239, opened 2016-12-20
Fix setup cgroup before prestart hook

moypray:
* User Case:
User could use prestart hook to add block devices to container. so the
hook should have a way to set the permissions of the devices.

Just move cgroup config operation before prestart hook will work.

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -280,9 +280,6 @@ func (p *initProcess) start() error {
 	ierr := parseSync(p.parentPipe, func(sync *syncT) error {
 		switch sync.Type {
 		case procReady:
-			if err := p.manager.Set(p.config.Config); err != nil {
-				return newSystemErrorWithCause(err, "setting cgroup config for ready process")
-			}
 			// set rlimits, this has to be done here because we lose permissions
 			// to raise the limits once we enter a user-namespace
 			if err := setupRlimits(p.config.Rlimits, p.pid()); err != nil {
@@ -290,6 +287,11 @@ func (p *initProcess) start() error {
 			}
 			// call prestart hooks
 			if !p.config.Config.Namespaces.Contains(configs.NEWNS) {
+				// Setup cgroup before prestart hook, so that the prestart hook could apply cgroup permissions.
+				if err := p.manager.Set(p.config.Config); err != nil {
+					return newSystemErrorWithCause(err, "setting cgroup config for ready process")
+				}
+
 				if p.config.Config.Hooks != nil {
 					s := configs.HookState{
 						Version: p.container.config.Version,
@@ -310,6 +312,10 @@ func (p *initProcess) start() error {
 			}
 			sentRun = true
 		case procHooks:
+			// Setup cgroup before prestart hook, so that the prestart hook could apply cgroup permissions.
+			if err := p.manager.Set(p.config.Config); err != nil {
+				return newSystemErrorWithCause(err, "setting cgroup config for procHooks process")
+			}
 			if p.config.Config.Hooks != nil {
 				s := configs.HookState{
 					Version: p.container.config.Version,


cyphar:
I'm not convinced this is correct. The reason why cgroups were written to be applied that late is so that hooks and other `runC` internal processes don't hit the resource constraints for the container. For example, setting the `pids.limit` resource to be very small then you'll start hitting troubles if hook scripts start using a lot of resources.

mrunalp:
@moypray Can't you setup the cgroups in your config.json?

moypray:
@cyphar , I think the hook is not under control of container cgroup. because hook is not a child process of 'runc init' but parent runC. I think if the "runc init" meet the limit of cgroup should be ok.  Is that right?

moypray:
@mrunalp  actually, if I use the config of container, will bring lots of troubles: such as config should be persistent and uniformity. (docker daemon will keep a cache in memory, I need to modify lots of codes to fix this, and would break the origin docker interface, It is terrible).  But thanks all the same.

hqhq:
@cyphar hook processes are invoked by runc, they are not constrained by container resources. Or are you concerning that the manipulation of cgroup resources by hooks could be malicious?

I think it's reasonable that we allow prestart hooks to customize cgroup resources.
Design LGTM.

cyphar:
@hqhq 

My concern is that a hook could do something like `runc exec` and I'm not sure if we should be enforcing cgroup limits on those processes. Not to mention that I don't think that hooks should be modifying stuff that was configured in `config.json` -- just seems like a footgun.

> I think it's reasonable that we allow prestart hooks to customize cgroup resources.

In which case this patch should drop `procReady` entirely because it's not used for anything else IIRC.

hqhq:
> My concern is that a hook could do something like runc exec and I'm not sure if we should be enforcing cgroup limits on those processes. 

If hook wants to run processes in container, why should they not be constrained by cgroup resources? What if that a long-running forked process which can live with container processes?

> Not to mention that I don't think that hooks should be modifying stuff that was configured in config.json -- just seems like a footgun.

I think the point is that hook authors should be responsible for the containers, most time they should be the same authors, like we permitted hooks to mess with container mounts.

> In which case this patch should drop procReady entirely because it's not used for anything else IIRC.

This PR very likely needs to be amended, I haven't look it through, just talking about design first :)

cyphar:
@hqhq 

> I think the point is that hook authors should be responsible for the containers, most time they should be the same authors, like we permitted hooks to mess with container mounts.

Fair enough. I'm just worried about people deciding to use hooks as a way to avoid touching the OCI specification (like the PR author is trying to do). But ultimately some usecases are more complicated than what can be described in `config.json`. I'm conflicted to tell the truth, let's wait to see what the other @opencontainers/runc-maintainers think.

moypray:
@cyphar , sorry for misunderstanding, I think the hook is running on host, the same with docker, runc.  and  customized by service provider themselves.  We should trust the hook.

moypray:
ping  @opencontainers/runc-maintainers , How do you feel it, just talking about the design, not this PR implement, thx~

cyphar:
@moypray Most maintainers are on holidays now (including myself). I'd wait for a while before pinging people again, just because everyone is AFK at the moment. :wink:

moypray:
@cyphar , OK, thanks for your replay.

hqhq:
@cyphar Can you check how would this affect process overhead of pid cgroup? I see we already have inevitable overhead for pid cgroup, since the test is pretty loose https://github.com/opencontainers/runc/pull/553 .

moypray:
@opencontainers/runc-maintainers , what do you think of this proposal?

> Can you check how would this affect process overhead of pid cgroup?   

@hqhq  I have done some testing based on my fix.  And I am sure that there will not be more go routines or threads created between the two positions of 'Set cgroup' operation. So it will not affect the pid cgroup.
Using new runC with my fix, We can run `docker run --pid-limits 2 busybox cat /sys/fs/cgroup/pids/pids.max` successfully, if my fix affects the pids cgroup, it would run with failure.

hqhq:
@opencontainers/runc-maintainers PTAL, design and code both look good to me.

3XX0:
See also the original issue: https://github.com/opencontainers/runc/issues/1044

3XX0:
@mrunalp This is what I was referring to for the GPU support in Kubernetes

mrunalp:
This makes sense to me. @moypray can you rebase the PR and address comments?
@3XX0 Thanks for the reminder.

libcontainer/process_linux.go
```diff
@@ -357,6 +359,10 @@ func (p *initProcess) start() error {
 			}
 			sentRun = true
 		case procHooks:
+			// Setup cgroup before prestart hook, so that the prestart hook could apply cgroup permissions.
+			if err := p.manager.Set(p.config.Config); err != nil {
+				return newSystemErrorWithCause(err, "setting cgroup config for ready process")
+			}
```
TomSweeneyRedHat: NIT.  Can we have slightly different error message text so it can be more easily determined where the error came from?  Perhaps appending " in procReady" to the first message at line 339 and " in procHooks" here in the second?

mrunalp:
@3XX0 Would you be willing to carry this over to a new rebased PR?

hqhq:
@moypray Please rebase, thanks.

moypray:
hi @mrunalp @hqhq   sorry for late replay.  I will rebase it ASAP.

moypray:
hi @mrunalp  @hqhq   done, Sorry again for late replay.

hqhq:
LGTM

3XX0:
@mrunalp Sorry, saw it a little late ;)
The patch looks good to me though!

moypray:
Ping @mrunalp

mrunalp:
LGTM

PR #1240, opened 2016-12-21
fix typo

YuPengZTE:
Signed-off-by: yupeng <yu.peng36@zte.com.cn>


hqhq:
According to: http://the-difference-between.com/unglamorous/unglamourous
I don't think it's typo.

So I'm closing it, thanks.

PR #1245, opened 2016-12-28
Fix integration when missing criu

WeiZhang555:
If criu isn't installed in the host, integration tests won't be able
to go on, it should skip checkpoint/restore tests instead of interrupt.

Output before:

```
$ make localintegration
bats -t tests/integration
/usr/lib/bats/bats-exec-suite: line 20: let: count+=: syntax error:
operand expected (error token is "+=")
Makefile:103: recipe for target 'localintegration' failed
```

### DIFF
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -18,7 +18,7 @@ HELLO_IMAGE="$TESTDATA/hello-world.tar"
 HELLO_BUNDLE="$BATS_TMPDIR/hello-world"
 
 # CRIU PATH
-CRIU="$(which criu)"
+CRIU="$(which criu || true)"
 
 # Kernel version
 KERNEL_VERSION="$(uname -r)"


tests/integration/helpers.bash
```diff
@@ -18,7 +18,7 @@ HELLO_IMAGE="$TESTDATA/hello-world.tar"
 HELLO_BUNDLE="$BATS_TMPDIR/hello-world"
 
 # CRIU PATH
-CRIU="$(which criu)"
+CRIU="$(which criu)" || true
```
cyphar: How does this fix the problem? As far as I can tell, you're setting the value of `CRIU` to be ` || true`. To be fair, that path probably doesn't exist, but it doesn't mean the logic makes sense.

WeiZhang555:
The root cause of this error is that bats will exit immediately once a command failed with `set -e`, if `criu` doesn't exist, `which criu` will fail and `bats` will be interrupted. 

The result of this logic is that if I don't install criu, I can't run the integration, that's wrong.

tests/integration/helpers.bash
WeiZhang555: In my understanding, the logic is `(CRIU="$(which criu)") || true` which will always set `$?=0`. If I don't do this, once which criu failed, `bats` will quit (`set -e`)

cyphar: > In my understanding, the logic is (CRIU="$(which criu)") || true

I don't think that's how it works (since environment variables aren't propagated from subshells). You should instead make it something like

```
CRIU="$(which criu || true)"
```

cyphar: > (CRIU="$(which criu)") || true

Yeah, I don't really like it. Can you instead use

```
CRIU="$(which criu || true)"
```

I also don't really like that `$CRIU` is now set to `criu not found` rather than an empty string, but what are you going to do...

WeiZhang555:
@cyphar Updated as you suggested.

WeiZhang555:
Rebased.

cyphar:
LGTM.

/cc @avagin

hqhq:
LGTM

PR #1252, opened 2017-01-04
remove `-i` option to avoid failure of jenkins in non-interactive mode.

FengtuWang:
Signed-off-by: Fengtu Wang <wangfengtu@huawei.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ localtest:
 	make localunittest localintegration
 
 unittest: runcimage
-	docker run -e TESTFLAGS -ti --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localunittest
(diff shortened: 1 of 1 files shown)


FengtuWang:
Use `make test` in jenkins cause `cannot enable tty mode on non tty input`. It seems that this option is not needed here.

FengtuWang:
My jenkins can't run an interactive shell.So when i use 'make test' command in jenkins's shell, it print error message 'cannot enable tty mode on non tty input' as '-i' option can only used in an interactive environment. Jenkins in our upstream CI didn't have such problem because it didn't use 'make test' command in the shell. It use 'docker run' command directly instand, no option '-i' as well. Maybe someone's jenkins can resolve such problem by some plugins, but not mine. As this option seems no use, i'd like to remove it.

FengtuWang:
I think we should keep '-i' option in  'make shell' command. This command seems to be designed for interactive situation, because it just start a container and enter bash. Other commands should be input manually.

hqhq:
LGTM

crosbymichael:
LGTM

PR #1259, opened 2017-01-07
Fix error shadow and error check warnings

mrunalp:

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -14,7 +14,7 @@ import (
 )
 
 func killContainer(container libcontainer.Container) error {
-	container.Signal(syscall.SIGKILL, false)
+	_ = container.Signal(syscall.SIGKILL, false)
 	for i := 0; i < 100; i++ {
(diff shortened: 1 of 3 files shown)


hqhq:
LGTM

delete.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func killContainer(container libcontainer.Container) error {
-	container.Signal(syscall.SIGKILL, false)
+	_ = container.Signal(syscall.SIGKILL, false)
```
crosbymichael: Its this golint or govet stuff?

crosbymichael:
LGTM

delete.go
hqhq: I think neither, @mrunalp how did you get these warnings?

mrunalp: I ran gometalinter. I hope that we could enable it for runc at some point.

Issue #349, opened 2015-10-21
runc should not emit errors on stderr (cannot distinguish from contained process stderr)


heavenlyhash:
Runc currently emits some errors on stderr (even if the logs are redirected to a file).   It would be desirable if these errors weren't emitted on stderr --  It's pretty tough (read: impossible, I think?) to separate these messages on stderr from something the executable inside the container might want to legitimately emit on stderr.

Some errors are also doubled up on both stderr and routed through through the logger.  These are also impossible to dedup, sadly, unless one gives up on streaming stderr to any callers (I can't figure out a way around the race between the error on the log channel and when it comes out on stderr except for waiting for the process to exit entirely... there's no "all clear" signal on the log channel we can wait for either).

Many of the offenders seem to be flowing through the `fatal` and `fatalf` functions in `utils.go`, which refer directly to `os.Stderr`.  Can we replace all of these with `logrus.Fatal` so the errors are both formatted and routed to the log file if set?

(Happy to work on a patch, opening ticket for discussion first just to check this sounds reasonable :) )

hqhq:
There is similar problem for container mechanism, syslog is not namespaced, so all containers share the syslog with host.

It is a problem that runc shares the stderr with container, but I don't think there is an easy way to fix the root cause. Let runc use `logrus` while leave stderr to container doesn't make much sense to me. Emit errors to stderr is human nature, and how do you handle nested containers? Should they share the stderr?

So I don't think it can fix this. What about we add some prefix to distinguish host and different containers?

heavenlyhash:
I think nested containers are a discussion topic that can probably best be discussed with similar strategies to IO stream handling for trees of contained processes in general.

heavenlyhash:
I found some tweaks to LinuxFactory.StartInitialization that makes the behavior for the previously double-reported error messages better... will fire a PR after I figure out where to put tests and have a chance to clean it up a bit.

cyphar:
This has been resolved since ~2017 with the introduction of proper terminal support in runc in #1018. This isn't exactly what you wanted but when using `--console-socket` the output is perfectly segregated (container output only goes through the allocated console) and in forwarding (i.e., not `runc run -d`) mode we use the same allocated console for stdout and stderr so the forwarding logic treats all container output as stdout.

Detached mode with host terminal fds will still mix output but that mode is really quite insecure. I'm also not sure what we could do to improve the situation for stderr in that case.

PR #1261, opened 2017-01-09
Add support for IO redirection on runc exec

georgethebeatle:
As mentioned in [this issue](https://github.com/opencontainers/runc/issues/349), separating the stdout/err streams of `runc exec/init` and the container is not possible because the streams of runc get inherited by the container process. This is not ideal for Garden because we don't want runc errors appearing in the "user's" application logs.

This PR attempts to separate the standard streams of runc from those of the container by adding 3 flags - one for stdin, stdout and stderr. If you pass one of these flags this essentially redirects the corresponding standard stream to the given file just before exec-ing the container process. After this redirection it is safe to assume that runc's streams are not written to by the container process and vice versa, the container's streams are not written to by runc.

This is a first stab at this behaviour but it seemed consistent with how the LISTEN_FDS behaved. 

Signed-off-by: Georgi Sabev <georgethebeatle@gmail.com>

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -84,6 +84,18 @@ following will output a list of processes running in the container:
 			Usage:  "disable the use of the subreaper used to reap reparented processes",
 			Hidden: true,
 		},
+		cli.StringFlag{
+			Name:  "stdin",
+			Usage: "redirect process stdin to given filepath",
+		},
+		cli.StringFlag{
+			Name:  "stdout",
+			Usage: "redirect process stdout to given filepath",
+		},
+		cli.StringFlag{
+			Name:  "stderr",
+			Usage: "redirect process stderr to given filepath",
+		},
 	},
 	Action: func(context *cli.Context) error {
 		if err := checkArgs(context, 2, minArgs); err != nil {
@@ -136,6 +148,9 @@ func execProcess(context *cli.Context) (int, error) {
 		container:       container,
 		consoleSocket:   context.String("console-socket"),
 		detach:          detach,
+		stdin:           context.String("stdin"),
+		stdout:          context.String("stdout"),
+		stderr:          context.String("stderr"),
 		pidFile:         context.String("pid-file"),
 	}
 	return r.run(p)
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -426,3 +426,47 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	}
 	return nil
 }
+
+func setupStdio() error {
+	stdinFD := os.Getenv("STDIN_FD")
+	stdoutFD := os.Getenv("STDOUT_FD")
+	stderrFD := os.Getenv("STDERR_FD")
+
+	if len(stdinFD) > 0 {
+		stdinFD, err := strconv.Atoi(stdinFD)
+		if err != nil {
+			return err
+		}
+
+		err = syscall.Dup2(stdinFD, 0)
+		if err != nil {
+			return err
+		}
+	}
+
+	if len(stdoutFD) > 0 {
+		stdoutFD, err := strconv.Atoi(stdoutFD)
+		if err != nil {
+			return err
+		}
+
+		err = syscall.Dup2(stdoutFD, 1)
+		if err != nil {
+			return err
+		}
+	}
+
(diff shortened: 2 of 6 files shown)


crosbymichael:
We have been using the `--log` flag to reading runc errors, does that work for you or are there still some different errors that are you seeing on the IO?  

What do you think about only writing runc errors to the log and not touching IO so we keep the same dup logic for the container?

williammartin:
We have been [using](https://github.com/cloudfoundry/guardian/blob/master/cmd/dadoo/main_linux.go#L62) the `--log` flag when execing but we still saw an error making it to the stderr. Annoyingly I think I lost the full trace but `runc init` exited (SIGABRT maybe?) and printed `runtime/cgo: pthread_create failed: Resource temporarily unavailable`. I'll try to find all the info.

Obviously, the underlying cause is our problem but the failure highlighted an area that wasn't behaving totally as expected. I think we'd be fine if all errors made it to log, but I don't know how hard that is to ensure in every case. We felt it was a simpler mental model to just split the streams and avoid the confusion anyway.

crosbymichael:
I think we can try to make sure, with a recover that runc errors don't hit the stdio for the container.  

Its worth a try because it reduces the complexity with having to pass additional fds or files via flags to runc.  We have already did a fairly good job moving much of the internal runc errors,etc to the log, we just need to find the remaining places.

Would this work for you all if we try that first?

williammartin:
I like the idea if it's possible. I can't say I'm familiar enough with golang behaviour but is it really possible for you to recover from all errors? I'm still trying to get a trace of our problem but I would expect for example, if runc hit OOM, the runtime would throw something unrecoverable like `fatal error: runtime: out of memory` into the container's stdio.

What do you think?

williammartin:
Any further thoughts on this? @crosbymichael

cyphar:
My concern with this is that it's starting to mess around with the recent `--console-socket` changes. I also don't like that we're adding even more crufty functions called `setupIO` (the code is already difficult enough to follow).

Is there a way we can make `--log` better? In principle I would like this idea because it means that a user could get away from the default setup of runC stdio being passed to it (which is a security concern), but I'm a bit worried about the interaction with console sockets and other parts of the console handling.

:weary:

teddyking:
We've been discussing this a bit more. I think we probably could look to improve `--log` but I guess our concern is that it may not be possible to always catch _everything_, which isn't great as this can lead to issues further up the stack for us.

You mentioned about adding crufty `setupIO` funcs and possibly messing around with the `--console-socket` changes, is there any other way we could look to implement this functionality in a neater and/or safer way?

williammartin:
Hey @cyphar Just following up on Ed's previous comment. Do you have any thoughts on how we could achieve the same functionality in a way that fits better with your larger `console-socket` body of work? 

We would like to help out and PR something more suitable if possible, but we don't feel like we have the full context.

cyphar:
Sorry, I've been travelling in the past few weeks. I just got back today, so I'll get back to this issue once I catch up to it in my backlog.

heavenlyhash:
Poll if there are any updates on this?  :)

I'm following here from #349 and this patch looks like exactly the solution I'm dreaming of: it would allow total clarity of what's from the contained process and what's not.

heavenlyhash:
FWIW, I just read the patch and it looks roughly good, with one nit:

We should add some tests that a container that execs

```bash
echo 1 >&1
echo 2 >&2
echo 3 >&1
echo 4 >&2
```

... actually emits "1\n2\n3\n4\n" in the correct order when stdout and stderr are redirected to the same file.

It looks like this will always open two different file descriptors for stdout and stderr, which may be not-quite-correct: IIUC, if the user redirects them to the same files, it's desirable to have the same filedecscriptors, as without this interleaving writes may not DWIM because they end up with separate write buffering in the kernel, and so on.  (Go's os/exec package has a couple interesting conditional branches for "if p.stderr == p.stdout` for this reason, iirc.)

kolyshkin:
The only time `runc --log` will print something to stdin/stdout is when it calls `panic()`, or go runtime fails in some way. This is unfortunate but rather unusual, meaning something is broken very badly.

It probably doesn't make sense to try to workaround for such cases.

PR #1265, opened 2017-01-10
Fix regression of exec command

### LINKED ISSUE #1264: runc exec requires command to be passed on CLI
It seems that [this commit](https://github.com/opencontainers/runc/commit/b517076907b7c18d7a1f296ec390cf3f275e2375) introduced a minArgs requirement of 2 for `runc exec`, with the second arg consisting of a process to run.

Is this actually a requirement here? Previously we were able to run `runc exec` by passing `-p process.json` containing the command to run etc.

Would it be ok to make the `<container command>` optional again?
We'd be happy to submit the PR.

Thanks!
Ed & Konstantinos

WeiZhang555:
Fixes #1264

Set minimal required args to 1 for `exec` command to roll back to
previous behavior, also modify help message a little bit.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -18,10 +18,11 @@ import (
 var execCommand = cli.Command{
 	Name:  "exec",
 	Usage: "execute new process inside the container",
-	ArgsUsage: `<container-id> <container command> [command options]
+	ArgsUsage: `<container-id> <command> [command options]  || -p process.json <container-id>
 
 Where "<container-id>" is the name for the instance of the container and
-"<container command>" is the command to be executed in the container.
+"<command>" is the command to be executed in the container.
+"<command>" can't be empty unless a "-p" flag provided.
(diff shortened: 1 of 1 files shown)


WeiZhang555:
ping @cyphar

crosbymichael:
LGTM

mrunalp:
LGTM

exec.go
```diff
@@ -86,7 +87,7 @@ following will output a list of processes running in the container:
 		},
 	},
 	Action: func(context *cli.Context) error {
-		if err := checkArgs(context, 2, minArgs); err != nil {
+		if err := checkArgs(context, 1, minArgs); err != nil {
```
cyphar: This should actually be conditional based on whether `-p` was provided. I'll write a patch to fix it.

PR #1274, opened 2017-01-16
libcontainer: init: only pass stateDirFd when creating a container

cyphar:
If we pass a file descriptor to the host filesystem while joining a
container, there is a race condition where a process inside the
container can ptrace(2) the joining process and stop it from closing its
file descriptor to the stateDirFd. Then the process can access the
*host* filesystem from that file descriptor. This was fixed in part by
5d93fed3d27f ("Set init processes as non-dumpable"), but that fix is
more of a hail-mary than an actual fix for the underlying issue.

To fix this, don't open or pass the stateDirFd to the init process
unless we're creating a new container. A proper fix for this would be to
remove the need for even passing around directory file descriptors
(which are quite dangerous in the context of mount namespaces).

There is still an issue with containers that have CAP_SYS_PTRACE and are
using the setns(2)-style of joining a container namespace. Currently I'm
not really sure how to fix it without rampant layer violation.

Fixes: CVE-2016-9962
Fixes: 5d93fed3d27f ("Set init processes as non-dumpable")

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -222,28 +222,32 @@ func (l *LinuxFactory) Type() string {
 // StartInitialization loads a container by opening the pipe fd from the parent to read the configuration and state
 // This is a low level implementation detail of the reexec and should not be consumed externally
 func (l *LinuxFactory) StartInitialization() (err error) {
-	var pipefd, rootfd int
-	for _, pair := range []struct {
-		k string
-		v *int
-	}{
-		{"_LIBCONTAINER_INITPIPE", &pipefd},
-		{"_LIBCONTAINER_STATEDIR", &rootfd},
-	} {
+	var (
+		pipefd, rootfd int
+		envInitPipe    = os.Getenv("_LIBCONTAINER_INITPIPE")
+		envStateDir    = os.Getenv("_LIBCONTAINER_STATEDIR")
+	)
 
-		s := os.Getenv(pair.k)
-
-		i, err := strconv.Atoi(s)
-		if err != nil {
-			return fmt.Errorf("unable to convert %s=%s to int", pair.k, s)
-		}
-		*pair.v = i
+	// Get the INITPIPE.
+	pipefd, err = strconv.Atoi(envInitPipe)
+	if err != nil {
+		return fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE=%s to int: %s", envInitPipe, err)
 	}
+
 	var (
 		pipe = os.NewFile(uintptr(pipefd), "pipe")
 		it   = initType(os.Getenv("_LIBCONTAINER_INITTYPE"))
 	)
 	defer pipe.Close()
+
+	// Only init processes have STATEDIR.
+	rootfd = -1
+	if it == initStandard {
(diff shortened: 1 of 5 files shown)


libcontainer/factory_linux.go
```diff
@@ -222,27 +222,34 @@ func (l *LinuxFactory) Type() string {
...
 	}
+
 	var (
 		pipe = os.NewFile(uintptr(pipefd), "pipe")
 		it   = initType(os.Getenv("_LIBCONTAINER_INITTYPE"))
 	)
+
+	// Only get the STATEDIR if we're an init process. It's a bit late to do
+	// anything about this if we've already brought in an fd (a racing process
+	// could've opened the fd by now because we're in the PID namespace).
```
hqhq: I don't quite understand the second sentence in this comment. Can you explain or rephrase so it can be more clear?

I think your point is to explain the effect if we set STATEDIR in exec process?

cyphar: If STATEDIR is set by a parent, then by the time we hit this Go code the race condition in CVE-2016-9962 could've been hit. That's what I'm trying to explain, which is why I don't attempt to close `STATEDIR` if we're an init process -- we're already stuffed if we have to attempt to close the fd.

hqhq: So you are explaining the effect of the case that if we set STATEDIR for exec process right? I think your comments in `newParentProcess` is sufficient and all you do is to not set STATEDIR or pass stateDirFD to exec process, so I think you don't have to explain why you do so in this late phase, which is why I'm confused by this comment.

cyphar: Eh, alright I'll drop the comment. The reason for writing it was to ensure that nobody would think that `STATEDIR` should ever be set or checked in `runc exec`.

cyphar: Fixed.

cyphar:
Fixed @hqhq's nit. PTAL, this is quite important (it extends the CVE-2016-9962 fix so that it also works for `CAP_SYS_PTRACE` containers!).

hqhq:
LGTM

cyphar:
Rebased.

avagin:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1274/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

crosbymichael:
LGTM

crosbymichael:
LGTM

mrueg:
@cyphar @crosbymichael Can you please tag a new release candidate that fixes the security issue, so distributions don't need to apply further patches?

cyphar:
@crosbymichael I'll make a proposal for a new rc release.

hqhq:
Can we include https://github.com/opencontainers/runc/pull/1237 in the new release, it fixes really annoying issue.

cyphar:
I don't really have time to review that today -- it would have to wait until the end of the week.

mrueg:
@hqhq @cyphar @crosbymichael Is there a way to get a new release out anytime soon? Tags are cheap, if #1237 is fixed, you could just release rc4.

Issue #1269, opened 2017-01-11
Add support for setting console size for runc exec -d -tty -console-socket ... 


teddyking:
Hi,

We'd like to be able to configure an initial console size when running a `runc exec -d -tty -console-socket /path/to/socket.sock ...`. We've started work on a possible PR to add this functionality but I wanted to open this up for discussion/feedback.

Our current thoughts are to add a `setConsoleSize` func just after opening up `/dev/ptmx` [here](https://github.com/opencontainers/runc/blob/eedebb8c40ca598ed45f4ea023c1f1f43d442189/libcontainer/console_linux.go#L12-L31). The values for the height/width would be defined by `consoleSize` in the process.json and passed down via initConfig.

Does this sound reasonable? We've been testing this out today and it seems to do what we want.

For context, we're looking to pull in the latest version of runc (incl. the new console work), but the changes are causing flakey behaviour in our tests. Being able to define the initial size for the console will help to alleviate this.

Thanks!
Ed & Petar

mrunalp:
I think that sounds reasonable. Could you open a PR for that?

teddyking:
Great thanks - we've opened up PR #1275 so I'll close this one off.

PR #1275, opened 2017-01-16
Set initial console size based on process spec

williammartin:
Signed-off-by: Petar Petrov <pppepito86@gmail.com>

This commit hopefully resolves issue #1269, setting the initial console size for TTYs based on the ConsoleSize properties in the process spec.

Two thoughts about this PR that would be good to get feedback on: Firstly, the height and width are passed through as separate primitives and turned into a `term.WinSize` right before use. We weren't sure of the pros and cons of this approach vs the `term` package all the way up the stack, or the runtime-spec `Box` or even whether we should define another process config struct.

Secondly, the bats test would ideally read the master fd from the `$CONSOLE_SOCKET` and read output directly from a `runc exec` invocation calling `stty -a` but there seems to be no convention for doing this and working around the existing `recvtty` infrastructure to allow these kind of tests seemed like a fairly sizable change, that might be best as its own PR.

### DIFF
--- a/libcontainer/process.go
+++ b/libcontainer/process.go
@@ -47,6 +47,10 @@ type Process struct {
 	// ExtraFiles specifies additional open files to be inherited by the container
 	ExtraFiles []*os.File
 
+	// Initial sizings for the console
+	ConsoleWidth  uint16
+	ConsoleHeight uint16
+
 	// Capabilities specify the capabilities to keep when executing the process inside the container
 	// All capabilities not specified will be dropped from the processes capability mask
 	Capabilities *configs.Capabilities
--- a/tests/integration/tty.bats
+++ b/tests/integration/tty.bats
@@ -116,3 +116,60 @@ function teardown() {
 	[[ ${lines[0]} =~ 1000 ]]
 	[[ ${lines[1]} =~ 5 ]]
 }
+
+@test "runc exec [tty consolesize]" {
+	# allow writing to filesystem
+	sed -i 's/"readonly": true/"readonly": false/' config.json
+
+	# run busybox detached
+	runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+	[ "$status" -eq 0 ]
+
+	# make sure we're running
+	testcontainer test_busybox running
+
+	tty_info_with_consize_size=$( cat <<EOF
+{
+    "terminal": true,
+    "consoleSize": {
+	    "height": 10,
+	    "width": 110
+    },
+    "args": [
+	    "/bin/sh",
+	    "-c",
+	    "/bin/stty -a > /tmp/tty-info"
+    ],
+    "cwd": "/"
+}
+EOF
+	)
+
+	# run the exec
+	runc exec --pid-file pid.txt -d --console-socket $CONSOLE_SOCKET -p <( echo $tty_info_with_consize_size ) test_busybox
+	[ "$status" -eq 0 ]
+
+	# check the pid was generated
+	[ -e pid.txt ]
+
+	#wait user process to finish
+	timeout 1 tail --pid=$(head -n 1 pid.txt) -f /dev/null
+
+	tty_info=$( cat <<EOF
+{
+    "args": [
+	"/bin/cat",
+	"/tmp/tty-info"
+    ],
+    "cwd": "/"
+}
+EOF
+	)
+
+	# run the exec
+	runc exec -p <( echo $tty_info ) test_busybox
+	[ "$status" -eq 0 ]
+
+	# test tty width and height against original process.json
+	[[ ${lines[0]} =~ "rows 10; columns 110" ]]
+}
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -108,6 +108,12 @@ func newProcess(p specs.Process) (*libcontainer.Process, error) {
 		NoNewPrivileges: &p.NoNewPrivileges,
 		AppArmorProfile: p.ApparmorProfile,
 	}
+
+	if p.ConsoleSize != nil {
+		lp.ConsoleWidth = uint16(p.ConsoleSize.Width)
+		lp.ConsoleHeight = uint16(p.ConsoleSize.Height)
+	}
+
 	if p.Capabilities != nil {
 		lp.Capabilities = &configs.Capabilities{}
 		lp.Capabilities.Bounding = p.Capabilities.Bounding
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -522,6 +522,8 @@ func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
 		cfg.Rlimits = process.Rlimits
 	}
 	cfg.CreateConsole = process.ConsoleSocket != nil
+	cfg.ConsoleWidth = process.ConsoleWidth
+	cfg.ConsoleHeight = process.ConsoleHeight
 	return cfg
 }
 
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -62,6 +62,8 @@ type initConfig struct {
 	ContainerId      string                `json:"containerid"`
 	Rlimits          []configs.Rlimit      `json:"rlimits"`
 	CreateConsole    bool                  `json:"create_console"`
+	ConsoleWidth     uint16                `json:"console_width"`
+	ConsoleHeight    uint16                `json:"console_height"`
 	Rootless         bool                  `json:"rootless"`
 }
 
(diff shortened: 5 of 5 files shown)


williammartin:
Looks like the build failed with some bad packaging. Can't see how this is anything to do with us - is it possible to kick off again?

cyphar:
I've re-kicked off the CI and the tests now pass.

utils_linux.go
```diff
@@ -81,6 +81,8 @@ func newProcess(p specs.Process) (*libcontainer.Process, error) {
 		Label:           p.SelinuxLabel,
 		NoNewPrivileges: &p.NoNewPrivileges,
 		AppArmorProfile: p.ApparmorProfile,
+		ConsoleWidth:    uint16(p.ConsoleSize.Width),
+		ConsoleHeight:   uint16(p.ConsoleSize.Height),
```
justincormack: This will silently take the lowest 16 bits of the specified value, which could be confusing. I think it is better to pass and store as `int` everywhere and error on console creation if it is too big (in future other platforms could have different limits).

libcontainer/console_linux.go
```diff
@@ -5,18 +5,23 @@ import (
 	"os"
 	"syscall"
 	"unsafe"
+
+	"github.com/docker/docker/pkg/term"
```
justincormack: I think there is a requirement to not depend on `docker/docker` to avoid circular dependencies. May be betetr to copy the one function into this repo.

utils_linux.go
williammartin: Yeh this seems reasonable. Have updated the PR - let me know if you think this is an appropriate solution. Thanks.

libcontainer/console_linux.go
williammartin: Yeh we wondered about this but thought we saw it elsewhere (maybe some legacy that was needing to be removed, or just a mistake). Have just done a little copying and refactoring in the updated PR.

williammartin:
The build failed because of the flake fixed by: https://github.com/opencontainers/runc/pull/1237

```
container_linux.go:247: starting container process caused "process_linux.go:272: running exec setns process for init caused \"exit status 26\""
```

Can someone kick off again please.

caniszczyk:
cc: @crosbymichael since it's on janky

cyphar:
For future reference, anyone inside the `#docker-maintainers` channel can restart the Janky jobs using the message `!rebuild runc#<pr number>`.

crosbymichael:
@williammartin just wondering, why do you have to do a resize at this point in time?  Why can't you resize after receiving the master pty?

williammartin:
@crosbymichael Just to be clear on your question, when you say "receiving the master pty", you mean when the caller of runc receives it, rather than when `runc exec` gets it from `runc init`?

crosbymichael:
When the caller of runc gets it

williammartin:
Well, it's intuitive that if this is specified in the runtime-spec, that the runtime should apply it before the process starts. More importantly, software that expects particular tty sizings should expect it to be set.

I'm wondering if part of the pain here is because `exec` doesn't have a `create/start` split i.e. right now the process might be finished by the time I receive the master pty. I'd need some way to let `runc` know "ok, I've set the tty size, carry on", which feels less intuitive and more complicated than this PR.

crosbymichael:
Well if your issue is with exec that is kinda a separate problem because consoles don't work at all right now because its too fast.

 See https://github.com/opencontainers/runc/issues/1302

williammartin:
While it's true that the console stuff is currently broken, it still seems intuitive that the runtime should set the specification for a caller before the process starts. Do you have some other thoughts on solving that?

crosbymichael:
Yes I agree that it should be set before the process starts.  I think it would be better to solve this how I want to fix the race for the exec case, if we can block until we receive the pty master, resize it, then close that connection causing runc to start the process I think that is a better flow than having to provide console sizes through runc.

craigfurman:
@crosbymichael now that #1302 is resolved, do you have any thoughts on this?

craigfurman:
This PR is a little smaller after our most recent rebase now that runc uses a pty library. Does this change anything?

Craig & @jszroberto

utils_linux.go
```diff
@@ -98,6 +98,13 @@ func getDefaultImagePath(context *cli.Context) string {
 // newProcess returns a new libcontainer Process with the arguments from the
 // spec and stdio from the current process.
 func newProcess(p specs.Process) (*libcontainer.Process, error) {
+	consoleWidth := uint(0)
+	consoleHeight := uint(0)
+	if p.ConsoleSize != nil {
+		consoleWidth = p.ConsoleSize.Width
+		consoleHeight = p.ConsoleSize.Height
+	}
```
cyphar: The defaults of `0` don't make sense given that you unconditionally set the console size when creating it. I don't know what the kernel will do when you tell it to resize a console to 0x0, but from [my quick reading](http://elixir.free-electrons.com/linux/v4.13.4/source/drivers/tty/tty_io.c#L2092) I expect that it isn't a no-op.

cyphar: The defaults of `0` don't make sense given that you unconditionally set the console size when creating it. I don't know what the kernel will do when you tell it to resize a console to 0x0, but from [my quick reading](http://elixir.free-electrons.com/linux/v4.13.4/source/drivers/tty/tty_io.c#L2092) I expect that it isn't a no-op.

libcontainer/process.go
```diff
@@ -47,6 +47,10 @@ type Process struct {
 	// ExtraFiles specifies additional open files to be inherited by the container
 	ExtraFiles []*os.File
 
+	// Initial sizings for the console
+	ConsoleWidth  uint
+	ConsoleHeight uint
```
cyphar: Since we have to use `uint16` in the `Resize` why not just make these `uint16` in the first place?

tests/integration/tty.bats
```diff
@@ -116,3 +116,67 @@ function teardown() {
...
+
+	# run busybox detached
+	runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+	[ "$status" -eq 0 ]
+
+	# check state
+	wait_for_container 15 1 test_busybox
+
+	# make sure we're running
+	testcontainer test_busybox running
```
cyphar: I believe we've stopped using this pattern in tests (`wait_for_container` and `testcontainer`), but I might be mistaken.

tests/integration/tty.bats
```diff
@@ -116,3 +116,67 @@ function teardown() {
...
+{
+	"terminal": true,
+	"consoleSize": {
+		"height": 10,
+		"width": 110
+	},
+	"args": [
+		"/bin/sh",
+		"-c",
+		"/bin/stty -a > $BATS_TMPDIR/tty-info"
```
cyphar: `$BATS_TMPDIR` is on the host not in the container. Since it's generally `/tmp` this won't cause obvious issues, but just use `/tmp` inside the container.

tests/integration/tty.bats
```diff
@@ -116,3 +116,67 @@ function teardown() {
...
+
+	#wait user process to finish
+	timeout 1 tail --pid=$(head -n 1 pid.txt) -f /dev/null
+
+# write a process.json that echoes the tty file info
+cat <<EOF >> $BATS_TMPDIR/read-tty-info.json
+{
+	"args": [
+	    "/bin/cat",
+	    "$BATS_TMPDIR/tty-info"
```
cyphar: Same here.

tests/integration/tty.bats
```diff
@@ -116,3 +116,67 @@ function teardown() {
...
+	# run the exec
+	runc exec -p $BATS_TMPDIR/read-tty-info.json test_busybox
+	[ "$status" -eq 0 ]
+
+	# test tty width and height against original process.json
+	[[ ${lines[0]} =~ "rows 10; columns 110" ]]
+
+	# clean process.jsons
+	rm $BATS_TMPDIR/write-tty-info.json
+	rm $BATS_TMPDIR/read-tty-info.json
```
cyphar: Also you could use process substitution to remove the need for temporary files, but that's not really a big issue.

BooleanCat:
@crosbymichael We have made the suggested changes above, let us know what you think.

Tom and @jszroberto

cyphar:
LGTM. Thanks all.

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1275/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

williammartin:
Thanks all

PR #1048, opened 2016-09-15
Use same state object for state and list

crosbymichael:
Signed-off-by: Michael Crosby <crosbymichael@gmail.com>

PR #1277, opened 2017-01-17
Update state JSON encoding fields to match spec

ijc25:
In #1048 the `cState` struct was consolidated into the `containerState`
struct which inadvertently changed the `bundlePath` and `rootfsPath`
fields in the JSON to `bundle` and `rootfs` respectively.

https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state
requires that the bundle path field be named `bundlePath` so reinstate
that name.

The specification does not cover the rootfs field name, but revert it
too for consistency.

/cc @crosbymichael

### DIFF
--- a/list.go
+++ b/list.go
@@ -31,9 +31,9 @@ type containerState struct {
 	// Status is the current status of the container, running, paused, ...
 	Status string `json:"status"`
 	// Bundle is the path on the filesystem to the bundle
-	Bundle string `json:"bundle"`
+	Bundle string `json:"bundlePath"`
 	// Rootfs is a path to a directory containing the container's root filesystem.
-	Rootfs string `json:"rootfs"`
+	Rootfs string `json:"rootfsPath"`
 	// Created is the unix timestamp for the creation time of the container in UTC
 	Created time.Time `json:"created"`
 	// Annotations is the user defined annotations added to the config.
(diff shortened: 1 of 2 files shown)


hqhq:
CI failed.

WeiZhang555:
I was also planning to do this but got no time, my idea is 

```
type containerState struct {
    specs.State

    Created time.Time `json:"created"`
}
```

my idea is we should *Reuse* `specs.State` and *Remove* `Rootfs`, because you can actually get rootfs path from `bundle/config.json`.

What do you think?

ijc25:
@hqhq hqhq `make test` failed for me locally both before and after my change so I missed the new error I introduced. Sorry, I should really have debugged the baseline failure first so I could have spotted the new problem.

@WeiZhang555 well, it sounds sensible to me, but I'm pretty new to this codebase/project...

ijc25:
While debugging the CI failure I realised that this also changes the output of `runc list --format json`. I'm unsure if that is desirable or not.

WeiZhang555:
@ijc25 Well, it will break the backward compatibility without doubt, but I think we'd better do it whatever, but it didn't follow the spec, it will diverge further with spec if we don't do it now.

crosbymichael:
I think i'd rather change the spec to `bundle` because the path part is just redundant

crosbymichael:
We updated the spec to just have `bundle` so we don't have to make this change and removes the backwards compat concerns.  

Thanks for the PR.

PR #1278, opened 2017-01-17
move error check out of the for loop

datawolf:
The `bufio.Scanner.Scan` method returns false either by reaching the
end of the input or an error. After Scan returns false, the Err method
will return any error that occurred during scanning, except that if it
was io.EOF, Err will return nil.

We should check the error when Scan return false(out of the for loop).

refer: [https://golang.org/pkg/bufio/#Scanner.Scan](https://golang.org/pkg/bufio/#Scanner.Scan)


mrunalp:
LGTM

dqminh:
LGTM

PR #1279, opened 2017-01-17
libcontainer: add support for Intel RDT/CAT in runc

xiaochenshen:
**v5 commits:**
4d2756c116315cd30b7ded82dfbb4bca78ceb627 libcontainer: add test cases for Intel RDT/CAT
692f6e1e27b2e4d38e1c9e0b463f7d1bfe3f7492 libcontainer: add support for Intel RDT/CAT in runc
af3b0d9dce34fb0e91c58d823d0ad48268b0f988 libcontainer/SPEC.md: add documentation for Intel RDT/CAT

**Changes in v5:**
Reworked code according to @crosbymichael and @hqhq's comments:
1. https://github.com/opencontainers/runc/pull/1279#issuecomment-319085278
2. https://github.com/opencontainers/runc/pull/1279#issuecomment-320008325
3. https://github.com/opencontainers/runc/pull/1279#issuecomment-319917615

- Obsoleted the design of "refactor resource manager interface": removed generic resourceManagers, isolated IntelRdt functions without touching cgroups code, removed the static map with static keys on resourceManagers set.
- Simplified IntelRdtManager implementation, removed "dead and empty" methods - GetPids(), GetAllPids(), Freeze(). Kept the required methods set - Apply(), GetStats(), Destroy(), GetPath(), and Set().
- Re-implemented GetStats(), isolated cgroups.Stats and intelrdt.Stats and kept type safety on the Stats.
- Changed test stuffs according to the functional rework.
- Added read-only L3 cache info (CbmMask, MinCbmBits, and NumClosids) into Intelrdt.Stats. Runc users could run "runc events" to fetch the information, which helps know more about L3 cache capabilities and how to set correct L3 cache schema in different Intel Xeon platforms.

---

**v4 commits**:
b1c836632b1d426a66b0e39abbac3df176177bd2 libcontainer: add support for Intel RDT/CAT in runc
48d8ffe6b84a8d647cf1622a0b02a544752ac791 libcontainer/SPEC.md: add documentation for Intel RDT/CAT
8851a0db77a65b42f72e6711356dd401204ad436 libcontainer: refactor resource manager interface
80648193d0add2b155844e558634b41eabb9116f vendor: specs-go: update specs for Intel RDT/CAT

**Changes in v4:**
Rebased to latest master branch.

---

**v3 commits**:
be25a2030bec8113434e53e0ee1b99a6e0bfc7ed libcontainer: add support for Intel RDT/CAT in runc
96fd05fc7ab0bcca414efc3e9e4894dbe1df0e5d libcontainer/SPEC.md: add documentation for Intel RDT/CAT
d5ac70d5ae68b514421a7c7f60475189191c70f1 libcontainer: refactor resource manager interface
c302b704bedd62e7655a401697f1a4a79085bcc0 vendor: specs-go: update specs for Intel RDT/CAT

**Changes in v3:**
Addressed comment https://github.com/opencontainers/runc/pull/1279#discussion_r116414850 from @yummypeng :
- Add parameter `path` to record IntelRdtPath in NewIntelRdtManager.

---

**v2 commits (for keeping records in code review):**
ab32d21313aa82dd6c88b24c1bfcbaf8a970dcea libcontainer: add support for Intel RDT/CAT in runc
5972fd56b5f59444a1a7260c84f915515622e297 libcontainer/SPEC.md: add documentation for Intel RDT/CAT
7f8b32107131150a581b62ad0f2665c8a03abbf1 libcontainer: refactor resource manager interface
9b03a51cbfaad7d0a2afe7c9667ac21d1799dcb0 vendor: specs-go: update specs for Intel RDT/CAT

**Changes in v2:**
Addressed the comments from @hqhq and @yummypeng :
- Added intelrdt config validator.
- Removed resctrl filesystem re-mount logic in isIntelRdtMounted().
- Optimized IsIntelRdtEnabled() to avoid unnecessary time-consuming mount points parsing.
- Changed Intel RDT related json items to snake_case style.
- Fixed some redundant codes with regard to intelRdtPath.
- Fixed potential null pointer issue in IntelRdtManager.Set().
- Fixed some typos (Sets, num_closids and etc.)
- Improved libcontainer/SPEC.md: added minimal supported kernel version, provided a more sophisticated config example.

---

**v1 commits (for keeping records in code review):**
4752fd2f639a38eb6ef6c2e030252098cd11d9f0 libcontainer: add support for Intel RDT/CAT in runc
c8486a7a8866b5ce4264ee0a17e54292a0f56fbf libcontainer/SPEC.md: add documentation for Intel RDT/CAT
aefd02ae210105573dabbb90ace3ab70943504ce libcontainer: refactor resource manager interface
8e8e5d96e6ad7dfefcd0f7d49a8d8c0b5cba0ef6 vendor: specs-go: update specs for Intel RDT/CAT

---

This PR fixes issue #433 

About Intel RDT/CAT feature:
Intel platforms with new Xeon CPU support Intel Resource Director Technology
(RDT). Cache Allocation Technology (CAT) is a sub-feature of RDT, which
currently supports L3 cache resource allocation.

This feature provides a way for the software to restrict cache allocation to a
defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
The different subsets are identified by class of service (CLOS) and each CLOS
has a capacity bitmask (CBM).

For more information about Intel RDT/CAT can be found in the section 17.17
of Intel Software Developer Manual.

About Intel RDT/CAT kernel interface:
In Linux 4.10 kernel or newer, the interface is defined and exposed via
"resource control" filesystem, which is a "cgroup-like" interface.

Comparing with cgroups, it has similar process management lifecycle and
interfaces in a container. But unlike cgroups' hierarchy, it has single level
filesystem layout.

Intel RDT "resource control" filesystem hierarchy:
```
mount -t resctrl resctrl /sys/fs/resctrl
tree /sys/fs/resctrl
/sys/fs/resctrl/
|-- info
|   |-- L3
|       |-- cbm_mask
|       |-- min_cbm_bits
|       |-- num_closids
|-- cpus
|-- schemata
|-- tasks
|-- <container_id>
    |-- cpus
    |-- schemata
    |-- tasks
```
For runc, we can make use of `tasks` and `schemata` configuration for L3 cache
resource constraints.

The file `tasks` has a list of tasks that belongs to this group (e.g.,
<container_id>" group). Tasks can be added to a group by writing the task ID
to the "tasks" file  (which will automatically remove them from the previous
group to which they belonged). New tasks created by fork(2) and clone(2) are
added to the same group as their parent. If a pid is not in any sub group, it
Is in root group.

The file `schemata` has allocation bitmasks/values for L3 cache on each socket,
which contains L3 cache id and capacity bitmask (CBM).
```
	Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
```
For example, on a two-socket machine, L3's schema line could be `L3:0=ff;1=c0`
which means L3 cache id 0's CBM is 0xff, and L3 cache id 1's CBM is 0xc0.

The valid L3 cache CBM is a *contiguous bits set* and number of bits that can
be set is less than the max bit. The max bits in the CBM is varied among
supported Intel Xeon platforms. In Intel RDT "resource control" filesystem
layout, the CBM in a group should be a subset of the CBM in root. Kernel will
check if it is valid when writing. e.g., 0xfffff in root indicates the max bits
of CBM is 20 bits, which mapping to entire L3 cache capacity. Some valid CBM
values to set in a group: 0xf, 0xf0, 0x3ff, 0x1f00 and etc.

For more information about Intel RDT/CAT kernel interface:  
https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt

An example for runc:
```
Consider a two-socket machine with two L3 caches where the default CBM is
0xfffff and the max CBM length is 20 bits. With this configuration, tasks
inside the container only have access to the "upper" 80% of L3 cache id 0 and
the "lower" 50% L3 cache id 1:

"linux": {
	"intelRdt": {
		"l3CacheSchema": "L3:0=ffff0;1=3ff"
	}
}
```

### DIFF
--- a/events.go
+++ b/events.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/intelrdt"
 
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -25,11 +26,12 @@ type event struct {
 
 // stats is the runc specific stats structure for stability when encoding and decoding stats.
 type stats struct {
-	CPU     cpu                `json:"cpu"`
-	Memory  memory             `json:"memory"`
-	Pids    pids               `json:"pids"`
-	Blkio   blkio              `json:"blkio"`
-	Hugetlb map[string]hugetlb `json:"hugetlb"`
+	CPU      cpu                `json:"cpu"`
+	Memory   memory             `json:"memory"`
+	Pids     pids               `json:"pids"`
+	Blkio    blkio              `json:"blkio"`
+	Hugetlb  map[string]hugetlb `json:"hugetlb"`
+	IntelRdt intelRdt           `json:"intel_rdt"`
 }
 
 type hugetlb struct {
@@ -94,6 +96,23 @@ type memory struct {
 	Kernel    memoryEntry       `json:"kernel,omitempty"`
 	KernelTCP memoryEntry       `json:"kernelTCP,omitempty"`
 	Raw       map[string]uint64 `json:"raw,omitempty"`
+}
+
+type l3CacheInfo struct {
+	CbmMask    string `json:"cbm_mask,omitempty"`
+	MinCbmBits uint64 `json:"min_cbm_bits,omitempty"`
+	NumClosids uint64 `json:"num_closids,omitempty"`
+}
+
+type intelRdt struct {
+	// The read-only L3 cache information
+	L3CacheInfo *l3CacheInfo `json:"l3_cache_info,omitempty"`
+
+	// The read-only L3 cache schema in root
+	L3CacheSchemaRoot string `json:"l3_cache_schema_root,omitempty"`
+
+	// The L3 cache schema in 'container_id' group
+	L3CacheSchema string `json:"l3_cache_schema,omitempty"`
 }
 
 var eventsCommand = cli.Command{
@@ -227,6 +246,13 @@ func convertLibcontainerStats(ls *libcontainer.Stats) *stats {
 	for k, v := range cg.HugetlbStats {
 		s.Hugetlb[k] = convertHugtlb(v)
 	}
+
+	if is := ls.IntelRdtStats; is != nil {
+		s.IntelRdt.L3CacheInfo = convertL3CacheInfo(is.L3CacheInfo)
+		s.IntelRdt.L3CacheSchemaRoot = is.L3CacheSchemaRoot
+		s.IntelRdt.L3CacheSchema = is.L3CacheSchema
+	}
+
 	return &s
 }
 
@@ -258,4 +284,12 @@ func convertBlkioEntry(c []cgroups.BlkioStatEntry) []blkioEntry {
 		})
 	}
 	return out
+}
+
+func convertL3CacheInfo(i *intelrdt.L3CacheInfo) *l3CacheInfo {
+	return &l3CacheInfo{
+		CbmMask:    i.CbmMask,
+		MinCbmBits: i.MinCbmBits,
+		NumClosids: i.NumClosids,
+	}
 }
--- a/libcontainer/SPEC.md
+++ b/libcontainer/SPEC.md
@@ -154,6 +154,90 @@ that no processes or threads escape the cgroups.  This sync is
 done via a pipe ( specified in the runtime section below ) that the container's
 init process will block waiting for the parent to finish setup.
 
+### IntelRdt
+
+Intel platforms with new Xeon CPU support Intel Resource Director Technology
+(RDT). Cache Allocation Technology (CAT) is a sub-feature of RDT, which
+currently supports L3 cache resource allocation.
+
+This feature provides a way for the software to restrict cache allocation to a
+defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
+The different subsets are identified by class of service (CLOS) and each CLOS
+has a capacity bitmask (CBM).
+
+It can be used to handle L3 cache resource allocation for containers if
+hardware and kernel support Intel RDT/CAT.
+
+In Linux 4.10 kernel or newer, the interface is defined and exposed via
+"resource control" filesystem, which is a "cgroup-like" interface.
+
+Comparing with cgroups, it has similar process management lifecycle and
+interfaces in a container. But unlike cgroups' hierarchy, it has single level
+filesystem layout.
+
+Intel RDT "resource control" filesystem hierarchy:
+```
+mount -t resctrl resctrl /sys/fs/resctrl
+tree /sys/fs/resctrl
+/sys/fs/resctrl/
+|-- info
+|   |-- L3
+|       |-- cbm_mask
+|       |-- min_cbm_bits
+|       |-- num_closids
+|-- cpus
+|-- schemata
+|-- tasks
+|-- <container_id>
+    |-- cpus
+    |-- schemata
+    |-- tasks
+
+```
+
+For runc, we can make use of `tasks` and `schemata` configuration for L3 cache
+resource constraints.
+
+The file `tasks` has a list of tasks that belongs to this group (e.g.,
+<container_id>" group). Tasks can be added to a group by writing the task ID
+to the "tasks" file  (which will automatically remove them from the previous
+group to which they belonged). New tasks created by fork(2) and clone(2) are
+added to the same group as their parent. If a pid is not in any sub group, it
+is in root group.
+
+The file `schemata` has allocation masks/values for L3 cache on each socket,
+which contains L3 cache id and capacity bitmask (CBM).
+```
+	Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
+```
+For example, on a two-socket machine, L3's schema line could be `L3:0=ff;1=c0`
+Which means L3 cache id 0's CBM is 0xff, and L3 cache id 1's CBM is 0xc0.
+
+The valid L3 cache CBM is a *contiguous bits set* and number of bits that can
+be set is less than the max bit. The max bits in the CBM is varied among
+supported Intel Xeon platforms. In Intel RDT "resource control" filesystem
+layout, the CBM in a group should be a subset of the CBM in root. Kernel will
+check if it is valid when writing. e.g., 0xfffff in root indicates the max bits
+of CBM is 20 bits, which mapping to entire L3 cache capacity. Some valid CBM
+values to set in a group: 0xf, 0xf0, 0x3ff, 0x1f00 and etc.
+
+For more information about Intel RDT/CAT kernel interface:  
+https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt
+
+An example for runc:
+```
+Consider a two-socket machine with two L3 caches where the default CBM is
+0xfffff and the max CBM length is 20 bits. With this configuration, tasks
+inside the container only have access to the "upper" 80% of L3 cache id 0 and
+the "lower" 50% L3 cache id 1:
+
+"linux": {
+	"intelRdt": {
+		"l3CacheSchema": "L3:0=ffff0;1=3ff"
+	}
+}
+```
+
 ### Security 
 
 The standard set of Linux capabilities that are set in a container
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	selinux "github.com/opencontainers/selinux/go-selinux"
 )
 
@@ -38,6 +39,9 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 		return err
 	}
 	if err := v.sysctl(config); err != nil {
+		return err
+	}
+	if err := v.intelrdt(config); err != nil {
 		return err
 	}
 	if config.Rootless {
@@ -148,6 +152,19 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 			}
 		}
 		return fmt.Errorf("sysctl %q is not in a separate kernel namespace", s)
+	}
+
+	return nil
+}
+
+func (v *ConfigValidator) intelrdt(config *configs.Config) error {
+	if config.IntelRdt != nil {
+		if !intelrdt.IsIntelRdtEnabled() {
+			return fmt.Errorf("intelRdt is specified in config, but Intel RDT feature is not supported or enabled")
+		}
+		if config.IntelRdt.L3CacheSchema == "" {
+			return fmt.Errorf("intelRdt is specified in config, but intelRdt.l3CacheSchema is empty")
+		}
 	}
 
 	return nil
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -21,6 +21,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/criurpc"
+	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runc/libcontainer/utils"
 
@@ -38,6 +39,7 @@ type linuxContainer struct {
 	root                 string
 	config               *configs.Config
 	cgroupManager        cgroups.Manager
+	intelRdtManager      intelrdt.Manager
 	initArgs             []string
 	initProcess          parentProcess
 	initProcessStartTime uint64
@@ -67,6 +69,9 @@ type State struct {
 
 	// Container's standard descriptors (std{in,out,err}), needed for checkpoint and restore
 	ExternalDescriptors []string `json:"external_descriptors,omitempty"`
+
+	// Intel RDT "resource control" filesystem path
+	IntelRdtPath string `json:"intel_rdt_path"`
 }
 
 // Container is a libcontainer container object.
@@ -163,6 +168,11 @@ func (c *linuxContainer) Stats() (*Stats, error) {
 	if stats.CgroupStats, err = c.cgroupManager.GetStats(); err != nil {
 		return stats, newSystemErrorWithCause(err, "getting container stats from cgroups")
 	}
+	if c.intelRdtManager != nil {
+		if stats.IntelRdtStats, err = c.intelRdtManager.GetStats(); err != nil {
+			return stats, newSystemErrorWithCause(err, "getting container's Intel RDT stats")
+		}
+	}
 	for _, iface := range c.config.Networks {
 		switch iface.Type {
 		case "veth":
@@ -192,6 +202,15 @@ func (c *linuxContainer) Set(config configs.Config) error {
 			logrus.Warnf("Setting back cgroup configs failed due to error: %v, your state.json and actual configs might be inconsistent.", err2)
 		}
 		return err
+	}
+	if c.intelRdtManager != nil {
+		if err := c.intelRdtManager.Set(&config); err != nil {
+			// Set configs back
+			if err2 := c.intelRdtManager.Set(c.config); err2 != nil {
+				logrus.Warnf("Setting back intelrdt configs failed due to error: %v, your state.json and actual configs might be inconsistent.", err2)
+			}
+			return err
+		}
 	}
 	// After config setting succeed, update config and states
 	c.config = &config
@@ -434,15 +453,16 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
 		return nil, err
 	}
 	return &initProcess{
-		cmd:           cmd,
-		childPipe:     childPipe,
-		parentPipe:    parentPipe,
-		manager:       c.cgroupManager,
-		config:        c.newInitConfig(p),
-		container:     c,
-		process:       p,
-		bootstrapData: data,
-		sharePidns:    sharePidns,
+		cmd:             cmd,
+		childPipe:       childPipe,
+		parentPipe:      parentPipe,
+		manager:         c.cgroupManager,
+		intelRdtManager: c.intelRdtManager,
+		config:          c.newInitConfig(p),
+		container:       c,
+		process:         p,
+		bootstrapData:   data,
+		sharePidns:      sharePidns,
 	}, nil
 }
 
@@ -461,6 +481,7 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, 
 	return &setnsProcess{
 		cmd:           cmd,
 		cgroupPaths:   c.cgroupManager.GetPaths(),
+		intelRdtPath:  state.IntelRdtPath,
 		childPipe:     childPipe,
 		parentPipe:    parentPipe,
 		config:        c.newInitConfig(p),
@@ -1519,6 +1540,10 @@ func (c *linuxContainer) currentState() (*State, error) {
 		startTime, _ = c.initProcess.startTime()
 		externalDescriptors = c.initProcess.externalDescriptors()
 	}
+	intelRdtPath, err := intelrdt.GetIntelRdtPath(c.ID())
+	if err != nil {
+		intelRdtPath = ""
+	}
 	state := &State{
 		BaseState: BaseState{
 			ID:                   c.ID(),
@@ -1529,6 +1554,7 @@ func (c *linuxContainer) currentState() (*State, error) {
 		},
 		Rootless:            c.config.Rootless,
 		CgroupPaths:         c.cgroupManager.GetPaths(),
+		IntelRdtPath:        intelRdtPath,
 		NamespacePaths:      make(map[configs.NamespaceType]string),
 		ExternalDescriptors: externalDescriptors,
 	}
--- a/libcontainer/container_linux_test.go
+++ b/libcontainer/container_linux_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/system"
 )
 
@@ -17,6 +18,11 @@ type mockCgroupManager struct {
 	allPids []int
 	stats   *cgroups.Stats
 	paths   map[string]string
+}
+
+type mockIntelRdtManager struct {
+	stats *intelrdt.Stats
+	path  string
 }
 
 func (m *mockCgroupManager) GetPids() ([]int, error) {
@@ -48,6 +54,26 @@ func (m *mockCgroupManager) GetPaths() map[string]string {
 }
 
 func (m *mockCgroupManager) Freeze(state configs.FreezerState) error {
+	return nil
+}
+
+func (m *mockIntelRdtManager) Apply(pid int) error {
+	return nil
+}
+
+func (m *mockIntelRdtManager) GetStats() (*intelrdt.Stats, error) {
+	return m.stats, nil
+}
+
+func (m *mockIntelRdtManager) Destroy() error {
+	return nil
+}
+
+func (m *mockIntelRdtManager) GetPath() string {
+	return m.path
+}
+
+func (m *mockIntelRdtManager) Set(container *configs.Config) error {
 	return nil
 }
 
@@ -118,6 +144,11 @@ func TestGetContainerStats(t *testing.T) {
 				},
 			},
 		},
+		intelRdtManager: &mockIntelRdtManager{
+			stats: &intelrdt.Stats{
+				L3CacheSchema: "L3:0=f;1=f0",
+			},
+		},
 	}
 	stats, err := container.Stats()
 	if err != nil {
@@ -129,13 +160,22 @@ func TestGetContainerStats(t *testing.T) {
 	if stats.CgroupStats.MemoryStats.Usage.Usage != 1024 {
 		t.Fatalf("expected memory usage 1024 but recevied %d", stats.CgroupStats.MemoryStats.Usage.Usage)
 	}
+	if intelrdt.IsIntelRdtEnabled() {
+		if stats.IntelRdtStats == nil {
+			t.Fatal("intel rdt stats are nil")
+		}
+		if stats.IntelRdtStats.L3CacheSchema != "L3:0=f;1=f0" {
+			t.Fatalf("expected L3CacheSchema L3:0=f;1=f0 but recevied %s", stats.IntelRdtStats.L3CacheSchema)
+		}
+	}
 }
 
 func TestGetContainerState(t *testing.T) {
 	var (
-		pid                 = os.Getpid()
-		expectedMemoryPath  = "/sys/fs/cgroup/memory/myid"
-		expectedNetworkPath = fmt.Sprintf("/proc/%d/ns/net", pid)
+		pid                  = os.Getpid()
+		expectedMemoryPath   = "/sys/fs/cgroup/memory/myid"
+		expectedNetworkPath  = fmt.Sprintf("/proc/%d/ns/net", pid)
+		expectedIntelRdtPath = "/sys/fs/resctrl/myid"
 	)
 	container := &linuxContainer{
 		id: "myid",
@@ -166,6 +206,12 @@ func TestGetContainerState(t *testing.T) {
 				"memory": expectedMemoryPath,
 			},
 		},
+		intelRdtManager: &mockIntelRdtManager{
+			stats: &intelrdt.Stats{
+				L3CacheSchema: "L3:0=f0;1=f",
+			},
+			path: expectedIntelRdtPath,
+		},
 	}
 	container.state = &createdState{c: container}
 	state, err := container.State()
@@ -184,6 +230,15 @@ func TestGetContainerState(t *testing.T) {
 	}
 	if memPath := paths["memory"]; memPath != expectedMemoryPath {
 		t.Fatalf("expected memory path %q but received %q", expectedMemoryPath, memPath)
+	}
+	if intelrdt.IsIntelRdtEnabled() {
+		intelRdtPath := state.IntelRdtPath
+		if intelRdtPath == "" {
+			t.Fatal("intel rdt path should not be empty")
+		}
+		if intelRdtPath != expectedIntelRdtPath {
+			t.Fatalf("expected intel rdt path %q but received %q", expectedIntelRdtPath, intelRdtPath)
+		}
 	}
 	for _, ns := range container.config.Namespaces {
 		path := state.NamespacePaths[ns.Type]
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -18,6 +18,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/configs/validate"
+	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/utils"
 
 	"golang.org/x/sys/unix"
@@ -81,6 +82,20 @@ func RootlessCgroups(l *LinuxFactory) error {
 		return &rootless.Manager{
 			Cgroups: config,
 			Paths:   paths,
+		}
+	}
+	return nil
+}
+
+// IntelRdtfs is an options func to configure a LinuxFactory to return
+// containers that use the Intel RDT "resource control" filesystem to
+// create and manage Intel Xeon platform shared resources (e.g., L3 cache).
+func IntelRdtFs(l *LinuxFactory) error {
+	l.NewIntelRdtManager = func(config *configs.Config, id string, path string) intelrdt.Manager {
+		return &intelrdt.IntelRdtManager{
+			Config: config,
+			Id:     id,
+			Path:   path,
 		}
 	}
 	return nil
@@ -150,6 +165,9 @@ type LinuxFactory struct {
 
 	// NewCgroupsManager returns an initialized cgroups manager for a single container.
 	NewCgroupsManager func(config *configs.Cgroup, paths map[string]string) cgroups.Manager
+
+	// NewIntelRdtManager returns an initialized Intel RDT manager for a single container.
+	NewIntelRdtManager func(config *configs.Config, id string, path string) intelrdt.Manager
 }
 
 func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, error) {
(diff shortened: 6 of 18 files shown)


xiaochenshen:
@cyphar @crosbymichael @hqhq @mrunalp @vishh

This PR will obsolete #1198 

To address @crosbymichael and @cyphar 's comments https://github.com/opencontainers/runc/pull/1198#issuecomment-263357031 and https://github.com/opencontainers/runc/pull/1198#issuecomment-263568672, the design is updated:

It adds a new "ResourceManager" structure as the base interface for all resource managers, such as cgroups manager and incoming IntelRdt manager.

All registered resource managers are consolidated in linuxContainer structure. We can apply to unified operations (e.g., Apply(), Set(), Destroy()) using all of the registered resource managers.

Currently, cgroups manager is the single resource manager in libcontainer. Linux kernel 4.10 will introduce Intel RDT/CAT feature, the kernel interface is exposed via "resource control" filesystem, which is a cgroup-like interface. In order to support Intel RDT/CAT in libcontainer, we need a new resource manager (IntelRdt manager) outside cgroups.

crosbymichael:
Thanks.  Design looks much better than before.

xiaochenshen:
Any comments for this PR? Thank you.

xiaochenshen:
@LK4D4 @cyphar @crosbymichael @hqhq @mrunalp @vishh 

Who can help forward me the simplest steps to build runc in my forked repo (github.com/xiaochenshen/runc)? Thank you.

When I tried to rebase the code against latest runc master branch to resolve the conflict, I found the Godeps is replaced by vndr tool recently. I have tried to build with "make" in github.com/opencontainers/runc, it works. But in my forked dir, I got following errors. It looks like vndr related issues.
```
# make
go build -i -ldflags "-X main.gitCommit="291bf601105c97dc1aa631fdfb7fca63c947319b" -X main.version=1.0.0-rc2" -tags "seccomp" -o runc .
# github.com/xiaochenshen/runc
./restore.go:111: cannot use spec (type *"github.com/xiaochenshen/runc/vendor/github.com/opencontainers/runtime-spec/specs-go".Spec) as type *"github.com/opencontainers/runc/vendor/github.com/opencontainers/runtime-spec/specs-go".Spec in field value
./utils_linux.go:154: cannot use spec (type *"github.com/xiaochenshen/runc/vendor/github.com/opencontainers/runtime-spec/specs-go".Spec) as type *"github.com/opencontainers/runc/vendor/github.com/opencontainers/runtime-spec/specs-go".Spec in field value
make: *** [runc] Error 2
```

cyphar:
@xiaochenshen Make sure that this project is cloned properly inside your `GOPATH`. In particular you should have something like this set up:

```bash
% mkdir ~/yourgopath
% export GOPATH=~/yourgopath
% mkdir -p $GOPATH/src/github.com/opencontainers/runc
% git clone https://github.com/opencontainers/runc $GOPATH/src/github.com/opencontainers/runc
# ...
% cd $GOPATH/src/github.com/opencontainers/runc
% make
# should work
```

xiaochenshen:
@cyphar Thank you for your guide. It works with github.com/opencontainers/runc.
But this issue happens only in my forked repo github.com/xiaochenshen/runc which is also under $GOPATH ($GOPATH/src/github.com/xiaochenshen/runc).
Should I modify vendor.conf file?

Or git clone my repo `https://github.com/xiaochen/runc` into the directory $GOPATH/src/github.com/opencontainers/runc? It looks weird, to some extent:
```
git clone https://github.com/xiaochen/runc $GOPATH/src/github.com/opencontainers/runc
```

xiaochenshen:
@crosbymichael /cc @opencontainers/runc-maintainers

To address your suggestion in https://github.com/opencontainers/runtime-spec/pull/630#pullrequestreview-25906691:
`IntelRdt` struct has been moved out of `Resources` and inside `Linux` now.

This PR has been updated accordingly.

xiaochenshen:
ping @crosbymichael @cyphar @mrunalp @hqhq 
/cc @rjnagal @vmarmol @dqminh @avagin

https://github.com/opencontainers/runtime-spec/pull/630 + this PR to fix issue #433. And https://github.com/opencontainers/runtime-spec/pull/630 has been merged.
 
Could you help code review this PR at your convenience? Thank you!
I just rebased the code to fix conflict with rootless cgroup manager

xiaochenshen:
@yummypeng Thank you for your help on code review. I will check up the issues one by one later.

yummypeng:
I tried to add the configuration of intelRdt in my `config.json`, then run a container in an environment without support of intelRdt, and unexpectedly, the container was started successfully. Seems there needs an validation of `Spec` maybe in https://github.com/opencontainers/runc/blob/master/libcontainer/configs/validate/validator.go#L24.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,556 @@
...
+	// The valid L3 cache CBM is a *contiguous bits set* and number of
+	// bits that can be set is less than the max bit. The max bits in the
+	// CBM is varied among supported Intel Xeon platforms. In Intel RDT
+	// "resource control" filesystem layout, the CBM in a group should
+	// be a subset of the CBM in root. Kernel will check if it is valid
+	// when writing.
+	// e.g., 0xfffff in root indicates the max bits of CBM is 20 bits,
+	// which mapping to entire L3 cache capacity. Some valid CBM values
+	// to set in a group: 0xf, 0xf0, 0x3ff, 0x1f00 and etc.
+	l3CacheSchema := container.IntelRdt.L3CacheSchema
```
yummypeng: This requires a check if `container.IntelRdt == nil`, otherwise this will panic.

yummypeng:
Since `resctrl` is a "cgroup-like" interface, why not mount it into container like cgroup does?

libcontainer/intelrdt/intelrdt.go
xiaochenshen: Yes, you are right. I will add check in case of null pointer.

xiaochenshen:
@yummypeng 
> Since resctrl is a "cgroup-like" interface, why not mount it into container like cgroup does?

There are discussions about that in obsolete #1198. You can refer to @crosbymichael and @cyphar 's comments: https://github.com/opencontainers/runc/pull/1198#issuecomment-263357031 and https://github.com/opencontainers/runc/pull/1198#issuecomment-263568672.

libcontainer/resourcemanager/resource_manager_linux.go
```diff
@@ -0,0 +1,34 @@
...
+	Freeze(state configs.FreezerState) error
+
+	// Destroys the resource set
+	Destroy() error
+
+	// Returns resource paths to save in a state file and to be able to
+	// restore the object later
+	GetPaths() map[string]string
+
+	// Set the resource as configured
```
xiaochenshen: Thanks for finding the typo.

libcontainer/container_linux.go
```diff
@@ -432,9 +451,14 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe,
 	if err != nil {
 		return nil, err
 	}
+	intelRdtPath, err := intelrdt.GetIntelRdtPath(c.ID())
```
xiaochenshen: Yes, you are right. I will fix it.

libcontainer/container_linux_test.go
```diff
@@ -184,6 +256,15 @@ func TestGetContainerState(t *testing.T) {
 	if memPath := paths["memory"]; memPath != expectedMemoryPath {
 		t.Fatalf("expected memory path %q but received %q", expectedMemoryPath, memPath)
 	}
+	if intelrdt.IsIntelRdtEnabled() {
+		path := state.IntelRdtPath
+		if path == "" {
+			t.Fatal("intel rdt path should not be empty")
+		}
+		if intelRdtPath := path; intelRdtPath != expectedIntelRdtPath {
```
xiaochenshen: Yes, I will fix it.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,556 @@
...
+ * 	}
+ * }
+ */
+
+type Manager interface {
+	resourcemanager.ResourceManager
+
+	// Returns Intel RDT "resource control" filesystem path to save in
+	// a state file and to be able to restore the object later
+	GetPath() string
```
xiaochenshen: Currently I kept GetPaths() here intentionally. In future, we may add more Intel RDT resource constrains besides L3 cache. If it is really not needed at that time, I will reorg the `Manager` interface just like you suggested.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,556 @@
...
+}
+
+func isIntelRdtMounted() bool {
+	_, err := getIntelRdtRoot()
+	if err != nil {
+		if !IsNotFound(err) {
+			return false
+		}
+
+		// If not mounted, we try to mount again:
```
xiaochenshen: I agree with you now. I will remove this.

I was thinking that it is no harm for it. Unlike cgroups, Intel RDT resource control filesystem will likely not be mounted during boot-up by default in popular Linux OS distributions.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,556 @@
...
+	path := filepath.Join(rootPath, "info", "l3")
+	cbmMask, err := getIntelRdtParamUint(path, "cbm_mask")
+	if err != nil {
+		return nil, err
+	}
+	minCbmBits, err := getIntelRdtParamUint(path, "min_cbm_bits")
+	if err != nil {
+		return nil, err
+	}
+	numClosid, err := getIntelRdtParamUint(path, "num_closid")
```
xiaochenshen: Thanks for finding this typo.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,556 @@
...
+// Applies configuration to the process with the specified pid
+func (m *IntelRdtManager) Apply(pid int) (err error) {
+	d, err := getIntelRdtData(m.Config, pid)
+	if err != nil && !IsNotFound(err) {
+		return err
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	path, err := d.join(m.Id)
```
xiaochenshen: Do you mean the lock is not necessary or the lock here will cause a deadlock?

libcontainer/resourcemanager/resource_manager_linux.go
```diff
@@ -0,0 +1,34 @@
+// +build linux
```
hqhq: Is this gonna break Solaris and Windows?

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,556 @@
...
+		if err := writeFile(path, "schemata", l3CacheSchema); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (raw *intelRdtData) join(id string) (string, error) {
+	path := filepath.Join(raw.root, id)
```
hqhq: What about nested container cases? You can refer to the cgroup `join` function to see how cgroup handle this.

events.go
```diff
@@ -95,6 +96,12 @@ type memory struct {
 	Raw       map[string]uint64 `json:"raw,omitempty"`
 }
 
+type intelRdt struct {
+	// The read-only default "schema" in root, for reference
+	L3CacheSchemaRoot string `json:"l3CacheSchemaRoot,omitempty"`
+	L3CacheSchema     string `json:"l3CacheSchema,omitempty"`
```
hqhq: json variables should be snake_case.

libcontainer/SPEC.md
```diff
@@ -154,6 +154,89 @@ that no processes or threads escape the cgroups.  This sync is
...
+This feature provides a way for the software to restrict cache allocation to a
+defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
+The different subsets are identified by class of service (CLOS) and each CLOS
+has a capacity bitmask (CBM).
+
+It can be used to handle L3 cache resource allocation for containers if
+hardware and kernel support Intel RDT/CAT.
+
+In Linux kernel, it is exposed via "resource control" filesystem, which is a
+"cgroup-like" interface.
```
hqhq: Can you specify the minimum linux kernel version that supports intelrdt feature?

hqhq:
@xiaochenshen Thanks for working on this, overall it looks good, can you address all the comments and rebase please? thanks.

events.go
xiaochenshen: Got it. I will fix it.

libcontainer/SPEC.md
xiaochenshen: No problem - 4.10 kernel.

```+Linux kernel 4.10 introduces Intel RDT/CAT support, it is exposed via "resource control" filesystem...```

File: libcontainer/intelrdt/intelrdt.go [hunk 13]
```diff
@@ -0,0 +1,556 @@
...
+}
+
+// Check if Intel RDT is enabled
+func IsIntelRdtEnabled() bool {
+	// 1. check if hardware and kernel support Intel RDT feature
+	// "rdt" flag is set if supported
+	isFlagSet, err := parseCpuInfoFile("/proc/cpuinfo")
+	if err != nil {
+		return false
+	}
```
xiaochenshen: Thanks, I agree with you.

libcontainer/resourcemanager/resource_manager_linux.go
xiaochenshen: @hqhq Which is better in your opinion? Thank you.
(1) Just remove this line (+// +build linux) in resource_manager_linux.go.
(2) Add new file resource_manager_unsupported.go for !linux cases:
```
// +build !linux

package resourcemanager
```

hqhq: After a second check, I think current implementation is fine, it's only imported by packages which are wrapped by `linux` build flag, sorry for the noise.

libcontainer/intelrdt/intelrdt.go
hqhq: Shouldn't this be part of default mount info as we did for cgroups in https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/example.go#L111-L116 and be handled as bind mount in [mountToRootfs](https://github.com/opencontainers/runc/blob/v1.0.0-rc3/libcontainer/rootfs_linux.go#L153)?

xiaochenshen: @hqhq 
Unlike cgroup and cgroup filesystem, Intel RDT and resource control (resctrl) filesystem heavily depend on hardware/CPU support.

Likely, resctrl filesystem will not be mounted to rootfs either (1) h/w doesn't support Intel RDT or (2) kernel version older than 4.10, doesn't support Intel RDT.

hqhq: I mean resctrl should also be bind mounted to container if we mount it on host by default, I didn't notice you're going to remove this, I'm OK with that, let users mount this on host and add that mount info to `config.json` specifically would be better.

xiaochenshen: @hqhq 
> let users mount this on host and add that mount info to config.json specifically would be better.

I agree that, if Intel RDT is supported by h/w and kernel, it is user's responsibility to mount the resctrl filesystem in host, and then to add mount info to config.json accordingly.

In most cases (e.g., non-Intel Xeon platforms), Intel RDT is not supported and resctrl filesystem is not enabled. In my opinion, we may not add "fixed" rectrl filesystem mount info into specs.Mount in libcontainer/specconv/example.go by default.

xiaochenshen:
@yummypeng 
>I tried to add the configuration of intelRdt in my config.json, then run a container in an environment without support of intelRdt, and unexpectedly, the container was started successfully. Seems there needs an validation of Spec maybe in https://github.com/opencontainers/runc/blob/master/libcontainer/configs/validate/validator.go#L24.

Thank you. I will add Intel RDT configuration check to fix this issue.

libcontainer/intelrdt/intelrdt.go
xiaochenshen: @hqhq 

I was thinking if Intel RDT/CAT is possible to support nested containers case.

Unlike cgroups' hierarchy, Intel RDT resource control filesystem supports only single level filesystem layout (see details in section "Intel RDT "resource control" filesystem hierarchy" in this PR).

libcontainer/intelrdt/intelrdt.go
xiaochenshen: @yummypeng 

For the nested containers case, I added comments in:
https://github.com/opencontainers/runc/pull/1279#discussion_r114473075

xiaochenshen:
@hqhq 
> @xiaochenshen Thanks for working on this, overall it looks good, can you address all the comments and rebase please? thanks.

Thank you very much for helping code review. Most of the comments from you and @yummypeng will be addressed in my new patch.

But one remaining issue is nested containers case. Could you share some docs or issue/PR# about nested containers for me to learn from? I will figure out how to support it with single level resctrl filesystem interface and number of groups limitation. Thank you.

yummypeng:
@xiaochenshen Sorry, I didn't notice that Intel RDT resource control filesystem supports only single level filesystem layout, if it's true, then it seems that nested containers case cannot be supported. 

@hqhq Any ideas?

hqhq:
@xiaochenshen Nested container means you can create a container in another container, for cgroup, we bind mount the cgroup paths of current process to paths `/sys/fs/cgroup/<subsystem>` in container, which means if you create a container con2 in container con1, path `/sys/fs/cgroup/cpu/con2` in container con1 is actually `/sys/fs/cgroup/cpu/con1/con2` on host.

So if resctrl doesn't support:
```
tree /sys/fs/resctrl
/sys/fs/resctrl/
|-- info
|   |-- L3
|       |-- cbm_mask
|       |-- min_cbm_bits
|       |-- num_closids
|-- cpus
|-- schemata
|-- tasks
|-- <container_id1>
|   |-- <container_id2>
|       |-- cpus
|       |-- schemata
|       |-- tasks    
|   |-- cpus
|   |-- schemata
|   |-- tasks
``` 

or
```
mount -t resctrl resctrl /sys/fs/resctrl
unshare -m && chroot <container_rootfs>
mount -t resctrl resctrl /sys/fs/resctrl/
(the two mount points both works)
```

Then I think we can't support Intel RDT/CAT feature in nested containers.

xiaochenshen:
@hqhq @yummypeng 

Thanks for coaching me how nested containers work.

Intel RDT resource control filesystem doesn't support hierarchy layout which is the nature feature in cgroup.
Actually it is the reason why original cgroup-based Intel RDT/CAT kernel patch is rejected by kernel cgroup community. You can find more information from: https://github.com/opencontainers/runc/issues/433#issuecomment-281394321

xiaochenshen:
@hqhq @yummypeng 
@crosbymichael @cyphar
/cc @rjnagal @vmarmol @mrunalp @dqminh @avagin 

I have rebased the code to address the comments from @hqhq and @yummypeng.
Please help to review. Thank you.

**Changes in v2:**
Addressed the comments from @hqhq and @yummypeng:
- Added intelrdt config validator.
- Removed resctrl filesystem re-mount logic in isIntelRdtMounted().
- Optimized IsIntelRdtEnabled() to avoid unnecessary time-consuming mount points parsing.
- Changed Intel RDT related json items to snake_case style.
- Fixed some redundant codes with regard to intelRdtPath.
- Fixed potential null pointer issue in IntelRdtManager.Set().
- Fixed some typos (Sets, num_closids and etc.)
- Improved libcontainer/SPEC.md: added minimal supported kernel version, provided a more sophisticated config example.

**v1 commits for code review:**
4752fd2f639a38eb6ef6c2e030252098cd11d9f0 libcontainer: add support for Intel RDT/CAT in runc
c8486a7a8866b5ce4264ee0a17e54292a0f56fbf libcontainer/SPEC.md: add documentation for Intel RDT/CAT
aefd02ae210105573dabbb90ace3ab70943504ce libcontainer: refactor resource manager interface
8e8e5d96e6ad7dfefcd0f7d49a8d8c0b5cba0ef6 vendor: specs-go: update specs for Intel RDT/CAT

libcontainer/configs/validate/validator.go
```diff
@@ -153,6 +157,19 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 	return nil
 }
 
+func (v *ConfigValidator) intelrdt(config *configs.Config) error {
+	if config.IntelRdt != nil {
+		if !intelrdt.IsIntelRdtEnabled() {
+			return fmt.Errorf("intelRdt is specified in config, but Intel RDT is not supported")
+		}
+		if config.IntelRdt.L3CacheSchema == "" {
```
yummypeng: So, the following configuration is illegal?  
```
"linux": {
	"intelRdt": {
		"l3CacheSchema": ""
	}
}
```
@hqhq WDYT?

ramkri123:
Is this already committed? I tried generating a patch in the latest runc and it failed.

Feedback - would be very useful to have a method for specifying a percentage of the cache and abstract the low level cache bitmask details from the user.

Overall a very useful feature and well done.

xiaochenshen:
@rmanyari 
> Overall a very useful feature and well done.

Thank you for trying this and giving us positive feedback.

> Is this already committed? I tried generating a patch in the latest runc and it failed.

It is still under code review.

> Feedback - would be very useful to have a method for specifying a percentage of the cache and abstract the low level cache bitmask details from the user.

This PR is for basic Intel RDT/CAT feature support in runc. I do have plan to improve the user interface in future. Currently, users need to understand some background knowledge (e.g., the cache bitmask).

Specifying a percentage of the cache is really a good suggestion, but it also has limitation.
In this example, the contexts of "50% percentage" on cache 0 and on cache 1 are different:
```
On a two-socket machine with two L3 caches where the default CBM is 0xfffff and the
max CBM length is 20 bits. With this configuration, tasks inside the container only
have access to the "upper" 50% of L3 cache id 0 and the "lower" 50% L3 cache id 1:

"linux": {
	"intelRdt": {
		"l3CacheSchema": "L3:0=ffc00;1=3ff"
	}
}
```

libcontainer/configs/validate/validator.go
xiaochenshen: @yummypeng 
> So, the following configuration is illegal?

In my opinion, if we specify "intelRdt" in config, we'd better specify a non-empty string for sub-item "l3CacheSchema".

ramkri123:
@xiaochenshen  
>>> Specifying a percentage of the cache is really a good suggestion, but it also has limitation.
In this example, the contexts of "50% percentage" on cache 0 and on cache 1 are different:

Not really if this is the first partition you are creating ... at that point you have full control over the cache. Happy to chat more over a call as needed.

Additionally, you need a way to specify the exact partition number. Intel CAT allows several partitions.

xiaochenshen:
@ramkri123 

> Not really if this is the first partition you are creating ... at that point you have full control over the cache. Happy to chat more over a call as needed.

You can find more details about Intel RDT/CAT in the proposal https://github.com/opencontainers/runc/issues/433 and Intel RDT/CAT kernel interface:
https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt

> Additionally, you need a way to specify the exact partition number. Intel CAT allows several partitions.

Sorry, I didn't catch up with your question. I would like to explain more about how Intel RDT/CAT works in runc:

Intel RDT/CAT is exposed via "resource control" filesystem, which is a "cgroup-like" interface. Comparing with cgroups, it has similar process management lifecycle and interfaces in a container. But unlike cgroups' hierarchy, it has single level filesystem layout. After a runc container is started, the Intel RDT "resource control" filesystem hierarchy looks like:
```
tree /sys/fs/resctrl
/sys/fs/resctrl/
|-- info
|   |-- L3
|       |-- cbm_mask
|       |-- min_cbm_bits
|       |-- num_closids
|-- cpus
|-- schemata
|-- tasks
|-- <container_id>
    |-- cpus
    |-- schemata
    |-- tasks
```

/<container_id> is the group/partition for the runc container. The /<container_id>/schemata which specified via config.json is written by IntelRdtManager.Set(), and tasks running inside this container are attached to /<container_id>/tasks by IntelRdtManager.Apply() or WriteIntelRdtTasks().

ramkri123:
@xiaochenshen 

Let us continue the discussion at opencontainers mailing list.

libcontainer/factory_linux.go
```diff
@@ -224,10 +247,15 @@ func (l *LinuxFactory) Load(id string) (Container, error) {
...
 		initArgs:             l.InitArgs,
 		criuPath:             l.CriuPath,
-		cgroupManager:        l.NewCgroupsManager(state.Config.Cgroups, state.CgroupPaths),
 		root:                 containerRoot,
 		created:              state.Created,
 	}
+	resourceManagers := make(map[string]resourcemanager.ResourceManager)
+	resourceManagers["cgroups"] = l.NewCgroupsManager(state.Config.Cgroups, state.CgroupPaths)
+	if intelrdt.IsIntelRdtEnabled() {
+		resourceManagers["intelrdt"] = l.NewIntelRdtManager(&state.Config, id)
```
yummypeng: `state.IntelRdtPath` is ignored when load container from disk. This will result in residual directory in `/sys/fs/resctrl`.

Steps to reproduce:
```
# runc create testcontainer
# ls /sys/fs/resctrl
cpus  info  schemata  tasks  testcontainer
# runc delete testcontainer
# ls /sys/fs/resctrl
cpus  info  schemata  tasks  testcontainer
```

Seems that `NewIntelRdtManager` needs one more parameter to record `IntelRdtPath`.

xiaochenshen: @yummypeng 

Thank you. I will add parameter `path` to record IntelRdtPath in NewIntelRdtManager.

xiaochenshen:
@yummypeng @hqhq 
@crosbymichael @cyphar
/cc @rjnagal @vmarmol @mrunalp @dqminh @avagin 

I have rebased the code to v3. Please help code review. Thank you.

**v3 commits**:
be25a2030bec8113434e53e0ee1b99a6e0bfc7ed libcontainer: add support for Intel RDT/CAT in runc
96fd05fc7ab0bcca414efc3e9e4894dbe1df0e5d libcontainer/SPEC.md: add documentation for Intel RDT/CAT
d5ac70d5ae68b514421a7c7f60475189191c70f1 libcontainer: refactor resource manager interface
c302b704bedd62e7655a401697f1a4a79085bcc0 vendor: specs-go: update specs for Intel RDT/CAT

**Changes in v3:**
Addressed comment https://github.com/opencontainers/runc/pull/1279#discussion_r116414850 from @yummypeng :
- Add parameter `path` to record IntelRdtPath in NewIntelRdtManager.

xiaochenshen:
@yummypeng @hqhq 
@crosbymichael @cyphar 
/cc @rjnagal @vmarmol @mrunalp @dqminh @avagin 

I have rebased the code against latest master branch. Please help code review at your convenience.
Thank you.

xiaochenshen:
Ping maintainers and contributors:

@hqhq @crosbymichael @cyphar @yummypeng 
/cc @rjnagal @vmarmol @mrunalp @dqminh @avagin 


Could you help code review at your convenience?
Thank you.

hqhq:
LGTM

@xiaochenshen Thanks for working on this, vendor probably needs rebase but github didn't complain so far.

hqhq:
@crosbymichael @cyphar I'm adding this to 1.0 milestone as it's part of runtime-spec 1.0.

crosbymichael:
Either things change a lot from the last time I looked at the design or I spaced it.

I still don't see why we have to try to fit this behind the cgroup manager interface.  It adds so much code, dead code, empty methods, etc that it should be clear that it probably is not the right place for this code.  Now we have a static map with static keys on a map and removed type safety on the stats.

I don't understand why this feature wasn't just a simple `SetRdt()` and `StatRdt()`.  

@cyphar what do you think?

hqhq:
@xiaochenshen Does intelRdt need the whole cgroup manager interface?

xiaochenshen:
@crosbymichael 
/cc @cyphar 
The general design has not changed from https://github.com/opencontainers/runc/pull/1279#issuecomment-273307523 which you have commented positive feeback in https://github.com/opencontainers/runc/pull/1279#issuecomment-277826103
```
It adds a new "ResourceManager" structure as the base interface for all resource managers,
such as cgroups manager and incoming IntelRdt manager.

All registered resource managers are consolidated in linuxContainer structure.
We can apply to unified operations (e.g., Apply(), Set(), Destroy()) using all of the
registered resource managers.

Currently, cgroups manager is the single resource manager in libcontainer.
Linux kernel 4.10 will introduce Intel RDT/CAT feature, the kernel interface is exposed
via "resource control" filesystem, which is a cgroup-like interface. In order to support
Intel RDT/CAT in libcontainer, we need a new resource manager (IntelRdt manager)
outside cgroups."
```

This design is also based on @cyphar 's suggestion in https://github.com/opencontainers/runc/issues/433#issuecomment-241212914
```
If we can't consolidate it with the cgroup code, then please make a new "resource manager"
structure where all Managers are consolidated. Then, we can just have generic Manager
code which applies to right operation using all of the registered managers. That way,
we don't have intelRdtWhatever sitting anywhere outside the "resource manager" code.
This would actually be a possible improvement over the current state of affairs (though
there's some cgroup code that you won't be able to refactor so this may result in the
patch being a bit harder to implement).
```

In this PR, commit 8851a0db77a65b42f72e6711356dd401204ad436 refactors current resource manager interface.
``` 
This patch adds a new "ResourceManager" structure as the base interface for
all resource managers, such as cgroups manager and incoming IntelRdt manager.

All registered resource managers are consolidated in linuxContainer structure.
We can apply to unified operations (e.g., Apply(), Set(), Destroy()) using
all of the registered resource managers.
```

As a result, IntelRdt manager doesn't depends on cgroups manager, IntelRdt manager and cgroups manager have same base interface "ResourceManager".

xiaochenshen:
@hqhq 
Thank you for code review on several iterations.

> @xiaochenshen Does intelRdt need the whole cgroup manager interface?

With this PR, IntelRdt manager doesn't depends on cgroups manager, IntelRdt manager and cgroups manager have same base interface "ResourceManager".

hqhq:
>> @xiaochenshen Does intelRdt need the whole cgroup manager interface?
>
> With this PR, IntelRdt manager doesn't depends on cgroups manager, IntelRdt manager and cgroups manager have same base interface "ResourceManager".

My point is you made "ResourceManager" the same as the old cgroup manager interface, so you have to implement all these functions in intelRdt manager, are these functions all needed by intelRdt which make them the base ResourceManager? Maybe that's "dead code, empty methods" which @crosbymichael mentioned.

Maybe you can have a subset of old cgroup manager functions as `ResourceManager` interface and have cgroup and intelRdt there own functions, so you don't have to implement functions you don't need such as `Freeze` and handle `GetStats` differently.

crosbymichael:
Yes, my point is that this is so much code to force this features to "be like cgroups" when its not.  We could do the same thing with two function calls in the libcontainer codebase to set and get this.  Its one field on the config struct and 1000+ lines to make it look like a cgroup.  I think we just take the simple route and just get the feature working without touching cgroups code.

xiaochenshen:
@crosbymichael @hqhq 
Thank you for your code review and comments.
I will try to re-work to align with your suggestions.

xiaochenshen:
@crosbymichael @hqhq
/cc @cyphar @yummypeng

## Design updates (2017-08-18)

**v5 commits:**
5f47f99f3016027ddb0ce28d811c56c16cbdc270 libcontainer: add test cases for Intel RDT/CAT
0e83c867e7f0febe09eae3dc6d5fb6ab7e5e3a52 libcontainer: add support for Intel RDT/CAT in runc
657b33c724fc1f302aad5c05aa448319009f9df4 libcontainer/SPEC.md: add documentation for Intel RDT/CAT

**Changes in v5:**
Reworked code according to @crosbymichael and @hqhq's comments:
1. https://github.com/opencontainers/runc/pull/1279#issuecomment-319085278
2. https://github.com/opencontainers/runc/pull/1279#issuecomment-320008325
3. https://github.com/opencontainers/runc/pull/1279#issuecomment-319917615

- Obsoleted the design of "refactor resource manager interface": removed generic resourceManagers, isolated IntelRdt functions without touching cgroups code, removed the static map with static keys on resourceManagers set.
- Simplified IntelRdtManager implementation, removed "dead and empty" methods - GetPids(), GetAllPids(), Freeze(). Kept the required methods set - Apply(), GetStats(), Destroy(), GetPath(), and Set().
- Re-implemented GetStats(), isolated cgroups.Stats and intelrdt.Stats and kept type safety on the Stats.
- Changed test stuffs according to the functional rework.
- Added read-only L3 cache info (CbmMask, MinCbmBits, and NumClosids) into Intelrdt.Stats. Runc users could run "runc events" to fetch the information, which helps know more about L3 cache capabilities and how to set correct L3 cache schema in different Intel Xeon platforms.

xiaochenshen:
@crosbymichael @hqhq 
/cc @cyphar @yummypeng 

Could you help code review at your convenience?
Thank you.

## Design updates (2017-08-18)

**v5 commits:**
5f47f99f3016027ddb0ce28d811c56c16cbdc270 libcontainer: add test cases for Intel RDT/CAT
0e83c867e7f0febe09eae3dc6d5fb6ab7e5e3a52 libcontainer: add support for Intel RDT/CAT in runc
657b33c724fc1f302aad5c05aa448319009f9df4 libcontainer/SPEC.md: add documentation for Intel RDT/CAT

**Changes in v5:**
Reworked code according to @crosbymichael and @hqhq's comments:
1. https://github.com/opencontainers/runc/pull/1279#issuecomment-319085278
2. https://github.com/opencontainers/runc/pull/1279#issuecomment-320008325
3. https://github.com/opencontainers/runc/pull/1279#issuecomment-319917615

- Obsoleted the design of "refactor resource manager interface": removed generic resourceManagers, isolated IntelRdt functions without touching cgroups code, removed the static map with static keys on resourceManagers set.
- Simplified IntelRdtManager implementation, removed "dead and empty" methods - GetPids(), GetAllPids(), Freeze(). Kept the required methods set - Apply(), GetStats(), Destroy(), GetPath(), and Set().
- Re-implemented GetStats(), isolated cgroups.Stats and intelrdt.Stats and kept type safety on the Stats.
- Changed test stuffs according to the functional rework.
- Added read-only L3 cache info (CbmMask, MinCbmBits, and NumClosids) into Intelrdt.Stats. Runc users could run "runc events" to fetch the information, which helps know more about L3 cache capabilities and how to set correct L3 cache schema in different Intel Xeon platforms.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -0,0 +1,545 @@
...
+			return fmt.Errorf("failed to write %v to %v: %v", pid, IntelRdtTasks, err)
+		}
+	}
+	return nil
+}
+
+// Check if Intel RDT is enabled
+func IsIntelRdtEnabled() bool {
+	// We have checked the flag before
+	if isIntelRdtEnabled {
```
crosbymichael: isn't this racy? you should probably use a `sync.Once` for initialization of this bool

crosbymichael:
LGTM

Can you fix my comment in a follow up PR?

mrunalp:
LGTM. We can fix up smaller issues in follow-on PRs. Thanks!

mrunalp:
LGTM

libcontainer/intelrdt/intelrdt.go
xiaochenshen: @crosbymichael 
Thank you for good suggestion. I will fix it in follow-up PR.

xiaochenshen:
@crosbymichael @mrunalp 

Thank you for your code review.
I will open a new PR to fix remaining issues.

libcontainer/intelrdt/intelrdt.go
xiaochenshen: I opened a new PR in https://github.com/opencontainers/runc/pull/1589 to fix this issue.

xiaochenshen:
@crosbymichael @hqhq @cyphar @mrunalp 
/cc @opencontainers/runc-maintainers @yummypeng

I planned to add Intel RDT/CAT support for Docker. More information can be found in #433 .
But I am confused with the "vendor" configuration among Docker related projects (containerd, moby and docker-cli). I found a lot of dependency issues because recent runc changes have not been merged into these projects yet.

Who could share me the tips or documents about "vendor" tools?
Thank you in advance.

cyphar:
There is a `vendor.conf` in the root of Docker's source tree. You need to just update the commit ID in that file, and then run `vndr github.com/opencontainers/runc` in the source tree (the tool is available at https://github.com/lk4d4/vndr).

xiaochenshen:
@cyphar 
Got it. Thank you.

PR #1280, opened 2017-01-18
user: fix the parameter error

datawolf:
The parameters passed to `GetExecUser` is not correct.
Consider the following code:

```
package main

import (
	"fmt"
	"io"
	"os"
)

func main() {
	passwd, err := os.Open("/etc/passwd1")
	if err != nil {
		passwd = nil
	} else {
		defer passwd.Close()
	}

	err = GetUserPasswd(passwd)
	if err != nil {
		fmt.Printf("%#v\n", err)
	}
}

func GetUserPasswd(r io.Reader) error {
	if r == nil {
		return fmt.Errorf("nil source for passwd-formatted
data")
	} else {
		fmt.Printf("r = %#v\n", r)
	}
	return nil
}
```

If the file `/etc/passwd1` is not exist, we expect to return
`nil source for passwd-formatted data` error, and in fact, the func
`GetUserPasswd` return nil.

The same logic exists in runc code. this patch fix it.

### DIFF
--- a/libcontainer/user/user.go
+++ b/libcontainer/user/user.go
@@ -199,18 +199,16 @@ type ExecUser struct {
 // files cannot be opened for any reason, the error is ignored and a nil
 // io.Reader is passed instead.
 func GetExecUserPath(userSpec string, defaults *ExecUser, passwdPath, groupPath string) (*ExecUser, error) {
-	passwd, err := os.Open(passwdPath)
-	if err != nil {
-		passwd = nil
-	} else {
-		defer passwd.Close()
+	var passwd, group io.Reader
+
+	if passwdFile, err := os.Open(passwdPath); err == nil {
+		passwd = passwdFile
+		defer passwdFile.Close()
 	}
 
-	group, err := os.Open(groupPath)
-	if err != nil {
-		group = nil
-	} else {
-		defer group.Close()
+	if groupFile, err := os.Open(groupPath); err == nil {
+		group = groupFile
+		defer groupFile.Close()
 	}
 
 	return GetExecUser(userSpec, defaults, passwd, group)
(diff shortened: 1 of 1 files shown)


libcontainer/user/user.go
```diff
@@ -199,18 +199,18 @@ type ExecUser struct {
...
 func GetExecUserPath(userSpec string, defaults *ExecUser, passwdPath, groupPath string) (*ExecUser, error) {
-	passwd, err := os.Open(passwdPath)
-	if err != nil {
-		passwd = nil
-	} else {
-		defer passwd.Close()
+	var passwd, group io.Reader
+
+	passwdFile, err := os.Open(passwdPath)
+	if err == nil {
```
WeiZhang555: It can be simplified more:

```
if passwdFile, err := os.Open(passwdPath); err == nil {
...
}
```

libcontainer/user/user.go
```diff
@@ -433,9 +431,11 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
 // that opens the groupPath given and gives it as an argument to
 // GetAdditionalGroups.
 func GetAdditionalGroupsPath(additionalGroups []string, groupPath string) ([]int, error) {
-	group, err := os.Open(groupPath)
-	if err == nil {
```
cyphar: The intention of this code was so that you could pass `""` as a path and you won't get any errors unless the specification of a user requires lookup -- effectively allowing the numerical specification of a user without specifying `/etc/passwd`.

Obviously I got that a bit wrong, but the current version your proposing doesn't retain that behaviour. Though maybe we should consider whether it makes sense for us to keep that behaviour.

datawolf: @cyphar, sorry for the late reply. If we pass `""` as a path or the rootfs does not contains file `/etc/group`, the `os.Open` will failed and return a `(*os.File)(nil)`, And we should not pass it to functions `GetAdditionalGroups`  and `ParseGroupFilter`, because `group != nil` in these functions always `true`.

For the case you mentioned, we can process it in `GetAdditionalGroupsPath`, Or we need not to keep that behaviour.

cyphar:
Apologies, I must've misunderstood the code on my first review.

LGTM.

crosbymichael:
LGTM

PR #1158, opened 2016-10-28
Check args numbers before application start

WeiZhang555:
Add a general args number validator for all client commands.

Signed-off-by: Zhang Wei zhangwei555@huawei.com

PR #1282, opened 2017-01-18
kill: make second argument optional

giuseppe:
commit b517076907b7c1 added a check that kill accepts two arguments.
Since the second argument is optional, change it back to accept the
shorter form "kill CONTAINER".

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -52,10 +52,10 @@ var signalMap = map[string]syscall.Signal{
 var killCommand = cli.Command{
 	Name:  "kill",
 	Usage: "kill sends the specified signal (default: SIGTERM) to the container's init process",
-	ArgsUsage: `<container-id> <signal>
+	ArgsUsage: `<container-id> [signal]
 
 Where "<container-id>" is the name for the instance of the container and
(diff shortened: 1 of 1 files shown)


WeiZhang555:
You are right, sorry...

Maybe the help message should be updated too? I made this change based on description of command help, that's why I think second one is mandatory.

giuseppe:
@WeiZhang555 I pushed another version that changes ArgsUsage as well ⬆️

WeiZhang555:
LGTM (though I'm not a maintainer)

giuseppe:
Thanks for the feedback.  I am tagging @hqhq @cyphar as they've already reviewed https://github.com/opencontainers/runc/pull/1158

mrunalp:
LGTM

hqhq:
LGTM

cyphar:
But now you can give `runc kill ctr ABC DEF HIJ` and it won't give an error. We should probably come up with a way of implementing gt and lt requirements...

giuseppe:
@cyphar I have opened a PR to add a max of two arguments for kill:

https://github.com/opencontainers/runc/pull/1305

PR #1285, opened 2017-01-21
Only wait for processes after delivering SIGKILL in signalAllProcesses

stevenh:
signalAllProcesses was making the assumption that the requested signal was SIGKILL, possibly due to the signal parameter being added at a later date, and hence it was safe to wait for all processes which is not the case.

BaseContainer.Signal(s os.Signal, all bool) exposes this functionality to consumers, so an arbitrary signal could be used which is not guaranteed to make the processes exit.

Correct the documentation for signalAllProcesses around the signal delivered and update it so that the wait is only performed on SIGKILL hence making it safe to process other signals without risk of blocking forever, while still maintaining compatibility to SIGKILL callers.

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -394,8 +394,8 @@ func setOomScoreAdj(oomScoreAdj int, pid int) error {
 }
 
 // signalAllProcesses freezes then iterates over all the processes inside the
-// manager's cgroups sending a SIGKILL to each process then waiting for them to
-// exit.
+// manager's cgroups sending the signal s to them.
+// If s is SIGKILL then it will also wait for each process to exit.
 func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	var procs []*os.Process
 	if err := m.Freeze(configs.Frozen); err != nil {
@@ -420,6 +420,11 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	if err := m.Freeze(configs.Thawed); err != nil {
 		logrus.Warn(err)
 	}
+
+	if s != syscall.SIGKILL {
+		return nil
+	}
+
 	for _, p := range procs {
 		if _, err := p.Wait(); err != nil {
 			logrus.Warn(err)
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"unsafe"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM
cc: @crosbymichael

cyphar:
There is also `SIGTERM` that we need to worry about, not to mention the other signals that _could_ kill a process. The "proper" fix for this (though a bit ugly) would be to check `/proc/[pid]/stat` and see whether the process is a zombie before we do `.Wait()`.

stevenh:
You actually don't want to do SIGTERM as it can be either ignored or fail to make the process quit.

cyphar:
@stevenh I know, which is why I said "**could**" kill the process. That's also why my suggested fix involves checking whether the `kill` was fatal (see whether the process exited and is now a zombie) -- I can't think of a nicer way (other than using a timeout which usually doesn't end well).

stevenh:
If the goal here is to avoid zombies, I'd argue that explicit waits where we "think" the process might exit are the wrong approach as that won't deal with unexpected exits, so doesn't solve the issue in all cases.

In addition, manually waiting on specific processes like this it may prevent their correct behaviour as it could prevent a parent process correctly receiving exit information from its children.

Instead installing a signal handler in the init process, as runc see appears to already do, might be the better option?

Alternatively using syscall.Wait4 with WNOHANG to check if the processes wish to report status and if so then call Wait() might be a good compromise.

What do you think?

stevenh:
So wait4 on Linux it doesn't support WNOHANG :(

I tried waitid, which is used inside Process.Wait(), and that errored with "no child processes" so I rechecked with runc (unpatched) and the wait calls actually fail and log the same warning, as they are sent from a process which isn't the parent.

I remember discovering this issue with Linux when we wrote our internal process management library a long time ago and the only way to wait for processes which aren't your children is to use netlink and I think that's only for notification it won't reap zombies.

In our use of libcontainer however, the issue is still be relevant, as our process is a persistent management daemon hence the initial process will be a child, so does block.

I do have a full implementation using waitid now so if you would like to me to push that up for review let me know.

cyphar:
@stevenh 

> I remember discovering this issue with Linux when we wrote our internal process management library a long time ago and the only way to wait for processes which aren't your children is to use netlink and I think that's only for notification it won't reap zombies.

But runC uses `PR_SET_SUBPEAPER` which effectively turns runC into a fake PID 1 process (meaning that all children of the process tree will reparent themselves to us). Now, there are downsides to this approach, but the point is that we should be getting `SIGCHLD`s (unless I've forgotten some intricacy about the PID namespace while I was on holiday).

> In our use of libcontainer however, the issue is still be relevant, as our process is a persistent management daemon hence the initial process will be a child, so does block.

Aside: runC is **not** a daemon. It's intended usage is of the form `runc create` and `runc start` -- which essentially ensure that the runC process doesn't stick around. Please keep this in mind. :smile_cat:

cyphar:
> > In our use of libcontainer however, the issue is still be relevant, as our process is a persistent management daemon hence the initial process will be a child, so does block.

> Aside: runC is **not** a daemon. It's intended usage is of the form `runc create` and `runc start` -- which essentially ensure that the runC process doesn't stick around. Please keep this in mind. :smile_cat: 

Ah sorry, I misread this and missed "In **our** usage of libcontainer" -- where "our" is not runC. In which case, have you taken a look at `PR_SET_CHILD_SUBREAPER`?

dqminh:
I dont think its the case of subreaper. We just exposed a somewhat invalid API here to signal all processes.

Dont you think its cleaner to just have `pids, err := signalAllProcesses()` which doesnt wait, then have `killAllProceses` which does `signalAllProcesses(manager, sigkill)`, then wait on the returned pids ?

cyphar:
@dqminh Yes, this API needs to be fixed. But we can't _only_ do `Wait` for `SIGKILL` -- because a process could die for a variety of other reasons. However, this isn't a problem _in runC_ because we have a signal handler for `SIGCHLD` as well as subreapers already set up.

I'm just trying to help explain to @stevenh a way of handling this in the caller -- we still need to refine the signalling API though.

dqminh:
> Yes, this API needs to be fixed. But we can't only do Wait for SIGKILL -- because a process could die for a variety of other reasons. However, this isn't a problem in runC because we have a signal handler for SIGCHLD as well as subreapers already set up.

That's right, But we still have the option to disable subreaper :p
This kill function is only called when sharing pid namespace which we always send a sigkill so i think its ok to always wait there.

stevenh:
@dqminh is not only used there is used by container.Signal() which is why we encountered the issue where our call to signal.Signal(syscall.INT, true) blocked forever as the processes in the container didn't exit.

I don't think its good idea to assume that consumers of the API have any additional processing. The API should clean up after itself.

The current method does ensure that children are cleaned up even if the caller hasn't explicitly done so, however as I pointed out its only for the duration of the call so personally I don't like it as its only part of the problem.

To be clear this PR is only to fix the blocking forever when using container.Signal(), while still maintaining signalAllProcesses behaviour for its internal consumer; fixing signal handling in general should be handled separately to this.

stevenh:
I've pushed an update to this which now always attempts to wait for the children no matter the signal.

While its more complex it is closer to the original for none kill signals, as it will reap children if they exited.

I've kept this as a separate commit for now so the difference can be seen, if this is the preferred method I'll squash.

crosbymichael:
Its probably better to handle the reaping at the higher layers and split up this api so that you SignalAllProcess, do some reaping if needed with a proper handler, then Destory or whatever else you want to do.

stevenh:
@crosbymichael I'd agree that would be ideal, so feel free to use this an interim fix until such time as that can be done.

I'm sure people will appreciate, we'd like to use official versions instead of having to maintain private patches until a bigger API change can be completed and integrated to address the wider issue.

crosbymichael:
LGTM

We can work on splitting this up after, merge.

hqhq:
LGTM

PR #1290, opened 2017-01-24
fixed some golint warnings

rainrambler:
fixed some golint warnings about the format of comment.

Signed-off-by: Wang Anyu <wanganyu@outlook.com>

### DIFF
--- a/libcontainer/devices/devices_unix.go
+++ b/libcontainer/devices/devices_unix.go
@@ -23,7 +23,7 @@ var (
 	ioutilReadDir = ioutil.ReadDir
 )
 
-// Given the path to a device and its cgroup_permissions(which cannot be easily queried) look up the information about a linux device and return that information as a Device struct.
+// DeviceFromPath: Given the path to a device and its cgroup_permissions(which cannot be easily queried) look up the information about a linux device and return that information as a Device struct.
 func DeviceFromPath(path, permissions string) (*configs.Device, error) {
 	fileInfo, err := osLstat(path)
 	if err != nil {
--- a/libcontainer/user/lookup.go
+++ b/libcontainer/user/lookup.go
@@ -11,7 +11,7 @@ var (
 	// ErrNoPasswdEntries: No matching password entries found in file.
 	ErrNoPasswdEntries = errors.New("no matching entries in passwd file")
 	// ErrNoGroupEntries: No matching group entries found in file.
-	ErrNoGroupEntries  = errors.New("no matching entries in group file")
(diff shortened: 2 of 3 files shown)


hqhq:
Commits should to be squashed.

libcontainer/user/lookup.go
```diff
@@ -6,11 +6,12 @@ import (
 )
 
 var (
-	// The current operating system does not provide the required data for user lookups.
+	// ErrUnsupported: The current operating system does not provide the required data for user lookups.
```
stevenh: This really should be reworded slightly so its not just eliminating the lint warning but also making it read well e.g.
```// ErrUnsupported is returned if the current operating system does not provide the required data for user lookups.```

File: libcontainer/user/lookup.go [hunk 2]
```diff
@@ -6,11 +6,12 @@ import (
...
 	ErrUnsupported = errors.New("user lookup: operating system does not provide passwd-formatted data")
-	// No matching entries found in file.
+	// ErrNoPasswdEntries: No matching password entries found in file.
```
stevenh: As above:
```ErrNoPasswdEntries is returned if no matching password entries is found in the passwd file.```

File: libcontainer/user/lookup.go [hunk 3]
```diff
@@ -6,11 +6,12 @@ import (
...
 	ErrNoPasswdEntries = errors.New("no matching entries in passwd file")
-	ErrNoGroupEntries  = errors.New("no matching entries in group file")
+	// ErrNoGroupEntries: No matching group entries found in file.
```
stevenh: ```// ErrNoGroupEntries is returned if no matching group entries is found in the group file.```

libcontainer/devices/devices_unix.go
```diff
@@ -23,7 +23,7 @@ var (
 	ioutilReadDir = ioutil.ReadDir
 )
 
-// Given the path to a device and its cgroup_permissions(which cannot be easily queried) look up the information about a linux device and return that information as a Device struct.
+// DeviceFromPath: Given the path to a device and its cgroup_permissions(which cannot be easily queried) look up the information about a linux device and return that information as a Device struct.
```
stevenh: I would suggest something along the lines of:
```// DeviceFromPath returns a config.Device representing the Linux device at path with the cgroup representation of permissions.```

crosbymichael:
@rainrambler if you have time to update this PR again please reopen when you are ready.

Thanks.

PR #1293, opened 2017-01-24
Resolve InitArgs to ensure init works

stevenh:
If a relative pathed exe is used for InitArgs init will fail to run if Cwd is not set the original path.

Prevent failure of init to run by ensuring that exe in InitArgs is an absolute path.

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -34,7 +34,15 @@ var (
 // InitArgs returns an options func to configure a LinuxFactory with the
 // provided init binary path and arguments.
 func InitArgs(args ...string) func(*LinuxFactory) error {
-	return func(l *LinuxFactory) error {
+	return func(l *LinuxFactory) (err error) {
+		if len(args) > 0 {
+			// Resolve relative paths to ensure that its available
+			// after directory changes.
+			if args[0], err = filepath.Abs(args[0]); err != nil {
+				return newGenericError(err, ConfigInvalid)
+			}
+		}
+
 		l.InitArgs = args
 		return nil
 	}


libcontainer/factory_linux.go
```diff
@@ -34,12 +34,39 @@ var (
...
+		if len(l.InitArgs) == 0 {
+			return newGenericError(fmt.Errorf("empty initargs"), ConfigInvalid)
+		}
+
+		if args[0], err = resolvePath(args[0]); err != nil {
+			return newGenericError(err, ConfigInvalid)
+		}
+
 		l.InitArgs = args
+
```
cyphar: Nit: Don't add a space here.

libcontainer/factory_linux.go
```diff
@@ -34,12 +34,39 @@ var (
 // InitArgs returns an options func to configure a LinuxFactory with the
 // provided init binary path and arguments.
 func InitArgs(args ...string) func(*LinuxFactory) error {
-	return func(l *LinuxFactory) error {
+	return func(l *LinuxFactory) (err error) {
+		if len(l.InitArgs) == 0 {
```
cyphar: `len(args)`? Also I'm not sure if this check is necessary because there are cases where you the container command won't be executed (where you just do `runc create` and then `runc exec`).

libcontainer/factory_linux.go
```diff
@@ -34,12 +34,39 @@ var (
...
+	f, err := filepath.EvalSymlinks(path)
+	if err != nil {
+		return "", err
+	}
+
+	if f, err = filepath.Abs(f); err != nil {
+		return "", err
+	}
+
+	if _, err := os.Stat(f); err != nil {
```
cyphar: This check is not a good idea -- we support creating a container and then creating the binary later (before `runc start` is called).

libcontainer/factory_linux.go
stevenh: Thanks refactored to fix this

libcontainer/factory_linux.go
```diff
@@ -35,6 +35,14 @@ var (
 // provided init binary path and arguments.
 func InitArgs(args ...string) func(*LinuxFactory) error {
 	return func(l *LinuxFactory) error {
+		if len(l.InitArgs) > 0 {
```
cyphar: Again, I'm confused -- why isn't this `len(args)`? `l.InitArgs` hasn't been set yet.

stevenh: We'll that will teach me to re-type instead of doing a partial cherry-pick won't it!

crosbymichael:
@stevenh why don't you just give it the abs path yourself?  That function is meant to set the args to exactly what the user provided.   So if you want to have an abs path, a symlink, or whatever, its up to you to give it a proper path.  We shouldn't resolve symlinks or change the data that is provided.

Also the api is setup so you can just write your only options func that you pass to the create method to change this.  It does not have to be changed in the core.

stevenh:
I'm not sure I'd agree as due to is internal use values which are totally valid when passed in become invalid due to actions taken internally, actions which aren't documented for the correct reasons (they are implementation details).

Now I came across the issue as I used the example code in libcontainers README:
```go
factory, err := libcontainer.New("/var/lib/container", libcontainer.Cgroupfs, libcontainer.InitArgs(os.Args[0], "init"))
if err != nil {
    logrus.Fatal(err)
    return
}
```
Now while this works fine if the application was invoked using an absolute path, however it breaks if it was relative, which is exactly what happened to me and had me scratching and adding debug to the library to determine the issue; especially since errors where not handled properly.

Now as you say each consumer of libcontainer could fix this by doing the lookup themselves, but why make every consumer do that when updating the internals to do it is simple and has no down sides I can see?

crosbymichael:
@stevenh  if you look at prior use or even the default you can see that it's using `/proc/self/exe` as the default and will do the correct thing.

However, I agree with you on your last comment since we change the cwd in some instances that we should resolve the abs.  

LGTM

hqhq:
LGTM

PR #1294, opened 2017-01-24
Ensure pipe is always closed on error in StartInitialization

stevenh:
Ensure that the pipe is always closed during the error processing of  StartInitialization.

Also:
* Fix a comment typo.
* Use newContainerInit directly as there's no need for i to be an initer.
* Move the comment about the behaviour of Init() directly above it, clarifying what happens for all defers.

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -243,16 +243,15 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 		pipe = os.NewFile(uintptr(pipefd), "pipe")
 		it   = initType(os.Getenv("_LIBCONTAINER_INITTYPE"))
 	)
+	defer pipe.Close()
+
 	// clear the current process's environment to clean any libcontainer
 	// specific env vars.
 	os.Clearenv()
 
-	var i initer
 	defer func() {
 		// We have an error during the initialization of the container's init,
 		// send it back to the parent process in the form of an initError.
-		// If container's init successed, syscall.Exec will not return, hence
-		// this defer function will never be called.
 		if werr := utils.WriteJSON(pipe, syncT{procError}); werr != nil {
 			fmt.Fprintln(os.Stderr, err)
 			return
@@ -261,18 +260,19 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 			fmt.Fprintln(os.Stderr, err)
 			return
 		}
-		// ensure that this pipe is always closed
-		pipe.Close()
 	}()
 	defer func() {
 		if e := recover(); e != nil {
 			err = fmt.Errorf("panic from initialization: %v, %v", e, string(debug.Stack()))
 		}
 	}()
-	i, err = newContainerInit(it, pipe, rootfd)
+
+	i, err := newContainerInit(it, pipe, rootfd)
 	if err != nil {
 		return err
 	}
+
+	// If Init succeeds, syscall.Exec will not return, hence none of the defers will be called.
 	return i.Init()
 }
 


libcontainer/factory_linux.go
```diff
@@ -239,19 +239,23 @@ func (l *LinuxFactory) StartInitialization() (err error) {
...
 		*pair.v = i
 	}
+
 	var (
 		pipe = os.NewFile(uintptr(pipefd), "pipe")
 		it   = initType(os.Getenv("_LIBCONTAINER_INITTYPE"))
 	)
+
+	// ensure that this pipe is always closed in the case of failure.
+	defer pipe.Close()
```
cyphar: If you want to make this a `defer` please put it inside the `defer func()` where the old `pipe.Close` used to be. I can't remember the exact reasoning behind it, but you need to make sure that you close `pipe.Close` immediately after sending the error.

libcontainer/factory_linux.go
```diff
@@ -261,19 +265,26 @@ func (l *LinuxFactory) StartInitialization() (err error) {
...
+
+	var i initer
 	i, err = newContainerInit(it, pipe, rootfd)
 	if err != nil {
 		return err
 	}
-	return i.Init()
+
+	if i, err = newContainerInit(it, pipe, rootfd); err != nil {
+		return
```
cyphar: We don't use this form of `return`. Please use `return err` _explicitly_, as well as `i, err :=`.

libcontainer/factory_linux.go
```diff
@@ -261,19 +265,26 @@ func (l *LinuxFactory) StartInitialization() (err error) {
...
+	}
+
+	err = i.Init()
+
+	return
```
cyphar: Same here. `return i.Init()`.

libcontainer/factory_linux.go
stevenh: That will already happen naturally.

libcontainer/factory_linux.go
stevenh: You can't use ```i, err :=``` as there are no new variables so it would fail compile.

Is there a reason for not using straight ```return``` as this is not only the idiomatic way, given the use of ```err``` in the ```defer``` not doing so could easily lead to additional coding errors where the defer fails to get the required error.

Given this I would argue that its both clearer and less error prone to use the short ```return``` format.

libcontainer/factory_linux.go
cyphar: When I said "immediately" I meant inside the function. I understand what you're saying, but especially because of `recover()` I don't like moving this outside of the other `defer`. `EOF`s are important inside runC, and I don't like splitting that from where the error handling is done.

libcontainer/factory_linux.go
cyphar: > You can't use i, err := as there are no new variables so it would fail compile.

Sorry, I missed that.

> Is there a reason for not using straight return as this is not only the idiomatic way

Because it's confusing to read.

> given the use of err in the defer not doing so could easily lead to additional coding errors where the defer fails to get the required error.

That will not happen with `return err`. However, it _can_ happen with the bare `return` format because it's less clear what the last value of `err` was. Then someone "refactors" the code and now you're forgetting to set an error.

Bare `return`s are one of the larger warts in Go IMO.

libcontainer/factory_linux.go
stevenh: Its still unneeded and putting it directly after the creation is both clearer and more idiomatic.

Here's a playground demo: https://play.golang.org/p/_KvjCwK2V4

stevenh: With the updated version I think this is the only outstanding point of contention, LMK your verdict.

cyphar: While I'm aware of the semantics of `defer`, my point is that you have to remember the semantics while reading the `defer`d functions (which also interact with errors). I mean, ultimately I don't care all that much it's just more stuff to keep track of in an already confusing function.

LK4D4:
LGTM

dqminh:
LGTM

PR #1296, opened 2017-01-24
Report init process errors during early startup

stevenh:
If the init process fails during StartInitialization() it returns the cause to the parent via the pipe.

This initial failure causes knock on failures in parent which are reported and the error from the init process is lost.

For the early portion of start() we now attempt to read the error from the init process, if one is found we report it in preference, as it provides more relevant information about the actual problem.

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -87,7 +87,7 @@ func (p *setnsProcess) start() (retErr error) {
 	p.messageSockPair.child.Close()
 	p.logFilePair.child.Close()
 	if err != nil {
-		return fmt.Errorf("error starting setns process: %w", err)
+		return parentError(p.messageSockPair.parent, err, "starting setns process")
 	}
 
 	waitInit := initWaiter(p.messageSockPair.parent)
@@ -110,15 +110,15 @@ func (p *setnsProcess) start() (retErr error) {
 
 	if p.bootstrapData != nil {
 		if _, err := io.Copy(p.messageSockPair.parent, p.bootstrapData); err != nil {
-			return fmt.Errorf("error copying bootstrap data to pipe: %w", err)
+			return parentError(p.messageSockPair.parent, err, "copying bootstrap data to pipe")
 		}
 	}
 	err = <-waitInit
 	if err != nil {
 		return err
 	}
-	if err := p.execSetns(); err != nil {
-		return fmt.Errorf("error executing setns process: %w", err)
+	if err = p.execSetns(); err != nil {
+		return parentError(p.messageSockPair.parent, err, "executing setns process")
 	}
 	for _, path := range p.cgroupPaths {
 		if err := cgroups.WriteCgroupProc(path, p.pid()); err != nil && !p.rootlessCgroups {
@@ -158,7 +158,7 @@ func (p *setnsProcess) start() (retErr error) {
 		return fmt.Errorf("error setting rlimits for process: %w", err)
 	}
 	if err := utils.WriteJSON(p.messageSockPair.parent, p.config); err != nil {
-		return fmt.Errorf("error writing config to pipe: %w", err)
+		return parentError(p.messageSockPair.parent, err, "writing config to pipe")
 	}
 
 	ierr := parseSync(p.messageSockPair.parent, func(sync *syncT) error {
@@ -353,6 +353,17 @@ func (p *initProcess) waitForChildExit(childPid int) error {
 	return nil
 }
 
+// parentError attempts to read any error from the parent so we have better context
+// about the cause of the failure. If no error is found then err is combined with cause
+// and returned.
+func parentError(pipe *os.File, err error, cause string) error {
+	if perr := parseSync(pipe, func(sync *syncT) error { return nil }); perr != nil {
+		return fmt.Errorf("%s: %w", cause, perr)
+	}
+
+	return fmt.Errorf("%s: %w", cause, err)
+}
+
 func (p *initProcess) start() (retErr error) {
 	defer p.messageSockPair.parent.Close() //nolint: errcheck
 	err := p.cmd.Start()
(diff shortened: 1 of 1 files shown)


libcontainer/process_linux.go
```diff
@@ -72,15 +72,15 @@ func (p *setnsProcess) start() (err error) {
 	p.childPipe.Close()
 	p.rootDir.Close()
 	if err != nil {
-		return newSystemErrorWithCause(err, "starting setns process")
+		return parentError(p.parentPipe, err, "starting setns process")
 	}
 	if p.bootstrapData != nil {
 		if _, err := io.Copy(p.parentPipe, p.bootstrapData); err != nil {
-			return newSystemErrorWithCause(err, "copying bootstrap data to pipe")
+			return parentError(p.parentPipe, err, "copying bootstrap data to pipe")
```
cyphar: For all of these, how can you be sure that there is a pending error on `p.parentPipe` or the process had died? Otherwise `.Decode` in `parseSync` will block, no?

libcontainer/process_linux.go
```diff
@@ -255,6 +255,19 @@ func (p *initProcess) execSetns() error {
 	return nil
 }
 
+// parentError attempts to read any error from the parent so we have better context
+// about the cause of the failure. If no error is found then a new system error for
+// err and cause is returned.
+func parentError(pipe *os.File, err error, cause string) error {
+	if ierr := parseSync(pipe, func(sync *syncT) error { return nil }); ierr != nil {
```
cyphar: And more importantly, I don't think I understand the logic of why you're using `parseSync` here -- what synchronisation call are you waiting for? How can you be sure that one was sent or that the child has died?

stevenh: I think there may be a miss-understanding of what parseSync does due to its name. I believe that it is actually just a method to read data from the pipe by aligning on messages that a prefixed by a syncT which act as a message "headers".

If my understanding is correct parseSync might be better named processMsgs.

The problem of if the child has died or not is a tricky one. Given the early nature of these failures its highly likely but not guaranteed, so it might be a good idea ensure it is dead by killing it as we're never going to use it again.

That however is a bigger issue as the whole of this function is fragile with regards to that, as only if it succeeds is the wait and hence full cleanup performed currently. Fixing this would require a much bigger restructure, which I would suggest should be done separately.

libcontainer/process_linux.go
stevenh: We can't but see my answer to your question below, which as you correctly state is the real issue in question.

libcontainer/process_linux.go
cyphar: > I think there may be a miss-understanding of what parseSync does due to its name. I believe that it is actually just a method to read data from the pipe by aligning on messages that a prefixed by a syncT which act as a message "headers".

`parseSync` parses `syncT`s from a given pipe. Sorry for the confusing name (I wrote it to abstract all of the `syncT` parsing code introduced previously).

> If my understanding is correct parseSync might be better named processMsgs.

That name is more confusing IMO.

> The problem of if the child has died or not is a tricky one. Given the early nature of these failures its highly likely but not guaranteed, so it might be a good idea ensure it is dead by killing it as we're never going to use it again.

Sure, and that's a round-about way of forcing an `EOF`. Still not sure if it makes sense.

> That however is a bigger issue as the whole of this function is fragile with regards to that, as only if it succeeds is the wait and hence full cleanup performed currently. Fixing this would require a much bigger restructure, which I would suggest should be done separately.

I think that's because you're misusing `parseSync` -- its purpose is to be used within the synchronisation state machine code. It's a bit odd how you're using it, and I doubt that it works as intended.

stevenh: Hmm, not sure we're on the same page yet.

I could manually read the pipe, slurping data until EOF looking for an procError but parseSync achieves just this, so it seems sensible to make use of this; did I miss something?

I can confirm that when init failed, I'd not included nsenter so the config read was failing with a parse error, this change correctly receives and outputs the error returned from init.

stevenh: @cyphar could you clarify your position on this one now please?

cyphar: @stevenh Can you please give an example of how you've tested that this improves the current error situation. I assumed that this patch would show it, but it appears that this error is handled properly by `master`. As far as I can tell this code is just prioritising `procError` over internal `libcontainer.Process` errors -- I'm not sure whether that makes sense.

```diff
diff --git a/libcontainer/factory_linux.go b/libcontainer/factory_linux.go
index 7f044bd45d5a..f88d8c25c719 100644
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -275,6 +275,7 @@ func (l *LinuxFactory) StartInitialization() (err error) {
        }

        // If Init succeeds, syscall.Exec will not return, hence none of the defers will be called.
+       return fmt.Errorf("LOLWHAT")
        return i.Init()
 }
```

Output in both branches:

```
% sudo ./runc run ctr
container_linux.go:259: starting container process caused "process_linux.go:385: container init caused \"LOLWHAT\""
```

stevenh: @cyphar the main one was not including nsenter, which was missing from the docs:
import _ "github.com/opencontainers/runc/libcontainer/nsenter"

AkihiroSuda:
needs rebase

kolyshkin:
I worked a lot to make sure early runc init (including nsenter) errors are making its way to the log. There are only a few cases when it's not so, and those are very rate.

@stevenh can you please provide an example of a failure which is not caught/logged without this PR, and is caught/logged with it? The failure itself can be injected into the code.

cyphar:
I think this was obsoleted when we fixed up the log forwarding logic. Now, you should get logs from the child and we don't need to handle it manually AFAIK.

stevenh:
Thanks for checking in on this, as it is very old, I'm going close and I'll re-raise if I encounter again.

PR #1297, opened 2017-01-25
Allow Destroy() on running containers

stevenh:
Allow destroying of running containers as documented by BaseContainer.Destroy() which states:
"Destroys the container after killing all running processes."

Also:
* Fix documentation typo for restoredState.

### DIFF
--- a/libcontainer/state_linux.go
+++ b/libcontainer/state_linux.go
@@ -126,13 +126,7 @@ func (r *runningState) transition(s containerState) error {
 }
 
 func (r *runningState) destroy() error {
-	t, err := r.c.runType()
-	if err != nil {
-		return err
-	}
-	if t == Running {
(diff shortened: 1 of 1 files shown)


stevenh:
This might be contentious one, but the docs seem clear that Destroy() should always kill all the processes and then destroy the container, with no mention of "unless its running".

Certainly tripped us up when trying to implement against the word of the docs.

crosbymichael:
No, the "docs" are wrong or they are referring to killing any additional processes in the container's cgroups if they are left around.  We wan the user to explicitly stop a container first before destroying it.  

If you would like to open a PR to modify the comment that would be great but we want to keep the same logic forcing a user to stop the container first if they want to delete it.

stevenh:
Should the function also be renamed then as the meaning of destroy is pretty clear, so even correcting the docs it would still be confusing.

PR #1300, opened 2017-01-26
Call defer tty.Close() earlier

hqhq:
We could fail and return in tty.recvtty() and miss
tty.Close(), which might cause broken console output.

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -239,6 +239,7 @@ func (r *runner) run(config *specs.Process) (int, error) {
 		r.destroy()
 		return -1, err
 	}
+	defer tty.Close()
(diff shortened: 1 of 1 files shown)


cyphar:
What do you mean by "broken console output"? Do you mean that the terminal is left in raw mode, or you actually get garbled output?

cyphar:
LGTM in any case, because you're right that this would leave the terminal in raw mode -- I guess I was worried about deadlocking (because we use a `WaitGroup` to delay exiting after the `tty` was set up).

dqminh:
LGTM

dqminh:
I dont think the CI failure is related, so this should be ok to merge.

PR #1301, opened 2017-01-26
Reap nsenter children to prevent zombies

stevenh:
During the bootstap processing nsenter it spawns a temporary child which then exits after processing. This child is created with CLONE_PARENT so its init parents responsibility to reap.

Update execSetns to reap these temporary children avoiding zombies.

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -148,7 +148,11 @@ func (p *setnsProcess) start() (err error) {
 // because setns support requires the C process to fork off a child and perform the setns
 // before the go runtime boots, we wait on the process to die and receive the child's pid
 // over the provided pipe.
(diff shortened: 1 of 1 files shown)


stevenh:
Not sure the failure is related to this change as its it doesn't look like its got to the new reaper yet.

Outstanding race?

crosbymichael:
I think reaping should be handled at the higher layers in here, this is what happens in runc and I don't think we have the same issues.  We probably shouldn't handle it here inside the lib.

cyphar:
I agree with what @crosbymichael said. If you want to make our signal handler code (`signals.go`) more generic so that you can use in in a higher layer then that sort of patch would be acceptable. But handling it transparently inside the library is definitely not correct because it means that higher levels have to go through more hoops to do things like collect return codes and so on.

PR #1307, opened 2017-02-03
Add --additional-gids to runc exec.

### LINKED ISSUE #1306: runc exec: cli flag to specify additional gids
`runc exec` only allows uid and primary gid as command line arguments and doesn't have an option to specify additional gids. You have to pass process.json (-p option) to specify additional gids.

This issue is to enhance runc exec to allow specifying additional gids on the command line as a new argument.

sumitsanghrajka:
This flag allows specifying additional gids for the process.
Without this flag, the user will have to provide process.json which allows additional gids.
Closes #1306

Signed-off-by: Sumit Sanghrajka <sumit.sanghrajka@gmail.com>

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -50,9 +50,9 @@ following will output a list of processes running in the container:
 			Name:  "user, u",
 			Usage: "UID (format: <uid>[:<gid>])",
 		},
-		cli.StringFlag{
+		cli.Int64SliceFlag{
 			Name:  "additional-gids, g",
-			Usage: "additional gids separated by comma",
+			Usage: "additional gids",
 		},
 		cli.StringFlag{
 			Name:  "process, p",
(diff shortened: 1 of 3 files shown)


GrantSeltzer:
urfave/cli provides the capability for an `IntSliceFlag` which can be used instead of forcing the user to specify by a comma and then splitting it out.

sumitsanghrajka:
@GrantSeltzer done.

GrantSeltzer:
@sumitsanghrajka Looks like the build is failing because you forgot to sign off on that last commit. Otherwise it looks good, I don't know if the int64 flag is necessary over the int one but I don't see why it'd be an issue other than for consistency.

sumitsanghrajka:
@GrantSeltzer gids can go upto 2^32 value, so int won't work hence using int64.

GrantSeltzer:
@sumitsanghrajka Ah, didn't catch that.

exec.go
```diff
@@ -204,5 +208,8 @@ func getProcess(context *cli.Context, bundle string) (*specs.Process, error) {
 		}
 		p.User.UID = uint32(uid)
 	}
+	for _, gid := range context.Int64Slice("additional-gids") {
+		p.User.AdditionalGids = append(p.User.AdditionalGids, uint32(gid))
```
cyphar: I would recommend doing a check to make sure `gid` isn't negative.

PR #1310, opened 2017-02-03
Correct container.Destroy() docs

stevenh:
Correct container.Destroy() docs to clarify that destroy can only operate on containers in specific states.


stevenh:
As discussed here: https://github.com/opencontainers/runc/pull/1297

crosbymichael:
LGTM

dqminh:
LGTM

PR #1312, opened 2017-02-06
libcontainer: selinux: fix DupSecOpt and DisableSecOpt

runcom:
`label.InitLabels` takes options as a string slice in the form of:

    user:system_u
    role:system_r
    type:container_t
    level:s0:c4,c5

However, `DupSecOpt` and `DisableSecOpt` were still adding a docker
specifc `label=` in front of every option. That leads to `InitLabels`
not being able to correctly init selinux labels in this scenario for
instance:

    label.InitLabels(DupSecOpt([%OPTIONS%]))

if `%OPTIONS` has options prefixed with `label=`, that's going to fail.
Fix this by removing that docker specific `label=` prefix.

### DIFF
--- a/libcontainer/selinux/selinux.go
+++ b/libcontainer/selinux/selinux.go
@@ -536,14 +536,14 @@ func DupSecOpt(src string) []string {
 		con["level"] == "" {
 		return nil
 	}
-	return []string{"label=user:" + con["user"],
-		"label=role:" + con["role"],
-		"label=type:" + con["type"],
-		"label=level:" + con["level"]}
+	return []string{"user:" + con["user"],
+		"role:" + con["role"],
+		"type:" + con["type"],
+		"level:" + con["level"]}
 }
 
 // DisableSecOpt returns a security opt that can be used to disabling SELinux
 // labeling support for future container processes
 func DisableSecOpt() []string {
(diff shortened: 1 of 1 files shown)


runcom:
@mrunalp @rhatdan PTAL

rhatdan:
:+1:

crosbymichael:
LGTM

crosbymichael:
On a side note, should we break this code out into a separate repo so that it can be maintained independently and have dedicated maintainers for this?

mrunalp:
LGTM

mrunalp:
@crosbymichael Yeah, that sounds good. I will see if we can get it under opencontainers.

rhatdan:
SGTM, and I volunteer to be the maintainer.  Others have asked for the SELinux go bindings, so getting this independent would probably be a good idea.

crosbymichael:
Ya, lets do that.  I think for something like this it would be better separate where the API can be versioned and have move implementations depending on it.  

Probably something like @rhatdan, @mrunalp, and @runcom  as maintainers?   Anyone else that I am missing?

rhatdan:
That is fine,  No one else unless I got someone from upstream SELinux to contribute on it.

PR #1313, opened 2017-02-06
libcontainer: rootfs_linux: support overlayfs

runcom:
As the runtime-spec allows it, we want to be able to specify overlayfs
mounts with:

    {
        "destination": "/etc/pki",
        "type": "overlay",
        "source": "overlay",
        "options": [
            "lowerdir=/etc/pki:/home/amurdaca/go/src/github.com/opencontainers/runc/rootfs_fedora/etc/pki"
        ]
    },

This patch takes care of allowing overlayfs mounts. Both RO and RW
should be supported.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -219,6 +219,24 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 			}
 		}
 		return nil
+	case "overlay":
+		// ensure that the destination of the bind mount is resolved of symlinks at mount time because
+		// any previous mounts can invalidate the next mount's destination.
+		// this can happen when a user specifies mounts within other mounts to cause breakouts or other
+		// evil stuff to try to escape the container's rootfs.
+		var err error
+		if dest, err = symlink.FollowSymlinkInScope(filepath.Join(rootfs, m.Destination), rootfs); err != nil {
+			return err
+		}
+		if err := checkMountDestination(rootfs, dest); err != nil {
+			return err
+		}
+		// update the mount with the correct dest after symlinks are resolved.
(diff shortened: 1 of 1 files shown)


runcom:
@rhatdan @mrunalp @rhvgoyal PTAL

libcontainer/rootfs_linux.go
```diff
@@ -219,6 +219,11 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 			}
 		}
 		return nil
+	case "overlay":
+		if err := mountPropagate(m, rootfs, mountLabel); err != nil {
+			return err
+		}
+		return label.SetFileLabel(m.Destination, mountLabel)
```
rhatdan: You should not call this.
The label would have to be specified in the overaly stuff above I believe.

runcom: removed

mrunalp:
Maybe we could move this to the default section so it could work for most other options. @crosbymichael WDYT?

crosbymichael:
Ya, probably better to do this in the default case and clean up anything else that does not require special handling.

runcom:
alright, the default case should work just fine for overlay then. If we need other stuff I'll add them.

crosbymichael:
@runcom i guess we mean, do we want to move the validation steps inside default?

runcom:
@crosbymichael well sure, I wasn't fully sure those were needed for other mounts other than binds, I'll reopen.

runcom:
can't reopen

crosbymichael:
you have to restore your branch

runcom:
yeah, I did another PR though (deeply sorry...) https://github.com/opencontainers/runc/pull/1314

PR #1314, opened 2017-02-06
libcontainer: rootfs_linux: support overlayfs

runcom:
As the runtime-spec allows it, we want to be able to specify overlayfs
mounts with:

    {
        "destination": "/etc/pki",
        "type": "overlay",
        "source": "overlay",
        "options": [
            "lowerdir=/etc/pki:/home/amurdaca/go/src/github.com/opencontainers/runc/rootfs_fedora/etc/pki"
        ]
    },

This patch takes care of allowing overlayfs mounts. Both RO and RW
should be supported.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -313,6 +313,19 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 			}
 		}
 	default:
+		// ensure that the destination of the mount is resolved of symlinks at mount time because
+		// any previous mounts can invalidate the next mount's destination.
+		// this can happen when a user specifies mounts within other mounts to cause breakouts or other
+		// evil stuff to try to escape the container's rootfs.
+		var err error
+		if dest, err = symlink.FollowSymlinkInScope(filepath.Join(rootfs, m.Destination), rootfs); err != nil {
+			return err
+		}
+		if err := checkMountDestination(rootfs, dest); err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


runcom:
@mrunalp @crosbymichael PTAL

crosbymichael:
@runcom actually a fun thing that I have been doing is that you can also mount with the destination as `/` to have it mount the rootfs for the contianer.  Its pretty cool.

runcom:
yeah, maybe use runtime config directly in docker to mount the rootfs (just wondering...)

mrunalp:
LGTM

libcontainer/rootfs_linux.go
```diff
@@ -313,6 +313,19 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 			}
 		}
 	default:
+		// ensure that the destination of the mount is resolved of symlinks at mount time because
+		// any previous mounts can invalidate the next mount's destination.
+		// this can happen when a user specifies mounts within other mounts to cause breakouts or other
+		// evil stuff to try to escape the container's rootfs.
+		var err error
+		if dest, err = symlink.FollowSymlinkInScope(filepath.Join(rootfs, m.Destination), rootfs); err != nil {
```
hqhq: s/filepath.Join(rootfs, m.Destination)/dest?

runcom: This is copy/pasted from the bind switch case..

hqhq: OK, we can fix that later then.

hqhq:
LGTM

Issue #1311, opened 2017-02-03
adding _LIBCONTAINER to environment breaks state


cyphar:
Because of some **very** questionable logic when determining the state of a container, if you have a container with `_LIBCONTAINER` anywhere in its environment then `runc state` will show the container as `created` not `running`.

This is because of this code (which isn't correct IMO):

```go
	// check if the process that is running is the init process or the user's process.
	// this is the difference between the container Running and Created.
	environ, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/environ", pid))
	if err != nil {
		return Stopped, newSystemErrorWithCausef(err, "reading /proc/%d/environ", pid)
	}
	check := []byte("_LIBCONTAINER")
	if bytes.Contains(environ, check) {
		return Created, nil
	}
	return Running, nil

```

cyphar:
This is causing issues with rootless containers when trying to make `!dumpable` also safe with rootless containers.

hqhq:
Can we check the status of fifo file to get if the container is in `created` state?

crosbymichael:
We can have something like `runc start` update that state or to keep this check dynamic we could maybe look at `/proc/<pid>/exe` readlilnk on it to see if its the runc binary( though this is hard to match up ).  

Checking if the fifo is there or not could work as we delete it after start.  Its probably our best option.

PR #1317, opened 2017-02-09
Fix state when _LIBCONTAINER in environment

hqhq:
Fixes: #1311

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1259,14 +1259,9 @@ func (c *linuxContainer) runType() (Status, error) {
 	if !exist || err != nil {
 		return Stopped, err
 	}
-	// check if the process that is running is the init process or the user's process.
-	// this is the difference between the container Running and Created.
-	environ, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/environ", pid))
-	if err != nil {
-		return Stopped, newSystemErrorWithCausef(err, "reading /proc/%d/environ", pid)
-	}
-	check := []byte("_LIBCONTAINER")
-	if bytes.Contains(environ, check) {
+	// We'll create exec fifo and blocking on it after container is created,
+	// and delete it after start container.
+	if _, err := os.Stat(filepath.Join(c.root, execFifoFilename)); err == nil {
 		return Created, nil
 	}
 	return Running, nil
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -191,17 +191,29 @@ func (c *linuxContainer) Start(process *Process) error {
 	if err != nil {
 		return err
 	}
-	return c.start(process, status == Stopped)
+	if status == Stopped {
+		if err := c.createExecFifo(); err != nil {
+			return err
+		}
+	}
+	if err := c.start(process, status == Stopped); err != nil {
+		if status == Stopped {
+			c.deleteExecFifo()
+		}
+		return err
+	}
+	return nil
 }
 
 func (c *linuxContainer) Run(process *Process) error {
 	c.m.Lock()
-	defer c.m.Unlock()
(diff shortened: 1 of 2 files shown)


crosbymichael:
Looks like this messes with C/R 

`not ok 3 checkpoint and restore`

hqhq:
@crosbymichael Yeah it's because restore always recreate that fifo, I pushed a commit to clean that up, ideally restore should've never created that in the first place if it won't use it, but add a new parameter to `factory.Create` seems overdo, not sure if there are other ways.

restore.go
```diff
@@ -137,7 +138,10 @@ func restoreContainer(context *cli.Context, spec *specs.Spec, config *configs.Co
 		if err != nil {
 			return -1, err
 		}
+		// TODO: restore should not create this fifo in the first place.
+		cleanupExecFifo(context)
```
cyphar: Yeah, I agree this a bit ugly. I think the problem boils down to the fact that `libcontainer.Factory.Create` is being used as both a "real" create (for `runc create`) but also as a "fake" create (in `runc restore`) in order to effectively make it look like `.Load` worked.

Maybe there's a nicer way of handling this? Should we move the stateful stuff out of `Create` into something separate? I'm not sure.

hqhq:
@crosbymichael @cyphar I've updated this PR, move exec fifo's creation out of factory.Create, PTAL.

libcontainer/container_linux.go
```diff
@@ -191,17 +191,28 @@ func (c *linuxContainer) Start(process *Process) error {
...
+	}
+	return nil
 }
 
 func (c *linuxContainer) Run(process *Process) error {
 	c.m.Lock()
-	defer c.m.Unlock()
 	status, err := c.currentStatus()
 	if err != nil {
 		return err
```
crosbymichael: Need an unlock here

libcontainer/container_linux.go
```diff
@@ -291,6 +302,37 @@ func (c *linuxContainer) Signal(s os.Signal, all bool) error {
...
+	if err != nil {
+		return err
+	}
+	rootgid, err := c.Config().HostGID()
+	if err != nil {
+		return err
+	}
+
+	fifoName := filepath.Join(c.root, execFifoFilename)
+	if _, err := os.Stat(fifoName); err == nil {
```
crosbymichael: For this, should we just call mkfifo and handle an eexist error?   Does it generate an eexist error ?

hqhq: I don't think so, `syscall.Mkfifo` will call `mknodat` syscall, and according to https://linux.die.net/man/2/mknodat ,there is no EEXIST error.

crosbymichael: ok, SGTM

libcontainer/container_linux.go
hqhq: Updated.

crosbymichael:
LGTM

libcontainer/container_linux.go
```diff
@@ -291,6 +303,37 @@ func (c *linuxContainer) Signal(s os.Signal, all bool) error {
...
+		return fmt.Errorf("exec fifo %s already exists", fifoName)
+	}
+	oldMask := syscall.Umask(0000)
+	if err := syscall.Mkfifo(fifoName, 0622); err != nil {
+		syscall.Umask(oldMask)
+		return err
+	}
+	syscall.Umask(oldMask)
```
cyphar: This is an aside, but is there a reason that we don't do `syscall.Chmod` manually? I don't really trust `umask(2)` when it comes to Go programs (where you can have goroutines run code in your current thread in-between different statements).

cyphar:
LGTM. We can clean up the `umask` stuff later.

Issue #986, opened 2016-08-15
Container Checkpoint Pre-Dump Enhancement


washraf:
Criu the tool used for container checkpoint and restoration contains a great enhancement for checkpoint time (The Actual service down time). https://criu.org/Iterative_migration#Pre-dump.

Runc should checkpoint the container using this method. Which is the valid method for check-pointing even in Virtual Machines. and this is must incase of container migration to decrease the migration down time.

marcosnils:
IMHO pre-dump shouldn't be set by default. I think it's best if we just allow the CRIU options in runC so the user decides how/when to do pre-dump migrations.

FWIW we've added the parameters in this branch https://github.com/opencontainers/runc/compare/master...marcosnils:pre_dump?expand=1 for the CMT project but we never submitted a PR because we didn't add any tests.

washraf:
@marcosnils @cyphar .. Any updates on this matter ??

marcosnils:
@washraf I just giving my opinion on how I would implement it in runC. I don't have time to craft a proper the PR from my branch. If you need this feature it'd be nice if you could craft a PR to include it in runC. As you'll probably use it and provide real feedback about the functionality.

cyphar:
I defer to the CRIU expert around here (@avagin). Feel free to write a PR, but his opinion is probably what holds the most water with regards to checkpoint-restore usecases.

x1022as:
Sorry for missing this issue before proposing a pr. #1001 is about adding pre-dump to runc checkpoint.

kolyshkin:
`runc checkpoint` supports `--pre-dump` option now

PR #1318, opened 2017-02-09
checkpoint: add pre-dump support

adrianreber:
This is based on existing pull requests to implement pre-dump support in
runc. In addition to rebasing this tries to address the open questions
in the previous pull request.

One questions was why the code was no longer waiting (cmd.Process.Wait())
for CRIU in the pre-dump case. This is related to how CRIU expects a
pre-dump to work. Once a pre-dump is started the CRIU RPC process goes
into a loop where it either expects further pre-dumps or the final dump.
From CRIU's point of view this makes sense as it expects the user (runc)
to make as many pre-dumps as necessary and then the final dump.
In its final implementation that is probably also what runc should do.
In this first approach the incremental checkpoints are created manually
by runc as well as the final checkpoint. Therefore the code is still not
waiting for CRIU to end, because it never will. But if we get the RPC
message that the pre-dump was successful CRIU is just terminated.

Instead of passing the pre-dump as additional variable to Checkpoint()
the information about doing a pre-dump is now part of criuOpts.

If referencing a parent image for pre-dump mode the option TrackMem is
now passed to CRIU as it is required.

Examples:

 1. Dump:
  runc checkpoint <container-id> --pre-dump --image-path parent
 2. Final Dump:
  runc checkpoint <container-id> --image-path image --parent-path ../parent
 3: Restore:
  runc restore <container-id> --image-path image

In one test the initial dump was 182M.  The final dump only 6.6M which
decreased the downtime to the time required to transfer 6.6M instead of
182M.


adrianreber:
This is a rebase of #1001 and related to #986

adrianreber:
@avagin can please have a look

hqhq:
@adrianreber AFAICS, https://github.com/opencontainers/runc/pull/1001 is still under discussion and just lack of review after it's last push, so you shouldn't cherry it if @x1022as is still interested in improving it.

avagin:
@adrianreber have you looked at the new libphaul? 
https://github.com/criupatchwork/criu/commit/3fdad140efc49540c513f60786c2d2c26d38587a

adrianreber:
@avagin, yes I looked at that GO implementation of p.haul and I think that in the end this should be the way to do pre-copy migration in runc. I also chatted with @xemul on IRC about it.

My goal with this PR (or #1001) was to get the initial pre-dump version merged and then look how the p.haul based approach works. The logic to do multiple pre-dumps until certain thresholds (number of iterations or number of dirty pages) have been reached sounds like an important improvement.

I have also some patches to use the FEATURE_CHECK RPC mechanism to verify that criu can actually do dirty page tracking which I wanted to use instead of comparing versions.

As a first step, however, I would like to see #1001 or this PR merged.

adrianreber:
Closing as #1001 has been changed to address all open reviewers questions. Thanks @x1022as

PR #1320, opened 2017-02-10
Add --preserve-file-descriptors=N to create

ijc25:
This preserves the given number of file descriptors on top of the 3 stdio, that
is, fds [3..3+N) will be passed through.

Since Socket Activation (via $LISTEN_FDS) also claims fd 3 onwards the use of
this new flag is incompatible with socket activation.

### DIFF
--- a/create.go
+++ b/create.go
@@ -46,6 +46,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "no-new-keyring",
 			Usage: "do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key",
 		},
+		cli.IntFlag{
+			Name:  "preserve-fds",
+			Usage: "Pass N additional file descriptors to the container (stdio + $LISTEN_FDS + N in total)",
+		},
 	},
 	Action: func(context *cli.Context) error {
 		if err := checkArgs(context, 1, exactArgs); err != nil {
--- a/run.go
+++ b/run.go
@@ -57,6 +57,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "no-new-keyring",
 			Usage: "do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key",
 		},
+		cli.IntFlag{
+			Name:  "preserve-fds",
+			Usage: "Pass N additional file descriptors to the container (stdio + $LISTEN_FDS + N in total)",
+		},
 	},
 	Action: func(context *cli.Context) error {
 		if err := checkArgs(context, 1, exactArgs); err != nil {
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -180,6 +180,7 @@ type runner struct {
 	shouldDestroy   bool
 	detach          bool
 	listenFDs       []*os.File
+	preserveFDs     int
 	pidFile         string
 	consoleSocket   string
 	container       libcontainer.Container
@@ -196,9 +197,15 @@ func (r *runner) run(config *specs.Process) (int, error) {
 		r.destroy()
 		return -1, err
 	}
+
 	if len(r.listenFDs) > 0 {
 		process.Env = append(process.Env, fmt.Sprintf("LISTEN_FDS=%d", len(r.listenFDs)), "LISTEN_PID=1")
 		process.ExtraFiles = append(process.ExtraFiles, r.listenFDs...)
+	}
+
+	baseFd := 3 + len(process.ExtraFiles)
+	for i := baseFd; i < baseFd+r.preserveFDs; i++ {
+		process.ExtraFiles = append(process.ExtraFiles, os.NewFile(uintptr(i), "PreserveFD:"+strconv.Itoa(i)))
 	}
 
 	rootuid, err := r.container.Config().HostUID()
@@ -353,6 +360,7 @@ func startContainer(context *cli.Context, spec *specs.Spec, create bool) (int, e
 		consoleSocket:   context.String("console-socket"),
 		detach:          context.Bool("detach"),
 		pidFile:         context.String("pid-file"),
+		preserveFDs:     context.Int("preserve-fds"),
 		create:          create,
 	}
 	return r.run(&spec.Process)


utils_linux.go
```diff
@@ -196,11 +197,20 @@ func (r *runner) run(config *specs.Process) (int, error) {
 		r.destroy()
 		return -1, err
 	}
+
+	if len(r.listenFDs) > 0 && r.preserveFDs > 0 {
+		return -1, fmt.Errorf("Preserve FDs and socket activation (LISTEN_FDS) are mutually incompatible")
+	}
```
ijc25: This feels like a bit of a cop out, but I couldn't think of a better way. The only alternative I thought of was to specify that preserveFds preserves N over and above `${LISTEN_FDS:-0}` but that seemed pretty fugly. It seems unlikely to me that one would write an application which uses both socket activation and expects other random open fds to be present upon exec.

create.go
```diff
@@ -46,6 +46,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "no-new-keyring",
 			Usage: "do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key",
 		},
+		cli.IntFlag{
+			Name:  "preserve-file-descriptors",
```
crosbymichael: This is a pretty long flag name, Maybe go for something like `--preserve-fd`?

utils_linux.go
crosbymichael: You an just start by using `len(process.ExtraFiles)`, adding 3 stdio and always preserve from what is already in the `ExtraFiles`

create.go
ijc25: Good idea. I think I'll go for `--preserve-fds` to keep it plural. Maybe `--nr-preserve-fds` would be clearer though (but then the length creeps back up...). `--nr-keep-fds`?

utils_linux.go
ijc25: @crosbymichael thanks, so to check I understand, you are suggesting that `--preserve-fds` should be cumulative over `SD_LISTEN` (or anything else which extends `process.ExtraFiles` in the future). IOW given `LISTEN_FDS=3` and `--preserve-fds=5` then we would preserve `[3, 11)` (in addition to stdio). That's 3, 4 & 5 from `LISTEN_FDS=3` and 6, 7, 8, 9 & 10 from `--preserve-fds=5`.

I'll do that.

crosbymichael:
I still think the `nr-` is kinda verbose.  I don't even know what `nr` means. ;)

dqminh:
@ijc25 i'm curious, do you have a usecase for this noted down somewhere ?

ijc25:
@crosbymichael `nr` is an abbreviation for `number` (maybe not as commonly used as I thought). I added it to indicate that the argument was the count of such things, rather than a list of them or whatever. I'll happily drop it if you prefer though.

@dqminh basically I'd like to pass an open socket or file fd to my application so I can reduce its privileges to not include opening sockets/files.

rhatdan:
I also don't like -nr, have not seen this in a CLI before.

crosbymichael:
LGTM

mrunalp:
Could you either squash or rework the commits?

create.go
```diff
@@ -46,6 +46,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "no-new-keyring",
 			Usage: "do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key",
 		},
+		cli.IntFlag{
+			Name:  "preserve-fds",
+			Usage: "Pass N additional file descriptors to the container (stdio + $LISTEN_FDS + N in total)",
+		},
```
hqhq: Shall we also add this to run.go? They almost have same options.

ijc25: Yes, I think that would make sense.

ijc25: They were even similar enough that `git show create.go | patch run.go` did the right thing ;-)

ijc25:
I see how adding the option to `run.go` has caused the `continuous-integration/travis-ci/pr` failure. [Failure](https://travis-ci.org/opencontainers/runc/jobs/202572518#L2541) is:

```
--- FAIL: TestExecIn (0.21s)

	utils_test.go:51: execin_test.go:42: unexpected error: container_linux.go:247: starting container process caused "process_linux.go:270: running exec setns process for init caused \"exit status 26\""

```

I see similar messages in some other recent master failures e.g. 
https://travis-ci.org/opencontainers/runc/jobs/198764647#L2632
https://travis-ci.org/opencontainers/runc/jobs/201936653#L2642
Is there some flakiness in CI or shall I continue to investigate?

crosbymichael:
LGTM

utils_linux.go
```diff
@@ -196,11 +197,17 @@ func (r *runner) run(config *specs.Process) (int, error) {
...
 	}
+
 	if len(r.listenFDs) > 0 {
 		process.Env = append(process.Env, fmt.Sprintf("LISTEN_FDS=%d", len(r.listenFDs)), "LISTEN_PID=1")
 		process.ExtraFiles = append(process.ExtraFiles, r.listenFDs...)
 	}
 
+	baseFd := 3 + len(process.ExtraFiles)
+	for i := baseFd; i < baseFd+r.preserveFDs; i++ {
+		process.ExtraFiles = append(process.ExtraFiles, os.NewFile(uintptr(i), "PreserveFD"))
```
mrunalp: Should the name be different for each fd instead of "PreserveFD"? Maybe "PreserveFD:i" ?

crosbymichael: It would be nicer for debugging to do that, It does not really make a difference functionality wise.

ijc25: Done.

cyphar:
@ijc25 It's not a flakiness in the CI, it was a race condition that we've been hitting a lot recently. It's been fixed in #1237.

run.go
```diff
@@ -57,6 +57,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "no-new-keyring",
 			Usage: "do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key",
 		},
+		cli.IntFlag{
+			Name:  "preserve-fds",
+			Usage: "Pass N additional file descriptors to the container (stdio + $LISTEN_FDS + N in total)",
```
cyphar: If we're going to have this sort of API, I would prefer that we make it more explicit than this. For example, make it so that you have to specify **which** file descriptors you want to pass (`IntSliceFlag`). This would also mean that cases where you don't want to use low file descriptors (see [this SQLite option for more reasons why that might matter](https://www.sqlite.org/compile.html#minimum_file_descriptor)).

cyphar: So it would look like

```
runc run --preserve-fd 512 --preserve-fd 1337 ctr
```

ijc25: The current interface doesn't allow you to pass through fds 0, 1, or 2 in this way so I think [the SQLlite reasoning](https://www.sqlite.org/compile.html#minimum_file_descriptor) for avoiding those fds doesn't really apply here.

I can take a look switching things to behave like you suggest but it would have to be contingent on Go's [`os/exec`](https://golang.org/pkg/os/exec/) `Cmd.ExtraFiles` supporting the necessary semantics (i.e. a bug slice full of `nil`s apart from at offset `512` and `1337` in your example). Personally I don't think it is terribly useful to be able to pass arbitrary fds, it's pretty easy (and arguably more controlled) for the caller to `dup2` things into the right place after `fork` but before `exec` of runc.

crosbymichael: For this i'm not sure defining individual fds is optimal.  If you have a large amount of fds to pass you would run out of space for all the `--preserve-fd 3 --preserve-fd 4 --preserve-fd 5...`

ijc25: That's a good point. I was a bit surprised that `IntSliceFlag` didn't support comma separated lists or even `N-M` spans (or if it does that the syntax wasn't more obvious...).

FWIW I did confirm that `Cmd.ExtraFiles` looks on first glance to DTRT with `nil` in the list, although `finalizeNamespace` and `CloseExecFrom` would probably need extending to handle a sparse fd space to be absolutely safe.

dqminh:
> @dqminh basically I'd like to pass an open socket or file fd to my application so I can reduce its privileges to not include opening sockets/files.

Thanks the the clarification ! I was wondering why reusing the `listen_*` interface was not reasonable enough, but i guess that setting the`listen_*` env variable is also abit more complex than just passing them as CLI arg, especially when we know the command to be run.

crosbymichael:
@dqminh ya, and it would be overloading the systemd functionality that people expect with listen_

crosbymichael:
Thanks for squashing 

LGTM

dqminh:
LGTM.

PR #1334, opened 2017-02-20
update all the vendored libs to the latest versions

mcuadros:
The current build is broken if you try to use libcontainer as library in other project, since is using a old version of runtime-spce.

I tried to update the vendoring, but I found that its using a old version. I just updated the vendoring to the latest version for all the packages and to use the latest Godep version (was a very old version for 1.5.x)

Also I updated the CI, to run the tests properly with the changes introduced by Godep

One test is [failing](https://travis-ci.org/opencontainers/runc/jobs/203409870) just tell me if the tests is wrong now and I should change it, or I should fix the behavior of the function:

```
time="2017-02-20T11:52:41Z" level=warning msg="os: process already finished" 
--- FAIL: TestExecInError (0.32s)
	execin_test.go:245: executable file not found error not delivered to stdio:
```


hqhq:
@mcuadros Thanks for doing this, can you separate these changes to different PRs? Such like move Godep update into a PR and show how you did it, so we can make sure you're doing it right. Otherwise it's hard to review all these changes.

Thanks.

cyphar:
> The current build is broken if you try to use libcontainer as library in other project, since is using a old version of runtime-spce.

If the issue is an old version of `runtime-spec` then please only update that vendor and not the entire vendor folder. The current state of this PR is effectively impossible for us to review.

Also, yeah we've been planning on fixing the issue that is us using `godep` for a while. Sorry that you had to use it to do this PR.

cyphar:
> One test is failing just tell me if the tests is wrong now and I should change it, or I should fix the behavior of the function:

If that test is failing then this PR has broken something with regards to how console handling is done internally. I can't really help more than that unless you break up the PR a bit more -- it's a bit difficult to figure out which part of the 100kline diff broke it.

Do the integration tests pass for you (`make integration`)?

mcuadros:
Is true that is hard to review, I will do it tomorrow morning. 

The main problem is that the old Godeps is complete obsolete, I cannot run any command on-top of it. I will try to do it by hand.

So I will make a PR updating the Godeps to the new version and after updating the libraries.

mcuadros:
I already split this is the first PR https://github.com/opencontainers/runc/pull/1337

PR #1335, opened 2017-02-20
update go version at travis-ci

mcuadros:
Travis is testing against 1.6.x and 1.7.x if the intent is test againts the last two stable version we should remove the 1.6 and add the 1.8

Also the tool being use for the validation of bash scripts (https://github.com/mvdan/sh) is only compatible with 1.7>

Also I added tip as an allowed to fail version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
 language: go
 go:
-  - 1.6.x
   - 1.7.x
-# - master
+  - 1.8.x
+  - tip
+
+matrix:
+  allow_failures:
+    - go: tip
 
 # `make ci` uses Docker.
 sudo: required
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.7.4
+FROM golang:1.8.0
 
 # libseccomp in jessie is not _quite_ new enough -- need backports version
 RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/backports.list


cyphar:
Personally I would /prefer/ if we keep testing Go `1.6.x` -- mainly because currently there are still stable distributions which use Go 1.6. Testing another version is fairly cheap compared to breaking things so downstreams need to deal with it.

Unfortunately the failures on `master` are because one of the tools we use (`shfmt`) are broken on `1.6.x`. :/

crosbymichael:
@cyphar I don't think we need to support go 1.6.  If people are using it then it can be their responsibility.  There are many improvements with 1.7-1.8 and security fixes that we can take advantage of.  

Also this is not a runtime issue, its a buildtime issue so its not an impact on users.  Since we are pre 1.0 I think its an ok move to make, there is no reason to live on an old build of Go and not take advantage of the improvements.

LGTM

cyphar:
@crosbymichael Fair enough.

LGTM.

mvdan:
FWIW, you can have the best of both worlds. You can keep 1.6.x and any older Go releases as long as you only run tools like `vet`, `gofmt` and `shfmt` on the latest Go release. Running them on older releases is flakier and wastes resources anyway.

mcuadros:
Compiling against old version of Go, IMHO, is a bad practice. We must help to the language evolve fast, updating to the latest versions ASAP.

mvdan:
@mcuadros sure, but that's unrelated to supporting older Go versions :) Most projects support the last 2-3 minor releases, but you could just as well do 4 or more while adopting the newest version as soon as it's out.

crosbymichael:
The problem is that go stdlib changes and when you use a new package it breaks older versions anyways.

PR #1338, opened 2017-02-22
ps/update: ensure container status at first

sak0:
ps: add check step
update: check status at first


crosbymichael:
Why are you making this change?  I does not make sense to me to remove it from an internal method and duplicate it in two other places.

hqhq:
This looks wrong to me, it'll make libcontainer vulnerable to use.

sak0:
@crosbymichael @hqhq 
thanks for review.
may i keep the change for `runc ps`,  it seems that ps a stopped container is useless.

PR #1339, opened 2017-02-22
Don't override system error

cpuguy83:
The error message added here provides no value as the caller already
knows all the added details. However it is covering up the underlying system error (typically `ENOTSUP`). There is no way to correctly handle this error before this change.

### DIFF
--- a/libcontainer/label/label_selinux.go
+++ b/libcontainer/label/label_selinux.go
@@ -142,7 +142,7 @@ func Relabel(path string, fileLabel string, shared bool) error {
 		fileLabel = c.Get()
 	}
 	if err := selinux.Chcon(path, fileLabel, true); err != nil {
-		return fmt.Errorf("SELinux relabeling of %s is not allowed: %q", path, err)
+		return err
 	}
 	return nil
 }


crosbymichael:
LGTM

hqhq:
What would the error message look like on Docker side after this change?

cpuguy83:
@hqhq https://github.com/opencontainers/runc/blob/master/libcontainer/system/xattrs_linux.go#L50
It's the raw `syscall.Errno`.

hqhq:
@cpuguy83 I know after this change, runc will return raw `syscall.Errno` in this case, just don't get it how this gonna help diagnose https://github.com/docker/docker/issues/31255 , current error message in that issue
> Received the error docker: Error response from daemon: SELinux relabeling of /var/lib/docker/volumes/footest/_data is not allowed: "operation not supported".

Looks pretty straight to me, what would it look like after this change? Do I miss something?

cpuguy83:
@hqhq Yes we know what the issue is, and are already handling it: https://github.com/docker/docker/blob/v1.13.1/container/container_unix.go#L199

But since this error got wrapped our check doesn't work anymore, and there's no way to handle it without string comparison.

hqhq:
@cpuguy83 Got it, thanks for clarifying.

LGTM

cyphar:
As an aside, we really should be switching to `github.com/pkg/errors` and use `errors.Wrap`.

PR #1344, opened 2017-02-24
fix cpu.cfs_quota_us changed when systemd daemon-reload using systemd.

xuxinkun:
Close https://github.com/docker/docker/issues/30121

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -269,6 +269,13 @@ func (m *Manager) Apply(pid int) error {
 			newProp("CPUShares", uint64(c.Resources.CpuShares)))
 	}
 
+	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
+	if c.Resources.CpuQuota != 0 && c.Resources.CpuPeriod != 0 {
+		cpuQuotaPerSecUSec := c.Resources.CpuQuota * 1000000 / c.Resources.CpuPeriod
+		properties = append(properties,
+			newProp("CPUQuotaPerSecUSec", uint64(cpuQuotaPerSecUSec)))
+	}
+
 	if c.Resources.BlkioWeight != 0 {
 		properties = append(properties,
 			newProp("BlockIOWeight", uint64(c.Resources.BlkioWeight)))


libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -269,6 +269,19 @@ func (m *Manager) Apply(pid int) error {
...
 	}
 
+	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
+	// The percentage cpuQuota refered specifies how much CPU time the unit
+	// shall get at maximum, relative to the total CPU time available on one CPU.
+	// The percentage should be greater than 1%.
+	if c.Resources.CpuQuota != 0 && c.Resources.CpuPeriod != 0 {
+		cpuQuota := c.Resources.CpuQuota * 100 / c.Resources.CpuPeriod
+		if cpuQuota == 0 {
+			return fmt.Errorf("cpu_quota divided by cpu_period must be greater than 0.01")
```
cyphar: Why? Where does this requirement come from (if it's a kernel requirement -- which I assume it is -- then we generally don't error out for cases like this)?

xuxinkun: Because the systemd can not accept that the percentage(cpu_quota divided by cpu_period) is less than 0.01. For example,  if the user let CpuQuota be 10 and CpuPeriod be 10000, the cpu.cfs_quota_us will be -1, not 10. So, I think it is necessary to warn the user that the percentage must be greater than 0.01.

cyphar: Can you provide a link to the systemd documentation which describes this option? I can't find it [here](https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html). If it's an option added recently, you'll need to make your change backwards compatible with older systemd versions (by conditionally adding it like we do with `Delegate` for example).

xuxinkun: Actually, I did not find it from document. By observing, I found the percentage cannot be lower than 0.01. I found that if i set the percentage lower than 0.01, the cpu.cfs_quota_us will not be set as it should be set. 
How about this idea? I delete this check and add docs about it to warn users.

xuxinkun:
@cyphar  I modified the code. Pls review it.

cyphar:
LGTM.

xuxinkun:
@dqminh May you help review?

crosbymichael:
LGTM

PR #1346, opened 2017-02-24
container.Processes(): make sure the container is not stopped

sak0:
Stopped is the status that denotes the container does not have a created or running process.
it seems that `runc ps` a stopped container is useless.

### DIFF
--- a/ps.go
+++ b/ps.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/opencontainers/runc/libcontainer"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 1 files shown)


ps.go
```diff
@@ -32,6 +33,13 @@ var psCommand = cli.Command{
 		if err != nil {
 			return err
 		}
+		status, err := container.Status()
+		if err != nil {
+			return err
+		}
+		if status == libcontainer.Stopped {
+			return fmt.Errorf("cannot ps a container that has run and stopped")
```
hqhq: cannot ps a container that has stopped.

Container can be in stopped status from created status, without running.

cyphar:
I don't feel this is necessary. It's not like `runc ps` gives you incorrect results in this case -- it accurately tells you there are no processes inside the container...

crosbymichael:
@sak0 is there a use case that this is causing issues with for you? If not, I would probably side with @cyphar and say that its really not necessary to do a status check here as it returns the correct values.

sak0:
thanks for all comments.
there is no specially issues util now. I'll close it.

PR #1349, opened 2017-02-25
tty: dup /dev/null over stdio for (detach && !terminal)

cyphar:
This cleans up some of the terminal=false semantics for detached
containers. One of the biggest issues with running `runc create` inside
a terminal is that it gave the container process access to a host tty.
While this is not a security issue in of itself, it is quite concerning
(and similar bugs have caused security issues in LXC in the past). Not
to mention that this meant that the user's shell was borked -- something
which should be fixed anyway.

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -3,114 +3,121 @@
 load helpers
 
 function setup() {
-  teardown_busybox
-  setup_busybox
+	teardown_busybox
+	setup_busybox
 }
 
 function teardown() {
-  teardown_busybox
+	teardown_busybox
+}
+
+function __test_checkpoint_and_restore() {
+	requires criu
+
+	opt_detach=""
+	opt_predump=""
+	while getopts "td" optch; do
+		case "$optch" in
+			d)
+				opt_detach="1"
+				;;
+			p)
+				opt_predump="1"
+				;;
+		esac
+	done
+
+	# criu does not work with external terminals so..
+	# setting terminal and root:readonly: to false
+	sed -i 's;"terminal": true;"terminal": false;' config.json
+	sed -i 's;"readonly": true;"readonly": false;' config.json
+	sed -i 's/"sh"/"sh","-c","while :; do date; sleep 1; done"/' config.json
+
+	run_args=()
+	if [[ "$opt_detach" ]]; then
+		run_args+=("-d")
+	fi
+
+	# run busybox
+	if [[ "$opt_detach" ]]; then
+		runc run "${run_args[@]}" test_busybox
+		[ "$status" -eq 0 ]
+	else
+		(
+			runc run "${run_args[@]}" test_busybox
+			[ "$status" -eq 0 ]
+		) &
+	fi
+
+	# check state
+	wait_for_container 15 1 test_busybox
+
+	runc state test_busybox
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"running"* ]]
+
+	# if you are having problems getting criu to work uncomment the following dump:
+	#cat /run/opencontainer/containers/test_busybox/criu.work/dump.log
+
+	restore_args=()
+	checkpoint_args=()
+	if [[ "$opt_predump" ]]; then
+		mkdir parent-dir
+		restore_args=("--image-path" "./image-dir")
+		checkpoint_args=("--image-path" "./image-dir" "--parent-path" "./parent-dir")
+
+		# do a pre-dump checkpoint
+		runc --criu "$CRIU" checkpoint --pre-dump --image-path ./parent-dir test_busybox
+		[ "$status" -eq 0 ]
+
+		# busybox should still be running
+		runc state test_busybox
+		[ "$status" -eq 0 ]
+		[[ "${output}" == *"running"* ]]
+	fi
+
+	# checkpoint the running container
+	mkdir image-dir
+	runc --criu "$CRIU" checkpoint "${checkpoint_args[@]}" test_busybox
+	[ "$status" -eq 0 ]
+
+	# after checkpoint busybox is no longer running
+	runc state test_busybox
+	[ "$status" -ne 0 ]
+
+	# restore from checkpoint
+	if [[ "$opt_detach" ]]; then
+		runc --criu "$CRIU" restore "${restore_args[@]}" "${run_args[@]}" test_busybox
+		[ "$status" -eq 0 ]
+	else
+		(
+			runc --criu "$CRIU" restore "${restore_args[@]}" "${run_args[@]}" test_busybox
+			[ "$status" -eq 0 ]
+		) &
+	fi
+
+	# check state
+	wait_for_container 15 1 test_busybox
+
+	# busybox should be back up and running
+	runc state test_busybox
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"running"* ]]
+
 }
 
 @test "checkpoint and restore" {
-  requires criu
-
-  # criu does not work with external terminals so..
-  # setting terminal and root:readonly: to false
-  sed -i 's;"terminal": true;"terminal": false;' config.json
-  sed -i 's;"readonly": true;"readonly": false;' config.json
-  sed -i 's/"sh"/"sh","-c","while :; do date; sleep 1; done"/' config.json
-
-  (
-    # run busybox (not detached)
-    runc run test_busybox
-    [ "$status" -eq 0 ]
-  ) &
-
-  # check state
-  wait_for_container 15 1 test_busybox
-
-  runc state test_busybox
-  [ "$status" -eq 0 ]
-  [[ "${output}" == *"running"* ]]
-
-  # checkpoint the running container
-  runc --criu "$CRIU" checkpoint test_busybox
-  # if you are having problems getting criu to work uncomment the following dump:
-  #cat /run/opencontainer/containers/test_busybox/criu.work/dump.log
-  [ "$status" -eq 0 ]
-
-  # after checkpoint busybox is no longer running
-  runc state test_busybox
-  [ "$status" -ne 0 ]
-
-  # restore from checkpoint
-  (
-    runc --criu "$CRIU" restore test_busybox
-    [ "$status" -eq 0 ]
-  ) &
-
-  # check state
-  wait_for_container 15 1 test_busybox
-
-  # busybox should be back up and running
-  runc state test_busybox
-  [ "$status" -eq 0 ]
-  [[ "${output}" == *"running"* ]]
+	__test_checkpoint_and_restore
 }
 
-@test "checkpoint(pre-dump) and restore" {
-  requires criu
+@test "checkpoint and restore [detach]" {
+	__test_checkpoint_and_restore -d
+}
 
-  # criu does not work with external terminals so..
-  # setting terminal and root:readonly: to false
-  sed -i 's;"terminal": true;"terminal": false;' config.json
-  sed -i 's;"readonly": true;"readonly": false;' config.json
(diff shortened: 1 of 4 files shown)


cyphar:
@mrunalp Looks like `rh-atomic-bot` doesn't know that it needs to have `busybox` in `/busybox`.

mrunalp:
I will follow up on that.

hqhq:
ping @crosbymichael is there any specific semantics for detach mode or any other concerns about this?

this LGTM to me as I always think the io for detached mode behavior looks weird.

hqhq:
LGTM

cyphar:
I was planning on also extending the `c/r` tests, so that we can actually make sure that it works with:

1. `--console-socket` (which it /should/ though I'm not sure how we're meant to get the master pty).
2. `--detach` (now that it no longer passes the parent PTY).

cyphar:
/ping @mrunalp @crosbymichael PTAL. I've expanded the `c/r` tests to include this new case.

cyphar:
/ping @hqhq @mrunalp @crosbymichael PTAL. I've expanded the `c/r` tests to include this new case.

crosbymichael:
Sorry this is not correct at all.  The current function only dups over if the stdio is set as dev/null initially.  

If you do this there is no way to get input/outpt of a container that does not have a tty.  You can verify this easy by doing an `echo hi` and running the container in detached, you will not get any output for this patch.

crosbymichael:
You can use runc from the CLI and it works really well but features like detach are for use with a larger system on top of it.  Its weird if you think of launching it from a cli with detach when you don't care about the output but launching runc from anything else with detach is a way to have the container launched without keeping runc alive the entire time. 

How are you supposed to get pipes, fifos, anything as the container's STDIO if you don't pass it as runc's stdio on create/run -d?  There is no other way to do it and this is the most common way to pass STDIO to processes on unix.  This change just ignores that and sets everything into the black hole of `/dev/null` never to be seen again.   It breaks every container that ever outputs or gets anything piped into it without a tty. 

```
# broken
docker run alpine echo hi
#broken
echo hi | docker run alpine cat 
#broken
docker run alpine cat /somefile
```

mrunalp:
How about we allow callers the options to:
1. inherit runtime stdio which is today's behavior.
2. Pass fds for stdout/stderr of the container. This will simplify logging. The runtime can open files and pass them down to the container and higher level systems could just read those files.

cyphar:
@crosbymichael 

> You can verify this easy by doing an echo hi and running the container in detached, you will not get any output for this patch.

That was intentional, but the reason behind this change is that the current system of just passing host file descriptors isn't safe IMO. It (in most cases) exposes host ptys or host pipes to the container. For the pty case, this can result in a security issue (it happened with LXC), just plain confusion (see #774) and frustration (if you do it in a shell you can no longer use your shell because the container is now reading from your shell's stdin).

Also, you give the example of how output and input are broken by this patch, but the current state of `--detach` is **also** broken. So there is definitely something that we need to fix here.

Ultimately I would've **loved** if we'd made `--preserve-fds` generic enough that you could make it so that people who wanted to detach a container but still log to somewhere would be able to specify that `2` should be <something>. Because recently we've just been adding more and more hacks on top of the I/O handling and I'm starting to get sick of it.

cyphar:
Also the reason that `runc checkpoint` doesn't work with `--detach` is because we pass in file descriptors from the host. If that file descriptor is a pty or a pipe then generally `criu` can't record and recreate it later.

So "using it from the terminal" is not the only problem with the current setup. Not to mention that I generally think that having `terminal` be optional is a bit ... questionable (it causes an explosion of different setups that all need to be handled in different ways).

crosbymichael:
@cyphar what hacks?  How is `--detach` broken?  What is the reason you are thinking running `detach` in a shell is such an important usecase that it warrants breaking the I/O of the feature?

crosbymichael:
Lets discuss this on the OCI call this week.  If you are unable to make it then we can find a time to get some other maintainers and myself on a call this week that works well for your timezone.  I'll stay up late and make sure we are on the same page with it comes to the detaching of runc.  

Overall, breaking a feature because of a bug is ok, we all make mistakes.  Breaking a feature because of a usecase that you is unaware of is ok, its hard for any one person to know all the usecases of a low level tool.  However, breaking a feature intentionally because you don't like it or for whatever reason is irresponsible and unacceptable.

cyphar:
@crosbymichael 

> what hacks?

We now have `LISTEN_FDS`, `--preserve-fds`, the systemd-notify changes, `--console-socket`, the proxying method of `runc run`, and implicitly passing down file descriptors to the container process.

While all of these features are necessary on their own, together it smells of constantly applying more and more ad-hoc fixes to the I/O system. I'd hoped that after #1018 we would be able to start reducing the amount of individual features we have, and rather work on implementing them in a more generic way. I mean, every single time I have to open `utils_linux.go` or `tty.go` I sigh because it's going to be a long day of trying to understand how everything interacts -- that's not a sign of a well-designed setup.

As I mentioned in #1320, if we had something like `--pass-fd=NUM` then we could start reducing down these features into a single feature that is generic enough to do what you want.

> How is --detach broken? What is the reason you are thinking running detach in a shell is such an important usecase that it warrants breaking the I/O of the feature?

`--detach` in a shell (or `create`) not working as expected in a shell is not the only problem. It's symptomatic of a larger issue, which is that the current way we handle `--detach` (where we pass host file descriptors into the container process's control) has some issues that cause these sorts of breakages. For example, depending on how your *host terminal emulator* is set up, `runc checkpoint` may or may not work with `--detach`. As I mentioned above, there's also a lot of confusing things that start happening when you consider #774.

> However, breaking a feature intentionally because you don't like it or for whatever reason is irresponsible and unacceptable.

Sorry, when I said "intentional" I didn't fully understand what you meant when said that it was broken (in fact we use this feature in `cri-o` as well, so I was clearly operating in split-brain mode). I didn't mean to say that "breaking features is okay". Sorry for the misunderstanding.

crosbymichael:
@cyphar Ok, thank you for clarifying.  

I know what you mean and wasn't I the one staying we should inherit the console fd and not have a flag to the path? ;) 

I think there is a clear split between the detach usecase and using it in the shell.  In the shell, of course its going to be award because who launches services with `&` or daemonizing in their shell?  Its non-sense.  We only run into it because we are developers on runc and its quick to test things in your shell.  This is not what it is designed for or how it should be used.  Detaching is meant for use by things like docker, containerd, cri-o and its effortless for them to add whatever they want to the process for the stdio. It should be so effortless that I shouldn't even have to explain.  So when you take the shell awkwardness out of it and admit that its not a usecase for detach or create/start then it makes the feature much more usable and robust when looking at it.

As for checkpoint and restore on create/start, if it not working for you, you have a bug in your code.  containerd supports this today and all containers are launched detached.  We make sure to always create pipes and hand that to the container to support c/r.  From our experience, even if you want to log to a file, a pipe helps abstract host specific paths during c/r and the shim can redirect as needed.

----

I read over the latest comments in the rootless container pr.  You two are just hitting issues during development and testing and it is not real world use cases.  Unless garden is launching a bash shell and then starting containers I don't see how they are having issues.  I'm guess they are misunderstanding something and permissions on pipes/whatever they are providing the container are not correct.

Like @mrunalp suggestion above we could could do that for a nicer API for shell users but when calling runc pragmatically this is a non-issue.

As far as rethinking our IO options, there are somethings we cannot get rid of.  If we want to support systemd socket activation we have to support LISTEN_FDS, etc and we cannot overload that for other people's needs as it will cause more confusion.  

I agree its not ideal but keeping stdio being inherited is very simple and how the majority of unix programs work.  It allows things like this to be effortless and simple:

```systemd
[Service]
Type=forking
ExecStart=/usr/local/sbin/runc run -d --pid-file /run/containers/%i.pid %i
PIDFile=/run/containers/%i.pid
Delegate=yes
Restart=always
```

So overall lets come up with some proposals to unify the IO options and see what works without breaking existing usecases.  I'm fine with making changes to calling code pre 1.0 if we have a solid solution.  Lets just not focus on shell + detach as a usecase because it does not make sense at all and its as bad as all those tutorials telling you to run mysql in `screen` so that it stays running.

williammartin:
Ok so clearly, we shouldn't break existing behaviour just to make our (Garden's) current usage a little simpler. We were only validating against the problem we were bumping up against (attached `runc create`, no `runc start`) and may have hit problems further down the road anyway.

I'm not sure I agree that development, testing and operability are not "real world use cases". Through the very fact we banged our heads against a lack of simplicity here seems convincing enough to me that it's worth doing something here but obviously not at the expense of some other functionality.

Also, if I understood you correctly @mrunalp and @crosbymichael, you were both suggesting something like what we PRed already in https://github.com/opencontainers/runc/pull/1261  ?

williammartin:
@cyphar Absolutely. I will try to join the next call. I'm not suggesting to take that PR as-is, only clarifying that the suggestions seemed to be on the same line as what we already wanted.

crosbymichael:
What i don't understand is how hard is this for you to fix? 

```go
f, _ := os.Open("/dev/null")
cmd := exec.Command("runc", "create",...)
cmd.Stdin = f
cmd.Stdout =f
cmd.Stderr = f
cmd.Run()
```

That is all you have to do in your calling code to get the exact same functionality of this PR.  Whats the issue?

ijc25:
I did consider whether `--preserve-fds` could replace the `$LISTEN_FDS` special casing by having the caller include `LISTEN_FDS=N` in the `config.json`'s `Env` and passing `--preserve-fds=N` but when I looked it was the `$LISTEN_FDS` from the `runc` processes environment and not the `config.json` one which was the trigger for the special handling (which I suppose is to support the systemd service file model which Michael [mentions above](https://github.com/opencontainers/runc/pull/1349#issuecomment-282898553)) at which point it didn't seem like that idea would pan out.

I also didn't think having `--preserve-fds` default to `0` (i.e. not preserving stdio and requiring explicit opt in, whether by specifying N in the range 0..N or specifying particular fds as was suggest in #1320 is a bit moot) made much sense from a principal of least surprise PoV (Unix processes can typically expect something to be on those fds) and as a pretty radical change from the existing behaviour. Although I'm not sure if that was what was being suggested here or not.

williammartin:
@crosbymichael Sure, that would achieve the same affect against current runc master. However, our discussion was on the rootless PR and I reiterated above that we were working against Aleksa's rootless branch. Unless I'm misunderstanding ownership of `/dev/null` your snippet is not going to make any difference in the rootless world when `runc` attempts to `fchown` in `dupStdio`. In fact, not setting stdio on a golang `cmd` actually results in everything going to `/dev/null` anyway, which is exactly why we hit the issue in the first place.

Again, I'm not advocating for this PR if it breaks other behaviour and I can't speak to its merit outside of our specific rootless usage, only following up on your confusion about why Garden was hitting issues.

crosbymichael:
@williammartin ok, that makes much more sense.  The fchown check just needs to get smarter or done above this level.

From here, I think we can close this PR and continue the discussion on a proposal for IO handling or in this but it we should not merge a change like this to fix a fchown error. I'll work on fixing the fchown thing today.

crosbymichael:
@williammartin we are not going to break `exec -d` without `--tty`. It is a fairly simple fix to handling not fchowning `/dev/null`

williammartin:
@crosbymichael Thanks for taking the time on this. In general I'm not sure why the `fchown` happens at all since it feels like the caller should maybe be handling ownership of stdio but I'm sure there's a good reason that I just don't have the context to understand?

crosbymichael:
@williammartin ya, its only there for the non-detached case.  When runc creates pipe pairs for the container, it fchowns them for use in user namespaces.  It seems like this code is running during detach when the STDIO is already provided and should be owned correctly.

crosbymichael:
@williammartin i opened #1354 and verified the use cases for you, rootless, and how its currently working today.

Issue #1347, opened 2017-02-24
Unable to update kmem limit only


qiancai:
Due to commit 11222ee1f18f2ca2847a126dbda9cf057dd85f78 , it is not possible update item limit if not specified a default value in the config in the first place.

`# runc update --kernel-memory 50331648 root`
failed to set memory.kmem.limit_in_bytes, because either tasks have already joined this cgroup or it has children

However, it is fine to update other memory limits even if did not specify default values in the config.json.

`# runc update --memory-reservation 50331648 root`
`# runc update --memory 50331648 root`
`# runc update --memory-swap 50331648 root`

@crosbymichael PTAL

hqhq:
@qiancai Can you confirm if https://github.com/opencontainers/runc/pull/1350 is gonna fix this?

qiancai:
Yes, it fixed the problem.

PR #1350, opened 2017-02-25
Fix kmem accouting when use with cgroupsPath

hqhq:
Fixes: #1347
Fixes: #1083

The root cause of #1083 is because we're joining an
existed cgroup whose kmem accouting is not initialized,
and it has child cgroup or tasks in it.

Fix it by checking if the cgroup is first time created,
and we should enable kmem accouting if the cgroup is
craeted by libcontainer with or without kmem limit
configed. Otherwise we'll get issue like #1347

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -33,14 +33,18 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 	path, err := d.path("memory")
 	if err != nil && !cgroups.IsNotFound(err) {
 		return err
+	} else if path == "" {
+		return nil
 	}
 	if memoryAssigned(d.config) {
-		if path != "" {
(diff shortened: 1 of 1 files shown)


hqhq:
ping @crosbymichael @mrunalp

cyphar:
Hmm, this looks okay as a first step to fix that issue, but I'm a bit worried that this will break `runc update` (unless I'm misreading how the update code works).

cyphar:
LGTM. I thought this would break the `runc update` code, but I was wrong.

dqminh:
LGTM

dqminh:
I assume the errors from rh-atomic-bot is because it was misconfigured ? Doesnt look like it does anything.

cyphar:
I'm fairly sure that @rh-atomic-bot is broken because they didn't correctly script the Dockerfile we use. I would ignore it until it starts passing on PRs that we know work.

PR #1353, opened 2017-03-01
Container can be in stopped status from created status.

sak0:
container can be in stopped status from created status. even if it was't been run before.


hqhq:
LGTM

cyphar:
LGTM

PR #1354, opened 2017-03-01
Don't fchown when inheriting io

crosbymichael:
This is a fix for rootless containers and general io handling.  The
higher level systems must preparte the IO for the container in the
detach case and make sure it is setup correctly for the container's
process.

Replacement for #1349

Fixes detached use case for #774 

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"sync"
-	"syscall"
 
 	"github.com/docker/docker/pkg/term"
 	"github.com/opencontainers/runc/libcontainer"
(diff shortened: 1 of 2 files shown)


tty.go
```diff
@@ -62,19 +61,10 @@ func createStdioPipes(p *libcontainer.Process, rootuid, rootgid int) (*tty, erro
 	return t, nil
 }
 
-func dupStdio(process *libcontainer.Process, rootuid, rootgid int) error {
+func inheritStdio(process *libcontainer.Process, rootuid, rootgid int) error {
```
cyphar: Can we also drop the `rootuid` and `rootuid` arguments?

williammartin:
Functionality works for us here in Garden-land against master and the rootless branch. Thanks @crosbymichael

crosbymichael:
@cyphar updated

cyphar:
After trying to run this under #774, I found that there's still two places where we're doing `Fchown` of stdio fds:

```
% git grep Fchown
libcontainer/init_linux.go:             if err := syscall.Fchown(int(fd), u.Uid, int(s.Gid)); err != nil {
libcontainer/process_linux.go:          if err := syscall.Fchown(int(fd), rootuid, rootgid); err != nil {
```

So if you create a root terminal `sudo -E termite` or `sudo -E gnome-terminal` and then do `sudo -u cyphar ./runc --root /tmp/runc create ctr` it will fail with `EINVAL` because it ends up trying to `fchown` the stdio to an unmapped user.

If you want I can fix this in #774 or we can figure out whether it's safe to delete that code too.

crosbymichael:
@cyphar i can look at that independently also

cyphar:
Reading over `fixStdioPermissions` it looks like the code isn't _wrong_ but we should modify it so that it handles user namespaces properly (namely handle unmapped UIDs and GIDs).

hqhq:
LGTM

dqminh:
LGTM

PR #1356, opened 2017-03-02
Add separate console socket

### LINKED ISSUE #1302: runc exec -d with --tty too fast
With the new tty functionality if you are using `--console-socket` on runc exec, because exec has no create/start there is no way to get the console before a command like `true` exits or `echo hi` and get the output because the command will exit before runc returns the console master to the caller.  

We need the sending of the console to block on the caller reading the master instead of continuing.  I think its is because runc uses two sockets, one from the child to parent, and one for runc to its caller.  

@cyphar can we just use one socket for this?

crosbymichael:
Fixes #1302

### DIFF
--- a/libcontainer/console.go
+++ b/libcontainer/console.go
@@ -1,8 +1,6 @@
 package libcontainer
 
 import (
-	"encoding/json"
-	"fmt"
 	"io"
 	"os"
 )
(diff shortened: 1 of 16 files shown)


crosbymichael:
Ok, so currently I'm getting an error on start so I wanted to open this up to get a few eyes on it.  For some reason, maybe nsenter is eating the console socket or the initpipe is getting messed up.  I cannot see it right now but if anyone else has time to take a look please do.  I'll keep working on it but it could be something obvious that I'm missing.

crosbymichael:
@cyphar if you are checking out prs can you take a look at this one? I still have to fix a weird mount test pr but overall it looks good

cyphar:
Yup, will do.

libcontainer/console.go
```diff
@@ -17,57 +15,3 @@ type Console interface {
...
-	TerminalInfoType    uint8  = 'T'
-)
-
-// TerminalInfo is the structure which is passed as the non-ancillary data
-// in the sendmsg(2) call when runc is run with --console-socket. It
-// contains some information about the container which the console master fd
-// relates to (to allow for consumers to use a single unix socket to handle
-// multiple containers). This structure will probably move to runtime-spec
-// at some point. But for now it lies in libcontainer.
-type TerminalInfo struct {
```
cyphar: The reason for this was so that you could use a single `console-socket` and listen for multiple instances of `runc` passing file descriptors. Without this, you can't tell which container a particular `pty` comes from. This was one of the concerns raised about the scalability of `console-socket` -- how do you handle starting hundreds of containers simultaneously if you keep hammering `bind` and `socket`?

crosbymichael: Do we know for a fact that it is a slow point in the execution or just a premature optimization?

crosbymichael:
@mrunalp @cyphar we need this merged before we can think about integrating in docker because exec is broken without it.

mrunalp:
I'll review this by tomorrow.

libcontainer/console.go
cyphar: To be honest I didn't benchmark it, but it seemed like a useful optimisation because it allowed us to also pass around information about the container to the caller (rather than just a file descriptor). But if you feel that's not necessary, we can implement it in `cri-o` with a new socket for every `conmon` and presumably you can do the same thing in `containerd-shim`.

mrunalp:
LGTM

hqhq:
Looks much cleaner.

LGTM

libcontainer/console.go
hqhq: I think we can create a new socket for every container for now, at least before we figure out the double socket issue.

PR #1358, opened 2017-03-06
run: implement --delete-if-stopped

giuseppe:
"run -d container" and "run container" behave in different ways in
what they leave after the container exits, "run container" cleans up
the root directory while "run -d" requires an additional delete.

I'd like to have the possibility to run a container without caring if
there is already an existing one (stopped) with the same id.

In addition, the new option will simplify systemd services using runC
as it can be possible to have only one directive:

ExecStart=/usr/local/sbin/runc run --delete-if-stopped -d --pid-file /run/minecraft.pid minecraft-build-container

instead of requiring an additional ExecStopPost directive.

### DIFF
--- a/man/runc-run.8.md
+++ b/man/runc-run.8.md
@@ -26,3 +26,4 @@ command(s) that get executed on start, edit the args parameter of the spec. See
    --no-subreaper            disable the use of the subreaper used to reap reparented processes
    --no-pivot                do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk
    --no-new-keyring          do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key
+   --delete-if-stopped       if a stopped container with the same id exists, delete it
--- a/run.go
+++ b/run.go
@@ -61,6 +61,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "preserve-fds",
 			Usage: "Pass N additional file descriptors to the container (stdio + $LISTEN_FDS + N in total)",
 		},
+		cli.BoolFlag{
+			Name:  "delete-if-stopped",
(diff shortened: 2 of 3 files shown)


crosbymichael:
REJECT

I don't like opening up the scope of runc to these one off flags.  They are nasty to support in the long run and there are many ways that you can solve this problem in the calling code.  I don't think its runcs place to have logic like this and you can imagine the additional flags that people will want added to force runc to work perfect with their workflow.

I don't think an additional StopPost is horrible at all and I use it in my unit files.

```systemd

[Unit]
Description=Some container %i
After=network.target

[Service]
Type=forking
ExecStart=/usr/local/sbin/runc  run -d --pid-file /run/%i.pid %i
ExecStopPost=/usr/local/sbin/runc delete %i
WorkingDirectory=/var/lib/containers/bundles/%i
PIDFile=/run/%i.pid
Delegate=yes
Restart=always

[Install]
WantedBy=multi-user.target

```

giuseppe:
Got it.  I added the systemd example more of a side effect, while what I am really trying to accomplish is the possibility to run a runc container and not worry if runc didn't properly clean up the previous run.  In few cases I've observed that the `/run/runc/$ID` wasn't properly deleted and the service couldn't not restart.  Having an option like the proposed one will solve similar issues as runc will try to cleanup what was left on startup.  The alternative would be an `ExecStartPre=/bin/runc delete $NAME` so that the service won't get stuck if `/run/runc/$ID` is still present.

cyphar:
Reject, for the same reason that @crosbymichael has outlined above. The create-start-stop-delete flow is defined by the OCI spec and I don't really like adding flags that ignore parts of it. If you want you can do `runc delete -f ctr && runc run -d ctr` to override the container.

crosbymichael:
if its a bug that is causing it to not clean up properly then we can fix that.

Issue #897, opened 2016-06-08
selinux: switch to opencontainers/go-selinux


lucab:
While touching rkt code, I realized there is an embedded copy of libcontainer selinux package. The two versions are now [diverging](https://github.com/coreos/rkt/blob/master/pkg/selinux/selinux.go), and I would like to reconcile/contribute back.

Looking at the code, it looks like most of it isn't container specific and could be nicely split into its own module. Would you be ok in splitting this out in a self-standing selinux module? I'd be happy to submit current rkt changes into that and getting rid of existing forks.

cyphar:
WDYT @rhatdan

rhatdan:
Fine with me.

lucab:
@cyphar any news on this? Can the new repo see the light?

I'm certainly in favour of this living under the opencontainers umbrella, I just have some minor wish items:
 * having a clear set of MAINTAINERS to ping (as I guess not all runc people would have interests in that)
 * having tagged releases to help downstream consumers

cyphar:
I'm still waiting for a response from @caniszczyk and @crosbymichael.

caniszczyk:
The @opencontainers/tob have the authorization to approve new projects, they would need to sign off it and we can move forward. @philips do you see anything wrong with this? This is essentially a utility package

jonboulle:
This seems to be live (https://github.com/opencontainers/go-selinux/), except it hasn't been integrated back into this repository yet - is there a plan for that?

lucab:
The first half of this has happened after https://github.com/opencontainers/tob/pull/29, thus I've re-titled it for the remaining part.

crosbymichael:
We just need to do the work not to re-integrate this back into runc.  If someone has time it should be a very simple PR with the code vendored.

vbatts:
I'm sure a PR would be welcome ;-)

crosbymichael:
@vbatts naw

vbatts:
dang

hqhq:
@crosbymichael @vbatts I'll do it today.

rhatdan:
I would prefer the repository be renamed to selinux with a subdir of the golang bindings.
I have been requested to move container-selinux to this repo, but moving the policy to something named go-selinux, makes no sense.

PR #1365, opened 2017-03-09
Use opencontainers/selinux package

hqhq:
It's splitted as a separate project.

Fixes: #897 


hqhq:
We'll need https://github.com/opencontainers/go-selinux/pull/7 to be merged to fix CI failure.

rhatdan:
I would prefer the rename before this gets merged.

hqhq:
@rhatdan Sure, I'll rework this PR after rename and other fixes in opencontainers/selinux are merged.

hqhq:
Updated, all green now~

rhatdan:
I have requested one more change to the opencontainers/selinux package to put the go bindings into a golang subdir, so we can add a policy subdir and move container-selinux package into opencontainers/selinux package.

https://github.com/opencontainers/selinux/pull/10

crosbymichael:
@rhatdan is upstream good now?  I saw that issue is closed so are you happy with the naming of the project now so that we can update?

rhatdan:
@crosbymichael we have renamed a bunch of function calls.  Lets get these merged and then we will need to change the calls in libcontainer as well as vendor it in.

rhatdan:
Upstream has been merged, we can start to convert runc to use the new interfaces.

hqhq:
@rhatdan Thanks, I'll rework later today.

hqhq:
Updated.

libcontainer/configs/validate/validator.go
```diff
@@ -7,7 +7,7 @@ import (
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
-	"github.com/opencontainers/runc/libcontainer/selinux"
+	selinux "github.com/opencontainers/selinux/go-selinux"
```
rhatdan: @runcom says there is no need to use 

`selinux` "github.com/opencontainers/selinux/go-selinux"

Here, it will work fine without it.

rhatdan: Other then that this LGTM

runcom: yup, there shouldn't be the need of aliasing if `go-selinux` exports `selinux` package afaict.

hqhq: I think it's a explicit or implicit issue, and I'd prefer it to be explicit, we can wait for other maintainers' opinions :)

crosbymichael: go-imports also makes it explicit, its fine both ways and is not an issue at all

crosbymichael:
LGTM

hqhq:
ping @cyphar @mrunalp @dqminh

crosbymichael:

dqminh:
LGTM

PR #1366, opened 2017-03-09
Remove unused ExecFifoPath

hqhq:
In container process's Init function, we use
fd + execFifoFilename to open exec fifo, so this
field in init config is never used.


crosbymichael:
LGTM

dqminh:
LGTM

Issue #1367, opened 2017-03-09
Unable to create containers when cgroup subsystems are mounted in multiple places


teddyking:
We recently ran into an issue whereby runc was not able to create containers and was failing with the following error:

```
EOF
container_linux.go:259: starting container process caused "process_linux.go:283: applying cgroup configuration for process caused \"open /sys/fs/cpuset.cpus: no such file or directory\""
```

This was happening due to an additional cpu,memory cgroup subsystem mounted somewhere on the system (at `/cgroups` in this case):

```
# cat /proc/self/mountinfo | grep cgroup
24 18 0:19 / /sys/fs/cgroup rw,relatime - tmpfs none rw
32 23 0:25 / /cgroups rw,relatime - cgroup cgroup rw,cpu,memory
33 24 0:26 / /sys/fs/cgroup/devices rw,relatime - cgroup cgroup rw,devices
34 24 0:27 / /sys/fs/cgroup/cpuset rw,relatime - cgroup cgroup rw,cpuset
35 24 0:25 / /sys/fs/cgroup/cpu rw,relatime - cgroup cgroup rw,cpu,memory
36 24 0:25 / /sys/fs/cgroup/memory rw,relatime - cgroup cgroup rw,cpu,memory
```

It seems that this is only a problem when two mounts of the same subsystem are performed in a certain order (i.e in the example above, the `/cgroups` mount appears first in `/proc/self/mountinfo`).

Is runc able/supposed to support this sort of setup?

Cheers,
Ed & @craigfurman

cyphar:
Yeah, this is related to #798 and several other issues we've had in the past. IMO we need to rework how mountpoint checking is done (the code is basically unreadable and is [in my mind] fragile). In particular this code also had similar issues with mounting different cgroup versions (which has now been fixed but I don't like the fix :wink:).

In my mind, the perfect way for this to be handled (so that it can also accomodate #774) is this:

* Have a list of "implementations" of `cgroup` (or rather `resource`) handlers [which includes both cgroupv2 and cgroupv1].
* Go through the list, trying to see if they can set the limit for `$resource`. `continue` if any of them succeeds.
* If none of them could set the limit, check whether the limit is the "default" and output an error if it isn't.

The `Set` and `Apply` split probably won't work with the above idea, but I'm unsure whether exposing that cgroup interface in our Go interface makes much sense. Then there's the whole `kmemcg` fiasco.

[...]

PR #1372, opened 2017-03-15
Handle container creation when cgroups have already been mounted in another location

craigfurman:
As described in https://github.com/opencontainers/runc/issues/1367, if a non-cpuset subsystem has been been mounted earlier in the mount table at a disjoint location to the rest of the hierarchy (e.g. `/sys/fs/cgroup/{cpuset,cpu,memory...}`) then `runc create` will fail with:

```
container_linux.go:259: starting container process caused "process_linux.go:283: applying cgroup configuration for process caused \"open /sys/fs/cpuset.cpus: no such file or directory\""
```

To replicate this behaviour, using Ubuntu 14.04 with a 4.4 kernel (linux-generic-lts-xenial package), create a mount table that looks like this:

```
cgroup /cgroup cgroup rw,relatime,cpu 0 0
none /sys/fs/cgroup tmpfs rw,relatime 0 0
cgroup /sys/fs/cgroup/cpuset cgroup rw,relatime,cpuset 0 0
cgroup /sys/fs/cgroup/cpu cgroup rw,relatime,cpu 0 0
cgroup /sys/fs/cgroup/cpuacct cgroup rw,relatime,cpuacct 0 0
cgroup /sys/fs/cgroup/blkio cgroup rw,relatime,blkio 0 0
cgroup /sys/fs/cgroup/memory cgroup rw,relatime,memory 0 0
cgroup /sys/fs/cgroup/devices cgroup rw,relatime,devices 0 0
cgroup /sys/fs/cgroup/freezer cgroup rw,relatime,freezer 0 0
cgroup /sys/fs/cgroup/net_cls cgroup rw,relatime,net_cls 0 0
cgroup /sys/fs/cgroup/perf_event cgroup rw,relatime,perf_event 0 0
cgroup /sys/fs/cgroup/net_prio cgroup rw,relatime,net_prio 0 0
cgroup /sys/fs/cgroup/hugetlb cgroup rw,relatime,hugetlb 0 0
cgroup /sys/fs/cgroup/pids cgroup rw,relatime,pids 0 0
```

Note that /cgroup can't have subsystem cpuset, otherwise everything will be fine.

On runc master:

```
→ runc create test
container_linux.go:259: starting container process caused "process_linux.go:283: applying cgroup configuration for process caused \"open /sys/fs/cpuset.cpus: no such file or directory\""
```

Using this PR, creation completes successfully. It succeeds because we no longer use the "cgroup root" (the dir that is assumed to contain all cgroup subsystem mountpoints) to work out the highest cpuset subsystem dir. Instead we look for the closest mountpoint ancestor of the cgroup dir we are operating on.

The following shell functions may come in handy for setting up a mount table that looks like this:

```
function cmount() { mount -t cgroup -o cpu cgroup /cgroup && mount -t tmpfs none /sys/fs/cgroup && cgroups-mount; }

function cumount { cgroups-umount && umount /sys/fs/cgroup && umount /cgroup; }
```

We weren't sure how to integration test this, as providing a mount table like above seems to break the integration tests in docker, precisely due to this bug but if anyone has any ideas we are happy to update.

We think it would be ideal if runc took a "cgroup root" as configuration (it could be optional), but didn't want to change too much in this PR.

Cheers,
@williammartin and Craig

### DIFF
--- a/libcontainer/cgroups/fs/cpuset.go
+++ b/libcontainer/cgroups/fs/cpuset.go
@@ -57,10 +57,11 @@ func (s *CpusetGroup) ApplyDir(dir string, cgroup *configs.Cgroup, pid int) erro
 	if dir == "" {
 		return nil
 	}
-	root, err := getCgroupRoot()
+	mountInfo, err := ioutil.ReadFile("/proc/self/mountinfo")
 	if err != nil {
 		return err
 	}
+	root := filepath.Dir(cgroups.GetClosestMountpointAncestor(dir, string(mountInfo)))
 	// 'ensureParent' start with parent because we don't want to
 	// explicitly inherit from parent, it could conflict with
 	// 'cpuset.cpu_exclusive'.
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -66,6 +66,21 @@ func isSubsystemAvailable(subsystem string) bool {
 	return avail
 }
 
+func GetClosestMountpointAncestor(dir, mountinfo string) string {
+	deepestMountPoint := ""
+	for _, mountInfoEntry := range strings.Split(mountinfo, "\n") {
+		mountInfoParts := strings.Fields(mountInfoEntry)
+		if len(mountInfoParts) < 5 {
+			continue
+		}
+		mountPoint := mountInfoParts[4]
+		if strings.HasPrefix(mountPoint, deepestMountPoint) && strings.HasPrefix(dir, mountPoint) {
+			deepestMountPoint = mountPoint
+		}
+	}
+	return deepestMountPoint
+}
+
 func FindCgroupMountpointDir() (string, error) {
 	f, err := os.Open("/proc/self/mountinfo")
 	if err != nil {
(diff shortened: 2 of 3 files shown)


williammartin:
That failed Travis test looks to be a flake since I've seen on other Travis builds. I tried the `!rebuild runc#1372` but didn't look like I was allowed.

cyphar:
Wait, is the only cgroup that we have problems with `cpuset`? That doesn't sound right.

craigfurman:
@cyphar it's because [cpuset subsystem impl gets the cgroup root again](https://github.com/opencontainers/runc/blob/d27094036343c79ca29023ad530fa5be3d9b06c2/libcontainer/cgroups/fs/cpuset.go#L60)

craigfurman:
Hi @cyphar , I know you're busy, but did you get a chance to check out the explanation above?

cyphar:
LGTM. Sorry for the delay.

williammartin:
No problem. Hmm, need one more LGTM on this, any idea who would be good to review?

crosbymichael:
LGTM

PR #1373, opened 2017-03-16
fix minor issue

moypray:
When failed to attach veth pair, should remove the veth device

### DIFF
--- a/libcontainer/network_linux.go
+++ b/libcontainer/network_linux.go
@@ -171,7 +171,7 @@ func (v *veth) create(n *network, nspid int) (err error) {
 			netlink.LinkDel(veth)
 		}
 	}()
-	if err := v.attach(&n.Network); err != nil {
+	if err = v.attach(&n.Network); err != nil {
 		return err
 	}
 	child, err := netlink.LinkByName(n.TempVethPeerName)


crosbymichael:
LGTM

mrunalp:
LGTM

libcontainer/network_linux.go
```diff
@@ -171,7 +171,7 @@ func (v *veth) create(n *network, nspid int) (err error) {
 			netlink.LinkDel(veth)
 		}
 	}()
-	if err := v.attach(&n.Network); err != nil {
+	if err = v.attach(&n.Network); err != nil {
```
cyphar: I'm confused, what did this actually fix? Is there a reason we are setting `err` in the parent scope to `nil` in this round-about way?

hqhq: I don't think it fix anything.

moypray: Ahh... I thought the `err` variable was the one returned, after checking the Golang grammar I think I was wrong. Actually the origin code is correct. This change fixed nothing. Sorry to misunderstand.

cyphar:
Reverted in #1374. This change just makes the code more complicated because the `=` looks intentional.

PR #1381, opened 2017-03-21
 Add criupath testcase and refactor code

chchliang:
Signed-off-by: chchliang <chen.chuanliang@zte.com.cn>

### DIFF
--- a/libcontainer/factory_linux_test.go
+++ b/libcontainer/factory_linux_test.go
@@ -3,6 +3,7 @@
 package libcontainer
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


libcontainer/factory_linux_test.go
```diff
@@ -47,6 +47,10 @@ func TestFactoryNew(t *testing.T) {
 	if factory.Type() != "libcontainer" {
 		t.Fatalf("unexpected factory type: %q, expected %q", factory.Type(), "libcontainer")
 	}
+
+	if lfactory.CriuPath != "criu" {
+		t.Fatalf("unexpected factory CriuPath: %q, expected %q", lfactory.CriuPath, "criu")
```
cyphar: What is the reason for this change?

chchliang: I think `defer syscall.Unmount(root, syscall.MNT_DETACH)` at last line of this function is wrong,by the way add `lfactory.CriuPath` check, like `factory.Type() != "libcontainer"`

chchliang:
@cyphar 
 I have refactored this code, `func checkFactory(factory Factory, root string) (*LinuxFactory, error) `

PR #1386, opened 2017-03-22
Error handling for OOM score adjust in LXD

CarltonSemple:
The OOM score for processes is not allowed to be changed by unprivileged system containers, so the function in `libcontainer/init_linux.go` has been modified to move on if there's a permission denied error.

For more detail, see https://github.com/lxc/lxd/issues/2994

Signed-off-by: Carlton-Semple <carlton.semple@ibm.com>

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -390,7 +390,16 @@ func setupRlimits(limits []configs.Rlimit, pid int) error {
 func setOomScoreAdj(oomScoreAdj int, pid int) error {
 	path := fmt.Sprintf("/proc/%d/oom_score_adj", pid)
 
-	return ioutil.WriteFile(path, []byte(strconv.Itoa(oomScoreAdj)), 0600)
+	// If there's a permission error, continue. In unprivileged LXD containers, this isn't allowed.
+	// See https://github.com/lxc/lxd/issues/2994 .
+	if err := ioutil.WriteFile(path, []byte(strconv.Itoa(oomScoreAdj)), 0600); err != nil {
+		if os.IsPermission(err) {
+			logrus.Warn(err)
+		} else {
+			return err
+		}
+	}
+	return nil
 }
 
 const _P_PID = 1


libcontainer/init_linux.go
```diff
@@ -390,7 +390,16 @@ func setupRlimits(limits []configs.Rlimit, pid int) error {
 func setOomScoreAdj(oomScoreAdj int, pid int) error {
 	path := fmt.Sprintf("/proc/%d/oom_score_adj", pid)
 
-	return ioutil.WriteFile(path, []byte(strconv.Itoa(oomScoreAdj)), 0600)
+	// If there's a permission error, continue. In unprivileged LXD containers, this isn't allowed.
+	// See https://github.com/lxc/lxd/issues/2994 .
+	if err := ioutil.WriteFile(path, []byte(strconv.Itoa(oomScoreAdj)), 0600); err != nil {
+		if os.IsPermission(err) {
```
crosbymichael: I think we want to check if we are running in a userns or not for error handling.

If we are in a userns, then don't even warn.
If we are not, then return the error regardless of an eperm

cyphar:
A more complete version of this code is already included in #774. I'm confused, both @brauner and @mwhudson know of that patch. @brauner asked me to split out https://github.com/opencontainers/runc/pull/774/commits/f31022e5aadac64e7c44b1dffb1e53fc8a0f2730 specifically for Canonical's backporting purposes.

cyphar:
Unless you're saying that LXD has some additional `oom_score_adj` restrictions?

cyphar:
Ah, I must've misread the linked issue. This is orthogonal to my patches that make it possible to set `oom_score_adj` in a rootless container, but will need to be re-done with #774. However, we can't just ignore errors -- my reading of the runtime specification considers that to be a violation (if one of the options specified in `config.json` cannot be applied we can't just ignore that error).

crosbymichael:
@cyphar i guess the alternative is that we fail, the user is required to remove the oom_score_adj from their spec and then they start again?

brauner:
@cyphar, @crosbymichael, I can see your point about failing when a non-zero oom_score_adj value has been requested in other cases it should be possible to move on. I don't know enough about your specs though.

crosbymichael:
@CarltonSemple what do you think of the discussion above as far as alternate implementations for this?

CarltonSemple:
@crosbymichael so I actually just realized Docker has a similar fix, and they use the `system.RunningInUserNS()` function from `runc/libcontainer/system` to determine whether there should be a warning.

https://github.com/docker/docker/blob/17.04.x/libcontainerd/oom_linux.go#L25

cyphar:
I believe this problem has been fixed as part of #774. Effectively, we now set `oom_score_adj` inside `nsexec.c` (through `/proc/self`) before we set ourselves to be non-dumpable. Feel-free to re-open it if it's still applicable.

AkihiroSuda:
needs rebase

AkihiroSuda:
Adding `system.RunningInUserNS()` check sounds good

AkihiroSuda:
@CarltonSemple  Are you still interested in this?

kolyshkin:
I think we can close this one.

PR #1387, opened 2017-03-22
Don't try to read freezer.state from the current directory

avagin:
If we try to pause a container on the system without freezer cgroups,
we can found that runc tries to open ./freezer.state. It is obviously wrong.

$ ./runc pause test
no such directory for freezer.state

$ echo FROZEN > freezer.state
$ ./runc pause test
container not running or created: paused

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1284,7 +1284,12 @@ func (c *linuxContainer) runType() (Status, error) {
 }
 
 func (c *linuxContainer) isPaused() (bool, error) {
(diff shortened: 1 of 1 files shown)


dqminh:
So the end result is the same ( because if freezer cgroup is not mounted, then we go into `os.IsNotExist(err)` and we return `false, nil` anyway ? This just saves an open call right ?

avagin:
@dqminh The difference is in a case when a current directory contains a freezer.state file. Without this changes, you can write anything in this file and runc will use it.

dqminh:
LGTM

hqhq:
LGTM

PR #1388, opened 2017-03-23
travis: set go_import_path to github.com/opencontainers/runc

avagin:
If someone forks runc and wants to check changes in travis, he will find
that the command 'make BUILDTAGS="${BUILDTAGS}"' fails, because
github.com/opencontainers/runc/ is used in many places to import
packages (e.g. libcontainer).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ matrix:
   allow_failures:
     - go: tip
 
+go_import_path: github.com/opencontainers/runc
+
 # `make ci` uses Docker.
 sudo: required
 services:


avagin:
@vbatts git-validation doesn't work if I pushed my changes into the my runc repo. Can we do something with this?
https://travis-ci.org/avagin/runc/jobs/214068100
```
0.01s$ git-validation -run DCO,short-subject -v
ERRO[0000] mm[git] cmd: "git --no-pager log --pretty=format:%H eec6c2a693ca..1e97db11aeac" 
2017/03/23 00:20:12 exit status 128
```

vbatts:
@avagin yes, i really want to see a reliable way to do these checks with travis. In the link you provided, the commit-range eec6c2a693ca..1e97db11aeac does not exist for your `travis` branch on https://github.com/avagin/runc.git 
I will look into the `TRAVIS_` env variables provided to see what might be more reliable than `TRAVIS_COMMIT_RANGE`

avagin:
@vbatts when I am preparing a patch series, I do git rebase -i origin/master and then git push --force avagin HEAD:branch. I think the --force option is a reason why TRAVIS_COMMIT_RANGE doesn't work in this case.

vbatts:
@avagin i'm currently at a loss for this use-case :grimacing:

avagin:
With this patch, anyone can push changes in his own repo and check that nothing failed in travis.
For example: https://travis-ci.org/avagin/runc/builds/241371998

avagin:
And here is an example how it looks like without this patch.
https://travis-ci.org/avagin/runc/jobs/241395776

avagin:

dqminh:
LGTM

crosbymichael:
LGTM

PR #1392, opened 2017-03-24
testcase code was refactored in `func TestValidate(t *testing.T)`

chchliang:
Signed-off-by: chchliang <chen.chuanliang@zte.com.cn>

### DIFF
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -9,14 +9,141 @@ import (
 )
 
 func TestValidate(t *testing.T) {
-	config := &configs.Config{
-		Rootfs: "/var",
+	for _, s := range []struct {
+		config   *configs.Config
+		testcase string
+		//expectError `true` return error is not nil,`false` is nil.
+		expectError bool
+	}{
+		{
+			config: &configs.Config{
+				Rootfs: "/var",
+			},
+			testcase:    "TestValidate",
+			expectError: false,
+		},
+		{
+			config: &configs.Config{
+				Rootfs:     "/var",
+				Namespaces: []configs.Namespace{},
+				Networks:   []*configs.Network{{Type: "loopback"}},
+			},
+			testcase:    "TestValidateNetworkWithoutNETNamespace",
+			expectError: true,
+		},
+		{
+			config: &configs.Config{
+				Rootfs:     "/var",
+				Namespaces: []configs.Namespace{},
+				Routes:     []*configs.Route{{Gateway: "255.255.255.0"}},
+			},
+			testcase:    "TestValidateNetworkRoutesWithoutNETNamespace",
+			expectError: true,
+		},
+		{
+			config: &configs.Config{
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

cyphar:
I don't really like this because it means that it's no longer obvious what test actually failed. Also, `expected` should be a boolean not a string.

chchliang:
@cyphar 
I don't think so. because every attributes have validated,if fail then return different error, add comments to each test mock.

image

hqhq:
What about expected error to occur but it was nil?

chchliang:
@hqhq 
Expected value reversal, i haved changed, thanks.

hqhq:
@chchliang You might've read me wrong, my question is following your answer to @cyphar 's question, if test failed because it expected error to occur but it was not, in your code, it's not obvious what test actually failed.

And neither `true` means return error or not are bad ideas, because `expected` variable is bad, your can use something like `expectError` to be explicit.

chchliang:
@hqhq 
Yes, variable expected means only return error or not, i will change the variable name,so  easy  to understand what it means.

cyphar:
@chchliang 

> I don't think so. because every attributes have validated,if fail then return different error, add comments to each test mock.

Okay, but the error doesn't necessarily tell you what test failed. For example, you could get multiple tests that produce the same error -- how am I (as a developer) meant to disambiguate them?

If you want  to refactor this code, make a common function that's shared between the different `TestXxx` functions so that `go test` remains sane (I'm NACK on this PR at the moment). Thanks.

cyphar:
Maybe something like this:

```
func testValidateHelper(t *testing.T, config *configs.Config, shouldFail bool) {
  // ...
}
```

Which means you could reduce most of the tests into a one-liner.

chchliang:
@cyphar 
I think that add attribute `testcase` to illustrate  function when test fail. WDYT?

image


image

cyphar:
@chchliang You could do that, but I think a helper would be much easier to read.

cyphar:
Another benefit of having separate functions is that you can specify which test to run with `go test` (`got test -run TestBlah`). Combining the tests makes you lose that benefit.

chchliang:
@cyphar 
Yes, each has its advantages and disadvantages, after refactored, developer only set test data.

cyphar:
While that might be true, it's only like 4 extra lines of boilerplate that a developer has to copy. I'll just be blunt -- I NACK this because it's not actually helpful as-is IMO. Maybe if you used sub-tests this would be more acceptable, but I still don't understand why "separate tests in separate functions, as per idiomatic Go" is a disadvantage.

chchliang:
@cyphar 
I have modified, refactored with sub-tests.

crosbymichael:
Lets keep the tests the same way.  There are advantages to having separate tests and as @cyphar pointed out its not that much more boilerplate to add them.

Thanks for the PR @chchliang

PR #1394, opened 2017-03-29
libcontainer: rewrite cmsg to use sys/unix

cyphar:
The original implementation is in C, which increases cognitive load and
possibly might cause us problems in the future. Since sys/unix is better
maintained than the syscall standard library switching makes more sense.


cyphar:
@crosbymichael This is at least the start of the "switch everything to `sys/unix`" transition. I'm really surprised how big `sys/unix` is (it's ~122k lines).

cyphar:
@crosbymichael This is at least the start of the "switch everything to `sys/unix`" transition. I'm really surprised how big `sys/unix` is (it's ~122k lines).

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1396, opened 2017-04-04
Set container state only once during start

harche:
Irrespective of 'isInit' is set to true or false, set container state during start only once. 


rhatdan:
LGTM

dqminh:
LGTM

mrunalp:
LGTM

PR #1399, opened 2017-04-06
restore: apply resource limits

avagin:
When C/R was implemented, it was enough to call manager.Set to apply
limits and to move a task. Now .Set() and .Apply() have to be called
separately.

Fixes: 8a740d5391a7 ("libcontainer: cgroups: don't Set in Apply")

Cc: @cyphar

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -949,6 +949,10 @@ func (c *linuxContainer) criuApplyCgroups(pid int, req *criurpc.CriuReq) error {
 		return err
 	}
 
+	if err := c.cgroupManager.Set(c.config); err != nil {
+		return newSystemError(err)
+	}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
IDK if you need it elsewhere or not for sure, but you're doing 
	return c.cgroupManager.Set(c.config)
Near line 187 in the Set function and I don't see a call to Apply elsewhere.  Given your PR description, do you need to add a corresponding apply() call for that Set call too?

avagin:
@TomSweeneyRedHat .Apply() adds processes into cgroups. container.Set() is called when a container is running, so all processes are in required cgroups already.

hqhq:
LGTM

mrunalp:
LGTM

PR #1401, opened 2017-04-11
validator.go: check the given net namespace is host net namespace

coolljt0725:
If the network namespace is not a symlink, it still could
be a host network namespace. For exampe, in docker, the
host network namespace path is `/var/run/docker/netns/default`,
and it's a mount point of host network namespace.

```
$ mount | grep default
nsfs on /run/docker/netns/default type nsfs (rw)
```
This will make the docker can set `sysctl` when use
host network namespace and this will cause security
issue if we update the runc of docker.

Signed-off-by: Lei Jitang <leijitang@huawei.com>

relate to https://github.com/opencontainers/runc/pull/1221
cc @sameo @mrunalp

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -4,10 +4,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/selinux"
+	"github.com/vishvananda/netns"
 )
 
 type Validator interface {
@@ -164,7 +166,10 @@ func isSymbolicLink(path string) (bool, error) {
 
 // checkHostNs checks whether network sysctl is used in host namespace.
 func checkHostNs(sysctlConfig string, path string) error {
-	var currentProcessNetns = "/proc/self/ns/net"
+	var (
+		currentProcessNetns = "/proc/self/ns/net"
+		destOfContainer     string
+	)
 	// readlink on the current processes network namespace
 	destOfCurrentProcess, err := os.Readlink(currentProcessNetns)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


dqminh:
@coolljt0725 can we write some test for this ?

libcontainer/configs/validate/validator.go
```diff
@@ -179,14 +184,35 @@ func checkHostNs(sysctlConfig string, path string) error {
...
+		}
+		defer origns.Close()
 
-	// readlink on the path provided in the struct
-	destOfContainer, err := os.Readlink(path)
-	if err != nil {
-		return fmt.Errorf("read soft link %q error", path)
+		containerns, err := netns.GetFromPath(path)
+		if err != nil {
+			return fmt.Errorf("faild to get netns from path %v:%v", path, err)
```
TomSweeneyRedHat: nit 'faild' to 'failed'

coolljt0725: fixed

coolljt0725:
@dqminh  test added

dqminh:
LGTM

cyphar:
I'm not a fan of joining the namespace in the validator, it seems like something that might break quite easily (especially because Go is multithreaded, and `LockOSThread` doesn't always do what you might think it does). Can't we use the `nsfs` `ioctl(2)`s in some way to achieve the same thing?

In addition, joining namespaces in the validator very strongly feels like it will break rootless containers (though setting those sysctls probably would fail in the first place, it's not obvious to me that this won't restrict rootless containers in some way).

crosbymichael:
Do we care about doing this type of validation? It adds a large dependency for the validation and I think over time, we have been moving validation like this out of runc and just rely on the underlying system to returns errors.

hqhq:
Setns in goroutine is a bad idea, any other ways we can do this validation?

dqminh:
I think its more reasonable / simpler to move these checks to the upper layer ( be it containerd/docker etc. ) since they have better access to what the user actually want using whatever flags user set up (i.e. --host in docker for example)

Doing these checks here is pretty complex compared to that. WDYT about just moving current logic to the upper layer, and add new logic there ?

crosbymichael:
I don't think this should be done in runc.  I think it will cause more problems than what it adds especially this late in the release cycle while we are trying for a 1.0 

REJECT

dqminh:
I have the same concern as @crosbymichael

Closing this for now. I think changes post 1.0 to move this logic entirely out of runc into upper layer would be nice, cc @mrunalp.

PR #1405, opened 2017-04-13
Makefile: `make install` requires ./runc

vbatts:
as if it weren't already implied. :-)

Signed-off-by: Vincent Batts <vbatts@redhat.com>


crosbymichael:
Nooooo! ;)

This would mean that when I do `sudo make install` it will search for `go` in my system path that does not exist.

vbatts:
@crosbymichael only if `./runc`

crosbymichael:
Personally I would like to keep it as `make` makes and `make install` installs and keep the split.

vbatts:
right on

PR #1406, opened 2017-04-14
Revert back to using /sbin

crosbymichael:
This was changed in
https://github.com/opencontainers/runc/commit/d2f49696#diff-b67911656ef5d18c4ae36cb6741b7965R7
and is causing install problems for people building runc and having it
installed in /bin and /sbin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
 PREFIX := $(DESTDIR)/usr/local
-BINDIR := $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


crosbymichael:
This change was probably made during development and forgotten/missed in code review.

cyphar:
It was intentional (it's not longer software that only admins can use), but I didn't realise this would cause issues.

LGTM.

crosbymichael:
Ya, the issue was `sbin` was trumping `bin` so people were installing, saying it does not work, etc.  And runc was never being updated on their system because the new version was being installed in `bin` and all the applications depending on runc were looking in sbin, including bash.

mrunalp:
LGTM

PR #1408, opened 2017-04-15
vendor: clean up to be better written

cyphar:
vndr doesn't support non-top-level imports, and in addition we really
should be using tagged releases far more than we currently are
(*especially* when it come to the OCI specs).


hqhq:
Needs rebase.

cyphar:
It also seems to break tests. Not sure why.

cyphar:
There. I've also fixed the tests @hqhq.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1409, opened 2017-04-18
Add timeout for syscall.Openat

keloyang:
Openat will be blocked until the fifo on the other side is  opened, but in some
abnomal scenario(e.g. containerd is killed), Openat maybe be blocked all the time.

### DIFF
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"syscall"
+	"time"
 
 	"github.com/opencontainers/runc/libcontainer/apparmor"
 	"github.com/opencontainers/runc/libcontainer/configs"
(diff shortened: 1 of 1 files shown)


keloyang:
###test case
add the patch for containerd:
diff --git a/runtime/process.go b/runtime/process.go
index bc36e05..8338ce3
--- a/runtime/process.go
+++ b/runtime/process.go
@@ -431,6 +431,8 @@ func (p *process) Start() error {
                        cmd  = exec.Command(p.container.runtime, args...)
                )
                go func() {
+                       time.Sleep(2*time.Second)
+                       os.Exit(1)
                        out, err := cmd.CombinedOutput()
                        if err != nil {
                                errC <- fmt.Errorf("%s: %q", err.Error(), out)
then start a container:
docker run --rm -ti euleros bash
on the other console:
[root@localhost docker-bin]# date;ps -eaf|grep exe
Tue Apr 18 07:27:30 EDT 2017
root     15973 15958  0 07:27 pts/0    00:00:00 /proc/self/exe init
root     16120 23108  0 07:27 pts/2    00:00:00 grep --color=auto exe
[root@localhost docker-bin]# kill -9 15973
[root@localhost docker-bin]# date;ps -eaf|grep exe
Tue Apr 18 07:29:40 EDT 2017
root     17161 23108  0 07:29 pts/2    00:00:00 grep --color=auto exe
[root@localhost docker-bin]# date;ps -eaf|grep exe
Tue Apr 18 07:29:41 EDT 2017
root     17212 17197  2 07:29 pts/0    00:00:00 /proc/self/exe init
root     17231 23108  0 07:29 pts/2    00:00:00 grep --color=auto exe
...
Tue Apr 18 07:31:26 EDT 2017
root     17212 17197  0 07:29 pts/0    00:00:00 /proc/self/exe init
root     17971 23108  0 07:31 pts/2    00:00:00 grep --color=auto exe
[root@localhost docker-bin]# date;ps -eaf|grep exe
Tue Apr 18 07:31:45 EDT 2017
root     18111 23108  0 07:31 pts/2    00:00:00 grep --color=auto exe
two minutes later，"/proc/self/exe init" will exit，if not add this patch for runc， "/proc/self/exe init" will be blocked all the time.

cyphar:
This patch doesn't make sense to me -- the whole point of the `create` and `start` split is that you can `create` a container and run a bunch of hooks separately to it (no matter how long that might take) and then `start` it. In fact, people can use `create` and then just use `exec` (without using `start`) -- which means this patch will break that workflow.

I don't think that we should have timeouts here because it seems like you're trying to fix a problem in containerd (it doesn't clean up containers that have already been started if containerd dies) inside runC rather than inside containerd.

williammartin:
Agree with Cyphar. In Garden, we never call `start` and exclusively `exec` to get processes running in the container. This would break us.

keloyang:
Ok, I got it ,thanks.

PR #1414, opened 2017-04-21
Update examples on README to allow rootless execution

### LINKED ISSUE #1413: runc run fails with "mkdir /run/runc: permission denied"
I followed the instructions found on README.md:

1. Create the filesystem: `docker export $(docker create busybox) | tar -C rootfs -xvf -`
2. Create the config: `runc spec`
3. Run the image: `runc run mycontainerid`

Got the error: `mkdir /run/runc: permission denied`

Am I missing something?

jwendell:
Closes #1413.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ Assuming you have an OCI bundle from the previous step you can execute the conta
 The first way is to use the convenience command `run` that will handle creating, starting, and deleting the container after it exits.
 
 ```bash
+# run as root
 cd /mycontainer
-
 runc run mycontainerid
 ```
 
@@ -165,8 +165,8 @@ Now we can go though the lifecycle operations in your shell.
 
 
 ```bash
+# run as root
 cd /mycontainer
-
 runc create mycontainerid
 
 # view the container is created and in the "created" state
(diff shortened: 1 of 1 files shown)


crosbymichael:
Maybe a separate section for rootless instead of overwriting the default?  What do you think? 

We really need to work on our docs and getting started so thank you for the PR

jwendell:
@crosbymichael I don't have a strong opinion on that. Thing is that currently the instructions are broken in the sense they are not saying you need to execute those commands as root. So, if we go this way of creating another section for rootless we need to fix the default section.

I'm happy to just fix the current doc adding the need of root if you prefer.

crosbymichael:
Well its implied that you have to run them as root because that has been the status quo for many many years ;)

cyphar:
I agree with @crosbymichael, it'd be nice if the docs split out the "normal" operation (run as root) and the new rootless mode.

jwendell:
OK, Just pushed a new version.

@crosbymichael It was not so clear to me, that's my first contact with runc and I was strictly following the tutorial on README, and tried to execute the commands as an ordinary user.

cyphar:
You need to include a `Signed-off-by` line in your commit. You can pass the `-s` flag to `git commit --amend` and git will include it automatically.

jwendell:
oops, sorry. Fixed.

README.md
```diff
@@ -107,6 +107,7 @@ After a root filesystem is populated you just generate a spec in the format of a
 To find features and documentation for fields in the spec please refer to the [specs](https://github.com/opencontainers/runtime-spec) repository.
 
 ```bash
+# run as root
```
cyphar: Nit: You don't need to run this as root.

jwendell: That's correct. Fixed.

cyphar:
LGTM

crosbymichael:
LGTM

Thanks @jwendell

PR #1420, opened 2017-04-25
ignore vim .swp files in the tree

tych0:
Signed-off-by: Tycho Andersen <tycho@docker.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ vendor/pkg
 contrib/cmd/recvtty/recvtty
 man/man8
 release
+*.swp


crosbymichael:
You should add these to your `.gitignore_global` in your home dir to do the same thing system wide ;)

kolyshkin:
I was about to create a similar PR (only using `.*.swp` not `*.swp`) since I saw a PR in this repo that accidentally added such files (and not because I was about to add such files myself).

@crosbymichael let me know if you reconsider :)

Issue #1421, opened 2017-04-26
setting Resources.MemorySwap unlimited after type change is messy


sjenning:
https://github.com/opencontainers/runc/pull/1375 changed `Resources.MemorySwap` to a `uint64`.

https://github.com/opencontainers/runc/blob/master/libcontainer/configs/cgroup_unix.go#L53-L54

Comment on that field is set to `-1` for unlimited, which can't be done with a `uint64`

This is blocking a vendored dep bump in kube: https://github.com/kubernetes/kubernetes/issues/43856

@derekwaynecarr

sjenning:
ok, well... I found this https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/memory.go#L107-L116

Apparently the solution is `uint64(-1)` which seems messy

sjenning:
Even more messy,  `uint64(-1)` doesn't even compile.  You have to trick it with:

```
unlimited := -1
res.MemorySwap = uint64(unlimited)
```

derekwaynecarr:
@mrunalp @vishh -- i find this awkward to say the least.

dqminh:
Yah i agree its a bit awkward, `math.MaxUint64` should do the job as well though.

justincormack:
The kernel specifically accepts `-1` in its parser, using `page_counter_memparse(args, "-1",...)` but we no longer allow this since this change, so it no longer matches the kernel behaviour, even if it is a "sort of" `uint64` it is a special one that also accepts `-1`.

In addition the maximum meaningful value is 2^63 due to the kernel/user memory split.

In addition some older kernels (RHEL 7.2 at least) will set the limits to `0` if they are larger than 2^63, and then immediately crash, so setting them to `2^64-1` makes them crash, while the previous `-1` worked correctly.

So we should revert this whole change.

cc @tiborvass

hqhq:
@justincormack `uint64` is actually the type kernel uses for memory limit, it accepts `-1` because we take this value as string in cgroup API, we can do the same thing in our user space tools if we see fit, or we can force change `-1` to uint64 which also works.

> In addition the maximum meaningful value is 2^63 due to the kernel/user memory split.
>
> In addition some older kernels (RHEL 7.2 at least) will set the limits to 0 if they are larger than 2^63, and then immediately crash, so setting them to 2^64-1 makes them crash, while the previous -1 worked correctly.

These are points I didn't realize, I'll double check these, doesn't make a lot of sense to me though.

[...]

PR #1422, opened 2017-04-26
define MemoryUnlimited const for users

sjenning:
Fixes https://github.com/opencontainers/runc/issues/1421

Currently, a user wanting to set `MemorySwap` to unlimited has to do a messy trick:
```
unlimited := -1
res.MemorySwap = uint64(unlimited)
```
as `uint64(-1)` doesn't compile.

This PR defines a `MemoryUnlimited` constant to make this cleaner.

@derekwaynecarr @mrunalp

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -104,14 +104,12 @@ func setKernelMemory(path string, kernelMemoryLimit uint64) error {
 }
 
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
-	ulimited := -1
-
-	// If the memory update is set to uint64(-1) we should also
-	// set swap to uint64(-1), it means unlimited memory.
-	if cgroup.Resources.Memory == uint64(ulimited) {
+	// If the memory update is set to MemoryUnlimited we should also
+	// set swap to MemoryUnlimited.
+	if cgroup.Resources.Memory == configs.MemoryUnlimited {
 		// Only set swap if it's enbled in kernel
 		if cgroups.PathExists(filepath.Join(path, cgroupMemorySwapLimit)) {
-			cgroup.Resources.MemorySwap = uint64(ulimited)
+			cgroup.Resources.MemorySwap = uint64(configs.MemoryUnlimited)
 		}
 	}
 
@@ -126,7 +124,7 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 		// When update memory limit, we should adapt the write sequence
 		// for memory and swap memory, so it won't fail because the new
 		// value and the old value don't fit kernel's validation.
-		if cgroup.Resources.MemorySwap == uint64(ulimited) || memoryUsage.Limit < cgroup.Resources.MemorySwap {
+		if cgroup.Resources.MemorySwap == uint64(configs.MemoryUnlimited) || memoryUsage.Limit < cgroup.Resources.MemorySwap {
 			if err := writeFile(path, cgroupMemorySwapLimit, strconv.FormatUint(cgroup.Resources.MemorySwap, 10)); err != nil {
 				return err
 			}
--- a/libcontainer/configs/cgroup_unix.go
+++ b/libcontainer/configs/cgroup_unix.go
@@ -2,6 +2,10 @@
 
 package configs
 
+import (
+	"math"
+)
+
 type FreezerState string
 
 const (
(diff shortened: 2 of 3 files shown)


derekwaynecarr:
ideally we could have this in the rc and not just master...

mrunalp:
@sjenning Could you sign the commit?

libcontainer/cgroups/fs/memory.go
```diff
@@ -104,14 +104,12 @@ func setKernelMemory(path string, kernelMemoryLimit uint64) error {
...
 
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
-	ulimited := -1
-
-	// If the memory update is set to uint64(-1) we should also
-	// set swap to uint64(-1), it means unlimited memory.
-	if cgroup.Resources.Memory == uint64(ulimited) {
+	// If the memory update is set to MemoryUnlimited we should also
+	// set swap to MemoryUnlimited.
+	if cgroup.Resources.Memory == uint64(configs.MemoryUnlimited) {
```
wking: With this commit, `MemoryUnlimited` is defined as a `unit64`, so I think you can drop the redundant cast from this line (and the later lines which also have `uint64(configs.MemoryUnlimited)`.

sjenning:
@mrunalp @wking signed the commit and removed the unneeded type cast

cyphar:
@sjenning "Sign" refers the `Signed-off-by` line in your commit message. You can add it with the `-s` flag to `git commit --amend` (it signals that you can attest the [Developer Certificate of Origin](https://developercertificate.org/)).

sjenning:
@cyphar oops! fixed.

mrunalp:
LGTM

hqhq:
@sjenning Sorry for the confusion. `uint64(-1)` is not some random trick, I use it because `-1` is what cgroup api defines for unlimited usage, and we also export it to `runc update` and use this trick in https://github.com/opencontainers/runc/blob/v1.0.0-rc3/update.go#L228 , so what `memory.go` does just pairs that.

I agree this can be more explicit, but I don't see how `^uint64(0)` could be less messy, why not just use `math.MaxUint64`? And if you do this change, you may also need to fix what we did in `update.go`.

Thanks for doing this.

sjenning:
@hqhq thanks for the review.  I've made the requested changes.

libcontainer/cgroups/fs/memory.go
```diff
@@ -104,14 +104,12 @@ func setKernelMemory(path string, kernelMemoryLimit uint64) error {
...
-	// If the memory update is set to uint64(-1) we should also
-	// set swap to uint64(-1), it means unlimited memory.
-	if cgroup.Resources.Memory == uint64(ulimited) {
+	// If the memory update is set to MemoryUnlimited we should also
+	// set swap to MemoryUnlimited.
+	if cgroup.Resources.Memory == configs.MemoryUnlimited {
 		// Only set swap if it's enbled in kernel
 		if cgroups.PathExists(filepath.Join(path, cgroupMemorySwapLimit)) {
-			cgroup.Resources.MemorySwap = uint64(ulimited)
+			cgroup.Resources.MemorySwap = uint64(configs.MemoryUnlimited)
```
dqminh: i think you dont need to cast this to uint64 anymore

libcontainer/cgroups/fs/memory.go
```diff
@@ -126,7 +124,7 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 		// When update memory limit, we should adapt the write sequence
 		// for memory and swap memory, so it won't fail because the new
 		// value and the old value don't fit kernel's validation.
-		if cgroup.Resources.MemorySwap == uint64(ulimited) || memoryUsage.Limit < cgroup.Resources.MemorySwap {
+		if cgroup.Resources.MemorySwap == uint64(configs.MemoryUnlimited) || memoryUsage.Limit < cgroup.Resources.MemorySwap {
```
dqminh: same as the above, casting is not necessary.

libcontainer/configs/cgroup_unix.go
```diff
@@ -33,6 +37,8 @@ type Cgroup struct {
 	*Resources
 }
 
+const MemoryUnlimited = math.MaxUint64
```
justincormack: This is not the value that is actually used by the kernel, which is 9223372036854771712. Some older kernels will set any value larger than this to 0, and then crash, eg RHEL 7.2.

justincormack: Also this value may well be totally incorrect on 32 bit machines, not sure what happens there.

hqhq:
I've carried this PR in https://github.com/opencontainers/runc/pull/1494 , thanks for your contribution @sjenning .

PR #1428, opened 2017-04-28
Remove redundant declaration of namespace slice

harche:
Namespaces slice is declared twice. Let's keep list 
namespaces at one place instead of duplicating it. Should
there be any change in future in the number of namespaces
or anything like that, we just have to change in one 
location.

### DIFF
--- a/libcontainer/configs/namespaces_unix.go
+++ b/libcontainer/configs/namespaces_unix.go
@@ -64,12 +64,12 @@ func IsNamespaceSupported(ns NamespaceType) bool {
 
 func NamespaceTypes() []NamespaceType {
 	return []NamespaceType{
+		NEWUSER, // Keep user NS always first, don't move it.
+		NEWIPC,
+		NEWUTS,
 		NEWNET,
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

libcontainer/configs/namespaces_unix.go
```diff
@@ -64,12 +64,12 @@ func IsNamespaceSupported(ns NamespaceType) bool {
 
 func NamespaceTypes() []NamespaceType {
 	return []NamespaceType{
+		NEWUSER, // Keep user NS always first, don't move it.
```
crosbymichael: Can you finish re-ordering this list to match the one that was deleted?

harche: @crosbymichael done. Thanks.

TomSweeneyRedHat:
LGTM

crosbymichael:
LGTM

Thanks @harche

mrunalp:
LGTM

PR #1433, opened 2017-05-02
tests: don't call wait_for_container after synchronous operations

avagin:
It is useless and potentially we can skip bugs.

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -47,7 +47,6 @@ EOF
     # run a detached busybox to work with
     runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_kmem
     [ "$status" -eq 0 ]
-    wait_for_container 15 1 test_cgroups_kmem
 
     # update kernel memory limit
     runc update test_cgroups_kmem --kernel-memory 50331648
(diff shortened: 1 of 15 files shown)


cyphar:
The reason for `wait_for_container` is to avoid races in tests between detached container runs and operations on the detached container. I don't have a problem with removing them from the checkpoint/restore tests if you're sure they're not necessary but I'm fairly sure that we do need them for most of the other tests.

avagin:
@cyphar I think I don't understand the problem. Could you give more details?

Do you want to say that "runc run -d test" exits before the container achieves the running state? It sounds weird.

avagin:
Cc: @mikebrow

avagin:
@cyphar I tried to put sleep in different places and everything works fine.

cyphar:
Hmmm, okay. I'm wondering if this is something that we fixed after the `--console-socket` changes (where init will block until the socket is sent) and thus we no longer have races like that. In which case, I'm fine with dropping it.

avagin:
@cyphar thank you for the feedback. I think that we want to detect this sort of issues in future, so it probably better to remove this hack.

mikebrow:
yeah the wait stuff was to fix the timing `feature` where runc run would return before the container can be managed with a subsequent runc call.  If the call is fully synchronous now the feature is broken and we can delete the waits :-)

cyphar:
@mikebrow Yeah, that "feature" sounds like a bug to me. We should drop it then IMO. In addition I've recently found a bunch of race conditions in our state transition code that we'll need to sort out as well.

cyphar:
LGTM.

dqminh:
LGTM

PR #1434, opened 2017-05-03
WIP - ignore * when reading from /dev/console

TomasTomecek:
This is my attempt to initiate fix of https://github.com/moby/moby/issues/27202. This PR started from code posted inside this comment: https://github.com/moby/moby/issues/27202#issuecomment-294493514

Using the current code I just wanted to make sure that if I ignore sighups (actually all errors), whether I get output from systemd. I did. The next step should be to decide how to fix this issue. Probably the best way would be to use epoll, which would obviously work only on linux: not sure whether if running on linux we should use epoll and other systems should use current upstream solution.

The issue is that systemd closes `/dev/console` because of [kernel SAK](https://www.kernel.org/doc/Documentation/SAK.txt) -- it's arguable whether this behaviour is a bug/feature in kernel, systemd or runc.

Signed-off-by: Tomas Tomecek <ttomecek@redhat.com>

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/docker/docker/pkg/term"
 	"github.com/opencontainers/runc/libcontainer"
@@ -77,7 +78,26 @@ func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) error {
 	console := libcontainer.ConsoleFromFile(f)
 	go io.Copy(console, os.Stdin)
 	t.wg.Add(1)
-	go t.copyIO(os.Stdout, console)
+	go func() {
+		// t.copyIO(os.Stdout, console)
+		buf := make([]byte, 1)
+		for {
+			nr, err := console.Read(buf)
+			if nr > 0 {
+				nw, ew := os.Stdout.Write(buf[0:nr])
+				if ew != nil {
+					break
+				}
+				if nr != nw {
+					break
+				}
+			}
+			time.Sleep(5 * time.Millisecond)
+			if err != nil {
+				// this is the place where we should care only about sighup
+			}
+		}
+	}()
 	state, err := term.SetRawTerminal(os.Stdin.Fd())
 	if err != nil {
 		return fmt.Errorf("failed to set the terminal from the stdin: %v", err)


cyphar:
Eugh, I really don't like that we have to rewrite `io.Copy`. :disappointed:

cyphar:
As an aside, while the `/dev/console` behaviour might make sense on the host, it makes literally no sense inside a container -- since `/dev/console` isn't actually the "console" proper. It's just a symlink we use because some legacy apps want to write to `/dev/console` for some stupid reason (such as systemd it seems). And I don't know of **any** container runtime that bindmounts the host's `/dev/console` -- because it would be a horrific security issue.

So (as much as I hate punting issues like this because the systemd folks tend to get annoyed) why can't systemd only enable this SAK security behaviour if `/dev/console` is actually a console (it has the major and minor numbers of a console)?

cyphar:
I've left a comment on the systemd issue.

However, one more thing to note is that changing this here won't actually fix the bug. When you run `runc` in a detached mode (like `containerd` does) then `runc` doesn't stick around -- the file descriptors are handed to the parent process and are managed by it (this is especially true as of #1018 which I don't think Docker has updated to yet). So you would also need to fix this problem there.

TomasTomecek:
@cyphar thank you for commenting inside the systemd issue. As I understand this now, both sides resist to fix the issue in their codebases because you and Lennart think it is a hack, is this correct? I'm not sure how to proceed.

This https://github.com/moby/moby/issues/9212 seems to be the same issue as the systemd issue. So Lennart seems to be correct that every init has this behaviour.

crosbymichael:
Is there a reason why you even run systemd with a `--tty`?  Just run it without and let it create a `/dev/console`?

TomasTomecek:
I would love to see output from systemd. systemd acts as a true daemon and "closes" its `std{in,out,err}`:

```
[root@runc /]# ls -lha /proc/1/fd
total 0
dr-x------ 2 root root  0 May  9 09:29 .
dr-xr-xr-x 9 root root  0 May  9 09:29 ..
lrwx------ 1 root root 64 May  9 09:29 0 -> /dev/null
lrwx------ 1 root root 64 May  9 09:29 1 -> /dev/null
lrwx------ 1 root root 64 May  9 09:29 2 -> /dev/null
```

Hence `/dev/console` is likely the only way to get logs from systemd.

hqhq:
TBH we are using the same hack in our internal version but for containerd-shim, I don't quite like it though.

TomasTomecek:
closing in favor of #1446

Issue #1429, opened 2017-05-02
regex for checking id 


harche:
The regex for validating the container id, https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L31 lets anyone specify an unlimited string length for an id. 

For all practical purposes id of 64 characters is more than sufficient (which is what docker daemon generates too, e.g. `8b72c0494d86930f8eb5c2f972b75813e225bceb3d3f9c7713be1977b8031456`). Should we have a tighter regex to enforce that? Or having unrestricted string length is ok for id string?

IMO, it should be replaced with something like, 
`
^[\w+-\.]{1,64}$
`

Any thoughts?

cyphar:
I'm not sure why we would intentionally limit the length of container IDs, it's not something that makes much sense to me. Practically speaking there is a limit to the length of the container ID (since we store state information in a directory, the maximum file name length on the filesystem is a factor) but I don't see why we should artificially limit something.

Is there a security or other argument for this?

harche:
@cyphar I wrote that from the security point of view only. At this point, I don't know if it opens any security related issue, but having something defined of arbitrary length and completely unbounded makes me little uncomfortable. But again, I could be just overthinking about it.

harche:
@cyphar In fact you do check for legth of id,
https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L333

Then why not to handle it in regex itself instead of doing if else later on?

harche:
So in go version `1.7.4` that I am using `^[\w+-\.]{1,1000}$` is valid but `^[\w+-\.]{1,1024}$` is not. So it limits the characters to 1000, not 1024. But if we do this, we can avoid one `if else` at https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L333

harche:
So in go version `1.7.4` that I am using `^[\w+-\.]{1,1000}$` is valid but `^[\w+-\.]{1,1024}$` is not. So it limits the characters to 1000, not 1024.

But if we do this, we can avoid one if else at https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L333

cyphar:
@harche Apparently we switched away from regex [in the past](https://github.com/opencontainers/runc/commit/de57f78590cefdf08249118e2d710908cc81b16a). I'm not entirely sure why we are limiting the length though -- @mrunalp?

[...]

PR #1435, opened 2017-05-04
Issue #1429 : Removing check for id string length

harche:
Based on the discussions in Issue #1429, we should not 
be checking the length of the id string


crosbymichael:
LGTM

dqminh:
LGTM

PR #1442, opened 2017-05-09
Move libcontainer to x/sys/unix

clnperez:
Since syscall is outdated and broken for some architectures,
use x/sys/unix instead.

There are still some dependencies on the syscall package that will
remain in syscall for the forseeable future:

Errno
Signal
SysProcAttr

Additionally, os still uses syscall, so it needs to be kept for anything
returning *os.ProcessState, such as process.Wait.

Signed-off-by: Christy Perez <christy@linux.vnet.ibm.com>


clnperez:
This fixes several tests that were failing on POWER (ppc64le) -- including some criu tests. I tested using the same test suite but with a modified dockerfile and makefile to pull in the right power bits, and ran `make test`.

clnperez:
On a closer look, there's not that much code outside of libcontainer that doesn't also need to be changed, so WDYT about replacing syscall everywhere else as well?

clnperez:
Hm. Tests failing again. Hold off on this for now.

libcontainer/specconv/spec_linux.go
```diff
@@ -710,7 +711,7 @@ func setupSeccomp(config *specs.LinuxSeccomp) (*configs.Seccomp, error) {
 		return nil, nil
 	}
 
-	// No default action specified, no syscalls listed, assume seccomp disabled
+	// No default action specified, no unixs listed, assume seccomp disabled
```
wking: I think this (and a few later entries) should keep `syscalls`.  In #1445 I replaced `sycalls.` with `unix.` (including the trailing period), and while that can still have false positives (e.g. when the word is at the end of a comment sentence), it won't accidentally do things like this.

clnperez: Oops. Yep. I thought I caught all the spots I needed to keep syscalls. I'll take a closer look.

clnperez:
Fixed those error message and comment false positives @wking. PTAL.

wking:
On Thu, May 11, 2017 at 07:59:15AM -0700, christyp wrote:
> Fixed those error message and comment false positives @wking. PTAL.

I haven't looked at the rest of this PR, but as of 67d40a82, the only
difference between it and #1445 for
libcontainer/specconv/spec_linux.go is a blank line in the import
block.  So that looks fine to me.

wking:
Travis isn't happy.  It looks like the [tests][1] [hung][2]?

[1]: https://travis-ci.org/opencontainers/runc/jobs/231199433#L2779
[2]: https://travis-ci.org/opencontainers/runc/jobs/231199432#L2772

libcontainer/factory_linux.go
```diff
@@ -303,7 +304,7 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 		return err
 	}
 
-	// If Init succeeds, syscall.Exec will not return, hence none of the defers will be called.
+	// If Init succeeds, unix.Exec will not return, hence none of the defers will be called.
```
dqminh: this should still be syscall.Exec

libcontainer/init_linux.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/opencontainers/runc/libcontainer/user"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/vishvananda/netlink"
+
+	sysunix "golang.org/x/sys/unix"
```
dqminh: why do we use sysunix import name here ?

libcontainer/system/setns_linux.go
```diff
@@ -1,40 +0,0 @@
...
-	ns, exists := setNsMap[fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)]
-	if !exists {
-		return fmt.Errorf("unsupported platform %s/%s", runtime.GOOS, runtime.GOARCH)
-	}
-	_, _, err := syscall.RawSyscall(ns, fd, flags, 0)
-	if err != 0 {
-		return err
-	}
-	return nil
-}
```
dqminh: thanks for removing unused functions !

list.go
```diff
@@ -7,7 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"syscall"
+	"syscall" // avoid use if possible
```
dqminh: i think this comment is not required.

libcontainer/init_linux.go
clnperez: Weird. I swear there was a conflict here. I'll change it to unix, thanks.

clnperez:
Strange that this failed with go 1.8, but not 1.7, and not go tip. Is that failure something that anyone has seen before?

clnperez:
For reference (because I'm about to rebase) the failure seen on go 1.8 is:
```
--- FAIL: TestNotifyOnOOM (0.00s)
	notify_linux_test.go:99: expected event control to be closed
	notify_linux_test.go:103: expected event fd to be closed
```

clnperez:
😆 So *this* time -- that same test passed with 1.8 and failed with `go tip` and 1.7. I'll take a look at that test.

clnperez:
I added the error to the debug output so that I could see what error it's getting, since not `unix.EBADF` -- but this time all three go versions' tests passed. WDYT @dqminh, @wking?

wking:
On Mon, May 15, 2017 at 12:12:51PM -0700, christyp wrote:
> … but this time all three go versions' tests passed. WDYT @dqminh,
> @wking?

Maybe amend the commit to get a new commit date (and commit hash), and
repeat until you can trigger the failure again?  It certainly seems
like a flaky-test issue to me, but I'm not a maintainer so I can't
speak to whether fixing the test is a blocker for this PR or not.

clnperez:
Welp, I tried several times and just can't get this to fail again.

clnperez:
Does this look okay to merge? I can't get the test to fail again so it must be flaky.

crosbymichael:
@clnperez this needs to be rebased

libcontainer/devices/devices_linux.go
```diff
@@ -36,13 +37,13 @@ func DeviceFromPath(path, permissions string) (*configs.Device, error) {
...
-		fileModePermissionBits |= syscall.S_IFCHR
+		fileModePermissionBits |= unix.S_IFCHR
 		devType = 'c'
 	default:
-		fileModePermissionBits |= syscall.S_IFBLK
+		fileModePermissionBits |= unix.S_IFBLK
 		devType = 'b'
 	}
-	stat_t, ok := fileInfo.Sys().(*syscall.Stat_t)
+	stat_t, ok := fileInfo.Sys().(*unix.Stat_t)
```
crosbymichael: May need to verify this line, i'm not sure if the cast will work

crosbymichael: Ya, you will have to revert this line.

`panic: interface conversion: interface {} is *syscall.Stat_t, not *unix.Stat_t`

clnperez: i thought i had fixed that issue. hold tight on reviews while i take a look at my rebase.

crosbymichael: It also looks like you have an unrelated commit `	shfmt'ing files`

clnperez: i put that in there b/c the tests failed (but also kept it separate b/c it is unrelated)

clnperez:
@crosbymichael PTAL. Three questions:
- Can I leave the other usages of Stat_t as-is (`unix.Stat_t`)?
- What should I do about the unrelated shfmt commit? The CI fails me I don't fix those up. 
- Should I squash the two code changes (or three...) into one commit? I was originally planning on only updating libcontainer but there weren't many more files outside of it that needed switching over.

clnperez:
Tests seem a little flaky. All passing now.

crosbymichael:
LGTM

I don't know anything about why the tests are failing because of shfmt.

`unix.Stat_t` is food for all the other cases because you are using it in the syscall and not what FileInfo returns.

Thanks!

dqminh:
LGTM

PR #1444, opened 2017-05-11
WIP: libcontainer/specconv/spec_linux: defaults should not be a no-op

wking:
It has been since it landed in 9fac1832, but [the spec currently references mount(8) for these options][1] and [mount(8) has][2]:

> * defaults
>
>     Use the default options: rw, suid, dev, exec, auto, nouser, and async.
>
>     Note that the real set of all default mount options depends on kernel and filesystem type.  See the beginning of this section for more details.

I exepect that “real set” paragraph applies to:

> Note that filesystems also have per-filesystem specific default mount options (see for example tune2fs -l output for extN filesystems).

For what its worth, util-linux 2.28.2 seems to ignore `defaults` instead of clearing bits:

    # strace -o /tmp/trace mount -t tmpfs -o ro,defaults - /tmp/a
    # grep 'mount(' /tmp/trace
    mount("-", "/tmp/a", "tmpfs", MS_MGC_VAL|MS_RDONLY, NULL) = 0

While a single-bit clear option does unset an earlier bit:

    # strace -o /tmp/trace mount -t tmpfs -o ro,rw - /tmp/a
    # grep 'mount(' /tmp/trace
    mount("-", "/tmp/a", "tmpfs", MS_MGC_VAL, NULL) = 0

but the spec is currnently punting to the util-linux mount(8) page and not to the util-linux implementation.

Related to opencontainers/runtime-spec#771.

[2]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -641,7 +641,7 @@ func parseMountOptions(options []string) (int, []int, string, int) {
 		"async":         {true, syscall.MS_SYNCHRONOUS},
 		"atime":         {true, syscall.MS_NOATIME},
 		"bind":          {false, syscall.MS_BIND},
-		"defaults":      {false, 0},
+		"defaults":      {true, syscall.MS_NODEV | syscall.MS_NOEXEC | syscall.MS_NOSUID | syscall.MS_RDONLY | syscall.MS_SYNCHRONOUS},
 		"dev":           {true, syscall.MS_NODEV},
 		"diratime":      {true, syscall.MS_NODIRATIME},
 		"dirsync":       {false, syscall.MS_DIRSYNC},


wking:
I've [posted a request for clarification][1] to the util-linux list to resolve the differences between my reading of the man page and the util-linux implementation.

[1]: http://marc.info/?l=util-linux-ng&m=149450344322927&w=2

wking:
Updated mount(8) wording is still up in the air, but [Karel Zak confirmed that runC's current `defaults` is correct and my interpretation here is wrong][1].

[1]: http://marc.info/?l=util-linux-ng&m=149450548823490&w=2

PR #1445, opened 2017-05-11
libcontainer/specconv/spec_linux: Use golang.org/x/sys/unix

wking:
`syscall` [is deprecated][1] and `sys/unix` [contains newer flags like `MS_LAZYTIME`][2] which we need to implement all the [filesystem-independent mount options referenced from the spec][3].

Related to opencontainers/runtime-spec#771.

[1]: https://golang.org/pkg/syscall/
[2]: https://godoc.org/golang.org/x/sys/unix
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L68

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -9,13 +9,13 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/seccomp"
 	libcontainerUtils "github.com/opencontainers/runc/libcontainer/utils"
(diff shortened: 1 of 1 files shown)


tianon:
This is a dup of #1442, isn't it?

wking:
This is a subset of #1442, ping @clnperez.

wking:
On Wed, May 10, 2017 at 10:09:03PM -0700, Tianon Gravi wrote:
> This is a dup of #1442, isn't it?

It's a smaller chunk (just what I need for
opencontainers/runtime-spec#771).  But it looks like Travis is happy
with #1442 at the moment [1], so if it's easier to review the more
comprehensive PR that works for me.

[1]: https://travis-ci.org/opencontainers/runc/builds/230550007

crosbymichael:
This is a dup of #1442, closing

PR #1447, opened 2017-05-12
Clean up unix vs linux usage

justincormack:
FreeBSD does not support cgroups or namespaces, which the code suggested, and is not supported
in runc anyway right now. So clean up the file naming to use `_linux` where appropriate.


dqminh:
LGTM

crosbymichael:
LGTM

PR #1449, opened 2017-05-15
fix panic when Linux is nil

Mashimiao:
Linux is not always not nil.
If Linux is nil, panic will occur.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -184,20 +184,7 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	}
 
 	exists := false
-	if config.RootPropagation, exists = mountPropagationMapping[spec.Linux.RootfsPropagation]; !exists {
-		return nil, fmt.Errorf("rootfsPropagation=%v is not supported", spec.Linux.RootfsPropagation)
-	}
 
-	for _, ns := range spec.Linux.Namespaces {
-		t, exists := namespaceMapping[ns.Type]
-		if !exists {
-			return nil, fmt.Errorf("namespace %q does not exist", ns)
-		}
-		if config.Namespaces.Contains(t) {
-			return nil, fmt.Errorf("malformed spec file: duplicated ns %q", ns)
-		}
-		config.Namespaces.Add(t, ns.Path)
-	}
 	if config.Namespaces.Contains(configs.NEWNET) {
 		config.Networks = []*configs.Network{
 			{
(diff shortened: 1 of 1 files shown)


crosbymichael:
Instead of having this check everywhere shouldn't we make sure that its non-nil at the beginning?

Mashimiao:
You mean change spec object after loading config.json depends on whether Linux{} is set or not ?

crosbymichael:
@Mashimiao validate once after we read the spec and if it is `nil` fail before we start executing.

Mashimiao:
On 05/17/2017 12:57 AM, Michael Crosby wrote:
> @Mashimiao <https://github.com/mashimiao> validate once after we read the spec and if it is |nil| fail before we start executing.

If so, this will be a problem.
The runtime-spec says linux entry is not necessary, just optional even if paltform.os is Linux.

crosbymichael:
@Mashimiao oh, ok.  Never mind then ;)

libcontainer/specconv/spec_linux.go
```diff
@@ -180,20 +180,11 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	}
 
 	exists := false
-	if config.RootPropagation, exists = mountPropagationMapping[spec.Linux.RootfsPropagation]; !exists {
-		return nil, fmt.Errorf("rootfsPropagation=%v is not supported", spec.Linux.RootfsPropagation)
+	linuxSpecific := false
+	if spec.Linux != nil {
+		linuxSpecific = true
 	}
```
cyphar: This variable is not necessary since you only use it once. As an aside, the nice way of writing this would be

```go
linuxSpecific := spec.Linux != nil
```

justincormack:
Yes, I noticed this panics recently and hadn't opened an issue yet, it does need fixing...

justincormack:
A simpler solution would be just to set it if nil to the empty config.

justincormack:
Needs a rebase.

It would be really nice to fix this as it is a stupid panic and makes testing harder.

libcontainer/specconv/spec_linux.go
Mashimiao: removed such variable.

Mashimiao:
@justincormack  rebased.

Mashimiao:
@hqhq I think we need re-open this PR, because #1551 does not fix Linux nil problem totally.

crosbymichael:
@Mashimiao please open a new issue with specifically what is still breaking so we can resolve it.  We don't have the information here to go on.

Mashimiao:
OK, I submit a new PR #1559  to fix the problem

PR #1451, opened 2017-05-16
Ignore error when force deleting a non-existing container

runcom:
This patch mimics the behavior of "rm -rf" so that if a container
doesn't exist and you force delete it, it won't error out.


runcom:
ping @crosbymichael @mrunalp @hqhq @dqminh

mrunalp:
LGTM

crosbymichael:
LGTM

PR #1453, opened 2017-05-19
Propagate the correct argv0 when re-execing

petrosagg:
This allows runc to be used as a target for docker's reexec module that depends on a correct argv0 to select which process entrypoint to invoke. Without this patch, when runc re-execs argv0 is set to "/proc/self/exe" and the reexec module doesn't know what to do with it.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -40,6 +40,7 @@ type linuxContainer struct {
 	config               *configs.Config
 	cgroupManager        cgroups.Manager
 	intelRdtManager      intelrdt.Manager
+	initPath             string
 	initArgs             []string
 	initProcess          parentProcess
 	initProcessStartTime uint64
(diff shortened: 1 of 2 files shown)


crosbymichael:
I remember us having issues with this in the past and `/proc/self/exe` was the most robust and efficient way to do a reexec.

crosbymichael:
@petrosagg what is the actual error that you are seeing and how can you reproduce this?

cyphar:
@crosbymichael It looks like this just sets the `argv[0]` (not the `path` for `exec`), which isn't /wrong/ necessarily. But `runc` shouldn't have any issues with the name being `/proc/self/exe` so I'd also like some more information about what issue this is mean to be solving.

petrosagg:
The problem it's trying to solve isn't with runc itself, but it makes runc easier to integrate in a busybox-like system. What I mean is something like this: https://github.com/resin-os/balena/blob/17.06-resin/cmd/balena/main.go#L22-L36

In Balena we make a single binary that contains docker-cli, moby, containerd, runc etc, so when runc re-execs we need to know the correct arg0. This patch is meant to address this usecase

petrosagg:
I rebased the commit on top of master and re-wrote its message to something more accurate

gireeshpunathil:
I am just wondering the implications (with  or without this change actually)  in case if one is using a relative path to spawn `runc` originally, and then change the `cwd` within the process and then try to re-exec itself. Won't the relative mention of its path name cause issue as the base location has now changed? just a thought.

petrosagg:
Changing the cwd won't affect the re-exec process because `/proc/self/exe` is a special file that is basically the image of the binary in memory, which stays the same regardless of cwd. Even if the original binary gets deleted from the filesystem, `/proc/self/exe` can be used to re-exec.

crosbymichael:
Thanks, makes sense

LGTM

cyphar:
LGTM.

PR #1460, opened 2017-05-26
libcontainer/specconv/spec_linux: Add support for (no)lazytime

wking:
Part of catching runC up with the spec, which [punts valid options to `mount(8)`][1].  Part of opencontainers/runtime-spec#771.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L68

PR #1461, opened 2017-05-26
libcontainer/specconv/spec_linux: Add support for silent and loud

wking:


wking:

PR #1463, opened 2017-05-26
libcontainer/specconv/spec_linux: Add support for (no)acl

wking:

This is a filesystem-specific entry in `mount(8)`, but it's represented by a `MS_*` flag in `mount(2)` so we need an entry in the translation table.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -639,6 +639,7 @@ func parseMountOptions(options []string) (int, []int, string, int) {
 		clear bool
 		flag  int
 	}{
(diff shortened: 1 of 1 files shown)


wking:
I'm not sure why Travis is unhappy.  It looks like the tests passed on Go 1.8 but failed on [1.7][] and [tip][] with a `checkpoint --pre-dump and restore` issue?  Is that test reliable?

[1.7]: https://travis-ci.org/opencontainers/runc/jobs/236437672#L2971
[1.8]: https://travis-ci.org/opencontainers/runc/jobs/236437673
[tip]: https://travis-ci.org/opencontainers/runc/jobs/236437674#L2978

wking:

PR #1470, opened 2017-05-30
Use symlink xattr functions from x/sys/unix

tklauser:
Update x/sys/unix to get the symlink xattr syscall wrappers Lgetxattr, Llistxattr and Lsetxattr (introduced in
golang/sys@b90f89a1e7a9c1f6b918820b3daa7f08488c8594) and switch libcontainer to use these instead of providing own wrappers.


cyphar:
LGTM. I should probably also move `umoci` to use `golang.org/x/sys/unix`...

crosbymichael:
LGTM

PR #1474, opened 2017-06-02
Use NLA_* constants from x/sys/unix instead of syscall

tklauser:
Use the NLA_ALIGNTO and NLA_HDRLEN constants from x/sys/unix instead of
syscall, as the syscall package shouldn't be used anymore (except for a
few exceptions).

This also makes the syscall_NLA_HDRLEN workaround for gccgo unnecessary.

### DIFF
--- a/libcontainer/message_linux.go
+++ b/libcontainer/message_linux.go
@@ -3,9 +3,8 @@
 package libcontainer
 
 import (
-	"syscall"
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

cyphar:
LGTM

PR #1475, opened 2017-06-02
Add mips support

vstefanovic:


crosbymichael:
LGTM

libcontainer/system/syscall_linux_mipsx.go
```diff
@@ -0,0 +1,25 @@
+// +build linux,mips linux,mipsle linux,mips64 linux,mips64le
+
+package system
+
+import (
+	"syscall"
```
cyphar: I wonder if these are provided by `golang.org/x/sys/unix` (and so we could just switch `syscall_linux` to use that)?

mrunalp: @vstefanovic Could you reply here? (otherwise PR looks good). Thanks!

vstefanovic: Unlikely, there's no Setuid and Setgid in golang.org/x/sys/unix/syscall_linux_*.go.

cyphar:
LGTM. That's a shame (/cc @tklauser -- maybe they should be added to `golang.org/x/sys/unix`?).

tklauser:
@cyphar thanks for the ping. Yes, would make sense to add them to `x/sys/unix'. I'll send a corresponding CL.

tklauser:
@cyphar Regarding adding `Setuid/Setgid` to `x/sys/unix', also see golang/go#1435. It looks like they intentionally didn't add it and added placeholder functions returning `EOPNOTSUPP` with the following comment:

```
// On linux Setuid and Setgid only affects the current thread, not the process.
// This does not match what most callers expect so we must return an error
// here rather than letting the caller think that the call succeeded.
```

cyphar:
@tklauser I would assume that people would expect (in Go) that it would be implemented the same way as `nptl(7)` (using internal signals -- which is obviously a massive pain that shouldn't be necessary). But reading that issue, I guess that Russ Cox doesn't feel that it's worth it? That's a shame -- because `syscall` is still "broken" in the same way. [As an aside, we don't really care about it being broken too much in `runc` because I believe that we handle it properly (since we eventually hit an `exec`).]

tklauser:
@cyphar FWIW, the `syscall` package has the exact same implementation as `x/sys/unix` (i.e. returning `EOPNOTSUPP`). I agree that the current situation with projects having to re-implement Setuid/Setgid themselves is a shame. Let's see whether I can find some time to dig deeper into this and contribute some work towards making `Setuid/Setgid/...` available in Go in a manner that matches `nptl(7)` behavior.

libcontainer/system/syscall_linux_mipsx.go
tklauser: Couldn't the implementation from `libcontainer/system/syscall_linux_64.go` have been used here? I'll sent a follow-up PR in #1613.

cyphar:
Thanks so much @tklauser. :heart:

PR #1478, opened 2017-06-07
Use Prctl() from x/sys/unix instead of own wrapper

tklauser:
Use unix.Prctl() instead of reimplemnting it as system.Prctl().


crosbymichael:
LGTM

cyphar:
LGTM.

PR #1479, opened 2017-06-08
libcontainer/console_linux.go: Make SaneTerminal public

wking:
And use it only in local tooling that is forwarding the pseudoterminal master.  That way runC no longer has an opinion on the `onlcr` setting for folks who are creating a terminal and detaching.  They'll use `--console-socket` and can setup the pseudoterminal however they like without runC having an opinion.  With this commit, the only cases where runC still applies SaneTerminal is when *it* is the process consuming the master descriptor.

Previous discussion in #1146 (which landed `saneTerminal` in the first place) and also [on IRC earlier today][1].  From that IRC discussion, this PR takes the “whoever gets the master descriptor can set it up as they see fit” approach.  And alternative floated by @crosbymichael was to [expose some subset of the termios structure in the config][2] which had later [support][3] from @mrunalp.  After further discussion, both @cyphar ([here][4]) and @crosbymichael ([here][5]) seem ok with the “setup the master descriptor as you see fit” approach I've taken here (although that doesn't mean they'll approve of the way I've implemented that approach).

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-06-07.log.html#t2017-06-07T17:17:44
[2]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-06-07.log.html#t2017-06-07T20:47:27
[3]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-06-07.log.html#t2017-06-07T21:03:25
[4]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-06-07.log.html#t2017-06-07T23:17:56
[5]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-06-07.log.html#t2017-06-07T23:22:26

### DIFF
--- a/contrib/cmd/recvtty/recvtty.go
+++ b/contrib/cmd/recvtty/recvtty.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/urfave/cli"
 )
@@ -98,6 +99,9 @@ func handleSingle(path string) error {
 	// Get the master file descriptor from runC.
 	master, err := utils.RecvFd(socket)
 	if err != nil {
+		return err
+	}
+	if err = libcontainer.SaneTerminal(master); err != nil {
 		return err
 	}
 
--- a/libcontainer/console_linux.go
+++ b/libcontainer/console_linux.go
@@ -21,9 +21,6 @@ func newConsole() (Console, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := saneTerminal(master); err != nil {
-		return nil, err
-	}
 	console, err := ptsname(master)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


wking:
I've pushed 147fc3a → 830c0d7 to [fix][1]:

```
=== RUN   TestExecInTTY
--- FAIL: TestExecInTTY (0.29s)
	execin_test.go:335: unexpected carriage-return in output
```

[1]: https://travis-ci.org/opencontainers/runc/jobs/240647031#L2843

wking:
Travis is complaining about the [checkpoint issue][1] I've been getting occasionally, which I think is unrelated to this PR (see also [here][2] and [here][3]).  This time it's only for Go 1.8.

[1]: https://travis-ci.org/opencontainers/runc/jobs/240648562#L2949
[2]: https://github.com/opencontainers/runc/pull/1463#issuecomment-304338428
[3]: https://github.com/opencontainers/runc/pull/1460#issuecomment-304769891

crosbymichael:
LGTM

ping @mrunalp

mrunalp:
LGTM

PR #1483, opened 2017-06-09
Update criu to 3.1

crosbymichael:
This updates criu to the latest release to help resolve some of the
issues we are seeing on the CI.

Signed-off-by: Michael Crosby <crosbymichael@gmail.com>

PR #1489, opened 2017-06-14
libcontainer/container_linux: Consider process state (running, zombie, etc.) in runType

wking:
As @avagin [pointed][1] [out][2], we're currently failing to distinguish between running (stopped, disk-sleeping, etc.) and dead/zombie processes when determining the container state.  This PR rerolls our runType approach to make that distinction.

Now that we need two values from `/proc/[pid]/stat`, I've rerolled `libcontainer/system/proc` to use a `Stat_t` object (like the [`sys/x/unix` `Stat_t` API][3]).  Then we can get both the start-time and process state from the same filesystem read.

Additional details in the commit messages for the curious.

[1]: https://github.com/opencontainers/runc/pull/1483#issuecomment-307527789
[2]: https://github.com/opencontainers/runc/pull/1483#issuecomment-307529294
[3]: https://godoc.org/golang.org/x/sys/unix#Stat_t


wking:
Hooray, [Travis is happy][1] :).  Can someone more familiar with the checkpoint tooling comment on whether [the appropriate post-checkpoint state is `Stopped`][2] (with a zombie process?).

[1]: https://travis-ci.org/opencontainers/runc/jobs/243359123
[2]: https://github.com/opencontainers/runc/pull/1489/files#diff-4dcc5378706c75d50b2b7c827d8546e8L183

avagin:
@wking Thank you for the patch. I looked at it without details and it looks good for me. I'm going to review it carefully a bit later.
> Can someone more familiar with the checkpoint tooling comment on whether the appropriate post-checkpoint state is Stopped (with a zombie process?).

I am not sure that I understand what do you mean here. Could you elaborate?

wking:
On Tue, Jun 20, 2017 at 11:17:25AM -0700, Andrew Vagin wrote:
> > Can someone more familiar with the checkpoint tooling comment on
> > whether the appropriate post-checkpoint state is Stopped (with a
> > zombie process?).
>
> I am not sure that I understand what do you mean here. Could you
> elaborate?

The change I make to checkpoint_test.go [1] is after a container has
been checkpointed (via container.Checkpoint [2]).  When the state is
checked [3], the container process was a zombie, which, with this PR,
means the status was ‘Stopped’ (before this PR, a zombie container
process was reported as ‘Running’ [4]).  Interestingly, there's an
earlier call to container.Checkpoint with different options in that
test [5], and in that case the container process is ‘Running’ both
before and after this PR [6].  I don't understand container.Checkpoint
well enough to know why the call in [2] (but not the call in [5])
seems to be zombi-fying the container process (the difference may be
PreDump being set in the earlier call [7]).  And since I don't
understand why the container process is (at least sometimes) being
zombi-fied, I'm not sure how reliable my checkpoint_test.go change [1]
is.

[1]: https://github.com/opencontainers/runc/pull/1489/files#diff-4dcc5378706c75d50b2b7c827d8546e8L183
[2]: https://github.com/opencontainers/runc/blob/035b57895dcd7284a61aacfb9dbfd545a1a425ba/libcontainer/integration/checkpoint_test.go#L173
[3]: https://github.com/opencontainers/runc/blob/035b57895dcd7284a61aacfb9dbfd545a1a425ba/libcontainer/integration/checkpoint_test.go#L178
[4]: https://github.com/opencontainers/runc/blob/035b57895dcd7284a61aacfb9dbfd545a1a425ba/libcontainer/integration/checkpoint_test.go#L183
[5]: https://github.com/opencontainers/runc/blob/035b57895dcd7284a61aacfb9dbfd545a1a425ba/libcontainer/integration/checkpoint_test.go#L145
[6]: https://github.com/opencontainers/runc/blob/035b57895dcd7284a61aacfb9dbfd545a1a425ba/libcontainer/integration/checkpoint_test.go#L155
[7]: https://github.com/opencontainers/runc/blob/035b57895dcd7284a61aacfb9dbfd545a1a425ba/libcontainer/integration/checkpoint_test.go#L141

avagin:
My reproducer for the problem from  is still working. You need to stop the init process, I use gdb for that, and then start tests.

```
[root@fc24 runc]# gdb -p 1
(gdb) 
[1]+  Stopped                 gdb -p 1
[root@fc24 runc]# bats tests/integration/checkpoint.bats 
 ✗ checkpoint and restore
   (in test file tests/integration/checkpoint.bats, line 35)
     `[ "$status" -ne 0 ]' failed
   runc list (status=0):
   runc spec (status=0):
   
   runc run -d --console-socket /tmp/console.sock test_busybox (status=0):
   
   runc state test_busybox (status=0):
   {
     "ociVersion": "1.0.0-rc5-dev",
     "id": "test_busybox",
     "pid": 23063,
     "status": "running",
     "bundle": "/tmp/busyboxtest",
     "rootfs": "/tmp/busyboxtest/rootfs",
     "created": "2017-06-20T22:50:35.560498491Z",
     "owner": ""
   }
   runc --criu /usr/sbin/criu checkpoint test_busybox (status=0):
   time="2017-06-21T01:50:35+03:00" level=error msg="container is not destroyed" 
   runc state test_busybox (status=0):
   {
     "ociVersion": "1.0.0-rc5-dev",
     "id": "test_busybox",
     "pid": 23063,
     "status": "running",
     "bundle": "/tmp/busyboxtest",
     "rootfs": "/tmp/busyboxtest/rootfs",
     "created": "2017-06-20T22:50:35.560498491Z",
     "owner": ""
   }
   runc list (status=0):
   ID             PID         STATUS      BUNDLE             CREATED                          OWNER
   test_busybox   23063       running     /tmp/busyboxtest   2017-06-20T22:50:35.560498491Z   root
   runc kill test_busybox KILL (status=0):
   
   Command "eval __runc state 'test_busybox' | grep -q 'stopped'" failed 10 times. Output: 
   runc delete test_busybox (status=1):
   time="2017-06-21T01:50:46+03:00" level=error msg="cannot delete container test_busybox that is not stopped: running\n\n" 
   cannot delete container test_busybox that is not stopped: running
```

avagin:
Oops, I forgot to compile runc. Actually everything works as expected with this patch.

avagin:
@wking  you changes in checkpoint_test.go looks correct.

> Interestingly, there's an earlier call to container.Checkpoint with different options in that
test [5], and in that case the container process is ‘Running’ both before and after this PR [6]. 

container.Checkpoint(preDumpOpts) doesn't stop a container, this call enables memory tracker and dumps memory for processes without stopping them. Then we call container.Checkpoint() in a second time and this call stops processes and dumps them. The big part of memory is already dumped, so we have to dump only changed memory. This optimization allows us to reduce a downtime of a container, when it is migrated from one host to another host: https://criu.org/Iterative_migration

avagin:
LGTM

wking:
I just pushed 407e33a → 2bea4c8 fixing a `%ds` → `%d` typo.  Sorry for the churn :/.  @avagin, can you check the update and re-LGTM?

avagin:
LGTM

libcontainer/container_linux.go
```diff
@@ -1365,40 +1365,17 @@ func (c *linuxContainer) refreshState() error {
...
-		return Stopped, newSystemErrorWithCausef(err, "sending signal 0 to pid %d", pid)
+	stat, err := system.Stat(pid)
+	if err != nil {
+		return Stopped, nil
 	}
-	// check if the process is still the original init process.
-	exist, err := c.doesInitProcessExist(pid)
-	if !exist || err != nil {
-		return Stopped, err
+	if stat.StartTime != c.initProcessStartTime || stat.State == system.Zombie || stat.State == system.Dead {
```
TomSweeneyRedHat: Should there be a test with the state set as a Zombie or Dead to make sure that it gets picked off?  Apologies if I missed it, and not a hold up for this PR, suggestion for later if missing.

wking: > Should there be a test with the state set as a Zombie or Dead to make sure that it gets picked off?

I'm not sure what you mean.  Can you give more detail on the case you feel is not covered?

wking: > Should there be a test with the state set as a Zombie or Dead to make sure that it gets picked off?

Ah, I think you mean a unit test where we run this on a dead/zombie process.  The `checkpoint_test.go` change is doing that already, although there's too much going on there for it to be a *unit* test.  If you have another way to manufacture a zombie process for testing, I'm happy to add a more focused test.

TomSweeneyRedHat: I was indeed thinking of a unit test, trying to ensure more code coverage.  I unfortunately don't have a good way to emulate a zombie/dead process.  I was just thinking of faking it out in a test to change the state of the process.  I didn't see that happening for Dead/Zombie in the in the checkpoint_tests.go in particular, only that the process wasn't Stopped.  Anywho, no need to address now, especially with this PR, just a point to mull.

TomSweeneyRedHat:
LGTM, shoulda done this when I commented earlier today.

crosbymichael:
LGTM

PR #1491, opened 2017-06-21
Use Eventfd() from golang.org/x/sys/unix

tklauser:
Update golang.org/x/sys/unix to get Eventfd() and the corresponding EFD_* flags introduced in golang/sys@0b25a408) and switch libcontainer to use Eventfd() instead of manually reimplementing it.  Also use the correct corresponding unix.EFD_CLOEXEC flag instead of unix.FD_CLOEXEC (which can have a different value on some architectures and thus might lead to unexpected behavior).


crosbymichael:
LGTM

crosbymichael:
@tklauser Thanks for working on these.  It really helps.

mrunalp:
LGTM

cyphar:
I second @crosbymichael -- thanks a lot for these @tklauser!

PR #1375, opened 2017-03-20
Use uint64 for resources to keep consistency with runtime-spec

hqhq:
Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>

PR #1492, opened 2017-06-23
Revert memory fields to signed integers

### LINKED ISSUE #1421: setting Resources.MemorySwap unlimited after type change is messy

tiborvass:
Reverts commit 8430cc4f488f1925df01e517358c2a4cce8de176

Conflicted with commits 4f9cb13b644285c39c077feb20a254f3074d2d77
and 1e7e276aff03893ccb243311e2cbeb9edeae8198

The kernel specifically accepts -1 in its parser, using
page_counter_memparse(args, "-1",...) but we no longer allow this since
this change, so it no longer matches the kernel behaviour, even if it is
a "sort of" uint64 it is a special one that also accepts -1.

In addition the maximum meaningful value is 2^63 due to the kernel/user
memory split.

In addition some older kernels (RHEL 7.2 at least) will set the limits
to 0 if they are larger than 2^63, and then immediately crash, so
setting them to 2^64-1 makes them crash, while the previous -1 worked
correctly.

Signed-off-by: Tibor Vass <tibor@docker.com>

Fixes #1421
Closes #1422
Reverts #1375

Related to https://github.com/opencontainers/runtime-spec/commit/ec9449187bc752ef728727bd479e547d66fade44

In fact, that spec change seems incorrect for the reasons detailed above. I think we should also revert the spec.

cc @hqhq @crosbymichael @tonistiigi @justincormack

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -332,8 +332,8 @@ func removePath(p string, err error) error {
 	return nil
 }
 
-func CheckCpushares(path string, c uint64) error {
-	var cpuShares uint64
+func CheckCpushares(path string, c int64) error {
+	var cpuShares int64
 
 	if c == 0 {
 		return nil
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -55,7 +55,7 @@ func (s *CpuGroup) ApplyDir(path string, cgroup *configs.Cgroup, pid int) error 
 
 func (s *CpuGroup) SetRtSched(path string, cgroup *configs.Cgroup) error {
 	if cgroup.Resources.CpuRtPeriod != 0 {
-		if err := writeFile(path, "cpu.rt_period_us", strconv.FormatUint(cgroup.Resources.CpuRtPeriod, 10)); err != nil {
+		if err := writeFile(path, "cpu.rt_period_us", strconv.FormatInt(cgroup.Resources.CpuRtPeriod, 10)); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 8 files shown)


hqhq:
@tiborvass Can https://github.com/opencontainers/runc/pull/1494 fix your problem?

crosbymichael:
I think we should go with this one.  Having the handle and mess with some MemoryUnlimited const to work around this initial change makes things messy and not very straight forward.  I don't see a benefit of syncing up these two as the cgroups api is string based and accepts -1

justincormack:
Spec PR opened https://github.com/opencontainers/runtime-spec/pull/876

hqhq:
> In addition the maximum meaningful value is 2^63 due to the kernel/user
memory split.

I don't quite understand this, what do you mean by "due to the kernel/user memory split", if you mean address space split, we only have 47 bits for user space on x86_64 (different arch have different splits, but they all have only parts of the whole 64 bits), and the largest physical memory we can support is 64TB (see https://github.com/torvalds/linux/blob/v4.12-rc6/Documentation/x86/x86_64/mm.txt the 'direct mapping' slot), though we have 5 level page table which can support 56 bits user space, there's no real hardware for that yet AFAIK.

Disregarding the limited physical memory we can support, the address space we are talking about is for one process, but the memory limit we set here is for container, which can have lots of processes.

Anyway, the point is, it's not about real world cases, it's about technically right or wrong. If Linux kernel can accept that value, we should be able to set it, that's why http://marc.info/?l=linux-mm&m=137541397220314&w=2 was merged in the first place, not because people want to use that large value. The major reason I want it to be uint64, is that if we can do
```
# echo 18446744073709541615 > memory.limit_in_bytes
```
On Linux, why we can't set that value with OCI runtime?

Since the spec change is merged, I won't be the blocker here, I can live with int64 since the problem is  not in the real world for now, and it makes runc code a bit more complex, we can change that in OCI once it is.

dqminh:
This can be closed now wth #1495

PR #1493, opened 2017-06-23
Do not cast -1 to uint64 in EnableKernelMemoryAccounting

tiborvass:
This fixes a runc bug that is very annoying on kernels such as RHEL 7.2
that have the following bug: when setting memory.kmem.limit_in_bytes to
a bigger value than the default max value (bigger than 2^63-1), the
buggy kernel sets it to 0, instead of setting it to the max value. This
results in the process being unable to run with a 0 byte limit on kmem.

That kernel bug exposes a runc bug introduced in 8430cc4, that casted an
int64(-1) to uint64 in EnableKernelMemoryAccounting, which is called
every time the user specifies any memory-related cgroup setting as
determined by the memoryAssigned function.

This patch is the smallest fix for this clearly unwanted behavior.
Reverting 8430cc4, although desirable because int64 values are casted to
uint64 in many other places introducing similar bugs, it is out of scope
for this patch since it would require a spec change.

Signed-off-by: Tibor Vass <tibor@docker.com>

Related #1492 #1375
cc @crosbymichael @justincormack @tonistiigi @hqhq

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -73,14 +73,14 @@ func EnableKernelMemoryAccounting(path string) error {
 	// until a limit is set on the cgroup and limit cannot be set once the
 	// cgroup has children, or if there are already tasks in the cgroup.
 	for _, i := range []int64{1, -1} {
-		if err := setKernelMemory(path, uint64(i)); err != nil {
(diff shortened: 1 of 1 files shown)


crosbymichael:
Lets revert the entire commit for now in your other PR.

PR #1494, opened 2017-06-23
Fix unlimited memory issue on old kernel

### LINKED ISSUE #1421: setting Resources.MemorySwap unlimited after type change is messy

hqhq:
Fixes #1421

On old kernels (before 3.11), max memory limit is
LLONG_MAX, which is 2^63-1 on most platforms, so we
should not use uint64(-1) for unlimited memory.

Keep using `-1` for cgroup API when people want to
set memory unlimited to fix this.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -5,7 +5,6 @@ package fs
 import (
 	"bufio"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -88,7 +87,8 @@ func setKernelMemory(path string, kernelMemoryLimit uint64) error {
 		// kernel memory is not enabled on the system so we should do nothing
 		return nil
 	}
-	if err := ioutil.WriteFile(filepath.Join(path, cgroupKernelMemoryLimit), []byte(strconv.FormatUint(kernelMemoryLimit, 10)), 0700); err != nil {
+
+	if err := writeMemoryEntries(path, cgroupKernelMemoryLimit, kernelMemoryLimit); err != nil {
 		// Check if the error number returned by the syscall is "EBUSY"
 		// The EBUSY signal is returned on attempts to write to the
 		// memory.kmem.limit_in_bytes file if the cgroup has children or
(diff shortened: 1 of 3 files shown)


justincormack:
This should fix the immediate problem we found during testing, so we should merge once confirmed, as the current code does not work at all on older kernels (you cannot set a memory limit at all).

I still think that having the memory limits as `uint64` is not the best solution, as `-1` is an allowed value, and the maximum allowed value is `2^63`, due to the kernel/userspace split which means that no value larger than this is valid, so an `int64` is a much better fit to the externally presented interface from the kernel. There is nowhere that the internals of the value being a `uint64` are exposed, so there is no `syscall` API to these values, only a string API.

dqminh:
> I still think that having the memory limits as uint64 is not the best solution, as -1 is an allowed value, and the maximum allowed value is 2^63, due to the kernel/userspace split which means that no value larger than this is valid, so an int64 is a much better fit to the externally presented interface from the kernel. 

Yah its a bit weird right now because essentially we are doing double-cast, where we translate user input from -1 into uint64, than cast back to -1 again.

The argument before was that we should stick to the internal kernel's value presentation (i.e. uint64), but maybe we should actually just conform to the kernel external interface for these settings (i.e int64)

crosbymichael:
@dqminh ya that sounds like a good idea.  But does this also need a spec change?

crosbymichael:
I think we should change these back to -1.  This PR fixes it but it has to add a lot of code to work around the issue.  We have been running with -1 for years now and it has worked correctly.  Lets either revert the kmem accounting or the entire thing and not add more code around this as cgroups accepts -1

justincormack:
I am happy to do a PR to change the spec back to `int64`, was planning to do one anyway, but I think it helps to discuss here too, and `runc` is currently badly broken.

crosbymichael:
@justincormack tibor already opened a PR for this: #1492

justincormack:
@crosbymichael I meant a PR to change the runtime spec, not `runc`. Will do one now.

justincormack:
Spec PR opened https://github.com/opencontainers/runtime-spec/pull/876

dqminh:
This can be closed now with #1495

PR #1495, opened 2017-06-24
Update spec to master, switch to int64 for memory limits

justincormack:
Update memory specs to use int64 not uint64

replace #1492 #1494
fix #1422

Since https://github.com/opencontainers/runtime-spec/pull/876 the memory
specifications are now `int64`, as that better matches the visible interface where
`-1` is a valid value. Otherwise finding the correct value was difficult as it
was kernel dependent.

Also in the update of the spec, `Platform` was removed, so remove some code.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -73,14 +73,14 @@ func EnableKernelMemoryAccounting(path string) error {
 	// until a limit is set on the cgroup and limit cannot be set once the
 	// cgroup has children, or if there are already tasks in the cgroup.
 	for _, i := range []int64{1, -1} {
-		if err := setKernelMemory(path, uint64(i)); err != nil {
+		if err := setKernelMemory(path, i); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func setKernelMemory(path string, kernelMemoryLimit uint64) error {
+func setKernelMemory(path string, kernelMemoryLimit int64) error {
 	if path == "" {
 		return fmt.Errorf("no such directory for %s", cgroupKernelMemoryLimit)
 	}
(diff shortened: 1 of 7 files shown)


justincormack:
cc @crosbymichael

libcontainer/cgroups/fs/memory.go
```diff
@@ -128,29 +126,29 @@ func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
 		// When update memory limit, we should adapt the write sequence
 		// for memory and swap memory, so it won't fail because the new
 		// value and the old value don't fit kernel's validation.
-		if cgroup.Resources.MemorySwap == uint64(ulimited) || memoryUsage.Limit < cgroup.Resources.MemorySwap {
-			if err := writeFile(path, cgroupMemorySwapLimit, strconv.FormatUint(cgroup.Resources.MemorySwap, 10)); err != nil {
+		if cgroup.Resources.MemorySwap == -1 || memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap) {
```
tiborvass: `memoryUsage.Limit < uint64(cgroup.Resources.MemorySwap)` is weird

tiborvass: I think I understand what's going on: since memoryUsage.Limit is a uint64 (since read from fs), we have to choose between casting that to int64, or casting an int64 that's different from -1, to uint64. Would be great to add a comment about the casting.

Which brings up the question: have we ensured that the int64s are never `< -1` ?

hqhq: `-1` should be the only valid negative value, but I don't think we should ensure that in runc, cgroup API will return "invalid argument" in this case anyway.

crosbymichael:
@justincormack can you rebase this to pick up the CI fix?

justincormack:
rebased

dqminh:
LGTM

crosbymichael:
LGTM

PR #1496, opened 2017-06-25
Fix checkpoint/restore tests with newer kernel

dqminh:
Story for the stack gap problem: https://github.com/xemul/criu/issues/322

This upgrades criu to a testing commit to fix our test failures. We should upgrade again when criu does a release but this commit seems to fix the tests for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ RUN cd /tmp \
     && rm -rf /tmp/bats
 
 # install criu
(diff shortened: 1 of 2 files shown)


dqminh:
cc @crosbymichael @avagin 

This also probably should replace #1483

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1497, opened 2017-06-27
compilation error in windows

SagarPuneria:
move the code from lookup.go to lookup_unix.go

### DIFF
--- a/libcontainer/user/lookup.go
+++ b/libcontainer/user/lookup.go
@@ -2,8 +2,6 @@ package user
 
 import (
 	"errors"
-
-	"golang.org/x/sys/unix"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


crosbymichael:
@SagarPuneria  thanks for the PR but I'm not sure we care about having runc compile on windows.  It is highly unlikely that there would be a version of runc that will work with windows containers.  It should be a totally different implementation, maybe a `runw` :S.  Bad name but you get the point.

I think the community as a whole have taking the route of separate OCI runtimes that service a single platform to keep their code bases small and focused.

sunjayBhatia:
@SagarPuneria @crosbymichael we have started implementing an OCI runtime for Windows containers with [`winc`](https://github.com/cloudfoundry-incubator/winc)

It is a work in progress as we work towards implementing [Garden](https://github.com/cloudfoundry/garden) support on Windows Server 2016 and is not a pure OCI runtime as of yet, but the goal is eventual parity (where it makes sense) with `runc`

PR #1502, opened 2017-06-29
runc only works on Linux so remove putative Solaris and unsupported main

justincormack:
Solaris runc equivalent is not open source. The unsupported option does
not build as other code is Linux specific.


cyphar:
LGTM. I'm really hoping that the illumos community steps up and creates a proper free software OCI implementation. :cry:

justincormack:
@cyphar indeed, so am I.

crosbymichael:
LGTM

PR #1504, opened 2017-07-03
Use Prctl() and ioctl wrapper functions from x/sys/unix

tklauser:
Update golang.org/x/sys/unix once again to get some more newly added wrapper functions and switch libcontainer to use them instead of manually reimplemented local wrappers.

PR #1505, opened 2017-07-03
*: fix shfmt

cyphar:
Recent changes in upstream shfmt have started causing our scripts to no
longer be "correctly formatted". Fix up with `shfmt -w`.

### DIFF
--- a/script/check-config.sh
+++ b/script/check-config.sh
@@ -202,7 +202,7 @@ flags=(
 	NF_NAT_IPV4 IP_NF_FILTER IP_NF_TARGET_MASQUERADE
 	NETFILTER_XT_MATCH_{ADDRTYPE,CONNTRACK}
 	NF_NAT NF_NAT_NEEDED
-
+	
 	# required for bind-mounting /dev/mqueue into containers
 	POSIX_MQUEUE
 )


cyphar:
This fixes the failures in #1504.

crosbymichael:
LGTM

should we pin the commit for shfmt so contributors don't run into this type of thing on their PRs when upstream changes?

cyphar:
Most likely, but we'd have to replace the `go get -u` with the good ol' fashioned `git pull && make && make install` stuff.

dqminh:
LGTM

mvdan:
Darn, this is a regression :) And yes, you should pin versions. You can use binaries from https://github.com/mvdan/sh/releases too, if you wish.

mvdan:
Fixed here: https://github.com/mvdan/sh/commit/4c56b4cfd0a28bd05e9324be1f43ec4cc2a076f5

dqminh:
@mvdan i think pinning 1.3.1 is not working for us now since we made changes that is incompatible with it. Do you want to do a release then we will pin it here ?

mvdan:
I plan on doing 2.0 very soon - probably in the next week - but I can't do it right now. Want to make sure I'm not missing anything related to breaking changes. And I haven't fuzzed master thoroughly yet.

I don't plan on doing any more refactors like the one that introduced this regression however, so you should be fine.

cyphar:
@mvdan I was wondering if this was a regression, because it's odd to require trailing whitespace. I'll re-send an un-fix for this. :wink:

Issue #1443, opened 2017-05-10
Should libcontainer automatically reap runc:[1:CHILD]?


LittleLightLittleFire:
This would mean there is no need to install a subreaper.

This avoids boobytrapping anyone trying to use os.Exec and would mean downstream projects like containerd no longer have to have a [custom reaper](https://github.com/containerd/containerd/blob/master/reaper/reaper.go) to prevent race conditions with `wait(2)`.

Something like this perhaps?
```
diff --git a/libcontainer/init_linux.go b/libcontainer/init_linux.go
index 99cc02c..282d0c7 100644
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -29,7 +29,8 @@ const (
 )
 
 type pid struct {
-	Pid int `json:"pid"`
+	Pid  int `json:"pid"`
+	Pid2 int `json:"pid2"`
 }
 
 // network is an internal struct used to setup container networks.
diff --git a/libcontainer/nsenter/nsexec.c b/libcontainer/nsenter/nsexec.c
index 0ad6883..01ef2a8 100644
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -543,7 +543,7 @@ void nsexec(void)
 	 */
 	case JUMP_PARENT: {
 			int len;
-			pid_t child;
+			pid_t child, old_child = -1;
 			char buf[JSON_MAX];
 			bool ready = false;
 
@@ -607,18 +607,18 @@ void nsexec(void)
 					}
 					break;
 				case SYNC_RECVPID_PLS: {
-						pid_t old = child;
+						old_child = child;
 
 						/* Get the init_func pid. */
 						if (read(syncfd, &child, sizeof(child)) != sizeof(child)) {
-							kill(old, SIGKILL);
+							kill(old_child, SIGKILL);
 							bail("failed to sync with child: read(childpid)");
 						}
 
 						/* Send ACK. */
 						s = SYNC_RECVPID_ACK;
 						if (write(syncfd, &s, sizeof(s)) != sizeof(s)) {
-							kill(old, SIGKILL);
+							kill(old_child, SIGKILL);
 							kill(child, SIGKILL);
 							bail("failed to sync with child: write(SYNC_RECVPID_ACK)");
 						}
@@ -667,7 +667,7 @@ void nsexec(void)
 			}
 
 			/* Send the init_func pid back to our parent. */
[...]

PR #1506, opened 2017-07-05
Pass back the pid of runc:[1:CHILD] so we can wait on it

LittleLightLittleFire:
Fixes https://github.com/opencontainers/runc/issues/1443

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -141,6 +141,16 @@ func (p *setnsProcess) execSetns() error {
 		p.cmd.Wait()
 		return newSystemErrorWithCause(err, "reading pid from init pipe")
 	}
+
+	// Clean up the zombie parent process
+	firstChildProcess, err := os.FindProcess(pid.PidFirstChild)
+	if err != nil {
+		return err
+	}
+
+	// Ignore the error in case the child has already been reaped for any reason
+	_, _ = firstChildProcess.Wait()
+
 	process, err := os.FindProcess(pid.Pid)
 	if err != nil {
 		return err
@@ -224,6 +234,16 @@ func (p *initProcess) execSetns() error {
 		p.cmd.Wait()
 		return err
 	}
+
+	// Clean up the zombie parent process
+	firstChildProcess, err := os.FindProcess(pid.PidFirstChild)
+	if err != nil {
+		return err
+	}
+
+	// Ignore the error in case the child has already been reaped for any reason
+	_, _ = firstChildProcess.Wait()
+
 	process, err := os.FindProcess(pid.Pid)
 	if err != nil {
 		return err
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -31,7 +31,8 @@ const (
 )
 
 type pid struct {
-	Pid int `json:"pid"`
+	Pid           int `json:"pid"`
+	PidFirstChild int `json:"pid_first"`
 }
 
 // network is an internal struct used to setup container networks.
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -542,7 +542,7 @@ void nsexec(void)
 	 */
 	case JUMP_PARENT: {
 			int len;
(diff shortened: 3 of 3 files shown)


libcontainer/process_linux.go
```diff
@@ -141,6 +141,16 @@ func (p *setnsProcess) execSetns() error {
 		p.cmd.Wait()
 		return newSystemErrorWithCause(err, "reading pid from init pipe")
 	}
+
+	// Clean up the zombie parent process
+	firstChildProcess, err := os.FindProcess(pid.PidFirstChild)
+	if err != nil {
```
cyphar: I'm not convinced that propagating the error here makes sense (especially since you might get `ECHILD` -- I haven't checked the Go source yet). Maybe we should do something like this

```go
if firstChildProcess, err := os.FindProcess(pid.PidFirstChild); err == nil {
	if _, err := firstChildProcess.Wait(); err != nil {
		return err
	}
}
```

Ditto for the `pid.Pid`. @mrunalp @dqminh @hqhq WDYT?

cyphar: Also are we sure it's not possible to get `PidFirstChild = -1`?

LittleLightLittleFire: os.FindProcess can't fail on Unix systems. https://godoc.org/os#FindProcess

> On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists.

LittleLightLittleFire: > Also are we sure it's not possible to get `PidFirstChild = -1?`

Pretty sure it is impossible since it'd mean the grandchild'd pid has been received. This implies that the child has already been created.

I can just change the nsenter function to differentiate between `child` and `grandchild`.

LittleLightLittleFire: Anyways, let me know what changes you want to see.

hqhq: What would `process.Wait` return if the process doesn't exist?

cyphar: In which case the `Wait` is what will fail if the process doesn't exist -- so maybe some thing more like

```go
if _, err := firstChildProcess.Wait(); err != nil {
	if err != unix.ECHILD && err != unix.ESRCH {
		return err
	}
}
```

Maybe?

LittleLightLittleFire: Shouldn't firstChildProcess (runc:[1:CHILD]) always exist? (thus `wait` can't fail)?
If it doesn't exist we won't be able to get the grandchild's pid.

libcontainer/process_linux.go
```diff
@@ -141,6 +141,16 @@ func (p *setnsProcess) execSetns() error {
...
+		return err
+	}
+	if _, err := firstChildProcess.Wait(); err != nil {
```
cyphar: As discussed IRL, basically this should either just ignore the error:

```
_, _ = firstChildProcess.Wait()
```

Or explicitly special-case `unix.ESRCH` or `unix.ECHILD`, but I'm not convinced that makes much more sense than just ignoring the error (if you `grep` for `Wait` you'll find it).

libcontainer/process_linux.go
```diff
@@ -224,6 +234,16 @@ func (p *initProcess) execSetns() error {
 		p.cmd.Wait()
 		return err
 	}
+
+	// Clean up the zombie parent process
+	firstChildProcess, err := os.FindProcess(pid.PidFirstChild)
+	if err != nil {
+		return err
+	}
+	if _, err := firstChildProcess.Wait(); err != nil {
```
cyphar: Same comment here.

LittleLightLittleFire:
Pull request updated

cyphar:
LGTM.

crosbymichael:
LGTM

c4milo:
does this mean we can run runc --detach using systemd with Type=notify, without leaving zombie processes around?

cyphar:
@c4milo Probably, I would have to test it to find out, but this was intended to solve the more general intermediate zombie issue.

PR #1511, opened 2017-07-06
update to github.com/sirupsen/logrus v1.0.0

mcandre:
Signed-off-by: Andrew Pennebaker <apennebaker@datapipe.com>

Fix some dependencies by updating to latest logrus, now hosted at github.com/sirupsen/logrus.


mcandre:
Odd, the unit tests are passing in CI, but the Makefile appears to fail the local task anyway.

cyphar:
They're not passing. [Here's the highlighted failure](https://travis-ci.org/opencontainers/runc/jobs/250959798#L3372).

dqminh:
yah, this is because logrus switch the default log output to stderr, while the execintest code only set the buffer to stdout. Change https://github.com/opencontainers/runc/blob/master/libcontainer/integration/execin_test.go#L237 to Stderr should make the test pass.

crosbymichael:
@mcandre you can revert your last commit to fix the shell changes.  If you rebase on master the validation will pass.  Sorry for the trouble.

dqminh:
@mcandre i can carry this change if you need help for it. Just need a few small fixes i think.

dqminh:
Closed in favor of #1526

PR #1514, opened 2017-07-10
libcontainer/container.go: Fix small marshaling error

baude:
When running runc kill <container> KILL, an error was always being throw like:

   json: cannot unmarshal string into Go struct field State.init_process_start of type uint64

Converting the uint64 to string for json output resolves the issue.

### DIFF
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -54,7 +54,7 @@ type BaseState struct {
 	InitProcessPid int `json:"init_process_pid"`
 
 	// InitProcessStartTime is the init process start time in clock cycles since boot time.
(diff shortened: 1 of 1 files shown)


jwhonce:
@mheon please take a look.  Thanks.

mrunalp:
You are probably dealing with configs created by different versions of runc.

cyphar:
Yeah, that's what this looks like to me as well. Personally I'm not ready to provide any form of stability guarantees of our current state format (especially since we want to switch to OCI-only internal information in the future) so this change doesn't make much sense to me.

Can you reproduce this issue without mixing runc versions?

mrunalp:
Closing as this looks like mismatched versions of runc and config.json.

PR #1516, opened 2017-07-11
list: fix various problems with owner field

### LINKED ISSUE #1515: runc list missing username (!?) incorrectly (!?) casts uid (int) to string
### Summary
In the following the `owner` value is rendered as `𰨅` (U+30A05 == 199173).  This only happens on one (Ubuntu) workstation and does not arise on a second (Ubuntu) machine. I raise it because, while there appears to be several issues, at least one appears to be manifest by `runc`.

### Issues
On the problematic instance, the `owner` value is rendered by `runc list` as `𰨅`:
- `𰨅` corresponds to `U+30A05` == `199173`
- 199173 is the uid of the user account logged-in to the instance
- however the `runc` commands are being run as `root` (sudo) not the user
- `runc` appears to fail in looking up the username of the 199173 user (not root?)
- this results in `runc` `owner.Name` string value of `"\U00030a05"`
- which appears to be then rendered by `runc list` as if it were a character (`𰨅`) not the string

### Observation
consistently across `runc list --format`

```
sudo ./runc list --format table

ID          PID         STATUS      BUNDLE                                                                                                              CREATED                          OWNER
daz01       23235       running     [[OBFUSCASTED]]/go/src/github.com/opencontainers/runc/mycontainer   2017-07-10T22:09:45.474632798Z   𰨅
```
```
sudo ./runc list --format json

[{"ociVersion":"1.0.0-rc5-dev","id":"container01","pid":23235,"status":"running","bundle":"[[OBFUSCATED]]/go/src/github.com/opencontainers/runc/mycontainer","rootfs":"[[OBFUSCATED]]/go/src/github.com/opencontainers/runc/mycontainer/rootfs","created":"2017-07-10T22:09:45.474632798Z","owner":"𰨅"}]
```
### Exploration

I don't understand how this is arising but this uid corresponds to the logged in user's uid even though the `runc` commands are being run as `root`:

`os.Geteuid()` == `unix.Getuid()` == `199173`

Subsequently, the `owner.Name` is assigned a Unicoded encoded string value of `"\U00030a05"` (==199173) and then this is incorrectly interpreted as a character and rendered as `𰨅`

This appears to arise during `lookup.go` attempt to determine username from `/etc/passwd`

list.go:136
```
github.com/opencontainers/runc/libcontainer/user.User {Name: "", Pass: "", Uid: 0, Gid: 0, Gecos: "", Home: "", Shell: ""}
```

`user.LookupUid` is returning an err (`delve` isn't surfacing this to me) and so...

list.go:138
```
github.com/opencontainers/runc/libcontainer/user.User {Name: "\U00030a05", Pass: "", Uid: 0, Gid: 0, Gecos: "", Home: "", Shell: ""}
```

cyphar:
If a container is owned by a UID that is not listed in /etc/passwd,
previously we would cast the UID to a string which contained a character
with the unicode value of the UID. This is clearly wrong, switch to
using fmt.Sprintf as intended.

In addition, notate unknown users with a leading '#' in the column. This
is necessary to ensure that a user is not under the impression that the
UID is the same as a numeric username.

It appears as though these semantics were not fully thought out when
implementing them for rootless containers. It is not necessary (and
could be potentially dangerous) to set the owner of /run/ctr/$id to be
the root inside the container (if user namespaces are being used).

Instead, just use the e{g,u}id of runc to determine the owner.

Fixes #1515 

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -162,14 +162,6 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := l.Validator.Validate(config); err != nil {
 		return nil, newGenericError(err, ConfigInvalid)
 	}
-	uid, err := config.HostRootUID()
-	if err != nil {
-		return nil, newGenericError(err, SystemError)
-	}
-	gid, err := config.HostRootGID()
-	if err != nil {
-		return nil, newGenericError(err, SystemError)
-	}
 	containerRoot := filepath.Join(l.Root, id)
 	if _, err := os.Stat(containerRoot); err == nil {
 		return nil, newGenericError(fmt.Errorf("container with id exists: %v", id), IdInUse)
@@ -179,7 +171,7 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := os.MkdirAll(containerRoot, 0711); err != nil {
 		return nil, newGenericError(err, SystemError)
 	}
-	if err := os.Chown(containerRoot, uid, gid); err != nil {
+	if err := os.Chown(containerRoot, unix.Geteuid(), unix.Getegid()); err != nil {
 		return nil, newGenericError(err, SystemError)
 	}
 	if config.Rootless {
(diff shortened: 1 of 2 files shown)


DazWilkin:
Nice! Thank you.

crosbymichael:
LGTM

hqhq:
LGTM

PR #1517, opened 2017-07-12
Update runtime-spec to rc6

crosbymichael:


crosbymichael:
I'm going to wait for the oom change in the spec before updating runc

PR #1519, opened 2017-07-13
libcontainer: use additional functions and constants from x/sys/unix

tklauser:
Use ParseSocketControlMessage and ParseUnixRights from
golang.org/x/sys/unix instead of their syscall equivalent.

Use PR_SET_NO_NEW_PRIVS defined in golang.org/x/sys/unix instead of
manually defining it.


hqhq:
LGTM

dqminh:
LGTM

PR #1522, opened 2017-07-14
Fix failures during make localintegration

leitwolf7:
make localintegration fails with Ubuntu 17.04 on x86.

Problem is a glitch in helpers.bash that fails when criu is not available,
and 2 string comparision issue in exec.bats.

Signed-off-by: leitwolf7 <leitwolf@wolke7.net>


hqhq:
@leitwolf7 Can you base? It's partially fixed by another PR.

leitwolf7:
Rebased and included fix from an other PR.

leitwolf7:
Closing, will create a new clean PR.

PR #1455, opened 2017-05-19
tty: move IO of master pty to be done with epoll

dqminh:
Fix #1434
Fix #1446

Using edge-triggered event with epoll to make sure that we can still cope with
the other side temporarilt goes away.

PR #1526, opened 2017-07-19
Updated logrus to v1

stevenh:
Updated logrus to use v1 which includes a breaking name change Sirupsen -> sirupsen.

This includes a manual edit of the docker term package to also correct the name there too.


stevenh:
This replicates the changes done by https://github.com/opencontainers/runc/pull/1511 but also updates vendor libs too.

stevenh:
For reference this uses latest commit on 1-0-stable due to missing v1.0.2 tag.

dqminh:
fwiw, i added a new change to https://github.com/opencontainers/runc/pull/1455 that removes usages of docker/term so we dont have to do manual edit here.

stevenh:
Thanks @dqminh if you prod me when that's merged I'll rebase to remove that part 👍

dmcgowan:
@dqminh can we get this merged and rebase #1455? Either that or poke some of the maintainers to review the other PR.

dqminh:
@dmcgowan i did a poke on the other PR. I didn't like this too much because it has a manual edit on a dependency.

One other way is just to move the docker/pkg dependencies wholesale here (i.e. not import them anymore). we only use 2 small packages there. `containerd` does have some similar `mount` code too, probably we can also extract them out ?

dmcgowan:
@dqminh the circular nature of these imports is pretty obnoxious. I am trying to push forward this change containerd/moby/docker*, wondering if it will be better to vendor logrus twice until everything is updated, I really hate manual vendor edits as well.

dmcgowan:
Actually the manual edits can just be removed for now. A vendor will still import as `Sirupsen`, but the go compiler will correctly use `sirupsen`. The vndr tool only shows a warning but that is fine, that can be fixed in a follow up when the upstreams are fixed.

dqminh:
LGTM

i thought about this abit more and i think it would not be the end of the world if we merge this first before #1455. Since #1455 remove the term package anyway, and we dont update vendor that often (i dont see any pending PR that updates vendor ), i think merging this first has negilibile effect, we just have to be careful when updating vendor, but i hope we can merge #1455 soon :stuck_out_tongue:

crosbymichael:
LGTM

PR #1529, opened 2017-07-20
Support multiple users/groups mapped for the rootless case

giuseppe:
Take advantage of the newuidmap/newgidmap tools to allow multiple users/groups to be mapped into the new user namespace in the rootless case.

This is what I've tried here (only the relevant bits showed for config.json):

```json
	"args": [
	    "cat", "/proc/self/uid_map"
	],
...
        "uidMappings": [
            {
                "hostID": 1000,
                "containerID": 0,
                "size": 1
            },
            {
                "hostID": 110000,
                "containerID": 1,
                "size": 255
            }
        ],
        "gidMappings": [
            {
                "hostID": 1000,
                "containerID": 0,
                "size": 1
            }
        ],
```

```bash
$ whoami
gscrivano

$ grep ^gscrivano /etc/subuid
gscrivano:110000:65536

$ runc run foo
         0       1000          1
         1     110000        255
```

### DIFF
--- a/libcontainer/configs/validate/rootless.go
+++ b/libcontainer/configs/validate/rootless.go
@@ -36,37 +36,27 @@ func (v *ConfigValidator) rootless(config *configs.Config) error {
 	return nil
 }
 
-func rootlessMappings(config *configs.Config) error {
-	rootuid, err := config.HostRootUID()
-	if err != nil {
-		return fmt.Errorf("failed to get root uid from uidMappings: %v", err)
+func hasIDMapping(id int, mappings []configs.IDMap) bool {
+	for _, m := range mappings {
+		if id >= m.ContainerID && id < m.ContainerID+m.Size {
+			return true
+		}
 	}
+	return false
+}
+
+func rootlessMappings(config *configs.Config) error {
 	if euid := geteuid(); euid != 0 {
 		if !config.Namespaces.Contains(configs.NEWUSER) {
 			return fmt.Errorf("rootless containers require user namespaces")
 		}
-		if rootuid != euid {
-			return fmt.Errorf("rootless containers cannot map container root to a different host user")
-		}
 	}
 
-	rootgid, err := config.HostRootGID()
-	if err != nil {
-		return fmt.Errorf("failed to get root gid from gidMappings: %v", err)
+	if len(config.UidMappings) == 0 {
+		return fmt.Errorf("rootless containers requires at least one UID mapping")
 	}
-
-	// Similar to the above test, we need to make sure that we aren't trying to
-	// map to a group ID that we don't have the right to be.
-	if rootgid != getegid() {
-		return fmt.Errorf("rootless containers cannot map container root to a different host group")
-	}
-
-	// We can only map one user and group inside a container (our own).
-	if len(config.UidMappings) != 1 || config.UidMappings[0].Size != 1 {
-		return fmt.Errorf("rootless containers cannot map more than one user")
-	}
-	if len(config.GidMappings) != 1 || config.GidMappings[0].Size != 1 {
-		return fmt.Errorf("rootless containers cannot map more than one group")
+	if len(config.GidMappings) == 0 {
+		return fmt.Errorf("rootless containers requires at least one UID mapping")
 	}
 
 	return nil
@@ -104,11 +94,28 @@ func rootlessMount(config *configs.Config) error {
 		// Check that the options list doesn't contain any uid= or gid= entries
 		// that don't resolve to root.
 		for _, opt := range strings.Split(mount.Data, ",") {
-			if strings.HasPrefix(opt, "uid=") && opt != "uid=0" {
-				return fmt.Errorf("cannot specify uid= mount options in rootless containers where argument isn't 0")
+			if strings.HasPrefix(opt, "uid=") {
+				var uid int
+				n, err := fmt.Sscanf(opt, "uid=%d", &uid)
+				if n != 1 || err != nil {
+					// Ignore unknown mount options.
+					continue
+				}
+				if !hasIDMapping(uid, config.UidMappings) {
+					return fmt.Errorf("cannot specify uid= mount options for unmapped uid in rootless containers")
+				}
 			}
-			if strings.HasPrefix(opt, "gid=") && opt != "gid=0" {
-				return fmt.Errorf("cannot specify gid= mount options in rootless containers where argument isn't 0")
+
+			if strings.HasPrefix(opt, "gid=") {
+				var gid int
+				n, err := fmt.Sscanf(opt, "gid=%d", &gid)
+				if n != 1 || err != nil {
+					// Ignore unknown mount options.
+					continue
+				}
+				if !hasIDMapping(gid, config.GidMappings) {
+					return fmt.Errorf("cannot specify gid= mount options for unmapped gid in rootless containers")
+				}
 			}
 		}
 	}
--- a/libcontainer/configs/validate/rootless_test.go
+++ b/libcontainer/configs/validate/rootless_test.go
@@ -66,28 +66,6 @@ func TestValidateRootlessMappingUid(t *testing.T) {
 	if err := validator.Validate(config); err == nil {
 		t.Errorf("Expected error to occur if no uid mappings provided")
 	}
-
-	config = rootlessConfig()
-	config.UidMappings[0].HostID = geteuid() + 1
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur if geteuid() != mapped uid")
-	}
-
-	config = rootlessConfig()
-	config.UidMappings[0].Size = 1024
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur if more than one uid mapped")
-	}
-
-	config = rootlessConfig()
-	config.UidMappings = append(config.UidMappings, configs.IDMap{
-		HostID:      geteuid() + 1,
-		ContainerID: 0,
-		Size:        1,
-	})
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur if more than one uid extent mapped")
-	}
 }
 
 func TestValidateRootlessMappingGid(t *testing.T) {
@@ -97,28 +75,6 @@ func TestValidateRootlessMappingGid(t *testing.T) {
 	config.GidMappings = nil
 	if err := validator.Validate(config); err == nil {
 		t.Errorf("Expected error to occur if no gid mappings provided")
-	}
-
-	config = rootlessConfig()
-	config.GidMappings[0].HostID = getegid() + 1
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur if getegid() != mapped gid")
-	}
-
-	config = rootlessConfig()
-	config.GidMappings[0].Size = 1024
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur if more than one gid mapped")
-	}
-
-	config = rootlessConfig()
-	config.GidMappings = append(config.GidMappings, configs.IDMap{
-		HostID:      getegid() + 1,
-		ContainerID: 0,
-		Size:        1,
-	})
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur if more than one gid extent mapped")
 	}
 }
 
@@ -149,6 +105,18 @@ func TestValidateRootlessMountUid(t *testing.T) {
 	if err := validator.Validate(config); err != nil {
 		t.Errorf("Expected error to not occur when setting uid=0 in mount options: %+v", err)
 	}
+
+	config.Mounts[0].Data = "uid=2"
+	config.UidMappings[0].Size = 10
+	if err := validator.Validate(config); err != nil {
+		t.Errorf("Expected error to not occur when setting uid=2 in mount options and UidMapping[0].size is 10")
+	}
+
+	config.Mounts[0].Data = "uid=20"
+	config.UidMappings[0].Size = 10
+	if err := validator.Validate(config); err == nil {
+		t.Errorf("Expected error to occur when setting uid=20 in mount options and UidMapping[0].size is 10")
+	}
 }
 
 func TestValidateRootlessMountGid(t *testing.T) {
@@ -175,6 +143,18 @@ func TestValidateRootlessMountGid(t *testing.T) {
 	config.Mounts[0].Data = "gid=0"
 	if err := validator.Validate(config); err != nil {
 		t.Errorf("Expected error to not occur when setting gid=0 in mount options: %+v", err)
+	}
+
+	config.Mounts[0].Data = "gid=5"
+	config.GidMappings[0].Size = 10
+	if err := validator.Validate(config); err != nil {
+		t.Errorf("Expected error to not occur when setting gid=5 in mount options and GidMapping[0].size is 10")
+	}
+
+	config.Mounts[0].Data = "gid=11"
+	config.GidMappings[0].Size = 10
+	if err := validator.Validate(config); err == nil {
+		t.Errorf("Expected error to occur when setting gid=11 in mount options and GidMapping[0].size is 10")
 	}
 }
 
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -1,3 +1,4 @@
+
 #define _GNU_SOURCE
 #include <endian.h>
 #include <errno.h>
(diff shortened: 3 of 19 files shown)


teddyking:
Hey @giuseppe, I opened a PR looking to add this same functionality a couple of months ago: https://github.com/opencontainers/runc/pull/1403. There's a discussion ongoing in the linked PR, would be great to hear your thoughts on that as well. It looks like we were both heading down the same path here.

libcontainer/configs/validate/rootless.go
```diff
@@ -91,27 +68,3 @@ func rootlessCgroup(config *configs.Config) error {
 
 	return nil
 }
-
-// mount verifies that the user isn't trying to set up any mounts they don't have
-// the rights to do. In addition, it makes sure that no mount has a `uid=` or
-// `gid=` option that doesn't resolve to root.
-func rootlessMount(config *configs.Config) error {
```
cyphar: I'm not convinced that dropping this makes sense. Even in the multiple-uid case, really we should be making it so that we emit an error if the `uid=...` or `gid=...` options reference unmapped users. The errors are far too cryptic if you ignore it.

libcontainer/configs/validate/rootless_test.go
```diff
@@ -98,84 +76,6 @@ func TestValidateRootlessMappingGid(t *testing.T) {
...
-
-	config.Mounts[0].Data = "gid=5"
-	if err := validator.Validate(config); err == nil {
-		t.Errorf("Expected error to occur when setting gid=5 in mount options")
-	}
-
-	config.Mounts[0].Data = "gid=0"
-	if err := validator.Validate(config); err != nil {
-		t.Errorf("Expected error to not occur when setting gid=0 in mount options: %+v", err)
-	}
```
cyphar: ... so all of these removed tests should stay, and we should add new tests making sure that multiple-mapping cases are still treated sanely.

libcontainer/nsenter/nsexec.c
```diff
@@ -192,22 +194,74 @@ static void update_setgroups(int pid, enum policy_t setgroup)
 	}
 }
 
-static void update_uidmap(int pid, char *map, size_t map_len)
+static int attempt_mapping_tool(const char *app, int pid, char *map, size_t map_len)
```
cyphar: Please format this in the Linux kernel style (sans the 80-column restriction).

libcontainer/nsenter/nsexec.c
```diff
@@ -192,22 +194,74 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+{
+#define MAX_ARGV 20
+	char *argv[MAX_ARGV];
+	char pid_fmt[16];
+	int argc = 0;
+	size_t i;
+	int child;
+
+	sprintf (pid_fmt, "%d", pid);
```
cyphar: `asprintf`?

libcontainer/nsenter/nsexec.c
```diff
@@ -192,22 +194,74 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+	child = fork();
+	if (child < 0)
+		bail("failed to fork");
+
+	if (child == 0) {
+		execv (app, argv);
+	}
+	else {
+		int status;
+		while (waitpid (child, &status, 0) < 0 && errno == EINTR);
```
cyphar:
```
do
    waitpid(child, &status, 0);
while (errno == EINTR);
```

If not, just put the `;` on a bare newline so it's more clear it's an empty loop.

libcontainer/nsenter/nsexec.c
```diff
@@ -192,22 +194,74 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+}
+
+static void update_uidmap(int pid, uint8_t is_rootless, char *map, size_t map_len)
 {
 	if (map == NULL || map_len <= 0)
 		return;
 
-	if (write_file(map, map_len, "/proc/%d/uid_map", pid) < 0)
+	if (write_file(map, map_len, "/proc/%d/uid_map", pid) < 0) {
+		if(errno == EPERM && is_rootless && attempt_mapping_tool ("/usr/bin/newuidmap", pid, map, map_len) == 0)
```
cyphar: I'm not sure how I feel about hard-coding the `/usr/bin/newuidmap`. At this point, `$PATH` should still be trustworthy (so `execvp` could work), but maybe @crosbymichael knows better than me.

main.go
```diff
@@ -60,6 +60,14 @@ func main() {
 	}
 	v = append(v, fmt.Sprintf("spec: %s", specs.Version))
 	app.Version = strings.Join(v, "\n")
+
+	root := os.Getenv("XDG_RUNTIME_DIR")
```
cyphar: This is not correct, as it will mean that if you run `runc` as root with `XDG_RUNTIME_DIR` set, then the spawned containers won't be visible by other `runc` management processes. Instead, this should be conditional based on whether you're an unprivileged user. Even then I'm still not convinced, because what if someone sets up `/run/runc` to be like `/tmp` (sticky-bit and world-writeable) in order for all users to access the same store of container state?

I agree that `runc` should obey `XDG_RUNTIME_DIR`, but I'm not entirely convinced that this is correct (especially since changing `XDG_RUNTIME_DIR` will make it more difficult for other people to manage the `runc` containers spawned by someone with `XDG_RUNTIME_DIR` set.

@crosbymichael do you have any ideas on how this should work nicely?

libcontainer/nsenter/nsexec.c
```diff
@@ -597,8 +651,8 @@ void nsexec(void)
 						update_setgroups(child, SETGROUPS_DENY);
 
 					/* Set up mappings. */
-					update_uidmap(child, config.uidmap, config.uidmap_len);
-					update_gidmap(child, config.gidmap, config.gidmap_len);
+					update_uidmap(child, config.is_rootless, config.uidmap, config.uidmap_len);
+					update_gidmap(child, config.is_rootless, config.gidmap, config.gidmap_len);
```
cyphar: Why not always try to use `newuidmap` and `newgidmap` rather than making it conditional on `is_rootless`? (Though maybe you're right here.)

libcontainer/nsenter/nsexec.c
giuseppe: IMO it would be wasteful to allocate dynamic memory when we know it will be just to format an `int`

libcontainer/nsenter/nsexec.c
giuseppe: yes, fixed and add an error check

libcontainer/nsenter/nsexec.c
giuseppe: I've changed it to `execvp` so that PATH is honored.

libcontainer/nsenter/nsexec.c
giuseppe: I am using `newuidmap` and `newgidmap` only after an `EPERM`.  When running as `root` that should never happen

main.go
giuseppe: I've changed it to be used only when `geteuid() !=0`.  I think a shared directory for all users is a very uncommon use case, while most users would probably prefer to have runc working out of the box and behave like other applications that use `XDG_RUNTIME_DIR`.  The first case is still possible though by overriding the root directory, and I think it is preferable to require extra configuration for an uncommon use like a shared directory for all users.

giuseppe:
@teddyking yes, we got to something very similar :)  An extra point for considering this the right approach, and for the maintainers to pick what they prefer from both implementations

libcontainer/nsenter/nsexec.c
cyphar: Right, so then the `is_rootless` check is unecessary (which is what I meant)?

libcontainer/nsenter/nsexec.c
cyphar: It's just because I don't like that you've hard-coded the length of `pid_fmt` to be `11`.

Personally I like @teddyking's approach to the formatting a bit more (we send most of the information from Go with a netlink message rather than formatting everything in C). I prefer that because the context in which this C code is run is quite fragile and I get less worried if we do things in Go when possible. I'll read through his PR again and point out which bits seem good to me.

giuseppe: AFAICS #1403 is doing a very similar thing for preparing the argv for `newuidmap/newgidmap`

The main difference is that here I am not using `strtok`  and parsing the string inline without allocating any memory.  I've avoided input validation here as it would be superfluous since newuidmap/newgidmap do that already.

libcontainer/nsenter/nsexec.c
williammartin: One thing we'd quite like to bring over from #1403 is the ability to specify `--newuidmap` and `--newgidmap` binary locations. It's seems to be in line with the `--criuPath` and it's our preferred way to be explicit in Garden.

libcontainer/nsenter/nsexec.c
giuseppe: sure I can remove that.  I've pushed a new version where I drop it.

giuseppe:
is there anything I can do to get this moving forward?

cyphar:
Yeah, there's a few things left that we can take from #1403. I'll list them as a checklist below.

* [ ] Switch to passing the (`$PATH` expanded) path of `newuidmap` and `newgidmap` via `netlink`. I don't think they need to be user-visible options at the moment, but I don't like hardcoding them in nsexec.
* [ ] Add some tests for this functionality, if possible (you could use what #1403 has as inspiration). If you don't feel like doing it, I can work on it after this is merged.

I'll also add some in-line comments.

libcontainer/nsenter/nsexec.c
```diff
@@ -191,22 +194,83 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+				bail("failed to waitpid");
+			}
+			if (WIFEXITED(status))
+				return WEXITSTATUS(status);
+                  }
+	}
+	return -1;
+}
+
+static void update_uidmap(int pid, uint8_t is_rootless, char *map, size_t map_len)
```
cyphar: `is_rootless` is still here, even though it isn't used anymore.

libcontainer/nsenter/nsexec.c
```diff
@@ -191,22 +194,83 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+		if(errno == EPERM
+                   && try_mapping_tool ("newuidmap", pid, map, map_len) == 0) {
+			return;
+		}
 		bail("failed to update /proc/%d/uid_map", pid);
+	}
 }
 
-static void update_gidmap(int pid, char *map, size_t map_len)
+static void update_gidmap(int pid, uint8_t is_rootless, char *map, size_t map_len)
```
cyphar: ... and it's still here too.

libcontainer/nsenter/nsexec.c
```diff
@@ -191,22 +194,83 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
 		return;
 
-	if (write_file(map, map_len, "/proc/%d/gid_map", pid) < 0)
+	if (write_file(map, map_len, "/proc/%d/gid_map", pid) < 0) {
+		if(errno == EPERM
+                   && try_mapping_tool ("newgidmap", pid, map, map_len) == 0) {
+			return;
+		}
 		bail("failed to update /proc/%d/gid_map", pid);
+	}
```
cyphar: Completely stylistic, but I think this is more readable (same applies for `update_uidmap`):

```
if (write_file(map, map_len, "/proc/%d/gid_map", pid) < 0) {
	if (errno != EPERM)
		bail("failed to update /proc/%d/gid_map", pid);
	if (try_mapping_tool("newgidmap", pid, map, map_len))
		bail("failed to use newgid map on %d", pid);
}
```

main.go
```diff
@@ -61,6 +61,13 @@ func main() {
 	}
 	v = append(v, fmt.Sprintf("spec: %s", specs.Version))
 	app.Version = strings.Join(v, "\n")
+
+	root := "/run/runc"
+	if os.Geteuid() != 0 {
+		runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
+		root = runtimeDir + "/runc"
+	}
```
cyphar: This doesn't handle the case where `XDG_RUNTIME_DIR` is unspecified safely. If it's not set, just use `/run/runc` as usual.

libcontainer/configs/validate/rootless.go
```diff
@@ -104,11 +95,24 @@ func rootlessMount(config *configs.Config) error {
 		// Check that the options list doesn't contain any uid= or gid= entries
 		// that don't resolve to root.
 		for _, opt := range strings.Split(mount.Data, ",") {
-			if strings.HasPrefix(opt, "uid=") && opt != "uid=0" {
-				return fmt.Errorf("cannot specify uid= mount options in rootless containers where argument isn't 0")
+			if strings.HasPrefix(opt, "uid=") {
+				uid, _ := strconv.Atoi(strings.Replace(opt, "uid=", "", 1))
```
cyphar: I don't like the ignoring of errors, nor `strings.Replace` here. Can we just do something like

```
var uid int
n, err := fmt.Sscanf(opt, "uid=%d", &uid)
if n != 1 || err != nil {
    // Ignore unknown mount options.
    continue
}
if !hasIDMapping(uid, config.UidMappings) {
    return fmt.Errorf("cannot specify uid= mount options for unmapped uid in rootless containers")
}
```

And the same for the other option.

libcontainer/nsenter/nsexec.c
```diff
@@ -191,22 +194,83 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+
+		sprintf (pid_fmt, "%d", pid);
+
+		argv[argc++] = (char *) app;
+		argv[argc++] = pid_fmt;
+		/*
+		 * Convert the map string into a list of argument that
+		 * newuidmap/newgidmap can understand.
+		 */
+		for (i = 0; i < map_len - 1; ) {
```
cyphar: Is it possible to do this with strtok?

libcontainer/nsenter/nsexec.c
```diff
@@ -191,22 +194,83 @@ static void update_setgroups(int pid, enum policy_t setgroup)
...
+		bail("failed to fork");
+
+	if (child == 0) {
+#define MAX_ARGV 20
+		char *argv[MAX_ARGV];
+		char pid_fmt[16];
+		int argc = 0;
+		size_t i;
+
+		sprintf (pid_fmt, "%d", pid);
```
cyphar: While it would never be an issue, use `snprintf(pid_fmt, 16, ...)`.

cyphar:
Also I believe this needs a rebase, because the test failures are happening due to `shfmt` (which we no longer use).

giuseppe:
I've taken some of the changes from the other PR (to pass the uid/gid executable name from the golang part) and addressed the other issues.  I'll try to add some more tests later today or next week

About the failing tests, the PR is rebased but it is still failing, I can see other open PRs have the same issue so probably master is affected as well.

cyphar:
#1578 fixes the CI.

libcontainer/configs/validate/rootless.go
```diff
@@ -104,11 +94,24 @@ func rootlessMount(config *configs.Config) error {
 		// Check that the options list doesn't contain any uid= or gid= entries
 		// that don't resolve to root.
 		for _, opt := range strings.Split(mount.Data, ",") {
-			if strings.HasPrefix(opt, "uid=") && opt != "uid=0" {
-				return fmt.Errorf("cannot specify uid= mount options in rootless containers where argument isn't 0")
+			var uid int
+			n, err := fmt.Sscanf(opt, "uid=%d", &uid)
```
cyphar: We still need the `string.Contains` blocks from before, because now `gid=...` won't correctly be detected (the first `Sscanf` fails).

cyphar:
That's my final must-be-fixed-before-merge comment. I'm planning on doing some further cleanups (including getting some tests in place) after this is merged.

giuseppe:
@ciphar fixed here ⬆️

cyphar:
LGTM.

cyphar:
/me does some more testing. Thanks @giuseppe. Tentative LGTM, after I do some tests. :heart:

cyphar:
@giuseppe I've found a few bugs in the current `newgidmap` handling, as well as some stylistic issues. Do you mind if I push the fixup to your branch and then LGTM it?

giuseppe:
@cyphar Thanks for the fixes! They LGTM

cyphar:
@giuseppe I've pulled some of the changes from #1403, and rewrote some others to work better in the test framework. Now all of this code is tested as an integration suite run. :smile_cat:

cyphar:
LGTM.

cyphar:
LGTM (again, needed rebase).

crosbymichael:
LGTM

AkihiroSuda:
Added `runc spec` support for optionally generating a config with multiple uidMappings and gidMappings: https://github.com/opencontainers/runc/pull/1692

PR #1530, opened 2017-07-21
libcontainer: one more switch from syscall to x/sys/unix

tklauser:
Refactor DeviceFromPath in order to get rid of package syscall and
directly use the functions from x/sys/unix. This also allows to get rid
of the conversion from the OS-independent file mode values (from the os
package) to Linux specific values and instead let's us use the raw
file mode value directly.


cyphar:
LGTM, thanks @tklauser.

dqminh:
LGTM

PR #1533, opened 2017-07-24
Remove the code that close negative descriptor

keloyang:
There is a possibility that  the code close negative descriptor,  it don't lead to a issue, but the code is ugly.


dqminh:
LGTM

crosbymichael:
LGTM

PR #1535, opened 2017-07-27
Use CRIU VERSION RPC if available

adrianreber:
With this runC also uses RPC to ask CRIU for its version. CRIU supports a VERSION RPC since CRIU 3.0 and using the RPC interface does not require parsing the console output of CRIU (which could change anytime).

For older CRIU versions which do not yet have the VERSION RPC runC falls back to its old CRIU output parsing mode.

Once CRIU 3.0 is the minimum version required for runC the old code can be removed.


mrunalp:
@avagin ping

libcontainer/container_linux.go
```diff
@@ -592,47 +592,52 @@ func (c *linuxContainer) checkCriuFeatures(criuOpts *CriuOpts, rpcOpts *criurpc.
...
-		var version string
-		if sp := strings.Index(string(out), "GitID"); sp > 0 {
-			version = string(out)[sp:ep]
-		} else {
-			return fmt.Errorf("Unable to parse the CRIU version: %s", c.criuPath)
-		}
+	// If the version of criu has already been determined there is no need
+	// to ask criu for the version again. Use the value from c.criuVersion.
+	if c.criuVersion == 0 {
+		var x, y, z int
```
avagin: It may be better to introduce a function for this code?

libcontainer/container_linux.go
```diff
@@ -590,60 +590,109 @@ func (c *linuxContainer) checkCriuFeatures(criuOpts *CriuOpts, rpcOpts *criurpc.
...
 
 	// If the version of criu has already been determined there is no need
 	// to ask criu for the version again. Use the value from c.criuVersion.
-	if c.criuVersion == 0 {
-		var x, y, z int
+	if c.criuVersion != 0 {
+		return compareCriuVersion(c.criuVersion, minVersion)
+	}
 
-		out, err := exec.Command(c.criuPath, "-V").Output()
```
avagin: In a previous patch you shifted this code and now you shifted it back. Try to avoid such things.

adrianreber:
@avagin: Thanks for having a look. You are right, there was unnecessary code movement which I removed and I also moved the version parsing into a separate function.

adrianreber:
Not sure why travis failed. My other PR (#1541) which includes these patches was successful. So I guess the failure is unrelated to this PR?

Any other comments?

avagin:
I'm agree that the failure is unrelated.
```
--- FAIL: TestNotifyMemoryPressure (0.00s)
	notify_linux_test.go:99: expected event control to be closed, but received error errno 0
....
```

LGTM

cyphar:
I've retriggered the CI, and will merge if it succeeds.

cyphar:
LGTM.

PR #1536, opened 2017-07-27
Add CFS data to libcontainer cgroups

14rcole:
Add information about the completely fair scheduler in cpu cgroups to libcontainer/cgroups. and report the information as a part of `runc events`

### DIFF
--- a/events.go
+++ b/events.go
@@ -74,8 +74,14 @@ type cpuUsage struct {
 	User   uint64   `json:"user"`
 }
 
+type cfs struct {
+	Period uint64 `json:"period,omitempty"`
+	Quota  int64  `json:"quota,omitempty"`
+}
+
 type cpu struct {
 	Usage      cpuUsage   `json:"usage,omitempty"`
+	CFS        cfs        `json:"cfs,omitempty"`
 	Throttling throttling `json:"throttling,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


14rcole:
@mrunalp PTAL

mrunalp:
LGTM
@hqhq PTAL

events.go
```diff
@@ -74,8 +74,14 @@ type cpuUsage struct {
...
 }
 
+type cfs struct {
+	Period uint64 `json:"period,omitempty"`
+	Quota  int64  `json:"quota,omitempty"`
+}
+
 type cpu struct {
 	Usage      cpuUsage   `json:"usage,omitempty"`
+	Scheduling cfs        `json:"cfs,omitempty"`
```
crosbymichael: Do we want to make that stats output and this the same?  You have `Scheduling` field, `cfs` tag, and `CpuCfs`, `cpu_cfs` in the two different places.

I think we should pick one and stick with it

14rcole: I think these should all be CpuCfs or cpu_cfs.  I'll change them

14rcole: actually, CPU.CpuCfs is redunant, that one should be CFS.  So the fields will be called Cfs and the json fields, in order to match the others, will be cpu_cfs

hqhq:
@14rcole CI failure is relevant.

hqhq:
@14rcole Why do you need to get these CFS data from `runc event`? They are technically not statistics, but configurations, you should get them from higher level tools like `docker inspect`, or refer to `config.json` on runc level.

cyphar:
I agree with @hqhq, this doesn't make a lot of sense to me. Events should specifically be for statistics and other information the kernel is providing that isn't part of the configuration (such as the current usage, OOM events, etc).

hqhq:
@14rcole I'll close this for now, let us know if you have more concerns, or create an issue so we can discuss and figure out the right approach to fulfill your needs.

cyphar:
@hqhq I think that the issue might come up if you use `runc update` (and so the original `config.json` is no longer accurate). IMO the best solution for this would be to provide `runc state` in a way that it spits out the "current applied configuration". At the moment this is not possible (internally we store `libcontainer/config` rather than OCI configurations) but that's something else that myself and @crosbymichael were planning on getting done. It would be a _lot_ of work to do though.

hqhq:
@cyphar Yeah `runc state` is a better place for such information.

PR #1539, opened 2017-07-31
 recvtty socket file replaced by pid-file during creation of container

rajasec:
Opened the socket by 
sudo ./contrib/cmd/recvtty/recvtty -m null /tmp/testfilesock

ls -ltr /tmp/filesock
srwxr-xr-x 1 root root 0 Aug  1 22:02 /tmp/testfilesock

In another terminal, created the container using the following command.
sudo ./runc create --console-socket /tmp/testfilesock --pid-file /tmp/testfilesock test
ls -ltr /tmp/filesock

Since os.Rename called at last, it replaces the socket file with pid-file
So when we try to use this socket for second container, socket file has been replaced by normal file, so it throws out an error saying "address in in use"
This problem comes when we pass the same name for both console-socket and pid-file.

Added the condition, so it will error out if the file exists.

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -171,6 +171,9 @@ func createPidFile(path string, process *libcontainer.Process) error {
 	if err != nil {
 		return err
 	}
+	if _, err := os.Stat(path); err == nil {
+		return fmt.Errorf("file %s already exists", path)
+	}
 	var (
 		tmpDir  = filepath.Dir(path)
 		tmpName = filepath.Join(tmpDir, fmt.Sprintf(".%s", filepath.Base(path)))


utils_linux.go
```diff
@@ -171,6 +171,9 @@ func createPidFile(path string, process *libcontainer.Process) error {
 	if err != nil {
 		return err
 	}
+	if _, err := os.Stat(path); err == nil {
+		return fmt.Errorf("file %s already exists", path)
```
cyphar: I'm not sure this way of solving the issue makes sense. There might be cases where you want to overwrite a stale pidfile. Also, this should be `Lstat` AFAICS (if `path` is a broken symlink this check won't be hit).

cyphar: Then again, maybe making `runc` not indiscriminately override existing files would be a good idea.

dqminh: yah overwriting pid files is usually a norm. The easy way here is probably just detect at the flag parsing time (i.e. socket file is not the same as pid file )

cyphar: I'm a bit worried that is going to be a moving target we have to remember to update each time we add a new flag. Maybe if we check the d_type and if it's not a normal file we error out (the `Rename` would cause it to be overridden anyway which is counter-intuitive if you want the `pid` to be written to a FIFO for example).

rajasec: with respect to d_type, I'm not sure whether d_type supports in all file systems, if the file system is xfs ?

dqminh: you can probably detect if the file is a socket using its filemode. However i feel this is a bit overkill considering that this is more or less an user misconfiguration that can be very easily detected much earlier in the code by just comparing the flags, or perhaps even adding some documentation will suffice.

We only have two file flags now, i dont think its very confusing.

rajasec: As the host files are getting replaced, adding documentation will not be suffice here.
we can ensure runc not override not existing files.

crosbymichael:
I'm not sure this is worth fixing in runc.  Its common, as @dqminh to overwrite pid files.  I think the right thing here is to not do that as a user as it makes no sense.

Issue #1457, opened 2017-05-22
subsystems list duplicated across files


yastij:
as states the TODO in `runc/libcontainer/cgroups/rootless/rootless.go`, we duplicate an array that contains a bunch of cgroups (duplicate : `runc/libcontainer/cgroups/fs/apply_raw.go` ) : 

```go
	subsystems = subsystemSet{
		&CpusetGroup{},
		&DevicesGroup{},
		&MemoryGroup{},
		&CpuGroup{},
		&CpuacctGroup{},
		&PidsGroup{},
		&BlkioGroup{},
		&HugetlbGroup{},
		&NetClsGroup{},
		&NetPrioGroup{},
		&PerfEventGroup{},
		&FreezerGroup{},
		&NameGroup{GroupName: "name=systemd", Join: true},
	}
```
this should be refactored

cyphar:
The refactoring work is much more significant than I outline in `rootless.go`. The correct way to handle cgroups (or rather the _most_ correct way) would be to use `access(2)` to figure out whether any particular cgroup subsystem setup will fail and then bubble it up so that cgroups will only error out in rootless containers in cases where it is _truly_ necessary. But that will require having a better understanding of whether a particular configuration changes the "default" in the `Manager` level of the cgroup code.

The reason I didn't implement it that way is because it's not code that anyone will enjoy reviewing or touching (and rootless is actually just doing a `noop` of most operations).

Honestly we should torch this whole thing. Especially the systemd part. :wink:

cyphar:
From memory @williammartin (or someone from CF) mentioned they might work on this at some point.

yastij:
@cyphar maybe we should implement it as you said, if no one at CF ends up working on it, i'll submit a PR if you want

williammartin:
Unfortunately the work we'd like to do here keeps getting bumped down the backlog. Right now it's estimated for the end of July but it's a roll of the dice when it actually gets done. @yastij if you submit a PR that would be excellent as far as CF is concerned.

yastij:
i think that i'll start work on this one, will submit a PR for review ;)

teddyking:
@yastij did you get a chance to start work on this at all? If not we may be in a position to work on it pretty soon.

yastij:
@teddyking got caught on some k8s and docker issues, but would like to be involved on this one. ping me if you or someone submits a PR before me. Thanks !

PR #1540, opened 2017-08-01
Support cgroups with limits as rootless

williammartin:
Following on from the rootless cgroups discussion in https://github.com/opencontainers/runc/issues/1457, this PR provides an initial implementation for rootless cgroups that fulfils Garden's needs. This does not enable all cgroup related functiionality, but we think it's a good starting point for other currently disabled features to work.	

In more detail, this PR removes the current rootless cgroup implementation, and using cgroups in rootless mode should now work broadly the same as rootful, providing that runc has permissions on the cgroup path. 

## Differences with Rootful

Typically, without permissions on the cgroup path, we would expect an error when `applying` (entering a pid into a controller), however we now support a kind of "opportunistic" cgroup usage when no limits have been set and no cgroup path has been provided. Since child processes are entered into the same cgroup as their parent, this is effectively the same (regarding cgroup enforcement) as not providing a cgroup path and not setting limits. Attempting to `set` (change a resource limit) will still result in a permission denied error either during creation or via `runc update` but hopefully with a slightly more informative error.

The devices cgroup doesn't have all functionality available for setting limits. This is because there is a requirement on `CAP_SYS_ADMIN` for the `devices.allow` and `devices.deny` files. We're not sure of a way to provide different device white/blacklisting per container, but a possible solution for some use cases is to set a static list in a parent cgroup that is inherited. This works because `applying` does not require `CAP_SYS_ADMIN`.

## Disabled features

We haven't enabled OOM notification or Memory pressure notificiation but this is hopefully as simple as removing the rootless conditional: https://github.com/opencontainers/runc/blob/e775f0fba3ea329b8b766451c892c41a3d49594d/libcontainer/container_linux.go#L528

We haven't enabled CRIU features because we aren't familiar with what is required. 

## Notes

The BATS added seem to cover the right features, however, we aren't super happy with the changes to the Makefile. Thoughts on how to solve the requirement that a cgroup exists, chowned to rootless in a nicer way would be much appreciated.

Signed-off-by: Ed King <eking@pivotal.io>


Makefile
```diff
@@ -96,10 +96,14 @@ integration: runcimage
 localintegration: all
 	bats -t tests/integration${TESTFLAGS}
 
+rootlessintegrationstub:
+	mkdir -p /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/rootless_cgroup
+	chown rootless:rootless -R /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/rootless_cgroup
+	make localrootlessintegration
+
 rootlessintegration: runcimage
-	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) --cap-drop=ALL -u rootless $(RUNC_IMAGE) make localintegration
```
cyphar: Why have you removed `--cap-drop=all`? That was necessary to ensure that rootless didn't depend on extra capabilities.

Makefile
```diff
@@ -96,10 +96,14 @@ integration: runcimage
 localintegration: all
 	bats -t tests/integration${TESTFLAGS}
 
+rootlessintegrationstub:
+	mkdir -p /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/rootless_cgroup
+	chown rootless:rootless -R /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/rootless_cgroup
```
cyphar: It is crucial that when we add new tests for opportunistic functionality in rootless that we make sure that the base functionality works. So if we want to add tests for this (and also `newuidmap` in the future) we need to add them as effectively re-runs of the rootless tests under different conditions (with some extra tests enabled).

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -142,11 +142,19 @@ func (m *Manager) Apply(pid int) (err error) {
...
+			if os.IsPermission(err) && m.Cgroups.Path == "" {
+				// If we didn't set a cgroup path, then let's defer the error here
+				// until we know whether we have set limits or not.
+				// If we hadn't set limits, then it's ok that we couldn't join this cgroup, because
+				// it will have the same limits as its parent.
+				continue
+			}
 			return err
 		}
+		m.Paths[sys.Name()] = p
```
cyphar: IIRC we moved this before `Apply` to handle some race conditions, maybe @hqhq remembers this better than me.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -198,6 +206,10 @@ func (m *Manager) Set(container *configs.Config) error {
 	for _, sys := range subsystems {
 		path := paths[sys.Name()]
 		if err := sys.Set(path, container.Cgroups); err != nil {
+			if path == "" {
+				// cgroup never applied
+				return fmt.Errorf("cannot set limits on the %s cgroup, as the container has not joined it. The cgroup controller may not be mounted or you may not have previously had permissions to create the cgroup subdirectory", sys.Name())
```
cyphar: This error is *way* too long.

libcontainer/process_linux.go
```diff
@@ -83,8 +83,7 @@ func (p *setnsProcess) start() (err error) {
 	if err = p.execSetns(); err != nil {
 		return newSystemErrorWithCause(err, "executing setns process")
 	}
-	// We can't join cgroups if we're in a rootless container.
-	if !p.config.Rootless && len(p.cgroupPaths) > 0 {
```
cyphar: This check is still correct in some cases, but I guess erroring out is acceptable if someone explicitly asked for an impossible cgroup configuration (now that we could in principle nest things). I would like to see a test for this though.

cyphar:
Just to be clear, I'm very impressed that there were only two changes necessary to make the `cgroups/fs` driver opportunistic and this looks like it's on the right track. My main gripes are just with how we're testing stuff.

yastij:
Thanks so much guys for the PR !

libcontainer/cgroups/fs/apply_raw.go
cyphar: @williammartin What if you keep the `m.Paths[sys.Name()] = p` where it is, but do `delete(m.Paths, sys.Name())` if there is a permission error that is ignore-able?

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -142,11 +142,19 @@ func (m *Manager) Apply(pid int) (err error) {
 			}
 			return err
 		}
-		m.Paths[sys.Name()] = p
 
 		if err := sys.Apply(d); err != nil {
+			if os.IsPermission(err) && m.Cgroups.Path == "" {
```
cyphar: Another thing I just thought of is that we should only be permissive if we're running in rootless mode (though I imagine that such cases are unlikely, and passing `config.Rootless` down here might be a bit ugly).

libcontainer/cgroups/fs/apply_raw.go
teddyking: Yep, no problem.

libcontainer/process_linux.go
teddyking: We're not clear under what circumstances the rootless check still makes sense? Can you give an example please?

`p.cgroupPaths` is loaded from the `state.json` which best we can tell has to be the same as the cgroup paths the init process is in, so barring people doing weird things, if you succeeded on create, you should succeed now?

libcontainer/cgroups/fs/apply_raw.go
teddyking: That seems reasonable, we can try it out

libcontainer/cgroups/fs/apply_raw.go
teddyking: We hadn't done this because we couldn't think of a way in which root could get a permission error. Maybe from a Mandatory Access Control?

Makefile
teddyking: It seems `--cap-drop=ALL` doesn't do anything useful when also using `--privileged`. With or without it, once we switch to the rootless user our caps are:

```
CapInh: 0000003fffffffff
CapPrm: 0000000000000000
CapEff: 0000000000000000
CapBnd: 0000003fffffffff
CapAmb: 0000000000000000
```

So, as the tests are right now in master, and in this PR, we don't have any effective caps. If we want to guard against picking up caps somehow from files, we could perhaps figure out a way to drop from the bounding/inheritable set as well but I'm not sure that really makes sense within the test context.

Makefile
teddyking: Not quite sure we understand this, could you elaborate please?

teddyking:
Thanks very much for the review. Hope we responded to all the comments.

--
@williammartin and Me.

Makefile
cyphar: While we have the `require rootless_cgroup` class of tests, and the rest of the tests that don't have that requirement, I'm worried about not explicitly testing the case where this setup code is not run (in other words, the case were `runc` has no permissions to touch any cgroups). I appreciate that code is still tested, but I just want to be cautious about testing "opportunistic" functionality like rootless cgroup handling or (in the future) `newuidmap` handling.

libcontainer/cgroups/fs/apply_raw.go
cyphar: Yeah, I think you're right. We can fix it later if we hit it some other time.

libcontainer/process_linux.go
cyphar: I'm actually not sure what I meant either, sorry about that. This change is fine, but I would like to see a `runc exec` test with rootless cgroups in use, to make sure this works fine.

teddyking:
We've rebased and made a few changes, specifically:

cyphar:
First-pass this looks good, squashing into relevant commits would be preferred. I'm going to test this out over the next few days, and then LGTM if it all looks good. Thanks so much for working on this @williammartin, @jszroberto, and @teddyking! :heart:

williammartin:
@cyphar No problem, thanks for keeping on top of this and providing useful pointers!

cyphar:
The issues with test extensibility and making sure opportunistic features are tested correctly are solved in #1529. I can help you rebase this PR's tests on the new `tests/rootless.sh` setup once it's merged.

teddyking:
Awesome, sounds good to me. I've squashed the commits here into one so hopefully it shouldn't be too difficult to rebase once #1529 is merged in.

teddyking:
We've rebased on master and taken a stab at updating this PR's tests to fit the new `tests/rootless.sh` way of donig things.
We had a few issues with the `update.bats` tests (and I've just seen the jenkins build has failed on these ... will take a look at that now).

Here's a brief overview of the changes since last push:

* Add $TESTFLAGS to `tests/rootless.sh`
* Add "cgroups" to ALL_FEATURES with corresponding enable/disable_ funcs
* Remove `requires root` from cgroup integration tests that previously had it
* In `helpers.bash` we updated the runc_spec function to ensure that the config.json included a Linux.Resources object, as this is expected by the sed command in `setup()` func in `updates.bash`
* This does raise the question as to whether `runc spec --rootless` should generate this by default?

cyphar:
@teddyking Cool, thanks. I'm currently at OSS but I'll take a look at this next week.

teddyking:
@cyphar great! I pushed a fix for the failing travis build, but it's now failing on the go 1.8.x job only for a totally unrelated reason (error downloading busybox.tar.xz), so it probably just needs to be kicked off again.

tests/rootless.sh
```diff
@@ -55,6 +55,21 @@ function disable_idmap() {
...
 }
 
+# FEATURE: Opportunistic cgroups support, allowing a rootless container to set
+#          resource limits on condition that cgroupsPath is set to a path the
+#          rootless user has permissions on.
+
+function enable_cgroup() {
+	# Set up cgroups for use in rootless containers.
+	mkdir -p /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/runc-cgroups-integration-test
+	chown rootless:rootless -R /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/runc-cgroups-integration-test
```
cyphar: I would prefer if we did `chown rootless:rootless -R /sys/fs/cgroup/*/runc-cgroups-integration-test` here.

tests/rootless.sh
```diff
@@ -55,6 +55,21 @@ function disable_idmap() {
...
+}
+
+function disable_cgroup() {
+	# Remove cgroups used in rootless containers.
+	[ -d /sys/fs/cgroup/devices/runc-cgroups-integration-test ] && rmdir /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/runc-cgroups-integration-test
```
cyphar: And `rmdir /sys/fs/cgroup/*/runc-cgroups-integration-test` here.

tests/integration/cgroups.bats
```diff
@@ -78,3 +74,54 @@ EOF
...
     fi
 }
+
+@test "runc create (no limits + no cgrouppath + no permission) succeeds" {
+    runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_permissions
+    [ "$status" -eq 0 ]
+}
+
+@test "runc create (rootless + no limits + cgrouppath + no permission) fails with permission error" {
+    requires rootless
```
cyphar: I will think about this a little more when I get back from OSS, but I am not sure that we'd want the "fails with permission error" tests.

tests/integration/update.bats
```diff
@@ -53,10 +52,10 @@ function check_cgroup_value() {
...
 # TODO: test rt cgroup updating
 @test "update" {
-    # XXX: currently cgroups require root containers.
-	# XXX: Also, this test should be split into separate sections so that we
-	#      can skip kmem without skipping update tests overall.
-    requires cgroups_kmem root
+    # XXX: Also, this test should be split into separate sections so that we
+    #      can skip kmem without skipping update tests overall.
+    requires cgroups_kmem
+    [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
```
cyphar: Minor nit: swap the order of the checks to match the rest of the tests.

tests/rootless.sh
```diff
@@ -55,6 +55,21 @@ function disable_idmap() {
 	[ -e /usr/bin/newgidmap ] && mv /usr/bin/{,unused-}newgidmap
 }
 
+# FEATURE: Opportunistic cgroups support, allowing a rootless container to set
+#          resource limits on condition that cgroupsPath is set to a path the
+#          rootless user has permissions on.
+
+function enable_cgroup() {
+	# Set up cgroups for use in rootless containers.
+	mkdir -p /sys/fs/cgroup/{blkio,cpu,cpuacct,cpuset,devices,freezer,hugetlb,memory,net_cls,net_prio,openrc,perf_event,pids,systemd}/runc-cgroups-integration-test
```
cyphar: We might want to source this list from `/proc/self/cgroups`, but I can do that in a follow-up if you prefer.

cyphar:
Note to myself: We should add some documentation on how to use the lxcfs PAM module to allow a "normal" user to set up delegated cgroups on their host.

tests/rootless.sh
teddyking: Sure, will leave that up to you to decide.

tests/integration/cgroups.bats
williammartin: Why wouldn't you want these? Test bloat?

cyphar: If it ends up working in the future (which may happen if systemd decides to enable the cgroupv2 delegation code), then the test will break. On the other hand, we could just include these and deal with that later.

williammartin: I can buy that. As long as getting a permission error isn't the desired behaviour right now (rather than say, some other error that might bubble up) and we don't want this test for regression, then I say kill the test and save test bloat/potential confusion later.

cyphar: There's a delegation section in the [cgroupv2 documentation](https://www.kernel.org/doc/Documentation/cgroup-v2.txt). Effectively you can set `-o nsdelegate` on the root mount of `cgroupv2` and then you get permissions to write to a sub-cgroup when you're in a cgroup namespace. I did a quick search for the actual patch but couldn't find it.

cyphar:
Okay, the one thing left is the notification scheme for OOM and memory pressure. Effectively I think that we should not allow someone to register for notifications if we did not join a cgroup (and inherited the original cgroups) -- so we'd have to save somewhere whether the cgroup manager (silently) failed to setup cgroups.

cyphar:
Okay, the one thing left is the notification scheme for OOM and memory pressure. Effectively I think that we should not allow someone to register for notifications if we did not join a cgroup (and inherited the original cgroups) -- so we'd have to save somewhere whether the cgroup manager (silently) failed to setup cgroups.

I'll work on this after we merge this, since it's not a blocker IMO.

cyphar:
LGTM. The other issues can be fixed in future versions.

crosbymichael:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1540/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

williammartin:
Great stuff. Thanks @cyphar and all!

PR #1541, opened 2017-08-01
checkpoint: support lazy migration

adrianreber:
With the help of userfaultfd CRIU supports lazy migration. Lazy
migration means that memory pages are only transferred from the
migration source to the migration destination on page fault.

This enables to reduce the downtime during process or container
migration to a minimum as the memory does not need to be transferred
during migration.

Lazy migration currently depends on userfaultfd being available on the
current Linux kernel and if the used CRIU version supports lazy
migration. Both dependencies can be checked by querying CRIU via RPC if
the lazy migration feature is available. Using feature checking instead
of version comparison enables runC to use CRIU features from the
criu-dev branch. This way the user can decide if lazy migration should
be available by choosing the right kernel and CRIU branch.

To use lazy migration the CRIU process during dump needs to dump
everything besides the memory pages and then it opens a network port
waiting for remote page fault requests:

```
 # runc checkpoint httpd --lazy-pages --page-server 0.0.0.0:27 \
  --status-fd /tmp/postcopy-pipe
```

In this example CRIU will hang/wait once it has opened the network port
and wait for network connection. As runC waits for CRIU to finish it
will also hang until the lazy migration has finished. To know when the
restore on the destination side can start the '--status-fd' parameter is
used:

```
 #️ runc checkpoint --help | grep status
  --status-fd value   criu writes \0 to this FD once lazy-pages is ready
```

The parameter '--status-fd' is directly from CRIU and this way the
process outside of runC which controls the migration knows exactly when
to transfer the checkpoint (without memory pages) to the destination and
that the restore can be started.

On the destination side it is necessary to start CRIU in 'lazy-pages'
mode like this:

```
 # criu lazy-pages --page-server --address 192.168.122.3 --port 27 \
  -D checkpoint
```

and tell runC to do a lazy restore:

```
 # runc restore -d --image-path checkpoint --work-path checkpoint \
  --lazy-pages httpd
```

If both processes on the restore side have the same working directory
'criu lazy-pages' creates a unix domain socket where it waits for
requests from the actual restore. runC starts CRIU restore in lazy
restore mode and talks to 'criu lazy-pages' that it wants to restore
memory pages on demand. CRIU continues to restore the process and once
the process is running and accesses the first non-existing memory page
the 'criu lazy-pages' server will request the page from the source
system. Thus all pages from the source system will be transferred to the
destination system. Once all pages have been transferred runC on the
source system will end and the container will have finished migration.

This can also be combined with CRIU's pre-copy support. The combination
of pre-copy and post-copy (lazy migration) provides the possibility to
migrate containers with minimal downtimes.

Some additional background about post-copy migration can be found in
these articles:

 https://lisas.de/~adrian/?p=1253
 https://lisas.de/~adrian/?p=1183


adrianreber:
This includes the same commits as #1535 as it needs them. The only new commits are the two latest commits.

adrianreber:
@rppt: FYI

adrianreber:
Updated with the review results of #1535

adrianreber:
@avagin: Do you have any comments on this?

crosbymichael:
@kolyshkin can you also take a look at this change?  Thanks!

avagin:
@adrianreber we need a test case for this.  How it will be integrated with phaul? (https://github.com/xemul/criu/tree/criu-dev/phaul)

avagin:
The lazy migration feature is in a criu development branch and we are going to release it in CRIU 3.4 (Sep 2017).

adrianreber:
@avagin, I was also thinking about the test case but as the container is not totally destroyed before complete memory migration my first attempts failed as it was not possible to restore the same container with the same name. I will try to change the name, that might work. Let me try to add a working test-case for lazy migration into runC. I will update this PR.

About p.haul: Not that it really belongs here, but p.haul feels kind of abandoned, especially with the difficulties of integrating p.haul and all the container engines in a useful way. So I am also leaning more towards replacing the pre-copy code in runC with CRIU's go migration library. So the main question is p.haul in its current form still alive or not? For me it seems not very alive in its current form.

xemul:
@adrianreber , the py version of the p.haul (that sits in a separate repo) is indeed abandoned. Mostly for the reasons you've mentioned -- too hard to integrate python code with anything else. At the same time the go p.haul, that sits in the criu repo is the place where support for criu live migration features will (well, should) go.

xemul:
@avagin , 3.4 is going to be in August. And lazy pages are thus aimed at 3.5.

adrianreber:
@avagin test case now included

adrianreber:
Rebased after recent breakage. Any reviewers?

libcontainer/container_linux.go
```diff
@@ -758,6 +773,19 @@ func (c *linuxContainer) addMaskPaths(req *criurpc.CriuReq) error {
...
 		req.Opts.ExtMnt = append(req.Opts.ExtMnt, extMnt)
 	}
+	return nil
+}
+
+func waitForCriuLazyServer(r *os.File, status string) error {
+
+	data := make([]byte, 1)
+	count, _ := r.Read(data)
+	logrus.Debugf("%d:%s", count, status)
```
crosbymichael: can you remove these debug statements?

libcontainer/container_linux.go
```diff
@@ -758,6 +773,19 @@ func (c *linuxContainer) addMaskPaths(req *criurpc.CriuReq) error {
 		}
 		req.Opts.ExtMnt = append(req.Opts.ExtMnt, extMnt)
 	}
+	return nil
+}
+
+func waitForCriuLazyServer(r *os.File, status string) error {
+
+	data := make([]byte, 1)
+	count, _ := r.Read(data)
```
crosbymichael: Can you add error handling for various calls in this function?  Read, OpenFile, and Write are all missing

crosbymichael:
LGTM

mrunalp:
LGTM

tests/integration/checkpoint.bats
```diff
@@ -122,3 +122,111 @@ function teardown() {
...
+  # Double fork taken from helpers.bats
+  # We need to start 'runc checkpoint --lazy-pages' in the background,
+  # so we double fork in the shell.
+  (runc --criu "$CRIU" checkpoint --lazy-pages --page-server 0.0.0.0:${port} --status-fd ${lazy_pipe} --work-path ./work-dir --image-path ./image-dir test_busybox & ) &
+  # Sleeping here. This is ugly, but not sure how else to handle it.
+  # The return code of the in the background running runc is needed, if
+  # there is some basic error. If the lazy migration is ready can
+  # be handled by $lazy_pipe. Which probably will always be ready
+  # after sleeping two seconds.
+  sleep 2
```
avagin: It isn't only ugly, it is probably a signal that the interface is not designed properly

avagin: @adrianreber I think we can discuss this interface on LPC

adrianreber: sure

PR #1542, opened 2017-08-02
makefile: enable -buildmode=pie

cyphar:
Go has supported PIC builds for a while now, and given the security
benefits of using PIC binaries we should really enable them. There also
appears to be some indication that non-PIC builds have been interacting
oddly on ppc64le (the linker cannot load some shared libraries), and
using PIC builds appears to solve this problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,14 +29,14 @@ SHELL := $(shell command -v bash 2>/dev/null)
 .DEFAULT: runc
 
 runc: $(SOURCES)
-	$(GO) build $(EXTRA_FLAGS) -ldflags "-X main.gitCommit=${COMMIT} -X main.version=${VERSION} $(EXTRA_LDFLAGS)" -tags "$(BUILDTAGS)" -o runc .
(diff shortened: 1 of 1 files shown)


crosbymichael:
ping @justincormack

cyphar:
As an aside, why are we using `go build -i`? This breaks the stock build on my machine (I'm not sure why this only started happening now).

```
go install runtime/internal/sys: mkdir /usr/lib64/go/1.8/pkg/linux_amd64_shared: permission denied                                       
make: *** [Makefile:30: runc] Error 1
```

crosbymichael:
@cyphar not sure when that was added.  I don't remember it at all

crosbymichael:
LGTM

justincormack:
LGTM. Have been building it like that for LinuxKit for some months now without issue (also as static pie)

cyphar:
@justincormack Yeah, we've been building it like this with openSUSE as well for a while and it works there as well.

cyphar:
Rebased on top of #1555. PTAL.

crosbymichael:
LGTM

hqhq:
LGTM

PR #1550, opened 2017-08-07
Backward compatability for InitProcessStartTime and Swappiness

f4tq:
We have a project that embeds runc as a library.

Recent commits post v1.0.0-rc3 broke the ability to read 1000s of CoreOS and Ubuntu hosts we have that rely on earlier versions of runc by way of Docker.  

This is a simple PR which allow the older default values '-1' for swappiness, and a string rep of a uint64 to be read using mainline runc.

Docker on CoreOS and Ubuntu all use either runc v1.0.0-rc2 or v1.0.0-rc3. 

I've provided tests of course.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -187,7 +187,7 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 	if cgroup.Resources.MemorySwappiness == nil || int64(*cgroup.Resources.MemorySwappiness) == -1 {
 		return nil
 	} else if *cgroup.Resources.MemorySwappiness <= 100 {
-		if err := writeFile(path, "memory.swappiness", strconv.FormatUint(*cgroup.Resources.MemorySwappiness, 10)); err != nil {
+		if err := writeFile(path, "memory.swappiness", strconv.FormatUint(*(*uint64)(cgroup.Resources.MemorySwappiness), 10)); err != nil {
 			return err
 		}
 	} else {
--- a/libcontainer/cgroups/fs/memory_test.go
+++ b/libcontainer/cgroups/fs/memory_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
 const (
@@ -236,7 +237,7 @@ func TestMemorySetMemorySwappinessDefault(t *testing.T) {
 	defer helper.cleanup()
 
 	swappinessBefore := 60 //default is 60
-	swappinessAfter := uint64(0)
+	swappinessAfter := configs.Swappiness(0)
 
 	helper.writeFileContents(map[string]string{
 		"memory.swappiness": strconv.Itoa(swappinessBefore),
@@ -252,7 +253,7 @@ func TestMemorySetMemorySwappinessDefault(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to parse memory.swappiness - %s", err)
 	}
-	if value != swappinessAfter {
+	if value != uint64(swappinessAfter) {
 		t.Fatalf("Got the wrong value (%d), set memory.swappiness = %d failed.", value, swappinessAfter)
 	}
 }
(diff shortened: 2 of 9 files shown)


cyphar:
Is it not possible for you to match the runc versions that you're using as a library with the version that Docker is using? This patch adds quite a lot of hacks that I'm not really willing to accept the maintainence burden of, especially since we don't provide any guarantees about our internal state represenation.

f4tq:
Hacks is strong word isn't it?  

We have a tool spanning a bunch of environments that have different versions of CoreOS/Ubuntu/runc.  The tool treats runc as a library.  The only reason that's not possible is due to the type switch that happened just prior to this PR.
  
Unfortunately, docker relies on various versions of rc-xx versions of runc across these environments
.  
All this PR does is tolerate the type switch that was made to structures. 

I can always fork but I thought I'd contribute something useful in case someone else wants to treat runc as a library...

cyphar:
> Hacks is strong word isn't it?

Sorry, it wasn't meant to be a value judgement. But having a bunch of wrapper types to trick the JSON library into treating certain values as strings (or `int64`) when they're not feels like a bit of a hack to me. It's a somewhat elegant hack, but a hack nonetheless.

> Unfortunately, docker relies on various versions of rc-xx versions of runc across these environments

Yeah, it's unfortunate that Docker doesn't pin proper releases (though it's also our fault for not doing enough releases). For what it's worth, Docker's runc version is also a fork (https://github.com/docker/runc). Note that on paper, using the same version of `runc` as a library on machines that have other `runc` versions concurrently modifying the same state directory, should never have worked in the first place (though I'm sure that doesn't help you at all).

I'm not going to suggest that you build with different library versions for each of those environments (though that is what I would recommend, but I understand why that probably wouldn't work for you).

> I can always fork but I thought I'd contribute something useful in case someone else wants to treat runc as a library...

People using `libcontainer` as a library is something that we want to support more of (I know several projects that use `libcontainer` directly). Unfortunately there is a _lot_ of legacy code and other work going on which makes it difficult for us to commit to saying that "vX will be compatible with state concurrently written by vY". Is it not possible for you to vendor this project at a point when it worked for you -- or are there other changes that you need to include as well?

As an example of future plans, we (or more accurately, "I") have a plan to completely switch the configuration API from `libcontainer/config.Config` to OCI configurations (for the internal interfaces) as well as switching the state information to only include the OCI configuration -- to remove a bunch of the lossy conversions and legacy code lying around. Those changes will be fundamentally backwards incompatible if you're running `runc` with Docker that has an outdated version. Given those plans, my inkling is that fixing this problem temporarily is not going to have any net long-term benefits aside from potentially making maintenance harder.

Don't think I'm not grateful that you want to contribute (I am, and appreciate that you wanted to contribute before you just forked the project), it's just that there are other considerations at play. I appreciate that this PR won't break any current configurations or setups (and in fact will fix your setup), but there is a non-zero maintenance burden for every piece of code merged into a project.

But maybe some other other maintainers have an alternate opinion, and we can merge this now and then revert it in later versions. I'm willing to be overruled. :smile_cat:

f4tq:
I appreciate your thoughts on this.
FWIW, I'm simply trying to read the contents of the configs, not run anything with runc.  I didn't want to reinvent all the work that was done in runc transforming the json into go structures.

It may be helpful for you to know that the tool we've built uses runc (as a library) to ingest the json state across many installations/clusters of heterogenous mesos, kubernetes and systemd-nspawn systems running related containers so we can do maintenance.    

Once such case is to determine process trees emanating from a given host's containers regardless of origin (mesos,kube,etc) but with regard to quality of service they may need to provide.  We then map their collective listening ports and existing connections so as to create SYN blocks (iptables) that induce various schedulers (like marathon etc) to route traffic away from these hosts while allowing existing work to complete.  It's been working great...

I understand the challenges you're facing and I understand you have many concerns.  

I think the fact that runc produces json reps of state on the filesystem is great and very useful.  

It'd be great if that had carried some weight as an interface with the world...

I hope you change your mind.
Over & out,
Chris

crosbymichael:
@f4tq 

What we ended up doing with docker is writing a migration step for the spec and runc state file on boot of the new version.

It has custom marshal/unmarshal functions that we read in the spec/state and then write it back with the correct types so that the new version of runc can correctly read and manage the container again after an upgrade.  

You can see our code here to do this for specific docker versions:

https://github.com/crosbymichael/upgrade

mrunalp:
We can guarantee config.json to be stable after 1.0. Closing this after discussion with other maintainers.

PR #1551, opened 2017-08-08
fix panic when Linux is nil

crosbymichael:
Linux is not always not nil.
If Linux is nil, panic will occur.

Closes #1449

This is a rebase of #1449

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -184,20 +184,6 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	}
 
 	exists := false
-	if config.RootPropagation, exists = mountPropagationMapping[spec.Linux.RootfsPropagation]; !exists {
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM

libcontainer/specconv/spec_linux.go
```diff
@@ -566,53 +569,54 @@ func createDevices(spec *specs.Spec, config *configs.Config) error {
...
 		}
-		config.Devices = append(config.Devices, device)
 	}
 	return nil
 }
 
 func setupUserNamespace(spec *specs.Spec, config *configs.Config) error {
-	if len(spec.Linux.UIDMappings) == 0 {
-		return nil
-	}
```
dqminh: this removal seems unintended. i think if spec.Linux is nil here we shoud just bail to be the same as the old code.

crosbymichael: The nil check should be on 613 now

crosbymichael: nvm, fixed it

cyphar:
LGTM.

hqhq:
LGTM

PR #1554, opened 2017-08-11
release: import umoci's release.sh script

cyphar:
This script is far easier to use than the previous `make release`
target, not to mention that it also automatically signs all of the
artefacts and makes everything really easy to do for maintainers.

[I'm going to re-release all of our previous releases using this script.]

Ref: openSUSE/umoci#163

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,4 @@ before_install:
 script:
   - git-validation -run DCO,short-subject -v
   - make BUILDTAGS="${BUILDTAGS}"
-  - make BUILDTAGS="${BUILDTAGS}" clean validate test
+  - make BUILDTAGS="${BUILDTAGS}" clean ci
--- a/Makefile
+++ b/Makefile
@@ -43,31 +43,7 @@ static: $(SOURCES)
 	CGO_ENABLED=1 $(GO) build $(EXTRA_FLAGS) -tags "$(BUILDTAGS) cgo static_build" -ldflags "-w -extldflags -static -X main.gitCommit=${COMMIT} -X main.version=${VERSION} $(EXTRA_LDFLAGS)" -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
 
 release:
-	@flag_list=(seccomp selinux apparmor static); \
-	unset expression; \
-	for flag in "$${flag_list[@]}"; do \
-		expression+="' '{'',$${flag}}"; \
-	done; \
-	eval profile_list=("$$expression"); \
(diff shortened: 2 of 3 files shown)


cyphar:
In case you're wondering what the final release artifacts would look like for our last release, here they are:

```
% tree release
release
└── 1.0.0-rc4
    ├── runc.amd64
    ├── runc.amd64.asc
    ├── runc.sha256sum
    ├── runc.tar.xz
    └── runc.tar.xz.asc

1 directory, 5 files
```

The really nice thing is that it also provides a signed version of the source code (which is great for openSUSE because OBS supports verifying PGP signatures of source code automatically).

hqhq:
This new script is much nicer but the former `make release` was supposed to build runc with all combinations of build flags, don't know if there are any users depending on that.

cyphar:
@hqhq I don't think anyone actually wants that, I've personally never used it (nor have I ever actually published them either). Since we're statically compiling the binaries for release, we might as well enable everything. With the `ambient` stuff maybe there was an argument to not always including everything, but now-a-days all of the features are still optional at run-time.

hqhq:
@cyphar It's proposed in https://github.com/opencontainers/runc/issues/899 /cc @crosbymichael should we keep building runc with different build configurations when `make release`?

cyphar:
Yeah, and I agreed to it in the past as well. I think I've changed my mind on this topic, after having published several releases.

crosbymichael:
I think its fine to have a single build now that most of the features are switched on via the spec.

dqminh:
Im +1 for just doing a single build with all tags.

crosbymichael:
And we expect distros that ship only selinux or apparmor to produce their own builds with only the features that they support right?  

This is more of our reference build of `runc`.

hqhq:
@cyphar OK, I'm +1 for single build as well, needs rebase.

cyphar:
> And we expect distros that ship only selinux or apparmor to produce their own builds with only the features that they support right?

Yeah, distros should generally be building from source (and most importantly, not statically) so the binaries we ship aren't really relevant to them. The nice thing about `release.sh` for distributions is that it will generate signed archives of the source code (which provides better provenance than just pulling an unsigned tag from GitHub -- though I do sign my tags as well).

cyphar:
Rebased.

hqhq:
LGTM

crosbymichael:
LGTM

cyphar:
/me rebuilds all of the latest releases with the new script.

PR #1555, opened 2017-08-13
makefile: drop usage of --install

cyphar:
The "go build -i" invocation may slightly help with incremental
recompilation, but it will cause builds to fail if $GOROOT is not
writeable by the current user. While this does appear to work sometimes,
it's a concern for external build systems where "-i" causes build errors
for no real gain.

Given the size of the runc project, --install is not really giving us
much anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 	    localtest localunittest localintegration \
 	    test unittest integration
 
+GO := go
+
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

crosbymichael:
LGTM

PR #1558, opened 2017-08-15
Update state after update

hqhq:
state.json should be a reflection of the container's
realtime state, including resource configurations,
so we should update state.json after updating container
resources.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -186,8 +186,17 @@ func (c *linuxContainer) Set(config configs.Config) error {
 	if status == Stopped {
 		return newGenericError(fmt.Errorf("container not running"), ContainerNotRunning)
 	}
+	if err := c.cgroupManager.Set(&config); err != nil {
+		// Set configs back
+		if err2 := c.cgroupManager.Set(c.config); err2 != nil {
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

libcontainer/container_linux.go
```diff
@@ -186,8 +186,17 @@ func (c *linuxContainer) Set(config configs.Config) error {
 	if status == Stopped {
 		return newGenericError(fmt.Errorf("container not running"), ContainerNotRunning)
 	}
+	if err := c.cgroupManager.Set(&config); err != nil {
+		// Set configs back
+		if err2 := c.cgroupManager.Set(c.config); err2 != nil {
```
kolyshkin, line 191: I am not sure about external users of libcontainer, but runc's own `update.go` modifies container config in place, meaning `&config` and `c.config` is exactly the same thing, meaning this "rollback" does not make sense.

@hqhq do you remember why you introduced it?

hqhq: @kolyshkin Libcontainer was supposed to be used without runc, so the config parameter can be different from c.config, if I remember right, before I added `runc update`, docker was using this `Set` function of libcontainer directly for `docker update`, this rollback was valid at that time.

PR #1559, opened 2017-08-16
fix panic when Linux is nil for rootless case

Mashimiao:
congfig.Sysctl setting is duplicated.
when contianer is rootless and Linux is nil, runc will panic.


cyphar:
LGTM.

crosbymichael:
LGTM

PR #1560, opened 2017-08-17
Check error return values

tklauser:
Both tty.resize and notifySocket.setupSocket return an error which isn't
handled in the caller. Fix this and either log or propagate the errors.

Found using https://github.com/mvdan/unparam


crosbymichael:
LGTM

cyphar:
LGTM.

PR #1561, opened 2017-08-17
Add AutoDedup option to CriuOpts

nseps:
Memory image deduplication, very useful for incremental dumps.

See: https://criu.org/Memory_images_deduplication

Signed-off-by: Nikolas Sepos <nikolas.sepos@gmail.com>


crosbymichael:
Is there are newer version of criu that we should bump in our dependency list to take advantage of this?

dqminh:
@crosbymichael the current criu we tracked in dockerfile should already have this feature. This PR however doesn't enable autodedup anywhere though, just make it possible to have the option. Should we also add a flag to enable this in checkpoint/restore command @nseps ?

nseps:
@dqminh Currently i'm just using libcontainer directly from golang. That is why this PR just makes the option available. I plan on following this with "auto-dedup" flag for runc but feel free to add it :)

nseps:
Now that i think of it, maybe it would be better if the runc flag was part of this PR. What do you think?

crosbymichael:
@nseps ya, that works if you don't mind adding it

dqminh:
LGTM

crosbymichael:
LGTM

Thanks!

PR #1562, opened 2017-08-18
nsenter: improve namespace creation and SELinux IPC handling

cyphar:
nsenter: move namespace creation after userns creation

Technically, this change should not be necessary, as the kernel
documentation claims that if you call clone(flags|CLONE_NEWUSER), the
new user namespace will be the owner of all other namespaces created in
@ flags. Unfortunately this isn't always the case, due to various
additional semantics and kernel bugs.

One particular instance is SELinux, which acts very strangely towards
the IPC namespace and mqueue. If you unshare the IPC namespace *before*
you map a user in the user namespace, the IPC namespace's internal
kern-mount for mqueue will be labelled incorrectly and the container
won't be able to access it. The only way of solving this is to unshare
IPC *after* the user has been mapped and we have changed to that user.
I've also heard of this happening to the NET namespace while talking to
some LXC folks, though I haven't personally seen that issue.

This change matches our handling of user namespaces to be the same as
how LXC handles these problems.

Closes #959
Closes #975

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -811,24 +811,29 @@ void nsexec(void)
 				join_namespaces(config.namespaces);
 
 			/*
-			 * Unshare all of the namespaces. Now, it should be noted that this
-			 * ordering might break in the future (especially with rootless
-			 * containers). But for now, it's not possible to split this into
-			 * CLONE_NEWUSER + [the rest] because of some RHEL SELinux issues.
-			 *
-			 * Note that we don't merge this with clone() because there were
-			 * some old kernel versions where clone(CLONE_PARENT | CLONE_NEWPID)
-			 * was broken, so we'll just do it the long way anyway.
-			 */
-			if (unshare(config.cloneflags) < 0)
-				bail("failed to unshare namespaces");
-
-			/*
 			 * Deal with user namespaces first. They are quite special, as they
 			 * affect our ability to unshare other namespaces and are used as
 			 * context for privilege checks.
+			 *
+			 * We don't unshare all namespaces in one go. The reason for this
+			 * is that, while the kernel documentation may claim otherwise,
+			 * there are certain cases where unsharing all namespaces at once
+			 * will result in namespace objects being owned incorrectly.
+			 * Ideally we should just fix these kernel bugs, but it's better to
+			 * be safe than sorry, and fix them separately.
+			 *
+			 * A specific case of this is that the SELinux label of the
+			 * internal kern-mount that mqueue uses will be incorrect if the
+			 * UTS namespace is cloned before the USER namespace is mapped.
+			 * I've also heard of similar problems with the network namespace
+			 * in some scenarios. This also mirrors how LXC deals with this
+			 * problem.
 			 */
 			if (config.cloneflags & CLONE_NEWUSER) {
+				if (unshare(CLONE_NEWUSER) < 0)
+					bail("failed to unshare user namespace");
+				config.cloneflags &= ~CLONE_NEWUSER;
+
 				/*
 				 * We don't have the privileges to do any mapping here (see the
 				 * clone_parent rant). So signal our parent to hook us up.
@@ -854,7 +859,20 @@ void nsexec(void)
 					if (prctl(PR_SET_DUMPABLE, 0, 0, 0, 0) < 0)
 						bail("failed to set process as dumpable");
 				}
+
+				/* Become root in the namespace proper. */
+				if (setresuid(0, 0, 0) < 0)
+					bail("failed to become root in user namespace");
 			}
+
+			/*
+			 * Unshare all of the namespaces. Note that we don't merge this
+			 * with clone() because there were some old kernel versions where
+			 * clone(CLONE_PARENT | CLONE_NEWPID) was broken, so we'll just do
+			 * it the long way.
+			 */
+			if (unshare(config.cloneflags) < 0)
+				bail("failed to unshare namespaces");
 
 			/*
 			 * TODO: What about non-namespace clone flags that we're dropping here?


crosbymichael:
LGTM

libcontainer/nsenter/nsexec.c
```diff
@@ -890,6 +899,15 @@ void nsexec(void)
 				bail("failed to set uid to root");
 
 			/*
+			 * At this point we are now able to do the previously deferred
+			 * CLONE_NEWIPC (see the unshare(config.cloneflags) comment above).
+			 */
+			if (leftover_cloneflags) {
+				if (unshare(leftover_cloneflags) < 0)
+					bail("failed to unshare leftover namespaces");
```
rhvgoyal: How `unshare(CLONE_NEWUSER | CLONE_NEWIPC) is different from `unshare(CLONE_NEWUSER)` followed by `unshare(CLONE_NEWIPC)`. IIUC, in both the cases new user namespace will be owner of IPC namespace.

mrunalp:
LGTM

cyphar:
**[do not merge yet]**

I think the `uid_map` handling probably needs to be corrected.

libcontainer/nsenter/nsexec.c
cyphar: While I agree that that is true on paper, it isn't. There's a kernel bug (as discussed in #959) which results in SELinux labels (on `/dev/mqueue`) not being settable in a user namespace unless the IPC namespace was created after the user namespace was set up fully (including the `uid_map` and `setuid` stuff).

rhvgoyal: That issue really does not give exact technical details. I am looking at kernel code, and owner of new ipc namespace is going to be newly created user namespace. If that's the case, then we should be able to just do.

unshare(CLONE_NEWUSER | CLONE_NEWIPC)
set_uid_gid_map
switch to root inside container
mount_dev_mqueue_and_label.

I am wondering why above workflow is not working and if it is a kernel bug which needs fixing.

rhvgoyal: I tried reproducing problem on fedora 26 and I can reproduce it. I see following error message in journal.

`SELinux: mount invalid.  Same superblock, different security settings for (dev mqueue, type mqueue)`

rhvgoyal: Looking at strace() output. Looks like first we tried selinux context mount of mqueue and that failed. I think due to above kernel message. And the we tried a non-context mount and tried lsexattr() and that lsetxattr failed. 

29777 mount("mqueue", "/root/runc-testing/rootfs/dev/mqueue", "mqueue", MS_NOSUID|MS_NODEV|MS_NOEXEC, "context=\"system_u:object_r:svirt"... <unfinished ...>
29777 <... mount resumed> )             = -1 EINVAL (Invalid argument)

29777 mount("mqueue", "/root/runc-testing/rootfs/dev/mqueue", "mqueue", MS_NOSUID|MS_NODEV|MS_NOEXEC, NULL <unfinished ...>
29777 <... mount resumed> )             = 0

29777 lsetxattr("/root/runc-testing/rootfs/dev/mqueue", "security.selinux", "system_u:object_r:svirt_sandbox_"..., 47, 0) = -1 EPERM (Operation not permitted)

rhvgoyal: I think first failure happens because new IPC namespace was created, it creates an internal mount point of mqueue fs. And when user space tries to mount it with context=, it fails because it does not match the security context provided when kernel created internal mount point. IIUC, following kernel code is creating that internal mount point.

mq_init_ns() {
      ns->mq_mnt = kern_mount_data(&mqueue_fs_type, ns);
}

rhvgoyal: More debugging. I think following check in selinux fails.

selinux_inode_setxattr() {
          if (!inode_owner_or_capable(inode)) {
          }
}

This inode should belong to /rootfs/dev/mqueue (one belonging mqueuefs). IIUC, calling thread is the one which is already inside container and has effective fsuid=1000. But interestingly inod->i_uid is 0. And 0 is not mapped inside container, so kuid_has_mapping() check fails too.

So question is, how did inode->i_uid is 0. IIUC, it is the fsuid=1000 which created this mqueue directory and mqueue mount point. I would think that inode->i_uid should been 1000 instead? What am I missing?

cyphar: @rhvgoyal I'll admit I never went through debugging this particular issue, but I agree with your conclusion. It looks like the reason why it's done that way is because of the `mq_open` (and related `mq_*` syscalls) that don't go through the VFS but need access to the `mqueue` mountpoint -- hence the mount is created each time a new IPC namespace is created.

More importantly, it looks like [we actually cannot ever set mount options if we are not in a user namespace](http://elixir.free-electrons.com/linux/latest/source/security/selinux/hooks.c#L840) -- `mqueue` is not whitelisted. So `xattr` is the only way to set them, and you need to pass [this `uid`-based check](http://elixir.free-electrons.com/linux/latest/source/security/selinux/hooks.c#L3170). And in order for the `uid`-based check to pass you need to have your user namespace set up properly (namely `kuid_has_mapping` needs to work).

cyphar: @rhvgoyal Ah, you saw the same thing as me (I was debugging in parallel :smile_cat:).

> This inode should belong to /rootfs/dev/mqueue (one belonging mqueuefs). IIUC, calling thread is the one which is already inside container and has effective fsuid=1000. But interestingly inode->i_uid is 0.

This will only be true in rootless containers. In a container that uses user namespaces but is running as root, the `unshare(...)` will happen with an `fsuid` of `0` (which will carry on to `mqueue`).

> And 0 is not mapped inside container, so kuid_has_mapping() check fails too.

Yeah. I think making `kuid_has_mapping` pass is the only way to be sure that things work out the way we want. But we could try something like `setfsuid(rootuid)` which should not affect anything else.

rhvgoyal: Hmm..., I think it ties back to clone(CLONE_NEWUSER | CLONE_NEWIPC) call. This is called by process with fsuid=0. And kernel creates new ipc mount namespace, and also creates an internal mount point of mqueuefs. That in turn instantiates root inode and assigns i_uid from the calling thread. 
mqueue_get_inode() {
         inode->i_uid = current_fsuid();
        inode->i_gid = current_fsgid();
}

And that's how root inode of mqueuefs gets i_uid=0. When container process later tries to change the label, it fails as host uid 0 is not mapped inside container. 

So now I atleast understand the problem.

And Mrunal's patch is helping because unshare(CLONE_NEWIPC) is called by container process with fsuid=1000. That means mqeueufs root inode will get i_uid=1000 and container process will have the privileges to change selinux label.

cyphar: @rhvgoyal We can set our `fsuid` though (see my above comment). I'm working on another patch which try to do it that way.

mrunalp:
@cyphar Any update?

cyphar:
@mrunalp I've just pushed the new patch now. I have code to handle the full `uid_map` schema, as well as some basic code that uses `setfsuid` (which on paper should solve the problem) but I don't have a machine to test this on.

I'd be interested to know whether this fixes the problem. However I had some discussions with @brauner at OSS, and it looks like trying to work around these sorts of bugs is a waste of time and we should always do a `CLONE_NEWUSER` and mapping before doing all other unshares.

mrunalp:
I'll check this out. Thanks!

crosbymichael:
Can you look into the ci failure on this? 

```
--- FAIL: TestExecInUserns (0.32s)
	utils_test.go:51: execin_test.go:571: unexpected error: container_linux.go:295: starting container process caused "process_linux.go:302: running exec setns process for init caused \"signal: segmentation fault (core dumped)\""
```

mrunalp:
@cyphar ping

cyphar:
I think I'm going to rewrite this one quite significantly, to just delay *all* namespace unsharing until after user namespaces are set up. This is what LXC does and I think they're right about not trusting that the kernel does the right thing in all cases here.

cyphar:
I've rebased this to just delay unsharing of all namespaces. PTAL.

cyphar:
Test failure is because of spec validator.

dqminh:
LGTM.

I think spec validator should be fixed now. Do you want to rebase ?

cyphar:
Yes, I will rebase.

/ping @mrunalp to verify that this fixes the issue originally reported.

cyphar:
This has been rebased. Ping @mrunalp and the rest of @opencontainers/runc-maintainers.

mrunalp:
I'll test this out and get back. Thanks!

iavael:
@mrunalp any updates here?

giuseppe:
@rhatdan I confirm this solves the issue we have seen.

LGTM

giuseppe:
could this be finally merged?

crosbymichael:
LGTM

rhatdan:
LGTM
This fixes an issue that is causing us to have to disable SELinux when using UserNS, not ideal in podman.

mrunalp:
Reviewing right now.

cyphar:
:tada:

PR #1564, opened 2017-08-22
Fix cgroup: add check task exists after writing freezer.state

moypray:
Description:
  Example: we start a container which just run `sleep 3` command in
container. After 3 seconds, if we try to pause it (container exit
process in shim/containerd is not finished), the cgroup file exists.

So we could update the cgroup file successfully, but the container
process `sleep 3` has already exited.

Then we got an 'Exited' container with paused state in docker.

So I think we should check if the `tasks` is empty or not after update
`freezer.state`.

### DIFF
--- a/libcontainer/cgroups/fs/freezer.go
+++ b/libcontainer/cgroups/fs/freezer.go
@@ -43,6 +43,14 @@ func (s *FreezerGroup) Set(path string, cgroup *configs.Cgroup) error {
 			}
 			time.Sleep(1 * time.Millisecond)
 		}
+		tasks, err := readFile(path, "tasks")
+		if err != nil {
+			return err
+		}
+		if strings.TrimSpace(tasks) == "" {
+			return fmt.Errorf("Error: no tasks running in freeze cgroup")
+		}
+
 	case configs.Undefined:
 		return nil
 	default:


hqhq:
The process can still exit after you check `tasks` file so you didn't actually fix any issue here. I think it's more like an event synchronization issue which should not be fixed in runc.

moypray:
@hqhq from my view, if a process is running and freezed by freezer, I think it could not be exited, because it could not be scheduled in kernel after freezing it.
I am not sure if the process exit, but parent do not call `wait`, could it be visible in cgroup tasks? If could not,  I think the fix should be working.

If something was wrong ,correct me.

hqhq:
`Set` can both freeze and resume.

hqhq:
@moypray I understand that people want to see a paused container after `runc pause` succeed instead of an exited container, but we should not do it in libcontainer, libcontainer is supposed to honor cgroup APIs, some users use it without runc or docker, and freezer cgroup allows to freeze a container without any processes, so I opened a new PR https://github.com/opencontainers/runc/pull/1565 to do this check in `pause.go`, I think it also fixes your issue, and probably some docker issues.

moypray:
OK, this is fine, close this one.

PR #1565, opened 2017-08-22
Check container process after pause succeed

hqhq:

### DIFF
--- a/pause.go
+++ b/pause.go
@@ -2,7 +2,11 @@
 
 package main
 
-import "github.com/urfave/cli"
+import (
+	"fmt"
+
+	"github.com/urfave/cli"
+)
 
 var pauseCommand = cli.Command{
 	Name:  "pause",
@@ -24,6 +28,20 @@ Use runc list to identiy instances of containers and their current status.`,
 		}
 		if err := container.Pause(); err != nil {
 			return err
+		}
+		processes, err := container.Processes()
+		if err != nil {
+			return err
+		}
+		// It's possible that container is running when we checked the current
+		// status in `container.Pause()` then container process exited, and we
+		// paused container successfully before `runc delete` had deleted cgroup
+		// directory.
+		// And it's also wanted that after `runc pause` succeed, we should get
+		// a paused container instead of an exited container, so check container
+		// processes after `container.Pause()` succeed.
+		if len(processes) == 0 {
+			return fmt.Errorf("container process already exited.")
 		}
 
 		return nil


cyphar:
I'm not sure I understand the race you're describing in your comment. Is it something like

```
% runc create blah
% runc start blah
% runc delete blah &
% runc pause blah # this completes before the operation above?
```

On the one hand, I'm not entirely convinced that's a bug, but maybe there's some other state-related issue that occurs? Or is this patch only about returning an error in that case?

hqhq:
@cyphar The race condition happens with docker, it's something like:
```
docker run -ti --name foo ubuntu sleep 3s
<after 3 seconds>
docker pause foo
```

It's possible that `docker pause` succeed but you end up with a exited container, because when `runc pause` doing status check, the process is still there, but it exit after that, so pause succeed, and containerd will send pause event to docker first, then send exit event. It'll cause confusion to users as they might make a bad decision since they think the container must be paused if the pause operation succeed.

hqhq:
@cyphar A test case failed in rootless mode, did you see that before? It doesn't seem like related to this PR.

hqhq:
@cyphar And yes this patch only about returning an error in that case, it makes sure if `runc pause` succeed, you'll have a paused container.

cyphar:
@hqhq I've seen this error before, it appears to happen randomly. I've restarted the build, and it passes now.

```
time="2017-08-23T01:27:26Z" level=error msg="bad file descriptor" 
```

I'll open a bug about it.

crosbymichael:
What are you supposed to do with the error from pause?  Technically, pause succeeded and worked.  The error here would be a little weird and will have to write code client side to handle it.  

How do you see clients handling this?  Is it `pause`s responsibility to say that the container exited?

hqhq:
@crosbymichael 

> What are you supposed to do with the error from pause? How do you see clients handling this? 

Clients can handle it the same way as they handle `pause` a not-running container.

> Technically, pause succeeded and worked.

Technically, I think pause didn't succeed and work from runc or docker's perspective in this case (maybe it seems to be succeeded in libcontainer, that's why I did this check in pause.go), when runc and docker pause a container, they actually want to pause the process(es) in the container, because we don't allow them to pause a not-running container, if pause succeeded, they'll think the processes are paused and certainly they can't exit before they resume, and now it is not true.

mrunalp:
I think that this logic is best left to the caller as @crosbymichael technically pause succeeded.

hqhq:
@mrunalp So do you think it makes sense that we tell users a successful pause operation doesn't mean they'll get a paused container (that's all the problem I'm saying the race condition)? I just think it does nothing but confuse if we return true and we know the container is not exist at that time, as I said, pause only succeeded in libcontainer, not in runc's perspective.

kolyshkin:
I think that runc is too low level (and it's not a daemon) and thus upper-level runtimes (like cri-o or containerd) should handle such things. The situation is the same as with, say, `mv` vs `rm`. Imagine that someone runs `mv a b` to rename an existing file, and someone else runs `rm -f b` right after that. It is not `mv`'s responsibility to check that `b` exists after the operation.

PR #1566, opened 2017-08-22
createDevices: don't add device to config.Devices if the device is al…

keloyang:
The devices in config.Devices comes from runc and docker, and they may be duplicate.
such as /dev/random and /dev/urandom.

It may lead to start a container(with user namespace ) failed.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -508,9 +508,18 @@ func stringToDeviceRune(s string) (rune, error) {
 	}
 }
 
+func deviceIsExist(dev []*configs.Device, path string) bool {
+	for _, d := range dev {
+		if d.Path == path {
+			return true
+		}
+	}
+	return false
+}
+
 func createDevices(spec *specs.Spec, config *configs.Config) error {
 	// add whitelisted devices
-	config.Devices = []*configs.Device{
+	defaultDevices := []*configs.Device{
 		{
 			Type:     'c',
 			Path:     "/dev/null",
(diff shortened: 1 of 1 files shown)


libcontainer/specconv/spec_linux.go
```diff
@@ -508,6 +508,15 @@ func stringToDeviceRune(s string) (rune, error) {
 	}
 }
 
+func deviceIsExist(dev []*configs.Device, path string) bool {
+	for _, d := range dev {
+		if d.Path == path {
```
cyphar: I'd prefer if this check did something more like `reflect.DeepEqual(d, <the device being checked in the caller>)`. My main concern about this patch is whether we're going to miss something if a user added additional permissions to the device that we implicitly add.

However, I actually don't like that we have pre-configured devices *at all* because it feels like working around a spec issue...

keloyang: We should use  user's additional devices at first , if not duplicate, than add the pre-configured device.
@cyphar WDYT ?  thanks for your suggestion, I have changed the code.

hqhq: @keloyang You didn't get @cyphar 's point, you should do more check than just `d.Path` to confirm the device is exist, and the critical elements for a device should be `type`, `major` and `minor`.

BTW, I agree we should not have duplicated devices for runc to create, but why would this fail with user ns enabled? Doesn't runc just do bind mount and the former mount point would be overlaid by the later one?

mrunalp:
@keloyang Can you update the PR per the suggestions made?

PR #1567, opened 2017-08-22
bindMountDeviceNode: skip if the source of a bind mount does not exist

keloyang:
Some device like /dev/fuse is define in docker side as default, but /dev/fuse may not exist in os,  so we don't start a user namespace container in a os which don't has /dev/fuse.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -475,7 +475,17 @@ func createDevices(config *configs.Config) error {
 	return nil
 }
 
+func exist(name string) bool {
+	_, err := os.Stat(name)
+	return err == nil
+}
+
 func bindMountDeviceNode(dest string, node *configs.Device) error {
+	if exist(node.Path) {
+		// skip if the source of a bind mount does not exist to avoid a failed starting
+		return nil
+	}
+
 	f, err := os.Create(dest)
 	if err != nil && !os.IsExist(err) {
 		return err


libcontainer/rootfs_linux.go
```diff
@@ -475,7 +475,17 @@ func createDevices(config *configs.Config) error {
 	return nil
 }
 
+func isNotExist(name string) bool {
+	_, err := os.Stat(name)
+	return err != nil && !os.IsExist(err)
+}
+
 func bindMountDeviceNode(dest string, node *configs.Device) error {
+	if isNotExist(node.Path) {
```
crosbymichael: I would make this simpler and have the function be `exits(path)`

```go
func exits(name string) bool {
    _, err := os.Stat(name)
    return err == nil 
}
if !exists(node.Path) {
...
}
```

keloyang:
@crosbymichael updated, thanks for your review.

cyphar:
I don't really like this. If a user tells us to mount something that doesn't exist, we should always error out. Otherwise we're just going to confuse them by creating a container that doesn't have the mount they want. Also I believe this is a violation of the runtime-spec.

keloyang:
I want to error out at the beginning, but it lead to start a userns container failed if  the host has no /dev/fuse.  Considering this is not fatal, I skip it.
@cyphar are you mean that we should error out and remove /dev/fuse as default in docker?

crosbymichael:
@cyphar 's point makes sense as well as long as we communicate this breaking change to users where they can update their code when they rupdate runc.

hqhq:
I dont' think we should do such thing in runc either, upper level tools should not add such device if the creation would fail somehow.

PR #1569, opened 2017-08-24
init: delay seccomp application as late as possible

cyphar:
This further reduces the number of syscalls that a user needs to enable
in their seccomp profile.

This mirrors the standard_init_linux.go seccomp code, which only applies
seccomp early if NoNewPrivileges is enabled. Otherwise it's done
immediately before execve to reduce the amount of syscalls necessary for
users to enable in their seccomp profiles.

### DIFF
--- a/libcontainer/setns_init_linux.go
+++ b/libcontainer/setns_init_linux.go
@@ -47,7 +47,10 @@ func (l *linuxSetnsInit) Init() error {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


cyphar:
Not sure why GitHub isn't showing this, but the tests have passed on Travis.

crosbymichael:
LGTM

mrunalp:
LGTM

Issue #1557, opened 2017-08-14
--systemd-cgroup broken with memory cgroup enabled


qiancai:
Bisect indicated that commit 3d9074ead33a5c27dc20bb49457c69c6d2ae6b57 breaks --systemd-cgroup.

```
# runc --systemd-cgroup run root
container_linux.go:265: starting container process caused "process_linux.go:284: applying cgroup configuration for process caused \"No such device or address\""
```
Reproducer is below,

```
# runc spec
# Edit config.json to add memory section under resource.
        "linux": {
                "resources": {
                        "devices": [
                                {
                                        "allow": false,
                                        "access": "rwm"
                                }
                        ],
                        "memory": {
                                "limit": 104847600,
                                "reservation": 52428800
                        }

                },
# runc --systemd-cgroup run root
```

@justincormack

PR #1573, opened 2017-08-25
Fix systemd cgroup after memory type changed

hqhq:
Fixes: #1557

I'm not quite sure about the root cause, looks like
systemd still want them to be uint64.


TomSweeneyRedHat:
Assuming happy tests, LGTM.

crosbymichael:
LGTM

cyphar:
LGTM. systemd, amirite?

crosbymichael:
@cyphar yes, urite

PR #1575, opened 2017-08-26
signal: ignore tty.resize errors

### LINKED ISSUE #1571: exec: "bad file descriptor" occurs in tests
This appears to happen on random tests (sometimes in rootless sometimes in normal mode). I started noticing it a week or so ago. The error is basically just `EBADF`. I'll see if I can reproduce it, it seems to only happen on `runc exec`s.

```
not ok 33 mask paths [directory]
# (in test file tests/integration/mask.bats, line 46)
#   `[[ "${output}" == "" ]]' failed
# runc list (status=0):
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/console.sock test_busybox (status=0):
# 
# runc exec test_busybox ls /testdir (status=0):
# time="2017-08-24T07:59:50Z" level=error msg="bad file descriptor" 
# runc list (status=0):
# ID             PID         STATUS      BUNDLE             CREATED                          OWNER
# test_busybox   20467       running     /tmp/busyboxtest   2017-08-24T07:59:50.815627018Z   root
# runc kill test_busybox KILL (status=0):
# 
# runc delete test_busybox (status=0):
# 
```

cyphar:
Fixes a race that occurred very frequently in testing where the tty of
the container may be closed by the time that runc gets to sending
SIGWINCH. This failure mode is not fatal, but it would cause test
failures due to expected outputs not matching. On further review it
appears that the original addition of these checks in 4c5bf649d01e
("Check error return values") was actually not necessary, so partially
revert that change.

The particular failure mode this resolves would manifest as error logs
of the form:

  time="2017-08-24T07:59:50Z" level=error msg="bad file descriptor"

Fixes: 4c5bf649d01e ("Check error return values")
Fixes #1571

### DIFF
--- a/signals.go
+++ b/signals.go
@@ -74,16 +74,15 @@ func (h *signalHandler) forward(process *libcontainer.Process, tty *tty, detach 
 		}
 	}
 
-	// perform the initial tty resize.
-	if err := tty.resize(); err != nil {
-		logrus.Error(err)
-	}
+	// Perform the initial tty resize. Always ignore errors resizing because
+	// stdout might have disappeared (due to races with when SIGHUP is sent).
+	_ = tty.resize()
+	// Handle and forward signals.
 	for s := range h.signals {
 		switch s {
 		case unix.SIGWINCH:
-			if err := tty.resize(); err != nil {
-				logrus.Error(err)
-			}
+			// Ignore errors resizing, as above.
+			_ = tty.resize()
 		case unix.SIGCHLD:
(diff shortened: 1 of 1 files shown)


tklauser:
Thanks @cyphar! Didnt't consider this possibility when submitting the original patch. Your fix LGTM.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1578, opened 2017-08-31
travis: drop shfmt install

cyphar:
It looks like we missed this in 5930d5b42728 ("Remove shfmt"), which was
causing CI to break (since it looks like the repo has moved or something
like that). Since we're no longer using shfmt, drop it completely from
the repo.


dqminh:
LGTM

crosbymichael:
LGTM

PR #1583, opened 2017-09-05
use conditional assignment for BUILDTAGS

lsm5:
This change would allow usage of additional BUILDTAGS while building
rpms. Like so:

BUILDTAGS="seccomp selinux" make all

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>


crosbymichael:
Doesn't this already work today via `make BUILDTAGS='seccomp apparmor'`

lsm5:
ha, stupid me, it does, closing this..

PR #1586, opened 2017-09-07
Apply cgroups earlier

crosbymichael:
This applies cgroups earlier for container creation before the init
process starts running and forking off any additional processes.


dqminh:
LGTM

hqhq:
LGTM @crosbymichael Did you encounter any bug? It won't change anyghing as the internal processes `runc-init` creates would be exited when we do `Apply`, and it might slightly impact resource consumption which might not be expected by users, but I noticed that containerd has moved shim process into container cgroup, so it'll be less a problem when users realized that fact.

PR #1589, opened 2017-09-08
libcontainer: intelrdt: use init() to avoid race condition

xiaochenshen:
This is the follow-up PR of #1279 to fix remaining issues:

Use init() to avoid race condition in IsIntelRdtEnabled().
Add also rename some variables and functions.

### DIFF
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -130,13 +130,28 @@ var (
 	intelRdtRootLock sync.Mutex
 
 	// The flag to indicate if Intel RDT is supported
-	isIntelRdtEnabled bool
+	isEnabled bool
 )
 
 type intelRdtData struct {
 	root   string
 	config *configs.Config
 	pid    int
+}
+
+// Check if Intel RDT is enabled in init()
+func init() {
+	// 1. Check if hardware and kernel support Intel RDT/CAT feature
+	// "cat_l3" flag is set if supported
+	isFlagSet, err := parseCpuInfoFile("/proc/cpuinfo")
+	if !isFlagSet || err != nil {
+		isEnabled = false
+		return
+	}
+
+	// 2. Check if Intel RDT "resource control" filesystem is mounted
+	// The user guarantees to mount the filesystem
+	isEnabled = isIntelRdtMounted()
 }
 
 // Return the mount point path of Intel RDT "resource control" filesysem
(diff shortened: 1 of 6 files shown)


libcontainer/intelrdt/intelrdt.go
```diff
@@ -370,22 +371,21 @@ func WriteIntelRdtTasks(dir string, pid int) error {
 
 // Check if Intel RDT is enabled
 func IsIntelRdtEnabled() bool {
```
cyphar: This function should really just be called `IsEnabled` like `go-selinux` (the package name already tells us that it's `intelrdt`, there's no need to add it to the function name as well).

libcontainer/intelrdt/intelrdt.go
```diff
@@ -370,22 +371,21 @@ func WriteIntelRdtTasks(dir string, pid int) error {
...
-	// We have checked the flag before
-	if isIntelRdtEnabled {
-		return true
-	}
+	// We only need to set the flag once
+	once.Do(func() {
```
cyphar: If we only need to do this once, why not just do it in an `init` and set the flag when the program starts?

xiaochenshen: @cyphar 

Thank you for code review.

Intel RDT/CAT is a hardware-dependent feature. We wish add zero/minimal overhead to the hardware platform which doesn't support this feature. So there are several orthogonal call paths in libcontainer to check the capability:
1. In loadFactory() --> libcontainer.New().
2. In LinuxFactory.Create() and Load().
3. In ConfigValidator.Validate()
4. In libcontainer and intelrdt test cases.

Taking loadFactory() --> libcontainer.New() for example. It calls IsIntelRdtEnabled() to check the capability. If not support, libcontainer.New() will not add intelRdtManager as a factory constructor option. In subsequent container creating, running and destroying life cycle, intelRdtManager handles Intel RDT operations. If intelRdtManager is nil, we will not touch Intel RDT related functionality.

From my understanding, keeping IsIntelRdtEnabled() as a "utility" function is no harm in this case. We could check the capability in outside package without adding intelRdtManager hooks. Please correct me if I am wrong.

Thank you.

cyphar: I don't think you understood my question. I was asking why don't we set the `isIntelRdtEnabled` flag inside an `init` rather than setting it when you do the check (and then just making `IsEnabled` check that flag and not touch it). Something like

```
func init() {
  // code to check if it's enabled
  isIntelRdtEnabled = true // or whatever
}

func IsEnabled() {
  return isIntelRdtEnabled
}
```

So you don't need to do the `sync.Do` stuff.

libcontainer/intelrdt/intelrdt.go
xiaochenshen: @cyphar 

Thank you. I will correct the function name.

libcontainer/intelrdt/intelrdt.go
xiaochenshen: @cyphar 
> package intelrdt
> func init() {

I got it. Thank you for suggestion.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1590, opened 2017-09-08
libcontainer: intelrdt: add update command support

xiaochenshen:
Add runc update command support for Intel RDT/CAT.

for example:
runc update --l3-cache-schema "L3:0=f;1=f" container-id

For more information about Intel Resource Director Technology (RDT)
and Cache Allocation Technology (CAT), please refer to #1279 

### DIFF
--- a/update.go
+++ b/update.go
@@ -9,6 +9,8 @@ import (
 	"strconv"
 
 	"github.com/docker/go-units"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/urfave/cli"
 )
@@ -111,6 +113,10 @@ other options are ignored.
 		cli.IntFlag{
 			Name:  "pids-limit",
 			Usage: "Maximum number of pids allowed in the container",
+		},
+		cli.StringFlag{
+			Name:  "l3-cache-schema",
+			Usage: "The string of Intel RDT/CAT L3 cache schema",
 		},
 	},
 	Action: func(context *cli.Context) error {
@@ -253,6 +259,34 @@ other options are ignored.
 		config.Cgroups.Resources.MemoryReservation = *r.Memory.Reservation
 		config.Cgroups.Resources.MemorySwap = *r.Memory.Swap
 		config.Cgroups.Resources.PidsLimit = r.Pids.Limit
+
+		// Update Intel RDT/CAT
+		if val := context.String("l3-cache-schema"); val != "" {
+			if !intelrdt.IsEnabled() {
+				return fmt.Errorf("Intel RDT: l3 cache schema is not enabled")
+			}
+
+			// If intelRdt is not specified in original configuration, we just don't
+			// Apply() to create intelRdt group or attach tasks for this container.
+			// In update command, we could re-enable through IntelRdtManager.Apply()
+			// and then update intelrdt constraint.
+			if config.IntelRdt == nil {
+				state, err := container.State()
+				if err != nil {
+					return err
+				}
+				config.IntelRdt = &configs.IntelRdt{}
+				intelRdtManager := intelrdt.IntelRdtManager{
+					Config: &config,
+					Id:     container.ID(),
+					Path:   state.IntelRdtPath,
+				}
+				if err := intelRdtManager.Apply(state.InitProcessPid); err != nil {
+					return err
+				}
+			}
+			config.IntelRdt.L3CacheSchema = val
+		}
 
 		return container.Set(config)
 	},
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -213,8 +213,7 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 		newgidmapPath: l.NewgidmapPath,
 		cgroupManager: l.NewCgroupsManager(config.Cgroups, nil),
 	}
-	c.intelRdtManager = nil
-	if intelrdt.IsEnabled() && c.config.IntelRdt != nil {
+	if intelrdt.IsEnabled() {
 		c.intelRdtManager = l.NewIntelRdtManager(config, id, "")
 	}
 	c.state = &stoppedState{c: c}
@@ -256,8 +255,7 @@ func (l *LinuxFactory) Load(id string) (Container, error) {
 	if err := c.refreshState(); err != nil {
 		return nil, err
 	}
-	c.intelRdtManager = nil
-	if intelrdt.IsEnabled() && c.config.IntelRdt != nil {
+	if intelrdt.IsEnabled() {
 		c.intelRdtManager = l.NewIntelRdtManager(&state.Config, id, state.IntelRdtPath)
 	}
 	return c, nil
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -401,6 +401,10 @@ func GetIntelRdtPath(id string) (string, error) {
 
 // Applies Intel RDT configuration to the process with the specified pid
 func (m *IntelRdtManager) Apply(pid int) (err error) {
+	// If intelRdt is not specified in config, we do nothing
+	if m.Config.IntelRdt == nil {
+		return nil
+	}
 	d, err := getIntelRdtData(m.Config, pid)
 	if err != nil && !IsNotFound(err) {
 		return err


update.go
```diff
@@ -254,6 +259,18 @@ other options are ignored.
 		config.Cgroups.Resources.MemorySwap = *r.Memory.Swap
 		config.Cgroups.Resources.PidsLimit = r.Pids.Limit
 
+		if intelrdt.IsEnabled() && config.IntelRdt != nil {
```
crosbymichael: Wouldn't it be better to just set the value and let the libcontainer rtd code validate and return an error if its not enabled?  Right now, it would just fail silently and no one would know if its set or not

xiaochenshen: @crosbymichael 

I have made changes according to your suggestion to avoid silent failure.
There are several cases in error paths:
(1) IsEnabled == false && intelRdt is specified in config
(2) IsEnabled == true && intelRdt is not specified in config
(3) IsEnabled == true && intelRdt is specified in config && update value is invalid

(1): It is handled by validate.Validator(). runc create/run will fail. It never pass to update command.
(2): We'd better return an error that the update command is failed before container.Set(). Just passing through container.Set() can't handle this case:
(3): We just set the value, intelrdt.Set() will return an error that the update value is invalid.

crosbymichael:
You are making things wayyy more complicated than they need to be. 

```go
if val := context.String("l3-cache-schema"); val != "" {
    config.IntelRtd = &configs.IntelRdt{
        L3CacheSchema: val,
    }
}
```

xiaochenshen:
@crosbymichael 

Thank you. I have updated the code according to your suggestion.
And also added error return in front to avoid silent failure passing through container.Set().

update.go
```diff
@@ -254,6 +260,17 @@ other options are ignored.
 		config.Cgroups.Resources.MemorySwap = *r.Memory.Swap
 		config.Cgroups.Resources.PidsLimit = r.Pids.Limit
 
+		if val := context.String("l3-cache-schema"); val != "" {
+			// Avoid silent failure in container.Set()
+			if !intelrdt.IsEnabled() || config.IntelRdt == nil {
```
crosbymichael: Why  would `config.IntelRdt == nil` be a case for an error?

xiaochenshen: @crosbymichael 
>Why would config.IntelRdt == nil be a case for an error?

Sorry, this is not straightforward.

1. If linux.intelRdt is not specified in config.json, always config.IntelRdt == nil (https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L253-L258)
2. In update.go, config := container.Config() (https://github.com/opencontainers/runc/blob/master/update.go#L150), we get `config.IntelRdt == nil`, which implies linux.intelRdt is not specified in config (even though intelrdt.IsEnabled() == true).
3. In my understanding, this is an error case in updateCommand because intelRdt is not managed in container processes management life cycle - no Apply(), Destroy() and etc., only invoke container.Set() --> IntelRdtManager.Set() in updateCommand.

crosbymichael:
LGTM

update.go
```diff
@@ -254,6 +260,17 @@ other options are ignored.
...
+				return fmt.Errorf("l3 cache schema is not enabled or not specified in config, update value '%s' is discarded", val)
+			}
+			if config.IntelRdt == nil {
```
xiaochenshen: @crosbymichael 
This if condition `if config.IntelRdt == nil` is always false, because error returns above when `config.IntelRdt == nil`.
I will remove this if condition lines.

Then the code will look like:
```
        Action: func(context *cli.Context) error {
                ...

                if val := context.String("l3-cache-schema"); val != "" {
                        // Avoid silent failure in container.Set()
                        if !intelrdt.IsEnabled() || config.IntelRdt == nil {
                                return fmt.Errorf("l3 cache schema is not enabled or not specified in config, update value '%s' is discarded", val)
                        }
                        config.IntelRdt.L3CacheSchema = val
                }

                return container.Set(config)
        },
```

update.go
```diff
@@ -254,6 +260,17 @@ other options are ignored.
...
+				config.IntelRdt = &configs.IntelRdt{}
+			}
```
xiaochenshen: @crosbymichael 
This if condition `if config.IntelRdt == nil` is always false, because error returns above when `config.IntelRdt == nil`.
I will remove this if condition lines.

Then the code will look like:
```
        Action: func(context *cli.Context) error {
                ...

                if val := context.String("l3-cache-schema"); val != "" {
                        // Avoid silent failure in container.Set()
                        if !intelrdt.IsEnabled() || config.IntelRdt == nil {
                                return fmt.Errorf("l3 cache schema is not enabled or not specified in config, update value '%s' is discarded", val)
                        }
                        config.IntelRdt.L3CacheSchema = val
                }

                return container.Set(config)
        },
```

update.go
cyphar: That sounds like a bug to me. If someone wants to add an intelrdt limit after the fact (where they didn't specify any options in the original configuration) that should be possible -- is this a kernel bug or just a limitation of `runc` at the moment? We allow you to do that with all of the other cgroups from memory.

update.go
```diff
@@ -254,6 +259,14 @@ other options are ignored.
 		config.Cgroups.Resources.MemorySwap = *r.Memory.Swap
 		config.Cgroups.Resources.PidsLimit = r.Pids.Limit
 
+		if val := context.String("l3-cache-schema"); val != "" {
+			// Avoid silent failure in container.Set()
+			if !intelrdt.IsEnabled() || config.IntelRdt == nil {
+				return fmt.Errorf("l3 cache schema is not enabled or not specified in config, update value '%s' is discarded", val)
```
cyphar: This error should be shorter, just remove the last bit `, update value '%s' is discarded`.

xiaochenshen: @cyphar 

Thank you for review. This sounds fine to me.

update.go
xiaochenshen: @cyphar 

> That sounds like a bug to me. If someone wants to add an intelrdt limit after the fact (where they didn't specify any options in the original configuration) that should be possible -- is this a kernel bug or just a limitation of `runc` at the moment? We allow you to do that with all of the other cgroups from memory.

This is what I intended to do for Intel RDT/CAT support in `runc` as a **tradeoff**. In my opinion, this is neither a kernel bug nor a runc limitation. Let me explain here:

Unlike cgroups' hierarchy, Intel RDT resource control filesystem in kernel supports only single level filesystem layout. (see details in section "Intel RDT "resource control" filesystem hierarchy" in #433 ). For Cache Allocation Technology (CAT) feature, it only supports limited number of groups. It is hardware limitation by nature. The limitation is indicated in /sys/fs/resctrl/info/L3/num_closids. 

In `runc` case, we could only create limited number <container_id> directories under /sys/fs/resctrl via `Apply()`. So we need a **tradeoff**: we `Apply()` intelrdt constraint on a container **only** if hardware and kernel support Intel RDT feature and intelRdt is specified in original configuration. Otherwise, if hardware and kernel support the feature, we will run out of the limitation quickly when started a lot of containers instances, even though the user doesn't care about Intel RDT/CAT feature or doesn't want to add intelrdt constraint for some container instances.

cyphar: @xiaochenshen Alright, that makes sense (as in why it's not that way at the moment) -- but that just sounds like we need code to create a new `rdtgroup` if we're doing an update? I think just using `Apply` might be enough, but I would have to check the code again.

xiaochenshen: @cyphar 
> but that just sounds like we need code to create a new rdtgroup if we're doing an update? I think just using Apply might be enough, but I would have to check the code again.

I will have a look how to do it in update command. Thank you for the suggestion.

xiaochenshen:
@crosbymichael @cyphar 
I have submitted 2 new commits according to @cyphar 's suggestion.
Please help review. Thank you.

NOTE:
Without the first patch, either Intel RDT is not enabled by hardware and kernel, or intelRdt is not specified in original config, we don't init IntelRdtManager in the container to handle intelrdt constraint. It is a tradeoff that Intel RDT has hardware limitation to support only limited number of groups.

The first patch makes a minor change to support update command:
Whether or not intelRdt is specified in config, we always init IntelRdtManager in the container if Intel RDT is enabled. If intelRdt is not specified in original config, we just don't Apply() to create intelrdt group or attach tasks for this container.

In the second patch, in update command, we could re-enable through IntelRdtManager.Apply() and then update intelrdt constraint.

xiaochenshen:
Ping @crosbymichael @cyphar 

Could you have a look at your convenience? Thank you.
Please find the explanation in "NOTE" in my previous comment for how it works in update command.

mrunalp:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1590/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

crosbymichael:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1590/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

PR #1600, opened 2017-09-25
Bump console and sys deps

crosbymichael:
This removes the ClearONLCR in the console package so that clients get a default tty on the system that they create it.


hqhq:
LGTM

dqminh:
LGTM

PR #1603, opened 2017-10-02
rootfs: preserve old mount flags when remounting bindmount

cyphar:
Fixes the case where a bind-mount is being defined in a user namespaced
container which requires a remount. Previously this code would attempt
to do a MS_REMOUNT that potentially dropped mount flags that were
inherited through the MS_BIND. Resolve this by intentionally adding the
mount flags to the mount configuration in that scenario.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -247,9 +247,32 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 		// bind mount won't change mount options, we need remount to make mount options effective.
 		// first check that we have non-default options required before attempting a remount
 		if m.Flags&^(unix.MS_REC|unix.MS_REMOUNT|unix.MS_BIND) != 0 {
+			// Make a copy of the previous mount flags, because in a user
+			// namespace we are not allowed to drop mount options from the host
+			// bindmount.
+			var statfs unix.Statfs_t
+			if err := unix.Statfs(dest, &statfs); err != nil {
+				return err
+			}
+
 			// only remount if unique mount options are set
 			if err := remount(m, rootfs); err != nil {
-				return err
+				if !os.IsPermission(err) {
+					return err
+				}
+
+				// We only inherit bits that are locked with CL_UNPRIVILEGED,
+				// if the remount failed with EPERM (meaning we are likely in a
+				// user namespace).
+				var lockedBits int64
+				lockedBits = unix.MS_RDONLY | unix.MS_NODEV |
+					unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NOATIME |
+					unix.MS_RELATIME | unix.MS_STRICTATIME
+				m.Flags |= int(statfs.Flags & lockedBits)
+
+				if werr := remount(m, rootfs); werr != nil {
+					return err
+				}
 			}
 		}
 


cyphar:
This is a continuation of #1572, and fixes the problems I mentioned in this thread.

/cc @tych0 @justincormack

libcontainer/rootfs_linux.go
```diff
@@ -247,6 +247,15 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 		// bind mount won't change mount options, we need remount to make mount options effective.
 		// first check that we have non-default options required before attempting a remount
 		if m.Flags&^(unix.MS_REC|unix.MS_REMOUNT|unix.MS_BIND) != 0 {
+			// Make a copy of the previous mount flags, because in a user
+			// namespace we are not allowed to drop mount options from the host
+			// bindmount.
+			var statfs unix.Statfs_t
+			if err := unix.Statfs(dest, &statfs); err != nil {
```
tych0: This still seems like the wrong thing to be doing for me -- it seems like someone higher up (i.e. the person/program generating the runc config) should be providing these options. If we do this here, it's probably worth masking off the non-protected bits (i.e. everything besides the *ATIME, NODEV, NOEXEC, and NOSUID bits).

cyphar: Ideally, yes. But there are a lot of users of runc, and quite a few of them have produced incorrect configurations. This is a case that not a lot of people seem to be aware of, so making it "just work" on this level seems like it's much more effective (though maybe some of the other maintainers don't agree).

But yes, we should mask off the non-protected bits.

mrunalp:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1603/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

justincormack:
There are several issues with this...

1. we remount with different options based on whether we get permission denied the first time. I don't really think this is acceptable from a security point of view, there should be one thing we should do and a clear failure if we cannot. If user namespaces absolutely must be special cased, this should be done only based on whether the config has specified a user namespace, or based on a clear instruction from the user, not as a fallback for an error case, as this can lead to exploitable security issues on the non user namespace path.
2. This is all in the `type="bind"` case. `bind` is not even a type that is recognised by the kernel, and is marked as optional, so whether this code is called at all depends randomly on whether you happen to set the `type` on your bind mounts to `bind` or not. Historically `docker` does for example, and other runtimes may well do, but it is not required.
3. Which comes to the spec. The spec is currently very clear that this stuff is passed straight to the kernel as is. The spec does not even say you can do a `remount`, and the onus is all client side. If you think this should change then the spec should be changed first. It is definitely true that the spec does not capture historical usage patterns, but this should be locked down.

cyphar:
@justincormack 

> I don't really think this is acceptable from a security point of view, there should be one thing we should do and a clear failure if we cannot.

I'm not sure I understand what precise class of security issue you're concerned about. We already have code that is conditional on `EPERM` (in the rootless path for example). If the original (more privileged) operation succeeds then it won't hit the retry. If we do a retry with a less privileged operation, then it's effectively the same as not doing the first operation because it failed. My reason for not casing it on whether we have a user namespace is to handle the cases where runc is being run inside a user namespace, or other potential cases where the restrictions might change from the kernel side in the future (plus the code would be a bit gnarly because the mount function currently doesn't take the full configuration).

But yeah, we can do it with a different if, it's not that big of a deal -- I'm just a little confused by what you're referring to as a security issue.

> This is all in the type="bind" case.

Unfortunately this is a problem with the entirety of `rootfs_linux.go`, and is not specific to this PR.

> The spec is currently very clear that this stuff is passed straight to the kernel as is. The spec does not even say you can do a remount, and the onus is all client side.

Not to be too weasel-wordy, but the spec doesn't say that the options *must* be passed to the kernel as-is (it just says that the mounts *must* be done in-order). Not to mention that `bind` is not actually a string option, so there is necessary processing in either case.

FWIW -- I do agree that this is ugly, but the "client" in this context is not me -- it's containerd and Docker which are currently generating broken configurations. I've also seen some other `runc` wrappers do the wrong thing here as well. The problem is that the `bind` option _already_ does quite a bit of magic (it handles binding and adding options -- something that I would argue is not really guaranteed by the spec either, but is effectively necessary for users).

justincormack:
I feel that all fallbacks are a potential security risk, as we should be aiming for a predictable single code path that is well defined. The fallback code path will not be well tested.

We could define what happens if you use "bind" in the spec (thats is what I meant by historical use), but I think it would be better to fix the clients and remove all the code in this "bind" section at some point if this is feasible.

cyphar:
Yeah, @justincormack is right here. Closing in favour of writing patches for `runc` users.

PR #1604, opened 2017-10-04
libcontainer: create Cwd when it does not exist

AkihiroSuda:
The benefit for doing this within runc is that it works well with
userns.
Actually, runc already does the same thing for mount points.

discussed in https://github.com/containerd/containerd/pull/1585 (cc @crosbymichael @estesp )

If this change is fine, I'll open a PR to runtime-spec correspondingly.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -40,7 +40,8 @@ func needsSetupDev(config *configs.Config) bool {
 // prepareRootfs sets up the devices, mount points, and filesystems for use
 // inside a new mount namespace. It doesn't set anything as ro. You must call
 // finalizeRootfs after this function to finish setting up the rootfs.
-func prepareRootfs(pipe io.ReadWriter, config *configs.Config) (err error) {
+func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
+	config := iConfig.Config
 	if err := prepareRoot(config); err != nil {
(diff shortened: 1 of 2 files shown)


libcontainer/rootfs_linux.go
```diff
@@ -76,6 +77,13 @@ func prepareRootfs(pipe io.ReadWriter, config *configs.Config) (err error) {
 		}
 	}
 
+	// Setup Cwd before running the pre-start hooks, so that hooks can play with Cwd.
```
crosbymichael: why not do this right before chdir to the cwd when the rootfs is already setup and every other mount is done and we are jailed inside the rootfs?

cyphar: That's what I was about to comment :wink:. That way we don't need to use `CleanPath` we can just use `MkdirAll` and nothing else. If a hook wants to mess with the cwd, they could create it themselves.

AkihiroSuda:
Updated PR. PTAL. @crosbymichael @cyphar 

However, I didn't do this right before `Chdir()` in `finalizeNamespace()`, because the rootfs is remounted as read-only during `finalize.Rootfs()`.

cyphar:
@AkihiroSuda That's fine, this should (semantically) be part of `prepareRootfs` anyway.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1606, opened 2017-10-08
specconv: emit an error when using MS_PRIVATE with --no-pivot

### LINKED ISSUE #1602: runc with --no-pivot retains global mount context
When a runc container is started with `--no-pivot`, the mount namespace of the process running under runc retains some form of the mount namespace, resulting in the inability to unmount filesystems mounted before runc started from the system.  This happens regardless of the unmounted filesystem not appearing in `/proc/$PID/mounts`

---

### Expected

Filesystem not referenced by runc container to be unmounted when attempted to be unmounted from the system.

### Actual

Filesystem remains mounted until runc containers are stopped.

---

### Reproduction

Reproducible: Always

Runc version: v1.0.0-rc4
Kernel: 4.12.10-1-ARCH

This issue was originally noticed with physical devices, on Ubuntu Xenial with runc 1.0.0-rc2, but the following reproduction is on the above, as listed.

#### Prepare loopback mount

```
$ truncate -s 1G lofile
$ mkdir lomount
$ mkfs.xfs lofile
$ mount -o loop lofile lomount
```

```
# ps auxf | grep loop
root     22940  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [loop0]
root     22948  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-buf/loop0]
root     22952  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-data/loop0]
root     22953  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-conv/loop0]
root     22954  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-cil/loop0]
root     22956  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-log/loop0]
root     22958  0.0  0.0      0     0 ?        S    13:14   0:00  \_ [xfsaild/loop0]
```

```
# mount | grep lomount
/root/loopfs on /root/lomount type xfs (rw,relatime,attr2,inode64,noquota)
```

#### Prepare and run container

```
# docker pull gcr.io/google_containers/pause:0.8.0
# mkdir rootfs
# docker create gcr.io/google_containers/pause:0.8.0
$HASH
# docker export $HASH | tar -C rootfs -x
# oci-runtime-tool generate --args /pause --rootfs-path $(pwd)/rootfs > config.json
$ runc run --no-pivot pause
```

```
# cat /proc/$PAUSE_PID/mounts
/dev/mapper/home / ext4 rw,relatime,data=ordered 0 0
proc /proc proc rw,relatime 0 0
tmpfs /dev tmpfs rw,nosuid,size=65536k,mode=755 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666 0 0
shm /dev/shm tmpfs rw,nosuid,nodev,noexec,relatime,size=65536k 0 0
mqueue /dev/mqueue mqueue rw,nosuid,nodev,noexec,relatime 0 0
sysfs /sys sysfs ro,nosuid,nodev,noexec,relatime 0 0

# cat /proc/$PAUSE_PID/mountinfo 
274 229 254:2 /nakato/tmp/runc/x/usr/bin/rootfs / rw,relatime - ext4 /dev/mapper/home rw,data=ordered
275 274 0:63 / /proc rw,relatime - proc proc rw
276 274 0:64 / /dev rw,nosuid - tmpfs tmpfs rw,size=65536k,mode=755
277 276 0:65 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
278 276 0:66 / /dev/shm rw,nosuid,nodev,noexec,relatime - tmpfs shm rw,size=65536k
279 276 0:62 / /dev/mqueue rw,nosuid,nodev,noexec,relatime - mqueue mqueue rw
280 274 0:67 / /sys ro,nosuid,nodev,noexec,relatime - sysfs sysfs ro

# cat /proc/$PAUSE_PID/mountstats 
device /dev/mapper/home mounted on / with fstype ext4
device proc mounted on /proc with fstype proc
device tmpfs mounted on /dev with fstype tmpfs
device devpts mounted on /dev/pts with fstype devpts
device shm mounted on /dev/shm with fstype tmpfs
device mqueue mounted on /dev/mqueue with fstype mqueue
device sysfs mounted on /sys with fstype sysfs
```

#### Unmount filesystem

```
# umount /root/lomount
```

```
# mount | grep lomount
```
No output, looks unmounted.

```
# ps auxf | grep loop
root     22940  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [loop0]
root     22948  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-buf/loop0]
root     22952  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-data/loop0]
root     22953  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-conv/loop0]
root     22954  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-cil/loop0]
root     22956  0.0  0.0      0     0 ?        S<   13:14   0:00  \_ [xfs-log/loop0]
root     22958  0.0  0.0      0     0 ?        S    13:14   0:00  \_ [xfsaild/loop0]
```
Kernel threads are still here, it's still mounted.

#### Stop the container

I hit Ctrl+C on the container window to stop the container.  Runc and pause have exited.

```
# ps auxf | grep loop
```
No output, filesystem is now unmounted.

---

Some of the repercussions of this issue is the inability to unmount and xfs_repair a physical disk that was mounted before runc started a namespaced process.  This means having to unmount the fs, stop all runc containers, then start all runc processes back up, or remove the filesystem from being mounted on boot and rebooting the system.

I was also able to reproduce this behavior with ext4.  The kernel thread for ext4 is `jbd2/loop`

This issue cannot be reproduced if pivot occurs.

cyphar:
Due to the semantics of chroot(2) when it comes to mount namespaces, it
is not generally safe to use MS_PRIVATE as a mount propgation when using
chroot(2). The reason for this is that this effectively results in a set
of mount references being held by the chroot'd namespace which the
namespace cannot free. pivot_root(2) does not have this issue because
the @old_root can be unmounted by the process.

Ultimately, --no-pivot is not really necessary anymore as a commonly
used option since f8e6b5af5e12 ("rootfs: make pivot_root not use a
temporary directory") resolved the read-only issue. But if someone
really needs to use it, MS_PRIVATE is never a good idea.

Fixes #1602 

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -203,6 +203,9 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 		if config.RootPropagation, exists = mountPropagationMapping[spec.Linux.RootfsPropagation]; !exists {
 			return nil, fmt.Errorf("rootfsPropagation=%v is not supported", spec.Linux.RootfsPropagation)
 		}
+		if config.NoPivotRoot && (config.RootPropagation&unix.MS_PRIVATE != 0) {
+			return nil, fmt.Errorf("rootfsPropagation of [r]private is not safe without pivot_root")
+		}
 
 		for _, ns := range spec.Linux.Namespaces {
 			t, exists := namespaceMapping[ns.Type]


cyphar:
I'm not sure if it's better that we have an error here or if we have some sort of warning message. I didn't want to add a warning message here initially because it would be the first logging we did as part of `specconv` (which doesn't really feel right).

crosbymichael:
What if we moved this into `runc` so that it happens before setup and clone?

cyphar:
@crosbymichael You mean in `/utils.go`? Maybe, though I think that some `specconv` users might actually want to know this as well. But I can switch this to a warning in `/utils.go` if you prefer.

crosbymichael:
Oh, sorry.  I didn't read the filename correctly.

LGTM

hqhq:
LGTM

PR #1608, opened 2017-10-10
add additional-gids to runc exec

crosbymichael:
Closes #1307 

This is a carry of #1307

### DIFF
--- a/tests/integration/exec.bats
+++ b/tests/integration/exec.bats
@@ -120,8 +120,8 @@ function teardown() {
 
   wait_for_container 15 1 test_busybox
 
-  runc exec --user 1000:1000 --additional-gids 100 test_busybox id
+  runc exec --user 1000:1000 --additional-gids 100 --additional-gids 99 test_busybox id 
   [ "$status" -eq 0 ]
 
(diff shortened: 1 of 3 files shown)


crosbymichael:
@cyphar how should this work in userns?  See travis failures for context.

tests/integration/exec.bats
```diff
@@ -112,3 +112,19 @@ function teardown() {
 
   [[ "${output}" == "uid=1000 gid=1000"* ]]
 }
+
+@test "runc exec --additional-gids" {
+  # --user can't work in rootless containers that don't have idmap.
+  [[ "$ROOTLESS" -ne 0 ]] && requires rootless_idmap
```
cyphar: `--additional-gids` cannot ever work in a rootless container, because `setgroups` is disabled. This should just be `requires root`. In this case, the `specconv` error is correct here.

crosbymichael:
Ok, updated to skip that test when rootless is running.  Good to review now

cyphar:
LGTM.

cyphar:
As an aside, we probably should add (privileged) user namespace integration tests for cases like this. But we can work on that in the future.

mrunalp:
LGTM

Issue #1609, opened 2017-10-12
libcontainer: Potential infinite loop when setting freezer cgroup state


sboeuf:
When using libcontainer library, we can end up calling into signalAllProcesses which leads to the following call chain:
https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L335
==>https://github.com/opencontainers/runc/blob/master/libcontainer/init_linux.go#L473
==> https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/apply_raw.go#L224
==> https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/freezer.go#L29

This can potentially lead to a deadlock by running in an infinite loop if the state never gets the one that is expected (FREEZING instead of FROZEN).

This code needs two things:
- Call into `writeFile()` everytime we don't get the expected state, because reading from the doc here: https://www.kernel.org/doc/Documentation/cgroup-v1/freezer-subsystem.txt, it appears that if a new task pops up, it will set the state to FREEZING instead of FROZEN. We have to set FROZEN again to make sure the new task is set to FROZEN state too.
- Add a timeout. After some time, we might want to consider this loop as failing instead of waiting forever.

sboeuf:
@crosbymichael @cyphar @mrunalp any thoughts about this ?

crosbymichael:
SGTM

PR #1610, opened 2017-10-12
libcontainer: cgroups: Write freezer state after every state check

sboeuf:
This commit ensures we write the expected freezer cgroup state after
every state check, in case the state check does not give the expected
result. This can happen when a new task is created and prevents the
whole cgroup to be FROZEN, leaving the state into FREEZING instead.

This patch prevents the case of an infinite loop to happen.

Fixes https://github.com/opencontainers/runc/issues/1609


crosbymichael:
LGTM

mrunalp:
LGTM

PR #1612, opened 2017-10-16
RFC: libcontainer: add rdmacg support

mitake:
This commit adds an initial support for rdmacg of linux, that enables
controlling rdma related resources (currently HCA handles and HCA
objects).

Signed-off-by: Hitoshi Mitake <mitake.hitoshi@lab.ntt.co.jp>

This is a very early stage PR. I'm really glad if I can hear opinions about the idea of managing RDMA related resources from runc.

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -31,6 +31,7 @@ var (
 		&PerfEventGroup{},
 		&FreezerGroup{},
 		&NameGroup{GroupName: "name=systemd", Join: true},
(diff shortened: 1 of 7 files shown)


cyphar:
You'd want to submit a corresponding PR to opencontainers/runtime-spec -- because ultimately the `runc` configuration is sourced from an OCI configuration.

mitake:
@cyphar thanks for your comment. I'll create a PR for the spec side later.

mitake:
I'm closing this because @paravmellanox has the good reference implementation.

PR #1615, opened 2017-10-17
libcontainer: intelrdt: fix a GetStats() issue

xiaochenshen:
This fixes a GetStats() issue introduced in #1590:
If Intel RDT is enabled by hardware and kernel, but intelRdt is not
specified in original config, GetStats() will return error unexpectedly
because we haven't called Apply() to create intelrdt group or attach
tasks for this container. As a result, runc events command will have no
output.

### DIFF
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -443,6 +443,11 @@ func (m *IntelRdtManager) GetPath() string {
 
 // Returns statistics for Intel RDT
 func (m *IntelRdtManager) GetStats() (*Stats, error) {
+	// If intelRdt is not specified in config
+	if m.Config.IntelRdt == nil {
+		return nil, nil
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	stats := NewStats()


libcontainer/intelrdt/intelrdt.go
```diff
@@ -443,6 +443,11 @@ func (m *IntelRdtManager) GetPath() string {
 
 // Returns statistics for Intel RDT
 func (m *IntelRdtManager) GetStats() (*Stats, error) {
+	// If intelRdt is not specified in config
+	if m.Config.IntelRdt == nil {
+		return nil, nil
```
crosbymichael: Does the calling code handle a `nil, nil` return or will it panic?

xiaochenshen: @crosbymichael 
> Does the calling code handle a nil, nil return or will it panic?

It will not panic. In caller linuxContainer.Stats(): libcontainer.Stats.IntelRdtStats will be assigned to nil when intelRdtManager.GetStats() returns nil, nil.
And in events command, the caller convertLibcontainerStats() will handle the case when libcontainer.Stats.IntelRdtStats = nil

crosbymichael:
LGTM

cyphar:
LGTM.

PR #1424, opened 2017-04-26
Update libseccomp-golang dependency for filter generation bugfix

mheon:
A bug in Seccomp filter handling was recently identified in Moby related to handling of syscall arguments in Seccomp filters (https://github.com/moby/moby/issues/32714). The bug was in the `libseccomp-golang` bindings, and has been fixed there. This PR updates the bindings to include this fix, and provides integration tests to catch regressions in this behavior.

The minimum supported version of `libseccomp` is bumped from v2.1.0 to v2.2.0 by associated changes to the bindings. Support for v2.1.0 was never very good (some features had to be gated off because of a library bug), though it is still the version provided by a few major distributions. If this change is contentious, I can see about backing out the changes that require v2.2.0.

PR #1616, opened 2017-10-17
Fix breaking change in Seccomp profile behavior

mheon:
Multiple conditions were previously allowed to be placed upon the same syscall argument. My earlier patch #1424 fixed an issue in the Libseccomp bindings with conditional syscall matching, but this broke multiple conditions being placed upon the same argument. This patch restores the old behavior for rules which would be broken by the prior change.

Note that the prior behavior here is still not correct behavior - the conditions for these rules will still be logical-OR instead of logical-AND as they should be. I looked into generating rules that would act correctly, but that proved to be unreasonably complex in some cases.

I haven't added a warning for the badly-behaved profiles, though I am open to doing so.

### DIFF
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -22,6 +22,11 @@ var (
 	actErrno = libseccomp.ActErrno.SetReturnCode(int16(unix.EPERM))
 )
 
+const (
+	// Linux system calls can have at most 6 arguments
+	syscallMaxArguments int = 6
+)
+
 // Filters given syscalls in a container, preventing them from being used
 // Started in the container init process, and carried over to all child processes
 // Setns calls, however, require a separate invocation, as they are not children
@@ -45,11 +50,11 @@ func InitSeccomp(config *configs.Seccomp) error {
 	for _, arch := range config.Architectures {
 		scmpArch, err := libseccomp.GetArchFromString(arch)
 		if err != nil {
-			return err
+			return fmt.Errorf("error validating Seccomp architecture: %s", err)
 		}
 
 		if err := filter.AddArch(scmpArch); err != nil {
-			return err
+			return fmt.Errorf("error adding architecture to seccomp filter: %s", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mheon:
@mrunalp PTAL, this should fix the issue identified by the CRI-O tests

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -170,29 +170,55 @@ func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
...
+				return fmt.Errorf("error creating seccomp syscall condition for syscall %s: %s", call.Name, err)
 			}
 
 			conditions = append(conditions, newCond)
 		}
 
-		if err = filter.AddRuleConditional(callNum, callAct, conditions); err != nil {
-			return err
+		hasMultipleArgs := false
+		for _, count := range argCounts {
```
TomSweeneyRedHat: I might be missing a golang nuance, but could you just do "len(argCounts) > 1"?

TomSweeneyRedHat:
LGTM assuming happy test.  One minor question.  Also do we have a trello card or issue created to go back and get to the correct behaviour at some point in time?

libcontainer/seccomp/seccomp_linux.go
mheon: len(argCounts) should always be 6, as it's a 6-long array of uints, one for each argument possible in a Linux sycall. Each one holds the number of conditions that check that specific argument. If we had reduce in Go, I could make this a one-liner, but as things are needs to check each individual field in the array so it needs to iterate.

TomSweeneyRedHat: Ah, I see the 6 now at line 184.  Missed that, was thinking this was dynamically allocated.  Thanks for the 411.

cyphar:
@TomSweeneyRedHat This seems like a user configuration issue from the `runc` side and I don't really see an issue in supporting this for the foreseeable future. If you're referring to `cri-o` not generating these broken seccomp filters, I sure would hope there's something tracking that. :wink:

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -170,29 +170,55 @@ func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
...
 	if len(call.Args) == 0 {
 		if err = filter.AddRule(callNum, callAct); err != nil {
-			return err
+			return fmt.Errorf("error adding seccomp filter rule for syscall %s: %s", call.Name, err)
 		}
 	} else {
-		// Conditional match - convert the per-arg rules into library format
+		// If two or more arguments have the same condition,
+		// Revert to old behavior, adding each condition as a separate rule
+		argCounts := make([]uint, 6)
```
cyphar: While I recognise that currently Linux only has 6-argument syscalls, and it's unlikely we'll ever have 7-argument syscalls, can we make this a global `MaxSyscallArguments` somewhere?

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -170,29 +170,55 @@ func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
...
 		conditions := []libseccomp.ScmpCondition{}
 
 		for _, cond := range call.Args {
+			argCounts[cond.Index] += 1
```
cyphar: And I'm worried this will panic if someone gives an invalid index as a configuration. We should error out here if `cond.Index >= MaxSyscallArguments`.

mheon: I think we can actually just move the line down below the logic that creates a Libseccomp condition, which has logic to catch invalid indexes. Should resolve the issue without introducing additional checks.

libcontainer/seccomp/seccomp_linux.go
mheon: Should this constant live in runc or the libseccomp bindings? I think the bindings would be the logical place, but it would require revendoring once the change lands.

cyphar: I agree with putting it in the libseccomp bindings, but let's keep in `runc` for now until the change lands.

mheon:
@cyphar Made requested changes. Can you check the constant to see if it's what you were looking for?

cyphar:
Yup, that looks good.

LGTM (the test failure is due to Travis' repos being broken at the moment -- I 'll re-trigger the tests later).

mrunalp:
LGTM

PR #1619, opened 2017-10-25
WIP: Better testsuite for specconv

fntlnz:
The aim of this is to improve the current `specconv` testsuite so that edge cases are spotted early and future refactorings are easier.

This is a work in progress, opened the PR early so that I can take benefit of CI and comments.

Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/libcontainer/specconv/spec_linux_test.go
+++ b/libcontainer/specconv/spec_linux_test.go
@@ -3,12 +3,175 @@
 package specconv
 
 import (
+	"os"
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer/configs/validate"
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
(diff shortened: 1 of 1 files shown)


cyphar:
Note that we are planning on dropping `specconv` "eventually", and just using OCI configurations throughout the entirety of `runc` and `libcontainer` (the existence of `specconv` is because `libcontainer` existed before the OCI -- and rewriting it is more expensive than just wrapping it).

fntlnz:
@cyphar all right, do you  think that it's still worth having fewer more tests anyway given that?

cyphar:
Yeah, it'll still be useful -- just wanted to make sure you're aware that the code will likely go away with said dropping of `specconv`. It's likely we won't be dropping `specconv` for quite a while yet.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1621, opened 2017-10-26
tests: add missing cgroups_kmem requirement

monstermunchkin:
Since the defined config.json contains kmem settings, the test will try
writing to memory.kmem.* and fail. Therefore, it needs to require
cgroups_kmem.

Signed-off-by: Thomas Hipp <thipp@suse.de>


cyphar:
LGTM. Though I would personally prefer if we had much better split-out cgroup tests.

crosbymichael:
LGTM

PR #1622, opened 2017-10-27
Remove pkg/symlink from docker/docker and use cyphar/filepath-securejoin

vdemeester:
`runc` shouldn't depend on docker and be more self-contained — removing github.com/pkg/symlink dep is the first step to not depend on docker anymore.

The next one is `pkg/mount` 👼

/cc @jhowardmsft @crosbymichael @hqhq @mrunalp


cyphar:
If we're gonna stop using the upstream `pkg/symlink` -- we should switch entirely to https://github.com/cyphar/filepath-securejoin. And it's versioned properly.

dqminh:
Swithing to https://github.com/cyphar/filepath-securejoin looks OK to me.

vdemeester:
@cyphar @dqminh oh nice, I would be ok with that too 👼 😉

vdemeester:
I'll update using it 👼

cyphar:
I can work up a PR to switch Docker to use `filepath-securejoin` as well.

vdemeester:
@cyphar @dqminh done 👼

cyphar:
There's still a reference to `symlink.FollowSymlinkInScope` in `rootfs_linux`. And also it looks like the files aren't `gofmt`'d correctly.

vdemeester:
ah yep :scream:

vdemeester:
oh @cyphar.. some unit-tests are failing :sweat_smile:

libcontainer/rootfs_linux.go
```diff
@@ -240,7 +240,7 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 		// any previous mounts can invalidate the next mount's destination.
 		// this can happen when a user specifies mounts within other mounts to cause breakouts or other
 		// evil stuff to try to escape the container's rootfs.
-		if dest, err = symlink.FollowSymlinkInScope(dest, rootfs); err != nil {
```
cyphar: The correct ordering of arguments would be:

```
if dest, err = securejoin.SecureJoin(rootfs, m.Destination); err != nil {
```

However it feels like this section would be a better place to fix:

```
164 var (
165  	dest = m.Destination
166 )
167 if !strings.HasPrefix(dest, rootfs) {
168 	dest = filepath.Join(rootfs, dest)
169 }
```

And then the two calls for `symlink.FollowSymlinkInScope` shouldn't be necessary.

/cc @crosbymichael

thaJeztah:
@vdemeester can you update the PR description to match the updated change?

libcontainer/rootfs_linux.go
vdemeester: ohhh :sweat_smile:

vdemeester:
@thaJeztah oh right :+1:

cyphar:
Yeah, just to be clear: `SecureJoin(a, b)` is basically semantically similar to `filepath.Join(a, b)` except it scopes the output to `a`. The reason it is different to the usual `EvalSymlinksInScope` model that Docker used is because people who use `EvalSymlinksInScope` almost always also use `Join` -- so might as well combine the two (plus `Join` does a `Clean` which makes the results less accurate).

vdemeester:
@cyphar updated 👼

cyphar:
LGTM.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1623, opened 2017-10-28
Remove global device mknod cgroup rules

### LINKED ISSUE #1618: Default linux cgroup devices.allow rules too general
In [runc/libcontainer/specconv/spec_linux.go](https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go), there are two default rules, `c *:* m` and `b *:* m`, commented "allow mknod for any device".  I have a use case where I am scheduling Docker containers using the SLURM resource manager, and specifically the character device rule is causing issues.

When SLURM restricts access to GPU devices (which are character devices), it writes `rwm` rules to `devices.deny`.  When we run the a Docker job through SLURM, we are using Docker's `--cgroup-parent` option to set SLURM's cgroup restrictions as the parent of Docker's policies, ensuring that the SLURM restrictions are propagated.

This is where the issue come's in.  When running the container, runc is trying to set the aforementioned `c *:* m` rule for Docker, but this is in conflict with one or more of the parent GPU deny rules (set by SLURM), so the container fails to run.

I have tried building a custom version of runc that gets rid of the `c *:* m` rule, but it seems like this causes all container runs to completely break (hang).

It seems like runc should not be setting device rules that are this global by default, since it can cause conflicts with other cgroup policies like I am observing.  This could also cause issues when users try to set explicit device rules with moby's `--device-cgroup-rule` flag [(introduced here.)](https://github.com/moby/moby/pull/22563).

Similarly, the `b *:* m` rule could cause issues with block device cgroup rules defined in the parent policy.

Does anyone have insight into what devices actually need mknod device permissions in order for containers to run properly?  I think these default rules should be restricted unless there is some explicit reason for them to be so global.

dasizeman:
These global rules seem like they aren't needed and they cause conflicts with parent cgroup policies.

Fixes #1618


cyphar:

dasizeman:
I have signed off on the commit, thanks.

crosbymichael:
One of the reasons is that many things mknod devices and install fails.  Things like java mknod fuse, etc.  Mknod is safe across the board as long as rw is blocked via cgroups. 

Also mknod is save because there is nothing stopping an image from bringing in device nodes so we have a wildcard to allow mknod and block access

dasizeman:
I see, so it seems like really the issue I was originally having in #1618 is because SLURM shouldn't be blocking mknod on the devices in the parent policy.

@crosbymichael Do you have any more details/references on how you are sure that global mknod is safe, and maybe some examples of things that fail without it?  I would like to understand this better and also I am going to need a compelling argument if I want a change in SLURM to be accepted.

crosbymichael:
@dasizeman 

I think this will give you all the history and information that you need.  Everything from kernel devs to users involved ;)

https://github.com/moby/moby/pull/2979

rhatdan:
Personally I would prefer that this was off.  There are hardly any containers that would ever need mknod, and I believe in relying on as many security mechanisms as possible to protect against container break out.  Perhaps having separate capabilities for build and run would be helpful.  But when far less then 1 % of all containers needs mknod, we should not enable it by default.  But that is water under the bridge at this point for docker/moby I am afraid.

cyphar:
@crosbymichael I agree with you from the Docker side of things, but it feels as though these default should be added to the specs that Docker generates rather than having them in `runc` -- especially since if you wanted to make different `mknod` policies you'd need to mess around with `devices.deny` a fair bit in order to make `runc` act nicely. Not to mention that tools (like `umoci`) that generate root filesystems for runtime bundles might do filtering.

And while a process inside a container cannot `open(...)` the device inode, I have found some quite worrying kernel bugs (which I've reported and am discussing upstream) that would make it theoretically possible to escalate the ability to access an inode into being able to trick another (more privileged) process into opening it.

dasizeman:
@crosbymichael Thanks for pointing me toward that PR, it did a great job of catching me up on the history.

I actually tend to agree with @cyphar here.  Based on the original need for these permissions in Docker, it seems like they only need to be present at image build time.  The correct course of action in my opinion would be to remove them in runc, and have Docker only implicitly enable them during image builds, not when running containers (if they were needed at runtime, they could be set like any other special permission).

If everyone agrees, I would be happy to help with getting the needed changes merged on the moby side of things.

crosbymichael:
I'm fine with it as long as we update projects that rely on runc so that no regressions are introduced.  This change is not about security, its about where the responsibility lies.  Also the moby change should not apply to just build time, it should apply to all, I'm not comfortable saying that people are not ever going to drop into a bash session and run an apt-get install and it breaks after this change.

dasizeman:
Good point about apt-get installs from the shell, I forgot about that.  I think you're right that it's a responsibility issue, and I think that the responsibility of allowing mknod to enable certain package installs should lie with moby, not runc.  Perhaps the best move forward here is to get this merged in runc and update any downstream projects.

Unfortunately, on the moby side of things, this is a more complicated issue.  These mknod permissions are required for certain use cases, but at the same time they are really an inconsistency as far as container security goes, and they are only around to allow certain package installs, while simultaneously breaking `--cgroup-parent`.

I will open an issue in moby to start a discussion on how to integrate this change if it were merged.

@crosbymichael What other projects are there downstream?  I believe moby actually uses runc indirectly through containerd correct?  Also, what would the process be for releasing a change like this and making sure we make the needed changes in downstream projects?  If we can't figure out a good way to "solve" the inconsistency on the moby side anyway, then I would imagine it may not be worth going through the process of integrating this right now, since there would be no motivating use case (my original use case with Docker would still be broken if no change was made on that side).

cyphar:
@dasizeman 

> Unfortunately, on the moby side of things, this is a more complicated issue. These mknod permissions are required for certain use cases, but at the same time they are really an inconsistency as far as container security goes: they allow certain package installs, while simultaneously breaking `--cgroup-parent`.

[Insert common complaint that `--cgroup-parent` is usually misused.]

This is actually situation that can be detected somewhat trivially by whoever is making decisions on what devices should be allowed. `devices.list` gives you the effective set of devices that are allowed -- and userspace (whether that be `runc` or Docker) should be able to figure out whether `a *:* m` is going to be permitted in a sub-cgroup (of course you'd need to replicate the kernel parsing code and other set-related logic for `devices`). We could add some functions in `libcontainer` than Docker could then use, since we already have a lot of code in `runc` that does these sorts of things.

There are some slight complications for blacklist `devices` cgroups -- but in that situation there's nothing we can do in userspace other than just try the bad configuration and then adjust afterwards. And ultimately, if you're using a blacklist you're already doing things wrong.

> What other projects are there downstream? I believe moby actually uses runc indirectly through containerd correct?

That is correct, but there are other downstreams. `cri-o` uses `runc`. Cloud Foundry has also switched to using `runc`. There's also tools like [`orca-build`](https://github.com/cyphar/orca-build) which use `runc` (and there are certainly many others).

cyphar:
I'm removing this from 1.0 since we don't have time to land other commits.

kolyshkin:
I'm inclined to close this one, as every time we remove some permissions we end up realizing some upper-level tools depend on them and/or have no way to re-add those when needed, thus bringing in the unnecessary turmoil. Last time it happened then we tried to remove tun/tap device from the default set (see https://github.com/opencontainers/runc/pull/3468, https://github.com/opencontainers/runc/pull/4555).

Feel free to reopen or file a new PR/issue if you disagree.

PR #1625, opened 2017-10-28
We should not mount /dev read-only

rhatdan:
/dev is on a tmpfs.  It is not part of the host file system, so it should not be read-only.

This stops certain containers from creating device nodes in /dev as well as /dev/log.

Breaks systemd, which wants to create /dev/log for the journal.  Also will also break any
tool that creates a tty device or any of the other devices allowed to be created by the device cgroup.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -127,18 +127,6 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 // finalizeRootfs sets anything to ro if necessary. You must call
 // prepareRootfs first.
 func finalizeRootfs(config *configs.Config) (err error) {
-	// remount dev as ro if specified
-	for _, m := range config.Mounts {
-		if libcontainerUtils.CleanPath(m.Destination) == "/dev" {
-			if m.Flags&unix.MS_RDONLY == unix.MS_RDONLY {
-				if err := remountReadonly(m); err != nil {
-					return newSystemErrorWithCausef(err, "remounting %q as readonly", m.Destination)
-				}
-			}
-			break
-		}
-	}
-
 	// set rootfs ( / ) as readonly
 	if config.Readonlyfs {
 		if err := setReadonly(); err != nil {


cyphar:
This is a configuration-specific thing, and removing the code you removed will break configurations which explicitly make `/dev/` read-only. If you have a Docker issue (or a default configuration issue) then I'd recommend changing the default configuration or whatever configuration generator you're using.

rhatdan:
@cyphar So your saying this would only happen if someone explicitly sets /dev to readonly.  Not if someone sets the read-only flag in the spec file?

I was digging around for where this would be set and I read this code wrong.  I see no reason for anyone to set /dev on a tmpfs to be readonly, but if they explicitly asked for it then this code should run.

dqminh:
@rhatdan yah this is user explicitly asks us to mount /dev readonly. The readonly flag in spec file is for rootfs ( a few lines below your change )

Can you also post the reproduction step for the issue ?

rhatdan:
Well I see this with docker.
```
# docker run --read-only -ti fedora mount | grep "/dev " 
tmpfs on /dev type tmpfs (ro,context="system_u:object_r:container_file_t:s0:c211,c719",mode=755)
```

runcom:
seems like docker is asking to remount it readonly when the whole rootfs is readonly:
```go
541     if s.Root.Readonly {
542         for i, m := range s.Mounts {
543             switch m.Destination {
544             case "/proc", "/dev/pts", "/dev/mqueue": // /dev is remounted by runc
545                 continue
546             }
547             if _, ok := userMounts[m.Destination]; !ok {
548                 if !stringutils.InSlice(m.Options, "ro") {
549                     s.Mounts[i].Options = append(s.Mounts[i].Options, "ro")
550                 }
551             }
552         }
553     }
```

rhatdan:
So if I add "/dev" to that list we should be ok. Although I don't know what the comment is alluding to.

544             case "/proc", "/dev/pts", "/dev/mqueue", "/dev":

rhatdan:
Opened https://github.com/moby/moby/pull/35344 on moby.

PR #1633, opened 2017-11-01
Create the xattr test file in user's home dir

BooleanCat:
Running tests on a Mac with a VirtualBox Linux docker host we see TestXattr fail because the test is unable to set the extended attributes on /go/src/github.com/opencontainers/runc. Extended attributes do not seem to work on VBox shared file systems (which is what
VirtualBox mounts the runc directory as).

We propose creating the test file in the user's home directory which is more likely to be a file system that supports extended attributes

We think this change is uncontroversial because of this statement in the README: https://github.com/opencontainers/runc/blame/10dc96bb9e7e3b07511efdc830560c1384d92f79/README.md#L72

Signed-off-by: Danail Branekov <danailster@gmail.com>

### DIFF
--- a/libcontainer/xattr/xattr_test.go
+++ b/libcontainer/xattr/xattr_test.go
@@ -4,13 +4,19 @@ package xattr_test
 
 import (
 	"os"
+	"os/user"
+	"path/filepath"
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer/xattr"
 )
 
 func TestXattr(t *testing.T) {
-	tmp := "xattr_test"
+	u, err := user.Current()
+	if err != nil {
+		t.Fatal("failed")
+	}
+	tmp := filepath.Join(u.HomeDir, "xattr_test")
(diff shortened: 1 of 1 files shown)


danail-branekov:
We noticed that after running the tests there is a leftover file `libcontainer/state.json`. We would like to git-ignore it so that the repository does not get dirty. Could you suggest whether we do that as part of this PR, or raise another one?

dqminh:
@BooleanCat actually we should be able to remove all xattr related functions here in libcontainer, now that selinux has been moved to a separated package. Nothing in libcontainer/runc should use xattr otherwise.

BooleanCat:
@dqminh Would you accept a separate PR to remove the Xattr functions and tests from this repo? Which package has selinux been moved to?

dqminh:
@BooleanCat yah we can close this one and have a new PR for removing xattr related packages from here.

selinux is moved here https://github.com/opencontainers/selinux

BooleanCat:
👍

Issue #1635, opened 2017-11-02
runc update  error when there is a nested cpuset cgroup


Ace-Tang:
I try this command `runc --root /run/containerd/runc/default/ update --cpuset-cpus "0,1" docker2`, and get error
```
failed to write 0,1 to cpuset.cpus: write /sys/fs/cgroup/cpuset/default/docker2/cpuset.cpus: device or resource busy
```

this is because a nested cpuset cgroup existed, cause child cpusets must contain a subset of the parents CPU and Memory resources, so we cannot simply change the data in parent cgroup file.

this case occurs in a real business scene. here is my test environment, docker2 is a container cpuset cgroup, t1, t2, l1, l2, c3 is a nested cpuset cgroup.
```
root@www:/sys/fs/cgroup/cpuset/default/docker2# tree
...
├── cpuset.sched_load_balance
├── cpuset.sched_relax_domain_level
├── notify_on_release
├── t1
│   ...
│   ├── cpuset.memory_migrate
│   ├── cpuset.memory_pressure
│   ├── cpuset.memory_spread_page
│   ├── cpuset.memory_spread_slab
│   ├── cpuset.mems
│   ├── cpuset.sched_load_balance
│   ├── cpuset.sched_relax_domain_level
│   ├── l1
│   │   ├── cgroup.clone_children
│   │   ├── cgroup.procs
│   │   ├── cpuset.cpu_exclusive
│   │   ├── cpuset.cpus
│   │   ├── cpuset.effective_cpus
│   │   ├── cpuset.effective_mems
│   │   ├── cpuset.mem_exclusive
│   │   ├── cpuset.mem_hardwall
│   │   ├── cpuset.memory_migrate
│   │   ├── cpuset.memory_pressure
│   │   ├── cpuset.memory_spread_page
│   │   ├── cpuset.memory_spread_slab
│   │   ├── cpuset.mems
│   │   ├── cpuset.sched_load_balance
│   │   ├── cpuset.sched_relax_domain_level
│   │   ├── notify_on_release
│   │   └── tasks
│   ├── notify_on_release
│   └── tasks
├── t2
│    ...
│   ├── cpuset.memory_spread_slab
│   ├── cpuset.mems
│   ├── cpuset.sched_load_balance
│   ├── cpuset.sched_relax_domain_level
│   ├── l2
│   │   ├── c3
│   │   │   ├── cgroup.clone_children
│   │   │   ├── cgroup.procs
[...]

PR #1636, opened 2017-11-02
cgroups: support udpate nested cpuset cgroup

Ace-Tang:
cpuset cgroup form a nested hierarchy visible in a virtual
file system, and child cpusets must contain a subset of the
parents CPU and Memory resources. When root user get permission
to make nested cpuset cgroup, this patch will allow update
operation in this case.

Fixes: #1635

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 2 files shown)


Ace-Tang:
here is my test, runc build from commit `0232e38342a8d230c2745b67c17050b2be70c6bc`
```
root@www:/sys/fs/cgroup/cpuset/default/docker2# runc -v
runc version 1.0.0-rc4+dev
commit: 0232e38342a8d230c2745b67c17050b2be70c6bc
spec: 1.0.0

# runc --root /run/containerd/runc/default/ update --cpuset-cpus "0,2" docker2
failed to write 0,2 to cpuset.cpus: write /sys/fs/cgroup/cpuset/default/docker2/cpuset.cpus: device or resource busy
```

`./runc` build with the patch
```
root@www:/home/ace/gosrc/src/github.com/opencontainers/runc# ./runc  --root /run/containerd/runc/default/ update --cpuset-cpus "0,2" docker2
root@www:/home/ace/gosrc/src/github.com/opencontainers/runc# 

root@www:/sys/fs/cgroup/cpuset/default/docker2# cat cpuset.cpus 
0,2
```

hqhq:
The logic "update one cgroup's cpuset means update that cgroup and all it's children" is not right, we should not do this by default.

Use an extra option makes more sense, or an easier way, use a script to update all it's child cgroups first, the administrator(or orchestrator) should know the whole hierarchy.

PR #1640, opened 2017-11-03
enable unit test on arm64 platform

jongwu:
Currently, unit test can't be done on arm64 platform
due to multi-arch issue.

Fix points:
golang:1.8.0 doesn't support arm64.
We use golang:1.8 to replace it.

Signed-off-by: Jianyong Wu <jianyong.wu@arm.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.8.0
+FROM golang:1.8
 
 # libseccomp in jessie is not _quite_ new enough -- need backports version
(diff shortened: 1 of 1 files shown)


jongwu:
@avagin @caniszczyk @crosbymichael @cyphar can you help to check?

tianon:
(for context, Go 1.8.5 added official arm64 release binaries, which is why this works)

cyphar:
LGTM.

avagin:
LGTM

dqminh:
LGTM

PR #1641, opened 2017-11-04
specconv.Example(): add /proc/scsi to masked paths

AkihiroSuda:
Port over https://github.com/moby/moby/pull/35399

### DIFF
--- a/libcontainer/specconv/example.go
+++ b/libcontainer/specconv/example.go
@@ -116,6 +116,7 @@ func Example() *specs.Spec {
 				"/proc/timer_stats",
 				"/proc/sched_debug",
 				"/sys/firmware",
+				"/proc/scsi",
 			},
 			ReadonlyPaths: []string{
 				"/proc/asound",


vielmetti:
This is related to CVE-2017-16539.

libcontainer/specconv/example.go
```diff
@@ -116,6 +116,7 @@ func Example() *specs.Spec {
 				"/proc/timer_stats",
 				"/proc/sched_debug",
 				"/sys/firmware",
```
cyphar: Paths should be sorted.

vielmetti:
There exists a kernel patch too: https://marc.info/?l=linux-scsi&m=150982199728895&w=2

libcontainer/specconv/example.go
AkihiroSuda: Since the existing list is not sorted, I wonder sorting in this PR might have some negative impact on git commit traceability.

cyphar: Hmm, fair enough.

cyphar:
LGTM.

crosbymichael:
LGTM

cyphar:
This also should have a `runtime-tools` patch.

AkihiroSuda:
looks like runtime-tools doesn't set masked paths by default? https://github.com/opencontainers/runtime-tools/search?utf8=✓&q=maskedpaths&type=

cc @Mashimiao @q384566678

Mashimiao:
@AkihiroSuda  currently, yes.  runtime-tools will supply which kind of default config as template, which has not been determined. Discussion is [here](https://github.com/opencontainers/runtime-tools/pull/379). 
If we decided to supply good default config as template for users, I think we should have this path.

PR #1642, opened 2017-11-07
enable integration test on arm64 platform

lubinsz:
Currently, integration test can't be done on arm64 platform due to several issues.
 Fix points:
 1, add busybox.tar with arm64 format
 2, add hello-world.tar with arm64 format

Signed-off-by: Bin Lu <bin.lu@arm.com>


lubinsz:
@tianon @cyphar @avagin @dqminh 
Hi all, can you help to check it?
Thanks.

Dockerfile
```diff
@@ -50,9 +50,16 @@ RUN mkdir -p /usr/src/criu \
 
 # setup a playground for us to spawn containers in
 ENV ROOTFS /busybox
-RUN mkdir -p ${ROOTFS} \
-    && curl -o- -sSL 'https://github.com/docker-library/busybox/raw/a0558a9006ce0dd6f6ec5d56cfd3f32ebeeb815f/glibc/busybox.tar.xz' | tar xfJC - ${ROOTFS}
+RUN mkdir -p ${ROOTFS}
 
+RUN case $(uname -m) in \
```
tianon: I would always recommend against using `uname` for platform detection, both because it will often return the wrong result (building `arm/v7` images on an `arm64` box, for example) and because for many ARM 32-bit platforms it gets [a little out of hand](https://github.com/tianon/gosu/issues/6#issuecomment-132239368), and would recommend instead using something like `dpkg --print-architecture` or `apk --print-arch` to get a simple and consistent string that represents the architecture of the image/userspace instead.

lubinsz:
Hi @tianon ,
I have updated it. Using 'dpkg --print-architecture' now.
Thanks.

Dockerfile
lubinsz: @tianon 
Can you check it?
Thanks.

lubinsz:
@avagin 
@caniszczyk 
@crosbymichael 
@cyphar 
@dqminh 
@hqhq 
@mrunalp 
@rjnagal 
Hi all,
Could you take a look at it? 
Thanks.

tests/integration/helpers.bash
```diff
@@ -14,7 +14,15 @@ BUSYBOX_IMAGE="$BATS_TMPDIR/busybox.tar"
 BUSYBOX_BUNDLE="$BATS_TMPDIR/busyboxtest"
 
 # hello-world in tar format
-HELLO_IMAGE="$TESTDATA/hello-world.tar"
+case $(dpkg --print-architecture) in
```
hqhq: If we use dpkg, non debian systems won't be able to run local integration tests, are there any other standard way to detect the osarch?

cyphar: `$(go env GOARCH)` seems like it would be a much better idea.

tests/integration/helpers.bash
```diff
@@ -14,7 +14,15 @@ BUSYBOX_IMAGE="$BATS_TMPDIR/busybox.tar"
...
+arm64)
+	HELLO_IMAGE="$TESTDATA/hello-world-aarch64.tar"
+;;
+*)
+	HELLO_IMAGE="$TESTDATA/hello-world.tar"
+;;
```
cyphar: It isn't necessary to have two different filenames for this.

Dockerfile
```diff
@@ -50,9 +50,16 @@ RUN mkdir -p /usr/src/criu \
...
+RUN mkdir -p ${ROOTFS}
 
+RUN case $(dpkg --print-architecture) in \
+    arm64) \
+      curl -o- -sSL 'https://github.com/docker-library/busybox/raw/23fbd9c43e0f4bec7605091bfba23db278c367ac/glibc/busybox.tar.xz' | tar xfJC - ${ROOTFS} \
+      ;; \
+    *) \
+      curl -o- -sSL 'https://github.com/docker-library/busybox/raw/a0558a9006ce0dd6f6ec5d56cfd3f32ebeeb815f/glibc/busybox.tar.xz' | tar xfJC - ${ROOTFS} \
+      ;; \
+    esac
```
cyphar: This should be a script that both this and `helpers.bash` call into. No point in replicating this logic in more than one place.

tests/integration/helpers.bash
```diff
@@ -270,7 +278,14 @@ function setup_busybox() {
 		BUSYBOX_IMAGE="/testdata/busybox.tar"
 	fi
 	if [ ! -e $BUSYBOX_IMAGE ]; then
-		curl -o $BUSYBOX_IMAGE -sSL 'https://github.com/docker-library/busybox/raw/a0558a9006ce0dd6f6ec5d56cfd3f32ebeeb815f/glibc/busybox.tar.xz'
+		case $(dpkg --print-architecture) in
```
cyphar: Please use `$(go env GOARCH)`.

tests/integration/helpers.bash
lubinsz: OK. I will update it.

tests/integration/helpers.bash
lubinsz: ok, I will update it.

tests/integration/helpers.bash
lubinsz: @cyphar I uploaded the file of 'hello-world-aarch64.tar' in 'runc/tests/integration/testdata/'.

The file of 'hello-world.tar' can't be used on Arm64 platform, because the binary files with x86 format can't be used on Arm platform.

cyphar: Ah, I didn't notice that you included the `.tar` in `testdata` (I thought we always downloaded the rootfs archive).

Dockerfile
lubinsz: @cyphar 
Hi, I have moved all the multi-arch related code into 'tests/integration/multi-arch.bash'.
Please check it.
Thanks.

lubinsz:
@cyphar 
Hi, what's your opinion?

lubinszARM:
@avagin
@caniszczyk
@crosbymichael
@cyphar
@dqminh
@hqhq
@mrunalp
@rjnagal
Hi all,
We need to enable integration test in arm servers. So could you help to check it?
Thanks.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1648, opened 2017-11-10
 rootfs: preserve owner of tmpfs mount destination

cyphar:
Previously we only used SecureJoin for "tmpfs" and "bind" mounts
directly. Instead we should be using it for all mount types to avoid
possible issues with other mount exploits. In addition, always do the
necessary verification of mounts to further avoid /proc over-mount
hijinks.

Previously we would only preserve the mode of a tmpfs mount destination
if it already existed, but we would change the owner to root:root.
Preserve the owner to remain more consistent.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -199,16 +199,17 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string) error {
 		}
 		return nil
 	case "tmpfs":
-		copyUp := m.Extensions&configs.EXT_COPYUP == configs.EXT_COPYUP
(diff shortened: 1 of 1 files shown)


cyphar:
I've just double-checked the spec and it looks like providing a host absolute path for a mountpoint destination is actually not allowed by the spec (i.e. we are currently not doing the right thing). The test failure is something I'm working on fixing now.

@crosbymichael Am I right? I seem to recall that Docker uses this, so disabling this would break Docker (possibly).

crosbymichael:
I don't think Docker uses full paths for the dest, its always the path inside the container

PR #1651, opened 2017-11-15
systemd: adjust CPUQuotaPerSecUSec to compensate for systemd internal handling

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1509467#c6

@derekwaynecarr @jupierce @mrunalp

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -271,6 +271,13 @@ func (m *Manager) Apply(pid int) error {
 	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
 	if c.Resources.CpuQuota != 0 && c.Resources.CpuPeriod != 0 {
 		cpuQuotaPerSecUSec := uint64(c.Resources.CpuQuota*1000000) / c.Resources.CpuPeriod
+		// systemd converts CPUQuotaPerSecUSec (microseconds per CPU second) to CPUQuota
+		// (integer percentage of CPU) internally.  This means that if a fractional percent of
+		// CPU is indicated by Resources.CpuQuota, we need to round up to the nearest
+		// 10ms (1% of a second) such that child cgroups can set the cpu.cfs_quota_us they expect.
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
This was a fun one.

LGTM

sjenning:
wait on this, my rounding is wrong for quotas that don't need adjustment.  i'll fix in a minute.

sjenning:
ok should be good now

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -271,6 +271,13 @@ func (m *Manager) Apply(pid int) error {
 	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
 	if c.Resources.CpuQuota != 0 && c.Resources.CpuPeriod != 0 {
 		cpuQuotaPerSecUSec := uint64(c.Resources.CpuQuota*1000000) / c.Resources.CpuPeriod
+		// systemd converts CPUQuotaPerSecUSec (microseconds per CPU second) to CPUQuota
+		// (integer percentage of CPU) internally.  This means that if a fractional percent of
+		// CPU is inidicated by Resources.CpuQuota, we need to round up to the nearest
```
TomSweeneyRedHat: nit inidicated to indicated

sjenning:
@mrunalp can I get review/approval on this? I need one other approver as well.

dqminh:
Its a bit ugly though, do you think we should just let systemd create the slice without limits, then set the limits ourselves ?

sjenning:
@dqminh it is not a good idea to change values in cgroups that systemd thinks it controls.  We have encountered issues where things like a daemon-reload will cause those values to be lost.

dqminh:
LGTM

@sjenning yah i think i had some similar bugs like that in the past I thought that would be fixed in recent version of systemd though ( but im not too sure as we dont run systemd cgroup now). Probably need more investigations though, meanwhile i think this works fine as a workaround.

mrunalp:
LGTM

PR #1653, opened 2017-11-16
Remove cgo usage in libcontainer/mount

tklauser:
The getmntinfo C function on FreeBSD uses an internal static buffer to store the results and returns a pointer to this buffer. This is not thread safe. Instead do the lookup using getfsstat (which getmntinfo also does) but allocate the buffer in the caller.

This allows to get rid of cgo usage in libcontainer.

Requires an update of golang.org/x/sys which is done in the first commit.


crosbymichael:
The change looks fine but shouldn't we just remove the freebsd files altogether?

tklauser:
> The change looks fine but shouldn't we just remove the freebsd files altogether?

Sent PR #1664 to do that and remove all FreeBSD specific bits.

PR #1657, opened 2017-11-20
Treat EROFS in cgroups setup as skippable error

rutsky:
Currently cgroup setup ignores permissions errors.
AFAIK this helps with rootless containers: if user have permissions to
change cgroups, it will have them setup, otherwise rootless container
will use parent process cgroups (as it was with rootless containers
before cgroups cgroups support was introduced).
If cgroup is mounted in read-only mode (e.g. inside Docker container),
operations will return not permission error, but EROFS - this patch
treats EROFS as skippable error in cgroups setup.

Signed-off-by: Vladimir Rutsky <rutsky@google.com>

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"syscall"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
@@ -100,6 +101,22 @@ type cgroupData struct {
 	pid       int
 }
 
+func isIgnorableError(err error) bool {
+	if os.IsPermission(err) {
+		return true
+	}
+
+	if perr, ok := err.(*os.PathError); ok {
+		switch perr.Err.(syscall.Errno) {
+		// Read-only file system.
+		case syscall.EROFS:
+			return true
+		}
+	}
+
+	return false
+}
+
 func (m *Manager) Apply(pid int) (err error) {
 	if m.Cgroups == nil {
 		return nil
@@ -145,7 +162,7 @@ func (m *Manager) Apply(pid int) (err error) {
 		m.Paths[sys.Name()] = p
 
 		if err := sys.Apply(d); err != nil {
-			if os.IsPermission(err) && m.Cgroups.Path == "" {
+			if isIgnorableError(err) && m.Cgroups.Path == "" {
 				// If we didn't set a cgroup path, then let's defer the error here
 				// until we know whether we have set limits or not.
 				// If we hadn't set limits, then it's ok that we couldn't join this cgroup, because


rutsky:
Since it changes your code, @williammartin, can you take a look at this PR?

williammartin:
@rutsky Your summary of why we do the permission check for rootless containers is correct and it doesn't look like this PR will break that. I'm not super sure about the EROFS check logic and the expected behaviour though because I'm not familiar with Docker. If the whole cgroup tree is mounted read-only inside a docker container, how do non-rootless containers work?

The only thing I'd note is that `isPermissionError` should maybe be something like `isIgnorableError`, because the read-only nature is no longer related to permissions imo.

cyphar:
@williammartin 

> If the whole cgroup tree is mounted read-only inside a docker container, how do non-rootless containers work?

This PR is about *inside* a container, where cgroupfs is mounted as ro -- obviously it's rw on the host.

williammartin:
Yeh I got that, but I think I answered my own question which is that the referenced issue https://github.com/opencontainers/runc/issues/1658 is dealing with unprivileged docker containers, so rootless containers inside are the only thing that would ever work anyways.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -144,8 +145,24 @@ func (m *Manager) Apply(pid int) (err error) {
 		}
 		m.Paths[sys.Name()] = p
 
+		isPermissionError := func(err error) bool {
```
cyphar: Please define this as an actual function (preferably called `isIgnorableError`) rather than using the `a := func(...) {...}` form.

rutsky: Done. PTAL.

rutsky:
@williammartin thanks for the review!

Relatively offtopic to this PR question: what if cgroups setup will fail due to permission error on non-rootless container? E.g. if runc will be run as root and due to some reasons cgroup setup will return permission error (not sure if this is possible, but also I don't see why this is impossible), wouldn't this lead to security issues or unconstrained (in terms of resources) processes inside container?

cyphar:
@rutsky That is something that I brought up in #1540 as a point of concern, but the only places where (currently) you'll get an EPERM in `cgroupfs` are places that don't affect a process that has `CAP_DAC_OVERRIDE` (the errors are all related to permission checks done in addition to the VFS checks). All other errors are usally `EINVAL` or similar.

However, we should only use `isIgnorableError` if we're running as rootless (something I wanted to do in the original PR) but I believe the problem at the time was that the `libcontainer/cgroup` API didn't pass down the top-level `*configs.Config` so getting `config.Rootless` was non-trivial. If you can fix this, that'd be greatly appreciated.

AkihiroSuda:
ping @rutsky 

This PR seems the good alternative for https://github.com/opencontainers/runc/pull/1691 , and +1 for honoring `config.Rootless`.

AkihiroSuda:
@cyphar Could you set `rootless-containers` label to this one as well?

cyphar:
Done. I will go through a proper review of all the `rootless-containers` PRs next week (I'm at a conference this week).

cyphar:
This change looks okay in principle, but I think that (harrowing back to my comments in #1540) we should be only applying the `isIgnorable` check if we are running in rootless mode. So I reckon that the `Manager` should have a field that says whether we are rootless.

@rutsky I can carry this PR for you and do it, if you are no longer interested in updating this.

cyphar:
Closing in favour of #1759 which carries this.

PR #1661, opened 2017-11-21
Ensure container tests do not write on the host

danail-branekov:
TestGetContainerStateAfterUpdate creates its state.json file on the current
directory which turns out to be the host runc directory. Thus whenever
the test completes it leaves the state.json file behind thus
a) poluting the local git repository
b) changing the host file system violating the principle of doing
everything in an isolated container environment

This change would create a new temporary (in-container) directory and use it as
linuxContainer.root

Signed-off-by: Tom Godkin <tgodkin@pivotal.io>

### DIFF
--- a/libcontainer/container_linux_test.go
+++ b/libcontainer/container_linux_test.go
@@ -4,6 +4,7 @@ package libcontainer
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


libcontainer/container_linux_test.go
```diff
@@ -281,8 +282,16 @@ func TestGetContainerStateAfterUpdate(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	rootDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal("failed")
```
hqhq: You can just print the error with `t.Fatal(err)`.

libcontainer/container_linux_test.go
```diff
@@ -281,8 +282,16 @@ func TestGetContainerStateAfterUpdate(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	rootDir, err := ioutil.TempDir("", "")
```
cyphar: Please give it a prefix. If you can't think of one make it just the test name.

danail-branekov:
@hqhq, @cyphar Thanks, both issues are now addressed

hqhq:
LGTM

Thanks.

cyphar:
LGTM.

PR #1654, opened 2017-11-16
remove placeholder for non-linux platforms

dqminh:
runc currently only support Linux platform, and since we dont intend to expose
the support to other platform, removing all other platforms placeholder code.

`libcontainer/configs` still being used in https://github.com/moby/moby/blob/master/daemon/daemon_windows.go so
keeping it for now.

cc @opencontainers/runc-maintainers @justincormack @jhowardmsft

PR #1664, opened 2017-11-24
libcontainer: drop FreeBSD support

tklauser:
runc is not supported on FreeBSD, so remove all FreeBSD specific bits.

As suggested by @crosbymichael in #1653


cyphar:
LGTM, thanks!

dqminh:
LGTM

i am doing the same thing in https://github.com/opencontainers/runc/pull/1654 but we can merge this first so the other change is smaller !

Issue #1662, opened 2017-11-21
Collect events from multiple containers via a single command


danail-branekov:
Collect events from multiple containers via a single command

Collecting events from all containers running on a single host is a significant part of any containers monitoring use case. When there are lots of containers running on a single host, one needs to invoke `runc events` per every container and this is expensive. It would be great if one could get the stats of multiple containers via a single command.

We (Cloud Foundry) are willing to PR such a change so that one could execute something like 
```runc events --stats <comma separated list of containers>```
and receive a JSON array of `Stats`. 

With `runc` output backwards compatibility in mind I would imagine that the user should request events from multiple containers explicitly, via e.g. a dedicated flag. For example, the command would look like this
```runc events --stats --containers=container1,container2,container3```

And here goes a sample output:
```json
[
	{"type":"stats", "id":"container1","data":{…}},
	{"type":"stats", "id":"container2","data":{…}},
	{"type":"stats", "id":"container3","data":{…}},
]
```

What do you guys think?

danail-branekov:
On the other hand, the events produced by `runc events --interval` are not grouped in an array, each line is a separate stat. Therefore maybe it makes sense to *not* group the stats from different containers into an array, but simply print every one of them on a separate line. Such an approach would simplify the implementation as the single argument (`container_id`) could be treated as a comma separated list of container ids which is backwards compatible.

williammartin:
I think I'd prefer a command that did:

`runc events --stats --container A --container B --container C` 

or

`runc events --stats A B C`

What do you think? I've never been super keen on CSV args.

hqhq:
runc is supposed to be a simple and low level tool that handles container runtime, any high level tools or even a script work based on runc could fulfill your needs, so I don't think we should poke this hole in runc, otherwise people would ask for `runc kill cont1 cont2` and `runc start cont1 cont2` and so on.

I'll close the issue, feel free to comment if you have more thoughts.

williammartin:
@hqhq Understood about opening up the product surface area. For us, the collection of stats is a much more frequent occurrence than `kill` or `start` etc since we collect stats for each container every 30s or something. This means that every interval we are having to spawn many runc processes when one would do. 

Do you think the performance implications are not enough to open that pandoras box? Do you have any other suggestions?

cyphar:
[...]

PR #1666, opened 2017-11-30
Support multiple containers in events subcommand

danail-branekov:
This PR implements #1662

`runc events` can now accept multiple container IDs, for example `runc events containerA containerB`. It outputs the stats from all the specified containers in separate JSON objects, i.e. are not be grouped in a JSON array.
* If any of the specified containers does not exist or is in `Stopped` state, an error is returned
* `runc events --interval` watches all specified containers and exists when any of them is deleted
* The PR introduces a test to verify that OOM events are being sent to the output

### DIFF
--- a/events.go
+++ b/events.go
@@ -4,8 +4,10 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -22,6 +24,10 @@ type event struct {
 	Type string      `json:"type"`
 	ID   string      `json:"id"`
 	Data interface{} `json:"data,omitempty"`
+}
+
+func newEvent(eventType, id string, data interface{}) *event {
+	return &event{Type: eventType, ID: id, Data: data}
 }
 
 // stats is the runc specific stats structure for stability when encoding and decoding stats.
@@ -118,9 +124,9 @@ type intelRdt struct {
 var eventsCommand = cli.Command{
 	Name:  "events",
 	Usage: "display container events such as OOM notifications, cpu, memory, and IO usage statistics",
-	ArgsUsage: `<container-id>
+	ArgsUsage: `<container-ids>
 
-Where "<container-id>" is the name for the instance of the container.`,
+Where "<container-ids>" are the names for the container instances.`,
 	Description: `The events command displays information about the container. By default the
 information is displayed once every 5 seconds.`,
 	Flags: []cli.Flag{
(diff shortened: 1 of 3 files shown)


williammartin:
@hqhq and @cyphar any thoughts on this?

events.go
```diff
@@ -293,3 +297,118 @@ func convertL3CacheInfo(i *intelrdt.L3CacheInfo) *l3CacheInfo {
...
+			}
+
+			for {
+				select {
+				case <-signal:
+					return
+				case _, ok := <-oomNotification:
+					if !ok {
+						// the channel was closed because the container stopped and the cgroups no longer exist.
+						close(signal)
```
hqhq: Why cancel all oom monitors when one container got stopped?

tests/integration/events.bats
```diff
@@ -107,3 +131,145 @@ function teardown() {
...
+  wait
+
+  [ -e events.log ]
+
+  run cat events.log
+  printf '%s\n' "${lines[@]}"
+  expect_oom_for_container "test_busybox" "${lines[@]}"
+}
+
+function expect_stats_for_container() {
```
cyphar: Both of these functions can be implemented using grep AFAICS, which would be far more readable and wouldn't require having separate functions.

tests/integration/events.bats
```diff
@@ -107,3 +131,145 @@ function teardown() {
...
+
+@test "events --interval test_busybox (should report oom events)" {
+  # XXX: currently cgroups require root containers.
+  requires root
+
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  # set memory limit to 10M
+  run eval "echo 10000000 > /sys/fs/cgroup/memory/test_busybox/memory.limit_in_bytes"
```
cyphar: Why not use `runc update`?

tests/integration/events.bats
```diff
@@ -107,3 +131,145 @@ function teardown() {
...
+  runc events --stats non-existent-container-1 non-existent-container-2 test_busybox
+  [ "$status" -ne 0 ]
+  [[ "${lines[0]}" =~ "non-existent-container-1" ]]
+  [[ "${lines[0]}" =~ "does not exist" ]]
+  [[ "${lines[1]}" =~ "non-existent-container-2" ]]
+  [[ "${lines[1]}" =~ "does not exist" ]]
+  [[ "${lines[@]}" != *"test_busybox"* ]]
+}
+
+@test "events --interval non-existent-container-1 non-existent-container-2 test_busybox (bails when any container doesn't exist)" {
```
cyphar: This is a duplicate of the previous test.

events.go
danail-branekov: At CloudFoundry we do not have a use case for `runc events --interval` that could drive the implementation to address questions like:
* Should runc exit when a container exits?
* Should runc keep running forever, or it should exit once all monitored containers exit? 

That is why we took the easiest implementation approach to make `runc` exit when one of the monitored container is gone. Happy to revisit the code if you could suggest some reasonable behaviour.

cyphar: I would expect that `runc` would exit when all of the containers you asked to monitor have died. Since this implementation already uses `sync.WaitGroup` it's odd that it's not done in the more straight-forward way IMO.

williammartin: I sort of feel like if I was running `runc events --interval 5s container1 container2 ... container100`, I would not want to keep track of "disappearing containers" that are no longer emitting metrics. Another view would be that I said "give me the events for all these containers" but now runc can't fulfil that, so bail out instead of confusing me.

I could also totally buy that it should carry on "best-effort".

hqhq: > I said "give me the events for all these containers" but now runc can't fulfil that

I would simply think runc is still fulfilling that, you can't get more events from some containers because they are exited and no events are generated from them any more.

`runc events` don't exit until all monitored containers are gone makes more sense to me.

tests/integration/events.bats
BooleanCat: The previous test is making an assertion about the output of `runc events --stats` whereas this one concerns `runc events --interval`

tests/integration/events.bats
BooleanCat: Done.

events.go
williammartin: Alright cool. Happy to defer to both of you and fix this up. Thanks for the feedback.

danail-branekov: We've changed the behaviour so that one container stopping only prevents metrics from that container being emitted to `runc events`. Also, `runc events` exits when all containers exit.

danail-branekov:
@cyphar All comments have been addressed, please review

danail-branekov:
@hqhq and @cyphar, any thoughts?

BooleanCat:
@cyphar @hqhq I'd like to revive this. We've hopefully addressed all comments raised in this PR, do you have any more thoughts on this? Getting this merged would help us (Cloud Foundry's Garden) decrease the overhead of gathering metrics for our application containers.

Cheers!

danail-branekov:
@crosbymichael Well, we _could_ implement something that collects metrics from cgroups directly but it would feel weird to use runc APIs _and_ operating system internals such as cgroups at the same time. After all, runc is all about providing nice container management API and container runtimes are supposed to only provide a thin integration layer on top of it.
Is there something about such a change that makes you feel uncomfortable?

Callisto13:
@cyphar @hqhq do you have any further thoughts on this?

ostenbom:
Ping @crosbymichael @cyphar @hqhq, this'd still help us if we could get this merged!

BooleanCat:
Ping

kolyshkin:
`events` is not part of runtime-spec, I think we may even need to drop it (or move the functionality to a separate tool).

PR #1675, opened 2017-12-11
RFC: libcontainer: remove dependency on libapparmor

### LINKED ISSUE #1674: RFC: drop dependency on libapparmor and reimplement aa_change_exec in Go
libapparmor is integrated in libcontainer using cgo but is only used to call a single function: `aa_change_onexec` in https://github.com/opencontainers/runc/blob/master/libcontainer/apparmor/apparmor.go#L35.

The dependency on libapparmor requires additional packages to be installed for runc/libcontainer to be built and run, which might cause confusion for some users (e.g see #1486).

Looking at  `aa_change_onexec` in libapparmor it seems like is simple enough (writing a string to a file in /proc/<n>/attr/...) to be re-implemented locally in libcontainer in plain Go.

I have an untested PR ready locally which I can submit if people think this is a worthwhile thing to do. In the meantime I'll try to figure out how to test this properly (i.e. getting runc to run with an apparmor profile).

tklauser:
libapparmor is integrated in libcontainer using cgo but is only used to
call a single function: aa_change_onexec. It turns out this function is
simple enough (writing a string to a file in /proc/<n>/attr/...) to be
re-implemented locally in libcontainer in plain Go.

This allows to drop the dependency on libapparmor and the corresponding
cgo integration.

Fixes #1674

### DIFF
--- a/libcontainer/apparmor/apparmor.go
+++ b/libcontainer/apparmor/apparmor.go
@@ -2,15 +2,10 @@
 
 package apparmor
 
-// #cgo LDFLAGS: -lapparmor
-// #include <sys/apparmor.h>
-// #include <stdlib.h>
-import "C"
 import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"unsafe"
 )
 
 // IsEnabled returns true if apparmor is enabled for the host.
@@ -24,16 +19,36 @@ func IsEnabled() bool {
 	return false
 }
 
+func setprocattr(attr, value string) error {
+	// Under AppArmor you can only change your own attr, so use /proc/self/
+	// instead of /proc/<tid>/ like libapparmor does
+	path := fmt.Sprintf("/proc/self/attr/%s", attr)
+
+	f, err := os.OpenFile(path, os.O_WRONLY, 0)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	_, err = fmt.Fprintf(f, "%s", value)
+	return err
+}
+
+// changeOnExec reimplements aa_change_onexec from libapparmor in Go
+func changeOnExec(name string) error {
+	value := "exec " + name
+	if err := setprocattr("exec", value); err != nil {
+		return fmt.Errorf("apparmor failed to apply profile: %s", err)
+	}
+	return nil
+}
+
 // ApplyProfile will apply the profile with the specified name to the process after
 // the next exec.
 func ApplyProfile(name string) error {
 	if name == "" {
 		return nil
 	}
-	cName := C.CString(name)
-	defer C.free(unsafe.Pointer(cName))
-	if _, err := C.aa_change_onexec(cName); err != nil {
-		return fmt.Errorf("apparmor failed to apply profile: %s", err)
-	}
-	return nil
+
+	return changeOnExec(name)
 }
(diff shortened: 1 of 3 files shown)


tklauser:
As noted in #1674, I did not fully test this yet. I'm trying to get runc running with an apparmor profile to test this first.

libcontainer/apparmor/apparmor.go
```diff
@@ -24,15 +21,29 @@ func IsEnabled() bool {
...
 	if name == "" {
 		return nil
 	}
-	cName := C.CString(name)
-	defer C.free(unsafe.Pointer(cName))
-	if _, err := C.aa_change_onexec(cName); err != nil {
+
+	// mimick aa_change_onexec from libapparmor in Go
+	value := "exec " + name
+	if err := setprocattr(unix.Gettid(), "exec", value); err != nil {
```
cyphar: This looks /okay/ overall, but I would prefer if we had a function called `ChangeOnExec` (or whatever) which wraps these two lines -- to make it clearer what we are mimicing.

libcontainer/apparmor/apparmor.go
```diff
@@ -24,15 +21,29 @@ func IsEnabled() bool {
...
+	_, err = fmt.Fprintf(f, "%s", value)
+	return err
+}
+
 // ApplyProfile will apply the profile with the specified name to the process after
 // the next exec.
 func ApplyProfile(name string) error {
 	if name == "" {
 		return nil
 	}
```
AkihiroSuda: Should check `!strings.Contains(name, "/")`

cyphar: Why? Profile names can have `/` in them?

libcontainer/apparmor/apparmor.go
tklauser: Agree. Does it need to be exported or could it also be `changeOnExec`?

libcontainer/apparmor/apparmor.go
tklauser: Would something like the following be sufficient?

```
if strings.Contains(name, "/") {
    return fmt.Errorf("apparmor profile name cannot contain '/'")
}
```

libcontainer/apparmor/apparmor.go
tklauser: PTAL

cyphar: Yeah, there's no need to export it I guess.

cyphar:
This looks fine. I will test it though (though I imagine it will work fine -- the only really ugly parts of the AppArmor API are related to profile loading and unloading).

libcontainer/apparmor/apparmor.go
AkihiroSuda: At least ".." should be prohibited to prevent potential filesystem contamination

AkihiroSuda: np, I misread the code

libcontainer/apparmor/apparmor.go
```diff
@@ -24,16 +21,34 @@ func IsEnabled() bool {
 	return false
 }
 
+func setprocattr(tid int, attr, value string) error {
+	path := fmt.Sprintf("/proc/%d/attr/%s", tid, attr)
```
cyphar: While this works, I think this should be `/proc/self/attr/%s`. Under AppArmor you can only change your own `attr`, so I think `/proc/self` makes more sense than providing a `tid` argument.

tklauser: Updated, PTAL.

cyphar:
LGTM. Tested this with a few profiles and they are applied properly.

crosbymichael:
LGTM

PR #1678, opened 2017-12-14
libcontainer: Do not wait for signalled processes if subreaper is set

### LINKED ISSUE #1677: Don't wait for signalled processes if subreaper is enabled
In case the consumer of libcontainer sets a subreaper, we should not wait for processes when signalling them (https://github.com/opencontainers/runc/blob/master/libcontainer/init_linux.go#L467-L517). Otherwise, we have a race about the reaping, and in case the reaping is done from this function, our subreaper will not reap the expected process, meaning that we're gonna miss the exit code related to this process.

sboeuf:
When a subreaper is enabled, it might expect to reap a process and
retrieve its exit code. That's the reason why this patch is giving
the possibility to define the usage of a subreaper as a consumer of
libcontainer. Relying on this information, libcontainer will not
wait for signalled processes in case a subreaper has been set.

Fixes #1677

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -494,6 +494,16 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 		logrus.Warn(err)
 	}
 
+	subreaper, err := system.GetSubreaper()
+	if err != nil {
+		// The error here means that PR_GET_CHILD_SUBREAPER is not
+		// supported because this code might run on a kernel older
+		// than 3.4. We don't want to throw an error in that case,
+		// and we simplify things, considering there is no subreaper
+		// set.
+		subreaper = 0
+	}
+
 	for _, p := range procs {
 		if s != unix.SIGKILL {
 			if ok, err := isWaitable(p.Pid); err != nil {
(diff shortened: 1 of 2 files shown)


sboeuf:
@mrunalp @crosbymichael @sameo 
This is an attempt to fix #1677

crosbymichael:
Instead of using a bool set my the user just use `PR_GET_CHILD_SUBREAPER` to see if it has been set or not.

sboeuf:
@crosbymichael  oh yeah I haven't thought about this way, but this will simplify things :)
A detail though, do you expect to check `PR_GET_CHILD_SUBREAPER` every time we want to know if a subreaper is set, or do you want to make this check only once and store this value ?

crosbymichael:
Probably safer each time to check it.

sboeuf:
@crosbymichael sounds fair to me !

sboeuf:
Just updated it !

libcontainer/init_linux.go
```diff
@@ -494,6 +494,11 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 		logrus.Warn(err)
 	}
 
+	subreaper, err := system.GetSubreaper()
+	if err != nil {
+		return err
```
crosbymichael: One thing, this will fail on kernels older than 3.4 that don't support it.  I would maybe ignore the error and set subreaper to `0` in this case.

sboeuf: Good point. We don't want to fail in such a case since the application is still supposed to work.

sboeuf: @crosbymichael I think I should make this a bit smarter, by assigning `subreaper = 0` only when the error is `EINVAL`. WDYT ?

sboeuf: No nevermind, in case of `PR_GET_CHILD_SUBREAPER`, there is no specific errno set.

crosbymichael: Sounds good but I don't know what returning this error here would accomplish in the execution flow.  Either works for me.

sboeuf: I have updated the PR with the simple case. We fallback to original behavior in case getting subreaper status fails.

crosbymichael:
LGTM

sboeuf:
@mrunalp @cyphar WDYT ?

mrunalp:
LGTM

PR #1682, opened 2018-01-04
Pin version of gojsonschema in tests

### LINKED ISSUE #1680: Integration test failing on master
Running runc tests with `make test`, I see a failing integration test:

```
bats -t tests/integration/spec.bats
1..3
ok 1 spec generation cwd
ok 2 spec generation --bundle
not ok 3 spec validator
# (in test file tests/integration/spec.bats, line 92)
#   `./validate src/runtime-spec/schema/config-schema.json config.json' failed
# runc spec (status=0):
```

I can see this test is also failing in your CI. I've looked into the cause of this and found something interesting. There's a `go get` in the test script [here](https://github.com/opencontainers/runc/blob/eb5bd4fa6adfe9f2e41a5a5a2463857ae5f7a0c6/tests/integration/spec.bats#L81) that pulls github.com/xeipuuv/gojsonschema.

It seems a change to that dependency is breaking this test because if I add this after the go get:

`( cd "${GOPATH}/src/github.com/xeipuuv/gojsonschema" && git reset --hard 6637feb73ee44cd4640bb3def285c29774234c7f )`

... then the test passes. Seems like [this commit](https://github.com/xeipuuv/gojsonschema/commit/83a7f6369d552cba07cdeea7dc622f1161a07deb) is the culprit.

I'll figure out if the issue lives in runc or gojsonschema then submit a PR to the appropriate repo.

BooleanCat:
Fixes #1680 (partially). This is suggested as a temporary fix to runc so that CI can go green and development can continue.

We will detail a potential problem we've found with opencontainers json schema validation in a email to the mailing list.

Signed-off-by: Will Martin <wmartin@pivotal.io>

### DIFF
--- a/tests/integration/spec.bats
+++ b/tests/integration/spec.bats
@@ -81,6 +81,9 @@ function teardown() {
   run bash -c "GOPATH='$GOPATH' go get github.com/xeipuuv/gojsonschema"
   [ "$status" -eq 0 ]
 
+  run git -C "${GOPATH}/src/github.com/xeipuuv/gojsonschema" reset --hard 6637feb73ee44cd4640bb3def285c29774234c7f
+  [ "$status" -eq 0 ]
+
   GOPATH="$GOPATH" go build src/runtime-spec/schema/validate.go
   [ -e ./validate ]
 


tests/integration/spec.bats
```diff
@@ -81,6 +81,9 @@ function teardown() {
   run bash -c "GOPATH='$GOPATH' go get github.com/xeipuuv/gojsonschema"
   [ "$status" -eq 0 ]
 
+  run git -C "${GOPATH}/src/github.com/xeipuuv/gojsonschema" reset --hard 6637feb73ee44cd4640bb3def285c29774234c7f
```
stevvooe: Unit tests should not modify local environments like this. This will lose local changes.

BooleanCat: But gojsonschema isn't the thing under test, it's a dependency. The same pattern is already used to pin runtime-spec: https://github.com/opencontainers/runc/blob/master/tests/integration/spec.bats#L76

hqhq:
LGTM

tests/integration/spec.bats
stevvooe: This line makes two basically incorrect assumptions that will break:

1. An alternative version of gojsonschema may be in use that could fix this. We can never test with this.
2. A `GOPATH` may consist of multiple paths.

Just because the same pattern is in use doesn't mean that it is correct.

runcom:
Can we merge as a temporary stopgap to unblock other PRs? Many other repo added a workaround like this for now, we can create an issue to revert and fully fix this once we know what to do.

@mrunalp ptal

crosbymichael:
LGTM

A fix after this should be moving these out of the bats files.  It makes no sense that deps are scattered and hidden in various files.  ( @BooleanCat  not your fault, its good that you kept it consistent for now )

PR #1683, opened 2018-01-07
Fix race against systemd

vikaschoudhary16:
- T0: runc triggers a systemd unit creation asynchronously from [here](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/apply_systemd.go#L298)
- T1: runc then moves ahead and starts creating cgroup paths(.scope directories), [here](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/apply_systemd.go#L348). Kernel creates .scope directory and cgroup.procs file(along with other default files) in the directory automatically, in an atomic manner.
- T3: systemd execution thread which was invoked at time `T0`, is still in the process of unit creation. systemd also trying to create cgroup paths and deletes the `.scope` directory which is created at time `T1` by runc from [here](https://github.com/systemd/systemd/blob/v219/src/shared/cgroup-util.c#L1630) in the code

Fixes https://github.com/openshift/origin/issues/16246

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -295,9 +295,12 @@ func (m *Manager) Apply(pid int) error {
 		}
 	}
 
-	if _, err := theConn.StartTransientUnit(unitName, "replace", properties, nil); err != nil && !isUnitExists(err) {
+	statusChan := make(chan string)
+	if _, err := theConn.StartTransientUnit(unitName, "replace", properties, statusChan); err != nil && !isUnitExists(err) {
 		return err
 	}
+
+	<-statusChan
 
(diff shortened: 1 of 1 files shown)


runcom:
Looks ok, /cc @cyphar

runcom:
You need to sign your commit @vikaschoudhary16

mrunalp:
LGTM
We have a similar fix in cri-o for waiting on transient units.

hqhq:
LGTM

hqhq:
CI issues is addressed in https://github.com/opencontainers/runc/pull/1682

smileusd:
@vikaschoudhary16 I still see this issue on runc v1.1.2, containerd 1.6.6, systemd 145
https://github.com/containerd/containerd/issues/7506

PR #1688, opened 2018-01-13
main: support rootless mode in userns

AkihiroSuda:
Running rootless containers in userns is useful for mounting
filesystems (e.g. overlay) with mapped euid 0, but without actual root
privilege.

Usage: (Note that `unshare --mount` requires `--map-root-user`)
```console
  user$ mkdir lower upper work rootfs
  user$ curl http://dl-cdn.alpinelinux.org/alpine/v3.7/releases/x86_64/alpine-minirootfs-3.7.0-x86_64.tar.gz | tar Cxz ./lower || ( true; echo "mknod errors were ignored" )
  user$ unshare --mount --map-root-user
  mappedroot# runc spec --rootless
  mappedroot# sed -i 's/"readonly": true/"readonly": false/g' config.json
  mappedroot# mount -t overlay -o lowerdir=./lower,upperdir=./upper,workdir=./work overlayfs ./rootfs
  mappedroot# runc run foo
```

Note: unprivileged overlay is only supported in Ubuntu and a few distros (http://kernel.ubuntu.com/git/ubuntu/ubuntu-artful.git/commit/fs/overlayfs?h=Ubuntu-4.13.0-25.29&id=0a414bdc3d01f3b61ed86cfe3ce8b63a9240eba7)

### DIFF
--- a/libcontainer/system/linux.go
+++ b/libcontainer/system/linux.go
@@ -3,6 +3,7 @@
 package system
 
 import (
+	"os"
 	"os/exec"
 	"syscall" // only for exec
 	"unsafe"
@@ -119,6 +120,22 @@ func UIDMapInUserNS(uidmap []user.IDMap) bool {
 		return false
 	}
 	return true
+}
+
+// GetParentNSeuid returns the euid within the parent user namespace
+func GetParentNSeuid() int {
+	euid := os.Geteuid()
+	uidmap, err := user.CurrentProcessUIDMap()
+	if err != nil {
+		// This kernel-provided file only exists if user namespaces are supported
+		return euid
+	}
+	for _, um := range uidmap {
+		if um.ID <= euid && euid <= um.ID+um.Count-1 {
+			return um.ParentID + euid - um.ID
+		}
+	}
+	return euid
 }
 
 // SetSubreaper sets the value i as the subreaper setting for the calling process
--- a/libcontainer/system/linux.go
+++ b/libcontainer/system/linux.go
@@ -3,13 +3,11 @@
 package system
 
 import (
-	"bufio"
-	"fmt"
-	"os"
 	"os/exec"
 	"syscall" // only for exec
 	"unsafe"
 
+	"github.com/opencontainers/runc/libcontainer/user"
 	"golang.org/x/sys/unix"
 )
 
@@ -102,29 +100,22 @@ func Setctty() error {
 }
 
 // RunningInUserNS detects whether we are currently running in a user namespace.
-// Copied from github.com/lxc/lxd/shared/util.go
+// Originally copied from github.com/lxc/lxd/shared/util.go
 func RunningInUserNS() bool {
-	file, err := os.Open("/proc/self/uid_map")
+	uidmap, err := user.CurrentProcessUIDMap()
 	if err != nil {
 		// This kernel-provided file only exists if user namespaces are supported
 		return false
 	}
-	defer file.Close()
+	return UIDMapInUserNS(uidmap)
+}
 
-	buf := bufio.NewReader(file)
-	l, _, err := buf.ReadLine()
-	if err != nil {
-		return false
-	}
-
-	line := string(l)
-	var a, b, c int64
-	fmt.Sscanf(line, "%d %d %d", &a, &b, &c)
+func UIDMapInUserNS(uidmap []user.IDMap) bool {
 	/*
 	 * We assume we are in the initial user namespace if we have a full
 	 * range - 4294967295 uids starting at uid 0.
 	 */
-	if a == 0 && b == 0 && c == 4294967295 {
+	if len(uidmap) == 1 && uidmap[0].ID == 0 && uidmap[0].ParentID == 0 && uidmap[0].Count == 4294967295 {
 		return false
 	}
 	return true
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -234,7 +234,7 @@ func isRootless(context *cli.Context) (bool, error) {
 	// So we use system.GetParentNSeuid() here.
 	//
 	// TODO(AkihiroSuda): how to support nested userns?
-	return system.GetParentNSeuid() != 0, nil
+	return system.GetParentNSeuid() != 0 || system.RunningInUserNS(), nil
 }
 
 func createContainer(context *cli.Context, id string, spec *specs.Spec) (libcontainer.Container, error) {
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -16,6 +16,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/specconv"
+	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/runtime-spec/specs-go"
 
@@ -217,19 +218,37 @@ func createPidFile(path string, process *libcontainer.Process) error {
 	return os.Rename(tmpName, path)
 }
 
-// XXX: Currently we autodetect rootless mode.
-func isRootless() bool {
-	return os.Geteuid() != 0
+func isRootless(context *cli.Context) (bool, error) {
+	if context != nil {
+		b, err := parseBoolOrAuto(context.GlobalString("rootless"))
+		if err != nil {
+			return false, err
+		}
+		if b != nil {
+			return *b, nil
+		}
+		// nil b stands for "auto detect"
+	}
+	// Even if os.Geteuid() == 0, it might still require rootless mode,
+	// especially when running within userns.
+	// So we use system.GetParentNSeuid() here.
+	//
+	// TODO(AkihiroSuda): how to support nested userns?
+	return system.GetParentNSeuid() != 0, nil
 }
 
 func createContainer(context *cli.Context, id string, spec *specs.Spec) (libcontainer.Container, error) {
+	rootless, err := isRootless(context)
+	if err != nil {
+		return nil, err
+	}
 	config, err := specconv.CreateLibcontainerConfig(&specconv.CreateOpts{
 		CgroupName:       id,
 		UseSystemdCgroup: context.GlobalBool("systemd-cgroup"),
 		NoPivotRoot:      context.Bool("no-pivot"),
 		NoNewKeyring:     context.Bool("no-new-keyring"),
 		Spec:             spec,
-		Rootless:         isRootless(),
+		Rootless:         rootless,
 	})
 	if err != nil {
 		return nil, err
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -44,7 +44,11 @@ checkpointed.`,
 			return err
 		}
 		// XXX: Currently this is untested with rootless containers.
-		if isRootless() {
+		rootless, err := isRootless(context)
+		if err != nil {
+			return err
+		}
+		if rootless {
 			return fmt.Errorf("runc checkpoint requires root")
 		}
 
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/sirupsen/logrus"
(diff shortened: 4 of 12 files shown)


AkihiroSuda:
cc @cyphar

utils_linux.go
```diff
@@ -219,7 +220,12 @@ func createPidFile(path string, process *libcontainer.Process) error {
 
 // XXX: Currently we autodetect rootless mode.
 func isRootless() bool {
-	return os.Geteuid() != 0
+	// Even if os.Geteuid() == 0, it might still require rootless mode,
+	// especially when running within `unshare -m -r`.
+	// So we use system.GetParentNSeuid() here.
+	//
+	// TODO(AkihiroSuda): how to support nested userns?
```
AkihiroSuda: Rather than checking UID, it might be better to attempt some lightweight operation that requires "real" root privilege.

Any thought?

frezbo:
Does this PR address this issue: https://github.com/opencontainers/runc/issues/1658

AkihiroSuda:
@frezbo no

libcontainer/system/linux.go
```diff
@@ -130,6 +130,33 @@ func RunningInUserNS() bool {
 	return true
 }
 
+// GetParentNSeuid returns the euid within the parent user namespace
+func GetParentNSeuid() int {
```
cyphar: I think @brauner had a patch like this for Ubuntu? This code should also check `/proc/self/setgroups` and ensure that it's set to `deny`.

AkihiroSuda: As commented above, I think we need to find an alternative check to support nested userns

utils_linux.go
cyphar: The simplest test would be to do `mknod`, but I'm not sure I think this is the best idea in the world to be honest. There are plans in the future to provide the ability for previously disallowed syscalls to be allowed through the whole SECCOMP syscall emulation mechanism that's being worked on, so trying to pin down "we are in a user namespace because X doesn't work" is a bit of a stretch.

AkihiroSuda: `func isRootless() { return os.Geteuid() != 0 || readFile("/proc/self/setgroups") == "deny") }` is fine then?

AkihiroSuda:
Updated PR

- Parser commit eb671b is shared with https://github.com/opencontainers/runc/pull/1692 (bea27f10925086bd973d585e1aa10be883d84ba8)
- Add global bool flag `--force-rootless` to support nested userns.

I think this PR is mergeable now.

frezbo:
@AkihiroSuda I can't seem to build your branch, here;s the error:
```runc (aws:default)(git:unshare-m-r)$ make
go build -buildmode=pie  -ldflags "-X main.gitCommit="6f6720a754f7cc69274762586d678b85aecfbeba" -X main.version=1.0.0-rc4+dev " -tags "seccomp" -o runc .
# github.com/AkihiroSuda/runc
./signals.go:137:28: cannot use ws (type "github.com/AkihiroSuda/runc/vendor/golang.org/x/sys/unix".WaitStatus) as type "github.com/opencontainers/runc/vendor/golang.org/x/sys/unix".WaitStatus in argument to utils.ExitStatus
./utils_linux.go:230:9: undefined: system.GetParentNSeuid
./utils_linux.go:239:7: cannot use spec (type *"github.com/AkihiroSuda/runc/vendor/github.com/opencontainers/runtime-spec/specs-go".Spec) as type *"github.com/opencontainers/runc/vendor/github.com/opencontainers/runtime-spec/specs-go".Spec in field value
make: *** [Makefile:32: runc] Error 2
runc (aws:default)(git:unshare-m-r)$ 
```

AkihiroSuda:
@frezbo mv AkihiroSuda opencontainers

AkihiroSuda:
Any thought?
This is the (only) blocker for rootless containerd: https://github.com/containerd/containerd/pull/2006

AkihiroSuda:
On second thought I came up with a ternary CLI flag `--rootless=auto/true/false` rather than boolean `--force-rootless=true/false`. But I'm not sure such a ternary flag is acceptable.

cyphar:
I am currently on vacation unfortunately. I will do a review of this when I get back.

AkihiroSuda:
rebased

AkihiroSuda:
> On second thought I came up with a ternary CLI flag --rootless=auto/true/false rather than boolean --force-rootless=true/false. But I'm not sure such a ternary flag is acceptable.

Any thought about this?

Although `isRootless()` auto-detection might not be perfect, I think this PR can be mergeable when we agree on the CLI flag.
The auto-detection logic can be improved later.

cc @ehotinger (https://github.com/genuinetools/img/issues/69#issuecomment-375413286)

cyphar:
I think a tertiary flag is okay. I will need to re-review this.

AkihiroSuda:
@cyphar updated to use ternary flag

AkihiroSuda:
Is there any additional task needed?

cyphar:
LGTM, I've played around with this and it works pretty well. There isn't really a nice way of detecting whether we are in a user namespace in a non-destructive way (unfortunately) -- so this will have to do for now.

utils_linux.go
```diff
@@ -217,19 +218,37 @@ func createPidFile(path string, process *libcontainer.Process) error {
...
+			return *b, nil
+		}
+		// nil b stands for "auto detect"
+	}
+	// Even if os.Geteuid() == 0, it might still require rootless mode,
+	// especially when running within userns.
+	// So we use system.GetParentNSeuid() here.
+	//
+	// TODO(AkihiroSuda): how to support nested userns?
+	return system.GetParentNSeuid() != 0, nil
```
giuseppe: would it be safer to assume "yes" everytime it is running in an userNS (i.e. doesn't have the full 0-4294967295 range available in /proc/self/uid_map)?

cyphar: This is actually a good point. I think doing both (and if either is true, we assume we are in rootless mode) would work as well. The main problem with the 4294967295 check is that you can create user namespaces that have a full mapping but still require rootless tricks (and as I said above there's no real way of checking if that is true).

giuseppe: is that because it is still theoretically possible to setup the full range and have root mapped to some other user in the host namespace?  If I am not missing something, that sounds quite difficult to achieve as it requires an intermediate user namespace with at least two mappings (root in the intermediate namespace mapped to non root in the host) that still has enough privileges to setup the full range for the final userNS.

In any case, I fully agree that having both checks is safer

cyphar: > is that because it is still theoretically possible to setup the full range and have root mapped to some other user in the host namespace?

That is also possible, but it's actually much simpler than that. Even if you have a 1-to-1 mapping of all users in a new user namespace certain operations will still fail purely because you are not in `&init_user_ns` (such as `mknod(2)`) -- because capabilities are scoped to your user namespace and your host UID is irrelevant.

While this is an edge-case it means that in general we have no way of checking whether we are in a user namespace or the host user namespace. The closest thing we have is `ioctl(NS_GET_PARENT)` but returns `-EPERM` even if there is no parent -- I would argue that it should return `-ENOENT` in that case but @ebiederm might not agree.

giuseppe: In the kernel the inode for the initial user namespace is hardcoded, and it seems like a reliable (and undocumented AFAICS) way to detect if we are running in the init userNS:

```
$ cat > userns_checker.c << EOF
#include <sys/types.h>
#include <sys/stat.h>
#include <unistd.h>
#include <stdio.h>
#include <string.h>

int main ()
{
  struct stat sb;
  if (stat ("/proc/self/ns/user", &sb) < 0)
    return -1;
  printf ("is main userns? %s\n", 0xEFFFFFFDU == sb.st_ino ? "yes" : "no");
  return 0;
}
EOF

$ gcc -o userns_checker userns_checker.c

$ ./userns_checker 
is main userns? yes
$ unshare -r ./userns_checker 
is main userns? no
```

What do you think?

ebiederm:
Giuseppe Scrivano <notifications@github.com> writes:

> @giuseppe commented on this pull request.
>
> -------------------------------------------------------------------------------------------------------------------
>
> In utils_linux.go:
>
>> +	if context != nil {
> +		b, err := parseBoolOrAuto(context.GlobalString("rootless"))
> +		if err != nil {
> +			return false, err
> +		}
> +		if b != nil {
> +			return *b, nil
> +		}
> +		// nil b stands for "auto detect"
> +	}
> +	// Even if os.Geteuid() == 0, it might still require rootless mode,
> +	// especially when running within userns.
> +	// So we use system.GetParentNSeuid() here.
> +	//
> +	// TODO(AkihiroSuda): how to support nested userns?
> +	return system.GetParentNSeuid() != 0, nil
>
> In the kernel the inode for the initial user namespace is hardcoded, and it seems like a reliable (and undocumented AFAICS) way to
> detect if we are running in the init userNS:
>
> $ cat > userns_checker.c << EOF
> #include <sys/types.h>
> #include <sys/stat.h>
> #include <unistd.h>
> #include <stdio.h>
> #include <string.h>
>
> int main ()
> {
>   struct stat sb;
>   if (stat ("/proc/self/ns/user", &sb) < 0)
>     return -1;
>   printf ("is main userns? %s\n", 0xEFFFFFFDU == sb.st_ino ? "yes" : "no");
>   return 0;
> }
> EOF
>
> $ gcc -o userns_checker userns_checker.c
>
> $ ./userns_checker 
> is main userns? yes
> $ unshare -r ./userns_checker 
> is main userns? no
>
> What do you think?

It is not safe to hard code inode numbers.  They are an implementation
detail. 

Eric

utils_linux.go
AkihiroSuda: Rebased PR and added the 4294967295 check. 
The 0xEFFFFFFD check can be follow-up PR after discussion.

cyphar: As far as I can tell this is not guaranteed -- it just happens to be that `ns_alloc_inum` returns the same thing on each boot. But any kernel change or module that results in `proc_alloc_inum` being called before `&init_user_ns` is instantiated is going to result in that inode number changing.

Not to mention that (as Eric said) special inode numbers in general are not something you should depend on (a perfect example is "detecting" `chroot(2)` by the inode number of `/` -- something that always breaks every time someone tries it with a new filesystem because they all use different inode numbers for `/`).

cyphar:
@ebiederm How do you feel about making `NS_GET_PARENT` return `-ENOENT` if the user is in the namespace and the namespace is the init_ns?

cyphar:
LGTM on the updated patchset.

AkihiroSuda:
Added commit c93815738a53acae4998c5c9bd84dae18f864b42 for allowing `setgroups`.
(Already reviewed by @cyphar in https://github.com/genuinetools/img/issues/96#issuecomment-391459986 )

AkihiroSuda:
@dqminh PTAL?

cyphar:
LGTM again.

crosbymichael:
LGTM

libcontainer/system/linux.go
```diff
@@ -102,34 +101,43 @@ func Setctty() error {
...
-	var a, b, c int64
-	fmt.Sscanf(line, "%d %d %d", &a, &b, &c)
+func UIDMapInUserNS(uidmap []user.IDMap) bool {
 	/*
 	 * We assume we are in the initial user namespace if we have a full
 	 * range - 4294967295 uids starting at uid 0.
 	 */
-	if a == 0 && b == 0 && c == 4294967295 {
+	if len(uidmap) == 1 && uidmap[0].ID == 0 && uidmap[0].ParentID == 0 && uidmap[0].Count == 4294967295 {
 		return false
```
thaJeztah: Looks like this broke ARM builds in Moby;

```
linux_amd64_netgo:/usr/local/go/pkg/linux_amd64_netgo"    -e GOARM=6 "docker-dev:master" hack/make.sh binary
04:21:31 
04:21:32 Removing bundles/
04:21:32 
04:21:32 ---> Making bundle: binary (in bundles/binary)
04:21:32 Building: bundles/binary-daemon/dockerd-18.06.0-ce-dev
04:22:29 # github.com/docker/docker/vendor/github.com/opencontainers/runc/libcontainer/system
04:22:29 vendor/github.com/opencontainers/runc/libcontainer/system/linux.go:119:89: constant 4294967295 overflows int
```

PR #1690, opened 2018-01-14
Add detailed descriptions and examples of the bundle

Taeung:
For more clear descriptions, modify man-pages(`create`, `run`) and their descriptions.

### DIFF
--- a/create.go
+++ b/create.go
@@ -16,7 +16,7 @@ are starting. The name you provide for the container instance must be unique on
 your host.`,
 	Description: `The create command creates an instance of a container for a bundle. The bundle
 is a directory with a specification file named "` + specConfig + `" and a root
-filesystem.
+filesystem(default path: "rootfs").
 
 The specification file includes an args parameter. The args parameter is used
 to specify command(s) that get run when the container is started. To change the
--- a/man/runc-create.8.md
+++ b/man/runc-create.8.md
@@ -11,7 +11,10 @@ your host.
 # DESCRIPTION
    The create command creates an instance of a container for a bundle. The bundle
 is a directory with a specification file named "config.json" and a root
-filesystem.
+filesystem(default path: "rootfs"). For example, The bundle directory can be like below:
+
+       # ls mycontainer/
+       config.json  rootfs/
 
 The specification file includes an args parameter. The args parameter is used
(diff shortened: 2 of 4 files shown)


Taeung:
In addition, what do you think about the below examples ?
```patch
diff --git a/man/runc-create.8.md b/man/runc-create.8.md
index 1bf2e693..49fc9c98 100644
--- a/man/runc-create.8.md
+++ b/man/runc-create.8.md
@@ -27,3 +27,18 @@ command(s) that get executed on start, edit the args parameter of the spec. See
    --pid-file value          specify the file to write the process id to
    --no-pivot                do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk
    --no-new-keyring          do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key
+
+# EXAMPLE
+To create a container with the 'busybox' root filesystem:
+        # mkdir -p /mycontainer/rootfs
+        # cd /mycontainer
+        # docker export $(docker create busybox) | tar -C rootfs -xvf -
+        # runc spec
+
+        # ls /mycontainer
+        config.json  rootfs/
+
+        # runc create test
+        # runc list
+        ID     PID     STATUS    BUNDLE         CREATED                          OWNER
+        test   17257   created   /mycontainer   2018-01-14T14:38:36.138596574Z   root
```

cyphar:
@Taeung 

Regarding this:

```
# docker export $(docker create busybox) | tar -C rootfs -xvf -
```

I personally don't like referencing Docker in OCI documentation (I realise that this is in the README, but I also think the README should be changed too). One option is to reference tools like `skopeo` and `umoci` which are fairly widely used these days for image operations. The relevant command would be:

```
% skopeo copy docker://busybox oci:busybox
% umoci unpack --image busybox /mycontainer
% # and then you use /mycontainer/rootfs
```

The https://runc.io/ website used to mention the above, but the website now redirects to GitHub. Another option is to just download a BusyBox (or other distribution) rootfs and unpack it using `curl` and `tar`.

Taeung:
OK, I understood it !
Based on your comment, I'll change this PR.

AkihiroSuda:
@Taeung Are you still interested in this?

kolyshkin:
No reply from the author; closing.

PR #1692, opened 2018-01-15
rootless: optional support for generating config with subuid map

AkihiroSuda:

`runc spec --rootless --rootless-subuid` generates a config with multiple uidMappings and gidMappings. (See https://github.com/opencontainers/runc/pull/1529)

### DIFF
--- a/libcontainer/specconv/example.go
+++ b/libcontainer/specconv/example.go
@@ -2,8 +2,11 @@ package specconv
 
 import (
 	"os"
+	"sort"
 	"strings"
 
+	"github.com/opencontainers/runc/libcontainer/system"
+	"github.com/opencontainers/runc/libcontainer/user"
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
updated to support execution within userns

cyphar:
I'm not a *huge* fan of this, though I might still be convinced. `runc spec --rootless` was already a bit too much in terms of trying to provide more configuration with regards to config generation (something users should do with `oci-runtime-tools`). But since this sources information from `/etc/subuid` it might be a different story.

However, there is another problem, which is that in many cases users shouldn't be mapping all of their allocated subuids/subgids for each container. They should be using *independent* sets of uids and gids (this is something that Docker gets very, very wrong -- and we shouldn't be repeating that mistake). And example of this done more correctly is rkt. With that in mind, I'm not sure that you could automatically decide what the best sub-range is of a user's `/etc/sub{uid,gid}` sets...

AkihiroSuda:
How about adding UID/GID range fields to RootlessOpts?
If empty entire spaces will be used.

cc @jessfraz

jessfraz:
I'm not sure the best approach here to be honest, but not making the same mistakes as docker sounds good haha

cyphar:
Just to note that LXC also does the right thing here -- they allocate sub-sections of the available `/etc/sub{uid,gid}` ranges. But I'm not sure whether they deal with collisions by storing stuff on-disk -- which would be nice to avoid. I'll ping @brauner elsewhere and see how they do this or if they have any tips.

libcontainer/specconv/example.go
```diff
@@ -177,16 +230,66 @@ func ToRootless(spec *specs.Spec) {
...
+					})
+				uNextContainerID += subuid.Count
+			}
+			gNextContainerID := 1
+			sort.Sort(subIDSorter(ctx.SubGIDs))
+			for _, subgid := range ctx.SubGIDs {
+				spec.Linux.GIDMappings = append(spec.Linux.GIDMappings,
+					specs.LinuxIDMapping{
+						HostID:      uint32(subgid.SubID),
+						ContainerID: uint32(uNextContainerID),
```
AkihiroSuda: typo: will fix immediately

AkihiroSuda: done (cc @jessfraz )

AkihiroSuda:
Removed CLI and added godoc ,as this seems controversial, although already used in img and rootless BuildKit.

AkihiroSuda:
rebased

AkihiroSuda:
Any thought?

libcontainer/specconv/example.go
```diff
@@ -155,10 +158,66 @@ func Example() *specs.Spec {
...
+	// Add all subuids/subgids to spec.Linux.{U,G}IDMappings.
+	// Note that in many cases users shouldn't be mapping all of their
+	// allocated subuids/subgids for each container.
+	// They should be using independent sets of uids and gids if possible.
+	//
+	// MapAllSubIDs requires newuidmap(1) and newgidmap(1) with suid bit.
+	//
+	// When running in userns, MapAllSubIDs is ignored and
+	// /proc/self/[ug]id_map entries are used.
+	MapAllSubIDs bool
```
AkihiroSuda: I'm going to add more options such as `KeepNetworkNamespace` as well after this PR gets merged.

libcontainer/specconv/example.go
```diff
@@ -155,10 +158,66 @@ func Example() *specs.Spec {
...
+// RootlessOpts is an optional spec for ToRootless
+type RootlessOpts struct {
+	// Add all subuids/subgids to spec.Linux.{U,G}IDMappings.
+	// Note that in many cases users shouldn't be mapping all of their
+	// allocated subuids/subgids for each container.
+	// They should be using independent sets of uids and gids if possible.
+	//
+	// MapAllSubIDs requires newuidmap(1) and newgidmap(1) with suid bit.
+	//
+	// When running in userns, MapAllSubIDs is ignored and
```
AkihiroSuda: As this is very confusing, I'm temporary closing this PR. 

Please also see https://github.com/opencontainers/runc/issues/1837

PR #1693, opened 2018-01-15
 rootless: allow setgroups if `--force-mapping-tool` is set

AkihiroSuda:
Previously, `/proc/PID/setgroups` was always set to `"deny"` in rootless mode, as recent Linux kernel does not allow unprivileged users to update `/proc/PID/uid_map` without doing so.

However, we can mitigate this limitation by using `newuidmap(1)` and `newgidmap(1)`, although it requires these binaries (not `runc` itself) to have suid bit.

This PR adds a new flag `--force-mapping-tool` to use these binaries compulsory.

Now `apt` works within a rootless container, without ptrace hack:

Example config.json: https://gist.github.com/AkihiroSuda/0a46ad9ba6f392b27ac4c8d372721207

Demo (Tested on Ubuntu 17.10, kernel 4.13.0-25-generic x86_64):
```
$ mkdir rootfs
$ wget -O /tmp/ubuntu-base-17.10-base-amd64.tar.gz http://cdimage.ubuntu.com/ubuntu-base/releases/17.10/release/ubuntu-base-17.10-base-amd64.tar.gz
$ tar xzf /tmp/ubuntu-base-17.10-base-amd64.tar.gz -C ./rootfs || echo "mknod error were ignored"
$ cp /etc/resolv.conf rootfs/etc
$ vi config.json
(you need to set your own uidMappings and gidMappings)
$ runc run --force-mapping-tool foo
root@runc:/# chmod 1777 /tmp
root@runc:/# apt update
root@runc:/# apt install cowsay
root@runc:/# /usr/games/cowsay hello rootless apt
 ____________________
< hello rootless apt >
 --------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

If depending on the mapping tool is problem, we can use `ptrace(2)` hacks instead: https://github.com/AkihiroSuda/runrootless

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -47,6 +47,7 @@ type linuxContainer struct {
 	criuPath             string
 	newuidmapPath        string
 	newgidmapPath        string
+	forceMappingTool     bool
 	m                    sync.Mutex
 	criuVersion          int
 	state                containerState
@@ -1749,8 +1750,9 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
 					Value: []byte(c.newgidmapPath),
 				})
 			}
-			// The following only applies if we are root.
-			if !c.config.Rootless {
+			// The following only applies if we are root, unless we are
+			// advised to use the mapping tool.
+			if !(c.config.Rootless && !c.forceMappingTool) {
 				// check if we have CAP_SETGID to setgroup properly
 				pid, err := capability.NewPid(os.Getpid())
 				if err != nil {
@@ -1763,6 +1765,12 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
 					})
 				}
 			}
+		}
+		if c.forceMappingTool {
+			r.AddData(&Boolmsg{
+				Type:  ForceMappingToolAttr,
+				Value: true,
+			})
 		}
 	}
 
--- a/libcontainer/message_linux.go
+++ b/libcontainer/message_linux.go
@@ -77,13 +77,13 @@ func (msg *Boolmsg) Serialize() []byte {
 	native.PutUint16(buf[0:2], uint16(msg.Len()))
 	native.PutUint16(buf[2:4], msg.Type)
 	if msg.Value {
-		buf[4] = 1
+		native.PutUint32(buf[4:8], uint32(1))
 	} else {
-		buf[4] = 0
+		native.PutUint32(buf[4:8], uint32(0))
 	}
 	return buf
 }
 
 func (msg *Boolmsg) Len() int {
-	return unix.NLA_HDRLEN + 1
+	return unix.NLA_HDRLEN + 4 // alignment
 }
--- a/libcontainer/message_linux.go
+++ b/libcontainer/message_linux.go
@@ -10,16 +10,17 @@ import (
 // list of known message types we want to send to bootstrap program
 // The number is randomly chosen to not conflict with known netlink types
 const (
-	InitMsg         uint16 = 62000
-	CloneFlagsAttr  uint16 = 27281
-	NsPathsAttr     uint16 = 27282
-	UidmapAttr      uint16 = 27283
-	GidmapAttr      uint16 = 27284
-	SetgroupAttr    uint16 = 27285
-	OomScoreAdjAttr uint16 = 27286
-	RootlessAttr    uint16 = 27287
-	UidmapPathAttr  uint16 = 27288
-	GidmapPathAttr  uint16 = 27289
+	InitMsg              uint16 = 62000
+	CloneFlagsAttr       uint16 = 27281
+	NsPathsAttr          uint16 = 27282
+	UidmapAttr           uint16 = 27283
+	GidmapAttr           uint16 = 27284
+	SetgroupAttr         uint16 = 27285
+	OomScoreAdjAttr      uint16 = 27286
+	RootlessAttr         uint16 = 27287
+	UidmapPathAttr       uint16 = 27288
+	GidmapPathAttr       uint16 = 27289
+	ForceMappingToolAttr uint16 = 27290
 )
 
 type Int32msg struct {
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -88,6 +88,7 @@ struct nlconfig_t {
 	size_t uidmappath_len;
 	char *gidmappath;
 	size_t gidmappath_len;
+	uint8_t force_mapping_tool;
 };
 
 /*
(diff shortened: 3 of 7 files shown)


libcontainer/message_linux.go
```diff
@@ -77,13 +78,13 @@ func (msg *Boolmsg) Serialize() []byte {
...
 	} else {
-		buf[4] = 0
+		native.PutUint32(buf[4:8], uint32(0))
 	}
 	return buf
 }
 
 func (msg *Boolmsg) Len() int {
-	return unix.NLA_HDRLEN + 1
+	return unix.NLA_HDRLEN + 4 // alignment
```
AkihiroSuda: This alignment issue had not appear because no multiple `Boolmsg`s had been serialized at once.

frezbo:
@AkihiroSuda  looks like the *newuidmap* and *newgidmap* already has suid bit set, I tested on my local fedora 27 machine, so the suid bit should be ok. I wonder what all privileges these require so we can do a setcap and remove the suid bit.

craigfurman:
Interestingly, users do not need an entry in `/etc/subgid` at all in order to create a user namespace and setuid to 0 without writing `deny` into `/proc/$pid/setgroups`. This actually renders negative group ACLs (e.g. denying a directory to a group by using mode 0707) useless.

I raised this as [an issue](https://bugs.launchpad.net/ubuntu/+source/shadow/+bug/1729357), but there's disagreement on how best to tackle it.

It's a bit of a shame that we have a kernel-level check that's subverted by popular setuid executables. I do see the point that negative group ACLs aren't very useful, but we're kind of in the worst of both worlds now, where we have `/proc/` interface complexity (setgroups file) that's subverted by newgidmap!

cyphar:
@craigfurman I think the logic is that you need to have `setgroups` allowed in order for a "normal" Linux distribution to work as-expected in a user namespace. The `707` issue is most definitely quite concerning, but I expect that the argument from the shadow folks will be

> If you specify that users can set their GID in `/etc/subgid`, you are implicitly allowing `setgroups` and this is an *administrative* decision that is similar to adding a user to `sudoers` or changing `/etc/passwd` such that a user has `uid=0`.

Personally I can see why it is useful to allow `setgroups` (I know that the LXC folks use this quite heavily so you can boot full distributions inside containers -- something that would become significantly more difficult without this), though I do somewhat agree that it's more dangerous than it is being sold as.

craigfurman:
@cyphar you don't need an entry in `/etc/subgid` to be able to do this though

craigfurman:
And yeah tbh I totally value allowing setgroups more than I value negative group ACLs... it's certainly more useful for us container folks! :)

cyphar:
I added a comment to the bug you mentioned with an idea for how we could extend `/etc/subgid` to fix this issue. I can work on a patch for this soon, and I think that it's probably a good idea to get a CVE assigned.

craigfurman:
Nice, thanks for replying to that thread. I'll continue the conversation on there.

AkihiroSuda:
@cyphar @craigfurman Does this PR LGTY?

libcontainer/container_linux.go
```diff
@@ -1749,8 +1750,9 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
 					Value: []byte(c.newgidmapPath),
 				})
 			}
-			// The following only applies if we are root.
-			if !c.config.Rootless {
+			// The following only applies if we are root, unless we are
+			// advised to use the mapping tool.
+			if !(c.config.Rootless && !c.forceMappingTool) {
```
craigfurman: What do you think of `if !c.config.Rootless || c.forceMappingTool`? Fewer `!`s is usually less confusing. I haven't run this, you should test it rather than take my word for it of course

AkihiroSuda:
> However it is no longer as obvious whether this actually will help with setgroups(2) in the single-user-being-mapped case (the admin would need to explicitly set cyphar:1000:1 in /etc/subuid).

For single-user-being-mapped case we would use ptrace hack instead.
So I think this PR is still fine.

cyphar:
But if we have more than one user being mapped then we would already fall-back to `newuidmap` and `newgidmap` (because the write attempts will fail) -- this is what I was confused about.

AkihiroSuda:
s/force-mapping-tools/keep-proc-setgroups-untouched/ then?

cyphar:
We could do this without a CLI flag actually. You just need to figure out whether the mapping contains more than one user (because the single-user case is now the same with or without the setuid tools). If it does, then we assume we don't want to write anything to `/proc/self/setgroups`. Otherwise we write as usual.

How does that sound?

AkihiroSuda:
@cyphar updated, PTAL

libcontainer/container_linux.go
```diff
@@ -1801,8 +1801,11 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
 					Value: []byte(c.newgidmapPath),
 				})
 			}
-			// The following only applies if we are root.
-			if !c.config.Rootless {
+			gidMapCount := 0
+			for _, m := range c.config.GidMappings {
+				gidMapCount += m.Size
+			}
```
cyphar: This isn't quite right. It will miss if we have a single mapping of more than one group. Really, what we should do is have a helper function that does something like:

```
func requiresMappingTool(c *configs.Config) bool {
  gidMap := []configs.IDMap{
    {ContainerID: 0, HostID: os.Getegid(), Size: 1},
  }
  return !reflect.DeepEquals(c.GidMappings, gidMap)
}
```

AkihiroSuda: > It will miss if we have a single mapping of more than one group. 

Why? Does not `gidMapCount += m.Size` catch such case?

libcontainer/nsenter/nsexec.c
```diff
@@ -678,17 +678,16 @@ void nsexec(void)
...
-					if (config.is_rootless && config.is_setgroup) {
-						kill(child, SIGKILL);
-						bail("cannot allow setgroup in an unprivileged user namespace setup");
-					}
 
 					if (config.is_setgroup)
 						update_setgroups(child, SETGROUPS_ALLOW);
-					if (config.is_rootless)
+					else if (config.is_rootless)
 						update_setgroups(child, SETGROUPS_DENY);
```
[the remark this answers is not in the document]
AkihiroSuda: done

libcontainer/container_linux.go
cyphar: It doesn't catch if the `HostID` is not `os.Getegid`.

AkihiroSuda: thanks for pointing out, updated

cyphar:
LGTM. I've tested it out a bit and it appears to work pretty well. Thanks @AkihiroSuda!

AkihiroSuda:
@opencontainers/runc-maintainers anyone PTAL?

haampie:
@AkihiroSuda, I think this example has gone broken on recent versions of runc, can you verify that?

AkihiroSuda:
--force-mapping-tool is no longer needed

haampie:
Yeah, I'm aware of that, but I cannot get `apt-get` update to work inside a rootless container with a setup very similar as posted in your gist above, see #2517. I don't see obvious issues with the configuration, would be interested to know if you can still get your example from above to work in a rootless container.

haampie:
Sorry for the noise, I've figured it out, see the issue.

PR #1702, opened 2018-01-25
chroot when no mount namespaces is provided

crosbymichael:
We shouldn't skip the rootfs code when NEWNS is not specified because it will place the process in the host's `/`, instead, perform a simple `chroot` so that the process is still placed inside the specified rootfs. 

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -100,8 +100,10 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 
 	if config.NoPivotRoot {
 		err = msMoveRoot(config.Rootfs)
-	} else {
+	} else if config.Namespaces.Contains(configs.NEWNS) {
 		err = pivotRoot(config.Rootfs)
+	} else {
+		err = chroot(config.Rootfs)
 	}
 	if err != nil {
 		return newSystemErrorWithCause(err, "jailing process inside rootfs")
@@ -702,6 +704,10 @@ func msMoveRoot(rootfs string) error {
 	if err := unix.Mount(rootfs, "/", "", unix.MS_MOVE, ""); err != nil {
 		return err
 	}
+	return chroot(rootfs)
+}
+
+func chroot(rootfs string) error {
 	if err := unix.Chroot("."); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


justincormack:
I think this behaviour should be mentioned in the (Linux) spec, but I do think it is totally reasonable, if you dont specify a mount namespace, still making the root of the filesystem the rootfs seems sane, as hardly anything would work otherwise, if it was still the system root. For a start you would not even be in the right directory, and paths would not be useful, and you would have no way of knowing what the mountpoint of the rootfs was. So in favour of this, even if use cases are quite niche.

wking:
> I think this behaviour should be mentioned in the (Linux) spec…

+1 for this.  The spec also does not cover runc's `--no-pivot` semantics, which results in [downstream compatibility issues][1].

[1]: https://github.com/kubernetes-incubator/cri-o/issues/1260#issuecomment-358739239

cyphar:
LGTM in principle, though I would like to have this clarified in the spec -- and also we should generally dissuade people from using this (as we all know, `chroot(2)` pins the old mounts).

It should also be noted that the current behaviour of `pivot_root(2)` (that it changes the root of all processes in the same mount namespace) is considered a bug and that the "right" behaviour would be more like `chroot` (but given how long `pivot_root` has been around it's a pipe dream to sit around waiting for them to fix the `pivot_root` issues) -- so in an ideal world we would still be doing a `pivot_root` here even without mount namespaces.

crosbymichael:
ya, good luck waiting for that pivot root change to happen ;) 

I'll think about a good way to word this in the spec but in the meantime I think this change is safer than what happens today of the rootfs code does not run and we get placed in the host root, ignoring `Rootfs.Path`

cyphar:
@crosbymichael Yup, I completely agree.

mrunalp:
LGTM

AkihiroSuda:
What is the usecase of this?

AkihiroSuda:
> It should also be noted that the current behaviour of pivot_root(2) (that it changes the root of all processes in the same mount namespace) is considered a bug and that the "right" behaviour would be more like chroot

@cyphar Why?

cyphar:
@AkihiroSuda My view on that has flipped since I wrote the comment -- it turns out that the man page is more than a decade old and was written before `pivot_root(2)` was merged. The man-page maintainer contacted the kernel "containers" mailing lists and said he's going to rewrite the `pivot_root(2)` page entirely and make reference to tricks that LXC (and runc) use such as `pivot_root(".", ".")`.

PR #1703, opened 2018-01-25
libcontainer/state_linux_test: Add a testTransitions helper

wking:
The helper DRYs up the transition tests and makes it easy to get complete coverage for invalid transitions.

I'm also using `t.Run()` for subtests.  `Run()` is [new in Go 1.7][1], but runc dropped support for 1.6 back in e773f96b (#1335).

[1]: https://blog.golang.org/subtests

PR #218, opened 2015-08-21
Handling Pausing from freezer state

rajasec:
@crosbymichael @mrunalp 
With current runc implementation, container pause is handled from the config file, Added the contents to handle from the freezer subsystem and its freezer.state from the respective container.
This is to handle #217 which has been raised

Signed-off-by: Rajasekaran <rajasec79@gmail.com>

PR #1704, opened 2018-01-25
libcontainer: Add support for the "pausing" status

wking:
Builds on #1703; review that first.

The freezer controller [distinguishes between freezing and frozen][1]. We've had a `Pausing` since dbb515f0 (2014-07-08), but had [never implemented it][2].  This pull-request adds the missing implementation.

I haven't updated `libcontainer/cgroups`, which has used a blocking freeze change since 3e8849fa (2014-06-04). The cgroups interface doesn't distinguish between blocking and non-blocking sets, so there's no way to say “I want to start freezing, but don't need to block until completion”.

I haven't added integration tests either.  All of our current tests (`state.bats` and `pause.bats`) assume a complete transition to `paused`.  That makes sense because `runc pause …` is based on the blocking `libcontainer/cgroups` implementation.  The only way to see `pausing` would be to have a parallel state call during the pause call.  Here's how the timing could work out:

1. Call `pause`.
2. Call `state`, which reads `THAWED` and outputs `running` (or `created`, or whatever).
3. `pause` starts freezing the controller.
4. Call `state`, which reads `FREEZING` and outputs `pausing`.
5. `pause` finishes freezing the controller.
6. Call `state`, which reads `FROZEN` and outputs `paused`.
7. `pause` exits.

I'm not sure how to get a test where the `state` call consistently fires at 4 and not at 2 or 6.

[1]: https://www.kernel.org/doc/Documentation/cgroup-v1/freezer-subsystem.txt
[2]: https://github.com/opencontainers/runc/pull/218/files#r38837239

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1588,11 +1588,13 @@ func (c *linuxContainer) currentStatus() (Status, error) {
 // out of process we need to verify the container's status based on runtime
 // information and not rely on our in process info.
 func (c *linuxContainer) refreshState() error {
-	paused, err := c.isPaused()
+	paused, err := c.pausedState()
 	if err != nil {
 		return err
 	}
-	if paused {
+	if paused == Pausing {
+		return c.state.transition(&pausingState{c: c})
+	} else if paused == Paused {
 		return c.state.transition(&pausedState{c: c})
 	}
 	t, err := c.runType()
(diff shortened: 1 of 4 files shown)


wking:
Rebased onto master with 2f8ae9cd → 87b4cc38 now that #1703 has landed.  No changes to the tip commit, but the rebase removes the #1703 changes from GitHub's PR diff.

kolyshkin:
We've removed "pausing" state in PR #3360

PR #1711, opened 2018-02-04
Update build dependencies in Dockerfile

dqminh:
This upgrade golang version in dockerfile to 1.10, image version from jessie to stretch, and criu version to v3.7. Because of the bump from jessie -> stretch, libseccomp is also bumped to 2.3.1-2.1


cyphar:
DCOs are missing.

dqminh:
@cyphar updated :man_facepalming: Also updated golang version to 1.10 instead of 1.9.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1724, opened 2018-02-19
libcontainer/capabilities_linux: Drop os.Getpid() call

wking:
gocapability has supported 0 as “the current PID” since syndtr/gocapability#2.  libcontainer was ported to that approach in 444cc298 (docker/libcontainer#358), but the change was clobbered by 22df5551 which landed via 5b73860e (docker/libcontainer#388).  This pull-request restores the changes from 444cc298.

### DIFF
--- a/libcontainer/capabilities_linux.go
+++ b/libcontainer/capabilities_linux.go
@@ -4,7 +4,6 @@ package libcontainer
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


cyphar:
LGTM.

crosbymichael:
LGTM

PR #1726, opened 2018-02-20
*: Avoid creation panics when 'process' and 'linux' are unset

wking:
`process` has been optional since opencontainers/runtime-spec#701, and `linux` has been optional since at least opencontainers/runtime-spec#414.  This pull-request makes a few pivots so runc can handle something like the spec's [minimal config][1].  I actually tested with:

```
{
  "ociVersion": "1.0.0",
  "root": {
    "path": "/"
  }
}
```

because the runtime crashes without access to a container-side `/proc` (I have syndtr/gocapability#14 open to address part of this, but [other][2] [parts][3] would take more work).  With the changes in this PR, `runc create test` succeeded with that configuration.  It also mucked up my local terminal creation (although I rebooted without digging into exactly why), so take care when testing this yourself.

There's some previous discussion on moving the `LookPath` call starting [here][4].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/schema/test/config/good/minimal.json
[2]: https://github.com/opencontainers/runc/blob/595bea022f077a9e17d7473b34fbaf1adaed9e43/libcontainer/utils/utils_unix.go#L14
[3]: https://github.com/opencontainers/runc/blob/595bea022f077a9e17d7473b34fbaf1adaed9e43/libcontainer/standard_init_linux.go#L163-L167
[4]: https://github.com/opencontainers/runc/pull/827#discussion_r63464201

### DIFF
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -3,6 +3,7 @@
 package libcontainer
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -152,12 +153,6 @@ func (l *linuxStandardInit) Init() error {
 	if unix.Getppid() != l.parentPid {
 		return unix.Kill(unix.Getpid(), unix.SIGKILL)
 	}
-	// Check for the arg before waiting to make sure it exists and it is
-	// returned as a create time error.
-	name, err := exec.LookPath(l.config.Args[0])
-	if err != nil {
-		return err
-	}
 	// Close the pipe to signal that we have completed our init.
 	l.pipe.Close()
 	// Wait for the FIFO to be opened on the other side before exec-ing the
@@ -185,6 +180,13 @@ func (l *linuxStandardInit) Init() error {
 		if err := seccomp.InitSeccomp(l.config.Config.Seccomp); err != nil {
 			return newSystemErrorWithCause(err, "init seccomp")
 		}
+	}
+	if len(l.config.Args) == 0 {
+		return errors.New("no process arguments configured")
+	}
+	name, err := exec.LookPath(l.config.Args[0])
+	if err != nil {
+		return err
 	}
 	if err := syscall.Exec(name, l.config.Args[0:], os.Environ()); err != nil {
 		return newSystemErrorWithCause(err, "exec user process")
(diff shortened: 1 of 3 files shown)


libcontainer/standard_init_linux.go
```diff
@@ -186,6 +181,13 @@ func (l *linuxStandardInit) Init() error {
 			return newSystemErrorWithCause(err, "init seccomp")
 		}
 	}
+	if len(l.config.Args) == 0 {
```
crosbymichael: We want to validate this on creation not after we have done all this work.  Change the above code to if process == nil error out

utils_linux.go
```diff
@@ -353,17 +357,21 @@ func (r *runner) terminate(p *libcontainer.Process) {
...
 	}
-	if (!detach || !config.Terminal) && r.consoleSocket != "" {
+	if (!detach || !terminal) && r.consoleSocket != "" {
 		return fmt.Errorf("cannot use console socket if runc will not detach or allocate tty")
 	}
 	return nil
 }
 
 func validateProcessSpec(spec *specs.Process) error {
+	if spec == nil {
```
crosbymichael: this should be an error case

libcontainer/standard_init_linux.go
crosbymichael: In other words, there is no reason to do 90% of container creation if we get to the end and we cannot run the container because process == nil

cyphar:
@wking Regarding the `/proc` issue -- this is something I've been thinking about for a while. To cut a very long story short, there doesn't appear to be a way (given the current set of kernel facilities at our disposal) to safely create a container without mounting at least `/proc` in the container's PID namespace -- so it breaking really shouldn't be a surprise IMO. However there are also other issues that crop up with regards to `/proc` when you think about malicious container processes -- and at the moment there is literally no kernel facility that can protect against some kinds of attacks (AFAICS, and I've discussed this with some LXC folks and they agree...).

libcontainer/standard_init_linux.go
wking: > In other words, there is no reason to do 90% of container creation if we get to the end and we cannot run the container because process == nil

The point of this PR is allowing `create` for containers with no `process` (catching up with opencontainers/runtime-spec#701).  You can't `start` those containers, which is why we have this check here, but we certainly *don't* want to error those out during `create`.

AkihiroSuda:
needs rebase

kolyshkin:
The second and the third commits are obsoleted by 
 - PR #1926 (commit q16d55f17)
 - PR #1551 (commit 527dc5ac)
 
 The first one and the last one doesn't make much sense to me. Yet, if needed, please open a new PR.

PR #1728, opened 2018-02-21
Provide install instruction that uses "go get"

vsoch:
Is there any substantial difference to using go get instead of git clone? I first did the git clone (and of course) cloned one directory level off, resulting in the wrong paths and a bunch of errors. This is of course user error, but given that Go is a dependency anyway, would it make sense to control this clone step to prevent the user (me<--) from messing it  up?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,15 @@ make
 sudo make install
 ```
 
+You can also use `go get` to install to your `GOPATH`, assuming that you have a `github.com` parent folder already created under `src`:
+
+```bash
+go get github.com/opencontainers/runc
+cd $GOPATH/src/github.com/opencontainers/runc
+make
+sudo make install
+```
+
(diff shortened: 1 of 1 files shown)


crosbymichael:
Content looks good.  Can you sign your commit for the DCO?

vsoch:
I must admit I don't know what that means :) What do I need to do to sign it?

vsoch:
ahh DCO == robot? --> https://github.com/probot/dco

crosbymichael:
https://github.com/opencontainers/runc/blob/master/CONTRIBUTING.md#sign-your-work

vsoch:
oh no I didn't use my full name! I failed. I will try again.

crosbymichael:
Sorry, it cannot be two commits, each one must have the signoff.  You need to amend your original commit

Issue #1721, opened 2018-02-16
runc create hangs when passed a pipe for stdout or stderr


deitch:
I found this by playing with various options in golang `cmd := exec.Command("runc")` and changing `cmd.Stdout`, etc. to try and capture output from a container run. It works if I make stdout `os.Stdout` or an actual `file`, but anything else causes it to hang.

For example:

* `cmd.Stdout = io.MultiWriter(os.Stdout)`
* `cmd.Stdout = io.MultiWriter(file)`
* `pipe := cmd.StdoutPipe()` - hangs when you do `io.Copy(os.Stdout, pipe)`

A sample with this is [here](https://gist.github.com/deitch/ad6cb7b0801abe920be3aed1839cb722)

But it turns out I can recreate the problem simply:

```
$ runc create echo | cat
```

The above hangs. 

Is this because of some strange pty/tty handling? How would I work around it if I need to capture stdout from a `runc create` and possibly send it to other areas? E.g. `runc create echo | tree /tmp/foo`

deitch:
cc @justincormack

cyphar:
What is your `config.json`? In particular, do you have `terminal` enabled or disabled?

deitch:
Unset, which, according to the docs, means `false`. Not that I ever really understood the import of the setting. :-)

deitch:
As far as I can tell, `terminal: false` means not to create a new pty for the container, but rather to inherit from whatever was passed to `runc`. When I run `runc -d` (or `runc create` and then `runc start`), it runs in detached mode - `runc` starts it and exits - and that causes something to hang. I am unsure what. 

If I try to do it as `runc run` (no `-d` or `create`+`start`), then it appears to work. 

1. Did I understand correctly?
2. If so, what is the correct way to get the container to use an already existing `stdin`/`stdout`/`stderr` but still have the lifecycle control of `runc create` followed by `runc start`?

cyphar:
(We really need to document this somewhere other than issues, or I should come up with a canned reply to questions about `terminal`.)

There are three main cases that can occur with pty handling:

* `runc create` or `runc run -d` indicates that `runc` will not stick around -- there will be no `runc` code running after the container is set up. This means that `runc` cannot do any IO forwarding -- it has to be set up by the caller. The way this restriction manifests depends on the terminal setting.
[...]

PR #1730, opened 2018-02-22
Add docs for terminals

deitch:
As discussed with @cyphar , a first attempt to document how to use terminals with runc. See discussion [here](https://github.com/opencontainers/runc/issues/1721) particularly [this comment](https://github.com/opencontainers/runc/issues/1721#issuecomment-366493547)

I am sure there are some errors here, so happy to take updates.

### DIFF
--- a/docs/terminals.md
+++ b/docs/terminals.md
@@ -1,0 +1,314 @@
+# Terminals and Standard IO #
+
+*Note that the default configuration of `runc` (foreground, new terminal) is
+generally the best option for most users. This document exists to help explain
+what the purpose of the different modes is, and to try to steer users away from
+common mistakes and misunderstandings.*
+
+In general, most processes on Unix (and Unix-like) operating systems have 3
+standard file descriptors provided at the start, collectively referred to as
+"standard IO" (`stdio`):
+
+* `0`: standard-in (`stdin`), the input stream into the process
+* `1`: standard-out (`stdout`), the output stream from the process
+* `2`: standard-error (`stderr`), the error stream from the process
+
+When creating and running a container via `runc`, it is important to take care
+to structure the `stdio` the new container's process receives. In some ways
+containers are just regular processes, while in other ways they're an isolated
+sub-partition of your machine (in a similar sense to a VM). This means that the
+structure of IO is not as simple as with ordinary programs (which generally
+just use the file descriptors you give them).
+
+## Other File Descriptors ##
+
+Before we continue, it is important to note that processes can have more file
+descriptors than just `stdio`. By default in `runc` no other file descriptors
+will be passed to the spawned container process. If you wish to explicitly pass
+file descriptors to the container you have to use the `--preserve-fds` option.
+These ancillary file descriptors don't have any of the strange semantics
+discussed further in this document (those only apply to `stdio`) -- they are
+passed untouched by `runc`.
+
+It should be noted that `--preserve-fds` does not take individual file
+descriptors to preserve. Instead, it takes how many file descriptors (not
+including `stdio` or `LISTEN_FDS`) should be passed to the container. In the
+following example:
+
+```
+% runc run --preserve-fds 5 <container>
+```
+
+`runc` will pass the first `5` file descriptors (`3`, `4`, `5`, `6`, and `7` --
+assuming that `LISTEN_FDS` has not been configured) to the container.
+
+In addition to `--preserve-fds`, `LISTEN_FDS` file descriptors are passed
+automatically to allow for `systemd`-style socket activation. To extend the
+above example:
+
+```
+% LISTEN_PID=$pid_of_runc LISTEN_FDS=3 runc run --preserve-fds 5 <container>
+```
+
+`runc` will now pass the first `8` file descriptors (and it will also pass
+`LISTEN_FDS=3` and `LISTEN_PID=1` to the container). The first `3` (`3`, `4`,
+and `5`) were passed due to `LISTEN_FDS` and the other `5` (`6`, `7`, `8`, `9`,
+and `10`) were passed due to `--preserve-fds`. You should keep this in mind if
+you use `runc` directly in something like a `systemd` unit file. To disable
+this `LISTEN_FDS`-style passing just unset `LISTEN_FDS`.
+
+**Be very careful when passing file descriptors to a container process.** Due
+to some Linux kernel (mis)features, a container with access to certain types of
+file descriptors (such as `O_PATH` descriptors) outside of the container's root
+file system can use these to break out of the container's pivoted mount
+namespace. [This has resulted in CVEs in the past.][CVE-2016-9962]
+
+[CVE-2016-9962]: https://nvd.nist.gov/vuln/detail/CVE-2016-9962
+
+## <a name="terminal-modes" /> Terminal Modes ##
+
+`runc` supports two distinct methods for passing `stdio` to the container's
+primary process:
+
+* [new terminal](#new-terminal) (`terminal: true`)
+* [pass-through](#pass-through) (`terminal: false`)
+
+When first using `runc` these two modes will look incredibly similar, but this
+can be quite deceptive as these different modes have quite different
+characteristics.
+
+By default, `runc spec` will create a configuration that will create a new
+terminal (`terminal: true`). However, if the `terminal: ...` line is not
+present in `config.json` then pass-through is the default.
+
+*In general we recommend using new terminal, because it means that tools like
+`sudo` will work inside your container. But pass-through can be useful if you
+know what you're doing, or if you're using `runc` as part of a non-interactive
+pipeline.*
+
+### <a name="new-terminal"> New Terminal ###
+
+In new terminal mode, `runc` will create a brand-new "console" (or more
+precisely, a new pseudo-terminal using the container's namespaced
+`/dev/pts/ptmx`) for your contained process to use as its `stdio`.
+
+When you start a process in new terminal mode, `runc` will do the following:
+
+1. Create a new pseudo-terminal.
+2. Pass the slave end to the container's primary process as its `stdio`.
+3. Send the master end to a process to interact with the `stdio` for the
+   container's primary process ([details below](#runc-modes)).
+
+It should be noted that since a new pseudo-terminal is being used for
+communication with the container, some strange properties of pseudo-terminals
+might surprise you. For instance, by default, all new pseudo-terminals
+translate the byte `'\n'` to the sequence `'\r\n'` on both `stdout` and
+`stderr`. In addition there are [a whole range of `ioctls(2)` that can only
+interact with pseudo-terminal `stdio`][tty_ioctl(4)].
+
+> **NOTE**: In new terminal mode, all three `stdio` file descriptors are the
+> same underlying file. The reason for this is to match how a shell's `stdio`
+> looks to a process (as well as remove race condition issues with having to
+> deal with multiple master pseudo-terminal file descriptors). However this
+> means that it is not really possible to uniquely distinguish between `stdout`
+> and `stderr` from the caller's perspective.
+
+[tty_ioctl(4)]: https://linux.die.net/man/4/tty_ioctl
+
+### <a name="pass-through"> Pass-Through ###
+
+If you have already set up some file handles that you wish your contained
+process to use as its `stdio`, then you can ask `runc` to pass them through to
+the contained process (this is not necessarily the same as `--preserve-fds`'s
+passing of file descriptors -- [details below](#runc-modes)). As an example
+(assuming that `terminal: false` is set in `config.json`):
+
+```
+% echo input | runc run some_container > /tmp/log.out 2>& /tmp/log.err
+```
+
+Here the container's various `stdio` file descriptors will be substituted with
+the following:
+
+* `stdin` will be sourced from the `echo input` pipeline.
+* `stdout` will be output into `/tmp/log.out` on the host.
+* `stderr` will be output into `/tmp/log.err` on the host.
+
+It should be noted that the actual file handles seen inside the container may
+be different [based on the mode `runc` is being used in](#runc-modes) (for
+instance, the file referenced by `1` could be `/tmp/log.out` directly or a pipe
+which `runc` is using to buffer output, based on the mode). However the net
+result will be the same in either case. In principle you could use the [new
+terminal mode](#new-terminal) in a pipeline, but the difference will become
+more clear when you are introduced to [`runc`'s detached mode](#runc-modes).
+
+## <a name="runc-modes" /> `runc` Modes ##
+
+`runc` itself runs in two modes:
+
+* [foreground](#foreground)
+* [detached](#detached)
+
+You can use either [terminal mode](#terminal-modes) with either `runc` mode.
+However, there are considerations that may indicate preference for one mode
+over another. It should be noted that while two types of modes (terminal and
+`runc`) are conceptually independent from each other, you should be aware of
+the intricacies of which combination you are using.
+
+*In general we recommend using foreground because it's the most
+straight-forward to use, with the only downside being that you will have a
+long-running `runc` process. Detached mode is difficult to get right and
+generally requires having your own `stdio` management.*
+
+### Foreground ###
+
+The default (and most straight-forward) mode of `runc`. In this mode, your
+`runc` command remains in the foreground with the container process as a child.
+All `stdio` is buffered through the foreground `runc` process (irrespective of
+which terminal mode you are using). This is conceptually quite similar to
+running a normal process interactively in a shell (and if you are using `runc`
+in a shell interactively, this is what you should use).
+
+Because the `stdio` will be buffered in this mode, some very important
+peculiarities of this mode should be kept in mind:
+
+* With [new terminal mode](#new-terminal), the container will see a
+  pseudo-terminal as its `stdio` (as you might expect). However, the `stdio` of
+  the foreground `runc` process will remain the `stdio` that the process was
+  started with -- and `runc` will copy all `stdio` between its `stdio` and the
+  container's `stdio`. This means that while a new pseudo-terminal has been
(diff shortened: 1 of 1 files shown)


cyphar:
I will take a thorough look over this when I get back from vacation.

deitch:
Hey @cyphar . How was vacation/holiday? Did I capture it well? Happy to make the next person to walk down my path have an easier walk. :-)

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Terminal Modes
+
+`runc` supports two distinct methods for passing stdio to the container's primary process:
+
+* pass-through
+* new terminal
+
+### Pass-Through
+In pass-through mode, the `stdio` passed to the `runc` call will be passed as is to the primary container. This means that if you do the following:
```
cyphar: This isn't totally accurate. In the case of non-detached, the `stdio` passed to the container are actually pipes (which are then read from by the still-running `runc` process). The reason this might be important to someone is so they know that `fd=2` in your example isn't *actually* the file `/tmp/log.out`.

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
...
+
+```
+$ echo input | runc run some_container > /tmp/log.out 2>& /tmp/log.err
+```
+
+Then the stdout of `some_container` will appear in the file `/tmp/log.out`, the stderr will appear in `/tmp/log.err`, and the process will receive input of `input`. `runc`'s stdout was redirected to `/tmp/log.out`, and `runc`, in turn, passed that handle to the container itself.
+
+To invoke pass-through mode, configure your container's `config.json` with `terminal: false` (which is the default).
```
cyphar: It's not the default if you use `runc spec`.

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
...
+## Which Mode To Use
+`runc` itself runs in two modes:
+
+* foreground
+* detached
+
+You can use either terminal mode with either runc mode. However, there are considerations that may indicate preference for one mode over another.
+
+### Foreground
+When running in foreground, `runc` and the container you invoke are most like running a regular foreground process. This is the scenario most likely to be useful for running in pass-through terminal mode. As the example above showed:
```
cyphar: > This is the scenario most likely to be useful for running in pass-through terminal mode.

Unfortunately it's not as simple as this. It really depends on how you're using `pass-through`. If you want to set up your own `stdio` then actually detached makes more sense in a lot of cases (unless you don't want to bother with writing the exit-code handling).

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
...
+You run in detached mode by doing one of the following:
+
+* `runc -d`
+* `runc create` followed by `runc start`, i.e. container lifecycle management
+
+#### Pass-Through Detached
+When running runc detached with terminal Pass-Through, the stdin/stdout/stderr of the calling process will be passed to `runc` and from there to the container's process. However, since `runc` detaches, there are several side effects to take into account:
+
+* The stdout/stderr of the process can become intermixed with the stdout/stderr of the calling process.
+* Any input into the calling process can cause indeterminate issues with which process actually receives the input.
```
cyphar: These notes appear to only be meaningful for shell users (and quite a few other points also seem quite shell-focused). Really this should just be a caution to not use pass-through detached in a shell -- rather than a warning about pass-through detached in general.

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+Two problems occurred above:
+
+1. The output from the `curl` command and the output of the container `foo` became intermingled.
+2. Our typing `curl https://containerd.io` actually was passed to both our shell and the container `foo`. The results can be unpredictable. Our example here was simulated and might not even work.
+
+Further, a third, possibly temporary problem exists. Pass-Through Detached only works correctly when stdout/stderr are either an actual terminal or a file. If you use _anything_ else, e.g. a pipe, it will hang, for reasons unknown. There is an [open issue](https://github.com/opencontainers/runc/issues/1721) on this.
+
+For this reason, we **strongly recommend** using New Terminal mode when running `runc` detached.
```
cyphar: ... unless you set up your own stdio outside the container.

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
...
+
+## How To Get the FD for New Terminal
+When running `runc` in New Terminal mode, i.e. `terminal:true` in `config.json` - whether foreground or detached `runc` - you **must** tell `runc` what process will control the master end of the pty. To do that:
+
+1. Open a Unix-domain socket
+2. Pass the path to the socket to `runc create <container> --console-socket <path_to_socket>`
+3. Listen on the socket to get the master file descriptor
+4. Use that file descriptor to control stdio
+
+Note the one shortcoming: the single descriptor means that you have mixed stdout/stderr. There is no way to separate them. Essentially, this is a _console_, which has no separate stdout and stderr.
```
cyphar: (Though this shortcoming is identical to regular terminal emulators -- take a look at `/proc/self/fd/[12]` in your shell right now.)

docs/terminals.md
```diff
@@ -0,0 +1,141 @@
+# Terminals and Standard IO
+All processes on Unix (and Unix-like) operating systems have 3 standard file descriptors (fd) passed to them at start, collectively standard-IO (`stdio`):
+
+* `0`: standard-in (`stdin`), the input stream into the process
+* `1`: standard-out (`stdout`), the output stream from the process
+* `2`: standard-error (`stderr`), the error stream from the process
```
cyphar: I would recommend at least mentioning `--preserve-fds` here, with the note that it is **separate** to all of this handling for `stdio` and that those file descriptors are passed verbatim in all cases.

docs/terminals.md
deitch: I am not sure I knew that, although it makes sense. Updating.

docs/terminals.md
deitch: Well, it still is the default if no `terminal:` setting is specific. `runc spec` just happens to actually specify it in the `config.json`. But I don't mind updating it.

docs/terminals.md
deitch: My inclination would be to leave this part. It doesn't say "only", but rather "most likely". I am coming at this as someone who hadn't messed around with `runc` before, banged my head on the wall to understand, and am trying to save others the headache. Simplifications, along with options to get deeper, are the best way to explain it to people IMHO.

docs/terminals.md
deitch: Is that true? I could have a process with its own stdout/stderr/stdin that is not a shell, and if it calls runc detached with pass-through, then the intermixing will happen, and anything fed into stdin will be indeterminate. Granted, stdin is used most often in shells, but it is not exclusive. I think the issue you described to me (slack? wherever it was...) actually is a general one.

docs/terminals.md
deitch: Good point. Fixing.

docs/terminals.md
deitch: Of course. It really is only something you can separate when you start a new process with separate descriptors. Happy to add a comment here.

docs/terminals.md
deitch: OK. I just added something at the end.

deitch:
@cyphar I updated with most of the changes requested; have another look. 

> I'm not sure if you saw the security worries about using pass-through (detached) in a shell, but please include those (if you need clarification on them feel free to ping me). This is actually the most significant issue with using detached pass-through in a shell.

Please do clarify?

deitch:
Hi @cyphar ; this is almost 2 months old. Can we review the changes and get it in?

cyphar:
Sorry for the long wait. I've added some comments. However, I have a feeling at this point we might as well merge it (with the `--preserve-fds` section fixed) and I can reword the relevant parts in my own time (since I'm sure it's annoying to re-send drafts).

cyphar:
Eugh, sorry. Wrong button.

deitch:
> However, I have a feeling at this point we might as well merge it (with the --preserve-fds section fixed) and I can reword the relevant parts in my own time (since I'm sure it's annoying to re-send drafts).

Sure. I will do an immediate update to the best of my knowledge and send it back to you.

docs/terminals.md
deitch: Please do. I think I need to see how you rewrite it once it is in.

docs/terminals.md
```diff
@@ -0,0 +1,177 @@
...
+  quitChan <- struct{}{}
+}()
+
+// Only close the master fd once we've stopped copying.
+<-quitChan
+c.Close()
+```
+
+## Other File Descriptors
+Completely separately from stdio - file descriptors `0`,`1`,`2` - you also can pass other file descriptors to the container process "as is" using `--preserve-fds <descriptor>`. Thus, for example, if your calling process has an open handle to the file `/var/foo` as file descriptor `5`, and you want to pass it to your container process, you can call `runc`:
```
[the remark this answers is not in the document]
deitch: Fixing now.

docs/terminals.md
```diff
@@ -0,0 +1,177 @@
...
+You can use either terminal mode with either runc mode. However, there are considerations that may indicate preference for one mode over another.
+
+### Foreground
+When running in foreground, `runc` and the container you invoke are most like running a regular foreground process. This is the scenario most likely to be useful for running in pass-through terminal mode. As the example above showed:
+
+```
+$ echo input | runc run some_container > /tmp/log.out 2>& /tmp/log.err
+```
+
+Nonetheless, you still have the option, when running in foreground runc mode, of using New Terminal mode. However, if the process is short-lived, you will not have much time to retrieve the console's fds from the socket and doing anything with it.
```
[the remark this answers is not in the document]
deitch: OK, please do.

deitch:
My one request, please add your patches soon after merging? I finally understand most of it, want to grasp the entire thing! :-)

deitch:
Updated the `--preserve-fds` comment. Back at you @cyphar .

cyphar:
LGTM. I will do a rewrite of the relevant parts after it's merged.

cyphar:
LGTM. I will do a rewrite of the relevant parts after it's merged.

cyphar:
PullApprove appears to be broken? /cc @caniszczyk

caniszczyk:
LGTM

caniszczyk:
@cyphar weird try to LGTM again

dqminh:
@cyphar does it make more sense to just commit directly into the branch instead of us having to merge with some incompleted / incorrect parts ?

docs/terminals.md
```diff
@@ -0,0 +1,179 @@
...
+
+Note the one shortcoming: the single descriptor means that you have mixed stdout/stderr. There is no way to separate them. Essentially, this is a _console_, which has no separate stdout and stderr. This is the same issue you have with any console, or even any normal terminal window. To see the issue:
+
+```
+/ # ls -l /proc/self/fd/[12]
+lrwx------    1 root     root            64 Mar 19 07:09 /proc/self/fd/1 -> /0
+lrwx------    1 root     root            64 Mar 19 07:09 /proc/self/fd/2 -> /0
+```
+
+The following sample code, partially taken from the reference implementation [recvtty](https://github.com/opencontainers/runc/blob/master/contrib/cmd/recvtty/recvtty.go), shows how to accept a file descriptor for the socket. It does not contain the usual `err` handling and `defer`red closing.
```
dqminh: i would prefer to just point directly to `recvtty.go` rather than attaching sample code here so there are one less place to keep change.

docs/terminals.md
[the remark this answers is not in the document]
dqminh: I would prefer to remove the inaccurate part if we like to merge this now.

cyphar:
@dqminh Ah, I didn't even check if @deitch has given us push access. I'll send some corrections to his branch then re-LGTM it.

cyphar:
Reject

cyphar:
@deitch I have done a fairly extensive rewrite of the document, and now I'm pretty happy with how it stands. Since it still has your name on it, I'd like your blessing before we merge it -- does the current document help explain how things work?

cyphar:
LGTM (obviously) -- but wait for @deitch's response before merging.

deitch:
> I'd like your blessing before we merge it

Aleksa is the expert here! Not sure asking my opinion makes sense... :-)

I will do a quick review.

docs/terminals.md
```diff
@@ -0,0 +1,291 @@
...
+% runc run --preserve-fds 5 <container>
+```
+
+`runc` will pass the first `5` file descriptors (`3`, `4`, `5`, `6`, and `7` --
+assuming that `LISTEN_FDS` has not been configured) to the container.
+
+In addition to `--preserve-fds`, `LISTEN_FDS` file descriptors are passed
+automatically to allow for `systemd`-style socket activation. They always take
+up the first file descriptors after `stdio` (and `--preserve-fds` always follow).
+To disable this, simply unset `LISTEN_FDS`.
```
deitch: 1. what happens if `LISTEN_FDS` are set and `--preserve-fds`? An example like the above might help?
2. By default, isn't `LISTEN_FDS` unset anyways?

docs/terminals.md
```diff
@@ -0,0 +1,291 @@
...
+
+### <a name="pass-through"> Pass-Through ###
+
+If you have already set up some file handles that you wish your contained
+process to use as its `stdio`, then you can ask `runc` to pass them through to
+the contained process (this is not necessarily the same as `--preserve-fds`'s
+passing of file descriptors -- [details below](#runc-modes)). As an example:
+
+```
+% echo input | runc run some_container > /tmp/log.out 2>& /tmp/log.err
```
deitch: Hmm... isn't `runc`'s default mode `terminal: true`?

docs/terminals.md
```diff
@@ -0,0 +1,291 @@
...
+handle the `stdio` of the container.
+
+#### Detached Pass-Through ####
+
+In detached mode, pass-through actually does what it says on the tin -- the
+`stdio` file descriptors of the `runc` process are passed through (untouched)
+to the container's `stdio`. The purpose of this option is to allow a user to
+set up `stdio` for a container themselves and then force `runc` to just use
+their pre-prepared `stdio` (without any pseudo-terminal funny business). *If
+you don't see why this would be useful, don't use this option.*
```
deitch: I love this line!

docs/terminals.md
```diff
@@ -0,0 +1,291 @@
...
+the only holder of the pseudo-terminal master file descriptor will be whatever
+process read the file descriptor from the socket).
+
+> **NOTE**: Currently `runc` doesn't support abstract socket addresses (due to
+> it not being possible to pass an `argv` with a null-byte as the first
+> character). In the future this may change, but currently you must use a valid
+> path name.
+
+In order to help users make use of detached new terminal mode, we have provided
+a [Go implementation in the `go-runc` bindings][containerd/go-runc.Socket].
```
deitch: This reference is very helpful.

docs/terminals.md
```diff
@@ -0,0 +1,291 @@
...
+(which can then use that file descriptor as though they opened it). When using
+`runc` in detached new terminal mode, this is how a user gets access to the
+pseudo-terminal's master file descriptor.
+
+To this end, there is a new option (which is required if you want to use `runc`
+in detached new terminal mode): `--console-socket`. This option takes the path
+to a Unix domain socket which `runc` will connect to and send the
+pseudo-terminal master file descriptor down (after which `runc` will exit and
+the only holder of the pseudo-terminal master file descriptor will be whatever
+process read the file descriptor from the socket).
```
deitch: Maybe add a paragraph that says something like the following?

So to use `runc` in "Detached New Terminal" mode:

1. Set up a unix domain socket
2. Run `runc` in detached new terminal mode, passing it the path to the socket as the argument to `--console-socket`
3. Retrieve the file descriptor for the new pseudo-terminal from the socket
4. Use the file descriptor to communicate with stdio of the container, now running detached

deitch:
Thanks @cyphar , really better. I made one small suggestion, and 1-2 places I was a bit confused.

In the end, this is complex, so it will be confusing. But this is a really good starting point.

docs/terminals.md
cyphar: > 1. What happens if LISTEN_FDS are set and --preserve-fds? An example like the above might help?

The first `LISTEN_FDS` file descriptors come first, then the `--preserve-fds` ones (the `--help` for `runc run` explains it a bit). I could try to add an example, but `LISTEN_FDS` is a bit difficult to given an example of without having a `systemd` service and all the rest of that fun stuff.

> 2. By default, isn't LISTEN_FDS unset anyways?

Yup, but it is important to mention (because it's quite related to `--preserve-fds`).

docs/terminals.md
cyphar: Yup, though in this section I thought it was implied. I can add a one-liner about it.

docs/terminals.md
cyphar: Sure, I'll explain what the binding does (roughly).

docs/terminals.md
deitch: That helps, thanks.

docs/terminals.md
```diff
@@ -0,0 +1,314 @@
...
+To this end, there is a new option (which is required if you want to use `runc`
+in detached new terminal mode): `--console-socket`. This option takes the path
+to a Unix domain socket which `runc` will connect to and send the
+pseudo-terminal master file descriptor down. The general process for getting
+the pseudo-terminal master is as follows:
+
+1. Create a unix domain socket at some path, `$socket_path`.
+2. Call `runc run` or `runc create` with the argument `--console-socket
+   $socket_path`.
+3. Using `recvmsg(2)` retreive the file descriptor sent using `SCM_RIGHTS` by
```
deitch: One nit: `retrieve` instead of `retreive` on the above line.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1731, opened 2018-02-22
Abstract recvtty to a single function

deitch:
Right now, to use a console socket to retrieve the file descriptor for the container's console when `terminal: true`, you need to figure out how to do it from reverse-engineering `recvtty.go`.

This PR creates a `func` in `utils.go` so a go client easily can do:

```go
import 	"github.com/opencontainers/runc/libcontainer/utils"

	// Open a socket.
	c, err := utils.ConsoleSocket(path)

       // c is a Console. so it implements Reader, Writer, Closer
      c.Read() 
      c.Write() 
      c.Close()
```

It also changes `recvtty.go` to leverage it.

Open to suggestion of course.

### DIFF
--- a/contrib/cmd/recvtty/recvtty.go
+++ b/contrib/cmd/recvtty/recvtty.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/containerd/console"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


deitch:
Also discussed with @cyphar earlier.

deitch:
@cyphar this also is ~2 months old, per our discussion back then. Please review?

dqminh:
Not sure if its necessary to extract this out because this may encourage people to pull in the utils lib instead of other more friendly way such as using a [runc-bindings](https://github.com/containerd/go-runc).

For example, there is https://godoc.org/github.com/containerd/go-runc#Socket to interact with the socket.

deitch:
@dqminh it has been 4+ months since I opened it, so I cannot recall all the details about what I was after at the time, except that I had come across several apps that had to do a chunk of work to interact with it, essentially replicating what was in `recvtty.go`.

> For example, there is https://godoc.org/github.com/containerd/go-runc#Socket to interact with the socket.

What would the moral equivalent of the above be using `go-runc`? How would I connect/read/write/close using that?

dqminh:
@deitch To use go-runc's Socket interface, it should roughly be:

```
socket, _ = runc.NewConsoleSocket(path)
console, _ = socket.ReceiveMaster()
```

console is https://godoc.org/github.com/containerd/console#Console which you can read/write/close/resize etc.

You can see some example of the usages in containerd https://github.com/containerd/containerd/blob/356b0d59646253bc040e8d6365418f20badd59eb/runtime/linux/proc/init.go#L213-L220

deitch:
Oh that is nice. It essentially provides what I was trying to do here, then? If so, I think we can close this out.

dqminh:
@deitch indeed, its quite similar to the proposed api here. I'm going to close this. Thanks for the help !

Issue #1734, opened 2018-02-23
Drop in-container /proc requirement?


wking:
Spun off from #1726, it would be nice to not require `/proc` inside containers.  Current blockers:

* [ ] capabilities loading, but with syndtr/gocapability#14 landed I'll be able to fix that.
* [ ] [closing extra file descriptors][1], but we can handle that before leaving the host mount namespace.
* [ ] [opening the `start`-signal FIFO][2], but we can inherit a socket through from outside withough reopening.

Do we need `/proc` for anything else?  @cyphar [seems to imply we do][3] (or I'm reading him wrong).  This is definitely an edge case, but I'd like to get it working.

[1]: https://github.com/opencontainers/runc/blob/595bea022f077a9e17d7473b34fbaf1adaed9e43/libcontainer/utils/utils_unix.go#L14
[2]: https://github.com/opencontainers/runc/blob/595bea022f077a9e17d7473b34fbaf1adaed9e43/libcontainer/standard_init_linux.go#L163-L167
[3]: https://github.com/opencontainers/runc/pull/1726#issuecomment-367885718

cyphar:
So, the short version of my opinion on this is that removing the `/proc` requirement from inside the container would be quite cool (and it also would allow us to beef up security around malicious `pid1` containers -- something that all container runtimes are lacking these days).

My comment in #1726 was a bit of a jumbled mess. The thrust of my point is that there are several theoretical security issues in `runc` at the moment related to `/proc` -- and while we can fix some of these issues, in order to fix all of them we need some extra kernel primitives we don't have at the moment.

> opening the start-signal FIFO, but we can inherit a socket through from outside withough reopening.

Can you clarify what you mean here? The O_PATH and re-opening trick is necessary to protect against container escapes -- we could send it via a unix socket, but that feels more complicated. My goal would be to get `openat` to accept `AT_EMPTY_PATH` -- but this is something that I'd need to get Al Viro and Eric Biederman to accept.

> closing extra file descriptors, but we can handle that before leaving the host mount namespace.

This can be done without needing `/proc` at all -- you can do what `rpm` does, where you do a `fnctl(fd, F_GETFD)`. The downside though is that you don't know which FDs are actually mapped, so it's quite expensive to check all of them (it would be nice if there was a syscall API for this -- but I wouldn't bet on this being a popular request).

> @cyphar seems to imply we do (or I'm reading him wrong).

[...]

PR #1735, opened 2018-02-23
libcontainer: Use gocapability's NewPid2

wking:
For `newContainerCapList`, we don't care what the current values are.  Using `NewPid2` saves us a few syscalls at no cost.

For `bootstrapData`, this avoids crashing when there is no container-side `/proc`.  Previously you'd get:

    container_linux.go:… starting container process caused "open /proc/self/status: no such file or directory"

The `IsNotExist` check is because we do need to load the effective set.  But [`capsV3`'s `Load()` does that first with the `capget()` call][1].  After the `capget` call [it hits `/proc` for bounding and ambient capabilities][2].  We don't need those, so a not-exist error isn't a problem.  This is a fairly tight binding to the current gocapability implementation, but we vendor gocapability, so I'm not too worried about it.  If it becomes an issue we can follow up with a `LoadType(which CapType)` so we can explicitly ask to only load the effective set.

Part of #1734.

[1]: https://github.com/syndtr/gocapability/blob/33e07d32887e1e06b7c025f27ce52f62c7990bc0/capability/capability_linux.go#L399
[2]: https://github.com/syndtr/gocapability/blob/33e07d32887e1e06b7c025f27ce52f62c7990bc0/capability/capability_linux.go#L406-L434

### DIFF
--- a/libcontainer/capabilities_linux.go
+++ b/libcontainer/capabilities_linux.go
@@ -71,7 +71,7 @@ func newContainerCapList(capConfig *configs.Capabilities) (*containerCapabilitie
 		}
 		ambient = append(ambient, v)
 	}
-	pid, err := capability.NewPid(0)
+	pid, err := capability.NewPid2(0)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


kolyshkin:
Obsoleted by #2029 (commit eb86f603).

PR #1738, opened 2018-02-24
libcontainer/configs/config: Clear hook environ variables on empty Env

wking:
The runtime spec [has][1]:

> * env (array of strings, OPTIONAL) with the same semantics as IEEE Std 1003.1-2008's environ.

And running `execle` or similar with `NULL` `env` results in an empty environent:

```console
$ cat test.c
#include <unistd.h>

int main()
{
  return execle("/usr/bin/env", "env", NULL, NULL);
}
$ cc -o test test.c
$ ./test
…no output…
```

[Go's `Cmd.Env`][2], on the other hand, has:

> If Env is nil, the new process uses the current process's environment.

This commit works around that by setting ~~a single dummy environment variable~~ `[]string{}` in those cases to avoid leaking the runtime environment into the hooks.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#posix-platform-hooks
[2]: https://golang.org/pkg/os/exec/#Cmd

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -480,6 +480,9 @@ func (c Command) Run(s *specs.State) error {
 		Stdout: &stdout,
 		Stderr: &stderr,
 	}
+	if cmd.Env == nil {
+		cmd.Env = []string{}
+	}
 	if err := cmd.Start(); err != nil {
 		return err
 	}


libcontainer/configs/config.go
```diff
@@ -320,6 +320,9 @@ func (c Command) Run(s HookState) error {
 		Stdout: &stdout,
 		Stderr: &stderr,
 	}
+	if cmd.Env == nil {
+		cmd.Env = []string{"_GO_DOES_NOT_PROVIDE_A_WAY_TO_CLEAR_ENV="}
```
cyphar: Does `cmd.Env = []string{}` also not work?

wking: > Does `cmd.Env = []string{}` also not work?

Looks like [it does][1] :).  Fixed with 5f6ed84c → 96e87700.

[1]: https://play.golang.org/p/UZDaL_Ukiht

wking:
> Ah sorry, was mislead by the comments above...

I've updated the initial comment to avoid misleading others.

kolyshkin:
This makes sense but needs a test case (in e.g. `tests/integration/hooks.bats`)

lifubang:
Carried in #4323

kolyshkin:
Closing in favor of #4323

Issue #1710, opened 2018-02-02
'prestart/poststart' hooks are done in the 'create' operation


liangchenye:
We are working on the [lifecycle test](https://github.com/opencontainers/runtime-tools/pull/558) and found 'runc' calls 'prestart/poststart' hooks in the 'create' phase.

It is not compliant with the spec: 
https://github.com/opencontainers/runtime-spec/blob/master/config.md#prestart
https://github.com/opencontainers/runtime-spec/blob/master/config.md#poststart

There are similar OLD issues #1574 and #1135.

liangchenye:
Another issue found in https://github.com/opencontainers/runtime-tools/pull/569.
If the poststart hook fails, 'runc' fails in 'create a container', but it should just record a warning and continue to work according to the spec.

liangchenye:
New issue found in opencontainers/runtime-tools#569.
If we have more than one hooks in poststart and one of them fails, the rest hooks should continue to run. But it is not now.

PR #568, opened 2016-02-17
Move pre-start hooks after container mounts

mrunalp:
Today mounts in pre-start hooks get overriden by the default mounts.
Moving the pre-start hooks to after the container mounts and before
the pivot/move root gives better flexiblity in the hooks.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

PR #1741, opened 2018-02-25
libcontainer: Set 'status' in hook stdin

wking:
Finish off the work started in #1201 to bring hook stdin into compliance with runtime-spec.

Also:

* Drop `HookState`, since there's no need for a local alias for `specs.State`.
* ~~Add `len` guards to avoid computing the state when `.Hooks` is non-`nil` but the particular phase we're looking at is empty.~~ I had to drop these to satisfy @cyphar [here][4].
* Also set `c.initProcess` in `newInitProcess` to support `OCIState` calls from within `initProcess.start()`.  I think the cyclic references between `linuxContainer` and `initProcess` are unfortunate, but didn't want to address that here.
* I've also left the timing of the `Prestart` hooks alone, although [the spec calls for them to happen before start][1] (not as part of creation, #1710).  Once the timing gets fixed we can drop the `initProcessStartTime` hacks which `initProcess.start` currently needs.

I'm not sure why we trigger the prestart hooks in response to both [`procReady`][2] and [`procHooks`][3].  But we've had two prestart rounds in `initProcess.start` since #568.  I've left that alone too.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#prestart
[2]: https://github.com/opencontainers/runc/blob/b50fa98d9e5ec3ff58028c46767129607067f961/libcontainer/process_linux.go#L343
[3]: https://github.com/opencontainers/runc/blob/b50fa98d9e5ec3ff58028c46767129607067f961/libcontainer/process_linux.go#L374
[4]: https://github.com/opencontainers/runc/pull/1741#discussion_r233331570

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -25,6 +25,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runc/libcontainer/utils"
+	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/sirupsen/logrus"
@@ -154,6 +155,12 @@ func (c *linuxContainer) State() (*State, error) {
 	c.m.Lock()
 	defer c.m.Unlock()
 	return c.currentState()
+}
+
+func (c *linuxContainer) OCIState() (*specs.State, error) {
+	c.m.Lock()
+	defer c.m.Unlock()
+	return c.currentOCIState()
 }
 
 func (c *linuxContainer) Processes() ([]int, error) {
@@ -349,13 +356,9 @@ func (c *linuxContainer) start(process *Process) error {
 		c.initProcessStartTime = state.InitProcessStartTime
 
 		if c.config.Hooks != nil {
-			bundle, annotations := utils.Annotations(c.config.Labels)
-			s := configs.HookState{
-				Version:     c.config.Version,
-				ID:          c.id,
-				Pid:         parent.pid(),
-				Bundle:      bundle,
-				Annotations: annotations,
+			s, err := c.currentOCIState()
+			if err != nil {
+				return err
 			}
 			for i, hook := range c.config.Hooks.Poststart {
 				if err := hook.Run(s); err != nil {
@@ -493,7 +496,7 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
 	if err != nil {
 		return nil, err
 	}
-	return &initProcess{
+	init := &initProcess{
 		cmd:             cmd,
 		childPipe:       childPipe,
 		parentPipe:      parentPipe,
@@ -504,7 +507,9 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, parentPipe, c
 		process:         p,
 		bootstrapData:   data,
 		sharePidns:      sharePidns,
-	}, nil
+	}
+	c.initProcess = init
+	return init, nil
 }
 
 func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, parentPipe, childPipe *os.File) (*setnsProcess, error) {
@@ -1537,13 +1542,9 @@ func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Proc
 		}
 	case notify.GetScript() == "setup-namespaces":
 		if c.config.Hooks != nil {
-			bundle, annotations := utils.Annotations(c.config.Labels)
-			s := configs.HookState{
-				Version:     c.config.Version,
-				ID:          c.id,
-				Pid:         int(notify.GetPid()),
-				Bundle:      bundle,
-				Annotations: annotations,
+			s, err := c.currentOCIState()
+			if err != nil {
+				return nil
 			}
 			for i, hook := range c.config.Hooks.Prestart {
 				if err := hook.Run(s); err != nil {
@@ -1733,6 +1734,27 @@ func (c *linuxContainer) currentState() (*State, error) {
 				ns := configs.Namespace{Type: nsType}
 				state.NamespacePaths[ns.Type] = ns.GetPath(pid)
 			}
+		}
+	}
+	return state, nil
+}
+
+func (c *linuxContainer) currentOCIState() (*specs.State, error) {
+	bundle, annotations := utils.Annotations(c.config.Labels)
+	state := &specs.State{
+		Version:     specs.Version,
+		ID:          c.ID(),
+		Bundle:      bundle,
+		Annotations: annotations,
+	}
+	status, err := c.currentStatus()
+	if err != nil {
+		return nil, err
+	}
+	state.Status = status.String()
+	if status != Stopped {
+		if c.initProcess != nil {
+			state.Pid = c.initProcess.pid()
 		}
 	}
 	return state, nil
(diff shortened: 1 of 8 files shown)


cyphar:
> I'm not sure why we trigger the prestart hooks in response to both procReady and procHooks. But we've had two prestart rounds in initProcess.start since #568.

That is almost definitely a bug, since the change you refer to was meant to *move* the hooks to be run earlier rather than duplicate running them. I'm a little confused why nobody has opened a bug about this (maybe pre-start just isn't used as often as post-start). /cc @mrunalp

> Once the timing gets fixed we can drop the initProcessStartTime hacks which initProcess.start currently needs.

I'm not sure this is correct, but will take a proper look when I get a chance.

wking:
> However, even with this PR, the runtime-tools validation/hooks_stdin.t test still fails with an empty status, as mentioned in opencontainers/runtime-tools#608.

Can you post the error you see after this patch?  In the list in my topic post here I point out some issues I'm not addressing; maybe you're hitting #1710?

dongsupark:
Basically the same as one from https://github.com/opencontainers/runtime-tools/pull/608#issuecomment-390181397.

```
TAP version 13
not ok 1 - the state of the container MUST be passed to "prestart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"\" in the stdin of prestart hook, expected \"created\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 2 - the state of the container MUST be passed to "poststart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"\" in the stdin of poststart hook, expected \"running\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 3 - the state of the container MUST be passed to "poststop" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* status in the stdin of poststop hook should not be empty",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
1..3
```

dongsupark:
@wking Sorry, ignore my comment above. There was a test setup issue.

Now with a correct test setup, I'm getting the following errors:

```
TAP version 13
not ok 1 - the state of the container MUST be passed to "prestart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"creating\" in the stdin of prestart hook, expected \"created\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
not ok 2 - the state of the container MUST be passed to "poststart" hook over stdin
  ---
  {
    "error": "1 error occurred:\n\n* wrong status \"created\" in the stdin of poststart hook, expected \"running\"",
    "reference": "https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#posix-platform-hooks"
  }
  ...
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x4f5a02]

goroutine 1 [running]:
github.com/opencontainers/runtime-tools/vendor/github.com/hashicorp/go-multierror.(*Error).Error(0x0, 0x7dfc20, 0xc420224480)
	/home/dpark/go/src/github.com/opencontainers/runtime-tools/vendor/github.com/hashicorp/go-multierror/multierror.go:15 +0x22
main.main()
	/home/dpark/go/src/github.com/opencontainers/runtime-tools/validation/hooks_stdin.go:154 +0x11ba
```

The first two failures happen because of https://github.com/opencontainers/runc/issues/1710, right?
The last panic is rather a runtime-tools issue. I'll create a PR to fix it in runtime-tools.

wking:
> The first two failures happen because of #1710, right?

Yeah.

cyphar:
So it looks like this has been merged into #1811. I'm going to close this and I'll get #1811 merged (hopefully for 1.0).

libcontainer/container_linux.go
```diff
@@ -359,14 +366,10 @@ func (c *linuxContainer) start(process *Process, isInit bool) error {
...
 
-		if c.config.Hooks != nil {
-			bundle, annotations := utils.Annotations(c.config.Labels)
-			s := configs.HookState{
-				Version:     c.config.Version,
-				ID:          c.id,
-				Pid:         parent.pid(),
-				Bundle:      bundle,
-				Annotations: annotations,
+		if c.config.Hooks != nil && len(c.config.Hooks.Poststart) > 0 {
```
cyphar: The `len` additions is really not necessary, and I think it should be removed from all the patches.

If you want, I can just combine this and #1811 in a new PR that I will rebase and fix so we can get 1.0 sooner.

wking: > The `len` additions is really not necessary...

From my commit message and topic post:

>  Add `len` guards to avoid computing the state when `.Hooks` is non-`nil` but the particular phase we're looking at is empty.

Are you saying that state computation is so cheap that it's not worth a `len(c.config.Hooks.Poststart) > 0` guard?

cyphar: Yeah, I don't think the state computation is expensive enough that it makes sense to have likely-to-be-incorrectly-copy-pasted checks.

wking: > ... likely-to-be-incorrectly-copy-pasted checks.

Wouldn't we expect unit testing to catch this?  If it's not covered now, I think it should be.  The runtime-tools compliance suite should already cover it.

wking: And building the status hits the filesystem for at least [this][1], and probably more.

[1]: https://github.com/opencontainers/runc/blob/4769cdf607b0d1cbd73e5407d9655da7511fe193/libcontainer/process_linux.go#L66

wking: So they seem like a slam dunk to me, but if you're still unconvinced when I wake back up, I'll drop the guards ;).

cyphar: `/proc` reading is quite cheap, and it should be noted that it will only be hit if `config.Hooks != nil` which IIRC would require having at least one hook defined... I mean, whatever. My main grip with it is that it looks a bit ugly and I guarantee someone will screw it up in a copy-paste.

wking: > `/proc` reading is quite cheap...

Compared to, say hitting a disk.  Certainly not compared to a `len` call.

> My main gripe with it is that it looks a bit ugly and I guarantee someone will screw it up in a copy-paste.

And get that screw-up through CI and review?  I'm sceptical ;).  But whatever, I'm not a maintainer.  Rebased onto master (because I'm touching it anyway) and dropped the `len` guards with 6ff5804 -> e238686.

cyphar:
LGTM.

cyphar:
This is one of two final PRs for 1.0. Thanks!

cyphar:
/cc @mrunalp 

You said you wanted to test this.

cyphar:
LGTM LGTM LGTM LGTM.

(Yoohoo! PullApprove! Time to wake up!)

cyphar:
I'll wait until @mrunalp checks this before merging.

RenaudWasTaken:
> Does this break you folks as well? This one is pretty critical for OCI compliance.

Thanks for asking!
My understanding of this PR is that you are adding the status field in the struct / json a hook can read from stdin.
In this case, this is a non breaking change in the "hook API" and it won't affect us, as we only rely on getting the `Bundle` to read the spec.

cyphar:
@RenaudWasTaken Yeah, that's right.

Merging since the change is won't break existing users.

PR #1743, opened 2018-02-26
libcontainer: setupUserNamespace is always called

### LINKED ISSUE #1742: Devices are mounted with wrong uid/gid
When creating a container with a device attached, the device does not inherit the correct uid/gid since version 1.0.0-rc4.
The test is based on a config.json generated by

```
docker run -it --device /dev/snd alpine
```
 
With runc version 1.0.0-rc4 :

```
$ sudo runc run test-v1.0.0-rc4

/ # ls -als /dev/snd
     0 crw-rw----    1 root     root      116,   2 Feb 26 10:35 controlC0
     0 crw-rw----    1 root     root      116,  10 Feb 26 10:35 hwC0D0
```

With runc version 1.0.0-rc3:

```
$ sudo runc run test-v1.0.0-rc3

/ # ls -als /dev/snd
     0 crw-rw----    1 root     986       116,   2 Feb 26 10:38 controlC0
     0 crw-rw----    1 root     986       116,  10 Feb 26 10:38 hwC0D0
```

ynirk:
The setupUserNamespace is always called even if the usernamespace is not set. This results having wrong uid/gid set on devices.

This fix add a test to check if usernamespace is set befor calling
setupUserNamespace.

Fixes #1742

Signed-off-by: Julien Lavesque <julien.lavesque@gmail.com>

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -192,9 +192,6 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	if err := createDevices(spec, config); err != nil {
 		return nil, err
 	}
-	if err := setupUserNamespace(spec, config); err != nil {
-		return nil, err
-	}
 	c, err := createCgroupConfig(opts)
 	if err != nil {
 		return nil, err
@@ -224,6 +221,11 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 				{
 					Type: "loopback",
 				},
+			}
+		}
+		if config.Namespaces.Contains(configs.NEWUSER) {
+			if err := setupUserNamespace(spec, config); err != nil {
+				return nil, err
 			}
 		}
 		config.MaskPaths = spec.Linux.MaskedPaths


cyphar:
The code in question is still broken with user namespaces, because the logic in `setupUserNamespaces` is wrong. It should only be setting the owner to root if and only if the uid or gid are unmapped in the user namespaces set (so if the uid is unmapped for a given device, set the uid to root).

However in general I also think that the the logic of "magically change the owners" doesn't make much sense either -- so maybe this section should just be dropped entirely.

crosbymichael:
LGTM

Also I think @cyphar point makes sense about changing the logic in a follow up PR.  I think it was originally like this because the userns was erroring when no uid/gid was set on a device or the device was owned by a weird user before.  We would have to test and see if we can remove it fully.

crosbymichael:
LGTM

crosbymichael:
@cyphar this should probably go in RC5 and 1.0.  It was a bug merged in that caused this because we had a check before that wouldn't run this code unless userns was specified .

mrunalp:
LGTM

cyphar:
@crosbymichael Alright, I'll send out a REJECT and then re-send the release mail. Merging.

LGTM (for the record).

PR #1746, opened 2018-02-28
Fix make shell

tiborvass:
The "shell" rule in the Makefile uses docker to run a bash session,
however it was depending on the "all" rule which assumes non-docker local
development. This commit fixes it by making it depend on the "runcimage" rule.


cyphar:
LGTM.

crosbymichael:
LGTM

Issue #1745, opened 2018-02-27
`TestExecInTTY` and `TestCheckpoint` fail on master on fresh ubuntu droplet


tiborvass:
Here are the logs of the tests, let me know if you need more information.

OS: Ubuntu 16.04.3
Kernel: `4.4.0-77-generic` but also tried on ubuntu's hwe kernel: `4.13.0-36-generic`

```
=== RUN   TestCheckpoint
--- FAIL: TestCheckpoint (1.31s)
	checkpoint_test.go:18: === /tmp/criu284986317/restore.log ===
	checkpoint_test.go:29: (00.000046) Added /sys/fs/cgroup/systemd:/sys/fs/cgroup/systemd ext mount mapping
	checkpoint_test.go:29: (00.000087) Added /sys/fs/cgroup/memory:/sys/fs/cgroup/memory ext mount mapping
	checkpoint_test.go:29: (00.000093) Added /sys/fs/cgroup/pids:/sys/fs/cgroup/pids ext mount mapping
	checkpoint_test.go:29: (00.000098) Added /sys/fs/cgroup/net_cls,net_prio:/sys/fs/cgroup/net_cls,net_prio ext mount mapping
	checkpoint_test.go:29: (00.000103) Added /sys/fs/cgroup/cpu,cpuacct:/sys/fs/cgroup/cpu,cpuacct ext mount mapping
	checkpoint_test.go:29: (00.000107) Added /sys/fs/cgroup/hugetlb:/sys/fs/cgroup/hugetlb ext mount mapping
	checkpoint_test.go:29: (00.000112) Added /sys/fs/cgroup/perf_event:/sys/fs/cgroup/perf_event ext mount mapping
	checkpoint_test.go:29: (00.000116) Added /sys/fs/cgroup/blkio:/sys/fs/cgroup/blkio ext mount mapping
	checkpoint_test.go:29: (00.000121) Added /sys/fs/cgroup/devices:/sys/fs/cgroup/devices ext mount mapping
	checkpoint_test.go:29: (00.000125) Added /sys/fs/cgroup/cpuset:/sys/fs/cgroup/cpuset ext mount mapping
	checkpoint_test.go:29: (00.000129) Added /sys/fs/cgroup/freezer:/sys/fs/cgroup/freezer ext mount mapping
	checkpoint_test.go:29: (00.000134) Added /dev/null:/dev/null ext mount mapping
[...]

PR #1750, opened 2018-02-28
tests: allow to load kernel modules from a test container

avagin:
CRIU needs to load a few modules to checkpoint/resume containers.

https://github.com/opencontainers/runc/issues/1745

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,13 +65,13 @@ localtest:
 	make localunittest localintegration localrootlessintegration
 
 unittest: runcimage
-	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localunittest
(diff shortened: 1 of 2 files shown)


tiborvass:
@avagin this fixed my issue, now `make test` succeeds for me too, thank you!

dqminh:
LGTM

Makefile
```diff
@@ -65,13 +65,13 @@ localtest:
 	make localunittest localintegration localrootlessintegration
 
 unittest: runcimage
-	docker run -e TESTFLAGS -t --privileged --rm -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localunittest
```
cyphar: Should `/lib/modules` be read-only? I'm a little concerned about making it read-write (in case something breaks and we overwrite some critical modules).

avagin: You are right. Will fix.

cyphar:
LGTM. Thanks @avagin.

mrunalp:
LGTM

PR #1752, opened 2018-03-06
cgroups/fs: fix NPE on Destroy than no cgroups are set

dennwc:
Currently Manager accepts nil cgroups when calling Apply, but it will panic then trying to call Destroy with the same config.


crosbymichael:
LGTM

cyphar:
LGTM.

PR #1753, opened 2018-03-06
libcontainer/specconv/spec_linux: Support empty 'type' for bind mounts

wking:
From the “Creating a bind mount” section of [`mount(2)`][1]:

> If mountflags includes MS_BIND (available since Linux 2.4), then perform a bind mount…
>
> The filesystemtype and data arguments are ignored.

This pull request adds support for configurations that leave the [OPTIONAL `type`][2] unset for bind mounts.  There's a related spec-example change in flight with opencontainers/runtime-spec#954, although my personal preference would be a more explicit spec for the whole mount structure (opencontainers/runtime-spec#771).

[1]: http://man7.org/linux/man-pages/man2/mount.2.html
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config.md#L102

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -269,13 +269,17 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 func createLibcontainerMount(cwd string, m specs.Mount) *configs.Mount {
 	flags, pgflags, data, ext := parseMountOptions(m.Options)
 	source := m.Source
-	if m.Type == "bind" {
+	device := m.Type
+	if flags|unix.MS_BIND != 0 {
+		if device == "" {
+			device = "bind"
+		}
 		if !filepath.IsAbs(source) {
 			source = filepath.Join(cwd, m.Source)
 		}
 	}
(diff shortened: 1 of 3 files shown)


cyphar:
Looks okay, probably needs a test though. And I think `rootfs_linux` really needs a good cleanup after this anyway.

wking:
> … probably needs a test though.

What sort of test were you looking for?  An integration test?  Something in `spec_linux_test.go`?  I've pushed 060f9a37 → bc225e7c dropping our only internal explicit `Type = "bind"`, does that satisfy your need for testing (do we already test `notifySocket.setupSpec` somewhere?)?

cyphar:
> What sort of test were you looking for? An integration test? Something in spec_linux_test.go?

An integration test would be nice, but if that's too overkill for simply testing bind-mounts, then a `spec_linux_test.go` would probably be sufficient.

wking:
I've pushed bc225e7 → 0aa6e4e5 adding an integration test for this.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1754, opened 2018-03-07
Add timeout while waiting for StartTransinetUnit completion signal

vikaschoudhary16:
This PR adds one second timeout for waiting on channel for `StartTransientUnit` completion signal.

This channel was introduced to avoid a rare race between systemd and runc where systemd could delete pids cgroup created by runc. https://github.com/opencontainers/runc/pull/1683

If for some reason signal is not received like the case [here](https://bugzilla.redhat.com/show_bug.cgi?id=1548358#c22), runc will hang forever. Adding a timeout will help recover in such a case.

In the timeout handling, we have two options either return Error or continue. Since originally purpose of introducing channel was to avoid the race mentioned above, one second time duration should be enough in ensuring the same and to continue seems a better option of the two.

/cc @derekwaynecarr @sjenning

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -17,6 +17,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fs"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/sirupsen/logrus"
 )
 
 type Manager struct {
@@ -300,7 +301,11 @@ func (m *Manager) Apply(pid int) error {
 		return err
 	}
 
-	<-statusChan
+	select {
+	case <-statusChan:
+	case <-time.After(time.Second):
+		logrus.Warnf("Timed out while waiting for StartTransientUnit completion signal from dbus. Continuing...")
+	}
 
 	if err := joinCgroups(c, pid); err != nil {
 		return err


libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -300,7 +300,10 @@ func (m *Manager) Apply(pid int) error {
 		return err
 	}
 
-	<-statusChan
+	select {
+	case <-statusChan:
+	case <-time.After(time.Second):
```
cyphar: Should have some sort of warning at least. I would also recommend investigating *why* we don't get the status from systemd. In the issue you said:

> Not sure how dbus works from within the containerized env or how we might fix this.

I'm not sure what containerised environment you're using, but dbus should be running as a daemon inside whatever container you're spawning `runc` inside as it normally would on your host. Can you reproduce the issue using `runc` on the host?

vikaschoudhary16: @cyphar added warning. 
> what containerised environment ...

It occured in containerized OpenShift. I am trying to reproduce it on my local machine. Will update on that further.

sjenning: I believe the way this has worked in the past is that the origin node container runs as privileged, which implies a shared IPC and PID namespace with the host such that the containerized node can get dbus signals from the host dbus.

sjenning:
@cyphar we are trying to run down the root cause, but in the meantime, I think we should merge this.  Even if we get the dbus issue figured out, I'm not sure if we should consider dbus reliable.  If, for whatever reason, it drops the message or systemd fails to send the signal we expect, we'll are still hung.  Said another way, this probably should have never been an unconditional wait on the channel with no timeout.

Do you agree?

mrunalp:
LGTM

sjenning:
@hqhq could we get lgtm on this since you approved the previous PR to which the fix applies?

hqhq:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runc/pull-request/1754/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runc)

derekwaynecarr:
i agree this should have had a timeout from its inception, i would like to understand the root cause.

Issue #1523, opened 2017-07-16
bind mounting /etc/resolv.conf fails 


yuvipanda:
I'm trying to run a rootless container on a Linux machine.

*How to reproduce*

```
$ skopeo --insecure-policy copy docker://opensuse/amd64:42.2 oci:opensuse:latest
$umoci unpack --rootless --image singleuser:latest bundle
$ runc  --root /tmp/^Cnc run  test
container_linux.go:265: starting container process caused "process_linux.go:348: container init caused \"rootfs_linux.go:57: mounting \\\"/etc/resolv.conf\\\" to rootfs \\\"/home/yuvipanda/code/ferry-commute/bundle/rootfs\\\" at \\\"/home/yuvipanda/code/ferry-commute/bundle/rootfs/etc/resolv.conf\\\" caused \\\"operation not permitted\\\"\""
```

I suspect that the problem is related to the fact that my $HOME is mounted with encfs and has nodev & nosuid.

```
/home/.ecryptfs/yuvipanda/.Private on /home/yuvipanda type ecryptfs (rw,nosuid,nodev,relatime,ecryptfs_fnek_sig=bf2ce9e5622d0c40,ecryptfs_sig=ae8560bd6ae1015e,ecryptfs_cipher=aes,ecryptfs_key_bytes=16,ecryptfs_unlink_sigs)
```

yuvipanda:
hmm, might not be related to nodev / nosuid. I mounted a tmpfs and it still has the same problem.

stracing yields:

```
25813 mount("/mnt/ramdisk/bundle/rootfs", "/mnt/ramdisk/bundle/rootfs", 0xc4200beac8, MS_BIND|MS_REC, NULL) = 0
25813 stat("/mnt/ramdisk/bundle/rootfs/proc", {st_mode=S_IFDIR|0755, st_size=40, ...}) = 0
25813 mount("proc", "/mnt/ramdisk/bundle/rootfs/proc", "proc", 0, NULL) = 0
25813 stat("/mnt/ramdisk/bundle/rootfs/dev", {st_mode=S_IFDIR|0755, st_size=300, ...}) = 0
25813 mount("tmpfs", "/mnt/ramdisk/bundle/rootfs/dev", "tmpfs", MS_NOSUID|MS_STRICTATIME, "mode=755,size=65536k") = 0
25813 fchmodat(AT_FDCWD, "/mnt/ramdisk/bundle/rootfs/dev", 0755) = 0
[...]

PR #1759, opened 2018-03-16
rootless: cgroup: treat EROFS as a skippable error

cyphar:
In some cases, /sys/fs/cgroups is mounted read-only. In rootless
containers we can consider this effectively identical to having cgroups
that we don't have write permission to -- because the user isn't
responsible for the read-only setup and cannot modify it. The rules are
identical to when /sys/fs/cgroups is not writable by the unprivileged
user.

An example of this is the default configuration of Docker, where cgroups
are mounted as read-only as a preventative security measure.

Fixes #1657
Reported-by: Vladimir Rutsky <rutsky@google.com>

<hr>

I've included a change related to `oom_score_adj` here as well, because
@jonboulle mentioned that he's been having permission issues recently and I
discovered that we haven't been following the spec. So I fixed that too.


libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	libcontainerUtils "github.com/opencontainers/runc/libcontainer/utils"
+	"github.com/pkg/errors"
```
jonboulle: remove line 6

jonboulle:
thanks! I can confirm that with the combination of 
- this patch (once the build is fixed (-:) 
- `privileged=true` (thanks to #1658)
- a `resolv.conf` mount workaround (thanks to #1523, which is still a pain)

that runc with rootless containers works again in my new environment (which happens to be GKE).

cyphar:
Fixed the build issue.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -35,7 +36,7 @@ var (
 	HugePageSizes, _ = cgroups.GetHugePageSize()
 )
 
-var errSubsystemDoesNotExist = errors.New("cgroup: subsystem does not exist")
+var errSubsystemDoesNotExist = fmt.Errorf("cgroup: subsystem does not exist")
```
AkihiroSuda: why?

cyphar: We have a `github.com/pkg/errors` import, so we can't import `errors` (without renaming and that's just ugly). I could do `errors.Errorf` but then you'd get a stack trace in the package which isn't really useful. `fmt.Errorf` is otherwise essentially the same as `errors.New`.

AkihiroSuda: Why not use https://godoc.org/github.com/pkg/errors#New

cyphar: > New also records the stack trace at the point it was called.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -100,6 +102,27 @@ type cgroupData struct {
 	pid       int
 }
 
+// isIgnorableError returns whether err is a permission error (in the loose
+// sense of the word). This includes EROFS (which for an unprivileged user is
+// basically a permission error) and EACCES (for similar reasons) as well as
+// the normal EPERM.
+func isIgnorableError(err error) bool {
```
hqhq: Does this only make sense for rootless container? If so, can we add rootless check in it?

cyphar: The rootless check is already done at the only caller. We could pass `isRootless` as a bool if you think it'll look cleaner, but we can't just do an `os.Geteuid()` here because (in theory) you can run as rootless in other contexts too.

hqhq: > We could pass isRootless as a bool if you think it'll look cleaner

That'll be better, my concern is that people may use this function in other circumstances and that's actually not what they expected.

AkihiroSuda:
Any chance to get this merged?

crosbymichael:
LGTM

AkihiroSuda:

libcontainer/cgroups/fs/apply_raw.go
cyphar: Oh, I forgot to fix this. I'll send a follow-up.

PR #1760, opened 2018-03-19
rootless: set sticky bit if using XDG_RUNTIME_DIR

### LINKED ISSUE #1694: rootless: set sticky bits to the files under XDG_RUNTIME_DIR
According to https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html :

> To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.

For rootless runc, setting sticky bits seems the easiest solution.

cyphar:
According to the XDG specification[1], in order to avoid the possibility of
our container states being auto-pruned every 6 hours we need to set the
sticky bit. Rather than handling all of the users of --root, we just
create the directory and set the sticky bit during detection, as it's
not expensive.

[1]: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Fixes #1694

### DIFF
--- a/main.go
+++ b/main.go
@@ -67,6 +67,15 @@ func main() {
 		runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
 		if runtimeDir != "" {
 			root = runtimeDir + "/runc"
+			// According to the XDG specification, we need to set anything in
+			// XDG_RUNTIME_DIR to have a sticky bit if we don't want it to get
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #1763, opened 2018-03-22
Implement setns for cgroupns by path

alban:
The OCI runtime spec allows to specify a cgroupns path to enter:
https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#namespaces

Example:
```
    "namespaces": [
        {
            "type": "cgroup",
            "path": "/proc/1234/ns/cgroup"
        },
```

-----

Issue found while trying to implement a test for this:
https://github.com/opencontainers/runtime-tools/issues/572#issuecomment-374949214

### DIFF
--- a/libcontainer/configs/namespaces_linux.go
+++ b/libcontainer/configs/namespaces_linux.go
@@ -7,12 +7,13 @@ import (
 )
 
 const (
-	NEWNET  NamespaceType = "NEWNET"
-	NEWPID  NamespaceType = "NEWPID"
-	NEWNS   NamespaceType = "NEWNS"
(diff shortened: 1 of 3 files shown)


cyphar:
Thanks for this, but we're already working on it as part of #1184 (which I'm working on carrying at the moment). In short (you can read the discussion in the thread), the "simple" way of implementing cgroup namespaces doesn't work as you'd expect because you will end up with having the root of the cgroup namespace being the cgroup the process was in before `runc` put it in a separate cgroup.

I'm closing this as a dup of #1184 (and I am going to carry it in the near future).

PR #1768, opened 2018-03-27
add gitignore line for supporting goland IDE

sangyun-han:
I added gitignore option `.idea` for supporting Jetbrain goland IDE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ vendor/pkg
 contrib/cmd/recvtty/recvtty
 man/man8
 release
+.idea


ijc:
Rather than having all projects in the world carrying the union of ignore files for all possible IDEs it is a better idea to simply add the ones relevant to the IDE you use to your global `~/.gitignore` (or `.config/git/ignore`) or, if you want it local to a given checkout to `.git/info/exclude`. See `gitignore(5)` for details of how the various sources of ignore combine.

PR #1769, opened 2018-03-27
add type casting for uint64 parameter

sangyun-han:
Add type casting for major.unix method. Its input parameter is uint64, but the parameter in unix.Stat_t is int32.


cyphar:

sangyun-han:
I added `Signed-off-by`. 
Thank you for the comment.

dqminh:
LGTM.

I believe this is only uint32 for MIPS though.

cyphar:
Closing since #1824 was merged (which is basically the same thing). Thanks.

PR #1770, opened 2018-03-30
support hierarchy setting for memory cgroup subsystem

ddongchen:
We should support setting hierarchy for memory cgroup subsystem.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -184,6 +184,11 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 			return err
 		}
 	}
+	if cgroup.Resources.UseHierarchy {
+		if err := writeFile(path, "memory.use_hierarchy", "1"); err != nil {
+			return err
+		}
+	}
 	if cgroup.Resources.MemorySwappiness == nil || int64(*cgroup.Resources.MemorySwappiness) == -1 {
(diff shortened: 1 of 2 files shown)


cyphar:

cyphar:
In addition, you're adding `libcontainer` changes without any other corresponding changes -- users won't be able to set that value.

As for the *spirit* of the change, I think that the idea is fine, but the execution is a bit more complicated than just setting `use_hierarchy`. In particular, I believe you need to set it in the root of the cgroup in order for the setting to be useful, which means that we have to modify global cgroup policy. Most distributions already set `use_hierarchy=1` by default (Ubuntu used to be well-known for not doing it, but they do it now). So I'm wondering in which cases you actually have had a distribution that was using the (broken) hierarchy in memcg.

ddongchen:
@cyphar, The distribution in use is fedora, the **use_hierarchy** is 0 by default(maybe the version is too old, it's linux kernel 3.10). Therefore, we need to set the **use_hierarchy** for root cgroup or the related child cgroup. 
Besides, we should support the user to set **0** for **use_hierarchy**, if they do not want to static the children's usage.

ddongchen:
I signed it!

ddongchen:
I added Signed-off-by.
Thank you for the comment.

AkihiroSuda:
Fedora switched away to cgroup v2.
Can we close this?

PR #1771, opened 2018-03-30
libcontainer: Don't set container state to running when exec'ing

### LINKED ISSUE #1767: libcontainer: Do not change container state when exec'ing a process
I was trying to exec a process on a container that was not `running` from Kata Containers, and I have realized that we get an issue because of the way our agent relies on `libcontainer`. Indeed, our agent is a long lasting process that keeps `libcontainer.Container` containers and `libcontainer.Process` processes pointers in memory.
The case of the `exec` being started on a `created` container does not support this behavior. It assumes the container status will be asked from the disk everytime. That's a problem in our case and I found how to fix this, by removing [this](https://github.com/opencontainers/runc/blob/v1.0.0-rc5/libcontainer/container_linux.go#L380-L383).

Basically, this __else statement__ modifies the in-memory status of the container but never saves/stores it on disk. That's a problem when you rely exclusively on in-memory statuses like we do in case of our agent.

Now, I have opened this issue to suggest that we could remove this code part, but I might be missing a real use case for that, and I am looking for guidance here.

@crosbymichael @mrunalp @cyphar please let me know what you think about this.

sboeuf:
There is no reason to set the container state to "running" as a
temporary value when exec'ing a process on a container in "created"
state. The problem doing this is that consumers of the libcontainer
library might use it by keeping pointers in memory. In this case,
the container state will indicate that the container is running, which
is wrong, and this will end up with a failure on the next action
because the check for the container state transition will complain.

Fixes #1767

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -377,10 +377,6 @@ func (c *linuxContainer) start(process *Process, isInit bool) error {
 				}
 			}
 		}
-	} else {
-		c.state = &runningState{
-			c: c,
-		}
 	}
 	return nil
 }


sboeuf:
@mrunalp is there any test I can run to make sure this does not break some expected behavior ?

crosbymichael:
I'll run this patch on the containerd stress test server and see if we have any issues.

sboeuf:
Thank you @crosbymichael !

crosbymichael:
I'll let it run all weekend and let you know.  so far so good as far as identifying any regressions on the `runc` consumer side.

sboeuf:
@crosbymichael that's great, I really don't want to break anything and such testing seems very appropriate to confirm if we can merge this.
Thank you !

mrunalp:
Looks fine :+1: will wait for the test results :)

sboeuf:
@crosbymichael any results ?

crosbymichael:
No issues

crosbymichael:
LGTM

sboeuf:
Cool :)

sboeuf:
@mrunalp PTAL

mrunalp:
LGTM

PR #1775, opened 2018-04-04
kill: add an option --signal to specify a signal

dongsupark:
According to the command line interface of [runtime-tools](https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md#kill), signal needs to be set by an option `--signal`. However, `runc kill` does not have the option, but just takes the first argument as the signal.

Let's add a new option `--signal` to `runc kill`. Then we are able to make runtime-tools call simply a kill command with `--signal` in [the validation utils](https://github.com/opencontainers/runtime-tools/blob/master/validation/util/container.go#L172..L174).

Signed-off-by: Dongsu Park <dongsu@kinvolk.io>

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -14,7 +14,7 @@ import (
 var killCommand = cli.Command{
 	Name:  "kill",
 	Usage: "kill sends the specified signal (default: SIGTERM) to the container's init process",
-	ArgsUsage: `<container-id> [signal]
+	ArgsUsage: `<container-id>
 
 Where "<container-id>" is the name for the instance of the container and
 "[signal]" is the signal to be sent to the init process.
@@ -23,18 +23,22 @@ EXAMPLE:
 For example, if the container id is "ubuntu01" the following will send a "KILL"
 signal to the init process of the "ubuntu01" container:
 	 
-       # runc kill ubuntu01 KILL`,
+       # runc kill --signal KILL ubuntu01`,
 	Flags: []cli.Flag{
 		cli.BoolFlag{
 			Name:  "all, a",
 			Usage: "send the specified signal to all processes inside the container",
+		},
+		cli.StringFlag{
+			Name:  "signal",
+			Usage: "specify a signal to be sent to processes",
 		},
 	},
 	Action: func(context *cli.Context) error {
 		if err := checkArgs(context, 1, minArgs); err != nil {
 			return err
 		}
-		if err := checkArgs(context, 2, maxArgs); err != nil {
+		if err := checkArgs(context, 1, maxArgs); err != nil {
 			return err
 		}
 		container, err := getContainer(context)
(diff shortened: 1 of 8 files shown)


cyphar:
This will break existing users of `runc` (such as `Docker`, `cri-o` and so on). They all have `runc` versions pinned, but it is something to keep in mind.

crosbymichael:
runtime-tools does not specify the interface, if anything, runtime-tools is incorrect.  The spec is source of truth and it specifically has it as an arg:

https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#kill

This shows how bad and out of sync runtime-tools and the maintainers are in the org and this needs to be fixed because they are off on the side doing/specing their own things. 

@dongsupark I'm closing this one and I'm going to follow up on runtime-tools and getting people inline.

wking:
> The spec is source of truth and it specifically has it as an arg:
>
> https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#kill

The spec is explicitly [*not* a command-line API spec][1]:

> Note: these operations are not specifying any command-line APIs, and the parameters are inputs for general operations.

For a backwards compatible switch to the command-line API spec, you could continue to support both approaches for some deprecation period:

    runc kill [--signal SIGNAL] ID [SIGNAL]

You'd error out if `SIGNAL` was set in both places and log a deprecation notice if it was set as a positional argument.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L85

crosbymichael:
@dongsupark runtime-tools should be updated.  runc should not change as its correct with the spec and because it would break backwards compat with current consumers.  I don't think we want to support two ways of doing the same thing.

wking:
> runc should not change as its correct with the spec and because it would break backwards compat with current consumers.

As I pointed out [above][1], I think we could make this transition fairly painlessly without much code overhead during the deprecation window.

> … I don't think we want to support two ways of doing the same thing.

Agreed in the long term.  But I prefer the CLI spec approach here, because the chosen signal is optional, and optional input seems like a better fit for [command-specific options][2] ;).  And, as mentioned in opencontainers/runtime-tools@5c87cc6469f41f, the `--signal` approach is closer to `kill(1)` as defined in [POSIX][3] (although POSIX has only the short `-s`), [util-linux][4], [procps-ng][5], and [coreutils][6].  I think long-term similarity with those widely-adopted commands is worth temporarily having two ways of doing this in runc.

[1]: https://github.com/opencontainers/runc/pull/1775#issuecomment-378675629
[2]: https://github.com/opencontainers/runtime-tools/blob/v0.5.0/docs/command-line-interface.md#global-usage
[3]: http://pubs.opengroup.org/onlinepubs/9699919799.2016edition/utilities/kill.html
[4]: http://man7.org/linux/man-pages/man1/kill.1.html
[5]: http://man7.org/linux/man-pages/man1/kill.1@@procps-ng.html
[6]: https://www.gnu.org/software/coreutils/manual/html_node/kill-invocation.html

PR #1776, opened 2018-04-06
Detect whether Delegate is available on both slices and scopes

filbranden:
Starting with systemd 237, in preparation for cgroup v2, delegation is only now available for scopes, not slices.

Update libcontainer code to detect whether delegation is available on both and use that information when creating new slices.

Fixes kubernetes/kubernetes#61474.

/cc @adelton, @jmontleon, @derekwaynecarr.


derekwaynecarr:
FYI @mrunalp and @sjenning

filbranden:
@crosbymichael Will do, I'll update the commit description and force-push it here.

@adelton The problem you reported in kubernetes/kubernetes#61474 is:

```
E0321 09:23:23.527391   27717 node_container_manager.go:52] Failed to create "/kubepods" cgroup
F0321 09:23:23.527441   27717 kubelet.go:1363] Failed to start ContainerManager Delegation not available for unit type
```

Which this commit should fix.

The problem you reported in the comment above (kubelet startup hanging) is actually fixed by #1772.

So I think you need both :-)

While #1772 by fixes the problem of kubelet hanging on systemd v236 and below, that commit by itself won't fix the problem you're seeing starting kubelet on systemd v238, then this commit will fix that too. Can you try with both of them cherry-picked?

Thanks!
Filipe

adelton:
Thank you. Now I took https://github.com/opencontainers/runc/pull/1776 and cherry-picked 8ab251f298b9a003b894d4acbfd8b4f74e36b4f1 from https://github.com/opencontainers/runc/pull/1772 on top of that, and copied `libcontainer/*` content to Kubernetes' `vendor/github.com/opencontainers/runc/libcontainer/`. I still see the

```
E0410 02:40:53.328962   15638 node_container_manager.go:53] Failed to create "/kubepods" cgroup
F0410 02:40:53.329024   15638 kubelet.go:1302] Failed to start ContainerManager Delegation not available for unit type
```

message and node not getting ready.

filbranden:
Thanks for testing @adelton 

I'm working with the hypothesis that systemd v238 will not return an error that Delegate is read-only (which is what the code is checking) but that the attribute does not exist (which is effectively the change that went in on v237.)

Will perform some additional tests using busctl and update this code to do that instead. (In fact, I think I used to have a "go run" test case I could use on v238 too, let me dig that up.)

I'll let you know when I have a new version you can test.

Cheers,
Filipe

filbranden:
Yes that was it... I just added `"org.freedesktop.DBus.Error.InvalidArgs"` to the list of errors to check and now my mini test case of creating a single slice works...

@adelton Can you please test this again in your environment?

Note that #1772 is now merged and I rebased this branch on top of that one, so if you just sync the vendor/ tree in Kubernetes to where this latest commit of mine is, you should be good to go... It would be nice to confirm that that's indeed fixed before we merge this!

Cheers,
Filipe

cyphar:
> Starting with systemd 237, in preparation for cgroup v2, delegation is only now available for scopes, not slices.

Why on earth... cgroupv2 still isn't ready yet and they're already removing features from systemd and breaking things "just for fun" because "everyone should just switch to cgroupv2 -- ignore that important controllers are totally absent from cgroupv2, and systemd doesn't support setting up different paths for different controllers because of cgroupv2". Not to mention that the scope/slice distinction is entirely an invention of systemd.

All of that being said, if they've broken it then we will need this patch (once rebased). Although I believe that adding `Delegate=true` will just result in a warning right -- it won't cause the slice creation to fail?

filbranden:
> it won't cause the slice creation to fail?

I tested it with a small 15-line Go program creating a slice. It fails with `"org.freedesktop.DBus.Error.InvalidArgs"`, you'll see my latest version of the commit includes that... So in a way it's already *tested* in that I created a slice on systemd v238 with it...

I'd still prefer to wait for @adelton to confirm that it does fix the issue in Kubernetes, just to make sure there's nothing else lurking there...

Cheers,
Filipe

mrunalp:
@umohnani is also testing this out. Thanks!

cgwalters:
This patch fixed https://bugzilla.redhat.com/show_bug.cgi?id=1558425 for me.    See also https://pagure.io/atomic-wg/issue/452#comment-505389

filbranden:
Awesome, thanks for testing! I think we're good to go... Merge?

filbranden:
Once this one is in, I'll update kubernetes/kubernetes#61926 to include everything including this PR to fix the issue in Kubernetes code base too.

Cheers!
Filipe

umohnani8:
Works for me also!

mrunalp:
LGTM

crosbymichael:
LGTM

PR #1777, opened 2018-04-11
Only configure networking when creating a net ns

nalind:
When joining an existing namespace, don't default to configuring a loopback interface in that namespace.

Its creator should have done that, and we don't want to fail to create the container when we don't have sufficient privileges to configure the network namespace.


cyphar:
LGTM.

crosbymichael:
LGTM

Issue #1780, opened 2018-04-14
Making systemd StartTransientUnit synchronous (mini post-mortem on that)


filbranden:
Opening an issue here so we can try to understand the sequence to events, the current situation and what to do next.

cc @derekwaynecarr @vikaschoudhary16 @sjenning @mrunalp 

1. The original behavior of libcontainer systemd cgroup driver was to send systemd the StartTransientUnit request via D-Bus, but not wait for a reply.
2. This caused a race condition, in that it's possible that runc/libcontainer would [create the directories](https://github.com/opencontainers/runc/blob/f753f300ae6a725becddbacf15955a0a03b895cb/libcontainer/cgroups/systemd/apply_systemd.go#L380) using [`os.MkdirAll`](https://golang.org/pkg/os/#MkdirAll) (which creates the directories, if they don't exist.)
3. systemd (at least v219, version in RHEL 7) will decide which controllers get created (which might not be the whole set) and will [remove](https://github.com/systemd/systemd/blob/v219/src/shared/cgroup-util.c#L1630) controllers that weren't supposed to exist.
4. Due to this race, a controller directory might get created by runc/libcontainer, while later be removed by systemd (still in the process of creating the cgroups for the scope.)
5. In order to fix this issue, PR #1683 attempted to make StartTransientUnit synchronous, so that the whole creation of the scope/slice by systemd is completed before runc/libcontainer does any cgroup processing.
  a. I believe this approach is appropriate. Waiting for the operation to complete before proceeding is the correct behavior here.
  b. The problem with PR #1683 was that the StartTransientUnit call breaks on errors only when `err != nil && !isUnitExists(err)`, so in the case where the unit already exists, blocking on the channel is a mistake, since there's no pending operation from systemd. (Personal note: this would have been unlikely to have happened on a language with proper exception handling.)
6. The bug introduced in step 5(b) started causing trouble in `kubelet` (Kubernetes), since it does call libcontainer on the same cgroup multiple times, and the latter calls will trigger the `isUnitExists(err)` situation that will make the libcontainer code now block on a channel that will never receive any message. So `kubelet` startup gets stuck there. (BTW, I spent a couple days trying to troubleshoot that myself too, seeing `kubelet` mysteriously fail to initialize when using systemd cgroup driver.)
7. To address that bug, PR #1754 introduced a timeout on blocking on the channel. If an answer was not received within one second, we would stop waiting on the channel.
[...]

PR #1781, opened 2018-04-14
Make channel for StartTransientUnit buffered

filbranden:
So that, if a timeout happens and we decide to stop blocking on the operation, the writer will not block when they try to report the result of the operation.

This should address Issue #1780 and it's a follow up for PR #1683, PR #1754 and PR #1772. Also relevant is kubernetes/kubernetes#61926 (which is likely to need to be updated to contain this one after it's merged.)

@derekwaynecarr @sjenning @vikaschoudhary16 @mrunalp 

This might need some more testing... Do you have a reproducer that triggers the problem? It would need to trigger the timeout in libcontainer to cause the bug that will later cause the deadlock. If there's a way to make the cgroup unit creation take longer than 1s, that might be enough...

Thanks!
Filipe


cyphar:

filbranden:
> You need to add a `Signed-off-by:` line to your commit(s)

Ah yes sorry about that... I did know about it, I just keep forgetting it (it turns out right now this is the only project I'm contributing to that needs it...) Anyways, fixed now. Thanks!

sjenning:
Sorry, I'm not understanding what issue this is fixing

filbranden:
Hi @sjenning 

If we [time out waiting for systemd to reply through D-Bus](https://github.com/opencontainers/runc/blob/63e6708c74c1cc46091ec92ea9df5aca4d82e803/libcontainer/cgroups/systemd/apply_systemd.go#L327), then no one will consume any message in `statusChan`.

Later on, when the D-Bus reply is received, [this code in go-systemd will try to update the channel to indicate completion](https://github.com/coreos/go-systemd/blob/v16/dbus/methods.go#L35). But as no one is consuming it, it will block forever.

The situation gets worse because the code in go-systemd is holding the `c.jobListener` lock, so essentially everything in go-systemd gets deadlocked at this point.

This fix here just makes the channel buffered with one slot, so that when go-systemd updates it, it doesn't get blocked. If we timed out, the message is lost (which is fine) but no one gets blocked.

See also the discussion here: https://github.com/kubernetes/kubernetes/pull/61926#discussion_r181460028, where @derekwaynecarr mentions the jobListener lock (that's how I tracked this down.)

I also tried to summarize the whole situation (including the history of PRs) on #1780, so try to go through that one if it's still unclear to you...

Thanks!
Filipe

filbranden:
Ping @sjenning @derekwaynecarr 

I'd like to have this one figured out so we can wrap up kubernetes/kubernetes#61926 update of the vendored libcontainer this week...

Cheers,
Filipe

mrunalp:
LGTM

PR #1788, opened 2018-04-23
Add runc API for building CLI extensions

crosbymichael:
This change adds an `api` and `commands` package that abstracts out the libcontainer implementation from what CLI consumers see.  It allows other producers of runtimes, that atleast write the runtime in Go to implement the `API` interface and get the CLI code for free.  The command package also allows for extensions and I have placed things like update, events, and the runc init command there as it should be out of the API spec.

This allows us to solidify the CLI as a spec and things like runtime-tools and other runtimes to depend on a consistent API for OCI containers and runtime-spec. This also allows us to write consistent client code for interacting with OCI runtimes based on the API interface. 

Please review and give your thoughts.  I think we will probably want to make the API interface smaller, such as removing checkpoint/restore into an extension API as not all OCI runtimes will implement this.


api/api.go
```diff
@@ -0,0 +1,118 @@
...
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+var (
+	ErrNotImplemented = errors.New("command not implemented")
+	ErrEmptyID        = errors.New("container id cannot be empty")
+)
+
+type API interface {
```
mrunalp: Should we call API as ContainerOperations?

crosbymichael: Ya, we can work on naming and how big this interface should be for general OCI compatibility.

crosbymichael:
You can still review, ci is failing because golang lint issues:

https://github.com/golang/lint/issues/397

api/api.go
crosbymichael: What do you think about adding `context.Context` to these methods?  It's not really important with the implementation side, but if we want to build clients the context is useful for `exec.CommandContext` to support timeouts and such.

crosbymichael:
Pushed a change moving checkpoint/restore to a separate interface blocked by an `ErrNotImplemented`

api/command/create.go
```diff
@@ -0,0 +1,100 @@
...
+			if err != nil {
+				return err
+			}
+			opts := api.CreateOpts{
+				PidFile:       pidFile,
+				ConsoleSocket: context.String("console-socket"),
+				NoPivot:       context.Bool("no-pivot"),
+				NoNewKeyring:  context.Bool("no-new-keyring"),
+				PreserveFDs:   context.Int("preserve-fds"),
+				Detach:        context.Bool("detach"),
```
Random-Liu: There doesn't seem to be `detach` flag for `runc create`.

crosbymichael: Create detaches by default so this value shouldn't matter either way but Ill remove it for correctness

api/api.go
```diff
@@ -0,0 +1,121 @@
...
+	Delete(id string, opts DeleteOpts) error
+	Kill(id string, sig syscall.Signal, opts KillOpts) error
+	List() ([]Container, error)
+	PS(id string) ([]int, error)
+	Pause(id string) error
+	Resume(id string) error
+	Run(id string, opts CreateOpts) (*CreateResult, error)
+	Start(id string) error
+	State(id string) (*Container, error)
+	Exec(id string, opts ExecOpts) (*CreateResult, error)
```
dqminh: I think `Run` and `Exec` returns CreateResult is not very intuitive, probably `CommandResult` is a better name.

api/api.go
```diff
@@ -0,0 +1,121 @@
...
+
+var (
+	ErrNotImplemented = errors.New("command not implemented")
+	ErrEmptyID        = errors.New("container id cannot be empty")
+)
+
+type ContainerOperations interface {
+	Create(id string, opts CreateOpts) (*CreateResult, error)
+	Delete(id string, opts DeleteOpts) error
+	Kill(id string, sig syscall.Signal, opts KillOpts) error
```
dqminh: not sure if this is intended to be consumed cross-platform. If yes, probably this can be replaced with `os.Signal` instead of current linux-specific `syscall.Signal`

mrunalp:
This looks fine overall. We should probably put it in CI for containerd and cri-o to make sure there are no regressions.

mrunalp:
I will open a PR in cri-o to test this out.

caniszczyk:
ping @opencontainers/runc-maintainers maintainers, this came up on the OCI dev call today (as towards moving to a v1.0 release)

crosbymichael:
I'm closing this.  dont have time to finish and probably there are better ways.

PR #1802, opened 2018-05-11
Update man pages

rst0git:
Hello,
This PR aims to add some missing options in the manual pages.

However, to avoid similar inconsistencies in future, we have to automate the process of generating the markdown files.

Here is a simple shell script which does that:
```bash
#!/bin/bash

# Get a list of all commands from runc --help
commands=$(runc --help | sed -n '/^COMMANDS:$/,/^ *$/p' | awk '{print $1;}' | sed '/^COMMANDS\|help/d')

# generate markdown page for each command
for cmd in $commands;
do
    runc $cmd --help | sed 's/^USAGE:$/SYNOPSIS:/g' | sed 's/^\([A-Z ]\+\):$/# \1/g' > runc-$cmd.8.md
done

# generate markdown page for runc
runc --help | sed 's/^USAGE:$/SYNOPSIS:/g' | sed 's/^\([A-Z ]\+\):$/# \1/g' > runc.8.md
```

### DIFF
--- a/man/runc-resume.8.md
+++ b/man/runc-resume.8.md
@@ -9,4 +9,4 @@ resumed.
 
 # DESCRIPTION
    The resume command resumes all processes in the instance of the container.
-Use runc list to identiy instances of containers and their current status.
+Use runc list to identify instances of containers and their current status.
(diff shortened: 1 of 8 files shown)


man/runc-resume.8.md
```diff
@@ -9,4 +9,4 @@ resumed.
 
 # DESCRIPTION
    The resume command resumes all processes in the instance of the container.
-Use runc list to identiy instances of containers and their current status.
+Use runc list to identity instances of containers and their current status.
```
ehotinger: seems like all instances of this should be `identify`

rst0git: Thanks for catching that!

cyphar:
Can you include the script in the `script/` directory so we can run it regularly (we could also add it to CI to fail if the output doesn't match).

PR #1805, opened 2018-05-22
fix systemd cpu quota for -1

derekwaynecarr:
if a caller passes -1 for cpu quota, no quota setting should be applied at all with systemd.

after #1651, passing -1 caused a side-effect of the value being converted to 10000 incorrectly.

this fixes an issue with the systemd cgroup driver in kubernetes when cfs quota is disabled.

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -5,6 +5,7 @@ package systemd
 import (
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
If the more appropriate change is to have Kubernetes send 0 whether cgroupfs or systemd driver, it would be good to get clarity from maintainers.

derekwaynecarr:
it looks like passing 0 would be consistent across cgroupfs and systemd as any non-zero value results in writing out to cfs_quota_us, and since the default cfs_quota_us is -1, writing nothing would be fine.

but if for example, we ever want to change from quota set to unset on an existing container, passing 0 would not induce a write with the cgroupfs driver, so maybe -1 is the right approach after all.

derekwaynecarr:
Ok, hold on this until I make sure that -1 is consistently applied on create and update of existing cgroups.

derekwaynecarr:
ok, this is ready to go, and ensures that we can set and unset quota.

PTAL @mrunalp @sjenning

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -295,13 +296,19 @@ func (m *Manager) Apply(pid int) error {
...
+		// corresponds to USEC_INFINITY in systemd
+		// if USEC_INFINITY is provided, CPUQuota is left unbound by systemd
+		// always setting a property value ensures we can apply a quota and remove it later
+		cpuQuotaPerSecUSec := uint64(math.MaxUint64)
+		if c.Resources.CpuQuota > 0 {
+			// systemd converts CPUQuotaPerSecUSec (microseconds per CPU second) to CPUQuota
+			// (integer percentage of CPU) internally.  This means that if a fractional percent of
+			// CPU is indicated by Resources.CpuQuota, we need to round up to the nearest
+			// 10ms (1% of a second) such that child cgroups can set the cpu.cfs_quota_us they expect.
+			cpuQuotaPerSecUSec := uint64(c.Resources.CpuQuota*1000000) / c.Resources.CpuPeriod
```
derekwaynecarr: knew i messed something up, this shadows the var

derekwaynecarr:
ok, all set, sorry for spam.

sjenning:
I verified this fix.  With cpu-cfs-quota="true" limits were enforced at the pod and container level.  When set to "false" the quota is not set (-1).

mrunalp:
Pointer to systemd code -- https://github.com/systemd/systemd/blob/master/src/core/cgroup.c#L522

mrunalp:
LGTM

PR #1806, opened 2018-05-25
cgroup: clean up isIgnorableError for skippable EROFS

cyphar:
Include a rootless argument for isIgnorableError to avoid people
accidentally using isIgnorableError when they shouldn't (we don't ignore
any errors when running as root as that really isn't safe).

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -106,11 +106,17 @@ type cgroupData struct {
 // sense of the word). This includes EROFS (which for an unprivileged user is
 // basically a permission error) and EACCES (for similar reasons) as well as
 // the normal EPERM.
-func isIgnorableError(err error) bool {
+func isIgnorableError(rootless bool, err error) bool {
+	// We do not ignore errors if we are root.
+	if !rootless {
+		return false
+	}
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

Thanks~

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -106,11 +106,17 @@ type cgroupData struct {
 // sense of the word). This includes EROFS (which for an unprivileged user is
 // basically a permission error) and EACCES (for similar reasons) as well as
 // the normal EPERM.
-func isIgnorableError(err error) bool {
+func isIgnorableError(rootless bool, err error) bool {
+	// We do not ignore errors if we are root.
+	if !rootless {
```
rhatdan: Should we be smarter and see if you have SYS_ADMIN Caps?

cyphar: That's something that can be handled by having a more clever `isRootless` function earlier on. Not to mention that some of the permission errors would be caused by not having `CAP_DAC_OVERRIDE` -- so `CAP_SYS_ADMIN` checks wouldn't be sufficient.

dqminh:
LGTM

PR #1807, opened 2018-05-25
sd-notify: do not hang when NOTIFY_SOCKET is used with create

giuseppe:
if NOTIFY_SOCKET is used, do not block the main runc process waiting for events on the notify socket.  Bind mount the parent directory of the notify socket, so that "start" can create the socket and it is still accessible from the container.

### DIFF
--- a/notify_socket.go
+++ b/notify_socket.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"path/filepath"
+	"strconv"
+	"time"
 
+	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runtime-spec/specs-go"
-
-	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
@@ -27,12 +29,12 @@ func newNotifySocket(context *cli.Context, notifySocketHost string, id string) *
 	}
 
 	root := filepath.Join(context.GlobalString("root"), id)
-	path := filepath.Join(root, "notify.sock")
+	socketPath := filepath.Join(root, "notify", "notify.sock")
 
 	notifySocket := &notifySocket{
 		socket:     nil,
 		host:       notifySocketHost,
-		socketPath: path,
+		socketPath: socketPath,
 	}
 
 	return notifySocket
@@ -44,13 +46,19 @@ func (s *notifySocket) Close() error {
 
 // If systemd is supporting sd_notify protocol, this function will add support
 // for sd_notify protocol from within the container.
-func (s *notifySocket) setupSpec(context *cli.Context, spec *specs.Spec) {
-	mount := specs.Mount{Destination: s.host, Source: s.socketPath, Options: []string{"bind"}}
+func (s *notifySocket) setupSpec(context *cli.Context, spec *specs.Spec) error {
+	pathInContainer := filepath.Join("/run/notify", path.Base(s.socketPath))
+	mount := specs.Mount{
+		Destination: path.Dir(pathInContainer),
+		Source:      path.Dir(s.socketPath),
+		Options:     []string{"bind", "nosuid", "noexec", "nodev", "ro"},
+	}
 	spec.Mounts = append(spec.Mounts, mount)
-	spec.Process.Env = append(spec.Process.Env, fmt.Sprintf("NOTIFY_SOCKET=%s", s.host))
+	spec.Process.Env = append(spec.Process.Env, fmt.Sprintf("NOTIFY_SOCKET=%s", pathInContainer))
+	return nil
 }
 
-func (s *notifySocket) setupSocket() error {
+func (s *notifySocket) bindSocket() error {
 	addr := net.UnixAddr{
 		Name: s.socketPath,
 		Net:  "unixgram",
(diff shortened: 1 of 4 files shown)


dqminh:
@giuseppe can you describe a minimal test case for this ? Would be nice to setup an integration test too.

With this, do we even need to pass notifySocket to signalHandler anymore ?

giuseppe:
@dqminh currently if the `NOTIFY_SOCKET` environment variable is set for "runc create" then runc just hangs indefinitely.
`signalHandler.forward()` is still the caller for the new process.  Would you prefer to move it somewhere else?

rhatdan:
@giuseppe Couldn't you move the wait for ever functionality out of the run and into the start?

giuseppe:
I cannot do that as the socket needs to be opened by `create` so that it can be bind mounted into the container.  We would need to unlink it before creating it again with `start` (and the container would not see it)

rhatdan:
@dqminh We are waiting for reply from you.
@avagin @caniszczyk @crosbymichael @cyphar @hqhq, @mrunalp @rjnagal @vmarmol   PTAL           

Would love to get podman support for sd-notify so we can run podman/runc with sd-notify containers.

mrunalp:
The cost of the extra process falls on the users of the socket so I am fine with this as there is no simpler way.

mrunalp:
LGTM

rhatdan:
@dqminh @avagin @caniszczyk @crosbymichael @cyphar @hqhq, @rjnagal @vmarmol PTAL

rhatdan:
Could someone @dqminh @avagin @caniszczyk @crosbymichael @cyphar @hqhq, @rjnagal @vmarmol PTAL???

dqminh:
Hi, sorry for the delay !

Having to spawn one more process is unfortunate, but i cant think of a better way if we want to keep the same flow where the forwarding of notification is setup when we first create the container. (However in this case perhaps we should not overload init and instead give the process a separate operation ).

Another way i can think of is to have `Create` just handling the mounting of socket in host to socket in container and setup correct env variable (perhaps set MAINPID ?), but dont actually start handling readiness notification at all. Instead we defer the handling of notifications' forwarding to `Start`. So `Start` will be something like
1. read state of the container if required ( path to notify socket on the host needs to be here somewhere)
2. if notify socket path exists, then start listener
3. notify the container process that it should start executing
4. start forwarding messages

In this case, `NOTIFY_SOCKET=something runc create` will be completed immediately, but `runc start` of such container will also act as the sd_notify forwarder hence will stay alive until such forwarding completed. Basically a change in behavior, but we dont need to execute another long-live process without context attached.

Does that make sense ?

giuseppe:
what stopped me from doing that in the first iteration was that we cannot create the socket once the container is already created (and the socket bind mounted), so that the listener had to be created from create.
I've done an additional change now, where we bind mount the parent directory of the socket, so we can create the socket later.
With this change we lose the ability to keep in the container the same file name specified in the host `NOTIFY_SOCKET`, in the container it is always `/run/notify/notify.sock` but I don't think it really matters for the applications.

signals.go
```diff
@@ -70,7 +70,7 @@ func (h *signalHandler) forward(process *libcontainer.Process, tty *tty, detach
 			h.notifySocket.run(pid1)
 			return 0, nil
 		} else {
-			go h.notifySocket.run(0)
+			h.notifySocket.run(os.Getpid())
```
dqminh: question: does it matter here that we set PID to `runc` process ? Cant we always set it to the container's process instead ?

dqminh:
@giuseppe Thanks ! This looks great and much simpler than before.

signals.go
giuseppe: if we keep runc running, the container's process won't be a direct child of systemd and systemd won't get the SIGCHLD once it terminates.  Differently, it will get the SIGCHLD for the runc process and can link it to the service being terminated.

giuseppe:
does the new version look fine? @dqminh @mrunalp

TomSweeneyRedHat:
Re-ping @dqminh @mrunalp th.  This is part of the fix for a Podman issue https://github.com/containers/libpod/issues/746

giuseppe:
ping, can this be finally reviewed?

crosbymichael:
I don't know enough about this to be able to give a good review or even test.  Someone other maintainers will have to as I wouldn't feel comfortable either way. Sorry.

crawford:
What does this need in order to be merged? I'd love to have sdnotify support and I'm happy to help out with this PR where I can.

mrunalp:
LGTM

mrunalp:
@dqminh Could you take a look?

giuseppe:
any hope this patch can get into the runc release?  We are already using it on runc distributed by Fedora and it solves the issue we were having with systemd.  /cc @cyphar @dqminh @mrunalp

rhatdan:
Yes we would really like to get this in before 1.0.

giuseppe:
rebased :arrow_up:

Could someone please take a look?

mrunalp:
LGTM

rhatdan:
Another month has gone by, Can someone review this PR @cyphar @dqminh @mrunalp

caniszczyk:

giuseppe:
can someone please take a look?

filbranden:
@giuseppe In #2014 I'm adding support for integration tests in an environment where systemd is running and also a Travis-CI job to run those tests on PRs.

Do you think you might be able to add an integration test that would test the NOTIFY_SOCKET feature, perhaps on top of that PR?

I guess having a test case pass would make it easier to get this merged.

Cheers!
Filipe

giuseppe:
@filbranden thanks for the suggestion.  I've looked into it but I don't see any way for using UNIX datagram sockets from a busybox container so that I could communicate back with the systemd socket

filbranden:
@giuseppe Yes busybox will probably not be enough... Can we consider a slightly larger minimal image that would have enough tools (like a somewhat more complete `nc`) to perform the test? Something based on Alpine perhaps?

Or, alternatively, can we consider copying a binary into the unpacked tree of busybox image before running the `runc` test on it?

Let me know if I can help in any way.

PR #2014 has been merged, so rebasing on top of latest head should already get you the tests including systemd and the Travis CI config to run them as part of every pending PR...

Cheers!
Filipe

giuseppe:
yes, using Alpine will help, then I can run the equivalent of:

```
$ systemd-run --unit foo --service-type=notify -p NotifyAccess=all podman run --rm alpine sh -c 'apk add nmap-ncat; echo READY=1 | ncat -u -U $NOTIFY_SOCKET ; sleep 100'
$ systemctl is-active foo
```

giuseppe:
The PR was opened almost 2 years ago.

Since there is no interest in fixing the issue, I am closing it.

saschagrunert:
> The PR was opened almost 2 years ago.
> 
> Since there is no interest in fixing the issue, I am closing it.

Hey @giuseppe, now is the time where we're also interested in fixing this bug. May I ask you to re-open the issue that we can finally merge it? :innocent:

rhatdan:
Yes @giuseppe Let's reopen and see if we can get @AkihiroSuda to review.

AkihiroSuda:
Does crun already have this? Can we have an integration test?

notify_socket.go
```diff
@@ -71,45 +79,89 @@ func (s *notifySocket) setupSocket() error {
...
+	fileChan := make(chan []byte)
+	go func() {
+		for {
+			buf := make([]byte, 512)
+			r, err := n.socket.Read(buf)
+			if err != nil {
+				return
+			}
+			got := buf[0:r]
+			if !bytes.HasPrefix(got, []byte("READY=")) {
```
AkihiroSuda: We need to compare the bytes per line, not per fixed 512 bytes?

giuseppe: there is just one line in the message, so we should not split the lines

giuseppe:
> Does crun already have this? Can we have an integration test?

yes crun already supports it,.  As well as the runc distributed by Fedora/RHEL, where we are backporting this patch.

notify_socket.go
AkihiroSuda: Is there a spec document for that?

giuseppe: I've looked at what systemd-nspawn does and yes it indeed splits by line (it also supports `STATUS` which is not currently done here).  I've adapted the implementation.

Thanks for pointing it out.

notify_socket.go
```diff
@@ -71,46 +79,90 @@ func (s *notifySocket) setupSocket() error {
...
-					client.Write([]byte(newPid))
-				}
-				return
+		}
+	}()
+
+	for {
+		select {
+		case <-ticker.C:
+			_, err := os.Stat(filepath.Join("/proc", strconv.Itoa(pid1)))
```
AkihiroSuda: Can't we use waitpid? Or pidfd for recent kernel?

notify_socket.go
```diff
@@ -71,46 +79,90 @@ func (s *notifySocket) setupSocket() error {
...
+	fileChan := make(chan []byte)
+	go func() {
+		for {
+			buf := make([]byte, 4096)
+			r, err := n.socket.Read(buf)
+			if err != nil {
+				return
+			}
+			got := buf[0:r]
+			for _, line := range bytes.Split(got, []byte{'\n'}) {
```
AkihiroSuda: needs to handle a line that spans multiple `[4096]byte` packets

notify_socket.go
giuseppe: The process is not a child of the "runc start" session.  Pidfd would work but we still need this for older kernels

notify_socket.go
giuseppe: that is not needed for systemd notify.

From the `sd_notify(3)` man page:

```
These functions send a single datagram with the state string as payload to the
`AF_UNIX` socket referenced in the `$NOTIFY_SOCKET` environment variable
```

AkihiroSuda: Thanks, could you add a code comment for that

giuseppe: sure, done in the last revision.

AkihiroSuda:
LGTM, but I'd like to see integration test suite if possible. Can be follow-up PR if difficult.

mrunalp:
LGTM

Issue #1715, opened 2018-02-08
runc does not implement runtime-spec's 'creating' status


wking:
In opencontainers/runtime-tools#557, @liangchenye [pointed out][1] that runc doesn't currently support [the spec's `creating` status][2].  #1704 is in-flight as part of status cleanup, and I was planning to work up a `creating` PR after #1704 settled.  I'm filing this placeholder issue in the meantime, in case the runc maintainers want to add it to [the 1.0.0 milestone][3].  Or maybe someone else wants to work up the `creating` PR for me ;).

While working on this, we may also want to address `runc create`'s apparently early exit.  More information on that in opencontainers/runtime-tools#563.

[1]: https://github.com/opencontainers/runtime-tools/pull/557#discussion_r163762617
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L19
[3]: https://github.com/opencontainers/runc/milestone/4

PR #1811, opened 2018-05-30
libcontainer: call Prestart, Poststart hooks from correct places

dongsupark:
So far Prestart, Poststart hooks have been called from the context of create, which does not satisfy the runtime spec. That's why the runtime-tools validation tests like `hooks_stdin.t` have failed. Let's move call sites of Prestart, Poststart to correct places.

Unfortunately as for the Poststart hook, in practice it's not possible to tell whether a specific call site is from create context or run context. That's why we needed to allow Create and Run methods to accept another parameter `action` (of type `CtAct`). Doing that, it's possible to set a variable `initProcess.IsCreate` that allows us distinguish Create from Run. That's why the first commit `libcontainer: move CtAct to libcontainer` is needed.

I tested runtime-tools' `validation/hooks_stdin.t` with this PR of runc, and it worked fine. Though if there would be unexpected breakage from changing like this, please let me know.

It depends on a pending PR https://github.com/opencontainers/runc/pull/1741.
See also https://github.com/opencontainers/runc/issues/1710.

/cc @wking @liangchenye

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -230,7 +230,7 @@ func (c *linuxContainer) Set(config configs.Config) error {
 	return err
 }
 
-func (c *linuxContainer) Start(process *Process) error {
+func (c *linuxContainer) Start(process *Process, action CtAct) error {
 	c.m.Lock()
 	defer c.m.Unlock()
 	if process.Init {
@@ -238,7 +238,7 @@ func (c *linuxContainer) Start(process *Process) error {
 			return err
 		}
 	}
-	if err := c.start(process); err != nil {
+	if err := c.start(process, action == CT_ACT_CREATE); err != nil {
 		if process.Init {
 			c.deleteExecFifo()
 		}
@@ -247,8 +247,8 @@ func (c *linuxContainer) Start(process *Process) error {
 	return nil
 }
 
-func (c *linuxContainer) Run(process *Process) error {
-	if err := c.Start(process); err != nil {
+func (c *linuxContainer) Run(process *Process, action CtAct) error {
+	if err := c.Start(process, action); err != nil {
 		return err
 	}
 	if process.Init {
@@ -332,8 +332,8 @@ type openResult struct {
 	err  error
 }
 
-func (c *linuxContainer) start(process *Process) error {
-	parent, err := c.newParentProcess(process)
+func (c *linuxContainer) start(process *Process, isCreate bool) error {
+	parent, err := c.newParentProcess(process, isCreate)
 	if err != nil {
 		return newSystemErrorWithCause(err, "creating new parent process")
 	}
@@ -438,7 +438,7 @@ func (c *linuxContainer) includeExecFifo(cmd *exec.Cmd) error {
 	return nil
 }
 
-func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
+func (c *linuxContainer) newParentProcess(p *Process, isCreate bool) (parentProcess, error) {
 	parentInitPipe, childInitPipe, err := utils.NewSockPair("init")
 	if err != nil {
 		return nil, newSystemErrorWithCause(err, "creating new init pipe")
@@ -467,7 +467,7 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 	if err := c.includeExecFifo(cmd); err != nil {
 		return nil, newSystemErrorWithCause(err, "including execfifo in cmd.Exec setup")
 	}
-	return c.newInitProcess(p, cmd, messageSockPair, logFilePair)
+	return c.newInitProcess(p, cmd, messageSockPair, logFilePair, isCreate)
 }
 
 func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, childLogPipe *os.File) (*exec.Cmd, error) {
@@ -509,7 +509,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, chi
 	return cmd, nil
 }
 
-func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, messageSockPair, logFilePair filePair) (*initProcess, error) {
+func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, messageSockPair, logFilePair filePair, isCreate bool) (*initProcess, error) {
 	cmd.Env = append(cmd.Env, "_LIBCONTAINER_INITTYPE="+string(initStandard))
 	nsMaps := make(map[configs.NamespaceType]string)
 	for _, ns := range c.config.Namespaces {
@@ -533,6 +533,7 @@ func (c *linuxContainer) newInitProcess(p *Process, cmd *exec.Cmd, messageSockPa
 		process:         p,
 		bootstrapData:   data,
 		sharePidns:      sharePidns,
+		IsCreate:        isCreate,
 	}
 	c.initProcess = init
 	return init, nil
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -263,8 +263,24 @@ func (c *linuxContainer) Exec() error {
 	return c.exec()
 }
 
-func (c *linuxContainer) exec() error {
+func (c *linuxContainer) exec() (errExec error) {
 	path := filepath.Join(c.root, execFifoFilename)
+
+	if c.config.Hooks != nil && len(c.config.Hooks.Poststart) > 0 {
+		defer func() {
+			s, err := c.currentOCIState()
+			if err != nil {
+				errExec = err
+				return
+			}
+			for i, hook := range c.config.Hooks.Poststart {
+				if err := hook.Run(s); err != nil {
+					errExec = newSystemErrorWithCausef(err, "running poststart hook %d", i)
+					return
+				}
+			}
+		}()
+	}
 
 	fifoOpen := make(chan struct{})
 	select {
@@ -362,12 +378,9 @@ func (c *linuxContainer) start(process *Process, isCreate bool) error {
 			if err != nil {
 				return err
 			}
-			for i, hook := range c.config.Hooks.Poststart {
+			for i, hook := range c.config.Hooks.Prestart {
 				if err := hook.Run(s); err != nil {
-					if err := ignoreTerminateErrors(parent.terminate()); err != nil {
-						logrus.Warn(err)
-					}
-					return newSystemErrorWithCausef(err, "running poststart hook %d", i)
+					return newSystemErrorWithCausef(err, "running prestart hook %d", i)
 				}
 			}
 		}
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -370,6 +370,19 @@ func (p *initProcess) start() error {
 		sentResume bool
 	)
 
+	// runc start
+	if !p.IsCreate && p.config.Config.Hooks != nil {
+		s, err := p.container.currentOCIState()
+		if err != nil {
+			return err
+		}
+		for i, hook := range p.config.Config.Hooks.Poststart {
+			if err := hook.Run(s); err != nil {
+				return newSystemErrorWithCausef(err, "running poststart hook %d", i)
+			}
+		}
+	}
+
 	ierr := parseSync(p.messageSockPair.parent, func(sync *syncT) error {
 		switch sync.Type {
 		case procReady:
@@ -378,30 +391,13 @@ func (p *initProcess) start() error {
 			if err := setupRlimits(p.config.Rlimits, p.pid()); err != nil {
 				return newSystemErrorWithCause(err, "setting rlimits for ready process")
 			}
-			// call prestart hooks
 			if !p.config.Config.Namespaces.Contains(configs.NEWNS) {
-				// Setup cgroup before prestart hook, so that the prestart hook could apply cgroup permissions.
 				if err := p.manager.Set(p.config.Config); err != nil {
 					return newSystemErrorWithCause(err, "setting cgroup config for ready process")
 				}
 				if p.intelRdtManager != nil {
 					if err := p.intelRdtManager.Set(p.config.Config); err != nil {
 						return newSystemErrorWithCause(err, "setting Intel RDT config for ready process")
-					}
-				}
-
-				if p.config.Config.Hooks != nil {
-					s, err := p.container.currentOCIState()
-					if err != nil {
-						return err
-					}
-					// initProcessStartTime hasn't been set yet.
-					s.Pid = p.cmd.Process.Pid
-					s.Status = "creating"
-					for i, hook := range p.config.Config.Hooks.Prestart {
-						if err := hook.Run(s); err != nil {
-							return newSystemErrorWithCausef(err, "running prestart hook %d", i)
-						}
 					}
 				}
 			}
@@ -411,27 +407,12 @@ func (p *initProcess) start() error {
 			}
 			sentRun = true
 		case procHooks:
-			// Setup cgroup before prestart hook, so that the prestart hook could apply cgroup permissions.
 			if err := p.manager.Set(p.config.Config); err != nil {
 				return newSystemErrorWithCause(err, "setting cgroup config for procHooks process")
 			}
 			if p.intelRdtManager != nil {
 				if err := p.intelRdtManager.Set(p.config.Config); err != nil {
 					return newSystemErrorWithCause(err, "setting Intel RDT config for procHooks process")
-				}
-			}
-			if p.config.Config.Hooks != nil {
-				s, err := p.container.currentOCIState()
-				if err != nil {
-					return err
-				}
-				// initProcessStartTime hasn't been set yet.
-				s.Pid = p.cmd.Process.Pid
-				s.Status = "creating"
-				for i, hook := range p.config.Config.Hooks.Prestart {
-					if err := hook.Run(s); err != nil {
-						return newSystemErrorWithCausef(err, "running prestart hook %d", i)
-					}
 				}
 			}
 			// Sync with child.
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -231,6 +231,7 @@ type initProcess struct {
 	process         *Process
 	bootstrapData   io.Reader
 	sharePidns      bool
+	IsCreate        bool
 }
 
 func (p *initProcess) pid() int {
--- a/create.go
+++ b/create.go
@@ -3,6 +3,7 @@ package main
 import (
 	"os"
 
+	"github.com/opencontainers/runc/libcontainer"
 	"github.com/urfave/cli"
 )
 
@@ -62,7 +63,7 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 		if err != nil {
 			return err
 		}
-		status, err := startContainer(context, spec, CT_ACT_CREATE, nil)
+		status, err := startContainer(context, spec, libcontainer.CT_ACT_CREATE, nil)
 		if err != nil {
 			return err
 		}
--- a/exec.go
+++ b/exec.go
@@ -149,7 +149,7 @@ func execProcess(context *cli.Context) (int, error) {
 		consoleSocket:   context.String("console-socket"),
 		detach:          detach,
 		pidFile:         context.String("pid-file"),
-		action:          CT_ACT_RUN,
+		action:          libcontainer.CT_ACT_RUN,
 		init:            false,
 		preserveFDs:     context.Int("preserve-fds"),
 		logLevel:        logLevel,
--- a/libcontainer/container.go
+++ b/libcontainer/container.go
@@ -14,6 +14,13 @@ import (
 
 // Status is the status of a container.
 type Status int
+type CtAct uint8
+
+const (
+	CT_ACT_CREATE CtAct = iota + 1
+	CT_ACT_RUN
+	CT_ACT_RESTORE
+)
 
 const (
 	// Created is the status that denotes the container exists but has not been run yet.
@@ -128,7 +135,7 @@ type BaseContainer interface {
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
 	// SystemError - System error.
-	Start(process *Process) (err error)
+	Start(process *Process, action CtAct) (err error)
 
 	// Run immediately starts the process inside the container.  Returns error if process
 	// fails to start.  It does not block waiting for the exec fifo  after start returns but
@@ -139,7 +146,7 @@ type BaseContainer interface {
 	// ConfigInvalid - config is invalid,
 	// ContainerPaused - Container is paused,
 	// SystemError - System error.
-	Run(process *Process) (err error)
+	Run(process *Process, action CtAct) (err error)
 
 	// Destroys the container, if its in a valid state, after killing any
 	// remaining running processes.
--- a/libcontainer/integration/checkpoint_test.go
+++ b/libcontainer/integration/checkpoint_test.go
@@ -113,7 +113,7 @@ func testCheckpoint(t *testing.T, userns bool) {
 		Init:   true,
 	}
 
-	err = container.Run(&pconfig)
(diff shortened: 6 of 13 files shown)


libcontainer/container_linux.go
```diff
@@ -276,6 +276,21 @@ func (c *linuxContainer) Exec() error {
 func (c *linuxContainer) exec() error {
 	path := filepath.Join(c.root, execFifoFilename)
 
+	defer func() {
+		if c.config.Hooks != nil && len(c.config.Hooks.Poststart) > 0 {
```
wking: I don't know how cheap goroutines are, but you may want:

```go
if c.config.Hooks != nil && len(c.config.Hooks.Poststart) > 0 {
  defer func(0 {
    ...
  }()
}
```

To avoid spawning one when there are no hooks.

libcontainer/container_linux.go
```diff
@@ -276,6 +276,21 @@ func (c *linuxContainer) Exec() error {
...
+			s, err := c.currentOCIState()
+			if err != nil {
+				return
+			}
+			for i, hook := range c.config.Hooks.Poststart {
+				if err := hook.Run(s); err != nil {
+					fmt.Printf("%v", newSystemErrorWithCausef(err, "running poststart hook %d", i))
```
wking: Printing (to stdout?) is not a great API for callers.  Can we collect these and return any failed-hook [warnings][1] to the caller?

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L63

libcontainer/container_linux.go
```diff
@@ -366,17 +381,15 @@ func (c *linuxContainer) start(process *Process, isInit, isCreate bool) error {
...
 		c.initProcessStartTime = state.InitProcessStartTime
 
-		if c.config.Hooks != nil && len(c.config.Hooks.Poststart) > 0 {
+		if c.config.Hooks != nil && len(c.config.Hooks.Prestart) > 0 {
 			s, err := c.currentOCIState()
 			if err != nil {
 				return err
 			}
-			for i, hook := range c.config.Hooks.Poststart {
+			s.Status = "created"
```
wking: Do we still need this clobber now that these are called in the right place and no longer need to work around #1715?

libcontainer/container_linux.go
dongsupark: @wking Right. I'll make it return warnings to the caller.

libcontainer/container_linux.go
dongsupark: @wking Yes, it makes sense. Will remove this line.
Thanks for pointing to #1715, which looks like another can of worms. ;-)

dongsupark:
Rebased.
Added another commit to fix build errors under libcontainer/integration.

Addressed review comments.
* moved the defer function to inside of the if clause.
* made the defer function return warnings to the caller of exec().
* removed unnecessary setting of status.

flx42:
Dusting this off a little bit. The situation of hooks in the OCI runtime spec and in `runc` still seems messy today.

@dongsupark with your change, this comment becomes false:
https://github.com/opencontainers/runc/blob/70ca035aa6ecfc496e13365fdef20383408501ba/libcontainer/rootfs_linux.go#L80-L87
It means no OCI hook would be able to do runtime -> container mounts since the switch to the new root is done as part of `create`, before any hook is called (according to the spec). We rely on this feature to enable GPU support as a pre-start hook today.

wking:
> We rely on this feature to enable GPU support as a pre-start hook today.

Can't you stick to behavior laid out in the runtime spec and handle this sort of thing with mount propagation?  For example, mount a shared directory into your container, and then use your hooks to fill content into that shared directory.

wking:
> That forces you to shove everything inside a single directory.

You can mount multiple directories.  And you may be able to make bind chains like:

1. Bind host `/dev/a/b/c` to `/some/share/a/b/c` in the host mount namespace.
2. Bind `/share/a/b/c` to `/dev/a/b/c` in the container mount namespace.

I haven't actually tested that, maybe it doesn't work.

flx42:
Our CUDA stack expects devices to be in `/dev`, not in subdirectories.

flx42:
Also, it also mean the OCI spec must have been modified with a bind-mount with shared propagation. Thus, instead of one prestart hook to enable GPU support for everyone, we need everyone to modify their OCI spec in order to get our prestart hook to work.

wking:
> Our CUDA stack expects devices to be in `/dev`, not in subdirectories.

That doesn't seem like a critical issue, just adjust the paths in my example to match whatever you need.

> Also, it also mean the OCI spec must have been modified with a bind-mount with shared propagation. Thus, instead of one prestart hook to enable GPU support for everyone, we need everyone to modify their OCI spec in order to get our prestart hook to work.

Agreed.  But this seems like it would be straightforward enough to handle at a higher level, with a config similar to [what libpod uses for hooks][1].  And an approach like this would avoid having to update the spec to say "runtimes which implement containerization with a pivot, chroot or similar have to run the pre-start hooks somewhere inside the `create` workflow when ${THESE_THINGS} have happened but ${THESE_OTHER_THINGS} have not.  Runtimes based on VMs and other things are off the hook".  That sounds hard to specify clearly ;).

[1]: https://github.com/containers/libpod/blob/v0.9.1.1/pkg/hooks/docs/oci-hooks.5.md

flx42:
> That doesn't seem like a critical issue, just adjust the paths in my example to match whatever you need.

I think it is, unless `/dev` in the container is a bind-mount with shared propagation, how can I make `/dev/nvidia` appear inside the container without a subdirectory? 

> Agreed. But this seems like it would be straightforward enough to handle at a higher level, with a config similar to what libpod uses for hooks.

Possible, but then we would better standardize this higher level construct. We don't want to have to integrate in each different container runtime. In other words, CNI but for devices :)

> "runtimes which implement containerization with a pivot, chroot or similar have to run the pre-start hooks somewhere inside the create workflow when ${THESE_THINGS} have happened but ${THESE_OTHER_THINGS} have not. Runtimes based on VMs and other things are off the hook".

I think the spec is vague here. @alban mentioned to me that `create` should do the `pivot/chroot` per the following wording in the spec:
> All of the properties configured in config.json except for process MUST be applied

From https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#create
But does `configured` means for the [`Root`](https://github.com/opencontainers/runtime-spec/blob/master/config.md#root) struct? Does that mean [setting up the mount-point](https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L652)? Or does that mean `pivot_root`/`chroot`?

alban:
> It means no OCI hook would be able to do runtime's namespace -> container mounts since the switch to the new root is done as part of create, before any hook is called (according to the spec). We rely on this feature to enable GPU support as a pre-start hook today.

If the rootfs is configured as propagation mode "slave"/"rslave", the OCI hook, running in the runtime's namespace could add bind mounts from the host `/dev/foo` into the `$ROOTFS_CONTAINER/my-dev-foo` and it will be propagated into `/my-dev-foo` in the container mount namespace.

As you noted in https://github.com/opencontainers/runtime-spec/issues/973#issuecomment-405644551, it seems to be the default in runc. systemd-nspawn also uses "rslave".

dongsupark:
@flx42 Thanks for comments. I almost forgot about this PR. ;-)
Like @wking @alban mentioned, I suppose you can probably figure out how to deal with your issue via mount propagation.

BTW I'm still curious, what happened to the other PR, https://github.com/opencontainers/runc/pull/1741, on which this PR depends on.
I also have not heard of any feedback from maintainers, whether I'm doing the right thing with this PR or not.

cyphar:
I think this might be needed for 1.0 purely because it's a pretty big spec-related change. However I know that the cri-o folks use this already so maybe we should be careful about changing it to be spec-compliant.

/cc @opencontainers/runc-maintainers @mrunalp @rhatdan

cyphar:
LGTM, though I would like input from the cri-o folks.

mrunalp:
I'll review and get back on this tomorrow.

cyphar:
Sure @mrunalp. This is the final patch I'm going to be pushing for in 1.0. As soon as this is merged I will prepare the 1.0 branch and send out the vote email.

cyphar:
LGTM.

libcontainer/process_linux.go
```diff
@@ -364,21 +379,6 @@ func (p *initProcess) start() error {
 						return newSystemErrorWithCause(err, "setting Intel RDT config for ready process")
 					}
 				}
-
-				if p.config.Config.Hooks != nil && len(p.config.Config.Hooks.Prestart) > 0 {
```
wking: Removing the pre-start hooks here makes the earlier `// Setup cgroup before prestart hook...` and `// call prestart hooks` comments obsolete.  Now that the pre-start hooks are firing before the start signal (after all creation has finished), maybe we can just drop those comments?

libcontainer/process_linux.go
```diff
@@ -395,20 +395,6 @@ func (p *initProcess) start() error {
 					return newSystemErrorWithCause(err, "setting Intel RDT config for procHooks process")
 				}
 			}
-			if p.config.Config.Hooks != nil && len(p.config.Config.Hooks.Prestart) > 0 {
```
wking: Here again, this removal obsoletes some of the preceding comments.  And maybe `procHooks` as a whole can go away now?

libcontainer/process_linux.go
```diff
@@ -345,6 +345,21 @@ func (p *initProcess) start() error {
 		sentResume bool
 	)
 
+	// runc start
+	if !p.IsCreate {
+		if p.config.Config.Hooks != nil && len(p.config.Config.Hooks.Poststart) > 0 {
```
wking: nit: you can flatten these two to remove one level of nesting:

```go
if !p.IsCreate && p.config.Config.Hooks != nil && len(p.config.Config.Hooks.Poststart) > 0 {
```

libcontainer/process_linux.go
cyphar: Reading the PR again, the changes have stopped making sense. I think I'm going to have to properly carry this PR...

wking: > I think I'm going to have to properly carry this PR...

Any chance we can just land #1741?  I think we all agree on that part of this anyway.

cyphar: I mean, we can land that too (this PR includes it as the first commit). But my reason for wanting this in 1.0 is because currently `runc` doesn't correctly implement hooks -- and that's a bit of an issue if we want to be at least mostly spec compliant.

wking: > I mean, we can land that too (this PR includes it as the first commit). But my reason for wanting this in 1.0 is because currently runc doesn't correctly implement hooks -- and that's a bit of an issue if we want to be at least mostly spec compliant.

Right.  I just don't think we have to jump straight to compliant in a single PR.  #1741 moves us a tiny bit closer, and doesn't seem to break anything, so lets land it.  Then you or @dongsupark can pick up the timing fix from this PR and get that landed after you've worked it out.

cyphar: My main reasoning is that time-to-merge for time-critical PRs is usually the main blocker due to timezones. The plan for runc 1.0 is that I send out the release notification this week, and I am at KiwiCon from Thursday on.

But if you reopen the PR and rebase it now I will LGTM it.

wking: > But if you reopen the PR and rebase it now I will LGTM it.

I cannot re-open on my own (I need someone with write access to the repo to do that).  Once you re-open, I'll rebase (does it even need a rebase?  When I rebase locally I see no conflicts).

dongsupark: Either way I'm fine.

If https://github.com/opencontainers/runc/pull/1741 gets first merged, then I can rebase this PR #1811 on top of that. Also I can fix other issues with obsolete comments.

Or if you want to properly carry this PR with the other one altogether, then you can do so. Go ahead. :-)

cyphar:
I think it's necessary to rewrite some pretty large parts of this (in particular, `procHooks` needs to be removed and a few other such synchronisation bits needs to be cleaned up significantly). I will try to reuse some of these patches but this particular patchset is pretty hard to keep track of (I'v never really been a fan of `process_linux.go` and `container_linux.go` -- they're really hard to understand).

cyphar:
I will play with it today and if I can't get it working with a cleaner set of changes, we can just use this as a stop-gap. Don't get me wrong, I **really** appreciate you working on this and I'm sorry that discussion and merging of this was delayed for so long.

RenaudWasTaken:
Hello!

As my colleague @flx42 mentioned, We rely on this feature to enable GPU support and this patch **will** break our users (nvidia-docker, CRIO, containerd).

The issue we are facing is that we rely on the ability to mount a number of things from the runtime's namespace to the container's namespace:
- [libraries](https://github.com/NVIDIA/libnvidia-container/blob/56704b8dd297bf4daf82a2da4b270dc7f14e0008/src/nvc_info.c#L67-L111)
- [binaries](https://github.com/NVIDIA/libnvidia-container/blob/56704b8dd297bf4daf82a2da4b270dc7f14e0008/src/nvc_info.c#L53-L65)
- [procfs files](https://github.com/NVIDIA/libnvidia-container/blob/deccb2801502675bd283c6936861814dbca99ecd/src/nvc_mount.c#L232)
- [ipc](https://github.com/NVIDIA/libnvidia-container/blob/deccb2801502675bd283c6936861814dbca99ecd/src/nvc_mount.c#L131)

@wking suggested a solution for [devices above](https://github.com/opencontainers/runc/pull/1811#issuecomment-420451147) however, that would break all our users today.
We would also have no real way to cleanup what's on the host, since unmount doesn't propagate.

Is there a solution here that wouldn't break us?

wking:
> @wking suggested a solution for devices above however, that would break all our users today

Wouldn't injecting resources via bind mounts be identical for your users vs. your current approach?  Of course, you'd need to deploy updated hooks before deploying updated runcs, but that seems doable.  Am I missing something?

wking:
> We would also have no real way to cleanup what's on the host, since unmount doesn't propagate.

Yeah, and I don't see a way around this part :/.  I still think you should work with orchestrators to standardize a way to register pre-create [config manipulators][1] so you could inject all of these into `mounts` instead of mounting (or whatever) yourself.  But that's obviously not something that will get done this week.

[1]: https://github.com/opencontainers/runc/pull/1811#issuecomment-420457808

RenaudWasTaken:
> Wouldn't injecting resources via bind mounts be identical for your users
vs. your current approach? Of course, you'd need to deploy updated hooks
before deploying updated runcs, but that seems doable. Am I missing
something?

Wouldn't your solution require the Oci specs to have been modified with a
bind-mount with shared propagation? In that case it would break all
existing images...

wking:
> Wouldn't your solution require the Oci specs to have been modified with a bind-mount with shared propagation?

You're just pushing content in, so slave should work fine, right?  @alban [mentioned][1] rslave being the default for runc and systemd.  So if you can find a bind-mounted directory, you can probably use it for the bind-mount-chain hack.  It would be much more straightforward to get these injected in the runtime config's `mounts` or `devices` though.

As far as the spec is concerned, 1.0.1 (the latest release) doesn't really [say anything about bind mounts][2].  Things have improved a bit since, we now [mention (r)bind][3], but even the master spec only talks about propagation for [the root filesystem][4].

[1]: https://github.com/opencontainers/runc/pull/1811#issuecomment-420567145
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#mounts
[3]: https://github.com/opencontainers/runtime-spec/blob/1722abf79c2f8f2675f47367f827c6491472cf27/config.md#mounts
[4]: https://github.com/opencontainers/runtime-spec/blob/1722abf79c2f8f2675f47367f827c6491472cf27/config-linux.md#rootfs-mount-propagation

RenaudWasTaken:
> You're just pushing content in, so slave should work fine, right? @alban mentioned rslave being the default for runc and systemd. So if you can find a bind-mounted directory, you can probably use it for the bind-mount-chain hack

Wouldn't hijacking a random bind mounted directory end up breaking some consumers of that directory?

> It would be much more straightforward to get these injected in the runtime config's mounts or devices though.

Unfortunately the hook mechanism doesn't allow that (because it's too late) and they can't be contained in the docker image spec.

wking:
> Wouldn't hijacking a random bind mounted directory end up breaking some consumers of that directory?

It's not pretty, but you can pick names that don't collide.  And maybe unbind the intermediates in your hook (if that works, it may mitigate cleanup concerns).

> Unfortunately the hook mechanism doesn't allow that (because it's too late)...

*Runtime hooks* are too late.  Runtime-config manipulators seem like the long-tetm fix to me, but they're a [negotiate with orchestrators][1] idea that doesn't exist yet.

[1]: https://github.com/opencontainers/runc/pull/1811#issuecomment-439257276

wking:
> The point is that OCI should probably have more hooks so that folks like NVIDIA can get more hook points while still being able to obey the intention of the OCI spec...

I don't think we need runtime-spec additions for this use-case.  Couldn't the NVIDIA case be addressed with [`mounts`][1] like:

```
"mounts": [
  {
    "destination": "/usr/lib/mylib.so",
    "source": "/usr/lib/mylib.so",
    "options": ["bind"]
  },
  ...
],
```

and [`devices`][2] like:

```
"devices": [
  {
    "path": "/dev/mydev",
    "type": "c",
    "major": 123,
    "minor": 456,
    "fileMode": 438,
    "uid": 0,
    "gid": 0
  }, 
  ...
]
```

The difference is that existing orchestrators provide registries for external commands to add entries to `hooks` when they build the runtime config, but they do not provide registries to add entries to `mounts` or `devices`.  Registries like that seem like they'd be easy to add (just like the current hook registries but with more generic JSON-property targets?).  They just don't exist now.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#mounts
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-linux.md#devices

cyphar:
@wking 

> I don't think we need runtime-spec additions for this use-case.

We do need more hooks if folks need to hook into creation.

> Couldn't the NVIDIA case be addressed with mounts [...] and devices like: [...]

Maybe, but that would require their hooks to be able to modify the `config.json` which isn't really possible at the moment. Please remember that the use-case we have here is plugging into arbitrary OCI runtimes -- and in the case of `cri-o` and `podman` this is done through various annotations to add these hooks to the OCI spec. Requiring each OCI runtime user to also add NVIDIA-specific code to generate the correct spec is missing the point of having hooks...

> but they do not provide registries to add entries to mounts or devices

I'm not sure I see what the benefit is in doing it this way. Sure, it means that the OCI runtime would be in charge of creating the mounts and devices, but I think that's actually a downside. I always understood that the idea of hooks is to allow them to modify the container state in specific ways outside of the OCI spec -- so why require all OCI callers to now add more code to handle this usecase (which will *only* fix the devices and mounts problems -- who knows what other folks might want to do in hooks) when we could just add more hooks to service everyone.

wking:
> We do need more hooks if folks need to hook into creation.

We need *orchestrator* hooks (somewhere in the image-to-runtime-`create`-call conversion pipeline); we don't need new *runtime* hooks.  I've put together a proof-of-concept for the config-manipulation approach [here][1] in case it helps to have something concrete to look at (also linked from [the mailing list][2]).  I'll shut up here now; if anyone's interested in that approach, maybe we can continue it in the mailing list sub-thread?

[1]: https://github.com/wking/libpod/tree/config-filters
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/pWEp_UK6aAk/6_nHAypqAQAJ

dongsupark:
Rebased, and addressed review comments about comments and unnecessary if conditions.

BTW I'm fine even if it won't be merged for the upcoming release.

AkihiroSuda:
needs rebase

AkihiroSuda:
can we have integration tests?

dongsupark:
@AkihiroSuda Thanks, Akihiro.
Sure, I can rebase it again.
I will also try to add integration tests probably in this week.

Though I have not been following recent discussions regarding the potential regressions that Nvidia folks mentioned.
Have we somehow reached a consensus?

h-vetinari:
> Though I have not been following recent discussions regarding the potential regressions that Nvidia folks mentioned.
Have we somehow reached a consensus?

AFAICT, @RenaudWasTaken from Nvidia has been pinging for feedback in opencontainers/runtime-spec#1008, but there hasn't been any (publically visible) progress since two months.

@cyphar @vbatts @alban

dongsupark:
@cyphar 
As discussed, we should somehow review and merge https://github.com/opencontainers/runtime-spec/pull/1008.

kad:
Is this patchset up to date? 


```
root@nuc1/mycontainer # ls -l /tmp/hook.log
ls: cannot access '/tmp/hook.log': No such file or directory
root@nuc1/mycontainer # ./runc-pr1811 create myid
root@nuc1/mycontainer # ls -l /tmp/hook.log
-rw-r--r-- 1 root root 141 Oct 23 23:19 /tmp/hook.log
root@nuc1/mycontainer # ./runc-pr1811 state myid
{
  "ociVersion": "1.0.1-dev",
  "id": "myid",
  "pid": 93793,
  "status": "created",
  "bundle": "/mycontainer",
  "rootfs": "/mycontainer/rootfs",
  "created": "2019-10-23T20:19:46.68684718Z",
  "owner": ""
}
root@nuc1/mycontainer # rm /tmp/hook.log
root@nuc1/mycontainer # ./runc-pr1811 start myid
root@nuc1/mycontainer # ./runc-pr1811 state myid
{
  "ociVersion": "1.0.1-dev",
  "id": "myid",
  "pid": 93793,
  "status": "running",
  "bundle": "/mycontainer",
  "rootfs": "/mycontainer/rootfs",
  "created": "2019-10-23T20:19:46.68684718Z",
  "owner": ""
}
root@nuc1/mycontainer # ls -l /tmp/hook.log
ls: cannot access '/tmp/hook.log': No such file or directory
root@nuc1/mycontainer # ./runc-pr1811 delete myid
root@nuc1/mycontainer # ./runc-pr1811 -v
runc version 1.0.0-rc9+dev
commit: e27dc4fbc2395fd30fddefcbcc8192f13eec5385
spec: 1.0.1-dev
root@nuc1/mycontainer #
```

Source code used: mater+this PR:
```
kad@nuc1~/go/src/github.com/opencontainers/runc $ git describe
v1.0.0-rc9-17-ge27dc4fb
kad@nuc1~/go/src/github.com/opencontainers/runc $ git log --oneline origin/master..
e27dc4fb (HEAD -> hooks) libcontainer: call Prestart, Poststart hooks from correct places
15708e5d libcontainer/integration: update call sites of container.Run()
a8dcfb98 libcontainer: move CtAct to libcontainer
kad@nuc1~/go/src/github.com/opencontainers/runc $
```

vbatts:
@RenaudWasTaken @flx42 @cyphar 
This topic came up on the call today. The runtime-spec conversation has merged https://github.com/opencontainers/runtime-spec/pull/1008, and we'll have a spec version shortly (like 1.0.2).
Can we rebase this and push it through?

cyphar:
I can carry it if @dongsupark is tired of constantly rebasing this PR. :P

RenaudWasTaken:
I was waiting for a new version of the runtime spec before pushing a PR to runc (at least the create-runtime hook is trivial).
@cyphar since prestart hooks are deprecated now do we want to change them?

vbatts:
How should we deprecate the old ones?

cyphar:
@RenaudWasTaken The prestart hooks should keep their old behaviour so old NVIDIA hooks (from before the merge to the newer hooks) don't break on newer runc versions. We should keep them in runc until they're dropped from the spec (which isn't going to happen any time soon).

RenaudWasTaken:
> The prestart hooks should keep their old behaviour so old NVIDIA hooks (from before the merge to the newer hooks) don't break on newer runc versions

In that case I think this PR can be closed, I can probably open PRs to add the other hooks so that you can review them and we will rebase them once the spec has a new release.

cyphar:
Yeah, I'll close in favour of newer PRs being opened that implement the now-in-the-spec semantics.

Thanks for all your hard work on this @dongsupark -- I know it's taken us a long time to solve this problem, but hopefully it'll all be over quite soon. :smile_cat:

PR #1812, opened 2018-06-01
Fix race in runc exec

mrunalp:
There is a race in runc exec when the init process stops just before
the check for the container status. It is then wrongly assumed that
we are trying to start an init process instead of an exec process.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -140,6 +140,7 @@ func execProcess(context *cli.Context) (int, error) {
 		detach:          detach,
 		pidFile:         context.String("pid-file"),
 		action:          CT_ACT_RUN,
+		init:            false,
 	}
 	return r.run(p)
 }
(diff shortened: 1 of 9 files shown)


mrunalp:
@crosbymichael @cyphar @dqminh PTAL

mrunalp:
This is triggered by healthchecks in Dockerfile and kubernetes readiness/liveness probes. Credit to Ulrich Obergfell for pinpointing this issue.

exec.go
```diff
@@ -140,6 +140,7 @@ func execProcess(context *cli.Context) (int, error) {
 		detach:          detach,
 		pidFile:         context.String("pid-file"),
 		action:          CT_ACT_RUN,
+		init:            false,
```
dqminh: nit: should not need this.

dqminh:
LGTM

cyphar:
LGTM.

Issue #1755, opened 2018-03-08
rootfsPropagation=shared does not work


alban:
Tested with runc from git today (`git describe` = `v1.0.0-rc5-17-g9facb87f`).

How to test with `rootfsPropagation=shared`:

```

oci-runtime-tool generate --privileged --linux-rootfs-propagation=shared --process-terminal=true --rootfs-path=/home/alban/distro-trees/f26/ > config.json
sudo strace -f -e unshare,clone,mount,pivot_root,chdir,fchdir runc run c1

[pid  1602] mount("", "/", 0xc42009520c, MS_SHARED, NULL) = 0
[pid  1602] mount("", "/home", 0xc4200957d6, MS_PRIVATE, NULL) = 0
[pid  1602] mount("/home/alban/distro-trees/f26/", "/home/alban/distro-trees/f26/", 0xc4200957d7, MS_BIND|MS_REC, NULL) = 0
[pid  1602] mount("proc", "/home/alban/distro-trees/f26/proc", "proc", 0, NULL) = 0
[pid  1602] mount("tmpfs", "/home/alban/distro-trees/f26/dev", "tmpfs", MS_NOSUID|MS_STRICTATIME, "mode=755,size=65536k") = 0
[pid  1602] mount("devpts", "/home/alban/distro-trees/f26/dev/pts", "devpts", MS_NOSUID|MS_NOEXEC, "newinstance,ptmxmode=0666,mode=0"...) = 0
[pid  1602] mount("shm", "/home/alban/distro-trees/f26/dev/shm", "tmpfs", MS_NOSUID|MS_NODEV|MS_NOEXEC, "mode=1777,size=65536k") = 0
[pid  1602] mount("mqueue", "/home/alban/distro-trees/f26/dev/mqueue", "mqueue", MS_NOSUID|MS_NODEV|MS_NOEXEC, NULL) = 0
[pid  1602] mount("sysfs", "/home/alban/distro-trees/f26/sys", "sysfs", MS_RDONLY|MS_NOSUID|MS_NODEV|MS_NOEXEC, NULL) = 0
[pid  1602] chdir("/home/alban/distro-trees/f26/") = 0
[pid  1602] fchdir(9)                   = 0
[pid  1602] pivot_root(".", ".")        = 0
[pid  1602] fchdir(8)                   = 0
[pid  1602] mount("", ".", 0xc420095a64, MS_REC|MS_SLAVE, NULL) = 0
[pid  1602] chdir("/")                  = 0
[...]

PR #1813, opened 2018-06-06
libcontainer: fix a bug when setting shared rootfs propagation mode

dongsupark:
So far when the input mount flags contain `MS_SHARED`, the flag has not been applied to the container rootfs. That's because we call `rootfsParentMountPrivate()` after applying the original mount flags. Though we actually need to mount the container rootfs with `MS_PRIVATE`, to avoid failure from `pivot_root()` in the Linux kernel.

Thus if the mount flags contain `MS_SHARED`, we need to skip `rootfsParentMountPrivate()`, and call `chroot()` instead of `pivot_root()`.

With this fix, `validation/linux_rootfs_propagation_shared.t` of runtime-tools works well finally.

/cc @alban @cyphar 

Fixes https://github.com/opencontainers/runc/issues/1755

Signed-off-by: Dongsu Park <dongsu@kinvolk.io>

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -99,7 +99,11 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 		return newSystemErrorWithCausef(err, "changing dir to %q", config.Rootfs)
 	}
 
-	if config.NoPivotRoot {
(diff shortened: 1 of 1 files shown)


dongsupark:
Well, this PR has other issues. chroot is not exactly an ideal approach.
I'm going to think about it again, and try to address it in a different way.

PR #1814, opened 2018-06-07
SELinux labels are tied to the thread

rhatdan:
We need to lock the threads for the SetProcessLabel to work,
should also call SetProcessLabel("") after the container starts
to go back to the default SELinux behaviour.

Once you call SetProcessLabel, then any process executed by runc
will run with this label, even if the process is for setup rather
then the container.

It is always safest to call the SELinux calls just before the exec of the
container, so that other processes do not get started with the incorrect label.


libcontainer/standard_init_linux.go
```diff
@@ -187,7 +185,16 @@ func (l *linuxStandardInit) Init() error {
 			return newSystemErrorWithCause(err, "init seccomp")
 		}
 	}
-	if err := syscall.Exec(name, l.config.Args[0:], os.Environ()); err != nil {
+	runtime.LockOSThread()
```
crosbymichael: Did you try moving `runtime.LockOSThread()` up higher instead of moving the selinux code down?  I'm worried about seccomp and selinux in this flow.

cyphar: I agree, and would just put `runtime.LockOSThread` to the very beginning of `init` for both files and keep the rest of the code the same as before.

libcontainer/setns_init_linux.go
```diff
@@ -72,5 +70,14 @@ func (l *linuxSetnsInit) Init() error {
...
-	return system.Execv(l.config.Args[0], l.config.Args[0:], os.Environ())
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	if err := label.SetProcessLabel(l.config.ProcessLabel); err != nil {
+		return err
+	}
+	err := system.Execv(l.config.Args[0], l.config.Args[0:], os.Environ())
+	// Ignore error, there is nothing we can do if this fails
+	label.SetProcessLabel("")
```
cyphar: I don't think that this line makes sense. Your commit message says "after the container execs", but `execv(2)` **does not return** unless there was some error when attempting to execute the process -- which means that unsetting the process label will only have an impact for the few milliseconds it takes to error out and exit.

If you really want to make this change, I would make it a `defer` to keep the code cleaner.

libcontainer/standard_init_linux.go
```diff
@@ -187,7 +185,16 @@ func (l *linuxStandardInit) Init() error {
...
+	defer runtime.UnlockOSThread()
+	if err := label.SetProcessLabel(l.config.ProcessLabel); err != nil {
+		return errors.Wrap(err, "set process label")
+	}
+
+	err = syscall.Exec(name, l.config.Args[0:], os.Environ())
+	// Ignore error, there is nothing we can do if this fails
+	label.SetProcessLabel("")
```
cyphar: Ditto for this line.

libcontainer/standard_init_linux.go
rhatdan: The issue that triggered this PR was this AVC.
```
podman run -ti --blkio-weight-device /dev/sda1:500 alpine /bin/sh

triggered

type=AVC msg=audit(1528390506.883:851): avc:  denied  { append } for  pid=6511 comm="sh" name="blkio.weight_device" dev="cgroup" ino=56622 scontext=system_u:system_r:container_t:s0:c203,c914 tcontext=system_u:object_r:cgroup_t:s0 tclass=file
```
This avc shows a that runc launched a comm="sh" that is attempting to append to the blkio.weight_device file in the cgroup directory.  The problem is the shell is running as the container label, container_t.  Since this is part of the setup it should not be confined by the container type.  
It is my theory that runc launched a process after the SELinux label was set, which ran with the container label.  Moving the SetProcessLabel down in the stack means this will not happen.  I hav e no problem locking the threads much higher, if that is what you want.

libcontainer/setns_init_linux.go
rhatdan: If we can guarantee that no code will ever be execed after the system.Execv, then I am fine with removing the code.  Basically this line tells SELinux to go back to normal behaviour, IE DONT set the SELINUX Labels.  I was not sure if the Exec worked the same way in Go and the thread would not go back into a pool which go could use to launch other processes.

libcontainer/standard_init_linux.go
crosbymichael: I think it would be better to lock sooner before changing where selinux is being set.

rhatdan: I will lock at the top, but that does not eliminate the risk over where SELinux is set.

crosbymichael:
@rhatdan what risk where selinux is set?  Will we run into issues of a seccomp profile blocking us setting the process label by moving it after we finalize the seccomp profile?

rhatdan:
The risk is an executable (helper app) gets started between where it is set and the exec. If you can guarantee that will not happen then we could move it back.  The AVC I attached showed some kind of helper app was attempting to set the BLK IO Control in Cgroups which was running with the wrong label.  
SECCOMP should not block the SELinux call, since this is something a non privileged user is allowed to do.

rhatdan:
Now maybe the SetProcessLabel("") after the Exec might have fixed the helper shell problem, but I would figure this script was running before the container.

When are the hooks run?  They could also potentially be an issue.

libcontainer/setns_init_linux.go
cyphar: > I was not sure if the Exec worked the same way in Go and the thread would not go back into a pool which go could use to launch other processes.

`system.Execv` calls an `execve(2)` under the hood which does everything that entails (kill all the other threads and execute the process).

cyphar:
> SECCOMP should not block the SELinux call, since this is something a non privileged user is allowed to do.

But the seccomp profile could be used to block it (maybe the person who is running the container doesn't want any SELinux calls in their container). We generally want to keep seccomp profile application **as late** as possible to avoid requiring users to enable more syscalls because `runc` needs them internally before their application code ever runs.

> The risk is an executable (helper app) gets started between where it is set and the exec. If you can guarantee that will not happen then we could move it back. The AVC I attached showed some kind of helper app was attempting to set the BLK IO Control in Cgroups which was running with the wrong label.

From memory, cgroups are not set by this process (which is the `runc init` process). They are all set by the parent process, which shouldn't have these SELinux labels applied. Also I don't see any AVC attachment so I can't really see what logs you're talking about.

> When are the hooks run? They could also potentially be an issue.

Hooks are not spawned from this process either.

rhatdan:
Sorry I thought I attached this yesterday.

```
podman run -ti --blkio-weight-device /dev/sda1:500 alpine /bin/sh
```
Caused 
type=SYSCALL msg=audit(1528390506.883:851): arch=c000003e syscall=2 success=no exit=-30 a0=7f1c99cb3320 a1=441 a2=1b6 a3=0 items=0 ppid=6503 pid=6511 auid=0 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=pts0 ses=11 comm="sh" exe="/bin/busybox" subj=system_u:system_r:container_t:s0:c203,c914 key=(null)
type=AVC msg=audit(1528390506.883:851): avc:  denied  { append } for  pid=6511 comm="sh" name="blkio.weight_device" dev="cgroup" ino=56622 scontext=system_u:system_r:container_t:s0:c203,c914 tcontext=system_u:object_r:cgroup_t:s0 tclass=file

rhatdan:
Actually now that I look at this, it could be a leaked file descriptor?

rhatdan:
Ok I moved them somewhat back and added a defer incase one of the functions returns an error. on the SetProcessLabel("")
Any idea where the cgroup settings are being done?

crosbymichael:
The runc parent process sets up cgroups, not the containerized process.

https://github.com/opencontainers/runc/blob/master/libcontainer/process_linux.go#L278

rhatdan:
Yes but I can not see where this would leak.  Any ways I don't believe that the AVC's are related to this PR, but I still feel that this PR is necessary to make sure the SELinux labels get set correctly to the containers.

cyphar:
I think that setting the labels this late might break `"no_new_privs": false` if the SELinux stuff is seccomp'd (we set `seccomp` much earlier in that case -- I completely forgot about this until I reviewed the file again). Can you change it to apply before seccomp in both cases, something like:

```
if !noNewPrivileges {
  // set SELinux label
  // set seccomp
}
// rest of code
if noNewPrivileges {
  // set SELinux label
  // set seccomp
}
```

?

rhatdan:
Finally getting back to this one.  @cyphar I did the patch a few months ago the way you suggested.

cyphar:
LGTM, thanks (though I don't have an SELinux system to test on -- it seems about right given our previous discussion).

crosbymichael:
LGTM

Issue #941, opened 2016-07-13
i686 test failure: "user_test.go:441: constant 2147483648 overflows int"


onlyjob:
~~~~
# github.com/opencontainers/runc/libcontainer/user 
src/github.com/opencontainers/runc/libcontainer/user/user_test.go:441: constant 2147483648 overflows int 
~~~~

PR #1821, opened 2018-06-16
Really fix TestGetAdditionalGroups on i686 (Closes: #941)

onlyjob:
Although #941 has been closed it was never fixed:

~~~~
 src/github.com/opencontainers/runc/libcontainer/user/user_test.go:448:36: constant 2147483648 overflows int
~~~~

https://buildd.debian.org/status/fetch.php?pkg=runc&arch=i386&ver=1.0.0%7Erc5%2Bdfsg1-1&stamp=1529067561&raw=0

### DIFF
--- a/libcontainer/user/user.go
+++ b/libcontainer/user/user.go
@@ -473,7 +473,7 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
 				return nil, fmt.Errorf("Unable to find group %s", ag)
 			}
 			// Ensure gid is inside gid range.
-			if gid < minId || gid > maxId {
+			if gid < minId || gid >= maxId {
 				return nil, ErrRange
 			}
 			gidMap[gid] = struct{}{}
--- a/libcontainer/user/user_test.go
+++ b/libcontainer/user/user_test.go
@@ -445,7 +445,7 @@ this is just some garbage data
 	if utils.GetIntSize() > 4 {
 		tests = append(tests, foo{
 			// groups with too large id
-			groups:   []string{strconv.Itoa(1 << 31)},
+			groups:   []string{strconv.Itoa(1<<31 - 1)},
 			expected: nil,
 			hasError: true,
 		})


onlyjob:
Should be OK now.

libcontainer/user/user.go
```diff
@@ -473,7 +473,7 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
 				return nil, fmt.Errorf("Unable to find group %s", ag)
 			}
 			// Ensure gid is inside gid range.
-			if gid < minId || gid > maxId {
```
cyphar: Are you sure this change is correct?

onlyjob: Well, it works and passes tests on amd64 and i386 as far as I can tell. I think it is correct. Can we get someone to review?

cyphar: It's not so much that it passes tests, it's that you're now making it impossible for someone to have an additional group that is equal to `maxId` and I was wondering whether this you think it's actually semantically correct (I'm not entirely sure -- though I will double-check).

(I am the one doing the review. :wink:)

cyphar:
CI is failing because you haven't `gofmt`'d your source. You can do this with `go fmt ./...` or `gofmt -s -w <file>`.

libcontainer/user/user.go
onlyjob: This is very related to failure and I _had to_ change `>` to `>=` because of the following test failure on _amd64_ introduced as direct result of integer overflow fix:

~~~~
--- FAIL: TestGetAdditionalGroups (0.00s)
        user_test.go:459: Parse(user.foo{groups:[]string{"2147483647"}, expected:[]int(nil), hasError:true}) expects error but has none
~~~~

That looks perfectly reasonable to be but may be therei is a better way.

onlyjob:
I don't know about `gofmt`. I'd rather produce no-noise patch. I've tested the change on _amd64_ and _i386_. Please do `gofmt` as you like.

libcontainer/user/user.go
cyphar: Ah okay. I'll take another look at whether or not this is semantically okay and get back to you.

cyphar:
`gofmt` comes with every Go installation. It won't produce any more noise because all of our source code is already `gofmt`'d and we don't merge patches unless they have been `gofmt`'d.

In case you're interested, the problem is this line:

```
groups:   []string{strconv.Itoa( 1<<31 -1 )},
```

The space around `1<<31 - 1` makes it incorrect formatting. If you run `gofmt` I promise it will only reformat that line. :smile:

onlyjob:
Gofmt'd, thanks. :)

AkihiroSuda:
ping @cyphar

PR #1825, opened 2018-06-17
Stop using unix.SIGUNUSED which has been removed from golang.org/x/sys

onlyjob:
Bug-Debian: https://bugs.debian.org/889704

Fixes FTBFS `undefined: unix.SIGUNUSED`.


cyphar:
LGTM.

PR #1827, opened 2018-06-18
libcontainer: improve "kernel.{domainname,hostname}" sysctl handling

cyphar:
These sysctls are namespaced by CLONE_NEWUTS, and we need to use
"kernel.domainname" if we want users to be able to set an NIS domainname
on Linux. However we disallow "kernel.hostname" because it would
conflict with the "hostname" field and cause confusion (but we include a
helpful message to make it clearer to the user).

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -151,6 +151,16 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 				return fmt.Errorf("sysctl %q is not allowed in the hosts network namespace", s)
 			}
 		}
+		if config.Namespaces.Contains(configs.NEWUTS) {
+			switch s {
+			case "kernel.domainname":
+				// This is namespaced and there's no explicit OCI field for it.
+				continue
+			case "kernel.hostname":
+				// This is namespaced but there's a conflicting (dedicated) OCI field for it.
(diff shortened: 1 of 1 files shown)


cyphar:
This is a part of moby/moby#37302.

libcontainer/configs/validate/validator.go
```diff
@@ -151,6 +151,16 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 				return fmt.Errorf("sysctl %q is not allowed in the hosts network namespace", s)
 			}
 		}
+		if config.Namespaces.Contains(configs.NEWUTS) {
+			switch s {
+			case "kernel.domainname":
+				// This is namespaced and there's no explicit OCI field for it.
+				continue
```
dqminh: i dont think we need to explicitly single out `kernel.domainname` here since its not a TODO for OCI either right ?

cyphar: This code is **allowing** `kernel.domainname` which is why it needs to be singled out (the comment explains why it's allowed and `kernel.hostname` isn't). If you try to use `kernel.domainname` without this patch it will fail because `runc` thinks its not namespaced and fails container creation.

dqminh:
LGTM

hqhq:
LGTM

PR #1830, opened 2018-06-26
Pass GOMAXPROCS to init processes

crosbymichael:
This will help runc's init to not spawn many threads on large systems when
launched with max procs by the caller.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -461,6 +461,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
+	cmd.Env = append(cmd.Env, fmt.Sprintf("GOMAXPROCS=%s", os.Getenv("GOMAXPROCS")))
 	cmd.ExtraFiles = append(cmd.ExtraFiles, p.ExtraFiles...)
 	if p.ConsoleSocket != nil {
 		cmd.ExtraFiles = append(cmd.ExtraFiles, p.ConsoleSocket)


cyphar:
Does this actually help? From memory `GOMAXPROCS` doesn't actually restrict the number of threads spawned from syscalls blocking (we don't use `go` anywhere in `runc init` from memory).

cyphar:
Are you sure this makes a difference in the number of threads? From memory `GOMAXPROCS` doesn't actually restrict the number of threads spawned from syscalls blocking (we don't use `go` anywhere in `runc init` from memory).

crosbymichael:
@cyphar for the init, once go boots it will spawn many.  If a system has 128cores, it will spawn 128 threads out of the gate. While it won't limit things blocking on syscalls it will limit what go initially eats up.

libcontainer/container_linux.go
```diff
@@ -465,6 +465,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if p.ConsoleSocket != nil {
 		cmd.ExtraFiles = append(cmd.ExtraFiles, p.ConsoleSocket)
 		cmd.Env = append(cmd.Env,
+			fmt.Sprintf("GOMAXPROCS=%s", os.Getenv("GOMAXPROCS")),
```
dqminh: should this apply to only console socket ?, also we should check if GOMAXPROCS is indeed there before setting the env.

crosbymichael: oops

crosbymichael:
My findings could be wrong but this is what i'm using to debug more:

```go
package main

import (
        "fmt"
        "io/ioutil"
        "os"
        "runtime"
)

func check() {
        n := runtime.GOMAXPROCS(0)
        var s runtime.MemStats
        runtime.ReadMemStats(&s)
        pid := os.Getpid()
        files, err := ioutil.ReadDir(fmt.Sprintf("/proc/%d/task", pid))
        if err != nil {
                panic(err)
        }
        fmt.Println(n, s.MCacheInuse, len(files))
}

func init() {
        check()
        runtime.GOMAXPROCS(2)
        check()
}

func main() {
        check()
}

```

```bash
michael|~/scratch > go run num.go
4 6944 5
2 3472 6
2 3472 6
michael|~/scratch > GOMAXPROCS=1 go run num.go
1 1736 3
2 3472 5
2 3472 5

```

crosbymichael:
Maybe this does not make as much sense for a short lived process like the init.  What do you think @cyphar @dqminh ?

cyphar:
@crosbymichael 

>  If a system has 128 cores, it will allocate for 128.

*Aaahh*. Alright, that makes sense. I will see if I have access to a machine with that many cores. On my laptop (4-core machine) using `GOMAXPROCS=1 go run yourscript.go` I reduce the number of tasks from `5` to `3` -- so it definitely does make a difference. Though it should be noted that `runtime.GOMAXPROCS(...)` doesn't appear to kill any of the already-running threads (though that might just be because I'm on a "low" core count machine -- and most of the threads are spawned from syscall blocking and not for the M-to-N scheduling shenanigans).

dqminh:
LGTM.

I think effect will probably not that significant, but it's ok to do regardless. Particularly we dont really need high number of go routines.

libcontainer/container_linux.go
```diff
@@ -461,6 +461,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	if cmd.SysProcAttr == nil {
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 	}
+	cmd.Env = append(cmd.Env, fmt.Sprintf("GOMAXPROCS=%s", os.Getenv("GOMAXPROCS")))
```
cyphar: Won't this result in `GOMAXPROCS=` in the normal case of `runc` was called without `GOMAXPROCS` set?

dqminh:

cyphar:
LGTM -- I thought this was already merged TBH.

kolyshkin:
Wonder why we don't pass `GOMAXPROCS=1` (since this is what `runc init` sets explicitly in `init()` anyway?

PR #1832, opened 2018-06-29
linux: drop check for /proc as invalid dest

giuseppe:
it is now allowed to bind mount /proc when the PID namespace is shared
with the host.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -413,7 +413,7 @@ func checkMountDestination(rootfs, dest string) error {
 		if err != nil {
 			return err
 		}
-		if path == "." || !strings.HasPrefix(path, "..") {
+		if path != "." && !strings.HasPrefix(path, "..") {
 			return fmt.Errorf("%q cannot be mounted because it is located inside %q", dest, invalid)
 		}
 	}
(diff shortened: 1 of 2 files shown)


giuseppe:
tests fixed!

rhatdan:

giuseppe:
@dqminh would you mind to take a quick look at this patch as well? :-)

mrunalp:
@cyphar @crosbymichael @dqminh PTAL

libcontainer/rootfs_linux.go
```diff
@@ -53,7 +53,7 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 			}
 		}
 
-		if err := mountToRootfs(m, config.Rootfs, config.MountLabel); err != nil {
+		if err := mountToRootfs(m, config.Rootfs, config.MountLabel, config.Namespaces.Contains(configs.NEWNS)); err != nil {
```
crosbymichael: Why are you using NEWNS here when the description of the PR talks about NEWPID?

giuseppe: sorry that was a leftover of a cleanup.  The patch was originally written to address only the "chroot" case but I've changed it to address also another problem: `/proc` could not be bind mounted from the host.  This is useful for rootless containers when the PID namespace is shared with the host (and the container cannot mount a new `/proc`).

I've pushed a new version which is much simpler now.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #1833, opened 2018-07-02
rootless: fix Docker-in-LXD regression

AkihiroSuda:

"Step 1" of https://github.com/opencontainers/runc/issues/1837

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -237,12 +237,31 @@ func isRootless(context *cli.Context) (bool, error) {
 		}
 		// nil b stands for "auto detect"
 	}
-	// Even if os.Geteuid() == 0, it might still require rootless mode,
-	// especially when running within userns.
-	// So we use system.GetParentNSeuid() here.
+	if os.Geteuid() != 0 {
+		return true, nil
+	}
+	if !system.RunningInUserNS() {
+		// euid == 0 in the initial namespace
+		return false, nil
+	}
+	// euid == 0 in userns.
+	// TODO: In this case, we should not enable the entire "rootless mode", but we should allow ignoring cgroup permission errors.
+	// See https://github.com/opencontainers/runc/issues/1837
 	//
-	// TODO(AkihiroSuda): how to support nested userns?
-	return system.GetParentNSeuid() != 0 || system.RunningInUserNS(), nil
+	// * When runc is executed via Docker-in-LXD, isRootless() returns false, because systemd in LXD does not set $USER.
+	//   If Docker was launched manually in LXD, isRootless() still returns false, because $USER would be set to "root".
+	//   Note that returning false is the expected behavior for Docker-in-LXD, because LXD sets up cgroups.
+	//
+	// * When runc is executed via rootless img, buildkit, and a few other tools, isRootless() returns true,
+	//   because they don't change environment variables after unsharing the userns and mapping UID=0 to the current user.
+	//
+	// Corner case:
+	// * When runc is executed in Docker-in-rootless-Docker ("rootless dind"), as the root in the contaienr,
+	//   isRootless() returns false, and unlikely to work, unless cgroups is configured.
+	//   The dockerd in the container would need to specify `runc --rootless` explicitly in this case.
+	//   (And user would need to launch dockerd with --rootless explicitly, probably)
+	u := os.Getenv("USER")
+	return u != "" && u != "root", nil
 }
 
 func createContainer(context *cli.Context, id string, spec *specs.Spec) (libcontainer.Container, error) {
--- a/libcontainer/system/linux.go
+++ b/libcontainer/system/linux.go
@@ -3,7 +3,6 @@
 package system
 
 import (
-	"os"
 	"os/exec"
 	"syscall" // only for exec
 	"unsafe"
@@ -120,22 +119,6 @@ func UIDMapInUserNS(uidmap []user.IDMap) bool {
 		return false
 	}
 	return true
-}
-
-// GetParentNSeuid returns the euid within the parent user namespace
-func GetParentNSeuid() int64 {
-	euid := int64(os.Geteuid())
-	uidmap, err := user.CurrentProcessUIDMap()
-	if err != nil {
-		// This kernel-provided file only exists if user namespaces are supported
-		return euid
-	}
-	for _, um := range uidmap {
-		if um.ID <= euid && euid <= um.ID+um.Count-1 {
-			return um.ParentID + euid - um.ID
-		}
-	}
-	return euid
 }
 
 // SetSubreaper sets the value i as the subreaper setting for the calling process
--- a/libcontainer/system/unsupported.go
+++ b/libcontainer/system/unsupported.go
@@ -19,9 +19,3 @@ func RunningInUserNS() bool {
 func UIDMapInUserNS(uidmap []user.IDMap) bool {
 	return false
 }
-
-// GetParentNSeuid returns the euid within the parent user namespace
-// Always returns os.Geteuid on non-linux
-func GetParentNSeuid() int {
-	return os.Geteuid()
-}


AkihiroSuda:
@cyphar PTAL

utils_linux.go
```diff
@@ -237,12 +237,12 @@ func isRootless(context *cli.Context) (bool, error) {
...
-	// especially when running within userns.
-	// So we use system.GetParentNSeuid() here.
+	// if os.Geteuid() != 0: regardless of whether we are in a userns or in the initial namespace,
+	// we need to enable the rootless mode.
 	//
-	// TODO(AkihiroSuda): how to support nested userns?
-	return system.GetParentNSeuid() != 0 || system.RunningInUserNS(), nil
+	// if system.RunningInUserNS() == true: regardless of whether the userns was created by a regular
+	// user or the root (in the initial namespace), we need to enable the rootless mode.
+	return os.Geteuid() != 0 || system.RunningInUserNS(), nil
```
cyphar: This change makes sense -- because the parent namespace's euid is not actually use for permission checks. However, I think that this check is still not entirely correct -`os.Geteuid() == 0` is too strict of a check -- I think we should be checking whether the current process has `CAP_SYS_ADMIN`.

AkihiroSuda: Also, on second thought, I'm not sure we should always automatically enable rootless mode when running in userns. Maybe this has been breaking Docker-in-LXD usecase? (Although the corresponding commit is not used in Docker yet)

cyphar: We can always ask @brauner. They do have some runc patches from memory.

AkihiroSuda: Maybe better to just have `return system.GetParentNSeuid() != 0`, as in the initial revision of the code?
https://github.com/opencontainers/runc/pull/1688#discussion_r186716892

AkihiroSuda: cc @giuseppe

brauner: Yes, it is conceptually very different to be root in a user namespace then to be fully unprivileged. In other words, you need to always account for the fact that you might be privileged in a userns first and then you need to work around the quirks where the kernel checks your privileges against the init userns [1]. Uid `0` in @lxc is meaningless without the information about whether or not this is tied to a user namespace. The runtime basically requires a three-dimensional permission model: capabilities, uids, usernses.

[1]: Usually a "ask for forgiveness not permission" model works fine but note that this will be broken with `4.18` where we enabled `mknod()` in user namespaces as long as you're privileged towards that userns. But any device node your create will be unusable since `open()` will fail on it. To give it another spin starting with `4.18` `open()` will be a more privileged operation than `mknod()` for devices. So if you lived by that assumption you might want to change this.

AkihiroSuda: Actually, looks like my commit breaks Docker in LXD (sorry):
```
$ lxc launch ubuntu:18.04 foo -c security.nesting=true
$ lxc file push /usr/local/sbin/runc foo/usr/bin/docker-runc
...
foo# docker run -it --rm busybox
docker: Error response from daemon: OCI runtime create failed: cannot specify gid= mount options for unmapped gid in rootless containers: unknown.
```

Could you suggest logic for `isRootless`?

For LXD compatibility, we could remove userns detection from `isRootless()`, but then we will break img/buildkit/buildah/podman probably.

brauner: So I need to know some stuff:
- Do you have a `fork()`/`clone()`+  `exec()` model whereby the child process is the container and runs in a new set of namespaces or something else?
- Where is `IsRootless()` supposed to be called? In the parent namespaces or the child namespaces (presupposing a particular model in this question so feel free to reformulate as approriate)?
- What decisions are based on `IsRootless()`?

AkihiroSuda: > Do you have a fork()/clone()+ exec() model whereby the child process is the container and runs in a new set of namespaces or something else?

`img` and [RootlessKit](https://github.com/AkihiroSuda/rootlesskit) (used by [BuildKit](https://github.com/moby/buildkit/blob/80675875ca98274bf5d7d50d7bb64050f6b9a1b5/hack/dockerfiles/test.Dockerfile#L153), [containerd PR](https://github.com/containerd/containerd/pull/2006) and [Docker POC](https://github.com/moby/moby/issues/37375)) do this.

I think Buildah and Podman do similar stuff but haven't looked into their code deeply yet.

In addition to user namespace, we sets `CLONE_NEWNS` and optionally `CLONE_NEWNET`.

However, unlike Docker-in-LXD, we don't do `chroot` nor `pivot_root`.
Also, we keep environment variables such as `$USER`.
i. e. we we execute `runc` as UID=0 but `$USER="penguin"` in userns+mntns(+netns).


> Where is IsRootless() supposed to be called? In the parent namespaces or the child namespaces (presupposing a particular model in this question so feel free to reformulate as approriate)?

Child namespaces, from POV of `img`.

But note that executing `runc` in the initial namespace (as non-root UID) needs to be also allowed.


> What decisions are based on IsRootless()?

https://github.com/opencontainers/runc/issues/1837

AkihiroSuda:
Updated PR, please refer to step 1 of https://github.com/opencontainers/runc/issues/1837

utils_linux.go
brauner: Sorry, just saw your answers. Thanks.

> However, unlike Docker-in-LXD, we don't do chroot nor pivot_root.

Out of curiosity, only for the fully unprivileged case?

utils_linux.go
```diff
@@ -237,12 +236,16 @@ func isRootless(context *cli.Context) (bool, error) {
...
+	// We don't check EUID/UID and namespaces here. See https://github.com/opencontainers/runc/issues/1837
+	// * When runc is executed in userns via Docker-in-LXD, isRootless() returns false, because LXD would set $USER to "root".
+	// * When runc is executed in userns via rootless img, buildkit, and a few other tools, isRootless() returns true, because they don't change environment variables after unsharing the userns and mapping UID=0 to the current user.
+	// * When runc is executed as a regular user in the initial namespace, isRootless() returns true
+	// * When runc is executed as the root in the initial namespace, isRootless() returns false
+	// Corner cases:
+	// * When runc is executed in userns via Docker-in-rootless-Docker ("rootless dind"), as the root in the contaienr, isRootless() returns false, and unlikely to work. The dockerd in the container would need to specify runc --rootless explicitly in this case. (And user would need to launch dockerd with --rootless explicitly, probably)
+	// * When runc is executed in "rootless dind", as a non-root in the contaienr, isRootless() returns true and likely to work.
+	u := os.Getenv("USER")
```
danail-branekov: The `USER` environment variable seems to be set by the login process (see [man login(1)](http://man7.org/linux/man-pages/man1/login.1.html). Garden runs as server (i.e. it is not logging in) and therefore `USER` is not set when runc is invoked which makes `isRootless` return false regardless of whether we want rootless or not (`os.Getenv` returns an empty string if the var is not set)

In order to support both cases via checking the `USER` value only if it is set and fallback to the effective user ID if not, i.e. something like
```golang

func isRootless(context *cli.Context) (bool, error) {
  if context != nil {
      ...
  }

  user, userSet := os.LookupEnv("USER")
  if userSet {
    return user != "" && user != "root", nil
  }

  return system.GetParentNSeuid() != 0 || system.RunningInUserNS(), nil
}
```

What do you think?

cc @ostenbom @teddyking @julz

giuseppe: @danail-branekov this looks like a safer version, I think it makes sense to fallback to the euid when `USER` is not set

AkihiroSuda: Thanks, but it doesn't fix Docker-in-LXD regression, because Docker-in-LXD executes runc without `$USER`, as euid=0 in userns, but it expects `isRootless()` to return false...

AkihiroSuda: > Garden runs as server

How do you run rootless garden as a server? via `systemd --user`?

`systemd --user` seems to set `$USER`, at least on Ubuntu 18.04?

```console
suda@suda-ws01:~$ systemctl show-environment
LANG=en_US.UTF-8
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

```console
suda@suda-ws01:~$ systemctl --user show-environment
HOME=/home/suda
LANG=en_US.UTF-8
LOGNAME=suda
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
SHELL=/bin/bash
USER=suda
XDG_RUNTIME_DIR=/run/user/1001
```

AkihiroSuda:
I also made more precise version ("Step 2" of https://github.com/opencontainers/runc/issues/1837), but very huge change: https://github.com/AkihiroSuda/runc/commit/dc7a32a6ca56c3016fbe0880df53fbea19e7a5ba

So I suggest merging this PR (#1833) first if it doesn't break CloudFoundry.

utils_linux.go
ostenbom: We run garden with monit, which doesn't set `USER`. We also run the server as a user with maximum user id that does not exist in `/etc/passwd/` or elsewhere, so it would be best if we could run runc in rootless mode without setting `$USER`. On the other hand we can always run with the `--rootless` flag, so I guess this isn't too big of a problem for us. 

To clarify, in the issue you said `When runc is executed in userns via Docker-in-LXD, isRootless() returns false, because LXD would set $USER to "root".` - I don't understand in which case `$USER` *is* being set in Docker-in-LXD here?

AkihiroSuda: >because LXD would set $USER to "root”

sorry this was my wrong assumption

danail-branekov: >> because LXD would set $USER to "root”
> sorry this was my wrong assumption

So if Docker-in-LXD does NOT set $USER at all (i.e. $USER is either NOT on the environment, or set to en empty string when runc is run), then your PR would make `isRootless` return false, i.e. it would run rootful. Is that what you want to achieve? Is that what fixes the regression?

If yes, could you please update the comment in the code to something like
```
When runc is executed in userns via Docker-in-LXD, isRootless() returns false, because LXD would not set $USER.
``` 
so that we do not get confused?

AkihiroSuda: updated. also added EUID check

brauner: LXD does not set anything at all unless explicitly instructed via environment variables. As soon as the basic container setup is done everything is in the hands of the binary that is executed. In the regular case for LXD and LXC this means a full-blown init system.

danail-branekov: > updated. also added EUID check

Oh, great! The EUID check makes our CF Garden acceptance tests quite happy 👍 

Thanks!

brauner:
Sorry, I obviously lack context and this more curiosity than anything else but I still don't fully understand why there is the conceptual necessity to have something like `IsRootless()` in a container runtime. @lxc has support for fully unprivileged containers  for a long time (which is what you cover with the term "rootless") but it is not a different concept. It is just a specific point in a permissions-space that is managed continuously. So if you instruct @lxc with:
```
lxc.idmap = u 1000 1000 1
lxc.idmap = g 1000 1000 1

lxc.init.uid = 1000
lxc.init.gid = 1000
```
It'll just know what's up and manage your permissions accordingly:
[asciicast](https://asciinema.org/a/190267)

AkihiroSuda:
> I still don't fully understand why there is the conceptual necessity to have something like IsRootless() in a container runtime.

* When `IsRootless() == true`, runc ignores cgroup permission errors, because generally we don't have the write access to cgroups. (And we don't use `lxcfs` or `pam_cgfs`.) 

* When `IsRootless() == true`, runc uses `$XDG_RUNTIME_DIR` instead of `/run`.

The current implementation also does a bunch of stuff that is not needed at all when we are already in userns, but these stuff will be removed in "Step 2" of https://github.com/opencontainers/runc/issues/1837
(POC: https://github.com/AkihiroSuda/runc/commits/decompose-rootless)

AkihiroSuda:

dqminh:
LGTM

kolyshkin:
ping @opencontainers/runc-maintainers PTAL

crosbymichael:
This logic is getting more and more complex everytime it changes

AkihiroSuda:
The logic will be simplified in "Step 2" of #1837 
https://github.com/AkihiroSuda/runc/commit/dc7a32a6ca56c3016fbe0880df53fbea19e7a5ba

utils_linux.go
```diff
@@ -237,12 +237,31 @@ func isRootless(context *cli.Context) (bool, error) {
...
+	//   because they don't change environment variables after unsharing the userns and mapping UID=0 to the current user.
 	//
-	// TODO(AkihiroSuda): how to support nested userns?
-	return system.GetParentNSeuid() != 0 || system.RunningInUserNS(), nil
+	// Corner case:
+	// * When runc is executed in Docker-in-rootless-Docker ("rootless dind"), as the root in the contaienr,
+	//   isRootless() returns false, and unlikely to work, unless cgroups is configured.
+	//   The dockerd in the container would need to specify `runc --rootless` explicitly in this case.
+	//   (And user would need to launch dockerd with --rootless explicitly, probably)
+	u := os.Getenv("USER")
```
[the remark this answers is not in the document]
frezbo: I would second @giuseppe on using `XDG_RUNTIME_DIR`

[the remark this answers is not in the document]
AkihiroSuda: Sorry what is the point 2.2?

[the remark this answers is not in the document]
giuseppe: the second bullet point in step 2:

"Switch the cgroup manager to libcontainer.RootlessCgroupfs"

We should detect cgroup availability explicitly by probably trying mkdir /sys/fs/cgroup/foo/bar or something similar. I guess the overhead is negligible.
Or just remove libcontainer.RootlessCgroupfs manager and ignore all errors.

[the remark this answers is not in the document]
AkihiroSuda: Is it better to just close this PR and open the step2 PR now?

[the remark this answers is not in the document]
giuseppe: fine by me, but I think some runc maintainer should ensure it can get reviewed promptly.  If that is not possible, this much simpler PR can be an intermediate workaround

AkihiroSuda:
Opened the "Step 2" PR. PTAL.
https://github.com/opencontainers/runc/pull/1862

I'm closing this "Step 1" PR.

PR #1834, opened 2018-07-02
rootless: disable unsupported security stuff

AkihiroSuda:

### DIFF
--- a/libcontainer/specconv/example.go
+++ b/libcontainer/specconv/example.go
@@ -217,5 +217,13 @@ func ToRootless(spec *specs.Spec) {
 	spec.Mounts = mounts
 
 	// Remove cgroup settings.
+	// NOTE: cgroups is supported when the permission is properly configured: https://github.com/opencontainers/runc/pull/1540
+	// The caller can set cgroups configuration after calling ToRootless().
 	spec.Linux.Resources = nil
+	spec.Linux.CgroupsPath = ""
+
+	// Disable unsupported security stuff
+	spec.Process.User.AdditionalGids = nil
+	spec.Process.ApparmorProfile = ""
+	spec.Process.SelinuxLabel = ""
 }


AkihiroSuda:
cc @cyphar @williammartin @giuseppe

giuseppe:
I am not a maintainer, but LGTM

libcontainer/specconv/example.go
```diff
@@ -217,5 +217,13 @@ func ToRootless(spec *specs.Spec) {
...
 	// Remove cgroup settings.
+	// NOTE: cgroups is supported when the permission is properly configured: https://github.com/opencontainers/runc/pull/1540
+	// The caller can set cgroups configuration after calling ToRootless().
 	spec.Linux.Resources = nil
+	spec.Linux.CgroupsPath = ""
+
+	// Disable unsupported security stuff
+	spec.Process.User.AdditionalGids = nil
+	spec.Process.ApparmorProfile = ""
+	spec.Process.SelinuxLabel = ""
```
cyphar: We can't just disable this stuff like that -- if a user has specified them in the configuration then we have to either fail or try to do what they've asked. If we just ignore it (which is what this change does) then we could lead users into thinking their container has security properties it doesn't.

Also `spec.Process.User.AdditionalGids = nil` will break containers that are using `newgidmap` to allow for `setgroups` and the mapping of multiple users (unless I'm mistaken). The same with `CgroupsPath` -- if a privileged process has set up cgroups for us then we can use them.

cyphar:
NACK. I'm pretty sure this'll break configurations that we shouldn't be breaking -- I'm a little confused that the tests are passing when this change breaks `cgroupsPath` handling (something that we added in #1540).

libcontainer/specconv/example.go
AkihiroSuda: I'm closing but `AdditionalGids` is not supported either way at the moment even when newgidmap is used

https://github.com/opencontainers/runc/blob/3a079311a7e9afeab4b722616cefd2f9b4129104/libcontainer/init_linux.go#L286-#L294

rhatdan:
SELinux labeling is supported for non privilged users.  SELinux does not care about whether you are UID=0 or any other UID.

PR #1836, opened 2018-07-02
Add osusergo flag to static build

kolyshkin:
This should fix the following (very legitimate) warnings on static build:

> /tmp/go-link-818454663/000019.o: In function `mygetgrouplist':
> /usr/lib/go-1.10/src/os/user/getgrouplist_unix.go:15: warning: Using
> 'getgrouplist' in statically linked applications requires at runtime the
> shared libraries from the glibc version used for linking
>
> /tmp/go-link-818454663/000018.o: In function `mygetgrgid_r':
> /usr/lib/go-1.10/src/os/user/cgo_lookup_unix.go:38: warning: Using
> 'getgrgid_r' in statically linked applications requires at runtime the
> shared libraries from the glibc version used for linking
>
> ...

as well as segfaults in the resulting binary.

For more details, check https://github.com/golang/go/issues/23265


crosbymichael:
LGTM

cyphar:
Oh awesome. I didn't know that Go's stdlib actually had a way of disabling the `glibc` `os/user` implementation.

LGTM.

PR #1838, opened 2018-07-03
criu tests: rename criu feature check

adrianreber:
Upstream renamed the feature check for lazy migration support from
'lazy_pages' to 'uffd'. The lazy migration test case was therefore
not running at all. This enables the lazy migration test case in runc
again.

The test will, however, not run in travis as the kernel is too old.
But it works again locally.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #1840, opened 2018-07-10
cr: don't dump network devices and their configuration

avagin:
RunC doesn't manage network devices and their configurations,
so it is impossible to describe external dependencies to restore them
back.

This means that all users have to set --empty-ns network, so let's do
this by default.

Signed-off-by: Andrei Vagin <avagin@openvz.org>

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -125,7 +125,8 @@ var namespaceMapping = map[specs.LinuxNamespaceType]int{
 }
 
 func setEmptyNsMask(context *cli.Context, options *libcontainer.CriuOpts) error {
-	var nsmask int
+	/* Runc doesn't manage network devices and their configuration */
+	nsmask := unix.CLONE_NEWNET
 
 	for _, ns := range context.StringSlice("empty-ns") {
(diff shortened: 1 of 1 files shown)


checkpoint.go
```diff
@@ -127,6 +127,9 @@ var namespaceMapping = map[specs.LinuxNamespaceType]int{
 func setEmptyNsMask(context *cli.Context, options *libcontainer.CriuOpts) error {
 	var nsmask int
 
+	/* Runc doesn't manage network devices and their configuration */
+	nsmask = unix.CLONE_NEWNET
+
```
kolyshkin: Perhaps just
```go
- 	var nsmask int
￼+ 	/* Runc doesn't manage network devices and their configuration */
￼+ 	nsmask := unix.CLONE_NEWNET
```
?

kolyshkin:
LGTM

crosbymichael:
LGTM

avagin:
@cyphar @mrunalp @dqminh @hqhq Could one of you review this pr? Thanks!

hqhq:
LGTM

PR #1841, opened 2018-07-11
Fix the problem TESTFLAGS is not to be used in Makefile correctly

KentaTada:
Currently, To pass argument "TESTFLAGS" for Makefile to execute "make test" is meaningless because TESTFLAGS is used in both localunittest and localintegration. So, this commit modifies Makefile to handle test targets for "unittest" and "integration" correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,25 +66,25 @@ localtest:
 	make localunittest localintegration localrootlessintegration
 
 unittest: runcimage
-	docker run -e TESTFLAGS -t --privileged --rm -v /lib/modules:/lib/modules:ro -v $(CURDIR):/go/src/$(PROJECT) $(RUNC_IMAGE) make localunittest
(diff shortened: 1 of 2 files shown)


crosbymichael:
@KentaTada is the problem that TESTFLAGs is not being passed down or that it's not used in the docker case?

KentaTada:
@crosbymichael  I'm sorry I didn't make it clear enough.
The problem is that TESTFLAG is not being passed down to localintegration correctly because TESTFLAG are used in both localunittest and localintegration. For example,  'make test TESTFLAGS="-run=TestValidateNetworkWithoutNETNamespace"' will test 'bats -t tests/integration-run=TestValidateNetworkWithoutNETNamespace' as localintegration test but it will fail. So, I added TESTPATH for localintegration and removed unused TESTFLAG. I think this new flag TESTPATH will be used like 'make test TESTFLAGS="-run=TestValidateNetworkWithoutNETNamespace" TESTPATH="/ps.bats"'.

crosbymichael:
LGTM

KentaTada:
@crosbymichael  Thank you for your review.

/cc @rjnagal @vmarmol @mrunalp @dqminh @hqhq @cyphar
Please review this commit.

dqminh:
LGTM

PR #1826, opened 2018-06-17
specconv: fix null spec.Process making runc panic

jingxiaolu:
`Process` of `Spec struct` is a pointer:
```go
	Process *Process `json:"process,omitempty"`
```

When calling `CreateLibcontainerConfig()` with  `opts.Spec.Process: nil`, runc will panic.

I added a new test case `TestNullProcess` in `libcontainer/specconv`, it will failed with:
```sh
=== RUN   TestNullProcess
--- FAIL: TestNullProcess (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xc8 pc=0x540029]

goroutine 20 [running]:
testing.tRunner.func1(0xc42007e9c0)
        /usr/lib/golang/src/testing/testing.go:622 +0x29d
panic(0x57c280, 0x861a20)
        /usr/lib/golang/src/runtime/panic.go:489 +0x2cf
github.com/opencontainers/runc/libcontainer/specconv.CreateLibcontainerConfig(0xc42003af70, 0x3, 0xc42007ad00, 0xc42007c040)
        /home/workspace/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go:244 +0x7b9
github.com/opencontainers/runc/libcontainer/specconv.TestNullProcess(0xc42007e9c0)
        /home/workspace/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux_test.go:456 +0xa8
testing.tRunner(0xc42007e9c0, 0x5b9440)
        /usr/lib/golang/src/testing/testing.go:657 +0x96
created by testing.(*T).Run
        /usr/lib/golang/src/testing/testing.go:697 +0x2ca
FAIL    github.com/opencontainers/runc/libcontainer/specconv    0.009s
```

So I wrote this PR.

Please help to review and share your comments.

Signed-off-by: Jingxiao Lu <lujingxiao@huawei.com>

PR #1848, opened 2018-07-20
Fix panic with config.json with empty process

alban:
When config.json does not have a "process" section at all, runc paniced
with this error:
> panic: runtime error: invalid memory address or nil pointer dereference

According to the spec, "process" is optional:
https://github.com/opencontainers/runtime-spec/blob/master/config.md#process
> process (object, OPTIONAL) specifies the container process. This property is REQUIRED when start is called.

So the desired behaviour is:
- "runc create": succeed without error
- "runc start": return a proper error without panic

This patch fixes the panic in "runc create". Further work is needed to
allow create but not start.

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -392,6 +392,9 @@ func (r *runner) checkTerminal(config *specs.Process) error {
 }
 
 func validateProcessSpec(spec *specs.Process) error {
+	if spec == nil {
+		return fmt.Errorf("process property must not be empty")
+	}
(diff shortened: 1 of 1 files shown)


wking:
I think this, like #1826, is a dup of #1726.

AkihiroSuda:
should have been covered in https://github.com/opencontainers/runc/commit/16d55f17a853c643022445bb39a03e42f2865f19

Feel free to reopen if not.

Issue #1786, opened 2018-04-18
Restoring a checkpointed container into an existing namespace is not possible


adrianreber:
I am currently trying to fix (better implement) the non-existing functionality to restore a container into an existing network namespace. I already had a short discussion with @avagin on IRC but wanted to use this place to hopefully come to a correct solution.

I am starting a container and I want it to join an existing network namespace:

```
	{
		"type": "network",
		"path": "/run/netns/test"
	},
```

This works just like it should. The container is running and uses the network namespace 'test' as specified above. My next step is to checkpoint the container, which works and then to restore the container, which seems to work. Upon closer inspection I see that the restored contained is running in a network namespace but not the one I specified ('test') but a network namespace created by CRIU during restore. I would have probably not detected the problem if it would have been a PID namespaces, but as I have setup a veth pair in the 'test' network namespace with an IP address it became clear that the CRIU restored container is running in another network namespace.

With older versions of CRIU the veths have been restored correctly but the latest CRIU version (using criu-dev branch) seems to have a problem. But this is not the problem I am trying to understand right now.

For me it seems wrong that runc is told to use the network namespace 'test' and I configured it correctly, but CRIU just uses another namespace.

Luckily CRIU is prepared for cases like this. It has the possibility to join an existing namespace:

```
  -J|--join-ns NS:{PID|NS_FILE}[,OPTIONS]
			Join existing namespace and restore process in it.
			Namespace can be specified as either pid or file path.
			OPTIONS can be used to specify parameters for userns:
			    user:PID,UID,GID
```

This is also exported via CRIU's RPC interface so it can easily be used with the following patch:

```
@@ -1161,6 +1162,17 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
                },
        }
 
+       nsPath := c.config.Namespaces.PathOf(configs.NEWNET)
+       if nsPath != "" {
+               // A network namespace path has been set. Tell CRIU to restore
+               // the processes into that namespace. This only works if the
+               // processes have been dumped *without* a network namespace.
+               join := new(criurpc.JoinNamespace)
+               join.Ns = proto.String("net")
+               join.NsFile = proto.String(c.config.Namespaces.PathOf(configs.NEWNET))
+               req.Opts.JoinNs = append(req.Opts.JoinNs, join)
+       }
+
        for _, m := range c.config.Mounts {
                switch m.Device {
                case "bind":
```

[...]

PR #1849, opened 2018-07-24
Add support to checkpoint and restore into external network namespaces

adrianreber:
With the initial CRIU patches merged to support checkpointing and restoring a container into a external defined network namespace, these are the necessary runc changes to honour an external network namespace (something like ```"path": "/run/netns/test"```).

Details about the existing problem are described at https://github.com/opencontainers/runc/issues/1786

There are still two CRIU patches under review to complete CRIU's RPC interface which is used by runc.

Once all the necessary patches are merged and a new CRIU release is available with these patches the CRIU version in runc's travis test can be updated and from that point on the newly added test case in ```checkpoint.bats``` should be automatically activated. Currently it detects the missing CRIU functionality and is skipped. (@avagin FYI)

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -657,7 +657,7 @@ func (c *linuxContainer) checkCriuFeatures(criuOpts *CriuOpts, rpcOpts *criurpc.
 		Features: criuFeat,
 	}
 
-	err := c.criuSwrk(nil, req, criuOpts, false)
+	err := c.criuSwrk(nil, req, criuOpts, false, nil)
 	if err != nil {
 		logrus.Debugf("%s", err)
 		return fmt.Errorf("CRIU feature check failed")
@@ -770,7 +770,7 @@ func (c *linuxContainer) checkCriuVersion(minVersion int) error {
 		Type: &t,
 	}
 
-	err := c.criuSwrk(nil, req, nil, false)
+	err := c.criuSwrk(nil, req, nil, false, nil)
 	if err != nil {
 		return fmt.Errorf("CRIU version check failed: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


crosbymichael:
This fixes issues when the container is running in the host namespace right?

adrianreber:
I guess no. If starting a container with
```
	{
		"type": "network",
		"path": "/run/netns/test"
	},
```
the container is running in the specified network namespace. If I checkpoint and restore the container it is no longer running in that specified network namespace, but in some network namespace CRIU created. So if the network namespace is configured before starting (or restoring) runc the restored container is running somewhere else and all the settings in the specified network namespace are not used.

According to the code (I have not verified it) CRIU can restore the network adapter correctly in the network namespace, but it will be a different namespace than defined in ```config.json```.

The use case is, I am setting up a network namespace, and starting a container. After restore the container is still running in exactly that namespace and not some random namespace created by CRIU.

I also had a look how LXC does it and it seems LXC does not really care about the name of the network namespace. But as runc has the option to specify an external network namespace it makes sense for the restored container to run again in that network namespace.

Is that a better explanation?

crosbymichael:
LGTM

adrianreber:
Please do not merge yet, we are still discussing the details on the CRIU level.

adrianreber:
I force-pushed the final version. Sorry for the confusion. All necessary patches are now part of CRIU and as soon as runc updates its CRIU version the new test case will be run. Depending on the outstanding reviews this is now ready to be merged.

CC @avagin  (as he was heavily involved in getting the interface between CRIU and runc right. Thanks.)

crosbymichael:
ping @avagin

libcontainer/container_linux.go
```diff
@@ -928,6 +929,33 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
...
+		if err == nil {
+			// CRIU expects the information about an external namespace
+			// like this: --external net[<inode>]:<key>
+			// This <key> needs to be the same during restore.
+			var netns syscall.Stat_t
+			err = syscall.Stat(nsPath, &netns)
+			if err != nil {
+				return err
+			}
+			criuExternal := fmt.Sprintf("net[%d]:%s", netns.Ino, path.Base(nsPath))
```
avagin: I think we should not use path.Base(nsPath) as a key, it doesn't contain any useful information and it probably can be changed, if we will restore a container on another host. I think we can use a constant string for this. For example, it can be "extRootNetNS".

libcontainer/container_linux.go
```diff
@@ -1150,6 +1180,37 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
...
+		err := c.checkCriuVersion(31100)
+		if err == nil {
+			// CRIU wants the information about an existing network namespace
+			// like this: --inherit-fd fd[<fd>]:<key>
+			// The <key> needs to be the same as during checkpointing.
+			// We are always using the path.Base(nsPath) as the key in this.
+			netns, err := os.Open(nsPath)
+			defer netns.Close()
+			if err != nil {
+				logrus.Error("If a specific network namespace is defined it must exist.")
```
avagin: I think we can print an error in this message
logrus.Error("If a specific network namespace is defined it must exist: %s", err)

libcontainer/container_linux.go
```diff
@@ -1150,6 +1180,37 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
...
+				logrus.Error(err)
+				return fmt.Errorf("Requested network namespace %v does not exist", nsPath)
+			}
+			inheritFd := new(criurpc.InheritFd)
+			inheritFd.Key = proto.String(path.Base(nsPath))
+			inheritFd.Fd = proto.Int32(int32(4 + len(extraFiles)))
```
avagin: We need to add a comment here, which explains what this 4 means. 0-2 is the standard file descriptors, 3 is a swrk socket.

adrianreber:
@avagin With your suggested changes force pushed.

crosbymichael:
LGTM

avagin:
LGTM

Issue #1851, opened 2018-07-26
[enhancement] Refactor ps.go to use psgo library


marcov:
Refactor `ps.go` to use the [psgo](http://github.com/containers/psgo) library instead of parsing output of the `ps`  command. 

This moves the complexity of parsing the `ps` command output outside of runc and has the advantage to extract a richer set of information related to the containers' processes than what's current available using `ps`.

PR #1852, opened 2018-07-26
Refactor ps.go to use psgo library

marcov:
* Refactor `ps.go` to use the psgo APIs instead of `exec.Command("ps", psArgs...)`
 * Add to runc ps the *--print-descriptors* boolean flag to retrieve all
   available format descriptors supported by psgo
 * Update tests and documentation
 * Add psgo to the vendored libraries (required by tests)

Fixes: #1851
Signed-off-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/man/runc-ps.8.md
+++ b/man/runc-ps.8.md
@@ -1,13 +1,21 @@
+% RUNC-PS(8) runc-ps Man Pages
+
 # NAME
-   runc ps - ps displays the processes running inside a container
+**runc** **ps** - ps displays the processes running inside a container
 
 # SYNOPSIS
-   runc ps [command options] <container-id> [ps options]
+**runc** **ps** [_command options_] <_container-id_> [_format descriptors_]
 
 # OPTIONS
-   --format value, -f value     select one of: table(default) or json
+**--format** _value_, **-f** _value_	select one of: _table_ (default) or _json_
+   
+**--print-descriptors**			print the list of supported format descriptors
+
+[_format descriptors_] is a list of space separated AIX style format descriptors. When no format descriptors are specified, **runc** **ps** will use a set of descriptors to simulate `ps -ef` output. These are: _user_, _pid_, _ppid_, _pcpu_, _etime_, _tty_, _time_, _comm_.
 
 The default format is table.  The following will output the processes of a container
 in json format:
+```
+ # runc ps -f json <container-id>
+```
 
-    # runc ps -f json <container-id>
--- a/ps.go
+++ b/ps.go
@@ -6,25 +6,36 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"os/exec"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
+	"github.com/containers/psgo"
+	"github.com/opencontainers/runc/libcontainer"
 	"github.com/urfave/cli"
 )
 
 var psCommand = cli.Command{
 	Name:      "ps",
 	Usage:     "ps displays the processes running inside a container",
-	ArgsUsage: `<container-id> [ps options]`,
+	ArgsUsage: `<container-id> [format descriptors]`,
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:  "format, f",
 			Value: "table",
 			Usage: `select one of: ` + formatOptions,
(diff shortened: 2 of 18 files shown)


man/runc-ps.8.md
```diff
@@ -1,13 +1,21 @@
+% RUNC-PS(8) runc-ps Man Pages
+
 # NAME
-   runc ps - ps displays the processes running inside a container
+**runc** **ps** - ps displays the processes running inside a container
 
 # SYNOPSIS
-   runc ps [command options] <container-id> [ps options]
+**runc** **ps** [_command options_] <_container-id_> [_format descriptors_]
```
AkihiroSuda: `runc ps` no longer accept `[ps options]`?
This would break compatibility.

marcov: That's correct, with the proposed change `runc ps` would not accept `[ps options]` anymore. This is because the `ps` command execution is replace with calls to the  psgo API, and psgo is not meant to be a clone of `ps`.

AkihiroSuda: can we fallback to the `ps` mode when a ps option is specified?

marcov: It looks like a good compromise. Should we fallbak to `ps` also when no options are specified?

@vrothberg WDYT?

vrothberg: I understand the request but I don't think that's very user-friendly and hard to explain (and hard to understand) when `runc-ps` would return which kind of output in which kind of format.

In case of podman we decided to break the compatibility with flags but gained the flexibility of extracting much more data that comes in handy when debugging (e.g., seccomp, security labels, capabilities, host pid/user/group, etc.).

I *think* that it's a good trade-off but I don't know if there are users in the wild relying on how `runc-ps` works now. As there's not yet a v1.0.0, we might be able to change it :)

cyphar: @AkihiroSuda Do you know of anyone using `runc ps -- -ef` or similar options? This is an issue I was aware of when I proposed doing this to @marcov, but given the history of this feature I'm not sure it's widely used.

(One other possible solution would be to take a look at how `ps` does the mapping of flags to fields and we could just do that formatting for users -- since `psgo` supports `ps`-like formatting flags.)

AkihiroSuda: cc @Random-Liu

vrothberg: Friendly ping.

cyphar: /cc @opencontainers/runc-maintainers 

Do we need to support ps flags? In which case would we be fine with having a dumb conversion from ps flags to the columns we need to ask psgo to give us?

lifubang: I think need to support PS flags, especially when there is no PS command in a container. You can see https://github.com/opencontainers/runc/pull/1879

cyphar:
This actually presents us with the possibility to fix a long-standing bug in `runc ps`. Currently `runc ps` assumes that the first field of `ps` is the PID -- but this is obviously not always the case. So currently:

```
% runc ps ctr -efZ
LABEL                           UID        PID  PPID  C STIME TTY          TIME CMD
unexpected pid 'avahi': strconv.ParseInt: parsing "avahi": invalid syntax
```

Which is a problem. I think that if we just implement ps-flag parsing and have some functions in `psgo` that do this, this will be easily mergeable. I will work on this.

AkihiroSuda:
What's current status?

cyphar:
@AkihiroSuda There isn't an implementation of the psflags yet. I spoke to @marcov about it yesterday and gave an outline of how it should be done. Hopefully one of us will have a few spare cycles to implement all of the parsing.

For context, the `ps` flag parsing code is **very** complicated and so hopefully we can implement this in a less awful manner...

AkihiroSuda:
Is this PR needed for v1.0?

cyphar:
No.

kolyshkin:
I think we won't use psgo in runc, and have to rely on a `ps` binary for backward compatibility.

cyphar:
Also in the intervening time psgo had some vulnerabilities that made me a little less comfortable using it.

There are some alternative options for converting pids that we could implement ourselves but I'm not sure how much effort it is worth.

PR #1861, opened 2018-08-08
WIP - Support for logging from children processes

marcov:
Add support for children processes logging (including nsexec).
A pipe is used to send logs from children to parent in JSON.
The JSON format used is the same used by logrus JSON formatted,
i.e. children process can use standard logrus APIs.

This improves debugging and can also be used to report warning messages from the children.

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -43,12 +43,20 @@ type parentProcess interface {
 	externalDescriptors() []string
 
 	setExternalDescriptors(fds []string)
+
+	getChildLogs()
+}
+
+type pipePair struct {
+	r *os.File
+	w *os.File
 }
 
 type setnsProcess struct {
 	cmd           *exec.Cmd
 	parentPipe    *os.File
 	childPipe     *os.File
+	logPipe       pipePair
 	cgroupPaths   map[string]string
 	intelRdtPath  string
 	config        *initConfig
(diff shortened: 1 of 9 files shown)


libcontainer/standard_init_linux.go
```diff
@@ -187,6 +188,10 @@ func (l *linuxStandardInit) Init() error {
 			return newSystemErrorWithCause(err, "init seccomp")
 		}
 	}
+
+	logPipe, _ := (logrus.StandardLogger().Out).(*(os.File))
```
cyphar: Two things:

* I'm not sure it's necessary to do `(a).(*os.File)` -- I think `a.(*os.File)` will work just as fine (and `gofmt` should complain at you.

* You shouldn't ignore the `ok` value here. Even if it is impossible to occur, you should either make it panic or you should check that `logPipe != nil` (which is the error case -- because that would also panic).

* Also we should probably set `logPipe` to be `O_CLOEXEC` which would remove the requirement to close it like this.

main.go
```diff
@@ -138,6 +138,12 @@ func main() {
 		updateCommand,
 	}
 	app.Before = func(context *cli.Context) error {
+
+		// do nothing if logrus was already initialized in init.go
+		if logrus.StandardLogger().Out != logrus.New().Out {
```
cyphar: I'm not really a fan of this -- why not set up logging for `init` here (or setting a global flag) rather than doing it this way?

libcontainer/process_linux.go
```diff
@@ -43,12 +43,20 @@ type parentProcess interface {
 	externalDescriptors() []string
 
 	setExternalDescriptors(fds []string)
+
+	getChildLogs()
```
cyphar: I don't really like this name -- maybe `forwardChildLogs` or `proxyChildLogs` to make it clear that this method actually pipes directly to logrus.

cyphar:
Closing in favour of #2034. Thanks @marcov for kicking this one off.

Issue #1837, opened 2018-07-03
Don't always enable rootless mode in userns


AkihiroSuda:
## Problem

In #1688 we broke "Docker-in-LXD":

```
$ lxc launch ubuntu:18.04 foo -c security.nesting=true
$ lxc shell foo
foo# curl -fsSL get.docker.com -o get-docker.sh && sh get-docker.sh
foo# exit
$ lxc file push /usr/local/sbin/runc foo/usr/bin/docker-runc
$ lxc shell foo
foo# cat /proc/self/uid_map
(we are in userns here)
foo# docker run -it --rm busybox
docker: Error response from daemon: OCI runtime create failed: cannot specify gid= mount options for unmapped gid in rootless containers: unknown.
```

This is caused because runc enables the "rootless mode" when running in user namespace, but "Docker-in-LXD" does not expect runc to enable the rootless mode.

## What "rootless mode" does actually

1. [Honor `$XDG_RUNTIME_DIR`](https://github.com/opencontainers/runc/blob/0e561642f81e84ebd0b3afd6ec510c75a2ccb71b/main.go#L70)
2. [Switch the cgroup manager to `libcontainer.RootlessCgroupfs`](https://github.com/opencontainers/runc/blob/dd56ece8236d6d9e5bed4ea0c31fe53c7b873ff4/utils_linux.go#L46)
3. [Disable cgroup-specific features such as `runc ps`](https://github.com/opencontainers/runc/blob/0e561642f81e84ebd0b3afd6ec510c75a2ccb71b/ps.go#L32) and [OOM notification](https://github.com/opencontainers/runc/blob/dd56ece8236d6d9e5bed4ea0c31fe53c7b873ff4/libcontainer/container_linux.go#L613)
4. [Disable `runc checkpoint`](https://github.com/opencontainers/runc/blob/0e561642f81e84ebd0b3afd6ec510c75a2ccb71b/checkpoint.go#L51) and [`runc restore`](https://github.com/opencontainers/runc/blob/0e561642f81e84ebd0b3afd6ec510c75a2ccb71b/restore.go#L103)
5. [Make sure `config.json` contains userns and id mappings if euid != 0](https://github.com/opencontainers/runc/blob/b222ea4469dd5e304f3f6cf7a2482860285639a2/libcontainer/configs/validate/rootless.go#L41)
[...]

PR #1862, opened 2018-08-10
Disable rootless mode except RootlessCgMgr when executed as the root in userns (fix Docker-in-LXD regression)

AkihiroSuda:
This PR decomposes  `libcontainer/configs.Config.Rootless bool` into `RootlessEUID bool` and
`RootlessCgroups bool`, so as to make "runc-in-userns" to be more compatible with "rootful" runc.

`RootlessEUID` denotes that runc is being executed as a non-root user (euid != 0) in
the current user namespace. `RootlessEUID` is almost identical to the former `Rootless`
except cgroups stuff.

`RootlessCgroups` denotes that runc is unlikely to have the full access to cgroups.
`RootlessCgroups` is set to false if runc is executed as the root (euid == 0) in the initial namespace.
Otherwise `RootlessCgroups` is set to true.
(Hint: if `RootlessEUID` is true, `RootlessCgroups` becomes true as well)

When runc is executed as the root (euid == 0) in an user namespace (e.g. by Docker-in-LXD, Podman, Usernetes),
`RootlessEUID` is set to false but `RootlessCgroups` is set to true.
So, "runc-in-userns" behaves almost same as "rootful" runc except that cgroups errors are ignored.

This PR does not have any impact on CLI flags and `state.json`.

Note about CLI:
* Now `runc --rootless=(auto|true|false)` CLI flag is only used for setting `RootlessCgroups`.
* Now `runc spec --rootless` is only required when `RootlessEUID` is set to true.
  For runc-in-userns, `runc spec`  without `--rootless` should work, when sufficient numbers of
  UID/GID are mapped.

Note about `$XDG_RUNTIME_DIR` (e.g. `/run/user/1000`):
* `$XDG_RUNTIME_DIR` is ignored if runc is being executed as the root (euid == 0) in the initial namespace, for backward compatibility.
  (`/run/runc` is used)
* If runc is executed as the root (euid == 0) in an user namespace, `$XDG_RUNTIME_DIR` is honored if `$USER != "" && $USER != "root"`.
  This allows unprivileged users to allow execute runc as the root in userns, without mounting writable `/run/runc`.

Note about `state.json`:
* `rootless` is set to true when `RootlessEUID == true && RootlessCgroups == true`.

- - -

* "Step 2" of https://github.com/opencontainers/runc/issues/1837
* Replaces https://github.com/opencontainers/runc/pull/1833 ("Step 1")
Fix https://github.com/opencontainers/runc/issues/1837  https://github.com/projectatomic/libpod/issues/1147

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -186,12 +186,19 @@ type Config struct {
 	// callers keyring in this case.
 	NoNewKeyring bool `json:"no_new_keyring"`
 
-	// Rootless specifies whether the container is a rootless container.
-	Rootless bool `json:"rootless"`
-
 	// IntelRdt specifies settings for Intel RDT/CAT group that the container is placed into
 	// to limit the resources (e.g., L3 cache) the container has available
 	IntelRdt *IntelRdt `json:"intel_rdt,omitempty"`
+
+	// RootlessEUID is set when the runc was launched with non-zero EUID.
+	// Note that RootlessEUID is set to false when launched with EUID=0 in userns.
+	// When RootlessEUID is set, runc creates a new userns for the container.
+	// (config.json needs to contain userns settings)
+	RootlessEUID bool `json:"rootless_euid,omitempty"`
+
+	// RootlessCgroups is set when unlikely to have the full access to cgroups.
+	// When RootlessCgroups is set, cgroups errors are ignored.
+	RootlessCgroups bool `json:"rootless_cgroups,omitempty"`
 }
 
 type Hooks struct {
--- a/rootless_linux.go
+++ b/rootless_linux.go
@@ -1,0 +1,57 @@
+// +build linux
+
+package main
+
+import (
+	"os"
+
+	"github.com/opencontainers/runc/libcontainer/system"
+	"github.com/urfave/cli"
+)
+
+func shouldUseRootlessCgroupManager(context *cli.Context) (bool, error) {
+	if context != nil {
+		b, err := parseBoolOrAuto(context.GlobalString("rootless"))
+		if err != nil {
+			return false, err
+		}
+		if b != nil {
+			return *b, nil
+		}
+		// nil b stands for "auto detect"
+	}
+	if context.GlobalBool("systemd-cgroup") {
+		return false, nil
+	}
+	if os.Geteuid() != 0 {
+		return true, nil
+	}
+	if !system.RunningInUserNS() {
+		// euid == 0 , in the initial ns (i.e. the real root)
+		return false, nil
+	}
+	// euid = 0, in a userns.
+	// As we are unaware of cgroups path, we can't determine whether we have the full
+	// access to the cgroups path.
+	// Either way, we can safely decide to use the rootless cgroups manager.
+	return true, nil
+}
+
+func shouldHonorXDGRuntimeDir() bool {
+	if os.Getenv("XDG_RUNTIME_DIR") == "" {
+		return false
+	}
+	if os.Geteuid() != 0 {
+		return true
+	}
+	if !system.RunningInUserNS() {
+		// euid == 0 , in the initial ns (i.e. the real root)
+		// in this case, we should use /run/runc and ignore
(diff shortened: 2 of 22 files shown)


rhatdan:
@giuseppe PTAL

libcontainer/configs/config.go
```diff
@@ -186,12 +186,19 @@ type Config struct {
...
 	NoNewKeyring bool `json:"no_new_keyring"`
 
-	// Rootless specifies whether the container is a rootless container.
-	Rootless bool `json:"rootless"`
-
 	// IntelRdt specifies settings for Intel RDT/CAT group that the container is placed into
 	// to limit the resources (e.g., L3 cache) the container has available
 	IntelRdt *IntelRdt `json:"intel_rdt,omitempty"`
+
+	// RootlessEUID specifies is set when the runc was launched with non-zero EUID.
```
giuseppe: please drop "specifies" or "is set"

rootless_linux.go
```diff
@@ -0,0 +1,57 @@
...
+	if !system.RunningInUserNS() {
+		// uid == 0 , in the initial ns (i.e. the real root)
+		// in this case, we should use /run/runc and ignore
+		// $XDG_RUNTIME_DIR (e.g. /run/user/0) for backward
+		// compatibility.
+		return false
+	}
+	// euid = 0, in a userns.
+	u := os.Getenv("USER")
+	return u != "" && u != "root"
```
giuseppe: could this be something like?

```
	u, ok := os.LookupEnv("USER")
	return !ok || u != "root"
```

Otherwise it breaks current users like Podman (and probably Buildah as well) that don't pass `$USER` down to the runtime

libcontainer/configs/config.go
AkihiroSuda: updated

rootless_linux.go
AkihiroSuda: updated

kolyshkin:
this needs a rebase

AkihiroSuda:
rebased

AkihiroSuda:
@giuseppe LGTY?

giuseppe:
@AkihiroSuda yes, I've tested it with podman and it works well for me

giuseppe:
an error I've seen (not introduced by this patch but that could probably be fixed as part of it), is:

create a rootless user namespace with multiple IDs mapped in, then create a runc container that has:

```
    "uidMappings": [
      {
        "hostID": 1,
        "containerID": 0,
        "size": 1
      },
      {
        "hostID": 0,
        "containerID": 1,
        "size": 1
      }
    ],
    "gidMappings": [
      {
        "hostID": 1,
        "containerID": 0,
        "size": 1
      },
      {
        "hostID": 0,
        "containerID": 1,
        "size": 1
      }
    ],
```

if I try to exec (from the first user namespace) into the container, I get:

```
$ runc exec foo echo hi
exec failed: container_linux.go:337: starting container process caused "process_linux.go:90: adding pid 5652 to cgroups caused \"failed to write 5652 to cgroup.procs: write /sys/fs/cgroup/systemd/user.slice/user-1000.slice/user@1000.service/gnome-terminal-server.service/foo/cgroup.procs: permission denied\""

```

AkihiroSuda:
@giuseppe should be fixed in the latest revision

AkihiroSuda:
ping @cyphar

AkihiroSuda:
This PR seems also needed for Docker-on-Chromebook https://www.reddit.com/r/Crostini/comments/9jabhq/docker_now_working/

caniszczyk:

AkihiroSuda:
This is being discussed at the dev ML: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/jOaNYre5M40

crosbymichael:
I'm reviewing and testing this today

crosbymichael:
LGTM

mrunalp:
I'll review this today.

mrunalp:
LGTM

PR #1865, opened 2018-08-12
Makefile: allow BUILDTAGS to be overridden

liamstask:
This capability is explicitly mentioned in ReadMe.md but currently does not work


tianon:
I'm able to successfully override `BUILDTAGS` via `make BUILDTAGS=xxx` -- can you maybe describe what you've tried in more detail?

AkihiroSuda:
It seems we already have this fix: https://github.com/opencontainers/runc/blob/981dbef514940c6ce90e8d277ffbefced785271f/Makefile#L16

PR #1869, opened 2018-08-16
rootless: make /sys/fs/cgroup/* read-only

AkihiroSuda:
The default rootless spec bind-mounts /sys with "rbind,ro" (because mounting sysfs
requires netns to be unshared), however, it does not make subfilesystems mounted
under /sys read-only.

So, files under /sys/fs/cgroup were unexpectedly writable when they are chmod/chowned
via privileged helpers such as pam_cgfs.

This patch fixes the issue by mounting /sys/fs/cgroup/* as read-only explicitly.

### DIFF
--- a/libcontainer/specconv/example.go
+++ b/libcontainer/specconv/example.go
@@ -2,8 +2,11 @@ package specconv
 
 import (
 	"os"
+	"path/filepath"
+	"sort"
 	"strings"
 
+	"github.com/opencontainers/runc/libcontainer/mount"
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
@@ -189,13 +192,10 @@ func ToRootless(spec *specs.Spec) {
 	}}
 
 	// Fix up mounts.
-	var mounts []specs.Mount
-	for _, mount := range spec.Mounts {
-		// Ignore all mounts that are under /sys.
-		if strings.HasPrefix(mount.Destination, "/sys") {
-			continue
-		}
-
+	mountInfo, _ := mount.GetMounts()
+	fixupRootlessBindSys(spec, mountInfo)
+	for i := range spec.Mounts {
+		mount := &spec.Mounts[i]
 		// Remove all gid= and uid= mappings.
 		var options []string
 		for _, option := range mount.Options {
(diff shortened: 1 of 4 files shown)


libcontainer/specconv/example.go
```diff
@@ -189,33 +193,66 @@ func ToRootless(spec *specs.Spec) {
...
+	}
+	mounts = append(mounts, []specs.Mount{
+		// Add the sysfs mount as an rbind.
+		// Note:
+		// * "ro" does not make submounts read-only recursively.
+		// * rbind work for sysfs but bind does not.
+		{
+			Source:      "/sys",
+			Destination: "/sys",
+			Type:        "none",
```
giuseppe: should this be "bind"?

AkihiroSuda: why?

giuseppe: this is the only place where I see it used.  In the runtime-specs example `bind` is used as the type for a bind mount (https://github.com/opencontainers/runtime-spec/blob/master/config.md#example-linux), both Docker and Podman generate "bind".

Anyway since it is only in the example and not in the specs, and you are just moving this code snippet, it is probably fine to leave it as it is.

libcontainer/specconv/example.go
```diff
@@ -162,6 +162,10 @@ func ToRootless(spec *specs.Spec) {
 	var namespaces []specs.LinuxNamespace
 
 	// Remove networkns from the spec.
+	//
+	// TODO: removal of networkns should be optional,
```
crosbymichael: FYI.  When someone puts TODOs in PRs, it makes reviewers nervous.  

todo when?
by who? 
what is the issue now? 
what breaks while this isn't done?

giuseppe:
Thanks, LGTM

libcontainer/specconv/example.go
cyphar: 'type: "none"` is entirely valid. There was a bug previous in `runc` when we didn't handle this properly, but the type field (when it comes to `mount(2)`) doesn't affect bindmounts -- what matters is that the `bind` option is set.

libcontainer/specconv/example.go
```diff
@@ -189,33 +189,66 @@ func ToRootless(spec *specs.Spec) {
...
 }
+
+// fixupRootlessBindSys bind-mounts /sys.
+// This is required when netns is not unshared.
+func fixupRootlessBindSys(spec *specs.Spec) {
+	// Fix up mounts.
+	var mounts []specs.Mount
+	for _, mount := range spec.Mounts {
+		// Ignore all mounts that are under /sys.
+		if strings.HasPrefix(mount.Destination, "/sys") {
```
cyphar: I'm not sure if this is _always_ fine -- if someone wants to mount a fake filesystem over parts of `/sys` then I don't see why we would block it (let alone ignore it).

Ultimately this is used during config generation so it's not the end of the world, but it is something to keep in mind.

libcontainer/specconv/example.go
```diff
@@ -189,33 +189,66 @@ func ToRootless(spec *specs.Spec) {
...
+	}...)
+	spec.Mounts = mounts
+	var maskedPaths []string
+	for _, masked := range spec.Linux.MaskedPaths {
+		// Ignore masked paths that are under /sys. (should be safe for rootless mode)
+		//
+		// Without this hack, `df` fails with "df: /sys/firmware/efi/efivars"
+		// because the efivars entry exists in mtab (because /sys is rbind-mounted)
+		// but /sys/firmware is masked.
+		if strings.HasPrefix(masked, "/sys") {
```
cyphar: I'm not sure this is fine for rootless mode -- if you unmask paths you could be leaking information to containers (so while they might not be able to write to the masked paths, they are able to read them). I mean, you probably can't read most of `/sys` that is masked anyway, but I'm not sure it's a great idea.

AkihiroSuda:
addressed comments, although it became more complicated

AkihiroSuda:
@giuseppe latest revision still LGTY?

giuseppe:
@AkihiroSuda yes, latest version LGTM

AkihiroSuda:
@cyphar PTAL

AkihiroSuda:
ping @cyphar

AkihiroSuda:
On second thought maybe we should rather suggest users not mounting `/sys` when rootless && netns is not unshared.

Rootless BuildKit already removed `/sys` mount in https://github.com/moby/buildkit/pull/689

For Rootless Moby, `sysfs` mount (not bind-mount) is preserved, because netns is unshared via RootlessKit before starting `dockerd`: 
https://github.com/moby/moby/pull/38050

I'm going to close this PR but I may change my mind later.
RFC. @giuseppe @cyphar @jessfraz

PR #1871, opened 2018-08-17
cr: don't restore net namespace by default

Ace-Tang:
since runc don't manage net device and their configuration, checkpoint
also don't dump net namespace by default, so set 'nsmask = unix.CLONE_NEWNET'
by default in restore. Or if user do not pass 'empty-ns network', criu will
cost extra time in restore.

### DIFF
--- a/restore.go
+++ b/restore.go
@@ -109,6 +109,9 @@ using the runc checkpoint command.`,
 			return err
 		}
 		options := criuOptions(context)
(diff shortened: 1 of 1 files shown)


avagin:
LGTM, thanks!

crosbymichael:
Is there a way to tell criu to use an external namespace? Just trying to think how this all fits together.

crosbymichael:
LGTM

PR #496, opened 2016-01-21
Remove usage of GetMounts from GetCgroupMounts

LK4D4:
GetMounts is very cpu-expensive. I'll change other funcs in this package
to reuse code from GetCgroupMounts later.

PR #1872, opened 2018-08-17
Respect container's cgroup path

ostenbom:
Hi there!

We have a testing environment where we have multiple instances of garden/runc running, where each instance creates and destroys containers for its tests. Each of these testing nodes are run in isolation on the same host, so each have their own cgroup directories to minimise interference.

We have noticed that occasionally, when runc is trying to write properties to a cgroup, it chooses a neighbours cgroup by accident.

We think this is to do with the implementation of [`FindCgroupMountpointAndRoot`](https://github.com/opencontainers/runc/blob/20aff4f0488c6d4b8df4d85b4f63f1f704c11abd/libcontainer/cgroups/utils.go#L30), which finds the *first* occurrence of a subsystem in the mount table. When we run multiple instances in parallel, this is not always the right one.

We'd like to improve this function to respect the containers cgroup path, which we can find from the containers spec.

Looking forward to your feedback,

Cheers, the garden team.

Signed-off-by: Danail Branekov <danailster@gmail.com>
Signed-off-by: Oliver Stenbom <ostenbom@pivotal.io>

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -22,31 +22,41 @@ const (
 )
 
 // https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
-func FindCgroupMountpoint(subsystem string) (string, error) {
-	mnt, _, err := FindCgroupMountpointAndRoot(subsystem)
+func FindCgroupMountpoint(cgroupPath, subsystem string) (string, error) {
+	mnt, _, err := FindCgroupMountpointAndRoot(cgroupPath, subsystem)
 	return mnt, err
 }
 
-func FindCgroupMountpointAndRoot(subsystem string) (string, string, error) {
+func FindCgroupMountpointAndRoot(cgroupPath, subsystem string) (string, string, error) {
 	// We are not using mount.GetMounts() because it's super-inefficient,
 	// parsing it directly sped up x10 times because of not using Sscanf.
 	// It was one of two major performance drawbacks in container start.
 	if !isSubsystemAvailable(subsystem) {
 		return "", "", NewNotFoundError(subsystem)
 	}
+
 	f, err := os.Open("/proc/self/mountinfo")
 	if err != nil {
 		return "", "", err
 	}
 	defer f.Close()
 
-	scanner := bufio.NewScanner(f)
+	return findCgroupMountpointAndRootFromReader(f, cgroupPath, subsystem)
+}
+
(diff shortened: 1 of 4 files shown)


crosbymichael:
LGTM

ostenbom:
Bump, need another LGTM or some feedback here!

BooleanCat:
@cyphar WDYT?

libcontainer/cgroups/utils.go
```diff
@@ -22,37 +22,41 @@ const (
...
 	}
-	f, err := os.Open("/proc/self/mountinfo")
+
+	mountinfo, err := ioutil.ReadFile("/proc/self/mountinfo")
 	if err != nil {
 		return "", "", err
 	}
-	defer f.Close()
 
-	scanner := bufio.NewScanner(f)
```
cyphar: My main concern with this change is that it switches away from `bufio` to storing the entire contents of `/proc/self/mountinfo` in memory. From memory, we switched away from doing this a long time ago in #496 because of performance problems -- I think effectively reverting this change without a solid justification is probably less-than-ideal.

gcapizzi:
@cyphar we've switched back to a `Scanner` based solution, how does it look?

cyphar:
LGTM.

Last call for 1.0 merges!

cyphar:
LGTM.

Last call for 1.0 merges!

cyphar:
Dammit, PullApprove is broken again -- I guess they want to stop us from releasing. :wink:

/cc @caniszczyk

cyphar:
LGTM

crosbymichael:
LGTM

libcontainer/cgroups/utils.go
```diff
@@ -22,31 +22,41 @@ const (
...
+
+func findCgroupMountpointAndRootFromReader(reader io.Reader, cgroupPath, subsystem string) (string, string, error) {
+	scanner := bufio.NewScanner(reader)
 	for scanner.Scan() {
 		txt := scanner.Text()
-		fields := strings.Split(txt, " ")
-		for _, opt := range strings.Split(fields[len(fields)-1], ",") {
-			if opt == subsystem {
-				return fields[4], fields[3], nil
+		fields := strings.Fields(txt)
```
kolyshkin, line 51: Replacing `strings.Split` with `strings.Fields` is a noticeable performance hit. This is why we need benchmarks.

(this is no longer a problem, just looking through older changes)

PR #1873, opened 2018-08-20
When doing a copyup, /tmp can not be a shared mount point

rhatdan:
MOVE_MOUNT will fail under certain situations.

You are not allowed to MS_MOVE if the parent directory is shared.

```
man mount
...
   The move operation
       Move a mounted tree to another place (atomically).  The call is:

              mount --move olddir newdir

       This  will cause the contents which previously appeared under olddir to
       now be accessible under newdir.  The physical location of the files  is
       not changed.  Note that olddir has to be a mountpoint.

       Note  also that moving a mount residing under a shared mount is invalid
       and unsupported.  Use findmnt -o TARGET,PROPAGATION to see the  current
       propagation flags.
```

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -152,6 +152,26 @@ func finalizeRootfs(config *configs.Config) (err error) {
 	return nil
 }
 
+// /tmp has to be mounted as private to allow MS_MOVE to work in all situations
+func prepareTmp(topTmpDir string) (string, error) {
+	tmpdir, err := ioutil.TempDir(topTmpDir, "runctop")
+	if err != nil {
+		return "", err
+	}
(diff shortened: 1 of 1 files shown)


rhatdan:
I did not find a "tmp" directory that belonged to runc at this low of a level, so  I created a new tmpdir for each copyup.  Since this will probably not happen many times, it should not be a big issue.  If  someone can suggest  a better way to handle the libcontainer state.  So we would only create the parent directory once, I would love to see it.

rhatdan:
@calavera @avagin @caniszczyk @mrunalp @crosbymichael @cyphar @dqminh @hqhq @rjnagal PTAL

mrunalp:
LGTM

rhatdan:
@calavera @avagin @caniszczyk @mrunalp @crosbymichael @cyphar @dqminh @hqhq @rjnagal 
Can I get anone to comment/review this?

hqhq:
LGTM

PR #1874, opened 2018-08-22
Remove unused veth setup code

mrunalp:
Networking is setup by plugins for users of runc so it makes sense
to get rid of the veth strategy.


cyphar:
LGTM. I love diffs that are pure red.

crosbymichael:
LGTM

PR #1876, opened 2018-08-27
Modify empty value handling of global options --log

KentaTada:
runc cannot handle empty arguments of global option "--log" correctly. For example, "runc --log --debug run con1" will create the log file named "--debug" and I think it should not be assumed.
So, this commit modifies empty arguments handling for global option "--log"  to use "/dev/null" as runc help specified.

### DIFF
--- a/main.go
+++ b/main.go
@@ -142,6 +142,14 @@ func main() {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
 		if path := context.GlobalString("log"); path != "" {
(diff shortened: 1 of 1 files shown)


crosbymichael:
Either we have a bug in the cli lib or you are calling it wrong and it's user error.  I don't think we should have this type of handling.  We can try updating the cli package to see if they have any fixes for this but the flag parsing is working correctly in my opinion

PR #1879, opened 2018-08-28
fix: may display pids in other containers

lifubang:
Signed-off-by: Lifubang <lifubang@aliyun.com>

This bug will appear in a very small probability.

If there is a bash script for start a runc container:
bash file: st
```
#!/bin/bash
redis-server --port $1
```

After I start two runc container.
```
root@dockerdemo:/opt/runctest/redis# runc ps redis0
root     28808     1  0 16:23 ?        00:00:00 /bin/bash /st 28609
root     28837 28808  0 16:23 ?        00:00:05 redis-server *:28609
root@dockerdemo:/opt/runctest/redis# runc ps redis1
root     28874     1  0 16:23 ?        00:00:00 /bin/bash /st 28808
root     28904 28874  0 16:23 ?        00:00:05 redis-server *:28808
```

If redis1's port (28808) is just equals to redis0's pid(28808). There'll be a bug.
```
root@dockerdemo:/opt/runctest/redis# runc ps redis0 -C st -o cmd,ppid,pid
CMD                          PPID   PID
/bin/bash /st 28808             1 28874
```
The PID 28874 is owner to redis1, not redis0.
If I kill pid 28874, something will be wrong.

This is because strings.Fields is just split by white space, If there is a white space in cmd, there will be error. So, we need to compare the pid with a strict mode.

The other solution is to limit [ps options] of user's runc ps command. But I think it is not convenient for users.


cyphar:
This is a perfect opportunity to plug #1852 which would fix this issue, and many others.

lifubang:
It's a good job for #1852, but if he can compatible with ps command, it will be better, especially when there were no ps command in the container.

lifubang:
refer to https://salsa.debian.org/debian/procps/blob/master/ps/parser.c
psarglexer.go is a small lexer to analysis ps args. So runc ps can compatible with "ps" command safely.

PR #1880, opened 2018-08-28
libcontainer: CurrentGroupSubGIDs -> CurrentUserSubGIDs

AkihiroSuda:
subgid is defined per user, not group (see [`subgid(5)`](http://man7.org/linux/man-pages/man5/subgid.5.html))

This commit also adds support for specifying subuid owner with a numeric UID.

### DIFF
--- a/libcontainer/user/lookup_unix.go
+++ b/libcontainer/user/lookup_unix.go
@@ -5,6 +5,7 @@ package user
 import (
 	"io"
 	"os"
+	"strconv"
 
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Note that these functions are not used by runc itself, and looks like only consumed by RootlessKit: https://github.com/search?q=currentgroupsubgids&type=Code

Maybe we can just remove these functions instead.

libcontainer/user/lookup_unix.go
```diff
@@ -115,22 +116,23 @@ func CurrentGroup() (Group, error) {
...
 	}
-	return ParseSubIDFileFilter("/etc/subuid",
-		func(entry SubID) bool { return entry.Name == u.Name })
+	filter := func(entry SubID) bool {
+		return entry.Name == u.Name || entry.Name == strconv.Itoa(u.Uid)
+	}
+	return ParseSubIDFileFilter(fileName, filter)
 }
 
-func CurrentGroupSubGIDs() ([]SubID, error) {
```
crosbymichael: is this going to break anyone?

AkihiroSuda: AFAIK no https://github.com/opencontainers/runc/pull/1880#issuecomment-416763955

AkihiroSuda:
@cyphar PTAL?

cyphar:
LGTM.

crosbymichael:
LGTM

Issue #1415, opened 2017-04-22
Weirdness in TTY


jwendell:
1. Create a terminal-less container (`runc create ...`)
2. In another terminal, start it (`runc start ...`)
3. Go back to terminal 1 and press <ENTER> some times
4. See the weirdness

Notes:
1. I'm running the example from README (busybox image, "sh" entry point)
2. After start the container (step 2 above) I see in the first terminal: `sh: can't access tty; job control turned off`

cyphar:
@jwendell If you've disabled terminal creation (`"terminal": false` in your configuration) this is to be expected. What's happened is that your container has inherited its IO file descriptors from your shell, and your shell is still using them. So when you enter anything into your shell there is a race between your shell and the container process that determines which process reads the bytes.

There are several ways to use runc and runc accomodates all of them, but if you use the wrong combination it will cause you problems. To be fair this is not a well-documented area because of how much stuff has changed in the past few months and we're still not sure if the current system is the best way of  doing things.

`terminal=false` is an unfortunate option, but is generally used in cases where the parent process of the container has already set up the IO of the container and doesn't want `runc` to touch it. `runc create`/`runc start` (or `runc run -d` which is sort of like a combined version of those commands) results in the runc process no longer being alive afterwards. And due to the semantics of double-forking your shell will interpret that as "the child has finished running, it's my job to continue running now". **If you're using just a raw `runc create` from a shell you're most likely doing something wrong.** You should either use it with `--console-socket`, or set the stdio of `runc create` to files that you want to be linked to the container's stdio.

`terminal=true` is the nicer option to use, because it tells runc to create a new terminal inside the container's `devpts` setup. This results in much more sane behaviour from a variety of tools and it better emulates how a shell on a normal system acts. Unfortunately most users hit problems with it when they (mistakenly) try to use `runc create` in a shell with `terminal=true`. This won't work (unless you provide `--console-socket` which then requires you to use a separate program to capture the master `pty` file descriptor) and causes a lot of confusion.

Does that better explain things? As I mentioned in other threads, this really needs better documentation and is on [my backlog of things to do](https://todo.cyphar.com/b/EcHgEG5XXp3PSnsHQ/oci/6sYCpfjfJojmSG2Lz).

untoreh:
@cyphar 
> You should either use it with --console-socket, or set the stdio of runc create to files that you want to be linked to the container's stdio.

that is `runc run -d 1>out 2>err` works with `terminal=false` 
[...]

PR #1881, opened 2018-08-30
fix host's bash crash when create & start a busybox container with terminal=false

lifubang:
When create and start a container with terminal=false but container's command needs a tty, the host's bash ui will be crashed.
After start, the hosts bash can't input anything and became unreadable.  Please see #1415
Even delete this container in another window, this bash can't recover.

I think this is a bug. The best solution is to change tty allocate from CREATE to START. But it's very difficult. The easiest way is preventing inherit os.Stdin to these types of containers, because you do need a tty when you need interact with the container.

Command history:
```
root@dockerdemo:/opt/runctest/busybox# runc list
ID          PID         STATUS      BUNDLE                CREATED                          OWNER
redis       22444       running     /opt/runctest/redis   2018-08-30T11:27:26.724172165Z   root
```
```
root@dockerdemo:/opt/runctest/busybox# runc create busybox
root@dockerdemo:/opt/runctest/busybox# runc list
ID          PID         STATUS      BUNDLE                  CREATED                          OWNER
busybox     23419       created     /opt/runctest/busybox   2018-08-30T11:38:53.9521968Z     root
redis       22444       running     /opt/runctest/redis     2018-08-30T11:27:26.724172165Z   root
```
```
root@dockerdemo:/opt/runctest/busybox# runc start busybox
sh: can't access tty; job control turned off
/ # root@dockerdemo:/opt/runctest/busybox# [4;5RRRRR: command not found
root@dockerdemo:/opt/runctest/busybox# 
sh: [4: not found
sh: 5: not found
/ # [4;5Rroot@dockerdemo:/opt/runctest/busybox# 
sh: [4: not found
sh: 5R: not found
/ # -bash: syntax error near unexpected token `;'
root@dockerdemo:/opt/runctest/busybox# 
/ # [4;5Rroot@dockerdemo:/opt/runctest/busybox# lsNetid S
```

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -71,6 +71,12 @@ func inheritStdio(process *libcontainer.Process) error {
 	return nil
 }
 
+func inheritStdioWithoutIn(process *libcontainer.Process) error {
+	process.Stdout = os.Stdout
+	process.Stderr = os.Stderr
+	return nil
+}
+
 func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) error {
 	f, err := utils.RecvFd(socket)
 	if err != nil {
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -151,7 +151,7 @@ func destroy(container libcontainer.Container) {
 }
 
 // setupIO modifies the given process config according to the options.
-func setupIO(process *libcontainer.Process, rootuid, rootgid int, createTTY, detach bool, sockpath string) (*tty, error) {
+func setupIO(process *libcontainer.Process, rootuid, rootgid int, act CtAct, createTTY, detach bool, sockpath string) (*tty, error) {
 	if createTTY {
 		process.Stdin = nil
 		process.Stdout = nil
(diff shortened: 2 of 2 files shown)


crosbymichael:
I'm not sure about this one, i'll have to look more

lifubang:
Just like docker run without -i -t: 
docker run --name=shtest --entrypoin
=sh docker.acmcoder.com/public/busybox

It doesn't inherit is.stdin

lifubang:
Furthermore, I think `runc create` shouldn't run `runc init`, it just need to create necessary dir and files. Just like `docker create`, `ctr c creat` . Then every step will be more easy.
BTW, `runc create` create init process waiting exec.fifo file. It is fragile. If I run `cat exec.fifo`, the container will start automatically without run `runc start`.

crosbymichael:
No, runc is not docker and it will be different.  We need runc to function this way so things like docker can do what it needs to do.

lifubang:
yes, you are right.

PR #1883, opened 2018-08-31
fix delete other file bug when container id is ..

lifubang:

There is a big bug which can delete files not belonging to the container when use runc delete --force.
Command history is:

/opt/runc/runcroot is ruc's root dir.
```
root@dockerdemo:/opt/runc# ls -lh
total 16K
drwxr-xr-x 2 root root 4.0K Aug 31 11:19 dbback
drwx------ 2 root root 4.0K Aug 31 10:49 runcroot
-rw-r--r-- 1 root root    5 Aug 31 11:19 test
-rw-r--r-- 1 root root    6 Aug 31 11:19 test1
```
While **dbback dir** have very important db backup files.

run two container:
```
root@dockerdemo:/opt/runctest/redis# runc --root /opt/runc/runcroot run -d redis
root@dockerdemo:/opt/runctest/redis# 6:C 31 Aug 03:24:24.741 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo

root@dockerdemo:/opt/runctest/redis# runc --root /opt/runc/runcroot run -d ..0
root@dockerdemo:/opt/runctest/redis# 6:C 31 Aug 03:24:33.669 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo

root@dockerdemo:/opt/runctest/redis# runc --root /opt/runc/runcroot list
ID          PID         STATUS      BUNDLE                CREATED                          OWNER
..0         4170        running     /opt/runctest/redis   2018-08-31T03:24:33.664173003Z   root
redis       4121        running     /opt/runctest/redis   2018-08-31T03:24:24.73616634Z    root
root@dockerdemo:/opt/runctest/redis#
```

backup a state.json to /opt/runc
```
root@dockerdemo:/opt/runctest/redis# runc --root /opt/runc/runcroot list
ID          PID         STATUS      BUNDLE                CREATED                          OWNER
..0         4170        running     /opt/runctest/redis   2018-08-31T03:24:33.664173003Z   root
redis       4121        running     /opt/runctest/redis   2018-08-31T03:24:24.73616634Z    root
root@dockerdemo:/opt/runctest/redis# cp /opt/runc/runcroot/redis/state.json /opt/runc/
root@dockerdemo:/opt/runctest/redis# ls /opt/runc
dbback  runcroot  state.json  test  test1
root@dockerdemo:/opt/runctest/redis#
```

When I want to delete ..0, but I typed a wrong word: ..0 -> .. lost a 0
```
root@dockerdemo:/opt/runctest/redis# runc --root /opt/runc/runcroot delete --force ..
root@dockerdemo:/opt/runctest/redis# runc --root /opt/runc/runcroot list
root@dockerdemo:/opt/runctest/redis# ls /opt/runc
runcroot
root@dockerdemo:/opt/runctest/redis# ls /opt/runc/runcroot/
root@dockerdemo:/opt/runctest/redis#
```
**Everything is deleted, include my dbback dir.**

This is because there is no strict id validate method in Container Load function in "libcontainer/factory_linux.go", especially when delete.

Please check it, thanks.

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -359,23 +359,7 @@ func (l *LinuxFactory) loadState(root, id string) (*State, error) {
 }
 
 func (l *LinuxFactory) validateID(id string) error {
-	if !idRegex.MatchString(id) || id == ".." || id == "." {
-		return newGenericError(fmt.Errorf("invalid id format: %v", id), InvalidIdFormat)
-	}
-
-	//For unforeseen invalid id situations, can checked by is SubDir?
-	rootPath, err := filepath.Abs(l.Root)
-	if err != nil {
-		return err
-	}
-
-	containerRoot := filepath.Join(l.Root, id)
-	rootCheckPath, err := filepath.Abs(filepath.Join(containerRoot, ".."))
-	if err != nil {
-		return err
-	}
-
-	if rootPath != rootCheckPath {
+	if id == "." || !idRegex.MatchString(id) || utils.CleanPath(id) != id {
 		return newGenericError(fmt.Errorf("invalid id format: %v", id), InvalidIdFormat)
 	}
 
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -11,6 +11,7 @@ import (
 	"runtime/debug"
 	"strconv"
 
+	"github.com/cyphar/filepath-securejoin"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fs"
 	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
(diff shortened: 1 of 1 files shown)


cyphar:
Jesus. Yeah, that's a very bad bug.

I'm not a fan of this patch (it looks a bit overly complicated -- `filepath.*` validation is quite hard to get right). I would recommend doing something like checking whether `utils.CleanPath(containerid) == containerd` and erroring out if that's not the case.

However there are more places that this needs to be fixed -- most of `libcontainer/factory_linux.go` appears to use `filepath.Join` in an unsafe way.

lifubang:
@cyphar Thanks for your apply.
utils.CleanPath(containerid) is still unsafe for containerid. For example:
runc --root /opt/runc/runcroot delete --force .

Because utils.CleanPath(".") = "."

So, I said:
//For other unforeseen invalid id situations, can checked by is SubDir?
Because you don't know what is other unforeseen invalid id situations.

lifubang:
@cyphar There are 3 filepath.join in libcontainer/factory_linux.go
1. https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L198 Because there is validateID before filepath.join, so it is safe.
2. https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L232 There is no validateID before filepath.join, so I add it.
3. https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L342 Because stateFilename is a const, so it is safe.
Please check it.

cyphar:
@lifubang 

You could fix that by doing

```
string(os.Separator)+containerId == utils.CleanPath(string(os.Separator)+containerid)
```

My reason for wanting to do it this way is that it makes it clearer that we are making sure that the name doesn't contain any special lexical path tokens. However, since container IDs must match a regular expression that doesn't include `/` we could make it even easier and just disallow container IDs that are equal to `.` or `..` (without any `filepath` checks). We could even add that to the regular expression if we wanted to.

crosbymichael:
LGTM

PR #1888, opened 2018-09-07
libcontainer: fix InitArgs options func to set factory.InitPath

nickethier:
Greetings!

From the godocs it looks like `InitArgs` should be setting the full path and args for the init process:
>InitArgs returns an options func to configure a LinuxFactory with the provided init binary **path and arguments.**

This isn't the case so here is a PR to fix that + a unit test.

Thanks!
-Nick

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -452,8 +452,7 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 }
 
 func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.Cmd, error) {
-	cmd := exec.Command(c.initPath, c.initArgs[1:]...)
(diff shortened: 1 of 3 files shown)


BooleanCat:
I don't think we should be changing the InitPath to something other than /proc/self/exe. How else would runc syncronise containerisation of the init process without this unless you provide another binary that can syncronize in the same way?

From my reading of runc's code the flow looks like:

1. Set InitPath to `/proc/self/exe`, set the args to [`init`, ... other args]
2. runc invokes a child process with `/proc/self/exe init other args`
3. runc and the child communication syncronisation for a bit while namespaces and such are set up
4. the process defined in `other_args[0]` is execed

I *think* the intention of `InitArgs` is to enable you to change the init process that's eventually execed by the runc init child. Modying InitPath seems like a separate concern - herhaps the issue here is simply misleading wording in the comment?

PR #1891, opened 2018-09-14
clean mountpoint path

Ace-Tang:
clean mountpoint source and destination path in 'createLibcontainerMount'.

### why I do this

1. clean path seems reasonable in linux, produce a canonicalized pathname. (I remove the absolute word, it was add by mistake, I do not want to resolve path to absolute path, but only do clean)

2. actually, I failed to restore a checkpointed container caused I passed a pathname like '/foo/bar/' in -v,  note I am not using docker(docker will do the path clean in the high level). 
https://github.com/checkpoint-restore/criu/issues/547

Please check if the modification make sense !

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -280,8 +280,8 @@ func createLibcontainerMount(cwd string, m specs.Mount) *configs.Mount {
 	}
 	return &configs.Mount{
 		Device:           device,
-		Source:           source,
-		Destination:      m.Destination,
+		Source:           libcontainerUtils.CleanPath(source),
+		Destination:      libcontainerUtils.CleanPath(m.Destination),
 		Data:             data,
 		Flags:            flags,
 		PropagationFlags: pgflags,


cyphar:
This isn't always the most reasonable thing to do, because `filepath.Clean` lexically resolves paths but the kernel does not resolve paths in the same way. For instance if you have `/foo/../bar`, with `/foo` being a symlink to `/baz/blah`, then `filepath.Clean` will give you `/bar` while the kernel will resolve it as `/baz/bar`.

I don't mind this change that much but if we're going to do this then we should use `libcontainer/utils.CleanPath` which handles non-absolute paths much better than `filepath`.

Ace-Tang:
@cyphar , I agree with you on the `filepath.Clean`  do with `resolve symlink path` things, and I updated.

danishprakash:
@cyphar this is still an issue, your suggestion to use `libcontainer/utils.CleanPath` makes sense and I see that an update was made within this PR for the same. Are we not moving forward with the cleaning of the mount paths for some reason? I can send in a patch if it was due to inactivity.

Issue #1890, opened 2018-09-13
Panic if `config.json` is missing "linux" configuration.  Should fail gracefully instead.


marler8997:
Here's the stacktrace, but the fix would likely be checking that spec.Linux is not nil before dereferenceing it in https://github.com/opencontainers/runc/blob/master/libcontainer/specconv/spec_linux.go#L260

panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xd8 pc=0x558fe4aac809]

goroutine 1 [running]:
github.com/opencontainers/runc/vendor/github.com/urfave/cli.HandleAction.func1(0xc00018b890)
	/home/marler8997/go/src/github.com/opencontainers/runc/vendor/github.com/urfave/cli/app.go:478 +0x23c
panic(0x558fe4dfba80, 0x558fe507dda0)
	/home/marler8997/go/src/runtime/panic.go:513 +0x1bd
github.com/opencontainers/runc/libcontainer/specconv.CreateLibcontainerConfig(0xc00018b2b0, 0x558fe4ade782, 0xe, 0x0)
	/home/marler8997/go/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go:260 +0x7e9
main.createContainer(0xc00017e140, 0x7ffe97d3cb61, 0xb, 0xc000162150, 0xb, 0x0, 0x558fe48222d3, 0x0)
	/home/marler8997/go/src/github.com/opencontainers/runc/utils_linux.go:253 +0x146
main.startContainer(0xc00017e140, 0xc000162150, 0x2, 0x0, 0x0, 0xc000128600, 0x8)
	/home/marler8997/go/src/github.com/opencontainers/runc/utils_linux.go:426 +0x10c
main.glob..func13(0xc00017e140, 0x0, 0x0)
	/home/marler8997/go/src/github.com/opencontainers/runc/run.go:76 +0xb0
reflect.Value.call(0x558fe4de72c0, 0x558fe4e578d0, 0x13, 0x558fe4adb2ec, 0x4, 0xc0000e1850, 0x1, 0x1, 0xc0000840c0, 0xc00013e680, ...)
	/home/marler8997/go/src/reflect/value.go:447 +0x44b
[...]

PR #1894, opened 2018-09-17
Fix issue #1890: config.json with no linux config should not crash

marler8997:
Fix https://github.com/opencontainers/runc/issues/1890

Just moved the check for `spec.Linux.IntelRdt` into the `if spec.Linux != nil` block.


crosbymichael:
LGTM

cyphar:
LGTM

cyphar:
LGTM.

PR #1895, opened 2018-09-20
tty: close epollConsole on errors

giuseppe:
make sure epollConsole is closed before returning an error.  It solves
a hang when using these commands with a container that uses a
terminal:

runc run foo &
ssh root@localhost runc exec foo echo hello


rhatdan:
LGTM

crosbymichael:
LGTM

rhatdan:
@avagin @caniszczyk @cyphar @dqminh @hqhq @mrunalp @rjnagal @vmarmol PTAL

mrunalp:
LGTM

PR #1896, opened 2018-09-20
Overlayfs can be fooled to copyup on append in readonly mode.

rhatdan:
This change converts all mount points.

all mount points (host, container or any other namespace) will be converted to ro

Currently with an overlayfs back end if I do

podman run --name test fedora bash -c "echo hello >> /bin/sh"
5f393f523c1fb97d47ac6dfd1304e18a5c73a0893b28c09d1e2fb31e39570b94
bash: /bin/sh: Read-only file system
C /run
A /run/.containerenv
A /run/secrets
C /usr
C /usr/bin
C /usr/bin/bash

After this fix we see.

bc609feec03cdd544bf8fd2e1935687d5b95d9445fe6a0c57a295ee17e9c841e
bash: /bin/sh: Read-only file system
C /run
A /run/.containerenv
A /run/secrets

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -653,7 +653,7 @@ func prepareRoot(config *configs.Config) error {
 }
 
 func setReadonly() error {
-	return unix.Mount("/", "/", "bind", unix.MS_BIND|unix.MS_REMOUNT|unix.MS_RDONLY|unix.MS_REC, "")
+	return unix.Mount("/", "/", "bind", unix.MS_REMOUNT|unix.MS_RDONLY, "")
 }
 
 func setupPtmx(config *configs.Config) error {


rhatdan:
@rhvgoyal @adelton FYI

rhatdan:
@avagin @caniszczyk @crosbymichael @cyphar @dqminh @hqhq @mrunalp @rjnagal @vmarmol PTAL

cyphar:
A couple of things.

* The original code is not correct because it doesn't make the entire root filesystem read-only. It only makes the `/` mount read-only. There is an argument over which is the correct thing to do, but at the very least there should be a clarification in the spec about what `readonly: true` means. Because currently there is no kernel facility to recursively make all child mounts read-only -- you have to parse `/proc/self/mountinfo` and do it manually.

* This change is not correct -- `MS_BIND|MS_REMOUNT` has a special case in the kernel and this change will break rootless containers. In particular, it doesn't require owning the superblock of the filesystem -- it just goes through `change_mount_flags` which is a purely mount namespace path. However there is obviously a bug here, and I think it is that `overlayfs` doesn't correctly handle `change_mount_flags` (instead it requires a full `do_remount_sb` -- which is what this change is doing). However, this change is not correct because it breaks bind-mounted root filesystems (unless I'm mistaken) and it also breaks rootless containers which generally don't own the superblock of the filesystem.

I will take a look at this because I believe there is an underlying `overlayfs` kernel bug here, but NACK on this particular patch.

rhatdan:
@rhvgoyal Confirmed that the latest upstream kernel fixes this issue.
" dwalsh: latest upstream kernel first checks whether write access on filesystem is allowed or not. And later it does copy up. And that's why this problem is not present in 4.19-rc1 onwards."

So I will close this PR and just wait for the upstream kernel.

rhatdan:
Thanks @cyphar for review.

cyphar:
So I didn't see you comment that it was fixed in `4.19` and I spent most of today debugging it. For reference (to make the several hours I spent figuring out what was causing this bug worthwhile), the precise bug was that `vfs_open` called `d_real` (which does a copy-up on overlayfs) without doing a `mnt_want_write` (which would block writes through a readonly mount).

There were previous `mnt_want_write` calls but they were either ignored (as in `do_last`) or not hit during the path in this bug (as in `lookup_open`), and subsequent `mnt_want_write` calls happened after the `d_real` window when the copy-up already occurred (and `mnt_want_write` is called on the underlying mounts within `overlayfs` so they were not helpful).

The fix is to not call `d_real` before doing `mnt_want_write`. My solution (before I saw your comment) was to do an `mnt_want_write` in `vfs_open` but the solution in `4.19` is much simpler and removes the need for `d_real` in `vfs_open` entirely (`do_dentry_open` already handles `mnt_want_write` correctly).

PR #1897, opened 2018-09-21
tty: clean up epollConsole closing

cyphar:
ec0d23a92ffe ("tty: close epollConsole on errors") fixed a significant
issue, but the cleanup was not ideal (especially if the function is
changed in future to add additional error conditions to those currently
present). Using the defer-named-error trick avoids this issue and makes
the code more readable.

Fixes #1895

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -71,7 +71,7 @@ func inheritStdio(process *libcontainer.Process) error {
 	return nil
 }
 
-func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) error {
+func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) (Err error) {
 	f, err := utils.RecvFd(socket)
 	if err != nil {
 		return err
@@ -89,6 +89,11 @@ func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) error {
 	if err != nil {
 		return err
 	}
+	defer func() {
+		if Err != nil {
+			epollConsole.Close()
+		}
+	}()
 	go epoller.Wait()
 	go io.Copy(epollConsole, os.Stdin)
 	t.wg.Add(1)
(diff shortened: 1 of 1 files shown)


giuseppe:
LGTM

rhatdan:
LGTM
@avagin @caniszczyk @crosbymichael @dqminh @hqhq @mrunalp @rjnagal @vmarmol PTAL

tty.go
```diff
@@ -71,7 +71,7 @@ func inheritStdio(process *libcontainer.Process) error {
 	return nil
 }
 
-func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) error {
+func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) (Err error) {
```
crosbymichael: First time I have ever seen `Err` instead of `err`

rhatdan: Same here.  I kind of like it to differentiate and stop mistakes.

cyphar: So, this is quite widely used in a few places. The reason for not using `err` is because if you use a variable name already used in the function it can be a little bit confusing whether the last-assigned value of `err` is what's tested or the `return ___` which triggered the `defer`. Add in scopes like

```go
func baz() (err error) {
    if _, err := foo(); err == nil {
        defer func() {
            if err != nil {
               // does this ever execute?
            }
        }()
    }
}
```

And it just becomes complicated to reason about (even though the semantics are actually defined in the Go language spec -- in the above case [it does not execute](https://play.golang.org/p/wJyeNhe_vI5)).

By using a variable that is not used within the function at all, you eliminate these sorts of confusions and you also have a clear way of telling whether you are referencing the local variable or the actual return value. I use this quite a bit in `umoci` and I stole the idea from somewhere else (I think it's even used within Docker from memory).

cyphar:
Last call for pre-1.0 merges!

crosbymichael:
LGTM

PR #1907, opened 2018-10-11
[fix] execute prepareRootfs for a new mount namespace only

verm666:
Without that if condition runc creates a bunch of mounts for non NEWNS containers and don't clean after all.


verm666:
bug was introduced in 91ca331474b6b7859a85678d8a4dd4b876345ddd

(cc: @crosbymichael)

cyphar:
I get that `runc` will create mounts in the host mount namespace, but they are all inside the `chroot` right? The only other option would be to just fail if you have mounts specified with no `NEWNS`.

verm666:
> The only other option would be to just fail if you have mounts specified with no NEWNS.

Makes sense

mikebrow:
@verm666 the travis fix is already merged, https://github.com/opencontainers/runc/pull/1908

You can rebase or just tweak your commit and push -f, I don't have access or i'd just hit the restart button for ya :-)

crosbymichael:
looking

crosbymichael:
What if others depend on this functionality? What is the point of preventing it?

verm666:
> What is the point of preventing it?
Each container restart produces additional "leaked" mounts, and eventually you can see in mount(8) output thousands of lines. It's not critical, it doesn't affect containers in any bad way but it seems like a bug for me.

crosbymichael:
But givin the configuration, that is expected and should be handled. Atleast that is my POV, what do you think @cyphar ?

verm666:
Thank @mikebrow will rebase!

cyphar:
I agree with @crosbymichael -- you are asking us to mount things in a chroot setup. This is far less than ideal, and I would tell people to simply not do this (after all, this is not a secure way to set up a container) but I don't see the benefit of disabling it entirely.

If there was an underlying security issue (like we were mounting over host paths) then obviously we'd need to do something about it, but I'm not so sure this is as significant of an issue. Though, there is an argument that there are some `rename`-based `chroot` attacks which might cause some worry...

mikebrow:
my rebase a pr flow:

first rebase master

$ git checkout master
$ git fetch upstream master --tags
$ git merge upstream/master
$ git push origin master

Then checkout your pr's branch and rebase that pr against master:
$ git checkout yourBranch
$ git rebase master
Follow rebase instructions for example if there was a merge conflict you'll need to edit the conflict file:
$ atom confictFile
$ git add conflictFile
$ git rebase --continue

once rebase is complete:
$ git push -f origin yourBranch

cyphar:

cyphar:
(Note that I'm still not in favour of this change -- I'm just telling you what is causing the CI to fail.)

PR #1916, opened 2018-10-22
Add support for cgroup namespace

### LINKED ISSUE #1884: Sporadic runc exec failures
As of runc 1.0.0-rc5+dev we've started noticing increase rate in sporadic errors during k8s [liveness proves](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/). 
When the test fails, it produces the following error:
```sh
false Error:  OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:138: adding pid 3101 to cgroups caused \"failed to write 3101 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/docker/1d7b29b96dfd7bc97c6a2d6cbff82b00509cdcc4dbf2ac72ef5dd2bef9db7067/cgroup.procs: invalid argument\"": unknown
```
This might be related to #1326 and moby/moby#31230, but without a root cause or a resolution we can't be sure. 
When playing with runc code, it appears that the error disappears if we retry running `cgroups.EnterPid` on failure (which might indicate this is a transient/race issue).

We've found that it's easier to reproduce this issue if we run `perf` together with excessive docker exec load (we tried this on ubuntu xenial and other OSs). 
1.  Run perf
```sh
perf trace --no-syscalls --event 'sched:*'
```
2. Run the following code (similar to how k8s is invoking the liveness tests) 
```go
package main

import (
	"fmt"
	"github.com/fsouza/go-dockerclient"
	"time"
	"strings"
)

func main() {
	client, err := docker.NewClient("unix:///var/run/docker.sock")
	if err != nil {
		panic(err)
	}
	go func() {
		for {
			fmt.Println("Running", time.Now().String())
			time.Sleep(time.Hour)
		}
	}()
	const container = "test-container"
	client.RemoveContainer(docker.RemoveContainerOptions{
		ID:    container,
		Force: true,
	})
	_, err = client.CreateContainer(docker.CreateContainerOptions{
		Name: container,
		Config: &docker.Config{
			Image:        "busybox",
			Cmd:          []string{"sleep", "10000"},
			Tty:          true,
		},
	})
	if err != nil {
		panic(err)
	}
	if err := client.StartContainer(container, nil); err != nil {
		panic(err)
	}
	var tty bool

	for {
		tty = !tty
		exec, err := client.CreateExec(docker.CreateExecOptions{AttachStdout: true, AttachStderr: true, Tty: tty, Container: container, Cmd: []string{"echo", "A"}})
		if err != nil {
			panic(err)
		}
		m := logger{}
		err = client.StartExec(exec.ID, docker.StartExecOptions{Tty: tty, OutputStream: m, ErrorStream: m})
		if err != nil && strings.Contains(err.Error(), "invalid") {
			fmt.Println(err.Error())
			panic(err)
		} else if err != nil {
			fmt.Println("err", err, time.Now())
		}
	}
}

type logger struct{}
func (m logger) Write(p []byte) (n int, err error) {
	if strings.Contains(string(p), "cgroup") {
		fmt.Println(time.Now().String(), "Error:", string(p))
	}
	return len(p), nil
}
```

Runc version
```
runc version 1.0.0-rc5+dev
commit: 69663f0bd4b60df09991c08812a60108003fa340-dirty
spec: 1.0.0
```

crosbymichael:
Carrying [#781](https://github.com/opencontainers/runc/pull/781). Thanks for @cyphar 's work. :smile:

Closes #1184
Fixes #1884

Cgroup namespace can be configured in `config.json` as other
namespaces. Here is an example:

```
"namespaces": [
	{
		"type": "pid"
	},
	{
		"type": "network"
	},
	{
		"type": "ipc"
	},
	{
		"type": "uts"
	},
	{
		"type": "mount"
	},
	{
		"type": "cgroup"
	}
],
```

Note that if you want to run a container which has shared cgroup ns with
another container, then it's strongly recommended that you set
proper `CgroupsPath` of both containers(the second container's cgroup
path must be the subdirectory of the first one). Or there might be
some unexpected results.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -42,6 +42,12 @@ enum sync_t {
 	SYNC_ERR = 0xFF,	/* Fatal error, no turning back. The error code follows. */
 };
 
+/*
+ * Synchronisation value for cgroup namespace setup.
+ * The same constant is defined in process_linux.go as "createCgroupns".
+ */
+#define CREATECGROUPNS 0x80
+
 /* longjmp() arguments. */
 #define JUMP_PARENT 0x00
 #define JUMP_CHILD  0xA0
@@ -640,7 +646,6 @@ void nsexec(void)
 	case JUMP_PARENT:{
 			int len;
 			pid_t child, first_child = -1;
-			char buf[JSON_MAX];
 			bool ready = false;
 
 			/* For debugging. */
@@ -716,6 +721,18 @@ void nsexec(void)
 							kill(child, SIGKILL);
 							bail("failed to sync with child: write(SYNC_RECVPID_ACK)");
 						}
+
+						/* Send the init_func pid back to our parent.
+						 *
+						 * Send the init_func pid and the pid of the first child back to our parent.
+						 * We need to send both back because we can't reap the first child we created (CLONE_PARENT).
+						 * It becomes the responsibility of our parent to reap the first child.
+						 */
+						len = dprintf(pipenum, "{\"pid\": %d, \"pid_first\": %d}\n", child, first_child);
+						if (len < 0) {
+							kill(child, SIGKILL);
+							bail("unable to generate JSON for child pid");
+						}
 					}
 					break;
 				case SYNC_CHILD_READY:
@@ -759,23 +776,6 @@ void nsexec(void)
 					bail("unexpected sync value: %u", s);
 				}
 			}
-
-			/*
-			 * Send the init_func pid and the pid of the first child back to our parent.
-			 *
-			 * We need to send both back because we can't reap the first child we created (CLONE_PARENT).
-			 * It becomes the responsibility of our parent to reap the first child.
-			 */
-			len = snprintf(buf, JSON_MAX, "{\"pid\": %d, \"pid_first\": %d}\n", child, first_child);
-			if (len < 0) {
-				kill(child, SIGKILL);
-				bail("unable to generate JSON for child pid");
-			}
-			if (write(pipenum, buf, len) != len) {
-				kill(child, SIGKILL);
-				bail("unable to send child pid to bootstrapper");
-			}
-
 			exit(0);
 		}
 
@@ -862,14 +862,17 @@ void nsexec(void)
 				if (setresuid(0, 0, 0) < 0)
 					bail("failed to become root in user namespace");
 			}
-
 			/*
-			 * Unshare all of the namespaces. Note that we don't merge this
-			 * with clone() because there were some old kernel versions where
-			 * clone(CLONE_PARENT | CLONE_NEWPID) was broken, so we'll just do
-			 * it the long way.
+			 * Unshare all of the namespaces. Now, it should be noted that this
+			 * ordering might break in the future (especially with rootless
+			 * containers). But for now, it's not possible to split this into
+			 * CLONE_NEWUSER + [the rest] because of some RHEL SELinux issues.
+			 *
+			 * Note that we don't merge this with clone() because there were
+			 * some old kernel versions where clone(CLONE_PARENT | CLONE_NEWPID)
+			 * was broken, so we'll just do it the long way anyway.
 			 */
-			if (unshare(config.cloneflags) < 0)
+			if (unshare(config.cloneflags & ~CLONE_NEWCGROUP) < 0)
 				bail("failed to unshare namespaces");
 
 			/*
@@ -956,6 +959,18 @@ void nsexec(void)
 			if (!config.is_rootless_euid && config.is_setgroup) {
 				if (setgroups(0, NULL) < 0)
 					bail("setgroups failed");
+			}
+
+			/* ... wait until our topmost parent has finished cgroup setup in p.manager.Apply() ... */
+			if (config.cloneflags & CLONE_NEWCGROUP) {
+				uint8_t value;
+				if (read(pipenum, &value, sizeof(value)) != sizeof(value))
+					bail("read synchronisation value failed");
+				if (value == CREATECGROUPNS) {
+					if (unshare(CLONE_NEWCGROUP) < 0)
+						bail("failed to unshare cgroup namespace");
+				} else
+					bail("received unknown synchronisation value");
 			}
 
 			s = SYNC_CHILD_READY;
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -22,6 +22,10 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// Synchronisation value for cgroup namespace setup.
+// The same constant is defined in nsexec.c as "CREATECGROUPNS".
+const createCgroupns = 0x80
+
 type parentProcess interface {
 	// pid returns the pid for the running process.
 	pid() int
@@ -225,12 +229,17 @@ func (p *initProcess) externalDescriptors() []string {
 	return p.fds
 }
 
-// execSetns runs the process that executes C code to perform the setns calls
-// because setns support requires the C process to fork off a child and perform the setns
-// before the go runtime boots, we wait on the process to die and receive the child's pid
-// over the provided pipe.
-// This is called by initProcess.start function
-func (p *initProcess) execSetns() error {
+// getChildPid receives the final child's pid over the provided pipe.
+func (p *initProcess) getChildPid() (int, error) {
+	var pid pid
+	if err := json.NewDecoder(p.parentPipe).Decode(&pid); err != nil {
+		p.cmd.Wait()
+		return -1, err
+	}
+	return pid.Pid, nil
+}
+
+func (p *initProcess) waitForChildExit(childPid int) error {
 	status, err := p.cmd.Process.Wait()
 	if err != nil {
 		p.cmd.Wait()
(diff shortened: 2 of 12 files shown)


crosbymichael:
For #1884, We worked on a repo and I have verified that this fixes the issue.  Without the patch, you can repo in about 2-5min, with the patch, it runs for hours without error.  I'm confident that this fixes the issue.

libcontainer/nsenter/nsexec.c
```diff
@@ -704,18 +723,46 @@ void nsexec(void)
...
+						    != sizeof(s)) {
+							kill(first_child,
+							     SIGKILL);
 							kill(child, SIGKILL);
 							bail("failed to sync with child: write(SYNC_RECVPID_ACK)");
 						}
+
+						/* Send the init_func pid back to our parent. */
+						/*
+						 *                       * Send the init_func pid and the pid of the first child back to our parent.
```
mikebrow: comment block mishap..

libcontainer/nsenter/nsexec.c
```diff
@@ -201,7 +207,8 @@ static void update_setgroups(int pid, enum policy_t setgroup)
 		 * open(2) or write(2) will return ENOENT. This is fine.
 		 */
 		if (errno != ENOENT)
-			bail("failed to write '%s' to /proc/%d/setgroups", policy, pid);
+			bail("failed to write '%s' to /proc/%d/setgroups",
+			     policy, pid);
```
mikebrow: harder to see what's new when the diff includes lots of odd extra spaces and new lines throughout

libcontainer/nsenter/nsexec.c
```diff
@@ -704,18 +723,46 @@ void nsexec(void)
...
+						 *                                                                       * We need to send both back because we can't reap the first child we created (CLONE_PARENT).
+						 *                                                                                               * It becomes the responsibility of our parent to reap the first child.
+						 *                                                                                                                       */
+						len =
+						    snprintf(buf, JSON_MAX,
+							     "{\"pid\": %d, \"pid_first\": %d}\n",
+							     child,
+							     first_child);
+						if (len < 0) {
+							kill(child, SIGKILL);
```
mikebrow: len of child and first > (4096 - 20ish?) maybe dump those pids to log? put another way.. how could the length of pid / first be > 12chars? Maybe they are not valid pids?

libcontainer/nsenter/nsexec.c
```diff
@@ -854,20 +887,28 @@ void nsexec(void)
...
+			 * Note that we don't merge this with clone() because there were
+			 * some old kernel versions where clone(CLONE_PARENT | CLONE_NEWPID)
+			 * was broken, so we'll just do it the long way anyway.
+			 */
 			/*
-			 * Unshare all of the namespaces. Note that we don't merge this
-			 * with clone() because there were some old kernel versions where
-			 * clone(CLONE_PARENT | CLONE_NEWPID) was broken, so we'll just do
-			 * it the long way.
+			   if (config.cloneflags & CLONE_NEWCGROUP)
```
mikebrow: comment mishap

libcontainer/nsenter/nsexec.c
cyphar: Also this can be done with `dprintf` without the need for an intermediate buffer. I would recommend doing that.

libcontainer/nsenter/nsexec.c
```diff
@@ -854,22 +886,29 @@ void nsexec(void)
...
+			 *
+			 * Note that we don't merge this with clone() because there were
+			 * some old kernel versions where clone(CLONE_PARENT | CLONE_NEWPID)
+			 * was broken, so we'll just do it the long way anyway.
 			 */
-			if (unshare(config.cloneflags) < 0)
+			uint32_t flags = config.cloneflags;
+			if (config.cloneflags & CLONE_NEWCGROUP)
+				flags &= ~CLONE_NEWCGROUP;
+			if (unshare(flags) < 0)
```
cyphar: I would just do `config.cloneflags & ~CLONE_NEWCGROUP` and not have the temporary `flags` variable. But that's just me...

libcontainer/process_linux.go
```diff
@@ -22,6 +22,10 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// Synchronisation value for cgroup namespace setup.
+// The same constant is defined in nsexec.c as "CREATECGROUPNS".
+const createCgroupns byte = (1 << 7)
```
cyphar: I really would prefer if we represented both the same way. Either `0x80` or `1<<7` everywhere.

crosbymichael:
So as far as the whitespace, I thought we all agreed on using `indent -linux` for our C files so that they stay consistent, same as gofmt.

crosbymichael:
Resolved the comments, let me know if you want me to revert the formatting changes.

AkihiroSuda:
Max line length seems changed in recent indent?

crosbymichael:
@AkihiroSuda I think it has always been 80chars

crosbymichael:
Whats the next steps here?  I have verified the issues that this fixes and it would be nice to get this in.  Let me know what changes you want or not

libcontainer/process_linux.go
```diff
@@ -297,19 +292,47 @@ func (p *initProcess) start() error {
...
+	if err := p.waitForChildExit(childPid); err != nil {
+		return newSystemErrorWithCause(err, "waiting for our first child to exit")
+	}
+
+	defer func() {
+		if err != nil {
+			// TODO: should not be the responsibility to call here
+			p.manager.Destroy()
+		}
+	}()
```
dqminh: this block is odd because we do most of the same thing earlier. Shouldnt we only need to apply the cgroup namespace and wait for child here ?

crosbymichael: do you mean, just this defer?

dqminh: No, basically most of the block except the new part about setting up cgroup namespace and wait for first child to exit. Here is the current block we have just a few lines above: https://github.com/opencontainers/runc/blob/master/libcontainer/process_linux.go#L276-L295

Now this is interesting. We certainly put the parent into cgroup before we spawning the child accoring to our current code, so i wonder where the leak can be. Also #1884 is about `runc exec, shouldnt we apply the same thing in `setnsProcess` too ? I think that is the root cause of #1884, isnt it ?

cyphar:
It's always been 80chars, but I think when I ran the formatting it was with 120 characters (which is becoming a more common "sane" width). My only complaint with the formatting is that it makes stuff harder to read (function arguments are wrapped in odd ways and so on -- something that you don't see in actual Linux code). I would exclude them and we can re-establish a format in a separate PR.

libcontainer/process_linux.go
cyphar: I think that the old `Apply` code should be removed since there's not much point in applying it twice -- maybe everything including the `intelRdt` cleanup as well as the comments should be just wholesale moved here... Not sure.

> We certainly put the parent into cgroup before we spawning the child accoring to our current code, so i wonder where the leak can be.

I'm not sure that's current -- we have `err := p.cmd.Start()` before the cgroup application code. Though, I find it hard to believe that `runc init` can do all of its setup before going into the Go code before Go runs through 5 lines of code (especially since Linux's scheduler is biased to the currently-running process -- so theoretically the parent process should get more work done in the short race after fork+exec).

crosbymichael:
Sounds good to me, i'll make the changes.

crosbymichael:
Ok, pushed the formatting changes

cyphar:
LGTM, and I hope we can get this in 1.0 but if we can't we can always do it in the next release. I've tested it out and it appears to work for most basic uses.

<hr>

Last call for pre-1.0 merges!

crosbymichael:
I think we should merge this for 1.0, its a new feature but the implementation also fixes a nasty bug under stress

cyphar:
Yeah, I agree.

mrunalp:
LGTM

mrunalp:
will merge once pull approve returns.

giuseppe:
It seems that this change introduced a problem with the systemd cgroupfs backend.

I've added this snippet to the default "runc spec" config file (under `linux/resources`):

```json
      "memory": {
        "kernel": 41943040,
        "disableOOMKiller": false
      }
```

now when I try to run a container I get:

```bash
# runc --systemd-cgroup run foo
container_linux.go:344: starting container process caused "process_linux.go:311: applying cgroup configuration for process caused \"failed to set memory.kmem.limit_in_bytes, because either tasks have already joined this cgroup or it has children\""
```

libcontainer/process_linux.go
giuseppe: having the `Apply` after we have PIDs in the cgroup breaks `memory.kmem.limit_in_bytes` which can be set only when the cgroup is empty.

Is it safe to remove the second `Apply`?

kolyshkin: I still don't understand why there are two identical `defer` blocks in the same `start()` method.

thaJeztah: @kolyshkin I saw https://github.com/opencontainers/runc/pull/2238 is suggesting to remove one

PR #1918, opened 2018-10-25
rootless: fix running with /proc/self/setgroups set to deny

giuseppe:
This is a regression from 06f789cf26774dd64cb2a9cc0b3c6a6ff832733b
when the user namespace was configured without a privileged helper.
To allow a single mapping in an user namespace, it is necessary to set
/proc/self/setgroups to "deny".

For a simple reproducer, the user namespace can be created with
"unshare -r".

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"os"
 	"strings"
@@ -300,11 +301,18 @@ func setupUser(config *initConfig) error {
 		return err
 	}
 
+	setgroups, err := ioutil.ReadFile("/proc/self/setgroups")
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
 	// This isn't allowed in an unprivileged user namespace since Linux 3.19.
 	// There's nothing we can do about /etc/group entries, so we silently
 	// ignore setting groups here (since the user didn't explicitly ask us to
 	// set the group).
-	if !config.RootlessEUID {
+	allowSupGroups := !config.RootlessEUID && strings.TrimSpace(string(setgroups)) != "deny"
+
+	if allowSupGroups {
 		suppGroups := append(execUser.Sgids, addGroups...)
 		if err := unix.Setgroups(suppGroups); err != nil {
 			return err


baude:
this regression has broken podman in fedora ... expediting review of this PR would be very much appreciated.

AkihiroSuda:
Sorry and thanks, can we also have integration test?

rhatdan:
LGTM

giuseppe:
> can we also have integration test?

sure, what would be the best way to integrate "unshare -r" as wrapper when creating a user namespace?  Do we already have any tests that run in a user namespace not created by runc?

AkihiroSuda:
LGTM (IANAM)

Integration test can be probably investigated in separate issue/PR

libcontainer/init_linux.go
```diff
@@ -300,11 +301,18 @@ func setupUser(config *initConfig) error {
...
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
 	// This isn't allowed in an unprivileged user namespace since Linux 3.19.
 	// There's nothing we can do about /etc/group entries, so we silently
 	// ignore setting groups here (since the user didn't explicitly ask us to
 	// set the group).
-	if !config.RootlessEUID {
+	allowSupGroups := !config.RootlessEUID && strings.TrimSpace(string(setgroups)) != "deny"
```
cyphar: I would prefer `== "allow"`.

cyphar:
My nit aside, lgtm. Doing a test with `unshare -r` wouldn't be a bad idea -- you can probably try to do this by messing around a bit with `tests/rootless.sh` and some `helpers.bash` trickery.

libcontainer/init_linux.go
giuseppe: I've used the `!= deny` so that when `/proc/self/setgroups` doesn't exist (and setgroups is an empty string), `allowSupGroups` is still true.

mrunalp:
LGTM

crosbymichael:
LGTM

cyphar:
(LGTM)

AkihiroSuda:
@giuseppe @cyphar 

Is there any remaining regression issue? Is this ready for runc v1.0.0? (ref: Today's dev ML conversation)

giuseppe:
@AkihiroSuda not that I am aware of.  I am using upstream runc with rootless podman/buildah and they work fine

PR #1919, opened 2018-10-27
libcontainer: intelrdt: add support for Intel RDT/MBA Software Controller in runc

xiaochenshen:
This PR is the runc part of Intel RDT/MBA Software Controller support.
The runtime-spec part: https://github.com/opencontainers/runtime-spec/pull/992

MBA Software Controller feature is introduced in Linux kernel v4.18.
It is a software enhancement to mitigate some limitations in MBA which
describes in kernel documentation. It also makes the interface more user
friendly - we could specify memory bandwidth in "MBps" (Mega Bytes per
second) as well as in "percentages".

The kernel underneath would use a software feedback mechanism or a
"Software Controller" which reads the actual bandwidth using MBM
counters and adjust the memory bandwidth percentages to ensure:
"actual memory bandwidth < user specified memory bandwidth".

We could enable this feature through mount option "-o mba_MBps":
mount -t resctrl resctrl -o mba_MBps /sys/fs/resctrl

In runc, we handle both memory bandwidth schemata in unified format:
"MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;..."
The unit of memory bandwidth is specified in "percentages" by default,
and in "MBps" if MBA Software Controller is enabled.

For more information about Intel RDT and MBA Software Controller:
https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt

### DIFF
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -28,7 +28,8 @@ import (
  *
  * Memory Bandwidth Allocation (MBA) provides indirect and approximate throttle
  * over memory bandwidth for the software. A user controls the resource by
- * indicating the percentage of maximum memory bandwidth.
+ * indicating the percentage of maximum memory bandwidth or memory bandwidth
+ * limit in MBps unit if MBA Software Controller is enabled.
  *
  * More details about Intel RDT CAT and MBA can be found in the section 17.18
  * of Intel Software Developer Manual:
@@ -95,7 +96,7 @@ import (
  *
  * Memory bandwidth schema:
  * It has allocation values for memory bandwidth on each socket, which contains
- * L3 cache id and memory bandwidth percentage.
+ * L3 cache id and memory bandwidth.
  * 	Format: "MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;..."
  * For example, on a two-socket machine, the schema line could be "MB:0=20;1=70"
  *
@@ -105,6 +106,18 @@ import (
  * "info/MB/bandwidth_gran". The available bandwidth control steps are:
  * min_bw + N * bw_gran. Intermediate values are rounded to the next control
  * step available on the hardware.
+ *
+ * If MBA Software Controller is enabled through mount option "-o mba_MBps":
+ * mount -t resctrl resctrl -o mba_MBps /sys/fs/resctrl
+ * We could specify memory bandwidth in "MBps" (Mega Bytes per second) unit
+ * instead of "percentages". The kernel underneath would use a software feedback
+ * mechanism or a "Software Controller" which reads the actual bandwidth using
+ * MBM counters and adjust the memory bandwidth percentages to ensure:
+ * "actual memory bandwidth < user specified memory bandwidth".
+ *
+ * For example, on a two-socket machine, the schema line could be
+ * "MB:0=5000;1=7000" which means 5000 MBps memory bandwidth limit on socket 0
+ * and 7000 MBps memory bandwidth limit on socket 1.
  *
  * For more information about Intel RDT kernel interface:
  * https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt
@@ -165,6 +178,8 @@ var (
 	isCatEnabled bool
 	// The flag to indicate if Intel RDT/MBA is enabled
 	isMbaEnabled bool
+	// The flag to indicate if Intel RDT/MBA Software Controller is enabled
+	isMbaScEnabled bool
 )
 
 type intelRdtData struct {
@@ -197,7 +212,12 @@ func init() {
 			isCatEnabled = true
 		}
 	}
-	if isMbaFlagSet {
+	if isMbaScEnabled {
+		// We confirm MBA Software Controller is enabled in step 2,
+		// MBA should be enabled because MBA Software Controller
+		// depends on MBA
+		isMbaEnabled = true
+	} else if isMbaFlagSet {
 		if _, err := os.Stat(filepath.Join(intelRdtRoot, "info", "MB")); err == nil {
 			isMbaEnabled = true
 		}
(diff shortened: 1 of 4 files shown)


libcontainer/intelrdt/intelrdt.go
```diff
@@ -232,6 +251,11 @@ func findIntelRdtMountpointDir() (string, error) {
 				return "", fmt.Errorf("Error found less than 3 fields post '-' in %q", text)
 			}
 
+			// Check if MBA Software Controller is enabled through mount option "-o mba_MBps"
+			if strings.Contains(postSeparatorFields[2], "mba_MBps") {
+				isMbaScEnabled = true
```
hqhq: I'm a little bit confused by the name, why do we call "mega bytes per second" limit control for MBA "software controller"? Is percentage limit controlled by hardware controller or something?
Second, how do we use this flag in runc? I don't see it except for setting isMbaEnabled flag, doesn't seem to be enough.

xiaochenshen: @hqhq Thank you for kind code review.
> I'm a little bit confused by the name, why do we call "mega bytes per second" limit control for MBA "software controller"? Is percentage limit controlled by hardware controller or something?

The name "MBA Software Controller" stands for an enhancement of MBA feature in kernel. The motivation of "MBA Software Controller" in kernel is to mitigate some limitations in MBA which described in kernel documentation (search "Software Controller" in https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt for more details) and to make the interface more user friendly - we could specify memory bandwidth limit in "MBps" (Mega Bytes per second), it is more straightforward than memory bandwidth limit in "percentages" from end user's perspective.

The kernel underneath uses a software feedback mechanism or a "Software Controller" which reads the actual bandwidth using Memory Bandwidth Monitoring (MBM, an Intel RDT monitoring feature) counters and adjust the memory bandwidth percentages to ensure: "actual memory bandwidth < user specified memory bandwidth". In other words, MBA Software Controller also makes use of percentage limit control in hardware internally.

MBA Software Controller depends on MBA and MBM hardware capabilities, If both MBA and MBM are enabled by hardware, we could enable MBA Software Controller through mount option "-o mba_MBps":
mount -t resctrl resctrl -o mba_MBps /sys/fs/resctrl

>Second, how do we use this flag in runc? I don't see it except for setting isMbaEnabled flag, doesn't seem to be enough.

In this PR, we add new flag "isMbaScEnabled" to indicate if MBA Software Controller is enabled. In step 2 of intelrdt.init(), this flag is set if mount option "mba_MBps" for resctrl filesystem is found during parsing /proc/self/mountinfo. In step 3 of intelrdt.init(), if "isMbaScEnabled" is true, we could mark "isMbaEnabled" true immediately. We don't need to double check if MBA is enabled through checking if /sys/fs/resctrl/info/MB/ is available.

In runc, both memory bandwidth schemata of original MBA and MBA Software Controller are in ``unified`` format: ``MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;...``. the only difference is that the unit of memory bandwidth is specified in "percentages" by default, and in "MBps" if MBA Software Controller is enabled. So we deal with both original MBA and MBA Software Controller cases with exactly the same logic in runc.

In addition, we export function IsMbaScEnabled() to check if flag "isMbaScEnabled" is true or false. Currently it is only called by TestIntelRdtSetMemBwScSchema() in MBA Software Controller unit test in runc. But it will be useful to the caller in upper layer software (e.g., docker) in future. For example, the caller could check if MBA Software Controller is enabled. If yes, the memory bandwidth will be passed in "MBps" unit, otherwise the memory bandwidth will be passed in "percentages" unit.

cyphar:
Rejected -- please don't bump the runtime-spec to a non-released version -- while Go does encourage doing this we absolutely do not want to be using a spec that's not released. You're only updating it for a comment change, so it's really not necessary.

(As an aside I **really** don't like the format for `intelrdt` because of how much magic is involved. But whatever, that's water under the bridge now.)

xiaochenshen:
@cyphar Thank you for code review.
> Rejected -- please don't bump the runtime-spec to a non-released version -- while Go does encourage doing this we absolutely do not want to be using a spec that's not released. You're only updating it for a comment change, so it's really not necessary.

The vendor update in commit https://github.com/opencontainers/runc/pull/1919/commits/64daa60f9742035de190ecdefc7700f5555a247b only changes the code comments to avoid confusion to Intel RDT users. I am OK if we only add the second commit this time because there is no real code dependency.

BTW, do you have any suggestion that how we could add a new "version tag" in runtime-spec which includes the minor change? We could vendor update to a new runtime-spec "version tag" in runc in future? Thank you.

> (As an aside I really don't like the format for intelrdt because of how much magic is involved. But whatever, that's water under the bridge now.)

More and more Cloud Service Provider (CSP) customers are deploying Intel RDT features in their servers. The native Intel RDT support in OCI container is an imperative demand from customers. But the container user scenarios and usages on top of kernel resctrl filesystem are varied. The format of "intelRdt" in runtime-spec config is "passing through" the schemata format in kernel resctrl filesystem. Our goal is to provide the maximum flexibility to Intel RDT end users in OCI container user scenarios.

xiaochenshen:
@cyphar Really appreciated your code review and comments.
> It also changes some of the types for VMs -- which is unrelated to this change. This is more about the principle than anything else, we don't use non-released spec versions no matter how minor the actual changes are.

I understood now. I will remove the vendor update commit in this PR.

> We can do it some time in the future, but I think that's a separate topic to this PR.

Thank you for the suggestion.

xiaochenshen:
@cyphar @hqhq @crosbymichael @mrunalp @rjnagal @vmarmol @dqminh 

Per @cyphar's suggestion, I have removed the runtime-spec vendor update commit in this PR.
Could you help code review commit https://github.com/opencontainers/runc/pull/1919/commits/95af9eff82157317817d3d550626f3e5db240d8a at your convenience? Thank you!

Issue #1898, opened 2018-09-25
[Proposal] Include env var to disable kernel memory accounting when kernel <= 3.10


ryarnyah:
When using RedHat or CentOS, base kernel is 3.10 (with some backport and CONFIG_MEMCG_KMEM..).
We encounter on my current project the same error as described is attached issues.

To avoid reboot all ours nodes in production every week (or recompile kernel on every update) we propose to disable it with an env var `DISABLE_KERNEL_MEMORY_ACCOUNTING` in `libcontainer/cgroups/fs/memory.go`.

See: https://github.com/ryarnyah/runc/tree/1898-add-optional-kernel-accounting

See: https://github.com/scality/kubernetes/commit/b04b0506e49b1b60ea7c8b74a5ca2edbf341cd6c
See: https://github.com/kubernetes/kubernetes/issues/61937
See: https://github.com/opencontainers/runc/pull/1350
See: https://github.com/moby/moby/issues/29638

cyphar:
If you remove all kernel memory accounting options from `config.json` then `runc` shouldn't be touching kernel memory. If we are touching kernel memory for a good reason, we should not touch it if we get `ENOENT` (I don't think we have code for that at the moment -- but it is how I would implement it).

NACK on using an environment variable for this. It feels like a very hacky way of solving the problem, leads to OCI compliance issues (now `runc` will ignore kernel memory settings entirely if the env-flag is set), requires users to manually set the environment variable, and is just hiding a deeper problem.

Auto-detecting by seeing if we get `ENOENT` would be a far better way of solving it (because presumably users would only pass `RUNC_DISABLE_KERNEL_MEMORY_LIMIT` if kernel memory limits aren't exposed by their kernel -- which is something we can detect without asking the user to tell us about it).

cyphar:
Sorry, I misunderstood what the issue was (it's been a while since the last `kmem` bugs were fixed). So there's a kernel bug in pre-4.0 kernels that's causing setting `memory.kmem` to result in a kernel memory leak. Obviously this is very bad (and RedHat should probably work on fixing it -- given that it is a problem in their kernels as far as I understand).

The wish for an environment variable or a flag makes some more sense now, but I'm still not really thrilled about it (we added `--no-new-keyring` for similar reasons and I'm still not a huge fan of it). Kernel version tests aren't really very useful because distributions do backports, so those aren't an option.

[...]

Issue #1725, opened 2018-02-20
Enabling kmem accounting can break applications on CentOS7


TvdW:
After upgrading from Docker (CE) 17.05.0 to 17.12.0, these kernel messages started showing up on my machines:

```
[43073.004575] SLUB: Unable to allocate memory on node -1 (gfp=0x8020)
[43073.022211]   cache: ip6_dst_cache(0:18479a49f76b66461d8fedd17a6eba407f08cc8960bd0246d39df8eb21f20b4f), object size: 448, buffer size: 448, default order: 2, min order: 0
[43073.061420]   node 0: slabs: 74, objs: 2610, free: 0
[43073.077875]   node 1: slabs: 88, objs: 3141, free: 0
```

I was able to reproduce this on Docker 17.06.0, and eventually traced it to the commit introduced by https://github.com/opencontainers/runc/pull/1350 which enables kmem accounting for all containers. But as Docker helpfully suggests, kmemcg are experimental before linux 4.0:

```
$ sudo docker update --kernel-memory 1000g 18479a49f76b
You specified a kernel memory limit on a kernel older than 4.0. Kernel memory limits are experimental on older kernels, it won't work as expected and can cause your system to be unstable.
```

After a few more tests I was able to reproduce the issue on 17.05.0, by passing `--kernel-memory 1000g` to `docker run`. The kernel log slowly fills up (10 messages per hour?) with SLUB warnings, and containers seem far less stable than normal (i.e. they crash).

## Steps to reproduce

1. CentOS 7, latest kernel (3.10.0-693.17.1.el7.x86_64)
2. Docker 17.06.0+ or runc with equivalent settings
3. kmem limit unset, mem limit 40G, with 80G free memory left for page caches
4. An application that very heavily uses the local disk to cause caches to build up (in my case Apache Cassandra)

Eventually, these messages will start popping up in the kernel logs, and in rare cases it leads to an application getting killed/crashed.

--------------

With all of the above said, I'm 99% sure this is a kernel bug related to running an ancient kernel, and runc's patch would at best be a workaround. If only I had a way to get redhat's attention so they can fix it 😃

hqhq:
@TvdW The root cause is there are kernel memory limit bugs in 3.10, if you don't want to use kernel memory limit because it's not stable on your kernel, the best solution would be to disable kernel memory limit on your kernel.

[...]

PR #1920, opened 2018-10-31
libcontainer/cgroups: do not enable kmem on broken kernels     

kolyshkin:
Commit fe898e7862f94 (PR #1350) enables kernel memory accounting for all cgroups created by libcontainer even if kmem limit is not configured.
    
Kernel memory accounting is known to be broken in RHEL7 kernels,
including the latest RHEL 7.5 kernel. It does not support reclaim
and can lead to kernel oopses while removing cgroup (merging it
with its parent). Unconditionally enabling kmem acct on RHEL7
leads to bugs:
   
* https://github.com/opencontainers/runc/issues/1725
* https://github.com/kubernetes/kubernetes/issues/61937
* https://github.com/moby/moby/issues/29638
    
I am not aware of any good way to figure out whether the kernel
memory accounting in the given kernel is working or broken.
For the lack of a better way, let's check if the running kernel
is RHEL7, and disable initial setting of kmem.

Alternative to https://github.com/opencontainers/runc/issues/1898


thaJeztah:
Wondering if it should be a compile time option, so that packagers for a specific distro can enable/disable it

crosbymichael:
cc @mrunalp

kolyshkin:
Apparently the kernel bug(s) we work around here are **not** fixed in latest-n-greatest RHEL7.6 kernels (see latest comments in https://bugzilla.redhat.com/show_bug.cgi?id=1507149).

kolyshkin:
> Alternative to #1898

cc @ryarnyah

mrunalp:
I prefer #1921 and have LGTM'ed it. Thanks!

cyphar:
NACK. I prefer #1921 as well.

kolyshkin:
Closed in favor of #1921

PR #1921, opened 2018-11-01
libcontainer: ability to compile without kmem

kolyshkin:
Commit fe898e7862f94 (PR #1350) enables kernel memory accounting
for all cgroups created by libcontainer -- even if kmem limit is
not configured.

Kernel memory accounting is known to be broken in some kernels,
specifically the ones from RHEL7 (including RHEL 7.5). Those
kernels do not support kernel memory reclaim, and are prone to
oopses. Unconditionally enabling kmem acct on such kernels lead
to bugs, such as

* https://github.com/opencontainers/runc/issues/1725
* https://github.com/kubernetes/kubernetes/issues/61937
* https://github.com/moby/moby/issues/29638

This commit gives a way to compile runc without kernel memory setting
support. To do so, use something like

	make BUILDTAGS="seccomp nokmem"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 PROJECT := github.com/opencontainers/runc
-BUILDTAGS := seccomp
+BUILDTAGS ?= seccomp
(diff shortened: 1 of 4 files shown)


kolyshkin:
This is an alternative to #1920

crosbymichael:
I think this could be a safe change in addition to #1920.  What do you think?

mrunalp:
I prefer this over #1920

kolyshkin:
> I think this could be a safe change in addition to #1920. What do you think?

If you approve #1920 then I guess this one is not needed -- the only kernels we see this issue is RHEL7 ones -- everyone else, including ancient Ubuntu 16.04, is using more recent kernels with kmem controller rewritten.

mrunalp:
LGTM

cyphar:
LGTM.

cyphar:
(and pullapprove is broken again! yay!)

PR #1922, opened 2018-11-02
Makefile: rm cgo tag

kolyshkin:
There is no need to explicitly add `cgo` build tag, it is set by
by go tools if cgo is enabled.

Fixes: ecd64631019d0c08


crosbymichael:
LGTM

mrunalp:
LGTM

PR #1925, opened 2018-11-06
test: fix TestDupNamespaces fail to test dup-ns error

Ace-Tang:
add Root in created spec, or `TestDupNamespaces` is not test dup-namespace ,but test `lack of Root`,  error message is 'Root must be specified'

if not add Root field, test will fail at 
```
if spec.Root == nil {
        return nil, fmt.Errorf("Root must be specified")
    }
```

### DIFF
--- a/libcontainer/specconv/spec_linux_test.go
+++ b/libcontainer/specconv/spec_linux_test.go
@@ -4,6 +4,7 @@ package specconv
 
 import (
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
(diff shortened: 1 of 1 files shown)


Ace-Tang:
I check for the tests, find that when bug fixed in ` TestDupNamespaces`, non-pointer error can be tested by this test, since the test alway exited cause lack of  `Root`  , /cc @cyphar

cyphar:
LGTM, thanks!

cyphar:
Last call for pre-1.0 merges!

crosbymichael:
LGTM

PR #1926, opened 2018-11-06
libcontainer: fix potential panic if spec.Process is nil

Ace-Tang:
for the code logic, pointer 'spec.Process' should be judge first
to avoid panic.


cyphar:
Dammit, this code is so full of nil pointer derefs...

LGTM.

Ace-Tang:
Yes, maybe a refactor is need ^ ^

Ace-Tang:
I think I can add more test for `CreateLibcontainerConfig`, consider for many nil pointer before ^ ^

crosbymichael:
LGTM

PR #1928, opened 2018-11-10
rootless: fix potential panic in shouldUseRootlessCgroupManager

Ace-Tang:
move `context.GlobalBool("systemd-cgroup") ` inside if condition `context != nil`
```
if context != nil {
.....

        if context.GlobalBool("systemd-cgroup") {
			return false, nil
	}
}
```


crosbymichael:
LGTM

cyphar:
LGTM.

PR #1934, opened 2018-11-18
fix: may kill other process when container has been stopped

lifubang:

If a container A has been stopped, and after a long time, the host start a new process B whose process id is just equal to container A's process id.
When `runc kill A 9` again, process B will be killed.
So, we need to check container's status when we want to kill it.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -374,10 +374,18 @@ func (c *linuxContainer) Signal(s os.Signal, all bool) error {
 	if all {
 		return signalAllProcesses(c.cgroupManager, s)
 	}
-	if err := c.initProcess.signal(s); err != nil {
-		return newSystemErrorWithCause(err, "signaling init process")
+	status, err := c.currentStatus()
+	if err != nil {
+		return err
 	}
-	return nil
+	// to avoid a PID reuse attack
+	if status == Running || status == Created {
+		if err := c.initProcess.signal(s); err != nil {
+			return newSystemErrorWithCause(err, "signaling init process")
+		}
+		return nil
+	}
+	return newGenericError(fmt.Errorf("container not running"), ContainerNotRunning)
 }
 
 func (c *linuxContainer) createExecFifo() error {


cyphar:
Yeah, that's obviously a bug. But I think a better solution would be to compare against `InitProcessStartTime` which should be different when there is a PID reuse attack (this is why we use it for other operations). I agree that doing this other check is probably also reasonable, but I think that using `InitProcessStartTime` would be more consistent.

As an aside, there was [a kernel patch sent recently]() which would help avoid this problem entirely in a race-free way by holding open a file descriptor to `/proc/$pid` and then using `ioctl(fd, PROC_FD_SIGNAL, SIGKILL)`.

lifubang:
> As an aside, there was [a kernel patch sent recently](https://lore.kernel.org/lkml/20181118111751.6142-1-christian@brauner.io/) which would help avoid this problem entirely in a race-free way by holding open a file descriptor to `/proc/$pid` and then using `ioctl(fd, PROC_FD_SIGNAL, SIGKILL)`.

It's a good news for taking us more security kernel patch. Thanks for all of you.

> Yeah, that's obviously a bug. But I think a better solution would be to compare against `InitProcessStartTime` which should be different when there is a PID reuse attack (this is why we use it for other operations). I agree that doing this other check is probably also reasonable, but I think that using `InitProcessStartTime` would be more consistent.

In https://github.com/opencontainers/runc/blob/322760b15f4bee132a3a2e5da2133d5d7fb967a4/kill.go#L45 

`containerStatus, err := container.Status()` contains `ProcessStartTime` compare.
Shall we need to change the code to just only use `ProcessStartTime` compare?

lifubang:
I refactor the code refer to https://github.com/opencontainers/runc/blob/1a40f04e15f47e40c7f9fa862fa462d34807fdc7/libcontainer/container_linux.go#L196-L202 . 
And add a lock. I think we need a lock here.

lifubang:
Thank you for your review. I remove the lock now.

cyphar:
The current patch is missing the hunk I suggested. I can push a patch for it if you like.

cyphar:
Sorry, let me clarify. I think we should do both. But yeah, while testing I also noticed that you never hit the `newSystemErrorWithCause(err, "fetching init process start time")` case. I will look into that.

cyphar:
Never mind, `currentStatus` does actually do the `initProcessStartTime` check inside `runType`.

LGTM.

libcontainer/container_linux.go
```diff
@@ -371,6 +371,14 @@ func (c *linuxContainer) start(process *Process) error {
 }
 
 func (c *linuxContainer) Signal(s os.Signal, all bool) error {
+	status, err := c.currentStatus()
+	if err != nil {
+		return err
+	}
+	// to avoid a PID reuse attack
+	if status == Stopped {
```
cyphar: This check should be `status == Running || status == Created`. According to the runtime-spec:

> Attempting to send a signal to a container that is neither created nor running MUST have no effect on the container and MUST generate an error.

Aside from this, LGTM.

cyphar:
I've added this to 1.0 because it's actually a spec violation.

lifubang:
@cyphar I update the code, because if the container use hosts pid namespace, we should let `runc kill -a` run success.

lifubang:
You can see: 
https://github.com/opencontainers/runc/blob/87a188996e229bf382c27865584765d1a50c0211/libcontainer/state_linux.go#L41-L47

cyphar:
LGTM.

crosbymichael:
LGTM

PR #1936, opened 2018-11-19
Small fixes for CRIU based test cases

adrianreber:
This removes unnecessary lines from checkpoint.bats like:

 sed -i 's;"readonly": true;"readonly": false;' config.json

and adds (and corrects) comments which are leftover from older
versions of checkpoint.bats.


adrianreber:
There is one more line were `readonly` is disabled. This should not be necessary and I opened https://github.com/checkpoint-restore/criu/issues/575 for tracking this.

crosbymichael:
LGTM

PR #1937, opened 2018-11-21
*: release v1.0.0~rc6

cyphar:


crosbymichael:
LGTM

mikebrow:
looks good to me.. just to add.. if we need to bump the spec to add additional hooks and / or other needed items to get runc into spec into compliance.. then let's do that too.

cyphar:
One of my proposals is to do exactly that, but I'd like a little bit more consensus from everyone on what the best way forward is. Personally I think adding more hooks and transitioning folks would be the best method, but that's just going to be painful.

I think that we should add an OCI version field to the state JSON passed to hooks so this is less painful in the future...

mikebrow:
> add an OCI version field to the state JSON passed to hooks

was going to suggest that pattern as well.

PR #1941, opened 2018-11-26
add a mount options check for type bind in linux

lifubang:
For a fresh man with runc/containerd, mount a direct to a container may be a common action to us.
If there is a mount like this:
```
{
  "destination": "/context",
  "type": "bind",
  "source": "/root"
}
```
When run `runc run -d redis8`, the response will be:
container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"rootfs_linux.go:58: mounting \\\"/root\\\" to rootfs \\\"/opt/runctest/redis/rootfs\\\" at \\\"/opt/runctest/redis/rootfs/context\\\" caused \\\"no such device\\\"\""

I think this is hard for lots of people to realize what's wrong.
So, maybe we can response some more clear messages.

Such issue is also in containerd: https://github.com/containerd/containerd/issues/2831

I don't know this is reasonable, @maintainers, if you have time, please check, thanks.


cyphar:
I understand the argument for a warning like this, but I'd prefer that we instead just had warnings for `type: "bind"` or `type: "rbind"` without a corresponding option entry. This would be more generic and would handle most mistakes.

lifubang:
Thank you for your review. Updated.

libcontainer/specconv/spec_linux.go
```diff
@@ -190,6 +190,11 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 	exists := false
 	for _, m := range spec.Mounts {
+		if m.Type == "bind" || m.Type == "rbind" {
+			if len(m.Options) == 0 {
+				return nil, fmt.Errorf("Should add options when mount type is (r)bind.")
```
mikebrow: /s/"Should add options when mount type is (r)bind."/"Must include options when mount type is (r)bind for: %v", m/

mikebrow: should probably have a test case for this

mikebrow: and it's not an error .. should be a warning

mikebrow: the runtime spec points to the linux mount manual for details. 

A read of the mount man shows that we should probably be using the mount options of the source if not specified. Thus I would not expect an error to the thrown, that's probably a bug in our implementation.

crosbymichael: We have always defered a lot of things to the underlying syscall.  We cannot replicate every error from ever syscall so the strategy has been, call mount with no options and it will fail anyways.

mikebrow: that makes sense.. so if the syscall is "improved" then we'll just naturally pick that up

crosbymichael: yes, or else it would be a never ending battle trying to catch up or stay in sync with the underlying system

lifubang: > /s/"Should add options when mount type is (r)bind."/"Warning no options specified for(r)bind mount: %v", m/

Changed.

> should probably have a test case for this

Yes, I have add it.

libcontainer/specconv/spec_linux.go
```diff
@@ -190,6 +190,23 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 	exists := false
 	for _, m := range spec.Mounts {
+		// check mount options for (b)bind type mount
+		if m.Type == "bind" || m.Type == "rbind" {
+			if len(m.Options) == 0 {
+				return nil, fmt.Errorf("no options specified for %v type mount: [destination=%v, source=%v].", m.Type, m.Destination, m.Source)
```
mikebrow: suggest making this a warning "logrus.Warnf" vs throwing an error..

As Mr. Crosby said we can let the sys call throw the error.

libcontainer/specconv/spec_linux.go
```diff
@@ -190,6 +190,23 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
...
+			} else {
+				blBind := false
+				for _, opt := range m.Options {
+					if opt == "bind" || opt == "rbind" {
+						blBind = true
+						break
+					}
+				}
+				if !blBind {
+					return nil, fmt.Errorf("options should contain (r)bind for %v type mount: [destination=%v, source=%v].", m.Type, m.Destination, m.Source)
```
mikebrow: same

lifubang: use "logrus.Warnf" now. But we need `runc --log /dev/stdout` to display warn message, otherwise we just can see the error message `container_linux.go:344: starting container process caused "process_linux.go:424: container init caused \"rootfs_linux.go:58: mounting \\\"/etc/test\\\" to rootfs \\\"/opt/mygo/src/github.com/opencontainers/runctest/rootfs\\\" at \\\"/opt/mygo/src/github.com/opencontainers/runctest/rootfs/etc/test\\\" caused \\\"no such device\\\"\""`.

libcontainer/specconv/example.go
```diff
@@ -107,6 +107,12 @@ func Example() *specs.Spec {
 				Source:      "cgroup",
 				Options:     []string{"nosuid", "noexec", "nodev", "relatime", "ro"},
 			},
+			{
+				Destination: "/srv",
+				Type:        "bind",
+				Source:      "/srv",
+				Options:     []string{"bind", "ro"},
```
cyphar: `Example()` is used by quite a few projects, I'd prefer to not bind-mount host paths into containers by default.

libcontainer/specconv/spec_linux.go
```diff
@@ -190,6 +191,23 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 	exists := false
 	for _, m := range spec.Mounts {
+		// check mount options for (b)bind type mount
+		if m.Type == "bind" || m.Type == "rbind" {
+			if len(m.Options) == 0 {
+				logrus.Warnf("no options specified for %v type mount: [destination=%v, source=%v].", m.Type, m.Destination, m.Source)
```
cyphar: This warning should be the same as the other one. The code should be something like:

```
hasBindOpt := false
for _, opt := m.Options {
  // ...
}
if !hasBindOpt {
  logrus.Warnf(...)
}
```

In particular you don't need to special-case `len(m.Options) == 0`.

libcontainer/specconv/example.go
lifubang: deleted.

libcontainer/specconv/spec_linux.go
lifubang: Updated.
I still have a suggestion. How about use `fprintf(stderr, "....")` instead of `logrus.Warnf`?
Because user must use `--log /dev/stdout` to get the warning. And rootless container don't have access to `/dev/stdout`.

andrewrynhard:
Any way we can get this in? This is pretty easy to mess up. It is not intuitive that `type: bind` also requires an option of `bind` or `rbind`. I often find myself forgetting that and seeing the `unknown device` error (or something along those lines).

Just an idea, couldn't the option be automatically injected based on the type? I guess I am failing to understand the significance of `type: bind`.

cyphar:
This might be worthwhile to have, but just to clarify -- it's not that `type: "bind"` additionally requires a `"bind"` argument. It's that `type: "bind"` does nothing and you must have a `"bind"` argument. This mirrors the Linux kernel API (bind mounts are created with the `MS_BIND` flag, not a filesystem type of `bind`).

The main concern is that it's adding extra semantics on top of the Linux API when really we should be staying as close as possible to the Linux API. Though admittedly the only reason why people think `type: "bind"` does something is because our examples include it, so this is kind of our own fault :confused:.

andrewrynhard:
> @andrewrynhard
> 
> This might be worthwhile to have, but just to clarify -- it's not that `type: "bind"` additionally requires a `"bind"` argument. It's that `type: "bind"` does nothing and you must have a `"bind"` argument. This mirrors the Linux kernel API (bind mounts are created with the `MS_BIND` flag, not a filesystem type of `bind`).
> 
> The main concern is that it's adding extra semantics on top of the Linux API when really we should be staying as close as possible to the Linux API. Though admittedly the only reason why people think `type: "bind"` does something is because our examples include it, so this is kind of our own fault 😕.

I can understand this. I would opt for staying closer to the Linux semantics in that case.

PR #1942, opened 2018-11-30
Modify check-config.sh in accordance with Moby Project updates

KentaTada:
Some kernel configs are missing like below. So, I updated runC's check-config.sh in accordance with latest Moby's check-config.sh which excepts "Network Drivers" and "Storage Drivers"

https://github.com/moby/moby/commit/2fb2361b29aeaa0d0f931f55db372bc8f56b4c60

Or we should remove all network related configs from runC's check-config.sh.
Could you review my commits?


crosbymichael:
LGTM

dqminh:
LGTM

PR #1944, opened 2018-12-03
cr: get pid from criu notify when restore

Ace-Tang:
when restore container from a checkpoint directory, we should get
pid from criu notify, since c.initProcess has not been created.

## why I do this

I fail to update runc version in pouch project, https://github.com/alibaba/pouch/pull/2516, I check the restore code, in `criuNotifications` function,  `currentOCIState()` can not get pid since c.initProcess has not been created.
```
    case notify.GetScript() == "setup-namespaces":
        if c.config.Hooks != nil {
            s, err := c.currentOCIState()
            if err != nil {
                return nil
            }
```

I do not figure out how criu notify can get pid, I want invite @avagin and @adrianreber help to check whether the fix in pr is correct.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1554,6 +1554,7 @@ func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Proc
 			if err != nil {
 				return nil
 			}
+			s.Pid = int(notify.GetPid())
 			for i, hook := range c.config.Hooks.Prestart {
 				if err := hook.Run(s); err != nil {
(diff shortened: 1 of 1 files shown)


adrianreber:
So it seems #1741 is the reason for the problem. That PR basically removes, by calling out to a new function, that s.Pid is set.

I also see in #1741 that at a few places s.Pid is set just like it is done in this PR.

So it seems that this PR is doing the right thing. :+1:

Ace-Tang:
Thanks @adrianreber

cyphar:
LGTM, seems reasonable.

crosbymichael:
LGTM

Issue #1326, opened 2017-02-14
Running container killed with failure to write to cgroup.procs


sboschman:
On our Jenkins CI infrastructure we run Maven builds inside a Docker container. Unfortunately once in a while the build container crashes during the execution of the Maven build with a failure writing a pid to the cgroup.proc file.

>Feb` 13 23:56:41 myhost dockerd-current[5659]: time="2017-02-13T23:56:41.695619134+01:00" level=error msg="Error running exec in container: rpc error: code = 13 desc = invalid header field value \"oci runtime error: exec failed: container_linux.go:247: starting container process caused \\\"process_linux.go:87: adding pid 21890 to cgroups caused \\\\\\\"failed to write 21890 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/system.slice/docker-f2ea7bd5f37f4d5719fec4a05fdb58401207c98b6abbfa02e497af4bc167ec08.scope/cgroup.procs: invalid argument\\\\\\\"\\\"\\n\

> Feb 14 04:34:52 myhost dockerd-current[5659]: time="2017-02-14T04:34:52.084545467+01:00" level=error msg="Error running exec in container: rpc error: code = 13 desc = invalid header field value \"oci runtime error: exec failed: container_linux.go:247: starting container process caused \\\"process_linux.go:87: adding pid 36656 to cgroups caused \\\\\\\"failed to write 36656 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/system.slice/docker-2b412ea4871f6b1ed33547224bac344ea677f1814604a501a44e42ce84b64854.scope/cgroup.procs: invalid argument\\\\\\\"\\\"\\n\""

> Feb 14 06:20:22 myhost dockerd-current[5659]: time="2017-02-14T06:20:22.751841415+01:00" level=error msg="Error running exec in container: rpc error: code = 13 desc = invalid header field value \"oci runtime error: exec failed: container_linux.go:247: starting container process caused \\\"process_linux.go:87: adding pid 30239 to cgroups caused \\\\\\\"failed to write 30239 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/system.slice/docker-1418520af414c01054dce8ca3777b616289ee41a0de7ad135af8e2e740472a49.scope/cgroup.procs: invalid argument\\\\\\\"\\\"\\n\""```

[...]

PR #1950, opened 2018-12-17
Resilience in adding of exec tasks to cgroups

BooleanCat:
This is related to https://github.com/opencontainers/runc/issues/1884

This fix allows more leniency for setting up cgroups on exec when cgroup namespaces are not used.

The issue linked above was partly resolved by https://github.com/opencontainers/runc/pull/1916 - we can avoid the issue entirely by using cgroup namespaces. Cloud Foundry's use of runc does not yet use cgroup namespaces and we encounter the issue above frequently.

We believe that since not using cgroup namespaces is a valid configuration of a config.json, runc should be more resilient when we choose not to. This PR will attempt to write to cgroup.procs a few times when execing, only retrying on seeing EINVAL (which, unless you're using realtime processes, means the task's state in the kernel is still TASK_NEW: https://elixir.bootlin.com/linux/v4.8/source/kernel/sched/core.c#L8286)

We replaced the ioutil.WriteFile with a Open/Write dance because we don't want to open the file each time we attempt to write to cgroup.procs, and we reduce the surface for error types changing underneath us within golang.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	units "github.com/docker/go-units"
+	"golang.org/x/sys/unix"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/utils.go
```diff
@@ -463,11 +464,40 @@ func WriteCgroupProc(dir string, pid int) error {
...
-	// Don't attach any pid to the cgroup if -1 is specified as a pid
-	if pid != -1 {
-		if err := ioutil.WriteFile(filepath.Join(dir, CgroupProcesses), []byte(strconv.Itoa(pid)), 0700); err != nil {
-			return fmt.Errorf("failed to write %v to %v: %v", pid, CgroupProcesses, err)
+	// Dont attach any pid to the cgroup if -1 is specified as a pid
+	if pid == -1 {
+		return nil
+	}
+
+	cgroupProcessesFile, err := os.OpenFile(filepath.Join(dir, CgroupProcesses), os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0700)
```
BooleanCat: We also don't necessarily want to use all of these flags to open the file, `O_WRONLY` is probably enough. However we chose not to simplify for two reasons:

1. The previous ioutil.WriteFile does use all three of these flags
2. There was a test that failed when we tried to remove O_CREATE (probably it's a cgroup test that isn't using a real cgroup)

We'd be happy to change this to only O_WRONLY but thought we'd check first.

BooleanCat:
_Friendly bump_

BooleanCat:
@cyphar @crosbymichael Hey friends! Currently Garden is shipping with this in place as a patch since it was causing persistent application push failures in some cloud foundry deployments (which is fixed with this change).

Would be great to get this merge in so we can stop relying on patching runc! Please let me know what you think, thanks!

crosbymichael:
Reviewing this today.  Sorry, it got lost on me.

Thanks for the ping!

libcontainer/cgroups/utils.go
crosbymichael: Ya, the `O_CREATE` is for setting up a mock cgroup in the tests.

crosbymichael:
LGTM

mrunalp:
LGTM

redbaron:
Does it solve https://github.com/opencontainers/runc/issues/1326 ?

BooleanCat:
We've had this running in prod for a while now (even as a gift patch before this got merged) in cloud foundry and have not seen it reoccur.

PR #1953, opened 2018-12-27
Set default CRIU WorkDirectory = ImagesDirectory

rst0git:
CRIU is using both images and work directory to store files during
checkpoint and restore. The work directory contains files such as logs
and irmap cache.

Currently runc sets the default value of the work directory to
`filepath.Join(c.root, "criu.work")`, and in case of failure outputs
the following message:

    # runc restore --image-path /tmp/test/ mycontainer
    criu failed: type NOTIFY errno 0
    log file: /run/runc/mycontainer/criu.work/restore.log

However, this is not very useful because the directory
`/run/runc/mycontainer/criu.work` is removed in case of failure, and
therefore the `restore.log` file does not exist.

This patch changes runc to follow the default behaviour of CRIU by
setting the working directory to be the same as the images directory,
and therefore preserve the log file in case of failure.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -901,7 +901,7 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 	}
 
 	if criuOpts.WorkDirectory == "" {
-		criuOpts.WorkDirectory = filepath.Join(c.root, "criu.work")
+		criuOpts.WorkDirectory = criuOpts.ImagesDirectory
 	}
 
 	if err := os.Mkdir(criuOpts.WorkDirectory, 0755); err != nil && !os.IsExist(err) {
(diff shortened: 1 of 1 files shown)


rst0git:
cc @adrianreber

adrianreber:
This can potentially break runc checkpoint users. I just checked podman and there work directory and image directory is explicitly set for checkpoint and restore operations, so this change will not break podman. Maybe you should also have a look at how docker calls runc for checkpointing and restoring.

rst0git:
Thank you for the feedback!

> Maybe you should also have a look at how docker calls runc for checkpointing and restoring.

Moby is using the checkpoint/restore mechanism of containerd, and this change should not break containerd because the criu's work directory is explicitly set. @Ace-Tang would you be able to confirm that?

Ace-Tang:
From the code, looks like not broke criu function. I test with containerd criu related unit test, it also good.
How about use `runc restore --work-path` to specify work path. 
By the way, default `criu WorkDirectory` locates under root path seems not good.

PR #1954, opened 2018-12-31
cr: Add support for --skip-in-flight

### LINKED ISSUE #1521: Support for CRIU's --skip-in-flight
Hello,

I was trying to checkpoint a container but the dump failed because of this:

```
00.254096) Error (criu/sk-inet.c:159): In-flight connection (l) for 24f02
(00.254101) Error (criu/sk-inet.c:161): In-flight connections can be ignored with the --skip-in-flight option.
(00.254105) ----------------------------------------
(00.254121) Error (criu/cr-dump.c:1334): Dump files (pid: 15266) failed with -1
(00.254613) Unlock network
(00.254620) Unfreezing tasks into 1
(00.260844) Error (criu/cr-dump.c:1657): Dumping FAILED.
```

I've been reading about the **--skip-in-flight** option for CRIU and sounds like this would be the fix. However, runC is not able to pass this as an option to CRIU. Is there anyway that this can be done with what has currently been implemented? Or does it need to be added to runC?

rst0git:
CRIU does not know how to checkpoint/restore TCP connections that are not yet completely established.

With this option CRIU will skip the checkpoint of such in-flight TCP connections, assuming that the TCP stack on the client side can handle the re-connect gracefully.

Closes #1521


adrianreber:
I like this :+1: The CI failure seems unrelated, maybe you can re-trigger the CI run.

crosbymichael:
LGTM

avagin:
I think we can close this pr. We can specify any additional options in a criu config:
https://github.com/opencontainers/runc/blob/master/docs/checkpoint-restore.md#checkpointrestore-annotations

PR #1962, opened 2019-01-11
rootfs: umount all procfs and sysfs with --no-pivot

giuseppe:
When creating a new user namespace, the kernel doesn't allow to mount
a new procfs or sysfs file system if there is not already one instance
fully visible in the current mount namespace.

When using --no-pivot we were effectively inhibiting this protection
from the kernel, as /proc and /sys from the host are still present in
the container mount namespace.

A container without full access to /proc could then create a new user
namespace, and from there able to mount a fully visible /proc, bypassing
the limitations in the container.

A simple reproducer for this issue is:

unshare -mrfp sh -c "mount -t proc none /proc && echo c > /proc/sysrq-trigger"

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -748,6 +748,41 @@ func pivotRoot(rootfs string) error {
 }
 
 func msMoveRoot(rootfs string) error {
+	mountinfos, err := mount.GetMounts()
+	if err != nil {
+		return err
+	}
+
+	absRootfs, err := filepath.Abs(rootfs)
+	if err != nil {
+		return err
+	}
+
+	for _, info := range mountinfos {
+		p, err := filepath.Abs(info.Mountpoint)
+		if err != nil {
+			return err
+		}
+		// Umount every syfs and proc file systems, except those under the container rootfs
+		if (info.Fstype != "proc" && info.Fstype != "sysfs") || filepath.HasPrefix(p, absRootfs) {
+			continue
+		}
+		// Be sure umount events are not propagated to the host.
+		if err := unix.Mount("", p, "", unix.MS_SLAVE|unix.MS_REC, ""); err != nil {
+			return err
+		}
+		if err := unix.Unmount(p, unix.MNT_DETACH); err != nil {
+			if err != unix.EINVAL && err != unix.EPERM {
+				return err
+			} else {
+				// If we have not privileges for umounting (e.g. rootless), then
+				// cover the path.
+				if err := unix.Mount("tmpfs", p, "tmpfs", 0, ""); err != nil {
+					return err
+				}
+			}
+		}
+	}
 	if err := unix.Mount(rootfs, "/", "", unix.MS_MOVE, ""); err != nil {
 		return err
 	}


mrunalp:
LGTM

libcontainer/rootfs_linux.go
```diff
@@ -748,6 +749,41 @@ func pivotRoot(rootfs string) error {
...
+		// Umount every syfs and proc file systems, except those under the container rootfs
+		if (info.Fstype != "proc" && info.Fstype != "sysfs") || filepath.HasPrefix(p, absRootfs) {
+			continue
+		}
+		// Be sure umount events are not propagated to the host.
+		if err := unix.Mount("", p, "", unix.MS_SLAVE|unix.MS_REC, ""); err != nil {
+			return err
+		}
+		if err := unix.Unmount(p, unix.MNT_DETACH); err != nil {
+			if err.(syscall.Errno) != unix.EINVAL {
```
AkihiroSuda: Potentially we can also get `EPERM` here?

giuseppe: I guess, but I never saw that error message, when trying to umount `/proc` or `/sys` I got only `EINVAL`. I can amend the patch if you'd like

giuseppe: I went forward and amended the change in the updated version

libcontainer/rootfs_linux.go
```diff
@@ -748,6 +749,41 @@ func pivotRoot(rootfs string) error {
...
+		// Umount every syfs and proc file systems, except those under the container rootfs
+		if (info.Fstype != "proc" && info.Fstype != "sysfs") || filepath.HasPrefix(p, absRootfs) {
+			continue
+		}
+		// Be sure umount events are not propagated to the host.
+		if err := unix.Mount("", p, "", unix.MS_SLAVE|unix.MS_REC, ""); err != nil {
+			return err
+		}
+		if err := unix.Unmount(p, unix.MNT_DETACH); err != nil {
+			if err.(syscall.Errno) != unix.EINVAL && err.(syscall.Errno) != unix.EPERM {
```
cyphar: The `.(syscall.Errno)` is not necessary.

giuseppe: fixed

mrunalp:
LGTM

cyphar:
LGTM.

libcontainer/rootfs_linux.go
```diff
@@ -748,6 +748,41 @@ func pivotRoot(rootfs string) error {
 }
 
 func msMoveRoot(rootfs string) error {
+	mountinfos, err := mount.GetMounts()
+	if err != nil {
+		return err
+	}
+
+	absRootfs, err := filepath.Abs(rootfs)
```
kolyshkin: AFAICS this is not needed since rootfs is already validated by `(*ConfigValidator).rootfs()`

libcontainer/rootfs_linux.go
```diff
@@ -748,6 +748,41 @@ func pivotRoot(rootfs string) error {
...
+	if err != nil {
+		return err
+	}
+
+	for _, info := range mountinfos {
+		p, err := filepath.Abs(info.Mountpoint)
```
kolyshkin: not needed either as paths returned from the kernel are absolute and clean (unless I'm missing something).

I have addressed this in https://github.com/opencontainers/runc/pull/2255 (a separate commit to keep things simple), PTAL @giuseppe

PR #1963, opened 2019-01-15
Vendor in go-criu and use it for CRIU's RPC definition

adrianreber:
This PR vendors in CRIU's Go bindings. Right now this only replaces the manually updated CRIU RPC definition file. This PR does not change any functionality. It just deletes one file and replaces it with the same file from the go-criu repository.

Where possible we are planing to replace CRIU related runc code with functions provided by go-criu.


rst0git:
@adrianreber A commit that adds an interface for the `--tcp-close` option to RPC was merged in the criu-dev branch. Should we wait for the next CRIU release before adding support for this option in runc?

adrianreber:
@rst0git Concerning `--tcp-close`: I would make it dependent on if it is possible to detect if the installed CRIU supports it. Add a feature check via RPC to CRIU for `--tcp-close`, then update go-criu, the update runc. This way runc can always ask if CRIU supports `--tcp-close` via RPC, if CRIU supports runc can use it, independent if the feature (or interface) is only available in CRIU's development branch. Otherwise you would need to add a new version check  to make sure runc can use the RPC option for `--tcp-close`. I think CRIU versions which do not know about `--tcp-close` via RPC would probably just ignore the option and the user thinks `--tcp-close` was active and runc passes it down to CRIU, but CRIU would just ignore it. Not really a perfect user experience... But as probably not many people would be using it anyway, so probably not really important.

adrianreber:
Any more questions concerning this PR? From my point of view it is ready to be merged.

vendor.conf
```diff
@@ -3,6 +3,7 @@
 # using.
 github.com/opencontainers/runtime-spec 5684b8af48c1ac3b1451fa499724e30e3c20a294
 # Core libcontainer functionality.
+github.com/checkpoint-restore/go-criu master
```
crosbymichael: can you pin to a specific commit or tag and not master?

adrianreber: Good point. When I initially made this PR we did not have tags in go-criu, but other projects have asked for the same. Now that we have tags (which are following the CRIU releases) I changed it to v3.11

adrianreber:
Rebased, switch to tag v3.11 for criu in vendor.conf and force pushed.

crosbymichael:
LGTM

adrianreber:
Any further comments from other runc maintainers?

mrunalp:
LGTM

PR #1964, opened 2019-01-15
Document 'org.criu.config' annotation

adrianreber:
This documents the newly introduced 'org.criu.config' annotation.

See #1933


adrianreber:
@avagin, any comments regarding this documentation?

cyphar:
LGTM, seems reasonable to me.

PR #1966, opened 2019-01-22
ensure we call wait() on children created by nsenter

pepin:
we were ending up with 1 zombie for every container created, since
nothing was waiting on the intermediate child created by nsenter.

Signed-off-by: James Pepin <james.pepin@gmail.com>


vansante:
I have the same issue, so I tried the patch and it solved it.

So a :+1: from me.

crosbymichael:
This should be fixed by #2023

PR #1967, opened 2019-01-24
integration: fix mis-use of libcontainer.Factory

cyphar:
For some reason, libcontainer/integration has a whole bunch of incorrect
usages of libcontainer.Factory -- causing test failures with a set of
security patches that will be published soon. Fixing ths is fairly
trivial (switch to creating a new libcontainer.Factory once in each
process, rather than creating one in TestMain globally).


crosbymichael:
LGTM

mrunalp:
LGTM

PR #1968, opened 2019-01-30
Create bind mount mountpoints during restore

adrianreber:
runc creates all bind mount mountpoints when it starts a container, this commit also creates those mountpoints during restore. Now it is possible to restore a container using the same, but newly created rootfs just as during container start.

Additionally this PR wires the CRIU option `--ext-mount-map auto` through to make it available during container restore using `runc restore --autodetect-external-mounts`.

These changes are necessary for https://github.com/containers/libpod/issues/1618

CC: @avagin, @rst0git this are some of the necessary changes as described on the mailing list

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -19,6 +19,7 @@ import (
 	"syscall" // only for SysProcAttr and Signal
 	"time"
 
+	"github.com/cyphar/filepath-securejoin"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/criurpc"
@@ -1139,6 +1140,75 @@ func (c *linuxContainer) restoreNetwork(req *criurpc.CriuReq, criuOpts *CriuOpts
 	}
 }
 
+// makeCriuRestoreMountpoints makes the actual mountpoints for the
+// restore using CRIU. This function is inspired from the code in
+// rootfs_linux.go
+func (c *linuxContainer) makeCriuRestoreMountpoints(m *configs.Mount) error {
+	switch m.Device {
+	case "cgroup":
+		// Do nothing for cgroup, CRIU should handle it
+	case "bind":
+		// The prepareBindMount() function checks if source
+		// exists. So it cannot be used for other filesystem types.
+		if err := prepareBindMount(m, c.config.Rootfs); err != nil {
+			return err
+		}
+	default:
+		// for all other file-systems just create the mountpoints
+		dest, err := securejoin.SecureJoin(c.config.Rootfs, m.Destination)
+		if err != nil {
+			return err
+		}
+		if err := checkMountDestination(c.config.Rootfs, dest); err != nil {
+			return err
+		}
+		m.Destination = dest
+		if err := os.MkdirAll(dest, 0755); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// isPathInPrefixList is a small function for CRIU restore to make sure
+// mountpoints, which are on a tmpfs, are not created in the roofs
+func isPathInPrefixList(path string, prefix []string) bool {
+	for _, p := range prefix {
+		if strings.HasPrefix(path, p+"/") {
+			return false
+		}
+	}
+	return true
+}
+
+// prepareCriuRestoreMounts tries to set up the rootfs of the
+// container to be restored in the same way runc does it for
+// initial container creation. Even for a read-only rootfs container
+// runc modifies the rootfs to add mountpoints which do not exist.
+// This function also creates missing mountpoints as long as they
+// are not on top of a tmpfs, as CRIU will restore tmpfs content anyway.
+func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error {
+	// First get a list of a all tmpfs mounts
+	tmpfs := []string{}
+	for _, m := range mounts {
+		switch m.Device {
+		case "tmpfs":
+			tmpfs = append(tmpfs, m.Destination)
+		}
+	}
+	// Now go through all mounts and create the mountpoints
+	// if the mountpoints are not on a tmpfs, as CRIU will
+	// restore the complete tmpfs content from its checkpoint.
+	for _, m := range mounts {
+		if isPathInPrefixList(m.Destination, tmpfs) {
+			if err := c.makeCriuRestoreMountpoints(m); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
 func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 	c.m.Lock()
 	defer c.m.Unlock()
@@ -1250,6 +1320,12 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 			// All open FDs need to be transferred to CRIU via extraFiles
 			extraFiles = append(extraFiles, netns)
 		}
+	}
+
+	// This will modify the rootfs of the container in the same way runc
+	// modifies the container during initial creation.
+	if err := c.prepareCriuRestoreMounts(c.config.Mounts); err != nil {
+		return err
 	}
 
 	for _, m := range c.config.Mounts {
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -182,6 +182,33 @@ func mountCmd(cmd configs.Command) error {
 	return nil
 }
 
+func prepareBindMount(m *configs.Mount, rootfs string) error {
+	stat, err := os.Stat(m.Source)
+	if err != nil {
+		// error out if the source of a bind mount does not exist as we will be
+		// unable to bind anything to it.
+		return err
+	}
+	// ensure that the destination of the bind mount is resolved of symlinks at mount time because
+	// any previous mounts can invalidate the next mount's destination.
+	// this can happen when a user specifies mounts within other mounts to cause breakouts or other
+	// evil stuff to try to escape the container's rootfs.
+	var dest string
+	if dest, err = securejoin.SecureJoin(rootfs, m.Destination); err != nil {
+		return err
+	}
+	if err := checkMountDestination(rootfs, dest); err != nil {
+		return err
+	}
+	// update the mount with the correct dest after symlinks are resolved.
+	m.Destination = dest
+	if err := createIfNotExists(dest, stat.IsDir()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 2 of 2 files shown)


libcontainer/container_linux.go
```diff
@@ -1215,6 +1215,7 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 			OrphanPtsMaster: proto.Bool(true),
 			AutoDedup:       proto.Bool(criuOpts.AutoDedup),
 			LazyPages:       proto.Bool(criuOpts.LazyPages),
+			AutoExtMnt:      proto.Bool(criuOpts.AutodetectExternalMounts),
 		},
```
rst0git: Why not set AutoExtMnt by default, instead of adding a cli option (is there a case where we don't want to auto-detect of external mounts)?
LXC seems to do that and I haven't seen issues related to this, so far.

adrianreber: Basically I did not want to change any existing behaviour. No idea if setting it could break something. Always setting it would be even easier for me.

avagin: > Why not set AutoExtMnt by default, instead of adding a cli option (is there a case where we don't want to auto-detect external mounts)?

Because this option has "undefined" behavior and if you are going to migrate a container to another host, I don't recommend to use this option.

adrianreber: @avagin Thanks for letting us know that you do not recommend to use it. Concerning this pull request it does not hurt or break anything, so it could be added without a problem.

I am currently using it as there seems to be a problem between checkpointing and restoring with runc. During checkpoint runc let's CRIU know that it /dev/shm (in my case) is an external bind mount (correct). In the external mount mapping I have key of /dev/shm with a value of /dev/shm. During restore runc does not tell CRIU what to do with the /dev/shm mountpoint. So CRIU fails with something like: missing mapping for /dev/shm. Autodetect external mounts fixes it (or works around it). This can/needs to be fixed probably in a different way in runc.

avagin: First of all, we need to understand why an ext mount for /dev/shm isn't set on restore. Can you describe how I can reproduce this?

adrianreber: I figured it out, it was missing from the config.json during restore. If I do it correctly it actually works :wink: 

So I do not really need this option, but we could still add it to runc for people that need it.

adrianreber:
Any other comments? From my point of view `runc restore --autodetect-external-mounts` does not block this PR from being merged as it only optionally exposes a CRIU feature, but it does not change `runc`'s current behaviour.

rst0git:
LGTM

adrianreber:
@rst0git I just pushed one more commit which now also creates /proc and /sys on restore just as they are created during initial container creation. With this your migration test from https://github.com/containers/libpod/pull/2272 should now work.

rst0git:
@adrianreber I can confirm that the new commit resolves the issue, but is there a more generic solution, rather than specifying `proc` and `sysfs`?

adrianreber:
> @adrianreber I can confirm that the new commit resolves the issue, but is there a more generic solution, rather than specifying `proc` and `sysfs`?

Not sure, I just tried to do the same as in `rootfs_linux.go`.

libcontainer/container_linux.go
avagin: @adrianreber It is the main question whether we have users who need it or not. I think this option is only for experiments and it should be never used in production, so I would recommend to not add it in runsc.

libcontainer/rootfs_linux.go
```diff
@@ -182,6 +182,33 @@ func mountCmd(cmd configs.Command) error {
 	return nil
 }
 
+func prepareBindMounts(m *configs.Mount, rootfs string) error {
```
avagin: I think it should be prepareBindMount without s, because it handles only one mount

libcontainer/container_linux.go
```diff
@@ -1255,6 +1255,13 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 	for _, m := range c.config.Mounts {
 		switch m.Device {
 		case "bind":
+			// When runc starts a container it creates all mountpoints
+			// for bind mounts. Let's also create those mountpoints
+			// on restore to be able to restore a container created
+			// from an unmodified container rootfs.
+			if err := prepareBindMounts(m, c.config.Rootfs); err != nil {
```
avagin: The same problem exists for tmpfs mounts which are specified in a container config. And there can be dependencies between mounts. We can mount tmpfs into /mnt and than bind-mount a host directory into /mnt/bind. This code will try to create /mnt/bind on a container rootfs and it will fail of rootfs is read-only. 

The only right way to handle this is to snapshot a container rootfs.

adrianreber: But runc will also fail creating a container if the rootfs is read-only.

adrianreber: So at least for a read-only rootfs container we should to the same steps during restore as during creation. It does not really make sense to take a snapshot of a read-only rootfs container as it is expected to be the same, but runc will still try to modify it.

libcontainer/container_linux.go
adrianreber: I will remove this commit.

libcontainer/rootfs_linux.go
adrianreber: done

libcontainer/container_linux.go
avagin: @adrianreber I don't understand me. Here are steps for creating a container:
1. createIfNotExists(/mnt). A container rootfs has /mnt, so this function does nothing and exits without errors
2. mount tmpfs into /mnt
3. createIfNotExists(/mnt/A) - create a directory A on a tmpfs mount
4. bind-mount a host path to /mnt/A

In case of restore, /mnt/A will be created on a container rootfs, what looks wrong, doesn't it?

adrianreber: > In case of restore, /mnt/A will be created on a container rootfs, what looks wrong, doesn't it?

Not sure I understand you. Is this during restore or initial container creation?

If it is a tmpfs CRIU will anyway untar the content from the checkpoint into the tmpfs, so all tmpfs content will be handled correctly by CRIU internally.

If running the busybox container using podman, the downloaded container image (docker://busybox:latest) will not contain the directory /proc. runc will create /proc during container startup and mount proc at /proc.

If I checkpoint this container CRIU will remember that it mounted proc at /proc and try to restore the mount and fail.

I think there is a difference between changes done to the roof filesystem from the programs running in the container and changes done to the container by runc during startup.

Changes done to the container during startup, like creating /proc are necessary to restore the container.

I think restoring a container which does not modify the filesystem (or only using tmpfs) is a valid use case and because runc modifies the rootfs during startup we need to do the same modifications during restore. It seems unnecessary to have to transfer the changes done to the rootfs by runc during startup.

If the container (from within) modifies the root filesystem I agree we need to do a commit and export these changes to the new container. As long as the changes are done by runc, I think runc needs to handle the restore.

avagin: > Not sure I understand you. Is this during restore or initial container creation?

This is an initial container creation.  I just show you that when runsc start a continaer, /mnt/A is created on a tmpfs mount, but on restore (with these changes) /mnt/A will be created on a container rootfs. You are right, criu will restore the tmpfs mount, but it will be later.

adrianreber: Ah, okay, I get it. If a bind mount mountpoint is on a tmpfs runc will now modify the underlying rootfs, before CRIU mounts the tmpfs, understood. So skipping bind mount mountpoints would be useful, correct. Need to check what I need to do, to do that correctly.

Do you see any other problems if I can actually do the bind mount mountpoint on tmpfs correctly?

avagin: I don't understand why you don't want to save a commit of  the highest layer of rootfs together with criu dump files. If we merge these changes, it will fix your test, but the problem will not disappear for cases when a container is started with a read-write rootfs and processes create files on it.

adrianreber: Hmm... You are right, if the container modifies the rootfs the highest layer of the rootfs needs to be part of the checkpoint. What I am unsure about is the changes done to the rootfs by runc.

The correct thing to do would be that runc does not modify the rootfs at all. This, however, cannot be changed as it would break a lot of containers. This seems to be the root of this discussion.

My point is, that if runc changes the rootfs during creation (especially if it is a read-only container) then the same modification should happen during restore. Without this it will not be possible to migrate a container with a read-only rootfs, because runc will change read-only rootfs containers. Having to include the highest layer of a read-only rootfs container also seems wrong.

I understand your point of view and I would agree if runc would not modify the rootfs. Because runc modifies the rootfs during creation I think it can also be expected during restore.

Not sure how to continue this discussion. Maybe some comments from the runc maintainers would be helpful. How they see the changes to the rootfs done by runc. If these rootfs changes should also happen during restore or not. I guess my biggest problem with this is, is that runc does changes to the container, if this would not be case we would not need this whole discussion.

avagin: > Do you see any other problems

Now you enumerate only bind-mounts, /proc and /sys, but I think you need to create destination directories for tmpfs mounts too.

avagin:
> So skipping bind mount mountpoints would be useful, correct. Need to check what I need to do, to do that correctly.

This patch set will look good to me when you fix this issue. Thanks!

libcontainer/container_linux.go
adrianreber: > > Do you see any other problems
> 
> Now you enumerate only bind-mounts, /proc and /sys, but I think you need to create destination directories for tmpfs mounts too.

Yes, just tested it and it failed. Will update this PR.

adrianreber:
@avagin Thanks for your feedback.

adrianreber:
Updated and force pushed. This time mountpoints for all type of mounts are created during restore (just as during creation), but this time no mountpoints are made if the are overmounted by tmpfs.

adrianreber:
Any comments for someone else?

adrianreber:
Can I get two more reviews to get this merged if ready?

adrianreber:
Any more comments from reviewers for this?

mrunalp:
LGTM

mrunalp:

crosbymichael:
LGTM

libcontainer/container_linux.go
```diff
@@ -1139,6 +1140,75 @@ func (c *linuxContainer) restoreNetwork(req *criurpc.CriuReq, criuOpts *CriuOpts
...
+	}
+	return nil
+}
+
+// isPathInPrefixList is a small function for CRIU restore to make sure
+// mountpoints, which are on a tmpfs, are not created in the roofs
+func isPathInPrefixList(path string, prefix []string) bool {
+	for _, p := range prefix {
+		if strings.HasPrefix(path, p+"/") {
+			return false
```
avagin: According to the function name, it has to return true in this cases, doesn't it?

PR #1970, opened 2019-02-03
Make public ParseMountinfo

BooleanCat:
Whilst using libcontainer/mount.GetMounts() I had need of the underlying logic of parsing /proc/self/mountinfo for testing purposes. This PR makes that logic public, allowing for parsing of a Reader whose contents look like /proc/self/mountinfo.

### DIFF
--- a/libcontainer/mount/mount_linux.go
+++ b/libcontainer/mount/mount_linux.go
@@ -37,10 +37,10 @@ func parseMountTable() ([]*Info, error) {
 	}
 	defer f.Close()
 
-	return parseInfoFile(f)
+	return ParseMountinfo(f)
 }
 
-func parseInfoFile(r io.Reader) ([]*Info, error) {
+func ParseMountinfo(r io.Reader) ([]*Info, error) {
 	var (
 		s   = bufio.NewScanner(r)
 		out = []*Info{}


crosbymichael:
@caniszczyk for the pullapprove issues with billing

AkihiroSuda:
LGTM

kolyshkin:
@BooleanCat you can use "github.com/docker/docker/pkg/mount".GetMounts() (see https://github.com/moby/moby/blob/master/pkg/mount/mount.go)

I'm against making this one public, we might change it in the future and would have to maintain backward compatibility

AkihiroSuda:
@kolyshkin Not huge fan of introducing dependency on moby/moby in this repo. Maybe the package should be split out to a separate repo?

vrothberg:
How about moving the package from Moby here? I believe that (nearly) all projects needing it already have runc a code dependency. It would also avoid a circular dependencies between Moby and runc (on a project level).

AkihiroSuda:
Closing per https://github.com/opencontainers/runc/pull/2256 .
Moving `moby/sys` into `opencontainers` can be discussed in a separate issue.

kolyshkin:
@BooleanCat let me update my recommendation: please use `mountinfo` package from https://github.com/moby/sys; for example, see #2256

PR #1972, opened 2019-02-07
Update vendored golang.org/x/sys to latest

filbranden:
Update vendored repository from https://github.com/golang/sys/commit/7ddbeae9ae08c6a06a59597f0c9edbc5ff2444ce (September 2017) to https://github.com/golang/sys/commit/41f3e6584952bb034a481797859f6ab34b6803bd (February 2019).

This will be needed for the `unix.CGROUP_SUPER_MAGIC`, `unix.CGROUP2_SUPER_MAGIC` and `unix.TMPFS_MAGIC` constants for the result of `unix.Statfs()`, which can be used to detect whether the system is running with cgroupv2 only.

Sending this as a separate PR to make it easy to review it, since the only change is to `vendor.conf`, the other files are only updated as a decorrence of running `vndr` with the update.


crosbymichael:
LGTM

cyphar:
LGTM.

filbranden:
Thanks!

I'm expecting the next ones will not be easy and will require some back and forth, so taking this out of the way in an easy PR means I'll have less trouble with rebases and merge conflicts etc... In any case, thanks for the prompt review and merge!

Cheers,
Filipe

PR #1976, opened 2019-02-08
libcontainer: cgroup namespaces enabled utility function

### LINKED ISSUE #1975: Add utility function for checking if cgroup namespaces are enabled
In the review for adding cgroup namespace support to docker (https://github.com/moby/moby/pull/38377), it was suggested that checking if cgroup namespaces are enabled should be moved into a shared utility function.

rgulewich:
Move code for checking if cgroup namespaces are enabled into a utility function that can be used in other projects.

Fixes #1975.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -22,6 +22,19 @@ const (
 	CgroupProcesses  = "cgroup.procs"
 )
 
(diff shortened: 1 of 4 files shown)


cyphar:
I'm not a huge fan of this idea personally -- it's an incredibly trivial check and I don't see the benefit of making it a utility function.

rgulewich:
https://github.com/moby/moby/pull/38377 was merged without this: closing.

PR #1978, opened 2019-02-12
Remove detection for scope properties, which have always been broken

filbranden:
The detection for scope properties (whether scope units support `DefaultDependencies=` or `Delegate=`) has always been broken, since systemd refuses to create scopes unless at least one PID is attached to it (and this has been so since scope units were introduced in systemd v205, see systemd commit [v204-230-g6c12b52e1964](https://github.com/systemd/systemd/commit/6c12b52e19640747e96f89d85422941a23dc6b29).)

This can be seen in journal logs whenever a container is started with libpod:

```
Feb 11 15:08:07 myhost systemd[1]: libcontainer-12345-systemd-test-default-dependencies.scope: Scope has no PIDs. Refusing.
Feb 11 15:08:07 myhost systemd[1]: libcontainer-12345-systemd-test-default-dependencies.scope: Scope has no PIDs. Refusing.
```

Since this logic never worked, just assume both attributes are supported (which is what the code does when detection fails for this reason, since it's looking for an "unknown attribute" or "read-only attribute" to mark them as false) and skip the detection altogether.

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -72,13 +72,11 @@ const (
 )
 
 var (
-	connLock                        sync.Mutex
-	theConn                         *systemdDbus.Conn
-	hasStartTransientUnit           bool
-	hasStartTransientSliceUnit      bool
-	hasTransientDefaultDependencies bool
-	hasDelegateScope                bool
-	hasDelegateSlice                bool
+	connLock                   sync.Mutex
+	theConn                    *systemdDbus.Conn
+	hasStartTransientUnit      bool
+	hasStartTransientSliceUnit bool
+	hasDelegateSlice           bool
 )
 
 func newProp(name string, units interface{}) systemdDbus.Property {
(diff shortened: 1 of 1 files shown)


dqminh:
Maybe its possible for us to inject a dummy command/pid to the scope to detect this properly, however i'm not sure if its worth it, given that most stable distribution now ships systemd with capabilities that we are checking here i think.

@mrunalp @crosbymichael any thoughts ?

filbranden:
> Maybe its possible for us to inject a dummy command/pid to the scope to detect this properly, however i'm not sure if its worth it [...]

Yes, my thoughts exactly. Since the introduction of the feature, this detection never really worked (was always assumed "true") and nobody ever complained, so probably OK to keep this now...

I'm actually also looking at possibly removing the slice detection too (and just keeping Delegate=no there), since in latest systemd that's not possible and in older systemd it [didn't do much](https://lists.freedesktop.org/archives/systemd-devel/2019-February/042161.html). But that's for another PR. 😄 

Thanks for the quick review!
Filipe

mrunalp:
LGTM

crosbymichael:
LGTM

PR #1981, opened 2019-02-13
nsenter (CVE-2019-5736): remove unneeded parsing

brauner:
A __attribute__((constructor)) can get access to argc and argv. With
this there's no need to parse /proc/<pid>/{cmdline,environ}.

Fixes: CVE-2019-5736
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/libcontainer/nsenter/nsenter.go
+++ b/libcontainer/nsenter/nsenter.go
@@ -5,8 +5,8 @@ package nsenter
 /*
 #cgo CFLAGS: -Wall
 extern void nsexec();
-void __attribute__((constructor)) init(void) {
-	nsexec();
+void __attribute__((constructor)) init(int argc, char *argv[]) {
+	nsexec(argv);
 }
 */
 import "C"
--- a/libcontainer/nsenter/nsenter_gccgo.go
+++ b/libcontainer/nsenter/nsenter_gccgo.go
@@ -5,8 +5,8 @@ package nsenter
 /*
 #cgo CFLAGS: -Wall
 extern void nsexec();
-void __attribute__((constructor)) init(void) {
-	nsexec();
+void __attribute__((constructor)) init(int argc, char *argv[]) {
+	nsexec(argv);
 }
 */
 import "C"
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -535,9 +535,9 @@ void join_namespaces(char *nslist)
 }
 
 /* Defined in cloned_binary.c. */
-extern int ensure_cloned_binary(void);
+extern int ensure_cloned_binary(char *argv);
 
-void nsexec(void)
+void nsexec(char *argv[])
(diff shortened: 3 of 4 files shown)


libcontainer/nsenter/nsexec.c
```diff
@@ -535,9 +535,9 @@ void join_namespaces(char *nslist)
 }
 
 /* Defined in cloned_binary.c. */
-extern int ensure_cloned_binary(void);
+extern int ensure_cloned_binary(char *argv);
 
-void nsexec(void)
+void nsexec(int argc, char *argv)
```
cyphar: `char **argv`?

libcontainer/nsenter/nsenter_gccgo.go
```diff
@@ -5,8 +5,8 @@ package nsenter
 /*
 #cgo CFLAGS: -Wall
 extern void nsexec();
-void __attribute__((constructor)) init(void) {
-	nsexec();
+void __attribute__((constructor)) init(int argc, char *argv[]) {
+	nsexec(argv);
```
cyphar: You forgot to pass `argc` here.

libcontainer/nsenter/nsenter.go
```diff
@@ -5,8 +5,8 @@ package nsenter
 /*
 #cgo CFLAGS: -Wall
 extern void nsexec();
-void __attribute__((constructor)) init(void) {
-	nsexec();
+void __attribute__((constructor)) init(int argc, char *argv[]) {
+	nsexec(argv);
```
cyphar: And here.

libcontainer/nsenter/nsenter_gccgo.go
brauner: We don't really need `argc` but I can pass it if you want to make use of it in the future.

libcontainer/nsenter/nsenter.go
brauner: We don't really need `argc` but I can pass it if you want to make use of it in the future.

libcontainer/nsenter/nsenter_gccgo.go
brauner: Actually, let's remove `argc` from `nsexec()` for now.

libcontainer/nsenter/nsenter.go
brauner: Actually, let's remove `argc` from `nsexec()` for now.

brauner:
@cyphar, wait a bit. I want to verify a few things first.

brauner:
I'll re-open soonish.

libcontainer/nsenter/nsenter_gccgo.go
cyphar: My point was that `nsexec` was defined with `argc` but you didn't pass it. :P

PR #1983, opened 2019-02-14
Avoid repeated os.SetEnv calls in container init.

brandon-mabey:
In a Kubernetes cluster with a large number of services (>5000, causing a large amount of environment variables for the container) with guaranteed pods, the `runc init` process can take a long time to run. This is caused by two issues:

1. `runc init` calls `os.SetEnv` for every environment variable in the bundle, which can be expensive due to the underlying mutex locking and syscall logic. 
2. The cgroup is limited before the `runc init` process runs. This is outside the perview of runc I believe.

This PR targets the former by removing the `os.SetEnv` calls, which causes the `runc init` process to run much faster in cases where there are a large number of configured environment variables in the bundle.

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -79,47 +79,77 @@ func newContainerInit(t initType, pipe *os.File, consoleSocket *os.File, fifoFd 
 	if err := json.NewDecoder(pipe).Decode(&config); err != nil {
 		return nil, err
 	}
-	if err := populateProcessEnvironment(config.Env); err != nil {
+	environmentMap, err := populateEnvironmentMap(config.Env)
+	if err != nil {
 		return nil, err
 	}
+
 	switch t {
 	case initSetns:
 		return &linuxSetnsInit{
-			pipe:          pipe,
-			consoleSocket: consoleSocket,
-			config:        config,
+			pipe:           pipe,
+			consoleSocket:  consoleSocket,
+			config:         config,
+			environmentMap: environmentMap,
 		}, nil
 	case initStandard:
 		return &linuxStandardInit{
-			pipe:          pipe,
-			consoleSocket: consoleSocket,
-			parentPid:     unix.Getppid(),
-			config:        config,
-			fifoFd:        fifoFd,
+			pipe:           pipe,
+			consoleSocket:  consoleSocket,
+			parentPid:      unix.Getppid(),
+			config:         config,
+			fifoFd:         fifoFd,
+			environmentMap: environmentMap,
 		}, nil
 	}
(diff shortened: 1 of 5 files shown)


libcontainer/init_linux.go
```diff
@@ -79,47 +79,77 @@ func newContainerInit(t initType, pipe *os.File, consoleSocket *os.File, fifoFd
...
+		if err != nil {
+			return nil, err
 		}
+		mappedEnv[key] = pair
 	}
-	return nil
+
+	// Certain functions that are used later, such as exec.LookPath, require that the PATH
+	// environment variable is setup with the container's PATH.
+	os.Setenv("PATH", mappedEnv["PATH"])
```
AkihiroSuda: Could you be more specific? Wondering it may cause security issues.

kolyshkin: 1. This does not causes any security issues because this is what was done before the patch, too.
2. This appears to be an error because `mappedEnv["PATH"]` contains, for example, `PATH=/usr/bin:/usr/sbin` (rather than `/usr/bin:/usr/sbin`. As a result, the first PATH element will be wrong.

libcontainer/init_linux.go
```diff
@@ -319,10 +349,8 @@ func setupUser(config *initConfig) error {
 	}
 
 	// if we didn't get HOME already, set it based on the user's HOME
-	if envHome := os.Getenv("HOME"); envHome == "" {
-		if err := os.Setenv("HOME", execUser.Home); err != nil {
-			return err
-		}
+	if _, ok := environmentMap["HOME"]; !ok {
+		environmentMap["HOME"] = "HOME" + "=" + execUser.Home
```
kolyshkin, line 353: Instead, it's easier to just set HOME the old way (like it's done with PATH).

kolyshkin:
This looks interesting. Would be nice to add a benchmark.

kolyshkin:
Did some quick benchmark with 5000 env vars, comparing `os.Setenv` with passing env via cmd.Env. The latter is indeed faster!

```
name     old time/op    new time/op    delta
Exec-20    4.71ms ±33%    2.08ms ±13%  -55.88%  (p=0.000 n=10+9)

name     old alloc/op   new alloc/op   delta
Exec-20    1.43MB ± 0%    0.46MB ± 0%  -67.83%  (p=0.000 n=10+9)

name     old allocs/op  new allocs/op  delta
Exec-20     5.19k ± 0%     0.03k ± 0%  -99.46%  (p=0.000 n=10+10)
```

Will carry this one.

kolyshkin:
Closing in favor of #4325

PR #1984, opened 2019-02-15
nsenter: cloned_binary: "memfd" cleanups

### LINKED ISSUE #1979: [CVE-2019-5736]: fix only suit for kernel > 3.11 ?
https://github.com/opencontainers/runc/commit/0a8e4117e7f715d5fbeef398405813ce8e88558b is a good work to fix CVE-2019-5736, but I get some problem with the patch on kernel 3.10.

The background is that our production online use two major kernel 3.10 and 4.9. In patch, `memfd_create` supported until kernel 3.17, other method use `O_TMPFILE`, it is supported until 3.11. 

the other confused me is that with these code
```
/* Use our own wrapper for memfd_create. */
#if !defined(SYS_memfd_create) && defined(__NR_memfd_create)
#  define SYS_memfd_create __NR_memfd_create
#endif
#ifdef SYS_memfd_create
#  define HAVE_MEMFD_CREATE
/* memfd_create(2) flags -- copied from <linux/memfd.h>. */
#  ifndef MFD_CLOEXEC
#    define MFD_CLOEXEC       0x0001U
#    define MFD_ALLOW_SEALING 0x0002U
#  endif
int memfd_create(const char *name, unsigned int flags)
{
    return syscall(SYS_memfd_create, name, flags);
}
#endif
```
even I run code on kernel 2.6,  `HAVE_MEMFD_CREATE` always been defined, but `memfd_create` not work.

### LINKED ISSUE #1980: [CVE-2019-5736]: Runc uses more memory during start up after the fix
We observed higher memory usage (likely during container startup) after the fix for CVE https://github.com/opencontainers/runc/commit/0a8e4117e7f715d5fbeef398405813ce8e88558b.

We had a test that specifies 10m container cgroup limit, which never failed before, but now the container get oom-killed a lot. For example https://gubernator.k8s.io/build/kubernetes-jenkins/logs/ci-containerd-node-e2e-1-2/2500.
```
kernel: runc:[2:INIT] invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=998
kernel: runc:[2:INIT] cpuset=80e651c417ebd71d83e5023ee59b281e585497468bd71ee7c7b3ae6730d9ec8f mems_allowed=0
kernel: CPU: 0 PID: 333 Comm: runc:[2:INIT] Not tainted 4.4.64+ #1
kernel: Hardware name: Google Google Compute Engine/Google Compute Engine, BIOS Google 01/01/2011
kernel:  0000000000000000 ffff880003e87ca8 ffffffff9f317334 ffff880003e87d88
kernel:  ffff8800bb3e8000 ffff880003e87d18 ffffffff9f1a8fb4 ffff880003e87ce0
kernel:  ffffffff9f13e780 ffff8800bb3eb500 0000000000000206 ffff880003e87cf0
kernel: Call Trace:
kernel:  [<ffffffff9f317334>] dump_stack+0x63/0x8f
kernel:  [<ffffffff9f1a8fb4>] dump_header+0x65/0x1d4
kernel:  [<ffffffff9f13e780>] ? find_lock_task_mm+0x20/0xb0
kernel:  [<ffffffff9f13ef1d>] oom_kill_process+0x28d/0x430
kernel:  [<ffffffff9f1a3e6b>] ? mem_cgroup_iter+0x1db/0x390
kernel:  [<ffffffff9f1a6374>] mem_cgroup_out_of_memory+0x284/0x2d0
kernel:  [<ffffffff9f1a6de9>] mem_cgroup_oom_synchronize+0x2f9/0x310
kernel:  [<ffffffff9f1a1ab0>] ? memory_high_write+0xc0/0xc0
kernel:  [<ffffffff9f13f5f8>] pagefault_out_of_memory+0x38/0xa0
kernel:  [<ffffffff9f045a27>] mm_fault_error+0x77/0x150
kernel:  [<ffffffff9f046264>] __do_page_fault+0x414/0x420
kernel:  [<ffffffff9f046292>] do_page_fault+0x22/0x30
kernel:  [<ffffffff9f5b1f98>] page_fault+0x28/0x30
```

It seems to be caused by the memory spike introduced by binary copy. Should we always enforce a minimum memory limit for runc containers in the future?

My runc binary is statically linked:
```
$ ls -alh usr/local/sbin/runc 
-rwxr-xr-x 1 lantaol primarygroup 7.8M Feb 12 13:46 usr/local/sbin/runc
```

### LINKED ISSUE #1988: "runc init" goes into a loop on RHEL 7.6 (runc 09c8266bf2fcf9519a651b04ae54c967b9ab86ec)
After the last upgrade, docker hangs while starting containers.
htop shows the runc processes consuming 100% of the CPU time.

**Environment**
RHEL 7.6 (up to date)
from https://download.docker.com/linux/centos/7/x86_64/stable:
containerd.io-1.2.2-3.3.el7.x86_64.rpm 2019-02-11 16:03:53 22.1 MiB
docker-ce-cli-18.09.2-3.el7.x86_64
docker-ce-18.09.2-3.el7.x86_64

**Steps to reproduce the issue:**

    1. Install docker-ce-18.09.2-3.el7.x86_64, docker-ce-cli-18.09.2-3.el7.x86_64, containerd.io-1.2.2-3.3.el7.x86_64.rpm from https://download.docker.com/linux/centos/7/x86_64/stable

    2. docker run alpine

**Describe the results you received:**
"docker run alpine" hangs. htop shows the "runc init" process consuming 100% cpu.

**Describe the results you expected:**
Expected docker to run the alpine container and return.

**Output of `containerd --version`:**
```
[root@moykano ~]# containerd --version
containerd github.com/containerd/containerd 1.2.2 [9754871](https://github.com/containerd/containerd/commit/9754871865f7fe2f4e74d43e2fc7ccd237edcbce)
```
Note that containerd --version outputs the same result for 1.2.2-3 and 1.2.2-3.3

Downgrading containerd.io to **containerd.io-1.2.2-3.el7.x86_64.rpm 2019-01-09 21:07:30** solves the issue.

runc --version from 1.2.2-3 (GOOD):
```
runc version 1.0.0-rc6+dev
commit: 96ec2177ae841256168fcf76954f7177af9446eb
spec: 1.0.1-dev
```

runc --version from 1.2.2-3.3 (BAD):
```
runc version 1.0.0-rc6+dev
commit: 09c8266bf2fcf9519a651b04ae54c967b9ab86ec
spec: 1.0.1-dev
```

** Delta between versions **
https://github.com/opencontainers/runc/compare/96ec2177ae841256168fcf76954f7177af9446eb...09c8266bf2fcf9519a651b04ae54c967b9ab86ec

* Kernel level: Linux moykano 3.10.0-957.5.1.el7.x86_64 #1 SMP Wed Dec 19 10:46:58 EST 2018 x86_64 x86_64 x86_64 GNU/Linux  (rpm: kernel-3.10.0-957.1.3.el7.x86_64)

* lsb_release -a
  LSB Version:	:core-4.1-amd64:core-4.1-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-4.1-amd64:desktop-4.1-noarch:languages-4.1-amd64:languages-4.1-noarch:printing-4.1-amd64:printing-4.1-noarch
  Distributor ID:	RedHatEnterpriseWorkstation
  Description:	Red Hat Enterprise Linux Workstation release 7.6 (Maipo)
  Release:	7.6
  Codename:	Maipo

cyphar:
For a variety of reasons, sendfile(2) can end up doing a short-copy so
we need to just loop until we hit the binary size. Since /proc/self/exe
is tautologically our own binary, there's no chance someone is going to
modify it underneath us (or changing the size).

In order to get around the memfd_create(2) requirement, 0a8e411
("nsenter: clone /proc/self/exe to avoid exposing host binary to
container") added an O_TMPFILE fallback. However, this fallback was
flawed in two ways:

  * It required O_TMPFILE which is relatively new (having been added to
    Linux 3.11).

  * The fallback choice was made at compile-time, not runtime. This
    results in several complications when it comes to running binaries
    on different machines to the ones they were built on.

The easiest way to resolve these things is to have fallbacks work in a
more procedural way (though it does make the code unfortunately more
complicated) and to add a new fallback that uses mkotemp(3).

The usage of memfd_create(2) and other copying techniques is quite
wasteful, despite attempts to minimise it with _LIBCONTAINER_STATEDIR.
memfd_create(2) added ~10M of memory usage to the cgroup associated with
the container, which can result in some setups getting OOM'd (or just
hogging the hosts' memory when you have lots of created-but-not-started
containers sticking around).

The easiest way of solving this is by creating a read-only bind-mount of
the binary, opening that read-only bindmount, and then umounting it to
ensure that the host won't accidentally be re-mounted read-write. This
avoids all copying and cleans up naturally like the other techniques
used. Unfortunately, like the O_TMPFILE fallback, this requires being
able to create a file inside _LIBCONTAINER_STATEDIR (since bind-mounting
over the most obvious path -- /proc/self/exe -- is a *very bad idea*).

Unfortunately detecting this isn't fool-proof -- on a system with a
read-only root filesystem (that might become read-write during "runc
init" execution), we cannot tell whether we have already done an ro
remount. As a partial mitigation, we store a _LIBCONTAINER_CLONED_BINARY
environment variable which is checked *alongside* the protection being
present.

Fixes #1979 
Fixes #1980 
Fixes #1988

### DIFF
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -160,7 +160,7 @@ static char *read_file(char *path, size_t *length)
 
 	*length = 0;
 	for (;;) {
-		int n;
+		ssize_t n;
 
 		n = read(fd, buf, sizeof(buf));
 		if (n < 0)
@@ -403,6 +403,33 @@ out:
 	return ret;
 }
 
+static ssize_t fd_to_fd(int outfd, int infd)
+{
+	ssize_t total = 0;
+	char buffer[4096];
+
+	for (;;) {
+		ssize_t nread, nwritten = 0;
+
+		nread = read(infd, buffer, sizeof(buffer));
+		if (nread < 0)
+			return -1;
+		if (!nread)
+			break;
+
+		do {
+			ssize_t n = write(outfd, buffer + nwritten, nread - nwritten);
+			if (n < 0)
+				return -1;
+			nwritten += n;
+		} while(nwritten < nread);
+
+		total += nwritten;
+	}
+
+	return total;
+}
+
 static int clone_binary(void)
 {
 	int binfd, execfd;
@@ -435,8 +462,12 @@ static int clone_binary(void)
 
 	while (sent < statbuf.st_size) {
 		int n = sendfile(execfd, binfd, NULL, statbuf.st_size - sent);
-		if (n < 0)
-			goto error_binfd;
+		if (n < 0) {
+			/* sendfile can fail so we fallback to a dumb user-space copy. */
+			n = fd_to_fd(execfd, binfd);
+			if (n < 0)
+				goto error_binfd;
+		}
 		sent += n;
 	}
 	close(binfd);
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -217,8 +217,14 @@ enum {
 
 static int make_execfd(int *fdtype)
 {
-	int fd;
-	char template[] = "/tmp/runc-cloned-binary.XXXXXX";
+	int fd = -1;
+	char template[PATH_MAX] = {0};
+	char *prefix = secure_getenv("_LIBCONTAINER_STATEDIR");
+
+	if (!prefix || *prefix != '/')
+		prefix = "/tmp";
+	if (snprintf(template, sizeof(template), "%s/runc.XXXXXX", prefix) < 0)
+		return -1;
 
 	/*
 	 * Try memfd first, it's much nicer since it's easily detected thanks to
@@ -238,7 +244,7 @@ static int make_execfd(int *fdtype)
 	 * fd re-open it and clear O_EXCL).
 	 */
 	*fdtype = EFD_FILE;
-	fd = open("/tmp", O_TMPFILE | O_EXCL | O_RDWR | O_CLOEXEC, 0700);
+	fd = open(prefix, O_TMPFILE | O_EXCL | O_RDWR | O_CLOEXEC, 0700);
 	if (fd >= 0) {
 		struct stat statbuf = {};
 		bool working_otmpfile = false;
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -27,8 +27,10 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/statfs.h>
 #include <sys/vfs.h>
 #include <sys/mman.h>
+#include <sys/mount.h>
 #include <sys/sendfile.h>
 #include <sys/syscall.h>
 
@@ -67,6 +69,7 @@ int memfd_create(const char *name, unsigned int flags)
 #  define F_SEAL_WRITE  0x0008	/* prevent writes */
 #endif
 
+#define CLONED_BINARY_ENV "_LIBCONTAINER_CLONED_BINARY"
 #define RUNC_MEMFD_COMMENT "runc_cloned:/proc/self/exe"
 #define RUNC_MEMFD_SEALS \
 	(F_SEAL_SEAL | F_SEAL_SHRINK | F_SEAL_GROW | F_SEAL_WRITE)
@@ -88,29 +91,56 @@ static void *must_realloc(void *ptr, size_t size)
 static int is_self_cloned(void)
 {
 	int fd, ret, is_cloned = 0;
+	struct stat statbuf = {};
+	struct statfs fsbuf = {};
 
 	fd = open("/proc/self/exe", O_RDONLY|O_CLOEXEC);
 	if (fd < 0)
 		return -ENOTRECOVERABLE;
 
-	/* First check memfd. */
+	/*
+	 * Is the binary a fully-sealed memfd? We don't need CLONED_BINARY_ENV for
+	 * this, because you cannot write to a sealed memfd no matter what (so
+	 * sharing it isn't a bad thing -- and an admin could bind-mount a sealed
+	 * memfd to /usr/bin/runc to allow re-use).
+	 */
 	ret = fcntl(fd, F_GET_SEALS);
 	if (ret >= 0) {
 		is_cloned = (ret == RUNC_MEMFD_SEALS);
-	} else {
-		/*
-		 * Okay, we're a tmpfile -- or we're currently running on RHEL <=7.6
-		 * which appears to have a borked backport of F_GET_SEALS. Either way,
-		 * having a file which has no hardlinks indicates that we aren't using
-		 * a host-side "runc" binary and this is something that a container
-		 * cannot fake (because unlinking requires being able to resolve the
-		 * path that you want to unlink).
-		 */
-		struct stat statbuf = {};
-		if (fstat(fd, &statbuf) >= 0)
-			is_cloned = (statbuf.st_nlink == 0);
+		goto out;
 	}
 
+	/*
+	 * All other forms require CLONED_BINARY_ENV, since they are potentially
+	 * writeable (or we can't tell if they're fully safe) and thus we must
+	 * check the environment as an extra layer of defence.
+	 */
+	if (!getenv(CLONED_BINARY_ENV)) {
+		is_cloned = false;
+		goto out;
+	}
+
+	/*
+	 * Is the binary on a read-only filesystem? We can't detect bind-mounts in
+	 * particular (in-kernel they are identical to regular mounts) but we can
+	 * at least be sure that it's read-only. In addition, to make sure that
+	 * it's *our* bind-mount we check CLONED_BINARY_ENV.
+	 */
+	if (fstatfs(fd, &fsbuf) >= 0)
+		is_cloned |= (fsbuf.f_flags & MS_RDONLY);
+
+	/*
+	 * Okay, we're a tmpfile -- or we're currently running on RHEL <=7.6
+	 * which appears to have a borked backport of F_GET_SEALS. Either way,
+	 * having a file which has no hardlinks indicates that we aren't using
+	 * a host-side "runc" binary and this is something that a container
+	 * cannot fake (because unlinking requires being able to resolve the
+	 * path that you want to unlink).
+	 */
+	if (fstat(fd, &statbuf) >= 0)
+		is_cloned |= (statbuf.st_nlink == 0);
+
+out:
 	close(fd);
 	return is_cloned;
 }
@@ -227,15 +257,16 @@ static int make_execfd(int *fdtype)
 		return -1;
 
 	/*
-	 * Try memfd first, it's much nicer since it's easily detected thanks to
-	 * sealing and also doesn't require assumptions like /tmp.
+	 * Now try memfd, it's much nicer than actually creating a file in STATEDIR
+	 * since it's easily detected thanks to sealing and also doesn't require
+	 * assumptions about STATEDIR.
 	 */
 	*fdtype = EFD_MEMFD;
 	fd = memfd_create(RUNC_MEMFD_COMMENT, MFD_CLOEXEC | MFD_ALLOW_SEALING);
 	if (fd >= 0)
 		return fd;
-	if (errno != ENOSYS)
-		goto err;
+	if (errno != ENOSYS && errno != EINVAL)
+		goto error;
 
 #ifdef O_TMPFILE
 	/*
@@ -266,7 +297,7 @@ static int make_execfd(int *fdtype)
 		errno = EISDIR;
 	}
 	if (errno != EISDIR)
-		goto err;
+		goto error;
 #endif /* defined(O_TMPFILE) */
 
 	/*
@@ -281,7 +312,7 @@ static int make_execfd(int *fdtype)
 		close(fd);
 	}
 
-err:
+error:
 	*fdtype = EFD_NONE;
 	return -1;
 }
@@ -316,15 +347,83 @@ static int seal_execfd(int *fd, int fdtype)
 	return -1;
 }
 
+static int try_bindfd(void)
+{
+	int fd, ret = -1;
+	char template[PATH_MAX] = {0};
+	char *prefix = secure_getenv("_LIBCONTAINER_STATEDIR");
+
+	if (!prefix || *prefix != '/')
+		prefix = "/tmp";
+	if (snprintf(template, sizeof(template), "%s/runc.XXXXXX", prefix) < 0)
+		return ret;
+
+	/*
+	 * We need somewhere to mount it, mounting anything over /proc/self is a
+	 * BAD idea on the host -- even if we do it temporarily.
+	 */
+	fd = mkstemp(template);
+	if (fd < 0)
+		return ret;
+	close(fd);
+
+	/*
+	 * For obvious reasons this won't work in rootless mode because we haven't
+	 * created a userns+mntns -- but getting that to work will be a bit
+	 * complicated and it's only worth doing if someone actually needs it.
+	 */
+	ret = -EPERM;
+	if (mount("/proc/self/exe", template, "", MS_BIND, "") < 0)
+		goto out;
+	if (mount("", template, "", MS_REMOUNT | MS_BIND | MS_RDONLY, "") < 0)
+		goto out_umount;
+
+
+	/* Get read-only handle that we're sure can't be made read-write. */
+	ret = open(template, O_PATH | O_CLOEXEC);
+
+out_umount:
+	/*
+	 * Make sure the MNT_DETACH works, otherwise we could get remounted
+	 * read-write and that would be quite bad (the fd would be made read-write
+	 * too, invalidating the protection).
+	 */
+	if (umount2(template, MNT_DETACH) < 0) {
+		if (ret >= 0)
+			close(ret);
+		ret = -ENOTRECOVERABLE;
+	}
+
+out:
+	/*
+	 * We don't care about unlink errors, the worst that happens is that
+	 * there's an empty file left around in STATEDIR.
+	 */
+	unlink(template);
+	return ret;
+}
+
 static int clone_binary(void)
 {
-	int binfd, memfd;
+	int binfd, execfd;
 	struct stat statbuf = {};
 	size_t sent = 0;
 	int fdtype = EFD_NONE;
 
-	memfd = make_execfd(&fdtype);
-	if (memfd < 0 || fdtype == EFD_NONE)
+	/*
+	 * Before we resort to copying, let's try creating an ro-binfd in one shot
+	 * by getting a handle for a read-only bind-mount of the execfd.
+	 */
+	execfd = try_bindfd();
+	if (execfd >= 0)
+		return execfd;
+
+	/*
+	 * Dammit, that didn't work -- time to copy the binary to a safe place we
+	 * can seal the contents.
+	 */
+	execfd = make_execfd(&fdtype);
+	if (execfd < 0 || fdtype == EFD_NONE)
 		return -ENOTRECOVERABLE;
 
 	binfd = open("/proc/self/exe", O_RDONLY | O_CLOEXEC);
@@ -335,7 +434,7 @@ static int clone_binary(void)
 		goto error_binfd;
 
 	while (sent < statbuf.st_size) {
-		int n = sendfile(memfd, binfd, NULL, statbuf.st_size - sent);
+		int n = sendfile(execfd, binfd, NULL, statbuf.st_size - sent);
 		if (n < 0)
 			goto error_binfd;
 		sent += n;
@@ -344,14 +443,15 @@ static int clone_binary(void)
 	if (sent != statbuf.st_size)
 		goto error;
 
-	if (seal_execfd(&memfd, fdtype) < 0)
+	if (seal_execfd(&execfd, fdtype) < 0)
 		goto error;
-	return memfd;
+
+	return execfd;
 
 error_binfd:
 	close(binfd);
 error:
-	close(memfd);
+	close(execfd);
 	return -EIO;
 }
 
@@ -375,6 +475,11 @@ int ensure_cloned_binary(void)
 	if (execfd < 0)
 		return -EIO;
 
+	if (putenv(CLONED_BINARY_ENV "=1"))
+		goto error;
+
 	fexecve(execfd, argv, environ);
+error:
+	close(execfd);
 	return -ENOEXEC;
 }
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -36,18 +36,21 @@
 #if !defined(SYS_memfd_create) && defined(__NR_memfd_create)
 #  define SYS_memfd_create __NR_memfd_create
 #endif
-#ifdef SYS_memfd_create
-#  define HAVE_MEMFD_CREATE
 /* memfd_create(2) flags -- copied from <linux/memfd.h>. */
-#  ifndef MFD_CLOEXEC
-#    define MFD_CLOEXEC       0x0001U
-#    define MFD_ALLOW_SEALING 0x0002U
-#  endif
+#ifndef MFD_CLOEXEC
+#  define MFD_CLOEXEC       0x0001U
+#  define MFD_ALLOW_SEALING 0x0002U
+#endif
 int memfd_create(const char *name, unsigned int flags)
 {
+#ifdef SYS_memfd_create
 	return syscall(SYS_memfd_create, name, flags);
-}
+#else
+	errno = ENOSYS;
+	return -1;
 #endif
+}
+
 
 /* This comes directly from <linux/fcntl.h>. */
 #ifndef F_LINUX_SPECIFIC_BASE
@@ -64,11 +67,9 @@ int memfd_create(const char *name, unsigned int flags)
 #  define F_SEAL_WRITE  0x0008	/* prevent writes */
 #endif
 
-#ifdef HAVE_MEMFD_CREATE
-#  define RUNC_MEMFD_COMMENT "runc_cloned:/proc/self/exe"
-#  define RUNC_MEMFD_SEALS \
+#define RUNC_MEMFD_COMMENT "runc_cloned:/proc/self/exe"
+#define RUNC_MEMFD_SEALS \
 	(F_SEAL_SEAL | F_SEAL_SHRINK | F_SEAL_GROW | F_SEAL_WRITE)
-#endif
 
 static void *must_realloc(void *ptr, size_t size)
 {
@@ -92,23 +93,29 @@ static int is_self_cloned(void)
 	if (fd < 0)
 		return -ENOTRECOVERABLE;
 
-#ifdef HAVE_MEMFD_CREATE
+	/* First check memfd. */
 	ret = fcntl(fd, F_GET_SEALS);
-	is_cloned = (ret == RUNC_MEMFD_SEALS);
-#else
-	struct stat statbuf = {0};
-	ret = fstat(fd, &statbuf);
-	if (ret >= 0)
-		is_cloned = (statbuf.st_nlink == 0);
-#endif
+	if (ret >= 0) {
+		is_cloned = (ret == RUNC_MEMFD_SEALS);
+	} else {
+		/*
+		 * Okay, we're a tmpfile -- or we're currently running on RHEL <=7.6
+		 * which appears to have a borked backport of F_GET_SEALS. Either way,
+		 * having a file which has no hardlinks indicates that we aren't using
+		 * a host-side "runc" binary and this is something that a container
+		 * cannot fake (because unlinking requires being able to resolve the
+		 * path that you want to unlink).
+		 */
+		struct stat statbuf = {};
+		if (fstat(fd, &statbuf) >= 0)
+			is_cloned = (statbuf.st_nlink == 0);
+	}
+
 	close(fd);
 	return is_cloned;
 }
 
-/*
- * Basic wrapper around mmap(2) that gives you the file length so you can
- * safely treat it as an ordinary buffer. Only gives you read access.
- */
+/* Read a given file into a new buffer, and providing the length. */
 static char *read_file(char *path, size_t *length)
 {
 	int fd;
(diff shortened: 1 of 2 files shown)


cyphar:
This new setup would be an opportunity to fix #1980, by making the `O_TMPFILE` be the statedir of runc. But we'd have to allow you to disable `memfd_create(2)` usage at runtime which is just going to be ugly.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +190,149 @@ static int fetchve(char ***argv)
...
+	case EFD_FILE: {
+		/* Need to re-open "memfd" as an O_PATH to avoid execve(2) giving -EXTBSY. */
+		int newfd;
+		char *fdpath = NULL;
+
+		if (fchmod(*fd, 0100) < 0)
+			return -1;
+
+		while (asprintf(&fdpath, "/proc/self/fd/%d", *fd) < 0)
+			/* noop */;
```
brauner: Hm, but you know the fixed size for this. We have macros that automatically calculate this based on the integer type but you could do:
```c

/* /proc/       =    6
 *                +
 * <pid-as-str> =   INTTYPE_TO_STRLEN(pid_t) = 21
 *                +
 * /fd/         =    4
 *                +
 * <fd-as-str>  =   INTTYPE_TO_STRLEN(int) = 21
 *                +
 * \0           =    1
 */
#define RUNC_PROC_PID_FD_LEN
```
intstead of using `asprintf()`.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +190,149 @@ static int fetchve(char ***argv)
...
+	 * Our final option is to create a temporary file the old-school way, and
+	 * then unlink it so that nothing else sees it by accident.
+	 */
+	*fdtype = EFD_FILE;
+	if (*fdtype & allowed) {
+		fd = mkostemp(template, O_CLOEXEC);
+		if (fd >= 0) {
+			if (!unlink(template))
+				return fd;
+			close(fd);
```
brauner: You may want to consider:
```c
int _e_ = errno;
close(fd);
errno = _e_;
```

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +190,149 @@ static int fetchve(char ***argv)
...
-#ifdef HAVE_MEMFD_CREATE
-	int err = fcntl(memfd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
-	if (err < 0)
+	while (sent < statbuf.st_size) {
+		int n = sendfile(memfd, binfd, NULL, statbuf.st_size - sent);
+		if (n < 0)
+			goto error_binfd;
+		sent += n;
+	}
+	close(binfd);
```
brauner: `sendfile()` doesn't necessarily work correctly on all filesystems. If you're not on a tmpfs aka using `memfd_create()` you should probably have a `fd_to_fd()` or something. But *shrug* :)

brauner: (Referring back to our chat before. :))

cyphar: I will test it with some other filesystems, but as I discussed in our chat unless your filesystem defines a `f_op->splice_write` then it should work just as well as it would anywhere (and it's all handled in the VFS with `current->splice_pipe`.

brauner: I'm already tracking this thing down.

libcontainer/nsenter/cloned_binary.c
cyphar: I could also just use `PATH_MAX`, I guess.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -191,17 +190,113 @@ static int fetchve(char ***argv)
...
+}
+
+static int seal_execfd(int *fd, int fdtype)
+{
+	switch (fdtype) {
+	case EFD_MEMFD:
+		return fcntl(*fd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
+	case EFD_TMPFILE: /* fall through */
+	case EFD_FILE: {
+		/* Need to re-open "memfd" as an O_PATH to avoid execve(2) giving -EXTBSY. */
```
kolyshkin: neat: it's `-ETXTBSY`

libcontainer/nsenter/cloned_binary.c
```diff
@@ -205,52 +206,58 @@ static int make_execfd(int *fdtype)
 	 * sealing and also doesn't require assumptions like /tmp.
 	 */
 	*fdtype = EFD_MEMFD;
-	fd = memfd_create(RUNC_MEMFD_COMMENT, MFD_CLOEXEC | MFD_ALLOW_SEALING);
-	if (fd >= 0)
-		return fd;
-	if (errno != ENOSYS)
-		goto err;
+	if (*fdtype & allowed) {
```
kolyshkin: Perhaps it could be better write this as

```C
if (allowed & EFD_MEMFD)
```

not for the sake of optimization -- I hope compiler optimizes the dereference out -- but just for readability.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -205,52 +206,58 @@ static int make_execfd(int *fdtype)
...
-		if (fstat(fd, &statbuf) >= 0)
-			working_otmpfile = (statbuf.st_nlink == 0);
-
-		if (working_otmpfile)
-			return fd;
-
-		/* Pretend that we got EISDIR since O_TMPFILE failed. */
-		close(fd);
-		errno = EISDIR;
+	if (*fdtype & allowed) {
```
kolyshkin: ditto

libcontainer/nsenter/cloned_binary.c
```diff
@@ -205,52 +206,58 @@ static int make_execfd(int *fdtype)
...
 	 * Our final option is to create a temporary file the old-school way, and
 	 * then unlink it so that nothing else sees it by accident.
 	 */
 	*fdtype = EFD_FILE;
-	fd = mkostemp(template, O_CLOEXEC);
-	if (fd >= 0) {
-		if (!unlink(template))
-			return fd;
-		close(fd);
+	if (*fdtype & allowed) {
```
kolyshkin: ditto

kolyshkin:
Thanks for writing this, you beat me to it :) Left some minor comments in the code. The only issue remaining is `memfd_create()` is not used in a scenario when compiled with older includes (i.e. no __NR_memfd_create) and later run under a kernel that supports it. The fix would be to add something like
```C
/* Use our own wrapper for memfd_create. */
#if !defined(__NR_memfd_create)
#  ifdef __i386__
#    define __NR_memfd_create 356
#  elif __x86_64__
#    define __NR_memfd_create 319
#  elif __powerpc64__
#    define __NR_memfd_create 360
#  elif __aarch64__
#    define __NR_memfd_create 279
#  elif __arm__
#    define __NR_memfd_create 385
/* FIXME more arches here */
#  endif
#endif /* !__NR_memfd_create */
```
but in the light of #1988 and #1980 maybe it makes sense to just not use memfd_create() at all, only leaving other methods? I'm not proposinig it, just an idea to discuss...

cyphar:
@kolyshkin We don't do that for any other syscall and I really don't like the idea of hard-coding syscall numbers. Yeah, it's part of the Linux ABI and thus won't break, but I'm really not a fan of it.

libcontainer/nsenter/cloned_binary.c
brauner: So sendfile() works but I've added a fallback `fd_to_fd()`  in our code.

brauner:
@cyphar, did you think about dropping `O_TMPFILE`? We now have merged the old kernel pieces and now have

https://github.com/lxc/lxc/blob/2d8bd1db234462356c73b12f238a615b2d1cb550/src/lxc/rexec.c#L106

libcontainer/nsenter/cloned_binary.c
cyphar: I'm not sure we need it, we don't use errno here.

cyphar:
@brauner I don't have a strong opinion about whether I should keep `O_TMPFILE`. On the one hand having an extra fallback just means that the `mkostemp` path won't be tested as well, but on the other hand `O_TMPFILE` is in principle the "better" way of doing it.

cyphar:
Any strong opinions on this one (in particular, should we keep `O_TMPFILE` or drop it).

thaJeztah:
ping @justincormack as well; PTAL

mrunalp:
@cyphar I am in favor of retaining `O_TMPFILE`.

mrunalp:
I like the approach you have taken in the PR. I am on the fence on whether we should remove memfd_create right away as you provide a way to disable it.

mrunalp:
I am approving this but will be happy to re-approve even if majority thinks that we should drop memfd_create. Thanks!

mrunalp:
LGTM

kolyshkin:
TL;DR: it looks like `memfd_create()` is not properly ported to RHEL7 kernels (including the latest and greatest RHEL 7.6). **Tested on CentOS, not real RHEL**

I tested `memfd_create` and friends, using memfd_test.c from the kernel sources.

On CentOS 7.2 kernel (3.10.0-327.el7.x86_64) and it fails right there, giving EINVAL on MFD_ALLOW_SEALING flag.

I tested it on latest CentOS 7.6 kernel (3.10.0-957.5.1.el7.x86_64) and it fails mid-way on `SHARE_OPEN` test in

https://github.com/torvalds/linux/blob/f6163d67cc31b8f2a946c4df82be3c6dd918412d/tools/testing/selftests/memfd/memfd_test.c#L881

with:
```
memfd: SHARE-OPEN 
open(/proc/self/fd/3) failed: Text file busy
Aborted
```

Here's a relevant portion of strace:
```
write(1, "memfd: SHARE-OPEN \n", 19memfd: SHARE-OPEN 
)    = 19
memfd_create("kern_memfd_share_open", MFD_CLOEXEC|MFD_ALLOW_SEALING) = 3
ftruncate(3, 8192)                      = 0
fcntl(3, F_GET_SEALS)                   = 0
openat(AT_FDCWD, "/proc/self/fd/3", O_RDWR) = 4
fcntl(3, F_GET_SEALS)                   = 0
fcntl(3, F_ADD_SEALS, F_SEAL_WRITE)     = 0
fcntl(3, F_GET_SEALS)                   = 0x8 (seals F_SEAL_WRITE)
fcntl(4, F_GET_SEALS)                   = 0x8 (seals F_SEAL_WRITE)
fcntl(4, F_GET_SEALS)                   = 0x8 (seals F_SEAL_WRITE)
fcntl(4, F_ADD_SEALS, F_SEAL_SHRINK)    = 0
fcntl(3, F_GET_SEALS)                   = 0xa (seals F_SEAL_SHRINK|F_SEAL_WRITE)
fcntl(4, F_GET_SEALS)                   = 0xa (seals F_SEAL_SHRINK|F_SEAL_WRITE)
close(3)                                = 0
openat(AT_FDCWD, "/proc/self/fd/4", O_RDONLY) = 3
fcntl(3, F_GET_SEALS)                   = 0xa (seals F_SEAL_SHRINK|F_SEAL_WRITE)
fcntl(3, F_ADD_SEALS, F_SEAL_SEAL)      = -1 EPERM (Operation not permitted)
fcntl(3, F_GET_SEALS)                   = 0xa (seals F_SEAL_SHRINK|F_SEAL_WRITE)
fcntl(4, F_GET_SEALS)                   = 0xa (seals F_SEAL_SHRINK|F_SEAL_WRITE)
close(4)                                = 0
openat(AT_FDCWD, "/proc/self/fd/3", O_RDWR) = -1 ETXTBSY (Text file busy)
write(1, "open(/proc/self/fd/3) failed: Te"..., 45open(/proc/self/fd/3) failed: Text file busy
) = 45
```

Note that the above test case was part of initial implementation of memfd_create(). I have retried test with selinux disabled -- same failure.

I didn't look whether this is the way that this runc patch uses memfd_create (probably not).

All this gives me an impression that memfd_create() and friends are in bad shape even in latest RHEL7 kernels.

kolyshkin:
Now, for my great surprise, I ran the same test binary on Debian Jessie (which is 3.16 but have memfd stuff backported), and it passes with flying colors:
```
root@kir-deb-jessie:~# cat /etc/debian_version 
8.10
root@kir-deb-jessie:~# uname -a
Linux kir-deb-jessie 3.16.0-5-amd64 #1 SMP Debian 3.16.51-3+deb8u1 (2018-01-08) x86_64 GNU/Linux
root@kir-deb-jessie:~# ./memfd_test 
memfd: CREATE
memfd: BASIC
memfd: SEAL-WRITE
memfd: SEAL-SHRINK
memfd: SEAL-GROW
memfd: SEAL-RESIZE
memfd: SHARE-DUP 
memfd: SHARE-MMAP 
memfd: SHARE-OPEN 
memfd: SHARE-FORK 
memfd: SHARE-DUP (shared file-table)
memfd: SHARE-MMAP (shared file-table)
memfd: SHARE-OPEN (shared file-table)
memfd: SHARE-FORK (shared file-table)
memfd: DONE
```

kolyshkin:
Filed a bug to Red Hat: https://bugzilla.redhat.com/show_bug.cgi?id=1679829

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +197,155 @@ static int fetchve(char ***argv)
...
+	*fdtype = EFD_NONE;
+	return -1;
+}
+
+static int seal_execfd(int *fd, int fdtype)
+{
+	switch (fdtype) {
+	case EFD_MEMFD:
+		return fcntl(*fd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
+	case EFD_TMPFILE: /* fall through */
```
lifubang: is still working on this switch case 'EFD_TMPFILE'?

lifubang:
When use _LIBCONTAINER_DISABLE_MEMFD_CLONE=1 runc create test1
It will get an error:
```
nsenter: could not ensure we are a cloned binary: Permission denied
container_linux.go:344: starting container process caused "process_linux.go:293: copying bootstrap data to pipe caused \"write init-p: broken pipe\""
```
The error is raised in here:
https://github.com/opencontainers/runc/blob/89c07553177c66cfe07de11fcdc17b59621ead07/libcontainer/nsenter/cloned_binary.c#L374-L375 

I use root permission to run runc. May be need some special permission?

cyphar:
> nsenter: could not ensure we are a cloned binary: Permission denied

I'm really not sure, can you please provide some more information about your setup -- an `strace` log would also be super useful. I'm currently working on moving the `cgroup` setup code so that it's not necessary to disable `memfd_create` (and so I will remove the ability to disable it).

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +197,155 @@ static int fetchve(char ***argv)
...
+	/*
+	 * Try memfd first, it's much nicer since it's easily detected thanks to
+	 * sealing and also doesn't require assumptions like /tmp.
+	 */
+	if (allowed & EFD_MEMFD) {
+		*fdtype = EFD_MEMFD;
+		fd = memfd_create(RUNC_MEMFD_COMMENT, MFD_CLOEXEC | MFD_ALLOW_SEALING);
+		if (fd >= 0)
+			return fd;
+		if (errno != ENOSYS)
```
ijumps: What if os has `memfd_create` syscall and without seal support (I know that can happen on centos 7.2 with `3.10.0-327.36.3` kernel). It will return `EINVAL` without retry other ways.

By the way, is it reasonable to just allow use `memfd_create` without such sealing flags?

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +197,155 @@ static int fetchve(char ***argv)
 	return -EINVAL;
 }
 
+#define EFD_NONE     0x00
+#define EFD_MEMFD    0x01
+#define EFD_TMPFILE  0x02
+#define EFD_FILE     0x04
+#define EFD_ALL      (EFD_MEMFD | EFD_TMPFILE | EFD_FILE)
+
+static int make_execfd(int *fdtype, int allowed)
```
ijumps: Wondering that if we don't use in memory way like `EFD_MEMFD`, if we keep the copy runc on host file system, will it cause host in `out of disk` situation by writing back a very big file use the same way as CVE-2019-5736 do?

In memory method like `EFD_MEMFD` at least can be controlled by cgroup memory limit.

libcontainer/nsenter/cloned_binary.c
cyphar: I can make this check `errno != EINVAL` too.

> By the way, is it reasonable to just allow use memfd_create without such sealing flags?

No, because while we don't need them to protect against overwriting, we need working sealing support so that we can detect whether we are a cloned binary.

libcontainer/nsenter/cloned_binary.c
cyphar: The file is unlinked immediately, and will become free space as soon as the container has started. We're talking about a window of hundreds of milliseconds -- not to mention that this is a fallback for systems that don't have `memfd_create`.

I am pushing an updated patch that tries to use a read-only bind-mount before any of these copying tricks, to avoid most of these problems -- but running out of disk space shouldn't be a problem.

cyphar:
I would like to point out that one benefit of memfd sealing is that an administrator could actually bind-mount a sealed memfd over `/usr/bin/runc` -- and `runc` won't do another copy because it detects that it is already a sealed memfd.

cyphar:
Ready for review and merge.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -87,29 +89,62 @@ static void *must_realloc(void *ptr, size_t size)
  */
 static int is_self_cloned(void)
 {
-	int fd, ret, is_cloned = 0;
+	int fd, ret, is_cloned = false;
```
justincormack: you shouldn't init an `int` to `false`, use a `bool`.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +227,211 @@ static int fetchve(char ***argv)
...
+
+	if (!prefix || *prefix != '/')
+		prefix = "/tmp";
+	if (snprintf(template, sizeof(template), "%s/runc.XXXXXX", prefix) < 0)
+		return -1;
+
+	/*
+	 * Now try memfd, it's much nicer than actually creating a file in STATEDIR
+	 * since it's easily detected thanks to sealing and also doesn't require
+	 * assumptions about .
```
justincormack: a word missing here

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +227,211 @@ static int fetchve(char ***argv)
...
+			return fd;
+		close(fd);
+	}
+
+error:
+	*fdtype = EFD_NONE;
+	return -1;
+}
+
+static int seal_execfd(int *fd, int fdtype)
```
justincormack: It would be more idiomatic to return the new fd rather than modify in place, can still use -1 for errors.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +227,211 @@ static int fetchve(char ***argv)
...
+{
+	switch (fdtype) {
+	case EFD_MEMFD:
+		return fcntl(*fd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
+	case EFD_FILE: {
+		/* Need to re-open "memfd" as an O_PATH to avoid execve(2) giving -ETXTBSY. */
```
justincormack: It is confusing to call it "memfd" when it is not.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +227,211 @@ static int fetchve(char ***argv)
...
+	 * too, invalidating the protection).
+	 */
+	if (umount2(template, MNT_DETACH) < 0) {
+		if (ret >= 0)
+			close(ret);
+		ret = -ENOTRECOVERABLE;
+	}
+
+out:
+	unlink(template);
```
justincormack: It is unlikely to fail but you should check the return value here

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,49 +227,211 @@ static int fetchve(char ***argv)
...
-	if (memfd < 0)
+	struct stat statbuf = {};
+	size_t sent = 0;
+	int fdtype = EFD_NONE;
+
+	/*
+	 * Before we resort to copying, let's try creating an ro-binfd in one shot
+	 * by getting a handle for a read-only bind-mount of the execfd.
+	 */
+	memfd = try_bindfd();
```
justincormack: it is confusing to call this a memfd

libcontainer/nsenter/cloned_binary.c
cyphar: I was going to do that, but `fd = seal_execfd(fd, fdtype)` feels less idomatic to me.

libcontainer/nsenter/cloned_binary.c
cyphar: Ignoring it was intentional, if there's a temporary empty file in our runc state directory it's unlikely to matter -- and we don't have logging so we'd have to fail if `unlink` failed.

libcontainer/nsenter/cloned_binary.c
cyphar: I can change it to `retfd` or `rofd` if you prefer.

libcontainer/nsenter/cloned_binary.c
cyphar: It's in quotes for a reason (it's not *actually* a memfd but we're treating it like one), but sure I can reword it.

cyphar:
There, aside from the two comments I disagree with (the `unlink` thing, which I've commented -- and the `seal_execfd` interface which I don't think is more idiomatic and would just make the calling code more complicated with little benfit).

libcontainer/nsenter/cloned_binary.c
```diff
@@ -88,28 +90,61 @@ static void *must_realloc(void *ptr, size_t size)
...
+		goto out;
+	}
+
+	/*
+	 * Is the binary on a read-only filesystem? We can't detect bind-mounts in
+	 * particular (in-kernel they are identical to regular mounts) but we can
+	 * at least be sure that it's read-only. In addition, to make sure that
+	 * it's *our* bind-mount we check CLONED_BINARY_ENV.
+	 */
+	if (fstatfs(fd, &fsbuf) >= 0);
```
mrunalp: Is there an unintentional semi-colon at the end of if here?

cyphar: Yes, I'm not sure how I missed it (and I have GCC 8 which should be giving me warnings about misleading indentation...

libcontainer/nsenter/cloned_binary.c
cyphar: For the record, the issue was that AppArmor blocks writing to deleted files. If someone complains about this, I can fix it later but I don't feel the need to include `fd_to_fd()` at the moment.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,51 +227,217 @@ static int fetchve(char ***argv)
...
+	return -1;
+}
+
+static int try_bindfd(void)
+{
+	int fd, ret = -1;
+	char template[PATH_MAX] = {0};
+	char *prefix = secure_getenv("_LIBCONTAINER_STATEDIR");
+
+	if (!prefix || *prefix != '/')
```
Ace-Tang: I still cannot figure out, doesn't it want `*prefix == '/'` ? and same place in `make_execfd`

cyphar: If `_LIBCONTAINER_STATEDIR` is not present or doesn't start with a `/` then we use `/tmp`. That's what this check is doing.

cyphar:
@opencontainers/runc-maintainers This is ready (for real this time). :smile_cat:

libcontainer/nsenter/cloned_binary.c
cyphar: I've added `fd_to_fd()`.

Ace-Tang:
seems still something wrong,  build can be always successfully using master branch. But with this patch,  I only can build successfully on my ubuntu machine(4.15.0-45-generic), failed both on centos machine with kernel 4.9.93-010.ali3000.alios7.x86_64 and 3.10.0-327.ali2012.alios7.x86_64, donot know why. 
```
#make
go build -buildmode=pie  -ldflags "-X main.gitCommit="13b54de3c51f69fbfb137cf3be49b4ea433569e6" -X main.version=1.0.0-rc6+dev " -tags "seccomp" -o runc .
cloned_binary.c: In function 'make_execfd':
cloned_binary.c:228:20: error: 'O_TMPFILE' undeclared (first use in this function)
  fd = open("/tmp", O_TMPFILE | O_EXCL | O_RDWR | O_CLOEXEC, 0700);
                    ^
cloned_binary.c:228:20: note: each undeclared identifier is reported only once for each function it appears in
make: *** [runc] Error 2
```
and introduced by the second commit `nsenter: cloned_binary: expand and add pre-3.11 fallbacks`

cyphar:
Ah, we need to an `#ifdef O_TMPFILE`...

Ace-Tang:
Sloved !
```
I don't understand why it would fail on `4.9.93` -- `O_TMPFILE` was added in 3.11 but that's probably due to an old `glibc`
```
so which gcc version it should be larger than ? :confused:

cyphar:
It's not gcc, it's `glibc`. I'm really not sure, you would need to go into the git history. Not to mention that CentOS will have backports and so even then the version number would be useless.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -88,28 +91,61 @@ static void *must_realloc(void *ptr, size_t size)
 static int is_self_cloned(void)
```
lifubang: How about just use `CLONED_BINARY_ENV` to avoid potential loop that we have not found yet?
I don't know whether this is more safety or not!

For example:
```
static int is_self_cloned(void) {
    return getenv(CLONED_BINARY_ENV) && strcmp(getenv(CLONED_BINARY_ENV), "1")==0;
}
```

lifubang: I think there is just one disadvantage, even if we have protected run binary file, we will call ensure_clone_binary first.

cyphar: Because I don't want to be in a situation where someone passes `CLONED_BINARY_ENV` to `runc` and that results in an unsafe setup.

In fact, the *only* reason why `CLONED_BINARY_ENV` exists in this PR is to stop the case where `/usr/bin/runc` is on a read-only mount (or is an unlinked file) and we think that makes the binary safe -- but in those cases it is *not* safe (and we need to apply the protection).

I don't want to provide anyone the ability to opt out of this protection -- if you want to, you can bind-mount a sealed `memfd` over `/usr/bin/runc` if you want to "opt out" but apart from that there is no reason to opt-out now that the primary protection is a read-only bind-mount which takes up basically no memory.

lifubang: For this topic, I have no other opinions.

Ace-Tang:
@crosbymichael @cyphar , any update on this patch, can this be merged ?

crosbymichael:
LGTM

mrunalp:
LGTM

cyphar:
Ah dammit, this wasn't meant to be merged -- the bind-mount approach actually isn't fool-proof (if you can do a mount, you can get around it). However, the default profiles actually disallow mounting so this isn't the end of the world. I'll submit a follow-up PR that uses overlayfs rather than bind-mounts.

Ace-Tang:
Bad news... we plan to appy this patch to our production environment, looks like still need to wait

AkihiroSuda:
>  I'll submit a follow-up PR that uses overlayfs rather than bind-mounts.

It won't work in userns?

cyphar:
> It won't work in userns?

No, but that's why we have fallbacks. I've submitted #2006.

**However** I've come to the conclusion that running a privileged container with CAP_SYS_ADMIN is simply an insecure setup, so maybe it's better if we just leave it in this state -- where arguably it's not secure with CAP_SYS_ADMIN (but there's almost certainly other issues with running CAP_SYS_ADMIN).

libcontainer/nsenter/cloned_binary.c
```diff
@@ -192,51 +228,261 @@ static int fetchve(char ***argv)
...
+error:
+	*fdtype = EFD_NONE;
+	return -1;
+}
+
+static int seal_execfd(int *fd, int fdtype)
+{
+	switch (fdtype) {
+	case EFD_MEMFD:
+		return fcntl(*fd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
```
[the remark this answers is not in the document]
cyphar: If you want to submit a patch to do that, sure. But I spent several weeks working on it, and given painfully trivial the fix for the CVE is ("make a copy and execute that"), working on it for much more time feels like a waste of time (in fact, most of the time I spent working on it feels like wasted time).

If `F_ADD_SEALS` fails (with `memfd_create` and `MFD_ALLOW_SEALING` working) then the kernel is broken -- and while that can happen with some distributions I'd prefer to actually see it be an existing problem before trying to fix it.

PR #1986, opened 2019-02-17
switched travis to xenial

adrianreber:
The CRIU test for lazy migration was always skipped in Travis because
the kernel was too old. This switches Travis testing to dist: xenial
which provides a newer kernel which enables CRIU lazy migration testing.

Not sure if this is something runc cannot or does not want to do (yet).

Tests are all successful, so switching to a newer distribution in Travis seems doable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: go
 go:
   - 1.9.x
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #1987, opened 2019-02-18
fix CVE-2019-5736 by using as less memory as possible

lifubang:
I'm a code fan of opencontainers/runc project.
And my business is `Online Judge (OJ)`, which is based on docker.
To ensure fairness of the OJ examination, we need to use memory limit with 4m, 5m, or 10m.
So, I need to fix `CVE-2019-5736` by using as less memory as possible.
I open this PR to figure out my solution by using code, because my English is very poor.

My solution is to copy a new runc binary file with a random name before we start to run `runc init` command. And delete it after `runc init` started.

If there is a problem, please point out and discuss it.
Thanks.


libcontainer/container_linux.go
```diff
@@ -460,7 +461,42 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 	return c.newInitProcess(p, cmd, parentPipe, childPipe)
 }
 
+func runcInTmp() (string, error) {
+	runc, err := os.Readlink("/proc/self/exe")
+	if err != nil {
+		return "", err
+	}
+	id := uuid.New()
+	temp := fmt.Sprintf("/tmp/runc.%s", id)
```
cyphar: If we're going to do it this way, I would want to use `memfd_create` because using `/tmp` won't always work and `memfd_create` is a fool-proof way of doing it. But ultimately I could just rewrite #1984 entirely to not use C code.

cyphar:
#1984 now makes this no longer relevant -- since by default we try doing a read-only bind-mount which takes up no memory or storage (which is better than this approach).

PR #1992, opened 2019-02-20
Add option to checkpoint to specify SELinux label

adrianreber:
To checkpoint a process CRIU injects a parasite code into the process to
be checkpointed. This parasite code tries to communicate over a socket
with the main CRIU process. To SELinux this looks like a process inside
of a container tries to open a socket to another process.

This adds an option to runc to call setsockcreatecon() before running
CRIU to label all sockets with a certain label. Having the possibility
to specify a certain label makes it easier to write corresponding
policies.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -41,6 +41,7 @@ checkpointed.`,
 		cli.StringFlag{Name: "manage-cgroups-mode", Value: "", Usage: "cgroups mode: 'soft' (default), 'full' and 'strict'"},
 		cli.StringSliceFlag{Name: "empty-ns", Usage: "create a namespace, but don't restore its properties"},
 		cli.BoolFlag{Name: "auto-dedup", Usage: "enable auto deduplication of memory images"},
+		cli.StringFlag{Name: "selinux-socket-label", Usage: "SELinux label for CRIU sockets"},
 	},
 	Action: func(context *cli.Context) error {
 		if err := checkArgs(context, 1, exactArgs); err != nil {
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -26,6 +26,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/selinux/go-selinux/label"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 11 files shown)


adrianreber:
@rhatdan as discussed in https://github.com/containers/libpod/issues/2334

rhatdan:
I don't think we need an option here.  runc should just be setting the socket label to match the process label if the process label is set.  This indicates that the socket is to be used by the container process.
I don't like allowing the caller to set the label, since this is not the way SELinux works.  Potentially I could see passing a flag that says "don't set the label", but I really see that as a corner case.

adrianreber:
Unfortunately this does not work at all. I was using an instrumented CRIU which did the setsockcreatecon() just before creating the socket without noticing it. I think I have to do this directly in CRIU because if I do setsockcreatecon() in runc before calling CRIU and if I then do a getsockcreatecon() in CRIU I get a null. It seems setsockcreatecon() does not survive exec(), right?

rst0git:
> It seems setsockcreatecon() does not survive exec(), right?

From the man page of setsockcreatecon(3):

- The sockcreate context is automatically reset after the next execve(2), so a program doesn't need to  explicitly sanitize it upon startup.

> I think I have to do this directly in CRIU because if I do setsockcreatecon() in runc before calling CRIU and if I then do a getsockcreatecon() in CRIU I get a null.

Would it be possible to pass the value of `opts.SELinuxSocketLabel` to CRIU via `opts.lsm_profile`?

adrianreber:
> > It seems setsockcreatecon() does not survive exec(), right?
> 
> From the man page of setsockcreatecon(3):
> 
>     * The sockcreate context is automatically reset after the next execve(2), so a program doesn't need to  explicitly sanitize it upon startup.

Thanks for that pointer. So it cannot work with setsockcreatecon() from the parent application.

> > I think I have to do this directly in CRIU because if I do setsockcreatecon() in runc before calling CRIU and if I then do a getsockcreatecon() in CRIU I get a null.
> 
> Would it be possible to pass the value of `opts.SELinuxSocketLabel` to CRIU via `opts.lsm_profile`?

So far I tried setexeccon() before calling CRIU, but that failed and I kind of expected that, because CRIU is then running in the wrong context. The problem I also detected is that the restore process is running in the wrong context anyway, so this needs some work on CRIU. As far as I understand it `opts.lsm_profile` is only relevant during restore, but the current problem needs a certain label on the sockets during checkpoint. We could use `opts.lsm_profile` during checkpointing to pass the desired socket label, but I think a new option would be cleaner.

> Note that this option is also used in by [LXC](https://github.com/lxc/lxc/blob/a7547c5c34f5a5f4d754acee7e224602dfbda966/src/lxc/criu.c#L523-L526).

I have also looked at this code already :grin: .

adrianreber:
Closing this as it does not work and I have a change in CRIU to do it correctly: https://github.com/checkpoint-restore/criu/pull/648

PR #1995, opened 2019-02-25
exec: expose --preserve-fds

giuseppe:
The implementation is already there, we only need to add the CLI
option and pass it down.

### DIFF
--- a/tests/integration/exec.bats
+++ b/tests/integration/exec.bats
@@ -127,3 +127,14 @@ function teardown() {
 
   [[ ${output} == "uid=1000 gid=1000 groups=99(nogroup),100(users)" ]]
 }
+
+@test "runc exec --preserve-fds" {
+  # run busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  run bash -c "cat hello > preserve-fds.test; exec 3<preserve-fds.test; $RUNC --log /proc/self/fd/2 --root $ROOT exec --preserve-fds=1 test_busybox cat /proc/self/fd/3"
+  [ "$status" -eq 0 ]
+
+  [[ "${output}" == *"hello"* ]]
+}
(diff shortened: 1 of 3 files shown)


cyphar:
Seems reasonable, though I would like an integration test.

cyphar:
LGTM, assuming the tests pass.

giuseppe:
tests are passing now

cyphar:
LGTM.

mrunalp:
LGTM

mrunalp:
@caniszczyk pull-approve is taking a long time to detect LGTMs.

cyphar:
Tests are still failing.

tests/integration/exec.bats
```diff
@@ -127,3 +127,14 @@ function teardown() {
 
   [[ ${output} == "uid=1000 gid=1000 groups=99(nogroup),100(users)" ]]
 }
+
+@test "runc exec --preserve-fds" {
+  # run busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  run bash -c "cat hello > preserve-fds.test; exec 3<preserve-fds.test; $RUNC --log /proc/self/fd/2 --root $ROOT exec --preserve-fds=1 test_busybox cat /proc/self/fd/3"
```
kolyshkin, line 136: What I don't understand is why `cat hello` is not failing. There is no such file named `hello`.

giuseppe: I guess the error is just ignored but the file is still created by the shell.  I guess it had to be `echo` but too long ago to remember :)

kolyshkin: I mean, yes, `cat hello` should fail and this is ignored by bash since it does not run with `set -e`.

What I don't understand is how the check below (that checks that `$output` contains `hello` works at all.

kolyshkin: @giuseppe pointed out that `cat hello` produces the message like `cat: hello: no such file or directory`, and thus the check for *hello* in the $output suceeds!

PR #1996, opened 2019-02-26
updating link to code of conduct and security docs to org

vsoch:
Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

https://github.com/opencontainers/org/issues/3

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,7 @@ You can find official releases of `runc` on the [release](https://github.com/ope
 
 ### Security
 
-If you wish to report a security issue, please disclose the issue responsibly
-to security@opencontainers.org.
+If you wish to report a security issue, please see the protocol under [security][security].
 
 ## Building
 
@@ -268,3 +267,6 @@ WantedBy=multi-user.target
 ## License
 
 The code and docs are released under the [Apache 2.0 license](LICENSE).
+
+
+[security]: https://github.com/opencontainers/org/blob/master/security
--- a/vendor/github.com/opencontainers/runtime-spec/README.md
+++ b/vendor/github.com/opencontainers/runtime-spec/README.md
@@ -139,7 +139,7 @@ Read more on [How to Write a Git Commit Message][how-to-git-commit] or the Discu
 
 
 [charter]: https://www.opencontainers.org/about/governance
-[code-of-conduct]: https://github.com/opencontainers/tob/blob/master/code-of-conduct.md
+[code-of-conduct]: https://github.com/opencontainers/org/blob/master/.github/CODE_OF_CONDUCT.md
 [dev-list]: https://groups.google.com/a/opencontainers.org/forum/#!forum/dev
 [how-to-git-commit]: http://chris.beams.io/posts/git-commit
 [irc-logs]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/


README.md
```diff
@@ -18,8 +18,7 @@ You can find official releases of `runc` on the [release](https://github.com/ope
 
 ### Security
 
-If you wish to report a security issue, please disclose the issue responsibly
-to security@opencontainers.org.
+If you wish to report a security issue, please see the protocol under [security](security).
```
cyphar: Should be `[security]`.

vsoch: okay I think I got it right this time.

cyphar:
Please squash your commits.

README.md
cyphar: That works, as would `[security][security]`. :P

vsoch:
Can't we squash on merge?

cyphar:
We can, but I don't like doing it because it screws up the authors GPG signatures. But if you'd prefer that, I can do it (I do merge commits locally because I want to sign those too).

vsoch:
I definitely can - I’ve never squashed locally before. How do I do it?

cyphar:
The most generic way to do it is with an interactive git rebase (`git rebase -i`). Effectively you would do:

```
% git rebase -i origin/master
```

which will open your text editor, with a bunch of lines like this:

```
pick c6297bdbfacf updating link to code of conduct and security docs to org
pick defb213bc44c fixing security link
pick 89c3a9d3f036 fixing code of conduct link
```

Each line is one commit, in order -- and the idea is that the interactive rebase will apply each line in-order (`pick` just means to use that commit -- it comes from `git cherry-pick`).

This allows you to mess with the commits individually (you could re-order them or edit them), but since we want to squash them we can use `squash` -- but we also don't care about the commit messages for fix-up patches so we can just use `fixup` to avoid writing a new commit message. So you'd change the above lines to (you can use `f` instead of `fixup` if you want):

```
pick c6297bdbfacf updating link to code of conduct and security docs to org
fixup defb213bc44c fixing security link
fixup 89c3a9d3f036 fixing code of conduct link
```

Which will translate to:

1. Use the commit c6297bdbfacf.
2. Squash defb213bc44c into the previous commit (c6297bdbfacf) to create a new commit `$foo`.
3. Squash 89c3a9d3f036 into the previous commit (`$foo`) to create a new commit `$bar`.

And finally your current branch will now be `$bar`. If it's still a bit confusing, you could look up some tutorials online about interactive rebasing.

vsoch:
Cool I've never done that - this looks fun! Be back after futzing around a bit :)

vsoch:
okay I'm not sure I catch the $foo $bar mapping to fix/CoC-link and origin/master, but I did something and I'm going to try pushing it. Worst case scenario I'll just start over.

vsoch:
Nope, definitely failed. Can you explain the $foo $bar mapping?

vsoch:
Let's use this one instead https://github.com/opencontainers/runc/pull/1997

cyphar:
You don't need to worry about `$foo` or `$bar` -- you just need to change the lines like I said. The merge commit you have here is probably because you did a `git pull`. Fixing this is much less fun (I personally have `git pull --rebase` set up to be done by default, because it leads to less issues like that).

PR #1998, opened 2019-02-27
fix: cgroup path issue

lifubang:
At that time, if there is no user defined CgroupsPath, we use container's name as cgroup path.
It will cause cgroup rewrite by different containers with the same id in different root path.

For example:
**container test in /run/runc**
```
root@test:/opt/busybox# runc create test
root@test:/opt/busybox# runc list
ID          PID         STATUS      BUNDLE                                            CREATED                          OWNER
test        29438       created     /opt/busybox   2019-02-27T07:04:14.333865241Z   root
root@test:/opt/busybox# runc ps test
root     29438     1  0 15:04 ?        00:00:00 runc init
```

**container test in ./runc**
```
root@test:/opt/busybox# runc --root ./runc create test
root@test:/opt/busybox# runc --root ./runc list
ID          PID         STATUS      BUNDLE                                            CREATED                          OWNER
test        29556       created     /opt/busybox   2019-02-27T07:04:31.913650184Z   root
```

**Then cgroup content error**
```
root@test:/opt/busybox# runc ps test
root     29438     1  0 15:04 ?        00:00:00 runc init
root     29556     1  0 15:04 ?        00:00:00 runc init
```

**And rootless container can't create successful.**
```
test@test:~/busybox$ runc --root ./runc create test
container_linux.go:348: starting container process caused "process_linux.go:279: applying cgroup configuration for process caused \"failed to write 30288 to cgroup.procs: open /sys/fs/cgroup/cpuset/test/cgroup.procs: permission denied\""
```
Because this cgroup path has already created by root.

To fix this problem, we can append a suffix uuid string to the cgroup name.


AkihiroSuda:
> At that time, if there is no user defined CgroupsPath, we use container's name as cgroup path.
> It will cause cgroup rewrite by different containers with the same id in different root path.

Can we just document this behavior and call it a day?

cc @kolyshkin

kolyshkin:
For quite some time we have implemented a policy to not use an existing cgroup which has processes (see https://github.com/opencontainers/runc/pull/3223, https://github.com/opencontainers/runc/pull/3782/changes/82bc89cd10eb5fa5fdb5d77e41b7ea3c07f9d834, https://github.com/opencontainers/runc/issues/3780). I think this is sufficient to solve the problem described here.

PR #1999, opened 2019-02-27
fix rootless container: unrelated error with root flag

lifubang:
If we provide `--root`, we shold not check the XDG_RUNTIME_DIR env value.
For example:
```
test@test:~/busybox$ runc --root ./runc list
ERRO[0000] mkdir /run/user/0/runc: permission denied    
mkdir /run/user/0/runc: permission denied
```
This error is unrelated, because we use ./runc as root dir.

Actually, at that time, we use `--root` first.
But we check the value of XDG_RUNTIME_DIR and create related folder too early.

By the way, if we don't provide `--root`, we should provide the information of `XDG_RUNTIME_DIR env` when there is an permission error.

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,19 +64,12 @@ func main() {
 	v = append(v, fmt.Sprintf("spec: %s", specs.Version))
 	app.Version = strings.Join(v, "\n")
 
+	xdgRuntimeDir := ""
 	root := "/run/runc"
 	if shouldHonorXDGRuntimeDir() {
 		if runtimeDir := os.Getenv("XDG_RUNTIME_DIR"); runtimeDir != "" {
 			root = runtimeDir + "/runc"
-			// According to the XDG specification, we need to set anything in
-			// XDG_RUNTIME_DIR to have a sticky bit if we don't want it to get
-			// auto-pruned.
-			if err := os.MkdirAll(root, 0700); err != nil {
-				fatal(err)
-			}
-			if err := os.Chmod(root, 0700|os.ModeSticky); err != nil {
-				fatal(err)
-			}
+			xdgRuntimeDir = root
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM

cyphar:
The only thing I don't really like about this is that the default is now not obvious to users. Is it possible to delay `MkdirAll` until after the `--root` handling -- but still use `XDG_RUNTIME_DIR` to set the default `--root` value? (This might be a bit annoying to do though.)

lifubang:
>  (This might be a bit annoying to do though.)

a little, need to add a `xdgRuntimeDir` to save the `XDG_RUNTIME_DIR` env value to do check in `app.Before`.

lifubang:
use `!context.IsSet("root") && xdgRuntimeDir != ""` to check whether we should mkdir or not.

cyphar:
LGTM.

lifubang:
@AkihiroSuda @cyphar @mrunalp 
When I use rootless in my pc, the bug is still exist.
```
lifubang@acmcoder:/opt/mygo/src/github.com/opencontainers/runc$ ./runc --root /home/lifubang/test run -d test
ERRO[0000] mkdir /run/user/0/runc: permission denied    
mkdir /run/user/0/runc: permission denied
```

AkihiroSuda:
@lifubang You mean this PR isn't sufficient to solve the issue, or you mean this PR needs to be merged to solve the issue?

AkihiroSuda:
LGTM (if latter)

lifubang:
> @lifubang You mean this PR isn't sufficient to solve the issue, or you mean this PR needs to be merged to solve the issue?

I mean the later.  This PR needs to be merged to solve the issue

AkihiroSuda:
@mrunalp @cyphar still LGTY?

crosbymichael:
LGTM

PR #2000, opened 2019-02-28
fix preserve-fds flag may cause runc hang

lifubang:
`--preserve-fds` may caluse `runc create` and `runc run` hang.
If preserve-fds > 3, and /dev/fd/5 is not exist, runc will hang.
Because at that time, there is a socket pair for `runc init`, 
one of it's fd is 5.
So, when we pass preserve-fds flag, we need to check whether these fds is exist.

```
root@test:/opt/ubuntu# ls /dev/fd/ -alh
dr-x------ 2 root root  0 2月  28 16:36 .
dr-xr-xr-x 9 root root  0 2月  28 16:36 ..
lrwx------ 1 root root 64 2月  28 16:36 0 -> /dev/pts/2
lrwx------ 1 root root 64 2月  28 16:36 1 -> /dev/pts/2
lrwx------ 1 root root 64 2月  28 16:36 2 -> /dev/pts/2
lr-x------ 1 root root 64 2月  28 16:36 3 -> /proc/24479/fd
root@test:/opt/ubuntu# docker-runc run -d --preserve-fds 3 test1
^C^C^C^C^C
root@test:/opt/ubuntu#
```

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -3,7 +3,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -20,6 +19,7 @@ import (
 	"github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/coreos/go-systemd/activation"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


utils_linux.go
```diff
@@ -281,6 +281,11 @@ func (r *runner) run(config *specs.Process) (int, error) {
 	}
 	baseFd := 3 + len(process.ExtraFiles)
 	for i := baseFd; i < baseFd+r.preserveFDs; i++ {
+		_, err := os.Stat(fmt.Sprintf("/dev/fd/%d", i))
```
cyphar: Nit: `/proc/self/fd`.

utils_linux.go
```diff
@@ -281,6 +281,11 @@ func (r *runner) run(config *specs.Process) (int, error) {
...
+		if err != nil {
+			r.destroy()
+			return -1, err
```
cyphar: It'd be nice if you could wrap the error to explain what's going on -- a user isn't going to understand what `stat /proc/self/fd/150: no such file or directory` means. Something like:

```
return -1, errors.Wrap(err, "check that all preserve-fds are present")
```

lifubang: +1 Thanks.

cyphar:
LGTM.

crosbymichael:
LGTM

PR #2005, opened 2019-03-07
makefile: allow debug static build runc

Ace-Tang:
1. remove "-w" in build static runc to enable debug, the flag should
be set in EXTRA_LDFLAGS to make more sense, then user can
set different flags.
```
remove debug info
EXTRA_LDFLAGS="-s -w" make static

allow debug
make static
```
I think the changes can make sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 PROJECT := github.com/opencontainers/runc
 BUILDTAGS ?= seccomp
+EXTRA_LDFLAGS ?= -w
 COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
 COMMIT := $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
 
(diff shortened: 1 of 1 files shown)


seemethere (dismissed):

Ace-Tang:
Hi, @thaJeztah , for the first change, defualt `make` will increase 2m in binary size since it include debug info, or we can not debug unless change the code, for the second change, default `make` behavior won't be affected.

thaJeztah:
:+1: was looking at it for situations where `BUILDTAGS` could be overridden ik our build scripts, in case we'd have to add the `static_build` to it

Ace-Tang:
Oh, sorry, I think I make a mistake, that was `BUILDTAGS ?= seccomp static_build` also apply to `make`  behavior, I will revert the second change, and that will not effect you @thaJeztah , I think I should open a issue to see how can I use systemd cgroup driver with static build runc.

cyphar:
I would add an `EXTRA_LDFLAGS ?= -w` to avoid changing the default binary size. Go binaries are already large enough.

Ace-Tang:
Do you mean this
```
index 097ef8af..7f97b4d0 100644
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 BUILDTAGS ?= seccomp
+EXTRA_LDFLAGS ?= -w
 COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
 COMMIT := $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
```
I add a `EXTRA_LDFLAGS ?= -w` in makefile, so it won't effect default behavior, @cyphar

PR #2012, opened 2019-03-13
Need to setup labeling of kernel keyrings.

rhatdan:
Work is ongoing in the kernel to support different kernel
keyrings per user namespace.  We want to allow SELinux to manage
kernel keyrings inside of the container.

Currently when runc creates the kernel keyring it gets the label which runc is
running with ususally `container_runtime_t`, with this change the kernel keyring
will be labeled with the container process label container_t:s0:C1,c2.

Container running as container_t:s0:c1,c2 can manage keyrings with the same label.

This change required a revendoring or the SELinux go bindings.

github.com/opencontainers/selinux.

### DIFF
--- a/libcontainer/setns_init_linux.go
+++ b/libcontainer/setns_init_linux.go
@@ -34,6 +34,10 @@ func (l *linuxSetnsInit) Init() error {
 	defer runtime.UnlockOSThread()
 
 	if !l.config.Config.NoNewKeyring {
+		if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
+			return err
+		}
+		defer label.SetKeyLabel("")
 		// Do not inherit the parent's session keyring.
 		if _, err := keys.JoinSessionKeyring(l.getSessionRingName()); err != nil {
 			// Same justification as in standart_init_linux.go as to why we
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -48,6 +48,10 @@ func (l *linuxStandardInit) Init() error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 	if !l.config.Config.NoNewKeyring {
+		if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
+			return err
+		}
+		defer label.SetKeyLabel("")
 		ringname, keepperms, newperms := l.getSessionRingParams()
 
 		// Do not inherit the parent's session keyring.
--- a/vendor.conf
+++ b/vendor.conf
@@ -5,7 +5,7 @@ github.com/opencontainers/runtime-spec 29686dbc5559d93fb1ef402eeda3e35c38d75af4
 # Core libcontainer functionality.
 github.com/checkpoint-restore/go-criu v3.11
 github.com/mrunalp/fileutils ed869b029674c0e9ce4c0dfa781405c2d9946d08
-github.com/opencontainers/selinux v1.0.0-rc1
+github.com/opencontainers/selinux v1.2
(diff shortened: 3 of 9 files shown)


rhatdan:
@giuseppe FYI

vendor.conf
```diff
@@ -5,7 +5,7 @@ github.com/opencontainers/runtime-spec 29686dbc5559d93fb1ef402eeda3e35c38d75af4
 # Core libcontainer functionality.
 github.com/checkpoint-restore/go-criu v3.11
 github.com/mrunalp/fileutils ed869b029674c0e9ce4c0dfa781405c2d9946d08
-github.com/opencontainers/selinux v1.0.0-rc1
+github.com/opencontainers/selinux v1.2
```
thaJeztah: (not directly related to this PR); I noticed that these tags miss a trailing `.0`, and because of that are following the SemVer syntax (which may complicate using them with Go mod)

libcontainer/standard_init_linux.go
```diff
@@ -48,6 +48,10 @@ func (l *linuxStandardInit) Init() error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 	if !l.config.Config.NoNewKeyring {
+		if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
```
cyphar: Just to double-check -- are you sure this needs to be done *before* we create a new session? (Is `SetKeyLabel` setting what the label will be for all future keys or the label for the current key?)

libcontainer/setns_init_linux.go
```diff
@@ -34,6 +34,10 @@ func (l *linuxSetnsInit) Init() error {
 	defer runtime.UnlockOSThread()
 
 	if !l.config.Config.NoNewKeyring {
+		if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
```
cyphar: Since we are joining an existing keyring here, does it make sense to set a label (if the label is different for a different container, what will happen)?

vendor.conf
rhatdan: I can add another .0 if necessary.  in the next version.

libcontainer/standard_init_linux.go
rhatdan: Yes this says to the kernel, any new keyrings that get created by this process and its children should get created with the specified label.
When the new keyring gets created it gets created with the correct label.
Later on the defaulf SetKeyLabel("") tells the kernel to go back to the default labeling.

rhatdan: BTW I have been testing this patch with the updated kernel, which allows sepate kernel keyrings for each new UserNamespace.

vendor.conf
thaJeztah: Thanks! I noticed it earlier this week, and this PR made me remember that I saw that 🤗

libcontainer/standard_init_linux.go
mrunalp: I am not clear why we need to reset it back. Shouldn't it pass down to the container process?

rhatdan: Yes it will, the reset is on defer.  Not really important for runc, but if people vendor in this code, then we would want to make sure other code paths don't accidently set the label to this container.

cyphar: Yup, we've discussed the `defer` handling for SELinux in the past. Even though it doesn't really make sense for `runc` it avoids people vendoring it and shooting themselves in the foot (not that SELinux is the only thing that is a massive foot-gun in `libcontainer`).

cyphar:
LGTM. I haven't tested on a kernel with keyring support, but I take @rhatdan's word that this works as it should.

mrunalp:
LGTM

PR #2015, opened 2019-03-14
Use getenv not secure_getenv

justincormack:
secure_getenv is a Glibc extension and so this code does not compile
on Musl libc any more after this patch.

secure_getenv is only intended to be used in setuid binaries, in
order that they should not trust their environment. It simply returns
NULL if the binary is running setuid. If runc was installed setuid,
the user can already do anything as root, so it is game over, so this
check is not needed.


crosbymichael:
LGTM

cyphar:
LGTM.

PR #2016, opened 2019-03-15
fix cloned_binary fallback corner case

lifubang:
Because the `/run` folder in all my linux cloud server have `noexec` mount flag. So if cloned_binary fallback to temp file copy, there will be `nsenter: could not ensure we are a cloned binary: Permission denied` error.

I think we can check `noexec` mount flag first before we clone binary.

I think this is a corner case, so it is not important. It may just only effect kernel < 3.17 .

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -483,6 +483,7 @@ func (c *linuxContainer) commandTemplate(p *Process, childPipe *os.File) (*exec.
 	cmd.Env = append(cmd.Env,
 		fmt.Sprintf("_LIBCONTAINER_INITPIPE=%d", stdioFdCount+len(cmd.ExtraFiles)-1),
 		fmt.Sprintf("_LIBCONTAINER_STATEDIR=%s", c.root),
+		fmt.Sprintf("_LIBCONTAINER_USERHOME=%s", os.Getenv("HOME")),
 	)
 	// NOTE: when running a container with no PID namespace and the parent process spawning the container is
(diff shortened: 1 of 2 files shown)


cyphar:
I **really** don't like how complicated this fallback code is getting -- not to mention how complicated such a trivial fix has become. If `/run` is often mounted `noexec`, then we should not try to use it -- rather than this complicated fallback.

I'm of a mixed mind about using `$HOME` -- there are lots of users that don't have a `$HOME` and it's a bit odd to use `/root` for this. Also, what if the host has a read-only `/` -- this is the reason why I wanted to use `/run` in the first place (because in order for `runc` to work it must be able to write to `/run`). We could make our own `tmpfs` mount (like I did in #2006) but that's just adding more complexity and won't help rootless users.

lifubang:
> If `/run` is often mounted `noexec`, then we should not try to use it -- rather than this complicated fallback.

Runc's default state dir is in `/run`, users may not know what's the meaning of  `nsenter: could not ensure we are a cloned binary: Permission denied error.`
And for rootless user, the `XDG_RUNTIME_DIR` env often is `/run/user/0`.

rata:
Is this still relevant?

kolyshkin:
> Is this still relevant?

1. We moved binary cloning from C to Go (it now lives in libcontainer/exeseal.
2. In the new code we have a bunch of directories to try for tmpDir (see commit e089db3b).

So yes, it's not relevant; closing.

PR #2018, opened 2019-03-16
add prompt when rootless user has no read access to runc binary file

lifubang:
Another corner case:
If administrator have some protections of `CVE-2019-5736` for rootless container, it is safe enough, but we can't trust it.

 For example:
```
chmod 111 /usr/bin/runc
```

When we start a rootless container, it will get error:
```
nsenter: could not ensure we are a cloned binary: Permission denied
container_linux.go:345: starting container process caused "process_linux.go:293: copying bootstrap data to pipe caused \"write init-p: broken pipe\""
```

The user may not understand what has happened. I think we should add some more prompts:
```
you have no read access to runc binary file
nsenter: could not ensure we are a cloned binary: Permission denied
container_linux.go:345: starting container process caused "process_linux.go:293: copying bootstrap data to pipe caused \"write init-p: broken pipe\""
```

### DIFF
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -95,8 +95,10 @@ static int is_self_cloned(void)
 	struct statfs fsbuf = {};
 
 	fd = open("/proc/self/exe", O_RDONLY|O_CLOEXEC);
-	if (fd < 0)
+	if (fd < 0) {
+		fprintf(stderr, "you have no read access to runc binary file\n");
 		return -ENOTRECOVERABLE;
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

AkihiroSuda:
@cyphar PTAL

AkihiroSuda:
@kolyshkin @mrunalp PTAL

AkihiroSuda:
@cyphar @kolyshkin @mrunalp this one is easy to review

cyphar:
LGTM.

Issue #2022, opened 2019-03-21
Regression: zombie runc:[1:CHILD] processes


LittleLightLittleFire:
Whenever we execute code in nsenter.c and read back the child pid, we always need to be aware that the code in nsenter will spawn a zombie runc:[1:CHILD] process that needs to be cleaned up by the parent.

The code in process_linux.go was updated to support cgroup namespaces and uses the nsenter.c to spawn another process.

https://github.com/opencontainers/runc/blob/f56b4cbeadc407e715d9b2ba49e62185bd81cef4/libcontainer/process_linux.go#L235-L240

This does not clean up runc:[1:CHILD] as done here:

https://github.com/opencontainers/runc/blob/f56b4cbeadc407e715d9b2ba49e62185bd81cef4/libcontainer/process_linux.go#L155-L167

I've prepared a [PR](https://github.com/opencontainers/runc/pull/2023) to address this.

PR #2023, opened 2019-03-21
Fixes regression causing zombie runc:[1:CHILD] processes

LittleLightLittleFire:
Whenever processes are spawned using nsexec, a zombie runc:[1:CHILD]
process will always be created and will need to be reaped by the parent

Fixes https://github.com/opencontainers/runc/issues/2022


cyphar:
LGTM.

crosbymichael:
LGTM

PR #2025, opened 2019-03-26
Update Dockerfile

valerianpereira:


Dockerfile
```diff
@@ -34,10 +34,9 @@ RUN dpkg --add-architecture armel \
 # not require an entry in /etc/passwd to operate, one of the tests uses
 # `git clone` -- and `git clone` does not allow you to clone a
 # repository if the current uid does not have an entry in /etc/passwd.
-RUN useradd -u1000 -m -d/home/rootless -s/bin/bash rootless
-
 # install bats
-RUN cd /tmp \
+RUN useradd -u1000 -m -d/home/rootless -s/bin/bash rootless \
```
thaJeztah: Can you explain why you updated this? Combining this with the next step means that the `useradd` step will no longer be cached if bats is updated; also, given that there would be no overlap in files being touched in both steps, reducing steps/layers won't result in a smaller image

PR #2031, opened 2019-03-29
Add selinux validate in runc exec

### LINKED ISSUE #2030: Container Creation error: Container creation error: container_linux.go:345: starting container process caused "process_linux.go:430:  container init caused \"write /proc/self/attr/keycreate: invalid argument\""
distro:
```
Linux cdl-2080 4.15.0-29-generic #31~16.04.1-Ubuntu SMP Wed Jul 18 08:54:04 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
runc version:
```
runc version 1.0.0-rc6+dev
commit: 2b18fe1d885ee5083ef9f0838fee39b62d653e30
spec: 1.0.1-dev
```

lifubang:

1. Fix regression introduced by https://github.com/opencontainers/runc/pull/2012/commits/cd96170c10b35a37f90040fe55529185b9a5ec30 , the related commit is in https://github.com/opencontainers/selinux/commit/e5c68ba81868de0c57b61832c2197cb31008a40e#diff-08374585d1f5b66358d612f6292a3fae
There is no `if processLabel == "" {` check now.
Fixed by #2032

2. Fixes #2030 check nil for `selinuxLabel` before we `label.SetKeyLabel`, this is also introduced by https://github.com/opencontainers/runc/pull/2012/commits/cd96170c10b35a37f90040fe55529185b9a5ec30 ;
Because the validate is : https://github.com/opencontainers/runc/blob/84cba4cbc1767353ae9d999c61f62a50c538589f/libcontainer/configs/validate/validator.go#L96-L107
Fixed by #2032

3. Fix if we use `selinuxLabel` in `runc exec`, we should validate selinux again. If the system disable selinux, it will raise error `exec failed: container_linux.go:345: starting container process caused "write /proc/self/attr/keycreate: invalid argument"`. For example:
(1) use '--process-label' in `runc exec`;
(2) use `selinuxLabel` in `process.json` when use `--process` in `runc exec`;
(3) add `selinuxLabel` to config.json in bundle dir after the container started.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -224,5 +224,5 @@ func getProcess(context *cli.Context, bundle string) (*specs.Process, error) {
 		}
 		p.User.AdditionalGids = append(p.User.AdditionalGids, uint32(gid))
 	}
-	return p, nil
+	return p, validateProcessSpec(p)
 }
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -17,6 +17,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/specconv"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	selinux "github.com/opencontainers/selinux/go-selinux"
 
 	"github.com/coreos/go-systemd/activation"
 	"github.com/pkg/errors"
@@ -386,6 +387,9 @@ func validateProcessSpec(spec *specs.Process) error {
 	}
 	if len(spec.Args) == 0 {
 		return fmt.Errorf("args must not be empty")
+	}
+	if spec.SelinuxLabel != "" && !selinux.GetEnabled() {
+		return fmt.Errorf("selinux label is specified in config, but selinux is disabled or not supported")
 	}
 	return nil
 }


thaJeztah:
ping @rhatdan ptal

libcontainer/setns_init_linux.go
```diff
@@ -33,11 +34,18 @@ func (l *linuxSetnsInit) Init() error {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
+	// If config.json is modified after the container started, we should validate selinux again.
+	if l.config.ProcessLabel != "" && !selinux.GetEnabled() {
+		return fmt.Errorf("selinux label is specified in config, but selinux is disabled or not supported")
+	}
```
cyphar: From memory, the config.json is not re-parsed for `runc exec` -- we use the config that's in `/run/runc/$ctr/...` which means that there's no chance it could change between runs (other than `runc update` but that only really does `cgroup` changes).

libcontainer/setns_init_linux.go
```diff
@@ -33,11 +34,18 @@ func (l *linuxSetnsInit) Init() error {
...
+
 	if !l.config.Config.NoNewKeyring {
-		if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
-			return err
+		if l.config.ProcessLabel != "" {
+			if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
```
cyphar: @rhatdan I think that all of the `Set*Label` commands should be a no-op if `l.config.ProcessLabel` is `""` -- so as to avoid these sorts of problems.

cyphar:
Yeah, I didn't catch this when I reviewed #2012. Looks okay, minus my comments. We really should've fixed this for rc7... Guess it's time for an rc8 then...

libcontainer/setns_init_linux.go
lifubang: > From memory, the config.json is not re-parsed for `runc exec`

No, it re-parsed in here: https://github.com/opencontainers/runc/blob/84cba4cbc1767353ae9d999c61f62a50c538589f/exec.go#L170

libcontainer/setns_init_linux.go
lifubang: Yes, it does make sense for don't check nil in definition of `label.Set*Label`, because we maybe need to delete the process label, so we should check `l.config.ProcessLabel` is `""` for all the `Set*Label` call in `runc`.

rhatdan:
Yes, I think we need to check this in Runc.  What I don't understand is why this is blowing up, while the other calls do not.

But no SELinux calls should be made if the label is ""

rhatdan:
LGTM

libcontainer/setns_init_linux.go
lifubang: @cyphar Besides `re-parse` config.json bundle dir, there are 3 situations use selinux in `runc exec`, I have update the PR description.
(1) use '--process-label' in runc exec;
(2) add selinuxLabel in process.json when use --process in runc exec;
(3) add selinuxLabel to config.json after the container started.

lifubang: I move the `selinux` validate to `utils_linux.go` when `exec`, so we can do this check before we start `runc init`. Does it make sense?

cyphar: Yes, it would be much better to do it *a lot* earlier.

cyphar:
> What I don't understand is why this is blowing up, while the other calls do not.

I've figured it out -- it's because of older kernels and so this fix actually isn't sufficient (if you're using an older SELinux machine). I'm running AppArmor on a 5.0.3 kernel and I have a `/proc/self/attr/keycreate` file which will happily accept `""` as an argument (even though it's wrong for SELinux code to be touching it on an AppArmor machine).

So, either runc or go-selinux need to detect whether `/proc/self/attr/keycreate` is present before trying anything. I would argue this would both be cleaner and make more sense in go-selinux than runc -- because then we'll have to be aware of SELinux internals.

And if we're going to do that sort of check, I think it would also make sense to no-op go-selinux calls where the label is `""` -- unless `""` actually makes sense as an SELinux label?

cyphar:
I will submit a PR to go-selinux along the above lines, and then we can update this PR to re-vendor it and then remove some of the changes (but the "Is SELinux enabled" for `runc exec` change should stay).

rhatdan:
"" actuall makes sense as an SELinux label.  This tells go-selinux to turn back on the default labeling.

label.SetProcessLabel("system_u:system_r:container_t:s0:c1,c2)
defer label.SetProcessLabel("")

cyphar:
Hmm, so that means really `runc` should be gating all `Set.*Label` calls with `selinux.IsEnabled` (or the standard `Label != ""`)? We do this already, but not nearly as explicitly as it appears we should be -- because it's a bit weird for a SELinux library to write to `/proc/self/attr` even though SELinux isn't in use. It would end up with weird results for AppArmor profiles (for instance).

rhatdan:
We could embed the isselinuxenabled check in the label calls, and only check it if the "" is passed.

cyphar:
I will send a PR with these changes (the is-selinux-enabled checks and the backwards-compatibility changes) tomorrow.

lifubang:
Dear experts, after I see https://github.com/opencontainers/selinux/pull/49 , do we really need patch 49 in runc? Because consider:
```
label.SetProcessLabel(labelVar)
defer label.SetProcessLabel("")
```
If `labelVar` is not empty, it is checked whether selinux enabled or not by runc.
If `labelVar` is empty, the code becomes:
```
label.SetProcessLabel("")
defer label.SetProcessLabel("")
```
I think it has no meanings to the system even though selinux is enabled.

So I think we should check `labelVal` in runc, just like check `config.cwd` is empty or not.

If I'm wrong, please point out. Thanks.

lifubang:
If labelVar is empty, I think this means the user doesn't want to use selinux in runc.

cyphar:
@lifubang While it would make the operations a no-op, it's much better to do it in `go-selinux` so that we don't forget to do it in `runc` (like we forgot in #2012 and in several other places throughout runc). Function calls are cheaper than future maintenance work when someone forgets to add a check again.

lifubang:
> Function calls are cheaper than future maintenance work when someone forgets to add a check again.

I know what's the purpose now. Thank you for your explanation.

rhatdan:
I have opened this PR https://github.com/opencontainers/selinux/pull/49 to fix the issue in go-selinux.

lifubang:
As the regression is fixed by #2032 , so this PR becomes enhancement. The main purpose of this PR is:
1. If the user don't provide `selinuxLabel`, I think this means the user doesn't want to use selinux in runc, so we shouldn't call any selinux API at that time.
2. Add selinux validate if the user provide `selinuxLabel` when `runc exec`.

libcontainer/setns_init_linux.go
```diff
@@ -34,10 +34,12 @@ func (l *linuxSetnsInit) Init() error {
 	defer runtime.UnlockOSThread()
 
 	if !l.config.Config.NoNewKeyring {
-		if err := label.SetKeyLabel(l.config.ProcessLabel); err != nil {
-			return err
+		if l.config.ProcessLabel != "" {
```
cyphar: These checks aren't necessary anymore.

cyphar:
My only comment is we don't need any of the `ProcessLabel != ""` checks. Everything else looks fine. I'll send out the vote for an rc8 release once this is merged.

libcontainer/setns_init_linux.go
lifubang: updated

cyphar:
LGTM. Thanks @lifubang.

crosbymichael:
LGTM

PR #2032, opened 2019-04-02
Fix SELinux failures on disabled SELinux Machines

rhatdan:
On some machines when setting the SELinux key labels to "", we are seeing
failures that cause runc to fail.  Even if SELinux is disabled.

This check will ignore callers calling SELinux Set*Label functions with ""
when SELinux is disabled.


crosbymichael:
LGTM

mrunalp:
LGTM

cyphar:
(LGTM).

PR #2035, opened 2019-04-08
specconv: always set "type: bind" in case of MS_BIND

cyphar:
We discovered in umoci that setting a dummy type of "none" would result
in file-based bind-mounts no longer working properly, which is caused by
a restriction for when specconv will change the device type to "bind" to
work around rootfs_linux.go's ... issues.

However, bind-mounts don't have a type (and Linux will ignore any type
specifier you give it) because the type is copied from the source of the
bind-mount. So we should always overwrite it to avoid user confusion.

Ref: opencontainers/runc#1753
Ref: openSUSE/umoci#294


crosbymichael:
LGTM

mrunalp:
LGTM

PR #2037, opened 2019-04-09
Proposal: just signal all processes inside the container

lifubang:
`runc kill --help` says:
`--all, -a  send the specified signal to all processes inside the container`

But if we use shared cgroup path to create 2 containers a & b.
When we send signal KILL to container a by `runc kill -a a 9`,
 it will also kill container b's processes, not just only processes inside container a.

I think we should have a way to send a signal to just only all processes inside one container, not all processes in shared cgroup path.

This PR provides a way to implement this operation.
The method is to use sub cgroup path in freezer cgroup subsystem.

I think there are some discussion in:
https://github.com/containerd/containerd/issues/2558
https://github.com/containerd/containerd/pull/3149
We don't know when to use `runc kill -a cid` in `Shared cgroup path` and `Shared pid namespace`.
https://github.com/lifubang/runc/blob/dd2d6faab8ca804c7ab320003d69306917812c07/libcontainer/state_linux.go#L40-L46

And there is another effect:
If we use host pid namespace, we must use `signalAllProcess` to kill all processes inside the container when destroy the container, but if we have used shared cgroup path, other container's processes will be killed at that time.

Signed-off-by: Lifubang lifubang@acmcoder.com


kolyshkin:
Since v1.1.0, we no longer allow shared cgroups (see #3132, #3223), thus closing.

kolyshkin:
(and the latest shared pidns discussion happened in #4102)

PR #2038, opened 2019-04-10
`r.destroy` can defer exec in `runner.run` method.

imxyb:
Signed-off-by: Xiao YongBiao <xyb4638@gmail.com>

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -266,13 +266,17 @@ type runner struct {
 }
 
 func (r *runner) run(config *specs.Process) (int, error) {
-	if err := r.checkTerminal(config); err != nil {
-		r.destroy()
+	var err error
+	defer func() {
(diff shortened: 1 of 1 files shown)


imxyb:
@caniszczyk @cyphar PTAL, thanks!

crosbymichael:
LGTM

mrunalp:
LGTM

utils_linux.go
```diff
@@ -266,13 +266,17 @@ type runner struct {
 }
 
 func (r *runner) run(config *specs.Process) (int, error) {
```
cyphar: The correct way of handling this would've been to use `(_ int, Err error)` and then using the `error` in the defer. Otherwise, if you have a `return -1, fmt.Errorf()` you will treat it as a `nil`.

PR #2041, opened 2019-04-22
Change the permissions of the notify listener socket to rwx for everyone

jburianek:
When runc is started as a `Type=notify` systemd service, runc opens up its own listening socket inside the container to act as a proxy between the container and systemd for passing notify messages.

The domain socket that runc creates is only writeable by the user running runc however, so if the container has a different UID/GID then nothing inside the container will be able to write to the socket.

The fix is to change the permissions of the notify listener socket to 0777.

Signed-off-by: Joe Burianek <joe.burianek@pantheon.io>

### DIFF
--- a/notify_socket.go
+++ b/notify_socket.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"fmt"
 	"net"
+	"os"
 	"path/filepath"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -57,6 +58,12 @@ func (s *notifySocket) setupSocket() error {
 
 	socket, err := net.ListenUnixgram("unixgram", &addr)
 	if err != nil {
+		return err
+	}
+
+	err = os.Chmod(s.socketPath, 0777)
+	if err != nil {
+		socket.Close()
 		return err
 	}
 


rhatdan:
LGTM
@giuseppe PTAL

jburianek:
@giuseppe Have you had a chance to take a look at this?

Currently, unless the user that invokes runc and the user inside the container are the same, no sd_notify messages can be passed between the container and systemd.

utils_linux.go
```diff
@@ -419,7 +419,7 @@ func startContainer(context *cli.Context, spec *specs.Spec, action CtAct, criuOp
 	}
 
 	if notifySocket != nil {
-		err := notifySocket.setupSocket()
+		err := notifySocket.setupSocket(spec.Process.User)
```
cyphar: How does this handle user namespaces? I have a feeling you'd need to use the host-uid/gid here rather than the container one.

giuseppe: We could probably just change the mode to 0777, which is the same mode used on the host by systemd for `/run/systemd/notify`.  @jburianek what do you think?

giuseppe:
> @giuseppe Have you had a chance to take a look at this?

sorry, I was out of the office last week and I somehow missed the notification.  Left a comment

utils_linux.go
jburianek: @cyphar Good point, I forgot about user namespaces.

@giuseppe That seems reasonable, I've modified the PR with your suggestion.

giuseppe:
LGTM (I am not a maintainer)

jburianek:
Updated the PR title and comment to reflect the current state of this PR.

@cyphar Does this seem good to you? Or would you prefer I incorporate support for user namespaces based on your comment? Thanks.

jburianek:
Any more feedback on this? It'd be awesome to get this fixed.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #2044, opened 2019-04-24
fix a cgroup cpu quota/period order issue.

zq-david-wang:
When setting/updating child cpu cgroup cfs_quota_us and cfs_period_us, the order is significant because the child node should not have a higher quota/period ratio than its parent. This means if always setting period first would cause kernel error.
For example, say parent node has quota/period set to 10000/10000, and we want to change child to 1000/1000, if set period first, than after the period change and before the quota change, the ration is 10000/1000, and this setting make child node have larger quota ratio than its parent and rejected by kernel. 

This patch would get current child value first, and make sure the first setting always make the quota/period ratio lower than current (valid) ratio.
EDIT: 
A simpler strategy is to use a period->quota---->period sequence, if setting period failed, retry setting it after seting quota.

And there is an issue reported in kubernetes project
https://github.com/kubernetes/kubernetes/issues/76704

EDIT:
To trigger the issue, we need 
1. the parent cgroup node has a non-infinite cgroup quota
2. Before setting quota, the current child node's quota is non-infinite.
And when using  systemd cgroup driver with kubernetes/runc, both conditions are met...
kubelet would create a parent node for pod, and set its quota according to resource limit, and when create child node via systemd cgroup driver, systemd would set peirod to 100ms and quota to a calculated value other than -1.  Then further change to quota/period would be rejected by liinux kernel if child node have higher quota/period ratio after setting period and before setting quota.

### DIFF
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -69,6 +69,21 @@ func (s *CpuGroup) Set(path string, cgroup *configs.Cgroup) error {
 			return err
 		}
 	}
+	// The order of setting cfs_quota_us and cfs_period_us is significant, since setting cgroup child node
+	// quota/period ratio to a value higher than its parent would be rejected by linux kernel with invalid arugment.
+	// And setting quota/period is a two-steps no-automic operation in the middle of which the change would
+	// be rejected by kernel if the order of quota/peroid changes is fixed. E.g.
+	// Say parent -> child tree have quota/period set as following n1(100000/100000)->n2 (100000/100000)->n3(100000/100000)
+	// 1. setting n3 to 10000/10000 should have different order than setting n3 to 1000000/1000000, because in the middle,
+	// the quota/period ratio of 100000/10000 and 1000000/100000 is not valid
+	// 2. setting n1 to 10000/10000 or 1000000/1000000 is similiar to case 1
+	// 3. setting n2, a middle node in the cgroup tree,  to 10000/10000 is not possbile without setting quota to infinite first.
+	// To cover all valid cases, set quota to infinite first, and then change period, and then quota.
+	if cgroup.Resources.CpuQuota != 0 {
+		if err := writeFile(path, "cpu.cfs_quota_us", "-1"); err != nil {
+			return err
+		}
+	}
 	if cgroup.Resources.CpuPeriod != 0 {
 		if err := writeFile(path, "cpu.cfs_period_us", strconv.FormatUint(cgroup.Resources.CpuPeriod, 10)); err != nil {
 			return err


libcontainer/cgroups/fs/cpu.go
```diff
@@ -69,16 +69,37 @@ func (s *CpuGroup) Set(path string, cgroup *configs.Cgroup) error {
...
+		}
+	}
 	if cgroup.Resources.CpuPeriod != 0 {
 		if err := writeFile(path, "cpu.cfs_period_us", strconv.FormatUint(cgroup.Resources.CpuPeriod, 10)); err != nil {
 			return err
 		}
 	}
-	if cgroup.Resources.CpuQuota != 0 {
+	// if quota is already set, don't set it again.
+	if (!qs) && cgroup.Resources.CpuQuota != 0 {
```
cyphar: You can just do `if !qs && ...`.

zq-david-wang: fixed~

zq-david-wang:
Just realize that the code only address the situation for changing leaf node of a cgroup tree. When a parent node is changed, it should make it larger first.. And worse, if one want to change a node in the middle of the cgroup tree, it becoms a mission impossible... 

Say we have three cgroup nodes a, b, c, a is b's parent, and b is c's parent
a,b and c have same quota/period values, say 10000/10000, and we want to change b to 100000/100000. Changing quota first would make b's ratio 100000/10000, and its parent node a would complain, and if chaning period first, b's ration 10000/100000 would be lower than its child.....

zq-david-wang:
My understanding is that runc is mostly used  to manage leaf node of cgroup tree, right? If so, this fix should work "most of the time"...

cyphar:
Yes, runc only ever manages the leaf of tree. If the admin decides to misconfigure the cgroup tree to make runc crash there isn't much we can (or should) do about it.

zq-david-wang:
I have make changes to use a simpler strategy: set period first, if failed than retry if after setting quota.
This way, it should handle most of cases except the impossible one I mentioned before.

zq-david-wang:
https://github.com/opencontainers/runc/commit/c44aec9b23f89ca40434fe5f86693870ef3bf9f9
This commit make the initial value of cpu.cfs_quota_us non-infinite via systemd,

libcontainer/cgroups/fs/cpu.go
```diff
@@ -69,16 +69,30 @@ func (s *CpuGroup) Set(path string, cgroup *configs.Cgroup) error {
 			return err
 		}
 	}
+	// The order of setting cfs_quota_us and cfs_period_us is significant, since cgroup child node
+	// should not have a higher quota/period ratio than its parent.
+	// Use period->quota-->period sequence to make sure no matther parent or child node is changed,
```
crosbymichael: `matther` -> `matter`

crosbymichael:
One question, why not use re-order the logic on what is set first instead of this re-order bool?

zq-david-wang:
@crosbymichael  Thanks for the comments.
My first version of the code is to reorder the sequence to make the ratio smaller first by reading current node's period and comparing it with new value. If the new period is larger than current period, set period first, otherwise set quota first. 
But this only work when current cgroup node is a leaf of cgroup tree. Eventhough this is how runc is used mostly nowadays, I feel it would be better to address both situation incase one day runc is used to change some non-leaf node of cgroup tree.....

I can put the first version of code changes in another PR is necessary.

libcontainer/cgroups/fs/cpu.go
zq-david-wang: fixed~!

zq-david-wang:
@crosbymichael  @cyphar 
Just found out another solution, 
1. change node's cpu quota to -1
2. change period
3. change quota
This sequence can handle all situations no matter where the position of current node is in the cgroup tree, as long as the final quota/period ratio is valid regarding current node's parent and children.

Need your opinion on this.

AkihiroSuda:
Please add tests

PR #2047, opened 2019-05-01
Move systemd.Manager initialization into a function in that module

filbranden:
This will permit us to extend the internals of systemd.Manager to include further information about the system, such as whether cgroupv1, cgroupv2 or both are in effect.

Furthermore, it allows a future refactor of moving more of UseSystemd() code into the factory initialization function.


crosbymichael:
LGTM

mrunalp:
LGTM

PR #2049, opened 2019-05-02
libcontainer/utils: fix compilation on Darwin

### LINKED ISSUE #2048: compilation fails on Mac
If I try to install libcontainer/utils on Mac, I get:

```
go install -v ./libcontainer/utils
github.com/opencontainers/runc/libcontainer/utils
# github.com/opencontainers/runc/libcontainer/utils
libcontainer/utils/utils_unix.go:39:62: undefined: unix.SOCK_CLOEXEC
```

This constant is only defined for Linux machines.

kevinburke:
NewSockPair() only works on Linux due to the presence of the
SOCK_CLOEXEC constant, which is only present for Linux machines.

Fixes #2048.

### DIFF
--- a/libcontainer/utils/cmsg.go
+++ b/libcontainer/utils/cmsg.go
@@ -1,4 +1,4 @@
-// +build linux
+// +build !windows
 
 package utils
 
--- a/libcontainer/utils/utils_linux.go
+++ b/libcontainer/utils/utils_linux.go
@@ -1,5 +1,3 @@
-// +build !windows
-
 package utils
 
 import (


libcontainer/utils/utils_unix.go
```diff
@@ -1,4 +1,4 @@
-// +build !windows
+// +build !windows,!darwin
```
cyphar: You could just rename this file `utils_linux.go`.

rhatdan:
Yes lets rename this to utils_linux.go

cyphar:
I forgot to mention -- while I'm fine with accepting a minor change like this it should be noted that `runc` is a Linux-only project. We don't plan to target any other operating systems.

kevinburke:
OK, renamed

AkihiroSuda:
needs rebase

AkihiroSuda:
Closing because of inactivity, but feel free to reopen.

PR #2062, opened 2019-05-22
 Add support for specifying systemd timeout property

mrunalp:
Allow setting systemd timeout property for the container scope so that it gets terminated gracefully on system shutdown or reboot.

This is needed to fix https://github.com/kubernetes/kubernetes/issues/77873


main.go
```diff
@@ -109,6 +109,10 @@ func main() {
 			Name:  "systemd-cgroup",
 			Usage: "enable systemd cgroup support, expects cgroupsPath to be of form \"slice:prefix:name\" for e.g. \"system.slice:runc:434234\"",
 		},
+		cli.IntFlag{
+			Name:  "systemd-stop-timeout",
```
rhatdan: Should this be more flexible something like:
--systemd-options stop-timeout=50
Then it would be easier to add future values without having to add future options.

cyphar: I agree having a more extensible flag would be nicer.

libcontainer/cgroups/systemd/apply_nosystemd.go
```diff
@@ -18,7 +18,7 @@ func UseSystemd() bool {
 	return false
 }
 
-func NewSystemdCgroupsManager() (func(config *configs.Cgroup, paths map[string]string) cgroups.Manager, error) {
+func NewSystemdCgroupsManager() (func(config *configs.Cgroup, paths map[string]string, systemdProperities *configs.SystemdProperties) cgroups.Manager, error) {
```
giuseppe: nit: typo here s|systemdProperities|systemdProperties|

main.go
mrunalp: Sounds good :+1: I will update the PR to support that.

mrunalp:
Updated the PR.

rhatdan:
LGTM

rhatdan:
@avagin @caniszczyk @crosbymichael @cyphar @dqminh @hqhq @rjnagal @vmarmol PTAL

utils_linux.go
```diff
@@ -46,9 +47,33 @@ func loadFactory(context *cli.Context) (libcontainer.Factory, error) {
...
+				sp = &configs.SystemdProperties{}
+				systemdProperties := context.GlobalStringSlice("systemd-property")
+				for _, prop := range systemdProperties {
+					pair := strings.Split(prop, "=")
+					if len(pair) != 2 {
+						return nil, fmt.Errorf("systemd-property %q not of the form key=pair", prop)
+					}
+					switch pair[0] {
+					// TODO: Reuse SupportedStemdProperties here when we add more properties
+					case "stop-timeout":
```
crosbymichael: Any reason why we would want to switch on these one by one and not support a generic map so that any additional properties can be added?  

```bash
runc --systemd-property TimeoutStopUSec=1000000 --systemd-property TimeoutStartUSec=10000000
``` 

Seems like if we are going to add this, might as well make it generic so we can support all of them.  I'd probably just `map[string]interface{}` as the `newProp` func already handles the interface and makes a varint with it

mrunalp:
Closing in favor of #2224

PR #2063, opened 2019-05-28
Add public API interface

saschagrunert:
Main motivation of this commit is that dependent projects do not have to
shell out for using runc. This provides better performance and a general
type-safety.

This commit adds a public API to runc, which currently implements the
basic Version() API for demonstration purposes. A public interface
provides the possibility to mock the results by the consumer of the API.
The configuration of the API will be done via a modified builder
pattern, which is directly part of the public interface.

Later on, the all CLI commands within the `main` package should utilize
the configured API instance from the context to retrieve its results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,19 +29,23 @@ SHELL := $(shell command -v bash 2>/dev/null)
 
 .DEFAULT: runc
 
+LDFLAGS := -X $(PROJECT)/api.gitCommit=${COMMIT} \
+	-X $(PROJECT)/api.version=${VERSION} \
+	$(EXTRA_LDFLAGS)
+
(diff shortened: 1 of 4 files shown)


saschagrunert:
If you’re generally fine with the mentioned approach then I would continue with testing and further API implementations. :)

crosbymichael:
@saschagrunert have you seen https://github.com/containerd/go-runc ?

AkihiroSuda:
I'm closing this because I assume go-runc can cover the usecase, but feel free to reopen if not.

saschagrunert:
Sounds good, thanks for the response. :)

PR #2065, opened 2019-05-29
Fix cgroup hugetlb size prefix for kB

odinuge:
The hugetlb cgroup control files (introduced here in 2012:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=abb8206cb0773)
use "KB" and not "kB"
(https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/hugetlb_cgroup.c?h=v5.0#n349).

The behavior in the kernel has not changed since the introduction, and
the current code using "kB" will therefore fail on devices with small
amounts of ram (see
https://github.com/kubernetes/kubernetes/issues/77169) running a kernel
with config flag CONFIG_HUGETLBFS=y

As seen from the code in "mem_fmt" inside hugetlb_cgroup.c, only "KB",
"MB" and "GB" are used, so the others may be removed as well.

Here is a real world example of the files inside the
"/sys/kernel/mm/hugepages/" directory:
- "hugepages-64kB"
- "hugepages-2048kB"
- "hugepages-32768kB"
- "hugepages-1048576kB"

And the corresponding cgroup files:
- "hugetlb.64KB._____"
- "hugetlb.2MB._____"
- "hugetlb.32MB._____"
- "hugetlb.1GB._____"

Signed-off-by: Odin Ugedal <odin@ugedal.com>

Screenshot from 2019-05-29 22-37-33

Noticed this when tinkering around with kubernetes on a Raspberry PI (1GB ram) running Arch Linux (`Linux k8s-003 5.1.5-1-ARCH #1 SMP Sat May 25 13:23:49 MDT 2019 aarch64 GNU/Linux`)

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -409,14 +409,22 @@ func RemovePaths(paths map[string]string) (err error) {
 }
 
 func GetHugePageSize() ([]string, error) {
-	var pageSizes []string
-	sizeList := []string{"B", "KB", "MB", "GB", "TB", "PB"}
 	files, err := ioutil.ReadDir("/sys/kernel/mm/hugepages")
 	if err != nil {
-		return pageSizes, err
+		return []string{}, err
 	}
+	var fileNames []string
 	for _, st := range files {
-		nameArray := strings.Split(st.Name(), "-")
+		fileNames = append(fileNames, st.Name())
+	}
+	return getHugePageSizeFromFilenames(fileNames)
+}
+
+func getHugePageSizeFromFilenames(fileNames []string) ([]string, error) {
+	var pageSizes []string
+	sizeList := []string{"B", "KB", "MB", "GB", "TB", "PB"}
+	for _, fileName := range fileNames {
+		nameArray := strings.Split(fileName, "-")
 		pageSize, err := units.RAMInBytes(nameArray[1])
 		if err != nil {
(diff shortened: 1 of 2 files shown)


odinuge:
Looks like this is a result of a quick copy-paste of the `ByteSizes`, meant for making huge numbers more human readable, from the `docker/go-units` package. https://github.com/docker/go-units/blob/519db1ee28dcc9fd2474ae59fca29a810482bfb1/size.go#L37

mrunalp:
LGTM

odinuge:
Since the kernel only use "KB", "MB" and "GB", the most correct would be to only use them (including B, in order to make go-units work). For sizes larger than 1TB that would result in index-out-of-range, so that would also require a update of the `docker/go-units` (fixed here: https://github.com/docker/go-units/commit/a09cd47f892041a4fac473133d181f5aea6fa393).

What do you think @mrunalp?

Also, It would be nice having the list exported, so that other users (eg. kubernetes) can import it instead of copying it. :smile:

crosbymichael:
LGTM

odinuge:
Looks like you have to post LGTM again since I changed the PR @mrunalp. Would be nice to get this into master, so that we can fix the issue in k8s. :smile:

mrunalp:
LGTM

PR #2067, opened 2019-06-04
libcontainer: change seccomp test for clone syscall

KentaTada:
This commit changes the value of seccomp test for clone syscall.
Also hardcoded values should be changed because it is unclear to
understand what flags are tested.

Related issues:

* https://github.com/containerd/containerd/pull/3314
* https://github.com/moby/moby/pull/39308
* https://github.com/opencontainers/runtime-tools/pull/694

### DIFF
--- a/libcontainer/specconv/spec_linux_test.go
+++ b/libcontainer/specconv/spec_linux_test.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"testing"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/configs/validate"
 	"github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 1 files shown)


libcontainer/specconv/spec_linux_test.go
```diff
@@ -154,7 +156,7 @@ func TestSetupSeccomp(t *testing.T) {
 			expectedCloneSyscallArgs := configs.Arg{
 				Index:    0,
 				Op:       7, // SCMP_CMP_MASKED_EQ
-				Value:    2080505856,
+				Value:    2114060288,
```
cyphar: Why not change this to the `unix.*` values you have listed above?

KentaTada: @cyphar  There is no meaning in that. I should use unix.* value as you said.

KentaTada:
@cyphar I've updated expectedCloneSyscallArgs. Thanks!

crosbymichael:
LGTM

cyphar:
LGTM.

PR #2071, opened 2019-06-11
Allow to define `COMMIT` by env

judu:
Some package managers download the archive instead of cloning the git repo.
When they do that, the call to git fails.

This commit allows package managers to provide the COMMIT value via environment.


cyphar:
LGTM I guess, though it should be noted that we use `COMMIT_NO` for this in openSUSE.

crosbymichael:
LGTM

PR #2072, opened 2019-06-12
libcontainer: fix the way to get network stats

ZeroMagic:
Although `getNetworkInterfaceStats(interfaceName string)` was implemented in runc,  containerd does not transfer the interface name to runc. So we still can’t get the network stats from libcontainer’s Stats().

Because in a runc container, we can get network stats by `cat /proc/<init pid>/root/sys/class/net/<EthInterface>/statistics/rx_bytes`. So I leverage this point. 

With this PR, `runc events --stats <containerd-id>` can return network stats.


kolyshkin:
Perhaps we can remove the NIC stats code instead?

PR #2029, opened 2019-03-28
Update dependencies

thaJeztah:
Thought I'd do a round of updating dependencies, and where possible use tagged versions

Also formatting the vendor.conf file, and pin all deps by git-sha, to make it better readable, and to encourage pinning by sha, but align to a tagged release.

- https://github.com/pkg/errors/compare/v0.8.0...v0.8.1
- https://github.com/containerd/console/compare/2748ece16665b45a47f884001d5831ec79703880...0650fd9eeb50bab4fc99dceb9f2e14cf58f36e7f
  - containerd/console#27 console_linux: Fix race: lock Cond before Signal
- https://github.com/docker/go-units/compare/v0.2.0...v0.3.3
  - docker/go-units#8 Enhance FromHumanSize to parse float64 string
  - docker/go-units#20 Add `HumanSizeWithPrecision` function
- https://github.com/urfave/cli/compare/d53eb991652b1d438abdd34ce4bfa3ef1539108e...v1.20.0
  - previous version was somewhere between v1.18 and v1.19
- https://github.com/syndtr/gocapability/compare/db04d3cc01c8b54962a58ec7e491717d06cfcc16...d98352740cb2c55f81556b63d4a1ec64c5a319c2
  - syndtr/gocapability#14 capability: Deprecate NewPid and NewFile for NewPid2 and NewFile2
  - syndtr/gocapability#16 Fix capHeader.pid type
- https://github.com/mrunalp/fileutils/compare/ed869b029674c0e9ce4c0dfa781405c2d9946d08...7d4729fb36185a7c1719923406c9d40e54fb93c7
  - no significant changes, other than some linting fixes
- https://github.com/coreos/go-systemd/compare/v14..v19
  - coreos/go-systemd#248 dbus: add SetPropertiesSubscriber method
  - coreos/go-systemd#251 activation: add support for listeners with names
  - coreos/go-systemd#296 dbus: Fix API break from godbus
- https://github.com/godbus/dbus/compare/v3..v5.0.1
  - godbus/dbus#89 introduce MakeVariantWithSignature
- https://github.com/golang/protobuf/compare/18c9bb3261723cd5401db4d0c9fbc5c3b6c70fe8...v1.0.0
  - picking the first tagged release (newer versions exist, but I have bad experiences with bumping, lol)

PR #2073, opened 2019-06-18
Add support for Go Modules

odinuge:
This removes vndr, and swiches to native Go Modules instead. All modules
are kept on the old version.

Keeps the vendor/ dir, so everything is backwards compatible.

This is more or less only `go mod init`

This would make it way easier for other projects using runc as dependency to keep track of the dependencies (and their version) that runc require.

It would be very nice to get this in before the `v1.0.0` release!


vrothberg:
Just a drive-by-comment (I am no maintainer). To make sure that the code and the `./vendor` directory are in sync, we are using the following checks in the [github.com/containers](https://github.com/containers) projects:

### Makefile target
```Makefile
vendor:
	export GO111MODULE=on \
		$(GO) mod tidy && \
		$(GO) mod vendor && \
                $(GO) mod verify
```

### Shell script `tree_status.sh`
Checks if the git tree has changed and prints a report.

```bash
#!/usr/bin/env bash
set -e

STATUS=$(git status --porcelain)
if [[ -z $STATUS ]]
then
	echo "tree is clean"
else
	echo "tree is dirty, please commit all changes"
	echo ""
	echo "$STATUS"
	exit 1
fi
```

### CI (e.g., Travis)
Both are called in a CI job (`make vendor && ./hack/tree_status.sh`) to avoid inconsistencies from being committed.

thaJeztah:
similar discussion about validation in https://github.com/docker/distribution/pull/2941#issuecomment-503686124 (haven't checked if one has advantages over the other)

vrothberg:
> similar discussion about validation in [docker/distribution#2941 (comment)](https://github.com/docker/distribution/pull/2941#issuecomment-503686124) (haven't checked if one has advantages over the other)

Awesome!

vrothberg:
> Hi @vrothberg! Thanks for the tips! k8s use the same too, but with a bit more `hack`-files. smile
> 
> Added a simple check in [4f6014b](https://github.com/opencontainers/runc/commit/4f6014b4607eb06d3a807afafd34f949e5911cdf) now.

LGTM, thanks!

AkihiroSuda:
needs rebase

odinuge:
Thanks for the interest @nathanjsweet!

> do you need help to get this going?
I think this should be good to go now (tests fail because of some problems installing postgres in CI), but I am absolutely open for suggestions. I have stopped rebasing the PR when conflicts occur in await for some interest from the contributors/reviewers from OCI. Hopefully some of them will look at it, so we can get go modules into the project.

> Definitely bother the devs on the slack channel to get this across the finish line. To any of the OCI devs: it's harder to contribute to, use, etc this repository without go modules.

Agree! Have been using this in several projects myself, including k8s, and go modules (and a new release) will help the developer experience a lot!

odinuge:
I also think #2029 is kinda important, and that PR should be merged before this one.

odinuge:
ping @mrunalp

saschagrunert:
@odinuge can you please rebase on top of the latest master branch? Then I would give this one a review if you don't mind. :)

Makefile
```diff
@@ -118,6 +118,17 @@ validate:
 
 ci: validate test release
 
+vendor:
+	export GO111MODULE=on \
+	$(GO) mod tidy && \
+	$(GO) mod vendor && \
+	$(GO) mod verify
+
+vendor-verify: vendor
```
AkihiroSuda: this should be put to the `validate` target?

odinuge: Yes! Thanks for catching it. Updated the PR now.

odinuge:
There are some differences between the files added into the `vendor` dir from `golang.org/x/sys` between `go>=1.13` and `go<=1.12`. Should we run with go 1.13 in CI by default, or should we just skip vendoring all the deps?

vrothberg:
Go 1.13 became smarted in detecting which files are actually needed, so the CI must use exactly one version for the vendor checks. For https://github.com/containers/libpod and similar projects, we decided to use go 1.13 only which also impacts contributions to the code, so we added a `make vendor-in-container` target which does the vendoring in a `golang:1.13` container and is meant to be used on hosts that don't have go 1.13 installed.

.travis.yml
```diff
@@ -30,8 +30,8 @@ env:
 before_install:
```
saschagrunert: Please add go 1.13.x to the matrix as well.

odinuge: Yes, I think we can do it explicitly too.

The `tip` version is 1.13, but it currently fails because of the vendoring issue. Ref. comment from @vrothberg I think we should just move the vendoring check to `1.13`, since that is what most users use anyways.

saschagrunert: Yes, in the go.mod it always references 1.13 so I think this makes sense.

AkihiroSuda:
needs rebase

odinuge:
Hi @AkihiroSuda,

This will need a rebase every time a dep is changed, so ill rebase as soon as I get some feedback about it from the maintainers. :smile: Been open for 6-7 months now, so I am not sure if they want this merged or not. :/

AkihiroSuda:
LGTM (needs rebase)

odinuge:
> LGTM (needs rebase)

Thanks!! Should be good to go now @AkihiroSuda! Hopefully there will be no new dependency changes before we can get this into master. :smile:

.travis.yml
```diff
@@ -3,11 +3,21 @@ language: go
 go:
   - 1.11.x
   - 1.12.x
+  - 1.13.x
+  - 1.14.x
```
AkihiroSuda: can be a separate PR

odinuge: I guess I can do that, but there are some changes in the way go modules does vendoring, making tests fail for v1.13 and v1.14 if running `go mod vendor` in v1.12, and tests failing in v1.14 if running `go mod vendor` with v1.13.

https://golang.org/doc/go1.14#go-command

Do you still think I should split it?

AkihiroSuda: Please split. We may need some additional work for Go 1.14, because it changed the scheduler and EINTR stuff: https://golang.org/doc/go1.14#runtime

odinuge: Ahh, thant makes sense, thanks! Going back to v1.13 then.

vrothberg: We faced the very same issue and provide a `vendor-in-container` target to force _one_ version of go for vendoring (see https://github.com/containers/libpod/blob/master/Makefile#L633).

AkihiroSuda:
LGTM

hqhq:
LGTM

Use go modules makes sence to me, but I'm not sure why we are not using it in the first place, I think more maintainers should be aware of this change before merge. 
@crosbymichael @mrunalp @cyphar

h-vetinari:
> @hqhq: [...] but I'm not sure why we are not using it in the first place [...]

Go modules became mature with Go 1.13 (released Sept '19), and this PR has been open since June 19. Together with the usual turnaround time for reviews here, that already seems like more than enough explanation to me... 😅

hqhq:
> > @hqhq: [...] but I'm not sure why we are not using it in the first place [...]
> 
> Go modules became mature with Go 1.13 (released Sept '19), and this PR has been open since June 19. Together with the usual turnaround time for reviews here, that already seems like more than enough explanation to me... 😅

I see, thanks for explain. BTW does this mean we are not able to build runc without go 1.13 or newer?

odinuge:
> I see, thanks for explain. BTW does this mean we are not able to build runc without go 1.13 or newer?

We still keep the `vendor/` directory with all the dependencies, so it will still work the same for pre go 1.13.

h-vetinari:
Another conflict has arisen, unfortunately (due to merging #2222).

odinuge:
> Another conflict has arisen, unfortunately (due to merging #2222).

Ahh. Updated the PR now.

Since #2222 started building with go 1.14, and require it to pass, I switched back to running the vendoring in 1.14. If we vendor with go 1.13, go build will fail on go 1.14, making the tests fil.

AkihiroSuda:
LGTM

AkihiroSuda:
ping @opencontainers/runc-maintainers PTAL

AkihiroSuda:
I'll merge this on Wednesday unless there are any objections

cyphar:
LGTM. I do have some hesitation about some of the administrative aspects of Go modules (I've had issues with updating individual dependencies before, as well as problems with `go mod vendor` stripping out some source files that Debian needed) but we should still switch.

PR #2078, opened 2019-07-02
Rename Id ID

tomocy:
Both of Id and ID are used in the source code as the short for "identifier" and ID is preferred in Go(https://github.com/golang/go/wiki/CodeReviewComments#initialisms). Therefore, I have renamed Id ID.

### DIFF
--- a/vendor/github.com/vishvananda/netlink/nl/xfrm_state_linux.go
+++ b/vendor/github.com/vishvananda/netlink/nl/xfrm_state_linux.go
@@ -92,7 +92,7 @@ func (msg *XfrmStats) Serialize() []byte {
 
 type XfrmUsersaInfo struct {
 	Sel          XfrmSelector
-	Id           XfrmId
+	ID           XfrmId
 	Saddr        XfrmAddress
 	Lft          XfrmLifetimeCfg
 	Curlft       XfrmLifetimeCur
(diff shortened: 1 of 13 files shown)


vendor/github.com/vishvananda/netlink/nl/xfrm_state_linux.go
```diff
@@ -92,7 +92,7 @@ func (msg *XfrmStats) Serialize() []byte {
 
 type XfrmUsersaInfo struct {
 	Sel          XfrmSelector
-	Id           XfrmId
+	ID           XfrmId
```
thaJeztah: This is vendored code from a dependency, so should not be modified here (instead a pull request should be opened for that upstream in https://github.com/vishvananda/netlink/

vendor/github.com/vishvananda/netlink/xfrm_state_linux.go
```diff
@@ -45,11 +45,11 @@ func XfrmStateAdd(state *XfrmState) error {
 
 	msg := &nl.XfrmUsersaInfo{}
 	msg.Family = uint16(nl.GetIPFamily(state.Dst))
-	msg.Id.Daddr.FromIP(state.Dst)
+	msg.ID.Daddr.FromIP(state.Dst)
```
thaJeztah: Same for this file

AkihiroSuda:
@cyphar @mrunalp is this closable?

AkihiroSuda:
Closing because of inactivity

PR #2080, opened 2019-07-08
Update busybox source and fix runc exec bug

zhlhahaha:
Currently, the id verification in integration test failed on arm
platform due to the inconsistent /etc/group in the busybox images
for arm and x86. To be specific, the nogroup id in x86 is 99 while
that in arm is 65534.

99 is old id for nogroup, and no longer be used in recent system,
so sync the busybox image for arm and x86 to the image in busybox
github. Also change the id verification rule in integration test.

Signed-off-by: Howard Zhang <howard.zhang@arm.com>

### DIFF
--- a/tests/integration/exec.bats
+++ b/tests/integration/exec.bats
@@ -122,10 +122,10 @@ function teardown() {
 
   wait_for_container 15 1 test_busybox
 
-  runc exec --user 1000:1000 --additional-gids 100 --additional-gids 99 test_busybox id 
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

zhlhahaha:
ping @avagin @caniszczyk @cyphar @dqminh

caniszczyk:

hqhq:
LGTM

PR #2085, opened 2019-07-11
Make sure signalAllProcesses is invoked in the function of destroy when container shares pid namespace

keloyang:
# issue description
./runc spec
add pid ns config to config.json
```
{
    "type": "pid",
    "path": "/proc/1/ns/pid"
}
```

"./runc run hello" in a console and "./runc exec -t hello ash" in another console, and then exit container.
we will see the following erros,
```
/ # exit
ERRO[0036] Failed to remove paths: map[memory:/sys/fs/cgroup/memory/user.slice/hello blkio:/sys/fs/cgroup/blkio/user.slice/hello hugetlb:/sys/fs/cgroup/hugetlb/hello net_cls:/sys/fs/cgroup/net_cls,net_prio/hello freezer:/sys/fs/cgroup/freezer/hello cpuset:/sys/fs/cgroup/cpuset/hello cpu:/sys/fs/cgroup/cpu,cpuacct/user.slice/hello cpuacct:/sys/fs/cgroup/cpu,cpuacct/user.slice/hello pids:/sys/fs/cgroup/pids/user.slice/hello net_prio:/sys/fs/cgroup/net_cls,net_prio/hello perf_event:/sys/fs/cgroup/perf_event/hello name=systemd:/sys/fs/cgroup/systemd/user.slice/user-0.slice/session-2658.scope/hello devices:/sys/fs/cgroup/devices/user.slice/hello] 
```

# why
It's expect that signalAllProcesses is invoked when container shares pid namespace. share pid ns contains the following conditions:
```
{
    //no specify pid ns
}
{
    "type": "pid",
    "path": "/proc/${num}/ns/pid"
}
```
but the code can't do the second condition. this pr will add the second.

Signed-off-by: Shukui Yang <keloyangsk@gmail.com>

### DIFF
--- a/libcontainer/state_linux.go
+++ b/libcontainer/state_linux.go
@@ -38,7 +38,8 @@ type containerState interface {
 }
 
 func destroy(c *linuxContainer) error {
-	if !c.config.Namespaces.Contains(configs.NEWPID) {
+	if !c.config.Namespaces.Contains(configs.NEWPID) ||
+		c.config.Namespaces.PathOf(configs.NEWPID) != "" {
 		if err := signalAllProcesses(c.cgroupManager, unix.SIGKILL); err != nil {
 			logrus.Warn(err)
 		}


keloyang:
ping @calavera @fabiokung @vdemeester @albertoleal

libcontainer/state_linux.go
```diff
@@ -38,7 +38,8 @@ type containerState interface {
 }
 
 func destroy(c *linuxContainer) error {
-	if !c.config.Namespaces.Contains(configs.NEWPID) {
+	if !c.config.Namespaces.Contains(configs.NEWPID) ||
+		c.config.Namespaces.PathOf(configs.NEWPID) != "" {
```
mrunalp: Killing PID 1 in a PID namespace kills the other processes in the PID namespace.

keloyang:
@mrunalp thanks very much for your reply.

runc stop the processes in a container use two methed:
- container has a new pid ns, pid 1 process exit, and other processes in this pid ns will be killed.
- share pid ns, no pid 1 process exit, runc need find the other processes from container's cgroup, and kill them.

take the following ns config for example, and assume the process ${num} is a pid 1 process in a pid ns.
```
{
    "type": "pid",
    "path": "/proc/${num}/ns/pid"
}
```
if a container share "/proc/${num}/ns/pid" pid ns，when it exit，the process ${num} don't exit，no pid 1 process exit ,so the other processes in this container selected for its cgroup will not exit.

libcontainer/state_linux.go
keloyang: signalAllProcesses only kill the processes found from the container's cgroup, when a container share pid ns, it will not kill the pid 1 process in this pid ns.

hqhq:
I agree we should kill all processes when container exit while it's sharing pidns with other container. But why it's not handled in https://github.com/opencontainers/runc/blob/v1.0.0-rc8/signals.go#L99 ? We did call `signalAllProcesses` in wait method of initProcess https://github.com/opencontainers/runc/blob/v1.0.0-rc8/libcontainer/process_linux.go#L454

keloyang:
> why it's not handled 

If the child process has exited before invoking cmd.Wait,  cmd.Wait will return a error , the error is "waitid: no child processes". process.Wait() is involed after runc recv SIGCHLD, at that time the child process  is exited .

we can print err after cmd.Wait like the following, it will also resolve the issue:
```
 func (p *initProcess) wait() (*os.ProcessState, error) {
        err := p.cmd.Wait()
-       if err != nil {
+       fmt.Printf("%s\n", err.Error())
+       if err != nil &&
+               !strings.Contains(err.Error(), "no child processes") {
                return p.cmd.ProcessState, err
        }
```

now run runc again,
```
[root@centos0 container]# ./runc run hello
/ # exit            //this is comment, we need run "runc exec -t hello ash" in another console
waitid: no child processes
[root@centos0 container]# 

```

Docker is also involved in this issue，
```
[root@centos1 ~]# docker version
 Version:           19.03.9
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        9d988398e7
 Built:             Fri May 15 00:22:47 2020
 Experimental:      false

 Engine:
  Version:          19.03.9
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       9d988398e7
  Built:            Fri May 15 00:28:17 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

docker run --rm -ti --name test --pid container:test-1 busybox ash
/ # exit   

//  in another console run "docker exec -ti test ash", and the console will be hung after the container of test exited.
[root@centos1 ~]#  docker exec -ti test ash
/ #
```

ping @hqhq @crosbymichael @mrunalp

keloyang:

crosbymichael:
LGTM

I think this is safe to do as we still get this container's processes via the cgroup and it really has nothing to do with the pid namespace.

hqhq:
LGTM

keloyang:
ping @hqhq @crosbymichael All checks have passed

hqhq (approved):
LGTM

keloyang:
@hqhq thank you.
ping @opencontainers/runc-maintainers All checks have passed, PTAL, thank.

keloyang:
ping @crosbymichael PTAL, thanks.

crosbymichael (approved):
LGTM

PR #2086, opened 2019-07-17
fix permission denied

win-t:
when exec as root and `config.Cwd` is not owned by root, exec will fail
because root doesn't have the caps.

So, Chdir should be done before setting the caps.

to reproduce this error
1. create Dockerfile
```
FROM alpine:latest

RUN mkdir /lala && chown nobody:nobody /lala && chmod 0700 /lala
USER nobody
WORKDIR /lala

ENTRYPOINT ["sh", "-c", "mkfifo lala; cat lala"]
```

2. build and run the container
```
docker build -t asdf .
docker run -it --rm --name hai asdf
```

3. docker exec from another terminal
```
docker exec -it -u 0:0 hai sh
```

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -127,6 +127,12 @@ func finalizeNamespace(config *initConfig) error {
 		return errors.Wrap(err, "close exec fds")
 	}
 
+	if config.Cwd != "" {
+		if err := unix.Chdir(config.Cwd); err != nil {
+			return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
+		}
+	}
+
 	capabilities := &configs.Capabilities{}
 	if config.Capabilities != nil {
 		capabilities = config.Capabilities
@@ -153,11 +159,6 @@ func finalizeNamespace(config *initConfig) error {
 	}
 	if err := w.ApplyCaps(); err != nil {
 		return errors.Wrap(err, "apply caps")
-	}
-	if config.Cwd != "" {
-		if err := unix.Chdir(config.Cwd); err != nil {
-			return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
-		}
 	}
 	return nil
 }


libcontainer/init_linux.go
```diff
@@ -151,14 +151,14 @@ func finalizeNamespace(config *initConfig) error {
 	if err := system.ClearKeepCaps(); err != nil {
 		return errors.Wrap(err, "clear keep caps")
 	}
-	if err := w.ApplyCaps(); err != nil {
-		return errors.Wrap(err, "apply caps")
-	}
 	if config.Cwd != "" {
```
cyphar: You could move this before all of the capability and `setupUser` code (after `CloseExecFrom`).

win-t: okay

crosbymichael:
LGTM

siscia:
ping!

cyphar:
LGTM.

Random-Liu:
Does this introduce a regression?
See https://github.com/containerd/cri/pull/1397#issuecomment-586141463 and https://github.com/moby/moby/pull/36587

justincormack:
I believe this is incorrect as @Random-Liu suggests; I am unsure why you accepted a PR with a repro just from `docker exec` not `runc` as `docker exec` has some interesting properties in that it inherits some properties from the original run.

win-t:
@justincormack I can reproduce this just with old version of runc
it is just more easy with docker

cyphar:
I can see the argument for this new behaviour being wrong (in theory you could bypass some DAC controls in very odd setups by setting your working directory to be inside a path which is world-`rx` but has a parent which the use doesn't have `rx` permissions on.

I'm okay with reverting this, but I don't really feel that either behaviour is necessarily wrong. After all, we've always said that being able to run custom container configurations is equivalent to running as root.

kolyshkin:
At least we should not merge PRs with no runc repro / integration test case.

This was some arbitrary change that fixed one workflow and broke another (https://github.com/opencontainers/runc/pull/2685), and there is no test case to see if the fix from #2685 results in any regressions :(

kolyshkin:
OK here is the test case which fails before and passes after this PR.

```bats
@test "runc exec --user with no access to cwd" {
        requires root

        chown 42 rootfs/root
        chmod 700 rootfs/root

        update_config '   .process.cwd = "/root"
                        | .process.user.uid = 42
                        | .process.args |= ["sleep", "1h"]'

        [ "$status" -eq 0 ]

        runc exec --user 0 test_busybox true
        [ "$status" -eq 0 ]
}
```

PR #2087, opened 2019-07-25
cgroups/systemd: add cgroup-v2 path to the list when using hybrid mode

mauriciovasquezbernal:
Currently, the parent process of the container is moved to the right cgroup-v2 tree when systemd is using a hybrid model (last line with 0::):

```
$ runc --systemd-cgroup run myid
/ # cat /proc/self/cgroup
12:cpuset:/system.slice/runc-myid.scope
11:blkio:/system.slice/runc-myid.scope
10:devices:/system.slice/runc-myid.scope
9:hugetlb:/system.slice/runc-myid.scope
8:memory:/system.slice/runc-myid.scope
7:rdma:/
6:perf_event:/system.slice/runc-myid.scope
5:net_cls,net_prio:/system.slice/runc-myid.scope
4:freezer:/system.slice/runc-myid.scope
3:pids:/system.slice/runc-myid.scope
2:cpu,cpuacct:/system.slice/runc-myid.scope
1:name=systemd:/system.slice/runc-myid.scope
0::/system.slice/runc-myid.scope
```

However, if a second process is executed in the same container, it is not moved to the right cgroup-v2 tree:

```
$ runc exec myid /bin/sh -c 'cat /proc/self/cgroup'
12:cpuset:/system.slice/runc-myid.scope
11:blkio:/system.slice/runc-myid.scope
10:devices:/system.slice/runc-myid.scope
9:hugetlb:/system.slice/runc-myid.scope
8:memory:/system.slice/runc-myid.scope
7:rdma:/
6:perf_event:/system.slice/runc-myid.scope
5:net_cls,net_prio:/system.slice/runc-myid.scope
4:freezer:/system.slice/runc-myid.scope
3:pids:/system.slice/runc-myid.scope
2:cpu,cpuacct:/system.slice/runc-myid.scope
1:name=systemd:/system.slice/runc-myid.scope
0::/user.slice/user-1000.slice/session-8.scope
```

Having the processes of the container in its own cgroup-v2 is useful for any BPF programs that rely on bpf_get_current_cgroup_id(), like https://github.com/kinvolk/inspektor-gadget/ for instance.

This commit makes that processes executed with exec are placed into the right cgroup-v2 tree.  The implementation checks if systemd is using a hybrid mode (by checking if cgroups-v2 is mounted in /sys/fs/cgroup/unified), if yes, the path of the cgroup-v2 slice for this container is saved into the cgroup path list.


libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -69,6 +70,7 @@ var subsystems = subsystemSet{
 const (
 	testScopeWait = 4
 	testSliceWait = 4
+	cgroupUnified = "/sys/fs/cgroup/unified"
```
crosbymichael: Shouldn't we will look for the cgroups mountpoint and append `unified` to it in cause cgroups is not mounted under `/sys/fs`?

mauriciovasquezbernal: Please notice that this is `systemd` specific, in hybrid mode the cgroup-v2 is mounted on `/sys/fs/cgroup/unified` as stated here: https://systemd.io/CGROUP_DELEGATION.html#three-different-tree-setups-.

mauriciovasquezbernal:
Hello folks, any thoughts about this one?

Thanks!

/cc @filbranden, @crosbymichael @giuseppe @cyphar

dongsupark:
@cyphar 
It's what I mentioned during the talk in the hallway.

libcontainer/cgroups/systemd/apply_systemd.go
cyphar: Not to mention that these days, `/dev/cgroup` is dead and everyone mounts it to `/sys/fs/cgroup`.

AkihiroSuda: I think this variable should be named `cgroupHybridUnified` to avoid confusion

mauriciovasquezbernal: @AkihiroSuda I just renamed the variable. Do you have other feedback about this PR? Thanks.

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -430,10 +445,32 @@ func ExpandSlice(slice string) (string, error) {
 	return path, nil
 }
 
+// Check if cgroups-v2 is mounted on /sys/fs/cgroup/unified/ as stated in
+// https://systemd.io/CGROUP_DELEGATION.html#three-different-tree-setups-
+// Only the hybrid mode is taken into consideration as unified mode is not
+// supported in runc yet
+func isCgroupsv2Available() bool {
```
AkihiroSuda: `isHybridMode()` to avoid confusion with pure-v2 mode

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -321,6 +323,19 @@ func (m *Manager) Apply(pid int) error {
...
+	// If systemd is using cgroups-v2 add the slice path of this container to
+	// the paths so following process executed with exec join that cgroup-v2
+	// as well
+	if isCgroupsv2Available() {
+		// "" means cgroup-v2 path
+		cgroupsv2Path, err := getSubsystemPath(m.Cgroups, "")
+		if err != nil && cgroups.IsNotFound(err) {
+			return err
+		}
+		paths["cgroups-v2"] = cgroupsv2Path
```
AkihiroSuda: "hybrid" to avoid confusion with pure-v2 mode (RFC)

mauriciovasquezbernal:
@AkihiroSuda I changed the names according to your suggestion.

AkihiroSuda:
Please squash commits and sign

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -321,6 +323,19 @@ func (m *Manager) Apply(pid int) error {
 		}
 		paths[s.Name()] = subsystemPath
 	}
+
+	// If systemd is using cgroups-hybrid mode add the slice path of this
+	// container to the paths so following process executed with exec join
+	// that cgroup as well
+	if isHybridMode() {
+		// "" means cgroup-hybrid path
```
AkihiroSuda: L336  can be `paths[""] = ...` as well?

mauriciovasquezbernal: Yes, it can be. Updated.

AkihiroSuda:
LGTM

AkihiroSuda:

AkihiroSuda:
@mrunalp @kolyshkin PTAL

kolyshkin:
Should we ever try to support hybrid mode? If yes, it should at least be tested. Currently, we barely test v2.

dongsupark:
FYI, I have just realized that this PR needs a rebase.
Apparently code around cgroups/systemd has changed recently with v1.0.0-rc91.

kolyshkin:
I think we don't want to support this two-headed beast of hybrid mode, and recommend using cgroup v2 unified instead. IOW close this.

cyphar:
Yeah supporting hybrid mode will only lead to madness. cgroupv2 now supports all of the major controllers so hybrid mode doesn't really make sense anymore.

AkihiroSuda:
I agree that hybrid mode is meaningless for controllers like `cpu` and `memory`, but it is still useful for eBPF stuff. 
So I think it wouldn't hurt to merge this if somebody can rebase this. 

Anyway, we will need to close this if this remains unrebased.

mauriciovasquezbernal:
@AkihiroSuda @kolyshkin @cyphar I rebased this.

cyphar:
@kolyshkin Does this change make sense given all of the changes you made last year to how subsystem paths are handled in cgroupfs?

libcontainer/cgroups/v1_utils.go
```diff
@@ -116,6 +116,12 @@ func FindCgroupMountpoint(cgroupPath, subsystem string) (string, error) {
 		return "", errUnified
 	}
 
+	// If subsystem is empty it means that we are looking for the
+	// cgroups2 path
```
kolyshkin, line 120: I'd say "cgroup2 hybrid path" instead.

kolyshkin:
Looks like it still makes sense. Two things
 - fs cgroup driver is not fixed wrt hybrid mode;
 - I don't like the hack of using empty subsystem to denote the hybrid path much. Perhaps it can be better implemented in fs driver by adding `NameGroup{Name: "hybrid", Join: yes"}` to the list of subsystems, and modifying other places to support a special value of "hybrid".
 
 Finally, if we're doing it, we need a test case (apparently ubuntu 20.04 which is used on GHA is a good candidate for that).

kolyshkin:
(CI failure on centos7 is unrelated; it's https://github.com/opencontainers/runc/issues/2760)

AkihiroSuda:
> I don't like the hack of using empty subsystem to denote the hybrid path much. Perhaps it can be better implemented in fs driver by adding NameGroup{Name: "hybrid", Join: yes"} to the list of subsystems, and modifying other places to support a special value of "hybrid".

The empty string corresponds to /proc/PID/cgroup . Also, v1 named group has nothing to do with the hybrid hierarchy.

kolyshkin:
> The empty string corresponds to /proc/PID/cgroup

Yes, I understand that it comes from there.

> Also, v1 named group has nothing to do with the hybrid hierarchy.

... and yet we put the path to it into a map of paths for v1 controllers.

Anyway, I'm fine with the current approach -- it's a hack but at least it is documented in the code.

The other two issues remain:
 - fs cgroup driver is not fixed wrt hybrid mode;
 - need a simple test case.

mauriciovasquezbernal:
@kolyshkin I added a test and also implemented for the fs controller. I agree that using an empty string is not the cleanest solution, I tried to implement it in another way but I hit some issues. Unfortunately I don't have that much time to continue investigating other ways to implement this.

kolyshkin:
@mauriciovasquezbernal I am carrying this one in #3059 but this one can still be merged first (once rebased).

Note that the last patch needs to put `cgroup_hybrid` before `cgroup_*` (added by commit 44fcbfd68).

kolyshkin:
> @mauriciovasquezbernal I am carrying this one in #3059 but this one can still be merged first (once rebased).

Well, I'm actually wrong. This needs criu-3.16 which is still not released.

PR #2089, opened 2019-07-29
doc: First process in container needs `Init: true`

anx-astocker:
`Init` on the `Process` struct specifies whether the process is the first process in the container. This needs to be set to `true` when running the container.

### DIFF
--- a/libcontainer/README.md
+++ b/libcontainer/README.md
@@ -261,6 +261,7 @@ process := &libcontainer.Process{
 	Stdin:  os.Stdin,
 	Stdout: os.Stdout,
 	Stderr: os.Stderr,
+	Init:   true,
 }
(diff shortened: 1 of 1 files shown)


ChristianKreuzberger:
I can confirm this issue. On Ubuntu 18.04, when you don't set `init: true`, the process will be spawned outside of the container (or the container is somehow the parent system).

The integration test also uses `init: true` to spawn a new process:
https://github.com/opencontainers/runc/blob/6994ff2742f840d1bd11b741d4dac5a2a81f95c9/libcontainer/integration/utils_test.go#L160-L169

crosbymichael:
LGTM

mrunalp:
LGTM

PR #2098, opened 2019-08-06
man: fix man-pages

adrianreber:
The man-pages are using pre-formatted section to display the options for
all commands. The result on my system never looked correct:
```
OPTIONS
       --bundle value, -b value  path to the root [...]
          --console-socket value    path to an AF_UNIX [...]
```
The first line was always indented less than the other lines.

This commit makes the option block a pre-formatted block (as intended???) by
using 4 spaces instead of 3 spaces.

In addition the man-pages did not specify their name and section
correctly. This adds something like `% runc-run "8"` to all man-pages to
have correct title `runc-run(8)` instead of 'NAME()' and it also adds
the section to the title: 'System Manager's Manual'.

This also fixes the use of `>` and `<` at multiple places. The markdown
source files were using "`<container-id>`" and similar which was (most of
the time) rendered as `""`. On some systems it was rendered correctly.

### DIFF
--- a/man/runc-checkpoint.8.md
+++ b/man/runc-checkpoint.8.md
@@ -1,28 +1,30 @@
+% runc-checkpoint "8"
+
 # NAME
    runc checkpoint - checkpoint a running container
 
 # SYNOPSIS
-   runc checkpoint [command options] <container-id>
+   runc checkpoint [command options] `<container-id>`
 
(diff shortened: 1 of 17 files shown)


giuseppe:
LGTM

crosbymichael:
LGTM

mrunalp:
LGTM

PR #2099, opened 2019-08-08
 update cgroupfs fields checking logic of cpuacct

hzy:
`getCpuUsageBreakdown` function needs user field and system field, so we just need to check if they exist.
Judging by fields length may cause failure on some unusual linux distributions.

Signed-off-by: Zhiyuan Hong <hzylovelyl@gmail.com>

### DIFF
--- a/libcontainer/cgroups/fs/cpuacct.go
+++ b/libcontainer/cgroups/fs/cpuacct.go
@@ -84,21 +84,36 @@ func getCpuUsageBreakdown(path string) (uint64, uint64, error) {
 	if err != nil {
 		return 0, 0, err
 	}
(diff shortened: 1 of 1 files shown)


cyphar:
You don't need to open a new PR when you push new commits -- force-pushing will update the PR. Also please squash your commits (you can do this with `git rebase -i` and then changing the `go fmt last commit` line in the editor from `pick` to `fixup`).

AkihiroSuda:
Needs rebase and adding tests

kolyshkin:
@farawaaay will you be able to work on this? Needs a rebase, a squash of commits, and maybe some tests.

PR #2106, opened 2019-08-15
Recursively thaw freezer subcgroups during SIGKILL

### LINKED ISSUE #2105: runC can't kill containers containing cgroup-freezer frozen processes
As I understand it, runC's procedure for killing a container is to

1. Freeze the container's cgroup
2. Issue SIGKILL to the process tree
3. Unfreeze the container's cgroup
4. Wait for the processes to terminate

If at the time of this procedure there is a non-empty freezer sub-cgroup with freezer.state==FROZEN, runC will block forever on step 4, because the tasks in that sub-cgroup are still frozen. 

(I believe I encountered this issue in practice when nesting runC. Specifically, attempting to kill a k8s container that itself contained a runC subprocess that was in the middle of killing.)

percontation:
Fixes #2105 

This PR is offered as an example of what I think would work. Unfortunately, it's kinda breaking some abstractions, and someone who knows the runC codebase could probably reorganize the code to better accommodate the additional code.

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -278,6 +278,21 @@ func (m *Manager) Freeze(state configs.FreezerState) error {
 	return nil
 }
 
+// ThawAll unfreezes the container's freezer cgroup, and all subcgroups
+// recursively.
+func (m *Manager) ThawAll() error {
+	err := m.Freeze(configs.Thawed)
+	if err != nil {
+		return err
+	}
+	paths := m.GetPaths()
+	freezer, err := subsystems.Get("freezer")
+	if err != nil {
+		return err
+	}
+	return freezer.(*FreezerGroup).RecursiveThaw(paths["freezer"])
+}
+
 func (m *Manager) GetPids() ([]int, error) {
 	paths := m.GetPaths()
 	return cgroups.GetPids(paths["devices"])
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -475,6 +475,22 @@ func (m *Manager) Freeze(state configs.FreezerState) error {
 	return nil
 }
 
+func (m *Manager) ThawAll() error {
+	err := m.Freeze(configs.Thawed)
+	if err != nil {
+		return err
+	}
+	path, err := getSubsystemPath(m.Cgroups, "freezer")
+	if err != nil {
+		return err
+	}
+	freezer, err := subsystems.Get("freezer")
+	if err != nil {
+		return err
+	}
+	return freezer.(*fs.FreezerGroup).RecursiveThaw(path)
+}
+
 func (m *Manager) GetPids() ([]int, error) {
 	path, err := getSubsystemPath(m.Cgroups, "devices")
 	if err != nil {
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -442,6 +442,21 @@ func getHugePageSizeFromFilenames(fileNames []string) ([]string, error) {
 	return pageSizes, nil
 }
 
+// WalkCgroups calls walkFn on the cgroup directory at path and on all its
+// subcgroups.
+func WalkCgroups(path string, walkFn func(path string) error) error {
+	return filepath.Walk(path, func(p string, info os.FileInfo, iErr error) error {
+		dir, file := filepath.Split(p)
+		if file != CgroupProcesses {
+			return nil
+		}
+		if iErr != nil {
+			return iErr
+		}
+		return walkFn(dir)
+	})
+}
+
 // GetPids returns all pids, that were added to cgroup at path.
 func GetPids(path string) ([]int, error) {
 	return readProcsFile(path)
@@ -451,15 +466,7 @@ func GetPids(path string) ([]int, error) {
 // subcgroups.
 func GetAllPids(path string) ([]int, error) {
 	var pids []int
-	// collect pids from all sub-cgroups
-	err := filepath.Walk(path, func(p string, info os.FileInfo, iErr error) error {
-		dir, file := filepath.Split(p)
-		if file != CgroupProcesses {
-			return nil
-		}
-		if iErr != nil {
-			return iErr
-		}
+	err := WalkCgroups(path, func(dir string) error {
 		cPids, err := readProcsFile(dir)
 		if err != nil {
 			return err
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -492,7 +492,13 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 			logrus.Warn(err)
 		}
 	}
-	if err := m.Freeze(configs.Thawed); err != nil {
+
+	if s != unix.SIGKILL {
+		err = m.Freeze(configs.Thawed)
+	} else {
+		err = m.ThawAll()
+	}
+	if err != nil {
 		logrus.Warn(err)
 	}
 
--- a/libcontainer/cgroups/cgroups.go
+++ b/libcontainer/cgroups/cgroups.go
@@ -24,6 +24,9 @@ type Manager interface {
 	// Toggles the freezer cgroup according with specified state
 	Freeze(state configs.FreezerState) error
 
+	// Unfreezes the cgroup and all subcgroups
+	ThawAll() error
+
 	// Destroys the cgroup set
 	Destroy() error
 
--- a/libcontainer/cgroups/fs/freezer.go
+++ b/libcontainer/cgroups/fs/freezer.go
@@ -26,23 +26,23 @@ func (s *FreezerGroup) Apply(d *cgroupData) error {
 	return nil
 }
 
-func (s *FreezerGroup) Set(path string, cgroup *configs.Cgroup) error {
-	switch cgroup.Resources.Freezer {
+func _set(path string, state configs.FreezerState) error {
+	switch state {
 	case configs.Frozen, configs.Thawed:
 		for {
 			// In case this loop does not exit because it doesn't get the expected
 			// state, let's write again this state, hoping it's going to be properly
 			// set this time. Otherwise, this loop could run infinitely, waiting for
 			// a state change that would never happen.
-			if err := writeFile(path, "freezer.state", string(cgroup.Resources.Freezer)); err != nil {
+			if err := writeFile(path, "freezer.state", string(state)); err != nil {
 				return err
 			}
 
-			state, err := readFile(path, "freezer.state")
+			current, err := readFile(path, "freezer.state")
 			if err != nil {
 				return err
 			}
-			if strings.TrimSpace(state) == string(cgroup.Resources.Freezer) {
+			if strings.TrimSpace(current) == string(state) {
 				break
 			}
 
@@ -51,10 +51,19 @@ func (s *FreezerGroup) Set(path string, cgroup *configs.Cgroup) error {
 	case configs.Undefined:
 		return nil
 	default:
-		return fmt.Errorf("Invalid argument '%s' to freezer.state", string(cgroup.Resources.Freezer))
+		return fmt.Errorf("Invalid argument '%s' to freezer.state", string(state))
 	}
-
 	return nil
+}
+
+func (s *FreezerGroup) Set(path string, cgroup *configs.Cgroup) error {
+	return _set(path, cgroup.Resources.Freezer)
+}
+
+func (s *FreezerGroup) RecursiveThaw(path string) error {
+	return cgroups.WalkCgroups(path, func(dir string) error {
+		return _set(path, configs.Thawed)
+	})
 }
 
 func (s *FreezerGroup) Remove(d *cgroupData) error {
--- a/libcontainer/cgroups/systemd/apply_nosystemd.go
+++ b/libcontainer/cgroups/systemd/apply_nosystemd.go
@@ -54,6 +54,10 @@ func (m *Manager) Freeze(state configs.FreezerState) error {
 	return fmt.Errorf("Systemd not supported")
 }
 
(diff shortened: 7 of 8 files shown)


percontation:
So, this change could be refactored to not add `ThawAll()` as a method everywhere, and instead use the normal `Set` mechanism with a new `configs.ThawedRecursive` enum.

I think this would be less invasive and still fix the bug? Feedback on what's preferred welcome.

mruck:
@mrunalp any feedback? Happy to refactor this but want to know your thoughts first.

mruck:
@hqhq any thoughts?

hqhq:
I don't think runc is quite robust for nested containers, but for this fix, the idea looks good to me, child container should not block the exit of father container.

mruck:
Thanks for the feedback @hqhq!  As @percontation mentioned above, instead of adding `ThawAll()` as a method everywhere, we could use the normal `Set` mechanism with a new `configs.ThawedRecursive enum`.  Or did you have something else in mind?  Just want to figure out how this can be improved so it can be merged in.

hqhq:
@percontation @mruck Current change looks good to me, just needs rebase, thanks.

mruck:
@hqhq rebase done, let me know if there's anything else!

hqhq:
LGTM

@mrunalp @crosbymichael Is it OK that we only add functionalities on legacy cgroup now?

percontation:
It looks like the new cgroups objects have had the corresponding changes made for them as well here. (Not totally sure since I'm unfamiliar with the new cgroups stuff.) @mruck can you confirm?

libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -34,6 +34,22 @@ var unifiedSubsystems = subsystemSet{
...
+	}
+	path, err := getSubsystemPath(m.Cgroups, "freezer")
+	if err != nil {
+		return err
+	}
+	freezer, err := unifiedSubsystems.Get("freezer")
+	if err != nil {
+		return err
+	}
+	return freezer.(*fs.FreezerGroup).RecursiveThaw(path)
```
percontation: This needs to be a cast to `FreezerGroupV2`? And then of course `FreezerGroupV2` also needs it's new `RecursiveThaw` method.

hqhq: Yes, you can use fs cgroup method in cgroupv2, even in systemd cgroup, it'll work, but that'll be a dependency, and it'll be broken if one day we remove legacy cgroup support (though that most likely won't happen).

percontation: I'm not sure I understand. My impression is that, if this line here is a bad cast because the underlying type will be a `FreezerGroupV2`. If we change this to a `FreezerGroupV2` cast, and then add a `RecursiveThaw` method to `FreezerGroupV2` (by analogy to the change to `FreezerGroup`), then the content of this MR will support cgroupv2 and will *not* be broken if cgroupv1 support is removed.

hqhq: That's correct.

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -427,6 +427,22 @@ func (m *LegacyManager) Freeze(state configs.FreezerState) error {
...
+	}
+	path, err := getSubsystemPath(m.Cgroups, "freezer")
+	if err != nil {
+		return err
+	}
+	freezer, err := legacySubsystems.Get("freezer")
+	if err != nil {
+		return err
+	}
+	return freezer.(*fs.FreezerGroup).RecursiveThaw(path)
```
hqhq: And you don't need cast here.

mruck:
Updated so the change applies to cgroupsv2

mruck:
@hqhq let me know if anything else needs to be updated.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -301,6 +301,21 @@ func (m *Manager) Freeze(state configs.FreezerState) error {
...
+	err := m.Freeze(configs.Thawed)
+	if err != nil {
+		return err
+	}
+	paths := m.GetPaths()
+	freezer, err := subsystemsLegacy.Get("freezer")
+	if err != nil {
+		return err
+	}
+	return freezer.(*FreezerGroup).RecursiveThaw(paths["freezer"])
```
hqhq: As now you have RecursiveThaw for cgroupv2, you can use
```go
freezer, err := m.getSubsystems().Get("freezer")
if err != nil {
    return err
}
return freezer.RecursiveThaw(paths["freezer"])
```

libcontainer/cgroups/systemd/apply_systemd.go
hqhq: You can call `freezer.RecursiveThaw` directly.

libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -34,6 +34,22 @@ var unifiedSubsystems = subsystemSet{
...
+	}
+	path, err := getSubsystemPath(m.Cgroups, "freezer")
+	if err != nil {
+		return err
+	}
+	freezer, err := unifiedSubsystems.Get("freezer")
+	if err != nil {
+		return err
+	}
+	return freezer.(*fs.FreezerGroupV2).RecursiveThaw(path)
```
hqhq: Same here.

libcontainer/cgroups/fs/apply_raw.go
mruck: Maybe I'm misunderstanding, but in order to call `RecursiveThaw` directly don't we need to add `RecursiveThaw` as a method on the `subsystem` interface?  Which in turn now every implementation of the `subsystem` interface will need to have a dummy `RecursiveThaw` method?  I agree the cast isn't nice but the alternative also doesn't seem very palpable...maybe I'm not following?

hqhq: Right, my mistake, we could have a RecursiveSet method for subsystem, but we can do that when we see more needs.

hqhq:
LGTM

mruck:
Thanks! @mrunalp @crosbymichael any comments?

mruck:
@mrunalp @crosbymichael anything we can do to get this merged in?

mruck:
@mrunalp @crosbymichael hoping to get this merged in since we use this in production and @circosta looks like he ran into something similar

crosbymichael:
My worry with this change is that people can setup sub-cgroups that maybe related to the container but not bound to its lifecycle and this would be a breaking change if we recursed down the tree and changed settings.  

What usecase are you solving and how are you getting yourself into the position that is causing the error?

percontation:
> My worry with this change is that people can setup sub-cgroups that maybe related to the container but not bound to its lifecycle and this would be a breaking change if we recursed down the tree and changed settings.

Yes, the recursive calls only apply in the SIGKILL case of container shutdown. So, this patch doesn't cause any recursive "change of settings" of any process that runC wasn't about to send a SIGKILL at.

> What usecase are you solving and how are you getting yourself into the position that is causing the error?

We have an application that uses runC containers. Attempting to run this application on kubernetes-based infrastructure means that we're running nested runC. As described in #2105 running nested runC can result in the outer runC deadlocking as it attempts to exit (and this extends to cases beyond just nested runC; it applies to any runC container containing an application that wants to use the freezer cgroup).

crosbymichael:
@percontation so this would be in the case of killing the pod container while another container ( inside the pod) is still frozen?

percontation:
Yes. (Although, in our case, we're not intentionally suspending containers. The temporary freezing and unfreezing that runC does while sending signals is enough to cause the problem to exhibit from time to time.)

libcontainer/init_linux.go
```diff
@@ -492,7 +492,13 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 			logrus.Warn(err)
 		}
 	}
-	if err := m.Freeze(configs.Thawed); err != nil {
+
```
crosbymichael: Wouldn't it be a little weird that the code above for getting pids and actually freezing the container during this function is not recursive but the code for thawing the cgroups is recursive?

crosbymichael: Oops, I take this back, the GetAllPids function is recursive.  So it's just the Freeze of the current cgroup that is not

percontation: Yeah, the freeze of the current cgroup isn't recursive, but the unfreeze must be recursive. This is asymmetric, but correct. (Otherwise, the code below might block forever while waiting for the killed processes to become unfrozen, when really we wanted all those PIDs to die now regardless of whether or not they were frozen by a nested cgroup).

percontation: Or another way of thinking about it: because a task is frozen if any of its parent cgroups are switched to a frozen state,
1. freezing tasks with cgroups is inherently recursive
2. unfreezing with cgroups must be done with manual recursively if you actually wish to unfreeze all the involved tasks (irrespective of the wishes of subcgroups)

And, we do wish to unfreeze tasks against the wishes of any subcgroup, because that's what's required to SIGKILL an entire process tree.

In this particular case it'd be nice if the Linux kernel "respected" SIGKILL on a frozen task (instead of treating it like every other signal, where the task remains alive and blocked until it's unfrozen). But it doesn't, so runC must recursively unfreeze as part of its recursive killing.

mruck:
@crosbymichael is there anything else we can do to help so this gets merged?

libcontainer/init_linux.go
```diff
@@ -492,7 +492,13 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
...
+	if s != unix.SIGKILL {
```
crosbymichael: Why do we need this logic here?  I feel like branching on signals will cause some issues for users when the thaw works differently.

mruck: We only want to recursive thaw on sigkill because that's the scenario when we have processes lying around after the parent process has exited.  For example, if we have a nested container that is frozen because it's paused, I believe the desired behavior is for it to remain frozen/paused.

percontation: Note that below, there's similar logic that special-cases SIGKILL. This here is special-cased for the same reasons. In the SIGKILL case, runC is expecting everything everything to die (and the code below will wait for each process to die, which can hang forever without this patch).

For every other signal, runC is not expecting every process to die, so there's no problem if container processes remain frozen instead of dying.

crosbymichael:
@mrunalp @cyphar can you help review this one as well?

mruck:
@mrunalp @cyphar can we get your feedback? really hoping to get this upstreamed since we are running into this in production on k8s

mruck:
@crosbymichael any chance you can help me get the attention of @mrunalp @cyphar?

h-vetinari:
@mruck can you give this one more shot by rebasing?

percontation:
Rebases have not been entirely trivial since underlying cgroupsv2 code keeps changing, so it's probably not worth the effort until there's enough interest in reviewing the bugfix anyways ¯\_(ツ)_/¯

h-vetinari:
On the review side at least, there's one more new reviewer: @AkihiroSuda. He might be willing to give this a shot (in addition to the [LGTM](https://github.com/opencontainers/runc/pull/2106#issuecomment-531536920) from @hqhq; will have to be reconfirmed).

percontation:
Ah, cool. Yeah, if things actually start moving I'm happy to deal with conflicts.

libcontainer/cgroups/utils.go
```diff
@@ -494,6 +494,21 @@ func getHugePageSizeFromFilenames(fileNames []string) ([]string, error) {
 	return pageSizes, nil
 }
 
+// WalkCgroups calls walkFn on the cgroup directory at path and on all its
+// subcgroups.
+func WalkCgroups(path string, walkFn func(path string) error) error {
+	return filepath.Walk(path, func(p string, info os.FileInfo, iErr error) error {
```
kolyshkin, line 500: This is quite expensive as it calls stat(2) on every file. Might be better using filepath.WalkDir() but it's only available in go 1.16+.

libcontainer/cgroups/utils.go
```diff
@@ -494,6 +494,21 @@ func getHugePageSizeFromFilenames(fileNames []string) ([]string, error) {
...
+		dir, file := filepath.Split(p)
```
kolyshkin, line 501: This is also very expensive, especially since info.Name() is readily available.

IOW
```go
if info.Name() != CgroupProcesses {
    return nil
}

File: libcontainer/cgroups/utils.go [hunk 9]
```diff
@@ -494,6 +494,21 @@ func getHugePageSizeFromFilenames(fileNames []string) ([]string, error) {
...
+		if file != CgroupProcesses {
+			return nil
+		}
+		if iErr != nil {
```
kolyshkin, line 505: Error handling (if you really need it, I'm not sure) should be done before filename comparison.

libcontainer/cgroups/utils.go
kolyshkin: OK I haven't realized that
 - you've reused the GetAllPids code, inheriting all its issues
 - those issues (except for using WalkDir() instead of Walk() for go 1.16+) are already fixed by commit 5542a2c77d67605fa0fc9b562dfe8b3d19671b94

kolyshkin:
I think we can take this in after 1.0 release. Nothing bad in thawing a container which we're going to kill anyway :)

kolyshkin:
@percontation PTAL ^^^ (and this needs a rebase)

kolyshkin:
This PR is too old to be rebased and I'm still unsure what problem is it trying to solve*, thus closing. Besides, cgroups live in a separate repo now.

* if that's properly killing cgroup v1 container, please provide a reproducer first (in a new issue). For cgroup v2, we use cgroup.kill so it's a non-issue.

Feel free to open a new issue or PR though.

PR #2112, opened 2019-08-21
libcontainer: fix network stats collector

fuweid:
rx* should read the /sys/class/net/<EthInterface>/statistics/rx* and tx*
should read the /sys/class/net/<EthInterface>/statistics/tx*.

Signed-off-by: Wei Fu <fuweid89@gmail.com>

### DIFF
--- a/libcontainer/network_linux.go
+++ b/libcontainer/network_linux.go
@@ -52,15 +52,15 @@ func getNetworkInterfaceStats(interfaceName string) (*NetworkInterface, error) {
 	}
 	// Ingress for host veth is from the container. Hence tx_bytes stat on the host veth is actually number of bytes received by the container.
 	netStats := []netStatsPair{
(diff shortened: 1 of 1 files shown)


fuweid:
ping @crosbymichael

libcontainer/network_linux.go
```diff
@@ -52,15 +52,15 @@ func getNetworkInterfaceStats(interfaceName string) (*NetworkInterface, error) {
 	}
 	// Ingress for host veth is from the container. Hence tx_bytes stat on the host veth is actually number of bytes received by the container.
```
crosbymichael: Did you notice the comment here?  Do you think it is still valid?

fuweid: oops. I didn't see this comment 😂 

I think we use wrong function to collect network stat in vm-like container...

sorry for that.

PR #2113, opened 2019-08-21
libcontainer: initial support for cgroups v2

giuseppe:
add support to libcontainer for handling a cgroups v2 unified hierarchy.

The implementation tries to reuse the existing data structures already defined for cgroups v1.  It is not an exhaustive implementation as it doesn't cover reading/writing all the files available with cgroups v2 (e.g. there is no support yet for writing to `memory.low` or `memory.min`) that have no corresponding mapping in cgroups v1 and in the data structures already defined.  Also, there is no support yet for controllers that need eBPF such as `devices`.
The only exceptions with new fields added are `CpuWeight` and `CpuMax` (defined in `libcontainer/configs/cgroup_linux.go`).

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -18,7 +18,7 @@ import (
 )
 
 var (
-	subsystems = subsystemSet{
+	subsystemsLegacy = subsystemSet{
 		&CpusetGroup{},
 		&DevicesGroup{},
 		&MemoryGroup{},
@@ -32,6 +32,14 @@ var (
 		&PerfEventGroup{},
 		&FreezerGroup{},
 		&NameGroup{GroupName: "name=systemd", Join: true},
+	}
+	subsystemsUnified = subsystemSet{
+		&CpusetGroupV2{},
+		&FreezerGroupV2{},
+		&CpuGroupV2{},
+		&MemoryGroupV2{},
+		&IOGroupV2{},
+		&PidsGroupV2{},
 	}
 	HugePageSizes, _ = cgroups.GetHugePageSize()
 )
@@ -129,6 +137,13 @@ func isIgnorableError(rootless bool, err error) bool {
 	return errno == unix.EROFS || errno == unix.EPERM || errno == unix.EACCES
 }
 
+func (m *Manager) getSubsystems() subsystemSet {
+	if cgroups.IsCgroup2UnifiedMode() {
+		return subsystemsUnified
+	}
+	return subsystemsLegacy
+}
+
 func (m *Manager) Apply(pid int) (err error) {
 	if m.Cgroups == nil {
 		return nil
@@ -158,7 +173,7 @@ func (m *Manager) Apply(pid int) (err error) {
 		return cgroups.EnterPid(m.Paths, pid)
 	}
 
-	for _, sys := range subsystems {
+	for _, sys := range m.getSubsystems() {
 		// TODO: Apply should, ideally, be reentrant or be broken up into a separate
 		// create and join phase so that the cgroup hierarchy for a container can be
 		// created then join consists of writing the process pids to cgroup.procs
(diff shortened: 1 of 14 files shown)


rhatdan:
This is all part of an effort to move to cgroups V2.  It is planned right now that Fedora 31 is going to default to CgroupsV2 and we would like make sure we can get  Runc, Kubernernetes, Docker, Moby, Podman, Buildah, CRI-O, Containerd ... all working with CgroupsV2.

filbranden:
I had something different in mind.

I was hoping to also fix the issue that libcontainer is essentially bypassing systemd altogether while writing to the cgroup tree, and that's a big problem.

I was expecting we could fix that in the systemd cgroup driver as a whole, but when I attempted that, I bumped into the issue that support for older systemd versions is not straightforward.

So now my plan was to add all the support needed to kernel + systemd (looks like kernel 5.2 + systemd v244 [one more release in the future]) will be able to have everything we need.

Then we can have a third cgroup driver here (systemd_unified) that will do all writes through systemd, using D-Bus.

That should allow us to finally lock down the cgroup tree, gives us supports to devices (systemd implements that using eBPF) and keeps us forward-compatible with whatever comes next into the cgroup tree.

giuseppe:
I see the current PR just as a starting point for a proper cgroups v2 support, `devices` support is mandatory even for a PoC.

@filbranden is the refactoring for using only systemd doable on top of the current PR?  While it doesn't enable yet cgroups v2 for runc, this initial implementation  can already be used by at least cAdvisor and Kubernetes, that are vendoring libcontainer, to work on a cgroups v2 system.

For crun I've opted to not require systemd for managing cgroups.  I want to be able to support nested containers without requiring systemd inside of the containers as well.  I had to implement the `devices` controller using [eBPF](https://github.com/containers/crun/blob/master/src/libcrun/ebpf.c).     We could probably do the same thing for libcontainer.

filbranden:
Hey @guiseppe!

I think this approach looks good for the most part, the only thing is I think I'd prefer if we started by splitting the manager so we'd have one for unified vs. one for legacy, like I did in #2114. If you agree with that approach and if you like it, would you mind incorporating that into your set of commits? You're free to take my commit into your PR, or just rework it however you like.

> this initial implementation can already be used by at least cAdvisor and Kubernetes, that are vendoring libcontainer, to work on a cgroups v2 system.

Yes, this is awesome!

> For crun I've opted to not require systemd for managing cgroups. I want to be able to support nested containers without requiring systemd inside of the containers as well.

I think that is fine, but I think a systemd-less implementation would have to live in `libcontainer/cgroups/fs/` instead, that is the cgroup driver in libcontainer that doesn't have any dependencies on systemd, so that's the right place for that (in libcontainer, runc.)

> I had to implement the devices controller using eBPF. We could probably do the same thing for libcontainer.

I have some drafts of a devices implementation for systemd, e.g. [here](https://github.com/filbranden/runc/commit/a347be179b9e9d1f795186d6aa1c199f127066fe). I'd be happy to rewrite that on top of your patch. If you use the split managers, that would be fairly easy to do, without worrying about compatibility on legacy environments.

Let me know what you think. Happy to see this moving forward!

Cheers,
Filipe

giuseppe:
pushed an updated version

giuseppe:
cc @mrunalp @crosbymichael

0x0916:
It is time for us to make runc support cgroups v2. I am very happy to see it move forward.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -63,10 +71,11 @@ type subsystem interface {
...
 type Manager struct {
-	mu       sync.Mutex
-	Cgroups  *configs.Cgroup
-	Rootless bool // ignore permission-related errors
-	Paths    map[string]string
+	mu         sync.Mutex
+	Cgroups    *configs.Cgroup
+	Rootless   bool // ignore permission-related errors
+	Paths      map[string]string
+	Subsystems subsystemSet
```
crosbymichael: Is this set somewhere? I cannot find it.

giuseppe: I've added it as a way to customize the controller and allow to specify what subsystems are supported/used.
If it is `!= nil`, then `getSubsystems()` will use it: https://github.com/opencontainers/runc/pull/2113/files#diff-3638867461edf7257846ceba1724fb8aR142

Thinking more of it, probably it is not that useful to have so I can drop it if you prefer

giuseppe: I went forward and dropped it.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #2114, opened 2019-08-22
systemd cgroup driver: Split support for unified cgroup2 hierarchy into its own Manager

filbranden:
Split support for unified cgroup2 hierarchy into its own Manager in the systemd cgroup driver

Detection is done only once, as the manager is instantiated, and then the two implementations are for the most part separate.

This will allow us to iterate on the unified cgroup2 implementation without worrying about compatibility with the legacy implementation.

For now, the unified cgroup2 implementation consists of stubs that need to be filled up with actual implementation.

/cc @giuseppe @rhatdan 

I think I'd prefer if we started from a split like this, then added the bits needed for cgroup2 support. What do you think?

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -22,7 +22,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type Manager struct {
+type LegacyManager struct {
 	mu      sync.Mutex
 	Cgroups *configs.Cgroup
 	Paths   map[string]string
@@ -167,15 +167,28 @@ func NewSystemdCgroupsManager() (func(config *configs.Cgroup, paths map[string]s
 	if !systemdUtil.IsRunningSystemd() {
 		return nil, fmt.Errorf("systemd not running on this host, can't use systemd as a cgroups.Manager")
 	}
-	return func(config *configs.Cgroup, paths map[string]string) cgroups.Manager {
-		return &Manager{
-			Cgroups: config,
-			Paths:   paths,
-		}
-	}, nil
+	unified, err := IsCgroup2UnifiedMode()
+	if err != nil {
+		return nil, err
+	}
+	if unified {
+		return func(config *configs.Cgroup, paths map[string]string) cgroups.Manager {
+			return &UnifiedManager{
+				Cgroups: config,
+				Paths:   paths,
+			}
+		}, nil
+	} else {
+		return func(config *configs.Cgroup, paths map[string]string) cgroups.Manager {
+			return &LegacyManager{
+				Cgroups: config,
+				Paths:   paths,
+			}
+		}, nil
+	}
 }
 
-func (m *Manager) Apply(pid int) error {
+func (m *LegacyManager) Apply(pid int) error {
 	var (
 		c          = m.Cgroups
 		unitName   = getUnitName(c)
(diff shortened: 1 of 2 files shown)


rhatdan:
Works for me.  But @giuseppe is doing the work.

giuseppe:
@filbranden thanks for putting this up!

> Split support for unified cgroup2 hierarchy into its own Manager in the systemd cgroup driver
> 
> Detection is done only once, as the manager is instantiated, and then the two implementations are for the most part separate.

In the PR I've opened the detection is also done only once (using a `sync.Once`).
>
> This will allow us to iterate on the unified cgroup2 implementation without worrying about compatibility with the legacy implementation.
> 
> For now, the unified cgroup2 implementation consists of stubs that need to be filled up with actual implementation.

I am fine with some higher level refactoring and split the implementation into a new manager.

The biggest part seems to be how to we will handle the subsystems manager.  Are we going to have two versions of `GetStats()` and `Apply()`?

I thought of splitting it in something like:

`ApplyLegacy()` and `ApplyUnified()`, then `Apply` can be backward compatible and look like:

```go
func (s *CpuGroup) Apply(d *cgroupData) error {
       if cgroups.IsCgroup2UnifiedMode() {
	    return s.ApplyUnified(d)
       } else {
	    return s.ApplyLegacy(d)
       }
}

```

what do you think?

filbranden:
> I am fine with some higher level refactoring and split the implementation into a new manager.
> 
> The biggest part seems to be how to we will handle the subsystems manager. Are we going to have two versions of `GetStats()` and `Apply()`?

I feel like reusing the current `subsystems` with `&fs.*Group{}` is the wrong approach though.

Why not create a new `systemd.UnifiedCpuGroup` (and same for other cgroups) and have its `Apply()` and `GetStats()` do the "unified" version?

Yes, there's perhaps some opportunity for reuse of code currently in fs and in systemd legacy cgroup, but I guess then just refactoring those bits to export the relevant bits would be good.

Does that make sense?

giuseppe:
yes thanks, I'll refactor the code and split cgroups v2 stuff in new files under `cgroups/fs`.  I'll add also a new option for `fs.Manager` to support cgroupvs2.  It should be enough to configure the subsystems used.

I won't be available next week, so it might take some time before I submit an updated version.

filbranden:
Sounds good @giuseppe !

Yes, adding those to `cgroup/fs/` makes sense if you want to support the systemd-less case with an `fs.UnifiedManager` as well.

Whenever you have a new PR, I'm happy to review it.

Cheers!
Filipe

libcontainer/cgroups/systemd/apply_systemd.go
```diff
@@ -167,15 +167,28 @@ func NewSystemdCgroupsManager() (func(config *configs.Cgroup, paths map[string]s
...
+		return nil, err
+	}
+	if unified {
+		return func(config *configs.Cgroup, paths map[string]string) cgroups.Manager {
+			return &UnifiedManager{
+				Cgroups: config,
+				Paths:   paths,
+			}
+		}, nil
+	} else {
```
AkihiroSuda: nit: you don't need `else` here. Some lint tool may print a warning.

AkihiroSuda:
needs rebase

filbranden:
This has been incorporated in @giuseppe's #2113, so closing now.

PR #2116, opened 2019-08-23
Avoid the dependency on cgo through go-systemd/util package

filbranden:
Avoid the dependency on cgo through go-systemd/util package

This dependency is only needed in package "github.com/coreos/go-systemd/util" and we only use it for `IsRunningSystemd()`, which is a simple Go function that just stats a file.

Let's just borrow it here, so we remove the dependency and can remove that package from vendored build. (Licenses are compatible, Apache 2.0 here and there.)

This also removes dependencies on `dlopen` and on trying to find `libsystemd.so` or `libsystemd-login.so` in the system.

Tested that this still builds and works as expected.

(An alternative approach would be to fix coreos/go-systemd to expose `IsSystemdRunning()` under a different package out of util, and redirect the util implementation to that one. But it's so small that I think it's so small, straightforward, well defined and unlikely to change, that I think it's worth it just replicating it here.)

Cheers!
Filipe


crosbymichael:
LGTM

mrunalp:
LGTM

PR #2117, opened 2019-08-23
Remove libcontainer detection for systemd features

### LINKED ISSUE #2108: Cache the systemd detection results

filbranden:
Remove libcontainer detection for systemd features

Transient units (and transient slice units) have been available for quite a long time and RHEL 7 with systemd v219 (likely the oldest OS we care about at this point) supports that. A system running a systemd without these features is likely to break a lot of other stuff that runc/libcontainer care about.

Regarding delegated slices, modern systemd doesn't allow it and runc/libcontainer run fine on it, so we might as well just stop requesting it on older versions of systemd which allowed it. (Those versions never really changed behavior significantly when that option was passed anyways.)

Fixes #2108.

@youurayy Would you be willing to test this commit in your environment, confirm that it fixes the aforementioned issue and also everything else works as expected? You seem to be testing a fair number of distros to be able to catch unexpected issues resulting from this change... Thanks!

### DIFF
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -72,11 +72,8 @@ const (
 )
 
 var (
-	connLock                   sync.Mutex
-	theConn                    *systemdDbus.Conn
-	hasStartTransientUnit      bool
-	hasStartTransientSliceUnit bool
-	hasDelegateSlice           bool
+	connLock sync.Mutex
+	theConn  *systemdDbus.Conn
 )
 
 func newProp(name string, units interface{}) systemdDbus.Property {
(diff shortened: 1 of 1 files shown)


youurayy:
No probs, I will test on CentOS 7 and Ubuntu 18.04 today. (via my [k8s-hyperv](https://github.com/youurayy/k8s-hyperv) and [k8s-hyperkit](https://github.com/youurayy/k8s-hyperkit) projects)

youurayy:
works - tested on Ubuntu 18.04 and CentOS 7 1907

https://github.com/youurayy/runc/releases/tag/v1.0.0-rc8-slice-fix-2

filbranden:
@youurayy Thanks for confirming this works!

@cyphar What do you say we go with this approach?

Cheers,
Filipe

crosbymichael:
LGTM

mrunalp:
LGTM

PR #2125, opened 2019-09-06
cgroup: support mount of cgroup2

giuseppe:
convert a "cgroup" mount to "cgroup2" when the system uses cgroups v2
unified hierarchy.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -209,6 +209,80 @@ func prepareBindMount(m *configs.Mount, rootfs string) error {
 	return nil
 }
 
+func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns bool) error {
+	binds, err := getCgroupMounts(m)
+	if err != nil {
+		return err
+	}
+	var merged []string
+	for _, b := range binds {
+		ss := filepath.Base(b.Destination)
+		if strings.Contains(ss, ",") {
+			merged = append(merged, ss)
+		}
+	}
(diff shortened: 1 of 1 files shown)


giuseppe:
@filbranden FYI

giuseppe:
tested with:
```
# podman run --cgroupns private -ti --rm --memory 100M fedora cat /sys/fs/cgroup/memory.max
104857600

# podman run -ti --rm --memory 100M fedora sh -c 'cat /sys/fs/cgroup/`sed -e "s|0::||g" /proc/self/cgroup`/memory.max'
104857600
```

libcontainer/rootfs_linux.go
```diff
@@ -209,6 +209,74 @@ func prepareBindMount(m *configs.Mount, rootfs string) error {
...
+			if err := os.Symlink(mc, filepath.Join(rootfs, m.Destination, ss)); err != nil && !os.IsExist(err) {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func mountCgroupV2(m *configs.Mount, rootfs, mountLabel string, enableCgroupns bool) error {
+	cgroupPath := filepath.Join(rootfs, m.Destination)
```
TomSweeneyRedHat: any worries about symlinks here?

giuseppe: good point.  I've pushed a new version that uses `securejoin`

crosbymichael:
LGTM

mrunalp:
LGTM

PR #2129, opened 2019-09-23
Only allow proc mount if it is procfs

### LINKED ISSUE #2128: AppArmor can be bypassed by a malicious image that specifies a volume at /proc
A malicious volume can specify a volume mount on `/proc`. Since Docker populates the volume by copying data present in the image, it's possible to build a fake structure that will trick runc into believing it had successfully written to `/proc/self/attr/exec`:

https://github.com/opencontainers/runc/blob/7507c64ff675606c5ff96b0dd8889a60c589f14d/libcontainer/apparmor/apparmor.go#L23-L25

This is possible because apparmor.ApplyProfile is executed in the container rootfs, after pivot_root in prepareRootfs:

https://github.com/opencontainers/runc/blob/7507c64ff675606c5ff96b0dd8889a60c589f14d/libcontainer/standard_init_linux.go#L115-L117

checkMountDestinations is supposed to prevent mounting on top of `/proc`:

https://github.com/opencontainers/runc/blob/7507c64ff675606c5ff96b0dd8889a60c589f14d/libcontainer/rootfs_linux.go#L464-L469

... but the check does not work. I believe the reason is that the `dest` argument is resolved to an absolute path using `securejoin.SecureJoin` (before pivot_root), unlike the blacklist in checkMountDestinations, which is  relative to the rootfs:

https://github.com/opencontainers/runc/blob/7507c64ff675606c5ff96b0dd8889a60c589f14d/libcontainer/rootfs_linux.go#L414-L419

Minimal proof of concept (on Ubuntu 18.04):

```bash
mkdir -p rootfs/proc/self/{attr,fd}
touch rootfs/proc/self/{status,attr/exec}
touch rootfs/proc/self/fd/{4,5}

cat <<EOF > Dockerfile
FROM busybox
ADD rootfs /

VOLUME /proc
EOF

docker build -t apparmor-bypass .
docker run --rm -it --security-opt "apparmor=docker-default"  apparmor-bypass
# container runs unconfined
```

Not a critical bug on its own, but should get a CVE assigned.

Discovered by Adam Iwaniuk and disclosed during DragonSector CTF (https://twitter.com/adam_iwaniuk/status/1175741830136291328).

The CTF challenge mounted a file to `/flag-<random>` and denied access to it using an AppArmor policy. The bug could then be used to disable the policy and read the file: https://gist.github.com/leoluk/2513b6bbff8aa5cd623f3d7d7f20871a

crosbymichael:
Fixes #2128

This allows proc to be bind mounted for host and rootless namespace usecases but
it removes the ability to mount over the top of proc with a directory.

```bash
> sudo docker run --rm  apparmor
docker: Error response from daemon: OCI runtime create failed:
container_linux.go:346: starting container process caused "process_linux.go:449:
container init caused \"rootfs_linux.go:58: mounting
\\\"/var/lib/docker/volumes/aae28ea068c33d60e64d1a75916cf3ec2dc3634f97571854c9ed30c8401460c1/_data\\\"
to rootfs
\\\"/var/lib/docker/overlay2/a6be5ae911bf19f8eecb23a295dec85be9a8ee8da66e9fb55b47c841d1e381b7/merged\\\"
at \\\"/proc\\\" caused
\\\"\\\\\\\"/var/lib/docker/overlay2/a6be5ae911bf19f8eecb23a295dec85be9a8ee8da66e9fb55b47c841d1e381b7/merged/proc\\\\\\\"
cannot be mounted because it is not of type proc\\\"\"": unknown.

> sudo docker run --rm -v /proc:/proc apparmor

docker-default (enforce)        root     18989  0.9  0.0   1288     4 ?
Ss   16:47   0:00 sleep 20
```

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -19,7 +19,7 @@ import (
 	"syscall" // only for SysProcAttr and Signal
 	"time"
 
-	"github.com/cyphar/filepath-securejoin"
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
@@ -1176,7 +1176,7 @@ func (c *linuxContainer) makeCriuRestoreMountpoints(m *configs.Mount) error {
 		if err != nil {
 			return err
 		}
-		if err := checkMountDestination(c.config.Rootfs, dest); err != nil {
+		if err := checkProcMount(c.config.Rootfs, dest, ""); err != nil {
 			return err
 		}
 		m.Destination = dest
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -13,7 +13,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cyphar/filepath-securejoin"
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/mrunalp/fileutils"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
@@ -197,7 +197,7 @@ func prepareBindMount(m *configs.Mount, rootfs string) error {
 	if dest, err = securejoin.SecureJoin(rootfs, m.Destination); err != nil {
 		return err
 	}
-	if err := checkMountDestination(rootfs, dest); err != nil {
+	if err := checkProcMount(rootfs, dest, m.Source); err != nil {
 		return err
 	}
 	// update the mount with the correct dest after symlinks are resolved.
@@ -414,7 +414,7 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 		if dest, err = securejoin.SecureJoin(rootfs, m.Destination); err != nil {
 			return err
 		}
-		if err := checkMountDestination(rootfs, dest); err != nil {
+		if err := checkProcMount(rootfs, dest, m.Source); err != nil {
 			return err
 		}
 		// update the mount with the correct dest after symlinks are resolved.
@@ -461,12 +461,12 @@ func getCgroupMounts(m *configs.Mount) ([]*configs.Mount, error) {
 	return binds, nil
 }
 
-// checkMountDestination checks to ensure that the mount destination is not over the top of /proc.
+// checkProcMount checks to ensure that the mount destination is not over the top of /proc.
 // dest is required to be an abs path and have any symlinks resolved before calling this function.
-func checkMountDestination(rootfs, dest string) error {
-	invalidDestinations := []string{
-		"/proc",
-	}
+//
+// if source is nil, don't stat the filesystem.  This is used for restore of a checkpoint.
+func checkProcMount(rootfs, dest, source string) error {
+	const procPath = "/proc"
 	// White list, it should be sub directories of invalid destinations
 	validDestinations := []string{
 		// These entries can be bind mounted by files emulated by fuse,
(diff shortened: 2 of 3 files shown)


cyphar:
I'm not sure we should allow general procfs bind-mounts -- we won't be masking any of them and I'd be worried that all you'd now need to do is convince Docker to add extra `procfs` mounts. I think the old logic (don't allow mounts under procfs) was serviceable -- we just need to make it so that we do a `filepath.Rel` after the `securejoin` (none of this is symlink-race-safe but that isn't a problem here since we're still in container setup).

cyphar:
This looks alright (though I'm a bit iffy about allowing procfs bind-mounts inside `/proc` since it can also pretty easily defeat masked-paths -- especially given that we didn't allow it before). We do have to bind-mount the "host" proc for some peculiar nested container scenarios, but I'm not sure if we ever need to explicitly bind-mount something under `procfs` (I might be mis-remembering though).

crosbymichael:
The logic to allow it came from #1832, 

We can always revert that pr and that "should" fix the issue

cyphar:
Maybe I'm just being a little bit dense (it is pretty late at night here), but unless I'm mistaken the original check was:

```
if path == "." || !strings.HasPrefix(path, "..") { /* error */ }
```

Which blocks mounting on `/proc` or underneath it. This was changed in #1832 to:

```
if path != "." && !strings.HasPrefix(path, "..")  { /* error */ }
```

Which just allowed `/proc` but still disallowed any mounts under `/proc`. This PR changes it to:

```
if path == "." || !strings.HasPrefix(path, "..")  { /* error _unless_ it's procfs */ }
```

Which allows mount under `/proc` as long as its procfs. My point was that I don't think we need to allow it, since even #1832 didn't allow it. I'd say that the special procfs check should only apply for `path == "."` and disallow `!strings.HasPrefix(path, "..")` entirely.

libcontainer/mount/mount.go
```diff
@@ -21,3 +26,20 @@ func Mounted(mountpoint string) (bool, error) {
 	}
 	return false, nil
 }
+
+// FSType returns the file-system type of a mountpoint.
+//
+// ErrNotMounted is returned if the mountpoint refers to a path
+// that is not mounted.
+func FSType(mountpoint string) (string, error) {
```
cyphar: Wouldn't it be simpler to do an `statfs` and check whether `f_type == PROC_SUPER_MAGIC`?

crosbymichael: done

crosbymichael:
@cyphar updated

cyphar:
LGTM.

bennofs:
Does this also check for mounting to `/`? If I mount on top of `/`, the same issue should occur, right? (Because I can just create a fake proc directory in my mount on top of `/`)

crosbymichael:
@bennofs can you expand on what you mean by mounting to `/`?  What are you mounting and where?

crosbymichael:
ping @mrunalp @giuseppe @rhatdan Can you plz review this one?

bennofs:
@crosbymichael If you have a volume like this in your Dockerfile:

```
VOLUME /
```

That can also override `/proc`, no?

mrunalp:
Yeah, I will make a pass today.

cyphar:
@bennofs Maybe we should also include a check for direct-ancestors -- though it should be noted that if you volume-mount over `/` then the container won't start (so as an attack it's pretty useless). For example:

```
% docker run -it foo bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"open /dev/ptmx: no such file or directory\"": unknown.
% docker run foo bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"apply apparmor profile: apparmor failed to apply profile: open /proc/self/attr/exec: no such file or directory\"": unknown.
```

If you modify it to work around AppArmor, the process appears to die really early anyway:

```
% docker run foo bash
docker: Error response from daemon: OCI runtime start failed: container process is already dead: unknown.
```

giuseppe:
> ping @mrunalp @giuseppe @rhatdan Can you plz review this one?

the patch LGTM.

@crosbymichael @cyphar: I wonder if it possible to apply the profile before the `pivot_root` to be sure it happens on the host procfs

crosbymichael:
@giuseppe i'd rather take this approach because your suggestion about moving it before pivot_root won't work with `exec`'d processes

giuseppe:
I think this PR makes completely sense and we should have it.  I was only thinking of moving it as an additional countermeasure

cyphar:
CVE-2019-16884 has been assigned for this issue.

mrunalp:
LGTM

mrueg:
@cyphar @crosbymichael  can you release a new RC for this?

cyphar:
@mrueg This fix is not sufficient to fix the CVE completely -- see the discussion in #2128. We need to merge #2130 and opencontainers/selinux#59 first. Once that's done, I'll send out a vote for rc9.

rhatdan:
The SELinux fix is merged.

cyphar:
Yup, I'm updating #2130 to include a vendor bump (though strangely `vndr` appears to be breaking quite badly for me).

rhatdan:
Looks like everything is merged, are we cutting a new release?

cyphar:
Yes, I will send out the vote tomorrow morning (it's quite late here).

PR #2130, opened 2019-09-27
*: verify operations on /proc/... are on procfs

cyphar:
This is an additional mitigation for CVE-2019-16884. The primary problem
is that Docker can be coerced into bind-mounting a file system on top of
/proc (resulting in label-related writes to /proc no longer happening).

While we are working on mitigations against permitting the mounts, this
helps avoid our code from being tricked into writing to non-procfs
files. This is not a perfect solution (after all, there might be a
bind-mount of a different procfs file over the target) but in order to
exploit that you would need to be able to tweak a config.json pretty
specifically (which thankfully Docker doesn't allow).

Specifically this stops AppArmor from not labeling a process silently
due to /proc/self/attr/... being incorrectly set, and stops any
accidental fd leaks because /proc/self/fd/... is not real.

See opencontainers/runc#2128

### DIFF
--- a/libcontainer/utils/utils_unix.go
+++ b/libcontainer/utils/utils_unix.go
@@ -3,33 +3,57 @@
 package utils
 
 import (
-	"io/ioutil"
+	"fmt"
 	"os"
 	"strconv"
 
 	"golang.org/x/sys/unix"
 )
 
+// EnsureProcHandle returns whether or not the given file handle is on procfs.
+func EnsureProcHandle(fh *os.File) error {
+	var buf unix.Statfs_t
+	if err := unix.Fstatfs(int(fh.Fd()), &buf); err != nil {
+		return fmt.Errorf("ensure %s is on procfs: %v", fh.Name(), err)
+	}
+	if buf.Type != unix.PROC_SUPER_MAGIC {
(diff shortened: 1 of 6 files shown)


dqminh:
LGTM

libcontainer/utils/utils_unix.go
```diff
@@ -3,33 +3,54 @@
...
 	"strconv"
 
 	"golang.org/x/sys/unix"
 )
 
+// IsProcHandle returns whether or not the given file handle is on procfs.
+func IsProcHandle(fh *os.File) (bool, error) {
+	var buf unix.Statfs_t
+	err := unix.Fstatfs(int(fh.Fd()), &buf)
+	return buf.Type == unix.PROC_SUPER_MAGIC, err
```
dqminh: nit: `IsProcHandle` returns only bool is simpler, as caller has no way to handle error sanely so it should just abort with `/proc is not what we expected`

cyphar: I'll switch it to only return `error` (so at least you know why it failed).

dqminh:
LGTM

crosbymichael:
LGTM

Issue #2131, opened 2019-09-27
Runc shouldn't expect only 4 fileds in cpuacct.stat


skilxn-go:
In some kernel, cpuacct.stat is expand with some other useful fields. In this case, runc can't work correctly due to it limited this file to have only 4 fileds.

skilxnTL:
For k8s，it will lead imagegc's failed, and someother errors

cyphar:
#2132 is merged.

PR #2132, opened 2019-09-27
Support different field counts of cpuaact.stats

skilxnTL:
#2131


libcontainer/cgroups/fs/cpuacct.go
```diff
@@ -78,27 +78,41 @@ func getCpuUsageBreakdown(path string) (uint64, uint64, error) {
...
+	if len(fields)%2 != 0 {
+		return 0, 0, fmt.Errorf("failure - %s is expected to have even fields, actually %d", filepath.Join(path, cgroupCpuacctStat), len(fields))
 	}
-	if fields[2] != systemField {
-		return 0, 0, fmt.Errorf("unexpected field %q in %q, expected %q", fields[2], cgroupCpuacctStat, systemField)
+
+	statDict := make(map[string]string)
+	for i := 0; i < len(fields); i += 2 {
+		if _, ok := statDict[fields[i]]; ok {
+			return 0, 0, fmt.Errorf("failure - %s has %s appears more than once", filepath.Join(path, cgroupCpuacctStat), fields[i])
```
thaJeztah: "appears" should probably be "appear";

```suggestion
			return 0, 0, fmt.Errorf("failure - %s has %s appear more than once", filepath.Join(path, cgroupCpuacctStat), fields[i])
```

of perhaps (failure is a bit redundant?);

```suggestion
			return 0, 0, fmt.Errorf("duplicate field %q in %q", fields[i], filepath.Join(path, cgroupCpuacctStat))
```

libcontainer/cgroups/fs/cpuacct.go
```diff
@@ -78,27 +78,41 @@ func getCpuUsageBreakdown(path string) (uint64, uint64, error) {
...
+	// key1 value1
+	// key2 value2
+	// ...
 	data, err := ioutil.ReadFile(filepath.Join(path, cgroupCpuacctStat))
 	if err != nil {
 		return 0, 0, err
 	}
+
 	fields := strings.Fields(string(data))
-	if len(fields) != 4 {
```
thaJeztah: Wondering; so this function is now able to handle the additional fields, which adds a lot of complexity, but the additional fields are not used anywhere.

Is the problem you're trying to solve that there's _more than_ 4 fields, or is the problem that `user` and `system` are not the first two lines (which sounds unlikely)

If it's just that there's more than 4 fields, then I think the only thing needed is to change this line;

```go
if len(fields) < 4 {
```

And the error message;

```go
return 0, 0, fmt.Errorf("failure - %s is expected to have at least 4 fields", filepath.Join(path, cgroupCpuacctStat))
```

cyphar:
This change is necessary (and I'm surprised we hardcoded a check for 4 given that the file is designed for extension), but they way you've gone about it makes the code look even more confusing.

I think it would be much cleaner if we split the file line-by-line and then took the first field as the key and the rest of the line as the value (which is the format) -- or possibly splitting all the fields in the line and storing them as `map[string][]string`. That way you avoid all of this `i += 2` and `len(foo) % 2 == 0` complexity.

libcontainer/cgroups/fs/cpuacct.go
skilxnTL: How about now?

skilxnTL:
Based on @thaJeztah 's comments, I decide fix this problem using a minimum modify. 
Please help me to review this version, thx, @cyphar

crosbymichael:
LGTM

cyphar:
LGTM. I think switching to a map would've been a nicer cleanup, but this fixes the issue.

PR #2136, opened 2019-10-04
nsenter: cloned_binary: do not clone exe if on readonly rootfs

vaibhav-rustagi:
Fix for CVE-2019-5736 introduces a logic to clone the /proc/self/exe
so that no malicious container can modify the host binary. Due to this
bind mount was done in try_bindfd() which introduces three libmount
events. In systemd <= 232, for every libmount event, iteration over
all mounts present in /proc/self/mountinfo is done to check which
mount got changed. In case of large number of mounts already present,
systemd will require high cpu utilization for iteration over all the
mounts. In case of a read-only rootfs, cloning of /proc/self/exe is
not required and therefore, generation of libmount events can be
avoided.

This patch adds a check to see whether /proc/self/exe is on a
read-only rootfs or not. In case, it is present on a read-only
rootfs, cloning is not required.

Signed-off-by: Vaibhav Rustagi <vaibhavrustagi@google.com>

### DIFF
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -28,6 +28,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/statfs.h>
+#include <sys/statvfs.h>
 #include <sys/vfs.h>
 #include <sys/mman.h>
 #include <sys/mount.h>
@@ -86,15 +87,34 @@ static void *must_realloc(void *ptr, size_t size)
 /*
  * Verify whether we are currently in a self-cloned program (namely, is
  * /proc/self/exe a memfd). F_GET_SEALS will only succeed for memfds (or rather
- * for shmem files), and we want to be sure it's actually sealed.
+ * for shmem files), and we want to be sure it's actually sealed. In case the
+ * the program is on a read-only rootfs, do not clone.
  */
 static int is_self_cloned(void)
 {
 	int fd, ret, is_cloned = 0;
 	struct stat statbuf = {};
 	struct statfs fsbuf = {};
+	struct statvfs stat;
+	char abs_path[1024];
 
-	fd = open("/proc/self/exe", O_RDONLY|O_CLOEXEC);
+	memset(&stat, 0, sizeof(struct statvfs));
+	const char *path = "/proc/self/exe";
+
+	ssize_t len = readlink(path, abs_path, sizeof(abs_path)-1);
+	if (len != -1) {
+		/* readlink does not provide the path with null terminated character. */
+		abs_path[len] = '\0';
+
+		if (statvfs(abs_path, &stat) == 0) {
+			/* Checks whether a program is on a read-only rootfs. */
+			if (stat.f_flag & ST_RDONLY) {
+				return true;
+			}
+		}
+	}
+
+	fd = open(path, O_RDONLY|O_CLOEXEC);
 	if (fd < 0)
 		return -ENOTRECOVERABLE;
 


libcontainer/nsenter/cloned_binary.c
```diff
@@ -86,15 +87,34 @@ static void *must_realloc(void *ptr, size_t size)
...
+
+	memset(&stat, 0, sizeof(struct statvfs));
+	const char *path = "/proc/self/exe";
+
+	ssize_t len = readlink(path, abs_path, sizeof(abs_path)-1);
+	if (len != -1) {
+		/* readlink does not provide the path with null terminated character. */
+		abs_path[len] = '\0';
+
+		if (statvfs(abs_path, &stat) == 0) {
```
AkihiroSuda: When the actual length of the path is over 1024, this is not reliable

libcontainer/nsenter/cloned_binary.c
```diff
@@ -86,15 +87,34 @@ static void *must_realloc(void *ptr, size_t size)
...
+			/* Checks whether a program is on a read-only rootfs. */
+			if (stat.f_flag & ST_RDONLY) {
+				return true;
```
AkihiroSuda: The function name is confusing because now it returns true even when not cloned

cyphar:
NACK.

We already do this exact check (despite my better judgement) but it's hidden behind a requirement to set `_LIBCONTAINER_CLONED_BINARY` as an environment variable -- the reason for this is that this check is actually not bullet-proof and we normally want to bind-mount it ourselves to make sure an admin process won't make the root read-write after the container has started.

So if you just set `_LIBCONTAINER_CLONED_BINARY` in your container environment, it should fix your problem without the need for this strange patch (which can cause security problems).

More long-term the much more correct solution is for [my kernel fixes](https://lore.kernel.org/lkml/20190930183316.10190-1-cyphar@cyphar.com/) to be merged and backported, which will permit us to remove this logic entirely (because re-opens of `/proc/self/exe` with write permissions will fail completely).

libcontainer/nsenter/cloned_binary.c
```diff
@@ -86,15 +87,34 @@ static void *must_realloc(void *ptr, size_t size)
...
+ * for shmem files), and we want to be sure it's actually sealed. In case the
+ * the program is on a read-only rootfs, do not clone.
  */
 static int is_self_cloned(void)
 {
 	int fd, ret, is_cloned = 0;
 	struct stat statbuf = {};
 	struct statfs fsbuf = {};
+	struct statvfs stat;
+	char abs_path[1024];
```
cyphar: Just for the record, the best way of doing this is:

```
char abs_path[PATH_MAX+1] = {};
readlink(path, abs_path, PATH_MAX);
```

But more importantly, this method of checking is completely un-necessary. You can just do a `statfs("/proc/self/exe")`.

vaibhav-rustagi: Sorry for the late response. Regarding your suggestion to set '_LIBCONTAINER_CLONED_BINARY' to 1 will require change in every container that were currently running. The problem of high cpu usage was encountered when kubernetes pods running containers have a liveness/readiness probe [exec based]. I couldn't find a way to pass the environment variable to all the containers in the system instead of passing it through '-e'.

libcontainer/nsenter/cloned_binary.c
vaibhav-rustagi: I was not sure of having an absolute path corresponding to more than 1024 characters. But, yes using `PATH_MAX` for the array size will be a better approach as described by cyphar@.

libcontainer/nsenter/cloned_binary.c
cyphar: > Regarding your suggestion to set '_LIBCONTAINER_CLONED_BINARY' to 1 will require change in every container that were currently running.

Unless I'm very mistaken, you should be able to just set `_LIBCONTAINER_CLONED_BINARY=1` before you call into `runc` (you can probably do this by creating a wrapper around `/usr/bin/runc` or maybe even set it as an environment variable via a Kubelet configuration or through systemd). If you can't do that, I'd be willing to take a patch which allowed passing it.

> Also, could you specify the security concern you have in having an additional check whether /proc/self/exe is on a read-only rootfs or not?

If the host is read-only only temporarily (or if the attacker can trigger the host to temporarily make the host read-write) then you can trivially re-exploit CVE-2019-5736 (defeating the whole point of this protection). This is why we require the bind-mount to be *ours* and why we lazy-umount it (there are even ways of even working around **that** protection -- which is why we generally prefer doing it with `memfd`s -- but the mount-related attacks require too many privileges to be a realistic security concern).

Maybe this isn't a concern for Google, but I'd prefer to avoid opening the door to additional attacks for everyone -- `_LIBCONTAINER_CLONED_BINARY=1` is a hack that Google can use to disable it because you're sure that your environment is secure enough.

vaibhav-rustagi: I ran few experiments to pass the environment variable to `runc` either through `docker run -e` or through a containerd.service. However, in both cases `runc` was creating the mounts.

Example:
docker run -e "_LIBCONTAINER_CLONED_BINARY=1" centos

$ docker ps -a
7f72efed1fba        centos              "/bin/bash"         3 minutes ago       Exited (0) 3 minutes ago                       condescending_bhabha

In the dbus-messages, I could see:

sig     1570642924.135377       7713    :1.0    <none>  /org/freedesktop/systemd1/unit/run_2ddocker_2druntime_5cx2drunc_2dmoby_2d7f72efed1fba3b7ebd10935627b9362a198983d281c620ec4d8e500261266a8d_2drunc_2ek2xf7F_2emount       org.freedesktop.DBus.Properties PropertiesChanged
sig     1570642924.135499       7714    :1.0    <none>  /org/freedesktop/systemd1/unit/run_2ddocker_2druntime_5cx2drunc_2dmoby_2d7f72efed1fba3b7ebd10935627b9362a198983d281c620ec4d8e500261266a8d_2drunc_2ek2xf7F_2emount       org.freedesktop.DBus.Properties PropertiesChanged

which should be related to libmount events generated due to mounts created during cloning by runc.

For containerd.service:
[Service]
Environment="_LIBCONTAINER_CLONED_BINARY=1"

This also resulted in the same behavior as above.

To check the behavior with the environment variable set, I explicitly added the below line in `is_self_cloned`:

putenv(CLONED_BINARY_ENV "=1");

and the it results into the expected behavior of not cloning [not creating mounts] by runc.

AkihiroSuda:
Closing due to NACK, but feel free to submit a new proposal if you have an alternative idea.

PR #2139, opened 2019-10-12
checkpoint: Set descriptors.json file mode to 0600

rst0git:
Prevent unprivileged users from being able to read `descriptors.json`

cc: @avagin @adrianreber

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1111,7 +1111,7 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 			return err
 		}
 
-		err = ioutil.WriteFile(filepath.Join(criuOpts.ImagesDirectory, descriptorsFilename), fdsJSON, 0655)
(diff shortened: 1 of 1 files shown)


adrianreber:
Sounds correct!

crosbymichael:
LGTM

mrunalp:
LGTM

libcontainer/container_linux.go
```diff
@@ -1111,7 +1111,7 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 			return err
 		}
 
-		err = ioutil.WriteFile(filepath.Join(criuOpts.ImagesDirectory, descriptorsFilename), fdsJSON, 0655)
+		err = ioutil.WriteFile(filepath.Join(criuOpts.ImagesDirectory, descriptorsFilename), fdsJSON, 0600)
```
avagin: should we do the same for  os.Mkdir(criuOpts.WorkDirectory, 0755)?

adrianreber: Sounds like a good idea. No checkpoint files should be readable by anyone else but the user creating it.

PR #2140, opened 2019-10-15
Set unified mountpoint in find mnt func

crosbymichael:
This is needed for the fsv2 cgroups to work when there is a unified mountpoint.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -20,8 +20,9 @@ import (
 )
 
 const (
-	CgroupNamePrefix = "name="
-	CgroupProcesses  = "cgroup.procs"
+	CgroupNamePrefix  = "name="
+	CgroupProcesses   = "cgroup.procs"
+	unifiedMountpoint = "/sys/fs/cgroup"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe PTAL

giuseppe:
LGTM

mrunalp:
LGTM

cyphar:
Isn't `/sys/fs/cgroup/unified` meant to be the unified mountpoint? Or was this changed again?

AkihiroSuda:
> Isn't /sys/fs/cgroup/unified meant to be the unified mountpoint?

It is the hybrid mountpoint :P

libcontainer/cgroups/utils.go
```diff
@@ -50,6 +51,9 @@ func IsCgroup2UnifiedMode() bool {
 
 // https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
 func FindCgroupMountpoint(cgroupPath, subsystem string) (string, error) {
+	if IsCgroup2UnifiedMode() {
```
AkihiroSuda: Can we update godoc comment to make sure `subsystem` is ignored for unified mode? Or can we return `errors.New("subsystem must not be specified for unified mode")`?

AkihiroSuda:
ping @cyphar

libcontainer/cgroups/utils.go
crosbymichael: in another PR.  Trying to keep these changes small so they can be merged.

AkihiroSuda:
@dqminh mergeable?

PR #2148, opened 2019-10-18
cgroup2: cpuset_v2: skip Apply when no limit is specified

### LINKED ISSUE #2143: cgroup2: v1.0.0-rc.9 doesn't start up without adding +cpuset to /sys/fs/cgroup/cgroup.subtree_control
```
# /tmp/runc run foo
WARN[0000] signal: killed                               
ERRO[0000] container_linux.go:346: starting container process caused "process_linux.go:297: applying cgroup configuration for process caused \"open /sys/fs/cgroup/foo/cpuset.cpus.effective: no such file or directory\"" 
container_linux.go:346: starting container process caused "process_linux.go:297: applying cgroup configuration for process caused \"open /sys/fs/cgroup/foo/cpuset.cpus.effective: no such file or directory\""
```

runc: 1.0.0-rc.9
OS: Fedora 31 beta (cgroup2 unified)

`echo +cpuset > /sys/fs/cgroup/cgroup.subtree_control` solves the issue

AkihiroSuda:

Fix #2143

### DIFF
--- a/libcontainer/cgroups/fs/cpuset_v2.go
+++ b/libcontainer/cgroups/fs/cpuset_v2.go
@@ -22,11 +22,14 @@ func (s *CpusetGroupV2) Name() string {
 }
 
 func (s *CpusetGroupV2) Apply(d *cgroupData) error {
+	if d.config.Resources.CpusetCpus == "" && d.config.Resources.CpusetMems == "" {
+		return nil
+	}
 	dir, err := d.path("cpuset")
 	if err != nil && !cgroups.IsNotFound(err) {
 		return err
 	}
-	return s.ApplyDir(dir, d.config, d.pid)
+	return s.applyDir(dir, d.config, d.pid)
 }
 
 func (s *CpusetGroupV2) Set(path string, cgroup *configs.Cgroup) error {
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/fs/cpuset_v2.go
```diff
@@ -22,11 +22,14 @@ func (s *CpusetGroupV2) Name() string {
 }
 
 func (s *CpusetGroupV2) Apply(d *cgroupData) error {
+	if d.config.Resources.CpusetCpus == "" && d.config.Resources.CpusetMems == "" {
+		return nil
+	}
 	dir, err := d.path("cpuset")
```
AkihiroSuda: not directly related to this PR, but I'm not convinced of `d.path()` logic  in v2 `Apply()` functions.

We should check controller availability and maybe write `cgroup.subtree_control` as crun does?

AkihiroSuda: cc @giuseppe

giuseppe: yes, I think we should try to enable controllers that are needed

giuseppe:
LGTM

rhatdan:
@cyphar @mrunalp @crosbymichael PTAL

crosbymichael:
But should we still join the controllers like in v1, even if no limits are set? Do we still have to join everything to collect stats on the processes?

AkihiroSuda:
For unified hierarchy, I think joining (writing `cgroup.procs`) should be managed by `Manager` rather than each of subsystem, because it is impossible to join into a specific subsystem.

crosbymichael:
LGTM

AkihiroSuda:
I suggest removing `Apply()` functions from v2 controllers (https://github.com/opencontainers/runc/issues/2157), but I think this PR can be merged now, as a workaround until we make decision on #2157.

AkihiroSuda:
@mrunalp @hqhq @dqminh @cyphar PTAL?

hqhq:
LGTM

Issue #2156, opened 2019-10-26
cgroup2: GetAllSubsystems() should not refer /proc/cgroups for unified mode


AkihiroSuda:
https://github.com/opencontainers/runc/blob/c4d8e1688c816a8cef632a3b44a38611511b7140/libcontainer/cgroups/utils.go#L267-L291

`GetAllSubsystems()` should parse `/sys/fs/cgroup/cgroup.controllers` rather than `/proc/cgroups`, which is mostly for v1 (#2155 

Note that `/sys/fs/cgroup/cgroup.controllers` does not contain freezer and device controller

AkihiroSuda:
Fixed in #2160

PR #2160, opened 2019-10-27
cgroup2: do not parse /proc/cgroups

### LINKED ISSUE #2155: cgroup2: IOGroupV2.Name() should return "io" rather than "blkio"?
The currently implementation of ` IOGroupV2.Name()` returns "blkio"
https://github.com/opencontainers/runc/blob/c4d8e1688c816a8cef632a3b44a38611511b7140/libcontainer/cgroups/fs/io_v2.go#L21

I assume "blkio" was adopted because `/proc/cgroups` shows "blkio" when the IO controller V2 is enabled.
However, it is called "io" in `/sys/fs/cgroup/cgroup.subtree_control`.

So maybe it makes more sense to let  `IOGroupV2.Name()` return "io".

AkihiroSuda:
`/proc/cgroups` is meaningless for v2 and should be ignored.

https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features

* Now `GetAllSubsystems()` parses `/sys/fs/cgroup/cgroup.controller`, not `/proc/cgroups`.
  The function result also contains "pseudo" controllers: `{"devices", "freezer"}`.
  As it is hard to detect availability of pseudo controllers, pseudo controllers are always assumed to be available.

* Now `IOGroupV2.Name()` returns `"io"`, not `"blkio"`.

Fix #2155 #2156

### DIFF
--- a/libcontainer/cgroups/fs/io_v2.go
+++ b/libcontainer/cgroups/fs/io_v2.go
@@ -17,12 +17,11 @@ type IOGroupV2 struct {
 }
 
 func (s *IOGroupV2) Name() string {
-	// for compatibility with v1 blkio controller
-	return "blkio"
+	return "io"
 }
 
 func (s *IOGroupV2) Apply(d *cgroupData) error {
-	_, err := d.join("blkio")
+	_, err := d.join("io")
 	if err != nil && !cgroups.IsNotFound(err) {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #2161, opened 2019-10-28
Makefile: allow overriding `docker` command

AkihiroSuda:
e.g. `make CONTAINER_ENGINE="sudo podman" unittest` (for ease of cgroup2 testing)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 	    test unittest integration \
 	    cross localcross
 
+CONTAINER_ENGINE := docker
 GO := go
 
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,6 +3,7 @@
 	    test unittest integration \
 	    cross localcross
 
+DOCKER := docker
```
rhatdan: Can't you call this 
CONTAINER_ENGINE := docker

crosbymichael: yes, a more generic name would be better

AkihiroSuda: CONTAINER_ENGINE seems too long?
CONTAINER?

AkihiroSuda: As we already have `DOCKER_RUN_PROXY`, `DOCKER` seems fine.

AkihiroSuda: WDYT?

h-vetinari: Finishing your sentence a bit differently:

> @AkihiroSuda: As we already have `DOCKER_RUN_PROXY`...

... longer names like `CONTAINER_ENGINE` seem fine. ;-)

IMO, `DOCKER` is misleading and `CONTAINER` is too ambiguous. I like `CONTAINER_ENGINE` (but it could easily be another `CONTAINER_xxx` name).

AkihiroSuda: updated to use `CONTAINER_ENGINE`

AkihiroSuda:
updated

rhatdan:
LGTM
@avagin @caniszczyk @crosbymichael @cyphar @dqminh @hqhq @mrunalp @rjnagal @vmarmol PTAL

mrunalp:
LGTM

crosbymichael:
LGTM

PR #2168, opened 2019-11-07
cgroup2: ebpf: increase RLIM_MEMLOCK to avoid BPF_PROG_LOAD error

### LINKED ISSUE #2167: cgroup2: procHooks: failed to load program: operation not permitted
Moby is getting support for cgroup v2: https://github.com/moby/moby/pull/40174 (https://github.com/moby/moby/tree/ee30a1f5ad7c373bd9db3f0b5412b03082f76786)

But it doesn't work with runc b133feaeeb2e69ba94aa95eac3a455a143435ea9 (works with crun v0.10.4)

```console
$ docker run -it --rm hello-world
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:415: setting cgroup config for procHooks process caused \\\"failed to load program: operation not permitted\\\"\"": unknown.
```

Podman+runc doesn't hit this issue.

AkihiroSuda:
Fix #2167

### DIFF
--- a/libcontainer/cgroups/ebpf/ebpf.go
+++ b/libcontainer/cgroups/ebpf/ebpf.go
@@ -16,6 +16,13 @@ func LoadAttachCgroupDeviceFilter(insts asm.Instructions, license string, dirFD 
 	nilCloser := func() error {
 		return nil
 	}
+	// Increase `ulimit -l` limit to avoid BPF_PROG_LOAD error (#2167).
+	// This limit is not inherited into the container.
+	memlockLimit := &unix.Rlimit{
+		Cur: unix.RLIM_INFINITY,
+		Max: unix.RLIM_INFINITY,
+	}
+	_ = unix.Setrlimit(unix.RLIMIT_MEMLOCK, memlockLimit)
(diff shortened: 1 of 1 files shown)


KentaTada:
@AkihiroSuda I'm sorry to interrupt but I think that rootless runc can only raise up to the hard limit. Is it enough?

FYI: we modified our internal systemd settings instead of modifying runc when I experienced the similar situation about rtprio. But I have no idea to support this feature other than your commit in general.

AkihiroSuda:
Anyway rootless cannot use eBPF device controller and the error is ignored

KentaTada:
>Anyway rootless cannot use eBPF device controller and the error is ignored
As you commented, rootless runc cannot use eBPF in the first place.
Thank you!

giuseppe:
LGTM

crosbymichael:
LGTM

AkihiroSuda:
@dqminh @hqhq @cyphar @mrunalp PTAL?

hqhq:
LGTM

PR #2172, opened 2019-11-26
Make event types public

saschagrunert:
The event types are now part of a dedicated public `types` package
within runc to be able to unmarshal the output `runc events` directly.

cc @mrunalp @cyphar


cyphar:
Tentative LGTM, though I'm not sure if @crosbymichael has a strong opinion on where such API stuff should go. But I agree that this is something we should expose to avoid duplication in `runc` users.

crosbymichael:
LGTM

cyphar:
LGTM.

PR #2177, opened 2019-12-06
libcontainer: export and add new methods to allow cgroups manipulation

### LINKED ISSUE #2176: libcontainer:  export and add new methods to allow cgroups manipulation
In order to manipulate cgroups v1 and v2 from other packages and applications, we need next changes:

* export `createCgroupConfig` to create cgroup configs
* add new method in cgroup Manager to get and save cgroups configuration to disk

cc @mrunalp @egernst

devimc:
libcontainer: export createCgroupConfig

A `config.Cgroups` object is required to manipulate cgroups v1 and v2 using
libcontainer.
Export `createCgroupConfig` to allow API users to create `config.Cgroups`
objects using directly libcontainer API.

libcontainer: add method to get cgroup config from cgroup Manager

`configs.Cgroup` contains the configuration used to create cgroups. This
configuration must be saved to disk, since it's required to restore the
cgroup manager that was used to create the cgroups.
Add method to get cgroup configuration from cgroup Manager to allow API users
save it to disk and restore a cgroup manager later.

fixes #2176

### DIFF
--- a/libcontainer/cgroups/cgroups.go
+++ b/libcontainer/cgroups/cgroups.go
@@ -46,6 +46,9 @@ type Manager interface {
 
 	// Sets the cgroup as configured.
 	Set(container *configs.Config) error
+
+	// Gets the cgroup as configured.
+	GetCgroups() (*configs.Cgroup, error)
 }
 
 type NotFoundError struct {
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -467,3 +467,7 @@ func CheckCpushares(path string, c uint64) error {
 
 	return nil
 }
+
+func (m *Manager) GetCgroups() (*configs.Cgroup, error) {
+	return m.Cgroups, nil
+}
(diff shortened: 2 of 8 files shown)


devimc:
ping @mrunalp

libcontainer/cgroups/cgroups.go
```diff
@@ -46,6 +46,9 @@ type Manager interface {
 
 	// Sets the cgroup as configured.
 	Set(container *configs.Config) error
+
+	// Gets the cgroup as configured.
+	GetCgroups() (*configs.Cgroup, error)
```
egernst: Hey @devimc:

1. nit: s/splications/applications on your commit message.
2. Can you better explain the intended use case, highlighting why this is needed? Specific to the Kata case (in both v2-shim and CLI based configurations), what do we do with the cgroup config?  Do we need to store a copy of the information, or can it be obtained as needed later? It just isn't clear to me. thx!

devimc: done, thanks

mrunalp:
LGTM

crosbymichael:
LGTM

PR #1698, opened 2018-01-22
Avoid race when opening exec fifo

craigfurman:
When starting a container with `runc start` or `runc run`, the stub
process (runc[2:INIT]) opens a fifo for writing. Its parent runc process
will open the same fifo for reading. In this way, they synchronize.

If the stub process exits at the wrong time, the parent runc process
will block forever.

This can happen when racing 2 runc operations against each other: `runc
run/start`, and `runc delete`. It could also happen for other reasons,
e.g. the kernel's OOM killer may select the stub process.

This commit resolves this race by racing the opening of the exec fifo
from the runc parent process against the stub process exiting. If the
stub process exits before we open the fifo, we return an error.

Another solution is to wait on the stub process. However, it seems it
would require more refactoring to avoid calling wait multiple times on
the same process, which is an error.

Note: We aren't really sure how to integration test this in a sane way. In Garden, we wrote a test but it involves patching in:

```
diff --git a/libcontainer/standard_init_linux.go b/libcontainer/standard_init_linux.go
index 8a544ed5..84cd8765 100644
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 	"syscall" //only for Exec
+	"time"
 
 	"github.com/opencontainers/runc/libcontainer/apparmor"
 	"github.com/opencontainers/runc/libcontainer/configs"
@@ -169,6 +171,9 @@ func (l *linuxStandardInit) Init() error {
 	// user process. We open it through /proc/self/fd/$fd, because the fd that
 	// was given to us was an O_PATH fd to the fifo itself. Linux allows us to
 	// re-open an O_PATH fd through /proc.
+	if !strings.Contains(name, "init") {
+		time.Sleep(time.Hour)
+	}
 	fd, err := unix.Open(fmt.Sprintf("/proc/self/fd/%d", l.fifoFd), unix.O_WRONLY|unix.O_CLOEXEC, 0)
 	if err != nil {
 		return newSystemErrorWithCause(err, "open exec fifo")
```

to expose the race condition, and then performing `runc run` and `runc delete` operations. Hopefully someone has a better idea of how to get a more sensible test into `runc`.

[Fixes: #1697]

Cheers,
@williammartin & Craig

PR #2185, opened 2019-12-18
Fix race checking for process exit and waiting for exec fifo

### LINKED ISSUE #2183: Race condition in exec() exists waiting for fifo from container process
Container failures are seen downstream in Kubernetes e2e runs when very short-lived containers are quickly created/completed. https://github.com/kubernetes/kubernetes/issues/86312

The following error is returned from the container runtime:
```
Terminated:&ContainerStateTerminated{ExitCode:128,Signal:0,Reason:ContainerCannotRun,Message:OCI runtime start failed: container process is already dead: unknown
```

Even though the container actually ran successfully, based on log output.

@random-liu [tracked this down](https://github.com/kubernetes/kubernetes/issues/86312#issuecomment-566722798) to a race condition in the code that checks if a process is dead/zombie while waiting to open the fifo successfully. A sequence of events that could cause this is:

1. process starts, opens fifo in write mode
2. awaitFifoOpen opens fifo in read mode
3. process completes
4. 100ms timeout in awaitProcessExit fires, stats pid, gets error, closes isDead channel
5. select block in exec() takes the awaitProcessExit branch and returns a "container process is already dead" error
6. goroutine in awaitFifoOpen propagates the open fifo to the fifoOpened channel

We only see this under fairly heavy load, in which normal expectations around speed and sequence of goroutine execution can be challenged.

liggitt:
Fixes #2183, a race condition in checking for a dead process while waiting for the exec fifo to open:

* Moves the wait for the blocking fifo open (happy path) and the 100ms poll for a missing/zombie process into the same loop
* When the missing/zombie process check fires, does a non-blocking open/read of the fifo to see if the process already completed, and only returns the "container process is already dead" error if that fails.

To test this, I inserted a synthetic 150ms delay in the awaitFifoOpen path after opening the fifo (simulating a slow send on the fifoOpened channel losing the race to the 100ms dead process detection timeout), and the integration suite passed (this PR includes https://github.com/opencontainers/runc/pull/2186 which fixes an existing race condition in one of the integration tests).

```diff
diff --git a/libcontainer/container_linux.go b/libcontainer/container_linux.go
index bec72def..a5a4d1ad 100644
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -301,6 +301,8 @@ func awaitFifoOpen(path string) <-chan openResult {
        fifoOpened := make(chan openResult)
        go func() {
                result := fifoOpen(path, true)
+               // simulate a slow send on fifoOpened, racing with the 100ms dead process detection timeout
+               time.Sleep(150 * time.Millisecond)
                fifoOpened <- result
        }()
        return fifoOpened
```

When I insert the same simulated race in master, 14 integration tests fail with "container is already dead" errors.

Still working out a way to unit/integration test this without modifying non-test code like that.

cc @random-liu

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -265,22 +265,24 @@ func (c *linuxContainer) Exec() error {
 
 func (c *linuxContainer) exec() error {
 	path := filepath.Join(c.root, execFifoFilename)
+	pid := c.initProcess.pid()
+	blockingFifoOpenCh := awaitFifoOpen(path)
+	for {
+		select {
+		case result := <-blockingFifoOpenCh:
+			return handleFifoResult(result)
 
-	fifoOpen := make(chan struct{})
-	select {
-	case <-awaitProcessExit(c.initProcess.pid(), fifoOpen):
-		return errors.New("container process is already dead")
-	case result := <-awaitFifoOpen(path):
-		close(fifoOpen)
-		if result.err != nil {
-			return result.err
+		case <-time.After(time.Millisecond * 100):
+			stat, err := system.Stat(pid)
+			if err != nil || stat.State == system.Zombie {
+				// could be because process started, ran, and completed between our 100ms timeout and our system.Stat() check.
+				// see if the fifo exists and has data (with a non-blocking open, which will succeed if the writing process is complete).
+				if err := handleFifoResult(fifoOpen(path, false)); err != nil {
+					return errors.New("container process is already dead")
+				}
+				return nil
+			}
 		}
-		f := result.file
-		defer f.Close()
-		if err := readFromExecFifo(f); err != nil {
-			return err
-		}
-		return os.Remove(path)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


liggitt:
hmm, `TestCheckpoint` is failing in three of the travis jobs:
```
--- FAIL: TestCheckpoint (0.52s)
    checkpoint_test.go:260: Did not restore the pipe correctly: 
```
but passes locally:
```
=== RUN   TestCheckpoint
--- PASS: TestCheckpoint (0.51s)
```

liggitt:
cc @crosbymichael @mrunalp as approvers on the introducing PR (https://github.com/opencontainers/runc/pull/1698)

mrunalp:
I am away this week and have only phone access. This looks fine from a quick look but would let other maintainers review it deeper. Thanks!

Random-Liu:
Adding a check after detecting the process is gone sounds good.

crosbymichael:
LGTM

liggitt:
cc @cyphar since @mrunalp is OOO

mrunalp:
LGTM

liggitt:
Thank you all for the reviews. I'm working on getting this propagated into containerd/docker versions we use for CI in Kubernetes, since our test scenarios trigger this bug frequently. Is there an existing timeframe for the next tagged version of runc, or is that done on an as-needed basis?

cyphar:
> Is there an existing timeframe for the next tagged version of runc, or is that done on an as-needed basis?

It's mostly done on an as-needed basis, but right now we are all waiting (with bated breath) for opencontainers/runtime-spec#1008 to be merged so that runc can finally have a `1.0` release and we can switch to a more periodic release schedule. If you need a `1.0-rc10` we can do that, but right now we need to also deal with CVE-2019-19921.

liggitt:
That context is helpful, thanks. I don't have a good sense for how much work https://github.com/opencontainers/runtime-spec/pull/1008 will require in this repo before a 1.0 tag could be cut, or from downstream consumers before that 1.0 tag could be consumed... if it's O(weeks), an earlier RC tag would be appreciated if possible.

cyphar:
It will definitely be O(weeks) -- especially given the time of year. In that case, I will first try to get a fix for CVE-2019-19921 merged first (probably closer to a week unless we have long arguments about the solution) and then do an `rc10` release.

liggitt:
Thanks, much appreciated

h-vetinari:
> @cyphar: It's mostly done on an as-needed basis, but right now we are all waiting (with bated breath) for opencontainers/runtime-spec#1008 to be merged so that runc can finally have a `1.0` release and we can switch to a more periodic release schedule. If you need a `1.0-rc10` we can do that, but right now we need to also deal with CVE-2019-19921.

Considering that it's taken 2 years to sort out the hooks (more or less), shouldn't there be an `1.0-rc11` with the new hooks (once landed in spec and runc), to weed out potential implementation errors, see how the ecosystem can move away from the now-deprecated hooks to the new ones, and get feedback from other runtime authors? What's 2 additional month in this whole saga? ;)

cyphar:
> shouldn't there be an 1.0-rc11 with the new hooks (once landed in spec and runc), to weed out potential implementation errors, see how the ecosystem can move away from the now-deprecated hooks to the new ones, and get feedback from other runtime authors?

Yeah, that's mostly likely what we'll end up doing.

> What's 2 additional month in this whole saga? ;)

:sob: :sob: :sob: :sob:

liggitt:
@cyphar any estimate on getting this in a tagged release?

liggitt:
from https://github.com/opencontainers/runc/pull/2185#issuecomment-569271362

>  I will first try to get a fix for CVE-2019-19921 merged first (probably closer to a week unless we have long arguments about the solution) and then do an rc10 release.

For anyone following along, it looks like that is being discussed/tracked in https://github.com/opencontainers/runc/issues/2197#issuecomment-570055971

libcontainer/container_linux.go
```diff
@@ -295,38 +297,39 @@ func readFromExecFifo(execFifo io.Reader) error {
...
+	}
+	return openResult{file: f}
+}
+
+func handleFifoResult(result openResult) error {
+	if result.err != nil {
+		return result.err
+	}
+	f := result.file
+	defer f.Close()
```
tedyu: I wonder if this deferred close would function correctly w.r.t. the os.Remove call a few lines down.
On windows, we cannot rename/delete/move a file with open file handles.

Should the code be refactored so that this works on all platforms ?

Thanks

liggitt: This is moving existing code that was previously on line 279. It is also explicitly in a file that it only runs on Linux.

tedyu: Should have placed my previous comment on the original code.

PR #2205, opened 2020-01-13
Don't deny all devices when update cgroup resource

### LINKED ISSUE #2204: Update command causes 'Operation not permitted'
It's first seen in a kubernetes cluster with docker as container runtime. Our users reported that in some situation their bash script failed with message `can't create /dev/null: Operation not permitted`. But `/dev/null` is default device with permission `rwm`, After digging some logs, we found that it can be reproduced in runc by following steps.

1. Run a runc container like "busybox". Suppose this container is called A
2. run `while true;do echo >/dev/null;done` in container
3. runc update --cpu-share 1024 A
4. You will see `sh: can't create /dev/null: Operation not permitted`

The problem is located at https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/devices.go#L28. By default, if runc wants to allow some devices, it'll first deny all devices then allow specific devices. It's OK when launching a new container, but if you update   the resource of a container, it'll make a change that allowed devices becomes denied first, then allowed, during this period, user app will get `xxx Operation not permitted` error message.

mYmNeo:
Resolve #2204.

Signed-off-by: mYmNeo <thomassong2012@gmail.com>

### DIFF
--- a/libcontainer/cgroups/fs/devices.go
+++ b/libcontainer/cgroups/fs/devices.go
@@ -3,6 +3,8 @@
 package fs
 
 import (
+	"strings"
+
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
 	"github.com/opencontainers/runc/libcontainer/configs"
@@ -11,6 +13,12 @@ import (
 
 type DevicesGroup struct {
 }
+
+type Empty struct{}
+
+var (
+	defaultDevice = &configs.Device{Type: 'a', Major: -1, Minor: -1, Permissions: "rwm"}
+)
 
 func (s *DevicesGroup) Name() string {
 	return "devices"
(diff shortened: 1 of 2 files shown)


libcontainer/cgroups/fs/devices.go
```diff
@@ -78,3 +119,22 @@ func (s *DevicesGroup) Remove(d *cgroupData) error {
 func (s *DevicesGroup) GetStats(path string, stats *cgroups.Stats) error {
 	return nil
 }
+
+func readDevicesExceptDefault(path string) (allowed map[string]Empty, err error) {
+	cgroupData, err := readFile(path, "devices.list")
```
giuseppe: the same check won't be possible with cgroup v2.

Would it be possible to detect whether it is updating an existing cgroup in another way and return early if `cgroup.Resources.Devices` is not defined?

mYmNeo: >

> the same check won't be possible with cgroup v2.
> 
> Would it be possible to detect whether it is updating an existing cgroup in another way and return early if `cgroup.Resources.Devices` is not defined?

Cgroup v2 will use `devices_v2.go` to set configuration, so `readDevicesExceptDefault` won't affect cgroup v2. I'll check if cgroup v2 has interface to query allowed devices.

For current design, I think the answer is no. Because the `update` function only set `cpu, disk or memory`, but cgroup manager only support `Set` interface. This will set and replace all subsystem result.

mYmNeo: @giuseppe I checked cgroup v2 and found that cgroup v2 will not raise `Operation not permitted when invoking `update` function. So I think this patch is only used for cgroup v1

AkihiroSuda:
@giuseppe PTAL again?

libcontainer/cgroups/fs/devices.go
```diff
@@ -32,29 +40,63 @@ func (s *DevicesGroup) Set(path string, cgroup *configs.Cgroup) error {
...
+			if len(file) > 0 {
+				if err := fscommon.WriteFile(path, file, dev.CgroupString()); err != nil {
+					return err
+				}
 			}
 		}
 		return nil
 	}
 	if cgroup.Resources.AllowAllDevices != nil {
 		if *cgroup.Resources.AllowAllDevices == false {
```
cyphar, line 69: This isn't directly related to this patch (and is something we should fix separately), but I've hit this wall while trying to review it.

I've spent a while now trying to understand what the hell this whole `AllowAllDevices`, `AllowedDevices` and `DeniedDevices` stuff is supposed to do and how it interacts with `Devices`. It's been around forever in runc, isn't exposed in the runtime-spec, and isn't even generated by specconv. It appears to be only used for unit testing itself, which seems like a senseless thing to test. Honestly, this patch and many others would be much nicer if we just nuked this from orbit and just had `Devices`.

In the context of this patch, I'm not even sure how to review whether the semantics of the below code are correct -- if I'm correct in my understanding that nothing uses the code, how do we know if the changes make sense...

@crosbymichael Am I missing something? Will something really bad happen if we get rid of it? I remember the `initProcessTime` nightmare when we changed the type of a field in the libcontainer state JSON (which resulted in the whole "migration" fun) -- would something similar happen if we just purged this concept from the codebase?

@AkihiroSuda WDYT? (Since you were the last person to touch it, and I think @crosbymichael and I last looked at this code in 2016.)

libcontainer/cgroups/fs/devices.go
```diff
@@ -32,29 +40,63 @@ func (s *DevicesGroup) Set(path string, cgroup *configs.Cgroup) error {
...
 			if dev.Allow {
 				file = "devices.allow"
 			}
-			if err := fscommon.WriteFile(path, file, dev.CgroupString()); err != nil {
-				return err
+
+			// For the second time set, we don't deny all devices
+			if dev.Type == defaultDevice.Type && len(oldAllowedDevices) != 0 {
+				file = ""
+			}
```
cyphar, line 58: I _really_ don't like this, and after looking at it for a bit I think I'm convinced it isn't right. This code will ignore any update that contains `a` -- but there are valid use-cases for `a` other than "deny everything" (you can allow both character and block devices for a particular major number, for instance). But even if this check was `dev == defaultDevice` it'd still be wrong since you can also allow all devices (though you shouldn't) -- and that would be ignored.

I think the issue is that the caller shouldn't be specifying a default deny rule in the list which we then need to filter out -- that should be done as part of this module. So we figure out what rules we need to specify.

libcontainer/cgroups/fs/devices.go
```diff
@@ -32,29 +40,63 @@ func (s *DevicesGroup) Set(path string, cgroup *configs.Cgroup) error {
...
+				}
 			}
 
-			for _, dev := range cgroup.Resources.AllowedDevices {
-				if err := fscommon.WriteFile(path, "devices.allow", dev.CgroupString()); err != nil {
-					return err
+			newAllowedDevices := make(map[string]Empty)
+			for _, dev := range cgroup.AllowedDevices {
+				newAllowedDevices[dev.CgroupString()] = Empty{}
+			}
```
cyphar, line 80: I think this should loop over `cgroup.Devices` given my previous point about how `AllowedDevices` is actually not used by anything. But as I said, most of this logic needs to be torched -- I just wrote a very long block of text about how this patch doesn't handle denying no longer allowed devices, only to realise that this code block actually is used, and in theory does handle that problem...

cyphar:
After looking at this patch and the cgroup code, I've decided I'm just going to do a big cleanup and include a fix for this problem as part of it.

mYmNeo:
> After looking at this patch and the cgroup code, I've decided I'm just going to do a big cleanup and include a fix for this problem as part of it. Thanks for the head-start @mYmNeo -- and my apologies, as this really isn't the best bit of the codebase to touch if you want to have a fun time.

Looking forward to the fix patch.

cyphar:
#2391 is the follow-up for this PR.

PR #2207, opened 2020-01-14
rootfs: do not permit /proc mounts to non-directories

### LINKED ISSUE #2197: [CVE-2019-19921]: Volume mount race condition with shared mounts
Disclosed in https://github.com/opencontainers/runc/pull/2190.

Here's the original report to security@opencontainers.org:

Hi all,

an attacker who controls the container image for two containers that share a volume can race volume mounts during container initialization, by adding a symlink to the rootfs that points to a directory on the volume. The second container won't be able to see the actual mount, but it can race it by modifying the mount point on the volume.

This can be exploited for a full container breakout by racing readonly/mask mounts, allowing writes to dangerous paths like `/proc/sys/kernel/core_pattern`.

Example:

- The rootfs of container A has a symlink `/proc` -> `/evil/level1`
- Container A specifies a named volume mounted to `/evil`
- Container B, started before container A, shares this named volume and repeatedly swaps `/evil/level1` and `/evil/level1~`
- Container A mounts procfs to `/evil/level1~/level2`, but when it remounts `/proc/sys`, it does so at `/evil/level1/level2/sys`.

This can reliably be reproduced using runc and podman on Fedora 30 (takes about 0-5s to win the race for me): https://gist.github.com/leoluk/82965ad9df58247202aa0e1878439092

SELinux would ordinarily prevent the exploit by disallowing `container_t` from writing `usermodehelper_t`, but it can be disabled by symlinking `/proc/self/task/1/attr/exec` to something benign like `/proc/self/sched` (bypassing the procfs check). AppArmor can be disabled similarly.

Docker specifies the mounts in a different order and mounts procfs after it mounts the volumes, mounting over the /proc symlink, which  appears to prevent at least the /proc approach. I haven't tested other runc usage scenarios, for instance, k8s+cri-o might be vulnerable as well.

Fabian of Cure53 (in CC) created a minimal PoC that uses runc directly: https://gist.github.com/LiveOverflow/c937820b688922eb127fb760ce06dab9 

There are other container init steps after the volume mount that can be raced, obvious ones being utils.CloseExecFrom and the AppArmor/SELinux attrs but there might be others, especially in mountToRootfs (like tricking remount into mounting the rootfs as rshared if there's another volume that specifies the flag, but I haven't tried that).

This is similar to the vulnerability I reported that Adam Iwaniuk disclosed during their Dragon Sector CTF (https://github.com/opencontainers/runc/issues/2128) and a similar crun one (https://github.com/containers/crun/issues/111).

The fix for the mounts is probably what Aleksa outlined here, using /proc/self/fd to resolve the path: https://github.com/containers/crun/issues/111#issuecomment-536495867

cyphar:
mount(2) will blindly follow symlinks, which is a problem because it
allows a malicious container to trick runc into mounting /proc to an
entirely different location (and thus within the attacker's control for
a rename-exchange attack).

This is just a hotfix (to "stop the bleeding"), and the more complete
fix would be finish libpathrs and port runc to it (to avoid these types
of attacks entirely, and defend against a variety of other /proc-related
attacks). It can be bypased by someone having "/" be a volume controlled
by another container.

Fixes: CVE-2019-19921
Fixes #2197

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -299,6 +299,18 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 
 	switch m.Device {
 	case "proc", "sysfs":
+		// If the destination already exists and is not a directory, we bail
+		// out This is to avoid mounting through a symlink or similar -- which
+		// has been a "fun" attack scenario in the past.
+		// TODO: This won't be necessary once we switch to libpathrs and we can
+		//       stop all of these symlink-exchange attacks.
+		if fi, err := os.Lstat(dest); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+		} else if fi.Mode()&os.ModeDir == 0 {
+			return fmt.Errorf("filesystem %q must be mounted on ordinary directory", m.Device)
+		}
 		if err := os.MkdirAll(dest, 0755); err != nil {
 			return err
 		}


libcontainer/rootfs_linux.go
```diff
@@ -299,6 +299,20 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 
 	switch m.Device {
 	case "proc", "sysfs":
+		// If the destination already exists and is not a directory, we remove
+		// it. This is to avoid mounting through a symlink or similar -- which
+		// has been a "fun" attack scenario in the past.
+		// TODO: This won't be necessary once we switch to libpathrs and we can
+		//       stop all of these symlink-exchange attacks.
+		if fi, err := os.Lstat(dest); err != nil {
```
leoluk: Wouldn't there still be a TOCTOU race here?

cyphar: Yes -- but exploiting it requires `/` to be mounted in another container (because the parent directory of `/proc` and `/sys` is `/`). I'm not happy with this compromise, but the alternative patch I posted (which also didn't work) had similar limitations.

leoluk: I'm pretty sure that no container runtime even allows sharing `/`.

cyphar: Well, `runc` allows it. :wink:

libcontainer/rootfs_linux.go
```diff
@@ -299,6 +299,20 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
...
+			if !os.IsNotExist(err) {
+				return err
+			}
+		} else if fi.Mode()&os.ModeDir == 0 {
+			if err := os.Remove(dest); err != nil {
```
mrunalp: Instead of removing, should we just bail out and fail?

cyphar: Yeah that might be a better idea.

cyphar: Fixed.

giuseppe:
@cyphar I've added to crun an extra check to allow `procfs` and `sysfs` to be mounted only under the rootfs, so no `/foo/proc` to prevent the mount to happen on a volume where another container has `CAP_SYS_ADMIN` access (hopefully unlikely) and could do a `mount(MS_MOVE)` between the mount setup and the `MS_RDONLY`

cyphar:
Even given how crazy `VOLUME` is, I don't think you can trick Docker (or other such tools) into mounting `procfs` on `/foo/proc`. Also these are temporary restrictions, there's really no need to over-engineer them -- I want to drop this the second `runc` is ported to `libpathrs`.

Not to mention that if someone decides to create a container with `CAP_SYS_ADMIN` then they are running an insecure setup and it isn't our job to help protect them at that point -- there is literally nothing we can do to protect someone in that case (because the container can do many more malicious things than just `mount(MS_MOVE)`). If you have full-on `CAP_SYS_ADMIN` you can just re-mount anything as read-write.

giuseppe:
>  If you have full-on `CAP_SYS_ADMIN` you can just re-mount anything as read-write. I'm not sure that the protection you mentioned actually addresses the "you have a container with `CAP_SYS_ADMIN`" threat model.

true, I was thinking of a way to make the writeable `/proc` persistent `RW` when mounted on a shared volume, but I don't see a way without requiring a weird configuration (if possible at all with seccomp) to allow `mount(MS_MOVE)` but not allowing a remount/umount.

But even without `CAP_SYS_ADMIN`: if you can mount `/proc` on a shared bind mount, there will still be the possibility for a process in another container to open a file before the `/proc` mount is remounted read only.  Allowing `/proc` just in the rootfs limits the attack as the rootfs itself must be accessible from the other malicious container.

cyphar:
I see your point, but IMHO the threat model for OCI runtimes is not an attacker being able to control the configuration (because in that case, it's game over no matter what you do). The fact that higher-level runtimes allow for this (such as Docker's `VOLUME`) is something we have to deal with anyway, but I'd prefer to avoid giving anyone the impression that we can solve the general (or even semi-general) case here.

And while it isn't possible (AFAICS) to trick Docker into creating a `/foo/proc` mount for procfs (in such a way that it isn't also possible to mount anything), it is **very** possible to get Docker to make `/` a volume -- which as I mentioned above (and in the other issue) allows you to bypass all of these checks entirely (including the one you described).

cyphar:
Then again, all of our masked-path logic is very much tied to the actual path (not the filesystem) so there is an argument for making `procfs` only mountable on `/proc`. My main disagreement is that I don't think that change helps protect against the threat model it's trying to protect against -- because if you can provide a way to mount procfs to a custom path there are plenty of other ways of causing trouble.

hqhq:
LGTM

mrunalp:
LGTM

thaJeztah:
Will this be tagged as `v1.0.0-rc10` ?

thaJeztah:
oh, nevermind; I see the e-mail thread 👍

PR #2210, opened 2020-01-15
Exchange deprecated systemd resources with the appropriate for cgroupv2

### LINKED ISSUE #2164: cgroup2: systemd manager uses deprecated options
https://github.com/opencontainers/runc/blob/e57a774066af2f0adc76ffb6201588091cd872d5/libcontainer/cgroups/systemd/unified_hierarchy.go#L99-L132

According to [`systemd.resource-control(5)`](https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html), `CPUShares`, `MemoryLimit`, and `BlockIOWeight` are deprecated

Zyqsempai:
Fix #2164 

Exchanged deprecated systemd resources with the appropriate for cgroupv2

Signed-off-by: Boris Popovschi <zyqsempai@mail.ru>

### DIFF
--- a/libcontainer/cgroups/systemd/unified_hierarchy.go
+++ b/libcontainer/cgroups/systemd/unified_hierarchy.go
@@ -81,7 +81,7 @@ func (m *UnifiedManager) Apply(pid int) error {
 	properties = append(properties,
 		newProp("MemoryAccounting", true),
 		newProp("CPUAccounting", true),
-		newProp("BlockIOAccounting", true))
+		newProp("IOAccounting", true))
 
 	// Assume DefaultDependencies= will always work (the check for it was previously broken.)
 	properties = append(properties,
@@ -89,12 +89,12 @@ func (m *UnifiedManager) Apply(pid int) error {
 
 	if c.Resources.Memory != 0 {
 		properties = append(properties,
-			newProp("MemoryLimit", uint64(c.Resources.Memory)))
+			newProp("MemoryMax", uint64(c.Resources.Memory)))
 	}
 
-	if c.Resources.CpuShares != 0 {
+	if c.Resources.CpuWeight != 0 {
 		properties = append(properties,
-			newProp("CPUShares", c.Resources.CpuShares))
+			newProp("CPUWeight", c.Resources.CpuWeight))
 	}
 
 	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -117,11 +117,6 @@ func (m *UnifiedManager) Apply(pid int) error {
 			newProp("CPUQuotaPerSecUSec", cpuQuotaPerSecUSec))
 	}
 
-	if c.Resources.BlkioWeight != 0 {
-		properties = append(properties,
-			newProp("BlockIOWeight", uint64(c.Resources.BlkioWeight)))
```
AkihiroSuda: Do we have modern equivalent for this?

Zyqsempai: Yes, it's `IOWeight` but I tested it and faced some strange error `no such device or address`, so I will continue investigation related to this error, but until that we can merge this part.

AkihiroSuda: If systemd can translate the legacy value, let's keep the legacy code for IO.

mrunalp:
LGTM

AkihiroSuda:
@filbranden @giuseppe PTAL?

Zyqsempai:
@filbranden @giuseppe Ping

giuseppe:
LGTM but I am not a maintainer

filbranden:
Hey!

I haven't had time to look deeply into it. A problem in the past was whether older versions of systemd would have problems with this change.

I guess if you can test this on something like RHEL 7 / CentOS 7 (systemd v219 + some backports) to ensure this won't break on a systemd that old, that would be great!

You mentioned cgroupv2, so CentOS 7 doesn't really support cgroupv2, so it's not relevant there. (I haven't looked deeply into the code changes to check whether this affects cgroupv1 or cgroupv2 only.)

Having said that, otherwise, LGTM. As long as those issues are not of concern, you should be fine! Thanks for this!

Cheers,
Filipe

Zyqsempai:
@hqhq @crosbymichael PTAL and let's merge it;)

Zyqsempai:
@cyphar PTAL here also, it should be merged long time ago;)

Zyqsempai:
@mrunalp @dqminh maybe you can take a look at this one?

mrunalp:
LGTM

hqhq:
LGTM

PR #2217, opened 2020-01-22
VERSION: release 1.0.0~rc10

cyphar:


hqhq:
LGTM

dims:
Yes Please! eagerly waiting for the fix in https://github.com/opencontainers/runc/pull/2185

cyphar:

Since the vote already passed, I've pushed [the release tag](https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc10). It's quite silly that we need two LGTMs for releases, since our review process for releases is so formal...

cyphar:
@dims For future reference, I would suggest not including a reference to a PR inside a commit message -- GitHub has a (mis)feature where it cross-links commits that reference PRs but this means each time you force push it results in a bunch of annoying spam being included in the history of a PR or issue. Not your fault, it's because GitHub handles this poorly -- but I'd recommend in future that you only reference the PR inside the relevant PR being sent to upstream (GitHub handles that case correctly -- and only provides one cross-reference for the entire PR).

dims:
@cyphar good point. will avoid that. Looks like even when i don't use a full github url in the git commit, github still figures it out and causes spam ( https://github.com/containerd/containerd/commit/0db3c9b78007f1c7fab5d358a95f3dd086258251.patch )

mrunalp:
LGTM

PR #2218, opened 2020-01-24
fix(man): build without which

CameronNemo:


cyphar:
I don't really mind this change, but I'm wondering what systems don't have `which`?

CameronNemo:
xbps-src uses a minimal rootfs to perform builds in. for whatever reason, the void devs decided to drop which as a default installed package in those environments. It is still part of the base system for real installed systems / regular rootfs's though.

AkihiroSuda:
Covered in https://github.com/opencontainers/runc/pull/2228

PR #2222, opened 2020-02-05
travis: update configuration

cyphar:
Update the set of Go versions (and use 1.x to always test the latest
release), as well as making the cgroupv2 tests allowable failures (the
vagrant setup seems to break pretty often, causing flaky failures).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,22 @@
 dist: bionic
 language: go
 go:
-  - 1.11.x
   - 1.12.x
+  - 1.13.x
+  - 1.x
   - tip
 
 matrix:
   include:
     - go: 1.12.x
+      name: "cgroup-systemd"
       env:
         - RUNC_USE_SYSTEMD=1
       script:
         - make BUILDTAGS="${BUILDTAGS}" all
         - sudo PATH="$PATH" make localintegration RUNC_USE_SYSTEMD=1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,6 +32,7 @@ matrix:
         - ssh default sudo podman run --privileged --cgroupns=private test make localunittest
   allow_failures:
     - go: tip
+    - name: "cgroup-v2"
```
AkihiroSuda: can we add retry code to the v2 setup script instead?

cyphar: Maybe, but the v2 tests also take the longest to run -- I'm not a huge fan of the idea of having possibly-regular >1hr test runs.

h-vetinari: > @cyphar: I'm not a huge fan of the idea of having possibly-regular >1hr test runs.

Compared to how little the CI is exercised here, that seems like an acceptable trade-off for not breaking things...? "Acceptable failures" often quickly degenerate unless someone pays close attention to them. Wouldn't it be better to try to figure out what's causing the flakes and fixing that?

AkihiroSuda:
LGTM

AkihiroSuda:
LGTM

AkihiroSuda:
@mrunalp PTAL?

crosbymichael:
LGTM

PR #2224, opened 2020-02-07
Allow to set systemd scope properties via annotations

kolyshkin:
In case systemd is used as a cgroups driver for container, it creates a scope
(aka transient unit) dedicated to a container (usually named `runc-$ID.scope`). 

This PR adds an ability to set arbitrary properties for the systemd unit via
runtime spec annotations.

### Usage

Add the following to runtime spec (`config.json`):
    
```json
        "annotations": {
                "org.systemd.property.TimeoutStopSec": "300",
                "org.systemd.property.CollectMode":"'inactive-or-failed'"
        },
```

and start the container (e.g. `runc --systemd-cgroup run $ID`).

The above will set the following systemd parameters:
* `TimeoutStopSec` to 2 minutes and 3 seconds;
* `CollectMode` to "inactive-or-failed".
    
The values in spec should be in the gvariant format ([see docs here](https://developer.gnome.org/glib/stable/gvariant-text.html)).

To figure out which type systemd expects for a particular parameter,
please consult systemd sources.

### TODO
- [x] unit tests
- [x] documentation
- [x] allow *Sec parameters, simplify doc

### Notes
- This is needed to fix kubernetes/kubernetes#77873
- This is a carry of #2062, please see the initial discussion in there

### History
* v1: original version from #2062
* v2: use `map[string]interface{}` for `SystemdProperties` as suggesed in https://github.com/opencontainers/runc/pull/2062#discussion_r295835117, generalize cli options parsing
* v3: move `SystemdProperties` into `configs.Cgroup` to simplify code
* v4: cli option removed, setting redone via annotations in a generic fashion
* v5: allow time units with Sec suffix (in addition to USec)

### DIFF
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -1,5 +1,9 @@
 package configs
 
+import (
+	systemdDbus "github.com/coreos/go-systemd/dbus"
+)
+
 type FreezerState string
 
 const (
@@ -29,6 +33,11 @@ type Cgroup struct {
 
 	// Resources contains various cgroups settings to apply
 	*Resources
+
+	// SystemdProps are any additional properties for systemd,
+	// derived from org.systemd.property.xxx annotations.
+	// Ignored unless systemd is used for managing cgroups.
+	SystemdProps []systemdDbus.Property `json:"-"`
 }
 
 type Resources struct {
--- a/docs/systemd-properties.md
+++ b/docs/systemd-properties.md
@@ -1,0 +1,34 @@
+## Changing systemd unit properties
+
+In case runc uses systemd to set cgroup parameters for a container (i.e.
+`--systemd-cgroup` CLI flag is set), systemd creates a scope (a.k.a.
+transient unit) for the container, usually named like `runc-$ID.scope`.
+
+The systemd properties of this unit (shown by `systemctl show runc-$ID.scope`
+after the container is started) can be modified by adding annotations
+to container's runtime spec (`config.json`). For example:
+
+```json
+        "annotations": {
+                "org.systemd.property.TimeoutStopUSec": "uint64 123456789",
+                "org.systemd.property.CollectMode":"'inactive-or-failed'"
+        },
+```
+
+The above will set the following properties:
+
+* `TimeoutStopSec` to 2 minutes and 3 seconds;
+* `CollectMode` to "inactive-or-failed".
+
+The values must be in the gvariant format (for details, see
+[gvariant documentation](https://developer.gnome.org/glib/stable/gvariant-text.html)).
+
+To find out which type systemd expects for a particular parameter, please
+consult systemd sources. In particular, parameters with `USec` suffix are
+in microseconds, and those require an `uint64` typed argument. Since
+gvariant assumes int32 for a numeric values, the explicit type is required.
+
+**Note** that time-typed systemd parameter names must have the `USec`
+suffix, while they are documented with `Sec` suffix.
+For example, the stop timeout used in the example above must be
+set as `TimeoutStopUSec` but is shown and documented as `TimeoutStopSec`.
--- a/docs/systemd-properties.md
+++ b/docs/systemd-properties.md
@@ -24,11 +24,4 @@ The values must be in the gvariant format (for details, see
 [gvariant documentation](https://developer.gnome.org/glib/stable/gvariant-text.html)).
 
 To find out which type systemd expects for a particular parameter, please
-consult systemd sources. In particular, parameters with `USec` suffix are
-in microseconds, and those require an `uint64` typed argument. Since
-gvariant assumes int32 for a numeric values, the explicit type is required.
-
-**Note** that time-typed systemd parameter names must have the `USec`
-suffix, while they are documented with `Sec` suffix.
-For example, the stop timeout used in the example above must be
-set as `TimeoutStopUSec` but is shown and documented as `TimeoutStopSec`.
+consult systemd sources.
--- a/libcontainer/cgroups/systemd/apply_systemd.go
+++ b/libcontainer/cgroups/systemd/apply_systemd.go
@@ -245,6 +245,8 @@ func (m *LegacyManager) Apply(pid int) error {
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


utils_linux.go
```diff
@@ -28,6 +29,48 @@ import (
 
 var errEmptyID = errors.New("container id cannot be empty")
 
+var cliSystemdProperties = map[string]struct {
+	Name string // systemd property name
+	conv func(name, val string) (interface{}, error)
+}{
+	"stop-timeout": {"TimeoutStopUSec", convertUSec},
```
mrunalp: We don't want to explicitly add each property. We want to support arbitrary system properties. For e.g. `--systemd-property TimeoutStopUSec=1000`. This was requested on the original PR.

kolyshkin: Initially I also wanted to make it generic, but it seems it is both impossible and useless to support arbitrary systemd properties. Let me explain.

It is impossible because we do not know types of arguments. I know that systemd dbus uses at least booleans, strings, various types of integers, and even arrays. I don't see a way to pass the argument (to systemd dbus) without knowing its type.

It is useless because a limited amount of properties are supported for `StartTransientUnit()` systemd scope. Here's a quote from the docs (that might be obsoleted, but still):

> The properties array of StartTransientUnit() may take many of the settings that may also be configured in unit files. Not all parameters are currently accepted though, but we plan to cover more properties with future release. Currently you may set the Description, Slice and all dependency types of units, as well as RemainAfterExit, ExecStart for service units, TimeoutStopUSec and PIDs for scope units, and CPUAccounting, CPUShares, BlockIOAccounting, BlockIOWeight, BlockIOReadBandwidth, BlockIOWriteBandwidth, BlockIODeviceWeight, MemoryAccounting, MemoryLimit, DevicePolicy, DeviceAllow for services/scopes/slices. These fields map directly to their counterparts in unit files and as normal D-Bus object properties. The exception here is the PIDs field of scope units which is used for construction of the scope only and specifies the initial PIDs to add to the scope object.

From the above list, runc already sets `Slice`, `Description`, `DefaultDependencies`, and `PIDs`, as well as (I assume) all the cgroup parameters. Let's see what is left to set for a scope unit...

It seems the only property left is `TimeoutStopUSec`, there's no other option we can set.

Still, the doc can be obsoleted, or I might be wrong, so I modified the code in a way that allows us to add an option easily (in case of usec option that would be a line of code, in case of option with other type of arg you'd need to provide a validator+converter function, one for each type).

giuseppe: > We don't want to explicitly add each property

chould it be an annotation?

These runc-only features make interoperability with other OCI runtimes more difficult.  If it is an annotation, other runtimes that have no support for it (including older runc versions) will ignore the timeout.

giuseppe:
> We don't want to explicitly add each property. 

could it be an annotation?

These runc-only features make interoperability with other OCI runtimes more difficult.  If it is an annotation, runtimes that have no support for it (including older runc versions) will just ignore the timeout.

utils_linux.go
kolyshkin: OK the (seemingly up-to-date) list of properties available for transient units is here: https://github.com/systemd/systemd/blob/master/docs/TRANSIENT-SETTINGS.md

kolyshkin:
> could it be an annotation?
> 
> These runc-only features make interoperability with other OCI runtimes more difficult. If it is an annotation, runtimes that have no support for it (including older runc versions) will just ignore the timeout.

@giuseppe Do you have something like this in mind?

```json
    "annotations": {
        "org.systemd.property.TimeoutStopUSec": "x:120000000",
        "org.systemd.property.CollectMode": "s:inactive-or-failed"
    }
```

(here `x` and `s` are dbus type signatures (`x` is uint64, `s` is string). We can defnintely add some translation to improve the interface here.

kolyshkin:
Ended up having
```json
	"annotations": {
		"org.systemd.properties": "TimeoutStopUSec=uint64 123456789, CollectMode='inactive-or-failed', "
	},

```

libcontainer/configs/cgroup_linux.go
```diff
@@ -29,6 +33,11 @@ type Cgroup struct {
 
 	// Resources contains various cgroups settings to apply
 	*Resources
+
+	// SystemdProps are any additional properties for systemd,
+	// derived from org.systemd.properties annotations.
+	// Ignored unless systemd is used for managing cgroups.
+	SystemdProps []systemdDbus.Property `json:"-"`
```
kolyshkin: I don't like this type being used here :-\

kolyshkin:
Pushed the v4, updated the description accordingly.
@mrunalp @giuseppe PTAL

giuseppe:
> > 
> 
> @giuseppe Do you have something like this in mind?
> 
> ```json
>     "annotations": {
>         "org.systemd.property.TimeoutStopUSec": "x:120000000",
>         "org.systemd.property.CollectMode": "s:inactive-or-failed"
>     }
> ```

yes thanks.  Could it be possible to use the annotation as in the example above instead of having a comma separed list?

kolyshkin:
> Could it be possible to use the annotation as in the example above instead of having a comma separed list?

You're right, it's much better that way (although we have to iterate through all annotations).

Patch and description updated.

kolyshkin:
unit tests for `initSystemdProps()` added

kolyshkin:
Docs added, please see https://github.com/opencontainers/runc/blob/c6f5a5033af4678c6ed2586a1410face70ec7dd7/docs/systemd-properties.md

mrunalp:
LGTM

mrunalp:

kolyshkin:
I have added a way to specify time units in seconds (with their names as described in systemd documentation).

Not sure if this makes sense or not, since the implementation is a bit more lengthy than I thought. Anyway, it's a separate commit and can be removed.

Related systemd issue: https://github.com/systemd/systemd/issues/2047

kolyshkin:
Rebased; description updated; @opencontainers/runc-maintainers PTAL

kolyshkin:
@mrunalp @crosbymichael PTAL

mrunalp:
LGTM

hqhq:
LGTM

Issue #2196, opened 2019-12-31
CRIU tests are failing on CI


AkihiroSuda:
```
=== RUN   TestUsernsCheckpoint
--- FAIL: TestUsernsCheckpoint (0.57s)
    checkpoint_test.go:260: Did not restore the pipe correctly: 
=== RUN   TestCheckpoint
--- FAIL: TestCheckpoint (0.49s)
    checkpoint_test.go:260: Did not restore the pipe correctly: 
```

Seems Travis recently updated the kernel to an incompatible version: https://github.com/containerd/containerd/pull/3898

AkihiroSuda:
tests are temporarily disabled in #2198

PR #2226, opened 2020-02-10
restore: fix a race condition in process.Wait()

avagin:
@adrianreber  reported that the checkpoint test stated failing:
```
=== RUN   TestCheckpoint
--- FAIL: TestCheckpoint (0.38s)
    checkpoint_test.go:297: Did not restore the pipe correctly:
```

The problem here is when we start exec.Cmd, we don't call its wait
method. This means that we don't wait for cmd.goroutines and so we don't
know when all data will be read from process pipes.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1485,17 +1485,19 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 		return err
 	}
 	criuServer.Close()
+	// cmd.Process will be replaced by a restored init.
+	criuProcess := cmd.Process
 
 	defer func() {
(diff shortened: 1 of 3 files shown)


adrianreber:
@avagin Thanks for this fix. I have also prepared a few changes to make runc work with CRIU on Fedora 31, which I would submit after this here is merged.

It seems that #2196, which saw CRIU test failures, is not related to the Ubuntu kernel overlay bug after all but is actually fixed with this PR.

AkihiroSuda:
LGTM

avagin:
@crosbymichael @cyphar @mrunalp Hi! Maybe one of you can review this? Thanks!

mrunalp:
LGTM

PR #2228, opened 2020-02-12
Use "command -v" shell builtin instead of "which"

cpuguy83:
Took me awhile to figure out why it wasn't finding my go-md2man... centos image doesn't have `which` in it.
So this is just a little more convenient and one less required build dep.

### DIFF
--- a/man/md2man-all.sh
+++ b/man/md2man-all.sh
@@ -9,7 +9,7 @@ cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 	pwd
 }
 
-if ! ( which go-md2man &>/dev/null ); then
+if ! type go-md2man; then
 	echo "To install man pages, please install 'go-md2man'."
 	exit 0
 fi


mrunalp:
LGTM

mrueg:
See: https://github.com/opencontainers/runc/pull/2218

thaJeztah:
yes, looks like a duplicate of https://github.com/opencontainers/runc/pull/2218

man/md2man-all.sh
```diff
@@ -9,7 +9,7 @@ cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 	pwd
 }
 
-if ! ( which go-md2man &>/dev/null ); then
+if ! ( command -v go-md2man &>/dev/null ); then
```
kolyshkin: 1. I do not understand the need for a sub-shell here.

2. Maybe just use `type` and do not redirect stderr to /dev/null? This was the reason for the initial confusion :)

So, something like
```shell
if ! type go-md2man >/dev/null; then
...
```

kolyshkin:
> yes, looks like a duplicate of #2218

this one has better description

man/md2man-all.sh
AkihiroSuda: ping @cpuguy83

cpuguy83: Updated to not use a subshell.

AkihiroSuda:
LGTM

AkihiroSuda:
LGTM

AkihiroSuda:
I assume Mrunal's LGTM is still effective after rebase. Merging.

PR #2229, opened 2020-02-18
Add CreateRuntime, CreateContainer and StartContainer Hooks

### LINKED ISSUE #1710: 'prestart/poststart' hooks are done in the 'create' operation

RenaudWasTaken:
Hello!

This PR implements the new `CreateRuntime` hook as defined in https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks

This is a fairly simple change, `CreateRuntime` are basically `prestart` hooks with the requirement of being called after the now deprecated `prestart` hooks.

I've done of refactoring to not have too much duplication of code by adding a function: `(*Hooks) RunHooks(name HookName, spec *specs.State) error`.
This allows us to have a nicer statement in the code:
```golang
if err := c.config.Hooks.RunHooks(configs.Prestart, s); err != nil {
    return err
}

// Rather than
for i, hook := range c.config.Hooks.Prestart {
    if err := hook.Run(s); err != nil {
        return newSystemErrorWithCausef(err, "running prestart hook %d", i)
    }
}
```

I'm looking into how to implement the `createContainer` and `startContainer` changes but am not super familiar with the code base and would happily take pointers.

fixes #1710
cc @cyphar @mrunalp @vbatts

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,6 +67,19 @@ RUN mkdir -p /usr/src/criu \
     && cd - \
     && rm -rf /usr/src/criu
 
+# install skopeo
+RUN echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list \
+    && wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_Unstable/Release.key -O- | sudo apt-key add - \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends skopeo \
+    && rm -rf /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list \
+    && apt-get clean \
+    && rm -rf /var/cache/apt /var/lib/apt/lists/*;
+
+# install umoci
+RUN curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64 \
+    && chmod +x /usr/local/bin/umoci
+
 COPY script/tmpmount /
 WORKDIR /go/src/github.com/opencontainers/runc
 ENTRYPOINT ["/tmpmount"]
@@ -77,5 +90,10 @@ ENV ROOTFS /busybox
 RUN mkdir -p "${ROOTFS}"
 RUN . tests/integration/multi-arch.bash \
     && curl -fsSL `get_busybox` | tar xfJC - "${ROOTFS}"
+
+ENV DEBIAN_ROOTFS /debian
+RUN mkdir -p "${DEBIAN_ROOTFS}"
+RUN . tests/integration/multi-arch.bash \
+    && get_and_extract_debian "$DEBIAN_ROOTFS"
 
 COPY . .
--- a/Vagrantfile.centos7
+++ b/Vagrantfile.centos7
@@ -18,11 +18,15 @@ Vagrant.configure("2") do |config|
 
     # install yum packages
     yum install -y -q epel-release
-    yum install -y -q gcc git iptables jq libseccomp-devel make
+    yum install -y -q gcc git iptables jq libseccomp-devel make skopeo
     yum clean all
 
     # install Go
     curl -fsSL "https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz" | tar Cxz /usr/local
+
+    # Install umoci
+    curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64
+    chmod +x /usr/local/bin/umoci
 
     # install bats
     git clone https://github.com/bats-core/bats-core
--- a/Vagrantfile.fedora32
+++ b/Vagrantfile.fedora32
@@ -17,7 +17,7 @@ Vagrant.configure("2") do |config|
 config exclude kernel,kernel-core
 config install_weak_deps false
 update
-install iptables gcc make golang-go libseccomp-devel bats jq git-core criu
+install iptables gcc make golang-go libseccomp-devel bats jq git-core criu skopeo
 ts run
 EOF
     dnf clean all
@@ -31,10 +31,15 @@ EOF
     cat /root/rootless.key.pub >> /home/rootless/.ssh/authorized_keys
     chown -R rootless.rootless /home/rootless
 
+    # Install umoci
+    curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64
+    chmod +x /usr/local/bin/umoci
+
     # Add busybox for libcontainer/integration tests
     . /vagrant/tests/integration/multi-arch.bash \
-        && mkdir /busybox \
-        && curl -fsSL $(get_busybox) | tar xfJC - /busybox
+        && mkdir /busybox /debian \
+        && curl -fsSL $(get_busybox) | tar xfJC - /busybox \
+        && get_and_extract_debian /debian
 
     # Delegate cgroup v2 controllers to rootless user via --systemd-cgroup
     mkdir -p /etc/systemd/system/user@.service.d
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -239,10 +239,6 @@ const (
 	Poststop = "poststop"
 )
 
-var (
-	HookNameList = []HookName{Prestart, CreateRuntime, CreateContainer, StartContainer, Poststart, Poststop}
-)
-
 // TODO move this to runtime-spec
 // See: https://github.com/opencontainers/runtime-spec/pull/1046
 const (
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
-
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
@@ -176,7 +176,7 @@ type Config struct {
 
 	// Hooks are a collection of actions to perform at various container lifecycle events.
 	// CommandHooks are serialized to JSON, but other hooks are not.
-	Hooks *Hooks
+	Hooks Hooks
 
 	// Version is the version of opencontainer specification that is supported.
 	Version string `json:"version"`
@@ -203,17 +203,54 @@ type Config struct {
 	RootlessCgroups bool `json:"rootless_cgroups,omitempty"`
 }
 
-type Hooks struct {
+type HookName string
+type HookList []Hook
+type Hooks map[HookName]HookList
+
+const (
 	// Prestart commands are executed after the container namespaces are created,
 	// but before the user supplied command is executed from init.
-	Prestart []Hook
+	// Note: This hook is now deprecated
+	// Prestart commands are called in the Runtime namespace.
+	Prestart HookName = "prestart"
+
+	// CreateRuntime commands MUST be called as part of the create operation after
+	// the runtime environment has been created but before the pivot_root has been executed.
+	// CreateRuntime is called immediately after the deprecated Prestart hook.
+	// CreateRuntime commands are called in the Runtime Namespace.
+	CreateRuntime = "createRuntime"
+
+	// CreateContainer commands MUST be called as part of the create operation after
+	// the runtime environment has been created but before the pivot_root has been executed.
+	// CreateContainer commands are called in the Container namespace.
+	CreateContainer = "createContainer"
+
+	// StartContainer commands MUST be called as part of the start operation and before
+	// the container process is started.
+	// StartContainer commands are called in the Container namespace.
+	StartContainer = "startContainer"
 
 	// Poststart commands are executed after the container init process starts.
-	Poststart []Hook
+	// Poststart commands are called in the Runtime Namespace.
+	Poststart = "poststart"
 
 	// Poststop commands are executed after the container init process exits.
-	Poststop []Hook
-}
+	// Poststop commands are called in the Runtime Namespace.
+	Poststop = "poststop"
+)
+
+var (
+	HookNameList = []HookName{Prestart, CreateRuntime, CreateContainer, StartContainer, Poststart, Poststop}
+)
+
+// TODO move this to runtime-spec
+// See: https://github.com/opencontainers/runtime-spec/pull/1046
+const (
+	Creating = "creating"
+	Created  = "created"
+	Running  = "running"
+	Stopped  = "stopped"
+)
 
 type Capabilities struct {
 	// Bounding is the set of capabilities checked by the kernel.
@@ -228,32 +265,39 @@ type Capabilities struct {
 	Ambient []string
 }
 
-func (hooks *Hooks) UnmarshalJSON(b []byte) error {
-	var state struct {
-		Prestart  []CommandHook
-		Poststart []CommandHook
-		Poststop  []CommandHook
+func (hooks HookList) RunHooks(state *specs.State) error {
+	for i, h := range hooks {
+		if err := h.Run(state); err != nil {
+			return errors.Wrapf(err, "Running hook #%d:", i)
+		}
 	}
+
+	return nil
+}
+
+func (hooks *Hooks) UnmarshalJSON(b []byte) error {
+	var state map[HookName][]CommandHook
 
 	if err := json.Unmarshal(b, &state); err != nil {
 		return err
 	}
 
-	deserialize := func(shooks []CommandHook) (hooks []Hook) {
-		for _, shook := range shooks {
-			hooks = append(hooks, shook)
+	*hooks = Hooks{}
+	for n, commandHooks := range state {
+		if len(commandHooks) == 0 {
+			continue
 		}
 
-		return hooks
+		(*hooks)[n] = HookList{}
+		for _, h := range commandHooks {
+			(*hooks)[n] = append((*hooks)[n], h)
+		}
 	}
 
-	hooks.Prestart = deserialize(state.Prestart)
-	hooks.Poststart = deserialize(state.Poststart)
-	hooks.Poststop = deserialize(state.Poststop)
 	return nil
 }
 
-func (hooks Hooks) MarshalJSON() ([]byte, error) {
+func (hooks *Hooks) MarshalJSON() ([]byte, error) {
 	serialize := func(hooks []Hook) (serializableHooks []CommandHook) {
 		for _, hook := range hooks {
 			switch chook := hook.(type) {
@@ -268,9 +312,12 @@ func (hooks Hooks) MarshalJSON() ([]byte, error) {
 	}
 
 	return json.Marshal(map[string]interface{}{
-		"prestart":  serialize(hooks.Prestart),
-		"poststart": serialize(hooks.Poststart),
-		"poststop":  serialize(hooks.Poststop),
+		"prestart":        serialize((*hooks)[Prestart]),
+		"createRuntime":   serialize((*hooks)[CreateRuntime]),
+		"createContainer": serialize((*hooks)[CreateContainer]),
+		"startContainer":  serialize((*hooks)[StartContainer]),
+		"poststart":       serialize((*hooks)[Poststart]),
+		"poststop":        serialize((*hooks)[Poststop]),
 	})
 }
 
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -29,6 +29,7 @@ import (
 	"github.com/checkpoint-restore/go-criu/v4"
 	criurpc "github.com/checkpoint-restore/go-criu/v4/rpc"
 	"github.com/golang/protobuf/proto"
+	errorsf "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink/nl"
 	"golang.org/x/sys/unix"
@@ -379,13 +380,12 @@ func (c *linuxContainer) start(process *Process) error {
 			if err != nil {
 				return err
 			}
-			for i, hook := range c.config.Hooks.Poststart {
-				if err := hook.Run(s); err != nil {
-					if err := ignoreTerminateErrors(parent.terminate()); err != nil {
-						logrus.Warn(err)
-					}
-					return newSystemErrorWithCausef(err, "running poststart hook %d", i)
+
+			if err := c.config.Hooks[configs.Poststart].RunHooks(s); err != nil {
+				if err := ignoreTerminateErrors(parent.terminate()); err != nil {
+					logrus.Warn(errorsf.Wrapf(err, "Running Poststart hook"))
 				}
+				return err
 			}
 		}
 	}
@@ -1621,10 +1621,12 @@ func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Proc
 				return nil
 			}
 			s.Pid = int(notify.GetPid())
-			for i, hook := range c.config.Hooks.Prestart {
-				if err := hook.Run(s); err != nil {
-					return newSystemErrorWithCausef(err, "running prestart hook %d", i)
-				}
+
+			if err := c.config.Hooks[configs.Prestart].RunHooks(s); err != nil {
+				return err
+			}
+			if err := c.config.Hooks[configs.CreateRuntime].RunHooks(s); err != nil {
+				return err
 			}
 		}
 	case "post-restore":
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -362,6 +362,9 @@ func (p *initProcess) start() (retErr error) {
 	if err := p.createNetworkInterfaces(); err != nil {
 		return newSystemErrorWithCause(err, "creating network interfaces")
 	}
+	if err := p.updateSpecState(); err != nil {
+		return newSystemErrorWithCause(err, "updating the spec state")
+	}
 	if err := p.sendConfig(); err != nil {
 		return newSystemErrorWithCause(err, "sending config to init process")
 	}
@@ -378,9 +381,9 @@ func (p *initProcess) start() (retErr error) {
 			if err := setupRlimits(p.config.Rlimits, p.pid()); err != nil {
 				return newSystemErrorWithCause(err, "setting rlimits for ready process")
 			}
-			// call prestart hooks
+			// call prestart and CreateRuntime hooks
 			if !p.config.Config.Namespaces.Contains(configs.NEWNS) {
-				// Setup cgroup before prestart hook, so that the prestart hook could apply cgroup permissions.
+				// Setup cgroup before the hook, so that the prestart and CreateRuntime hook could apply cgroup permissions.
 				if err := p.manager.Set(p.config.Config); err != nil {
 					return newSystemErrorWithCause(err, "setting cgroup config for ready process")
 				}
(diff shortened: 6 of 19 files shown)


h-vetinari:
Cool, thanks for tackling this!

But - may I ask - why not add all the hooks from opencontainers/runtime-spec#1008 in one PR?

RenaudWasTaken:
@h-vetinari I'll be opening other PRs for these because they are less trivial 😛

libcontainer/configs/config.go
```diff
@@ -227,6 +249,23 @@ type Capabilities struct {
 	Ambient []string
 }
 
+func (hooks *Hooks) RunHooks(name HookName, spec *specs.State) error {
+	hooksMap := map[HookName][]Hook{
+		Prestart: hooks.Prestart,
+		CreateRuntime: hooks.CreateRuntime,
+		Poststart: hooks.Poststart,
+		Poststop: hooks.Poststop,
```
cyphar: This file has not been `gofmt`'d, which is why the CI is failing.

RenaudWasTaken: Sorry about that, fixed :)

katiewasnothere:
@RenaudWasTaken is there a blocker here?

RenaudWasTaken:
I just rebased my PR! Just getting a review for this specific piece of the hooks :)

libcontainer/configs/config.go
```diff
@@ -205,15 +205,37 @@ type Config struct {
...
 	// Prestart commands are executed after the container namespaces are created,
 	// but before the user supplied command is executed from init.
+	// Note: This hook is now deprecated
 	Prestart []Hook
 
+	// CreateRuntime commands MUST be called as part of the create operation after
+	// the runtime environment has been created but before the pivot_root has been executed.
+	// CreateRuntime is called immediately after the deprecated Prestart hook.
+	CreateRuntime []Hook
+
```
mikebrow: suggest adding TODOs for CreateContainer and StartContainer (or just one TODO to finish sync of changes to https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L130)

libcontainer/container_linux.go
```diff
@@ -1682,10 +1682,12 @@ func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Proc
...
-			for i, hook := range c.config.Hooks.Prestart {
-				if err := hook.Run(s); err != nil {
-					return newSystemErrorWithCausef(err, "running prestart hook %d", i)
-				}
+
+			if err := c.config.Hooks.RunHooks(configs.Prestart, s); err != nil {
+				return err
+			}
+			if err := c.config.Hooks.RunHooks(configs.CreateRuntime, s); err != nil {
+				return err
```
mikebrow: to be consistent should probably change all hook.Run() calls such as https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L369

RenaudWasTaken:
Updated, thanks for the review @mikebrow !

h-vetinari:
runtime-spec `1.0.2` was just released 🎉
So that means this PR, plus implementing the other hooks, and the potentially *truly* final `1.0.0-rc11` would be ready. 😊
(just a pity that `rc7` was already nicknamed "the eleventh hour" 😋)

h-vetinari:
@RenaudWasTaken Can you rebase this?

@crosbymichael @mrunalp @dqminh @hqhq @cyphar @AkihiroSuda
People [are](https://github.com/opencontainers/runc/pull/2258#issuecomment-605009013) [asking](https://github.com/opencontainers/runc/pull/2258#issuecomment-607217616) [about](https://github.com/google/cadvisor/pull/2469#issuecomment-610430446) [releasing](https://github.com/google/cadvisor/pull/2469#issuecomment-610432196) `rc11` already, and I'm guessing this PR should be included (at least the impending resolution of the hook issue was [announced](https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc10) with `rc10`)?

OTOH, if not all the hooks from opencontainers/runtime-spec#1008 are implemented for `rc11`, there'll have to be another rc anyway?

cyphar:
> (just a pity that rc7 was already nicknamed "the eleventh hour" yum)

Apologies -- it did look like it was going to be the final RC at the time. And what an innocent time that was. :wink: I'll change the nicknames once we do an rc11.

> People are asking about releasing rc11 already, and I'm guessing this PR should be included (at least the impending resolution of the hook issue was announced with rc10)?

I don't mind doing an rc11 now since it seems Kubernetes is asking for one. But whether or not we merge this (I haven't reviewed it yet) is a separate concern to rc11 -- if we need rc11 we can merge it without this because we'd need to implement the other hooks for the final rc before `1.0` anyway.

> OTOH, if not all the hooks from opencontainers/runtime-spec#1008 are implemented for rc11, there'll have to be another rc anyway?

Yes, there will have to be one RC which acts as the "feature freeze". Normally, RCs are all mean to be feature freezes but let's just say we've broken more than a couple of rules when it comes to how project releases should work. But as I've discussed here and on the OCI calls and mailing lists, we will change how runc does releases as soon as we've got `1.0.0` out of the door.

AkihiroSuda:
Maybe dumb question, but can we consider shipping the next release as **v11.0.0** to reflect the reality?

h-vetinari:
> @AkihiroSuda: Maybe dumb question, but can we consider shipping the next release as v11.0.0 to reflect the reality?

Maybe jumping to `v2.0.0-rc1` for the feature-freeze would be enough to indicate that this has progressed far beyond what `v1.0.0` was originally planned to be?

AkihiroSuda:
@RenaudWasTaken @cyphar What's current status?

RenaudWasTaken:
Hello!

Sorry for the lag, the past few weeks have been complicated :)
I've rebased this branch and will likely spend some time this week to implement another (or all) hook(s) in a separate PR.

RenaudWasTaken:
Small update, I've managed to take some to explore a bit more the code base and look into adding the other hooks.

While doing so, I noticed a few things that need to be added to this PR:
- Add the CreateRuntime to the Marshall / Unmarshall function
- Add the CreateRuntime to the speconv

Now that I'm a bit more familiar with the code base, it looks like the changes are less important than expected and might all go in this PR. More to come this weekend.

RenaudWasTaken:
Added the other hooks, I'll spend an hour or two to add some go tests during the weekend.
I did manually manage to test all the hooks (and they work) :D !

h-vetinari:
Awesome to see this being picked back up, thanks @RenaudWasTaken! Should also adapt the PR title accordingly.

libcontainer/configs/config.go
```diff
@@ -227,11 +263,32 @@ type Capabilities struct {
...
+		Prestart:        hooks.Prestart,
+		CreateRuntime:   hooks.CreateRuntime,
+		CreateContainer: hooks.CreateContainer,
+		StartContainer:  hooks.StartContainer,
+		Poststart:       hooks.Poststart,
+		Poststop:        hooks.Poststop,
+	}
+	for i, hook := range hooksMap[name] {
+		if err := hook.Run(spec); err != nil {
+			return fmt.Errorf("%s hook #%d: %v", HookToName[name], i, err)
```
AkihiroSuda, line 277: `errors.Wrapf`

RenaudWasTaken:
Added the tests, I'll be adding some integration tests, once I recover my machine (I was playing with libc.so on a remote machine, it didn't end well for the machine...)

RenaudWasTaken:
@cyphar I believe this is ready for review, could you take a stab at it?

I'd also like to add a full blown integration test, that would mount a dummy library in the CreateContainer and CreateRuntime hooks and run ldconfig in the StartContainer.

While writing that test I realized I needed to add a new image in the integration tests, at that point the PR started to be too big, I'll make a followup PR for that :)

libcontainer/process_linux.go
```diff
@@ -487,6 +487,15 @@ func (p *initProcess) startTime() (uint64, error) {
 }
 
 func (p *initProcess) sendConfig() error {
+	if p.config.Config.Hooks != nil {
```
RenaudWasTaken, line 490: Should we just send the OCI state regardless of if there are hooks? This might be a "gotcha" moment in the future when people try to use that state and realize it's nil.

cc @cyphar

RenaudWasTaken:
Added the integration test in the following PR: https://github.com/opencontainers/runc/pull/2402

libcontainer/configs/config.go
```diff
@@ -205,15 +205,55 @@ type Config struct {
...
+)
+
+var HookToName = map[HookName]string{
+	Prestart:        "prestart",
+	CreateRuntime:   "createRuntime",
+	CreateContainer: "createContainer",
+	StartContainer:  "startContainer",
+	Poststart:       "poststart",
+	Poststop:        "poststop",
+}
```
cyphar, line 255: It'd be simpler to just make `type HookName string` and define the constants as strings.

libcontainer/process_linux.go
cyphar: IMO it makes more sense just to always send it.

libcontainer/rootfs_linux.go
```diff
@@ -98,6 +98,15 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 		return newSystemErrorWithCausef(err, "changing dir to %q", config.Rootfs)
 	}
 
+	s := iConfig.SpecState
+	if s != nil {
+		s.Pid = unix.Getpid()
```
cyphar, line 103: This will always be equal to `1` (unless pid namespaces are disabled, obviously). Shouldn't this be the PID in the parent namespace -- which is already filled in? Then again, maybe the hook wants to know the PID of the container process in the namespace it's being run in? I'm not sure...

libcontainer/rootfs_linux.go
```diff
@@ -98,6 +98,15 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
...
+		s.Status = "creating"
```
cyphar, line 104: Are these status strings really not defined as constants anywhere?

cyphar:
Needs a rebase as well.

libcontainer/configs/config.go
```diff
@@ -227,11 +267,32 @@ type Capabilities struct {
...
+func (hooks *Hooks) RunHooks(name HookName, spec *specs.State) error {
+	hooksMap := map[HookName][]Hook{
+		Prestart:        hooks.Prestart,
+		CreateRuntime:   hooks.CreateRuntime,
+		CreateContainer: hooks.CreateContainer,
+		StartContainer:  hooks.StartContainer,
+		Poststart:       hooks.Poststart,
+		Poststop:        hooks.Poststop,
+	}
+	for i, hook := range hooksMap[name] {
```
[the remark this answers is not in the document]
cyphar: Also, this `RunHooks` helper is unnecessary. You could just make `type HookList []Hook` and then define a `Run` method for that. Then you could do

```go
if err := hooks.Prestart.Run(state); err != nil {
    // ...
}
```

libcontainer/configs/config.go
```diff
@@ -227,11 +267,32 @@ type Capabilities struct {
 	Ambient []string
 }
 
+func (hooks *Hooks) RunHooks(name HookName, spec *specs.State) error {
```
cyphar, line 270: Argument name should be `state` not `spec`.

cyphar: Also, this `RunHooks` helper is associated with the wrong structure IMHO. You could just make `type HookList []Hook` and then define a `Run` method for that. Then you could do

```go
if err := hooks.Prestart.Run(state); err != nil {
    // ...
}
```

No need for `config.HookName` everywhere.

libcontainer/configs/config.go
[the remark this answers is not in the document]
RenaudWasTaken: Right that was a solution too obvious for me :D ! I'll fix that

libcontainer/configs/config.go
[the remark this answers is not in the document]
RenaudWasTaken: > CommandHook/FunctionHook/Hook thing needs to just be removed and make everything just be a Hook with the semantics of CommandHook -- we don't support direct usage of libcontainer and having these function hooks (which are ignored when serialised meaning they don't work properly) is a bit silly.

The commandHooks are nice for testing purposes, I'll look into removing them in a followup PR though.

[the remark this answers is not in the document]
cyphar: Yup -- let's leave this for a separate PR.

libcontainer/rootfs_linux.go
RenaudWasTaken: My mental model is:
- Runtime namespace hooks get runtime/absolute information
- Container Namespace hooks get container/relative information

I can specify that in the hooks documentation to help with that

libcontainer/rootfs_linux.go
RenaudWasTaken: Good catch! I'll just add a const struct somewhere!

libcontainer/rootfs_linux.go
cyphar: Yeah that does make sense. It wouldn't hurt to mention it in the hook documentation.

RenaudWasTaken: Actually reading the spec, it always  "is the ID of the container process, as seen by the host."

cyphar: Thinking about it a little more -- I think that's a bug in the spec. Yes, there isn't an `exec` primitive in the specification, but you can manually share the PID namespace which would mean that the hook run inside the container won't be able to make use of the PID in the state JSON.

libcontainer/process_linux.go
RenaudWasTaken: Addressed as a new function:
- https://github.com/opencontainers/runc/pull/2229/files#diff-e6cd3268ea1b2c0ee41532976972b36fR353-R355
- https://github.com/opencontainers/runc/pull/2229/files#diff-e6cd3268ea1b2c0ee41532976972b36fR492-R501

Marking resolved if that's ok with you :)

libcontainer/configs/config.go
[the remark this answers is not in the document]
RenaudWasTaken: Addressed as requested: https://github.com/opencontainers/runc/pull/2229/files#diff-ec0220f977914208b5bf7c1ff362365fR205-R239

libcontainer/configs/config.go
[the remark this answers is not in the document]
RenaudWasTaken: I've had to change the Unmarshalling code for the purpose of using the new type, but will address the Marshalling code in the followup PR.

See what I did here: https://github.com/opencontainers/runc/pull/2229/files#diff-ec0220f977914208b5bf7c1ff362365fR278-R295

libcontainer/rootfs_linux.go
RenaudWasTaken: So keep `unix.Getpid()` for now and address it in the spec?

libcontainer/rootfs_linux.go
RenaudWasTaken: Resolved here: https://github.com/opencontainers/runc/pull/2229/files#diff-ec0220f977914208b5bf7c1ff362365fR248-R253

RenaudWasTaken:
Rebased!

libcontainer/rootfs_linux.go
cyphar: For now, yes. But please open a runtime-spec issue so we don't lose track of it.

RenaudWasTaken: https://github.com/opencontainers/runtime-spec/pull/1045

libcontainer/rootfs_linux.go
RenaudWasTaken: Also added a PR to the runtime spec here: https://github.com/opencontainers/runtime-spec/pull/1046/files

libcontainer/configs/config.go
RenaudWasTaken: Resolved, thanks!

libcontainer/container_linux.go
```diff
@@ -368,13 +369,13 @@ func (c *linuxContainer) start(process *Process) error {
 			if err != nil {
 				return err
 			}
-			for i, hook := range c.config.Hooks.Poststart {
-				if err := hook.Run(s); err != nil {
-					if err := ignoreTerminateErrors(parent.terminate()); err != nil {
-						logrus.Warn(err)
-					}
-					return newSystemErrorWithCausef(err, "running poststart hook %d", i)
+			s.Status = configs.Running
```
mrunalp, line 372: Why are we explicitly setting the state to `Running` here? There is a call to `c.currentOCIState()` a few lines above and we are overwriting that state here.

libcontainer/container_linux.go
```diff
@@ -1708,10 +1709,13 @@ func (c *linuxContainer) criuNotifications(resp *criurpc.CriuResp, process *Proc
 				return nil
 			}
 			s.Pid = int(notify.GetPid())
-			for i, hook := range c.config.Hooks.Prestart {
-				if err := hook.Run(s); err != nil {
-					return newSystemErrorWithCausef(err, "running prestart hook %d", i)
-				}
+			s.Status = configs.Creating
```
mrunalp, line 1712: Same comment here as above re: explicitly setting the state.

libcontainer/container_linux.go
RenaudWasTaken: Good catch! As part of the copy paste for these calls, I also copied the fact that we were manually setting the state.

Should we be setting the state anywhere?

tests/integration/helpers.bash
```diff
@@ -257,6 +260,12 @@ function requires() {
 				skip_me=1
 			fi
 			;;
+		cgroups_devices)
+			init_cgroup_paths
+			if [ ! -e "${CGROUP_DEVICES_BASE_PATH}/devices.list" ]; then
```
AkihiroSuda, line 265: This check should be removed for v2

RenaudWasTaken: This check is actually not needed (I added it when experimenting with things), thanks for pointing it out!

tests/integration/hooks.bats
```diff
@@ -0,0 +1,63 @@
...
+	requires no_systemd
+
+	# setup some dummy libs
+	gcc -shared -Wl,-soname,librunc-hooks-create-runtime.so.1 -o "$HOOKLIBCR.1.0.0"
+	gcc -shared -Wl,-soname,librunc-hooks-create-container.so.1 -o "$HOOKLIBCC.1.0.0"
+
+	current_pwd="$(pwd)"
+
+	# To mount $HOOKLIBCR we need to do that in the container namespace
+	create_runtime_hook=$(cat <<-EOF
```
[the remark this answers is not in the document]
RenaudWasTaken: So I tried to explain this test in the commit message:

> This patch adds a test based on real world usage of runc hooks
(libnvidia-container). We verify that mounting a library inside
a container and running ldconfig succeeds.

It's a test that does require some setup, but essentially it allows some extensive testing. In that we're not just testing that we can run the hook but that the setup around the hook is correct.
There are so many things that could lead to breaking the hooks that it's nice to have something that's close to real world example.

And the test essentially boils down to:
- create some dummy so files
- Run a container with the following hooks:
   - createRuntime: Mounts the lib inside the container
   - createContainer: Mounts the lib inside the container
   - startContainer: Runs ldconfig

Check the output of `ldconfig -p | grep librunc` to ensure that the libs that we mounted are picked up.

tests/integration/helpers.bash
RenaudWasTaken: Fixed!

tests/integration/hooks.bats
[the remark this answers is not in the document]
AkihiroSuda: SGTM then

RenaudWasTaken:
ping @cyphar @AkihiroSuda :)

Vagrantfile
```diff
@@ -31,10 +31,15 @@ EOF
...
+    curl -o /usr/local/bin/umoci -fsSL https://github.com/openSUSE/umoci/releases/download/v0.4.5/umoci.amd64
+    chmod +x /usr/local/bin/umoci
+
     # Add busybox for libcontainer/integration tests
     . /vagrant/tests/integration/multi-arch.bash \
-        && mkdir /busybox \
-        && curl -fsSL $(get_busybox) | tar xfJC - /busybox
+        && mkdir /busybox /debian \
+        && curl -fsSL $(get_busybox) | tar xfJC - /busybox \
+        && get_and_extract_debian /debain
```
AkihiroSuda, line 42: typo: debain

RenaudWasTaken: Fixed! Thanks!

libcontainer/configs/config.go
```diff
@@ -227,32 +264,39 @@ type Capabilities struct {
...
 	if err := json.Unmarshal(b, &state); err != nil {
 		return err
 	}
 
-	deserialize := func(shooks []CommandHook) (hooks []Hook) {
-		for _, shook := range shooks {
-			hooks = append(hooks, shook)
+	*hooks = Hooks{}
+	for n, commandHooks := range state {
+		if len(commandHooks) == 0 {
```
cyphar, line 286: This isn't necessary -- `range` over an empty array is a no-op.

tests/integration/hooks.bats
```diff
@@ -0,0 +1,63 @@
...
+	teardown_debian
+}
+
+@test "runc run (hooks library tests)" {
+	requires root
+	requires no_systemd
+
+	# setup some dummy libs
+	gcc -shared -Wl,-soname,librunc-hooks-create-runtime.so.1 -o "$HOOKLIBCR.1.0.0"
+	gcc -shared -Wl,-soname,librunc-hooks-create-container.so.1 -o "$HOOKLIBCC.1.0.0"
```
cyphar, line 33: As happy as I am to have `umoci` as a build dependency for runc, as far as I can tell the only reason you need debian is because you want to run `ldconfig` as a way to check if certain files exist in the container. Is it possible to just create files with special contents and check that the contents are correct -- that way you wouldn't need a debian image for testing? Or is there a particular reason you want to test `ldconfig` specifically?

libcontainer/configs/config.go
cyphar: Fair enough I guess, though I don't think there's anything that actually requires it to be empty?

RenaudWasTaken: I remember some tests needed to change, do you want me to update this piece of code?

tests/integration/hooks.bats
cyphar: To avoid blocking this PR for too long, I'd be happy to take it as-is and we can later change all of the tests to use Debian rather than busybox (this would also avoid having two methods of downloading and unpacking image as well). I'm not sure if that will be a painful change to make, so let's punt on it for now.

I don't want to over-simplify the use-cases you folks have (even though `ldconfig` is ultimately "just writing files") so I'd prefer to move the rest of the test suite rather than over-simplifying this particular test.

RenaudWasTaken: I've got some time set aside to refactor the Hooks after this, I can add some more to refactor the tests :) !

AkihiroSuda:
needs rebase

RenaudWasTaken:
Rebased :) !

Should we merge it and I'll create the followup two PRs:
- Transition tests to use the debian image
- Investigate removing the custom JSON serialization of the Hook structure

RenaudWasTaken:
Sorry I didn't realize the CI wasn't passing, pushed an update :) !

cyphar:
CentOS7 is failing because it doesn't use the Dockerfile and so `skopeo` isn't installed:

```
not ok 44 runc run (hooks library tests)
# (from function `get_and_extract_debian' in file tests/integration/multi-arch.bash, line 35,
#  from function `setup_debian' in file tests/integration/helpers.bash, line 449,
#  from function `setup' in test file tests/integration/hooks.bats, line 16)
#   `setup_debian' failed with status 127
# runc list (status=0):
# /vagrant/tests/integration/multi-arch.bash: line 35: skopeo: command not found
# runc list (status=0):
```

RenaudWasTaken:
Hopefully fixed in this new one!

cyphar:
Ah, you need to get `umoci` too. :wink: I'll do a final review of this today, but from a quick skim it should be good to merge and we can work on the other remaining topics (cleaning up the rest of the test suite to use one image and so on) later. Sorry about how long this took to get properly reviewed, and I'll buy you a beer if/when we meet up at a conference.

RenaudWasTaken:
Updated!

RenaudWasTaken:
> Sorry about how long this took to get properly reviewed, and I'll buy you a beer if/when we meet up at a conference.

No worries, I'm used to open source now and I sometime take a long time to review PRs :D !
I will however happily accept the beer (and am happy to buy a few too!) if we ever cross at FOSDEM/Kubecon or any other conference!

Dockerfile
```diff
@@ -67,6 +67,19 @@ RUN mkdir -p /usr/src/criu \
...
+RUN echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list \
+    && wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_Unstable/Release.key -O- | sudo apt-key add - \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends skopeo \
+    && rm -rf /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list \
+    && apt-get clean \
+    && rm -rf /var/cache/apt /var/lib/apt/lists/*;
+
+# install umoci
+RUN curl -o /usr/local/bin/umoci -fsSL https://github.com/openSUSE/umoci/releases/download/v0.4.5/umoci.amd64 \
```
cyphar, line 80: umoci has moved home to `github.com/opencontainers/umoci`.

Vagrantfile.centos7
```diff
@@ -18,12 +18,16 @@ Vagrant.configure("2") do |config|
...
     yum install -y -q epel-release
-    yum install -y -q gcc git iptables jq libseccomp-devel make
+    yum install -y -q gcc git iptables jq libseccomp-devel make skopeo
     yum clean all
 
     # install Go
     curl -fsSL "https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz" | tar Cxz /usr/local
 
+    # Install umoci
+    curl -o /usr/local/bin/umoci -fsSL https://github.com/openSUSE/umoci/releases/download/v0.4.5/umoci.amd64
```
cyphar, line 28: Needs to be updated here too.

Vagrantfile.fedora32
```diff
@@ -31,10 +31,15 @@ EOF
     cat /root/rootless.key.pub >> /home/rootless/.ssh/authorized_keys
     chown -R rootless.rootless /home/rootless
 
+    # Install umoci
+    curl -o /usr/local/bin/umoci -fsSL https://github.com/openSUSE/umoci/releases/download/v0.4.5/umoci.amd64
```
cyphar, line 35: And here.

libcontainer/configs/config.go
```diff
@@ -203,17 +203,54 @@ type Config struct {
...
 
 	// Poststop commands are executed after the container init process exits.
-	Poststop []Hook
-}
+	// Poststop commands are called in the Runtime Namespace.
+	Poststop = "poststop"
+)
+
+var (
+	HookNameList = []HookName{Prestart, CreateRuntime, CreateContainer, StartContainer, Poststart, Poststop}
```
cyphar, line 243: I'd prefer if this was only put in the `_test.go` file, since it's only used for tests.

RenaudWasTaken: Fixed

Vagrantfile.fedora32
RenaudWasTaken: Fixed

Vagrantfile.centos7
RenaudWasTaken: Fixed

Dockerfile
RenaudWasTaken: Fixed

RenaudWasTaken:
I'll have the followup PRs by next week!

Issue #2214, opened 2020-01-16
chroot() function appears to ignore its parameter


setharnold:
Hello, this `chroot()` wrapper appears to ignore the `rootfs` parameter. Is this an accidental oversight?

https://github.com/opencontainers/runc/blob/709377ca558df88ea538852c9310b700f140fc9b/libcontainer/rootfs_linux.go#L857

Thanks

cyphar:
This is an oversight, but it works okay because we `Chdir(rootfs)` earlier during setup.

setharnold:
Thanks cyphar; I'm glad to hear it's not a huge concern.

Thanks

cyphar:
Feel free to send a patch which either uses the path or removes the parameter.

PR #2237, opened 2020-02-27
Chroot rootfs path usage fix

Zyqsempai:
Fixes: #2214 

In order to avoid misunderstanding, it's better to use path parameter in chroot() func.

Signed-off-by: Boris Popovschi <zyqsempai@mail.ru>

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -867,10 +867,7 @@ func msMoveRoot(rootfs string) error {
 }
 
 func chroot(rootfs string) error {
-	if rootfs == "" {
-		rootfs = "."
-	}
-	if err := unix.Chroot(rootfs); err != nil {
+	if err := unix.Chroot(filepath.Clean(rootfs)); err != nil {
 		return err
 	}
 	return unix.Chdir("/")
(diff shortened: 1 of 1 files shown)


Zyqsempai:
@cyphar PTAL

libcontainer/rootfs_linux.go
```diff
@@ -867,7 +867,7 @@ func msMoveRoot(rootfs string) error {
 }
 
 func chroot(rootfs string) error {
-	if err := unix.Chroot("."); err != nil {
+	if err := unix.Chroot(rootfs); err != nil {
```
AkihiroSuda: needs validation?

AkihiroSuda: should error out if rootfs != "." ?

Zyqsempai: @AkihiroSuda Do you mean something like this?(check last commit)

AkihiroSuda: `if filepath.Clean(rootfs) != os.Getwd(){ return errors.New(..) }`? (Make sure to test it, ideally in CI)

Zyqsempai: os.Getwd() returns absolute wd path, on other hand rootfs path here can be relative and we will get error even if they are point at same dir.

AkihiroSuda: `if filepath.Abs(rootfs) != os.Getwd(){ return errors.New(..) }`. The argument of `Chroot()` should remain `"."`.

kolyshkin:
so, should we close this one?

PR #2238, opened 2020-02-27
Use named error return for initProcess#start

tedyu:
In initProcess#start, there is deferred func :
```
	defer func() {
		if err != nil {
```
Since the error return is not named, the above check doesn't cover all the cases.
e.g. a few lines below:
```
	if _, err := io.Copy(p.messageSockPair.parent, p.bootstrapData); err != nil {
		return newSystemErrorWithCause(err, "copying bootstrap data to pipe")
```
The check in deferred func doesn't check either err or, the actual returned error.

This PR changes the error return to be named.

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -279,7 +279,7 @@ func (p *initProcess) waitForChildExit(childPid int) error {
 	return nil
 }
 
-func (p *initProcess) start() error {
+func (p *initProcess) start() (retErr error) {
 	defer p.messageSockPair.parent.Close()
 	err := p.cmd.Start()
 	p.process.ops = p
@@ -290,6 +290,15 @@ func (p *initProcess) start() error {
 		p.process.ops = nil
 		return newSystemErrorWithCause(err, "starting init process command")
 	}
+	defer func() {
+		if retErr != nil {
+			p.manager.Destroy()
+			if p.intelRdtManager != nil {
+				p.intelRdtManager.Destroy()
+			}
+		}
+	}()
+
 	// Do this before syncing with child so that no children can escape the
 	// cgroup. We don't need to worry about not doing this and not being root
 	// because we'd be using the rootless cgroup manager in that case.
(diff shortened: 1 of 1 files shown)


tedyu:
@hqhq @cyphar
Mind taking a look ?

tedyu:
The deferred func starting line 352 is duplicate with the one starting line 304.

I think the second deferred func can be removed.

tedyu:
What happens is that the second deferred func is ineffective.
I checked every err variable after line 352: they are all local (newly declared) to the operation.

I will drop the second deferred func.

kolyshkin:
@tedyu thanks for the patch! Coincidentally, I was looking into this part of code as well and ended up with a similar patch.

Some discussion about the code is provided at https://github.com/opencontainers/runc/pull/1916#pullrequestreview-169551062 (also, this is a PR to blame) and yet it doesn't explain the double defer.

I think the defer needs to be moved up a bit (before `p.intelRdtManager.Apply()` or even before `p.manager.Apply()`), otherwise, if `p.intelRdtManager.Apply()` fails, `p.manager.Destroy()` is not called.

@crosbymichael @giuseppe PTAL

mrunalp:
LGTM

mrunalp:
@AkihiroSuda ptal

tedyu:
@kolyshkin 
I have lifted the deferred func earlier.
Please take a look.

tedyu:
@kolyshkin
Please let me know if there is anything else to be addressed.

thaJeztah:
@crosbymichael ptal 🤗

crosbymichael:
LGTM

tedyu:
@crosbymichael @thaJeztah 
Can this be merged ?

Thanks

thaJeztah:
I'm not a maintainer in this repository; @AkihiroSuda @mrunalp PTAL?

libcontainer/process_linux.go
```diff
@@ -290,6 +290,16 @@ func (p *initProcess) start() error {
 		p.process.ops = nil
 		return newSystemErrorWithCause(err, "starting init process command")
 	}
+	defer func() {
+		if retErr != nil {
+			// TODO: should not be the responsibility to call here
```
AkihiroSuda: Is this TODO still correct?

tedyu: @AkihiroSuda 
I have dropped this TODO.

AkihiroSuda:
LGTM

tedyu:
@AkihiroSuda 
Can you merge this ?

thanks

tedyu:
@mrunalp 
Can you merge this ?

thanks

tedyu:
Thanks @AkihiroSuda

AkihiroSuda:
> Signed-off-by: zyu

Please use your full name next time

libcontainer/process_linux.go
```diff
@@ -279,7 +279,7 @@ func (p *initProcess) waitForChildExit(childPid int) error {
 	return nil
 }
 
-func (p *initProcess) start() error {
+func (p *initProcess) start() (retErr error) {
```
cyphar: Nit: Please call this `Err`.

PR #2234, opened 2020-02-23
Dockerfile: some refactoring, and switch to "buster" variant

thaJeztah:
Debian buster is the current stable, so might be good to switch to that version; also did some refactoring of the dockerfile (see individual commits for details).

I'll open a follow-up PR to update to go 1.13, but thought to first make the switch to "buster" to keep changes smaller.

PR #2239, opened 2020-02-28
Update Golang 1.13.8, start testing on Go 1.14

thaJeztah:
Follow up to https://github.com/opencontainers/runc/pull/2234


kolyshkin:
This is very much needed, because despite having a bunch of versions in .travis.yml we only test on golang 1.12 (and do it many times, unnecessarily overloading travis).

.travis.yml
```diff
@@ -1,19 +1,19 @@
...
   include:
-    - go: 1.12.x
+    - go: 1.13.x
       env:
         - RUNC_USE_SYSTEMD=1
       script:
         - make BUILDTAGS="${BUILDTAGS}" all
         - sudo PATH="$PATH" make localintegration RUNC_USE_SYSTEMD=1
-    - go: 1.12.x
+    - go: 1.13.x
```
kolyshkin: One thing I don't understand here (maybe just because it's slow Saturday) is why there is only a single versoin of go (such as `go: 1.13.x`) in this matrix.

.travis.yml
```diff
@@ -1,19 +1,19 @@
 dist: bionic
 language: go
 go:
-  - 1.11.x
-  - 1.12.x
+  - 1.13.x
+  - 1.14.x
   - tip
```
cyphar: To avoid having to update this every time, we should have a `1.x` entry so we're always testing the latest release.

thaJeztah: Not sure. That way we wouldn't be testing "latest - 1", would that be an issue? (go supports last two versions, and not everyone may be on latest on day 1)

Also it would mean CI "silently" switching to a new Go version, which could be confusing why things all of a sudden start to fail

kolyshkin: we already have `tip` to test on the latest version, no need to have `1.x`.

cyphar: `tip` tests the latest build of Go master, not the latest released version. This would be like testing against nightly in the Rust world (useful, but not as useful as testing against the latest release).

I wasn't saying to remove the `latest -1` line, I was just saying we should add `1.x`.

kolyshkin: > I wasn't saying to remove the latest - 1 line (annoyingly Travis doesn't support this concept)

I just found they actually do! Travis is using `gimme` (https://github.com/travis-ci/gimme) for installing go versions, and this is what it gives to me
```console
[kir@kir-rhat runc]$ gimme -r oldstable
1.13.8
[kir@kir-rhat runc]$ gimme -r stable
1.14
[kir@kir-rhat runc]$ gimme -r 1.x
1.14
```

I also learned that `tip` means "get go it from git and compile" and I am not sure we want it.

So, since golang supports two versions, we can use `stable` and `oldstable` and catch the problems early. I think @thaJeztah would say that specifying versions explicitly is always better because we know what we use, and it won't break suddenly because of golang 1.23 (or whatever) is released. To that, I'd say (1) we have to support supported go releases, and if something breaks we need to take care ASAP (2) it is not ideal, but better than the current situation then we test everything on two unsupported go releases.

cyphar: :+1: to using `oldstable` and `stable`.

kolyshkin: OK, the bad news is for some reason `oldstable` is not working for Travis CI (which uses the very same version of `gimme` that I did above so I'm not sure why).

Reported earlier (perhaps to a wrong place)
https://github.com/travis-ci/gimme/issues/179

I have reported it as well (hopefully to the right place)
https://travis-ci.community/t/cant-specify-oldstable-as-go-version-in-travis-yml/7670

kolyshkin: Also filed https://github.com/travis-ci/gimme/issues/185.

thaJeztah:
superseded by https://github.com/opencontainers/runc/commit/3147c3200ead4d839148ea8408c504f758944fd7

.travis.yml
kolyshkin: This is finally allegedly fixed, so https://github.com/opencontainers/runc/pull/2538

PR #2252, opened 2020-03-09
Synchronize the call to linuxContainer.Signal()

### LINKED ISSUE #2251: linuxContainer.Signal() can race with another call to say Destroy()
linuxContainer.Signal() calls c.initProcess.signal(s). However, a concurrent call to linuxContainer.Destroy() will set c.initProcess = nil. This can lead to a nil pointer dereference, so the calls should be synchronized.

pkagrawal:
linuxContainer.Signal() can race with another call to say Destroy()
which clears the container's initProcess. This can cause a nil pointer
dereference in Signal().

This patch will synchronize Signal() and Destroy() by grabbing the
container's mutex as part of the Signal() call.

Closes #2251

Signed-off-by: Pradyumna Agrawal <pradyumnaa@vmware.com>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -379,6 +379,8 @@ func (c *linuxContainer) start(process *Process) error {
 }
 
 func (c *linuxContainer) Signal(s os.Signal, all bool) error {
+	c.m.Lock()
+	defer c.m.Unlock()
 	if all {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

mrunalp:
LGTM

thaJeztah:
should this be backported?

PR #2254, opened 2020-03-11
Makefile: set selinux and apparmor build tags

kolyshkin:
Both selinux and apparmor subsystem can detect whether it is enabled,
and act accordingly. Compiling it in by default should help avoid
some frustration cased by missing build tags.
    
This should not change anything in case BUILDTAGS is already set.
    
README.md is amended to clarify what BUILDTAGS are enabled by
default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 PROJECT := github.com/opencontainers/runc
-BUILDTAGS ?= seccomp
+BUILDTAGS ?= seccomp selinux apparmor
 COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
 COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,6 @@ make BUILDTAGS='seccomp apparmor'
 | seccomp   | Syscall filtering                  | libseccomp  |
 | selinux   | selinux process and mount labeling | <none>      |
 | apparmor  | apparmor profile support           | <none>      |
-| ambient   | ambient capability support         | kernel 4.3  |
 | nokmem    | disable kernel memory account      | <none>      |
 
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -13,7 +13,8 @@ GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 PROJECT := github.com/opencontainers/runc
-BUILDTAGS ?= seccomp
+SELINUX_TAG := $(shell selinuxenabled 2>/dev/null && echo selinux)
+BUILDTAGS ?= seccomp $(SELINUX_TAG)
```
AkihiroSuda: Can we add apparmor using `aa-enabled` for consistency?

kolyshkin: Good idea; done

AkihiroSuda:
LGTM

cyphar:
We should always build binaries with SELinux and AppArmor support because the code in runc will detect at runtime whether the machine supports SELinux or AppArmor. The only reason not to build with support of either is if the required shared libraries aren't installed on the system. Personally I think that these build-tags should be *opt-out* because they just cause headaches when people build binaries that have LSM support disabled for no good reason.

mikebrow:
Not against changing from opt in to opt out... upstreams could possibly be effected from a nit resource perspective, some test buckets, but easy to address.

kolyshkin:
> The only reason not to build with support of either is if the required shared libraries aren't installed on the system.

I'm not sure about apparmor, but `github.com/opencontainers/selinux` package is deliberately written to not depend on any shared libraries, so always adding `selinux` tag should not bring any regressions.

kolyshkin:
Updated to enable `selinux` and `apparmor` by default, and explain it a bit better in README.

AkihiroSuda:
LGTM

README.md
```diff
@@ -60,20 +58,23 @@ sudo make install
...
-| selinux   | selinux process and mount labeling | <none>      |
-| apparmor  | apparmor profile support           | <none>      |
-| ambient   | ambient capability support         | kernel 4.3  |
-| nokmem    | disable kernel memory account      | <none>      |
+| Build Tag | Feature                            | Enabled by default | Dependency |
+|-----------|------------------------------------|--------------------|------------|
+| seccomp   | Syscall filtering                  | yes                | libseccomp |
+| selinux   | selinux process and mount labeling | yes                | <none>     |
+| apparmor  | apparmor profile support           | yes                | <none>     |
+| ambient   | ambient capability support         | no                 | kernel 4.3 |
```
cyphar: This build tag no longer exists. In fact, us adding it in the first place introduced a privesc bug.

kolyshkin: Yep, fixed.

AkihiroSuda:
LGTM

cyphar:
LGTM.

PR #2255, opened 2020-03-13
Faster mountinfo parser

kolyshkin:
This is **part 1** of improving runc wrt /proc/self/mountinfo parsing.

It needs improvements because
 * the current code has 5 different implementations of runc parser
 * a single `runc run` command reads mountinfo from from 72 to 116 times (depending on whether `--systemd-cgroup` is set) on my system.
 * reading mountinfo from the kernel might be slow if there are many containers running (and thus many mounts)
 * reading mountinfo might be racy (see https://github.com/kolyshkin/procfs-test)

This PR
* updates the mountinfo parsing code (which was originally taken from Docker) to the recent code from Moby, making it way faster and versatile. For background on those improvements, see
  * moby/moby#36091
  * moby/moby#38977
  * https://github.com/moby/moby/pull/40684
 * converts existing users to benefit from it.

My plans are
 - [ ] part 2: switch the code that parses mountinfo on its own to use this package
 - [ ] part 3: look into cgroups code, trying to minimize calls to parse mountinfo


AkihiroSuda:
Do you have benchmark results?

kolyshkin:
> Do you have benchmark results?

I have just updated the description. Haven't performed any benchmarks for this one, but e.g.  https://github.com/moby/moby/pull/36091 claims 8x speed improvement :)

AkihiroSuda:
Maybe we should begin with splitting the repo first https://github.com/moby/moby/pull/40684#issuecomment-598516040

kolyshkin:
@AkihiroSuda is it OK to have runc depend on moby repos? I am asking because commit 3ca4c78b1a07 specifically says "get rid of docker/docker dependency". Or is a dependency upon a smaller thing is fine?

AkihiroSuda:
> Or is a dependency upon a smaller thing is fine?

I think fine

crosbymichael:
no, we don't want to dependency on moby/moby.  moby already imports runc and we don't want it to be the other way around.  keep the code here :)

kolyshkin:
> no, we don't want to dependency on moby/moby. moby already imports runc and we don't want it to be the other way around. keep the code here :)

I never proposed a dependency on moby/moby, this is out of the question.

I don't like copy-pasting code around either (as it creates a maintenance nightmare -- instead of fixing the code in one place, it has to be fixed in a few).

So, for now I have split out moby/moby/pkg/mount to two new packages:
 * github.com/moby/sys/mount
 * github.com/moby/sys/mountinfo
and made moby use it (part I: https://github.com/moby/moby/pull/40691, part II: https://github.com/moby/moby/pull/40694)

Anyway, moving along according to the new plan: https://github.com/opencontainers/runc/pull/2256https://github.com/opencontainers/runc/pull/2256

kolyshkin:
replaced by #2256

PR #2256, opened 2020-03-14
Use faster mountinfo parser (part 1)

kolyshkin:
This is **part 1** of improving runc wrt /proc/self/mountinfo parsing (and an alternative to https://github.com/opencontainers/runc/pull/2255).

It needs improvements because:
 * the current code has 5 different implementations of runc parser
 * a single `runc run` command reads mountinfo from from 72 to 116 times (depending on whether `--systemd-cgroup` is set) on my system.
 * reading mountinfo from the kernel might be slow if there are many containers running (and thus many mounts)
 * reading mountinfo might also be racy (see https://github.com/kolyshkin/procfs-test)

This PR
* removes the old `libcontainer/mount` mountinfo parser (which was taken from Docker in Nov 2017)
* makes use of moby/sys/mountinfo package which evolves from the same code as above but is fast (for background on improvements, see moby/moby#36091, moby/moby#38977, moby/moby/pull/40684)
 * converts existing users to benefit from the above.

The new parser itself is up to 8x faster than the one removed; in addition, it implements mountinfo filters that may result in additional savings (faster, less garbage to collect).

### History
* v2: switch to golang 1.13
* v3: switch to go modules (after  #2073 is merged)

My future plans (**not in this PR**) are
 - [ ] part 2: make other runc code that parses mountinfo on its own to use the above package
 - [ ] part 3: look into cgroups code, trying to minimize calls to parse mountinfo


kolyshkin:
This is failing because we still use golang 1.12 :(

kolyshkin:
Added a temporary commit switching to golang 1.13. Ultimately, we need to merge https://github.com/opencontainers/runc/pull/2234 and https://github.com/opencontainers/runc/pull/2239

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12-stretch
+FROM golang:1.14-stretch
```
AkihiroSuda: Why 1.14?

kolyshkin: Either 1.13 or 1.14 should work just fine. As 1.13 will eventually be unsupported, I'm jumping straight to 1.14. Can use 1.13 as well here, let me know if you want it that way.

Ideally we should merge https://github.com/opencontainers/runc/pull/2239 first and just rebase this one on top of it, but it's still marked as draft by @thaJeztah.

AkihiroSuda: 1.14 has breaking changes on `SIGURG`&`EINTR` stuff  https://golang.org/doc/go1.14#runtime

kolyshkin:
> Please roll back to go 1.13

done

kolyshkin:

kolyshkin:
IDK what's going on on travis, it doesn't like `oldstable` as go version. Works fine on my laptop though, and same gimme version.

My laptop:
```
$ gimme -r oldstable
1.13.8
$ gimme -V
v1.5.3
```

Travis:
```
gimme version
v1.5.3
...
gimme: version specifier 'oldstable' unknown
error: GIMME_GO_VERSION not recognized as valid
  got: oldstable
Failed to run gimme
The command "travis_setup_go" failed and exited with 86 during .
```

Anyway I'll just use 1.13 and 1.14

Dockerfile
thaJeztah: It was marked "draft", because I separated it from https://github.com/opencontainers/runc/pull/2234. Hopefully that one could be merged soon, and I'll rebase https://github.com/opencontainers/runc/pull/2239

kolyshkin:
Some glitch in CI, rebasing to re-trigger it

> gzip: stdin: not in gzip format
> 1452tar: Child returned status 1
> 1453tar: Error is not recoverable: exiting now
> 1454The command '/bin/sh -c mkdir -p /usr/src/criu     && curl -sSL https://github.com/checkpoint-restore/criu/archive/${CRIU_VERSION}.tar.gz | tar -v -C /usr/src/criu/ -xz --strip-components=1     && cd /usr/src/criu     && make install-criu     && rm -rf /usr/src/criu' returned a non-zero code: 2

kolyshkin:
CI is good now, was a glitch

.travis.yml
```diff
@@ -1,25 +1,24 @@
...
 
 matrix:
   include:
-    - go: 1.x
+    - go: stable
       name: "verify-dependencies"
       script:
         - make verify-dependencies
-    - go: 1.12.x
+    - go: stable
```
AkihiroSuda: stable -> 1.13.x ?

kolyshkin: No, currently `stable` means 1.14.x. I didn't want to use an explicit version here since it will need to be changed every once in a while, but an implicit version like `stable` can stay as is.

Do you want it to be explicit?

AkihiroSuda: We aren't ready for 1.14 yet

kolyshkin: this is CI only and it passes.

OK I'll switch to explicitly 1.13

kolyshkin: updated

AkihiroSuda:
LGTM

mrunalp:
LGTM

mrunalp:
@crosbymichael @cyphar @giuseppe ptal.

AkihiroSuda:
@mrunalp Can we merge this, or do you want one more LGTM?

PR #2257, opened 2020-03-17
Use signal map from x/sys/unix

kolyshkin:
### Remove signalmap, use unix.SignalNum
    
Since https://github.com/golang/sys/commit/d455e41777fca6e8a5a79e34a14b8368bc11d9ba, x/sys/unix has a function to convert signal name
to a number. Let's use it and drop home-grown signal map.
    
While at it, add a test case.

~~### runc kill: check numeric signals for sanity~~
    
~~If a signal is specified as a number, perform an additional~~
~~check that the signal is known, otherwise return an error.~~
    
~~Add a test case for it.~~


kill.go
```diff
@@ -58,10 +59,18 @@ signal to the init process of the "ubuntu01" container:
 func parseSignal(rawSignal string) (syscall.Signal, error) {
 	s, err := strconv.Atoi(rawSignal)
 	if err == nil {
-		return syscall.Signal(s), nil
+		signal := syscall.Signal(s)
+		if unix.SignalName(signal) == "" {
```
cyphar: While we are unlikely to see many new signals being defined, I don't see a reason to deny sending a signal if the kernel allows you to do it.

thaJeztah: Would a log entry be useful (to catch a potentially incorrect signal?)

crosbymichael: no, send the signal and let the kernel return an error, if  there even is one

kolyshkin: @crosbymichael see https://github.com/containerd/containerd/pull/4115 :)

kolyshkin: This is why I put this code in a separate commit (now removed).

@cyphar fixed

AkihiroSuda:
LGTM

kill.go
```diff
@@ -60,8 +61,12 @@ func parseSignal(rawSignal string) (syscall.Signal, error) {
...
 		return syscall.Signal(s), nil
 	}
-	signal, ok := signalMap[strings.TrimPrefix(strings.ToUpper(rawSignal), "SIG")]
-	if !ok {
+	sig := strings.ToUpper(rawSignal)
+	if !strings.HasPrefix(sig, "SIG") {
+		sig = "SIG" + sig
+	}
+	signal := unix.SignalNum(sig)
+	if signal == 0 {
```
crosbymichael: sig 0 is valid to check if a process is alive

tianon: Dropping the knowledge bombs :smile:

Looks like an oversight in the `golang.org/x/sys/unix` library's `SignalNum` library: (https://godoc.org/golang.org/x/sys/unix#SignalNum)

> SignalNum returns the syscall.Signal for signal named s, or 0 if a signal with such name is not found. The signal name should start with "SIG".

Given that there's not a name for "SIG zero" and that we catch the explicit "numeric input" case further up, this should be reasonable, right?

mikebrow: reasonable I think until someone adds the equivalent of a new SIGNONE or SIGPING text name option to the list of kill signals and someone adds it to golang unix.SignalNum() :-0

mikebrow: maybe a comment line or two explaining why the name lookup should fail for zero for now even though the raw 0 option should not fail..

thaJeztah: > reasonable I think until someone adds the equivalent of a new SIGNONE or SIGPING text name option to the list of kill signals and someone adds it to golang unix.SignalNum()

@tklauser fyi (discussion above)

kolyshkin: > sig 0 is valid to check if a process is alive

but it does not have a name, so...

@mikebrow I'm not sure if adding any comments to the code will improve it. It's pretty clear anyway -- we treat it as a number first, and pass on any valid number to Syscall.Signal(), letting it return an error if the number is wrong.

If it's not a number, we treat it as a name and use `unix.SignalNum` to resolve. The doc for the function clearly states that it returns `0` if a signal with such name is not found, so we check for `0`. 

All that is very clear from the code.

tklauser: > > reasonable I think until someone adds the equivalent of a new SIGNONE or SIGPING text name option to the list of kill signals and someone adds it to golang unix.SignalNum()
> 
> @tklauser fyi (discussion above)

That's indeed an oversight in `x/sys/unix`. The better option would have been to return `-1` in case no signal with the given name is found. But I think, we probably cannot change that now as we would break existing users :(

crosbymichael: ok, i see. this logic should be fine since it's a name lookup

crosbymichael:
LGTM

kill.go
cyphar: Thanks.

PR #2258, opened 2020-03-17
Retry writing to cgroup files on EINTR error

danail-branekov:
Golang 1.14 introduces asynchronous preemption which results into
applications getting frequent EINTR (syscall interrupted) errors when
invoking slow syscalls, e.g. when writing to cgroup files.

As writing to cgroups is idempotent, it is safe to retry writing to the
file whenever the write syscall is interrupted.

Note that this PR also bumps the golang version in the test image `Dockerfile` so that the fix can be verified.

### DIFF
--- a/libcontainer/cgroups/fscommon/fscommon_test.go
+++ b/libcontainer/cgroups/fscommon/fscommon_test.go
@@ -14,6 +14,10 @@ import (
 )
 
 func TestWriteCgroupFileHandlesInterrupt(t *testing.T) {
+	if cgroups.IsCgroup2UnifiedMode() {
+		t.Skip("cgroup v2 is not supported")
+	}
+
 	memoryCgroupMount, err := cgroups.FindCgroupMountpoint("", "memory")
 	if err != nil {
 		t.Fatal(err)
--- a/libcontainer/cgroups/fscommon/fscommon_test.go
+++ b/libcontainer/cgroups/fscommon/fscommon_test.go
@@ -1,0 +1,35 @@
+// +build linux
+
+package fscommon
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+)
+
+func TestWriteCgroupFileHandlesInterrupt(t *testing.T) {
+	memoryCgroupMount, err := cgroups.FindCgroupMountpoint("", "memory")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	cgroupName := fmt.Sprintf("test-eint-%d", time.Now().Nanosecond())
+	cgroupPath := filepath.Join(memoryCgroupMount, cgroupName)
+	if err := os.MkdirAll(cgroupPath, 0755); err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(cgroupPath)
+
+	for i := 0; i < 100000; i++ {
+		limit := 1024*1024 + i
+		if err := WriteFile(cgroupPath, "memory.limit_in_bytes", strconv.Itoa(limit)); err != nil {
+			t.Fatalf("Failed to write %d on attempt %d: %+v", limit, i, err)
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12-stretch
+FROM golang:1.14-stretch
```
AkihiroSuda: The default should be still kept <= 1.13. But you can override the version to 1.14 with `--build-arg` in CI.

libcontainer/cgroups/fscommon/fscommon_test.go
```diff
@@ -0,0 +1,35 @@
...
+	cgroupName := fmt.Sprintf("test-eint-%d", time.Now().Nanosecond())
+	cgroupPath := filepath.Join(memoryCgroupMount, cgroupName)
+	if err := os.MkdirAll(cgroupPath, 0755); err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(cgroupPath)
+
+	for i := 0; i < 100000; i++ {
+		limit := 1024*1024 + i
+		if err := WriteFile(cgroupPath, "memory.limit_in_bytes", strconv.Itoa(limit)); err != nil {
```
AkihiroSuda: `t.Skip()` on cgroup v2 env

mnitchev: added the skip in the beginning of the test

AkihiroSuda:
LGTM

kolyshkin:
@danail-branekov have you actually seen this (EINTR from ioutils.WriteFile) happening?

I'm asking because Golang 1.14 release notes say
> <...> programs that use packages like syscall or golang.org/x/sys/unix will see more slow system calls fail with EINTR errors.

and from this I would assume that if you're not using e.g. write(2) syscall directly, golang will either handle EINTR for you, or avoid it somehow (lock the goroutine doing the write). The question is, is this a wrong assumption?

kolyshkin:
It seems that
* write(2) should restart in case SA_RESTART flag is set for signal handlers, and
* golang sets all signal handlers with SA_RESTART
so, theoretically, EINTR should not happen during write.

But I don't know for sure.

mnitchev:
Hello @kolyshkin 

Yes, we've been seeing this fail in our CI the past two weeks after an upgrade to Go 1.14. The tests we're running are failing on `runc run`:
```
"runc run: exit status 1: container_linux.go:349: starting container process caused \"process_linux.go:449: container init caused \\\"process_linux.go:415: setting cgroup config for procHooks process caused \\\\\\\"failed to write \\\\\\\\\\\\\\\"67108864\\\\\\\\\\\\\\\" to \\\\\\\\\\\\\\\"/sys/fs/cgroup/memory/system.slice/garden.service/garden/f6e3016d-9e99-4b56-43fe-f27453cc5149/memory.memsw.limit_in_bytes\\\\\\\\\\\\\\\": write /sys/fs/cgroup/memory/system.slice/garden.service/garden/f6e3016d-9e99-4b56-43fe-f27453cc5149/memory.memsw.limit_in_bytes: interrupted system call\\\\\\\"\\\"\""
```
Worth noting is that we've only seen this fail on writing in the memory cgroup.
Regarding the fix. The next line in the [go doc you mentioned](https://github.com/golang/go/blob/master/doc/go1.14.html#L332-L333) says: ` Those programs will have to handle those errors in some way, most likely looping to try the system call again.`, which is why we assumed we needed to add the retry logic. If you run the test we've added without the retry in `fscommon.go` it fails consistently.

kolyshkin:
@mnitchev thank you for the explanation! Reproduced locally, filed an issue to golang: https://github.com/golang/go/issues/38033

AkihiroSuda:
ping @opencontainers/runc-maintainers . We need this to unblock migration to Go 1.14.

hqhq:
Looks like it's regression of golang, but this work around seems harmless.

hqhq:
LGTM

mnitchev:
Do you plan on releasing with this change soon?

AkihiroSuda:
I assume we need changes like this in more places before shipping the next release?

mnitchev:
It's probably worth taking a look, but from our perspective, we've only seen this type of error in this specific place in the codebase (for context our CI runs test every 20 minutes using pretty much all the runc operations). It also looks like the fscommon package is used when writing to all the cgroups.

AkihiroSuda:
@cyphar What do you think about the next release?

Issue #2253, opened 2020-03-11
Container is not destroyed with pre-dump


levanthanh3005:
Hi,
Could you please help me to check the error with runc:
My steps are:
mkdir /mycontainer
cd /mycontainer
mkdir rootfs
docker export $(docker create busybox) | tar -C rootfs -xvf -

runc spec

Then change config.json : as the example in github https://github.com/opencontainers/runc
"terminal": false,
"user": {
"uid": 0,
"gid": 0
},
"args": [
"sleep", "50000"
],

Then go to mycontainer
runc run -d busybox &> /dev/null < /dev/null
runc checkpoint --image-path cp busybox => it works

I reproduce the steps to make the busybox run again, and then run
runc checkpoint --pre-dump --image-path cp busybox
=> ERRO[0000] container is not destroyed

criu version is 3.13
runc version 1.0.0-rc9+dev
commit: c4d8e1688c816a8cef632a3b44a38611511b7140
spec: 1.0.1-dev

levanthanh3005:
Hi @adrianreber 
My complaining is about the behavior, 
"container is not destroyed" is correct, but it should not make any errors, I also try with other service with tcp connection and it makes the same error

adrianreber:
I will have a look at what is up with the message during checkpointing.

PR #2260, opened 2020-03-17
checkpoint: remove error message with --leave-running

adrianreber:
If checkpointing a container with '--leave-running' runc started to print the following message:

 `ERRO[0000] container is not destroyed`

The message is correct, because CRIU did not terminate the process, but as that was requested by the user the message is wrong.

So now the container is only destroyed if the user did not specify '--leave-running'.

Fixes: #2253

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -62,8 +62,15 @@ checkpointed.`,
 		if status == libcontainer.Created || status == libcontainer.Stopped {
 			fatalf("Container cannot be checkpointed in %s state", status.String())
 		}
-		defer destroy(container)
 		options := criuOptions(context)
+		if !options.LeaveRunning || !options.PreDump {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

mrunalp:
LGTM

checkpoint.go
```diff
@@ -62,8 +62,15 @@ checkpointed.`,
 		if status == libcontainer.Created || status == libcontainer.Stopped {
 			fatalf("Container cannot be checkpointed in %s state", status.String())
 		}
-		defer destroy(container)
 		options := criuOptions(context)
+		if !options.LeaveRunning || !options.PreDump {
```
kolyshkin, line 66: @adrianreber I'm afraid there should be `&&` in here.

PR #2262, opened 2020-03-19
libcontainer/cgroups: optimize Get*Cgroup for v2

kolyshkin:
Functions `GetOwnCgroup` and `GetInitCgroup` do this:

1. call `ParseCgroupFile()` to read and parse `/proc/*/cgroup` file
2. call `getControllerPath()`

In the case of cgroup2 unified hierarchy, the results from `ParseCgroupFile`
are not used by `getControllerPath`, so we can avoid the parsing
altogether.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -307,6 +307,10 @@ func GetAllSubsystems() ([]string, error) {
 
 // GetOwnCgroup returns the relative path to the cgroup docker is running in.
 func GetOwnCgroup(subsystem string) (string, error) {
+	if IsCgroup2UnifiedMode() {
+		return "/", nil
+	}
+
 	cgroups, err := ParseCgroupFile("/proc/self/cgroup")
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


kolyshkin:
> I think it is better to return ErrCgroup2Unimplemented

To make sure this code is not used anywhere from the cgroupv2 unified hierarchy code?

It makes sense to split the cgroup and cgroupv2 utils then. I'll take a look

AkihiroSuda:
If it is used in v2 code actually, it should return the actual cgroup path, not "/".

kolyshkin:
> If it is used in v2 code actually, it should return the actual cgroup path, not "/".

Its documentation says
> // GetOwnCgroup returns the relative path to the cgroup docker is running in.

kolyshkin:
Unfortunately, this code is still used by the unified hierarchy code, e.g. in `libcontainer/cgroups/systemd/unified_hierarchy.go`.

kolyshkin:
OK I think I know what needs to be done instead. Will open another PR soon

kolyshkin:
The alternative to this PR is https://github.com/opencontainers/runc/pull/2273/

PR #2264, opened 2020-03-21
Dockerfile: add -f to curl

kolyshkin:
When github is giving HTTP 404 (or other error), it still sends some
document (to be viewed in a browser), but we have tar on the other end
of the pipe, so the end result is like this:

> Step 13/13 : RUN . tests/integration/multi-arch.bash     && curl -o- -sSL `get_busybox` | tar xfJC - ${ROOTFS}
> 4872 ---> Running in ad84646a69f3
> 4873xz: (stdin): File format not recognized
> 4874tar: Child returned status 1
> 4875tar: Error is not recoverable: exiting now
> 4876The command '/bin/sh -c . tests/integration/multi-arch.bash     && curl -o- -sSL `get_busybox` | tar xfJC - ${ROOTFS}' returned a non-zero code: 2

Add -f switch to curl, so it will display an HTTP error and exit:

> curl: (22) The requested URL returned error: 404 Not Found
> xz: (stdin): File format not recognized
> tar: Child returned status 1
> tar: Error is not recoverable: exiting now

While at it, also
 * remove -v from criu untar (too much output)
 * remove -o - from curl (it does the same by default)

@thaJeztah PTAL

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ RUN cd /tmp \
 # install criu
 ENV CRIU_VERSION v3.12
 RUN mkdir -p /usr/src/criu \
-    && curl -sSL https://github.com/checkpoint-restore/criu/archive/${CRIU_VERSION}.tar.gz | tar -v -C /usr/src/criu/ -xz --strip-components=1 \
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

kolyshkin:
> LGTM; want me to include this into #2234?

Yes please

crosbymichael:
LGTM

PR #2267, opened 2020-03-23
Avoid duplicate calls to runner#destroy

tedyu:
From runner#run:
```
	status, err := handler.forward(process, tty, detach)
	if err != nil {
		r.terminate(process)
	}
	if detach {
		return 0, nil
	}
	r.destroy()
	return status, err
```

If forward() returns non-nil error, destroy() would be called twice: once on line 352 and once in the deferred func.

This PR avoids the duplicate call to destroy().

### DIFF
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -349,7 +349,9 @@ func (r *runner) run(config *specs.Process) (int, error) {
 	if detach {
 		return 0, nil
 	}
-	r.destroy()
+	if err == nil {
+		r.destroy()
+	}
 	return status, err
 }
 


crosbymichael:
Can you update this PR to sign your commit with the DCO?  You can see the steps in the contributing doc

tedyu:
@crosbymichael 
I have signed this PR.

utils_linux.go
```diff
@@ -349,7 +349,9 @@ func (r *runner) run(config *specs.Process) (int, error) {
 	if detach {
 		return 0, nil
 	}
-	r.destroy()
+	if err == nil {
+		r.destroy()
+	}
```
AkihiroSuda: Probably you can just drop `r.destroy()` here, and do `defer r.destroy()` without `if err != nil ` at the beginning of this function

tedyu: That would change the behavior for the following case where err == nil and detach == true:
```
	if detach {
		return 0, nil
	}
	r.destroy()
```
In the existing code, r.destroy() is not called in that case.

AkihiroSuda:
LGTM

tedyu:
@crosbymichael @mrunalp 
Can this be merged ?

crosbymichael:
LGTM

PR #2270, opened 2020-03-26
cgroupv2: don't try to set kmem for systemd case

kolyshkin:
To the best of my knowledge, it has been decided to drop the kernel
memory controller from the cgroupv2 hierarchy, so "kernel memory limits"
do not exist if we're using v2 unified.

So, we need to ignore kernel memory setting. This was already done in
non-systemd case (see commit 88e8350de), let's do the same for systemd.

This fixes the following error:

> container_linux.go:349: starting container process caused "process_linux.go:306: applying cgroup configuration for process caused \"open /sys/fs/cgroup/machine.slice/runc-cgroups-integration-test.scope/tasks: no such file or directory\""

@AkihiroSuda @mrunalp PTAL

### DIFF
--- a/libcontainer/cgroups/systemd/unified_hierarchy.go
+++ b/libcontainer/cgroups/systemd/unified_hierarchy.go
@@ -126,13 +126,7 @@ func (m *UnifiedManager) Apply(pid int) error {
 
 	properties = append(properties, c.SystemdProps...)
 
-	// We have to set kernel memory here, as we can't change it once
-	// processes have been attached to the cgroup.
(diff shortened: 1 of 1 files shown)


kolyshkin:
This PR does not come with a test case as I am currently working on updating our existing cgroup integration tests (those in tests/integration/update.bats) for v2, and will do a separate PR once it's ready for review.

AkihiroSuda:
LGTM, thanks

mrunalp:
LGTM

PR #2273, opened 2020-03-26
cgroup v2 cleanups

kolyshkin:
This is a set of small cleanups and fixes for cgroup v2 unified hierarchy + systemd. Please see individual commits for descriptions.

@AkihiroSuda PTAL

### DIFF
--- a/libcontainer/cgroups/systemd/unified_hierarchy.go
+++ b/libcontainer/cgroups/systemd/unified_hierarchy.go
@@ -4,7 +4,6 @@ package systemd
 
 import (
 	"bytes"
-	"fmt"
 	"io/ioutil"
 	"math"
 	"os"
@@ -135,7 +134,7 @@ func (m *UnifiedManager) Apply(pid int) error {
 		return err
 	}
 
-	path, err := getSubsystemPath(m.Cgroups, "")
+	path, err := getv2Path(m.Cgroups)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


kolyshkin:
I hope this is very reviewable and can be merged soon. I have more stuff in the same direction

libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -21,6 +20,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const UnifiedMountpoint = fs2.UnifiedMountpoint
```
AkihiroSuda: Can we just directly use `fs2.UnifiedMountpoint` without aliasing?

libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -197,14 +189,26 @@ func (m *UnifiedManager) GetUnifiedPath() (string, error) {
...
+}
+
 func createCgroupsv2Path(path string) (Err error) {
 	content, err := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
 	if err != nil {
 		return err
 	}
-	if !filepath.HasPrefix(path, "/sys/fs/cgroup") {
-		return fmt.Errorf("invalid cgroup path %s", path)
-	}
```
AkihiroSuda: Why remove?

AkihiroSuda:
A couple of nits, but overall LGTM

libcontainer/cgroups/systemd/unified_hierarchy.go
kolyshkin: Sure, I just thought it's cleaner that way. Will fix

libcontainer/cgroups/systemd/unified_hierarchy.go
kolyshkin: We're calling this function from a single place in the same file with path from `getv2Path()` which clearly sets prefix to `/sys/fs/cgroup/`.

I can reinstate the check but it seems redundant now.

libcontainer/cgroups/systemd/unified_hierarchy.go
kolyshkin: fixed

AkihiroSuda:
LGTM

mrunalp:
LGTM

PR #2268, opened 2020-03-24
update vendor

AkihiroSuda:
Notably, github.com/vishvananda/netlink hadn't been updated since 2015.

---
Also run `make verify-dependencies` with Go 1.14.x

Background: https://github.com/opencontainers/runc/pull/2073#issuecomment-596068439

PR #2276, opened 2020-03-27
cgroupv2: don't use GetCgroupMounts for criu c/r

kolyshkin:
When performing checkpoint or restore of cgroupv2 unified hierarchy,
there is no need to call getCgroupMounts() / cgroups.GetCgroupMounts()
as there's only a single mount in there.

This eliminates the last internal (i.e. runc) use of
`cgroups.GetCgroupMounts()` for v2 unified case. Unfortunately, there
are external ones (e.g. moby/moby) so we can't yet let it
return an error.

@adrianreber PTAL

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1093,6 +1093,11 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 			case "bind":
 				c.addCriuDumpMount(req, m)
 			case "cgroup":
+				if cgroups.IsCgroup2UnifiedMode() {
+					c.addCriuDumpMount(req, m)
+					continue
+				}
+				// cgroup v1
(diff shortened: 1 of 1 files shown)


adrianreber:
Makes sense. :+1:

kolyshkin:
CI failure is unrelated and is fixed by #2268

kolyshkin:
rebased (to include #2268) to fix CI failure

AkihiroSuda:
LGTM

kolyshkin:
Added WIP -- I overlooked another use of `GetPaths()`, will add a commit fixing it

kolyshkin:
> Added WIP -- I overlooked another use of GetPaths(), will add a commit fixing it

Nevermind, it was not in c/r code and can be dealt with separately.

mrunalp:
LGTM

PR #2279, opened 2020-03-31
Actually check for syscall.ENODEV when checking if a container is paused

yulianedyalkova:
It turns out that ioutil.Readfile wraps the error in a *os.PathError.
Since we cannot guarantee compilation with golang >= v1.13, we are
manually unwrapping the error.

Signed-off-by: Kieron Browne <kbrowne@pivotal.io>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1832,7 +1832,10 @@ func (c *linuxContainer) isPaused() (bool, error) {
 	data, err := ioutil.ReadFile(filepath.Join(fcg, filename))
 	if err != nil {
 		// If freezer cgroup is not mounted, the container would just be not paused.
-		if os.IsNotExist(err) || err == syscall.ENODEV {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

crosbymichael:
LGTM

kolyshkin:
Actually, we have switched to golang 1.13+ once we merged https://github.com/opencontainers/runc/pull/2256. Also, since go 1.12 is an unsupported version, it makes no sense to try supporting it.

Can we please use errors.Unwrap() instead (and maybe mark the file with `// +build go1.13`)?

kolyshkin:
Another (non-technical) problem with this commit is it is not signed-off-by by the author (and our DCO check fails to see it). I mean:

> Author: Yulia Nedyalkova <julianedialkova@hotmail.com>
> ....
>     Signed-off-by: Kieron Browne <kbrowne@pivotal.io>

kolyshkin:
Opened https://github.com/opencontainers/runc/pull/2280 to use errors.Unwrap()

PR #2280, opened 2020-04-01
Use errors.Unwrap() where possible

kolyshkin:
This set makes use of errors.Unwrap() where possible to get to an underlying error. The biggest motivation is to simplify the code.

The feature requires go 1.13 but since merging https://github.com/opencontainers/runc/pull/2256 we are already not supporting go 1.12 (which is an unsupported release anyway).

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -110,21 +110,13 @@ func isIgnorableError(rootless bool, err error) bool {
 	if !rootless {
 		return false
 	}
+	err = errors.Cause(err)
 	// Is it an ordinary EPERM?
-	if os.IsPermission(errors.Cause(err)) {
+	if os.IsPermission(err) {
 		return true
 	}
-
-	// Try to handle other errnos.
-	var errno error
-	switch err := errors.Cause(err).(type) {
-	case *os.PathError:
-		errno = err.Err
-	case *os.LinkError:
-		errno = err.Err
-	case *os.SyscallError:
-		errno = err.Err
-	}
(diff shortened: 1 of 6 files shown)


kolyshkin:
Let me know if you want me to merge most of the commits

AkihiroSuda:
LGTM

crosbymichael:
LGTM

kolyshkin:
Oh well, this might not be working as I thought it would. See this example: https://play.golang.org/p/-HyjwI4v_pz

IOW if an error is not wrapped, `errors.Unwrap()` returns `nil`! This is nuts :-\

AkihiroSuda:
We should revert this PR?

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -110,21 +110,13 @@ func isIgnorableError(rootless bool, err error) bool {
...
-	switch err := errors.Cause(err).(type) {
-	case *os.PathError:
-		errno = err.Err
-	case *os.LinkError:
-		errno = err.Err
-	case *os.SyscallError:
-		errno = err.Err
-	}
+	// Handle some specific syscall errors.
+	errno := errors.Unwrap(err)
```
kolyshkin, line 119: This code might not be working the way it should. If `err` is not wrapped, errno will be `nil`.

Will do a followup PR.

kolyshkin:
> We should revert this PR?

@AkihiroSuda while I don't see any code paths that might hit the problem I described above, it's better to not use `errors.Unwrap()` directly, there are better ways. Please see #2291.

PR #2281, opened 2020-04-01
cgroup v2: support rootless systemd

### LINKED ISSUE #2163: cgroup2: does not work with rootless podman
```console
$ podman --cgroup-manager=cgroupfs  run -it --rm --runtime=/usr/local/sbin/runc  docker.io/library/alpine  
Error: time="2019-10-30T00:13:24+09:00" level=warning msg="signal: killed"
time="2019-10-30T00:13:24+09:00" level=error msg="container_linux.go:346: starting container process caused \"process_linux.go:297: applying cgroup configuration for process caused \\\"mkdir /sys/fs/cgroup/libpod_parent/libpod-00b78ede55a4d847e53ddfe1e4c1b5498db0175ab22339705883f6b0dc136734: permission denied\\\"\""
container_linux.go:346: starting container process caused "process_linux.go:297: applying cgroup configuration for process caused \"mkdir /sys/fs/cgroup/libpod_parent/libpod-00b78ede55a4d847e53ddfe1e4c1b5498db0175ab22339705883f6b0dc136734: permission denied\"": OCI runtime permission denied error
```

```console
$ podman --cgroup-manager=systemd  run -it --rm --runtime=/usr/local/sbin/runc  docker.io/library/alpine
Error: time="2019-10-30T00:13:31+09:00" level=error msg="systemd cgroup flag passed, but systemd support for managing cgroups is not available"
systemd cgroup flag passed, but systemd support for managing cgroups is not available: OCI runtime error
```

Rootful mode seems working without problem (both cgroupfs and systemd).

* runc: d239ca84254c579a964101b4f8049b860fbfc135
* Podman: containers/libpod@a56131fef4a82824b397c94f6b7edcdb24769624
* Host: Ubuntu 19.10, kernel `5.3.0-19-generic #20-Ubuntu` with `systemd.unified_cgroup_hierarchy=1 cgroup_enable=memory swapaccount=1`

AkihiroSuda:
Tested with both Podman (master) and Moby (master), on Ubuntu 19.10 .

```console
$ podman --cgroup-manager=systemd run -it --rm --runtime=runc \
  --cgroupns=host --memory 42m --cpus 0.42 --pids-limit 42 alpine
/ # cat /proc/self/cgroup
0::/user.slice/user-1001.slice/user@1001.service/user.slice/libpod-132ff0d72245e6f13a3bbc6cdc5376886897b60ac59eaa8dea1df7ab959cbf1c.scope
/ # cat
/sys/fs/cgroup/user.slice/user-1001.slice/user@1001.service/user.slice/libpod-132ff0d72245e6f13a3bbc6cdc5376886897b60ac59eaa8dea1df7ab959cbf1c.scope/memory.max
44040192
/ # cat
/sys/fs/cgroup/user.slice/user-1001.slice/user@1001.service/user.slice/libpod-132ff0d72245e6f13a3bbc6cdc5376886897b60ac59eaa8dea1df7ab959cbf1c.scope/cpu.max
42000 100000
/ # cat
/sys/fs/cgroup/user.slice/user-1001.slice/user@1001.service/user.slice/libpod-132ff0d72245e6f13a3bbc6cdc5376886897b60ac59eaa8dea1df7ab959cbf1c.scope/pids.max
42
```

The behavior corresponds to crun v0.13.

To test with Moby, launch `dockerd-rootless.sh` with `--exec-opt native.cgroupdriver=systemd`.

Fix #2163

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,9 @@ matrix:
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration RUNC_USE_SYSTEMD=yes'
         # same setup but with fs2 driver instead of systemd
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration'
-        # rootless
+        # cgroupv2+systemd (rootless)
+        - sudo ssh default -t 'cd /vagrant && sudo make localrootlessintegration RUNC_USE_SYSTEMD=yes'
+        # same setup but with fs2 driver (rootless) instead of systemd
         - sudo ssh default -t 'cd /vagrant && sudo make localrootlessintegration'
   allow_failures:
     - go: tip
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,9 +28,25 @@ EOF
     # Add a user for rootless tests
     useradd -u2000 -m -d/home/rootless -s/bin/bash rootless
 
+    # Allow root to execute `ssh rootless@localhost` in tests/rootless.sh
+    ssh-keygen -t ecdsa -N "" -f /root/rootless.key
+    mkdir -m 0700 -p /home/rootless/.ssh
+    cat /root/rootless.key.pub >> /home/rootless/.ssh/authorized_keys
+    chown -R rootless.rootless /home/rootless
+
     # Add busybox for libcontainer/integration tests
     . /vagrant/tests/integration/multi-arch.bash \
         && mkdir /busybox \
         && curl -fsSL $(get_busybox) | tar xfJC - /busybox
+
+    # Delegate cgroup v2 controllers to rootless user via --systemd-cgroup
+    mkdir -p /etc/systemd/system/user@.service.d
+    cat > /etc/systemd/system/user@.service.d/delegate.conf << EOF
+[Service]
+# default: Delegate=pids memory
+# NOTE: delegation of cpuset requires systemd >= 244 (Fedora >= 32, Ubuntu >= 20.04). cpuset is ignored on Fedora 31.
+Delegate=cpu cpuset io memory pids
+EOF
+    systemctl daemon-reload
   SHELL
 end
--- a/libcontainer/cgroups/systemd/user.go
+++ b/libcontainer/cgroups/systemd/user.go
@@ -1,0 +1,106 @@
+// +build linux
+
+package systemd
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+	dbus "github.com/godbus/dbus/v5"
+	"github.com/opencontainers/runc/libcontainer/system"
+	"github.com/pkg/errors"
+)
+
+// NewUserSystemdDbus creates a connection for systemd user-instance.
+func NewUserSystemdDbus() (*systemdDbus.Conn, error) {
+	addr, err := DetectUserDbusSessionBusAddress()
+	if err != nil {
+		return nil, err
+	}
+	uid, err := DetectUID()
+	if err != nil {
+		return nil, err
+	}
+
+	return systemdDbus.NewConnection(func() (*dbus.Conn, error) {
+		conn, err := dbus.Dial(addr)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error while dialing %q", addr)
+		}
+		methods := []dbus.Auth{dbus.AuthExternal(strconv.Itoa(uid))}
+		err = conn.Auth(methods)
+		if err != nil {
+			conn.Close()
+			return nil, errors.Wrapf(err, "error while authenticating connection, address=%q, UID=%d", addr, uid)
+		}
+		if err = conn.Hello(); err != nil {
+			conn.Close()
+			return nil, errors.Wrapf(err, "error while sending Hello message, address=%q, UID=%d", addr, uid)
+		}
+		return conn, nil
+	})
+}
+
+// DetectUID detects UID from the OwnerUID field of `busctl --user status`
+// if running in userNS. The value corresponds to sd_bus_creds_get_owner_uid(3) .
+//
+// Otherwise returns os.Getuid() .
+func DetectUID() (int, error) {
+	if !system.RunningInUserNS() {
+		return os.Getuid(), nil
+	}
+	b, err := exec.Command("busctl", "--user", "--no-pager", "status").CombinedOutput()
+	if err != nil {
+		return -1, errors.Wrap(err, "could not execute `busctl --user --no-pager status`")
+	}
+	scanner := bufio.NewScanner(bytes.NewReader(b))
+	for scanner.Scan() {
+		s := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(s, "OwnerUID=") {
+			uidStr := strings.TrimPrefix(s, "OwnerUID=")
+			i, err := strconv.Atoi(uidStr)
+			if err != nil {
+				return -1, errors.Wrapf(err, "could not detect the OwnerUID: %s", s)
+			}
+			return i, nil
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return -1, err
+	}
+	return -1, errors.New("could not detect the OwnerUID")
+}
+
+// DetectUserDbusSessionBusAddress returns $DBUS_SESSION_BUS_ADDRESS if set.
+// Otherwise returns "unix:path=$XDG_RUNTIME_DIR/bus" if $XDG_RUNTIME_DIR/bus exists.
+// Otherwise parses the value from `systemctl --user show-environment` .
+func DetectUserDbusSessionBusAddress() (string, error) {
+	if env := os.Getenv("DBUS_SESSION_BUS_ADDRESS"); env != "" {
+		return env, nil
+	}
+	if xdr := os.Getenv("XDG_RUNTIME_DIR"); xdr != "" {
+		busPath := filepath.Join(xdr, "bus")
+		if _, err := os.Stat(busPath); err == nil {
+			busAddress := "unix:path=" + busPath
+			return busAddress, nil
+		}
+	}
+	b, err := exec.Command("systemctl", "--user", "--no-pager", "show-environment").CombinedOutput()
+	if err != nil {
+		return "", errors.Wrapf(err, "could not execute `systemctl --user --no-pager show-environment`, output=%q", string(b))
+	}
+	scanner := bufio.NewScanner(bytes.NewReader(b))
+	for scanner.Scan() {
+		s := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(s, "DBUS_SESSION_BUS_ADDRESS=") {
+			return strings.TrimPrefix(s, "DBUS_SESSION_BUS_ADDRESS="), nil
+		}
+	}
+	return "", errors.New("could not detect DBUS_SESSION_BUS_ADDRESS from `systemctl --user --no-pager show-environment`")
+}
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -14,6 +15,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/pkg/errors"
 )
 
 type unifiedManager struct {
@@ -89,7 +91,6 @@ func (m *unifiedManager) Apply(pid int) error {
 	var (
 		c          = m.cgroups
 		unitName   = getUnitName(c)
-		slice      = "system.slice"
 		properties []systemdDbus.Property
 	)
 
@@ -97,6 +98,10 @@ func (m *unifiedManager) Apply(pid int) error {
 		return cgroups.WriteCgroupProc(m.path, pid)
 	}
 
+	slice := "system.slice"
+	if m.rootless {
+		slice = "user.slice"
+	}
 	if c.Parent != "" {
 		slice = c.Parent
 	}
@@ -140,8 +145,12 @@ func (m *unifiedManager) Apply(pid int) error {
 	properties = append(properties, resourcesProperties...)
 	properties = append(properties, c.SystemdProps...)
 
-	if err := startUnit(unitName, properties); err != nil {
+	dbusConnection, err := getDbusConnection(m.rootless)
+	if err != nil {
 		return err
+	}
+	if err := startUnit(dbusConnection, unitName, properties); err != nil {
+		return errors.Wrapf(err, "error while starting unit %q with properties %+v", unitName, properties)
 	}
 
 	_, err = m.GetUnifiedPath()
@@ -161,13 +170,17 @@ func (m *unifiedManager) Destroy() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	dbusConnection, err := getDbusConnection(m.rootless)
+	if err != nil {
+		return err
+	}
 	unitName := getUnitName(m.cgroups)
-	if err := stopUnit(unitName); err != nil {
+	if err := stopUnit(dbusConnection, unitName); err != nil {
 		return err
 	}
 
 	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
+	err = os.Remove(m.path)
 	if err != nil && !os.IsNotExist(err) {
 		return err
 	}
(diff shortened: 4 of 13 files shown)


AkihiroSuda:
@giuseppe @filbranden @mrunalp PTAL

libcontainer/cgroups/systemd/user.go
```diff
@@ -0,0 +1,86 @@
...
+	if xdr == "" {
+		return -1, errors.New("could not detect the UID from XDG_RUNTIME_DIR")
+	}
+	s := strings.TrimPrefix(filepath.Clean(xdr), "/run/user/")
+	i, err := strconv.Atoi(s)
+	if err != nil {
+		return -1, errors.Wrap(err, "could not detect the UID from XDG_RUNTIME_DIR")
+	}
+	return i, nil
+}
```
giuseppe, line 66: how will it work with something like `XDG_RUNTIME_DIR=/tmp/foo`?

AkihiroSuda: Who will set such value for systemd-mode?

kolyshkin: Maybe it's slightly better to assume that XDG_RUNTIME_DIR ends up with a slash and a user id, i.e. something like

```go
s := path.Base(xdr)
```

That would still be an assumption, but less strict that the path must be `/run/user/$ID`

libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -209,10 +218,18 @@ func getv2Path(c *configs.Cgroup) (string, error) {
 		return "", err
 	}
 
+	if rootless {
+		uid, err := detectUID()
+		if err != nil {
+			return "", err
+		}
+		slice = fmt.Sprintf("user.slice/user-%d.slice/user@%d.service/%s", uid, uid, slice)
```
filbranden, line 226: Am I missing something or this will end up with `user.slice` twice in the name?

Like, when `c.Parent` is not defined, this will end up as `user.slice/user-1000.slice/user@1000.service/user.slice`, no?

If the idea is to have a slice of the user manager for these containers (which is definitely a good idea!), maybe it should have a name other than `user.slice` to avoid confusion? Perhaps `containers.slice`? I do think `machine.slice` is not a terrible idea for this (also a good one for Rootfull containers.)

filbranden, line 226: You can also query this path from the D-Bus interface:

```
$ busctl --user get-property org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager ControlGroup
s "/user.slice/user-1001.slice/user@1001.service"
```

I think that's better, having systemd resolve that for you. Also future-proof if this path ever changes. (It has changed in the past.)

libcontainer/cgroups/systemd/user.go
```diff
@@ -0,0 +1,86 @@
...
+
+// detectUserDbusSessionBusAddress returns $DBUS_SESSION_BUS_ADDRESS if set.
+// Otherwise parse the value from `systemctl --user show-environment` .
+func detectUserDbusSessionBusAddress() (string, error) {
+	if env := os.Getenv("DBUS_SESSION_BUS_ADDRESS"); env != "" {
+		return env, nil
+	}
+	b, err := exec.Command("systemctl", "--user", "show-environment").CombinedOutput()
```
filbranden, line 74: If your environment variables are borked, this won't work either...

```
$ DBUS_SESSION_BUS_ADDRESS=broken XDG_RUNTIME_DIR=/broken systemctl --user show-environment
Failed to connect to bus: Connection refused
```

The code used to connect to the user bus is [here](https://github.com/systemd/systemd/blob/v245/src/libsystemd/sd-bus/sd-bus.c#L1327-L1333), it depends on either `$DBUS_SESSION_BUS_ADDRESS` or `$XDG_RUNTIME_DIR` being properly set.

My 2¢ is that this should simply fail if the `$DBUS_SESSION_BUS_ADDRESS` environment variable is not available.

libcontainer/cgroups/systemd/user.go
```diff
@@ -0,0 +1,86 @@
...
+			conn.Close()
+			return nil, err
+		}
+		return conn, nil
+	})
+}
+
+// detectUID detects UID from $XDG_RUNTIME_DIR if running in userNS.
+// Otherwise returns os.Getuid() .
+func detectUID() (int, error) {
```
filbranden, line 51: I'm hoping you wouldn't need this... But looks like you still do for the external auth of the user D-Bus connection... Let me think about this one see if there's a better solution than this.

Having said that, see my comment below on requiring that the environment variables are properly set. So you could say that this is a hard requirement here. Looks like your code will fail if it doesn't match `/run/user/%d` which I think is a good thing.

libcontainer/cgroups/systemd/unified_hierarchy.go
AkihiroSuda: Yes, this is intentional for crun compatibility.

libcontainer/cgroups/systemd/user.go
AkihiroSuda: libsystemd (which is used by crun) seems extracting UID from the `"0::/user.slice/user-%d.slice/..."` string  in `/proc/self/cgroup`:
https://github.com/systemd/systemd/blob/6829d8ce6927c151e4391c6a5e69b76906fbd2be/src/libsystemd/sd-bus/sd-bus.c#L104
https://github.com/systemd/systemd/blob/ce4c73eb5d3c9a23628c450db0e25a529620adbe/src/basic/cgroup-util.c#L1361-L1395

I feel this seems less robust than `$XDG_RUNTIME_DIR`, but maybe we should do this?

libcontainer/cgroups/systemd/user.go
AkihiroSuda: Let's discuss this in https://github.com/opencontainers/runc/pull/2281#discussion_r401922044

libcontainer/cgroups/systemd/user.go
AkihiroSuda: `systemctl --user show-environment` works for me with `unset DBUS_SESSION_BUS_ADDRESS`. 

Actually, podman sets `XDG_RUNTIME_DIR` but does not set `DBUS_SESSION_BUS_ADDRESS`.

libcontainer/cgroups/systemd/user.go
AkihiroSuda: I'm :+1: for `path.Base(xdr)` proposed by @kolyshkin https://github.com/opencontainers/runc/pull/2281#discussion_r401778383

libcontainer/cgroups/systemd/unified_hierarchy.go
AkihiroSuda: Thanks, updated

libcontainer/cgroups/systemd/user.go
AkihiroSuda: Yet another way is to exec `busctl --user status` and grep `^UID=`

filbranden: `busctl --user status` ends up resolving it from `getsockopt(fd, SOL_SOCKET, SO_PEERCRED, ...)`, see [here](https://github.com/systemd/systemd/blob/v245/src/basic/socket-util.c#L735).

I believe you can do the same in Golang with `unix.GetsockoptUcred(int(fd), unix.SOL_SOCKET, unix.SO_PEERCRED)` from the `"golang.org/x/sys/unix"` package. Does that work?

filbranden: > libsystemd (which is used by crun) seems extracting UID from the `"0::/user.slice/user-%d.slice/..."` string in `/proc/self/cgroup`:

Yes, I think that's possibly reasonable. It's still reimplementing C code in Golang but I think that's preferrable than some of the alternatives.

But see my message above about getting the real UID from a `SO_PEERCRED` call...

libcontainer/cgroups/systemd/user.go
filbranden: Ah I see!

The [logic in sd-bus](https://github.com/systemd/systemd/blob/v245/src/libsystemd/sd-bus/bus-internal.h#L21) falls back to parsing `$XDG_RUNTIME_DIR`, so maybe use that same logic here too?

I think that's a superior approach to spawning an external command.

(BTW, @giuseppe maybe you'd want to address the issue of Podman exporting `$XDG_RUNTIME_DIR` but not `$DBUS_SESSION_BUS_ADDRESS` if that's applicable? Maybe it's not really applicable, just pinging you here in case you think it's worth looking into...)

libcontainer/cgroups/systemd/user.go
AkihiroSuda: > busctl --user status ends up resolving it from getsockopt(fd, SOL_SOCKET, SO_PEERCRED, ...), see here.

Seems not true for `OwnerUID`. `OwnerUID` seems extracted from cgroup.
https://github.com/systemd/systemd/blob/v245/src/libsystemd/sd-bus/bus-dump.c#L375-L377
https://github.com/systemd/systemd/blob/v245/src/libsystemd/sd-bus/bus-creds.c#L509-L526

Anyway, `SO_PEERCRED` isn't what we want here, because the kernel automatically translates the UID in the ucred struct.

AkihiroSuda: Updated PR to use `busctl --user status`. I think this is the most robust way. 

I guess eventually systemd may switch away from parsing cgroup and adopt more robust logic , but we won't be affected as long as we use `busctl`.

libcontainer/cgroups/systemd/user.go
AkihiroSuda: updated to use `$XDG_RUNTIME_DIR/bus`

libcontainer/cgroups/systemd/user.go
filbranden: Ah yes good point sorry my search was too shallow for it...

I think I just worry a bit about `busctl` not being available, it was first introduced as a debugging tool so I can see how distributions might not want to ship it or not ship it by default and it might go missing in the future...

Anyways, it's a fallback and we can iterate on it. Looks good.

filbranden:
LGTM

kolyshkin:
@AkihiroSuda I have a WIP progress patchset to remove Paths map from the systemd UnifiedManager, and simplify its initialization (meaning, for example, that you won't have to add `Rootless` as exported field). I'm not ready to create a PR yet but it looks like adding rootless support on top of my patch will be very easy. Can you please take a look?

Again, it's not ready yet but you can get the idea of what I'm trying to do:
https://github.com/kolyshkin/runc/commits/fs2-init-ctrl

AkihiroSuda:
@kolyshkin SGTM, do you want your commits to be merged before this PR or after?

kolyshkin:
@AkihiroSuda I think yours will be easier to port on top of mine than vise versa

AkihiroSuda:
If #2299 takes longer, can we merge this PR first?

kolyshkin:
> If #2299 takes longer, can we merge this PR first?

Sorry for the delay, it is ready now, and it will definitely be easier to work with the updated/refactored codebase.

kolyshkin:
Need a rebase

AkihiroSuda:
Needs https://github.com/opencontainers/runc/pull/2340

AkihiroSuda:
Also needs https://github.com/opencontainers/runc/pull/2331

.travis.yml
```diff
@@ -36,7 +36,10 @@ matrix:
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration RUNC_USE_SYSTEMD=yes'
         # same setup but with fs2 driver instead of systemd
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration'
-        # rootless
+        # cgroupv2+systemd (rootless)
+        # NOTE: setenforce 0 is required due to a `machinectl shell` issue: https://bugzilla.redhat.com/show_bug.cgi?id=1788616
```
AkihiroSuda, line 40: @rhatdan Any chance to take a look at https://bugzilla.redhat.com/show_bug.cgi?id=1788616 ?

rhatdan: This should not be recorded in the runc .travis file.  It has nothing to do with this.  There is no expectation that machinectl shell will work with runc.  It only works with systemd-nspawn at this point with a full systemd running inside.

AkihiroSuda: machinectl is used for creating user session for testing in rootless.sh

AkihiroSuda: Or, is it better to use `ssh rootless@localhost` for creating the session?

rhatdan: I would think that is more natural, if you want to match what users really do.

AkihiroSuda: updated

AkihiroSuda:
rebased

Vagrantfile
```diff
@@ -45,5 +51,15 @@ EOF
         && make install-criu \
         && cd - \
         && rm -rf /usr/src/criu
+
+    # Delegate cgroup v2 controllers to rootless user via --systemd-cgroup
+    mkdir -p /etc/systemd/system/user@.service.d/delegate.conf
+    cat > /etc/systemd/system/user@.service.d/delegate.conf << EOF
+[Service]
+# default: Delegate=pids memory
+# NOTE: delegation of cpuset requires systemd >= 244 (Fedora >= 32, Ubuntu >= 20.04). cpuset is ignored on Fedora 31.
```
AkihiroSuda, line 60: I tried to bump up Vagrantfile to use Fedora 32, but could not compile criu on Fedora 32.
Let me try that in yet another PR.

AkihiroSuda:
CI green. Ready to review/merge.

tests/rootless.sh
```diff
@@ -143,6 +147,13 @@ do
 	set -e
 	echo path: $PATH
 	export ROOTLESS_FEATURES="$enabled_features"
-	sudo -HE -u rootless PATH="$PATH" bats -t "$ROOT/tests/integration$ROOTLESS_TESTPATH"
+	if [[ -n "${RUNC_USE_SYSTEMD}" ]] ; then
+		# We use `ssh rootless@localhost` instaed of `sudo -u rootless` for creating systemd user session.
```
kolyshkin, line 151: typo: s/instaed/instead/

AkihiroSuda: fixed

Vagrantfile
AkihiroSuda: https://github.com/opencontainers/runc/pull/2363 is merged now

tests/integration/cgroups.bats
```diff
@@ -79,6 +79,8 @@ EOF
 @test "runc create (rootless + no limits + cgrouppath + no permission) fails with permission error" {
     requires rootless
     requires rootless_no_cgroup
+# systemd controls the permission, so error does not happen
+    requires no_systemd
```
h-vetinari, line 83: Comment missing indentation.

tests/integration/cgroups.bats
```diff
@@ -90,6 +92,8 @@ EOF
 @test "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error" {
     requires rootless
     requires rootless_no_cgroup
+# systemd controls the permission, so error does not happen
+    requires no_systemd
```
h-vetinari, line 96: Same here.

tests/integration/cgroups.bats
AkihiroSuda: fixed

tests/integration/cgroups.bats
AkihiroSuda: fixed

AkihiroSuda:
@kolyshkin @mrunalp PTAL

libcontainer/cgroups/systemd/user.go
```diff
@@ -0,0 +1,103 @@
...
+
+// DetectUID detects UID from the OwnerUID field of `busctl --user status`
+// if running in userNS. The value corresponds to sd_bus_creds_get_owner_uid(3) .
+//
+// Otherwise returns os.Getuid() .
+func DetectUID() (int, error) {
+	if !system.RunningInUserNS() {
+		return os.Getuid(), nil
+	}
+	b, err := exec.Command("busctl", "--user", "--no-pager", "status").CombinedOutput()
```
mrunalp, line 58: could we make a dbus call instead of execing here?

mrunalp: I am reading up on the previous conversation. This is fine for now.

AkihiroSuda: Calling dbus requires this function itself :)

AkihiroSuda:
rebased

AkihiroSuda:
@kolyshkin LGTY?

libcontainer/cgroups/systemd/user.go
kolyshkin: :chicken:&:egg:  problem :)

libcontainer/cgroups/systemd/user.go
```diff
@@ -0,0 +1,103 @@
...
+		if strings.HasPrefix(s, "OwnerUID=") {
+			uidStr := strings.TrimPrefix(s, "OwnerUID=")
+			i, err := strconv.Atoi(uidStr)
+			if err != nil {
+				return -1, errors.Wrapf(err, "could not detect the OwnerUID: %s", s)
+			}
+			return i, nil
+		}
+	}
+	return 0, nil
```
kolyshkin, line 74: nit: `return 0, scanner.Err()`

libcontainer/cgroups/systemd/v2.go
```diff
@@ -97,6 +99,9 @@ func (m *unifiedManager) Apply(pid int) error {
 		return cgroups.WriteCgroupProc(m.path, pid)
 	}
 
+	if m.rootless {
```
kolyshkin, line 102: nit: it would be better for readability to have it like
```go
slice := "system.slice"
if m.rootless {
  slice = "user.slice"
}
```

libcontainer/cgroups/systemd/v2.go
```diff
@@ -140,16 +145,20 @@ func (m *unifiedManager) Apply(pid int) error {
...
+		return errors.Wrapf(err, "error while starting unit %q with properties %+v", unitName, properties)
+	}
 
 	_, err = m.GetUnifiedPath()
 	if err != nil {
 		return err
 	}
 	if err := fs2.CreateCgroupPath(m.path, m.cgroups); err != nil {
-		return err
+		return errors.Wrapf(err, "error while creating directory %q", m.path)
```
kolyshkin, line 161: In many cases (except one) you'll end up printing the directory name twice here.

It's better to modify `fs2.CreateCgroupPath` to make sure path is always present in the error message (need to wrap error returned from `neededControllers` in there), and don't do any wrapping here.

kolyshkin:
Left a few nits, but can be addressed as a followup

LGTM

libcontainer/cgroups/systemd/user.go
AkihiroSuda: updated

libcontainer/cgroups/systemd/v2.go
AkihiroSuda: updated

libcontainer/cgroups/systemd/v2.go
AkihiroSuda: updated

AkihiroSuda:
addressed comments

kolyshkin:
still LGTM

kolyshkin:
LGTM (for pullapprove)

AkihiroSuda:
rebased

mrunalp:
LGTM

kolyshkin:
LGTM

PR #2282, opened 2020-04-01
libct/isPaused: don't use GetPaths from v2 code

kolyshkin:
Using `GetPaths` from cgroupv2 unified hierarchy code is deprecated
and this function will (hopefully) be removed.

Use `GetUnifiedPath()` for v2 case.

There are more cases like this that we still need to fix, but this one is a low hanging fruit :)

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1822,14 +1822,23 @@ func (c *linuxContainer) runType() Status {
 }
 
 func (c *linuxContainer) isPaused() (bool, error) {
-	fcg := c.cgroupManager.GetPaths()["freezer"]
-	if fcg == "" {
-		// A container doesn't have a freezer cgroup
-		return false, nil
-	}
-	pausedState := "FROZEN"
-	filename := "freezer.state"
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

mrunalp:
LGTM

libcontainer/container_linux.go
```diff
@@ -1822,14 +1822,23 @@ func (c *linuxContainer) runType() Status {
...
-	if cgroups.IsCgroup2UnifiedMode() {
+	var fcg, filename, pausedState string
+
+	if !cgroups.IsCgroup2UnifiedMode() {
+		fcg = c.cgroupManager.GetPaths()["freezer"]
+		if fcg == "" {
+			// A container doesn't have a freezer cgroup
+			return false, nil
+		}
+		filename = "freezer.state"
```
tedyu, line 1833: Just curious: the filenames are different for the unified versus non-unified cases. I assume this is intentional.

kolyshkin: @tedyu yes, cgroups v2 have different filenames for various controllers.

PR #2283, opened 2020-04-01
isPathInPrefixList return value should be reverted

tedyu:

Currently isPathInPrefixList returns false if match is found in the prefix list.

This PR reverts the return value so that the return value reflects the natural meaning of the method name.


AkihiroSuda:
LGTM

adrianreber:
Thanks. LGTM. :+1: 

This was actually mentioned in the original pull requests after it was merged and I totally forgot about it. Thanks for fixing.

crosbymichael:
LGTM

PR #2284, opened 2020-04-01
Avoid double close of criuServer

tedyu:

In linuxContainer#criuSwrk, criuServer is potentially closed twice.
See line 1468:
```
	defer criuServer.Close()
```
and the criuServer.Close() call following cmd.Start().

This PR removed the duplicate Close() call.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1488,7 +1488,6 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 	if err := cmd.Start(); err != nil {
 		return err
 	}
-	criuServer.Close()
 	// cmd.Process will be replaced by a restored init.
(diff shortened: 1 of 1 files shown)


kolyshkin:
@adrianreber PTAL

adrianreber:
LGTM. Thanks. Just recently I was confused by that when looking at the code. Good to have it cleaned up.

AkihiroSuda:
LGTM

crosbymichael:
LGTM

avagin:
> This PR removed the duplicate Close() call.

Why is this bad? In golang, this is completely fine, the close syscall will be called only once. But this patch is completely wrong. Now if CRIU will crash or unexpectedly exits, runsc will hang on reading from criuClient, because criuServer has not been closed in the current process.

tedyu:
To be restored via #2293

PR #2285, opened 2020-04-02
cgroupv2: set mem+swap to max if memory is set to max

kolyshkin:
In cgroupv1, command `runc update $ID --memory -1` sets both
memory and memory+swap to -1 (aka unlimited). This was introduced
by commit 18ebc51b3cc7 (Reset Swap when memory is set to unlimited,
Oct 19 2016).

This is not the case for cgroupv2. Fix it.

References:
 - https://github.com/opencontainers/runc/pull/1127
 - https://github.com/moby/moby/pull/22578

### DIFF
--- a/libcontainer/cgroups/fs2/memory.go
+++ b/libcontainer/cgroups/fs2/memory.go
@@ -32,13 +32,21 @@ func numToStr(value int64) (ret string) {
 }
 
 func setMemory(dirPath string, cgroup *configs.Cgroup) error {
-	if val := numToStr(cgroup.Resources.MemorySwap); val != "" {
-		if err := fscommon.WriteFile(dirPath, "memory.swap.max", val); err != nil {
+	mem := numToStr(cgroup.Resources.Memory)
+	memSwap := numToStr(cgroup.Resources.MemorySwap)
+	if mem == "max" {
+		// If the memory is set to max, set the mem+swap to max, too
+		memSwap = "max"
+	}
+
+	if memSwap != "" {
+		if err := fscommon.WriteFile(dirPath, "memory.swap.max", memSwap); err != nil {
 			return err
 		}
 	}
-	if val := numToStr(cgroup.Resources.Memory); val != "" {
-		if err := fscommon.WriteFile(dirPath, "memory.max", val); err != nil {
+
+	if mem != "" {
+		if err := fscommon.WriteFile(dirPath, "memory.max", mem); err != nil {
 			return err
 		}
 	}


kolyshkin:
The motivation for this PR is to make cgroupv2 behavior in this regard compatible with v1. Whether this v1 behavior is good or not I can't yet assess. Even after reading all the discussions linked above, I still do not understand whether it is useful to have unlimited RAM but limited RAM+swap. Probably not, but I'm not quite sure.

AkihiroSuda:
Do we need this for systemd driver as well?

AkihiroSuda:
cc @giuseppe

kolyshkin:
> Do we need this for systemd driver as well?

@AkihiroSuda well, systemd driver

1. does not set `MemorySwapMax=` for systemd unit via dbus (yet? I only found it today and need to take a closer look), it is only set via fsManager.
2. does not changes the systemd unit properties on `update` (IOW the `Set()` methodonly uses fs2 to update the limits).

I found both of these items today, and they both probably need to be taken care of (and I will file issues and try to fix this), but as of now, this PR fixes the issue for both fs2 and systemd drivers.

libcontainer/cgroups/fs2/memory.go
```diff
@@ -32,13 +32,21 @@ func numToStr(value int64) (ret string) {
...
-		if err := fscommon.WriteFile(dirPath, "memory.swap.max", val); err != nil {
+	mem := numToStr(cgroup.Resources.Memory)
+	memSwap := numToStr(cgroup.Resources.MemorySwap)
+	if mem == "max" {
+		// If the memory is set to max, set the mem+swap to max, too
+		memSwap = "max"
+	}
+
+	if memSwap != "" {
+		if err := fscommon.WriteFile(dirPath, "memory.swap.max", memSwap); err != nil {
```
giuseppe, line 43: to keep the same cgroup v1 behaviour (as the config.json file is configured for), we need to set the memory swap to `requested_memory_limit - requested_swap_limit`.  Since on cgroup v1 the swap limit accounts also for the memory

kolyshkin:
> 1. does not set `MemorySwapMax=` for systemd unit via dbus (yet? I only found it today and need to take a closer look), it is only set via fsManager.

created https://github.com/opencontainers/runc/pull/2288

> 2. does not changes the systemd unit properties on `update` (IOW the `Set()` methodonly uses fs2 to update the limits).

Filed https://github.com/opencontainers/runc/issues/2287

libcontainer/cgroups/fs2/memory.go
kolyshkin: Hmm, in this case maybe we don't need to touch swap at all, and this PR should be closed, right?

giuseppe: yes setting the `memory.max` should be enough on cgroup v2 as the swap max should not have any effect on it.

giuseppe: but since the PR is open, I think we could use it for setting the swap limit as it is expected to work on cgroup v1

kolyshkin: See #2288

kolyshkin:
This PR is wrong. Closed in favor of #2288

PR #2288, opened 2020-04-02
cgroupv2: fix setting MemorySwap

kolyshkin:
The resources.MemorySwap field from OCI is memory+swap, while cgroupv2
has a separate swap limit, so subtract memory from the limit (and make
sure values are set and sane).
    
Make sure to set MemorySwapMax for systemd, too. Since systemd does not
have MemorySwapMax for cgroupv1, it is only needed for v2 driver.

* v2: return -1 on any negative value, add unit test
* v3: treat any negative value other than -1 as error

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -623,3 +623,25 @@ func ConvertCPUQuotaCPUPeriodToCgroupV2Value(quota int64, period uint64) string 
 	}
 	return fmt.Sprintf("%d %d", quota, period)
 }
+
+// ConvertMemorySwapToCgroupV2Value converts MemorySwap value from OCI spec
+// for use by cgroup v2 drivers. A conversion is needed since Resources.MemorySwap
+// is defined as memory+swap combined, while in cgroup v2 swap is a separate value.
+func ConvertMemorySwapToCgroupV2Value(memorySwap, memory int64) (int64, error) {
+	if memorySwap == -1 || memorySwap == 0 {
+		// -1 is "max", 0 is "unset", so treat as is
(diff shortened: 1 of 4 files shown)


kolyshkin:
@giuseppe pointed out that in v1 this is mem+swap while in v2 this is swap only, so this might be wrong. Will take another look tomorrow.

kolyshkin:
@AkihiroSuda @giuseppe @mrunalp PTAL

libcontainer/cgroups/utils.go
```diff
@@ -623,3 +623,22 @@ func ConvertCPUQuotaCPUPeriodToCgroupV2Value(quota int64, period uint64) string
 	}
 	return fmt.Sprintf("%d %d", quota, period)
 }
+
+// ConvertMemorySwapToCgroupV2Value converts MemorySwap value for cgroup v2
+// drivers. Since Resources.MemorySwap is define as memory+swap combined,
+// while in cgroup v2 swap is a separate value, a conversion is needed.
+func ConvertMemorySwapToCgroupV2Value(memorySwap, memory int64) (int64, error) {
```
AkihiroSuda, line 630: Please add UT

kolyshkin: done

AkihiroSuda:
LGTM

kolyshkin:
I was wrong in treating any negative value as "max". In fact only `-1` should be treated as "max", and all other negative values are invalid.

Patch updated.

AkihiroSuda:
LGTM

crosbymichael:
LGTM

PR #2290, opened 2020-04-03
cleanupTmp doesn't need to return error

tedyu:

Since the return value from cleanupTmp is not checked, this PR drops the error return.


AkihiroSuda:
`defer cleanupTmp(tmpdir)` doesn't catch error anyway

tedyu:
@AkihiroSuda 
I have updated the PR.

PR #2291, opened 2020-04-03
Use errors.As() and errors.Is() to unwrap errors

kolyshkin:
My previous attempt (#2280) was a bit immature (though it worked).

Using errors.Unwrap() is not the best thing to do, since it returns
nil in case of an error which was not wrapped. More to say, 
errors package provides more elegant ways to check for underlying
errors, such as errors.As() and errors.Is().

Revert #2280, and use errors.Is() and errors.As()
    
Make use of errors.Is() and errors.As() where appropriate to check
the underlying error. The biggest motivation is to simplify the code.

The feature requires go 1.13 but since merging #2256 we are already
not supporting go 1.12 (which is an unsupported release anyway).

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -110,13 +110,21 @@ func isIgnorableError(rootless bool, err error) bool {
 	if !rootless {
 		return false
 	}
-	err = errors.Cause(err)
 	// Is it an ordinary EPERM?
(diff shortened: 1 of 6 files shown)


kolyshkin:
@AkihiroSuda @crosbymichael @mrunalp PTAL

AkihiroSuda:
LGTM

mrunalp:
LGTM

crosbymichael:
LGTM

PR #2294, opened 2020-04-03
Remove unused consts testScopeWait and testSliceWait

tklauser:
These are unused since commit 518c855833c4 ("Remove libcontainer
detection for systemd features")


AkihiroSuda:
LGTM

Issue #2298, opened 2020-04-07
cgroupv2 fs driver does not work with default cgroups path


kolyshkin:
When the cgroupv2 fs driver is used without setting `cgroupsPath`, it picks up a path from `/proc/self/cgroup`. On a host with systemd, such path looks like `/user.slice/user-1000.slice/session-4.scope`.

It fails to enable controllers for sub-cgroup. The error from the kernel is either `EBUSY` or `EOPNOTSUPP`. It has probably something to do with systemd, but I'm not sure.

Here's the error (with more debug added):
```console
# runc run -d --console-socket /tmp/console.sock test_cgroups_permissions
CreateCgroupPath: res="+cpuset +cpu +io +memory +hugetlb +pids"
CreateCgroupPath: echo $res > "/sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope/cgroup.subtree_control"
# time="2020-04-07T12:57:57-07:00" level=warning msg="signal: killed"
# time="2020-04-07T12:57:57-07:00" level=error msg="container_linux.go:349: starting container process caused \"process_linux.go:306: applying cgroup configuration for process caused \\\"write /sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope/cgroup.subtree_control: device or resource busy\\\"\""
```

### Here's the same thing, reproduced manually from a shell:
```console
$ sudo -s

# cat /proc/self/cgroup 
0::/user.slice/user-1000.slice/session-4.scope
# cd /sys/fs/cgroup/user.slice/user-1000.slice/session-4.scope

# cat cgroup.controllers 
cpuset cpu io memory hugetlb pids
# CONTROLS="+cpuset +cpu +io +memory +hugetlb +pids"
# echo $CONTROLS > cgroup.subtree_control
-bash: echo: write error: Device or resource busy

# for C in $CONTROLS; do echo $C; echo $C > cgroup.subtree_control; done
+cpuset
+cpu
+io
-bash: echo: write error: Operation not supported
+memory
-bash: echo: write error: Operation not supported
+hugetlb
-bash: echo: write error: Operation not supported
+pids

# uname -a
Linux f31-test 5.6.2-300.fc32.x86_64 #1 SMP Thu Apr 2 18:34:20 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

kolyshkin:
@AkihiroSuda do you have any idea why is that?

kolyshkin:
[...]

PR #2295, opened 2020-04-04
Initial integration tests for cgroupv2

kolyshkin:
This refactors integration test helpers a bit, and adds some initial tests for cgroupv2.

Split into logical small(er) commits for the sake of easier reviewing, so better to review it commit by commit.

This is far from complete, but is it mergeable and already helped to find a surprising number of bugs:
 * https://github.com/opencontainers/runc/pull/2272 cgroupv2: use "max" for negative values (and a followup: https://github.com/opencontainers/runc/pull/2300)
 * https://github.com/opencontainers/runc/pull/2270 cgroupv2: don't try to set kmem for systemd case
 * https://github.com/opencontainers/runc/pull/2288 cgroupv2: fix setting MemorySwap
 * https://github.com/opencontainers/runc/pull/2299 cgroupv2: fix fs2 driver initialization
 * https://github.com/opencontainers/runc/issues/2298 cgroupv2 fs driver does not work with default cgroups path (should be fixed by https://github.com/opencontainers/runc/pull/2305)
 * ~https://github.com/opencontainers/runc/issues/2310 cgroupv2: runc run --systemd-cgroup do not put container in proper cgroup (should be fixed by https://github.com/opencontainers/runc/pull/2311)~
 * https://github.com/opencontainers/runc/issues/2313 runc --systemd-cgroup run: need to check that StartTransientUnit succeeds

PR #2299, opened 2020-04-07
cgroupv2: fix fs2 driver initialization

kolyshkin:
The main goal of this patchset is to fix fs2 driver initialization. With these patches, the tests that I added in https://github.com/opencontainers/runc/pull/2295 are now passing with fs2 driver (i.e. when `RUNC_USE_SYSTEMD` is not set), except for one case (https://github.com/opencontainers/runc/issues/2298).

I also did some major cleanups, mostly for systemd driver. A high-level overview:
 - remove legacy paths map from v2 driver
 - untangle systemd v2 initialization;
 - reorganized systemd code into v1, v2, common, and unsupported.

@AkihiroSuda PTAL (this was discussed before in https://github.com/opencontainers/runc/pull/2281#issuecomment-609099674)


libcontainer/cgroups/fs2/fs2.go
```diff
@@ -8,75 +8,67 @@ import (
...
-	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/pkg/errors"
 )
 
+type manager struct {
+	config *configs.Cgroup
+	// dirPath is like "/sys/fs/cgroup/user.slice/user-1001.slice/session-1.scope"
+	dirPath string
```
AkihiroSuda, line 19: This example is confusing and looks as if it was related to rootless systemd

kolyshkin:
> `write /sys/fs/cgroup/cgroup.subtree_control: device or resource busy`

* if this is in test with `--systemd-cgroup`, then I don't understand why is this happening as I haven't changed anything;
* if this is in test without `--systemd-cgroup` but systemd is present, then `--systemd-cgroup` should be used;
* if this is in test without `--systemd-cgroup` and systemd is not present, then I don't have an idea about what is going on :(

AkihiroSuda:
the errors are happening on Travis
https://travis-ci.org/github/opencontainers/runc/jobs/672264460

kolyshkin:
The problem is we're enabling all the controllers no matter what limits are set. Two ways to fix this are:

1. enable all controllers on a "best effort" basis (i.e. ignore the errors during init time, return errors during set time)
2. only enable needed controllers.

I like 2 much better. Let's see if this can be done (and it will fix the above tests since looks like they do not set any limits)

kolyshkin:
Added a commit to enable controllers only when required and on a best-effort basis.

cgroup-v2 is passing now. The tests I add/enable in #2295 are passing, too, on my test box.

kolyshkin:
@crosbymichael @mrunalp PTAL

kolyshkin:
@cyphar PTAL (at least for those two patches that add/remove securejoin usage)

kolyshkin:
Rebased to resolve conflict after #2288 is merged

kolyshkin:
Let me rework the last patch (enable controllers on demand) -- while it works fine, I don't like that `neededControllers()` has the intimate knowledge of all subsystems in one place.

kolyshkin:
@AkihiroSuda please let me know if you want me to split into two PRs, as this one might be too big to review. I tried hard to split it into small logical commits but there are too many.

kolyshkin:
fixed a bug in last commit

libcontainer/cgroups/fs2/cpu.go
```diff
@@ -13,7 +13,15 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
+func isCpuSet(cgroup *configs.Cgroup) bool {
+	return cgroup.Resources.CpuWeight != 0 || cgroup.Resources.CpuMax != ""
+}
+
 func setCpu(dirPath string, cgroup *configs.Cgroup) error {
+	if !isCpuSet(cgroup) {
```
AkihiroSuda, line 21: Not sure whether this validation is useful.

libcontainer/cgroups/fs2/cpuset.go
```diff
@@ -7,7 +7,15 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
+func isCpusetSet(cgroup *configs.Cgroup) bool {
+	return cgroup.Resources.CpusetCpus != "" || cgroup.Resources.CpusetMems != ""
+}
+
 func setCpuset(dirPath string, cgroup *configs.Cgroup) error {
+	if !isCpusetSet(cgroup) {
```
AkihiroSuda, line 15: same as above

libcontainer/cgroups/fs2/create.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+func neededControllers(cgroup *configs.Cgroup) []string {
```
AkihiroSuda, line 13: Please add godoc that the list doesn't contain "pseudo" controllers, i.e. devices and freezer

libcontainer/cgroups/fs2/create.go
```diff
@@ -0,0 +1,103 @@
...
+	var list []string
+
+	if cgroup == nil {
+		return list
+	}
+
+	// list of all available controllers
+	const file = UnifiedMountpoint + "/cgroup.controllers"
```
AkihiroSuda, line 21: TODO (#2281):  this should be `/sys/fs/cgroup/user.slice/user-1001.slice/cgroup.controllers` in rootless systemd

libcontainer/cgroups/fs2/cpu.go
kolyshkin: The reason I added it is peculiar -- if someone will be modifying `setCpu()`, adding more fields to set, they will be reminded to also modify `isCpuSet()`. Having this check here _might_ also help to uncover bugs.

libcontainer/cgroups/systemd/common.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return false
+	}
+	return fi.IsDir()
+}
+
+// systemd represents slice hierarchy using `-`, so we need to follow suit when
+// generating the path of slice. Essentially, test-a-b.slice becomes
+// /test.slice/test-a.slice/test-a-b.slice.
+func ExpandSlice(slice string) (string, error) {
```
AkihiroSuda, line 38: Not new in this PR, but can we have UT for this?

libcontainer/cgroups/fs2/cpu.go
kolyshkin: E.g. someone adds `cpuFoobar` setting to setCpu(), but forgets to modify `isCpuSet()`. They add a unit test to check if `cpuFoobar` is applied, and the test fails! Hope it makes sense.

libcontainer/factory_linux.go
```diff
@@ -50,28 +50,51 @@ func InitArgs(args ...string) func(*LinuxFactory) error {
...
+			panic(errors.Errorf("invalid dir path %q", path))
+		}
+	}
+
+	return path
+}
+
+func systemdCgroupV2(l *LinuxFactory, rootless bool) error {
+	l.NewCgroupsManager = func(config *configs.Cgroup, paths map[string]string) cgroups.Manager {
+		return systemd.NewUnifiedManager(config, getUnifiedPath(paths), rootless)
```
AkihiroSuda, line 74: Can be another PR, but eventually we should remove `getUnifiedPath()` madness

libcontainer/cgroups/fs2/cpu.go
AkihiroSuda: Can be just code comment lines?

libcontainer/cgroups/fs2/create.go
kolyshkin: I'm not sure. We're getting the list of controllers available on the host, it does not matter rootless or not. Even if some controllers are not available for the rootless cgroup, we won't error out, so having them in this list is OK.

libcontainer/factory_linux.go
kolyshkin: Yes, totally agree, I plan to do that, but this PR is way too big already, let's do it step by step and one day we'll get there :)

libcontainer/cgroups/fs2/create.go
kolyshkin: done

libcontainer/cgroups/fs2/cpu.go
kolyshkin: > Can be just code comment lines?

What do you mean? Add a comment like "if you're modifying this, don't forget to modify isSetCpu"? In my experience, this never works. The added runtime check might actually help to uncover such a bug.

Or do you mean adding a comment stating the reason why this check is added?

kolyshkin: > Or do you mean adding a comment stating the reason why this check is added?

I have amended the commit telling why is*Set is added to set*:


4. Amend all setFoo() functions to call isFooSet(). While this might
   seem unnecessary, it might actually help to uncover a bug.
   Imagine someone:
    - adds a cgroup.Resources.CpuFoo setting;
    - modifies setCpu() to apply the new setting;
    - but forgets to amend isCpuSet() accordingly <-- BUG

   In this case, a test case modifying CpuFoo will help
   to uncover the BUG. This is the reason why it's added.

AkihiroSuda: I mean the former, but not a blocker

AkihiroSuda:
Looks good, but I'd like to merge the integration tests (https://github.com/opencontainers/runc/pull/2295) before this.

libcontainer/cgroups/systemd/common.go
kolyshkin: Will do (maybe in a separate PR?)

AkihiroSuda: Yes, can be a separate PR

kolyshkin:
Rebasing since #2295 is merged. I still need #2305 otherwise the tests will definitely fail.

kolyshkin:
- rebased on top of #2295
 - (temporarily) cherry-picked #2305
 - enabled the localintegration test

kolyshkin:
checkpoint/restore tests fail; let me try without

kolyshkin:
OK this breaks checkpoint for some reason. Looking

kolyshkin:
Disabled checkpoint tests (runc currently does not support cgroup v2 c/r). All others should pass

kolyshkin:
Something ugly with CI -- logs for failed jobs are empty. Restarting

kolyshkin:
This one is ready for merge.

@crosbymichael @mrunalp @filbranden PTAL

AkihiroSuda:
LGTM

AkihiroSuda:
LGTM (still)

mrunalp:
LGTM

AkihiroSuda:
This commit caused regression https://github.com/opencontainers/runc/issues/2339

PR #2306, opened 2020-04-10
[RFC] travis.yml: simplify vagrant ssh calls

kolyshkin:
Use `vagrant ssh` wrapper.

Unfortunately we still have to use sudo on the host to use ssh, otherwise we get this error (on Travis)::

> Permission denied @ rb_sysopen - /home/travis/.vagrant.d/data/machine-index/index.lock (Errno::EACCES)


kolyshkin:
I'm not sure if I like it or not.

Initial version was without `sudo` on the host (as usually you don't need to be root to use ssh :slightly_frowning_face:) but it fails (something about travis I guess).

@AkihiroSuda what do you think?

AkihiroSuda:
IIUC `vagrant ssh` had shlex issues, I prefer to use plain `ssh`.

kolyshkin:
Closing this one

PR #2307, opened 2020-04-11
Dockerfile: use bats-core

kolyshkin:
The bats testing framework we use for integration test is not maintained
since 2015 and was superceded by bats-core [1]. Also, we are using an
unreleased version (and as far as I remember depend on its functionality,
although I don't remember what is it exactly).

Unfortunately, Debian still packages very old version of bats from the old repo.

Use recent bats-core as installed by npm, as recommended in [3].

[1] https://github.com/sstephenson/bats/pull/269
[3] https://github.com/bats-core/bats-core/wiki/Install-Bats-Using-a-Package

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
 ARG GO_VERSION=1.13
-ARG BATS_VERSION=03608115df2071fff4eaaff1605768c275e5f81f
 ARG CRIU_VERSION=v3.13
 
 FROM golang:${GO_VERSION}-buster
@@ -33,6 +32,7 @@ RUN dpkg --add-architecture armel \
         libseccomp-dev:armhf \
         libseccomp-dev:ppc64el \
         libseccomp2 \
+        npm \
         pkg-config \
         protobuf-c-compiler \
         protobuf-compiler \
@@ -49,13 +49,7 @@ RUN dpkg --add-architecture armel \
 RUN useradd -u1000 -m -d/home/rootless -s/bin/bash rootless
 
 # install bats
-ARG BATS_VERSION
-RUN cd /tmp \
-    && git clone https://github.com/sstephenson/bats.git \
-    && cd bats \
-    && git reset --hard "${BATS_VERSION}" \
-    && ./install.sh /usr/local \
-    && rm -rf /tmp/bats
+RUN npm install -g bats
 
 # install criu
 ARG CRIU_VERSION


Dockerfile
```diff
@@ -33,6 +32,7 @@ RUN dpkg --add-architecture armel \
         libseccomp-dev:armhf \
         libseccomp-dev:ppc64el \
         libseccomp2 \
+	npm \
```
h-vetinari, line 35: Looks like this doesn't use the same indentation (tabs vs. spaces) as the surrounding lines.

Dockerfile
```diff
@@ -49,13 +49,7 @@ RUN dpkg --add-architecture armel \
...
 
 # install bats
-ARG BATS_VERSION
-RUN cd /tmp \
-    && git clone https://github.com/sstephenson/bats.git \
-    && cd bats \
-    && git reset --hard "${BATS_VERSION}" \
-    && ./install.sh /usr/local \
-    && rm -rf /tmp/bats
+RUN npm install -g bats
```
thaJeztah, line 52: Was there a reason for using `npm` for installing, and not from GitHub? the old method (with updated repo) should still work; https://github.com/bats-core/bats-core#installing-bats-from-source

kolyshkin: Just because it's simpler and someone (hopefully) maintains the package.

thaJeztah: Not a big fan; not only would we then depend on the `npm` package maintainer to maintain the package (in addition to the new `bats-core` org), it also adds a whole bunch of new dependencies _just_ to install `bats`;

```
docker build -t foo -<<EOF
FROM golang:1.13.10-buster
RUN apt-get update && apt-get install -y --no-install-recommends npm && apt-get clean && rm -rf /var/cache/apt /var/lib/apt/lists/*
EOF

docker history foo
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
947f3c9961e5        5 seconds ago       RUN /bin/sh -c apt-get update && apt-get ins…   82.6MB              buildkit.dockerfile.v0
<missing>           5 days ago          /bin/sh -c #(nop) WORKDIR /go                   0B
...
```

That's 83MB to install `npm`, which we only need to install `bats`

crosbymichael: I recently setup bats for a project, installing from the github repo was very simple.  I would rather us do that instead of using `npm`

kolyshkin:
Will install from git

Issue #1627, opened 2017-10-29
`runc exec` can't specify no-tty unless changing config.json or creating process.json


qingyunha:
```golang
    // set the tty
    if context.IsSet("tty") {
        p.Terminal = context.Bool("tty")
    }
```

kolyshkin:
What the bug reporter means is, if a container's config.json specifies `"terminal": true"` (in `process` section), there is no ability to have terminal-less `runc exec`.

In addition to that, `runc exec` does not create a proper terminal using /dev/ptmx, but reuses fd 0. This, combined with `runc exec ... < /dev/null`, leads to 

> ERRO[0000] exec failed: provided file is not a console

kolyshkin:
A few possible solutions are:

1. Do not derive terminal boolean property from config.json (it is for the container start/run, not for exec). This is the way crun does it.

2. If a terminal is requested, create a pty pair using /dev/ptmx (or `"github.com/containerd/console".NewPty`). This is also the way crun does it.

I'm going to implement 1 for now.

PR #2308, opened 2020-04-11
runc exec: don't enable terminal unless -t is set

kolyshkin:
If container's config.json have `"terminal": true` setting in its
"process" section, runc exec assumes that stdin (fd 0) is a terminal
and tries to use it.

This leads to the following error in case stdin is not a terminal:

> ERRO[0000] exec failed: provided file is not a console

So, even if -t/--tty is not set, exec uses stdin as a terminal.
It does not help that urfave/cli v1 parser we use does not allow
to use `-t no` or `-t false`.

Since the settings in config.json is probably for the container run/start,
not for the auxiliary process started inside a container with exec, do
not use a setting from there, only treating stdin as a terminal in case
`-t` is explicitly given.

Fixes https://github.com/opencontainers/runc/issues/1627

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -203,6 +203,7 @@ func getProcess(context *cli.Context, bundle string) (*specs.Process, error) {
 	p.Env = append(p.Env, context.StringSlice("env")...)
 
 	// set the tty
+	p.Terminal = false
 	if context.IsSet("tty") {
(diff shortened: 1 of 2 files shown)


mrunalp:
LGTM

AkihiroSuda:
LGTM

Issue #2310, opened 2020-04-12
[SELinux] cgroupv2: runc run --systemd-cgroup do not put container in proper cgroup


kolyshkin:
runc ps not working when runc run --systemd-cgroup is used

```console
# ../contrib/cmd/recvtty/recvtty -m null sock.sock &
# ./runc --systemd-cgroup run -d --console-socket sock.sock tees
# ../runc --systemd-cgroup list
ID          PID         STATUS      BUNDLE         CREATED                          OWNER
tees        26230       running     /vagrant/tst   2020-04-12T19:07:22.900616771Z   root
# ../runc --systemd-cgroup ps tees
UID          PID    PPID  C STIME TTY          TIME CMD
# ../runc ps tees
UID          PID    PPID  C STIME TTY          TIME CMD
```

Indeed, the process is not in the cgroup it is supposed to be in!

```console
# cat /sys/fs/cgroup/system.slice/runc-tees.scope/cgroup.procs 
# cat /proc/26230/cgroup 
0::/system.slice/sshd.service
```

:open_mouth: 

**Update:** the cause was old(er) container-selinux package in Fedora 31 vagrant image. Once updated, it works. The problem, though, is that runc (or crun) do not see an error from systemd which can't create a transient unit.

AkihiroSuda:
Is this a recent regression or depends on a specific config?

kolyshkin:
> Is this a recent regression or depends on a specific config?

Looks like it has always been that way. If you take a look at systemd's v1 manager (aka LegacyManager) Apply method, it calls `joinCgroups()` at the very end, and it will write the PID to all cgroups.

Same method for systemd v2 (aka UnifiedManager), it only calls `createCgroupsv2Path` but never writes pid. The code was different before my commit a949e4f22fa9 (PR #2273), but it still was not doing what it should (and alas I overlooked it).

kolyshkin:
> Looks like it has always been that way.

Indeed it was, since the very beginning -- see commit 524cb7c31 (PR #2113).

kolyshkin:
and since we were not doing any testing, this bug lived to the day.

Fix is easy though: https://github.com/opencontainers/runc/pull/2311

kolyshkin:
OK, the cause of the issue is vagrant image of Fedora 31 used for testing. It came with not-so-up-to-date `container-selinux` (or maybe `selinux-policy`?), which resulted in:

[...]

PR #2311, opened 2020-04-13
[WIP] cgroupv2: fix systemd driver not putting pid into cgroup

kolyshkin:
Apparently, the driver never cared to add pid to the cgroup.

Fixes: #2310

### DIFF
--- a/libcontainer/cgroups/systemd/unified_hierarchy.go
+++ b/libcontainer/cgroups/systemd/unified_hierarchy.go
@@ -151,6 +151,9 @@ func (m *UnifiedManager) Apply(pid int) error {
 	if err := createCgroupsv2Path(path); err != nil {
 		return err
 	}
+	if err := cgroups.WriteCgroupProc(path, pid); err != nil {
+		return err
+	}
 	m.Paths = map[string]string{
 		"pids":    path,
 		"memory":  path,


kolyshkin:
Let me test this first...

kolyshkin:
Temporary
 - rebased on top of https://github.com/opencontainers/runc/pull/2295
 - added a commit to unblock ps.bats test 
in order to validate the fix witn Travis CI.

kolyshkin:
Tested locally, I confirm it fixes the issue.

kolyshkin:
Hmm, I don't see Travis CI in the list of checks here :-\

kolyshkin:
Travis succeeds (see results here: https://travis-ci.org/github/opencontainers/runc/jobs/674240854)
```
ok 47 ps
ok 48 ps -f json
ok 49 ps -e -x
```

Removing temp patches.

kolyshkin:
This one is ready to be merged

@mrunalp @AkihiroSuda @crosbymichael PTAL

AkihiroSuda:
LGTM

libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -151,6 +151,9 @@ func (m *UnifiedManager) Apply(pid int) error {
 	if err := createCgroupsv2Path(path); err != nil {
 		return err
 	}
+	if err := cgroups.WriteCgroupProc(path, pid); err != nil {
```
giuseppe, line 154: isn't it done automatically by systemd?

I see we are specifying the PID here: https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/unified_hierarchy.go#L61-L63

kolyshkin: I see that systemd v1 driver (aka `LegacyManager`) is doing it explicitly:
* Apply -> joinCgroups -> join for all controllers except cpuset
* ... -> cpuset.ApplyDir for cpuset

https://github.com/opencontainers/runc/blob/ccbb3364d49d2dc6d9f057134570b0f382f6ceb7/libcontainer/cgroups/systemd/apply_systemd.go#L312

giuseppe: I guess it is done because some controllers are not supported by systemd (like cpuset).  On cgroup v2 it is enough to join the cgroup once.

kolyshkin: @giuseppe you're right, adding PIDs property should be enough, and yet it is not working. Investigating.

giuseppe: I'd investigate if `dbusConnection.StartTransientUnit` waits correctly that the systemd unit is created.  Does it eventually join the correct cgroup if you add a sleep here?

kolyshkin: See https://github.com/opencontainers/runc/issues/2310#issuecomment-613128742

kolyshkin:
See here: 
https://github.com/opencontainers/runc/issues/2310#issuecomment-613128742 (TL;DR: it was selinux).

I guess we can close this one, but we still need to figure out what to do about getting the error back from systemd, if it fails.

PR #2312, opened 2020-04-13
Properly remove intermediate directory

tedyu:
In createCgroupsv2Path , deferred func removes intermediate directory on error.
However, the path for intermediate directory was global.

This PR properly removes the intermediate directory.

### DIFF
--- a/libcontainer/cgroups/systemd/unified_hierarchy.go
+++ b/libcontainer/cgroups/systemd/unified_hierarchy.go
@@ -241,6 +241,7 @@ func createCgroupsv2Path(path string) (Err error) {
 				}
 			} else {
 				// If the directory was created, be sure it is not left around on errors.
+				current := current
 				defer func() {
 					if Err != nil {
 						os.Remove(current)


libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -240,10 +240,11 @@ func createCgroupsv2Path(path string) (Err error) {
 					return err
 				}
 			} else {
+				toBeRolledBack := current
```
kolyshkin, line 243: can we please just say
```go
current := current
```
and do it right before `defer` (i.e. after the comment)?

kolyshkin:
@tedyu good catch! I was working on that code and missed this :(
How have you found it?

Here's a repro:
- https://play.golang.org/p/Jh03_GIKnen
(remove the `current := current` line to see the bug)

kolyshkin:
> Properly remove intermediate directory

@tedyu also please prefix your commit title with `cgroupv2/systemd:` or something else to designate the part of sources the fix is touching.

tedyu:
@kolyshkin 
Your comment has been addressed.

I was browsing the code and noticed that variable current was not properly scoped.

kolyshkin:
> I was browsing the code and noticed that variable current was not properly scoped.

Awesome, good eye! I wish we had a linter for such things.

AkihiroSuda:
LGTM, good catch

crosbymichael:
LGTM

PR #2317, opened 2020-04-16
Defer netns.Close() after error check

tedyu:
In linuxContainer#Restore, 
```
			netns, err := os.Open(nsPath)
```
The deferred close of netns should be after err check.
Otherwise we would encounter nil-dereference.


AkihiroSuda:
LGTM

cyphar:
LGTM.

PR #2318, opened 2020-04-16
cgroupv2: use default allowed devices when linux resources is null

lifubang:
A strange behavior when we use runc directly(with file system cgroup v2).

With `busybox` rootfs, and the `config.json` is like this:
```
{
	"ociVersion": "1.0.2",
	"process": {
		"terminal": false,
		...
		"args": [
			"sleep", "100000"
		],
		...
	},
	...
	"linux": {
		...
		"resources": {
		},
		...
}
```
For this `config.json`, there is nothing in `resources` section.
when we run `runc run -d test`. It works fine.

But if we remove the `resources` section, when we run `runc run -d test`:
```
root@bogon runctest]# runc run -d test
WARN[0000] exit status 1                                
ERRO[0000] container_linux.go:349: starting container process caused "process_linux.go:439: container init caused \"rootfs_linux.go:114: reopening /dev/null inside container caused \\\"Failed to open /dev/null - open /dev/null: operation not permitted\\\"\"" 
container_linux.go:349: starting container process caused "process_linux.go:439: container init caused \"rootfs_linux.go:114: reopening /dev/null inside container caused \\\"Failed to open /dev/null - open /dev/null: operation not permitted\\\"\""
```
And for this situation, `crun` works fine.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -422,159 +422,158 @@ func CreateCgroupConfig(opts *CreateOpts) (*configs.Cgroup, error) {
 	c.Resources.AllowedDevices = AllowedDevices
 	if spec.Linux != nil {
 		r := spec.Linux.Resources
-		if r == nil {
-			return c, nil
-		}
-		for i, d := range spec.Linux.Resources.Devices {
-			var (
-				t     = "a"
-				major = int64(-1)
-				minor = int64(-1)
-			)
-			if d.Type != "" {
-				t = d.Type
+		if r != nil {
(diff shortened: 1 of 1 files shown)


lifubang:
I just only change some code in here:
https://github.com/opencontainers/runc/blob/cc183ca6626c7fdcab8fa3f6d10ec89d9b185ee9/libcontainer/specconv/spec_linux.go#L425-L427
to:
```
if r != nil {
```

kolyshkin:
@AkihiroSuda @mrunalp @crosbymichael PTAL

AkihiroSuda:
LGTM

Issue #2321, opened 2020-04-17
leakage of cgroupv2


lifubang:
Nowadays, the upstream of container runtime, such as docker, containerd and podman, don't join cgroup namespace by default. If we use cgroup v2, we can view the host's whole cgroup fs structure in the container.

How to reproduce:
1. use podman:
```
[root@bogon runctest]# podman --runtime runc run --rm -it busybox ls /sys/fs/cgroup
cgroup.controllers      cgroup.stat             cpuset.cpus.effective   io.pressure             sys
cgroup.max.depth        cgroup.subtree_control  cpuset.mems.effective   machine.slice           system.slice
cgroup.max.descendants  cgroup.threads          docker                  memory.pressure         user.slice
cgroup.procs            cpu.pressure            init.scope              runctest
[root@bogon runctest]# podman --runtime runc run --rm -it busybox ls /sys/fs/cgroup/machine.slice
cgroup.controllers                                                             libpod-207bddd76f2fb04a53cc2778613cf11967141615fe42c39941ddcb9878c00e57.scope
cgroup.events                                                                  libpod-6353fec0668b6154ba3d5c82aca31f3728e03038e59dcc1f9d197225105811fd.scope
cgroup.freeze                                                                  libpod-80f3340e25a1ea18d7d4f3d2016f64a3763dba10c8a9c2b5e855a03dd4d1550b.scope
cgroup.max.depth                                                               memory.current
cgroup.max.descendants                                                         memory.events
cgroup.procs                                                                   memory.events.local
cgroup.stat                                                                    memory.high
cgroup.subtree_control                                                         memory.low
cgroup.threads                                                                 memory.max
cgroup.type                                                                    memory.min
cpu.max                                                                        memory.oom.group
cpu.pressure                                                                   memory.pressure
cpu.stat                                                                       memory.stat
cpu.weight                                                                     memory.swap.current
cpu.weight.nice                                                                memory.swap.events
io.bfq.weight                                                                  memory.swap.max
io.latency                                                                     pids.current
io.max                                                                         pids.events
io.pressure                                                                    pids.max
io.stat
```
2. use runc:
```
[root@bogon runctest]# runc run -d test
root@bogon runctest]# runc exec -t test ls /sys/fs/cgroup
cgroup.controllers      cpu.pressure            memory.pressure
cgroup.max.depth        cpuset.cpus.effective   runctest
cgroup.max.descendants  cpuset.mems.effective   sys
cgroup.procs            docker                  system.slice
[...]

PR #2322, opened 2020-04-17
cgroupv2: default join cgroup namespace in runc example

lifubang:
I think the runc example should join cgroup namespace by default if we use cgroup v2.

### DIFF
--- a/libcontainer/specconv/example.go
+++ b/libcontainer/specconv/example.go
@@ -4,13 +4,14 @@ import (
 	"os"
 	"strings"
 
+	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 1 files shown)


lifubang:
related to #2321

giuseppe:
could it be done only on cgroup v2?  Old versions of the kernel might not have the cgroup namespace

lifubang:
releated to #

> could it be done only on cgroup v2? Old versions of the kernel might not have the cgroup namespace

OK, I'll do it later.
How do you think about #2321? This issue also exist in crun. 
@giuseppe @kolyshkin

crosbymichael:
LGTM

AkihiroSuda:
LGTM

PR #2324, opened 2020-04-17
libcontainer: fix Checkpoint wrt cgroupv2

kolyshkin:
Commit 9a0184b10f474b (PR #2259) meant to enable using cgroup v2 freezer
for criu >= 3.14, but it looks like it is doing something else instead.

The logic here is:

 - for cgroup v1, set `FreezeCgroup`, if available
 - for cgroup v2, only set `FreezeCgroup` for criu >= 3.14
 - do not use `GetPaths()` in case v2 is used
   (this method is obsoleted for v2 and will be removed)

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1015,14 +1015,24 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 		}
 	}
 
-	if !cgroups.IsCgroup2UnifiedMode() && c.checkCriuVersion(31400) == nil {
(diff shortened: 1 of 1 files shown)


kolyshkin:
@adrianreber @avagin PTAL

adrianreber:
@kolyshkin Thanks. Looks much better. The code which was merged with my PR is not really useful (totally wrong). This now looks correct.

AkihiroSuda:
LGTM

mrunalp:
LGTM

PR #2325, opened 2020-04-18
Nits

kolyshkin:
A few nitpicks to simplify the code.

### 1. Do not use `syscall` package where `unix` can be used.

In many places (not all of them though) we can use `unix.`
instead of `syscall.` as these are indentical.

In particular, x/sys/unix defines:

```go
type Signal = syscall.Signal
type Errno = syscall.Errno
type SysProcAttr = syscall.SysProcAttr

const ENODEV      = syscall.Errno(0x13)
```

and unix.Exec() calls syscall.Exec().

### 2. Use `errors.Is()` to check for EINTR.

This is a forgotten hunk from PR #2291.

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"github.com/urfave/cli"
 	"golang.org/x/sys/unix"
@@ -23,7 +22,7 @@ Where "<container-id>" is the name for the instance of the container and
 EXAMPLE:
 For example, if the container id is "ubuntu01" the following will send a "KILL"
 signal to the init process of the "ubuntu01" container:
-	 
+
        # runc kill ubuntu01 KILL`,
 	Flags: []cli.Flag{
 		cli.BoolFlag{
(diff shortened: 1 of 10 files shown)


kill.go
```diff
@@ -23,7 +22,7 @@ Where "<container-id>" is the name for the instance of the container and
 EXAMPLE:
 For example, if the container id is "ubuntu01" the following will send a "KILL"
 signal to the init process of the "ubuntu01" container:
-	 
+
```
kolyshkin, line 25: Sorry about that, but that's the only place with EOL whitespace in all the .go files in this repo, and in my mind it does not justify a separate commit to fix it, so I shoved it in here.

libcontainer/cgroups/fscommon/fscommon.go
```diff
@@ -41,20 +41,10 @@ func ReadFile(dir, file string) (string, error) {
 func retryingWriteFile(filename string, data []byte, perm os.FileMode) error {
 	for {
 		err := ioutil.WriteFile(filename, data, perm)
-		if isInterruptedWriteFile(err) {
+		if errors.Is(err, unix.EINTR) {
```
kolyshkin, line 44: Tested using `go 1.14.2`:

```
=== RUN   TestWriteCgroupFileHandlesInterrupt
INFO[0000] interrupted while writing 1112067 to /sys/fs/cgroup/memory/test-eint-651408428/memory.limit_in_bytes 
--- PASS: TestWriteCgroupFileHandlesInterrupt (0.75s)
=== RUN   TestGetCgroupParamsInt
--- PASS: TestGetCgroupParamsInt (0.00s)
PASS
```

kill.go
cyphar: All good.

cyphar:
LGTM.

AkihiroSuda:
LGTM

PR #2327, opened 2020-04-19
checkpoint: don't print error if --pre-dump is set

kolyshkin:
Obviously, either `--pre-dump` or `--leave-running` implies
that we don't want to remove the container.

Found while working on checkpoint integration tests.

Fixes: 87712d288 (PR #2260)

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -63,12 +63,8 @@ checkpointed.`,
 			fatalf("Container cannot be checkpointed in %s state", status.String())
 		}
 		options := criuOptions(context)
-		if !options.LeaveRunning || !options.PreDump {
-			// destroy prints out an error if we tell CRIU to
-			// leave the container running:
-			// ERRO[0000] container is not destroyed
-			// The message is correct, but we actually do not want
-			// to destroy the container in this case.
+		if !(options.LeaveRunning || options.PreDump) {
+			// destroy container unless we tell CRIU to keep it
 			defer destroy(container)
 		}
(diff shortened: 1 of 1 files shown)


kolyshkin:
@adrianreber PTAL

checkpoint.go
```diff
@@ -63,7 +63,7 @@ checkpointed.`,
 			fatalf("Container cannot be checkpointed in %s state", status.String())
 		}
 		options := criuOptions(context)
-		if !options.LeaveRunning || !options.PreDump {
+		if !options.LeaveRunning && !options.PreDump {
```
kolyshkin, line 66: or will it look better like this?
```go
if !(options.LeaveRunning || options.PreDump) {
```

adrianreber:
It seems you are fixing all of my recent pull requests. Thanks.

I think I initially only had leave-running implemented and tested that. Just before pushing the pull request I added the pre-dump logic and added it wrongly.

I think I actually wanted something like your second suggestion:
```go
  if !(options.LeaveRunning || options.PreDump) {
```

Definitely :+1:

checkpoint.go
adrianreber: For me that is more readable. Not really important to change it, just if you like to push it once more.

mrunalp:
LGTM

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda:
LGTM

PR #2329, opened 2020-04-19
cgroupv2: fix Poststop hooks could not run when using systemd driver

lifubang:
How to reproduce:
```
[root@localhost systemdtest]# runc --systemd-cgroup run -d test
[root@localhost systemdtest]# runc --systemd-cgroup delete -f test
ERRO[0000] cannot detect unified path
```
The error is in here: https://github.com/opencontainers/runc/blob/bf0a8e17471347407fe9e856d4f3ff61beaf2fea/libcontainer/state_linux.go#L63-L76
If we got an error, the Poststop hooks could not run because the func has returned.

The reason is here: https://github.com/opencontainers/runc/blob/bf0a8e17471347407fe9e856d4f3ff61beaf2fea/libcontainer/cgroups/utils.go#L472-L495
Because we delete the cgroup paths value in cgroup manager when destroy.

### DIFF
--- a/libcontainer/cgroups/systemd/unified_hierarchy.go
+++ b/libcontainer/cgroups/systemd/unified_hierarchy.go
@@ -167,6 +167,10 @@ func (m *UnifiedManager) Destroy() error {
 	if m.Cgroups.Paths != nil {
 		return nil
 	}
+	dirPath, err := m.GetUnifiedPath()
+	if err != nil {
+		return err
+	}
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/systemd/unified_hierarchy.go
```diff
@@ -167,18 +167,39 @@ func (m *UnifiedManager) Destroy() error {
...
 	dbusConnection, err := getDbusConnection()
 	if err != nil {
 		return err
 	}
-	dbusConnection.StopUnit(getUnitName(m.Cgroups), "replace", nil)
-	if err := cgroups.RemovePaths(m.Paths); err != nil {
-		return err
+	unitName := getUnitName(m.Cgroups)
+	statusChan := make(chan string, 1)
+	if _, err := dbusConnection.StopUnit(unitName, "replace", statusChan); err == nil {
```
lifubang, line 183: @kolyshkin 
I think we may not need `StopUnit` method. Because if the init process has been exited, the cgroup path will be deleted by systemd automatically.
WDYT

kolyshkin: Well, I think it is still needed, otherwise it will show up in systemctl output as "killed" or something like that (I haven't checked yet, just a hunch).

We can also take a look at crun implementation -- in there the unit is removed from systemd first, then killed. Might make more sense.

kolyshkin:
@lifubang can you please rebase it? (`libcontainer/cgroups/systemd/unified_hierarchy.go` is renamed to `libcontainer/cgroups/systemd/v2.go`)

lifubang:
`ERRO[0000] cannot detect unified path`
This error has been fixed by #2299 .
I will add waiting for `StopUnit` in #2331 .

PR #2332, opened 2020-04-20
Fix/improve checkpoint integration tests

kolyshkin:
This fixes the checkpoint/restore tests which were failing or hanging.

A copy/paste of the last commit message:

Commit a9e15e7 adds a check that stdin/out/err pipes
are restored correctly. Commit ec26065 copy/pastes
the same code to one more another test.

Problem is (as pointed out in commit 5369f9a) these tests
sometimes hang. I have also seen them fail.

Apparently, the code used to create pipes and open them to fds
is racy:

```shell
cat $fifo | cat $fifo &
pid=$!
exec 50</proc/$pid/fd/0
exec 51>/proc/$pid/fd/0
```

Since `cat | cat` is spawned asynchronously, by the time exec is used,
the second cat process (i.e. $pid) is already fork'ed but it might
not be exec'ed yet. As a result, we get this (`ls -l /proc/self/fd`):

```
lr-x------. 1 root root 64 Apr 20 02:39 50 -> /dev/pts/1
l-wx------. 1 root root 64 Apr 20 02:39 51 -> /dev/pts/1
```
or, sometimes, this:
```
lr-x------. 1 root root 64 Apr 20 02:45 50 -> /dev/pts/1
l-wx------. 1 root root 64 Apr 20 02:45 51 -> 'pipe:[215791]'
```

instead of expected set of pipes:

```
> lr-x------. 1 root root 64 Apr 20 02:45 50 -> 'pipe:[215791]'
> l-wx------. 1 root root 64 Apr 20 02:45 51 -> 'pipe:[215791]'
```

One possible workaround is to add `sleep 0.1` or so after cat|cat,
but it is outright ugly (besides, we already have one sleep in
the test code).

The solution is to not use any external processes to create pipes.
I admit this still looks not very comprehensible, but at least it
is easier than before, and it works.

While at it, remove code duplication, moving the setup and check
code into a pair of functions.

Finally, since the tests are working now, remove the skip.

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -29,11 +29,8 @@ function teardown() {
   for i in `seq 2`; do
     # checkpoint the running container
     runc --criu "$CRIU" checkpoint --work-path ./work-dir test_busybox
-    ret=$?
-    # if you are having problems getting criu to work uncomment the following dump:
-    #cat /run/opencontainer/containers/test_busybox/criu.work/dump.log
     cat ./work-dir/dump.log | grep -B 5 Error || true
-    [ "$ret" -eq 0 ]
+    [ "$status" -eq 0 ]
 
     # after checkpoint busybox is no longer running
     runc state test_busybox
@@ -216,8 +213,7 @@ function teardown() {
   # Killing the CRIU on the checkpoint side will let the container
   # continue to run if the migration failed at some point.
   __runc --criu "$CRIU" restore -d --work-path ./image-dir --image-path ./image-dir --lazy-pages test_busybox_restore <&60 >&51 2>&51
-  ret=$?
-  [ $ret -eq 0 ]
+  [ $? -eq 0 ]
   run grep -B 5 Error ./work-dir/dump.log -q
   [ "$status" -eq 1 ]
 
@@ -269,11 +265,10 @@ function teardown() {
     # checkpoint the running container; this automatically tells CRIU to
     # handle the network namespace defined in config.json as an external
     runc --criu "$CRIU" checkpoint --work-path ./work-dir test_busybox
-    ret=$?
     # if you are having problems getting criu to work uncomment the following dump:
     #cat /run/opencontainer/containers/test_busybox/criu.work/dump.log
     cat ./work-dir/dump.log | grep -B 5 Error || true
-    [ "$ret" -eq 0 ]
+    [ "$status" -eq 0 ]
 
     # after checkpoint busybox is no longer running
     runc state test_busybox
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -33,8 +33,7 @@ function teardown() {
     [ "$status" -eq 0 ]
 
     # after checkpoint busybox is no longer running
-    runc state test_busybox
-    [ "$status" -ne 0 ]
+    testcontainer test_busybox checkpointed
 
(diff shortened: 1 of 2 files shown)


kolyshkin:
@avagin @adrianreber PTAL. Also a question -- do you guys think we need to check the pipes in two tests?

avagin:
LGTM, thanks!

> do you guys think we need to check the pipes in two tests?

This is our check to be sure that processes have been restored correctly.

kolyshkin:
Fails with cgroupv2 as it needs a fix from runc side as well as patched criu. Will add.

adrianreber:
@kolyshkin Is the reason to skip cgroup2 tests on vagrant-fedora 31 that this test uses CRIU from the Fedora repository (3.13) and that it does not build CRIU itself like done in the Dockerfile (3.13 + cgroup2 patches)?

adrianreber:
> Is the reason to skip cgroup2 tests on vagrant-fedora 31 that this test uses CRIU from the Fedora 

I meant 'Is the reason to skip checkpoint/restore tests..."

kolyshkin:
@adrianreber there are two reasons, both are described in commit message.

I have actually already fixed it, just want the fix to be in a separate PR (and need time to create it). Hope to do it today.

kolyshkin:
@adrianreber added a commit to remove the sleep from the lazy mig test, PTAL

tests/integration/checkpoint.bats
```diff
@@ -161,8 +164,6 @@ function check_pipes() {
   cat $lazy_pipe
```
adrianreber, line 164: The test probably fails because this `cat` will now hang forever.

adrianreber:
Would be good if the sleep can be removed. @avagin and I discussed this probably one and a half year ago, but we came to the conclusion that it is complicated and did not follow up. Can't remember all the details any more. If your pipe magic can solve this I am :+1: .

kolyshkin:
> Would be good if the sleep can be removed. @avagin and I discussed this probably one and a half year ago, but we came to the conclusion that it is complicated and did not follow up. Can't remember all the details any more. If your pipe magic can solve this I am ￼ .

@adrianreber I've reverted the last commit for now, will address it later (or you can tackle on it).

The thing is, what advertized as
```
--status-fd value   criu writes \0 to this FD once lazy-pages is ready
```
is in fact a file not an fd:

```go
StatusFd:                context.String("status-fd"),
...
go waitForCriuLazyServer(statusRead, criuOpts.StatusFd)
...
func waitForCriuLazyServer(r *os.File, status string) error {
...
        fd, err := os.OpenFile(status, os.O_TRUNC|os.O_WRONLY, os.ModeAppend)
```

so we can not pass a file descriptor like I did in a (now reverted) commit.

kolyshkin:
OK this one is ready

AkihiroSuda:
needs rebase

kolyshkin:
Rebased

adrianreber:
As soon as CRIU 3.14 is released I will update it in Fedora 31 so you do not need to build the package from source for the vagrant-fedora-31 test environment.

kolyshkin:
CI failures are caused by the recent github instability issues (failure to download criu patches). Restarted CI

AkihiroSuda:
LGTM

thaJeztah:
@cyphar @mrunalp PTAL

mrunalp:
LGTM

Issue #2328, opened 2020-04-19
cgroupv2 checkpoint/restore is not working


kolyshkin:
One reason it is not working is cgroupv2 mount (and BTW cgroupv1 mounts in case `--cgroupns` is used) should not be treated like (presented to criu as) external bind mount(s).

* cgroupv1 + cgroupns mounts are real mounts (having `fstype` of `cgroup`)
* cgroupv2 mount is a real mount (`fstype` is `cgroup2`)

There might be other bugs, too.

kolyshkin:
I already have a patch to fix this, just need more time to create a PR

adrianreber:
When I implemented the freezer CRIU cgroup2 changes for runc and CRIU it was still working. That was a few weeks ago.

The integration tests are actually running checkpoint/restore using vagrant-fedora-31 with cgroup2 in the container and outside of the container: https://github.com/opencontainers/runc/blob/master/libcontainer/integration/checkpoint_test.go#L86

I am certain I do not understand the whole cgroup details as good as you, just wanted to mention that some checkpoint/restore tests using cgroup2 are working.

kolyshkin:
@adrianreber I'll take a closer look at it, but I think I found (and fixed) a genuine bug, please see #2335. The reason it works in your case is that you explicitly set `Device` to `"cgroup2"`, while in case of runc run/checkpoint/restore it says `"cgroup"` since it is taken right from contig.json.

Maybe I need to redo my fix differently, changing "cgroup" to "cgroupv2" in case we're using v2 unified hierarchy, but let's discuss it in #2335

PR #2335, opened 2020-04-21
Fix cgroupv2 checkpoint/restore

kolyshkin:
Fixes: https://github.com/opencontainers/runc/issues/2328

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,9 +15,27 @@ Vagrant.configure("2") do |config|
   config.vm.provision "shell", inline: <<-SHELL
     cat << EOF | dnf -y shell
 update
-install podman make golang-go libseccomp-devel bats jq
+install podman make golang-go libseccomp-devel bats jq \
+ patch protobuf protobuf-c protobuf-c-devel protobuf-compiler \
+ protobuf-devel protobuf-python libnl3-devel libcap-devel libnet-devel \
+ nftables-devel libbsd-devel gnutls-devel
 ts run
 EOF
     dnf clean all
+
+    # TODO: remove this after criu 3.14 is released
+    rpm -e --nodeps criu || true
+    CRIU_VERSION=v3.13
+    mkdir -p /usr/src/criu \
+        && curl -fsSL https://github.com/checkpoint-restore/criu/archive/${CRIU_VERSION}.tar.gz | tar -C /usr/src/criu/ -xz --strip-components=1 \
+        && cd /usr/src/criu \
+        && echo 1 > .gitid \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/4c27b3db4f4325a311d8bfa9a50ea3efb4d6e377.patch | patch -p1 \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/aac41164b2cd7f0d2047f207b32844524682e43f.patch | patch -p1 \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/6f19249b2565f3f7c0a1f8f65b4ae180e8f7f34b.patch | patch -p1 \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/378337a496ca759848180bc5411e4446298c5e4e.patch | patch -p1 \
+        && make install-criu \
+        && cd - \
+        && rm -rf /usr/src/criu
   SHELL
 end
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1085,18 +1085,19 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 		go waitForCriuLazyServer(statusRead, criuOpts.StatusFd)
 	}
 
-	//no need to dump these information in pre-dump
+	// no need to dump all this in pre-dump
 	if !criuOpts.PreDump {
+		hasCgroupns := c.config.Namespaces.Contains(configs.NEWCGROUP)
 		for _, m := range c.config.Mounts {
 			switch m.Device {
 			case "bind":
 				c.addCriuDumpMount(req, m)
 			case "cgroup":
-				if cgroups.IsCgroup2UnifiedMode() {
-					c.addCriuDumpMount(req, m)
+				if cgroups.IsCgroup2UnifiedMode() || hasCgroupns {
+					// real mount(s)
 					continue
 				}
-				// cgroup v1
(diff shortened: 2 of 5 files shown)


kolyshkin:
@adrianreber PTAL

kolyshkin:
Rebased

adrianreber:
I looked at the last three commits and this looks correct.

Good idea to rely on the CRIU tmpfs handling for cgroup1 non-cgroupns mode.

I tried to start a container with Podman using `--cgroupns=host` on a cgroup1 system and the mounted cgroup looks the same using the `mount` command as when I leave the parameter `--cgroupns=host` away. I always see a tmpfs as the base for the cgroup1 mount. Should I see something different or am I testing it wrong with and without cgroupns?

kolyshkin:
If I understand it right (I still doubt that),
* cgroup v1 without cgroupns is a set of bind mounts from the host to container rootfs
* cgroup v1 with cgroupns is a set of real mounts with fstype=cgroup

and it's not obvious how to distinguish between the two cases from inside the container, i.e. mounts look the same (except for maybe the source field)

kolyshkin:
Also, for some reason, cgroupv1+cgroupns checkpoint/restore works even without these patches, and I admit I don't quite understand why.

Vagrantfile
```diff
@@ -15,9 +15,27 @@ Vagrant.configure("2") do |config|
...
+        && curl -fsSL https://github.com/checkpoint-restore/criu/archive/${CRIU_VERSION}.tar.gz | tar -C /usr/src/criu/ -xz --strip-components=1 \
+        && cd /usr/src/criu \
+        && echo 1 > .gitid \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/4c27b3db4f4325a311d8bfa9a50ea3efb4d6e377.patch | patch -p1 \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/aac41164b2cd7f0d2047f207b32844524682e43f.patch | patch -p1 \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/6f19249b2565f3f7c0a1f8f65b4ae180e8f7f34b.patch | patch -p1 \
+        && curl -sSL https://github.com/checkpoint-restore/criu/commit/378337a496ca759848180bc5411e4446298c5e4e.patch | patch -p1 \
+        && make install-criu \
+        && cd - \
+        && rm -rf /usr/src/criu
```
AkihiroSuda, line 39: Why not in Dockerfile?

kolyshkin: Yes, we have the very same code in Dockerfile (this is mostly a copy-paste).

We are running these tests on the Vagrant host itself (not inside a container, i.e. `make localintegration` not `make integration`) and some of those tests use criu, so we have to have criu on the vagrant host, not inside a container.

Or do you mean we can copy the criu binary out of container we just built and use it? This is also possible (and I tried to do it with runc binary earlier, and you suggested to rebuild it on the host itself, so I assumed you have the same PoV about criu binary).

As noted in the comment, this will go away as soon as criu 3.14 is released and packaged for F31.

AkihiroSuda: If we want to run everything on Vagrant directly, can we eliminate Podman from Vagrantfile? Can be another PR though.

kolyshkin: Yes, we can followup on that. This just fixes the cgroupv2 checkpoint and covers it with tests.

See https://github.com/opencontainers/runc/pull/2342. Note the difference is we used to run unit tests on Debian (container) and now it's Fedora (host).

AkihiroSuda:
LGTM

libcontainer/container_linux.go
```diff
@@ -1173,16 +1174,21 @@ func (c *linuxContainer) restoreNetwork(req *criurpc.CriuReq, criuOpts *CriuOpts
...
 		if err := prepareBindMount(m, c.config.Rootfs); err != nil {
 			return err
 		}
+	case "cgroup":
+		if !cgroups.IsCgroup2UnifiedMode() {
+			// mount points are saved by CRIU as /sys/fs/cgroup is tmpfs
+			return nil
+		}
+		// /sys/fs/cgroup is a real mount, the mountpoint need to be created
+		fallthrough
```
adrianreber, line 1189: So I tried all (a lot) combinations. In no case is it necessary to create the base directory `/sys/fs/cgroup` as that exists as soon as sysfs is mounted. For V1 we always have a tmpfs on `/sys/fs/cgroup` which means that we do not need to create any directories for CRIU. I think the fallthrough is not necessary. But is also does not hurt.

kolyshkin: Interesting, maybe I was wrong here. And you're right, even if we create /sys/fs/cgroup, once /sys is mounted, our dir is shadowed. Let's retry without fallthrough.

The commit (including the commit message) changed accordingly.

kolyshkin: Yep, I was obviously wrong, thanks, @adrianreber!

Vagrantfile
```diff
@@ -15,9 +15,27 @@ Vagrant.configure("2") do |config|
   config.vm.provision "shell", inline: <<-SHELL
     cat << EOF | dnf -y shell
 update
-install podman make golang-go libseccomp-devel bats jq
+install podman make golang-go libseccomp-devel bats jq \
+ patch protobuf protobuf-c protobuf-c-devel protobuf-compiler \
+ protobuf-devel protobuf-python libnl3-devel libcap-devel libnet-devel \
+ nftables-devel libbsd-devel gnutls-devel
```
h-vetinari, line 21: Should the `# TODO: remove this after criu 3.14 is released` be noted for the additionall installs here as well?

kolyshkin: Yes, but this is very minor and it can be easily seen from the commit itself.

Besides, it is complicated to insert a comment in the middle of a single shell command.

kolyshkin:
Got a very weird SIGSEGV in CI, filed https://github.com/checkpoint-restore/criu/issues/1035 just in case. Seems more like a bug in golang runtime or protobuf package.

Restarted CI to check if the above is repeatable, or was a one-time thing.

AkihiroSuda:
LGTM

kolyshkin:
Added a couple of patches on top, @adrianreber PTAL (only the last 2 patches)

adrianreber:
@kolyshkin Last two patches sound reasonable.

AkihiroSuda:
LGTM

mrunalp:
LGTM

PR #2338, opened 2020-04-21
fix path error in systemd when stopped

### LINKED ISSUE #2337: systemd: ps/kill can't work when container is in stopped state
How to reproduce:
```
[root@bogon systemdtest]# runc --systemd-cgroup run -d test
[root@bogon systemdtest]# runc --systemd-cgroup list
ID          PID         STATUS      BUNDLE             CREATED                          OWNER
test        177063      running     /opt/systemdtest   2020-04-21T10:40:55.998063538Z   root
[root@bogon systemdtest]# runc --systemd-cgroup ps test
UID          PID    PPID  C STIME TTY          TIME CMD
root      177063       1  0 18:40 ?        00:00:00 sleep 20
[root@bogon systemdtest]# runc --systemd-cgroup list
ID          PID         STATUS      BUNDLE             CREATED                          OWNER
test        0           stopped     /opt/systemdtest   2020-04-21T10:40:55.998063538Z   root
[root@bogon systemdtest]# runc --systemd-cgroup kill -a test 0
WARN[0000] freezer not supported: open /sys/fs/cgroup/system.slice/runc-test.scope/cgroup.freeze: no such file or directory
ERRO[0000] lstat /sys/fs/cgroup/system.slice/runc-test.scope: no such file or directory
lstat /sys/fs/cgroup/system.slice/runc-test.scope: no such file or directory
[root@bogon systemdtest]# runc --systemd-cgroup ps test
ERRO[0000] container_linux.go:169: getting all container pids from cgroups caused "lstat /sys/fs/cgroup/system.slice/runc-test.scope: no such file or directory"
container_linux.go:169: getting all container pids from cgroups caused "lstat /sys/fs/cgroup/system.slice/runc-test.scope: no such file or directory"
[root@bogon systemdtest]#
```

lifubang:
fix #2337 

Because when the container is in stopped state, the cgroup path has been deleted by systemd driver.

### DIFF
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -392,3 +392,7 @@ func (m *manager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return freezer.(*FreezerGroup).GetState(dir)
 }
+
+func (m *manager) Exists() bool {
+	return cgroups.PathExists(m.paths["devices"])
+}
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -481,3 +481,7 @@ func (m *legacyManager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return freezer.(*fs.FreezerGroup).GetState(path)
 }
+
+func (m *legacyManager) Exists() bool {
+	return cgroups.PathExists(m.paths["devices"])
+}
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -370,3 +370,7 @@ func (m *unifiedManager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return fsMgr.GetFreezerState()
 }
+
+func (m *unifiedManager) Exists() bool {
+	return cgroups.PathExists(m.path)
+}
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -169,7 +169,17 @@ func (c *linuxContainer) OCIState() (*specs.State, error) {
 }
 
 func (c *linuxContainer) Processes() ([]int, error) {
-	pids, err := c.cgroupManager.GetAllPids()
+	var pids []int
+	status, err := c.currentStatus()
+	if err != nil {
+		return pids, err
+	}
+	// for systemd cgroup, the unit's cgroup path will be auto removed if container's all processes exited
+	if status == Stopped && !c.cgroupManager.Exists() {
+		return pids, nil
+	}
+
+	pids, err = c.cgroupManager.GetAllPids()
 	if err != nil {
 		return nil, newSystemErrorWithCause(err, "getting all container pids from cgroups")
 	}
@@ -385,12 +395,16 @@ func (c *linuxContainer) start(process *Process) error {
 func (c *linuxContainer) Signal(s os.Signal, all bool) error {
 	c.m.Lock()
 	defer c.m.Unlock()
-	if all {
-		return signalAllProcesses(c.cgroupManager, s)
-	}
 	status, err := c.currentStatus()
 	if err != nil {
 		return err
+	}
+	if all {
+		// for systemd cgroup, the unit's cgroup path will be auto removed if container's all processes exited
+		if status == Stopped && !c.cgroupManager.Exists() {
+			return nil
+		}
+		return signalAllProcesses(c.cgroupManager, s)
 	}
 	// to avoid a PID reuse attack
 	if status == Running || status == Created || status == Paused {
--- a/libcontainer/container_linux_test.go
+++ b/libcontainer/container_linux_test.go
@@ -50,6 +50,15 @@ func (m *mockCgroupManager) Destroy() error {
 	return nil
 }
 
+func (m *mockCgroupManager) Exists() bool {
+	paths := m.GetPaths()
+	if paths != nil {
+		_, err := os.Lstat(paths["devices"])
+		return err == nil
+	}
+	return false
+}
+
 func (m *mockCgroupManager) GetPaths() map[string]string {
 	return m.paths
 }
@@ -134,11 +143,27 @@ func (m *mockProcess) forwardChildLogs() {
 }
 
 func TestGetContainerPids(t *testing.T) {
-	container := &linuxContainer{
-		id:            "myid",
-		config:        &configs.Config{},
-		cgroupManager: &mockCgroupManager{allPids: []int{1, 2, 3}},
+	pid := 1
+	stat, err := system.Stat(pid)
+	if err != nil {
+		t.Fatalf("can't stat pid %d, got %v", pid, err)
 	}
+	container := &linuxContainer{
+		id:     "myid",
+		config: &configs.Config{},
+		cgroupManager: &mockCgroupManager{
+			allPids: []int{1, 2, 3},
+			paths: map[string]string{
+				"device": "/proc/self/cgroups",
+			},
+		},
+		initProcess: &mockProcess{
+			_pid:    1,
+			started: 10,
+		},
+		initProcessStartTime: stat.StartTime,
+	}
+	container.state = &runningState{c: container}
 	pids, err := container.Processes()
 	if err != nil {
 		t.Fatal(err)
--- a/tests/integration/ps.bats
+++ b/tests/integration/ps.bats
@@ -60,3 +60,27 @@ function teardown() {
   [[ ${lines[0]} =~ \ +PID\ +TTY\ +STAT\ +TIME\ +COMMAND+ ]]
   [[ "${lines[1]}" =~ [0-9]+ ]]
 }
+
+@test "ps after the container stopped" {
+  # ps requires cgroups
+  [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
+  set_cgroups_path "$BUSYBOX_BUNDLE"
+
+  # start busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  # check state
+  testcontainer test_busybox running
+
+  runc ps test_busybox
+  [ "$status" -eq 0 ]
+
+  runc kill test_busybox KILL
+  [ "$status" -eq 0 ]
+
+  retry 10 1 eval "__runc state test_busybox | grep -q 'stopped'"
+
+  runc ps test_busybox
+  [ "$status" -eq 0 ]
+}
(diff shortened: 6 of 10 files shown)


lifubang:
This pr just fix cgroup v2.
For cgroup v1, it also has this problem, shall we need to fix it in cgroup v1?

libcontainer/cgroups/systemd/v2.go
```diff
@@ -198,6 +198,9 @@ func (m *unifiedManager) GetUnifiedPath() (string, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	if m.path != "" {
+		if !cgroups.PathExists(m.path) {
```
kolyshkin, line 201: I would not do this in this function, but rather modify the callers to check that path exists.

kolyshkin: as this is a hot path and doing stat might not always be required.

lifubang:
> I'd rather check if path exists in callers

Force pushed.
And the label is "area/systemd", not "area/cgroupv2". @kolyshkin

lifubang:
travis-ci green now!

libcontainer/container_linux.go
```diff
@@ -163,8 +163,39 @@ func (c *linuxContainer) OCIState() (*specs.State, error) {
...
 func (c *linuxContainer) Processes() ([]int, error) {
-	pids, err := c.cgroupManager.GetAllPids()
+	var pids []int
+	status, err := c.currentStatus()
+	if err != nil {
+		return pids, err
+	}
+	if status == Stopped {
+		// for systemd cgroup, the unit's cgroup path will be auto removed if container's all processes exited
+		if check := c.cgroupPathNotExist(); check {
```
kolyshkin, line 193:
```suggestion
		if c.cgroupPathNotExist() {
```

libcontainer/container_linux.go
```diff
@@ -380,13 +411,19 @@ func (c *linuxContainer) start(process *Process) error {
...
-		return signalAllProcesses(c.cgroupManager, s)
-	}
 	status, err := c.currentStatus()
 	if err != nil {
 		return err
 	}
+	if all {
+		// for systemd cgroup, the unit's cgroup path will be auto removed if container's all processes exited
+		if status == Stopped {
+			if check := c.cgroupPathNotExist(); check {
```
kolyshkin, line 421: ditto

libcontainer/container_linux.go
```diff
@@ -163,8 +163,39 @@ func (c *linuxContainer) OCIState() (*specs.State, error) {
 	return c.currentOCIState()
 }
 
+func (c *linuxContainer) cgroupPathNotExist() bool {
```
kolyshkin, line 166: Actually, what about adding `Exists()` method to cgroupManager instead?

libcontainer/container_linux.go
```diff
@@ -163,8 +163,39 @@ func (c *linuxContainer) OCIState() (*specs.State, error) {
...
+		return pids, err
+	}
+	if status == Stopped {
+		// for systemd cgroup, the unit's cgroup path will be auto removed if container's all processes exited
+		if check := c.cgroupPathNotExist(); check {
+			return pids, err
+		}
+	}
+
+	pids, err = c.cgroupManager.GetAllPids()
```
thaJeztah, line 198: wouldn't this already be able to have a check for `os.IsNotExist(err)` ? (Also wondering if `c.cgroupManager.GetAllPids()` should ignore `os.IsNotExist(err)` by default (if the path doesn't exist, just return an empty list of pids ?

lifubang: > wouldn't this already be able to have a check for `os.IsNotExist(err)` ?

Yes, but in fs cgroup driver, there is no error when the container is in `stopped` state. And I think we should not ignore `IsNotExist` error when it is not in `stopped` state or it has joined a new pid namespace without path.

> (Also wondering if `c.cgroupManager.GetAllPids()` should ignore `os.IsNotExist(err)` by default (if the path doesn't exist, just return an empty list of pids ?

I think we should not ignore `IsNotExist` error when it is not in `stopped` state or it has joined a new pid namespace without path.

kolyshkin:
@lifubang needs a rebase

libcontainer/container_linux_test.go
```diff
@@ -134,11 +143,27 @@ func (m *mockProcess) forwardChildLogs() {
 }
 
 func TestGetContainerPids(t *testing.T) {
+	pid := 1
+	stat, err := system.Stat(pid)
+	if err != nil {
+		t.Fatalf("can't state pid %d", pid)
```
kolyshkin, line 149: nit: `s/state/stat/`
nit: include the actual error as well

libcontainer/container_linux_test.go
```diff
@@ -134,11 +143,27 @@ func (m *mockProcess) forwardChildLogs() {
...
 	container := &linuxContainer{
-		id:            "myid",
-		config:        &configs.Config{},
-		cgroupManager: &mockCgroupManager{allPids: []int{1, 2, 3}},
+		id:     "myid",
+		config: &configs.Config{},
+		cgroupManager: &mockCgroupManager{
+			allPids: []int{1, 2, 3},
+			paths: map[string]string{
+				"device": "/proc/self/cgroups",
```
kolyshkin, line 157: this looks kind of weird.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -425,3 +425,8 @@ func (m *manager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return freezer.(*FreezerGroup).GetState(dir)
 }
+
+func (m *manager) Exists() bool {
+	paths := m.GetPaths()
+	return cgroups.PathExists(paths["devices"])
```
kolyshkin, line 431: I think you can just do
```go
return cgroups.PathExists(m.Path("devices"))
```

libcontainer/cgroups/systemd/v1.go
```diff
@@ -485,3 +485,11 @@ func (m *legacyManager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return freezer.(*fs.FreezerGroup).GetState(path)
 }
+
+func (m *legacyManager) Exists() bool {
+	path, err := getSubsystemPath(m.cgroups, "devices")
```
kolyshkin, line 490: `getSubsystemPath` is super expensive, it parses the whole `/proc/self/mountinfo`.

I think you should be using `m.Path("devices")` here.

AkihiroSuda:
@lifubang PTAL. Also, can we have a test?

libcontainer/container_linux_test.go
lifubang: done

libcontainer/cgroups/fs/apply_raw.go
lifubang: done

libcontainer/cgroups/systemd/v1.go
lifubang: done

libcontainer/container_linux_test.go
lifubang: It's a mock data, I don't have any idea to improve it.

lifubang:
> @lifubang PTAL at Kir's comments. Also, can we have an integration test?

finished now, PTAL.

tests/integration/ps.bats
```diff
@@ -60,3 +60,28 @@ function teardown() {
   [[ ${lines[0]} =~ \ +PID\ +TTY\ +STAT\ +TIME\ +COMMAND+ ]]
   [[ "${lines[1]}" =~ [0-9]+ ]]
 }
+
+@test "ps after the container stopped" {
+  # ps is not supported, it requires cgroups
+  requires root
```
AkihiroSuda, line 66: should suppprt rootless cgroup

AkihiroSuda:
CI failing

libcontainer/container_linux_test.go
```diff
@@ -50,6 +50,15 @@ func (m *mockCgroupManager) Destroy() error {
 	return nil
 }
 
+func (m *mockCgroupManager) Exists() bool {
+	paths := m.GetPaths()
+	if paths != nil {
+		_, err := os.Lstat(paths["devices"])
```
kolyshkin, line 56: While it is technically OK to use `m.GetPaths()` here, and it's a mock code so it doesn't really matter, I'd still like to have `m.Path("devices")` used here, because since commit 714c91e9f73a1512808476eb532b4aa36bbb7530 we're not supposed to use GetPaths() for anything other than state save/restore.

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -392,3 +392,7 @@ func (m *manager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return freezer.(*FreezerGroup).GetState(dir)
 }
+
+func (m *manager) Exists() bool {
+	return cgroups.PathExists(m.paths["devices"])
```
kolyshkin, line 397: It probably doesn't matter in practice, but we're not supposed to access a map without holding a lock. This is why I have suggested using `m.Path("devices")` earlier -- it takes a lock before accessing m.paths.

Alternatively, you can leave this code as is but add taking a lock (same as `Path()`).

libcontainer/cgroups/systemd/v1.go
```diff
@@ -481,3 +481,7 @@ func (m *legacyManager) GetFreezerState() (configs.FreezerState, error) {
 	}
 	return freezer.(*fs.FreezerGroup).GetState(path)
 }
+
+func (m *legacyManager) Exists() bool {
+	return cgroups.PathExists(m.paths["devices"])
```
kolyshkin, line 486: ditto

libcontainer/container_linux.go
```diff
@@ -169,7 +169,17 @@ func (c *linuxContainer) OCIState() (*specs.State, error) {
...
 
 func (c *linuxContainer) Processes() ([]int, error) {
-	pids, err := c.cgroupManager.GetAllPids()
+	var pids []int
+	status, err := c.currentStatus()
+	if err != nil {
+		return pids, err
+	}
+	// for systemd cgroup, the unit's cgroup path will be auto removed if container's all processes exited
+	if status == Stopped && !c.cgroupManager.Exists() {
```
kolyshkin, line 178: Does it make sense to check for "Created" here as well? Or should it return an error in such case?

kolyshkin: No it does not ("created" container has a `runc init` process running so everything works as expected.

Also, "paused" container also has a process running.

So, it seems that the check for "stopped" here is sufficient.

mrunalp:
@lifubang could you update the PR?

AkihiroSuda:
@kolyshkin @mrunalp 
Let's merge this as-is and cover the nits in a follow-up PR.

mrunalp:
I am fine with that.

AkihiroSuda:
@kolyshkin WDYT?

PR #2340, opened 2020-04-21
fs2: fix cgroup.subtree_control EPERM on rootless + add CI

### LINKED ISSUE #2339: cgroup2:  \"open /sys/fs/cgroup/cgroup.subtree_control: permission denied\"": unknown.
Rootless runc on cgroup v2 (but without actual cgroup control) seems to have got broken recently

```console
$ docker --context=rootless run -it --rm alpine
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:305: applying cgroup configuration for process caused \"open /sys/fs/cgroup/cgroup.subtree_control: permission denied\"": unknown.
ERRO[0000] error waiting for container: context canceled 
```

```console
$ docker --context=rootless info
Client:
 Debug Mode: false
 Plugins:
  buildx: Build with BuildKit (Docker Inc., v0.3.1-31-g891d355)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 2
 Server Version: dev
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
 Logging Driver: json-file
 Cgroup Driver: none
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1d083fec4939ce58f23a48ad3a914adb826d592a
 runc version: 46be7b612e2533c494e6a251111de46d8e286ed5
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
  rootless
  cgroupns
 Kernel Version: 5.3.0-46-generic
 Operating System: Ubuntu 19.10
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 7.748GiB
 Name: suda-ws01
 ID: CWVR:KJQU:3CNT:IJF7:FMME:22Y7:GKFW:AFKJ:IVLQ:JOVW:3KZY:S25M
 Docker Root Dir: /home/suda/.local/share/docker
 Debug Mode: false
 Username: akihirosuda
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: No kernel memory limit support
WARNING: No kernel memory TCP limit support
```

```console
$ docker --context=rootless version
Client:
 Version:           20.03.0-dev
 API version:       1.41
 Go version:        go1.13.10
 Git commit:        abafad3df
 Built:             Tue Apr 21 15:06:56 2020
 Experimental:      true

Server:
 Engine:
  Version:          dev
  Go version:       go1.13.10
  Git commit:       ba8129b28a
  Built:            Tue Apr 21 15:06:37 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.3.0-511-g1d083fec
  GitCommit:        1d083fec4939ce58f23a48ad3a914adb826d592a
 runc:
  Version:          1.0.0-rc10+dev
  GitCommit:        46be7b612e2533c494e6a251111de46d8e286ed5
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

AkihiroSuda:
Fix #2339 (regression in 813cb3eb94659971c69c0dd19b2d5a9737891f15)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ matrix:
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration RUNC_USE_SYSTEMD=yes'
         # same setup but with fs2 driver instead of systemd
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration'
+        # rootless
+        - sudo ssh default -t 'cd /vagrant && sudo make localrootlessintegration'
   allow_failures:
     - go: tip
 
--- a/libcontainer/cgroups/fs2/create.go
+++ b/libcontainer/cgroups/fs2/create.go
@@ -95,15 +95,22 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 					}
 				}()
 			}
+			// Write cgroup.type explicitly.
+			// Otherwise ENOTSUP may happen.
+			cgType := filepath.Join(current, "cgroup.type")
+			_ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
 		}
 		// enable needed controllers
 		if i < len(elements)-1 {
 			file := filepath.Join(current, "cgroup.subtree_control")
-			if err := ioutil.WriteFile(file, []byte(allCtrs), 0755); err != nil {
-				// XXX: we can enable _some_ controllers doing it one-by one
-				// instead of erroring out -- does it makes sense to do so?
-				return err
+			if err := ioutil.WriteFile(file, []byte(allCtrs), 0644); err != nil {
+				// try write one by one
+				for _, ctr := range ctrs {
+					_ = ioutil.WriteFile(file, []byte(ctr), 0644)
+				}
 			}
+			// Some controllers might not be enabled when rootless or containerized,
+			// but we don't catch the error here. (Caught in setXXX() functions.)
 		}
 	}
 
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -53,7 +53,10 @@ func (m *manager) getControllers() error {
 
 	file := filepath.Join(m.dirPath, "cgroup.controllers")
 	data, err := ioutil.ReadFile(file)
-	if err != nil && !m.rootless {
+	if err != nil {
+		if m.rootless && m.config.Path == "" {
+			return nil
+		}
 		return err
 	}
 	fields := strings.Fields(string(data))
@@ -67,9 +70,22 @@ func (m *manager) getControllers() error {
 
 func (m *manager) Apply(pid int) error {
 	if err := CreateCgroupPath(m.dirPath, m.config); err != nil {
+		// Related tests:
+		// - "runc create (no limits + no cgrouppath + no permission) succeeds"
+		// - "runc create (rootless + no limits + cgrouppath + no permission) fails with permission error"
+		// - "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error"
+		if m.rootless {
+			if m.config.Path == "" {
+				cl, clErr := neededControllers(m.config)
+				if clErr == nil && len(cl) == 0 {
+					return nil
+				}
+				return errors.Wrap(err, "rootless needs no limits + no cgrouppath when no permission is granted for cgroups")
+			}
+		}
 		return err
 	}
-	if err := cgroups.WriteCgroupProc(m.dirPath, pid); err != nil && !m.rootless {
+	if err := cgroups.WriteCgroupProc(m.dirPath, pid); err != nil {
 		return err
 	}
 	return nil
@@ -199,7 +215,11 @@ func (m *manager) Set(container *configs.Config) error {
 		}
 	}
 	// devices (since kernel 4.15, pseudo-controller)
-	if err := setDevices(m.dirPath, container.Cgroups); err != nil {
+	//
+	// When m.Rootless is true, errors from the device subsystem are ignored because it is really not expected to work.
+	// However, errors from other subsystems are not ignored.
+	// see @test "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error"
+	if err := setDevices(m.dirPath, container.Cgroups); err != nil && !m.rootless {
 		errs = append(errs, err)
 	}
 	// cpuset (since kernel 5.0)
@@ -218,7 +238,7 @@ func (m *manager) Set(container *configs.Config) error {
 	if err := setFreezer(m.dirPath, container.Cgroups.Freezer); err != nil {
 		errs = append(errs, err)
 	}
-	if len(errs) > 0 && !m.rootless {
+	if len(errs) > 0 {
 		return errors.Errorf("error while setting cgroup v2: %+v", errs)
 	}
 	m.config = container.Cgroups
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -95,7 +95,7 @@ EOF
 
     runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_permissions
     [ "$status" -eq 1 ]
-    [[ ${lines[1]} == *"cannot set pids limit: container could not join or create cgroup"* ]]
+    [[ ${lines[1]} == *"rootless needs no limits + no cgrouppath when no permission is granted for cgroups"* ]] || [[ ${lines[1]} == *"cannot set pids limit: container could not join or create cgroup"* ]]
 }
 
 @test "runc create (limits + cgrouppath + permission on the cgroup dir) succeeds" {
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -68,6 +68,10 @@ EOF
 
     # check that initial values were properly set
     check_cgroup_value "cpuset.cpus" 0
+    if [[ "$CGROUP_UNIFIED" = "yes" ]] && ! grep -qw memory "$CGROUP_PATH/cgroup.controllers"; then
+    # This happen on containerized environment because "echo +memory > /sys/fs/cgroup/cgroup.subtree_control" fails with EINVAL
+        skip "memory controller not available"
+    fi
     check_cgroup_value $MEM_LIMIT 33554432
     check_cgroup_value $MEM_RESERVE 25165824
     check_cgroup_value "pids.max" 20
--- a/tests/rootless.sh
+++ b/tests/rootless.sh
@@ -82,6 +82,27 @@ function enable_cgroup() {
 		# handling.
 		[[ "$cg" == "cpuset" ]] && chown rootless:rootless "$CGROUP_MOUNT/$cg$CGROUP_PATH/cpuset."{cpus,mems}
 	done
+	# cgroup v2
+	if [[ -e "$CGROUP_MOUNT/cgroup.controllers" ]]; then
+		# Enable controllers. Some controller (e.g. memory) may fail on containerized environment.
+		set -x
+		for f in $(cat "$CGROUP_MOUNT/cgroup.controllers"); do echo +$f > "$CGROUP_MOUNT/cgroup.subtree_control"; done
+		set +x
+		# Create the cgroup.
+		mkdir -p "$CGROUP_MOUNT/$CGROUP_PATH"
+		# chown/chmod dir + cgroup.subtree_control + cgroup.procs + parent's cgroup.procs.
+		# See https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#delegation-containment
+		chown root:rootless "$CGROUP_MOUNT/$CGROUP_PATH" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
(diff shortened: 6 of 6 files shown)


kolyshkin:
Can we add a test case for this (apparently we don't, since #2299 CI was green)?

AkihiroSuda:
I will open separate PR for CI

AkihiroSuda:
added CI

.travis.yml
```diff
@@ -38,6 +38,8 @@ matrix:
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration RUNC_USE_SYSTEMD=yes'
         # same setup but with fs2 driver instead of systemd
         - sudo ssh default -t 'cd /vagrant && sudo make localintegration'
+        # rootless
+        - sudo ssh default -t 'cd /vagrant && sudo make localrootlessintegration'
```
kolyshkin, line 42: Do we want to run the same tests with `RUNC_USE_SYSTEMD=yes`?

AkihiroSuda: No until https://github.com/opencontainers/runc/pull/2281 gets merged.
And https://github.com/opencontainers/runc/pull/2281 will require this to be merged.

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -53,7 +53,10 @@ func (m *manager) getControllers() error {
 
 	file := filepath.Join(m.dirPath, "cgroup.controllers")
 	data, err := ioutil.ReadFile(file)
-	if err != nil && !m.rootless {
+	if err != nil {
+		if m.rootless && m.config.Path == "" {
+			return nil
```
kolyshkin, line 58: Does it make sense to keep trying to read the file every time the function is called, or should we maybe use `sync.Once()` here?

.travis.yml
kolyshkin: Right, not in this PR

tests/integration/cgroups.bats
```diff
@@ -95,7 +95,7 @@ EOF
 
     runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_permissions
     [ "$status" -eq 1 ]
-    [[ ${lines[1]} == *"cannot set pids limit: container could not join or create cgroup"* ]]
+    [[ ${lines[1]} == *"permission denied"* ]] || [[ ${lines[1]} == *"cannot set pids limit: container could not join or create cgroup"* ]]
```
kolyshkin, line 98: Can you tell what exact "permission denied" error are we getting here? If it's not descriptive enough, maybe makes sense to wrap it (merely "permission denied" looks way too generic).


We can wrap the generic "permission denied" error to avoid

libcontainer/cgroups/fs2/fs2.go
AkihiroSuda: I guess `sync.Once()` can be used, but this issue is not new in this PR

tests/integration/cgroups.bats
AkihiroSuda: updated

tests/integration/cgroups.bats
```diff
@@ -109,8 +109,10 @@ EOF
 }
 
 @test "runc exec (limits + cgrouppath + permission on the cgroup dir) succeeds" {
-   [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
-
+   [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup cgroups_v1
+# FIXME: how to chown/chmod for cgroup2 fs?
+# \"failed to write 86404 to cgroup.procs: write /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/cgroup.procs: permissi
```
AkihiroSuda, line 114:
```console
suda@suda-ws01:/sys/fs/cgroup$ sudo mkdir foo
suda@suda-ws01:/sys/fs/cgroup$ sudo chown suda.suda foo
suda@suda-ws01:/sys/fs/cgroup$ mkdir foo/bar
suda@suda-ws01:/sys/fs/cgroup$ echo $$ > foo/bar/cgroup.procs 
-bash: echo: write error: Permission denied
```

cc @giuseppe

Vagrantfile
```diff
@@ -19,5 +19,7 @@ install podman make golang-go libseccomp-devel bats jq
 ts run
 EOF
     dnf clean all
+    # Add a user for rootless tests
+    useradd -u2000 -m -d/home/rootless -s/bin/bash rootless
```
kolyshkin, line 23: BTW I am doing this (and more) in https://github.com/opencontainers/runc/pull/2342

AkihiroSuda: yes, copied from your PR :)

libcontainer/cgroups/fs2/create.go
```diff
@@ -100,8 +100,14 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
...
 			file := filepath.Join(current, "cgroup.subtree_control")
 			if err := ioutil.WriteFile(file, []byte(allCtrs), 0755); err != nil {
-				// XXX: we can enable _some_ controllers doing it one-by one
-				// instead of erroring out -- does it makes sense to do so?
+				if rootless {
+					// try write one by one
+					for _, ctr := range ctrs {
+						_ = ioutil.WriteFile(file, []byte(ctr), 0755)
+					}
+					// ignore unwritable files
```
kolyshkin, line 108: 1. We are writing to the same file here, so this comment is misleading.

2. Enabling controllers one by one makes sense if some controllers can't be enabled by the kernel (for example, pids and memory controller can't be enabled if we already have processes in this cgroup, while others can still be enabled).

I don't think that in case of rootless you can enable some controllers but not some others. Unless I'm wrong, it does not make sense to try to enable them individually here.

libcontainer/cgroups/fs2/fs2.go
kolyshkin: correct

tests/integration/cgroups.bats
kolyshkin: Looks like a kernel issue (or misfeature) to me. Files created in a new directory should be owned by the same UID/GID as the directory itself. Maybe there's something in v2 docs about it?

AkihiroSuda:
```console
$ ls -l foo/bar/cgroup.procs 
-rw-r--r-- 1 suda suda 0 Apr 22 07:59 foo/bar/cgroup.procs
```

Owned by me, but EPERM happens

AkihiroSuda: https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#delegation-containment

To write `/sys/fs/cgroup/foo/bar/cgroup.procs`, it seems I need to chown/chmod `/sys/fs/cgroup/cgroup.procs` and `/sys/fs/cgroup/cgroup.procs` as well. (The last one is nasty 😕 )

libcontainer/cgroups/fs2/create.go
AkihiroSuda: > Enabling controllers one by one makes sense if some controllers can't be enabled by the kernel 

And if constrained by the parent subtree_control

tests/integration/cgroups.bats
kolyshkin: I think as a user you're in a non-root cgroup already (`grep ^0: /proc/self/cgroup`) and this is the "source" cgroup to which you need access to, not the root one.

libcontainer/cgroups/fs2/create.go
```diff
@@ -95,13 +96,25 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 					}
 				}()
 			}
+			if rootless {
+				// Write cgroup.type explicitly.
+				// Otherwise ENOTSUP may happen.
```
kolyshkin, line 101: "may happen" where? When enabling controllers? Creating children cgroups?

Why the issue this code is trying to solve is rootless-only?

tests/integration/update.bats
```diff
@@ -68,6 +68,10 @@ EOF
 
     # check that initial values were properly set
     check_cgroup_value "cpuset.cpus" 0
+    if [[ "$CGROUP_UNIFIED" = "yes" ]] && ! grep memory "$CGROUP_PATH/cgroup.controllers"; then
```
kolyshkin, line 71: nit: `grep -qw`

tests/integration/update.bats
```diff
@@ -68,6 +68,10 @@ EOF
...
+    # This happen on containerized environment because "echo + memory > /sys/fs/cgroup/cgroup.subtree_control" fails with EINVAL
```
kolyshkin, line 72: nit: `s/echo + memory/echo +memory/`

tests/rootless.sh
```diff
@@ -82,6 +82,24 @@ function enable_cgroup() {
...
+		mkdir -p "$CGROUP_MOUNT/$CGROUP_PATH"
+		# chown/chmod dir + cgroup.subtree_control + cgroup.procs + parent's cgroup.procs.
+		# See https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#delegation-containment
+		chown root:rootless "$CGROUP_MOUNT/$CGROUP_PATH" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
+		chmod g+rw "$CGROUP_MOUNT/$CGROUP_PATH" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
+		# Fix up cgroup.type.
+		echo threaded > "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"
+		# Make sure cgroup.type doesn't contain "invalid". Otherwise write ops will fail with ENOTSUP.
+		# See http://man7.org/linux/man-pages/man7/cgroups.7.html
+		if grep invalid "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"; then
```
kolyshkin, line 99: nit: `grep -qw`

libcontainer/cgroups/fs2/create.go
AkihiroSuda: Updated. This seems needed when running the test in  acontainer (`make integration`), regardless to rootless or rootful.

AkihiroSuda:
Weird failure, fails only with idmap mode.

```
[03] run rootless tests ... (idmap)
path: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
1..78
ok 1 runc update --kernel-memory{,-tcp} (initialized) # skip test requires rootless_cgroup
ok 2 runc update --kernel-memory (uninitialized) # skip test requires rootless_cgroup
ok 3 runc create (no limits + no cgrouppath + no permission) succeeds
not ok 4 runc create (rootless + no limits + cgrouppath + no permission) fails with permission error
# (in test file tests/integration/cgroups.bats, line 86)
#   `[ "$status" -eq 1 ]' failed
...
```

tests/integration/update.bats
AkihiroSuda: done

tests/rootless.sh
AkihiroSuda: done

tests/integration/update.bats
AkihiroSuda: done

AkihiroSuda:
CI green. Ready for review.

kolyshkin:
Needs a rebase

AkihiroSuda:
rebased

AkihiroSuda:
@opencontainers/runc-maintainers PTAL? 

I'd like to get almost all cgroup v2 issues to be resolved before Docker 20.0X freeze (May 1) and Podman 2.0

.travis.yml
```diff
@@ -36,6 +36,8 @@ matrix:
...
```
kolyshkin, line 40: Just noticed it now -- we probably want to test it with systemd as well.

I.e. something like

```
# cgroupv2+systemd: test on vagrant host itself as we need systemd
- sudo ssh default -t 'cd /vagrant && sudo make localintegration localrootlessintegration RUNC_USE_SYSTEMD=yes'
# same setup but with fs2 driver instead of systemd
- sudo ssh default -t 'cd /vagrant && sudo make localintegration localrootlessintegration'
```

Or maybe it makes sense to introduce something like `localtest` as the above is starting to look horrific.

AkihiroSuda:
@kolyshkin 
No, the systemd PR is not merged yet: https://github.com/opencontainers/runc/pull/2281

Let's merge this PR now so that we can rebase #2281 on top of this and add test for that

.travis.yml
AkihiroSuda: https://github.com/opencontainers/runc/pull/2340#discussion_r412522250

kolyshkin:
LGTM

kolyshkin:
@mrunalp PTAL

mrunalp:
LGTM

PR #2341, opened 2020-04-21
runc checkpoint: fix --status-fd to accept fd

kolyshkin:
### 1
The command `runc checkpoint --lazy-server --status-fd $FD` actually
accepts a file name as an $FD. Make it accept a file descriptor,
like its name implies and the documentation states.

In addition, since runc itself does not use the result of CRIU status
fd, remove the code which relays it, and pass the FD directly to CRIU.

Note 1: runc should close this file descriptor itself after passing it
to criu, otherwise whoever waits on it might wait forever.

Note 2: due to the way criu swrk consumes the fd (it reopens
/proc/$SENDER_PID/fd/$FD), runc can't close it as soon as criu swrk has
started. There is no good way to know when criu swrk has reopened the
fd, so we assume that as soon as we have received something back, the
fd is already reopened.

### 2

Since the meaning of `--status-fd` has changed, the test case using
it needs to be fixed as well.

Modify the lazy migration test to remove "sleep 2", actually waiting
for the the lazy page server to be ready.

### While at it

- remove the double fork (using shell's background process is
sufficient here);

- check the exit code for "runc checkpoint" and "criu lazy-pages";

- remove the check for no errors in dump.log after restore, as we
are already checking its exit code.

### History

[v2: properly close status fd after spawning criu]
[v3: move close status fd to after the first read]


kolyshkin:
@adrianreber PTAL and let's discuss this here. Apparently the current `--status-fd` interface is not really good or easy to use. Since runc is still pre-1.0 (and I guess the only user of `--status-fd` is this test case), maybe we can make it better. Here's my raw brain dump, to spark the discussion:

* we can make `--status-fd` actually accepting fd rather than a filename
* we can write actual ASCII exit code here (easier to handle from shell than binary \0)
* we can rename it to `--status-file` if we want to keep using file
...

h-vetinari:
> @kolyshkin:  Since runc is still pre-1.0 (and I guess the only user of `--status-fd` is this test case)...

runc is pre-1.0 in version-numbering only. The whole situation with the hooks only took so long to resolve (>2 years!) because correcting an old mistake would have led to actual deployments being broken (not just downstream libraries). I can't speak to whether users of `--status-fd` actually exist, but if they do, then changing it is more difficult (depending on how easily the changes can be cushioned downstream).

adrianreber:
@kolyshkin That is difficult. I am not aware of (m)any lazy migration CRIU users and even more if anyone is using this via runc, but changing an existing interface is indeed difficult.

> * we can write actual ASCII exit code here (easier to handle from shell than binary \0)

At this place we are just passing through what we get from CRIU. Not sure how important the shell use-case is at all, but I would not change it and just forward what we get from CRIU.

I think the problem with this test was that we need to background runc to continue the shell script. The page-server mode means that runc/CRIU will continue to run until we have fetched the last memory page out of it. Now we need to know it the actual checkpoint was successful, we wait two seconds and check if a checkpoint exists (-e image-dir/inventory.img). Next step is to see if the page-server is ready by reading from the so-called status-fd file. Can this be solved if `--status-fd` is actually an FD instead of a pipe? I think the problem was that if the checkpoint failed for some reason, reading from the pipe would hang forever. The normal way to know if checkpointing failed is the return code of runc, but we do not have this because of the page-server.

Does this explanation help to understand the situation? Would it be easier if `--status-fd` is a real FD?

kolyshkin:
@adrianreber I agree that making runc to transform \0 from criu into something else (e.g. ASCII "0") is not a good idea. Let it stay the way it is.

I would still change `--status-fd` to work with fd rather than a file. It will simplify things and remove "runc" as a middleman. 

> The normal way to know if checkpointing failed is the return code of runc, but we do not have this because of the page-server.

It looks like we don't need a double fork (it is needed for `recvtty` but not for page server), and it's easy to get the return code.

Let me modify this a bit.

kolyshkin:
@adrianreber @avagin PTAL

kolyshkin:
Rebased; rm 'draft' status (forgot to do it earlier)

@adrianreber @avagin PTAL

libcontainer/container_linux.go
```diff
@@ -1072,27 +1052,34 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
...
+			flags, err := unix.FcntlInt(uintptr(fd), unix.F_GETFL, 0)
+			if err != nil {
+				return fmt.Errorf("invalid --status-fd argument %d: %w", fd, err)
+			}
+			// and writable
+			if flags&unix.O_WRONLY == 0 {
+				return fmt.Errorf("invalid --status-fd argument %d: not writable", fd)
+			}
+
+			rpcOpts.StatusFd = proto.Int32(int32(fd))
```
avagin, line 1076: do you close it after sending this request?

kolyshkin: will adding `defer unix.Close(fd)` be enough here?

patch updated.

AkihiroSuda:
needs rebase

kolyshkin:
Rebased. @avagin your comment is addressed, PTAL

@adrianreber PTAL

libcontainer/container_linux.go
kolyshkin: > will adding defer unix.Close(fd) be enough here?

No. We should close it as soon as we exec'd criu swrk.

Fixed now

avagin:
LGTM. Thank you for fixing this!

adrianreber:
LGTM. Thanks!

AkihiroSuda:
LGTM

kolyshkin:
Description updated. @mrunalp @cyphar PTAL

mrunalp:
LGTM

PR #2343, opened 2020-04-22
fix data inconsistent when runc update in systemd driven cgroup

### LINKED ISSUE #2287: runc --systemd-cgroup update does not update systemd scope
A repro

1. Prepare a bundle and a spec. Amend the config.json with
```json
        "pids": {
                "limit": 55
        }
```
in the `linux.resources` section, and a long-lived process such as sleep.

2. Start the container:

```console
# runc --systemd-cgroup run -d keke
```

3. Check its pids limit in cgroups scope, as well as in cgroupfs:

```console
# systemctl show runc-keke.scope | grep 'Task'
TasksCurrent=2
TasksAccounting=yes
TasksMax=55

# cat /sys/fs/cgroup/system.slice/runc-keke.scope/pids.max 
55
```

4. Update the limit:
```console
# runc --systemd-cgroup update keke --pids-limit 66
```

5. Check again:

```console
# systemctl show runc-keke.scope | grep 'Task'
TasksCurrent=2
TasksAccounting=yes
TasksMax=55

# cat /sys/fs/cgroup/system.slice/runc-keke.scope/pids.max 
66
```

As we can see, the kernel and the systemd is now in disagreement about the limit.

6. Check that the limit can in principal be updated:

```console
# systemctl show runc-keke.scope | grep 'Task'
TasksCurrent=2
TasksAccounting=yes
TasksMax=77
DropInPaths=/run/systemd/transient/runc-keke.scope.d/50-TasksMax.conf
# cat /sys/fs/cgroup/system.slice/runc-keke.scope/pids.max 
77
```

In the above example, we set/check/update pids.limit, but all the other resources behave in a similar way.

The cause of this is `Set()` method of `systemd.UnifiedManager`, which only uses fsManager to set the updated values.

lifubang:
fix #2287 
`runc --systemd-cgroup update does not update systemd scope`

As pointed out by @kolyshkin `The cause of this is Set() method of systemd.UnifiedManager, which only uses fsManager to set the updated values.`
So when we use systemd driven cgroup v2, we should use `SetUnitProperties` method to update container resource constraints.

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -34,6 +34,59 @@ func NewUnifiedManager(config *configs.Cgroup, path string, rootless bool) *unif
 	}
 }
 
+func genV2ResourcesProperties(c *configs.Cgroup) ([]systemdDbus.Property, error) {
+	var properties []systemdDbus.Property
+
+	if c.Resources.Memory != 0 {
+		properties = append(properties,
+			newProp("MemoryMax", uint64(c.Resources.Memory)))
+	}
+
+	swap, err := cgroups.ConvertMemorySwapToCgroupV2Value(c.Resources.MemorySwap, c.Resources.Memory)
+	if err != nil {
+		return nil, err
+	}
+	if swap > 0 {
+		properties = append(properties,
+			newProp("MemorySwapMax", uint64(swap)))
+	}
+
+	if c.Resources.CpuWeight != 0 {
+		properties = append(properties,
+			newProp("CPUWeight", c.Resources.CpuWeight))
+	}
+
+	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
+	if c.Resources.CpuQuota != 0 && c.Resources.CpuPeriod != 0 {
+		// corresponds to USEC_INFINITY in systemd
+		// if USEC_INFINITY is provided, CPUQuota is left unbound by systemd
+		// always setting a property value ensures we can apply a quota and remove it later
+		cpuQuotaPerSecUSec := uint64(math.MaxUint64)
+		if c.Resources.CpuQuota > 0 {
+			// systemd converts CPUQuotaPerSecUSec (microseconds per CPU second) to CPUQuota
+			// (integer percentage of CPU) internally.  This means that if a fractional percent of
+			// CPU is indicated by Resources.CpuQuota, we need to round up to the nearest
+			// 10ms (1% of a second) such that child cgroups can set the cpu.cfs_quota_us they expect.
+			cpuQuotaPerSecUSec = uint64(c.Resources.CpuQuota*1000000) / c.Resources.CpuPeriod
+			if cpuQuotaPerSecUSec%10000 != 0 {
+				cpuQuotaPerSecUSec = ((cpuQuotaPerSecUSec / 10000) + 1) * 10000
+			}
+		}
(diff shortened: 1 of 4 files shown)


lifubang:
If it makes sense, the same changes can be added to cgroup v1.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -263,11 +274,23 @@ func (m *unifiedManager) GetStats() (*cgroups.Stats, error) {
...
-	fsMgr, err := m.fsManager()
-	if err != nil {
+	var (
+		c        = m.cgroups
+		unitName = getUnitName(c)
+	)
+
+	if properties, err := genResourcesProperties(c); err != nil {
 		return err
+	} else {
```
kolyshkin, line 284: no need for "else" here since you have `return` above.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -82,55 +135,13 @@ func (m *unifiedManager) Apply(pid int) error {
...
-	if c.Resources.Memory != 0 {
-		properties = append(properties,
-			newProp("MemoryMax", uint64(c.Resources.Memory)))
-	}
-
-	swap, err := cgroups.ConvertMemorySwapToCgroupV2Value(c.Resources.MemorySwap, c.Resources.Memory)
-	if err != nil {
+	if resourcesProperties, err := genResourcesProperties(c); err != nil {
 		return err
+	} else {
```
kolyshkin, line 140: no need for `else` here

libcontainer/cgroups/systemd/v2.go
```diff
@@ -263,11 +274,23 @@ func (m *unifiedManager) GetStats() (*cgroups.Stats, error) {
...
+		dbusConnection, err := getDbusConnection()
+		if err != nil {
+			return err
+		}
+		if err := dbusConnection.SetUnitProperties(unitName, true, properties...); err != nil {
+			return err
+		}
+		return nil
 	}
-	return fsMgr.Set(container)
```
kolyshkin, line 270: some resources are not controlled by systemd so we still need to call this I'm afraid.

kolyshkin: and this is why CI is failing (systemd does not control cpuset)

lifubang: Force pushed now.
I don't know whether we should change to only write `cpuset` cgroup files or not?

lifubang:
> but as you say it also applies to cgroup v1, so it would be nice if you could fix that at the same time!

OK, I'll fix and test it in cgroup v1.

libcontainer/cgroups/systemd/v2.go
odinuge: I think the best is to keep it like it is now (with writing all of them), and then we can look at it in another PR/issue. This also applies to other resources not set via systemd like hugetlb. 

More info about it here: https://systemd.io/CGROUP_DELEGATION/

kolyshkin:
LGTM

kolyshkin:
@lifubang maybe we can add some simple tests, based on what I described in #2287.

lifubang:
> maybe we can add some simple tests

OK, I'll do it later.

lifubang:
Three tests has been added.

AkihiroSuda:
LGTM

kolyshkin:
LGTM

PR #2347, opened 2020-04-22
cgroupv2: allow to set EnableAllDevices=true

kolyshkin:
In this case we just do not install any eBPF rules
checking the devices.


AkihiroSuda:
Can we have integration test for devices?

kolyshkin:
@giuseppe PTAL

> Can we have integration test for devices?

Do you mean for this particular case, or setting permission for devices in general?

In case of latter, can you describe what you want to test?

AkihiroSuda:
The latter in the long term, but I think we can begin with the former one

giuseppe:
@mrunalp PTAL

mrunalp:
LGTM

AkihiroSuda:
LGTM, integration test can be added later.

PR #2352, opened 2020-04-25
fix runc events error in cgroup v2

### LINKED ISSUE #2351: cgroupv2: runc events is broken
How to reproduce:
```
[root@localhost ~]# runc events test
ERRO[0000] open /sys/fs/cgroup/user.slice/user-a.slice/aa-bb.scope/memory.oom_control: no such file or directory
open /sys/fs/cgroup/user.slice/user-a.slice/aa-bb.scope/memory.oom_control: no such file or directory
```

lifubang:
fix #2351 

Because there is no `oom notifier` in `cgroup v2` now.
I think we should use `Inotify` to watch `OOM` in `memory.events`.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -648,6 +648,13 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
 	if c.config.RootlessCgroups {
 		logrus.Warn("getting OOM notifications may fail if you don't have the full access to cgroups")
 	}
+	if cgroups.IsCgroup2UnifiedMode() {
+		path, err := c.cgroupManager.GetUnifiedPath()
+		if err != nil {
+			return nil, err
+		}
+		return notifyOnOOMV2(path)
+	}
 	return notifyOnOOM(c.cgroupManager.GetPaths())
 }
 
--- a/libcontainer/notify_linux_v2.go
+++ b/libcontainer/notify_linux_v2.go
@@ -1,0 +1,102 @@
+// +build linux
+
+package libcontainer
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"unsafe"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+func getValueFromCgroup(path, key string) (int, error) {
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+
+	lines := strings.Split(string(content), "\n")
+	for _, line := range lines {
+		arr := strings.Split(line, " ")
+		if len(arr) == 2 && arr[0] == key {
+			return strconv.Atoi(arr[1])
+		}
+	}
+	return 0, nil
+}
+
+func registerMemoryEventV2(cgDir, evName, cgEvName string) (<-chan struct{}, error) {
+	eventControlPath := filepath.Join(cgDir, evName)
+	cgEvPath := filepath.Join(cgDir, cgEvName)
+	fd, err := unix.InotifyInit()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to init inotify")
+	}
+	// watching oom kill
+	evFd, err := unix.InotifyAddWatch(fd, eventControlPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, errors.Wrap(err, "unable to add inotify watch")
+	}
+	// Because no `unix.IN_DELETE|unix.IN_DELETE_SELF` event for cgroup file system, so watching all process exited
+	cgFd, err := unix.InotifyAddWatch(fd, cgEvPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, errors.Wrap(err, "unable to add inotify watch")
+	}
+	ch := make(chan struct{})
+	go func() {
+		var (
+			buffer [unix.SizeofInotifyEvent + unix.PathMax + 1]byte
+			offset uint32
+		)
+		defer func() {
+			unix.Close(fd)
+			close(ch)
+		}()
+
+		for {
+			n, err := unix.Read(fd, buffer[:])
+			if err != nil {
+				logrus.Warnf("unable to read event data from inotify, got error: %v", err)
+				return
+			}
+			if n < unix.SizeofInotifyEvent {
+				logrus.Warnf("we should read at least %d bytes from inotify, but got %d bytes.", unix.SizeofInotifyEvent, n)
+				return
+			}
+			offset = 0
+			for offset <= uint32(n-unix.SizeofInotifyEvent) {
+				rawEvent := (*unix.InotifyEvent)(unsafe.Pointer(&buffer[offset]))
+				offset += unix.SizeofInotifyEvent + uint32(rawEvent.Len)
+				if rawEvent.Mask&unix.IN_MODIFY != unix.IN_MODIFY {
+					continue
+				}
+				switch int(rawEvent.Wd) {
+				case evFd:
+					oom, err := getValueFromCgroup(eventControlPath, "oom_kill")
+					if err != nil || oom > 0 {
+						ch <- struct{}{}
+					}
+				case cgFd:
+					pids, err := getValueFromCgroup(cgEvPath, "populated")
+					if err != nil || pids == 0 {
+						return
+					}
+				}
+			}
+		}
+	}()
+	return ch, nil
+}
+
+// notifyOnOOMV2 returns channel on which you can expect event about OOM,
+// if process died without OOM this channel will be closed.
+func notifyOnOOMV2(path string) (<-chan struct{}, error) {
+	return registerMemoryEventV2(path, "memory.events", "cgroup.events")
+}
--- a/tests/integration/events.bats
+++ b/tests/integration/events.bats
@@ -13,8 +13,8 @@ function teardown() {
 
 @test "events --stats" {
   # XXX: currently cgroups require root containers.
-  # TODO: support cgroup v2 memory.events
-  requires root cgroups_v1
+  requires root
+  init_cgroup_paths
 
   # run busybox detached
   runc run -d --console-socket $CONSOLE_SOCKET test_busybox
@@ -27,10 +27,10 @@ function teardown() {
   [[ "${lines[0]}" == *"data"* ]]
 }
 
-@test "events --interval default " {
+@test "events --interval default" {
   # XXX: currently cgroups require root containers.
-  # TODO: support cgroup v2 memory.events
-  requires root cgroups_v1
+  requires root
+  init_cgroup_paths
 
   # run busybox detached
   runc run -d --console-socket $CONSOLE_SOCKET test_busybox
@@ -55,10 +55,10 @@ function teardown() {
   [[ "${lines[0]}" == *"data"* ]]
 }
 
-@test "events --interval 1s " {
+@test "events --interval 1s" {
   # XXX: currently cgroups require root containers.
-  # TODO: support cgroup v2 memory.events
-  requires root cgroups_v1
+  requires root
+  init_cgroup_paths
 
   # run busybox detached
   runc run -d --console-socket $CONSOLE_SOCKET test_busybox
@@ -82,10 +82,10 @@ function teardown() {
   [ "$status" -eq 0 ]
 }
 
-@test "events --interval 100ms " {
+@test "events --interval 100ms" {
   # XXX: currently cgroups require root containers.
-  # TODO: support cgroup v2 memory.events
-  requires root cgroups_v1
+  requires root
+  init_cgroup_paths
 
   # run busybox detached
   runc run -d --console-socket $CONSOLE_SOCKET test_busybox
@@ -110,4 +110,41 @@ function teardown() {
 
   run eval "grep -q 'test_busybox' events.log"
   [ "$status" -eq 0 ]
+}
+
+@test "events oom " {
+  # XXX: currently cgroups require root containers.
+  requires root
+  init_cgroup_paths
+  
+  # we need the container to hit OOM, so disable swap
+  # ("swap" here is actually memory+swap)
+  DATA=$(cat <<EOF
+  "memory": {
+    "limit": 33554432,
+    "swap": 33554432
+  },
+EOF
+  )
+  DATA=$(echo ${DATA} | sed 's/\n/\\n/g')
+  sed -i "s/\(\"resources\": {\)/\1\n${DATA}/" ${BUSYBOX_BUNDLE}/config.json
+
+  # run busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  # spawn two sub processes (shells)
+  # the first sub process is an event logger that sends stats events to events.log
+  # the second sub process exec a memory hog process to cause a oom condition
+  # and waits for an oom event
+  (__runc events test_busybox > events.log) &
+  (
+    retry 10 1 eval "grep -q 'test_busybox' events.log"
+    __runc exec -d test_busybox sh -c 'test=$(dd if=/dev/urandom ibs=5120k)'
+    retry 10 1 eval "grep -q 'oom' events.log"
+    __runc delete -f test_busybox
+  ) &
+  wait # wait for the above sub shells to finish
+
+  grep -q '{"type":"oom","id":"test_busybox"}' events.log
 }


AkihiroSuda:
Thanks, please update https://github.com/opencontainers/runc/blob/master/tests/integration/events.bats as well

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,92 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
+)
+
+func checkOOM(path string) (bool, error) {
+	file, err := os.Open(path)
+	defer file.Close()
```
kolyshkin, line 19: move defer to after the `err != nil` check

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}
+		}
+	}()
+	return ch, nil
+}
+
+// notifyOnOOMV2 returns channel on which you can expect event about OOM,
+// if process died without OOM this channel will be closed.
+func notifyOnOOMV2(paths map[string]string) (<-chan struct{}, error) {
```
kolyshkin, line 85: We should not use `paths` for cgroupv2 code. in cgroupv2 there is only one path.

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return false, err
+	}
+	buff := bufio.NewReader(file)
+	for {
+		line, err := buff.ReadString('\n')
+		if err == io.EOF {
+			break
+		}
+		arr := strings.Fields(line)
```
kolyshkin, line 29: So, you are only analyzing the last line (which on my system starts with `oom_kill`).

What you should do (I think) is
1 read/parse the file
2 get the counters for `oom` and `oom_kill` lines
3 compare these with the previous readings (or `0` if there are none)
4 save the counters (to be used later on step 3)
5 if counters are increased, return `true`

libcontainer/container_linux.go
```diff
@@ -648,6 +648,9 @@ func (c *linuxContainer) NotifyOOM() (<-chan struct{}, error) {
 	if c.config.RootlessCgroups {
 		logrus.Warn("getting OOM notifications may fail if you don't have the full access to cgroups")
 	}
+	if cgroups.IsCgroup2UnifiedMode() {
+		return notifyOnOOMV2(c.cgroupManager.GetPaths())
```
kolyshkin, line 652: use `GetUnifiedPath()`

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/fsnotify/fsnotify"
```
kolyshkin, line 13: This package is not well maintained, and tries to work for all systems, sacrificing linux-specific features.

I havent't checked it, but can you take a look at how hard it would be to use `Inotify*` functions from `golang.org/x/sys/unix` directly?

lifubang:
> please update https://github.com/opencontainers/runc/blob/master/tests/integration/events.bats as well

enable cgroup v2 tests in events.bats now.

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+
+	oomCount := 0
+	lines := strings.Split(string(content), "\n")
+	for _, line := range lines {
+		arr := strings.Fields(line)
```
kolyshkin, line 30: `strings.Split(line, " ")` is 2x faster, if your input is standard (no extra spaces etc.), and in this case the input is standard.

AkihiroSuda:
LGTM

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return 0, nil
+}
+
+func registerMemoryEventV2(cgDir string, evName string, cgEvName string) (<-chan struct{}, error) {
+	eventControlPath := filepath.Join(cgDir, evName)
+	cgEvPath := filepath.Join(cgDir, cgEvName)
+	fd, err := unix.InotifyInit()
+	if err != nil {
+		return nil, err
```
kolyshkin, line 65: I'm afraid the error you get is very basic so you need to wrap it
```go
return errors.Wrap("unable to init inotify: %v", err)
```
or something like that

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	// watching oom kill
+	_, err = unix.InotifyAddWatch(fd, eventControlPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, err
```
kolyshkin, line 71: same here

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+		}
+	}
+	return oomCount, nil
+}
+
+func getPopulated(path string) (int, error) {
+	if _, err := os.Lstat(path); err != nil {
+		return 0, err
+	}
```
kolyshkin, line 44: Looks like if you remove the `os.Lstat()` code the functionality won't change -- you will get the same error from `ioutil.ReadFile` below.

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, line := range lines {
+		arr := strings.Split(line, " ")
+		if len(arr) == 2 && arr[0] == "populated" {
+			return strconv.Atoi(arr[1])
+		}
+	}
+	return 0, nil
+}
+
+func registerMemoryEventV2(cgDir string, evName string, cgEvName string) (<-chan struct{}, error) {
```
kolyshkin, line 60: nit: `(cgDir, evName, cgEvName string)`

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	_, err = unix.InotifyAddWatch(fd, eventControlPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, err
+	}
+	// Because no `unix.IN_DELETE|unix.IN_DELETE_SELF` event for cgroup file system, so watching all process exited
+	_, err = unix.InotifyAddWatch(fd, cgEvPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, err
```
kolyshkin, line 77: same here

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+		)
+		defer func() {
+			unix.Close(fd)
+			close(ch)
+		}()
+
+		for {
+			n, err := unix.Read(fd, buffer[:])
+			if err != nil {
+				logrus.Warn(err)
```
kolyshkin, line 93: Same here. The error you'll get is something like "interruped system call" with no context whatsoever. Please add some context.

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+				return
+			}
```
kolyshkin, line 95: You probably need to check if enough bytes are read
```go
if n < unix.SizeofInotifyEvent {
	logrus.Warn("short read from inotify")
	return
}
```

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+			offset = 0
+			for offset <= uint32(n-unix.SizeofInotifyEvent) {
+				rawEvent := (*unix.InotifyEvent)(unsafe.Pointer(&buffer[offset]))
+				bytes := (*[unix.PathMax]byte)(unsafe.Pointer(&buffer[unix.SizeofInotifyEvent]))
+				eventName := strings.TrimRight(string(bytes[0:rawEvent.Len]), "\000")
```
kolyshkin, line 100: 1. This is actually a file name, not an event name.

2. It might be easier to just save the two fds returned by `InotifyAddWatch` above and check if `rawEvent.Wd` is one of them. Let me take a look...

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+				offset += unix.SizeofInotifyEvent + uint32(rawEvent.Len)
+				if rawEvent.Mask&unix.IN_MODIFY == unix.IN_MODIFY {
```
kolyshkin, line 102: I'd rather have it as
```go
rawEvent := (*unix.InotifyEvent)(unsafe.Pointer(&buffer[offset]))
if rawEvent.Mask&unix.IN_MODIFY != unix.IN_MODIFY {
	continue
}
bytes := (*[unix.PathMax]byte)(unsafe.Pointer(&buffer[unix.SizeofInotifyEvent]))
eventName := strings.TrimRight(string(bytes[0:rawEvent.Len]), "\000")
switch eventName {
....
```

It's less indentation and better readability that way

libcontainer/notify_linux_v2.go
kolyshkin: That's right. You can just check the fd, no need to check the file name. Simpler code and faster checks.

```go
evFd, err := unix.InotifyAddWatch(fd, eventControlPath, unix.IN_MODIFY)
...
cgFd, err := unix.InotifyAddWatch(fd, cgEvPath, unix.IN_MODIFY)
...
for offset <= uint32(n-unix.SizeofInotifyEvent) {
  event := (*unix.InotifyEvent)(unsafe.Pointer(&buffer[offset]))
  offset += unix.SizeofInotifyEvent + uint32(event.Len)
  if event.Mask&unix.IN_MODIFY != unix.IN_MODIFY {
    continue
  }
  switch int(event.Wd) {
    case evFd:
      oom, err := getOOMCount(eventControlPath)
      ...
    case cgFd:
      pids, err := getPopulated(cgEvPath)
...
```

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	// Because no `unix.IN_DELETE|unix.IN_DELETE_SELF` event for cgroup file system, so watching all process exited
+	_, err = unix.InotifyAddWatch(fd, cgEvPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, err
+	}
+	ch := make(chan struct{})
+	go func() {
+		var (
+			buffer [unix.SizeofInotifyEvent + unix.PathMax]byte
```
kolyshkin, line 82: inotify(7) man page says
> Specifying a buffer of size
> `           sizeof(struct inotify_event) + NAME_MAX + 1`
> will be sufficient to read at least one event.

Yours is 1 byte short.

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+
+	oomCount := 0
+	lines := strings.Split(string(content), "\n")
+	for _, line := range lines {
+		arr := strings.Split(line, " ")
+		if len(arr) == 2 && (arr[0] == "oom" || arr[0] == "oom_kill") {
```
kolyshkin, line 31: Most probably you only need `"oom_kill"` here, since `"oom"' event means the cgroup was out of memory but wasn't necessarily killed.

kolyshkin:
Ideally I'd like to add a test case for OOM event, since currently we only test for those periodical stats.

Can you add a separate patch testing for OOM event? It's pretty easy: run a memory hog inside a container and wait for OOM event.

A memory hog can be created in sh like this:
`sh -c 'VAR=$(seq 1 100000000)'`

lifubang:
> Can you add a separate patch testing for OOM event? It's pretty easy:

I have add a test now, but not easy, we should disable memory swap, or else it’s very hard to cause oom.
And the meaning of swap value now is different in v1 and v2, it’s quite inconvenient to use. So I have post an issue.

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,102 @@
...
+	cgEvPath := filepath.Join(cgDir, cgEvName)
+	fd, err := unix.InotifyInit()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to init inotify")
+	}
+	// watching oom kill
+	evFd, err := unix.InotifyAddWatch(fd, eventControlPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, errors.Wrap(err, "unable to add watch inotify")
```
kolyshkin, line 44: `s/watch inotify/inotify watch/`

libcontainer/notify_linux_v2.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	// Because no `unix.IN_DELETE|unix.IN_DELETE_SELF` event for cgroup file system, so watching all process exited
+	cgFd, err := unix.InotifyAddWatch(fd, cgEvPath, unix.IN_MODIFY)
+	if err != nil {
+		unix.Close(fd)
+		return nil, errors.Wrap(err, "unable to add watch inotify")
```
kolyshkin, line 50: ditto

tests/integration/events.bats
```diff
@@ -111,3 +111,56 @@ function teardown() {
...
+  # XXX: currently cgroups require root containers.
+  requires root
+  init_cgroup_paths
+
+  # Set some initial known values
+  if [ "$CGROUP_UNIFIED" == "yes" ]; then
+  DATA=$(cat <<EOF
+  "memory": {
+    "limit": 33554433,
+    "kernel": 33554433,
```
kolyshkin, line 125: No need to specify kernel limit here, it serves no purpose (and is ignored for v2 anyway)

tests/integration/events.bats
```diff
@@ -111,3 +111,56 @@ function teardown() {
...
+    "swappiness": 0,
```
kolyshkin, line 126: same for swappiness

tests/integration/events.bats
```diff
@@ -111,3 +111,56 @@ function teardown() {
...
   [ "$status" -eq 0 ]
 }
+
+@test "events oom " {
+  # XXX: currently cgroups require root containers.
+  requires root
+  init_cgroup_paths
+
+  # Set some initial known values
+  if [ "$CGROUP_UNIFIED" == "yes" ]; then
```
kolyshkin, line 121: Let's merge https://github.com/opencontainers/runc/pull/2370 first so we won't have to have this switch here.

tests/integration/events.bats
```diff
@@ -111,3 +111,56 @@ function teardown() {
...
+  # spawn two sub processes (shells)
+  # the first sub process is an event logger that sends stats events to events.log
+  # the second sub process exec a memory hog process to cause a oom condition
+  # and waits for an oom event that includes test_busybox
+  # then kills the test_busybox container which causes the event logger to exit
+  (__runc events test_busybox > events.log) &
+  (
+    retry 10 1 eval "grep -q 'test_busybox' events.log"
+    __runc exec -d test_busybox sh -c 'VAR=$(seq 1 100000000)'
+    retry 10 1 eval "grep -q 'oom' events.log"
```
kolyshkin, line 158: maybe `grep -qF '"type":"oom"'` to be more strict about what we're looking for here.

tests/integration/events.bats
```diff
@@ -111,3 +111,56 @@ function teardown() {
...
+    __runc delete -f test_busybox
+  ) &
+  wait # wait for the above sub shells to finish
```
kolyshkin, line 161: Maybe add a small timeout in case runc won't ever exit?
`timeout 10 wait`

lifubang: Once  #2370 merged, I'll update it together with some other changes.

tests/integration/events.bats
lifubang: Yes, I just copy from v1, I'll remove `kernel` and `swappiness`.

tests/integration/events.bats
```diff
@@ -111,3 +111,56 @@ function teardown() {
...
+
+  [ -e events.log ]
```
kolyshkin, line 163: this check is redundant, grep below will print a message and return a non-zero exit code if the file is not found

tests/integration/events.bats
kolyshkin: @lifubang #2370 is merged

lifubang: @kolyshkin `timeout` and `wait` can't be used together.

tests/integration/events.bats
```diff
@@ -111,3 +111,40 @@ function teardown() {
   run eval "grep -q 'test_busybox' events.log"
   [ "$status" -eq 0 ]
 }
+
+@test "events oom " {
```
kolyshkin, line 115: nit: extra space before `"`

tests/integration/events.bats
```diff
@@ -111,3 +111,40 @@ function teardown() {
...
+  # XXX: currently cgroups require root containers.
+  requires root
+  init_cgroup_paths
+
+  # Set some initial known values
```
kolyshkin, line 120: Maybe replace this by
```go
# we need the container to hit OOM, so disable swap
# ("swap" here is actually memory+swap)
```

or something like that

tests/integration/events.bats
kolyshkin: Yes, you're right.

I am still concerned that this case might hang forever. Maybe we can run both `__runc` invocations with a timeout then. Something big, say 1 minute. Obviously, if OOM has not happened within a minute, the test has failed.

lifubang: Travis CI will return an error if there is no output after 150 seconds.
I met this error before I disable swap.

If you worry about this, I can write simple C program to ensure causing oom kill.

lifubang: @kolyshkin use `sh -c 'test=$(dd if=/dev/urandom ibs=5120k)'` to ensure oom kill.
Because the memory limit is 4M.

kolyshkin:
LGTM, thanks!

AkihiroSuda:
LGTM

PR #2355, opened 2020-04-27
libcontainer/stacktrace: fix inline func name

zhsj:
The TestCaptureTestFunc fails on gccgo, because captureFunc is inlined
there. Instead of using FuncForPC directly, using CallersFrames from
runtime package, which is available since go1.7

This commit has an API break for the stacktrace package, which removes
NewFrame func. But I think it's safe as the stacktrace is somehow an
internal package of libcontainer.

Signed-off-by: Shengjing Zhu <zhsj@debian.org>

### DIFF
--- a/libcontainer/stacktrace/capture.go
+++ b/libcontainer/stacktrace/capture.go
@@ -7,19 +7,22 @@ import "runtime"
 // skip is the number of frames to skip
 func Capture(userSkip int) Stacktrace {
 	var (
-		skip   = userSkip + 1 // add one for our own function
+		skip   = userSkip + 2 // add one for our own function, one for runtime.Callers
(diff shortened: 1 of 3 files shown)


zhsj:
@opencontainers/runc-maintainers could you take a look?

thaJeztah:
@kolyshkin @AkihiroSuda ptal

AkihiroSuda:
@cyphar PTAL?

kolyshkin (dismissed):
LGTM

zhsj:
ping?

rebased to pick up the new ci.

zhsj:
closing, as libcontainer/stacktrace is removed now.

PR #2358, opened 2020-04-27
cgroups/fs2: don't always parse /proc/self/cgroup

kolyshkin:
Function `defaultPath` always parses `/proc/self/cgroup`, but
the resulting value is not always used.

Avoid unnecessary reading/parsing by moving the code
to just before its use.

Modify the test case accordingly.

### DIFF
--- a/libcontainer/cgroups/fs2/defaultpath_test.go
+++ b/libcontainer/cgroups/fs2/defaultpath_test.go
@@ -17,8 +17,11 @@
 package fs2
 
 import (
+	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
 )
 
 func TestParseCgroupFromReader(t *testing.T) {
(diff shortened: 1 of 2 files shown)


libcontainer/cgroups/fs2/defaultpath_test.go
```diff
@@ -46,26 +47,33 @@ func TestParseCgroupFromReader(t *testing.T) {
 
 func TestDefaultDirPath(t *testing.T) {
```
AkihiroSuda, line 48: t.Skip on non-unified host

AkihiroSuda:
Looks good but a single nit

libcontainer/cgroups/fs2/defaultpath_test.go
kolyshkin: makes sense. updated

AkihiroSuda: where?

AkihiroSuda:
LGTM

mrunalp:
LGTM

PR #2359, opened 2020-04-28
docs: terminals: mention subreaper requirement

cyphar:
I realised that the terminal documentation which covers detached
terminals fails to mention that callers need to make themselves a
subreaper. Probably a good idea to mention this. I've also included a
minor comparison to LXC.


kolyshkin:
LGTM

AkihiroSuda:
LGTM

PR #2360, opened 2020-04-28
 Add initial support for rsvd accounting hugetlb cgroup

odinuge:
The previous non-rsvd max/limit_in_bytes does not account for reserved
huge page memory, making it possible for a processes to reserve all the
huge page memory, without being able to allocate it (due to cgroup
restrictions).

In practice this makes it possible to successfully mmap more huge page
memory than allowed via the cgroup settings, but when using the memory
the process will get a SIGBUS and crash. This is bad for applications
trying to mmap at startup (and it succeeds), but the program crashes
when starting to use the memory. eg. postgres is doing this by default.

This also keeps writing to the old max/limit_in_bytes, to make sure some
applications read the wrong value.

More info can be found here: https://lkml.org/lkml/2020/2/3/1153

---

Do we have to edit the runtime-spec in order to do this?

Also, this will fix https://github.com/zalando/patroni/issues/1393 (ref. the postgres part at the top ^)

### DIFF
--- a/libcontainer/cgroups/fs/hugetlb.go
+++ b/libcontainer/cgroups/fs/hugetlb.go
@@ -27,9 +27,28 @@ func (s *HugetlbGroup) Apply(d *cgroupData) error {
 	return nil
 }
 
+// HasReservationAccountingSupport checks if reservation accounting of huge pages in the hugetlb cgroup
+// is supported. This is supported from linux 5.7
+// https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/hugetlb.html
+func (s *HugetlbGroup) HasReservationAccountingSupport(path string) bool {
+	if len(HugePageSizes) == 0 {
+		return false
+	}
+	_, err := fscommon.ReadFile(path, strings.Join([]string{"hugetlb", HugePageSizes[0], "rsvd", "limit_in_bytes"}, "."))
+	return err == nil
+}
+
 func (s *HugetlbGroup) Set(path string, cgroup *configs.Cgroup) error {
+	supportsReservationAccounting := s.HasReservationAccountingSupport(path)
 	for _, hugetlb := range cgroup.Resources.HugetlbLimit {
 		if err := fscommon.WriteFile(path, strings.Join([]string{"hugetlb", hugetlb.Pagesize, "limit_in_bytes"}, "."), strconv.FormatUint(hugetlb.Limit, 10)); err != nil {
+			return err
+		}
+
+		if !supportsReservationAccounting {
+			continue
+		}
+		if err := fscommon.WriteFile(path, strings.Join([]string{"hugetlb", hugetlb.Pagesize, "rsvd", "limit_in_bytes"}, "."), strconv.FormatUint(hugetlb.Limit, 10)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


libcontainer/cgroups/fs/hugetlb.go
```diff
@@ -27,11 +27,30 @@ func (s *HugetlbGroup) Apply(d *cgroupData) error {
...
+func (s *HugetlbGroup) HasReservationAccountingSupport(path string) bool {
+	if len(HugePageSizes) == 0 {
+		return false
+	}
+	_, err := fscommon.ReadFile(path, strings.Join([]string{"hugetlb", HugePageSizes[0], "rsvd", "limit_in_bytes"}, "."))
+	return err == nil
+}
+
 func (s *HugetlbGroup) Set(path string, cgroup *configs.Cgroup) error {
+	supportsReservationAccounting := s.HasReservationAccountingSupport(path)
```
odinuge, line 42: Not sure if this is the best way to check, or should we try to "cache" the value like we do with `HugePageSizes`?

libcontainer/cgroups/fs2/hugetlb.go
```diff
@@ -39,15 +59,23 @@ func statHugeTlb(dirPath string, stats *cgroups.Stats) error {
 	}
 	hugetlbStats := cgroups.HugetlbStats{}
 
+	supportsReservationAccounting := HasReservationAccountingSupport(dirPath)
+
 	for _, pagesize := range hugePageSizes {
-		usage := strings.Join([]string{"hugetlb", pagesize, "current"}, ".")
+		filenamePrefix := strings.Join([]string{"hugetlb", pagesize}, ".")
```
kolyshkin, line 65: nit: maybe it would be better to have it as
```go
filenamePrefix := "hugetlb."+pagesize
```
(for readability)

libcontainer/cgroups/fs2/hugetlb.go
```diff
@@ -39,15 +59,23 @@ func statHugeTlb(dirPath string, stats *cgroups.Stats) error {
...
+
+		if supportsReservationAccounting {
+			filenamePrefix += ".rsvd"
+		}
+
+		usage := fmt.Sprintf("%s.current", filenamePrefix)
 		value, err := fscommon.GetCgroupParamUint(dirPath, usage)
 		if err != nil {
 			return errors.Wrapf(err, "failed to parse hugetlb.%s.current file", pagesize)
```
kolyshkin, line 74: The error message from `GetCgroupParamUint` already contain file name, so you can return the error as-is, no need to wrap it.

libcontainer/cgroups/fs2/hugetlb.go
```diff
@@ -39,15 +59,23 @@ func statHugeTlb(dirPath string, stats *cgroups.Stats) error {
...
 		}
 		hugetlbStats.Usage = value
 
-		fileName := strings.Join([]string{"hugetlb", pagesize, "events"}, ".")
+		fileName := fmt.Sprintf("%s.events", filenamePrefix)
```
kolyshkin, line 78: nit: using `fileName := filenamePrefix + ".events"` would be faster

but either way is fine

libcontainer/cgroups/fs2/hugetlb.go
kolyshkin: also the error now returns the wrong file name in case `supportsReservationAccounting` is set

libcontainer/cgroups/fs2/hugetlb.go
```diff
@@ -19,14 +20,33 @@ func isHugeTlbSet(cgroup *configs.Cgroup) bool {
 	return len(cgroup.Resources.HugetlbLimit) > 0
 }
 
+// HasReservationAccountingSupport checks if reservation accounting of huge pages in the hugetlb cgroup
+// is supported. This is supported from linux 5.7
+// https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/hugetlb.html
+func HasReservationAccountingSupport(dirPath string) bool {
+	hugePageSizes, err := cgroups.GetHugePageSize()
```
kolyshkin, line 27: Yes, I think it makes sense to do this check once, using `sync.Once`.

libcontainer/cgroups/fs/hugetlb_test.go
```diff
@@ -65,6 +70,58 @@ func TestHugetlbSetHugetlb(t *testing.T) {
 	}
 }
 
+func TestHugetlbSetHugetlbWithReservedAccounting(t *testing.T) {
+	helper := NewCgroupTestUtil("hugetlb", t)
```
kolyshkin, line 74: shouldn't this test be skipped `if !HasReservationAccountingSupport()`?

libcontainer/cgroups/fs/hugetlb.go
```diff
@@ -27,11 +27,30 @@ func (s *HugetlbGroup) Apply(d *cgroupData) error {
...
 }
 
+// HasReservationAccountingSupport checks if reservation accounting of huge pages in the hugetlb cgroup
+// is supported. This is supported from linux 5.7
+// https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/hugetlb.html
+func (s *HugetlbGroup) HasReservationAccountingSupport(path string) bool {
+	if len(HugePageSizes) == 0 {
+		return false
+	}
+	_, err := fscommon.ReadFile(path, strings.Join([]string{"hugetlb", HugePageSizes[0], "rsvd", "limit_in_bytes"}, "."))
```
kolyshkin, line 37: use `cgroups.PathExists` here

libcontainer/cgroups/fs2/hugetlb.go
kolyshkin: or not... since different cgroups can have different controls I guess...

libcontainer/cgroups/fs2/hugetlb.go
```diff
@@ -19,14 +20,33 @@ func isHugeTlbSet(cgroup *configs.Cgroup) bool {
...
+	if err != nil || len(hugePageSizes) == 0 {
+		return false
+	}
+	_, err = fscommon.ReadFile(dirPath, strings.Join([]string{"hugetlb", hugePageSizes[0], "rsvd", "max"}, "."))
```
kolyshkin, line 31: use `cgroups.PathExists()`

libcontainer/cgroups/fs2/hugetlb.go
kolyshkin: Fixing this should be done as a separate first patch I think.

kolyshkin:
> Do we have to edit the runtime-spec in order to do this?

I'm afraid yes. Reservation and use are two different properties, and we should not mix them together.

kolyshkin:
So, @odinuge, I think this should start with a PR to https://github.com/opencontainers/runtime-spec. Once merged, we can open a PR here (and most of the comments that I left reviewing this are still valid).

kolyshkin:
I'm working on reviving this PR now, once the spec is merged.

PR #2370, opened 2020-05-01
let runc disable swap in cgroup v2

### LINKED ISSUE #2369: cgroup2: can't set memory.swap.max to 0
How to reproduce:
1. set memory limit in `config.json` like this:
```
"memory": {
	 "limit": 33554433,
	 "swap":33554433,
	 "kernel": 33554433,
	 "swappiness": 0
}
```
set `swap`=`limit`, or `swap`=0
2. runc run -d test
3. look the value in `memory.swap.max`, it will get `max`, but not 0.

If we use `--systemd-cgroup`, we will also meet the same problem.

lifubang:
fix #2369

I think we should let cgroup fs/systemd driver can use 0 to disable memory swap in cgroup v2.

### DIFF
--- a/libcontainer/cgroups/fs2/memory.go
+++ b/libcontainer/cgroups/fs2/memory.go
@@ -45,10 +45,13 @@ func setMemory(dirPath string, cgroup *configs.Cgroup) error {
 	if err != nil {
 		return err
 	}
-	if val := numToStr(swap); val != "" {
-		if err := fscommon.WriteFile(dirPath, "memory.swap.max", val); err != nil {
-			return err
-		}
+	swapStr := numToStr(swap)
+	if swapStr == "" && swap == 0 && cgroup.Resources.MemorySwap > 0 {
+		// memory and memorySwap set to the same value -- disable swap
+		swapStr = "0"
+	}
+	if err := fscommon.WriteFile(dirPath, "memory.swap.max", swapStr); err != nil {
+		return err
 	}
 
 	if val := numToStr(cgroup.Resources.Memory); val != "" {
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -39,12 +39,12 @@ func genV2ResourcesProperties(c *configs.Cgroup) ([]systemdDbus.Property, error)
 		properties = append(properties,
 			newProp("MemoryMax", uint64(c.Resources.Memory)))
 	}
-
-	swap, err := cgroups.ConvertMemorySwapToCgroupV2Value(c.Resources.MemorySwap, c.Resources.Memory)
-	if err != nil {
-		return nil, err
-	}
-	if swap > 0 {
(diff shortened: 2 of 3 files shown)


libcontainer/cgroups/fs2/memory.go
```diff
@@ -45,7 +45,12 @@ func setMemory(dirPath string, cgroup *configs.Cgroup) error {
 	if err != nil {
 		return err
 	}
-	if val := numToStr(swap); val != "" {
+	// When we set `swap`, we should let us set `memory.swap.max` to 0
+	if swap == 0 && cgroup.Resources.MemorySwap > 0 {
```
kolyshkin, line 49: I would change the comment to something like
```go
// memory and memorySwap set to the same value -- disable swap
```

kolyshkin: and rewrite the whole block to avoid mentioning `fscommon.WriteFile(dirPath, "memory.swap.max"` twice.

For example
```go
swapStr := numToStr(swap)
if swapStr == "" && swap == 0 && cgroup.Resources.MemorySwap > 0 {
  // memory and memorySwap set to the same value -- disable swap
  swapStr = "0"
}
if err := fscommon.WriteFile(dirPath, "memory.swap.max", swapStr); err != nil {
  return err
}
```

libcontainer/cgroups/systemd/v2.go
```diff
@@ -44,7 +44,8 @@ func genV2ResourcesProperties(c *configs.Cgroup) ([]systemdDbus.Property, error)
 	if err != nil {
 		return nil, err
 	}
-	if swap > 0 {
+	// When we set `swap`, we should let us set `MemorySwapMax` to 0
+	if (c.Resources.MemorySwap > 0 && swap >= 0) || swap == -1 {
```
kolyshkin, line 48: I would do
```go
if (c.Resources.MemorySwap != 0) { // limit is set
￼	swap, err := cgroups.ConvertMemorySwapToCgroupV2Value(c.Resources.MemorySwap, c.Resources.Memory)
￼	if err != nil {
￼		return nil, err
￼	}
  properties = append(properties,
￼			newProp("MemorySwapMax", uint64(swap)))
￼	}
}
```

kolyshkin:
The commit should also include
```diff
-                return 0, errors.New("memory+swap limit should be > memory limit")
+                return 0, errors.New("memory+swap limit should be >= memory limit")
```

kolyshkin:
and if you can copy-paste the description from https://github.com/opencontainers/runc/issues/2369 into the commit message, so it would be possible to find out what's going out by just reading git log, without going to github, that'd be great!

kolyshkin:
LGTM, pullapprove ;)

kolyshkin:
@AkihiroSuda @mrunalp  PTAL

mrunalp:
LGTM

PR #2373, opened 2020-05-03
Logging nits

kolyshkin:
1. Remove a function `fatalf()` which were only used once.

2. Fix error "stuttering" (printing the error twice).

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -45,12 +45,20 @@ func checkArgs(context *cli.Context, expected, checkType int) error {
 	return nil
 }
 
+func logrusToStderr() bool {
+	l, ok := logrus.StandardLogger().Out.(*os.File)
+	return ok && l.Fd() == os.Stderr.Fd()
+}
+
 // fatal prints the error's details if it is a libcontainer specific error type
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
LGTM

utils.go
```diff
@@ -45,12 +45,20 @@ func checkArgs(context *cli.Context, expected, checkType int) error {
 	return nil
 }
 
+func logrusToStderr() bool {
+	l, ok := logrus.StandardLogger().Out.(*os.File)
+	return ok && l.Fd() == os.Stderr.Fd()
```
thaJeztah, line 50: Is this still needed? ISTR old versions of logrus writing to stdout, but they changed that (at least by default) in later versions

kolyshkin: It is needed because there is a way to redirect the logging to a file, and I did not want to introduce an additional "logging is redirected" flag to he code.

thaJeztah: Ah; gotcha

mrunalp:
LGTM

PR #2375, opened 2020-05-03
Close fd in case fd.Write() returns error

tedyu:
In waitForCriuLazyServer(), if fd.Write() returns error, current code would return without closing fd.

This PR closes the fd in above case to avoid leak.


AkihiroSuda:
LGTM

mrunalp:
LGTM

PR #2377, opened 2020-05-03
Simplify ticks, as the value is a constant

thaJeztah:
While checking dependencies, I stumbled on https://github.com/containerd/cgroups/pull/12, which removes the CGO dependency from the containerd/cgroups package. Given that that code was based on the one in libcontainer, looks like we can make the same change here. See the discussion on that PR for further details

See for example in the Musl libc source code https://git.musl-libc.org/cgit/musl/tree/src/conf/sysconf.c#n29

This removes the cgo dependency for the system package.


thaJeztah:
ping @kolyshkin @AkihiroSuda PTAL

AkihiroSuda:
LGTM

thaJeztah:
@kolyshkin LGTY? 🤗

libcontainer/cgroups/fs/cpuacct.go
```diff
@@ -20,7 +20,7 @@ const (
 	nanosecondsInSecond = 1000000000
 )
 
-var clockTicks = uint64(system.GetClockTicks())
+var clockTicks = system.GetClockTicks()
```
kolyshkin, line 23: I'd rather drop `system.GetClockTicks()` entirely and have this as
```go
const clockTicks = 100
```

(with a comment that explains where it is coming from)

kolyshkin: Having it as const will help compiler to optimize the calculations, and I doubt there are any external users of `system.GetClockTicks()`.

thaJeztah: It may still be used in some locations outside of runc, so we should probably be careful with removing `system.GetClockTicks()` altogether, but do you want me to put a const in this file (so that it doesn't depend on `libnetwork/system` ?

AkihiroSuda: Go API compatibility is not guaranteed across releases (at least until v1.0.0 GA), so we can safely remove `system.GetClockTicks()`.

thaJeztah:
Removed the function, and defined a local `const`

libcontainer/cgroups/fs/cpuacct.go
```diff
@@ -12,15 +12,17 @@ import (
...
 
 const (
 	cgroupCpuacctStat   = "cpuacct.stat"
 	nanosecondsInSecond = 1000000000
-)
 
-var clockTicks = uint64(system.GetClockTicks())
+	// The value comes from `C.sysconf(C._SC_CLK_TCK)`, and
+	// on Linux it's a constant which is safe to be hard coded,
+	// so we can avoid using cgo here.
```
AkihiroSuda, line 23: Please add link to `https://github.com/containerd/cgroups/pull/12#issuecomment-311327696`

thaJeztah: 🤦 I had it there, but thought it was redundant; let me update

thaJeztah: Added

AkihiroSuda:
LGTM

kolyshkin:
LGTM

thaJeztah:
Thanks!

PR #2379, opened 2020-05-04
Check the error return from startTime()

tedyu:
In linuxContainer#currentState(), the error return from initProcess.startTime() is ignored.

Looking at various Process implementations, not every implementation returns nil error. e.g. setnsProcess#startTime() may return non-nil error.

This PR adds error check for the return value.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1890,7 +1890,10 @@ func (c *linuxContainer) currentState() (*State, error) {
 	)
 	if c.initProcess != nil {
 		pid = c.initProcess.pid()
-		startTime, _ = c.initProcess.startTime()
+		var err error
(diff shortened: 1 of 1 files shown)


libcontainer/container_linux.go
```diff
@@ -1890,7 +1890,10 @@ func (c *linuxContainer) currentState() (*State, error) {
 	)
 	if c.initProcess != nil {
 		pid = c.initProcess.pid()
-		startTime, _ = c.initProcess.startTime()
+		var err error
+		if startTime, err = c.initProcess.startTime(); err != nil {
```
thaJeztah, line 1894: I don't think failing to get `startTime` is a critical error; if we return here because of that error, critical paths of the code base would no longer be able to function, so I don't think we should return an error here (at _most_ it could be logged, but perhaps it's not even important to do so)

PR #2380, opened 2020-05-04
libcontainer: RunningInUserNS() use sync.Once

thaJeztah:
Made similar changes in containerd, so thought I'd do the same here.

FWIW; perhaps we could use a sync.once on `CurrentProcessUIDMap` and `CurrentProcessGIDMap` which do the actual reading and parsing, but I wasn't sure if there would be situations where we needed to retry those if they failed (open to suggestions there)

### DIFF
--- a/libcontainer/system/linux.go
+++ b/libcontainer/system/linux.go
@@ -5,6 +5,7 @@ package system
 import (
 	"os"
 	"os/exec"
+	"sync"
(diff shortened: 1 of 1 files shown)


dims:
LGTM

AkihiroSuda:
LGTM

mrunalp:
LGTM

Issue #2383, opened 2020-05-06
Enable travis to run arm64, ppc64le and s390x


alicefr:
It might be useful to test runc for non-x86 architectures. Travis has the support for arm64, ppc64le and s390x. We can simply enable to run the tests also for other arches.

alicefr:
Here the travis blog how to enable multiple cpu architectures: https://docs.travis-ci.com/user/multi-cpu-architectures/

XiaodongLoong:
@alicefr 
I have raised a PR (https://github.com/opencontainers/runc/pull/2508) to run runc integration test on more CPU architecture.  Does travis support for mips64le?

alicefr:
@XiaodongLoong not sure, the best would be to try :). You can add the additional arch in the travis yaml on your branch and see if the job gets triggered.

alicefr:
@AkihiroSuda I'm changing job, so I don't have time to finish my PR. The tests are still failing for s390x, ppc64le and arm64. Not sure what is going wrong since when I run them directly on s390x, everything is successful. My latest updates can be found here https://github.com/alicefr/runc/tree/travis-multi-arch. I'm closing this issue since I'm not working on this anymore.

PR #2384, opened 2020-05-06
enable travis for multi-arch

alicefr:
Enable running the tests on arm64, ppc64le and s390x.

Fixes: https://github.com/opencontainers/runc/issues/2383

Signed-off-by: Alice Frosi <afrosi@de.ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,35 @@ matrix:
         - sudo ssh default -t 'cd /vagrant && sudo make localrootlessintegration RUNC_USE_SYSTEMD=yes'
         # same setup but with fs2 driver (rootless) instead of systemd
         - sudo ssh default -t 'cd /vagrant && sudo make localrootlessintegration'
+    - go: 1.13.x
+      name: "test-local-arm64"
+      arch: arm64
+      before_install:
+        - sudo apt-get install -y jq seccomp libseccomp-dev bats
+      script:
+        - make all
+        - sudo PATH="$PATH" make localintegration
+    - go: 1.13.x
+      name: "test-local-ppc64le"
+      arch: ppc64le
+      before_install:
+        - sudo apt-get install -y jq seccomp libseccomp-dev bats
+      script:
+        - make all
+        - sudo PATH="$PATH" make localintegration
+    - go: 1.13.x
+      name: "test-local-s390x"
+      arch: s390x
+      before_install:
+        - sudo apt-get install -y jq seccomp libseccomp-dev bats
+      script:
+        - make all
+        - sudo PATH="$PATH" make localintegration
   allow_failures:
     - go: tip
+    - name: "test-local-arm64"
+    - name: "test-local-ppc64le"
+    - name: "test-local-s390x"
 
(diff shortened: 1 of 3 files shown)


alicefr:
Vagrant is only avaiable for x86

alicefr:
Is it possible to run part of the test suite without vagrant?

kolyshkin:
Many tests are run not on Vagrant but inside a podman container. We only use Vagrant for cgroupv2 tests, which require a very recent kernel and distro.

kolyshkin:
You'll also need to add more arches to `tests/integration/multi-arch.bash` in order to be able to run integration tests (`/tests/integration/`).

alicefr:
it's a little bit more complicated because on the native architecture we're missing some dependencies in  https://github.com/opencontainers/runc/blob/master/Dockerfile#L15-L18

kolyshkin:
Yeah, I think the crossbuild should either be restricted to be run on amd64. Better yet, since we are now actually testing it on different platforms, we don't need it any longer.

The only thing that bugs me is in crossbuild we buld for three different ARMs (including 32-bit arm6 and arm8 aka hf), while in travis we only have arm64. Not sure how important it is bit maybe we can cross-build 32-bit arm on 64-bit arm?

AkihiroSuda:
If it is completed you can run tests outside Docker as in https://github.com/opencontainers/runc/blob/0f7ffbebeb9e3e697dd0efc6f4ab9dc14054960e/.travis.yml#L16-L22

alicefr:
@AkihiroSuda I'll try that out. Sorry I was on vacation and busy with other stuff. Not sure why the CI is using x86 images instead of the built one. I need to have a closer look.

AkihiroSuda:
Travis isn't responding, could you push again?

alicefr:
Let's check if by re-triggering travis will run

alicefr:
any idea what I'm doing wrong because travis doesn't start?

AkihiroSuda:
Not sure, could you try reopening a new PR?

alicefr:
yes, I can try!

alicefr:
Opened this: https://github.com/opencontainers/runc/pull/2477 . Let's see if this solves the issue

alicefr:
@AkihiroSuda it seems that travis didn't like the arch in the script. I removed that and travis CI started. There was a parsing issue. I'll close the second PR as it's a simple copy of this.

alicefr:
The PR is not ready yet. I had to modify the Dockerfile to get the right dependencies for x86. I also had to split the test for each architecture because travis matrix doesn't allow to define multiple architecture for the same entry. Right now, I have some issue with the tests on the other architecture. I need to investigate a little deeper. I guess they're getting a wrong image/binary

alicefr:
For ppc64le and s390x, I'm missing the hello-world tarball: https://github.com/opencontainers/runc/tree/master/tests/integration/testdata

AkihiroSuda:
@alicefr https://github.com/docker-library/hello-world

AkihiroSuda:
Travis not responding, please open a new one

PR #2386, opened 2020-05-07
Simplify cgroup paths handling in v2 via unified v1/v2 API

kolyshkin:
This unties the Gordian Knot of using GetPaths in cgroupv2 code.

The problem is, the current code uses GetPaths for three kinds of things:

1. Get all the paths to cgroup v1 controllers to save its state (see
   (*linuxContainer).currentState(), (*LinuxFactory).loadState()
   methods).

2. Get all the paths to cgroup v1 controllers to have the setns process
    enter the proper cgroups in `(*setnsProcess).start()`.

3. Get the path to a specific controller (for example,
   `m.GetPaths()["devices"]`).

Now, for cgroup v2 instead of a set of per-controller paths, we have only
one single unified path, and a dedicated function `GetUnifiedPath()` to get it.

This discrepancy between v1 and v2 cgroupManager API leads to the
following problems with the code:

 - multiple if/else code blocks that have to treat v1 and v2 separately;

 - backward-compatible GetPaths() methods in v2 controllers;

 - repeated writing of the PID into the same cgroup for v2;

Overall, it's hard to write the right code with all this, and the code
that is written is kinda hard to follow.

The solution is to slightly change the API to do the 3 things outlined
above in the same manner for v1 and v2:

1. Use `GetPaths()` for state saving and setns process cgroups entering.

2. Introduce and use `Path(subsys string)` to obtain a path to a
   subsystem. For v2, the argument is ignored and the unified path is
   returned.

This commit converts all the controllers to the new API, and modifies
all the users to use it.

### DIFF
--- a/libcontainer/cgroups/cgroups.go
+++ b/libcontainer/cgroups/cgroups.go
@@ -27,27 +27,22 @@ type Manager interface {
 	// Destroys the cgroup set
 	Destroy() error
 
-	// The option func SystemdCgroups() and Cgroupfs() require following attributes:
-	// 	Paths   map[string]string
-	// 	Cgroups *configs.Cgroup
-	// Paths maps cgroup subsystem to path at which it is mounted.
-	// Cgroups specifies specific cgroup settings for the various subsystems
-
-	// Returns cgroup paths to save in a state file and to be able to
-	// restore the object later.
-	GetPaths() map[string]string
-
-	// GetUnifiedPath returns the unified path when running in unified mode.
-	// The value corresponds to the all values of GetPaths() map.
-	//
-	// GetUnifiedPath returns error when running in hybrid mode as well as
-	// in legacy mode.
-	GetUnifiedPath() (string, error)
+	// Path returns a cgroup path to the specified controller/subsystem.
+	// For cgroupv2, the argument is unused and can be empty.
+	Path(string) string
 
 	// Sets the cgroup as configured.
 	Set(container *configs.Config) error
 
-	// Gets the cgroup as configured.
+	// GetPaths returns cgroup path(s) to save in a state file in order to restore later.
+	//
+	// For cgroup v1, a key is cgroup subsystem name, and the value is the path
+	// to the cgroup for this subsystem.
+	//
+	// For cgroup v2 unified hierarchy, a key is "", and the value is the unified path.
+	GetPaths() map[string]string
+
+	// GetCgroups returns the cgroup data as configured.
 	GetCgroups() (*configs.Cgroup, error)
 }
 
--- a/libcontainer/cgroups/fs/apply_raw.go
+++ b/libcontainer/cgroups/fs/apply_raw.go
@@ -209,15 +209,10 @@ func (m *manager) Destroy() error {
 	return nil
 }
 
-func (m *manager) GetPaths() map[string]string {
+func (m *manager) Path(subsys string) string {
 	m.mu.Lock()
-	paths := m.paths
-	m.mu.Unlock()
-	return paths
-}
-
-func (m *manager) GetUnifiedPath() (string, error) {
-	return "", errors.New("unified path is only supported when running in unified mode")
+	defer m.mu.Unlock()
+	return m.paths[subsys]
 }
 
 func (m *manager) GetStats() (*cgroups.Stats, error) {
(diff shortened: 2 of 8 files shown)


kolyshkin:
The only ugliness that is left is
1. function `func getUnifiedPath()` in `libcontainer/factory_linux.go`. It can be simplified but we'll lose backward compatibility (not sure if we need it so I'm not touching it).
2. a requirement to pass a string parameter to `Path()` for v2. I think it's an OK price to pay for the API which works for v1 and v2, otherwise we'd need an if/else and it will be uglier.
3. Public members of fs.Manager and systemd.LegacyManager. Will address it later.

kolyshkin:
@AkihiroSuda PTAL

libcontainer/cgroups/cgroups.go
```diff
@@ -27,27 +27,18 @@ type Manager interface {
...
 	Set(container *configs.Config) error
 
-	// Gets the cgroup as configured.
+	// GetPaths returns cgroup path(s) to save in a state file in order
+	// to restore it later.
+	GetPaths() map[string]string
+
+	// GetCgroups returns the cgroup data as configured.
 	GetCgroups() (*configs.Cgroup, error)
 }
```
AkihiroSuda, line 43:
```go
type ManagerV1 interface{
  Manager
  Paths() map[string]string
}

type ManagerV2 interface{
  Manager
  Path() string
}
```

kolyshkin: That was one of my ideas as well, yes, but it seems the approach in this  PR is better.

Problem with the above is we need an if/switch everywhere. With this PR the difference in code using it is minimized.

kolyshkin: Second problem with the approach above is the save state structure is different for v1 and v2, and that is another complication which is avoided by this PR quite elegantly.

AkihiroSuda: If we use the same interface for v1 and v, can we rename `GetPaths() map[string]string` to something like `StateJSON() (marshalled string, err error)` ?

kolyshkin: It will require merging the two json objects, as well as some additional steps in loadState(), which is not going to look nice I'm afraid.

I'll look into it.

AkihiroSuda: Just `State() map[string]string` maybe.

kolyshkin: Unfortunately we can't do it since there is one place where this GetPaths() is literally used as a set of paths:

https://github.com/opencontainers/runc/blob/64416d34f30eaf69af6938621137b393ada63a16/libcontainer/process_linux.go#L102..L103

(`struct setnsProcess` is defined at 
https://github.com/opencontainers/runc/blob/64416d34f30eaf69af6938621137b393ada63a16/libcontainer/process_linux.go#L102..L103)

AkihiroSuda:
LGTM

kolyshkin:
I think I'll split it into two PRs for the sake of easier review.

Part I: https://github.com/opencontainers/runc/pull/2387 (relatively easy to review)
Part II is this one.

libcontainer/cgroups/cgroups.go
```diff
@@ -27,27 +27,18 @@ type Manager interface {
...
+	// For cgroupv2, the argument is unused and can be empty.
+	Path(string) string
 
 	// Sets the cgroup as configured.
 	Set(container *configs.Config) error
 
-	// Gets the cgroup as configured.
+	// GetPaths returns cgroup path(s) to save in a state file in order
+	// to restore it later.
+	GetPaths() map[string]string
```
AkihiroSuda, line 39: Needs document for v2 key ("")

kolyshkin: It was documented in `struct State` in `libcontainer/container_linux.go`, but it doesn't hurt to repeat it here.

Thanks, fixed.

kolyshkin:

libcontainer/cgroups/fs/apply_raw.go
```diff
@@ -411,6 +404,12 @@ func CheckCpushares(path string, c uint64) error {
 	return nil
 }
 
+func (m *manager) GetPaths() map[string]string {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.paths
```
kolyshkin, line 410: Not sure if the locking here is needed. Might help in the case when another goroutine is populating the paths in `Apply()`, but practically I don't think this is a possibility. Nevertheless, nothing wrong to have it here.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -434,6 +429,13 @@ func setKernelMemory(c *configs.Cgroup) error {
 	}
 	return fs.EnableKernelMemoryAccounting(path)
 }
+
+func (m *legacyManager) GetPaths() map[string]string {
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
kolyshkin, line 435: ditto wrt locking

AkihiroSuda:
LGTM

kolyshkin:
@mrunalp @cyphar PTAL

mrunalp:
LGTM

PR #2390, opened 2020-05-08
cgroupv2: don't enable threaded mode by default

lifubang:
After this commit: https://github.com/opencontainers/runc/commit/60c647e3b8e0d04cc8fe85b957ba2addfcc4fc4f 
Runc enable `threaded` mode in cgroup v2 by default.
If the cgroupPath is set to a absolute path like `/docker/********`, the memory subsystem can't be used by this mode.
So, I think we should use `domain` mode by default.

### DIFF
--- a/libcontainer/cgroups/fs2/create.go
+++ b/libcontainer/cgroups/fs2/create.go
@@ -63,6 +63,13 @@ func neededControllers(cgroup *configs.Cgroup) ([]string, error) {
 	return list, nil
 }
 
+// containsDomainController returns whether the current config contains domain controller or not.
+// Refer to: http://man7.org/linux/man-pages/man7/cgroups.7.html
+// As at Linux 4.19, the following controllers are threaded: cpu, perf_event, and pids.
+func containsDomainController(cg *configs.Cgroup) bool {
+	return isMemorySet(cg) || isIoSet(cg) || isCpuSet(cg) || isHugeTlbSet(cg)
+}
+
 // CreateCgroupPath creates cgroupv2 path, enabling all the
 // needed controllers in the process.
 func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
@@ -95,10 +102,34 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 					}
 				}()
 			}
-			// Write cgroup.type explicitly.
-			// Otherwise ENOTSUP may happen.
-			cgType := filepath.Join(current, "cgroup.type")
-			_ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
+			cgTypeFile := filepath.Join(current, "cgroup.type")
+			cgType, _ := ioutil.ReadFile(cgTypeFile)
+			switch strings.TrimSpace(string(cgType)) {
+			// If the cgroup is in an invalid mode (usually this means there's an internal
+			// process in the cgroup tree, because we created a cgroup under an
+			// already-populated-by-other-processes cgroup), then we have to error out if
+			// the user requested controllers which are not thread-aware. However, if all
+			// the controllers requested are thread-aware we can simply put the cgroup into
+			// threaded mode.
+			case "domain invalid":
+				if containsDomainController(c) {
+					return fmt.Errorf("cannot enter cgroupv2 %q with domain controllers -- it is in an invalid state", current)
+				} else {
+					// Not entirely correct (in theory we'd always want to be a domain --
+					// since that means we're a properly delegated cgroup subtree) but in
+					// this case there's not much we can do and it's better than giving an
+					// error.
+					_ = ioutil.WriteFile(cgTypeFile, []byte("threaded"), 0644)
+				}
+			// If the cgroup is in (threaded) or (domain threaded) mode, we can only use thread-aware controllers
+			// (and you cannot usually take a cgroup out of threaded mode).
+			case "domain threaded":
+				fallthrough
+			case "threaded":
+				if containsDomainController(c) {
+					return fmt.Errorf("cannot enter cgroupv2 %q with domain controllers -- it is in %s mode", current, strings.TrimSpace(string(cgType)))
+				}
+			}
 		}
 		// enable needed controllers
 		if i < len(elements)-1 {
--- a/tests/rootless.sh
+++ b/tests/rootless.sh
@@ -99,13 +99,6 @@ function enable_cgroup() {
 		chown root:rootless "$CGROUP_MOUNT/$CGROUP_PATH" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
 		chmod g+rwx "$CGROUP_MOUNT/$CGROUP_PATH"
 		chmod g+rw "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
-		# Fix up cgroup.type.
-		echo threaded > "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"
-		# Make sure cgroup.type doesn't contain "invalid". Otherwise write ops will fail with ENOTSUP.
-		# See http://man7.org/linux/man-pages/man7/cgroups.7.html
-		if grep -qw invalid "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"; then
-			exit 1
-		fi
 	fi
 }
 


lifubang:
@AkihiroSuda PTAL

cyphar:
I'm not sure I understand why we're putting cgroupv2 into threaded mode -- @AkihiroSuda is this to work around some odd permission issue (and why does it fix an `EPERM`)? The description in 60c647e3b8e0d04cc8fe85b957ba2addfcc4fc4f and #2340 is pretty sparse.

libcontainer/cgroups/fs2/create.go
```diff
@@ -95,22 +95,24 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 					}
 				}()
 			}
-			// Write cgroup.type explicitly.
-			// Otherwise ENOTSUP may happen.
-			cgType := filepath.Join(current, "cgroup.type")
-			_ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
+			if rootless {
```
AkihiroSuda, line 98:
```go
cgType := filepath.Join(current, "cgroup.type")
cgTypeB, _ := ioutil.ReadFile(cgType)
if strings.TrimSpace(string(cgTypeB)) == "domain invalid" {
  _ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
}
```

AkihiroSuda:
@cyphar 
Writing `threaded` to `cgroup.type` is required when `cgroup.type` becomes `domain invalid`, mostly in nested environment
```console
$ sudo podman run -it --rm --privileged --runtime=crun alpine
/ # cd /sys/fs/cgroup/
/sys/fs/cgroup # cat cgroup.controllers 
cpuset cpu io memory pids
/sys/fs/cgroup # cat cgroup.subtree_control 
/sys/fs/cgroup # echo +cpu > cgroup.subtree_control
/sys/fs/cgroup # mkdir foo
/sys/fs/cgroup # cat foo/cgroup.type
domain invalid
```

cyphar:
Okay, but "domain invalid" means that the cgroup is in an invalid state (meaning that one of the cgroup rules has been violated). Putting a cgroup into threaded mode doesn't fix that -- it switches it into an alternative mode which only allows controllers which are thread-aware to be enabled (such as cpu). This is why we can't enable the memory controller.

IMHO, a more complete solution would be to figure out how to deal with the parent cgroup having child processes (which is a bit dodgy if we're going to move other programs on the system between cgroups) or to simply give an error if we hit `domain invalid` and explain what the issue is.

lifubang:
> or to simply give an error if we hit `domain invalid` and explain what the issue is.

Personally, I prefer this one. @AkihiroSuda WDYT? Because in [man7 cgroups](http://man7.org/linux/man-pages/man7/cgroups.7.html):
```
2. "Internal" processes are not permitted.  With the exception of the
          root cgroup, processes may reside only in leaf nodes (cgroups that
          do not themselves contain child cgroups).  The details are
          somewhat more subtle than this, and are described below.
```
```
Cgroups v2 "no internal processes" rule
       Cgroups v2 enforces a so-called "no internal processes" rule.
       Roughly speaking, this rule means that, with the exception of the
       root cgroup, processes may reside only in leaf nodes (cgroups that do
       not themselves contain child cgroups).  This avoids the need to
       decide how to partition resources between processes which are members
       of cgroup A and processes in child cgroups of A.

       For instance, if cgroup /cg1/cg2 exists, then a process may reside in
       /cg1/cg2, but not in /cg1.  This is to avoid an ambiguity in cgroups
       v1 with respect to the delegation of resources between processes in
       /cg1 and its child cgroups.  The recommended approach in cgroups v2
       is to create a subdirectory called leaf for any nonleaf cgroup which
       should contain processes, but no child cgroups.  Thus, processes
       which previously would have gone into /cg1 would now go into
       /cg1/leaf.  This has the advantage of making explicit the relation‐
       ship between processes in /cg1/leaf and /cg1's other children.

       The "no internal processes" rule is in fact more subtle than stated
       above.  More precisely, the rule is that a (nonroot) cgroup can't
       both (1) have member processes, and (2) distribute resources into
       child cgroups—that is, have a nonempty cgroup.subtree_control file.
       Thus, it is possible for a cgroup to have both member processes and
       child cgroups, but before controllers can be enabled for that cgroup,
       the member processes must be moved out of the cgroup (e.g., perhaps
       into the child cgroups).

       With the Linux 4.14 addition of "thread mode" (described below), the
       "no internal processes" rule has been relaxed in some cases.
```

AkihiroSuda:
SGTM, but when no domain controller is enabled, we can write "threaded" without retuning error

libcontainer/cgroups/fs2/create.go
```diff
@@ -95,10 +96,20 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
...
-			_ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
+			// There are 4 types: 'domain', 'threaded', 'domain threaded', and 'domain invalid'
+			// If got 'domain invalid', we should check parent's cgroup type.
+			// When no domain controller is enabled, we can write "threaded" without retuning erro
+			cgTypeFile := filepath.Join(current, "cgroup.type")
+			cgType, _ := ioutil.ReadFile(cgTypeFile)
+			if strings.TrimSpace(string(cgType)) == "domain invalid" {
+				cgTypeParentFile := filepath.Join(current, "../cgroup.type")
+				cgTypeParent, _ := ioutil.ReadFile(cgTypeParentFile)
+				if bytes.HasPrefix(cgTypeParent, []byte("domain")) {
```
AkihiroSuda, line 107: IIUC we don't need to check parent, we just need to check whether the current config contains domain controller

lifubang:
https://github.com/kolyshkin/runc/blob/1d143562d2d491d054e28815565fe371c461b8a6/tests/rootless.sh#L103
I think `rootless` doesn't need `threaded` mode in default?
I don't know whether my opinion is right or not. @AkihiroSuda

AkihiroSuda:
Let's check invalid cgroup.type and set threaded conditionally

tests/rootless.sh
```diff
@@ -100,7 +100,9 @@ function enable_cgroup() {
 		chmod g+rwx "$CGROUP_MOUNT/$CGROUP_PATH"
 		chmod g+rw "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
 		# Fix up cgroup.type.
-		echo threaded > "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"
+		if grep -qw invalid "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"; then
+			echo threaded > "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"
+		fi
 		# Make sure cgroup.type doesn't contain "invalid". Otherwise write ops will fail with ENOTSUP.
```
AkihiroSuda, line 106: You can drop L106-L110 now

AkihiroSuda:
LGTM

libcontainer/cgroups/fs2/create.go
```diff
@@ -63,6 +63,25 @@ func neededControllers(cgroup *configs.Cgroup) ([]string, error) {
...
+	}
+	if isIoSet(cgroup) {
+		return true
+	}
+	if isCpuSet(cgroup) {
+		return true
+	}
+	if isHugeTlbSet(cgroup) {
+		return true
+	}
```
kolyshkin, line 81: Maybe just
```go
func containsDomainController(cg *configs.Cgroup) bool {
  return isMemorySet(cg) || isIoSet(cg) || isCpuSet(cg) || isHugeTlbSet(cg)
}

File: libcontainer/cgroups/fs2/create.go [hunk 5]
```diff
@@ -95,10 +114,18 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 					}
 				}()
 			}
-			// Write cgroup.type explicitly.
-			// Otherwise ENOTSUP may happen.
-			cgType := filepath.Join(current, "cgroup.type")
-			_ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
+			// There are 4 types: 'domain', 'threaded', 'domain threaded', and 'domain invalid'.
+			// If got 'domain invalid', we should check whether the current config contains domain controller or not.
+			// When does not contain domain controller, we can write "threaded" without retuning erro.
```
kolyshkin, line 119: `s/When does not /If it does not /`

`s/ retuning erro./ returning error./`

tests/rootless.sh
kolyshkin: @lifubang ^^^

lifubang: This has been removed before LGTM.
I think we should keep these comments to let other people know why we need to write threaded to cgroups.type.

AkihiroSuda:
LGTM

AkihiroSuda:
@kolyshkin LGTY?

libcontainer/cgroups/fs2/create.go
```diff
@@ -95,10 +102,18 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
...
-			// Write cgroup.type explicitly.
-			// Otherwise ENOTSUP may happen.
-			cgType := filepath.Join(current, "cgroup.type")
-			_ = ioutil.WriteFile(cgType, []byte("threaded"), 0644)
+			// There are 4 types: 'domain', 'threaded', 'domain threaded', and 'domain invalid'.
+			// If got 'domain invalid', we should check whether the current config contains domain controller or not.
+			// If it does not contain domain controller, we can write "threaded" without retuning error.
+			cgTypeFile := filepath.Join(current, "cgroup.type")
+			cgType, _ := ioutil.ReadFile(cgTypeFile)
+			if strings.TrimSpace(string(cgType)) == "domain invalid" {
```
cyphar, line 110: One other thing we should consider is to give an error early if we are trying to join a cgroup in `threaded` mode but have domain controllers enabled in the configuration. That way we can avoid weird error messages and help users understand why cgroup setup is failing.

tests/rootless.sh
cyphar: Is the `echo threaded >` still needed? Seems a bit odd to run the entire test suite under `threaded`.

libcontainer/cgroups/fs2/create.go
lifubang: Yes, I'll find the best location to check and give the error to user.

libcontainer/cgroups/fs2/create.go
```diff
@@ -95,10 +102,18 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
...
+				if containsDomainController(c) {
+					return fmt.Errorf("Can't enter cgroup path %s, the cgroup type is 'domain invalid.", current)
```
kolyshkin, line 112: Nit: you have opening single quote before "domain invalid", but do not have a closing one.

tests/rootless.sh
```diff
@@ -100,11 +100,10 @@ function enable_cgroup() {
 		chmod g+rwx "$CGROUP_MOUNT/$CGROUP_PATH"
 		chmod g+rw "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.subtree_control" "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.procs" "$CGROUP_MOUNT/cgroup.procs"
 		# Fix up cgroup.type.
-		echo threaded > "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"
 		# Make sure cgroup.type doesn't contain "invalid". Otherwise write ops will fail with ENOTSUP.
 		# See http://man7.org/linux/man-pages/man7/cgroups.7.html
 		if grep -qw invalid "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"; then
-			exit 1
+			echo threaded > "$CGROUP_MOUNT/$CGROUP_PATH/cgroup.type"
```
kolyshkin, line 106: So, if we fixed this in `CreateCgroupPath`, there is no more need to set it here, or am I missing something?

libcontainer/cgroups/fs2/create.go
cyphar: You can do it right here, maybe something more like this (I've also cleaned up the error messages and comments):

```
cgTypeFile := filepath.Join(current, "cgroup.type")
cgType, _ := ioutil.ReadFile(cgTypeFile)
switch strings.TrimSpace(string(cgType)) {
// If the cgroup is in an invalid mode (usually this means there's an internal
// process in the cgroup tree, because we created a cgroup under an
// already-populated-by-other-processes cgroup), then we have to error out if
// the user requested controllers which are not thread-aware. However, if all
// the controllers requested are thread-aware we can simply put the cgroup into
// threaded mode.
case "domain invalid":
	if containsDomainController(c) {
		return fmt.Errorf("cannot enter cgroupv2 %q with domain controllers -- it is in an invalid state", current)
	} else {
		// Not entirely correct (in theory we'd always want to be a domain --
		// since that means we're a properly delegated cgroup subtree) but in
		// this case there's not much we can do and it's better than giving an
		// error.
		_ = ioutil.WriteFile(cgTypeFile, []byte("threaded"), 0644)
	}
// If the cgroup is in threaded mode, we can only use thread-aware controllers
// (and you cannot usually take a cgroup out of threaded mode).
case "threaded":
	if containsDomainController(c) {
		return fmt.Errorf("cannot enter cgroupv2 %q with domain controllers -- it is in threaded mode", current)
	}
}
```

lifubang: add `case "domain threaded":`

cyphar: I *think* we can safely use domain controllers in a `domain threaded` cgroup because it's the cgroup *above* a threaded cgroup. But honestly I'm not too sure -- I need to set up an Ubuntu 20.04 VM to test this stuff...

lifubang: Domain controllers can’t be used in a domain threaded cgroup. I have tested it in Fedora 31.

cyphar: Ah okay, thanks for testing it.

cyphar:
LGTM.

giuseppe:
@AkihiroSuda @kolyshkin PTAL

kolyshkin:
LGTM

Issue #2366, opened 2020-04-30
cgroup: devices updates appear to be broken


cyphar:
This affects both versions, but in quite different ways:

* [x] For cgroupv1, #2205 highlighted that on device cgroup updates, we temporarily block all devices. This results in spurious errors in the container (such as programs being unable to open `/dev/null`). We've seen this happen on customer systems under Kubernetes, so this is definitely a real issue.
  + This is actually a more complicated issue than it first appears because `runc` actually incorrectly implements the spec here -- technically `runc` actually is a black-list by default and users have to convert `runc` to be a white-list. Aside from not following the spec this is a worrying security stance.

* [ ] For cgroupv2, devices cgroup updates are implemented by appending a new BPF program to the cgroup. This means that only new denials have an effect, and thus it's incorrectly implemented. (EDIT: This also means that we "leak" eBPF programs and thus after 64+ applications we start getting errors -- see #2474.)
  + ~~Unfortunately this is a bit complicated to fix, but I have figured out how to do it. We need to make an eBPF map of type `BPF_MAP_TYPE_PROG_ARRAY`and then tail-call into it in a small stub eBPF program which we attach to the actual cgroup. This which will allow us to atomically update the devices cgroup rules (there is no way to atomically replace an eBPF program with `BPF_F_ALLOW_MULTI` -- and without any program, all device accesses would be permitted).~~
  + Ignore the above -- you cannot `bpf_tail_call` from cgroup programs. So we will need to instead implement it through an eBPF map (which we can atomically replace by mis-using `BPF_MAP_TYPE_ARRAY_OF_ARRAY`).
  + This is all slightly complicated by the fact that the entire API is fd-based (and we don't have our own monitor process so we can't stash away the fd). But luckily there is a lookup-by-id system which we can use to get the file descriptor, though the ids can be recycled so we'll need to be careful to make sure we don't start touching the wrong eBPF map -- and unlike eBPF programs, eBPF maps don't store information about when they were created.

Part of #2315.

cyphar:
Fixing the white-list problem actually isn't too bad -- we just need to re-calculate the device list (merging entries which overlap with each other) rather than blindly applying the provided list of devices in-order.

cyphar:
[...]

PR #2391, opened 2020-05-08
cgroup: devices: major cleanups and minimal transition rules

### LINKED ISSUE #2204: Update command causes 'Operation not permitted'

cyphar:
#2366 is mostly fixed by this, though the patch is unfortunately a little more complicated than you might've first expected (it includes an emulator for the devices controller). Unfortunately this logic is necessary in order to be able to calculate and verify the minimal transition rules for a cgroup update.

A similar issue affects the cgroupv2 devices setup, but that is a topic for another time (as the solution is drastically different).

Fixes #2204


cyphar:
@mYmNeo does this PR fix the original issue for you?

cyphar:
According to my testing, this does fix #2204.

libcontainer/cgroups/fs/devices.go
```diff
@@ -31,44 +31,24 @@ func (s *DevicesGroup) Set(path string, cgroup *configs.Cgroup) error {
...
-			file := "devices.deny"
-			if dev.Allow {
-				file = "devices.allow"
-			}
-			if err := fscommon.WriteFile(path, file, dev.CgroupString()); err != nil {
-				return err
-			}
-		}
-		return nil
+	// The devices list is a blacklist, so we must first deny all devices.
```
kolyshkin: do you mean `whitelist` here?

libcontainer/cgroups/fs2/devices.go
```diff
@@ -3,35 +3,27 @@
...
-		case 'w':
-			w = true
-		case 'm':
-			m = true
-		}
-	}
+func isRWM(perm string) bool {
+	r := strings.Contains(perm, "r")
+	w := strings.Contains(perm, "w")
+	m := strings.Contains(perm, "m")
```
kolyshkin, line 18: Looks like the old implementation is about 3x more efficient. Not that this is a critical part :smile: but I don't see a reason to replace it.

kolyshkin: I would do that for readability but to my mind both the old and the new one is fine.

libcontainer/configs/device.go
```diff
@@ -35,23 +28,153 @@ type Device struct {
...
+	var set uint
+	if strings.Contains(string(p), "r") {
+		set |= deviceRead
+	}
+	if strings.Contains(string(p), "w") {
+		set |= deviceWrite
+	}
+	if strings.Contains(string(p), "m") {
+		set |= deviceMknod
+	}
```
kolyshkin, line 54: 1. Is it OK to ignore unknown characters here?

2. I would rewrite it to do
```go
for _, c := range string(p) {
  switch c {
  case 'r':
    set |= deviceRead
  case 'w':
......
....
```

kolyshkin: > Is it OK to ignore unknown characters here?

oh, I see you have `IsValid()` below, nice.

AkihiroSuda:
Can we have integration tests?

libcontainer/configs/device.go
```diff
@@ -35,23 +28,153 @@ type Device struct {
...
+	return fmt.Sprintf("%c %s:%s %s", d.Type, major, minor, d.Permissions)
 }
 
-// deviceNumberString converts the device number to a string return result.
-func deviceNumberString(number int64) string {
-	if number == Wildcard {
-		return "*"
+func (d *DeviceRule) Mkdev() (uint64, error) {
+	if d.Major == Wildcard || d.Minor == Wildcard {
+		return 0, fmt.Errorf("cannot mkdev() device with wildcards")
```
kolyshkin, line 177: nit: no need to use formatting here since there's no arguments to format, so `errors.New()` is better.

libcontainer/configs/device.go
```diff
@@ -35,23 +28,153 @@ type Device struct {
...
+	Permissions DevicePermissions `json:"permissions"`
+
+	// Allow specifies whether this rule is allowed.
+	Allow bool `json:"allow"`
+}
+
+func (d *DeviceRule) CgroupString() string {
+	var (
+		major = fmt.Sprint(d.Major)
+		minor = fmt.Sprint(d.Minor)
```
kolyshkin, line 164: nit: better use `strconf.Itoa()` which doesn't have to reflect on the argument type and is faster.

libcontainer/cgroups/fs/devices.go
```diff
@@ -26,29 +31,74 @@ func (s *DevicesGroup) Apply(d *cgroupData) error {
...
 		}
 	}
+
+	// Final safety check -- ensure that the resulting state is what was
+	// requested. This is only really correct for white-lists, but for
+	// black-lists we can at least check that the cgroup is in the right mode.
+	//
+	// This safety-check is skipped for the unit tests because we cannot
+	// currently mock devices.list correctly.
+	if !s.testingSkipFinalCheck {
```
kolyshkin, line 91: nit:
```go
if s.testingSkipFinalCheck {
  return nil
}
....
```

mYmNeo:
@cyphar I've tested these patches, it works well. Thank you very much. 👍

libcontainer/cgroups/fs/devices.go
cyphar: Yup, will fix.

libcontainer/cgroups/fs2/devices.go
cyphar: Yeah, I'm not actually sure why I did this tbh. :thinking:

libcontainer/configs/device.go
cyphar: Yeah, I'll switch this to also use the iterator setup.

libcontainer/configs/device.go
cyphar: Sure, though I personally don't like mixing `fmt` and `errors` in the same file.

cyphar:
@AkihiroSuda Working on it.

libcontainer/cgroups/fs/devices.go
cyphar: I disagree, I think that an early-return like that makes it harder to read.

cyphar:
@crosbymichael I'd like to know if you're okay with the `/dev/console` rule removal as well as some of the core refactors. Quite a bit of this code predates even my involvement in `libcontainer`.

cyphar:
@AkihiroSuda Integration test added.

tests/integration/update.bats
```diff
@@ -312,3 +312,42 @@ EOF
...
+	mv config.json{.tmp,}
+
+	# Set up a temporary console socket and recvtty so we can get the stdio.
+	TMP_RECVTTY_DIR="$(mktemp -d "$BATS_TMPDIR/runc-tmp-recvtty.XXXXXX")"
+	TMP_RECVTTY_PID="$TMP_RECVTTY_DIR/recvtty.pid"
+	TMP_CONSOLE_SOCKET="$TMP_RECVTTY_DIR/console.sock"
+	CONTAINER_OUTPUT="$TMP_RECVTTY_DIR/output"
+	("$RECVTTY" --no-stdin --pid-file "$TMP_RECVTTY_PID" \
+				--mode single "$TMP_CONSOLE_SOCKET" &>"$CONTAINER_OUTPUT" &) &
+	sleep 0.5s
```
AkihiroSuda, line 330: maybe too short, `sleep 1s` ?

tests/integration/update.bats
```diff
@@ -312,3 +312,42 @@ EOF
...
+	# We trigger the update a few times to make sure we hit the race.
+	for _ in {1..12}
+	do
+		# TODO: Update "runc update" so we can change the device rules.
+		runc update --pids-limit 30 test_update_devices
+		[ "$status" -eq 0 ]
+	done
+
+	# Kill recvtty.
+	kill -9 "$(<"$TMP_RECVTTY_PID")"
```
AkihiroSuda, line 348: Is SIGKILL needed?

cyphar: We do `kill -9` for the regular `recvtty` setup, I just copied it.

libcontainer/cgroups/devices_emulator.go
```diff
@@ -0,0 +1,351 @@
+// +build linux
+
+package cgroups
```
AkihiroSuda, line 3: Maybe we want to create a dedicated package for `libcontainer/cgroups/devicesemulator`

libcontainer/cgroups/devices_emulator.go
```diff
@@ -0,0 +1,351 @@
...
+		err = errors.Wrap(e.rmRule(*rule), "remove 'deny' exception")
+	} else {
+		err = errors.Wrap(e.addRule(*rule), "add 'allow' exception")
+	}
+	return err
+}
+
+func (e *DevicesEmulator) deny(rule *deviceRule) error {
+	// This cgroup is configured as a white-list. Reset the entire emulator,
+	// and put is into white-list mode.
```
AkihiroSuda, line 202: put what?

cyphar: Typo, should be "put it".

libcontainer/cgroups/ebpf/devicefilter/devicefilter_test.go
```diff
@@ -83,71 +83,69 @@ block-2:
         19: Exit
 block-3:
         20: JNEImm dst: r2 off: -1 imm: 2 <block-4>
-        21: JNEImm dst: r4 off: -1 imm: 5 <block-4>
-        22: JNEImm dst: r5 off: -1 imm: 1 <block-4>
+        21: JNEImm dst: r4 off: -1 imm: 1 <block-4>
```
AkihiroSuda, line 86: How are these changes introduced?

cyphar: The removal of the `/dev/console` rule means that `5:1` is no longer in the default whitelist, which changes the code generated.

tests/integration/update.bats
cyphar: I didn't have issues with my testing, but sure I'll bump it up.

libcontainer/configs/device.go
```diff
@@ -35,23 +29,154 @@ type Device struct {
...
+func (p DevicePermissions) Intersection(o DevicePermissions) DevicePermissions {
+	lhs := p.toSet()
+	rhs := o.toSet()
+	return fromSet(lhs & rhs)
+}
+
+// IsEmpty returns whether the set of permissions in a DevicePermissions is
+// empty.
+func (p DevicePermissions) IsEmpty() bool {
+	return p == DevicePermissions("")
```
AkihiroSuda, line 99: Maybe needs `strings.TrimSpace` somewhere

cyphar: We could add it, but there's no point where we parse `DevicePermissions` that would allow a space to be added -- there's `IsValid` checks everywhere relevant. Honestly this whole `DevicePermissions` thing is to work around the fact that we (stupidly) embedded the string representation into the libcontainer state format...

cyphar:
The new integration test fails under the systemd cgroup driver. I'll need to figure out what's going on there.

libcontainer/cgroups/fs/devices.go
kolyshkin: To me earlier return is almost always more readable, but only marginally in this case so whatever works for you

tests/integration/update.bats
```diff
@@ -312,3 +312,46 @@ EOF
...
+	jq '.process.args = ["sh", "-c", "while true; do echo >/dev/null; done"]' config.json > config.json.tmp
+	mv config.json{.tmp,}
+
+	# Set up a temporary console socket and recvtty so we can get the stdio.
+	TMP_RECVTTY_DIR="$(mktemp -d "$BATS_TMPDIR/runc-tmp-recvtty.XXXXXX")"
+	TMP_RECVTTY_PID="$TMP_RECVTTY_DIR/recvtty.pid"
+	TMP_CONSOLE_SOCKET="$TMP_RECVTTY_DIR/console.sock"
+	CONTAINER_OUTPUT="$TMP_RECVTTY_DIR/output"
+	("$RECVTTY" --no-stdin --pid-file "$TMP_RECVTTY_PID" \
+				--mode single "$TMP_CONSOLE_SOCKET" &>"$CONTAINER_OUTPUT" &) &
```
kolyshkin, line 333: I know this comes from `helpers.bats` but the double `&` is not really needed here, you'll be fine with something like
```shell
("$RECVTTY" ... ) &
```

IOW inner `&` is not needed.

tests/integration/update.bats
```diff
@@ -312,3 +312,46 @@ EOF
...
+	sleep 1s
```
kolyshkin, line 334: I hate sleeps in the test code :( it always means we're not doing something right

In this case, I guess, there is no way to check that recvtty is ready.

Or can we maybe wait till "$TMP_CONSOLE_SOCKET" is created? Something like
```shell
retry 10 0.1 test -e "$TMP_CONSOLE_SOCKET"
```
?

tests/integration/update.bats
```diff
@@ -312,3 +312,46 @@ EOF
...
+
+	# Run the container in the background.
+	runc run -d --console-socket "$TMP_CONSOLE_SOCKET" test_update
+	cat "$CONTAINER_OUTPUT"
+	[ "$status" -eq 0 ]
+
+	# Trigger an update. This update doesn't actually change the device rules,
+	# but it will trigger the devices cgroup code to reapply the current rules.
+	# We trigger the update a few times to make sure we hit the race.
```
kolyshkin, line 343: Slow Saturday here... which race?

cyphar: That the container will try to write to `/dev/null` during the window between `a *:* rwm` being denied and the rest of the white-list being applied.

tests/integration/update.bats
cyphar: Yeah that'd work.

libcontainer/cgroups/devices_emulator.go
cyphar: I've put it in `libcontainer/cgroups/devices` so we can write `devices.Emulator`.

cyphar:
Ah, I think I know what the issue is. We start a transient unit under systemd even when we do `runc update` -- so systemd clears all of the state. We'll need to do something different if the container is already started (does systemd support an "update" to a transient unit).

This is a little scary because during the race window, systemd actually switches the devices cgroup **to be allow-all** (since that's the default). Oops.

cyphar:
Ah, I think I know what the issue is. We start a transient unit under systemd even when we do `runc update` -- so systemd clears all of the state. We'll need to do something different if the container is already started. In particular, we need to use `SetUnitProperties` if the container is already running.

This is quite scary because during the race window, systemd actually switches the devices cgroup **to be allow-all** (since that's the default). Which means that a container could theoretically get full access to any block or character device on the system. Oops. Yet another reason to not use the systemd cgroup manager I guess :wink:.

cyphar:
I just implemented the above, but it turns out that systemd has the same bug outlined above (if you tell it to reset `DevicesAllow` and then give it a new allow list, it will always write a deny-all blacklist first). So we'll just need to disable this test for systemd cgroups, since this behaviour is coming from systemd.

cyphar:
Since this PR solves a few actual security bugs, I might request a CVE. The systemd cgroup one is probably worth a CVE on its own because it means containers could do all sorts of worrying attacks...

cyphar:
Ah, of course. `DevicesAllow` doesn't allow wildcards for `/dev/{char,block}`. This is going to be quite ugly...

AkihiroSuda:
Can we use freezer to mitigate the systemd race?

cyphar:
"I used the cgroups to manage the cgroups."

Yeah, that's not a bad idea. Working on it.

cyphar:
Hurray, all green! :green_heart:

libcontainer/cgroups/systemd/v1.go
```diff
@@ -394,6 +418,13 @@ func (m *legacyManager) Set(container *configs.Config) error {
 		return err
 	}
 
+	// Unfreeze the cgroup.
```
AkihiroSuda, line 421: Seems covered in `defer m.Freeze(configs.Thawed)`

tests/integration/update.bats
AkihiroSuda: Please add code comments for the race?

libcontainer/cgroups/systemd/v1.go
cyphar: I wanted to make sure we unfroze the cgroup before we ran the fsManager (to limit the length the time the container remains frozen, as well as to avoid stepping on our toes with the fsManager also writing to the freezer cgroup).

However, thinking about this more the `defer` is incorrect (as is using `configs.Thawed` directly) because if you're dealing with a user-frozen container (or the user requested we create a frozen container) then we won't do what they're expecting. I'll fix that now.

tests/integration/update.bats
```diff
@@ -312,3 +312,66 @@ EOF
     check_cgroup_value "cpu.rt_period_us" 900001
     check_cgroup_value "cpu.rt_runtime_us" 600001
 }
+
+@test "update devices [minimal transition rules]" {
+    [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
+	# This test currently only makes sense on cgroupv1.
+    requires cgroups_v1
```
AkihiroSuda, line 319: indentation is broken

(off-topic: we should have `shfmt` in CI)

tests/integration/update.bats
```diff
@@ -312,3 +312,66 @@ EOF
...
+	# Kill recvtty.
+	kill -9 "$(<"$TMP_RECVTTY_PID")"
+
+	# There should've been no output from the container.
+	cat "$CONTAINER_OUTPUT"
+	[ -z "$(<"$CONTAINER_OUTPUT")" ]
+}
+
+@test "update paused container" {
+    [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
```
AkihiroSuda, line 360: indentation

AkihiroSuda:
LGTM

shfmt issue can be handled in follow-up PR

cyphar:
Hang on, I'm still working on fixing the tests -- the newest one is breaking (the auto-detection for the freezer cgroup isn't working). Should only take a moment to fix...

cyphar:
Alright, this is ready for review. Tests are implemented for the changes, and I enabled the `runc pause` tests for rootless containers which was missed some time ago.

kolyshkin:
LGTM

cyphar:
Rebased, now that the `GetPaths` changes have been merged.

AkihiroSuda:
LGTM

tests/integration/helpers.bash
```diff
@@ -200,75 +207,85 @@ function fail() {
...
+				skip=1
 			fi
 			;;
 		*)
-			fail "BUG: Invalid requires ${var}."
+			fail "BUG: Invalid requires $var."
 			;;
 		esac
+		if [ -n "$skip" ]; then
+			skip "test requires $var"
```
AkihiroSuda, line 287: nit: Can we avoid using the same identifier for the variable and the function. (Surprised to know they don't conflict)

cyphar: Sure. The reason they don't conflict is that variables and functions are in different namespaces.

cyphar: Done.

tests/integration/helpers.bash
```diff
@@ -200,75 +207,85 @@ function fail() {
 # support it, the test is skipped with a message.
 function requires() {
 	for var in "$@"; do
+		SKIP_ME=
```
AkihiroSuda, line 210: nit again: `local skip_me` (in lower characters)

AkihiroSuda:
LGTM

cyphar:
I've rebased and cleaned up the commit message for the `DeviceAllow` commit to better clarify that it was a security issue but since no released version of runc included it, it didn't deserve a proper security advisory.

kolyshkin:
LGTM

AkihiroSuda:
LGTM

tao12345666333:
When I try to vendor runc library v1.0.0-rc91 into Docker/Moby,  I found an unexpected behavior: https://github.com/moby/moby/pull/41178#issuecomment-663813218

After debugging, I found that the problem was introduced since this PR.  PTAL, Thanks!  https://github.com/moby/moby/pull/41178

AkihiroSuda:
Could you open a new issue?

tao12345666333:
sure.

PR #2393, opened 2020-05-09
Vagrantfile: use criu from stable repo

kolyshkin:
CRIU 3.14 has made its way to the F32 stable repo, let's use it.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,17 +13,14 @@ Vagrant.configure("2") do |config|
     v.cpus = 2
   end
   config.vm.provision "shell", inline: <<-SHELL
-    curl -OSs https://kojipkgs.fedoraproject.org/packages/criu/3.14/1.fc32/x86_64/criu-3.14-1.fc32.x86_64.rpm
     cat << EOF | dnf -y shell
 config exclude kernel,kernel-core
 config install_weak_deps false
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
LGTM

Vagrantfile
```diff
@@ -13,17 +13,14 @@ Vagrant.configure("2") do |config|
     v.cpus = 2
   end
   config.vm.provision "shell", inline: <<-SHELL
-    curl -OSs https://kojipkgs.fedoraproject.org/packages/criu/3.14/1.fc32/x86_64/criu-3.14-1.fc32.x86_64.rpm
     cat << EOF | dnf -y shell
 config exclude kernel,kernel-core
 config install_weak_deps false
-localinstall criu-3.14-1.fc32.x86_64.rpm
 update
-install iptables gcc make golang-go libseccomp-devel bats jq git-core
```
AkihiroSuda, line 22: Can we pin the version?

kolyshkin: > Can we pin the version?

Yes (`criu = 3.14` or something like that) but I don't see the reason why. If the next criu release fails, we want to know about it (and can certainly use a previous release once we do)

AkihiroSuda:
@mrunalp @cyphar PTAL

mrunalp:
LGTM

Issue #2367, opened 2020-04-30
cgroup2: Revert "CreateCgroupPath: only enable needed controllers"


AkihiroSuda:
`CreateCgroupPath: only enable needed controllers` (4b4bc995ad4ee4679ffbe38f1672be01c24256fc) seems problematic toward supporting nested containers. (#2356)

As many controller s as possible should be always enabled.

cc @kolyshkin

AkihiroSuda:
@kolyshkin WDYT?

kolyshkin:
Enabling a controller should negatively affect performance (since some accounting for the resource(s) controlled are to be done by the kernel) -- even in the case no limits are set.

So, this is a trade-off between best performance and versatility. Ideally, I'd like it to be configurable (i.e. there should be a way to just enable a controller but do not set any limits, which is currently impossible).

Until we have this ability, either way is fine with me (all controllers or on-demand). Since we used to have all controllers, let's do that.

AkihiroSuda:
resolved in https://github.com/opencontainers/runc/pull/2395

PR #2395, opened 2020-05-10
Partially revert "CreateCgroupPath: only enable needed controllers"

### LINKED ISSUE #2394: `runc update` broken in cgroup v2
If we update a resource which did not limited in the beginning, it will
have no effective.
For example:
We use default `config.json` to create a container, and then run:
```
runc update --cpu-quota 1 test
runc update --cpuset-cpus 1 test
```
It has no error input, but actually it has no effective.

lifubang:
fix #2394 

1. Partially revert "CreateCgroupPath: only enable needed controllers"
Because if we update a resource which did not limited in the beginning, it will
have no effective.
2. Returns err if we use an non enabled controller, or else the user may feel success, but actually
there are no effective.

For #2367 , I think we don't need fully revert https://github.com/opencontainers/runc/commit/4b4bc995ad4ee4679ffbe38f1672be01c24256fc , because rootless need some part in it to check errors.

### DIFF
--- a/libcontainer/cgroups/fs2/create.go
+++ b/libcontainer/cgroups/fs2/create.go
@@ -1,6 +1,7 @@
 package fs2
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -10,57 +11,56 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
-// neededControllers returns the string to write to cgroup.subtree_control,
-// containing the list of controllers to enable (for example, "+cpu +pids"),
-// based on (1) controllers available and (2) resources that are being set.
-//
-// The resulting string does not include "pseudo" controllers such as
-// "freezer" and "devices".
-func neededControllers(cgroup *configs.Cgroup) ([]string, error) {
-	var list []string
+func supportedControllers(cgroup *configs.Cgroup) ([]byte, error) {
+	const file = UnifiedMountpoint + "/cgroup.controllers"
+	return ioutil.ReadFile(file)
+}
 
+// needAnyControllers returns whether we enable some supported controllers or not,
+// based on (1) controllers available and (2) resources that are being set.
+// We don't check "pseudo" controllers such as
+// "freezer" and "devices".
+func needAnyControllers(cgroup *configs.Cgroup) (bool, error) {
 	if cgroup == nil {
-		return list, nil
+		return false, nil
 	}
 
 	// list of all available controllers
-	const file = UnifiedMountpoint + "/cgroup.controllers"
-	content, err := ioutil.ReadFile(file)
+	content, err := supportedControllers(cgroup)
 	if err != nil {
-		return list, err
+		return false, err
 	}
 	avail := make(map[string]struct{})
 	for _, ctr := range strings.Fields(string(content)) {
 		avail[ctr] = struct{}{}
 	}
 
-	// add the controller if available
-	add := func(controller string) {
-		if _, ok := avail[controller]; ok {
-			list = append(list, "+"+controller)
-		}
+	// check whether the controller if available or not
+	have := func(controller string) bool {
+		_, ok := avail[controller]
+		return ok
 	}
 
-	if isPidsSet(cgroup) {
-		add("pids")
+	if isPidsSet(cgroup) && have("pids") {
+		return true, nil
 	}
-	if isMemorySet(cgroup) {
-		add("memory")
+	if isMemorySet(cgroup) && have("memory") {
+		return true, nil
 	}
-	if isIoSet(cgroup) {
-		add("io")
+	if isIoSet(cgroup) && have("io") {
+		return true, nil
 	}
-	if isCpuSet(cgroup) {
-		add("cpu")
+	if isCpuSet(cgroup) && have("cpu") {
+		return true, nil
 	}
-	if isCpusetSet(cgroup) {
-		add("cpuset")
+	if isCpusetSet(cgroup) && have("cpuset") {
+		return true, nil
 	}
-	if isHugeTlbSet(cgroup) {
-		add("hugetlb")
+	if isHugeTlbSet(cgroup) && have("hugetlb") {
+		return true, nil
 	}
 
-	return list, nil
+	return false, nil
 }
 
 // containsDomainController returns whether the current config contains domain controller or not.
@@ -70,18 +70,19 @@ func containsDomainController(cg *configs.Cgroup) bool {
 	return isMemorySet(cg) || isIoSet(cg) || isCpuSet(cg) || isHugeTlbSet(cg)
 }
 
-// CreateCgroupPath creates cgroupv2 path, enabling all the
-// needed controllers in the process.
+// CreateCgroupPath creates cgroupv2 path, enabling all the supported controllers.
 func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 	if !strings.HasPrefix(path, UnifiedMountpoint) {
 		return fmt.Errorf("invalid cgroup path %s", path)
 	}
 
-	ctrs, err := neededControllers(c)
+	content, err := supportedControllers(c)
 	if err != nil {
 		return err
 	}
-	allCtrs := strings.Join(ctrs, " ")
+
+	ctrs := bytes.Fields(content)
+	res := append([]byte("+"), bytes.Join(ctrs, []byte(" +"))...)
 
 	elements := strings.Split(path, "/")
 	elements = elements[3:]
@@ -131,13 +132,14 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 				}
 			}
 		}
-		// enable needed controllers
+		// enable all supported controllers
 		if i < len(elements)-1 {
 			file := filepath.Join(current, "cgroup.subtree_control")
-			if err := ioutil.WriteFile(file, []byte(allCtrs), 0644); err != nil {
+			if err := ioutil.WriteFile(file, res, 0644); err != nil {
 				// try write one by one
-				for _, ctr := range ctrs {
-					_ = ioutil.WriteFile(file, []byte(ctr), 0644)
+				allCtrs := bytes.Split(res, []byte(" "))
+				for _, ctr := range allCtrs {
+					_ = ioutil.WriteFile(file, ctr, 0644)
 				}
 			}
 			// Some controllers might not be enabled when rootless or containerized,
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -76,8 +76,7 @@ func (m *manager) Apply(pid int) error {
 		// - "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error"
 		if m.rootless {
 			if m.config.Path == "" {
-				cl, clErr := neededControllers(m.config)
-				if clErr == nil && len(cl) == 0 {
+				if blNeed, nErr := needAnyControllers(m.config); nErr == nil && !blNeed {
 					return nil
 				}
 				return errors.Wrap(err, "rootless needs no limits + no cgrouppath when no permission is granted for cgroups")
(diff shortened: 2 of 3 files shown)


libcontainer/cgroups/fs2/fs2.go
```diff
@@ -189,57 +189,41 @@ func (m *manager) Set(container *configs.Config) error {
 	if err := m.getControllers(); err != nil {
 		return err
 	}
-	var errs []error
```
AkihiroSuda, line 192: why?

lifubang: I think if there is an error, we should return without try other controllers.

Do you mean we should let enabled controllers success when we update more than 1 controller in one command, for example:
```
runc update --cpuset-cpus 1 --memory 5m test
```
We should let memory limit set success?

lifubang: In v1, it doesn't retry any other subsystems when got an error in one subsystem.

lifubang:
CI error:
```
# time="2020-05-10T02:06:34Z" level=error msg="container_linux.go:348: starting container process caused: process_linux.go:438: container init caused: process_linux.go:404: setting cgroup config for procHooks process caused: failed to write \"33554432\" to \"/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/memory.max\": open /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/memory.max: permission denied"
```
I think this error is ignored in the past.
So, I think we need #2390 to fix this error.

libcontainer/cgroups/fs2/memory.go
```diff
@@ -50,8 +50,10 @@ func setMemory(dirPath string, cgroup *configs.Cgroup) error {
 		// memory and memorySwap set to the same value -- disable swap
 		swapStr = "0"
 	}
-	if err := fscommon.WriteFile(dirPath, "memory.swap.max", swapStr); err != nil {
-		return err
+	if swapStr != "" {
```
lifubang, line 53: CI error:
```
# time="2020-05-10T01:05:29Z" level=error msg="container_linux.go:348: starting container process caused: process_linux.go:438: container init caused: process_linux.go:404: setting cgroup config for procHooks process caused: failed to write \"\" to \"/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/memory.swap.max\": open /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/memory.swap.max: permission denied"
```
This error is introduced by #2370 . So we need to fix it in here.

lifubang:
Waiting for #2390 complete.

libcontainer/cgroups/fs2/create.go
```diff
@@ -10,7 +11,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
-// neededControllers returns the string to write to cgroup.subtree_control,
+// neededControllers returns the minimal string that should be write to cgroup.subtree_control,
```
kolyshkin, line 14: `string to write to` was OK

if you want `should be`, it should be `written`.

libcontainer/cgroups/fs2/memory.go
kolyshkin: This is a separate fix and it requires a separate patch.

Also, why haven't we caught this before?

libcontainer/cgroups/fs2/create.go
kolyshkin: if you're not using neededControllers, why keep it?

lifubang: Rootless needs it.
https://github.com/opencontainers/runc/blob/744bbe8128754ecb30a18465281ec3e02ee0766f/libcontainer/cgroups/fs2/fs2.go#L77-L85
How about move it to fs2.go and change the implementation way.

kolyshkin: Yes please.

kolyshkin:
@lifubang #2390 is now merged

lifubang:
How about modify this check:
https://github.com/opencontainers/runc/blob/4e2499e8de399f0cf281825d2e651d14ba58b21c/tests/integration/update.bats#L84-L87
Chang to check whether men controller is enabled or not in /sys/fs/cgroup/cgroups.controller?

libcontainer/cgroups/fs2/fs2.go
kolyshkin: make sense.

libcontainer/cgroups/fs2/create.go
```diff
@@ -10,24 +11,27 @@ import (
...
-// containing the list of controllers to enable (for example, "+cpu +pids"),
+func supportedControllers(cgroup *configs.Cgroup) ([]byte, error) {
+	const file = UnifiedMountpoint + "/cgroup.controllers"
+	return ioutil.ReadFile(file)
+}
+
+// neededControllers returns the number of controllers to enable,
 // based on (1) controllers available and (2) resources that are being set.
 //
 // The resulting string does not include "pseudo" controllers such as
```
kolyshkin, line 22: we don't have a string anymore so the comment needs to be fixed

libcontainer/cgroups/fs2/create.go
```diff
@@ -10,24 +11,27 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
-// neededControllers returns the string to write to cgroup.subtree_control,
-// containing the list of controllers to enable (for example, "+cpu +pids"),
+func supportedControllers(cgroup *configs.Cgroup) ([]byte, error) {
```
kolyshkin, line 14: I suggest to return `bool` here as this is what the caller needs.

libcontainer/cgroups/fs2/create.go
```diff
@@ -10,24 +11,27 @@ import (
...
+}
+
+// neededControllers returns the number of controllers to enable,
 // based on (1) controllers available and (2) resources that are being set.
 //
 // The resulting string does not include "pseudo" controllers such as
 // "freezer" and "devices".
-func neededControllers(cgroup *configs.Cgroup) ([]string, error) {
-	var list []string
+func neededControllers(cgroup *configs.Cgroup) (int, error) {
```
kolyshkin, line 24: Maybe return `bool` here as no one is interested in the number of controllers to be enabled, but rather whether any controllers need to be enabled or not.

kolyshkin: You can also optimize things by returning true as soon as there is a controller that needs to be enabled.

Something like

```go
￼	// 
￼	have := func(controller string) bool {
￼		if _, ok := avail[controller]; ok {
￼	￼	￼	return true
￼		}
￼		return false
￼	}

...

if isPidsSet(cgroup) && have("pids") {
￼		return true
}
if isMemorySet(cgroup) && have("memory") {
    return true
}
....
```
		add("memory")
	}
		add("pids")

kolyshkin: And maybe rename to `needAnyControllers()` or smth similar

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -175,57 +175,41 @@ func (m *manager) Set(container *configs.Config) error {
...
 	// pids (since kernel 4.5)
-	if _, ok := m.controllers["pids"]; ok {
```
kolyshkin, line 180: I believe this function was the only user of `controllers` map. If there aren't any left, remove it altogether.

Maybe as a separate patch, describing why there's no need for this map and check.

libcontainer/cgroups/fs2/memory.go
kolyshkin: Can we have this fix as a separate patch? And I still don't understand why we haven't caught it before.

kolyshkin:
Left a number of comments. It's better to split this into a few smaller patches as there are a few independent things being fixed.

lifubang:
> It's better to split this into a few smaller patches as there are a few independent things being fixed.

I quite agree with you. And the purpose of this patch is to let `runc update` work.
After this fix, you can do more refactoring work.

libcontainer/cgroups/fs2/memory.go
lifubang: We don’t check memorySwap==0

lifubang: @kolyshkin Do we still need to submit a separate patch?

libcontainer/cgroups/fs2/fs2.go
lifubang: agree with you

libcontainer/cgroups/fs2/memory.go
lifubang: The seprate patch: #2405

libcontainer/cgroups/fs2/fs2.go
lifubang: Oh, there is another user: `GetStats`!

libcontainer/cgroups/fs2/memory.go
lifubang: This part is fixed by #2410

lifubang:
As #2390 and #2410 has been merged, so set this PR ready for review.

libcontainer/cgroups/fs2/create.go
```diff
@@ -134,10 +138,11 @@ func CreateCgroupPath(path string, c *configs.Cgroup) (Err error) {
 		// enable needed controllers
```
AkihiroSuda, line 138: this comment is no longer correct?

AkihiroSuda:
LGTM, can we also have tests?

lifubang:
> LGTM, can we also have tests?

Yes, I think we should add some test cases. I'll add it later.

tests/integration/cgroups.bats
```diff
@@ -71,6 +72,24 @@ EOF
...
+    # run a detached busybox to work with
+    runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_mem
+    [ "$status" -eq 0 ]
+
+    # update memory limit
+    runc update test_cgroups_mem --memory 50331648
+
+    # check value of memory.max
+    [ "$status" -eq 0 ]
+    check_cgroup_value "memory.max" 50331648
```
AkihiroSuda, line 90: Just compare `/sys/fs/cgroup/cgroup.controllers`.
On rootless systemd, the content should correspond to `/sys/fs/cgroup/user.slice/user-${uid}.slice/cgroup.controllers`

tests/integration/cgroups.bats
```diff
@@ -110,6 +110,9 @@ EOF
 
     runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_permissions
     [ "$status" -eq 0 ]
+    if [ "$CGROUP_UNIFIED" != "no" ]; then
+        check_cgroup_value "cgroup.controllers" "$(cat /sys/fs/cgroup/cgroup.controllers)"
```
AkihiroSuda, line 114: should be compared against `/sys/fs/cgroup/user.slice/user-${uid}.slice/cgroup.controllers` when rootless && `RUNC_USE_SYSTEMD` is set

libcontainer/cgroups/fs2/create.go
AkihiroSuda: ping @lifubang

lifubang: changed now.

AkihiroSuda:
LGTM (if green)

libcontainer/cgroups/fs2/create.go
```diff
@@ -10,57 +11,58 @@ import (
...
 	}
 
-	// add the controller if available
-	add := func(controller string) {
+	// check whether the controller if available or not
+	have := func(controller string) bool {
 		if _, ok := avail[controller]; ok {
-			list = append(list, "+"+controller)
+			return true
 		}
```
kolyshkin, line 42: nit
```go
_, ok := avail[controller]
return ok
```

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -76,8 +76,8 @@ func (m *manager) Apply(pid int) error {
 		// - "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error"
 		if m.rootless {
 			if m.config.Path == "" {
-				cl, clErr := neededControllers(m.config)
-				if clErr == nil && len(cl) == 0 {
+				cl, clErr := needAnyControllers(m.config)
```
kolyshkin, line 79: nit: `cl` used to mean "controller list". Now it's not so maybe rename to `need` or something.

ditto for `clErr`.

tests/integration/cgroups.bats
```diff
@@ -110,6 +110,17 @@ EOF
 
     runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_permissions
     [ "$status" -eq 0 ]
+    if [ "$CGROUP_UNIFIED" != "no" ]; then
+        if [ -n "${RUNC_USE_SYSTEMD}" ] ; then
```
kolyshkin, line 114: nit:
```go
if [ "$CGROUP_UNIFIED" != "no" -a -n "${RUNC_USE_SYSTEMD}" ]; then
```

lifubang: I think no, because there is `else` for `if [ -n "${RUNC_USE_SYSTEMD}" ] ; then`.

kolyshkin: sorry, missed it

kolyshkin:
LGTM

@mrunalp @AkihiroSuda PTAL

AkihiroSuda:
LGTM

PR #2401, opened 2020-05-14
libct/cgroup: rm GetClosestMountpointAncestor using moby/sys/mountinfo parser

kolyshkin:
The function `GetClosestMountpointAncestor` is not very efficient,
does not really belong to cgroup package, and is only used once
(from fs/cpuset.go).

The removal is done in two stages/commits:

### 1

Prepare to remove it by replacing with the implementation based on
the `github.com/moby/sys/mountinfo` parser.

This commit is here to make sure the proposed replacement passes the
unit test.

Funny, but the unit test need to be slightly modified since it
supplies the wrong mountinfo (space as the first character, empty line
at the end).

Validated by
```
 $ go test -v -run Ance
 === RUN   TestGetClosestMountpointAncestor
 --- PASS: TestGetClosestMountpointAncestor (0.00s)
 PASS
 ok  	github.com/opencontainers/runc/libcontainer/cgroups	0.002s
```

### 2

Actually remove `GetClosestMountpointAncestor`, move the replacement
code to `libcontainer/cgroups/fs/cpuset.go`.

### DIFF
--- a/libcontainer/cgroups/fs/cpuset.go
+++ b/libcontainer/cgroups/fs/cpuset.go
@@ -62,7 +62,11 @@ func (s *CpusetGroup) ApplyDir(dir string, cgroup *configs.Cgroup, pid int) erro
 	if err != nil {
 		return err
 	}
-	root := filepath.Dir(cgroups.GetClosestMountpointAncestor(dir, string(mountInfo)))
+	root, err := cgroups.GetClosestMountpointAncestor(dir, string(mountInfo))
+	if err != nil {
+		return err
+	}
+	root = filepath.Dir(root)
 	// 'ensureParent' start with parent because we don't want to
 	// explicitly inherit from parent, it could conflict with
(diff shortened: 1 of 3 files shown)


kolyshkin:
OK, CI passed, adding the second commit

kolyshkin:
I quickly checked Docker and libpod sources -- this function is not used in there.

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda:
LGTM, but I guess we may eventually move back the function to libct/cg .

kolyshkin:
> I guess we may eventually move back the function to libct/cg

There is no need to mess with mountinfo for cgroup v2 case, and I don't think we'll make any major changes to cgroup v1 (except I'm still tinkering with how to decrease the ridiculous number of times we parse /proc/self/mountinfo).

mrunalp:
LGTM

PR #2402, opened 2020-05-14
Add integration tests for Oci hooks: CreateRuntime, CreateContainer and StartContainer

RenaudWasTaken:
Hello!

This PR is a followup of https://github.com/opencontainers/runc/pull/2229
It is best reviewed by looking at the following single commit: https://github.com/opencontainers/runc/pull/2402/commits/4ec35109a304df3c197a2816ec00f03e8a756fa8
All other commits are based of the above mentioned PR.

With these integration test, I wanted to check real world scenarios, that we depend on in libnvidia-container, the two that I had in mind were the following:
1) mount a library, run ldconfig and check that it picked up these library
2) mount a device, update the cgroups and check that the container has access to that device

Unfortunately the first use case requires a more complete container than busybox, which doesn't have ldconfig, or the hello container. I ended up installing skopeo and umoci for that.

Note that at first I wanted to use alpine to keep the image small, but alpine has it's own implementation of ldconfig (10 line shell script) which doesn't support printing the libraries it has in its cache.

Let me know if you think there is an easier path than adding a new image :)!
With this last PR (and the previous merged), I think we should have unblocked runc 1.0.0!

/assign @cyphar @AkihiroSuda @mikebrow 
/cc @mrunalp @h-vetinari

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,6 +71,24 @@ RUN mkdir -p /usr/src/criu \
     && cd - \
     && rm -rf /usr/src/criu
 
+# install skopeo
+RUN echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list \
+    && wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_Unstable/Release.key -O- | sudo apt-key add - \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends skopeo \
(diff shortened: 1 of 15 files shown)


kolyshkin:
I think the feature and its tests should be in a single PR (unless the feature is already merged, of course).

So maybe you can just add the commit 4ec3510 to #2229, update its description and close this one?

AkihiroSuda:
As Kir commented, please add these tests to #2229

cyphar:
Yeah, tests should be added to the same PR as the feature (if nothing else, to show that the code works). Reviewing tests is much simpler than code changes because if the tests pass and they look like they actually test the feature, then they're fine.

PR #2405, opened 2020-05-14
fix two swap limit regressions in cgroup v2

lifubang:
(1) never write empty string to memory.swap.max
Because the empty string means set swap to 0.

fixed by #2410 

For example:
with memory limit:
```
"memory": {
        "limit": 33554432,
        "swap": 33558528
 }
```
The value of `memory.swap.max` is 4096.
After we run `runc update --memory 33554432 test`, the value of `memory.swap.max` becomes 0.

(2) If the memory update is set to -1 we should also set swap to -1, it means unlimited memory.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -627,12 +627,17 @@ func ConvertCPUQuotaCPUPeriodToCgroupV2Value(quota int64, period uint64) string 
 // for use by cgroup v2 drivers. A conversion is needed since Resources.MemorySwap
 // is defined as memory+swap combined, while in cgroup v2 swap is a separate value.
 func ConvertMemorySwapToCgroupV2Value(memorySwap, memory int64) (int64, error) {
+	// If the memory update is set to -1 we should also
+	// set swap to -1, it means unlimited memory.
+	if memory == -1 {
+		return -1, nil
+	}
 	if memorySwap == -1 || memorySwap == 0 {
 		// -1 is "max", 0 is "unset", so treat as is
 		return memorySwap, nil
 	}
 	// sanity checks
-	if memory == 0 || memory == -1 {
+	if memory == 0 {
 		return 0, errors.New("unable to set swap limit without memory limit")
 	}
 	if memory < 0 {
--- a/libcontainer/cgroups/fs2/memory.go
+++ b/libcontainer/cgroups/fs2/memory.go
@@ -50,8 +50,11 @@ func setMemory(dirPath string, cgroup *configs.Cgroup) error {
 		// memory and memorySwap set to the same value -- disable swap
 		swapStr = "0"
 	}
-	if err := fscommon.WriteFile(dirPath, "memory.swap.max", swapStr); err != nil {
-		return err
+	// never write empty string to `memory.swap.max`, it means set to 0.
+	if swapStr != "" {
+		if err := fscommon.WriteFile(dirPath, "memory.swap.max", swapStr); err != nil {
+			return err
+		}
 	}
 
 	if val := numToStr(cgroup.Resources.Memory); val != "" {
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -588,9 +588,9 @@ func TestConvertMemorySwapToCgroupV2Value(t *testing.T) {
 			expErr:  true,
 		},
 		{
-			memswap: 300,
-			memory:  -1,
-			expErr:  true,
+			memswap:  300,
+			memory:   -1,
+			expected: -1,
 		},
 	}
 
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -75,7 +75,7 @@ EOF
         SYSTEM_MEM="max"
         SYSTEM_MEM_SWAP="max"
         # checking swap is currently disabled for v2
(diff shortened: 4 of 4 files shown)


libcontainer/cgroups/utils.go
```diff
@@ -627,12 +627,17 @@ func ConvertCPUQuotaCPUPeriodToCgroupV2Value(quota int64, period uint64) string
 // for use by cgroup v2 drivers. A conversion is needed since Resources.MemorySwap
 // is defined as memory+swap combined, while in cgroup v2 swap is a separate value.
 func ConvertMemorySwapToCgroupV2Value(memorySwap, memory int64) (int64, error) {
+	// If the memory update is set to -1 we should also
+	// set swap to -1, it means unlimited memory.
+	if memory == -1 {
```
kolyshkin, line 632: So, this basically means that if a user sets `memory: -1, swap: N`, the value of N for mem+swap is discarded, and the swap is set to unlimited. I don't think it's the right thing to do.

I also disagree that "unlimited memory" means "unlimited swap" as well. A configuration with no limit for RAM but some limit for swap is maybe unusual but not incorrect. Alas, we don't have a way to set a separate value for swap, when memory is set to -1, unless we set memory and swap to some values and then set memory to unlimited and don't touch swap.

I think the existing behavior (before this patch) is correct.

kolyshkin:
> (2) If the memory update is set to -1 we should also set swap to -1, it means unlimited memory.

I don't think so (see above)

libcontainer/cgroups/utils.go
lifubang: I copied it from v1:
https://github.com/opencontainers/runc/blob/6bb6d0b8943ad0aad57ba2575098b23078816246/libcontainer/cgroups/fs/memory.go#L76-L84

I think v2 should keep compatible with v1.



> So, this basically means that if a user sets `memory: -1, swap: N`, the value of N for mem+swap is silently discarded, and the swap is instead set to unlimited. I don't think it's the right thing to do.

Before this patch, it will print an error `unable to set swap limit without memory limit`.

cyphar: The behaviour of the v1 memory controller is different from v2. In that in v1 "MemorySwap" means memory+swap, while in v2 `MemorySwap` actually refers to just the amount of swap used. It therefore makes perfect sense that the behaviour for the configuration will be different -- in fact in v1 it's not possible to block a cgroup from using swap (unless you set `swappiness=0` in the cgroup and even then I'm not sure).

I think @kolyshkin is right.

cyphar: The issue is that the behaviour of the two cgroup versions is different -- but the meaning of `MemorySwap` in the runtime-spec is actually the original cgroupv1 meaning of `memory.memsw.limit_in_bytes` (which is the total memory + swap, as opposed to `memory.swap.max` in cgroupv2 which is just swap usage). If we'd done runtime-spec a little differently this wouldn't be such a headache. :wink:

However, the problem is that requesting `{Memory: -1, MemorySwap: N}` is an illogical request -- by the definition of the runtime-spec you're asking for unlimited memory but limited memory+swap. In cgroupv1 there wasn't a way to just limit swap usage, so for some reason the runc implementation ignores that this request is invalid and sets both options to be unlimited. I personally think giving an error is much more reasonable (and that we should see if changing cgroupv1 to give an error will cause issues with Docker or podman).

I agree with @lifubang that we should have the same behaviour (as much as we can) when we are converting cgroupv1 options to cgroupv2 settings, because not doing that will cause a lot more headaches in the future.

lifubang:
How about split this PR into 2 parts? Because #2395 needs (1).

libcontainer/cgroups/utils.go
kolyshkin: > by the definition of the runtime-spec you're asking for unlimited memory but limited memory+swap.

So, there are actually three ways to handle such a request for a cgroupv2 controller:

1. Return an error (current behavior).
2. Set memory and swap both to unlimited (proposed behavior).
3. Set memory and swap both to N (which just occurred to me).

I think either 1 or 3 is more reasonable than option 2, despite cgroupv1 controller doing 2.

kolyshkin:
> How about split this PR into 2 parts? Because #2395 needs (1).

Sure, (1) can be merged right away

libcontainer/cgroups/utils.go
lifubang: > I agree with @lifubang that we should have the same behaviour (as much as we can)

Yes, thanks.
I think at least the exit code of a command for cgroupv2 should be same with cgroupv1, but the value in `mem.swap.max` can be different.
So, for @kolyshkin 's last comment, I think we should not return an error for this case.
I have another thought, for the case `memory: -1, swap: N(N!=-1)`, we only set `mem.max` to `max`, but don't touch `mem.swap.max`.

kolyshkin: I traced the cgroupv1 behavior to PR #1127 which have a pretty long discussion about this case.

It's funny but I was trying to fix the exact issue we're discussing here in #2285 but got sidetracked by the fact that cgroupv2 handles swap as a separate limit.

Anyway, I read it all and I think this is what needs to be done, PTAL https://github.com/opencontainers/runc/pull/2426

kolyshkin:
I propose to close this one in favor of #2426

PR #2409, opened 2020-05-15
Update to latest go-criu

adrianreber:
This updates to the latest version of go-criu (4.0.0) which is based on CRIU 3.14.

As go-criu provides an existing way to query the CRIU binary for its version this also removes all the code from runc to handle CRIU version checking and now relies on go-criu.

An important side effect of this change is that this raises the minimum CRIU version to 3.0.0 as that is the first CRIU version that supports CRIU version queries via RPC in contrast to parsing the output of 'criu --version'

CRIU 3.0 has been released in April of 2017.

The big decision here is if we want to limit runc to CRIU >= 3.0. I guess older versions are not really working with runc anymore so that I do not see it as big problem.

I will update this PR to point to the 4.0.0 go-criu release once go-criu is tagged.

CC: @kolyshkin


go.sum
```diff
@@ -1,6 +1,10 @@
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/checkpoint-restore/go-criu v0.0.0-20191125063657-fcdcd07065c5 h1:950diauOSoCNaQfvLE0WaFadyI/ilKIjb6jEUQnm+hE=
 github.com/checkpoint-restore/go-criu v0.0.0-20191125063657-fcdcd07065c5/go.mod h1:TrMrLQfeENAPYPRsJuq3jsqdlRh3lvi6trTZJG8+tho=
+github.com/checkpoint-restore/go-criu v0.0.0-20200515115155-5fe4432ec9c4 h1:HAWVG+00NF6SNU9g16Epd/46Iq6QLBWWiH+SoegMtfU=
+github.com/checkpoint-restore/go-criu v0.0.0-20200515115155-5fe4432ec9c4/go.mod h1:TrMrLQfeENAPYPRsJuq3jsqdlRh3lvi6trTZJG8+tho=
```
kolyshkin, line 5: AFAICS these lines should not be there (and yeah I know that `go mod` is tough).

Maybe try running `go mod tidy`?

kolyshkin:
nit: maybe add "require criu >= 3.0" in the commit subject

kolyshkin:
LGTM

go.sum
kolyshkin: @adrianreber PTAL ^^^

adrianreber: I removed it manually as `go mod tidy` removed a couple more lines.

kolyshkin:
LGTM

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda:
LGTM

PR #2410, opened 2020-05-15
cgroupv2: never write empty string to memory.swap.max

lifubang:
Never write empty string to memory.swap.max
Because the empty string means set swap to 0.

For example:
with memory limit:
```
"memory": {
        "limit": 33554432,
        "swap": 33558528
 }
```
The value of `memory.swap.max` is 4096.
After we run `runc update --memory 33554432 test`, the value of `memory.swap.max` becomes 0.
This is a regression introduced by #2370 . I'm so sorry.


kolyshkin:
LGTM

kolyshkin:
@AkihiroSuda @mrunalp PTAL

mrunalp:
LGTM

PR #2387, opened 2020-05-08
cgroup refactoring

kolyshkin:
This is a set of patches aiming to prepare the codebase for #2386 

Please review commits one by one.

PR #2411, opened 2020-05-15
libct/cgroups/utils: fix/separate cgroupv1 code

kolyshkin:
~~_NOTE: this is based on and currently includes https://github.com/opencontainers/runc/pull/2401. I'll rebase once that one is merged._~~ _Rebased_

This tries to untangle the mess of code in `libcontainers/cgroups/utils.go`. There are multiple functions that do same or similar things, with multiple names, and it's not clear which ones should be used for cgroup v1, cgroup v2, or both.

Please review commit by commit.

Besides moving the code around, this PR reverts some attempts to make the functions work equally well for v1 and v2. The truth is, v2 is a very different beast and most of the functions that were v1-specific should stand that way.

The only real fix here is commit 775a5cbc119fa85573b6b7d36f9611a0f0e9138a:

> libct/criuApplyCgroups: don't set cgroup paths for v2
> 
> There is no need to have cgroupv1-specific controller paths on restore
> in case of cgroupv2.

I think the above was the last use of unholy multiple v1 paths in v2 code. :tada:


libcontainer/cgroups/v1_utils.go
```diff
@@ -0,0 +1,304 @@
...
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// Functions in this source file are specific to cgroupv1,
```
AkihiroSuda, line 12: Can we split pkg

kolyshkin: let me try

kolyshkin: I was able to do that. The problem, though, is all the existing users of runc need to adopt. I'm not sure how disruptive it is.

kolyshkin: oops, forgot the tests :)

AkihiroSuda:
LGTM

kolyshkin:
So, I checked moby/moby and found these cgroupv1-only functions are used from the code:
```
daemon/daemon_unix.go:  mnt, root, err := cgroups.FindCgroupMountpointAndRoot("", "cpu")
daemon/oci_linux.go:                    initPath, err := cgroups.GetInitCgroup("cpu")
daemon/oci_linux.go:                    _, err = cgroups.GetOwnCgroup("cpu")
pkg/sysinfo/sysinfo_linux.go:   cgMounts, err := cgroups.GetCgroupMounts(false)
pkg/sysinfo/sysinfo_linux.go:   _, err := cgroups.FindCgroupMountpoint("", "pids")
```

so this PR will obviously break it.

The reason is, all this stuff is very cgroupv1 specific and it makes no sense to use it when cgroupv2 is used on the host. In case it is still used in such scenario it means the code is doing something wrong and is broken.

 @thaJeztah please note the above (and I will be happy to help fix it in moby code).

kolyshkin:
Hmm, is this a regression caused by any of the recently merged PRs, or by this PR, or by some changes in Travis?

```
time="2020-05-19T22:16:58Z" level=warning msg="signal: killed"
--- FAIL: TestRunWithKernelMemory (0.23s)
   exec_test.go:713: runContainer failed with kernel memory limit: container_linux.go:353: starting container process caused: process_linux.go:437: container init caused: process_linux.go:403: setting cgroup config for procHooks process caused: kernel memory accounting not supported by this kernel
```

The error looks legit though

kolyshkin:
Multiple other errors similar to this one (missing cgroup files), and also 
```
# sh: can't create /dev/null: Operation not permitted
```

kolyshkin:
rebased to trigger CI

kolyshkin:
> @thaJeztah please note the above (and I will be happy to help fix it in moby code).

update: looks like the place which needs fixing is `WithCgroups` in `daemon/oci_linux.go`. Indeed, this code is probably not working right for cgroupv2. Everything else is mostly `s|cgroups|cgroups/cgroupv1|g`.

kolyshkin:
hmm, a single failure from Fedora32 on Vagrant ([logs here](https://travis-ci.org/github/opencontainers/runc/jobs/689070268)):
```
=== RUN   TestExecInTTY
    TestExecInTTY: execin_test.go:351: unexpected carriage-return in output
--- FAIL: TestExecInTTY (0.14s)
```
does not look like to be related. This comes from #1146. Hmm

kolyshkin:
OK I think this one is ready. @opencontainers/runc-maintainers PTAL

kolyshkin:
rebased

AkihiroSuda:
LGTM

kolyshkin:
@thaJeztah PTAL. I looked through the moby/moby code and after https://github.com/moby/moby/pull/41014 and https://github.com/moby/moby/pull/41016 the changes to moby/moby should be trivial.

kolyshkin:
updated the patchset to not move `GetAllSubsystems` out of cgroup package. No other changes.

AkihiroSuda:
LGTM

libcontainer/cgroups/cgroupv1/utils.go
```diff
@@ -0,0 +1,307 @@
...
+
+	return filepath.Join(mnt, relCgroup), nil
+}
+
+// ParseCgroupFile parses the given cgroup file, typically from
+// /proc/<pid>/cgroup, into a map of subgroups to cgroup names.
+//
+// This function is cgroupv1-specific and must not be used from
+// cgroupv2 code.
+func ParseCgroupFile(path string) (map[string]string, error) {
```
giuseppe, line 254: why is this function only for cgroup v1?  It can be used on cgroup v2 as well

kolyshkin: So, what it returns is a map of controllers to their paths, e.g.
```go
"pids" => "/user.slice/user-1000.slice"
"freezer" => "/"
"net_cls" => "/"
```
and so on. It is used to get the per-controller path. A caller has certain expectations (like the map keys being the controllers available).

For cgroupv2 (unified hierarchy), there is only one path, something like
```go
"" => "/user.slice/user-1000.slice"
```

so obviously a map is not needed here.

In fact we have cgroupv2-only implementation of parseCgroupFile here:
https://github.com/opencontainers/runc/blob/a891fee8c6fa7473893c5102b74922ac21f99415/libcontainer/cgroups/fs2/defaultpath.go#L75

but it's not public. If it is to be made public I think `GetOwnCgroup`/`GetInitCgroup` is a better API. If we want to have  `Get{Own,Init}Cgroup{,Path}` for v2, there are three ways:
1. have a separate v1 and v2 implementations (for v2, drop the `subsystem` argument)
2. have an implementation that works for both v1 and v2, ignoring the `subsystem` for v2 case.
3. have an implementation that works for both v1 and v2, checking the `subsystem` is available for v2 case.

I am very open to change all, just need to get more user cases. What use case do you have in mind?

giuseppe: even without considering the hybrid model and just considering the v2 unified hierarchy with all the controllers assigned to cgroup v2, there could still be v1 named hierarchies, so a process  belongs to a cgroup v2 hierarchy and also to named hierarchies.  systemd for example creates a named hierarchy even on cgroup v2 when running a container that uses an old version of systemd.

If someone is interested, I think returning a map would still make sense on cgroup v2 to cover this case.

Kubernetes is currently using `ParseCgroupFile` to parse `/proc/$PID/cgroup` for another process, so it would be useful if it allows to specify the file path.

kolyshkin: Here 
https://github.com/kubernetes/kubernetes/blob/fb9e1946b0af89998b3e366d49399f6365c94c56/pkg/kubelet/cm/container_manager_linux.go#L887-L894 

As you can see, the value returned by `ParseCgroupFile` is used in different ways for v1 and v2 -- this is the primary reason I want to make it v1-specific. I mean, ideally the code above should look like

```go
func getContainer(pid int) (string, error) {
if cgroups.IsCgroup2UnifiedMode() {
	return cgroupv2.GetProcessCgroup(pid)
}

cgs, err := cgroupv1.ParseCgroupFile(fmt.Sprintf("/proc/%d/cgroup", pid))
if err != nil {
	return "", err
}
...
```

but since we don't have `cgroupv2.GetProcessCgroup()` (yet?), I can redo this patchset to not move `ParseCgroupFile()` (for now?)

giuseppe: yes, I agree with you.  What do you think about `cgroupv2.ParseCgroupFile()` that is a wrapper around `cgroupv1.ParseCgroupFile()`?  The cgroupv1 version can still be used if someone cares about other hierarchies

kolyshkin: I have updated this PR to not move `ParseCgroupFile` to `cgroupv1` package.

AkihiroSuda:
needs rebase

kolyshkin:
rebased

kolyshkin:
I tried compiling kubernetes with runc vendored with the changes from this PR. The majority of issues were not related to this PR, but to some of the previous ones. In particular, I am puzzled about  since it removes `AllowAllDevices` used by kubernetes.

Here are the changes needed to compile kubernetes with this runc, in their entirety: https://gist.github.com/kolyshkin/460faa44d952de70fc6676043e065402

It consists of:
1. removal of `AllowAllDevices = true` due to https://github.com/opencontainers/runc/pull/2391/commits/b2bec9806f999f55c0d0f9ce1b1064ffbc62f4a9
2. changing the code to get/store an instance cgroup manager (very similar to https://github.com/google/cadvisor/pull/2469/commits/6a95c22c20877e66f166ba1224f90e7904e0f7a8, due to PR #2387)
3. changing the package name for Ids that were moved to `cgroupv1` package (caused by this PR).

I have also tried compiling cadvisor against this changes. The diff is trivial, mostly looks like this:
```diff
 type CgroupSubsystems struct {
        // Cgroup subsystem mounts.
        // e.g.: "/sys/fs/cgroup/cpu" -> ["cpu", "cpuacct"]
-       Mounts []cgroups.Mount
+       Mounts []cgroupv1.Mount
 
        // Cgroup subsystem to their mount location.
        // e.g.: "cpu" -> "/sys/fs/cgroup/cpu"
@@ -37,7 +42,7 @@ type CgroupSubsystems struct {
 // Get information about the cgroup subsystems those we want
 func GetCgroupSubsystems(includedMetrics container.MetricSet) (CgroupSubsystems, error) {
        // Get all cgroup mounts.
-       allCgroups, err := cgroups.GetCgroupMounts(true)
+       allCgroups, err := cgroupv1.GetCgroupMounts(true)
        if err != nil {
                return CgroupSubsystems{}, err
        }
```

giuseppe:
> I have also tried compiling cadvisor against this changes. The diff is trivial, mostly looks like this:
> 
> ```diff
>  type CgroupSubsystems struct {
>         // Cgroup subsystem mounts.
>         // e.g.: "/sys/fs/cgroup/cpu" -> ["cpu", "cpuacct"]
> -       Mounts []cgroups.Mount
> +       Mounts []cgroupv1.Mount
>  
>         // Cgroup subsystem to their mount location.
>         // e.g.: "cpu" -> "/sys/fs/cgroup/cpu"
> @@ -37,7 +42,7 @@ type CgroupSubsystems struct {
>  // Get information about the cgroup subsystems those we want
>  func GetCgroupSubsystems(includedMetrics container.MetricSet) (CgroupSubsystems, error) {
>         // Get all cgroup mounts.
> -       allCgroups, err := cgroups.GetCgroupMounts(true)
> +       allCgroups, err := cgroupv1.GetCgroupMounts(true)
>         if err != nil {
>                 return CgroupSubsystems{}, err
>         }
> ```

would that work on cgroup v2?

kolyshkin:
> would that work on cgroup v2?

The code in question is only called for v1.

There is only one place where `cgroupv1.` is used for cgroupv2 case, and it's only a data structure (`Mount`), not a function call. This obviously cries for refactoring, but will work as is.

kolyshkin:
> The code in question is only called for v1.

I was wrong, the changes I made made it possible to merely compile cadvisor. To make it work for cgroupv2, more changes are needed. I will look into it. Not a showstopper for this PR though, since cadvisor is not currently working with cgroupv2.

giuseppe:
> Not a showstopper for this PR though, since cadvisor is not currently working with cgroupv2.

there are surely issues left but cAdvisor is already used by the Kubelet on cgroup v2

thaJeztah:
oh, I saw your earlier comment https://github.com/opencontainers/runc/pull/2411#issuecomment-631069381, but didn't reply

> The reason is, all this stuff is very cgroupv1 specific and it makes no sense to use it when cgroupv2 is used on the host. In case it is still used in such scenario it means the code is doing something wrong and is broken.
>
> @thaJeztah please note the above (and I will be happy to help fix it in moby code).

Thanks for the heads-up; I'll continue following progress on this PR

kolyshkin:
> I'll continue following progress on this PR

After moby/moby#41014 and moby/moby#41016 the changes to adapt this are trivial (as trivial as  `s|cgroups\.|cgroups/cgroupv1.|g`), so not much to look at here.

AkihiroSuda:
@mrunalp @cyphar LGTY?

AkihiroSuda:

mrunalp:
There is still a case where we would end up using v1 code in v2 in k8s upstream. @giuseppe I know you were looking at updating this into cadvisor. Do you have your latest tree?

giuseppe:
> There is still a case where we would end up using v1 code in v2 in k8s upstream. @giuseppe I know you were looking at updating this into cadvisor. Do you have your latest tree?

get it to work into cAdvisor requires some substantial changes to how cAdvisor uses libcontainer, or recreate in cAdvisor the cgroupv1.Mount struct when running on cgroup v2.  I am worried that there is not much time for introducing changes given that the Kubernetes code freeze for 1.19 is in 10 days.

kolyshkin:
So, what we can do now is commit everything except the last patch (so the v1 code is moved to a separate file but not (yet) to a separate package, leaving this last step for later time.

mrunalp:
@kolyshkin Doing it in steps sounds good :+1:

kolyshkin:
OK so what I did is removed two commits:
1. one that moved code to a separate cgroupv1 package
2. once that made `type NotFoundError` private

@giuseppe PTAL

kolyshkin:
Phase 2 (move v1 to a separate package) can be done once we made sure all the users are (or can be) converted.

kolyshkin:
rebased

libcontainer/cgroups/utils.go
```diff
@@ -217,20 +217,12 @@ func getCgroupMountsHelper(ss map[string]bool, mi io.Reader, all bool) ([]Mount,
...
-		if err != nil {
-			return nil, err
-		}
-		m := Mount{
-			Mountpoint: unifiedMountpoint,
-			Root:       unifiedMountpoint,
-			Subsystems: availableControllers,
-		}
-		return []Mount{m}, nil
+		return nil, errUnified
```
giuseppe: also this change breaks the Kubelet and cAdvisor

kolyshkin: Fixed

PR #2412, opened 2020-05-15
remove cgroup path recursively in cgroup v2

### LINKED ISSUE #2403: cgroup2: Manager.Destroy() should recursively delete subgroups
```console
$ docker version
Client:
 Version:           20.03.0-dev
 API version:       1.41
 Go version:        go1.13.10
 Git commit:        5bf109154
 Built:             Thu May 14 05:59:55 2020
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          dev
  Go version:       go1.13.10
  Git commit:       4e102ab1f0
  Built:            Thu May 14 05:57:27 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.3.0-577-g7207226e
  GitCommit:        7207226e9d9aa6af9b48b681137139873b7c692d
 runc:
  Version:          1.0.0-rc10+dev
  GitCommit:        3f1e88699199d844dd211cbf5108eeff9444674e
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

```console
$ docker run -it --rm --privileged alpine mkdir /sys/fs/cgroup/foo
$ ls /sys/fs/cgroup/docker/
7643812e0a0390cc595543190844ab9c10e064646b563549c17e3c05ce519523  cpu.pressure           io.pressure          memory.stat
cgroup.controllers                                                cpuset.cpus            io.stat              memory.swap.current
cgroup.events                                                     cpuset.cpus.effective  io.weight            memory.swap.events
cgroup.freeze                                                     cpuset.cpus.partition  memory.current       memory.swap.max
cgroup.max.depth                                                  cpuset.mems            memory.events        pids.current
cgroup.max.descendants                                            cpuset.mems.effective  memory.events.local  pids.events
cgroup.procs                                                      cpu.stat               memory.high          pids.max
cgroup.stat                                                       cpu.uclamp.max         memory.low           rdma.current
cgroup.subtree_control                                            cpu.uclamp.min         memory.max           rdma.max
cgroup.threads                                                    cpu.weight             memory.min
cgroup.type                                                       cpu.weight.nice        memory.oom.group
cpu.max                                                           io.max                 memory.pressure
$ ls /sys/fs/cgroup/docker/7643812e0a0390cc595543190844ab9c10e064646b563549c17e3c05ce519523/
cgroup.controllers      cgroup.threads         cpuset.mems.effective  io.pressure          memory.max           pids.current
cgroup.events           cgroup.type            cpu.stat               io.stat              memory.min           pids.events
cgroup.freeze           cpu.max                cpu.uclamp.max         io.weight            memory.oom.group     pids.max
cgroup.max.depth        cpu.pressure           cpu.uclamp.min         memory.current       memory.pressure      rdma.current
cgroup.max.descendants  cpuset.cpus            cpu.weight             memory.events        memory.stat          rdma.max
cgroup.procs            cpuset.cpus.effective  cpu.weight.nice        memory.events.local  memory.swap.current
cgroup.stat             cpuset.cpus.partition  foo                    memory.high          memory.swap.events
cgroup.subtree_control  cpuset.mems            io.max                 memory.low           memory.swap.max
```

`Manager.Destroy()` should recursively delete subgroups.

lifubang:
fix #2403 

If we have subcgroups in a cgroup path, we should remove it recursively.

### DIFF
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -156,11 +156,35 @@ func (m *manager) Freeze(state configs.FreezerState) error {
 	return nil
 }
 
-func (m *manager) Destroy() error {
-	if err := os.Remove(m.dirPath); err != nil && !os.IsNotExist(err) {
+// removeCgroupPath aims to remove cgroup path recursively
+// Because there may be subcgroups in it.
+func removeCgroupPath(path string) error {
+	infos, err := ioutil.ReadDir(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = nil
+		}
 		return err
 	}
-	return nil
+	for _, info := range infos {
+		if info.IsDir() {
+			// We should remove subcgroups dir first
+			if err = removeCgroupPath(filepath.Join(path, info.Name())); err != nil {
+				break
+			}
+		}
+	}
+	if err == nil {
+		err = os.Remove(path)
+		if os.IsNotExist(err) {
+			err = nil
+		}
+	}
+	return err
+}
+
+func (m *manager) Destroy() error {
+	return removeCgroupPath(m.dirPath)
 }
 
 func (m *manager) Path(_ string) string {
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -51,3 +51,58 @@ function teardown() {
   runc delete --force notexists
   [ "$status" -eq 0 ]
 }
+
+@test "runc delete --force in cgroupv2 with subcgroups" {
+  requires cgroups_v2 root
+  set_cgroups_path "$BUSYBOX_BUNDLE"
+
+  # grant `rw` priviledge to `/sys/fs/cgroup`
+  cat "${BUSYBOX_BUNDLE}/config.json"\
+   | jq '.mounts |= map((select(.type=="cgroup") | .options -= ["ro"]) // .)'\
+   > "${BUSYBOX_BUNDLE}/config.json.tmp"
+  mv "${BUSYBOX_BUNDLE}/config.json"{.tmp,}
+
+  # run busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  # check state
+  testcontainer test_busybox running
+
+  # create a sub process
+  __runc exec -d test_busybox sleep 1d
+  [ "$status" -eq 0 ]
+
+  # find the pid of sleep
+  pid=$(__runc exec test_busybox ps -a | grep 1d | awk '{print $1}')
+  [[ ${pid} =~ [0-9]+ ]]
+
+  # create subcgroups
+  cat <<EOF > nest.sh
+  cd /sys/fs/cgroup
+  for f in \$(cat cgroup.controllers); do echo +\$f > cgroup.subtree_control; done
+  mkdir foo
+  cd foo
+  echo threaded > cgroup.type
+  echo ${pid} > cgroup.threads
+  cat cgroup.threads
+EOF
+  cat nest.sh | runc exec test_busybox sh
+  [[ ${output} =~ [0-9]+ ]]
+
+  # check create subcgroups success
+  [ -d $CGROUP_PATH/foo ]
+
+  # check cgroup.threads' value
+  runc exec test_busybox cat /sys/fs/cgroup/foo/cgroup.threads
+  [[ ${output} =~ [0-9]+ ]]
+
+  # force delete test_busybox
+  runc delete --force test_busybox
+
+  runc state test_busybox
+  [ "$status" -ne 0 ]
+
+  # check delete subcgroups success
+  [ ! -d $CGROUP_PATH/foo ]
+}


libcontainer/cgroups/fs2/fs2.go
```diff
@@ -158,7 +158,7 @@ func (m *manager) Freeze(state configs.FreezerState) error {
 }
 
 func (m *manager) Destroy() error {
-	if err := os.Remove(m.dirPath); err != nil && !os.IsNotExist(err) {
+	if err := os.RemoveAll(m.dirPath); err != nil && !os.IsNotExist(err) {
```
kolyshkin, line 161: 1. Theoretically speaking, we should only remove the directories, as cgroup file removes is not supported and will result in some error (possibly `EPERM`). I am not sure how RemoveAll() works, but ...

2. os.RemoveAll() doc says:

> It removes everything it can but returns the first error it encounters. If the path does not exist, RemoveAll returns nil (no error).

so 
* it should not return `IsNotExist()`, and thus the check for it is not needed
* it might return some other error, which we'll return.

All of this makes me think we have to write a removal function based on `filepath.Walk` or `ioutil.ReadDir`, which will only remove directories.

kolyshkin:
We need a test case -- create sub-cgroups, check they are actually removed.

libcontainer/cgroups/fs2/fs2.go
kolyshkin: But maybe I'm wrong and this works fine as it is. This is why we need a test case or two.

lifubang:
> We need a test case -- create sub-cgroups, check they are actually removed.

I add a test case in `delete.bats`, it test success in my computer(Fedora 31).
```
root@acm:/opt/mygo/src/github.com/opencontainers/runc# make localintegration TESTPATH=/delete.bats
go build "-mod=vendor" -buildmode=pie  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="15ec27f390f377780e1c332d24588e7e2ba7d3d3-dirty" -X main.version=1.0.0-rc10+dev " -o runc .
go build "-mod=vendor" -buildmode=pie  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="15ec27f390f377780e1c332d24588e7e2ba7d3d3-dirty" -X main.version=1.0.0-rc10+dev " -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
bats -t tests/integration/delete.bats
1..4
ok 1 runc delete
ok 2 runc delete --force
ok 3 runc delete --force with subcgroups in v2
ok 4 runc delete --force ignore not exist
root@acm:/opt/mygo/src/github.com/opencontainers/runc#
```
But failed in CI:
```
not ok 24 runc delete --force with subcgroups in v2
# (in test file tests/integration/delete.bats, line 74)
```
the line 72-74 is:
```
# run busybox detached
  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
  [ "$status" -eq 0 ]
```
So, we can't create `rw` cgroup mount in CI?
@kolyshkin PTAL, thanks.

tests/integration/delete.bats
```diff
@@ -47,6 +47,52 @@ function teardown() {
...
+ 
+  # run busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  # check state
+  testcontainer test_busybox running
+
+  # create subcgroups
+  runc exec -t test_busybox mkdir -p /sys/fs/cgroup/foo/bar
```
lifubang, line 80: How about change this line to `mkdir -p $CGROUP_PATH/foo/bar`?
I think it is enough for this test case.

libcontainer/cgroups/fs2/fs2.go
lifubang: > * it should not return `IsNotExist()`, and thus the check for it is not needed

Yes, removed now.

tests/integration/delete.bats
```diff
@@ -51,3 +51,49 @@ function teardown() {
...
+            "noexec",
+            "nodev",
+            "relatime",
+            "rw"
+    ]
+EOF
+    )
+  DATA=$(echo ${DATA} | sed 's/\n/\\n/g' | sed 's/\//\\\//g')
+  sed -i ":a;N;s/\n//g;ta" ${BUSYBOX_BUNDLE}/config.json
+  sed -i "s/\"destination\": \"\/sys\/fs\/cgroup\",[^}]*/${DATA}/" ${BUSYBOX_BUNDLE}/config.json
```
kolyshkin, line 75: maybe you can use `jq` here instead of `sed`, could be more readable.

tests/integration/delete.bats
```diff
@@ -51,3 +51,49 @@ function teardown() {
...
+
+  # force delete test_busybox
+  runc delete --force test_busybox
+
+  runc state test_busybox
+  [ "$status" -ne 0 ]
+
+  # check delete subcgroups success
+  [ ! -d $CGROUP_PATH/foo ]
+}
```
kolyshkin, line 99: nit: no newline at EOF

tests/integration/delete.bats
kolyshkin: In fact @AkihiroSuda already added a helper function that does it, see https://github.com/opencontainers/runc/pull/2416/

libcontainer/cgroups/fs2/fs2.go
AkihiroSuda: As Kir said, `filepath.Walk` is the right way.

tests/integration/delete.bats
```diff
@@ -51,3 +51,36 @@ function teardown() {
...
 }
+
+@test "runc delete --force in cgroupv2 with subcgroups" {
+  requires cgroups_v2 root
+  set_cgroups_path "$BUSYBOX_BUNDLE"
+
+  # grant `rw` priviledge to `/sys/fs/cgroup`
+  cat ${BUSYBOX_BUNDLE}/config.json\
+   | jq '.mounts |= map((select(.type=="cgroup") | .options -= ["ro"] | .options += ["rw"]) //.)'\
+   > ${BUSYBOX_BUNDLE}/config.json
```
lifubang, line 62: Why `>` can't work in here?

lifubang: It can work in my pc:
```
[root@localhost runc]# make localintegration TESTPATH=/delete.bats RUNC_USE_SYSTEMD=1
go build "-mod=vendor" -buildmode=pie  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="98a52396ea5807c141118704e0540b76edc5c03e" -X main.version=1.0.0-rc10+dev " -o runc .
go build "-mod=vendor" -buildmode=pie  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="98a52396ea5807c141118704e0540b76edc5c03e" -X main.version=1.0.0-rc10+dev " -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
bats -t tests/integration/delete.bats
1..4
ok 1 runc delete
ok 2 runc delete --force
ok 3 runc delete --force ignore not exist
ok 4 runc delete --force in cgroupv2 with subcgroups
[root@localhost runc]#
```

libcontainer/cgroups/fs2/fs2.go
lifubang: > As Kir said, `filepath.Walk` is the right way.

Yes, I looked the code of `os.RemoveAll`, it does removes everything it can. But ignore the `EPERM` error.
```
if err != syscall.EISDIR && err != syscall.EPERM && err != syscall.EACCES {
```
So, it does not worth to use `os.RemoveAll`.
Changed to `ioutil.ReadDir`.

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
 	return fmt.Errorf("Failed to remove paths: %v", paths)
 }
 
+// RemoveCgroupPath aims to remove cgroup path recursively
+// Because there may be subcgroups in it.
+func RemoveCgroupPath(path string) error {
```
AkihiroSuda, line 484: `RemovePathUnified`?

AkihiroSuda:
LGTM, though I expect the function name to be more descriptive

AkihiroSuda:
LGTM

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
...
+// RemovePathUnified aims to remove cgroup path recursively
+// Because there may be subcgroups in it.
+func RemovePathUnified(path string) error {
+	infos, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+	for _, info := range infos {
+		if info.IsDir() {
+			if err = RemovePathUnified(filepath.Join(path, info.Name())); err != nil {
```
lifubang, line 491: Maybe here should ignore NotExist error?

AkihiroSuda: Yes

AkihiroSuda:
LGTM

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
 	return fmt.Errorf("Failed to remove paths: %v", paths)
 }
 
+// RemovePathUnified aims to remove cgroup path recursively
+// Because there may be subcgroups in it.
+func RemovePathUnified(path string) error {
+	infos, err := ioutil.ReadDir(path)
+	if err == nil {
```
kolyshkin, line 486: 1 nit:
```go
if err != nil {
   if os.IsNotExist(err) {
      err = nil
   }
   return err
}
```

so
1. the rest of function will be less indented
2. we won't return ENOENT.

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
...
+		for _, info := range infos {
+			if info.IsDir() {
+				// We should remove subcgroups dir first
+				if err = RemovePathUnified(filepath.Join(path, info.Name())); err != nil && !os.IsNotExist(err) {
```
kolyshkin, line 490: with the above suggestion, we won't have to check for `IsNotExist` here.

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
...
+					break
+				}
+			}
+		}
+		if err == nil || os.IsNotExist(err) {
```
kolyshkin, line 495: ...and here

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
...
+			return os.Remove(path)
```
kolyshkin, line 496: I think we should avoid returning `ENOENT` from here as well (similar to what `os.RemoveAll()` does.

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
 	return fmt.Errorf("Failed to remove paths: %v", paths)
 }
 
+// RemovePathUnified aims to remove cgroup path recursively
+// Because there may be subcgroups in it.
+func RemovePathUnified(path string) error {
+	infos, err := ioutil.ReadDir(path)
```
kolyshkin, line 485: OK, I think we can optimize things for the common case here:

```go
err := os.Remove(path)
if err == nil || os.IsNotExist(err) {
  return nil
}
// remove subdirectories first
infos, err :=  ioutil.ReadDir(path)
....
```

libcontainer/cgroups/utils.go
```diff
@@ -479,6 +479,26 @@ func RemovePaths(paths map[string]string) (err error) {
 	return fmt.Errorf("Failed to remove paths: %v", paths)
 }
 
+// RemovePathUnified aims to remove cgroup path recursively
+// Because there may be subcgroups in it.
+func RemovePathUnified(path string) error {
```
kolyshkin, line 484: Since we're only using this function from `fs2`, why don't we move it to fs2 and make it private.

Maybe rename to `removeCgroupPath` or so.

kolyshkin:
left an optimization suggestion for `RemovePathUnified`, and a few nits.

Found another problem as well -- the test case added works even without the changes. I think we need to make sure the test case fails, then make it work, otherwise it's not clear what are we fixing here. Maybe the sub-cgroup need to have some limits set? IDK.

```console
$ git fetch lifubang 
...
$ git checkout removecgpath
Branch 'removecgpath' set up to track remote branch 'removecgpath' from 'lifubang'.
Switched to a new branch 'removecgpath'

$ git show libcontainer/ | patch -p1 -R # revert the "fixing" part
patching file libcontainer/cgroups/fs2/fs2.go
patching file libcontainer/cgroups/utils.go

$ sudo -s
# RUNC_USE_SYSTEMD=yes make localintegration TESTPATH=/delete.bats
go build "-mod=vendor" -buildmode=pie  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="430a1e88c359974ba441cf75b619626cece7c261-dirty" -X main.version=1.0.0-rc10+dev " -o runc .
go build "-mod=vendor" -buildmode=pie  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="430a1e88c359974ba441cf75b619626cece7c261-dirty" -X main.version=1.0.0-rc10+dev " -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
bats -t tests/integration/delete.bats
1..4
ok 1 runc delete
ok 2 runc delete --force
ok 3 runc delete --force ignore not exist
ok 4 runc delete --force in cgroupv2 with subcgroups
```

AkihiroSuda:
LGTM

mrunalp:
LGTM

libcontainer/cgroups/utils.go
kolyshkin: I still think it's a pretty good optimization. Will follow up.

tests/integration/delete.bats
```diff
@@ -51,3 +51,58 @@ function teardown() {
...
+  [ "$status" -eq 0 ]
+
+  # find the pid of sleep
+  pid=$(__runc exec test_busybox ps -a | grep 1d | awk '{print $1}')
+  [[ ${pid} =~ [0-9]+ ]]
+
+  # create subcgroups
+  cat <<EOF > nest.sh
+  cd /sys/fs/cgroup
+  for f in \$(cat cgroup.controllers); do echo +\$f > cgroup.subtree_control; done
```
kolyshkin, line 83: This gives an error on my system when trying to enable io controller.

Most probably just `+pids` is enough (and it won't error).

tests/integration/delete.bats
```diff
@@ -51,3 +51,58 @@ function teardown() {
...
+  # run busybox detached
+  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  [ "$status" -eq 0 ]
+
+  # check state
+  testcontainer test_busybox running
+
+  # create a sub process
+  __runc exec -d test_busybox sleep 1d
+  [ "$status" -eq 0 ]
```
kolyshkin, line 74: `__runc` does not set `$status`, so this check is checking and older status.

tests/integration/delete.bats
```diff
@@ -51,3 +51,58 @@ function teardown() {
...
+
+  # find the pid of sleep
+  pid=$(__runc exec test_busybox ps -a | grep 1d | awk '{print $1}')
```
kolyshkin, line 77: `busybox ps` does not have `-a`, but fortunately it is ignored.

tests/integration/delete.bats
```diff
@@ -51,3 +51,58 @@ function teardown() {
...
+  cd /sys/fs/cgroup
+  for f in \$(cat cgroup.controllers); do echo +\$f > cgroup.subtree_control; done
+  mkdir foo
+  cd foo
+  echo threaded > cgroup.type
+  echo ${pid} > cgroup.threads
+  cat cgroup.threads
+EOF
+  cat nest.sh | runc exec test_busybox sh
+  [[ ${output} =~ [0-9]+ ]]
```
kolyshkin, line 91: you don't check the exit status

tests/integration/delete.bats
```diff
@@ -51,3 +51,58 @@ function teardown() {
...
+  __runc exec -d test_busybox sleep 1d
+  [ "$status" -eq 0 ]
+
+  # find the pid of sleep
+  pid=$(__runc exec test_busybox ps -a | grep 1d | awk '{print $1}')
+  [[ ${pid} =~ [0-9]+ ]]
+
+  # create subcgroups
+  cat <<EOF > nest.sh
+  cd /sys/fs/cgroup
```
kolyshkin, line 82: makes sense to add `set -x -u -e`

tests/integration/delete.bats
```diff
@@ -51,3 +51,58 @@ function teardown() {
...
+EOF
+  cat nest.sh | runc exec test_busybox sh
+  [[ ${output} =~ [0-9]+ ]]
+
+  # check create subcgroups success
+  [ -d $CGROUP_PATH/foo ]
+
+  # check cgroup.threads' value
+  runc exec test_busybox cat /sys/fs/cgroup/foo/cgroup.threads
+  [[ ${output} =~ [0-9]+ ]]
```
kolyshkin, line 98: We have already checked it above (the output of `nest.sh` script), looks like this check is redundant.

PR #2415, opened 2020-05-19
fix for missing fmt package

knarfeh:
Signed-off-by: Jun He <hejun1874@gmail.com>

### DIFF
--- a/libcontainer/cgroups/fs/cpuset.go
+++ b/libcontainer/cgroups/fs/cpuset.go
@@ -5,6 +5,7 @@ package fs
 import (
 	"bytes"
 	"errors"
+	"fmt"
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/fs/cpuset.go
```diff
@@ -5,6 +5,7 @@ package fs
 import (
 	"bytes"
 	"errors"
+	"fmt"
```
lifubang, line 8: I think you should remove "fmt" here.
And change `fmt.Errorf` to `errors.Errorf`.

AkihiroSuda:
carried as this one is blocking other PRs: https://github.com/opencontainers/runc/pull/2418

PR #2416, opened 2020-05-19
cgroup2: exec: join the cgroup of the init process on EBUSY

### LINKED ISSUE #2356: cgroup2: how can we support nested containers with domain controllers?
https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#no-internal-process-constraint

> #### No Internal Process Constraint
> Non-root cgroups can distribute domain resources to their children only when they don’t have any processes of their own. In other words, only domain cgroups which don’t contain any processes can have domain controllers enabled in their “cgroup.subtree_control” files.

http://man7.org/linux/man-pages/man7/cgroups.7.html

> As at Linux 4.19, the following controllers are threaded: _cpu_, _perf_event_, and _pids_.

The constraint seems blocker toward supporting nested containers like dind and kind.

```console
$ sudo podman run -it --rm --runtime=runc --privileged --cgroupns=private alpine
/ # cd /sys/fs/cgroup/
/sys/fs/cgroup # cat cgroup.controllers 
cpu io memory pids
/sys/fs/cgroup # echo +cpu > cgroup.subtree_control 
/sys/fs/cgroup # echo +io > cgroup.subtree_control 
sh: write error: Not supported
/sys/fs/cgroup # echo +memory > cgroup.subtree_control 
sh: write error: Not supported
/sys/fs/cgroup # echo +pids > cgroup.subtree_control 
```

The situation is same on crun as well.

@giuseppe @kolyshkin @vbatts  Thoughts?
A workaround is to specify an entrypoint script that moves the processes in the namespaced-root cgroup to another cgroup.

AkihiroSuda:
Fix #2356.

~~Contains https://github.com/opencontainers/runc/pull/2418~~

- - -

Note: this behavior is different from crun v0.13.
crun joins "crun-exec` group rather than the init group: https://github.com/containers/crun/commit/af1080b175cac63fe1c42397bfe3cca982a93896
The next version of crun will switch to the init cgroup: https://github.com/containers/crun/pull/365

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -5,6 +5,7 @@ package libcontainer
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
@@ -12,12 +13,14 @@ import (
 	"strconv"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/logs"
 	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runc/libcontainer/utils"
 
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
@@ -66,6 +69,7 @@ type setnsProcess struct {
 	fds             []string
 	process         *Process
 	bootstrapData   io.Reader
+	initProcessPid  int
 }
 
 func (p *setnsProcess) startTime() (uint64, error) {
@@ -100,7 +104,25 @@ func (p *setnsProcess) start() (err error) {
 	}
 	if len(p.cgroupPaths) > 0 {
 		if err := cgroups.EnterPid(p.cgroupPaths, p.pid()); err != nil && !p.rootlessCgroups {
-			return newSystemErrorWithCausef(err, "adding pid %d to cgroups", p.pid())
+			// On cgroup v2 + nesting + domain controllers, EnterPid may fail with EBUSY.
+			// https://github.com/opencontainers/runc/issues/2356#issuecomment-621277643
+			// Try to join the cgroup of InitProcessPid.
+			if cgroups.IsCgroup2UnifiedMode() {
+				initProcCgroupFile := fmt.Sprintf("/proc/%d/cgroup", p.initProcessPid)
+				initCg, initCgErr := cgroups.ParseCgroupFile(initProcCgroupFile)
+				if initCgErr == nil {
+					if initCgPath, ok := initCg[""]; ok {
+						initCgDirpath := filepath.Join(fs2.UnifiedMountpoint, initCgPath)
+						logrus.Debugf("adding pid %d to cgroups %v failed (%v), attempting to join %q (obtained from %s)",
+							p.pid(), p.cgroupPaths, err, initCg, initCgDirpath)
+						// NOTE: initCgDirPath is not guaranteed to exist because we didn't pause the container.
+						err = cgroups.WriteCgroupProc(initCgDirpath, p.pid())
+					}
+				}
+			}
+			if err != nil {
+				return newSystemErrorWithCausef(err, "adding pid %d to cgroups", p.pid())
+			}
 		}
 	}
 	if p.intelRdtPath != "" {
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -136,3 +136,49 @@ EOF
     [ "$status" -eq 0 ]
     [[ ${lines[0]} == *"cgroups_exec"* ]]
 }
+
+@test "runc exec (cgroup v2 + init process in non-root cgroup) succeeds" {
+    requires root cgroups_v2
+
+    set_cgroups_path "$BUSYBOX_BUNDLE"
+    set_cgroup_mount_writable "$BUSYBOX_BUNDLE"
+
+    runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_group
+    [ "$status" -eq 0 ]
+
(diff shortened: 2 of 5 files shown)


AkihiroSuda:
CI is broken due to https://github.com/opencontainers/runc/pull/2415

AkihiroSuda:
@kolyshkin @cyphar @giuseppe PTAL

libcontainer/process_linux.go
```diff
@@ -100,7 +104,25 @@ func (p *setnsProcess) start() (err error) {
 	}
 	if len(p.cgroupPaths) > 0 {
 		if err := cgroups.EnterPid(p.cgroupPaths, p.pid()); err != nil && !p.rootlessCgroups {
-			return newSystemErrorWithCausef(err, "adding pid %d to cgroups", p.pid())
+			// On cgroup v2 + nesting + domain controllers, EnterPid may fail with EBUSY.
+			// https://github.com/opencontainers/runc/issues/2356#issuecomment-621277643
+			// Try to join the cgroup of InitProcessPid.
+			if cgroups.IsCgroup2UnifiedMode() {
```
giuseppe, line 110: should it not be done only with `err == EBUSY`?

AkihiroSuda: I guess it is worth trying to join pid1 even on other errors

AkihiroSuda:
@kolyshkin @mrunalp PTAL

tests/integration/cgroups.bats
```diff
@@ -125,3 +125,55 @@ EOF
     [ "$status" -eq 0 ]
     [[ ${lines[0]} == *"cgroups_exec"* ]]
 }
+
+@test "runc exec (cgroup v2 + init process in non-root cgroup) succeeds" {
+    requires root cgroups_v2
+
+    set_cgroups_path "$BUSYBOX_BUNDLE"
+# set_memory_limit is currently required to make sure "memory" appear in /sys/fs/cgroup/cgroup.controllers .
+# set_memory_limit will no longer be needed after merging "Partially revert "CreateCgroupPath: only enable needed controllers" (#2395)
```
mrunalp, line 134: #2395 has been merged.

libcontainer/process_linux.go
```diff
@@ -100,7 +104,25 @@ func (p *setnsProcess) start() (err error) {
...
+				initProcCgroupFile := fmt.Sprintf("/proc/%d/cgroup", p.initProcessPid)
+				initCg, initCgErr := cgroups.ParseCgroupFile(initProcCgroupFile)
+				if initCgErr == nil {
+					if initCgPath, ok := initCg[""]; ok {
+						initCgDirpath := filepath.Join(fs2.UnifiedMountpoint, initCgPath)
+						logrus.Warnf("adding pid %d to cgroups %v failed (%v), attempting to join %q (obtained from %s)",
```
mrunalp, line 116: I think we should drop this warning.

mrunalp:
Added one comment about removing the warning otherwise looks ready to merge. Thanks!

libcontainer/process_linux.go
AkihiroSuda: should be `Debugf` maybe?

mrunalp: yeah, debug is probably more acceptable here if we want to log anything at all.

AkihiroSuda: changed to Debugf

tests/integration/cgroups.bats
AkihiroSuda: updated

AkihiroSuda:
updated

mrunalp:
Thanks, will tag once tests pass again.

mrunalp:
LGTM

mrunalp:
@kolyshkin @cyphar ptal

tests/integration/helpers.bash
```diff
@@ -197,6 +197,15 @@ function set_resources_limit() {
   sed -i 's/\("linux": {\)/\1\n   "resources": { "pids": { "limit": 100 } },/'  "$bundle/config.json"
 }
 
+# Helper function to make /sys/fs/cgroup writable
+function set_cgroup_mount_writable(){
```
kolyshkin, line 201: nit: space before `{`.

I think there is another potential user of this function in commit 82fa19417986da834b7d6bdd7d3539c37a5735a4, maybe you can also make it use it while at it?

AkihiroSuda: done

tests/integration/delete.bats
```diff
@@ -55,13 +55,8 @@ function teardown() {
...
   set_cgroups_path "$BUSYBOX_BUNDLE"
-
-  # grant `rw` priviledge to `/sys/fs/cgroup`
-  cat "${BUSYBOX_BUNDLE}/config.json"\
-   | jq '.mounts |= map((select(.type=="cgroup") | .options -= ["ro"]) // .)'\
-   > "${BUSYBOX_BUNDLE}/config.json.tmp"
-  mv "${BUSYBOX_BUNDLE}/config.json"{.tmp,}
-
+  set_cgroup_mount_writable "$BUSYBOX_BUNDLE"
+
```
kolyshkin, line 59: nit: space at EOL

AkihiroSuda: fixed

AkihiroSuda:
fixed

libcontainer/process_linux.go
kolyshkin: @AkihiroSuda this decision (try to join pid1 on every error) can be problematic.

I've seen case (in https://github.com/opencontainers/runc/pull/4812#issuecomment-3076519132) in which the container init is moved to an entirely different cgroup (I'm not sure by whom, but suspect it's some GHA CI infra software). Once this happens, the container's limits, as configured in OCI spec, are no longer applied. Next, we call `runc exec` and it succeeds (thanks to the patch in this PR), but I'd rather have it failed, because something very bad has happened.

cyphar: I also somewhat question this idea in the first place -- administrative cgroup moves are generally not blocked by the cgroup infrastructure (this is an explicit design goal). I think you would only get this case with frozen cgroups, in which case an error is kind of reasonable IMHO?

PR #2418, opened 2020-05-19
fix "libcontainer/cgroups/fs/cpuset.go:63:14: undefined: fmt"

AkihiroSuda:
The compilation error had ocurred because of a bad rebase during #2401 and #2413

Replaces #2415

PR #2419, opened 2020-05-19
Remove "-buildmode=pie" from platforms that don't support it

tianon:
This sequence (and syntax) is taken directly from containerd's Makefile:
https://github.com/containerd/containerd/blob/4e08c2de67ec514b5602eea47804d41dfeabdc72/Makefile.linux#L21-L26

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,12 @@ VERSION := $(shell cat ./VERSION)
 ifneq ($(GO111MODULE),off)
 	MOD_VENDOR := "-mod=vendor"
 endif
-GO_BUILD := $(GO) build $(MOD_VENDOR) -buildmode=pie $(EXTRA_FLAGS) -tags "$(BUILDTAGS)" \
+ifeq ($(shell $(GO) env GOOS),linux)
+	ifeq (,$(filter $(shell $(GO) env GOARCH),mips mipsle mips64 mips64le ppc64))
+		GO_BUILDMODE := "-buildmode=pie"
+	endif
+endif
+GO_BUILD := $(GO) build $(MOD_VENDOR) $(GO_BUILDMODE) $(EXTRA_FLAGS) -tags "$(BUILDTAGS)" \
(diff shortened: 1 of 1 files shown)


tianon:
(CI failure appears to be due to https://github.com/opencontainers/runc/pull/2418 -- I'll keep an eye on that and rebase :+1:)

Makefile
```diff
@@ -18,7 +18,12 @@ VERSION := $(shell cat ./VERSION)
 ifneq ($(GO111MODULE),off)
 	MOD_VENDOR := "-mod=vendor"
 endif
-GO_BUILD := $(GO) build $(MOD_VENDOR) -buildmode=pie $(EXTRA_FLAGS) -tags "$(BUILDTAGS)" \
+ifeq ($(GOOS),linux)
+	ifneq ($(GOARCH),$(filter $(GOARCH),mips mipsle mips64 mips64le ppc64))
```
tianon, line 22: This isn't actually quite right because `GOOS` and `GOARCH` aren't necessarily set, so we need to query them.  I'll update shortly.

tianon:
All good now, and all green at https://travis-ci.org/github/opencontainers/runc/builds/688947950 although something in the reporting of that status back to GitHub seems mixed up. :sweat_smile:

kolyshkin:
> This sequence (and syntax) is taken directly from containerd's Makefile:

Technically speaking, this is no longer true.

LGTM otherwise

tianon:
That's fair, but it's still pretty close (definitely close enough to recognize readily, IMO) -- do you think the differences are large enough that I should remove that from the commit text?

kolyshkin:
> do you think the differences are large enough that I should remove that from the commit text?

No, it totally makes sense to mention it, but I would replace "is taken directly from" by "is modelled after", "is inspired by", "idea is stolen from" or something...

tianon:
Sure, totally fair -- updated!

AkihiroSuda:
LGTM

kolyshkin:
LGTM

PR #2420, opened 2020-05-19
Wait for criuProcess once

tedyu:
In linuxContainer#criuSwrk(), the deferred func may wait on criuProcess for a second time, if the following has been executed (around line 1640):
```
	st, err = criuProcess.Wait()
```
This PR adds check for st so that there is no redundant wait.

Note: there is already deferred criuClientCon.Close() on line 1476

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1403,11 +1403,14 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 	// cmd.Process will be replaced by a restored init.
 	criuProcess := cmd.Process
 
+	var criuProcessState *os.ProcessState
 	defer func() {
-		criuClientCon.Close()
-		_, err := criuProcess.Wait()
-		if err != nil {
-			return
+		if criuProcessState == nil {
+			criuClientCon.Close()
+			_, err := criuProcess.Wait()
+			if err != nil {
+				logrus.Warnf("wait on criuProcess returned %v", err)
+			}
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


kolyshkin:
1. I do not understand the meaning of
```go
                if err != nil {
                        return
                }
```
in the `defer func()`. Looks like it does nothing, so it does not make sense to have it at all (unless I am mistaken and it actually means something).

2. Since we are ignoring the error, there is nothing wrong to call Wait for a second time.

3. I would remove the second "criuClientCon.Close()" (as a separate patch, first in the series).

kolyshkin:
@avagin PTAL

tedyu:
w.r.t. #1, I have updated the PR to log the error.

libcontainer/container_linux.go
```diff
@@ -1504,11 +1504,14 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
...
 	defer func() {
-		criuClientCon.Close()
-		_, err := criuProcess.Wait()
-		if err != nil {
-			return
+		if st == nil {
+			criuClientCon.Close()
+			_, err := criuProcess.Wait()
+			if err != nil {
+				logrus.Warningf("wait on criuProcess returned %v", err)
```
kolyshkin, line 1513: nit: `s/Warningf/Warnf/` (both are OK but looks like `Warnf` is used everywhere else in runc code).

tedyu: Addressed.

tedyu:
Ping @avagin

libcontainer/container_linux.go
```diff
@@ -1504,11 +1504,14 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 	// cmd.Process will be replaced by a restored init.
 	criuProcess := cmd.Process
 
+	var st *os.ProcessState
```
avagin, line 1507: nit: this name is good when is used in a block of 20-lines code, but in this case, I would use something more descriptive (e g criuProcessState).

avagin:
LGTM

libcontainer/container_linux.go
tedyu: Your comment has been addressed.

kolyshkin:
LGTM

kolyshkin:
@AkihiroSuda @mrunalp PTAL

mrunalp:
LGTM

PR #2423, opened 2020-05-20
Fix setting some systemd limits, add more tests

kolyshkin:
This fixes a few systemd cgroup bugs for both v1 and v2, and adds tests for systemd limit values (which uncover the bugs being fixed).

In particular, the bug fixes are:
 * allow to set `TasksMax=infinity` (aka `pids.limit: -1`) [v1 and v2]
 * set `MemoryLow` (aka `memory.reservation`) [v2 only]
 * update `CPUQuotaPerSecUSec` for some cases [v1 only]

The integration test improvements are:
 * simplify `check_systemd_value` usage
 * add missing checks for systemd's `MemoryMax` / `MemoryLimit`
 * add checks for systemd's `MemoryLow` and `MemorySwapMax`
 * add checks for systemd's `TasksMax`
 * add cgroupv1 systemd CPU checks (`CPUShares` and `CPUQuotaPerSecUSec`)
 * add a test that sets `pids.limit=-1`

### DIFF
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -63,6 +63,7 @@ EOF
     case $CGROUP_UNIFIED in
     no)
         MEM_LIMIT="memory.limit_in_bytes"
+        SD_MEM_LIMIT="MemoryLimit"
         MEM_RESERVE="memory.soft_limit_in_bytes"
         MEM_SWAP="memory.memsw.limit_in_bytes"
         SYSTEM_MEM=$(cat "${CGROUP_MEMORY_BASE_PATH}/${MEM_LIMIT}")
@@ -70,6 +71,7 @@ EOF
         ;;
     yes)
         MEM_LIMIT="memory.max"
+        SD_MEM_LIMIT="MemoryMax"
         MEM_RESERVE="memory.low"
         MEM_SWAP="memory.swap.max"
         SYSTEM_MEM="max"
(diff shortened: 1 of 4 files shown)


kolyshkin:
hmm, is there no CI for drafts?

kolyshkin:
This one is ready. There might be more similar work like that needed but this part is done.

kolyshkin:
@AkihiroSuda @mrunalp PTAL

kolyshkin:
@lifubang PTAL

tests/integration/update.bats
```diff
@@ -285,7 +331,10 @@ EOF
     [ "$status" -eq 0 ]
     check_cgroup_value "cpu.cfs_period_us" 1000000
     check_cgroup_value "cpu.cfs_quota_us" 500000
+    check_systemd_value CPUQuotaPerSecUSec 500ms
+
     check_cgroup_value "cpu.shares" 100
+    check_systemd_value CPUShares 100
```
AkihiroSuda, line 337: nit: quote everything consistently?

kolyshkin: yeah, I'd rather have all these useless `""` removed but yeah, we can have them everywhere.

will fix

kolyshkin: fixed

AkihiroSuda:
LGTM

mrunalp:
LGTM

PR #2426, opened 2020-05-20
Fix some cases of swap setting

kolyshkin:
This fixes two nuances in memory+swap update handling, and finally enables cgroupv2 swap checks in `tests/integration/update.bats`.

### cgroupv1: don't ignore MemorySwap if Memory==-1
    
Commit 18ebc51b3cc3 "Reset Swap when memory is set to unlimited (-1)"
(see PR #1127) added handling of the case when a user updates the container
limits to set memory to unlimited (-1) but do not set any other limits.
Apparently, in this case, if swap limit was previously set, kernel fails
to set `memory.limit_in_bytes` to -1 if `memory.memsw.limit_in_bytes` is
not set to -1.
    
What the above commit fails to handle correctly is the request when
Memory is set to -1 and MemorySwap is set to some specific limit N
(where N > 0). In this case, the value of N is silently discarded
and MemorySwap is set to -1 instead.
    
This is wrong thing to do, as the limit set, even if incorrectly,
should not be ignored.
    
Fix this by only assigning MemorySwap == -1 in case it was not
explicitly set.

### cgroupv2: set mem+swap to max if mem set to max
    
... and mem+swap is not explicitly set otherwise.
    
This ensures compatibility with cgroupv1 controller which interprets
things this way. Previous attempts to fix this: #2285, #2405.
    
With this fixed, we can finally enable swap tests for cgroupv2.

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -74,9 +74,9 @@ func (s *MemoryGroup) Apply(d *cgroupData) (err error) {
 }
 
 func setMemoryAndSwap(path string, cgroup *configs.Cgroup) error {
-	// If the memory update is set to -1 we should also
-	// set swap to -1, it means unlimited memory.
-	if cgroup.Resources.Memory == -1 {
+	// If the memory update is set to -1 and the swap is not explicitly
+	// set, we should also set swap to -1, it means unlimited memory.
+	if cgroup.Resources.Memory == -1 && cgroup.Resources.MemorySwap == 0 {
 		// Only set swap if it's enabled in kernel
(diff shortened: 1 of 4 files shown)


kolyshkin:
@lifubang @cyphar @AkihiroSuda PTAL

kolyshkin:
rebased

kolyshkin:
ah, got a typo while rebasing... fixed now

kolyshkin:
@AkihiroSuda @mrunalp PTAL

mrunalp:
LGTM

AkihiroSuda:
LGTM

Issue #2396, opened 2020-05-11
add `shfmt` to CI


AkihiroSuda:
We should validate indentation issues of shell scripts (mostly bats) in CI
https://github.com/moby/moby/blob/master/hack/validate/shfmt

kolyshkin:
This is going to be a problem with bats files, and bats is not exactly shell. In particular, it complains like this:

> tests/integration/create.bats:25:1: "}" can only be used to close a block

since it does not recognize

> @test "runc create" {

as a function.

Surely we can find some workaround such as replacing `@test "some thing"` lines with `function some_thing()` and then replacing it back....

cyphar:
This is why I'm personally a bit worried about re-adding `shfmt` to the CI -- the "shell files" we want to test are actually not normal shell programs, and in addition we've had problems with `shfmt` changing their "best practice" rules in the past (breaking the CI). This might not be a problem anymore, but we should keep it in mind.

kolyshkin:
FIled https://github.com/mvdan/sh/issues/600

PR #2428, opened 2020-05-21
Cpu quota fixes

kolyshkin:
This fixes a few bugs wrt setting cpu quota, including a recent regression, adds more tests for cpu quota, and (hopefully) simplifies the code handling all this.

The bugs being fixed are:

1. The regression introduced by commit 06d7c1d2616052974e (PR #2423), which made
   it impossible to only set CpuQuota (without the CpuPeriod). Reported by @haircommander
    
2. ~~The discrepancy in how various cgroup drivers interpret the case of
       missing CpuPeriod. Before this, the fs2 driver assumed the default,
       while the fs driver assumed the previously set value. Now the
       behavior is the same.~~ I was wrong, all drivers assume previously set value.

3. An invalid configuration (CpuPeriod set without CpuQuota) was not rejected
  by some drivers.

4. The fs2 driver ignored `--cpu-quota=-1` when CpuPeriod was not set.

The test cases added are

1. CPU quota/period/weight/shares tests **for cgroup v2** :tada:

2. Unlimited CPU quota test.

### DIFF
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -269,25 +269,22 @@ EOF
     check_cgroup_value "cpu.shares" 100
     check_systemd_value "CPUShares" 100
 
-    # systemd driver does not allow to update quota and period separately
-    if [ -z "$RUNC_USE_SYSTEMD" ]; then
-        # update cpu period
-        runc update test_update --cpu-period 900000
-        [ "$status" -eq 0 ]
-        check_cgroup_value "cpu.cfs_period_us" 900000
+    # updating cpu period alone is not allowed
+    runc update test_update --cpu-period 900000
+    [ "$status" -eq 1 ]
 
-        # update cpu quota
-        runc update test_update --cpu-quota 600000
-        [ "$status" -eq 0 ]
-        check_cgroup_value "cpu.cfs_quota_us" 600000
-    else
-        # update cpu quota and period together
-        runc update test_update --cpu-period 900000 --cpu-quota 600000
-        [ "$status" -eq 0 ]
-        check_cgroup_value "cpu.cfs_period_us" 900000
-        check_cgroup_value "cpu.cfs_quota_us" 600000
-        check_systemd_value "CPUQuotaPerSecUSec" 670ms
-    fi
+    # update cpu quota
+    runc update test_update --cpu-quota 600000
+    [ "$status" -eq 0 ]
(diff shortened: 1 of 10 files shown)


kolyshkin:
First three (preparatory) commits: CI :heavy_check_mark: 

Adding more

kolyshkin:
This one is almost ready. Added missing tests for removing CPU quota.

kolyshkin:

tests/integration/update.bats
```diff
@@ -245,50 +245,79 @@ EOF
...
+	[ "$status" -eq 1 ]
 
-        # update cpu quota
-        runc update test_update --cpu-quota 600000
+	# update cpu quota
+	runc update test_update --cpu-quota 600000
+	[ "$status" -eq 0 ]
+	check_cpu_quota 600000 1000000 "600ms"
+
+        # remove cpu quota
```
AkihiroSuda, line 299: nit: indent

AkihiroSuda:
The indentation seems broken. Other parts look good.

tests/integration/update.bats
kolyshkin: thanks for catching this -- fixed.

NOTE that (as I noted in the changelog message) I have converted this code (and the code I've added) to use tab instead of 4 spaces since I like tabs much more than 4 spaces, and I am changing lots of code here now.

Long term, I don't know what to do about it. I want to avoid commits just changing the whitespace as this will make it harder to look through git history, but I don't like the current way of using 4 tabs either.

mrunalp:
LGTM

tests/integration/update.bats
mrunalp: Is there any tool we can run to ensure consistency?

cyphar: `shfmt` solves this problem (and we had it in CI for a few months a while ago) but see #2396 which tracks re-adding it.

mrunalp:
Needs rebase.

kolyshkin:
Rebased. I am also going to separate the regression fix out of this

kolyshkin:
> Rebased. I am also going to separate the regression fix out of this.

See #2434

kolyshkin:
container update from the command line is a mess; need more time to fix it

kolyshkin:
rebased on top of #2434

kolyshkin:
Going to rework this in a slightly less dramatic fashion :)

PR #2429, opened 2020-05-22
[do not merge] cpu quota tests

kolyshkin:
These are updated CPU quota tests extracted from #2428 without the fixes.

The sole purpose of this PR is to run CI.

### DIFF
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -245,50 +245,79 @@ EOF
     check_systemd_value "TasksMax" 20
 }
 
-@test "update cgroup v1 cpu limits" {
-    [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
-    requires cgroups_v1
+function check_cpu_quota() {
+	local quota=$1
+	local period=$2
+	local sd_quota=$3
 
-    # run a few busyboxes detached
-    runc run -d --console-socket $CONSOLE_SOCKET test_update
-    [ "$status" -eq 0 ]
+	if [ "$CGROUP_UNIFIED" = "yes" ]; then
+		if [ "$quota" = "-1" ]; then
+			quota="max"
+		fi
+		check_cgroup_value "cpu.max" "$quota $period"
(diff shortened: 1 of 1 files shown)


kolyshkin:
First found bug: `--cpu-period` is allowed by itself.

```
not ok 78 update cgroup cpu limits
3930# (in test file tests/integration/update.bats, line 292)
3931#   `[ "$status" -eq 1 ]' failed
```

the test code is
```sh
# updating cpu period alone is not allowed
runc update test_update --cpu-period 900000
[ "$status" -eq 1 ]
```

disabling this case

kolyshkin:
Second found bug is a regression introduced by commit 06d7c1d (PR #2423), which made
it impossible to only set CpuQuota (without the CpuPeriod). 

```
not ok 78 update cgroup cpu limits
# (in test file tests/integration/update.bats, line 296)
...
# runc update test_update --cpu-quota 600000 (status=1):
438# time="2020-05-22T22:13:46Z" level=warning msg="Setting back cgroup configs failed due to error: CPU quota and period should both be set, your state.json and actual configs might be inconsistent."
439# time="2020-05-22T22:13:46Z" level=error msg="CPU quota and period should both be set"
```

disabling it

kolyshkin:
Weird, I only found two bugs :)

eero-t:
> Weird, I only found two bugs :)

Kolyshkin, I think you missed one more bug from the above output:
> ```
> # runc update test_update --cpu-quota 600000 (status=1):
> 438# time="2020-05-22T22:13:46Z" level=warning msg="Setting back cgroup configs failed due to error: CPU quota and period should both be set, your state.json and actual configs might be inconsistent."
> 439# time="2020-05-22T22:13:46Z" level=error msg="CPU quota and period should both be set"
> ```

Based on the RunC code: https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L227

When trying to set a non-working config, I think you should get just the error, not the warning.  Restoring the original, working config should succeed!

(I bumped into this when adding to "update" command support for WIP functionality modeled on how BlockIO device limits work.)

PR #2434, opened 2020-05-26
cgroupv1: minimal fix for cpu quota regression

kolyshkin:
This is a quick-n-dirty fix the regression introduced by commit
06d7c1d (in PR #2423), which made it impossible to only set CpuQuota
(without the CpuPeriod). It partially reverts the above commit,
and adds a test case.

The proper fix is in #2428.


mrunalp:
LGTM

AkihiroSuda:
LGTM

PR #2437, opened 2020-05-28
cgroups/fs2: make removeCgroupPath faster

kolyshkin:
1. In case there are no sub-cgroups (which is probably a common case), a single rmdir should be faster than iterating through the list of files (as suggested at https://github.com/opencontainers/runc/pull/2412#discussion_r428376762)
    
2. Use unix.Rmdir() to save one more syscall since os.Remove() tries unlink(2) first which fails on a directory, and only then tries rmdir(2).
    
3. Re-use rmdir in the second remove attempt.

And some test case fixups:

1. __runc does not set $status, so the check is misleading.

2. Add set +eux to the nest.sh script so we can error out early, and see
   what is going on.

3. Doing "echo +io" > cgroup.controllers is giving an error on my
   machine ("sh: write error: Operation not supported"). It is probably
   fine to just enable pids controller.

4. Add status check for runc exec nest.sh

5. Remove the second check for cgroup.threads contents -- it was already
   checked earlier (the output of nest.sh script).

@lifubang PTAL

### DIFF
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -11,6 +11,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 )
 
 type manager struct {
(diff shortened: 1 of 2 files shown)


mrunalp:
LGTM

AkihiroSuda:
LGTM

PR #2443, opened 2020-05-30
cgroupv1/systemd.Set: don't enable kernel memory acct

kolyshkin:
This is a regression from commit 1d4ccc8e0 (PR #2343). We only need to enable
kernel memory accounting once, from the `(*legacyManager).Apply()`,
and there is no need to do it in `(*legacyManager).Set()`.

While at it, rename the method to better reflect what it's doing.

This saves us 1 call to mountinfo parser.

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -119,13 +119,6 @@ func genV1ResourcesProperties(c *configs.Cgroup) ([]systemdDbus.Property, error)
 			newProp("TasksMax", uint64(c.Resources.PidsLimit)))
 	}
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Can we have a test?

kolyshkin:
> Can we have a test?

I am not sure what to test here, or how to test it. What this commit fixes is extra code (to enable kmem acct) that was executed from `(*legacyManager).Set()`. There is no bug here, just some operation being performed which is useless.

Frankly speaking, I think it might make sense to disable kmem altogether, since the kmem controller is not actually working properly in the kernel (no kernel memory reclamation etc.) and this is the reason why it is removed from cgroup v2. But that's a totally different issue.

PR #2445, opened 2020-06-01
Replace sed with jq for more readable json manipulation in tests

### LINKED ISSUE #2417: tests: stop using sed for manipulating JSON
https://github.com/opencontainers/runc/blob/f369199ff61bafdb91b5308d8b30cb4c18270bd9/tests/integration/helpers.bash#L155-L158

The current sed scripts are hard to maintain, should be substituted with `jq` as in: https://github.com/opencontainers/runc/blob/f369199ff61bafdb91b5308d8b30cb4c18270bd9/tests/integration/helpers.bash#L87-L95

JFHwang:
Signed-off-by: John Hwang <John.F.Hwang@gmail.com>

Replaced sed in /tests/integration with jq for json manipulation.

An assumption I made in this refactor is that order of elements in an array doesn't matter. 

Also, I translated a few lines literally because I'm unsure of the nature of config.json.
ex) there is a recursive search for the key "gid" and "uid" because I'm unsure of whether they occur on multiple lines of a config.json, or just at the .process.user field.

Closes #2417

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -21,15 +21,7 @@ function setup() {
     set_cgroups_path "$BUSYBOX_BUNDLE"
 
     # Set some initial known values
-    DATA=$(cat <<-EOF
-    "memory": {
-        "kernel": 16777216,
-        "kernelTCP": 11534336
-    },
-EOF
-    )
-    DATA=$(echo ${DATA} | sed 's/\n/\\n/g')
-    sed -i "s/\(\"resources\": {\)/\1\n${DATA}/" ${BUSYBOX_BUNDLE}/config.json
+    update_config '.linux.resources.memory |= {"kernel": 16777216, "kernelTCP": 11534336}' ${BUSYBOX_BUNDLE}
 
     # run a detached busybox to work with
     runc run -d --console-socket $CONSOLE_SOCKET test_cgroups_kmem
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -82,27 +82,27 @@ function runc_spec() {
 	fi
 }
 
+# Helper function to reformat config.json file. Input uses jq syntax.
+function update_config() {
+	bundle="${2:-.}"
+	jq "$1" "$bundle/config.json" | awk 'BEGIN{RS="";getline<"-";print>ARGV[1]}' "$bundle/config.json"
+}
+
 # Shortcut to add additional uids and gids, based on the values set as part of
 # a rootless configuration.
 function runc_rootless_idmap() {
 	bundle="${1:-.}"
-	cat "$bundle/config.json" \
-		| jq '.mounts |= map((select(.type == "devpts") | .options += ["gid=5"]) // .)' \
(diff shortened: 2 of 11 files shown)


tests/integration/helpers.bash
```diff
@@ -70,6 +70,7 @@ function runc_spec() {
 	fi
 
 	runc spec "${args[@]}"
+	echo "args are ${args[@]}" >> /tmp/mylog
```
kolyshkin, line 73: forgotten debug hunks?

tests/integration/tty.bats
```diff
@@ -29,11 +42,20 @@ function teardown() {
...
+	mv "config.json"{.tmp,}
+
+
 
 	# run busybox
 	runc run test_busybox
 	[ "$status" -eq 0 ]
+	echo "Output is: $output" > /tmp/mylog
+	echo "line 0 is: ${lines[0]}" >> /tmp/mylog
+	echo "line 1 is: ${lines[1]}" >> /tmp/mylog
```
kolyshkin, line 57: ditto

tests/integration/helpers.bash
JFHwang: ooh. haha, yep

tests/integration/cgroups.bats
```diff
@@ -21,15 +21,10 @@ function setup() {
...
-        "kernelTCP": 11534336
-    },
-EOF
-    )
-    DATA=$(echo ${DATA} | sed 's/\n/\\n/g')
-    sed -i "s/\(\"resources\": {\)/\1\n${DATA}/" ${BUSYBOX_BUNDLE}/config.json
+    cat "${BUSYBOX_BUNDLE}/config.json" \
+	| jq '.linux.resources.memory |= {"kernel": 16777216, "kernelTCP": 11534336}' \
+	> "${BUSYBOX_BUNDLE}/config.json.tmp"
+    mv "${BUSYBOX_BUNDLE}/config.json"{.tmp,}
```
cyphar, line 27: if we're going to make this a pattern, can we use `sponge` so you don't have to have this `mv` step each time? `sponge` is like `tee` except it waits until EOF to overwrite the file (thus allowing you to replace files that are used as input to the pipeline).

cyphar: It also kind of feels like `cat | jq | sponge foo` should be a function or something.

cyphar:
I've been wanting a change like this for a while, thanks for working on it. My only point is that I'd like it to be a little bit more ergonomic (by using `sponge` and possibly making the `cat | jq | sponge` boilerplate a function). We're only ever going to use this to modify `config.json` so you can just hardcode that path.

tests/integration/cgroups.bats
JFHwang: Ok no problem! I'll look into it.

kolyshkin: There is a relevant discussion in here. Also, `sponge` can be implemented as a one-liner awk (in case it's not readily available on some distro we test on): https://github.com/stedolan/jq/issues/105

kolyshkin: Also, `cat` is not needed, jq is capable of reading from a file.

AkihiroSuda: +1 for sponge (`moreutils`), it seems universally available on all distros

JFHwang: sponge doesn't seem to work with the CI tests. I'm going to use the one-liner awk.

PR #2451, opened 2020-06-02
rootless: allow /dev/{tty,ptmx} to be present in linux.devices

### LINKED ISSUE #2450: [rootless+"privileged"] creating device nodes caused: open /..<snipped>../dev/tty: no such device or address
On rootless+cgroup2(systemd)+CRI, creating a "privileged" container (which is not really privileged) fails with `creating device nodes caused: open /..<snipped>.../dev/tty: no such device or address`.

This error is not reproducible with `docker (podman) run --privileged`.

The repro step is same as https://github.com/containers/crun/issues/382 but the error is different.

## Repro

https://github.com/AkihiroSuda/critest-rootless-cgroup2/tree/v0.0.3

runc dbe5acade3d0f84b5c1bd1184b1bbf5a533cdfa1

Ubuntu 20.04

```console
$ git clone https://github.com/AkihiroSuda/critest-rootless-cgroup2.git
$ cd critest-rootless-cgroup2
$ git checkout v0.0.3
$ ./bin-download.sh
$ cp -f /usr/local/sbin/runc ./bin/crun
$ ./containerd.sh
```

```console
$ vi pod-config.json container-config.json 
$ git diff
diff --git a/container-config.json b/container-config.json
index fd33c47..0e4401b 100644
--- a/container-config.json
+++ b/container-config.json
@@ -11,7 +11,7 @@
     "log_path":"my-container.log",
     "linux": {
         "security_context": {
-            "privileged": false
+            "privileged": true
         },
         "resources": {
             "cpu_shares": 128,
diff --git a/pod-config.json b/pod-config.json
index 92b8c62..2dd0548 100644
--- a/pod-config.json
+++ b/pod-config.json
@@ -9,7 +9,7 @@
     "linux": {
         "cgroup_parent": "user.slice",
         "security_context": {
-            "privileged": false
+            "privileged": true
         }
     }
 }
$ ./run.sh runc
$ ./run.sh runc
+ RUNTIME=runc
++ ./crictl.sh pods -q
+ [[ -n '' ]]
++ ./crictl.sh run --runtime=runc container-config.json pod-config.json
FATA[0002] Running container failed: starting the container "6451ca158f106fc8dd68f168993671ef60bbe195b2cd35e997f368270d65778a" failed: rpc error: code = Unknown desc = failed to create containerd task: OCI runtime create failed: container_linux.go:353: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:70: creating device nodes caused: open /run/.ro813277178/user/1001/containerd/io.containerd.runtime.v2.task/k8s.io/6451ca158f106fc8dd68f168993671ef60bbe195b2cd35e997f368270d65778a/rootfs/dev/tty: no such device or address: unknown 
+ id=
```

AkihiroSuda:
Fix #2450

Previously, rootless runc was failing when `.linux.devices` contains an entry for either `/dev/tty` or `/dev/ptmx`.

`/dev/tty`: `ENXIO` (no such device or address)
```console
$ rootlesskit runc run foo
WARN[0000] signal: killed                               
ERRO[0000] container_linux.go:353: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:70: creating device nodes caused: open /home/suda/tmp/runctest/rootfs/dev/tty: no such device or address 
[rootlesskit:child ] error: command [runc run foo] exited: exit status 1
[rootlesskit:parent] error: child exited: exit status 1
```

`/dev/ptmx`: `EBUSY` (device or resource busy)
```console
$ rootlesskit runc run foo
WARN[0000] signal: killed                               
ERRO[0000] container_linux.go:353: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:73: setting up ptmx caused: remove /home/suda/tmp/runctest/rootfs/dev/ptmx: device or resource busy 
[rootlesskit:child ] error: command [runc run foo] exited: exit status 1
[rootlesskit:parent] error: child exited: exit status 1
```

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -21,6 +21,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/system"
 	libcontainerUtils "github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/selinux/go-selinux/label"
+	"github.com/pkg/errors"
 
 	"golang.org/x/sys/unix"
 )
@@ -594,6 +595,36 @@ func createDevices(config *configs.Config) error {
 }
 
 func bindMountDeviceNode(dest string, node *configs.Device) error {
+	var st unix.Stat_t
+	if err := unix.Lstat(dest, &st); err != nil {
+		if !os.IsNotExist(err) {
+			return errors.Wrapf(err, "could not stat existing path for device node: %q", dest)
+		}
+	} else if st.Mode&unix.S_IFMT == unix.S_IFLNK {
+		// If the path is a symlink, we remove it.
+		if err := unix.Unlink(dest); err != nil {
+			return errors.Wrapf(err, "device node %q already exists and is a symlink", dest)
+		}
+	} else {
+		// If there already exists a device inode with the correct (type,
+		// major, minor), we can skip over setting it up. This logic only
+		// really makes sense for rootless containers.
+		var gotType configs.DeviceType
+		switch st.Mode & unix.S_IFMT {
+		case unix.S_IFBLK:
+			gotType = configs.BlockDevice
+		case unix.S_IFCHR:
+			gotType = configs.CharDevice
+		default:
+			return errors.Errorf("device node %q already exists but not a valid device: %#o", dest, st.Mode)
+		}
+		gotMajor := int64(unix.Major(st.Rdev))
+		gotMinor := int64(unix.Minor(st.Rdev))
+		if gotType != node.Type || gotMajor != node.Major || gotMinor != node.Minor {
+			return errors.Errorf("device node %q already exists but incorrect type [%v %d:%d] (wanted [%v %d:%d])", dest, gotType, gotMajor, gotMinor, node.Type, node.Major, node.Minor)
+		}
+		return nil
+	}
 	f, err := os.Create(dest)
 	if err != nil && !os.IsExist(err) {
 		return err
@@ -726,7 +757,15 @@ func setReadonly() error {
 func setupPtmx(config *configs.Config) error {
 	ptmx := filepath.Join(config.Rootfs, "dev/ptmx")
 	if err := os.Remove(ptmx); err != nil && !os.IsNotExist(err) {
-		return err
+		if errors.Is(err, unix.EBUSY) {
+			// EBUSY can happen when ptmx is bind-mounted, mostly when rootless && linux.devices contains ptmx
+			if umErr := unix.Unmount(ptmx, unix.MNT_DETACH); umErr == nil {
+				err = os.Remove(ptmx)
+			}
+		}
+		if err != nil {
+			return err
+		}
 	}
 	if err := os.Symlink("pts/ptmx", ptmx); err != nil {
 		return fmt.Errorf("symlink dev ptmx %s", err)
--- a/tests/integration/tty.bats
+++ b/tests/integration/tty.bats
@@ -231,3 +231,10 @@ EOF
 	runc kill test_busybox KILL
 	[ "$status" -eq 0 ]
 }
+
+@test "runc run [tty and ptmx in devices (#2450)]" {
+	update_config '.process.terminal=true | .process.args=["true"] | .linux.devices = [{"path":"/dev/tty","type":"c","major":5,"minor":0,"fileMode":8630},{"path":"/dev/ptmx","type":"c","major":5,"minor":2,"fileMode":8630}]'
+	# run busybox
+	runc run test_busybox
+	[ "$status" -eq 0 ]
+}


libcontainer/rootfs_linux.go
```diff
@@ -594,6 +595,10 @@ func createDevices(config *configs.Config) error {
 }
 
 func bindMountDeviceNode(dest string, node *configs.Device) error {
+	if _, stErr := os.Stat(dest); stErr == nil {
+		// already exist (#2450)
+		return nil
```
cyphar, line 600: I don't like this, but if we're going to do it we should actually check that the node matches the configuration (that the type/major/minor is the same). In addition, this should almost certainly be `Lstat`.

AkihiroSuda: updated

AkihiroSuda:
@cyphar @kolyshkin PTAL

libcontainer/rootfs_linux.go
```diff
@@ -594,6 +596,40 @@ func createDevices(config *configs.Config) error {
...
+		if !ok {
+			return errors.Errorf("device node %q is already populated but not a valid device with *syscall.Stat_t: %+v", dest, st.Mode())
+		}
+		major := int64(unix.Major(ust.Rdev))
+		minor := int64(unix.Minor(ust.Rdev))
+		if major != node.Major || minor != node.Minor {
+			return errors.Errorf("device node %q is already populated but has unexpected number %d:%d (expected %d:%d)", dest, major, minor, node.Major, node.Minor)
+		}
+		return nil
+	}
```
cyphar, line 632: This code would be much better written something like this:

```suggestion
	var st unix.Stat_t
	if err := os.Lstat(dest, &st); err != nil {
		if !os.IsNotExist(err) {
			return errors.Wrapf(err, "could not stat existing path for device node: %q", dest)
		}
	} else {
		var gotType configs.DeviceType
		switch st.Mode() & os.ModeType {
		case os.ModeSymlink:
			// If the path is a symlink, we remove it.
			// NOTE: These semantics might be incorrect, but I'm not convinced
			//       that the semantics in the original PR are correct either.
			//       Resolving symlinks is a pretty dangerous game and I'd
			//       prefer not to do it if possible.
			if err := unix.Unlink(dest); err != nil {
				return errors.Wrapf(err, "device node %q already exists and is a symlink", dest)
			}
			goto create
		case os.ModeDevice:
			gotType = configs.BlockDevice
		case os.ModeDevice | os.ModeCharDevice:
			gotType = configs.CharDevice
		default:
			return errors.Errorf("device node %q already exists but not a valid device: %o", dest, st.Mode())
		}

		gotMajor := int64(unix.Major(st.Rdev))
		gotMinor := int64(unix.Minor(st.Rdev))
		if gotType != node.Type || gotMajor != node.Major || gotMinor != node.Minor {
			return errors.Errorf("device node %q already exists but incorrect type [%s %d:%d] (wanted [%s %d:%d])", dest, gotType, gotMajor, gotMinor, node.Type, node.Major, node.Minor)
		}
		return nil
	}
create:
```

Though I'm not sure what the right handling of symlinks is. Each option seems suboptimal to me...

libcontainer/rootfs_linux.go
```diff
@@ -726,7 +762,15 @@ func setReadonly() error {
 func setupPtmx(config *configs.Config) error {
 	ptmx := filepath.Join(config.Rootfs, "dev/ptmx")
 	if err := os.Remove(ptmx); err != nil && !os.IsNotExist(err) {
-		return err
+		if errors.Is(err, unix.EBUSY) {
+			// EBUSY can happen when ptmx is bind-mounted, mostly when rootless && linux.devices contains ptmx
+			if umErr := unix.Unmount(ptmx, 0); umErr == nil {
```
cyphar, line 767: You should have a `MNT_DETACH` here.

libcontainer/rootfs_linux.go
AkihiroSuda: Can we error out if a symlink path has '/', otherwise keep the link?

AkihiroSuda:
updated

libcontainer/rootfs_linux.go
AkihiroSuda: doesn't work on rootless
```
not ok 79 runc run [tty and ptmx in devices (#2450)]
# (in test file tests/integration/tty.bats, line 238)
# runc list (status=0):
# 
# runc run test_busybox (status=1):
# time="2020-06-05T10:29:46Z" level=warning msg="exit status 1"
# time="2020-06-05T10:29:46Z" level=error msg="container_linux.go:353: starting container process caused: process_linux.go:450: container init caused: rootfs_linux.go:71: creating device nodes caused: device node \"/tmp/busyboxtest/rootfs/dev/tty\" already exists but not a valid device: 20666"
```

AkihiroSuda: modified to pass rootless

libcontainer/rootfs_linux.go
```diff
@@ -594,6 +595,37 @@ func createDevices(config *configs.Config) error {
 }
 
 func bindMountDeviceNode(dest string, node *configs.Device) error {
+	var st unix.Stat_t
+	if err := unix.Lstat(dest, &st); err != nil {
+		if !os.IsNotExist(err) {
+			return errors.Wrapf(err, "could not stat existing path for device node: %q", dest)
+		}
+	} else if os.FileMode(st.Mode)&os.ModeSymlink == os.ModeSymlink {
```
h-vetinari, line 603:
```suggestion
	} else if os.FileMode(st.Mode) & os.ModeSymlink == os.ModeSymlink {
```

cyphar: We require all code to follow `gofmt`, which your suggestion doesn't (yeah, I also think it looks a bit silly but that's what the Go dev's thought the Go style should look like).

libcontainer/rootfs_linux.go
```diff
@@ -594,6 +595,37 @@ func createDevices(config *configs.Config) error {
...
+		case os.ModeDevice:
+			if node.Type != configs.BlockDevice {
+				return errors.Errorf("device node %q already exists but incorrect type %v (wanted block type)", dest, node.Type)
+			}
+		case os.ModeDevice | os.ModeCharDevice:
+			if node.Type != configs.CharDevice {
+				return errors.Errorf("device node %q already exists but incorrect type %v (wanted block type)", dest, node.Type)
+			}
+		default:
+			// NOP, happens on rootless
```
cyphar, line 619: This doesn't make sense (and is why I changed it in my suggestion) -- what you're saying here is "if it's any other kind of inode, ignore it unless `node.Major` and `node.Minor` are non-zero because if the inode isn't a device then it won't have a major/minor number". I don't understand why you would want to do that.

The actual bug (and it was unrelated to rootless containers) was that I forgot that `os.Mode*` has different values from the Linux ones. The actual switch statement should've looked like:

```
switch st.Mode & unix.S_IFMT {
case unix.S_IFBLK:
	gotType := configs.BlockDevice
case unix.S_IFCHR:
	gotType := configs.CharDevice
default:
	// I'm not sure if you want to also handle regular files as well?
	// I'm still a little confused what the semantics are supposed to be.
	return fmt.Errorf(...)
```

And you should change back the rest of the hunk too.

cyphar: Also (because of the misunderstanding I mentioned above), this code will ignore the type of the device because it will always go through the `default` case in the `switch`.

libcontainer/rootfs_linux.go
cyphar, line 603: This is also wrong (it has the same misunderstanding). This should be `st.Mode&unix.S_IFMT == unix.S_IFLNK`.

libcontainer/rootfs_linux.go
```diff
@@ -594,6 +595,37 @@ func createDevices(config *configs.Config) error {
...
+		}
```
cyphar, line 620: This doesn't really make sense (and is why I changed it in my suggestion) -- what you're saying here is "if it's any other kind of inode, ignore it unless `node.Major` and `node.Minor` are non-zero because if the inode isn't a device then it won't have a major/minor number". I don't understand why you would want to do that.

The actual bug (and it was unrelated to rootless containers) was that I forgot that `os.Mode*` has different values from the Linux ones. This fix should've actually looked like:

```go
var st unix.Stat_t
if err := unix.Lstat(dest, &st); err != nil {
	if !os.IsNotExist(err) {
		return errors.Wrapf(err, "could not stat existing path for device node: %q", dest)
	}
} else if st.Mode&unix.S_IFMT == unix.S_IFLNK {
	// If the path is a symlink, we remove it.
	// NOTE: These semantics might be incorrect, but I'm not convinced
	//       that the semantics in the original PR are correct either.
	//       Resolving symlinks is a pretty dangerous game and I'd
	//       prefer not to do it if possible.
	if err := unix.Unlink(dest); err != nil {
		return errors.Wrapf(err, "device node %q already exists and is a symlink", dest)
	}
} else {
	// If there already exists a device inode with the correct (type,
	// major, minor), we can skip over setting it up. This logic only
	// really makes sense for rootless containers.
	var gotType configs.DeviceType
	switch st.Mode & unix.S_IFMT {
	case unix.S_IFBLK:
		gotType = configs.BlockDevice
	case unix.S_IFCHR:
		gotType = configs.CharDevice
	default:
		return errors.Errorf("device node %q already exists but not a valid device: %#o", dest, st.Mode)
	}
	gotMajor := int64(unix.Major(st.Rdev))
	gotMinor := int64(unix.Minor(st.Rdev))
	if gotType != node.Type || gotMajor != node.Major || gotMinor != node.Minor {
		return errors.Errorf("device node %q already exists but incorrect type [%s %d:%d] (wanted [%s %d:%d])", dest, gotType, gotMajor, gotMinor, node.Type, node.Major, node.Minor)
	}
	return nil
}
```

And you should change back the rest of the hunk too. (This version of the change is also incorrect because of the above misunderstanding -- it will effectively always ignore the type of the device and only compare the major:minor numbers because the bits in the switch cases are incorrect.)

libcontainer/rootfs_linux.go
cyphar, line 603: My bad on this one as well. This is also wrong (it has the same misunderstanding). This should be `st.Mode&unix.S_IFMT == unix.S_IFLNK`. And you might argue that we should handle regular files here too, I don't know what the intended semantics are but if unlink symlinks maybe we should also unlink regular files? Personally I'd prefer if we didn't do this (assuming it doesn't break any containers).

libcontainer/rootfs_linux.go
AkihiroSuda: updated

libcontainer/rootfs_linux.go
AkihiroSuda: updated

AkihiroSuda:
Another solution might be to just ignore `tty` and `ptmx` in `linux.devices` when rootless.

@cyphar @giuseppe @kolyshkin RFC

mrunalp:
@AkihiroSuda I think we shouldn't block on this for next rc. wdyt?

AkihiroSuda:
@mrunalp Not a blocker for rc91, but good to have if possible. (IIUC rc91 is currently blocked due to other PRs)

AkihiroSuda:
ping @cyphar

AkihiroSuda:
@cyphar WDYT? https://github.com/opencontainers/runc/pull/2451#issuecomment-640415147

AkihiroSuda:
Moved to rc92 milestone

ctalledo:
Hi Akihiro, I was looking into this problem too.

It seems to me the problem is that runc has a set of default devices that it always configures in the container (/dev/null, /dev/tty, etc.) and if the OCI spec it receives has those same devices, it's not detecting the redundancy. 

As a result, it creates a libcontainer config with redundant device configs. This hasn't been a problem without rootless (where mknod is used to setup the devices). But in rootless it fails on /dev/tty (and /dev/ptmx apparently too) because it sets up the device by creating a regular file under /dev/tty and bind-mounting the host's /dev/tty to the container's /dev/tty. When this operation is done redundantly, it fails the second time.

I am thinking an alternative fix is to catch the redundant device configs when creating the libcontainer config, and removing the redundancy by favoring the device config that came in the OCI spec over the default device config. 

Something like this:

```diff
diff --git a/libcontainer/specconv/spec_linux.go b/libcontainer/specconv/spec_linux.go
index 6fd98f02..7046ae9c 100644
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go

 func createDevices(spec *specs.Spec, config *configs.Config) error {
-       // Add default set of devices.
-       for _, device := range AllowedDevices {
-               if device.Path != "" {
-                       config.Devices = append(config.Devices, device)
+
+       // If a spec device is redundant with a default device, remove that default
+       // device (the spec one takes priority).
+       dedupedAllowDevs := []*configs.Device{}
+
+       if spec.Linux != nil {
+               for _, ad := range AllowedDevices {
+                       conflict := false
+                       for _, sd := range spec.Linux.Devices {
+                               if sd.Path == ad.Path {
+                                       conflict = true
+                               }
+                       }
+                       if !conflict {
+                               dedupedAllowDevs = append(dedupedAllowDevs, ad)
+                               if ad.Path != "" {
+                                       config.Devices = append(config.Devices, ad)
+                               }
+                       }
                }
        }
+
+       AllowedDevices = dedupedAllowDevs
+
```

What do you think?

cyphar:
I think @ctalledo is right, though I think you could simplify the patch down to the following (modifying `AllowedDevices` doesn't really make sense).

```diff
diff --git a/libcontainer/specconv/spec_linux.go b/libcontainer/specconv/spec_linux.go
index 6fd98f027539..6322c5588154 100644
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -651,9 +651,18 @@ func stringToDeviceRune(s string) (configs.DeviceType, error) {
 }
 
 func createDevices(spec *specs.Spec, config *configs.Config) error {
-       // Add default set of devices.
+       // Add default set of devices, but only if the device is not already
+       // configured in the provided spec (in that case, the specfile takes
+       // precedence).
+add_devices:
        for _, device := range AllowedDevices {
                if device.Path != "" {
+                       for _, specDevice := range spec.Linux.Devices {
+                               // XXX: Probably should do CleanPath here or something.
+                               if device.Path == specDevice.Path {
+                                       continue add_devices
+                               }
+                       }
                        config.Devices = append(config.Devices, device)
                }
        }
```

AkihiroSuda:
@ctalledo @cyphar Would you open a PR?

ctalledo:
Thanks @cyphar. 

> though I think you could simplify the patch down to the following (modifying AllowedDevices doesn't really make sense).

Actually, it turns out AllowedDevices is used in CreateCgroupConfig(), which is called right after createDevices(). This is why I felt modifying allowed devices was the proper thing to do.

> Would you open a PR?

Hi @AkihiroSuda: sure, I'll do this tomorrow. 

Thanks!
-Cesar

cyphar:
@ctalledo 

> Actually, it turns out AllowedDevices is used in CreateCgroupConfig(), which is called right after createDevices(). This is why I felt modifying allowed devices was the proper thing to do.

The issue is that `AllowedDevices` is a global and while the primary use-case of runc is as a binary operating on a single container, some people do use `libcontainer` directly and modifying `AllowedDevices` will result in memory unsafety as well as generally incorrect behaviour (if you start two different containers the `AllowedDevices` change will affect both).

But if that's the case then we should probably update the logic to only make use of `AllowedDevices` once.

ctalledo:
@cyphar 

> The issue is that AllowedDevices is a global and while the primary use-case of runc is as a binary operating on a single container, some people do use libcontainer directly and modifying AllowedDevices will result in memory unsafety as well as generally incorrect behaviour (if you start two different containers the AllowedDevices change will affect both).

Got it, thanks.

> But if that's the case then we should probably update the logic to only make use of AllowedDevices once.

That's the right path I think. Let me tweak it a bit and open a new PR.

ctalledo:
Hi @cyphar, @AkihiroSuda,

FYI, the PR for the alternative fix for #2450 is here: https://github.com/opencontainers/runc/pull/2522

Thanks!

AkihiroSuda:
Thanks

PR #2453, opened 2020-06-02
CI: add CentOS 7 (kernel 3.10, systemd 219)

### LINKED ISSUE #2444: Add Vagrant CI for several distros (e.g. CentOS 7)
- [X] Ubuntu 18.04 (Travis)
- [X] Fedora 32 (Vagrant on Travis)
- [x] CentOS 7 (Vagrant on Travis): https://github.com/opencontainers/runc/pull/2453

AkihiroSuda:
This PR ensures runc to be runnable on CentOS 7 with kernel 3.10, which is (unofficially) the oldest host we support.

Rootless tests are not covered in this PR and should be covered in follow-up PRs.

Part of  #2444

~~Contains #2454~~


kolyshkin:
Can you please add a description of what do you have in mind when adding it? It would help.

We might also need to add a semi-recent Ubuntu, as I recently found out we're not running the tests on a system with no mem+swap limit (https://github.com/opencontainers/runc/pull/2439).

Note that CentOS 7 has this controller (see e.g. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-modifying_control_groups)

.travis.yml
```diff
@@ -40,6 +35,19 @@ matrix:
...
+        - ln -sf Vagrantfile.centos7 Vagrantfile
+        - sudo vagrant up && sudo mkdir -p /root/.ssh && sudo sh -c "vagrant ssh-config >> /root/.ssh/config"
+      script:
+        # kernel 3.10 (frankenized)
+        - sudo ssh default 'uname -a'
+        - sudo ssh default -t 'sudo -i -- sh -c "cd /vagrant && make localunittest"'
+        - sudo ssh default -t 'sudo -i -- sh -c "cd /vagrant && make localintegration RUNC_USE_SYSTEMD=yes"'
+        - sudo ssh default -t 'sudo -i -- sh -c "cd /vagrant && make localintegration"'
+        - sudo ssh default -t 'sudo -i -- sh -c "cd /vagrant && make localrootlessintegration"'
+
```
kolyshkin, line 50: Maybe create a separate target for these four and use it? Something like `localtest`?

Funny, we already have one, it just needs a variant for systemd :)

tests/integration/spec.bats
```diff
@@ -79,15 +79,15 @@ function teardown() {
 
   SPEC_REF=$([[ -z "$SPEC_COMMIT" ]] && echo $SPEC_VERSION || echo $SPEC_COMMIT)
 
-  run git -C src/runtime-spec reset --hard "${SPEC_REF}"
+  run bash -c "cd src/runtime-spec && git reset --hard ${SPEC_REF}"
```
kolyshkin, line 82: What is the reason for this change? I guess old(er) git does not support `-C`?

AkihiroSuda: Yes, CentOS 7 is shipped with an ancient version of git (`git version 1.8.3.1`).

AkihiroSuda:
> Can you please add a description of what do you have in mind when adding it? It would help.

Done. This PR ensures runc to be runnable on CentOS 7 with kernel 3.10, which is (unofficially) the oldest host we support.

.travis.yml
AkihiroSuda: should be another PR perhaps

tests/integration/update.bats
```diff
@@ -133,7 +136,7 @@ EOF
         runc update test_update --memory-swap -1
         [ "$status" -eq 0 ]
         check_cgroup_value "$MEM_SWAP" $SYSTEM_MEM
-        check_systemd_value "$SD_MEM_SWAP" $SD_UNLIMITED
+        check_systemd_value "$SD_MEM_SWAP" $SD_UNLIMITED || check_systemd_value "$SD_MEM_SWAP" $SD_UNLIMITED_LEGACY
```
kolyshkin, line 139: This looks bad :(

Can we add a runtime check and set `SD_UNLIMITED` to the expected value?

Something like

```bash
SD_UNLIMITED="infinity"
# 
SD_VERSION=$(systemctl --version | awk '{print $2; exit}'
if [ "$SD_VERSION -lt 227 ]; then
   SD_UNLIMITED="18446744073709551615"
fi
```

tests/integration/update.bats
```diff
@@ -175,7 +178,7 @@ EOF
     runc update test_update --pids-limit -1
     [ "$status" -eq 0 ]
     check_cgroup_value "pids.max" max
-    check_systemd_value "TasksMax" $SD_UNLIMITED
+    check_systemd_value "TasksMax" $SD_UNLIMITED || check_systemd_value "TasksMax" $SD_UNLIMITED_LEGACY
```
kolyshkin, line 181: Please fix the commit description, it currently says this systemd change only affects `MemoryLimit`, but apparently it is also for `TasksMax` and `MemorySwapMax` (at least).

tests/integration/update.bats
AkihiroSuda: updated

tests/integration/update.bats
AkihiroSuda: updated

AkihiroSuda:
@kolyshkin @mrunalp PTAL

AkihiroSuda:
ping @kolyshkin @mrunalp

tests/rootless.sh
```diff
@@ -146,7 +146,7 @@ do
 		# https://bugzilla.redhat.com/show_bug.cgi?id=1788616
 	        ssh -t -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i $HOME/rootless.key rootless@localhost -- PATH="$PATH" RUNC_USE_SYSTEMD="$RUNC_USE_SYSTEMD" bats -t "$ROOT/tests/integration$ROOTLESS_TESTPATH"
 	else
-		sudo -HE -u rootless PATH="$PATH" bats -t "$ROOT/tests/integration$ROOTLESS_TESTPATH"
+		sudo -HE -u rootless PATH="$PATH" $(which bats) -t "$ROOT/tests/integration$ROOTLESS_TESTPATH"
```
kolyshkin, line 149: Just wondering if we still need PATH being set

PR #2457, opened 2020-06-04
cgroup2: fix `runc update --cpu-quota` w/o `--cpu-period`

### LINKED ISSUE #2456: cgroup2: `runc update --cpu-quota` w/o `--cpu-period` should keep the current period
`runc update --cpu-quota` w/o `--cpu-period`  is currently broken:

```console
$ git diff 0853956d235d20c328a4173b411e218900b81d48
diff --git a/tests/integration/update.bats b/tests/integration/update.bats
index 08736e42..8dff784b 100644
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -354,6 +354,31 @@ EOF
     check_systemd_value "CPUShares" 100
 }
 
+@test "update cgroup v2 cpu limits" {
+    requires cgroups_v2
+
+    # run a few busyboxes detached
+    runc run -d --console-socket $CONSOLE_SOCKET test_update
+    [ "$status" -eq 0 ]
+
+    # check that initial values were properly set
+    # period has six zeros.
+    check_cgroup_value "cpu.max" "500000 1000000"
+
+    # update cpu quota without explicit period
+    runc update test_update --cpu-quota 600000
+    [ "$status" -eq 0 ]
+    # make sure the initial period (six zeros) is preserved
+    check_cgroup_value "cpu.max" "600000 1000000"
+
+    # update cpu quota to -1 without explicit period
+    runc update test_update --cpu-quota -1
+    [ "$status" -eq 0 ]
+    # make sure quota is set to "max" and the initial period (six zeros) is preserved
+    check_cgroup_value "cpu.max" "max 1000000"
+}
+
 @test "update rt period and runtime" {
     requires cgroups_rt

$ sudo PATH=$PATH bats tests/integration/update.bats -f "update cgroup v2 cpu limits"
 ✗ update cgroup v2 cpu limits
   (from function `check_cgroup_value' in file tests/integration/helpers.bash, line 177,
    in test file tests/integration/update.bats, line 373)
     `check_cgroup_value "cpu.max" "600000 1000000"' failed
   runc list (status=0):
   runc list (status=0):
   runc list (status=0):
   runc spec (status=0):
   
   runc run -d --console-socket /tmp/console.sock test_update (status=0):
   
   /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/cpu.max
   current 500000 1000000 !? 500000 1000000
   runc update test_update --cpu-quota 600000 (status=0):
   
   /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup/cpu.max
   current 600000 100000 !? 600000 1000000
   runc list (status=0):
   ID            PID         STATUS      BUNDLE             CREATED                          OWNER
   test_update   117914      running     /tmp/busyboxtest   2020-06-03T16:36:45.591645549Z   root
   runc kill test_update KILL (status=0):
   
   runc delete test_update (status=0):
   
   runc list (status=0):
   runc list (status=0):

1 test, 1 failure
```

---

The comparison of `600000 100000 !? 600000 1000000` is failing.
(`cpu.max` is expected to have initially specified period=1,000,000 with six zeros, but the current implementation resets the period to the kernel default 100,000 with five zeros)

AkihiroSuda:
Fix #2456

`ConvertCPUQuotaCPUPeriodToCgroupV2Value()` is now called in the fs2 manager, not in the CLI module, because the conversion requires statting the previous "cpu.max" value.

related test: `TestUpdateCPUQuota` of Moby: https://github.com/moby/moby/blob/89382f2f20745b9e63bed6c066f104980dff4396/integration/container/update_linux_test.go#L86

### DIFF
--- a/libcontainer/cgroups/fs2/cpu.go
+++ b/libcontainer/cgroups/fs2/cpu.go
@@ -4,17 +4,21 @@ package fs2
 
 import (
 	"bufio"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/pkg/errors"
 )
 
 func isCpuSet(cgroup *configs.Cgroup) bool {
-	return cgroup.Resources.CpuWeight != 0 || cgroup.Resources.CpuMax != ""
+	return cgroup.Resources.CpuWeight != 0 || cgroup.Resources.CpuQuota != 0 || cgroup.Resources.CpuPeriod != 0
 }
 
 func setCpu(dirPath string, cgroup *configs.Cgroup) error {
@@ -29,15 +33,56 @@ func setCpu(dirPath string, cgroup *configs.Cgroup) error {
 		}
 	}
 
-	// NOTE: .CpuQuota and .CpuPeriod are not used here. Conversion is the caller's responsibility.
-	if cgroup.Resources.CpuMax != "" {
-		if err := fscommon.WriteFile(dirPath, "cpu.max", cgroup.Resources.CpuMax); err != nil {
+	// Convert .CpuQuota and .CpuPeriod into "cpu.max".
+	// Conversion requires the previous value of "cpu.max"
+	prevCpuMaxBytes, err := ioutil.ReadFile(filepath.Join(dirPath, "cpu.max"))
+	if err != nil {
+		return err
+	}
+	cpuMax, err := ConvertCPUQuotaCPUPeriodToCgroupV2Value(
+		strings.TrimSpace(string(prevCpuMaxBytes)),
+		cgroup.Resources.CpuQuota, cgroup.Resources.CpuPeriod)
+	if err != nil {
+		return err
+	}
+	if cpuMax != "" {
+		if err := fscommon.WriteFile(dirPath, "cpu.max", cpuMax); err != nil {
 			return err
 		}
 	}
-
 	return nil
 }
+
+// ConvertCPUQuotaCPUPeriodToCgroupV2Value generates cpu.max string.
+func ConvertCPUQuotaCPUPeriodToCgroupV2Value(prevCpuMax string, quota int64, period uint64) (string, error) {
+	if quota == 0 && period == 0 {
+		return prevCpuMax, nil
+	}
+
+	prevPair := strings.Fields(prevCpuMax)
+	if len(prevPair) < 2 {
+		return "", errors.Errorf("bad cpu.max: %q", prevCpuMax)
+	}
+	// prevQuotaStr is either a positive decimal integer or "max"
+	prevQuotaStr := prevPair[0]
+	prevPeriodStr := prevPair[1]
+	prevPeriod, err := strconv.Atoi(prevPeriodStr)
+	if err != nil {
+		return "", errors.Errorf("bad cpu.max: %q", prevCpuMax)
+	}
+
+	if period == 0 {
+		period = uint64(prevPeriod)
+	}
+	if quota < 0 {
+		return fmt.Sprintf("max %d", period), nil
+	}
+	if quota == 0 {
+		return fmt.Sprintf("%s %d", prevQuotaStr, period), nil
+	}
+	return fmt.Sprintf("%d %d", quota, period), nil
+}
+
 func statCpu(dirPath string, stats *cgroups.Stats) error {
 	f, err := os.Open(filepath.Join(dirPath, "cpu.stat"))
 	if err != nil {
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -127,6 +127,6 @@ type Resources struct {
 	// CpuWeight sets a proportional bandwidth limit.
 	CpuWeight uint64 `json:"cpu_weight"`
 
-	// CpuMax sets she maximum bandwidth limit (format: max period).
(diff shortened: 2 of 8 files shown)


tests/integration/update.bats
```diff
@@ -354,6 +354,36 @@ EOF
     check_systemd_value "CPUShares" 100
 }
 
+@test "update cgroup v2 cpu limits" {
```
kolyshkin, line 357: I don't think it makes sense to have separate tests for v1 and v2 -- runc should work the same way in both, only the check it different but I was able to unify the checks in https://github.com/opencontainers/runc/pull/2428/files#diff-4c21eb8cb6048713e619964ad6d1cab9.

libcontainer/configs/cgroup_linux.go
```diff
@@ -127,6 +127,6 @@ type Resources struct {
 	// CpuWeight sets a proportional bandwidth limit.
 	CpuWeight uint64 `json:"cpu_weight"`
 
-	// CpuMax sets she maximum bandwidth limit (format: max period).
-	CpuMax string `json:"cpu_max"`
+	// CpuMax is converted from CpuQuota and CpuPeriod.
+	// Specifying CpuMax directly is not supported currently.
```
kolyshkin, line 131: I think we should just remove it, since it is not in spec (as it was done in https://github.com/opencontainers/runc/pull/2428/commits/165e76c9253ff1860aeb96a26dcd6ae3813dcc1b)

kolyshkin:
So, let's discuss the fix. I see a few ways to fix this, and am still thinking about which one is the best.

1. Do not allow setting cpu quota without the period (and period without the quota). Naturally, these two numbers should go together, as they don't make sense one without the other (like numerator and denominator do not make sense one without the other in a fraction). I tried this in https://github.com/opencontainers/runc/pull/2423 and it turned out this is too radical, since there is a software depending on the current behavior of being able to specify one without the other.

2. When quota is set and period is not, use the default period (of 100000). This is how fs2 and systemd/v2 drivers currently work. There are two deficiencies here:
* this is obviously only working for initial container start (not for update, because the container's period might have been changed from the default earlier);
* this is only working when the period is not set (and does not handle the case when the period is set but the quota is not set).

3. When any one of {quota,period} is set, and the other is not, use the old (current, previously set) value for the other. Currently this is how cgoup v1 fs driver works.


It seems that the approach number 3 is the best, being a good compromise between the decency and the backword compatibility. So, we need to use it for the other three drivers (fs2 and systemd v1 and v2). Now, again, there are two ways of handling it:

1. Read the old/current value from the kernel (cgroup file). This is what this PR does for fs2.
2. Get the old/current value from state.json. This file is actually being read during `runc update`, but the value is being overwritten.

I have not yet decided which way is best, or maybe we should try a mixed approach (use the value from `state.json`, if available, otherwise from the kernel cgroup file.

@AkihiroSuda do you have any thoughts on this?

kolyshkin:
The fix looks OK to me overall. A similar one is required for cgroups/systemd/v1

kolyshkin:
> The fix looks OK to me overall

Actually not. Previously, if quota/period was not set, we were doing nothing. Now we read the old values, parse them, and write them back.

libcontainer/cgroups/fs2/cpu.go
```diff
@@ -29,15 +33,56 @@ func setCpu(dirPath string, cgroup *configs.Cgroup) error {
 		}
 	}
 
-	// NOTE: .CpuQuota and .CpuPeriod are not used here. Conversion is the caller's responsibility.
-	if cgroup.Resources.CpuMax != "" {
-		if err := fscommon.WriteFile(dirPath, "cpu.max", cgroup.Resources.CpuMax); err != nil {
+	// Convert .CpuQuota and .CpuPeriod into "cpu.max".
+	// Conversion requires the previous value of "cpu.max"
+	prevCpuMaxBytes, err := ioutil.ReadFile(filepath.Join(dirPath, "cpu.max"))
```
kolyshkin, line 38: We should not do anything if quota and period is not set. This read is not needed in this case.

libcontainer/cgroups/fs2/cpu.go
```diff
@@ -29,15 +33,56 @@ func setCpu(dirPath string, cgroup *configs.Cgroup) error {
...
+	}
+
+	prevPair := strings.Fields(prevCpuMax)
+	if len(prevPair) < 2 {
+		return "", errors.Errorf("bad cpu.max: %q", prevCpuMax)
+	}
+	// prevQuotaStr is either a positive decimal integer or "max"
+	prevQuotaStr := prevPair[0]
+	prevPeriodStr := prevPair[1]
+	prevPeriod, err := strconv.Atoi(prevPeriodStr)
```
kolyshkin, line 69: move the conversion to under `if period == 0` below

kolyshkin:
> A similar one is required for cgroups/systemd/v1

and cgroups/systemd/v2 as well. Looking into it...

AkihiroSuda:
Not sure how we can support systemd for the plan 3. The plan 1 might be the best, at least for cgroup2.

kolyshkin:
> Not sure how we can support systemd for the plan 3

For initial setting:
 - if period is not set, assume the default of `100000`
 - if quota is not set, reject it with an error

For update:
 - get the previous value from state.json and/or cgroupfs

My current work in progress is https://github.com/opencontainers/runc/pull/2458. I still haven't finished it but the issue #2456 is already fixed for all cgroup drivers. PTAL

kolyshkin:
Update to the plan.

1. For update, use the previous values from state.json if not explicitly set during update.
2. If period is not set, assume the default of 100000 us (100 ms, same for v1 and v2).
3. If quota is not set, assume -1 / unlimited.

This way, it is always possible to change either quota or period, and runc stays backward-compatible.

PR #2458, opened 2020-06-05
Cpu quota fixes, try II

kolyshkin:
_this is loosely based on https://github.com/opencontainers/runc/pull/2428 but I am trying to make things right this time_

Seting CPU quota and period independently does not make much sense, but historically runc allowed it and this needs to be supported to not break compatibility.

This PR fixes the following issues:

### 1. Issue https://github.com/opencontainers/runc/issues/2456

systemd/v1, systemd/v2 and fs2 cgroup drivers reset the period to default 100000 during `runc update` if it is not set explicitly. This makes the behavior inconsistent with the one when fs driver is used (it leaves the period at the old value).

The issue is fixed by reusing the old (previously set) value for period/quota, if a new period/quota is not specified during the update.

### 2. Issue https://github.com/opencontainers/runc/issues/2464

systemd/v1, systemd/v2 and fs2 cgroup drivers ignore the CPU period during start/run, if CPU quota is not set. The fs (v1) driver sets the period.

~~The issue is fixed by rejecting the configuration where quota is not set but the period is set during start. **NOTE:** this can bring some backward compatibility issues but I consider the chances are low.~~

Issue fixed by
 - changing the AND to OR in checking whether the CPU quota values should be applied, and
 - using defaults for both quota and period if those are not set (default quota is unlimited, default period is 100000).

### 3. Issue https://github.com/opencontainers/runc/issues/2463

systemd v2 driver ignores --cpu-quota during update if CPU period was not set.

Fixed by adding the default period to `addCpuQuota`

### Other changes

In addition, this PR enables CPU quota and CPU shares tests for v2, adds unlimited quota tests, and some corner test cases to cover all the issues above.

Fixes: #2456, #2463, #2464.

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -352,7 +352,7 @@ func stopUnit(dbusConnection *systemdDbus.Conn, unitName string) error {
 }
 
 func addCpuQuota(properties *[]systemdDbus.Property, quota int64, period uint64) {
-	if quota != 0 && period != 0 {
+	if quota != 0 || period != 0 {
 		// corresponds to USEC_INFINITY in systemd
 		cpuQuotaPerSecUSec := uint64(math.MaxUint64)
 		if quota > 0 {
@@ -360,6 +360,12 @@ func addCpuQuota(properties *[]systemdDbus.Property, quota int64, period uint64)
 			// (integer percentage of CPU) internally.  This means that if a fractional percent of
 			// CPU is indicated by Resources.CpuQuota, we need to round up to the nearest
 			// 10ms (1% of a second) such that child cgroups can set the cpu.cfs_quota_us they expect.
+			if period == 0 {
+				// assume the default kernel value of 100000 us (100 ms), same for v1 and v2.
+				// v1: https://www.kernel.org/doc/html/latest/scheduler/sched-bwc.html and
+				// v2: https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html
+				period = 100000
+			}
 			cpuQuotaPerSecUSec = uint64(quota*1000000) / period
 			if cpuQuotaPerSecUSec%10000 != 0 {
 				cpuQuotaPerSecUSec = ((cpuQuotaPerSecUSec / 10000) + 1) * 10000
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -3,6 +3,7 @@ package systemd
 import (
 	"bufio"
 	"fmt"
+	"math"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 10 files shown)


libcontainer/cgroups/systemd/common.go
```diff
@@ -349,3 +350,26 @@ func stopUnit(dbusConnection *systemdDbus.Conn, unitName string) error {
 	}
 	return nil
 }
+
+func addCpuQuota(properties *[]systemdDbus.Property, quota int64, period uint64) error {
+	// cpu.cfs_quota_us and cpu.cfs_period_us are controlled by systemd.
```
AkihiroSuda, line 355: and cpu.max

kolyshkin: right, this comment comes from cgroupv1, will fix

kolyshkin:
@AkihiroSuda @mrunalp @cyphar  PTAL

update.go
```diff
@@ -251,10 +251,36 @@ other options are ignored.
...
+		// If only one of {quota,period} is set and the other is not, leave
+		// the unset parameter at the old value (don't overwrite config).
+		p, q := *r.CPU.Period, *r.CPU.Quota
+		if (p == 0 && q == 0) || (p != 0 && q != 0) {
+			// both values are either set or unset (0)
+			config.Cgroups.Resources.CpuPeriod = p
+			config.Cgroups.Resources.CpuQuota = q
+		} else {
+			// one is set and the other is not
+			if p != 0 {
```
mrunalp, line 275: This is a very unlikely scenario. Should we just fail here?

kolyshkin: Not unlikely. I tried failing when one is set and the other isn't and it ended up not so good so I had to revert it, see https://github.com/opencontainers/runc/pull/2434

This allows to use `runc update $ID --cpu-quota` without `--cpu-period` and vice versa, in which case the old value from the missing setting is used (it is loaded into config earlier).

kolyshkin:
One more corner case I have missed:
```bash
runc --systemd-cgroup run -d xx44 # no cpu quota/limit setting in config
runc --systemd-cgroup update xx44 --cpu-period 5000
cat /sys/fs/cgroup/system.slice/runc-xx44.scope/cpu.max
max 100000
^^^ bug -- the new cpu-period setting is ignored :(
```

kolyshkin:
> One more corner case I have missed

Which is fortunately easy to fix (see the last commit).

kolyshkin:
> Which is fortunately easy to fix (see the last commit).

It was my own regression, so I fixed it in place. Adding more tests...

kolyshkin:
Found a couple of more bugs, fixed #2463 but chose not to fix https://github.com/opencontainers/runc/issues/2465 in here, otherwise, it will be endless.

kolyshkin:
@AkihiroSuda @mrunalp PTAL

kolyshkin:
Please wait; I want to polish this one a bit further

kolyshkin:
Dropped the patch that rejected cpu period without cpu quota on run/start, as it could bring some backward compatibility issues. Instead, we use default quota of `-1` aka `unlimited` aka `infinity`.

This one is ready for review.

libcontainer/cgroups/systemd/common.go
```diff
@@ -352,14 +352,20 @@ func stopUnit(dbusConnection *systemdDbus.Conn, unitName string) error {
...
 		// corresponds to USEC_INFINITY in systemd
 		cpuQuotaPerSecUSec := uint64(math.MaxUint64)
 		if quota > 0 {
 			// systemd converts CPUQuotaPerSecUSec (microseconds per CPU second) to CPUQuota
 			// (integer percentage of CPU) internally.  This means that if a fractional percent of
 			// CPU is indicated by Resources.CpuQuota, we need to round up to the nearest
 			// 10ms (1% of a second) such that child cgroups can set the cpu.cfs_quota_us they expect.
+			if period == 0 {
+				// assume the default kernel value of 100000 us (100 ms), same for v1 and v2.
+				// v1: https://www.kernel.org/doc/html/latest/scheduler/sched-bwc.html and
```
AkihiroSuda, line 365: nit: latest -> v5.6 to avoid potential deadlink?

AkihiroSuda:
@cyphar LGTY?

PR #2459, opened 2020-06-06
Set configs back when intelrdt configs cannot be set

tedyu:
linuxContainer#Set(), if call to intelRdtManager#Set() returns error, we try to set configs back.

However, only intelRdtManager#Set() is called.
This PR adds the call to cgroupManager#Set(), similar to the call on line 217.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -222,6 +222,9 @@ func (c *linuxContainer) Set(config configs.Config) error {
 	if c.intelRdtManager != nil {
 		if err := c.intelRdtManager.Set(&config); err != nil {
 			// Set configs back
+			if err2 := c.cgroupManager.Set(c.config); err2 != nil {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Could you add integration tests

tedyu:
Currently only container_linux_test.go exercises container.Set():
```
	if err := container.Set(newConfig); err != nil {
		t.Fatal(err)
```
@AkihiroSuda 
If you know how to inject error into intelRdtManager.Set() ?

kolyshkin:
> Could you add integration tests

those would require hardware capable of Intel RDT. This is definitely not available on Vagrant, but _might_ be available from Travis CI hosts (if they use recent Xeons). Some info about it here: https://www.kernel.org/doc/html/latest/x86/resctrl_ui.html

tedyu:
@kolyshkin 
Can this PR move forward ?

Thanks

Issue #2465, opened 2020-06-09
systemd drivers do not set CPUQuotaPeriodUSec


kolyshkin:
According to systemd.resource-control(5), there are two properties wrt cpu quota:

<blockquote>

**CPUQuota=**

Assign the specified CPU time quota to the processes executed. Takes a percentage value, suffixed with "%". The percentage specifies how much CPU time the unit shall get at maximum, relative to the total CPU time available on one CPU. Use values > 100% for allotting CPU time on more  than one CPU. This controls the "cpu.max" attribute on the unified control group hierarchy and "cpu.cfs_quota_us" on legacy. For details about these control group attributes, see Control Groups v2[2] and sched-bwc.txt[5].

Example: CPUQuota=20% ensures that the executed processes will never get more than 20% CPU time on one CPU.

**CPUQuotaPeriodSec=**

Assign the duration over which the CPU time quota specified by CPUQuota= is measured. Takes a time duration value in seconds, with an optional suffix such as "ms" for milliseconds (or "s" for seconds.) The default setting is 100ms. The period is clamped to the range supported by the kernel, which is [1ms, 1000ms]. Additionally, the period is adjusted up so that the quota interval is also at least 1ms. Setting CPUQuotaPeriodSec= to an empty value resets it to the default.

This controls the second field of "cpu.max" attribute on the unified control group hierarchy and "cpu.cfs_period_us" on legacy. For details about these control group attributes, see Control Groups v2[2] and CFS Scheduler[4].

Example: CPUQuotaPeriodSec=10ms to request that the CPU quota is measured in periods of 10ms.
</blockquote>

Currently, both systemd/v1 and systemd/v2 cgroup drivers set `CPUQuota` (known as `CPUQuotaPerSecUsec` on dbus and in `systemctl show` output) but does not set `CPUQuotaPeriodSec` (known as `CPUQuotaPeriodUSec` on dbus and in `systemctl show` output).

PR #2466, opened 2020-06-10
cgroups/systemd: add setting CPUQuotaPeriod prop

kolyshkin:
_This is currently based on and includes #2458 and will be rebased once that one is merged. Please only review the last commit._

For some reason, runc systemd drivers (both v1 and v2) never set
systemd unit property named `CPUQuotaPeriod` (known as
`CPUQuotaPeriodUSec` on dbus and in `systemctl show` output).

Set it, and add a check to all the integration tests. The check is less
than trivial because, when not set, the value is shown as "infinity" but
when set to the same (default) value, shown as "100ms", so in case we
expect 100ms (period = 100000 us), we have to _also_ check for
"infinity".

Fixes https://github.com/opencontainers/runc/issues/2465

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -21,6 +22,10 @@ var (
 	connOnce sync.Once
 	connDbus *systemdDbus.Conn
 	connErr  error
+
+	versionOnce sync.Once
+	version     int
+	versionErr  error
 )
 
 // NOTE: This function comes from package github.com/coreos/go-systemd/util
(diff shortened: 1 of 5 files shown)


kolyshkin:
On Ubuntu 18.04 we have this:

> Cannot set property CPUQuotaPeriodUSec, or unknown property.

Apparently systemd in ubuntu-18.04 did not have support for this parameter.

Looking at the systemd git, this was added in **systemd v242** (https://github.com/systemd/systemd/pull/9594).

I think we can check systemd version during controller initialization and set this conditionally.

AkihiroSuda:
https://github.com/opencontainers/runc/pull/2458 is merged, please rebase

kolyshkin:
Added systemd version detection and conditional setting. 

I'm not quite sure why this one fails. @AkihiroSuda can you please take a look?

From https://travis-ci.org/github/opencontainers/runc/jobs/698672952#L4689:

> make localrootlessintegration RUNC_USE_SYSTEMD=yes
> ...
> [01] run rootless tests ... ()
> ...
> not ok 79 update cgroup v1/v2 common limits
> (in test file /vagrant/tests/integration/update.bats, line 41)
>  `[ "$status" -eq 0 ]' failed
> ...
>  time="2020-06-15T19:58:00Z" level=error msg="container_linux.go:353: starting container process caused: process_linux.go:326: applying cgroup configuration for process caused: error while starting unit \"runc-cgroups-integration-test.scope\" with properties [{Name:Description Value:\"libcontainer container integration-test\"} {Name:Slice Value:\"machine.slice\"} {Name:PIDs Value:@au [56928]} {Name:Delegate Value:true} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:DefaultDependencies Value:false} {Name:DevicePolicy Value:\"strict\"} {Name:DeviceAllow Value:@a(ss) []} {Name:DeviceAllow Value:[\"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\"]} {Name:MemoryMax Value:@t 33554432} {Name:MemoryLow Value:@t 25165824} {Name:CPUWeight Value:@t 4} {Name:CPUQuotaPeriodUSec Value:@t 1000000} {Name:CPUQuotaPerSecUSec Value:@t 500000} {Name:TasksAccounting Value:true} {Name:TasksMax Value:@t 20}]: Permission denied"

kolyshkin:
Added a commit to temporarily disable setting cpu quota period to check that the error above is caused by it.

kolyshkin:
Still have the same error after commenting out setting `CPUQuotaPeriod`. Hmm, is the master broken or I am not seeing something?

> not ok 79 update cgroup v1/v2 common limits
> (in test file /vagrant/tests/integration/update.bats, line 41)
>  `[ "$status" -eq 0 ]' failed
> time="2020-06-15T21:54:35Z" level=error msg="container_linux.go:353: starting container process caused: process_linux.go:326: applying cgroup configuration for process caused: error while starting unit \"runc-cgroups-integration-test.scope\" with properties [{Name:Description Value:\"libcontainer container integration-test\"} {Name:Slice Value:\"machine.slice\"} {Name:PIDs Value:@au [56608]} {Name:Delegate Value:true} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:DefaultDependencies Value:false} {Name:DevicePolicy Value:\"strict\"} {Name:DeviceAllow Value:@a(ss) []} {Name:DeviceAllow Value:[\"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\", \"INVALID\"]} {Name:MemoryMax Value:@t 33554432} {Name:MemoryLow Value:@t 25165824} {Name:CPUWeight Value:@t 4} {Name:CPUQuotaPerSecUSec Value:@t 500000} {Name:TasksAccounting Value:true} {Name:TasksMax Value:@t 20}]: Permission denied"

libcontainer/cgroups/systemd/common.go
```diff
@@ -351,21 +356,61 @@ func stopUnit(dbusConnection *systemdDbus.Conn, unitName string) error {
 	return nil
 }
 
+func systemdVersion() (int, error) {
```
mrunalp, line 359: This seems heavy to call per runc invocation. Should we think about using build tags instead, so users can compile it for their version of systemd as needed? Another option may be a command-line flag. This applies to any other systemd feature detection as well.

AkihiroSuda: Let's just always assume systemd >= 242, and fall back to 241-compatible mode on error

kolyshkin: > This seems heavy to call per runc invocation

Does not look like heavy to me, and it's not per runc invocation, but only upon container creation or update (and in this case we talk to dbus already to create/update a unit, so there's no such thing as creating and new connection just to get the version).

Or am I missing something here?

The alternative is to exec systemctl and parse its output. I haven't checked which way is faster, but can look into it.

> Let's just always assume systemd >= 242, and fall back to 241-compatible mode on error

I was thinking about it, but the logic would look really ugly (got EPERM? remove cpu quota period and retry). I would consider it if there is no other way to know if systemd supports the feature or not, but it seems

kolyshkin:
Modified the patch to explicitly reuse dbus connection (which was implicitly reused before). This also fixed the rootless failure.

libcontainer/cgroups/systemd/common.go
```diff
@@ -21,6 +22,10 @@ var (
 	connOnce sync.Once
 	connDbus *systemdDbus.Conn
 	connErr  error
+
+	versionOnce sync.Once
```
kolyshkin, line 26: Note that all this sync.Once stuff is currently redundant, as we only call systemdVersion() once. It was added in anticipation there will be other systemd parameters like this that need a version guard, and since we call `systemdVersion()` when and where needed (right from `addCpuQuota`) rather than in advance, the future calls might come from a similar "when and where needed" places.

PR #2467, opened 2020-06-10
Write state.json atomically

mrunalp:
@AkihiroSuda @kolyshkin ptal

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1674,13 +1674,30 @@ func (c *linuxContainer) updateState(process parentProcess) (*State, error) {
 	return state, nil
 }
 
-func (c *linuxContainer) saveState(s *State) error {
-	f, err := os.Create(filepath.Join(c.root, stateFilename))
+func (c *linuxContainer) saveState(s *State) (retErr error) {
+	tmpFile, err := ioutil.TempFile(c.root, "state-")
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	return utils.WriteJSON(f, s)
+
+	defer func() {
+		if retErr != nil {
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
+		}
+	}()
+
+	err = utils.WriteJSON(tmpFile, s)
+	if err != nil {
+		return err
+	}
+	err = tmpFile.Close()
+	if err != nil {
+		return err
+	}
+
+	stateFilePath := filepath.Join(c.root, stateFilename)
+	return os.Rename(tmpFile.Name(), stateFilePath)
 }
 
 func (c *linuxContainer) deleteState() error {


libcontainer/container_linux.go
```diff
@@ -1674,13 +1674,36 @@ func (c *linuxContainer) updateState(process parentProcess) (*State, error) {
...
+	defer func() {
+		if retErr != nil {
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
+		}
+	}()
+
+	err = utils.WriteJSON(tmpFile, s)
+	if err != nil {
+		tmpFile.Close()
```
kolyshkin, line 1692: No need to close the file on error path as it is done in `defer`

libcontainer/container_linux.go
```diff
@@ -1674,13 +1674,36 @@ func (c *linuxContainer) updateState(process parentProcess) (*State, error) {
...
+		return err
+	}
+	err = tmpFile.Sync()
+	if err != nil {
+		tmpFile.Close()
```
kolyshkin, line 1697: ditto

kolyshkin:
Is this to prevent reading back a partially written `state.json`?

libcontainer/container_linux.go
mrunalp: fixed.

libcontainer/container_linux.go
mrunalp: fixed.

mrunalp:
> Is this to prevent reading back a partially written state.json?

Yes! :)

libcontainer/container_linux.go
```diff
@@ -1674,13 +1674,34 @@ func (c *linuxContainer) updateState(process parentProcess) (*State, error) {
...
+			tmpFile.Close()
+			os.Remove(tmpFile.Name())
+		}
+	}()
+
+	err = utils.WriteJSON(tmpFile, s)
+	if err != nil {
+		return err
+	}
+	err = tmpFile.Sync()
```
AkihiroSuda, line 1694: Do you need this?

mrunalp: Yes, we want to sync the file to make sure it is written to disk.

mrunalp: That applies if the root isn't a tmpfs.

mrunalp: I can drop it if we think we don't want to support that, but I don't see big downsides in keeping it.

cyphar:
Rather than open-coding this, why not ~~steal~~ use [the AtomicFileWriter from Docker et al](https://github.com/moby/moby/blob/master/pkg/ioutils/fswriters.go)?

libcontainer/container_linux.go
AkihiroSuda: Sync may block, and this file does not need persistency

kolyshkin: The file contents, even if not synced to the disk, will be read back from the page cache the way it was written. The only thing that fsync helps with is when the host will lose power (or the kernel will crash) before writing the file to disk. I am not sure this is needed.

mrunalp: Dropped the sync.

mrunalp:
@cyphar Yeah, that's what I considered using first, but didn't want to depend for just a few lines of code and now with the sync being dropped, I am not sure if we want to pull that in, since that code does sync - https://github.com/moby/moby/blob/master/pkg/ioutils/fswriters.go#L68

PR #2470, opened 2020-06-15
Fix #2469 omit memory.numa_stat when not available

### LINKED ISSUE #2469: memory.numa_stat not always available
memory.numa_stat is not always available, observed on:
```
/sys/fs/cgroup/memory # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.12.0
PRETTY_NAME="Alpine Linux v3.12"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"

/sys/fs/cgroup/memory # uname -sr
Linux 4.19.76-linuxkit
```
for details see: https://github.com/google/cadvisor/issues/2583

katarzyna-z:
Signed-off-by: Katarzyna Kujawa <katarzyna.kujawa@intel.com>

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -291,7 +291,9 @@ func getPageUsageByNUMA(cgroupPath string) (cgroups.PageUsageByNUMA, error) {
 	stats := cgroups.PageUsageByNUMA{}
 
 	file, err := os.Open(path.Join(cgroupPath, cgroupMemoryPagesByNuma))
-	if err != nil {
+	if os.IsNotExist(err) {
+		return stats, nil
+	} else if err != nil {
 		return stats, err
 	}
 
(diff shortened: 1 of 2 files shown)


libcontainer/cgroups/fs/memory.go
```diff
@@ -291,7 +291,9 @@ func getPageUsageByNUMA(cgroupPath string) (cgroups.PageUsageByNUMA, error) {
 	stats := cgroups.PageUsageByNUMA{}
 
 	file, err := os.Open(path.Join(cgroupPath, cgroupMemoryPagesByNuma))
-	if err != nil {
+	if os.IsNotExist(err) {
+		return stats, nil
+	} else if err != nil {
```
kolyshkin, line 296: `else` is redundant here because there's a `return` above it, but it's fine I guess.

PR #2472, opened 2020-06-16
Paths nits

kolyshkin:
This is a fixup for 

1. using `GetPaths()` instead of `Path()`
2. accessing `m.paths` map without lock

mostly introduced by #2338 and #2391.


kolyshkin:
@AkihiroSuda @lifubang @mrunalp @cyphar PTAL

Issue #2475, opened 2020-06-16
criu int test failures


kolyshkin:
Saw some C/R test failures while working on #2411. Excerpt from logs at https://travis-ci.org/github/opencontainers/runc/jobs/699060158:
```
not ok 12 checkpoint --lazy-pages and restore
2897# (from function `__runc' in file tests/integration/helpers.bash, line 54,
2898#  in test file tests/integration/checkpoint.bats, line 183)
2899#   `__runc --criu "$CRIU" restore -d --work-path ./image-dir --image-path ./image-dir --lazy-pages test_busybox_restore <&60 >&51 2>&51' failed
2900# runc list (status=0):
2901# ID          PID         STATUS      BUNDLE      CREATED     OWNER
2902# runc spec (status=0):
2903# 
2904# runc state test_busybox (status=0):
2905# {
2906#   "ociVersion": "1.0.2-dev",
2907#   "id": "test_busybox",
2908#   "pid": 3967,
2909#   "status": "running",
2910#   "bundle": "/tmp/busyboxtest",
2911#   "rootfs": "/tmp/busyboxtest/rootfs",
2912#   "created": "2020-06-16T19:53:26.677415824Z",
2913#   "owner": ""
2914# }
2915# Warn  (criu/kerndat.c:869): Can't keep kdat cache on non-tempfs
2916# runc list (status=0):
2917# ID             PID         STATUS      BUNDLE             CREATED                          OWNER
2918# test_busybox   3967        running     /tmp/busyboxtest   2020-06-16T19:53:26.677415824Z   root
2919# runc kill test_busybox KILL (status=0):
2920# 
2921# runc delete test_busybox (status=0):
2922# 
2923not ok 13 checkpoint and restore in external network namespace
2924# (from function `teardown_busybox' in file tests/integration/helpers.bash, line 453,
2925#  from function `teardown' in test file tests/integration/checkpoint.bats, line 14)
2926#   `teardown_busybox' failed
2927# runc list (status=0):
2928# ID          PID         STATUS      BUNDLE      CREATED     OWNER
2929# runc spec (status=0):
2930# 
2931# runc run -d --console-socket /tmp/console.sock test_busybox (status=0):
2932# 
2933# runc state test_busybox (status=0):
2934# {
2935#   "ociVersion": "1.0.2-dev",
2936#   "id": "test_busybox",
2937#   "pid": 4207,
2938#   "status": "running",
2939#   "bundle": "/tmp/busyboxtest",
[...]

PR #2476, opened 2020-06-17
fixups to tests/int/checkpoint.bats

kolyshkin:
When looking into https://github.com/opencontainers/runc/issues/2475 I noticed there are no CRIU errors shown for runc restore. Amend the script with showing errors like in other cases.

While at it
 - fix EOL whitespace
 - fix exit code check in `simple_cr`

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -17,7 +17,7 @@ function teardown() {
 function setup_pipes() {
 	# The changes to 'terminal' are needed for running in detached mode
 	update_config 	' (.. | select(.terminal? != null)) .terminal |= false
-			| (.. | select(.[]? == "sh")) += ["-c", "for i in `seq 10`; do read xxx || continue; echo ponG $xxx; done"]' 
+			| (.. | select(.[]? == "sh")) += ["-c", "for i in `seq 10`; do read xxx || continue; echo ponG $xxx; done"]'
 
 	# Create two sets of pipes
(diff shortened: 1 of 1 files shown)


kolyshkin:
@AkihiroSuda @mrunalp PTAL

tests/integration/checkpoint.bats
```diff
@@ -283,6 +284,7 @@ function simple_cr() {
 
   # checkpoint the running container
   runc --criu "$CRIU" checkpoint --work-path ./work-dir test_busybox
+  cat ./work-dir/dump.log | grep -B 5 Error || true
   [ "$status" -eq 0 ]
```
AkihiroSuda, line 288: this check doesn't make a sense

kolyshkin: Why? It totally does. We check that runc returns 0.

AkihiroSuda: The check should be before `cat .... true`

kolyshkin: No, it should be there is it.

First, the command `runc` is a shell function which uses bats `run` statement, which saves the exit code into `$status` (i.e. it does `status=$?` somewhere). We can check `$status` any time later, unless it's rewritten by another `run`, which is not the case here.

Second, in case `runc checkpoint` fails, we want to see the actual error. This is the reason why we do `cat | grep` before checking the status.

PR #2478, opened 2020-06-17
libct/cgroups/readProcsFile: return an error if read failed

kolyshkin:
Not sure why but the errors from scanner were ignored. Such errors
can happen if open(2) has succeeded but the subsequent read(2) fails.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -138,7 +138,7 @@ func readProcsFile(file string) ([]int, error) {
 			out = append(out, pid)
 		}
 	}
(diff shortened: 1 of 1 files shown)


kolyshkin:
Looking through the git history, this error was ignored since the beginning of times (see commit fc6d2317737af1a8cb339f80b31e3110a0a11ce1)

kolyshkin:
Tracked this down to commit https://github.com/moby/moby/commit/811d93326bc2d9451eb444e2343bb3063611de7a (PR https://github.com/moby/moby/pull/5976).

PR #2479, opened 2020-06-17
systemd: parse systemdVersion when only an int is returned

haircommander:
there have been cases observed where instead of `v$VER.0-$OS` the systemdVersion returned is just `$VER`. handle this case

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -365,20 +366,26 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
 			return
 		}
 
-		// verStr is like "v245.4-1.fc32" (including quotes)
-		if !strings.HasPrefix(verStr, `"v`) {
-			versionErr = fmt.Errorf("can't parse version %s", verStr)
-			return
-		}
-		// remove `"v` prefix and everything after the first dot
-		ver, err := strconv.Atoi(strings.SplitN(verStr[2:], ".", 2)[0])
-		if err != nil {
-			versionErr = err
-		}
-		version = ver
+		version, versionErr = systemdVersionAtoi(verStr)
+		return
 	})
 
 	return version, versionErr
+}
+
+func systemdVersionAtoi(verStr string) (int, error) {
+	// verStr should be of the form:
+	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32"
+	// all the input strings include quotes, and the output int should be 245
+	// thus, we unconditionally remove the `"v`
+	// and then match on the first integer we can grab
+	re := regexp.MustCompile(`"?v?([0-9]+)`)
+	matches := re.FindStringSubmatch(verStr)
+	if len(matches) < 2 {
+		return 0, errors.Errorf("can't parse version %s: incorrect number of matches %v", verStr, matches)
+	}
+	ver, err := strconv.Atoi(matches[1])
+	return ver, errors.Wrapf(err, "can't parse version %s", verStr)
 }
 
 func addCpuQuota(conn *systemdDbus.Conn, properties *[]systemdDbus.Property, quota int64, period uint64) {
--- a/libcontainer/cgroups/systemd/systemd_test.go
+++ b/libcontainer/cgroups/systemd/systemd_test.go
@@ -1,0 +1,32 @@
+package systemd
+
+import (
+	"testing"
+)
+
+func TestSystemdVersion(t *testing.T) {
+	var systemdVersionTests = []struct {
+		verStr      string
+		expectedVer int
+		expectErr   bool
+	}{
+		{`"219"`, 219, false},
+		{`"v245.4-1.fc32"`, 245, false},
+		{`"241-1"`, 241, false},
+		{`"v241-1"`, 241, false},
+		{"NaN", 0, true},
+		{"", 0, true},
+	}
+	for _, sdTest := range systemdVersionTests {
+		ver, err := systemdVersionAtoi(sdTest.verStr)
+		if !sdTest.expectErr && err != nil {
+			t.Errorf("systemdVersionAtoi(%s); want nil; got %v", sdTest.verStr, err)
+		}
+		if sdTest.expectErr && err == nil {
+			t.Errorf("systemdVersionAtoi(%s); wanted failure; got nil", sdTest.verStr)
+		}
+		if ver != sdTest.expectedVer {
+			t.Errorf("systemdVersionAtoi(%s); want %d; got %d", sdTest.verStr, sdTest.expectedVer, ver)
+		}
+	}
+}


haircommander:
fixes failure found [here](https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3868/test_pull_request_crio_integration_rhel/13239/artifacts/testout.txt)

haircommander:
oops, I messed up the quotes handling

libcontainer/cgroups/systemd/common.go
```diff
@@ -366,16 +366,24 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
 		}
 
 		// verStr is like "v245.4-1.fc32" (including quotes)
-		if !strings.HasPrefix(verStr, `"v`) {
-			versionErr = fmt.Errorf("can't parse version %s", verStr)
+		if strings.HasPrefix(verStr, `"v`) {
```
thaJeztah, line 369: instead of using two separate branches, would it perhaps be an option to unconditionally remove leading/trailing quotes (`"`) and `v`;

```
strings.Trim(verStr, `v"`)
```

and then look for the version before the (optional) dot?

libcontainer/cgroups/systemd/common.go
```diff
@@ -366,16 +366,24 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
 		}
 
 		// verStr is like "v245.4-1.fc32" (including quotes)
-		if !strings.HasPrefix(verStr, `"v`) {
```
kolyshkin, line 369: The comment above needs to be fixed as well, to mention that new form you discovered.

kolyshkin:
So, I guess, we should start handling the string by trimming the quotes, then trimming the `v` prefix, then trimming the `.whatever` suffix, then using Atoi. A shortcut suggested by @thaJeztah above also works (as we don't care if there's `v` at the end.

haircommander:
> left a suggestion
> 
> also curious if the situation where this problem happened is a standard distro, or if someone made a "whoopsie" when compiling :sweat_smile:

I ran into this on a rhel 7 box in AWS. I would not rule out a messed up compilation :laughing:

haircommander:
since there do not 

> Please add UT

tests added, and comments addressed, PTAL

libcontainer/cgroups/systemd/common.go
```diff
@@ -365,22 +365,24 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
...
-		}
-		version = ver
+		version, versionErr = systemdVersionAtoi(verStr)
+		return
 	})
 
 	return version, versionErr
 }
 
+func systemdVersionAtoi(verStr string) (int, error) {
```
kolyshkin, line 375: This commit breaks the build. I would not add the `Atoi` suffix to the function, as it's clear without it that it returns an integer.

libcontainer/cgroups/systemd/systemd_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+)
+
+func TestSystemdVersion(t *testing.T) {
+	var systemdVersionTests = []struct {
+		verStr       string
+		expectedVer  int
+		expectErrNil bool
+	} {
+		{"\"219\"", 219, true},
```
kolyshkin, line 13: You can use `"219"` form to avoid escaping quotes

libcontainer/cgroups/systemd/systemd_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+)
+
+func TestSystemdVersion(t *testing.T) {
+	var systemdVersionTests = []struct {
+		verStr       string
+		expectedVer  int
+		expectErrNil bool
```
kolyshkin, line 11: I would invert the meaning of this and call it "expectErr". Reads easier this way (especially at 7am:))

libcontainer/cgroups/systemd/common.go
```diff
@@ -365,22 +365,24 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
...
+	// verStr is either like "v245.4-1.fc32" or "245" (both including quotes)
+	// thus, we can unconditionally remove the leading "v, as well as
+	// everything after the `.`
+	ver, err := strconv.Atoi(strings.SplitN(strings.Trim(verStr, "\"v"), ".", 2)[0])
```
kolyshkin, line 379: using  backticks will be cleaner here I think
```go
 `"v`
```

libcontainer/cgroups/systemd/common.go
```diff
@@ -365,22 +365,24 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
...
+	if err != nil {
+		err = errors.Wrapf(err, "can't parse version %s", verStr)
+	}
+	return ver, err
```
kolyshkin, line 383: you can actually just remove the `if` and use `return ver, errors.Wrapf(err, "can't parse version %s", verStr)` since `errors. Wrapf` returns nil if err is nil

haircommander:
PTAL @kolyshkin @mrunalp

haircommander:
hold on, found another case

haircommander:
ref: https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/3891/test_pull_request_crio_e2e_fedora/13724/artifacts/e2e.log

libcontainer/cgroups/systemd/common.go
```diff
@@ -365,22 +366,24 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
...
 }
 
+func systemdVersionAtoi(verStr string) (int, error) {
+	// verStr should be of the form:
+	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32"
+	// all the input strings include quotes, and the output int should be 245
+	// thus, we unconditionally remove the `"v`
+	// and then match on the first integer we can grab
+	re := regexp.MustCompile(`[0-9]+`)
+	ver, err := strconv.Atoi(re.FindString(strings.Trim(verStr, `"v`)))
```
kolyshkin, line 383: Well, I guess if you end up using `regex.FindString`, you can do everything with it, i.e. there's no longer need to remove the prefix.

Or, if you want to be more strict about what you're parsing, keep the `Trim` (btw TrimPrefix will be better since we're no longer caring about the " at the end), and add `^` to the regex so it would want to see the numbers at the very beginning.

I would do the first way (leave the regex as is, remove strings.Trim as redundant).

haircommander:
PTAL @AkihiroSuda @kolyshkin @mrunalp

libcontainer/cgroups/systemd/common.go
haircommander: I went the pure regexp route, as suggested

libcontainer/cgroups/systemd/common.go
```diff
@@ -365,22 +366,28 @@ func systemdVersion(conn *systemdDbus.Conn) (int, error) {
...
 	return version, versionErr
 }
 
+func systemdVersionAtoi(verStr string) (int, error) {
+	// verStr should be of the form:
+	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32"
+	// all the input strings include quotes, and the output int should be 245
+	// thus, we unconditionally remove the `"v`
+	// and then match on the first integer we can grab
+	re := regexp.MustCompile(`"?v?([0-9]+)`)
```
kolyshkin, line 382: Strictly speaking, the quote is obligatory not optional, but I guess it does not matter much.

PR #2486, opened 2020-06-24
Move integration tests to Ubuntu Focal

RenaudWasTaken:
Hello!

Following up on the requests of https://github.com/opencontainers/runc/pull/2229
Here I moved all the integration tests from busybox to debian. This involved the following three operations:
- find and replace all occurrences of busybox by debian in the following three contexts:
   - teardown_busybox, setup_busybox
   - BUSYBOX_BUNDLE
   - runc run ./work-dir test_busybox
- Generate the debian spec using runc rather than umoci
   - This had the implication that we needed to update the hooks test to set the container root to read-write (by default runc generates a spec with a read-only root)
- Update some of the error messages that were slightly different with Debian (e.g: Permission denied) and error codes

/cc @cyphar @AkihiroSuda 

Thanks for reviewing!

Signed-off-by: Renaud Gaubert <rgaubert@nvidia.com>


RenaudWasTaken:
Hmm, I wasn't expecting debian not to have ps in the base image...

cyphar:
This has bitten me in the past as well. And now is the part where I pitch openSUSE :wink: -- the `opensuse/leap:15.1` image is smaller than the Debian one and has both `ldconfig` and `ps` in the base image.

RenaudWasTaken:
Hmm, I'll look into using the opensuse/leap image then!

cyphar:
One other thing you'll need to update is the Go "integration" tests in `libcontainer/integration` -- they use the busybox image as well.

RenaudWasTaken:
Looks like opensuse works on all bats the integration tests.
I'm hitting a small issue in the go integration tests related to groups:
```
--- FAIL: TestAdditionalGroups (0.26s)
utils_test.go:55: exec_test.go:487: unexpected error: container_linux.go:367: starting container process caused: setup user: Unable to find group plugdev
````

Will look into it a bit more tomorrow morning :) !

> It also wouldn't hurt to drop the hello image as well, which is also a slightly questionable thing we still have hanging around

I'm happy to drop that one :)

**Comment by cyphar** [comment]
Ah, right. Yeah you'll just need to pick a different group to run the user as (openSUSE doesn't have a `plugdev` group).

File: tests/integration/cgroups.bats [hunk 1]
```diff
@@ -6,24 +6,24 @@ function teardown() {
     rm -f $BATS_TMPDIR/runc-cgroups-integration-test.json
     teardown_running_container test_cgroups_kmem
     teardown_running_container test_cgroups_permissions
-    teardown_busybox
+    teardown_debian
```
mrunalp, line 9: Can we rename functions like these to be distro agnostic so we can change the implementation to use a different distro more isolated in the future?

cyphar: You can also rename `$DEBIAN_BUNDLE` to `$BUNDLE` for the same reason.

RenaudWasTaken: I can do that!

RenaudWasTaken: Taken care of. seems to be running locally, working through the CI now!

RenaudWasTaken:
Working through the last issues!

@cyphar it looks like I can't just `rm -rf` after a `umoci unpack --rootless` for the opensuse image. For some reason I need to chmod again the following dir: `./rootfs/var/lib/ca-certificates/"`. 

Are you ok with that workaround until we investigate a bit more what is going on there?

libcontainer/integration/init_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// init runs the libcontainer initialization code because of the busybox style needs
+// init runs the libcontainer initialization code because of the debian style needs
```
kolyshkin, line 14: Looks like this code is not really distro-dependent, so this comment should not be either.

libcontainer/integration/seccomp_test.go
```diff
@@ -60,7 +60,7 @@ func TestSeccompDenyGetcwdWithErrno(t *testing.T) {
 	}
 	ps, err := pwd.Wait()
 	if err == nil {
-		t.Fatal("Expecting error (negative return code); instead exited cleanly!")
+		t.Fatalf("Expecting error (negative return code); instead exited cleanly!")
```
kolyshkin, line 63: This hunk is useless; `t.Fatal` is more suitable here.

libcontainer/integration/seccomp_test.go
```diff
@@ -77,10 +77,10 @@ func TestSeccompDenyGetcwdWithErrno(t *testing.T) {
 		t.Fatalf("Getcwd should fail with negative exit code, instead got %d!", exitCode)
 	}
 
-	expected := "pwd: getcwd: No such process"
+	expected := "getcwd: cannot access parent directories: No such process"
 	actual := strings.Trim(buffers.Stderr.String(), "\n")
-	if actual != expected {
-		t.Fatalf("Expected output %s but got %s\n", expected, actual)
+	if !strings.Contains(actual, expected) {
+		t.Fatalf("Expected output to contain `%s` but got `%s`\n", expected, actual)
```
kolyshkin, line 83: Can use `%q` here which does the quoting.

libcontainer/integration/seccomp_test.go
RenaudWasTaken: Yep thanks for spotting this, I changed it while debugging :) !

libcontainer/integration/init_test.go
RenaudWasTaken: Thanks addressed!

libcontainer/integration/seccomp_test.go
RenaudWasTaken: Thanks addressed (and in other parts of the code)!

tests/integration/update.bats
```diff
@@ -284,7 +284,7 @@ function check_cpu_shares() {
 @test "update cgroup cpu limits" {
     [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
 
-    # run a few busyboxes detached
+    # run a few containeres detached
```
kolyshkin, line 287: nit: `s/containeres/containers/`

In fact maybe comments like this just need to be dropped, as the code (`runc run -d ...`) is obvious and does not really call to be commented.

kolyshkin:
PR subject and commit message should change debian to opensuse

RenaudWasTaken:
I've been fighting with the CI for a few days now :D !

I think I'm almost done, the most significant change that I've had to introduce is that it seems that just copying the rootfs that umoci extracts hangs in Vagrant (but not in the dockerfile).
See an example here: https://travis-ci.org/github/opencontainers/runc/jobs/701939941
I tracked it down, through plain dumb comment bisecting and CI runs to the specific copy instruction: https://github.com/opencontainers/runc/blob/1b94395c06577b36bae4afd2fe5da229f7a03284/tests/integration/helpers.bash#L454

I suspect it's due to the opensuse image since that issue isn't showing up in master and wasn't showing up when using the debian image.
Unfortunately I don't have a Vagrant capable machine so I haven't been able to thoroughly debug it (I suspect symlink shenanigans).

The workaround that I've employed is to unpack the rootfs (umoci unpack) for every test, but this has introduced a noticeable 2x-2.5x increase in the CI.  e.g:
- Fedora takes 50 minutes instead of 20 minutes
- Go takes 20 minutes instead of 10 minutes
- cgroup systemd takes 6 minutes instead of 3 minutes

tests/integration/update.bats
RenaudWasTaken: I took a stab at it and removed most of the references.

Looking back, I think this is probably a broader topic that should be tackled outside this PR. A lot of the comments are very similar in that they convey only the intent of the next line which is usually pretty obvious:
```
# Start the command.
runc start test_container

# Kill the container.
runc kill test_container KILL

# check state
testcontainer test_container running

# test state of container is paused
testcontainer test_container paused
```

Overall this seems like a discussion that's pretty close to a coding style discussion and might have a better place in an issue first so that there's consensus on how this should be addressed :) 

What do you think?

RenaudWasTaken:
Hmm so that's now to the point where the CI exceeds the time limit, I'll see if I manage to reproduce by some other way

cyphar:
umoci shouldn't make that huge of an impact (though because it has to extract tar archives, it will be a bit slower). I'll look into what might be making it slow. But yeah, since we aren't testing `umoci` here it would be best to just replicate the existing caching done for busybox.

tests/integration/update.bats
kolyshkin: > I took a stab at it and removed most of the references.

Thanks!

> What do you think?

I agree these comments do not add any value and should probably be removed. I also agree this is out of scope for this PR.

kolyshkin:
> cp -r "$DEBIAN_ROOTFS"/* "$DEBIAN_BUNDLE/" 

Wild idea, maybe try running it under strace (right there on CI) to see what is going on?

RenaudWasTaken:
Looks like a `cp -R -P $cache/* "$1"` yields a different result, I'm pretty sure we are dealing with some symlinks shenanigans now :D !

cyphar:
`cp -a` is more fool-proof than `cp -R` in this respect.

RenaudWasTaken:
Looks like CI is green and no significant increase with this latest caching commit!

RenaudWasTaken:
> cp -a is more fool-proof than cp -P (and -a implies -R) in this respect. I think the issue is that you're copying /* which can result in symlinks being resolved at the top-level of the directory. -a will preserve that.

I'll update the last commit, when I was copying stuff around, symlinks were definitely not at the root though. 

> But it would just be simpler to do umoci unpack $CACHED_BUNDLE and then do cp -a $CACHED_BUNDLE $BUNDLE without using * (so you don't run into spaces-in-filenames or too-many-arguments problems).

If I don't cache the result of `umoci unpack` I get a pretty high CI time, see https://travis-ci.org/github/opencontainers/runc/builds/703395550?utm_source=github_status&utm_medium=notification
Which is the direct result of the first commit: https://github.com/opencontainers/runc/pull/2486/commits

Let me update the `*`, I was testing things around :) 

> Wild idea, maybe try running it under strace (right there on CI) to see what is going on?

Hmm yeah that would have worked too :D !

cyphar:
>If I don't cache the result of umoci unpack I get a pretty high CI time

I was trying to say "instead of caching just `bundle/rootfs`, cache the entire `bundle` so you don't need to copy `*`" not that we shouldn't cache at all.

RenaudWasTaken:
> I was trying to say (a little clumsily) "instead of caching just bundle/rootfs, cache the entire bundle so you don't need to copy *" not that we shouldn't cache at all.

It looks like we have the same idea in mind :D  !
I think the latest version reflects that, let me know if I'm completely misreading your message and the code doesn't reflect that.

cyphar:
I'm tracking the performance of umoci in opencontainers/umoci#339. However, unless I'm mistaken you should see a ~38% performance boost if you just do `umoci raw unpack` (which disables `go-mtree` generation and doesn't generate `config.json` either).

RenaudWasTaken:
I managed to get caching running yesterday :) ! I don't think there is any significant difference with the master CI now!
This is ready for review! Let me know if I need to change anything :)

AkihiroSuda:
Does opensuse work on s390x and ppc64le?

AkihiroSuda:
Debian seems the best for multi-arch testing

cyphar:
If we use `opensuse/tumbleweed` then s390x and ppc64le are supported -- and I'd suggest using `tumbleweed` anyway. `opensuse/leap` supports ppc64le, but not s390x (this is because Leap doesn't build on s390x but I'll ask the Leap folks whether they plan to change this).

AkihiroSuda:
Where is the list of the supported architectures? Couldn't find on google :sob:

Dockerfile
```diff
@@ -76,8 +76,9 @@ RUN echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontain
     && apt-get clean \
     && rm -rf /var/cache/apt /var/lib/apt/lists/*;
 
-# install umoci
-RUN curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64 \
+# install umoci, retry with an exponential backoff strategy
+RUN curl --connect-timeout 5 --max-time 10 --retry 5 --retry-delay 0 --retry-max-time 40 -o /usr/local/bin/umoci \
```
cyphar, line 80: I will see if I can get `umoci` added to the `devel:kubic:libcontainers:stable` OBS project so we can eventually remove this and just install it like we do `skopeo`.

Dockerfile
```diff
@@ -76,8 +76,9 @@ RUN echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontain
...
+         -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64 \
```
cyphar, line 81: umoci 0.4.6 has been released.

RenaudWasTaken:
Hmm there's only a latest tag available for tumbleweed, is that ok?
https://hub.docker.com/r/opensuse/tumbleweed

Dockerfile
RenaudWasTaken: Addressed!

cyphar:
Yeah that's fine @RenaudWasTaken -- Tumbleweed is a rolling-release distribution.

cyphar:
@AkihiroSuda 

> Where is the list of the supported architectures? Couldn't find on google :sob:

I only just found this out, but you can find it by looking at [the tags tab on Docker Hub](https://hub.docker.com/r/opensuse/tumbleweed/tags) and you can see which architectures we build for [here](https://build.opensuse.org/repositories/openSUSE:Factory/opensuse-tumbleweed-image).

AkihiroSuda:
Will openSUSE support RISC-V and MIPS?

RenaudWasTaken:
Switched to tumbleweed

RenaudWasTaken:
So `umount` is not part of tumbleweed!

RenaudWasTaken:
It does seem like it's missing the `mount` command though....

RenaudWasTaken:
I opened a request on the opensuse base image to add the mount command: https://build.opensuse.org/request/show/819842

A few alternatives that comes to mind if we can't use the base image (untested):
- Docker in Docker to build a custom container before skopeo + umoci
- skopeo + umoci + chroot + zypper

RenaudWasTaken:
- Rebased
- Moved to Ubuntu as it seems to be the only image that matches all our requirements:
   - util commands: ps, mount, umount, ...
   - Arch support: x86_64, [ppc64le](https://hub.docker.com/r/ppc64le/ubuntu/), [s390x ](https://hub.docker.com/r/s390x/ubuntu/)

For RISC-V and MIPS it seems like there are no official images...

RenaudWasTaken:
Sorry I hadn't noticed that the CI was failing, fixed!

AkihiroSuda:
Lol github amplified my comment three times

RenaudWasTaken:
> Also, what was the problem of using Debian? It's the best operating system for multi arch.

debian does not have ps in the base image.
I have a few thoughts on how we would get that utility in the image (e.g: Docker in Docker or chroot) but nothing very clean...

AkihiroSuda:
Why can't we just aptget it?

RenaudWasTaken:
So the way the multi-arch script operates today is a simple:
```
skopeo copy docker://ubuntu:focal ...
umoci unpack --image ...
```

This works well in the Vagrant and Docker approach because it doesn't require spinning a container. If we wanted to apt-get a package inside that image, we have a few paths but they each seem to add a good amount of complexity on the CI:
- Docker in Docker: `docker build + skopeo copy on that image + umoci unpack`
- Some combination of `wget procps package in + chroot + dpkg`
- A separate pipeline where we build a custom debian container

It's possible that I'm missing a completely obvious solution though :D

kolyshkin:
This is very obsoleted now.

Issue #2474, opened 2020-06-16
api, cgroupv2: skip setting the devices cgroup


giuseppe:
The Kubelet uses libcontainer/cgroups to setup cgroups.  It would be nice to have a way to skip setting the devices cgroup at all.

mrunalp:
@kolyshkin @AkihiroSuda fyi

cyphar:
:grimacing: Is there a reason for wanting to skip the devices cgroup? The hard requirement for the devices cgroup is to make sure we have a fail-secure setup (if you don't set the devices whitelist, you will allow users to do all sorts of scary things).

giuseppe:
the Kubelet uses the libcontainer code to create the parent cgroups (e.g. /sys/fs/cgroup/kubepods).  Each container will have its own cgroup under `/sys/fs/cgroup/kubepods` where the device cgroup is configured.
I don't think it is a problem on cgroup v1, but on cgroup v2, every time the Kubelet restarts and tries to configure the cgroup, it leaks an eBPF program

cyphar:
Ah okay. Yeah we might need to add a way to configure that. The only important thing is that this should be strictly opt-out with a fairly large warning sign next to the configuration option.

kolyshkin:
Addressed this one in https://github.com/opencontainers/runc/pull/2490

cyphar:
As discussed in #2490, I believe that the cgroupv2 eBPF issue is a runc bug that we should fix anyway (tracked by #2366). Would the Kubernetes folks still need this if we fixed the cgroupv2 eBPF issue?

giuseppe:
the devices cgroup is not used at all by Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/container_manager_linux.go#L378-L385

Since the cost of running such eBPF program is close to 0, I am fine if the underlying issue is addressed instead of offering a different API.

odinuge:
> the devices cgroup is not used at all by Kubernetes:

Well, that is kinda true, but only applies to `kubelet` and its usage of `libcontainer`.

However, Kubernetes is implicitly using it via the CRI api since most CRI implementations use runc by default. As I understand the `cgroup.SkipDevices`, is only for users of the go api, not the binary.

[...]

PR #2490, opened 2020-06-30
libct/cgroups: add SkipDevices to Resources

kolyshkin:
The kubelet uses libct/cgroups code to set up cgroups. It creates a
parent cgroup (kubepods) to put the containers into.

The problem (for cgroupv2 that uses eBPF for device configuration) is
the hard requirement to have devices cgroup configured results in
leaking an eBPF program upon every kubelet restart.  program. If kubelet
is restarted 64+ times, the cgroup can't be configured anymore.

Work around this by adding a SkipDevices flag to Resources.

A check was added so that if SkipDevices is set, such a "container"
can't be started (to make sure it is only used for non-containers).

Fixes https://github.com/opencontainers/runc/issues/2474

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -204,7 +204,7 @@ func (m *manager) Apply(pid int) (err error) {
 		if err != nil {
 			// The non-presence of the devices subsystem is
 			// considered fatal for security reasons.
-			if cgroups.IsNotFound(err) && sys.Name() != "devices" {
+			if cgroups.IsNotFound(err) && (c.SkipDevices || sys.Name() != "devices") {
 				continue
 			}
 			return err
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -379,24 +379,27 @@ func (m *legacyManager) Set(container *configs.Config) error {
 		return err
 	}
 
-	// Figure out the current freezer state, so we can revert to it after we
-	// temporarily freeze the container.
-	targetFreezerState, err := m.GetFreezerState()
-	if err != nil {
-		return err
-	}
-	if targetFreezerState == configs.Undefined {
-		targetFreezerState = configs.Thawed
-	}
-
 	// We have to freeze the container while systemd sets the cgroup settings.
 	// The reason for this is that systemd's application of DeviceAllow rules
 	// is done disruptively, resulting in spurrious errors to common devices
 	// (unlike our fs driver, they will happily write deny-all rules to running
 	// containers). So we freeze the container to avoid them hitting the cgroup
 	// error. But if the freezer cgroup isn't supported, we just warn about it.
-	if err := m.Freeze(configs.Frozen); err != nil {
-		logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
+	targetFreezerState := configs.Undefined
+	if !m.cgroups.SkipDevices {
+		// Figure out the current freezer state, so we can revert to it after we
+		// temporarily freeze the container.
+		targetFreezerState, err = m.GetFreezerState()
+		if err != nil {
+			return err
+		}
+		if targetFreezerState == configs.Undefined {
+			targetFreezerState = configs.Thawed
+		}
+
+		if err := m.Freeze(configs.Frozen); err != nil {
+			logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
+		}
 	}
 
 	if err := dbusConnection.SetUnitProperties(getUnitName(container.Cgroups), true, properties...); err != nil {
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -298,24 +298,27 @@ func (m *unifiedManager) Set(container *configs.Config) error {
 		return err
 	}
 
-	// Figure out the current freezer state, so we can revert to it after we
-	// temporarily freeze the container.
-	targetFreezerState, err := m.GetFreezerState()
-	if err != nil {
-		return err
-	}
-	if targetFreezerState == configs.Undefined {
-		targetFreezerState = configs.Thawed
-	}
-
 	// We have to freeze the container while systemd sets the cgroup settings.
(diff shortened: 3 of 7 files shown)


libcontainer/configs/cgroup_linux.go
```diff
@@ -126,4 +126,12 @@ type Resources struct {
 
 	// CpuWeight sets a proportional bandwidth limit.
 	CpuWeight uint64 `json:"cpu_weight"`
+
+	// SkipDevices allows to skip configring device permissions.
```
mrunalp, line 130: typo: configuring

kolyshkin:
@cyphar PTAL

libcontainer/cgroups/fs/fs.go
```diff
@@ -207,6 +207,9 @@ func (m *manager) Apply(pid int) (err error) {
 			if cgroups.IsNotFound(err) && sys.Name() != "devices" {
```
cyphar, line 207: NIT: While it might be a bit long, I think `if cgroups.IsNotFound(err) && (c.SkipDevices || sys.Name() != "devices")` gets the intention of this logic across a little better. We also probably don't want to mask all `d.path` errors if `SkipDevices` is enabled -- just the devices cgroup ones.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -72,11 +72,13 @@ func genV1ResourcesProperties(c *configs.Cgroup, conn *systemdDbus.Conn) ([]syst
...
-	deviceProperties, err := generateDeviceProperties(r.Devices)
-	if err != nil {
-		return nil, err
+	if !c.SkipDevices {
+		deviceProperties, err := generateDeviceProperties(r.Devices)
+		if err != nil {
+			return nil, err
+		}
+		properties = append(properties, deviceProperties...)
 	}
```
cyphar, line 81: While I do understand why you changed this, please be aware that disabling this doesn't make systemd not do anything with the devices cgroup, it will actually place it into allow-everything mode.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -43,11 +43,13 @@ func genV2ResourcesProperties(c *configs.Cgroup, conn *systemdDbus.Conn) ([]syst
...
-	deviceProperties, err := generateDeviceProperties(r.Devices)
-	if err != nil {
-		return nil, err
+	if !c.SkipDevices {
+		deviceProperties, err := generateDeviceProperties(r.Devices)
+		if err != nil {
+			return nil, err
+		}
+		properties = append(properties, deviceProperties...)
 	}
```
cyphar, line 52: Same here.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -298,24 +300,27 @@ func (m *unifiedManager) Set(container *configs.Config) error {
...
+			return err
+		}
+		if targetFreezerState == configs.Undefined {
+			targetFreezerState = configs.Thawed
+		}
+
+		if err := m.Freeze(configs.Frozen); err != nil {
+			logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
+		}
 	}
```
cyphar, line 324: You need to add this conditional to the v1 systemd cgroup code.

libcontainer/configs/cgroup_linux.go
```diff
@@ -126,4 +126,11 @@ type Resources struct {
 
 	// CpuWeight sets a proportional bandwidth limit.
 	CpuWeight uint64 `json:"cpu_weight"`
+
+	// SkipDevices allows to skip configuring device permissions.
+	// Used by e.g. kubelet while creating a parent cgroup (kubepods)
+	// common for many containers.
+	//
+	// NOTE it is impossible to start a container which has this flag set.
```
cyphar, line 134: :+1: That does resolve my security concerns about this feature.

libcontainer/cgroups/fs/fs.go
kolyshkin: right, I'm not sure why I wrote it like this :-\

libcontainer/cgroups/systemd/v2.go
kolyshkin: right, will do

libcontainer/cgroups/systemd/v1.go
kolyshkin: Thank you, fixed.

kolyshkin:
Updated to address @cyphar review comments.

cyphar:
@kolyshkin 

One question about the justification for this change:

> The problem (for cgroupv2 that uses eBPF for device configuration) is
> the hard requirement to have devices cgroup configured results in
> leaking an eBPF program upon every kubelet restart. program. If kubelet
> is restarted 64+ times, the cgroup can't be configured anymore.

That's a bug in runc (#2366 is tracking that), and is something we need to fix anyway (we should be either doing `BPF_F_REPLACE` or manually replacing the program inatomically, or even possibly reimplementing our eBPF rules to use eBPF maps that we can atomically replace).

Is that the only reason for this change (looking back at #2474, it looks like I didn't catch this when discussing the need for this)? I'd be happy to have it anyway (since the requirement for the devices cgroup doesn't make sense in certain circumstances), but I just want to make sure that we're aware that this doesn't fix the underlying issue (which will be at least somewhat more involved).

kolyshkin:
I do agree this looks like a workaround for #2366.

From the other side, it does not make sense to require devices to be configured in cgroups that are not used to run containers, but rather to hold other cgroups (not sure what's the term here, parent cgroups? meta cgroups?). That is also the reason why I am thinking about renaming `SkipDevices` to `ParentCgroup` or something like that (although currently it only does what `SkipDevices` says).

libcontainer/cgroups/systemd/v1.go
kolyshkin: So, if the default is deny-all and a parent cgroup's ebpf is consulted first before the child cgroup, this apparently (
In view of https://github.com/kubernetes/kubernetes/pull/92862#pullrequestreview-444255169) results in no devices accessible.

I guess I'll have to revert to the original hunk here, leaving systemd unconfigured wrt devices, which should result in it using `DevicePolicy=auto`.

PR #2493, opened 2020-07-02
vendor: update cilium/ebpf v0.0.0-20200702112145-1c8d4c9ef775

thaJeztah:
full diff: https://github.com/cilium/ebpf/compare/a9f01edf17e3...1c8d4c9ef775

drops support for go1.12, and removes dependency on the golang.org/x/xerrors
transitional package.


dims:
LGTM

PR #2438, opened 2020-05-29
[DO NOT MERGE] cgroup v1: optimize code to avoid parsing mountinfo if possible

kolyshkin:
This set of patches brings down a number of times we parse /proc/self/mountinfo during container start+stop:

* with systemd: from 76 (49 after #2446) to 6 times.
* without systemd: from 116 (59 after #2446) to 16.

(there is some improvement for other operations as well but I haven't measured those)

Ideally I'd like to bring it down to less than 5 per runc run, but this is already a super good improvement.

**Update**: this is being splitted into multiple smaller PRs to facilitate easier reviewing. The following list is just for my convenience to make sure I haven't forgot anything.

- [x] cgroupv1/FindCgroupMountpoint: add a fast path | commit 38ab3b2 | PR #2499
- [x] cgroupv1/Apply: do not overuse path/getSubsystemPath | commit 89bd4c1 | PR #2494
- [x] cgroupv1/systemd: rework Apply/joinCgroups | commit 6f6c72d | PR #2494
- [x] cgroupv1/systemd: (re)use m.paths | commit 55c1ce2 | PR #2494
- [x] cgroups/fs: rm getSubsystems | commit db45ce7 | PR #2498
- [x] cgroups/fs/path: optimize | commit c7147ed | PR #2495
- [x] cgroups/fs: rm Remove method from controllers | commit 7ce1857 | PR #2498
- [x] cgroup/fs: rework Apply() | commit 576d617 | PR #2500
- [x] cgroupv1/freezer: don't use subsystemSet.Get() | commit 0bb5f5c | PR #2496
- [x] cgroups/fs/Freeze: simplify | commit cffb553 | PR #2496
- [x] cgroupv1: remove subsystemSet.Get() | commit 44a02d2 | PR #2498
- [x] cgroups/fs tests: unify TestInvalidXXXCgroupXXX | commit f78b570 | PR #2497
- [x] cgroup/fs: remove getCgroupRoot | commit 5aef675 | PR #2497

And here is the list of PRs this one was split into:
- [x] #2494 cgroups/v1: optimize path usage
- [x] #2495 libct/cgroups/fs: simple optimization
- [x] #2496 libct/cgroups/fs: simplify/speedup freezer code
- [x] #2497 libct/cgroups/fs: rm getCgroupRoot (or the alternative: #2507)
- [x] #2498 libct/cgroups/fs: code cleanups
- [x] #2499 cgroupv1/FindCgroupMountpoint: add a fast path
[...]

PR #2494, opened 2020-07-03
cgroups/v1: optimize path usage

kolyshkin:

Limit the use of `getSubsystemPath()` and `d.path()` functions, reusing the `m.paths` instead.

See individual commit descriptions for details.

Number of times mountinfo was parsed during `runc --systemd-cgroup run` of a busybox container running `sleep 1` went down from 50 to 19 as a result of these changes.

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -220,24 +220,24 @@ func (m *manager) Apply(pid int) (err error) {
 
 	var c = m.cgroups
 
+	m.paths = make(map[string]string)
+	if c.Paths != nil {
+		cgMap, err := cgroups.ParseCgroupFile("/proc/self/cgroup")
+		if err != nil {
+			return err
+		}
+		for name, path := range c.Paths {
+			// XXX(kolyshkin@): why this check is needed?
+			if _, ok := cgMap[name]; ok {
+				m.paths[name] = path
+			}
+		}
+		return cgroups.EnterPid(m.paths, pid)
+	}
+
 	d, err := getCgroupData(m.cgroups, pid)
 	if err != nil {
 		return err
-	}
-
-	m.paths = make(map[string]string)
-	if c.Paths != nil {
-		for name, path := range c.Paths {
-			_, err := d.path(name)
-			if err != nil {
-				if cgroups.IsNotFound(err) {
-					continue
-				}
-				return err
-			}
-			m.paths[name] = path
-		}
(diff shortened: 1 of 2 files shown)


mrunalp:
Needs rebase.

kolyshkin:
rebased

kolyshkin:
@AkihiroSuda PTAL

libcontainer/cgroups/fs/fs.go
```diff
@@ -220,26 +220,26 @@ func (m *manager) Apply(pid int) (err error) {
...
-	d, err := getCgroupData(m.cgroups, pid)
-	if err != nil {
-		return err
-	}
-
 	m.paths = make(map[string]string)
 	if c.Paths != nil {
 		for name, path := range c.Paths {
-			_, err := d.path(name)
+			cgMap, err := cgroups.ParseCgroupFile("/proc/self/cgroup")
```
kolyshkin, line 226: This obviously need to be taken out of the `for ... range c.Paths` loop!

kolyshkin: otherwise we parse /proc/self/cgroup a few times which is absolutely not needed

kolyshkin: updated

libcontainer/cgroups/fs/fs.go
```diff
@@ -220,26 +220,26 @@ func (m *manager) Apply(pid int) (err error) {
...
+			return err
+		}
 		for name, path := range c.Paths {
-			_, err := d.path(name)
-			if err != nil {
-				if cgroups.IsNotFound(err) {
-					continue
-				}
-				return err
+			// XXX(kolyshkin@): why this check is needed?
```
AkihiroSuda, line 230: What happens without this check?

kolyshkin: As far as I can see, nothing bad. The `paths` map is then used to write pid to cgroups, and `WriteCgroupProc` has a check that path exists.

There may be other scenarios (with software that uses runc/libcontainer) I don't envision, and I have broken stuff in the past, so I am trying to be extra careful here.

So, ideally, I'd remove this, but maybe later, thus `TODO`.

The new check is already 100x better than the old one, so this patch is a big improvement -- I'm just not sure if the check is needed at all.

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda (approved):

PR #2495, opened 2020-07-03
libct/cgroups/fs: simple optimization

kolyshkin:

This adds a very simple and straightforward optimization.

Should be very easy to review.


kolyshkin:
> This drastically reduces the number of calls to FindCgroupMountpoint
> during container start (from 116 to 62 in my setup).

Note this code was written and the measurements were done before merging https://github.com/opencontainers/runc/pull/2446. Now the savings should be less dramatic, but still major.

kolyshkin:
@AkihiroSuda @cyphar PTAL (easy to review)

PR #2497, opened 2020-07-03
libct/cgroups/fs: rm getCgroupRoot

kolyshkin:

I think this was added during the dark ages where cgroupfs can be
mounted to /cgroup, /dev/cgroup, or some other path. Now it have
finally settled on /sys/fs/cgroup, so let's remove the guessing
and use the standard path.
    
If you think this one is too extreme (because e.g. there are systems on which cgroup v1 mounts are not under `/sys/fs/cgroup` but some other path), please let us know, and see the alternative PR (#2507).

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -3,11 +3,9 @@
 package fs
 
 import (
-	"bufio"
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 	"sync"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
Do we have this officially in OCI Spec?

Wondering this might break some embedded systems (if the previous code was actually functional with a strange mount point)

kolyshkin:
> Do we have this officially in OCI Spec?

I was not able to find anything. I guess it is considered to be an implementation detail.

The closest thing I was able to find is

> The value of cgroupsPath MUST be either an absolute path or a relative path.
> 
> * In the case of an absolute path (starting with /), the runtime MUST take the path to be relative to the cgroups mount point.
> * In the case of a relative path (not starting with /), the runtime MAY interpret the path relative to a runtime-determined location in the cgroups hierarchy.

So, I consider this is an implementation detail and should not be in the spec.

> Wondering this might break some embedded systems (if the previous code was actually functional with a strange mount point)

It might. For example, Android mounts various v1 cgroups to `/dev/blkio`, `/dev/cpuctl`, `/dev/cpuset`, `/dev/memcg`, `/acct` etc. This won't work even with either the current code, since it has an assumption that all cgroups are mounted into directories that have the same parent dir.

I had an alternative to this patch -- that is, try the default (/sys/fs/cgroup) first and fall back to mountinfo parsing it if failed.

kolyshkin:
I traced the introduction of this way to get a cgroup root directory down to this commit:

https://github.com/moby/moby/pull/4327/commits/c44258630575f70231b11fb55bc4edc3fb677cab#diff-4ecfc3616c252e4c65bc87c539d9a459L18

Alas, it does not tell us why.

But let me work on a less radical alternative.

kolyshkin:
> But let me work on a less radical alternative.

Please see https://github.com/opencontainers/runc/pull/2507

kolyshkin:
Rebased to resolve merge conflicts

kolyshkin:
As much as I like this one, #2507 is now merged so let's close this one. Maybe revisit it in 5 years? :)

PR #2496, opened 2020-07-03
libct/cgroups/fs: simplify/speedup freezer code

kolyshkin:
_this is separated out from https://github.com/opencontainers/runc/pull/2438 in order to make review easier_

Make things faster and code more readable by simplification and code removal.

Please see individual commits for details.

PR #2498, opened 2020-07-03
libct/cgroups/fs: code cleanups

kolyshkin:

This contains some minor cleanups and optimizations, but mostly just code removal.

~~TODO: add~~
~~* https://github.com/opencontainers/runc/commit/db45ce7f21d33a54e14492b6ef6816e3eafcf1fc~~
~~* https://github.com/opencontainers/runc/commit/44a02d2dfe6d7c2db7e2fa4b326d41771ce656d5~~
~~once https://github.com/opencontainers/runc/pull/2496 is merged~~

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -250,7 +250,7 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 	stats := cgroups.NewStats()
 	for name, path := range m.paths {
 		sys, err := m.getSubsystems().Get(name)
-		if err == errSubsystemDoesNotExist || !cgroups.PathExists(path) {
+		if err == errSubsystemDoesNotExist {
 			continue
 		}
 		if err := sys.GetStats(path, stats); err != nil {
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -18,7 +18,7 @@ import (
 )
 
 var (
-	subsystemsLegacy = subsystemSet{
+	subsystems = subsystemSet{
 		&CpusetGroup{},
 		&DevicesGroup{},
 		&MemoryGroup{},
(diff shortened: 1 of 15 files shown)


kolyshkin:
For the reference, here's the benchmark used for the last commit
```go
package fs

import (
	"testing"
)

func initPaths() map[string]string {
	paths := make(map[string]string)

	for _, ss := range subsystems {
		paths[ss.Name()] = "/sys/fs/cgroup/" + ss.Name() + "/user.slice/user-1000.slice/session-881.scope/xx88"

	}

	return paths
}

func BenchmarkIterateOverPaths(b *testing.B) {
	paths := initPaths()

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for name, path := range paths {
			ss, err := subsystems.Get(name)
			if err == errSubsystemDoesNotExist || path == "" || ss == nil {
				continue
			}
		}
	}
}

func BenchmarkIterateOverSubsystems(b *testing.B) {
	paths := initPaths()

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for _, ss := range subsystems {
			path := paths[ss.Name()]
			if path == "" {
				continue
			}
		}
	}
}
```

libcontainer/cgroups/fs/fs.go
```diff
@@ -38,24 +38,11 @@ var (
...
-	}
-	return nil, errSubsystemDoesNotExist
-}
-
 type subsystem interface {
 	// Name returns the name of the subsystem.
 	Name() string
 	// Returns the stats, as 'stats', corresponding to the cgroup under 'path'.
 	GetStats(path string, stats *cgroups.Stats) error
-	// Removes the cgroup represented by 'cgroupData'.
```
mrunalp, line 57: This could be of use to external clients and maybe later on if/when we factor this out into a cgroups library.

kolyshkin: Theoretically, yes.

Practically,
1. I don't see it being used by kubernetes;
2. It uses `d.path` which is expensive, it requires to parse mountinfo (this is actually the main point of it being removed);
3. Its API is not in line with e.g. `Set` (which does not use `d.path` but the path is provided directly).

Also, if needed, it can be trivial to reimplement / re-add, as it doesn't do anything special, so I see no point in keeping it.

PR #2499, opened 2020-07-03
cgroupv1/FindCgroupMountpoint: add a fast path

kolyshkin:

In case cgroupPath is under the default cgroup prefix, let's try to
guess the mount point by adding the subsystem name to the default
prefix, and resolving the resulting path in case it's a symlink.

In most cases, given the default cgroup setup, this trick
should result in returning the same result faster, and avoiding
/proc/self/mountinfo parsing which is relatively slow and problematic.

Be very careful with the default path, checking it is
 - a directory;
 - a mount point;
 - has cgroup fstype.

If something is not right, fall back to parsing mountinfo.

While at it, remove the obsoleted comment about mountinfo parsing.  The
comment belongs to findCgroupMountpointAndRootFromReader(), but rather
than moving it there, let's just remove it, since it does not add any
value in understanding the current code.

### DIFF
--- a/libcontainer/cgroups/v1_utils.go
+++ b/libcontainer/cgroups/v1_utils.go
@@ -8,6 +8,10 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+	"golang.org/x/sys/unix"
 )
 
 // Code in this source file are specific to cgroup v1,
@@ -15,6 +19,7 @@ import (
 
 const (
 	CgroupNamePrefix = "name="
+	defaultPrefix    = "/sys/fs/cgroup"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/v1_utils.go
```diff
@@ -43,11 +49,59 @@ func IsNotFound(err error) bool {
...
+
+	if st.Sys().(*syscall.Stat_t).Dev == pst.Sys().(*syscall.Stat_t).Dev {
+		// parent dir has the same dev -- path is not a mount point
+		return ""
+	}
+
+	// (3) path should have 'cgroup' fs type.
+	fst := unix.Statfs_t{}
+	err = unix.Statfs(path, &fst)
+	if err != nil || fst.Type != unix.CGROUP_SUPER_MAGIC {
```
AkihiroSuda, line 87: Do we want to allow FUSE magic as well? (for LXCFS)

libcontainer/cgroups/v1_utils.go
```diff
@@ -19,6 +23,8 @@ const (
 
 var (
 	errUnified = errors.New("not implemented for cgroup v2 unified hierarchy")
+
+	defaultPrefix = "/sys/fs/cgroup"
```
kolyshkin, line 27: hmm, this should have been const.

libcontainer/cgroups/v1_utils.go
kolyshkin: Well, the code of FindCgroupMountpoint do not have support for lxcfs either, so it would be useless to add it here.

Besides, it seems like this feature is obsoleted (by cgroupns). From https://github.com/lxc/lxcfs

> Prior to the implementation of cgroup namespaces by Serge Hallyn LXCFS also provided a container aware cgroupfs tree. It took care that the container only had access to cgroups underneath it's own cgroups and thus provided additional safety. For systems without support for cgroup namespaces LXCFS will still provide this feature but it is mostly considered deprecated.

kolyshkin: > Well, the code of FindCgroupMountpoint do not have support for lxcfs either, so it would be useless to add it here

Unless, of course, LXCFS shadows /proc/self/mountinfo as well, in this case we might need to support it here as well.

Summoning @hallyn @brauner @stgraber -- what do you guys think?

stgraber: lxcfs cannot shadow /proc/self/* because those entries show up dynamically or are magic links so a bit incompatible with bind-mounts.

The fuse cgroupfs view is currently mostly unused thanks to cgns doing a much cleaner job but it may make a come back as we need a solution for running containers that are only cgroupv1 aware on cgroupv2-only hosts and there hasn't been much interest for an in-kernel solution to that...

kolyshkin: > we need a solution for running containers that are only cgroupv1 aware on cgroupv2-only hosts

This is hardly a runc use case.

Anyway, since supporting lxcfs requires more changes than just modifying this code, it is probably out of scope for this PR and can be implemented separately later.

libcontainer/cgroups/v1_utils.go
kolyshkin: fixed

kolyshkin:
CI is green but apparently github lost the notification. Restarted travis build.

kolyshkin:
@AkihiroSuda @cyphar PTAL

PR #2500, opened 2020-07-03
libct/cgroups/fs: rework Apply()

kolyshkin:

In manager.Apply() method, a path to each subsystem is obtained by
calling d.path(sys.Name()), and the sys.Apply() is called that does
the same call to d.path() again.

d.path() is an expensive call, so rather than to call it twice, let's
reuse the result.

This results the number of times we parse mountinfo during container
start from 62 to 34 on my setup.


AkihiroSuda:
needs rebase

kolyshkin:
Rebased to resolve a conflict caused by #2490 being merged.

PR #2331, opened 2020-04-20
check that StartTransientUnit/StopUnit succeeds

lifubang:
fix #2313(relate to #2310) #2344 

To fix #2313 
We can check whether `StartTransientUnit` succeeds or not with the value from the channel.
If we got `failed`, we should run `ResetFailedUnit` and return the error.

```
[root@localhost systemdtest]# runc --systemd-cgroup run -d test
WARN[0000] signal: killed                               
ERRO[0000] cannot detect unified path                   
ERRO[0000] container_linux.go:349: starting container process caused "process_linux.go:306: applying cgroup configuration for process caused \"StartTransientUnit job is not done: failed\"" 
container_linux.go:349: starting container process caused "process_linux.go:306: applying cgroup configuration for process caused \"StartTransientUnit job is not done: failed\""
```
The error `ERRO[0000] cannot detect unified path` relates to #2329 .

To fix #2344 
We should wait `StopUnit` succeeds.

Signed-off-by: lifubang <lifubang@acmcoder.com>

PR #2503, opened 2020-07-05
cgroup, systemd: cleanup cgroups 

giuseppe:
some hierarchies were created directly by .Apply() on top of systemd managed cgroups.  systemd doesn't manage these and as a result we leak these cgroups.

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -222,7 +222,14 @@ func (m *legacyManager) Destroy() error {
 		return err
 	}
 	unitName := getUnitName(m.cgroups)
-	if err := stopUnit(dbusConnection, unitName); err != nil {
+
+	err = stopUnit(dbusConnection, unitName)
+	// Both on success and on error, cleanup all the cgroups we are aware of.
+	// Some of them were created directly by Apply() and are not managed by systemd.
+	if err := cgroups.RemovePaths(m.paths); err != nil {
+		return err
+	}
+	if err != nil {
 		return err
 	}
 	m.paths = make(map[string]string)
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -12,24 +12,24 @@ function teardown() {
 }
 
 @test "runc delete" {
-  # run busybox detached
-  runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  runc run -d --console-socket $CONSOLE_SOCKET testbusyboxdelete
   [ "$status" -eq 0 ]
 
-  # check state
-  testcontainer test_busybox running
+  testcontainer testbusyboxdelete running
 
-  runc kill test_busybox KILL
+  runc kill testbusyboxdelete KILL
   [ "$status" -eq 0 ]
-  # wait for busybox to be in the destroyed state
-  retry 10 1 eval "__runc state test_busybox | grep -q 'stopped'"
+  retry 10 1 eval "__runc state testbusyboxdelete | grep -q 'stopped'"
 
-  # delete test_busybox
-  runc delete test_busybox
+  runc delete testbusyboxdelete
   [ "$status" -eq 0 ]
 
-  runc state test_busybox
+  runc state testbusyboxdelete
   [ "$status" -ne 0 ]
+
+  run find /sys/fs/cgroup -wholename '*testbusyboxdelete*' -type d
+  [ "$status" -eq 0 ]
+  [ "$output" = "" ] || fail "cgroup not cleaned up correctly: $output"
 }
 
 @test "runc delete --force" {
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -27,6 +27,9 @@ var (
 	versionOnce sync.Once
 	version     int
 	versionErr  error
+
+	isRunningSystemdOnce sync.Once
+	isRunningSystemd     bool
 )
 
 // NOTE: This function comes from package github.com/coreos/go-systemd/util
@@ -37,11 +40,11 @@ var (
 // checks whether /run/systemd/system/ exists and is a directory.
 // http://www.freedesktop.org/software/systemd/man/sd_booted.html
 func IsRunningSystemd() bool {
-	fi, err := os.Lstat("/run/systemd/system")
-	if err != nil {
-		return false
-	}
-	return fi.IsDir()
+	isRunningSystemdOnce.Do(func() {
+		fi, err := os.Lstat("/run/systemd/system")
+		isRunningSystemd = err == nil && fi.IsDir()
+	})
+	return isRunningSystemd
 }
 
 // systemd represents slice hierarchy using `-`, so we need to follow suit when


libcontainer/cgroups/systemd/v1.go
```diff
@@ -222,7 +223,14 @@ func (m *legacyManager) Destroy() error {
 		return err
 	}
 	unitName := getUnitName(m.cgroups)
-	if err := stopUnit(dbusConnection, unitName); err != nil {
+
+	err = stopUnit(dbusConnection, unitName)
+	// Both on success and on error, cleanup all the cgroups we are aware of.
+	// Some of them were created directly by Apply() and are not managed by systemd.
+	for _, path := range m.paths {
+		unix.Rmdir(path)
```
tedyu, line 231: There is rmdir(path string) in libcontainer/cgroups/fs2/fs2.go

Can that func be used ?
It seems we should return error if the cleanup isn't successful.

Thanks

giuseppe: it is a different package and rmdir is not exported.

I've added the error check.

tedyu:
lgtm

tests/integration/delete.bats
```diff
@@ -13,23 +13,27 @@ function teardown() {
...
-  # delete test_busybox
-  runc delete test_busybox
+  # delete the container
+  runc delete testbusyboxdelete
   [ "$status" -eq 0 ]
 
   runc state test_busybox
   [ "$status" -ne 0 ]
+
+  if test -n "$(find /sys/fs/cgroup -wholename '*testbusyboxdelete*')"; then
```
AkihiroSuda, line 34: Does busybox support `-wholename`? This flag doesn't shown up in `busybox find --help`

giuseppe: it seems to be supported:

```
$ podman run --rm -ti alpine find /usr -wholename '*cat*'
/usr/bin/bzcat
/usr/bin/fallocate
/usr/bin/lzcat
/usr/bin/lzopcat
/usr/bin/truncate
/usr/bin/xzcat
```

giuseppe:
@AkihiroSuda is the CI failure a flake?

AkihiroSuda:
```
$ sudo ssh default -t 'cd /vagrant && sudo make localunittest'
sudo: make: command not found
```

Could you try restarting CI? If the error persists, `make` would need to be explicitly dnf-installed in Vagrantfile.

kolyshkin:
I was too naive to think lstat(2) is a cheap syscall, and so I wrote a benchmark for `IsRunningSystemd`, to see how much the first commit in this PR helps:

```go
func BenchmarkIsRunningSystemd(b *testing.B) {
	var res bool
	for i := 0; i < b.N; i++ {
		res = IsRunningSystemd()
	}
	b.Logf("IsRunningSystemd: %v", res)
}
```

Before this PR:

```
BenchmarkIsRunningSystemd-8   	 1926032	       614 ns/op
```

After:

```
BenchmarkIsRunningSystemd-8   	637496562	         1.81 ns/op
```

Out of pure curiosity I also tried using `unix.Lstat` instead of `os.Lstat`:

```go
func IsRunningSystemd() bool {
        var st unix.Stat_t
        err := unix.Lstat("/run/systemd/system", &st)
        if err != nil {
                return false
        }
        return st.Mode&unix.S_IFDIR == unix.S_IFDIR
}
```

and the result is roughly 20% savings:

```
BenchmarkIsRunningSystemd-8   	 2321263	       506 ns/op
```

So, yes, using `sync.Once` totally makes sense here.

kolyshkin:
CI failure is a glitch (or we should not rely on github to provide `umoci` binary):

```
Step 13/24 : RUN curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64     && chmod +x /usr/local/bin/umoci
 ---> Running in e33fd55181d1
curl: (22) The requested URL returned error: 429 too many requests
The command '/bin/sh -c curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64     && chmod +x /usr/local/bin/umoci' returned a non-zero code: 22
Makefile:62: recipe for target 'runcimage' failed
```

CI restarted

tests/integration/delete.bats
```diff
@@ -13,23 +13,27 @@ function teardown() {
...
-  runc kill test_busybox KILL
+  runc kill testbusyboxdelete KILL
   [ "$status" -eq 0 ]
   # wait for busybox to be in the destroyed state
-  retry 10 1 eval "__runc state test_busybox | grep -q 'stopped'"
+  retry 10 1 eval "__runc state testbusyboxdelete | grep -q 'stopped'"
 
-  # delete test_busybox
-  runc delete test_busybox
+  # delete the container
```
kolyshkin, line 27: nit: maybe drop this comment (and the likes above -- "run busybox detached", "check state", "wait for busybox ..."), they are entirely useless.

kolyshkin:
Alas, the test case is not catching the bug (I have reverted the fixing part):
```
$ git show libcontainer/cgroups/systemd/v1.go | patch -p1 -R
patching file libcontainer/cgroups/systemd/v1.go

$ sudo make localintegration TESTPATH=/delete.bats
[sudo] password for kir: 
go build "-mod=vendor" "-buildmode=pie"  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="2185e9c4aa7b3e91b14e1073176ec3ee296ff665-dirty" -X main.version=1.0.0-rc91+dev " -o runc .
go build "-mod=vendor" "-buildmode=pie"  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="2185e9c4aa7b3e91b14e1073176ec3ee296ff665-dirty" -X main.version=1.0.0-rc91+dev " -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
bats -t tests/integration/delete.bats
1..4
ok 1 runc delete
ok 2 runc delete --force
ok 3 runc delete --force ignore not exist
ok 4 runc delete --force in cgroupv2 with subcgroups # skip test requires cgroups_v2
```

tests/integration/delete.bats
giuseppe: done

giuseppe:
> Alas, the test case is not catching the bug (I have reverted the fixing part):

are you using `sudo make localintegration TESTPATH=/delete.bats RUNC_USE_SYSTEMD=1` on cgroup v1?

kolyshkin:
> are you using sudo make localintegration TESTPATH=/delete.bats RUNC_USE_SYSTEMD=1 on cgroup v1?

Oh, I forgot `RUNC_USE_SYSTEMD=1`, most probably due to lack of espresso :-\

Yes, the test is working

libcontainer/cgroups/systemd/v1.go
```diff
@@ -222,7 +223,16 @@ func (m *legacyManager) Destroy() error {
 		return err
 	}
 	unitName := getUnitName(m.cgroups)
-	if err := stopUnit(dbusConnection, unitName); err != nil {
+
+	err = stopUnit(dbusConnection, unitName)
+	// Both on success and on error, cleanup all the cgroups we are aware of.
+	// Some of them were created directly by Apply() and are not managed by systemd.
+	for _, path := range m.paths {
+		if err := unix.Rmdir(path); err != nil && err != unix.ENOENT {
```
kolyshkin, line 231: :+1: for using rmdir(2)

tests/integration/delete.bats
```diff
@@ -12,24 +12,24 @@ function teardown() {
...
-  # delete test_busybox
-  runc delete test_busybox
+  runc delete testbusyboxdelete
   [ "$status" -eq 0 ]
 
-  runc state test_busybox
+  runc state testbusyboxdelete
   [ "$status" -ne 0 ]
+
+  if test -n "$(find /sys/fs/cgroup -wholename '*testbusyboxdelete*')"; then
```
kolyshkin, line 30: nit: maybe add `-type d` and print what's found, also check find exit code (to prevent false passing).

Something like
```bash
  run find /sys/fs/cgroup -wholename '*testbusyboxdelete*' -type d
  [ "$status" -eq 0 ]
  [ "$output" = "" ] || fail "cgroup not cleaned up correctly: $output"
```

which results in
```
1..4
not ok 1 runc delete
# (from function `fail' in file tests/integration/helpers.bash, line 226,
#  in test file tests/integration/delete.bats, line 40)
#   `[ "$output" = "" ] || fail "cgroup not cleaned up correctly: $output"' failed
....
# cgroup not cleaned up correctly: /sys/fs/cgroup/perf_event/system.slice/runc-testbusyboxdelete.scope
# /sys/fs/cgroup/freezer/system.slice/runc-testbusyboxdelete.scope
# /sys/fs/cgroup/hugetlb/system.slice/runc-testbusyboxdelete.scope
# /sys/fs/cgroup/net_cls,net_prio/system.slice/runc-testbusyboxdelete.scope
# /sys/fs/cgroup/cpuset/system.slice/runc-testbusyboxdelete.scope
...
```

feel free to ignore though

kolyshkin:
The regression is caused by commit bfa1b2aab39f7735db9319478b8f63b9adca26e3 (PR https://github.com/opencontainers/runc/pull/2331).

libcontainer/cgroups/systemd/v1.go
kolyshkin: On a second though, we have used `cgroups.RemovePaths()` which for some historical reasons does some ugly tricks, which might be required (on some older kernels? IDK), so maybe use it instead.

kolyshkin: One reason to use `cgroups.RemovePaths()` is in case a container creates sub-cgroups, `rmdir` might not work (depending on the kernel I guess).

giuseppe: thanks for the hint.  Fixed

tests/integration/delete.bats
giuseppe: amended

libcontainer/cgroups/systemd/v1.go
kolyshkin: We have a test with sub-cgroups but it's v2-specific. I am writing a similar one for v1

PR #2505, opened 2020-07-06
fix redundant source code copy issue

XiaodongLoong:
There is a -v flag to unittest, integration, rootlessintegration tests inside container in Makefile.
IMO, it do not have to copy source code to container.

Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,5 +95,3 @@ ENV DEBIAN_ROOTFS /debian
 RUN mkdir -p "${DEBIAN_ROOTFS}"
 RUN . tests/integration/multi-arch.bash \
     && get_and_extract_debian "$DEBIAN_ROOTFS"
-
-COPY . .


kolyshkin:
@thaJeztah PTAL

kolyshkin:
nit: `s/souce/source/` in the commit subject

cyphar:
There was a reason for doing this originally -- you could in theory do `docker run runc_image ...` and not have to deal with `-v` paths, but at the same time you should be able to have a `make shell` where the source code will be updated if you change it on the host. But to be honest, I don't remember the last time I used this feature so I'm fine with removing it.

XiaodongLoong:
@AkihiroSuda PTAL

thaJeztah:
> There was a reason for doing this originally -- you could in theory do docker run runc_image ... and not have to deal with -v paths, but at the same time you should be able to have a make shell where the source code will be updated if you change it on the host. But to be honest, I don't remember the last time I used this feature so I'm fine with removing it.

Yes, I was looking for a similar construct as we have in moby (where the bind-mount is optional in the Makefile), but didn't see one. There's definitely valid use-cases to think of, just that it currently doesn't seem to be used.

Do you think we should have a separate build-stage /  `--target` for that use-case?

XiaodongLoong:
I think the source files often change, and build new Docker image frequently with the source `COPY` operation.
When we check the PR by CI, the new image will be built.

thaJeztah:
yeah, it depends on the use-case (e.g. `-v` doesn't work when connecting to a remote machine)

kolyshkin:
> Do you think we should have a separate build-stage / --target for that use-case?

Could be addressed as a followup.

PR #2506, opened 2020-07-07
cgroupv1 removal nits

kolyshkin:
This set is a followup to https://github.com/opencontainers/runc/pull/2503. It does the following:

 - moves cleaning of paths map to `RemovePaths` so its callers don't have to;
 - adds `cgroups.RemovePath` (code move from `fs2.removeCgroupPath`);
 - optimizes `cgroups.RemovePaths()` by reusing `RemovePath` instead of `os.RemoveAll`;
 - adds a test case of cgroupv1 + sub-cgroups removal.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -230,6 +230,7 @@ func RemovePaths(paths map[string]string) (err error) {
 			}
 		}
 		if len(paths) == 0 {
+			paths = make(map[string]string)
 			return nil
 		}
 	}
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -207,6 +207,43 @@ func EnterPid(cgroupPaths map[string]string, pid int) error {
 	return nil
 }
 
+func rmdir(path string) error {
+	err := unix.Rmdir(path)
+	if err == nil || err == unix.ENOENT {
+		return nil
+	}
+	return &os.PathError{Op: "rmdir", Path: path, Err: err}
+}
+
+// RemovePath aims to remove cgroup path. It does so recursively,
+// by removing any subdirectories (sub-cgroups) first.
+func RemovePath(path string) error {
+	// try the fast path first
+	if err := rmdir(path); err == nil {
+		return nil
+	}
+
+	infos, err := ioutil.ReadDir(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = nil
+		}
+		return err
+	}
+	for _, info := range infos {
+		if info.IsDir() {
+			// We should remove subcgroups dir first
+			if err = RemovePath(filepath.Join(path, info.Name())); err != nil {
+				break
+			}
+		}
+	}
+	if err == nil {
+		err = rmdir(path)
+	}
+	return err
+}
+
 // RemovePaths iterates over the provided paths removing them.
 // We trying to remove all paths five times with increasing delay between tries.
 // If after all there are not removed cgroups - appropriate error will be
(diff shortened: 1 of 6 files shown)


kolyshkin:
@giuseppe @mrunalp @AkihiroSuda PTAL

cyphar:
New integration test fails on centos:

```
not ok 26 runc delete --force in cgroupv1 with subcgroups
# (in test file tests/integration/delete.bats, line 66)
# runc list (status=0):
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/console.sock test_busybox (status=1):
# time="2020-07-07T01:11:20Z" level=error msg="cgroup namespaces aren't enabled in the kernel"
# runc list (status=0):
```

kolyshkin:
Amended the test case to require cgroupns (testing without it is kinda cumbersome)

libcontainer/cgroups/utils.go
```diff
@@ -249,15 +250,24 @@ func RemovePath(path string) error {
...
 		for s, p := range paths {
-			os.RemoveAll(p)
-			// TODO: here probably should be logging
+			if err := RemovePath(p); err != nil {
+				// do not log intermediate iterations
+				switch i {
+				case 0:
+					logrus.WithError(err).Warnf("Failed to remove cgroup (will retry)")
+				case retries - 1:
+					logrus.WithError(err).Error("Failed to remove cgroup")
```
AkihiroSuda, line 267: Not sure we need these logrus. The error is returned to the caller.

kolyshkin: No it's not. The final error returned is `fmt.Errorf("Failed to remove paths: %v", paths)` (see the very end of the function below), it does not say anything about what happened.

This logging addresses the TODO above (`TODO: here probably should be logging`).

I think it makes sense to do what I did -- log the first iteration error as a warning, and last iteration error as an error.

Alternatively, I can try to rework this to return the error to the caller -- the problem is which one to return in case there are a few, during the last iteration. Just pile them up?

AkihiroSuda:
@mrunalp PTAL

libcontainer/cgroups/utils.go
mrunalp: We could use something like https://github.com/hashicorp/go-multierror to aggregate the errors.

kolyshkin: So, what this code does is
 - logs the first removal error as a warning (warning because it will retry);
 - retries a few times;
 - if removal still fails, logs the last error as an error.

I believe what the code does is the best approach, let me explain the logic that lead to this code.

1. In case the removal eventually succeeds, we might still be interested in why it failed initially -- this is why the first error is logged as a warning.
2. Most probably the removal error is the same for all tries, so it does not make sense to log it N times, or aggregate it.
3. In case it ultimately fails, since we do not return the error to the caller, it needs to be logged at least.

In any case, this is better than the current code and addresses the TODO in it.

kolyshkin: Previously, the error was ignored and we never knew if there was any retries, or what the removal error was.

Now, we log the first and the last error, so we still know what is going on, without any excessive logging.

kolyshkin: I took a look at whatever the current kernel implements, and normally it can only return EBUSY, so we'll log EBUSY once or twice. I still want to log the first error to distinguish between the "success from the first try" and "succeeded but after a retry cases". Can certainly demote the warning to debug if you like.

@mrunalp @cyphar @crosbymichael PTAL

AkihiroSuda:
@mrunalp @cyphar PTAL

PR #2507, opened 2020-07-07
libct/cgroups/GetCgroupRoot: make it faster

kolyshkin:
...by checking the default path first.

Quick benchmark shows it's about 5x faster on an idle system, and the
gain should be much more on a system doing mounts etc.

This is an alternative to #2497. I don't like it much because it adds code rather than removes it.

### DIFF
--- a/libcontainer/cgroups/fs/fs_test.go
+++ b/libcontainer/cgroups/fs/fs_test.go
@@ -295,3 +295,16 @@ func TestInvalidAbsoluteCgroupNameAndParent(t *testing.T) {
 		t.Errorf("SECURITY: cgroup path() is outside cgroup mountpoint!")
 	}
 }
+
+func TestTryDefaultCgroupRoot(t *testing.T) {
+	res := tryDefaultCgroupRoot()
+	exp := defaultCgroupRoot
+	if cgroups.IsCgroup2UnifiedMode() {
+		// checking that tryDefaultCgroupRoot does return ""
+		// in case /sys/fs/cgroup is not cgroup v1 root dir.
+		exp = ""
+	}
+	if res != exp {
+		t.Errorf("tryDefaultCgroupRoot: want %q, got %q", exp, res)
+	}
+}
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -68,6 +68,56 @@ func NewManager(cg *configs.Cgroup, paths map[string]string, rootless bool) cgro
 var cgroupRootLock sync.Mutex
 var cgroupRoot string
 
+const defaultCgroupRoot = "/sys/fs/cgroup"
+
+func tryDefaultCgroupRoot() string {
+	var st, pst unix.Stat_t
+
+	// (1) it should be a directory...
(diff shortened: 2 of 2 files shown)


kolyshkin:
CI failure (in checkpoint --lazy-pages) looks unrelated, but the cascade of errors after it failed led me to PR #2509.

kolyshkin:
CI restarted. The failure was
```
303612 checkpoint --lazy-pages and restore
3037# (from function `__runc' in file tests/integration/helpers.bash, line 57,
3038#  in test file tests/integration/checkpoint.bats, line 182)
3039#   `__runc --criu "$CRIU" restore -d --work-path ./image-dir --image-path ./image-dir --lazy-pages test_busybox_restore <&60 >&51 2>&51' failed
3040# runc list (status=0):
3041# ID          PID         STATUS      BUNDLE      CREATED     OWNER
3042# runc spec (status=0):
3043# 
3044# runc state test_busybox (status=0):
3045# {
3046#   "ociVersion": "1.0.2-dev",
3047#   "id": "test_busybox",
3048#   "pid": 3993,
3049#   "status": "running",
3050#   "bundle": "/tmp/busyboxtest",
3051#   "rootfs": "/tmp/busyboxtest/rootfs",
3052#   "created": "2020-07-07T04:09:18.881675684Z",
3053#   "owner": ""
3054# }
3055# Warn  (criu/kerndat.c:869): Can't keep kdat cache on non-tempfs
3056# runc list (status=0):
3057# ID             PID         STATUS      BUNDLE             CREATED                          OWNER
3058# test_busybox   3993        running     /tmp/busyboxtest   2020-07-07T04:09:18.881675684Z   root
3059# runc kill test_busybox KILL (status=0):
3060# 
3061# runc delete test_busybox (status=0):
3062# 
```

libcontainer/cgroups/fs/fs_test.go
```diff
@@ -295,3 +295,14 @@ func TestInvalidAbsoluteCgroupNameAndParent(t *testing.T) {
 		t.Errorf("SECURITY: cgroup path() is outside cgroup mountpoint!")
 	}
 }
+
+func TestTryDefaultCgroupRoot(t *testing.T) {
+	res := tryDefaultCgroupRoot()
+	exp := defaultCgroupRoot
+	if cgroups.IsCgroup2UnifiedMode() {
+		exp = ""
```
AkihiroSuda, line 303: why not t.skip?

kolyshkin: We are actually checking the case that `tryDefaultCgroupRoot()` correctly returns empty string when `/sys/fs/cgroup` is something other than cgroupv1 root. Let me add a comment about it.

mrunalp: Could you add a comment? Otherwise, PR looks good!

AkihiroSuda: ping @kolyshkin

kolyshkin: did that already but forgot to push. rebased, pushed

kolyshkin:
@AkihiroSuda @mrunalp PTAL

PR #2508, opened 2020-07-07
remove multi-arch.bash but use skopeo to get images directly

XiaodongLoong:
When I test runc on mips64le, I found that the integration and rootlessintegration test only support arm64 and amd64 cpu architecture.
I developed a new way to execute the integration test which can support more cpu arch.

Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>


cyphar:
#2486 is trying to clean up this issue (and should also allow us to remove multi-arch.sh). I don't really like this approach since using and requiring Docker just to get images is overkill. `umoci` and `skopeo` should be able to pull multi-arch images without issue.

XiaodongLoong:
> #2486 is trying to clean up this issue (and should also allow us to remove multi-arch.sh). I don't really like this approach since using and requiring Docker just to get images is overkill. `umoci` and `skopeo` should be able to pull multi-arch images without issue. 

Sorry, I ignored the  "overkill"  regulation and only simplify the process.

> But if you need MIPS support then we might have to hold off on switching to openSUSE since it doesn't support MIPS...

Yes, I need MIPS support indeed.

Makefile
```diff
@@ -58,12 +58,25 @@ lint:
 man:
 	man/md2man-all.sh
 
-runcimage:
+runcimage: | .busybox .hello-world .debian
 	$(CONTAINER_ENGINE) build $(CONTAINER_ENGINE_BUILD_FLAGS) -t $(RUNC_IMAGE) .
 
+.hello-world:
+	docker export $(shell docker create hello-world) -o $(CURDIR)/tests/integration/testdata/hello-world.tar
```
AkihiroSuda, line 65: docker -> $(CONTAINER_ENGINE)

XiaodongLoong: OK, I will fix it.

XiaodongLoong:
@cyphar I might have no "overkill" problem, I used docker in Makefile before starting container, it is same as executing integration test with docker.

Thanks!

kolyshkin:
Please clean this PR up (you probably did you do `git pull` without `--rebase`?), it should not contain any extra commits.

XiaodongLoong:
@kolyshkin OK, I'll do it. Thanks!

kolyshkin:
Needs a rebase

XiaodongLoong:
This PR can make the test cases run on more CPU architecture using the ability from docker.
@cyphar @AkihiroSuda @kolyshkin PTAL Thanks!

Makefile
```diff
@@ -58,12 +58,28 @@ lint:
...
+
+.busybox:
+	$(CONTAINER_ENGINE) export $(shell $(CONTAINER_ENGINE) create busybox) -o $(CURDIR)/tests/integration/testdata/busybox.tar
+	chmod a+r $(CURDIR)/tests/integration/testdata/busybox.tar
+	touch "$@"
+
+.debian:
+	$(CONTAINER_ENGINE) export $(shell $(CONTAINER_ENGINE) create debian) -o $(CURDIR)/tests/integration/testdata/debian.tar
+	chmod a+r $(CURDIR)/tests/integration/testdata/debian.tar
+	touch "$@"
```
kolyshkin, line 77: 1. You need to add something like `rm -f .debian .busybox .hello-world` to `clean` target. The downloaded images, too, I guess.

2. Also, it might be helpful to do
```
ln -sf $(CURDIR)/tests/integration/testdata/debian.tar $@
```
instead of just `touch`, as in the case a tarball will be removed, `make` will redo the target thanks to a broken symlink.

3. Using $(CURDIR) is redundant I think.

4. These three rules look similar, and it should be easy to generalize and merge those into one single rule.

5. Maybe move this code out to a separate Makefile in tests/integration/testdata, and just call it from the main Makefile.

Makefile
```diff
@@ -58,12 +58,28 @@ lint:
...
+
 test: unittest integration rootlessintegration
 
-localtest: localunittest localintegration localrootlessintegration
+localtest: | all .busybox .hello-world .debian
```
kolyshkin, line 81: This repetition can be avoided by having something like
```make
TEST_IMAGES ?= debian busybox hello-world
GET_IMAGE_TARGETS := $(addprefix ., $(TEST_IMAGES))
...
localtest: | $(GET_IMAGE_TARGETS)
```

Makefile
```diff
@@ -58,12 +58,28 @@ lint:
...
+	make localunittest localintegration localrootlessintegration
```
kolyshkin, line 82: No need to call make here.

Makefile
```diff
@@ -72,7 +88,7 @@ unittest: runcimage
 		-v $(CURDIR):/go/src/$(PROJECT) \
 		$(RUNC_IMAGE) make localunittest TESTFLAGS=$(TESTFLAGS)
 
-localunittest: all
+localunittest: | all .busybox .hello-world .debian
```
kolyshkin, line 91: Shouldn't it be ```localunittest: all | .busybox .hello-world .debian```?

Makefile
kolyshkin: Also, specifying these prerequisites here is redundant, as every one of (localunittest, localintegration, localrootlessintegration) already have those.

Vagrantfile.centos7
```diff
@@ -51,9 +51,9 @@ EOF
...
     useradd -u2000 -m -d/home/rootless -s/bin/bash rootless
 
-    # Add busybox for libcontainer/integration tests
-    . /vagrant/tests/integration/multi-arch.bash \
-        && mkdir /busybox \
-        && curl -fsSL $(get_busybox) | tar xfJC - /busybox
+    systemctl start docker
+    mkdir /busybox /debian
+    docker export $(docker create busybox) | tar -C /busybox/ -xvf -
+    docker export $(docker create debian) | tar -C /debian/ -xvf -
```
kolyshkin, line 57: Is there a way to maybe use something lighter than docker to just download/extract images from the repo? I frankly don't know

kolyshkin: there's https://raw.githubusercontent.com/moby/moby/master/contrib/download-frozen-image-v2.sh script

kolyshkin: https://github.com/containers/skopeo

kolyshkin: It's just I don't like much the idea of needing to run dockerd inside vagrant for the sole purpose of obtaining busybox (or the like) image.

What needs to be done instead is to download the images before running vagrant. Something like (in .travis.yml):
```diff
        - ln -sf Vagrantfile.fedora32 Vagrantfile
+       - make download-test-images
        - sudo vagrant up && sudo mkdir -p /root/.ssh && sudo sh -c "vagrant ssh-config >> /ro
```

(and then `vagrant up` will copy those as it copies everything from the current directory into /vagrant

kolyshkin:
I like the idea in general, the implementation can be improved though

Vagrantfile.centos7
cyphar: skopeo + umoci is what #2486 uses and (personal biases aside) I would much prefer that over using Docker just for the purposes of downloading like 3 images.

cyphar: I also want us to remove the need for three different images for our tests. Since it looks like folks want super esoteric architectures, we can just use Debian in #2486 and work around the missing packages.

XiaodongLoong: @cyphar 
I don't know umoci can  or can not download images for the CPU architecture supporting debian images. But umoci only support the ARM64 and AMD64 in the view of runc master branch. 
At first, I only want to make the test cases run on more CPU architecture.

cyphar: skopeo does the downloading and it does support multi-arch. umoci is a Go program and doesn't have any real architecture restrictions (aside from "you must be able to build a Go program on it"). Heck, you can even run umoci on MacOS.

All I'm suggesting is that this:

```
% docker export $(docker create busybox) | tar -C /busybox/ -xvf -
```

Becomes this:

```
% skopeo copy docker://busybox:latest oci:/tmp/busybox-image:latest
% umoci raw unpack --image /tmp/busybox-image:latest /busybox
```

That's it. And the latter will be more performant because `docker export $(docker create) | tar xvf -` requires extracting and generating tar archives multiple times (not to mention transferring the archive and all the rest of it)!

XiaodongLoong: OK, I got it.

XiaodongLoong:
@cyphar @kolyshkin  Thanks for your advice. I have pushed a new commit for this PR. 
There is still three test images( hello-world, busybox, debian), I suggest that we can open a new issue to run test using debian only.

XiaodongLoong:
@cyphar @kolyshkin  Thanks for your advice. I have pushed a new commit for this PR. 
There is still three test images( hello-world, busybox, debian), I suggest that we can open a new issue to run test using debian only.

Vagrantfile.fedora32
```diff
@@ -36,15 +36,10 @@ EOF
...
-    curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64
-    chmod +x /usr/local/bin/umoci
-
-    # Add busybox for libcontainer/integration tests
-    . /vagrant/tests/integration/multi-arch.bash \
-        && mkdir /busybox /debian \
-        && curl -fsSL $(get_busybox) | tar xfJC - /busybox \
-        && get_and_extract_debian /debian
+    # Add busybox and debian for libcontainer/integration tests
+    mkdir /busybox /debian
```
kolyshkin, line 40: nit:
```shell
for dist in busybox debian; do
    mkdir /${dist}
    tar --exclude './dev/*' -C /${dist} -xf /vagrant/tests/integration/testdata/${dist}.tar
done
```

tests/integration/helpers.bash
```diff
@@ -420,11 +418,9 @@ function setup_busybox() {
 	setup_recvtty
 	run mkdir "$BUSYBOX_BUNDLE"
 	run mkdir "$BUSYBOX_BUNDLE"/rootfs
-	if [ -e "/testdata/busybox.tar" ]; then
-		BUSYBOX_IMAGE="/testdata/busybox.tar"
-	fi
 	if [ ! -e $BUSYBOX_IMAGE ]; then
-		curl -o $BUSYBOX_IMAGE -sSL `get_busybox`
+		echo "busybox images does not exist"
```
kolyshkin, line 422: nit: `s/images/image/`

tests/integration/helpers.bash
```diff
@@ -435,6 +431,10 @@ function setup_hello() {
 	setup_recvtty
 	run mkdir "$HELLO_BUNDLE"
 	run mkdir "$HELLO_BUNDLE"/rootfs
+	if [ ! -e "$HELLO_IMAGE" ]; then
+		echo "hello-world images does not exist"
```
kolyshkin, line 435: ditto

tests/integration/helpers.bash
```diff
@@ -449,17 +449,15 @@ function setup_debian() {
 
 	setup_recvtty
 	run mkdir "$DEBIAN_BUNDLE"
-
-	if [ ! -d "$DEBIAN_ROOTFS/rootfs" ]; then
-		get_and_extract_debian "$DEBIAN_BUNDLE"
+	run mkdir "$DEBIAN_BUNDLE"/rootfs
+	if [ ! -e "$DEBIAN_IMAGE" ]; then
+		echo "debian images does not exist"
```
kolyshkin, line 454: ditto

tests/integration/testdata/Makefile
```diff
@@ -0,0 +1,9 @@
+TEST_IMAGES ?= debian busybox hello-world
+CONTAINER_ENGINE := docker
```
kolyshkin, line 2: As this one is already defined in the top-level Makefile, you just need to `export` it in there.

Here, maybe, set it only if unset, e.g. `?=` -- in this case, if `make` will be run directly, this can still be set, otherwise, a value from the top-level Makefile will be used.

Need to be checked by changing the value in the top-level makefile and running `make .download-test-images`

tests/integration/testdata/Makefile
```diff
@@ -0,0 +1,9 @@
...
+
+download_images:
```
kolyshkin, line 4: this is a phony target, needs to be marked as such

tests/integration/testdata/Makefile
```diff
@@ -0,0 +1,9 @@
...
+	$(foreach image, $(TEST_IMAGES),$(shell $(CONTAINER_ENGINE) export $(shell $(CONTAINER_ENGINE) create $(image)) -o $(image).tar))
+	chmod a+r $(foreach image, $(TEST_IMAGES), $(image).tar)
+
+clean:
```
kolyshkin, line 8: ditto

XiaodongLoong:
PR update, PTAL

tests/integration/helpers.bash
```diff
@@ -449,17 +449,15 @@ function setup_debian() {
...
-
-	# Use the cached version
-	if [ ! -d "$DEBIAN_BUNDLE/rootfs" ]; then
-		cp -r "$DEBIAN_ROOTFS"/* "$DEBIAN_BUNDLE/"
-	fi
-
+	tar --exclude './dev/*' -C "$DEBIAN_BUNDLE"/rootfs -xf "$DEBIAN_IMAGE"
 	cd "$DEBIAN_BUNDLE"
+	runc_spec
+	update_config '.root.readonly |= false'
```
kolyshkin, line 460: Not sure where this is coming from, can you explain?

XiaodongLoong: I fix a test error:
```
not ok 44 runc run (hooks library tests)
2994# (in test file tests/integration/hooks.bats, line 54)
2995# `[ "$status" -eq 0 ]' failed
2996# runc list (status=0):
2997# ID PID STATUS BUNDLE CREATED OWNER
2998# runc spec (status=0):
2999# 
3000# /tmp/bats.8004.src: line 47: warning: here-document at line 43 delimited by end-of-file (wanted `EOF')
3001# runc run test_debian (status=1):
3002# Running hook #0:: error running hook: exit status 1, stdout: , stderr: ldconfig: Can't link /lib/librunc-hooks-create-runtime.so.1 to librunc-hooks-create-runtime.so.1.0.0
3003# ldconfig: Can't link /lib/librunc-hooks-create-container.so.1 to librunc-hooks-create-container.so.1.0.0
3004# ldconfig: Can't create temporary cache file /etc/ld.so.cache~: Read-only file system
3005# runc list (status=0):
3006# ID PID STATUS BUNDLE CREATED OWNER
3007ok 45 kill detached busybox
```

kolyshkin: This should be done in the very test that fails. You have done it globally, not good.

kolyshkin:
Hey @XiaodongLoong, can you try implementing the following suggestion for this PR?
https://github.com/opencontainers/runc/pull/2508#discussion_r455489404

tests/integration/helpers.bash
XiaodongLoong: OK, I will fix it.

XiaodongLoong:
> Hey @XiaodongLoong, can you try implementing the following suggestion for this PR?
> [#2508 (comment)](https://github.com/opencontainers/runc/pull/2508#discussion_r455489404)

 Should I change the docker command inside runc/tests/integration/testdata/Makefile into skopeo+umoci ?
But it is just the same as using docker to run test cases.

kolyshkin:
> Should I change the docker command inside runc/tests/integration/testdata/Makefile into skopeo+umoci ?

Yes, that's what I'm thinking about. It should be faster and does not require dockerd to run.

Makefile
```diff
@@ -147,7 +153,7 @@ localcross:
 	CGO_ENABLED=1 GOARCH=arm64 CC=aarch64-linux-gnu-gcc         $(GO_BUILD) -o runc-arm64 .
 	CGO_ENABLED=1 GOARCH=ppc64le CC=powerpc64le-linux-gnu-gcc   $(GO_BUILD) -o runc-ppc64le .
 
-.PHONY: runc all recvtty static release dbuild lint man runcimage \
+.PHONY: runc all recvtty static release dbuild lint man runcimage .download-test-images \
```
kolyshkin, line 156: Now since we do have a file named `.download-test-images`, this target is no longer phony I guess.

tests/integration/hooks.bats
```diff
@@ -14,6 +14,7 @@ function setup() {
 
 	teardown_debian
 	setup_debian
+	update_config '.root.readonly |= false'
```
kolyshkin, line 17: as I noted earlier, it should be done
 - in a very test case that requires it
 - as a separate patch describing why it is done (since the linker in hooks test likes to create files)

XiaodongLoong: If this PR don't have this patch, CI verification failed. Can  I add some notes ?

XiaodongLoong:
@kolyshkin
I failed to install skope.
```
$ echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/  /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
277/home/travis/.travis/functions: line 109: /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list: Permission denied
278The command "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/  /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list" failed and exited with 1 during .
```

kolyshkin:
> I failed to install skopeo with Permission denied error.

Please see the .travis.yml file a few lines before your changes -- it calls apt, too, so you don't have to call it again.

Also, as you can see, it uses `sudo` and you should too.

`echo ... | sudo cat > /etc/apt/....`

tests/integration/hooks.bats
XiaodongLoong: I skiped the hook test case, but add some TODO notes for an new PR to fix this issue.

XiaodongLoong:
updated , @kolyshkin PTAL

kolyshkin:
@cyphar do you think we should proceed with this one? Overall I like it.

PR #2509, opened 2020-07-07
tests/int/checkpoint: fds and pids cleanup

kolyshkin:
1. Do not use hardcoded fd numbers, instead relying on bash feature of
   assigning an fd to a variable.

   This looks very weird, but the rule of thumb here is:
   - if this is `exec`, use `{var}` (i.e. no `$`);
   - otherwise, use as normal (`$var` or `${var}`).

2. Add killing the background processes and closing the fds to teardown.
   This is helpful in case of a test failure, in order to not affect the
   subsequent tests. Found in https://github.com/opencontainers/runc/pull/2507#issuecomment-655034055

3. Drop removing readonly flag for lazy cpt tests, it is no longer required
   (see https://github.com/checkpoint-restore/criu/pull/1033, https://github.com/checkpoint-restore/criu/issues/575)

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -144,9 +144,6 @@ function simple_cr() {
 
   setup_pipes
 
-  # This should not be necessary: https://github.com/checkpoint-restore/criu/issues/575
-  update_config '(.. | select(.readonly? != null)) .readonly |= false'
-
   # TCP port for lazy migration
   port=27277
 
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -12,6 +12,17 @@ function setup() {
 
 function teardown() {
   teardown_busybox
+  local pid fd
+
+  for pid in "${PIDS_TO_KILL[@]}"; do
+    kill -9 $pid || true
+  done
+  PIDS_TO_KILL=()
+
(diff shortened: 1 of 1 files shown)


kolyshkin:
@tianon @adrianreber @avagin PTAL

kolyshkin:
The `{fd}` feature requires bash-4.1. Documentation: https://www.gnu.org/software/bash/manual/html_node/Redirections.html

kolyshkin:
I've also found a commit dropping the readonly flag removal for cpt tests. Indeed it looks like  https://github.com/checkpoint-restore/criu/issues/575 is somehow fixed.

kolyshkin:
@AkihiroSuda @mrunalp PTAL (this should make our CI cleaner a bit)

kolyshkin:
> Indeed it looks like checkpoint-restore/criu#575 is somehow fixed.

Clarification on "somehow". The removal of "read-only" was preventing the following like from appearing in the logs:

> (00.069337) Error (criu/mount.c:1119): mnt: Can't create a temporary directory: Read-only file system

The thing is, this was not an error but actually warning, since the code was working around the read-only state, so clearing the read-only state was not needed. Later, https://github.com/checkpoint-restore/criu/pull/1033 demoted that error to a warning, so since criu 3.14 it is no longer appearing as an error. But even without this patch read-only could safely be removed.

PR #2520, opened 2020-07-17
vendor: update runtime-spec v1.0.3-0.20200728170252-4d89ac9fbff6

thaJeztah:


go.sum
```diff
@@ -27,6 +27,8 @@ github.com/mrunalp/fileutils v0.0.0-20171103030105-7d4729fb3618 h1:7InQ7/zrOh6Sl
 github.com/mrunalp/fileutils v0.0.0-20171103030105-7d4729fb3618/go.mod h1:x8F1gnqOkIEiO4rqoeEEEqQbo7HjGMTvyoq3gej4iT0=
 github.com/opencontainers/runtime-spec v1.0.3-0.20200520003142-237cc4f519e2 h1:9mv9SC7GWmRWE0J/+oD8w3GsN2KYGKtg6uwLN7hfP5E=
 github.com/opencontainers/runtime-spec v1.0.3-0.20200520003142-237cc4f519e2/go.mod h1:jwyrGlmzljRJv/Fgzds9SsS/C5hL+LL3ko9hs6T5lQ0=
+github.com/opencontainers/runtime-spec v1.0.3-0.20200710190001-3e4195d92445 h1:y8cfsJRmn8g3VkM4IDpusKSgMUZEXhudm/BuYANLozE=
```
kolyshkin, line 30: I think you need to run `go mod tidy` since the old version checksum is still present

libcontainer/process_linux.go
```diff
@@ -400,7 +400,7 @@ func (p *initProcess) start() (retErr error) {
 					}
 					// initProcessStartTime hasn't been set yet.
 					s.Pid = p.cmd.Process.Pid
-					s.Status = configs.Creating
```
kolyshkin, line 403: Maybe change `Status` type from `string` to `specs.ContainerState` instead of typecasting?

go.sum
thaJeztah: Arf, yes; updating

libcontainer/process_linux.go
thaJeztah: I was looking at that; I _think_ that would be a breaking change though, WDYT?

kolyshkin: I don't know :( but these typecasts everywhere just look wrong

@tianon @cyphar PTAL ^^^

cyphar: It won't be a breaking change -- the state information is serialised to JSON and the JSON form will remain the same (`json.Marshal` doesn't care about type aliases like `type ContainerState string`).

thaJeztah: Yeah, (Go) is a bit hairy in that respect; https://play.golang.org/p/LGJ3vs3ncu0

```go
package main

type Foo string

type Foobar struct {
	Bla Foo
}

const somethingFoo Foo = "something"

func main() {
	// This works
	_ = Foobar{Bla: "something"}

	// This works
	_ = Foobar{Bla: somethingFoo}

	// This won't work
	something := "something"
	_ = Foobar{Bla: something}
}
```

dims: whoa! mind blown! @thaJeztah

cyphar: Yeah, I've hit this in the past as well (I think `Foobar{Bla: "something"}` working is actually a mistake on the Go language side of things, because it leads to more edge-cases than would be ideal).

However I think it's fine to make a change like that -- the Go files aren't really what is being versioned by `SemVer`, it's the specification. While we would ideally not break upgrades, I think doing a major version bump on `runtime-spec` because of a type alias would be a little bit of an over-reaction. :wink:

kolyshkin:
I'd rather have https://github.com/opencontainers/runtime-spec/pull/1056 merged first and then redo this PR on top of it

libcontainer/container_linux.go
```diff
@@ -1863,7 +1863,7 @@ func (c *linuxContainer) currentOCIState() (*specs.State, error) {
 	if err != nil {
 		return nil, err
 	}
-	state.Status = status.String()
+	state.Status = specs.ContainerState(status.String())
```
thaJeztah, line 1866: Notice that `status.String()` can return states that are not in the OCI runtime spec (`"pausing"`, `"paused"`, `"unknown"`), and vice versa: the runtime spec defines `"creating"` which does not map to a libcontainer state ; https://github.com/opencontainers/runc/blob/ceefc3fe4ecb31fa89ed50ecb2bc0efc7dfac683/libcontainer/container.go#L31-L46


How should we deal with those?

thaJeztah:
Updated to get the latest runtime-spec, but left a comment above https://github.com/opencontainers/runc/pull/2520#discussion_r461804920

libcontainer/container_linux.go
kolyshkin: I think it's fine for runc to have a superset of statuses as defined by OCI spec. It's also fine to never return some statuses, like "creating".

kolyshkin:
Whoops, this needs a rebase @thaJeztah

thaJeztah:
ah! no worries; on it!

thaJeztah:
rebased

AkihiroSuda:
@cyphar LGTY?

PR #2522, opened 2020-07-21
Remove runc default devices that overlap with spec devices.

ctalledo:
Runc has a set of default devices that it includes in Linux containers
(e.g., /dev/null, /dev/random, /dev/tty, etc.)

However if the container's OCI spec includes all or a subset of those same devices,
runc is currently not detecting the redundancy, causing it to create a lib
container config that has redundant device configurations.

This causes a failure in rootless mode, in particular when the /dev/tty device
has a redundant config:

container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:70: creating device nodes caused: open /tmp/busyboxtest/rootfs/dev/tty: no such device or address"

The reason this fails in rootless mode only is that in this case runc sets up
/dev/tty not by doing mknod (it's not allowed within a user-ns) but rather by
creating a regular file under /dev/tty and bind-mounting the host's /dev/tty to
the container's /dev/tty. When this operation is done redundantly, it fails the
second time.

This change fixes this problem by ensuring runc checks for redundant devices
between the OCI spec it receives and the default devices it configures. If
a redundant device is detected, the OCI spec takes priority.

The change adds both a unit test and an integration test to verify the
behavior. Without this fix, this new integration test fails as shown above.

Signed-off-by: Cesar Talledo <ctalledo@nestybox.com>

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -19,6 +19,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/system"
+	"github.com/opencontainers/runc/libcontainer/utils"
 	libcontainerUtils "github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/selinux/go-selinux/label"
 
@@ -589,6 +590,12 @@ func createDevices(config *configs.Config) error {
 	useBindMount := system.RunningInUserNS() || config.Namespaces.Contains(configs.NEWUSER)
 	oldMask := unix.Umask(0000)
 	for _, node := range config.Devices {
+
+		// The /dev/ptmx device is setup by setupPtmx()
+		if utils.CleanPath(node.Path) == "/dev/ptmx" {
+			continue
+		}
+
 		// containers running in a user namespace are not allowed to mknod
 		// devices so we can just bind mount it from the host.
 		if err := createDeviceNode(config.Rootfs, node, useBindMount); err != nil {
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -241,13 +241,17 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	for _, m := range spec.Mounts {
 		config.Mounts = append(config.Mounts, createLibcontainerMount(cwd, m))
 	}
-	if err := createDevices(spec, config); err != nil {
-		return nil, err
-	}
-	c, err := CreateCgroupConfig(opts)
+
+	defaultDevs, err := createDevices(spec, config)
 	if err != nil {
 		return nil, err
 	}
+
+	c, err := CreateCgroupConfig(opts, defaultDevs)
+	if err != nil {
+		return nil, err
+	}
+
 	config.Cgroups = c
 	// set linux-specific config
 	if spec.Linux != nil {
@@ -410,7 +414,7 @@ func initSystemdProps(spec *specs.Spec) ([]systemdDbus.Property, error) {
 	return sp, nil
 }
 
-func CreateCgroupConfig(opts *CreateOpts) (*configs.Cgroup, error) {
+func CreateCgroupConfig(opts *CreateOpts, defaultDevs []*configs.Device) (*configs.Cgroup, error) {
 	var (
 		myCgroupPath string
 
@@ -616,9 +620,9 @@ func CreateCgroupConfig(opts *CreateOpts) (*configs.Cgroup, error) {
 			}
 		}
 	}
+
 	// Append the default allowed devices to the end of the list.
-	// XXX: Really this should be prefixed...
-	for _, device := range AllowedDevices {
+	for _, device := range defaultDevs {
 		c.Resources.Devices = append(c.Resources.Devices, &device.DeviceRule)
 	}
 	return c, nil
@@ -650,13 +654,26 @@ func stringToDeviceRune(s string) (configs.DeviceType, error) {
 	}
 }
 
-func createDevices(spec *specs.Spec, config *configs.Config) error {
-	// Add default set of devices.
-	for _, device := range AllowedDevices {
-		if device.Path != "" {
-			config.Devices = append(config.Devices, device)
+func createDevices(spec *specs.Spec, config *configs.Config) ([]*configs.Device, error) {
+	// If a spec device is redundant with a default device, remove that default
+	// device (the spec one takes priority).
+	dedupedAllowDevs := []*configs.Device{}
+
+next:
+	for _, ad := range AllowedDevices {
+		if ad.Path != "" {
+			for _, sd := range spec.Linux.Devices {
+				if sd.Path == ad.Path {
+					continue next
+				}
+			}
+		}
+		dedupedAllowDevs = append(dedupedAllowDevs, ad)
+		if ad.Path != "" {
+			config.Devices = append(config.Devices, ad)
 		}
 	}
+
 	// Merge in additional devices from the spec.
 	if spec.Linux != nil {
 		for _, d := range spec.Linux.Devices {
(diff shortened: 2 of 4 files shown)


tests/integration/dev.bats
```diff
@@ -0,0 +1,26 @@
...
+function setup() {
+	teardown_busybox
+	setup_busybox
+}
+
+function teardown() {
+	teardown_busybox
+}
+
+@test "runc run [redundant default dev]" {
```
AkihiroSuda, line 14: Could you add a test for ptmx as well? (It had a slightly code path than tty)

e.g. https://github.com/opencontainers/runc/pull/2451/files#diff-b890de221ad09cb84cc3c69482f2da36R235

ctalledo: Sure, I'll add the test for ptmx too.

ctalledo:
Hi @AkihiroSuda, I added a test for `/dev/ptmx` per your request. In fact that test caught a bug which my initial change did not address, so I had to add a second commit to the PR.

Please take a look when you get a chance.

Thanks!

libcontainer/rootfs_linux.go
```diff
@@ -589,6 +589,12 @@ func createDevices(config *configs.Config) error {
 	useBindMount := system.RunningInUserNS() || config.Namespaces.Contains(configs.NEWUSER)
 	oldMask := unix.Umask(0000)
 	for _, node := range config.Devices {
+
+		// The /dev/ptmx device is setup by setupPtmx()
+		if node.Path == "/dev/ptmx" {
```
AkihiroSuda, line 594: filepath.Clean(node.Path)

cyphar: Or even better `utils.CleanPath`.

ctalledo: Good catch, thanks. Fixed.

libcontainer/specconv/spec_linux.go
```diff
@@ -650,13 +654,29 @@ func stringToDeviceRune(s string) (configs.DeviceType, error) {
...
-	for _, device := range AllowedDevices {
-		if device.Path != "" {
-			config.Devices = append(config.Devices, device)
+func createDevices(spec *specs.Spec, config *configs.Config) ([]*configs.Device, error) {
+
+	// If a spec device is redundant with a default device, remove that default
+	// device (the spec one takes priority).
+	dedupedAllowDevs := []*configs.Device{}
+
+	if spec.Linux != nil {
```
kolyshkin, line 663: Hmm, I think you've changed the logic of this function in case `spec.Linux == nil`. Previously it did something (added entries to config.Devices), now it does not.

libcontainer/specconv/spec_linux.go
```diff
@@ -650,13 +654,29 @@ func stringToDeviceRune(s string) (configs.DeviceType, error) {
...
+		for _, ad := range AllowedDevices {
+			conflict := false
+			for _, sd := range spec.Linux.Devices {
+				if sd.Path == ad.Path {
+					conflict = true
```
kolyshkin, line 668: Maybe add `break` here, since it does not make sense to keep iterating over `spec.Linux.Devices` once you found a conflict?

libcontainer/specconv/spec_linux.go
```diff
@@ -650,13 +654,29 @@ func stringToDeviceRune(s string) (configs.DeviceType, error) {
...
+				}
+			}
+			if !conflict {
```
kolyshkin, line 671: I would use
```go
if conflict {
  break
}
dedupedAllowDevs = append(dedupedAllowDevs, ad)
....
```

here since it's more readable

kolyshkin: Of even
```go
next:
for _, ad := range AllowedDevices {
  for _, sd := range spec.Linux.Devices {
    if sd.Path == ad.Path {
      continue next
    }
  }
  dedupedAllowDevs = append(dedupedAllowDevs, ad)
  if ad.Path != "" {
    config.Devices = append(config.Devices, ad)
  }
}
```

kolyshkin: Maybe
```go
next:
for _, ad := range AllowedDevices {
  if ad.Path != "" {
    for _, sd := range spec.Linux.Devices {
      if sd.Path == ad.Path {
        continue next
      }
    }
  }
  dedupedAllowDevs = append(dedupedAllowDevs, ad)
  if ad.Path != "" {
    config.Devices = append(config.Devices, ad)
  }
}
```

as otherwise, if there's an entry in spec.Linux.Devices with Path = "", we'll skip all the AllowDevices with empty path as well, and this seems wrong. It seems your code has the same problem.

libcontainer/specconv/spec_linux.go
ctalledo: Hi @kolyshkin, thanks for the review. Good catch here, I've re-worked the code.

libcontainer/specconv/spec_linux.go
ctalledo: Good idea, thanks.

libcontainer/specconv/spec_linux.go
ctalledo: That looks nice ... thanks. Let me re-work this section of the code once more per your suggestion.

ctalledo: Hi @kolyshkin, I followed your suggestion and re-worked the loop. Thanks!

ctalledo:
Hi @cyphar, @kolyshkin, let me know if this PR looks good to you please. Thanks!

ctalledo:
Kind ping to @cyphar and @kolyshkin for approval. Thanks!

AkihiroSuda:
ping @cyphar @kolyshkin @mrunalp Let's have rc92 with this

libcontainer/specconv/spec_linux_test.go
```diff
@@ -648,3 +648,104 @@ func TestNullProcess(t *testing.T) {
...
+				t.Errorf("redundant dev was not deduplicated correctly: want %v, got %v", wantDev, configDev)
+			}
+		}
+	}
+
+	// Verify that createDevices() added the entry for /dev/ram0 in the config
+	found = false
+	for _, configDev := range conf.Devices {
+		if configDev.Path == "/dev/ram0" {
+			found = true
```
kolyshkin, line 745: nit: `break`

libcontainer/specconv/spec_linux.go
```diff
@@ -650,13 +654,27 @@ func stringToDeviceRune(s string) (configs.DeviceType, error) {
 	}
 }
 
-func createDevices(spec *specs.Spec, config *configs.Config) error {
-	// Add default set of devices.
-	for _, device := range AllowedDevices {
-		if device.Path != "" {
-			config.Devices = append(config.Devices, device)
+func createDevices(spec *specs.Spec, config *configs.Config) ([]*configs.Device, error) {
+
```
kolyshkin, line 658: nit: no need to have this empty line here.

libcontainer/specconv/spec_linux_test.go
```diff
@@ -648,3 +648,104 @@ func TestNullProcess(t *testing.T) {
...
+				t.Errorf("createDevices failed: returned a duplicated device entry: %v", defaultDevs)
+			}
+			found = true
+		}
+	}
+
+	// Verify that createDevices() placed all default devices are in the config
+	for _, allowedDev := range AllowedDevices {
+		found := false
+		if allowedDev.Path != "" {
```
kolyshkin, line 704: I'd rather have it as
```go
if allowedDev.Path == "" {
   continue
}

found := false
...
```

ctalledo: Good idea, I've fixed it.

libcontainer/specconv/spec_linux_test.go
ctalledo: Thanks!

libcontainer/specconv/spec_linux.go
ctalledo: Fixed, thanks!

thaJeztah:
ping @kolyshkin PTAL

ctalledo:
Hi @kolyshkin, would it be possible to get another review? Thanks!

cyphar (approved):
LGTM. I think the dedup devices code could still be made simpler, but this does solve the issue.

ctalledo:
Thanks @cyphar, @AkihiroSuda, @kolyshkin !

PR #2525, opened 2020-07-21
Tell CRIU to use an external pid namespace if necessary

adrianreber:
Trying to checkpoint a container out of a pod in cri-o fails with:

> `Error (criu/namespaces.c:1081): Can't dump a pid namespace without the process init`

Starting with the upcoming CRIU release 3.15, CRIU can be told to ignore the PID namespace during checkpointing and to restore processes into an existing network namespace.
    
With the changes from this PR and CRIU 3.15 it is possible to checkpoint a container out of a pod in cri-o.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -837,13 +837,19 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts) 
 }
 
 func (c *linuxContainer) criuSupportsExtNS(t configs.NamespaceType) bool {
+	var minVersion int
 	switch t {
 	case configs.NEWNET:
 		// CRIU supports different external namespace with different released CRIU versions.
 		// For network namespaces to work we need at least criu 3.11.0 => 31100.
-		return c.checkCriuVersion(31100) == nil
+		minVersion = 31100
+	case configs.NEWPID:
+		// For PID namespaces criu 31500 is needed.
+		minVersion = 31500
+	default:
+		return false
 	}
-	return false
+	return c.checkCriuVersion(minVersion) == nil
 }
 
 func (c *linuxContainer) criuNsToKey(t configs.NamespaceType) string {
@@ -976,6 +982,11 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 	// This basically means that CRIU will ignore the namespace
 	// and expect to be setup correctly.
 	if err := c.handleCheckpointingExternalNamespaces(&rpcOpts, configs.NEWNET); err != nil {
+		return err
+	}
+
+	// Same for possible external PID namespaces
+	if err := c.handleCheckpointingExternalNamespaces(&rpcOpts, configs.NEWPID); err != nil {
 		return err
 	}
 
@@ -1301,6 +1312,11 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 	// assigned to it, this now expects that the checkpoint will be restored in a
 	// already created network namespace.
 	if err := c.handleRestoringExternalNamespaces(req.Opts, &extraFiles, configs.NEWNET); err != nil {
+		return err
+	}
+
+	// Same for PID namespaces.
+	if err := c.handleRestoringExternalNamespaces(req.Opts, &extraFiles, configs.NEWPID); err != nil {
 		return err
 	}
 
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -836,6 +836,72 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts) 
 	}
 }
 
+func (c *linuxContainer) criuSupportsExtNS(t configs.NamespaceType) bool {
+	switch t {
+	case configs.NEWNET:
+		// CRIU supports different external namespace with different released CRIU versions.
+		// For network namespaces to work we need at least criu 3.11.0 => 31100.
+		return c.checkCriuVersion(31100) == nil
+	}
+	return false
+}
+
+func (c *linuxContainer) criuNsToKey(t configs.NamespaceType) string {
+	return "extRoot" + strings.Title(configs.NsName(t)) + "NS"
+}
+
+func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.CriuOpts, t configs.NamespaceType) error {
+	if !c.criuSupportsExtNS(t) {
+		return nil
+	}
+
+	nsPath := c.config.Namespaces.PathOf(t)
+	if nsPath == "" {
+		return nil
+	}
+	// CRIU expects the information about an external namespace
+	// like this: --external <TYPE>[<inode>]:<key>
+	// This <key> is always 'extRoot<TYPE>NS'.
+	var ns unix.Stat_t
+	if err := unix.Stat(nsPath, &ns); err != nil {
+		return err
+	}
+	criuExternal := fmt.Sprintf("%s[%d]:%s", configs.NsName(t), ns.Ino, c.criuNsToKey(t))
+	rpcOpts.External = append(rpcOpts.External, criuExternal)
+
+	return nil
+}
+
+func (c *linuxContainer) handleRestoringExternalNamespaces(rpcOpts *criurpc.CriuOpts, extraFiles *[]*os.File, t configs.NamespaceType) error {
+	if !c.criuSupportsExtNS(t) {
+		return nil
+	}
+
+	nsPath := c.config.Namespaces.PathOf(t)
+	if nsPath == "" {
+		return nil
+	}
+	// CRIU wants the information about an existing namespace
+	// like this: --inherit-fd fd[<fd>]:<key>
+	// The <key> needs to be the same as during checkpointing.
+	// We are always using 'extRoot<TYPE>NS' as the key in this.
+	nsFd, err := os.Open(nsPath)
+	if err != nil {
+		logrus.Errorf("If a specific network namespace is defined it must exist: %s", err)
+		return fmt.Errorf("Requested network namespace %v does not exist", nsPath)
+	}
+	inheritFd := new(criurpc.InheritFd)
+	inheritFd.Key = proto.String(c.criuNsToKey(t))
+	// The offset of four is necessary because 0, 1, 2 and 3 is already
+	// used by stdin, stdout, stderr, 'criu swrk' socket.
+	inheritFd.Fd = proto.Int32(int32(4 + len(*extraFiles)))
+	rpcOpts.InheritFd = append(rpcOpts.InheritFd, inheritFd)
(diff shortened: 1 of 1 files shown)


libcontainer/container_linux.go
```diff
@@ -1345,7 +1379,13 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 			req.Opts.InheritFd = append(req.Opts.InheritFd, inheritFd)
 		}
 	}
-	return c.criuSwrk(process, req, criuOpts, extraFiles)
+	err = c.criuSwrk(process, req, criuOpts, extraFiles)
+
+	for _, fd := range extraFiles {
```
kolyshkin, line 1398: It's not clear whether this is coming from? Seems like a separate bugfix -- if so, can you make a separate commit so this fix will have an explanation in its commit message

libcontainer/container_linux.go
```diff
@@ -836,6 +836,82 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	// This <key> is always 'extRoot<TYPE>NS'.
+	var ns unix.Stat_t
+	if err := unix.Stat(nsPath, &ns); err != nil {
+		return err
+	}
+	var ns_type string
+	var ns_key string
+	if t == configs.NEWNET {
+		ns_type = "net"
+		ns_key = "extRootNetNS"
```
kolyshkin: Maybe move this initialization to the above "switch" I proposed.

libcontainer/container_linux.go
```diff
@@ -836,6 +836,82 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+		return nil
+	}
+	// CRIU expects the information about an external namespace
+	// like this: --external <TYPE>[<inode>]:<key>
+	// This <key> is always 'extRoot<TYPE>NS'.
+	var ns unix.Stat_t
+	if err := unix.Stat(nsPath, &ns); err != nil {
+		return err
+	}
+	var ns_type string
```
kolyshkin, line 864: nit:
```go
var nsType, nsKey string
```

libcontainer/container_linux.go
adrianreber: It's not really a separate bugfix. Before this change the FDs in extraFiles were opened in this function and immediately `defer Close`d. Now that the FDs are opened in another function `defer Close()` no longer works (to figure this out took me much longer than it should have :wink: ). So all the open FDs need to be closed after CRIU is done with it, that is why I am explicitly closing them all. Maybe a comment would make this clearer.

kolyshkin: Right.

*/me goes off to get another espresso*

libcontainer/container_linux.go
```diff
@@ -836,6 +836,82 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	// not error out if it fails. runc will just behave as it used
+	// to do and ignore external namespaces.
+	if err := c.checkCriuVersion(minVersion); err != nil {
+		return nil
+	}
+	// CRIU wants the information about an existing namespace
+	// like this: --inherit-fd fd[<fd>]:<key>
+	// The <key> needs to be the same as during checkpointing.
+	// We are always using 'extRoot<TYPE>NS' as the key in this.
+	ns_fd, err := os.Open(nsPath)
```
kolyshkin, line 898: nit: `s/ns_fd/nsFd/` as per Go var naming convensions

adrianreber:
@kolyshkin thanks for the review. I tried to address all comments and pushed a new version.

libcontainer/container_linux.go
```diff
@@ -836,6 +836,80 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
 	}
 }
 
+func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.CriuOpts, minVersion int, t configs.NamespaceType) error {
```
kolyshkin, line 839: (I swear I left a comment about `minVersion` but I can't see it. Hmm, did I do it in my sleep?)

I think the `minVersion` should not be specified by the caller but internalized in this method, as it is clearly a function of NamespaceType.

IOW

```go
func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.CriuOpts,t configs.NamespaceType) error {
var minVersion int
if t == configs.NEWNET {
￼		nsType = "Net"
￼		minVersion = 31100
}
```
￼	} else {

libcontainer/container_linux.go
```diff
@@ -836,6 +836,80 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	var nsType string
+	if t == configs.NEWNET {
```
kolyshkin, line 841: to me, switch / case / case / default ... looks a tad better than if / else if / else if / else, but up to you.

adrianreber:
Both suggestion make sense. Especially as follow up patches will be simpler. minVersion is now part of the functions and replaced the if with a switch.

libcontainer/container_linux.go
```diff
@@ -836,6 +836,84 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
 	}
 }
 
+func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.CriuOpts, t configs.NamespaceType) error {
+	var nsType string
+	var minVersion int
+	switch configs.NEWNET {
```
kolyshkin, line 842: whoops, you probably mean to write `switch t` here.

libcontainer/container_linux.go
```diff
@@ -836,6 +836,84 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	criuExternal := fmt.Sprintf("%s[%d]:extRoot%sNS", strings.ToLower(nsType), ns.Ino, nsType)
+	rpcOpts.External = append(rpcOpts.External, criuExternal)
+
+	return nil
+}
+
+func (c *linuxContainer) handleRestoringExternalNamespaces(rpcOpts *criurpc.CriuOpts, extraFiles *[]*os.File, t configs.NamespaceType) error {
+	var key string
+	var minVersion int
+	switch configs.NEWNET {
```
kolyshkin, line 881: ditto

libcontainer/container_linux.go
```diff
@@ -836,6 +836,84 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	case configs.NEWNET:
+		nsType = "Net"
+		// CRIU supports different external namespace with different released CRIU versions.
+		// For network namespaces to work we need at least criu 3.11.0 => 31100.
+		minVersion = 31100
+	default:
+		return nil
+	}
```
kolyshkin, line 854: Now it's very clear we're doing the same thing twice, so maybe have something like

```go
// Checks if external namespace t is supported by the current criu.
// If yes, returns the nsType key, otherwise, an empty string is returned.
func criuExtNsSupported(t configs.NamespaceType) string {
  var (
    key string
    // CRIU supports different external namespace with different released CRIU versions.
    ver int
  )
  switch t {
    case configs.NEWNET:
      key = "Net"
      ver = 31100
    default:
      return ""
  }
  if err := c.checkCriuVersion(minVersion); err != nil {
    return ""
  }

  return key
}
```

and use it from both `handleCheckpointingExternalNamespaces` and `handleRestoringExternalNamespaces`?

Can be a `map[configs.NamespaceType]struct{minVersion int, key string}` instead of switch, or whatever works.

In fact we already have configs.NsName(), the only issue is capitalization, so maybe use it and `strings.Title()`.

IE

```go
func criuSupportsExtNS(t configs.NamespaceType) bool {
  switch t {
    case configs.NEWNET:
      return c.checkCriuVersion(31100) == nil
    ...
  }
  return false
}

func criuNsToKey(t configs.NamespaceType) {
  return "ExtRoot" + strings.Title(configs.NsName(t) + "NS"
}

...

func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.CriuOpts, t configs.NamespaceType) error {
  if !criuSupportsExtNs(t) {
    return nil
  }
  ...
  criuExternal := fmt.Sprintf("%s[%d]:%s", configs.NsName(t), ns.Ino, criuNsKey(t))
  ...
}

...
  inheritFd.Key = proto.String(criuNsKey(t))
```

libcontainer/container_linux.go
adrianreber: fixed

libcontainer/container_linux.go
adrianreber: fixed

libcontainer/container_linux.go
adrianreber: reworked like this

libcontainer/container_linux.go
```diff
@@ -836,6 +836,72 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	switch t {
+	case configs.NEWNET:
+		// CRIU supports different external namespace with different released CRIU versions.
+		// For network namespaces to work we need at least criu 3.11.0 => 31100.
+		return c.checkCriuVersion(31100) == nil
+	}
+	return false
+}
+
+func (c *linuxContainer) criuNsToKey(t configs.NamespaceType) string {
```
kolyshkin, line 855: looks like this does not have to be a method of `*linuxContainer`.

libcontainer/container_linux.go
```diff
@@ -836,6 +836,72 @@ func (c *linuxContainer) handleCriuConfigurationFile(rpcOpts *criurpc.CriuOpts)
...
+	nsFd, err := os.Open(nsPath)
+	if err != nil {
+		logrus.Errorf("If a specific network namespace is defined it must exist: %s", err)
+		return fmt.Errorf("Requested network namespace %v does not exist", nsPath)
+	}
+	inheritFd := new(criurpc.InheritFd)
+	inheritFd.Key = proto.String(c.criuNsToKey(t))
+	// The offset of four is necessary because 0, 1, 2 and 3 is already
+	// used by stdin, stdout, stderr, 'criu swrk' socket.
+	inheritFd.Fd = proto.Int32(int32(4 + len(*extraFiles)))
```
kolyshkin, line 903: Can be
```go
inheritFd := &criurpc.InheritFd{
  Key:  proto.String(c.criuNsToKey(t)),
  // The offset of four is necessary because 0, 1, 2 and 3 is already
  // used by stdin, stdout, stderr, 'criu swrk' socket.
  Fd: proto.Int32(int32(4 + len(*extraFiles))),
}
```

while you're at it (but not important).

kolyshkin:
Looks awesome, left a couple of very minor nits but fine to merge as is

PR #2528, opened 2020-07-25
cleancode: adjust and make it more readability

zvier:
Tiny adjust `LinuxFactory.StartInitialization()` method, I think this may be more readability and clean. 

First, get environment variables are together in `var` statement at the head of the method.  The init order are also consistent with the variable declaration.

And second, `it, pipe, consoleSocket, fifofd` are the args for `i, err := newContainerInit(it, pipe, consoleSocket, fifofd)` 

Signed-off-by: Jeff Zvier <zvier20@gmail.com>

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -335,35 +335,28 @@ func (l *LinuxFactory) Type() string {
 // StartInitialization loads a container by opening the pipe fd from the parent to read the configuration and state
 // This is a low level implementation detail of the reexec and should not be consumed externally
 func (l *LinuxFactory) StartInitialization() (err error) {
-	var (
-		pipefd, fifofd int
-		consoleSocket  *os.File
-		envInitPipe    = os.Getenv("_LIBCONTAINER_INITPIPE")
-		envFifoFd      = os.Getenv("_LIBCONTAINER_FIFOFD")
-		envConsole     = os.Getenv("_LIBCONTAINER_CONSOLE")
-	)
-
 	// Get the INITPIPE.
-	pipefd, err = strconv.Atoi(envInitPipe)
+	envInitPipe := os.Getenv("_LIBCONTAINER_INITPIPE")
+	pipefd, err := strconv.Atoi(envInitPipe)
 	if err != nil {
 		return fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE=%s to int: %s", envInitPipe, err)
 	}
-
-	var (
-		pipe = os.NewFile(uintptr(pipefd), "pipe")
-		it   = initType(os.Getenv("_LIBCONTAINER_INITTYPE"))
-	)
+	pipe := os.NewFile(uintptr(pipefd), "pipe")
 	defer pipe.Close()
 
 	// Only init processes have FIFOFD.
-	fifofd = -1
+	fifofd := -1
+	envInitType := os.Getenv("_LIBCONTAINER_INITTYPE")
+	it := initType(envInitType)
 	if it == initStandard {
+		envFifoFd := os.Getenv("_LIBCONTAINER_FIFOFD")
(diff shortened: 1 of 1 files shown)


libcontainer/factory_linux.go
```diff
@@ -336,33 +336,33 @@ func (l *LinuxFactory) Type() string {
...
-
-	var (
-		pipe = os.NewFile(uintptr(pipefd), "pipe")
-		it   = initType(os.Getenv("_LIBCONTAINER_INITTYPE"))
-	)
-	defer pipe.Close()
-
 	// Only init processes have FIFOFD.
 	fifofd = -1
+	it = initType(envInitType)
```
thaJeztah, line 351: Perhaps remove them from the `var(  )` block and just;

```suggestion
	fifofd := -1
	it := initType(envInitType)
```

libcontainer/factory_linux.go
```diff
@@ -336,33 +336,33 @@ func (l *LinuxFactory) Type() string {
...
 			return fmt.Errorf("unable to convert _LIBCONTAINER_FIFOFD=%s to int: %s", envFifoFd, err)
 		}
 	}
 
+	// Get the INITPIPE.
+	pipefd, err := strconv.Atoi(envInitPipe)
+	if err != nil {
+		return fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE=%s to int: %s", envInitPipe, err)
+	}
+	pipe = os.NewFile(uintptr(pipefd), "pipe")
```
thaJeztah, line 363: this could also be declared here inline instead of at the start

```suggestion
	pipe := os.NewFile(uintptr(pipefd), "pipe")
```

zvier: Ok, it seems more clean, and I also move consoleSocket declare down close to where it used.

libcontainer/factory_linux.go
```diff
@@ -336,33 +336,30 @@ func (l *LinuxFactory) Type() string {
...
 	var (
-		pipefd, fifofd int
-		consoleSocket  *os.File
-		envInitPipe    = os.Getenv("_LIBCONTAINER_INITPIPE")
-		envFifoFd      = os.Getenv("_LIBCONTAINER_FIFOFD")
-		envConsole     = os.Getenv("_LIBCONTAINER_CONSOLE")
+		envInitPipe = os.Getenv("_LIBCONTAINER_INITPIPE")
+		envInitType = os.Getenv("_LIBCONTAINER_INITTYPE")
+		envFifoFd   = os.Getenv("_LIBCONTAINER_FIFOFD")
+		envConsole  = os.Getenv("_LIBCONTAINER_CONSOLE")
```
kolyshkin, line 342: Now you don't even need `var` block, just `envXXX := os.Getenv(YYY)`.

In fact, I do not like that all four variables are initialized at the start -- I'd rather move them closer to the place of use.
For one thing, `envFifoFd` is only used in case `it == initStandard` so it can be initialized right there under the `if` block.

thaJeztah: Ah, yes, could be something like

```go
	if c := os.Getenv("_LIBCONTAINER_CONSOLE"); c != "" {
		console, err := strconv.Atoi(c)
		if err != nil {
			return fmt.Errorf("unable to convert _LIBCONTAINER_CONSOLE=%s to int: %s", c, err)
		}
		consoleSocket = os.NewFile(uintptr(console), "console-socket")
		defer consoleSocket.Close()
	}
```

zvier: ok，I get it, declare variables closer to the place where it use is also perfect.

kolyshkin (approved):
LGTM

kolyshkin:
@thaJeztah @AkihiroSuda PTAL

thaJeztah (approved):
LGTM, thanks!

AkihiroSuda (approved):

PR #2529, opened 2020-07-28
devices: correctly check device types

cyphar:
(mode&S_IFCHR == S_IFCHR) is the wrong way of checking the type of an
inode because the S_IF* bits are actually not a bitmask and instead must
be checked using S_IF*. This bug was neatly hidden behind a (major == 0)
sanity-check but that was removed by [me][1].

In addition, add a test that makes sure that HostDevices() doesn't give
rubbish results -- because [we broke this and fixed this before][2].

[1]: 24388be71e1a ("configs: use different types for .Devices and .Resources.Devices")
[2]: 3ed492ad33f3 ("Handle non-devices correctly in DeviceFromPath")

Fixes: b0d014d0e1a4 ("libcontainer: one more switch from syscall to x/sys/unix")

### DIFF
--- a/libcontainer/devices/devices_test.go
+++ b/libcontainer/devices/devices_test.go
@@ -2,11 +2,18 @@ package devices
 
 import (
 	"errors"
+	"io/ioutil"
 	"os"
 	"testing"
 
+	"github.com/opencontainers/runc/libcontainer/configs"
 	"golang.org/x/sys/unix"
 )
+
+func cleanupTest() {
+	unixLstat = unix.Lstat
+	ioutilReadDir = ioutil.ReadDir
+}
 
 func TestDeviceFromPathLstatFailure(t *testing.T) {
 	testError := errors.New("test error")
@@ -15,6 +22,7 @@ func TestDeviceFromPathLstatFailure(t *testing.T) {
 	unixLstat = func(path string, stat *unix.Stat_t) error {
 		return testError
 	}
+	defer cleanupTest()
 
 	_, err := DeviceFromPath("", "")
 	if err != testError {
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Can we have rc92 right after merging this?

cyphar:
Sure, especially since this breaks Docker.

libcontainer/devices/devices_test.go
```diff
@@ -61,3 +62,33 @@ func TestHostDevicesIoutilReadDirDeepFailure(t *testing.T) {
 		t.Fatalf("Unexpected error %v, expected %v", err, testError)
 	}
 }
+
+func HostDevicesAllValid(t *testing.T) {
```
AkihiroSuda, line 66: The function name needs to have `Test` prefix

tao12345666333:
Thanks!
This bug is too hidden. I have been debugging this problem for several days. :joy_cat:

libcontainer/devices/devices_test.go
cyphar: Oops.

libcontainer/devices/devices_test.go
```diff
@@ -55,9 +65,44 @@ func TestHostDevicesIoutilReadDirDeepFailure(t *testing.T) {
 
 		return []os.FileInfo{fi}, nil
 	}
+	defer func() {
```
AkihiroSuda, line 68: nit: I feel this should be defined as a named function like `func restoreInjectedFuncs()`

thaJeztah:
Thanks @cyphar 👍

PR #2533, opened 2020-07-30
use criu cgroup mode const from go-criu

XiaodongLoong:
Signed-off-by: liuxiaodong <liuxiaodong@loongson.cn>

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	criu "github.com/checkpoint-restore/go-criu/v4/rpc"
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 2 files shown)


kolyshkin:
@adrianreber PTAL

kolyshkin:
> and CRIU defaults to SOFT if nothing is set

so it's not just useless but also has wrong value. Yeah I guess it makes sense to remove it

kolyshkin:
Ideally though, we should get those values from go-criu, rather than hardcoding it ourselves. WDYT @adrianreber ?

adrianreber:
> Ideally though, we should get those values from go-criu, rather than hardcoding it ourselves. WDYT @adrianreber ?

Sounds like a good idea. I just checked and they are already exported in go-criu:

https://github.com/checkpoint-restore/go-criu/blob/master/rpc/rpc.pb.go#L25

runc could just use those values.

XiaodongLoong:
OK, I would use those values from go-criu.

kolyshkin:
@XiaodongLoong you forgot to change the commit message and the PR description. The code LGTM.

XiaodongLoong:
@kolyshkin I changed the commit message and PR description. PTAL

PR #2534, opened 2020-07-30
Pass location of CRIU binary to go-criu

adrianreber:
If the CRIU binary is in a non $PATH location and passed to runc via '--criu /path/to/criu', this information has not been passed to go-criu and since the switch to use go-criu for CRIU version detection, non $PATH CRIU usage was broken. This uses the newly added go-criu interface to pass the location of the binary to go-criu.


go.sum
```diff
@@ -1,6 +1,8 @@
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/checkpoint-restore/go-criu/v4 v4.0.2 h1:jt+rnBIhFtPw0fhtpYGcUOilh4aO9Hj7r+YLEtf30uA=
 github.com/checkpoint-restore/go-criu/v4 v4.0.2/go.mod h1:xUQBLp4RLc5zJtWY++yjOoMoB5lihDt7fai+75m+rGw=
+github.com/checkpoint-restore/go-criu/v4 v4.1.0 h1:WW2B2uxx9KWF6bGlHqhm8Okiafwwx7Y2kcpn8lCpjgo=
+github.com/checkpoint-restore/go-criu/v4 v4.1.0/go.mod h1:xUQBLp4RLc5zJtWY++yjOoMoB5lihDt7fai+75m+rGw=
```
kolyshkin, line 5: looks like you forgot to run `go mod tidy`

kolyshkin:
lgtm except for `go mod tidy` nit

go.sum
adrianreber: I was runnig `make vendor` and that already includes a `go mod tidy`, but running it again removes the old lines (4.0.2) indeed. Will update this PR.

Issue #2485, opened 2020-06-23
terminal: true with no stdin causes container startup failure


cyphar:
If you try to run a container that has `"terminal": true` configured in non-detached mode with a dummy stdin (like `/dev/null`), you get errors:

```
% sudo runc run -b bundle ctr </dev/null
WARN[0000] exit status 2                                
ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:459: container init caused: 
```

Found as part of opencontainers/umoci#296.

kolyshkin:
Most probably caused by assumption that fd 0 is a terminal (in `libcontainer/system.Setctty()`). I barely remember it was causing a real error before, now it's some weirdness:

With the following patch:
```diff
diff --git a/libcontainer/container_linux.go b/libcontainer/container_linux.go
index e958d159..f7cf9a64 100644
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -358,7 +358,7 @@ func (c *linuxContainer) start(process *Process) error {
        if err := parent.start(); err != nil {
                // terminate the process to ensure that it properly is reaped.
                if err := ignoreTerminateErrors(parent.terminate()); err != nil {
-                       logrus.Warn(err)
+                       logrus.Warnf("terminate parent: %v", err)
                }
                return newSystemErrorWithCause(err, "starting container process")
        }
diff --git a/libcontainer/standard_init_linux.go b/libcontainer/standard_init_linux.go
index f5be67c0..7002086a 100644
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -100,6 +100,7 @@ func (l *linuxStandardInit) Init() error {
                }
        }
 
+       return errors.New("after CreateConsole")
        // Finish the rootfs setup.
        if l.config.Config.Namespaces.Contains(configs.NEWNS) {
                if err := finalizeRootfs(l.config.Config); err != nil {
```

it breaks in 3 different ways (in 3 identical consecutive runs):

```console
[kir@kir-rhat tst]$ sudo ../runc --debug run xx44 < /dev/null
DEBU[0000] nsexec:601 nsexec started                    
DEBU[0000] child process in init()                      
DEBU[0000] logging has already been configured          
DEBU[0000] log pipe has been closed: EOF                
WARN[0000] terminate parent: exit status 2              
ERRO[0000] container_linux.go:363: starting container process caused: process_linux.go:450: container init caused: after CreateConsole 
[...]

PR #2535, opened 2020-07-31
Fix run/exec with terminal: true in case stdin is not a terminal

kolyshkin:
This fixes the following failure (and the likes):

```    
sudo runc run -b bundle ctr </dev/null
WARN[0000] exit status 2
ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:459: container init caused:
```
    
The "exit status 2" with no error message is caused by SIGHUP
which is sent to init by the kernel when we are losing the
controlling terminal. If we choose to ignore that, we get
the same result, this time caused by panic in console.Current(),
which is addressed by [1].
    
Otherwise, the issue here is simple: the code assumes stdin
is opened to a terminal, and fails to work otherwise. Some
standard Linux tools (e.g. stty, top) do the same (modulo panic),
while some others (reset, tput) use the trick of trying
all the three std streams (starting with stderr as it is least likely
to be redirected), and if all three fails, open /dev/tty.
    
This commit does a similar thing (see `initHostConsole`).
    
It also replaces the call to console.Current(), which may panic
(see [1]), by reusing the t.hostConsole.
    
Finally, a couple of simple test cases are added.

Alas, I'm not sure how to automate the test case when runc is
run without a controlling terminal. I have reproduced it manually:

```console
[root@kir-rhat runc]# cat /root/runme 
cd /home/kir/go/src/github.com/opencontainers/runc
./runc run -b tst xxx-$$
echo $?
[root@kir-rhat runc]# ssh -T root@localhost ./runme
time="2020-07-31T18:51:55-07:00" level=error msg="open /dev/tty: no such device or address"
1
[root@kir-rhat runc]# 
```
(which is much better than mysterious "container init caused" error, right?)
    
[1] https://github.com/containerd/console/pull/37

Fixes: https://github.com/opencontainers/runc/issues/2485

### DIFF
--- a/tty.go
+++ b/tty.go
@@ -12,16 +12,17 @@ import (
 	"github.com/containerd/console"
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/utils"
+	"github.com/pkg/errors"
 )
 
 type tty struct {
-	epoller   *console.Epoller
-	console   *console.EpollConsole
-	stdin     console.Console
-	closers   []io.Closer
-	postStart []io.Closer
-	wg        sync.WaitGroup
-	consoleC  chan error
+	epoller     *console.Epoller
+	console     *console.EpollConsole
+	hostConsole console.Console
+	closers     []io.Closer
+	postStart   []io.Closer
+	wg          sync.WaitGroup
+	consoleC    chan error
 }
 
 func (t *tty) copyIO(w io.Writer, r io.ReadCloser) {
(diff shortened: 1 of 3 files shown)


tty.go
```diff
@@ -99,18 +100,32 @@ func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) (Err error
...
+			continue
+		case nil:
+			hostConsole = c
+		default:
+			return errors.Wrap(err, "unable to get console")
+		}
 	}
-	if err := stdin.SetRaw(); err != nil {
+	if hostConsole == nil {
+		return errors.New("unable to find console (are all streams redirected?)")
```
cyphar, line 119: I think a nicer solution might be to just not do interrupt or resize handling if there isn't a console, but still allow the container to run. I don't really have an opinion on whether we should check `stdout` or `stderr` ... it feels a *little* wrong because `Ctrl-C` feels like an "input" thing but :shrug:.

kolyshkin: We need a terminal to set the raw mode etc -- and any terminal we have will do. Tools like `clear` and `tput` do the same thing -- they try stderr, stdout, stdin, and if all three fails, open("/dev/tty") should do the same thing.

I didn't do open("/dev/tty") as I am not sure where to close it, but I think I'll implement that later, to make cases like`runc run $ID top </dev/null >/dev/null 2>&1`work.

kolyshkin:
Reworked to open("/dev/tty") as a last resort. Without it, `runc init` can get SIGHUP in case all three streams are redirected.

In case we don't have a terminal at all, runc should fail gracefully and early.

Tested it without a terminal (which is kinda hard -- the only way I know is to use `ssh -T`):
```console
[root@kir-rhat ~]# ssh -T root@localhost cat ./runme
cd /home/kir/go/src/github.com/opencontainers/runc/tst
../runc run -b tst xxx-$$
echo $?

[root@kir-rhat ~]# ssh -T root@localhost ./runme
time="2020-07-31T16:35:47-07:00" level=error msg="chdir tst: no such file or directory"
1
```

tty.go
kolyshkin: > not do interrupt or resize handling if there isn't a console, but still allow the container to run

I disagree. If we say in config we want a console, and there's no [host] console we should bail out. This is what my current code does (see https://github.com/opencontainers/runc/pull/2535#issuecomment-667427446)

kolyshkin:
@cyphar @crosbymichael PTAL

kolyshkin:

tty.go
kolyshkin: @cyphar PTAL ^^^

crosbymichael:
LGTM

kolyshkin:
> LGTM

@crosbymichael we no longer use pullapprove for this repo, so you have to use github-native LGTM (i.e. go to 'commits' or 'file changes', click the 'review changes', select the 'approve' radio button...)

PR #2538, opened 2020-08-01
travis: don't hardcode golang versions

kolyshkin:
Golang is supporting two minor releases (currently 1.14.x and 1.13.x),
and thus once a new minor release is out, the oldest supported one
becomes unsupported (e.g. when 1.15 is out, 1.13 becomes unsupported).

Instead of hardcoding golang versions, let's specify "stable" and
"oldstable" (which currently equals to 1.14.x and 1.13.x). This way,
we'll be sure we're testing stuff using up to date and supported golang
releases.

This was not possible before due to:
 - https://github.com/travis-ci/gimme/issues/179
 - https://github.com/travis-ci/gimme/issues/185

but apparently it's finally fixed. Hooray!

Previous discussion on the topic:
 * https://github.com/opencontainers/runc/pull/2239#discussion_r393100037

Also, add `-mod=vendor` to `go fmt` and `go vet` since it's required
for go 1.13 to honor vendor subdir. This should speed up CI a little bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,19 @@ dist: bionic
 language: go
 os: linux
 go:
-  - 1.14.x
-  - 1.13.x
+  - stable
+  - oldstable
   - tip
 cache:
   directories:
   - /home/travis/.vagrant.d/boxes
 jobs:
   include:
-    - go: 1.14.x
+    - go: stable
       name: "verify-dependencies"
       script:
         - make verify-dependencies
-    - go: 1.13.x
+    - go: oldstable
       name: "cgroup-systemd"
       env:
         - RUNC_USE_SYSTEMD=1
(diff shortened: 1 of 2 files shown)


kolyshkin:
So, `stable` gives us:
```console
$ go version
go version go1.14.6 linux/amd64
```
and `oldstable`:
```console
$ go version
go version go1.13.14 linux/amd64
```
so I guess we're good!

kolyshkin:
@thaJeztah PTAL

AkihiroSuda:
I'm not sure we want this. I think we want to be aware of the tested versions explicitly.

thaJeztah:
Yeah, not a fan; I generally prefer explicit > implicit as well; even if that means a PR to update twice a year

dims:
I agree with @thaJeztah and @AkihiroSuda here

AkihiroSuda:
@kolyshkin btw `Makefile: fix go vet/fmt` commit LGTM, could you cherry-pick in another PR?

kolyshkin:
Closing as per comments above. The Makefile fix is separated out to #2545

PR #2541, opened 2020-08-04
add Go version and libseccomp version to `runc --version`

AkihiroSuda:
Printing these information would be helpful to debug runtime-related errors.

```console
$ runc --version
runc version 1.0.0-rc91+dev
commit: bdbc3b06b4690dbb29838cb8358969451f21e62e
spec: 1.0.2-dev
go: go1.14.6
libseccomp: 2.4.3
```

### DIFF
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -266,3 +266,8 @@ func parseStatusFile(path string) (map[string]string, error) {
 
 	return status, nil
 }
+
+// Version returns major, minor, and micro.
+func Version() (uint, uint, uint) {
+	return libseccomp.GetLibraryVersion()
+}
--- a/libcontainer/seccomp/seccomp_unsupported.go
+++ b/libcontainer/seccomp/seccomp_unsupported.go
@@ -22,3 +22,8 @@ func InitSeccomp(config *configs.Seccomp) error {
 func IsEnabled() bool {
 	return false
 }
+
+// Version returns major, minor, and micro.
+func Version() (uint, uint, uint) {
+	return 0, 0, 0
+}
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/logs"
@@ -62,6 +63,7 @@ func main() {
 		v = append(v, fmt.Sprintf("commit: %s", gitCommit))
 	}
 	v = append(v, fmt.Sprintf("spec: %s", specs.Version))
+	v = append(v, fmt.Sprintf("go: %s", runtime.Version()))
 	app.Version = strings.Join(v, "\n")
 
 	xdgRuntimeDir := ""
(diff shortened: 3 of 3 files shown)


cyphar:
Will this cause issues with Docker? They parse the output of `runc --version` and have in the past been very particular about the version output (I've had issues with Docker refusing to start if the output doesn't have the right number of fields -- as well as Docker giving you warnings if the fields aren't what it expects).

Aside from that, I'm in favour of the idea.

AkihiroSuda:
> Will this cause issues with Docker?

No, the latest version is robust enough.

https://github.com/moby/moby/blob/v19.03.12/daemon/info_unix.go#L224-L248

AkihiroSuda:
>  it also wouldn't hurt to include what Go build tags

This should be discussed in another issue, as it is not obvious how to obtain the tags and how it should be printed.

cyphar:
> This should be discussed in another issue, as it is not obvious how to obtain the tags and how it should be printed.

Well, you could just do it by adding `-X github.com/opencontainers/runc/buildTags=$(BUILD_TAGS)` to the compilation flags. A string is fine since build tags are space-separated anyway. But sure, we can do this in a separate PR (it would only be a couple of lines though).

AkihiroSuda:
>  A string is fine since build tags are space-separated anyway. But sure, we can do this in a separate PR (it would only be a couple of lines though).

crun uses systemd-ish form:
```console
$ crun --version
crun version 0.14.1.10-a0ec
commit: a0ec203cbce47f3d48b8fcc0aa784ba7d87cba17
spec: 1.0.0
+SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
```


Let's merge this PR as-is and discuss tags in another issue/PR

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -266,3 +266,8 @@ func parseStatusFile(path string) (map[string]string, error) {
 
 	return status, nil
 }
+
+// GetLibseccompVersion returns major, minor, and micro.
+func GetLibseccompVersion() (uint, uint, uint) {
```
kolyshkin, line 271: since 
 - the package name is already `seccomp`;
 - golang style discourages the `Get` prefix;

maybe just `Version` (or `LibVersion`, or `LibraryVersion`).

kolyshkin: > golang style discourages the `Get` prefix

This is what I mean: https://golang.org/doc/effective_go.html#Getters

KentaTada:
Available seccomp features also depend on the version of libseccomp-golang.
I think we may need to add the version of libseccomp-golang.

libcontainer/seccomp/seccomp_linux.go
AkihiroSuda: updated

AkihiroSuda:
> I think we may need to add the version of libseccomp-golang.

No, it is available in go.mod, and we don't want to duplicate the go.mod content in `runc --version`.

AkihiroSuda:
@cyphar PTAL

kolyshkin:
@mrunalp @crosbymichael PTAL

crosbymichael (approved):
LGTM

PR #2543, opened 2020-08-05
Release 1.0.0 rc92

mrunalp:


VERSION
```diff
@@ -1 +1 @@
-1.0.0-rc91+dev
+1.0.0-rc93-dev
```
AkihiroSuda, line 1: 92+dev

mrunalp: fixed.

dims:
LGTM (non-binding)

cyphar:
Uh... we needed more votes for a release @kolyshkin. :P

Oh well, I think we all agreed the whole 2/3rds vote thing was a pain anyway (technically according to the contributor guidelines we only need to have 2 LGTMs because releases aren't special -- at least with the current rules.) Preparing the release...

AkihiroSuda:
Where do we have the release guideline?

cyphar:
Release tagged: https://github.com/opencontainers/runc/releases/tag/v1.0.0-rc92

> Where do we have the release guideline?

It's not really written down anywhere (in fact, most of `MAINTAINERS_GUIDE.md` is completely inaccurate and nobody bothered to updated it -- it comes from the old Docker maintainers guide). The logic for the two-thirds requirement is that is what the OCI vote requirements are, and specifications have to get a two-thirds vote of maintainers to do a release. In theory this requirement has never applied to runc but we've done it that way for several years.

However, `umoci` now has a far more lax release system (allowing 3 LGTMs to be enough for a release) and we have discussed in other issues that we want to loosen the release rules in the future. In my view we should have something like:

  * 3 LGTMs required for any non-major release.
  * Major releases (`1.0.0`, `2.0.0` and so on -- as well as the first rc of each) require a two-thirds vote of all maintainers since it's a major change to the project.

We also should completely rewrite `MAINTAINERS_GUIDE.md` (as well as `CONTRIBUTING.md`) to correctly reflect how development works in runc today.

AkihiroSuda:
> a two-thirds vote of all maintainers

a two-thirds vote in favor from maintainers who voted within seven days?

PR #2549, opened 2020-08-09
Add shellcheck to bats files

kolyshkin:
This PR
 - fixes (or adds annotations to ignore) all the shellcheck warnings for .bats files;
 - add shellcheck to the CI cycle.

TODO (not in this PR):
 - fix shellcheck warnings for helpers.bash

~~This currently includes #2548; I'll rebase once it's merged.~~ rebased

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,7 @@ validate:
 	script/validate-gofmt
 	script/validate-c
 	$(GO) vet $(MOD_VENDOR) ./...
+	shellcheck tests/integration/*.bats
 
 ci: validate test release
 
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -170,6 +170,7 @@ function setup() {
     [[ ${lines[0]} == "0::/foo" ]]
 
 # teardown: remove "/foo"
+    # shellcheck disable=SC2016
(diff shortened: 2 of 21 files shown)


Makefile
```diff
@@ -122,7 +122,10 @@ validate:
 	script/validate-c
 	$(GO) vet ./...
 
-ci: validate test release
+shellcheck:
+	shellcheck tests/integration/*.bats
```
AkihiroSuda, line 126: Should be inside `make validate`?

AkihiroSuda: Also, apt-get install shellcheck in Dockerfile?

kolyshkin: good point, fixed

thaJeztah: Note that there's upstream docker images, so instead of installing it in the dockerfile, you could also run a separate container just for shell check https://hub.docker.com/r/koalaman/shellcheck-alpine/tags

kolyshkin: We do not use shellcheck from inside a Docker container. We run `make ci` on the host (in Travis-CI environment that already guarantees that `shellcheck` is installed), and in the Makefile the `ci` target depends on `validate`, which does some checks on the host. It's the `test` target that uses dockerfile.

So, no need to install `shellcheck` from Dockerfile (or use a docker image with shellcheck).

PR #2554, opened 2020-08-16
Allow searching for apparmor_parser in $PATH

saschagrunert:
Not all distributions ship the `apparmor_parser` in `/sbin/` (like NixOS).
To be able to use the AppArmor feature on such distributions as well we
have to check for the `apparmor_parser` in the `$PATH` environment
variable.


saschagrunert:
@AkihiroSuda you dismissed the review comment, I still added the check for `/sbin`

saschagrunert:
@mrunalp @cyphar please take a look, too :)

dims:
LGTM 👍

libcontainer/apparmor/apparmor.go
```diff
@@ -6,17 +6,28 @@ import (
...
+
+		// /sbin/ is not always in $PATH, so we check it explicitly
+		if _, err = os.Stat("/sbin/apparmor_parser"); err == nil {
+			return isEnabledByParameters()
+		}
+
+		// Fallback to checking $PATH
+		if _, err = exec.LookPath("apparmor_parser"); err == nil {
+			return isEnabledByParameters()
+		}
```
cyphar, line 30: I'm not sure that these checks are actually necessary -- the logic is "well, a box without `apparmor_parser` isn't a working AppArmor distribution" but I don't think that makes sense to enforce that in runc. All we care about is whether AppArmor is enabled in the kernel -- and if the user isn't actually using the profiles then the `config.json` won't contain an AppArmor profile setting anyway.

This check comes from Docker, which *does* use `apparmor_parser` (so this check makes a little bit more sense there) but we don't in runc so I think it'd just be simpler to drop these checks and just check `/sys/kernel` and `/sys/module`.

And as an aside, while we aren't executing any binaries here, `$PATH` lookups in runc make me nervous.

libcontainer/apparmor/apparmor.go
```diff
@@ -6,17 +6,28 @@ import (
...
+	"os/exec"
 
 	"github.com/opencontainers/runc/libcontainer/utils"
 )
 
 // IsEnabled returns true if apparmor is enabled for the host.
 func IsEnabled() bool {
 	if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil && os.Getenv("container") == "" {
-		if _, err = os.Stat("/sbin/apparmor_parser"); err == nil {
+		isEnabledByParameters := func() bool {
```
kolyshkin, line 17: The code looks way too complicated to me, especially with an introduced function.

We can do something like this instead
```go
if _, err := os.Stat("/sys/kernel/security/apparmor"); err != nil {
  return false
}
if os.Getenv("container") != "" {
  return false
}
if _, err = os.Stat("/sbin/apparmor_parser"); err != nil {
  if _, err = exec.LookPath("apparmor_parser"); err != nil {
    return false
  }
}
buf, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
return err == nil && len(buf) > 1 && buf[0] == 'Y'
```

libcontainer/apparmor/apparmor.go
```diff
@@ -6,17 +6,28 @@ import (
...
 			buf, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
 			return err == nil && len(buf) > 1 && buf[0] == 'Y'
```
cyphar, line 19: I think `bytes.HasPrefix(buf, []byte{'Y'})` is a bit less awful to read.

libcontainer/apparmor/apparmor.go
cyphar: See my comment -- I don't think checking for `apparmor_parser` is necessary. Also why do we have the `container` env check? Does LXC set that or something?

libcontainer/apparmor/apparmor.go
saschagrunert: Yep, changed the check to check for the "Y" prefix to not change the behavior.

saschagrunert:
I proposed to remove the check with respect to the discussions on https://github.com/opencontainers/runc/pull/2554#discussion_r471195294 and https://github.com/opencontainers/runc/pull/2554#discussion_r471195610.

libcontainer/apparmor/apparmor.go
saschagrunert: Yes I agree, I think we could omit the check for `apparmor_parser` and let higher level runtimes do this work.

libcontainer/apparmor/apparmor.go
kolyshkin: Dusting off git history, the check for container env var comes from https://github.com/moby/moby/pull/5534. And yes, `container=` was set at least by Docker's LXC driver.

I think it's obsoleted by now and should be removed.

libcontainer/apparmor/apparmor.go
```diff
@@ -13,10 +14,8 @@ import (
 // IsEnabled returns true if apparmor is enabled for the host.
 func IsEnabled() bool {
 	if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil && os.Getenv("container") == "" {
```
kolyshkin, line 16: As suggested in the comments above, maybe we need to drop the `container` env var check (as a separate commit)

libcontainer/apparmor/apparmor.go
kolyshkin: Me, too, thought about `HasPrefix`, and decided that `len(buf) > 1 && buf[0] == 'Y'` is good enough :)
Anyway, HasPrefix is good enough, too.

kolyshkin:
> The original commit where this check has been added is 44c7afa (cc @tifayuki)

The code was actually in a different repository back then. Tracking it back to the original repo, this was done in https://github.com/docker-archive/libcontainer/pull/532 which describes the reason for this (docker tries to exec this binary and fails).

More interestingly, this check was later removed (see https://github.com/docker-archive/libcontainer/pull/645) but somehow the change was not propagated to runc.

So, yes, I agree, there's no need to check for the binary existence

libcontainer/apparmor/apparmor.go
cyphar: I agree -- let's drop the `container` check too.

saschagrunert:
> We should also drop the `container` environment variable check.

Dropped that check as well :+1:

cyphar:
Now shellcheck is failing. You need to rebase on top of #2549. Sigh...

saschagrunert:
> Now shellcheck is failing. You need to rebase on top of #2549 (or rather, `master`).

Rebased on top of the latest master branch :heavy_check_mark:

saschagrunert:
Looks like the master is broken, too. :)

libcontainer/apparmor/apparmor.go
```diff
@@ -12,11 +13,9 @@ import (
 
 // IsEnabled returns true if apparmor is enabled for the host.
 func IsEnabled() bool {
-	if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil && os.Getenv("container") == "" {
-		if _, err = os.Stat("/sbin/apparmor_parser"); err == nil {
-			buf, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
-			return err == nil && len(buf) > 1 && buf[0] == 'Y'
-		}
+	if _, err := os.Stat("/sys/kernel/security/apparmor"); err == nil {
+		buf, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
```
kolyshkin, line 17: No need to call `os.Stat()` here as we already check for error from `ioutils.ReadFile`.

So this should be something like
```go
func IsEnabled() bool {
   buf, err := ioutil.ReadFile("/sys/module/apparmor/parameters/enabled")
   return err == nil && bytes.HasPrefix(buf, []byte("Y"))
}
```

kolyshkin:
> Looks like the master is broken, too. :)

Once #2561 is fixed you'll need to rebase this one; sorry for the trouble

libcontainer/apparmor/apparmor.go
cyphar: No, the `Stat` is necessary -- the paths are slightly different (one is for `apparmorfs` and the other is the module parameter check).

kolyshkin:
> Once #2561 is fixed you'll need to rebase this one; sorry for the trouble

@saschagrunert the fix is merged; please rebase

saschagrunert:
> > Once #2561 is fixed you'll need to rebase this one; sorry for the trouble
> 
> @saschagrunert the fix is merged; please rebase

Thanks for the hint, rebased on top of the latest master branch.

libcontainer/apparmor/apparmor.go
kolyshkin: Ah, I overlooked it, thanks!

dims:
LGTM

thaJeztah:
@kolyshkin @AkihiroSuda regarding https://github.com/opencontainers/runc/pull/2554#discussion_r471809575

> Dusting off git history, the check for container env var comes from [moby/moby#5534](https://github.com/moby/moby/pull/5534). And yes, `container=` was set at least by Docker's LXC driver.
> 
> I think it's obsoleted by now and should be removed.

Wondering if the env-var is still be relevant when running containerised (`docker-in-docker`); The `dind` script still sets it;
https://github.com/docker/docker/blob/ed89041433a031cafc0a0f19cfe573c31688d377/hack/dind#L13-L14

(originally was added to address https://github.com/moby/moby/pull/5529)

/cc @tianon PTAL

PR #2555, opened 2020-08-17
seccomp: first check via prctl, then parse the status file

saschagrunert:
This might be a premature optimization but it feels parsing the status
file implies a higher performance cost than executing the syscalls
first.


saschagrunert:
PTAL @AkihiroSuda @cyphar

AkihiroSuda:
I agree with Kir, let me close this

cyphar:
:+1: on Kir's point for why we do it this way -- but in addition, I'm not sold that there would be a measurable performance difference (in this code which gets called very infrequently). `procfs` file access is _very_ fast.

Issue #2559, opened 2020-08-19
CI: shellcheck is failing on master


AkihiroSuda:
https://travis-ci.org/github/opencontainers/runc/jobs/718839937 (https://github.com/opencontainers/runc/commit/a5847db387ae28c0ca4ebe4beee1a76900c86414)

```
...
shellcheck tests/integration/*.bats
In tests/integration/delete.bats line 64:
  for i in $(seq 1); do
  ^-^ SC2034: i appears unused. Verify use (or export if used externally).
In tests/integration/delete.bats line 65:
    runc run -d --console-socket $CONSOLE_SOCKET test_busybox
                                 ^-------------^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
    runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
In tests/integration/delete.bats line 92:
      eval path=\$${name}/foo
                  ^-----^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
      eval path=\$"${name}"/foo
In tests/integration/delete.bats line 93:
      echo $path
           ^---^ SC2154: path is referenced but not assigned.
           ^---^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
      echo "$path"
In tests/integration/delete.bats line 94:
      [ -d ${path} ] || fail "test failed to create memory sub-cgroup"
           ^-----^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
      [ -d "${path}" ] || fail "test failed to create memory sub-cgroup"
For more information:
  https://www.shellcheck.net/wiki/SC2034 -- i appears unused. Verify use (or ...
  https://www.shellcheck.net/wiki/SC2154 -- path is referenced but not assign...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
Makefile:121: recipe for target 'validate' failed
```

Seems caused by merging https://github.com/opencontainers/runc/pull/2506 after https://github.com/opencontainers/runc/pull/2549 without rebasing.

kolyshkin:
> Seems caused by merging #2506 after #2549 without rebasing.

That's right, the fix is coming

PR #2561, opened 2020-08-19
shellcheck CI files

kolyshkin:
#### 1. tests/int/delete.bats: fix shellcheck warnings.

Those are caused by merging #2506 after #2549 without a rebase.
Fixes: #2559 

The failures were:

```
In tests/integration/delete.bats line 64:
  for i in $(seq 1); do
  ^-^ SC2034: i appears unused. Verify use (or export if used externally).


In tests/integration/delete.bats line 65:
    runc run -d --console-socket $CONSOLE_SOCKET test_busybox
                                 ^-------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 


In tests/integration/delete.bats line 92:
      eval path=\$${name}/foo
                  ^-----^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      eval path=\$"${name}"/foo


In tests/integration/delete.bats line 93:
      echo $path
           ^---^ SC2154: path is referenced but not assigned.
           ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      echo "$path"


In tests/integration/delete.bats line 94:
      [ -d ${path} ] || fail "test failed to create memory sub-cgroup"
           ^-----^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      [ -d "${path}" ] || fail "test failed to create memory sub-cgroup"


In tests/integration/tty.bats line 73:
	runc run -d --console-socket $CONSOLE_SOCKET test_busybox
                                     ^-------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- i appears unused. Verify use (or ...
  https://www.shellcheck.net/wiki/SC2154 -- path is referenced but not assign...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

#### 2. tests/int: alt fix for shellcheck SC2034

Instead of ignoring it, use a throwaway variable `_`
(yes, just like in Go).

### DIFF
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -61,7 +61,7 @@ function teardown() {
 
   local subsystems="memory freezer"
 
-    runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+    runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
     [ "$status" -eq 0 ]
 
     testcontainer test_busybox running
@@ -88,9 +88,9 @@ EOF
 
     for s in ${subsystems}; do
       name=CGROUP_${s^^}
-      eval path=\$${name}/foo
-      echo $path
-      [ -d ${path} ] || fail "test failed to create memory sub-cgroup"
+      eval path=\$"${name}"/foo
+      # shellcheck disable=SC2154
+      [ -d "${path}" ] || fail "test failed to create memory sub-cgroup ($path not found)"
     done
 
     runc delete --force test_busybox
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -61,7 +61,6 @@ function teardown() {
 
   local subsystems="memory freezer"
 
-  for i in $(seq 1); do
     runc run -d --console-socket $CONSOLE_SOCKET test_busybox
     [ "$status" -eq 0 ]
 
@@ -102,8 +101,6 @@ EOF
     run find /sys/fs/cgroup -wholename '*testbusyboxdelete*' -type d
     [ "$status" -eq 0 ]
     [ "$output" = "" ] || fail "cgroup not cleaned up correctly: $output"
-
-  done
 }
 
 @test "runc delete --force in cgroupv2 with subcgroups" {
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -61,8 +61,7 @@ function simple_cr() {
 
   testcontainer test_busybox running
 
-  # shellcheck disable=SC2034
-  for i in $(seq 2); do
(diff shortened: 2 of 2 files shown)


tests/integration/delete.bats
```diff
@@ -61,8 +61,8 @@ function teardown() {
 
   local subsystems="memory freezer"
 
-  for i in $(seq 1); do
-    runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+  for _ in $(seq 1); do
```
AkihiroSuda, line 64: Why do we need this seq?

tests/integration/delete.bats
```diff
@@ -89,9 +89,10 @@ EOF
 
     for s in ${subsystems}; do
       name=CGROUP_${s^^}
-      eval path=\$${name}/foo
-      echo $path
-      [ -d ${path} ] || fail "test failed to create memory sub-cgroup"
+      eval path=\$"${name}"/foo
+      # shellcheck disable=SC2154
+      echo "$path"
```
AkihiroSuda, line 94: Can we remove this echo?

kolyshkin: Looks like we can, but the warning (and thus the `shellcheck disable=` still remains since shellcheck don't grok eval (https://github.com/koalaman/shellcheck/wiki/SC2034#indirection).

Updated

kolyshkin:
@AkihiroSuda @mrunalp PTAL

tests/integration/delete.bats
kolyshkin: Good question :) I wrote this code a month ago and I have no idea whatsoever :-\

My best guess is I wanted to repeat the test a few times (for some reason -- testing?), but yeah, it does not make any sense now.

Removing

kolyshkin: removed

saschagrunert (approved):

tests/integration/delete.bats
```diff
@@ -61,8 +61,7 @@ function teardown() {
 
   local subsystems="memory freezer"
 
-  for i in $(seq 1); do
-    runc run -d --console-socket $CONSOLE_SOCKET test_busybox
+    runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
```
AkihiroSuda, line 64: indent

AkihiroSuda (changes requested):

tests/integration/delete.bats
kolyshkin: please see the commit message

AkihiroSuda (approved):

Issue #2115, opened 2019-08-22
Optimize writeFile function usage in libcontainer/cgroups/fs


fristonio:
There are a lot of places in `libcontainer/cgroups/fs` where the usage of `writeFile` function can be optimised. For example here https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/blkio.go#L44-L51
the function call to `writeFile` is made for each entry in the slice `BlkioWeightDevice`, which in turn opens and closes file for each entry. This can be optimised by joining all the strings and then writing only once.
I haven't looked around much but this might be the case at other places too.

PS: Let me know if you need help with this, I will be happy to send a Pull Request.

mattbonnell:
I can take a stab at this, will work on a PR

kolyshkin:
> This can be optimised by joining all the strings and then writing only once.

Joining the strings might result in increased memory footprint. It might be better to find a solution to write a set of strings to a file in some way, or just do usual open / for ... write / close

mattbonnell:
> or just do usual open / for ... write / close

This is the approach I’m taking, yeah.

mattbonnell:
In light of @cyphar's investigations into the kernel source, it seems these interfaces can only accept one entry at a time, anyways. I think this issue can be closed.

kolyshkin:
Closing as per https://github.com/opencontainers/runc/pull/2568#issuecomment-682472705

PR #2568, opened 2020-08-27
 Optimize file-writing in libcontainer/cgroups/fs

mattbonnell:
This optimizes file-writing in the `libcontainer/cgroups/fs` package, addressing #2115

Originally, I sought to optimize file manipulation by opening and closing files once outside of each for-loop, as suggested [here](https://github.com/opencontainers/runc/issues/2115#issuecomment-681317084) by @kolyshkin. 

However, this test in `libcontainer/cgroups/fs/blkio_test.go` alerted me to the fact that only the last-written values are kept:
https://github.com/opencontainers/runc/blob/09ddc63afdde16d5fb859a1d3ab010bd45f08497/libcontainer/cgroups/fs/blkio_test.go#L140-L175

So, in this PR I refactored these looping write patterns so that only the last value in each slice is written to file.

I also define a `joinPath` helper to reduce duplication between `fscommon.WriteFile()` and `fscommon.ReadFile()`

### DIFF
--- a/libcontainer/cgroups/fs/blkio.go
+++ b/libcontainer/cgroups/fs/blkio.go
@@ -38,7 +38,9 @@ func (s *BlkioGroup) Set(path string, cgroup *configs.Cgroup) error {
 			return err
 		}
 	}
-	for _, wd := range cgroup.Resources.BlkioWeightDevice {
+
+	if len(cgroup.Resources.BlkioWeightDevice) > 0 {
+		wd := cgroup.Resources.BlkioWeightDevice[len(cgroup.Resources.BlkioWeightDevice)-1]
 		if err := fscommon.WriteFile(path, "blkio.weight_device", wd.WeightString()); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


cyphar:
> However, this test in `libcontainer/cgroups/fs/blkio_test.go` alerted me to the fact that only the last-written values are kept:

NACK, I think you've misunderstood what the comment in the test is saying. These files are part of the kernel API and writing to them (without `O_APPEND`) will result in the configuration being appended -- that is why we write each entry. However when we test this code we mock out the kernel interfaces and when you use ordinary files the old contents get removed -- this is not the actual behaviour of the kernel and your patch would break existing configurations.

Unless I'm mistaken it would be simpler to just generate one big string to write rather than doing it once for each option, but you'd have to check that the kernel allows that for these cgroup files.

mattbonnell:
> > However, this test in `libcontainer/cgroups/fs/blkio_test.go` alerted me to the fact that only the last-written values are kept:
> 
> NACK, I think you've misunderstood what the comment in the test is saying. These files are part of the kernel API and writing to them (without `O_APPEND`) will result in the configuration being appended -- that is why we write each entry. However when we test this code we mock out the kernel interfaces and when you use ordinary files the old contents get removed -- this is not the actual behaviour of the kernel and your patch would break existing configurations. A more complete test would use something like a FIFO and read from it in a different goroutine but that's kind of overkill for these tests IMHO.

Ah I see, that certainly makes more sense. I felt like I must be missing something.

> Unless I'm mistaken it would be simpler to just generate one big string to write rather than doing it once for each option, but you'd have to check that the kernel allows that for these cgroup files.

This approach was floated in #2115 but concerns were raised about the memory footprint this might have. The other approach I was exploring was just opening the files once and then writing each entry to the same file descriptor, as mentioned.

I’ll do some reading into the kernel docs to see which of these two approaches would be preferable and update the PR accordingly.

cyphar:
Yeah I saw @kolyshkin's comment after I posted here. :smile_cat: I don't really mind doing it either way, though I think the memory overhead will be minimal and not really worth worrying about -- the configuration is originally coming from a JSON blob we read into memory anyway.

mattbonnell:
> Yeah I saw @kolyshkin's comment after I posted here. 😸 I don't really mind doing it either way, though I think the memory overhead will be minimal and not really worth worrying about -- the configuration is originally coming from a JSON blob we read into memory anyway.

I'm trying to figure out what the expected format for the interface files should be. [These docs](https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/cgroup-v1/blkio-controller.rst) would seem to indicate that the entries should be newline-separated. If this is the case, then if the code is updated so that all the entries were written into the interface files at once (either by concatenating them and writing once or writing each to file while it's held open), tests like `TestBlkioSetMultipleWeightDevice` would need to be modified to check for the presence of both entries, separated by a newline, correct?

cyphar:
Looking at the implementation of `bfq_io_set_weight` I believe it only supports one entry at a time, so it's probably a better idea to just do multiple writes.

mattbonnell:
> Looking at the implementation of `bfq_io_set_weight` I believe it only supports one entry at a time, so it's probably a better idea to just do multiple writes.

Yeah, seems so. I think the original issue can thus be closed.

PR #2569, opened 2020-08-28
Use containers/common for seccomp handling

### LINKED ISSUE #2565: Make seccomp code re-usable
Hey runc maintainers :wave:, I really would like to re-use the runc seccomp profile validation in `libcontainer/seccomp/seccomp_linux.go`, but without loading the profile at all. This could be re-used for higher level profile validation in other libraries.

My proposal would be that we move most of the code into a different library or make it public in libcontainer. Both approaches would mean a refactoring of the API in any case. I'm thinking about two new functions:

- `ValidateSeccompProfile(profile string)` - unmarshal the profile into the `Seccomp` struct and try `BuildFilter()`
- `BuildFilter(profile *Seccomp)` - validates the profile further and builds the filter

`InitSeccomp(*Seccomp)` would then also use `BuildFilter()` but additionally runs `filter.Load()`.

What do you think about this? I think this way we could also work on better testing around the seccomp profile validation.

saschagrunert:
This change introduces the new dependency `containers/common`, which
mainly conains the logic from libcontainer/seccomp. We're also now able
to completely remove the internal `Seccomp` type and only rely on the
runtime-spec. The only visible API is now internal only (see
`internal/seccomp`).

Fixes #2565 

PTAL @rhatdan @mrunalp @AkihiroSuda @kolyshkin


saschagrunert:
cc @cyphar

go.sum
```diff
@@ -26,60 +87,243 @@ github.com/golang/protobuf v1.4.0-rc.4.0.20200313231945-b860323f09d0/go.mod h1:W
...
+gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v2 v2.2.4/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v2 v2.2.8/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v2 v2.3.0/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c h1:dUUwHk2QECo/6vqA44rthZ8ie2QXMNeKRTHCNY2nXvo=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
+gotest.tools v2.2.0+incompatible/go.mod h1:DsYFclhRJ6vuDpmuTbkuFWG+y2sxOXAzmJt81HFBacw=
+honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
+honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
```
kolyshkin, line 329: Not sure why so much changes in here... have you tried `go mod tidy`?

kolyshkin:
LGTM overall, and kudos for using `internal/`!

go.sum
saschagrunert: Yes, `go mod tidy` does not change anything inside `go.sum`. :shrug:

cyphar:
This would probably need a rewrite given #2750.

cyphar:
I'm also slightly concerned about this PR because it's changing the serialisation format of the `Config` structure -- which is stored as part of the state (and thus this will cause problems with `runc` upgrades -- we made this mistake before with `initProcessTime`).

saschagrunert:
> I'm also slightly concerned about this PR because it's changing the serialisation format of the `Config` structure -- which is stored as part of the state (and thus this will cause problems with `runc` upgrades -- we made this mistake before with `initProcessTime`).

Ah yeah so I guess we cannot remove the internal types in an easy way. I'm closing for now and will revisit later.

cyphar:
Yeah in the long-term we should work on removing the really ugly libcontainer serialisation but unfortunately for now it is needed and changing it has caused some /lovely/ bugs in the past.

PR #2570, opened 2020-08-28
Fix mount error when chmod RO tmpfs

### LINKED ISSUE #2246: chmod on read-only tmpfs
# How to reproduce
```
# docker run --rm -it --tmpfs /mnt:ro alpine
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"tmpfs\\\" to rootfs \\\"/var/lib/docker/overlay2/ee05948a3c3d2007ac920c0e8a7dcfdaed59a85fcaa59ff9eb791f5151dc92bf/merged\\\" at \\\"/mnt\\\" caused \\\"chmod /var/lib/docker/overlay2/ee05948a3c3d2007ac920c0e8a7dcfdaed59a85fcaa59ff9eb791f5151dc92bf/merged/mnt: read-only file system\\\"\"": unknown.
```
Tested on docker 19.03.2 with runc 1.0.1-dev.

# Related code
https://github.com/opencontainers/runc/blob/fc5759cf4fcf3f9c77c5973a24d37188dbcc92ee/libcontainer/rootfs_linux.go#L331-L377

# Use case
I'm using `registry:2` with s3 storage backend, thus `/var/lib/registry` mountpoint is useless for my setup. If I didn't specify a bind mount or tmpfs for this path, an anonymous volume would be created. I don't want to keep an empty folder for bind-mounting to `/var/lib/registry`, nor do I want a writable tmpfs. So I tried `--tmpfs /var/lib/registry:ro`, but failed.

My best effort without read-only tmpfs is `--tmpfs /var/lib/registry:nr_blocks=1,nr_inodes=2`.

EduardoVega:
Based on the comments, this will mount the filesystem rw, then chmod, and finally remount it with ro.

Fixes #2246

Signed-off-by: Eduardo Vega  <edvegavalerio@gmail.com>

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -383,6 +383,12 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 				return err
 			}
 		}
+		// Initially mounted rw in mountPropagate, remount to ro if flag set.
+		if m.Flags&unix.MS_RDONLY != 0 {
+			if err := remount(m, rootfs); err != nil {
+				return err
+			}
+		}
 		return nil
 	case "bind":
 		if err := prepareBindMount(m, rootfs); err != nil {
@@ -954,6 +960,12 @@ func mountPropagate(m *configs.Mount, rootfs string, mountLabel string) error {
 		flags = m.Flags
 	)
 	if libcontainerUtils.CleanPath(dest) == "/dev" {
+		flags &= ^unix.MS_RDONLY
+	}
+
+	// Mount it rw to allow chmod operation. A remount will be performed
+	// later to make it ro if set.
+	if m.Device == "tmpfs" {
 		flags &= ^unix.MS_RDONLY
 	}
 
--- a/tests/integration/mounts.bats
+++ b/tests/integration/mounts.bats
@@ -20,3 +20,12 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
 }
+
+@test "runc run [ro tmpfs mount]" {
+	update_config 	' .mounts += [{"source": "tmpfs", "destination": "/mnt", "type": "tmpfs", "options": ["ro", "nodev", "nosuid", "mode=755"]}] 
+			| .process.args |= ["grep", "^tmpfs /mnt", "/proc/mounts"]' 
+
+	runc run test_ro_tmpfs_mount
+	[ "$status" -eq 0 ]
+	[[ "${lines[0]}" == *'ro,'* ]]
+}


libcontainer/rootfs_linux.go
```diff
@@ -339,13 +339,17 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
...
+		isRo := m.Flags&unix.MS_RDONLY == 1
 		tmpDir := ""
 		stat, err := os.Stat(dest)
 		if err != nil {
 			if err := os.MkdirAll(dest, 0755); err != nil {
 				return err
 			}
 		}
+		if isRo {
+			m.Flags &= ^unix.MS_RDONLY
```
AkihiroSuda, line 351: m shouldn't be altered, please copy the struct

EduardoVega: I have changed a little bit the code. Now I am reusing a remount function that was in the code, so I did not require to copy the struct. Thanks @AkihiroSuda

AkihiroSuda (dismissed):

EduardoVega:
@AkihiroSuda I have squashed the commits, they were not initially.

AkihiroSuda (dismissed):

EduardoVega:
Fixes #2570

AkihiroSuda:
@kolyshkin PTAL

libcontainer/rootfs_linux.go
```diff
@@ -339,6 +339,7 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 		return label.SetFileLabel(dest, mountLabel)
 	case "tmpfs":
 		copyUp := m.Extensions&configs.EXT_COPYUP == configs.EXT_COPYUP
+		isRo := m.Flags&unix.MS_RDONLY == 1
```
kolyshkin, line 342: This check only works correctly because, by accident, the value of `unix.MS_RDONLY` is 1. Not that we will expect this to change, but if someone copy/pastes this code to check for the other flag, it won't work.

```suggestion
		isRo := m.Flags&unix.MS_RDONLY != 0
```

libcontainer/rootfs_linux.go
```diff
@@ -957,6 +963,11 @@ func mountPropagate(m *configs.Mount, rootfs string, mountLabel string) error {
 		flags &= ^unix.MS_RDONLY
 	}
 
+	isRo := m.Flags&unix.MS_RDONLY == 1
```
kolyshkin, line 966: ditto

tests/integration/mounts.bats
```diff
@@ -19,3 +19,12 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
 }
+
+@test "runc run [ro tmpfs mount]" {
+	update_config 	' .mounts += [{"source": "tmpfs", "destination": "/mnt", "type": "tmpfs", "options": ["ro", "nodev", "nosuid", "mode=755"]}] 
+			| .process.args |= ["grep", "/mnt", "/proc/mounts"]'
```
kolyshkin, line 25: Slightly more correct check would be something like
```suggestion
			| .process.args |= ["grep", "^tmpfs /mnt ", "/proc/mounts"]' 
```

tests/integration/mounts.bats
```diff
@@ -19,3 +19,12 @@ function teardown() {
...
 	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
 }
+
+@test "runc run [ro tmpfs mount]" {
+	update_config 	' .mounts += [{"source": "tmpfs", "destination": "/mnt", "type": "tmpfs", "options": ["ro", "nodev", "nosuid", "mode=755"]}] 
+			| .process.args |= ["grep", "/mnt", "/proc/mounts"]' 
+
+	runc run test_ro_tmpfs_mount
+	[ "$status" -eq 0 ]
+	[[ "${lines[0]}" == *'ro'* ]]
```
kolyshkin, line 29: maybe 
```suggestion
	[[ "${lines[0]}" == *'ro,'* ]]
```

libcontainer/rootfs_linux.go
```diff
@@ -957,6 +963,11 @@ func mountPropagate(m *configs.Mount, rootfs string, mountLabel string) error {
...
+	if isRo && m.Device == "tmpfs" {
+		flags &= ^unix.MS_RDONLY
```
kolyshkin, line 968: In fact you don't have to check the flag -- you can just go ahead and clear it.

Makes sense to add a comment explaining why are we doing it here.

libcontainer/rootfs_linux.go
```diff
@@ -383,6 +384,11 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 				return err
 			}
 		}
+		if isRo {
+			if err := remount(m, rootfs); err != nil {
```
kolyshkin, line 388: also makes sense to add a comment, something like
```go
// initially mounted rw in mountPropagate, remount to ro
```
should be sufficient I guess.

kolyshkin (changes requested):
left some suggestions

libcontainer/rootfs_linux.go
cyphar: I also personally don't think a variable is necessary in either case -- you only check the value once and to me `m.Flags&unix.MS_RDONLY != 0` is more obvious to read.

EduardoVega: I have removed the variable, the value will be checked only once before doing the remount.

libcontainer/rootfs_linux.go
EduardoVega: I removed the flag check and added a comment.

libcontainer/rootfs_linux.go
EduardoVega: done

tests/integration/mounts.bats
EduardoVega: done

tests/integration/mounts.bats
EduardoVega: done

EduardoVega:
Thanks. I have completed all requested changes.

kolyshkin (approved):
LGTM

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda (approved):

PR #2574, opened 2020-09-05
Add integration tests for cgroup devices

### LINKED ISSUE #2249: Need to add integration test suite for resource limitation (especially for devices)
https://github.com/opencontainers/runc/blob/master/tests/integration/cgroups.bats

The current test suite doesn't cover resource limitation stuff except `memory.kernel`.

EduardoVega:
Fixes #2249 

Signed-off-by: Eduardo Vega <edvegavalerio@gmail.com>

### DIFF
--- a/tests/integration/dev.bats
+++ b/tests/integration/dev.bats
@@ -33,3 +33,73 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "${lines[0]}" =~ "crw-rw-rw".+"1".+"0".+"0".+"5,".+"2".+"/dev/ptmx" ]]
 }
+
+@test "runc run [device cgroup deny]" {
+	requires root
+
+	update_config ' .linux.resources.devices = [{"allow": false, "access": "rwm"}]
+			| .linux.devices = [{"path": "/dev/kmsg", "type": "c", "major": 1, "minor": 11}]
+			| .process.args |= ["sh"]'
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_deny
+	[ "$status" -eq 0 ]
+
+	# test write
+	runc exec test_deny sh -c 'hostname | tee /dev/kmsg'
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *'Operation not permitted'* ]]
+
+	# test read
+	runc exec test_deny sh -c 'head -n 1 /dev/kmsg'
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *'Operation not permitted'* ]]
+}
+
+@test "runc run [device cgroup allow rw char device]" {
+	requires root
+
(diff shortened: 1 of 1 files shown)


tests/integration/dev.bats
```diff
@@ -33,3 +33,65 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "${lines[0]}" =~ "crw-rw-rw".+"1".+"root".+"root".+"5,".+"2".+"/dev/ptmx" ]]
 }
+
+@test "runc run [device cgroup no rules]" {
+	[[ "$ROOTLESS" -ne 0 ]] && requires root
+
```
AkihiroSuda, line 39: indent

AkihiroSuda:
Please squash commits

tests/integration/dev.bats
```diff
@@ -33,3 +33,46 @@ function teardown() {
...
+	# test read
+	runc exec test_allow sh -c 'head -n 1 /dev/kmsg'
+	[ "$status" -eq 0 ]
+}
+
+@test "runc run [device cgroup deny]" {
+	[[ "$ROOTLESS" -ne 0 ]] && requires root
+
+	update_config ' .linux.resources.devices = [{"allow": false, "access": "rwm"}]
+			| .linux.devices = [{"path": "/dev/kmsg", "type": "c", "major": 1, "minor": 11}]
```
AkihiroSuda, line 62: Can we have a test for block devices as well? (I think the previous revision had block device tests?)

EduardoVega: sure, let me add that back. 

When working on this, I saw some errors and I got questions about it
1. would you think that is a common scenario to add devices in a different path that the one in the host? for example host: `/dev/null` and in the container: `/dev/mynull`. Doing this in a container with user namespace, I would get this error `rootfs_linux.go:72: creating device nodes caused: no such file or directory`. If I am not wrong it happens because they are bind mounted usign the `device.Path` instead of created with mknod.

2. and also, is a common scenario to add specific deny rules besides deny all? when RUNC_USE_SYSTEMD is set to yes and I add something like `{"allow": true, "access": "rwm"}, {"allow": false, "access": "rw", "type": "c", "major": 1, "minor": 11}` I got `systemd doesn't support blacklist device rules -- applying temporary deny-all rule` and then I go this error `container init caused: rootfs_linux.go:123: reopening /dev/null inside container caused: Failed to open /dev/null - open /dev/null: operation not permitted`

thanks

AkihiroSuda: You can just skip the failing tests (with code comments) for userns and systemd

EduardoVega: Added one more integration test that tests the read and mknod operations for a block device

AkihiroSuda (approved):

AkihiroSuda:
@kolyshkin @cyphar PTAL

kolyshkin (approved):
LGTM

PR #2575, opened 2020-09-07
libcontainer: Store state.json before sync procRun

fuweid:
If the procRun state has been synced and the runc-create process has
been killed for some reason, the runc-init[2:stage] process will be
leaky. And the runc command also fails to parse root directory because
the container doesn't have state.json.

In order to make it possible to clean the leaky runc-init[2:stage]
process , we should store the status before sync procRun.

```before
current workflow:

[  child  ] <-> [   parent   ]

procHooks   --> [run hooks]
            <-- procResume

procReady   --> [final setup]
            <-- procRun

                ( killed for some reason)
                ( store state.json )
```

```expected
expected workflow:

[  child  ] <-> [   parent   ]

procHooks   --> [run hooks]
            <-- procResume

procReady   --> [final setup]
                store state.json
            <-- procRun
```

Signed-off-by: Wei Fu <fuweid89@gmail.com>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -369,18 +369,8 @@ func (c *linuxContainer) start(process *Process) error {
 		}
 		return newSystemErrorWithCause(err, "starting container process")
 	}
-	// generate a timestamp indicating when the container was started
-	c.created = time.Now().UTC()
-	if process.Init {
-		c.state = &createdState{
-			c: c,
-		}
-		state, err := c.updateState(parent)
-		if err != nil {
-			return err
-		}
-		c.initProcessStartTime = state.InitProcessStartTime
 
+	if process.Init {
 		if c.config.Hooks != nil {
 			s, err := c.currentOCIState()
 			if err != nil {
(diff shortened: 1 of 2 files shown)


fuweid:
ping @AkihiroSuda @kolyshkin PTAL, thanks!

fuweid:
ping @opencontainers/runc-maintainers PTAL, thanks~

kolyshkin (approved):
LGTM. Wonder what was the reason of runc-create's premature demise that you saw?

fuweid:
> LGTM. Wonder what was the reason of runc-create's premature demise that you saw?

@kolyshkin I didn't see what happened at the comment. We use kubernetes and containerd. The CRI StartContainer returned time out error and runc-init process was leaky and tried to open exec.fifo. 

It is hard to reproduce this issue. I meet few cases per few days in my production. When this case shows up, we see the system reclaims many more memory or OOM for the container during runc-create. But didn't find the root cause....

This patch is used to prevent runc-init leaky. I would like to share the root cause if I can get the clue from log monitor. :)

fuweid:
ping @AkihiroSuda ~

AkihiroSuda (approved):

PR #2580, opened 2020-09-09
tests/int: fix wrt busybox's nogroup->nobody change

kolyshkin:
Apparently, newer busybox image changed "nogroup" group name to "nobody",
breaking a few test cases that depended on the name.

Fix those tests by switching to numeric IDs.

### 1. tests/int: fix runc exec --preserve-fds

This test started to fail, as it expected the output from
`id` to be

	uid=1000 gid=1000 groups=100(users),65534(nogroup)

while the actual output is now

	uid=1000 gid=1000 groups=100(users),65534(nobody)

Apparently, busybox image changed group name.

As we're only interested in ids, not names, and to fix the test,
let's use `id -G`.

### 2. tests/int/dev.bats: fixes for new busybox

These tests expect group name to be "nogroup", while recent busybox
changed that to "nobody".

Use numeric uids/gids to fix.

### DIFF
--- a/tests/integration/dev.bats
+++ b/tests/integration/dev.bats
@@ -13,23 +13,23 @@ function teardown() {
 
 @test "runc run [redundant default /dev/tty]" {
 	update_config ' .linux.devices += [{"path": "/dev/tty", "type": "c", "major": 5, "minor": 0}]
-		      | .process.args |= ["ls", "-lL", "/dev/tty"]'
+		      | .process.args |= ["ls", "-lLn", "/dev/tty"]'
 
 	runc run test_dev
 	[ "$status" -eq 0 ]
 
 	if [[ "$ROOTLESS" -ne 0 ]]; then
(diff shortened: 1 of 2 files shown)


kolyshkin:
@AkihiroSuda @mrunalp PTAL

kolyshkin:
Found while looking at CI for https://github.com/opencontainers/runc/pull/2580

mrunalp (dismissed):

mrunalp:
@AkihiroSuda ptal

kolyshkin:
Another failure caused by the same change:

```
not ok 28 runc run [redundant default /dev/tty]
# (in test file tests/integration/dev.bats, line 22)
#   `[[ "${lines[0]}" =~ "crw-rw-rw".+"1".+"nobody".+"nogroup".+"5,".+"0".+"/dev/tty" ]]' failed
....
# crw-rw-rw-    1 nobody   nobody      5,   0 Sep  9 01:06 /dev/tty
```

AkihiroSuda (approved):

mrunalp (approved):

chuanchang:
I also hit the same issue, it's nobody not nogroup on RHEL-8.y.

PR #2583, opened 2020-09-10
restore: tell CRIU to use existing namespaces

adrianreber:
runc already tells CRIU to restore into an existing network or PID namespace if there is a path to a namespace specified in config.json.

PID and network have special handling in CRIU using CRIU's inherit_fd interface.

For UTS, IPC and MOUNT namespaces CRIU can join those existing namespaces using CRIU's join_ns interface.

This is especially interesting for environments where containers are running in a pod which already has running containers (pause for example) with namespaces configured and the restored container needs to join these namespaces.

CRIU has no support to join an existing CGROUP namespace (yet?) why restoring a container with a path specified to a CGROUP namespace will be aborted by runc.

CRIU would have support to restore a container into an existing time namespace, but runc does not yet seem to support time namespaces.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -879,6 +879,44 @@ func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.
 	return nil
 }
 
+func (c *linuxContainer) handleRestoringNamespaces(rpcOpts *criurpc.CriuOpts, extraFiles *[]*os.File) error {
+	for _, ns := range c.config.Namespaces {
+		switch ns.Type {
+		case configs.NEWNET, configs.NEWPID:
+			// If the container is running in a network or PID namespace and has
+			// a path to the network or PID namespace configured, we will dump
+			// that network or PID namespace as an external namespace and we
+			// will expect that the namespace exists during restore.
+			// This basically means that CRIU will ignore the namespace
+			// and expect it to be setup correctly.
+			if err := c.handleRestoringExternalNamespaces(rpcOpts, extraFiles, ns.Type); err != nil {
+				return err
+			}
+		default:
+			// For all other namespaces except NET and PID CRIU has
+			// a simpler way of joining the existing namespace if set
+			nsPath := c.config.Namespaces.PathOf(ns.Type)
+			if nsPath == "" {
+				continue
+			}
+			if ns.Type == configs.NEWCGROUP {
+				// CRIU has no code to handle NEWCGROUP
+				return fmt.Errorf("Do not know how to handle namespace %v", ns.Type)
+			}
+			// CRIU has code to handle NEWTIME, but it does not seem to be defined in runc
+
+			// CRIU will issue a warning for NEWUSER:
+			// criu/namespaces.c: 'join-ns with user-namespace is not fully tested and dangerous'
+			rpcOpts.JoinNs = append(rpcOpts.JoinNs, &criurpc.JoinNamespace{
+				Ns:     proto.String(configs.NsName(ns.Type)),
+				NsFile: proto.String(nsPath),
+			})
+		}
+	}
+
+	return nil
+}
+
 func (c *linuxContainer) handleRestoringExternalNamespaces(rpcOpts *criurpc.CriuOpts, extraFiles *[]*os.File, t configs.NamespaceType) error {
 	if !c.criuSupportsExtNS(t) {
 		return nil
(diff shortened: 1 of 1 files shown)


h-vetinari:
Xref https://github.com/opencontainers/runtime-spec/pull/1062

kolyshkin:
CI complains:

> script/validate-gofmt
> These files are not properly gofmt'd:
>  - libcontainer/container_linux.go
> 
> Please reformat the above files using "gofmt -s -w" and commit the result.

libcontainer/container_linux.go
```diff
@@ -879,6 +879,48 @@ func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.
...
+		switch ns.Type {
+		case configs.NEWNET:
+			// If the container is running in a network namespace and has
+			// a path to the network namespace configured, we will dump
+			// that network namespace as an external namespace and we
+			// will expect that the namespace exists during restore.
+			// This basically means that CRIU will ignore the namespace
+			// and expect to be setup correctly.
+			fallthrough
+		case configs.NEWPID:
```
kolyshkin, line 893: Maybe this instead?
```go
case configs.NEWNET, configs.NEWPID:
...
```

easier to read than `fallthrough`

libcontainer/container_linux.go
```diff
@@ -879,6 +879,48 @@ func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.
...
+			// Same for possible external PID namespaces
+			if err := c.handleRestoringExternalNamespaces(rpcOpts, extraFiles, ns.Type); err != nil {
+				return err
+			}
+		case configs.NEWIPC:
+			fallthrough
+		case configs.NEWUTS:
+			fallthrough
+		case configs.NEWNS:
```
kolyshkin, line 902: ditto

libcontainer/container_linux.go
```diff
@@ -879,6 +879,48 @@ func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.
...
+			// a simpler way of joining the existing namespace if set
+			nsPath := c.config.Namespaces.PathOf(ns.Type)
+			if nsPath == "" {
+				return nil
+			}
+			joinNamespace := &criurpc.JoinNamespace{
+				Ns: proto.String(configs.NsName(ns.Type)),
+				NsFile: proto.String(nsPath),
+			}
+			rpcOpts.JoinNs = append(rpcOpts.JoinNs, joinNamespace)
```
kolyshkin, line 913: Perhaps
```go
rpcOpts.JoinNs = append(rpcOpts.JoinNs, &criurpc.JoinNamespace{
￼				Ns: proto.String(configs.NsName(ns.Type)),
￼				NsFile: proto.String(nsPath),
￼			})
```

adrianreber:
@kolyshkin Thanks for the review. I changed the code and also used `gofmt` (I always forget that).

libcontainer/container_linux.go
```diff
@@ -879,6 +879,40 @@ func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.
...
+			if nsPath == "" {
+				return nil
+			}
+			rpcOpts.JoinNs = append(rpcOpts.JoinNs, &criurpc.JoinNamespace{
+				Ns:     proto.String(configs.NsName(ns.Type)),
+				NsFile: proto.String(nsPath),
+			})
+		default:
+			// CRIU has no code to handle NEWCGROUP
+			// CRIU has code to handle NEWTIME, but it does not seem to be defined un runc
```
rst0git, line 908: *in runc

adrianreber: thanks. fixed.

kolyshkin (approved):
LGTM

kolyshkin (changes requested):
Wait a minute...

libcontainer/container_linux.go
```diff
@@ -879,6 +879,40 @@ func (c *linuxContainer) handleCheckpointingExternalNamespaces(rpcOpts *criurpc.
...
+			// and expect it to be setup correctly.
+			if err := c.handleRestoringExternalNamespaces(rpcOpts, extraFiles, ns.Type); err != nil {
+				return err
+			}
+		case configs.NEWIPC, configs.NEWUTS, configs.NEWNS:
+			// For all other namespaces except NET and PID CRIU has
+			// a simpler way of joining the existing namespace if set
+			nsPath := c.config.Namespaces.PathOf(ns.Type)
+			if nsPath == "" {
+				return nil
```
kolyshkin, line 900: Hmm... Would you still like to handle the rest of namespaces?

adrianreber: What do you exactly mean?

adrianreber: Ah, I get it. I replaced `return nil` with `continue`

adrianreber:
With the changes tests are failing... Because the code is just wrong. Will fix it.

adrianreber:
Now it works as intended.

kolyshkin (approved):
LGTM

rst0git (approved):

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda (approved):

Issue #2563, opened 2020-08-20
Support native cgroupv2 spec


kolyshkin:
As https://github.com/opencontainers/runtime-spec/pull/1040 is merged, we should start working on implementing it in runc.

Relevant part of spec: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#unified

crun PR: containers/crun#459

kolyshkin:
crun implementation: https://github.com/containers/crun/pull/459

PR #2584, opened 2020-09-11
libct/cgroups: support Cgroups.Resources.Unified

kolyshkin:
Add support for unified resource map (as per [1]), and add some test
cases for the new functionality.
    
[1] https://github.com/opencontainers/runtime-spec/pull/1040

Fixes: #2563

The biggest issue, I think, is that the mapping of the "unified" parameters
to systemd properties is not implemented, meaning there can be cases
where systemd unit settings and in-kernel cgroup settings won't agree.

I don't think it can cause any issues, but it's just doesn't look right.

### DIFF
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -3,11 +3,14 @@
 package fs2
 
 import (
+	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/pkg/errors"
 )
@@ -206,7 +209,37 @@ func (m *manager) Set(container *configs.Config) error {
 	if err := setFreezer(m.dirPath, container.Cgroups.Freezer); err != nil {
 		return err
 	}
+	if err := m.setUnified(container.Cgroups.Unified); err != nil {
+		return err
+	}
 	m.config = container.Cgroups
+	return nil
+}
+
+func (m *manager) setUnified(res map[string]string) error {
+	for k, v := range res {
+		if strings.Contains(k, "/") {
+			return fmt.Errorf("unified resource %q must be a file name (no slashes)", k)
+		}
+		if err := fscommon.WriteFile(m.dirPath, k, v); err != nil {
+			errC := errors.Cause(err)
+			// Check for both EPERM and ENOENT since O_CREAT is used by WriteFile.
+			if errors.Is(errC, os.ErrPermission) || errors.Is(errC, os.ErrNotExist) {
+				// Check if a controller is available,
+				// to give more specific error if not.
+				sk := strings.SplitN(k, ".", 2)
+				if len(sk) != 2 {
+					return fmt.Errorf("unified resource %q must be in the form CONTROLLER.PARAMETER", k)
+				}
+				c := sk[0]
+				if _, ok := m.controllers[c]; !ok && c != "cgroup" {
+					return fmt.Errorf("unified resource %q can't be set: controller %q not available", k, c)
+				}
+			}
+			return errors.Wrapf(err, "can't set unified resource %q", k)
+		}
+	}
+
 	return nil
 }
 
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -705,6 +705,155 @@ func testRunWithKernelMemory(t *testing.T, systemd bool) {
 	}
 }
 
+func TestCgroupResourcesUnifiedErrorOnV1(t *testing.T) {
+	testCgroupResourcesUnifiedErrorOnV1(t, false)
+}
+
+func TestCgroupResourcesUnifiedErrorOnV1Systemd(t *testing.T) {
+	if !systemd.IsRunningSystemd() {
+		t.Skip("Systemd is unsupported")
+	}
+	testCgroupResourcesUnifiedErrorOnV1(t, true)
+}
+
+func testCgroupResourcesUnifiedErrorOnV1(t *testing.T, systemd bool) {
+	if testing.Short() {
+		return
+	}
+	if cgroups.IsCgroup2UnifiedMode() {
+		t.Skip("requires cgroup v1")
+	}
+	rootfs, err := newRootfs()
+	ok(t, err)
+	defer remove(rootfs)
+
+	config := newTemplateConfig(rootfs)
+	if systemd {
+		config.Cgroups.Parent = "system.slice"
+	}
+	config.Cgroups.Resources.Unified = map[string]string{
+		"memory.min": "10240",
+	}
+	_, _, err = runContainer(config, "", "true")
+	if !strings.Contains(err.Error(), cgroups.ErrV1NoUnified.Error()) {
+		t.Fatalf("expected error to contain %v, got %v", cgroups.ErrV1NoUnified, err)
+	}
+}
+
+func TestCgroupResourcesUnified(t *testing.T) {
+	testCgroupResourcesUnified(t, false)
+}
+
+func TestCgroupResourcesUnifiedSystemd(t *testing.T) {
+	if !systemd.IsRunningSystemd() {
+		t.Skip("Systemd is unsupported")
+	}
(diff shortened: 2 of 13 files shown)


kolyshkin:
That intermittent download failure, again
```
 ---> b2ec3d2c2048
Step 13/23 : RUN curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64     && chmod +x /usr/local/bin/umoci
 ---> Running in b2823a13a789
curl: (22) The requested URL returned error: 429 too many requests
The command '/bin/sh -c curl -o /usr/local/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.5/umoci.amd64     && chmod +x /usr/local/bin/umoci' returned a non-zero code: 22
Makefile:62: recipe for target 'runcimage' failed
```

Wonder if there's a better alternative. For some reason, this never happens to any other URLs downloaded from github

kolyshkin:
Rebased

libcontainer/integration/exec_test.go
```diff
@@ -705,6 +705,150 @@ func testRunWithKernelMemory(t *testing.T, systemd bool) {
...
+}
+
+func TestCgroupResourcesUnifiedSystemd(t *testing.T) {
+	if !systemd.IsRunningSystemd() {
+		t.Skip("Systemd is unsupported")
+	}
+	testCgroupResourcesUnified(t, true)
+}
+
+func testCgroupResourcesUnified(t *testing.T, systemd bool) {
```
AkihiroSuda, line 754: We should also have integration tests

kolyshkin: Sure, will do

kolyshkin: Added a few basic tests. Let me know if you had something else in mind.

kolyshkin:
@giuseppe @mrunalp PTAL

tests/integration/cgroups.bats
```diff
@@ -176,3 +178,37 @@ function setup() {
...
+}
+
+@test "runc run (cgroup v2 + unified resources)" {
+    requires root cgroups_v2
+
+    set_cgroups_path "$BUSYBOX_BUNDLE"
+    update_config ' .linux.resources.memory |= {"limit": 33554432}
+                  | .linux.resources.memorySwap |= {"limit": 33554432}
+                  | .linux.resources.unified |=
+                      {"memory.min": "131072", "memory.max": "10485760" }' \
```
AkihiroSuda, line 201: We should also test:
* `pids.max` (which doesn't conflict v1-compat config)
* `non.existent.value`
* `../../../etc/shadow`

tests/integration/cgroups.bats
```diff
@@ -176,3 +178,37 @@ function setup() {
...
+    set_cgroups_path "$BUSYBOX_BUNDLE"
+    set_resources_limit "$BUSYBOX_BUNDLE"
+    update_config '.linux.resources.unified |= {"memory.min": "131072"}' "$BUSYBOX_BUNDLE"
+
+    runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroups_unified
+    [ "$status" -ne 0 ]
+    [[ "$output" == *'invalid configuration'* ]]
+}
+
+@test "runc run (cgroup v2 + unified resources)" {
```
AkihiroSuda, line 194: Can we also have test for pure-unified config

tests/integration/cgroups.bats
kolyshkin: I am already testing the last two in the libcontainer/integration tests. Not sure how adding these cases to bats will add any value.

Not sure what you mean by "v1-compat config". The runtime-spec says 'resources.unified' can't be used on v1, and there's a code to enforce it (checking if "unified" is set, doesn't look into individual parameters). Do you mean you want to test this works if we set "pids.max"?

AkihiroSuda: All the current fields in `.linux.resources.unified` overrides the v1 config (`.linux.resources.{memory,memorySwap}`).

We should ensure that we can set unified fields that do not override v1.

kolyshkin: There is a code that rejects config if `.linux.resources.unified` set for cgroupv1, here:
https://github.com/opencontainers/runc/pull/2584/commits/b14556f2678eb12cbacdc12ae3ddf6dc952e3fbe#diff-252ac01f71681e9ed8e7ce0571cdb6acR349-R351
and here:
https://github.com/opencontainers/runc/pull/2584/commits/b14556f2678eb12cbacdc12ae3ddf6dc952e3fbe#diff-bf0fd9d73921163a3f447b24a8187ee7

I check that this code works in two tests, here:
https://github.com/opencontainers/runc/pull/2584/commits/b14556f2678eb12cbacdc12ae3ddf6dc952e3fbe#diff-bd87cca3665e082adb4ba55ee22e82b5
and here:
https://github.com/opencontainers/runc/pull/2584/commits/d0397b925803e3cc78e79ec7ef1da622529595c1#diff-1be821a9fd1920ed1cd1090d8a6bea39R182

The rejection code does not look into what's set under the `unified`, so it does not make sense to add some specific parameters in there -- anything is rejected.

tests/integration/cgroups.bats
kolyshkin: will do

kolyshkin: added; PTAL

kolyshkin:
Added bats test with more unified parameters (using only `resources.unified`), and libct/int test case using read-only file (`pids.current`) as a key to check for error.

AkihiroSuda (dismissed):

AkihiroSuda:
@giuseppe PTAL?

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -206,10 +209,42 @@ func (m *manager) Set(container *configs.Config) error {
...
 }
 
+func (m *manager) setUnified(res map[string]string) error {
+	for k, v := range res {
+		if strings.Contains(k, "/") {
+			return fmt.Errorf("unified resource %q must be a file name (no slashes)", k)
+		}
+		if err := fscommon.WriteFile(m.dirPath, k, v); err != nil {
+			errC := errors.Cause(err)
+			// cgroup2 returns EPERM (rather than ENOENT) for
```
giuseppe, line 226: is EPERM because the file is opened with `O_CREAT`?

giuseppe (commented):
just a question, otherwise LGTM

libcontainer/cgroups/fs2/fs2.go
kolyshkin: Right. `fscommon.WriteFile` use `ioutil.WriteFile` which calls `os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)`.

Let me fix the comment.

Wonder if we should not use `O_CREAT` as it makes no sense for cgroupfs.

kolyshkin: fixed

giuseppe:
LGTM

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -206,10 +209,41 @@ func (m *manager) Set(container *configs.Config) error {
...
+					return fmt.Errorf("unified resource %q must be in the form CONTROLLER.PARAMETER", k)
+				}
+				c := sk[0]
+				if _, ok := m.controllers[c]; !ok && c != "cgroup" {
+					return fmt.Errorf("unified resource %q can't be set: controller %q not available", k, c)
+				}
+			}
+			return errors.Wrapf(err, "can't set unified resource %q", k)
+		}
+
```
mrunalp, line 241: nit: space

mrunalp (approved):
We need follow up for systemd but this looks like a good start. Thanks!

libcontainer/cgroups/fs2/fs2.go
kolyshkin: fixed

mrunalp (approved):

mrunalp:
@AkihiroSuda ptal

AkihiroSuda (approved):

AkihiroSuda:
@kolyshkin Could you open an issue for tracking systemd issue?

kolyshkin:
> Could you open an issue for tracking systemd issue?

* https://github.com/opencontainers/runc/issues/2602
* https://github.com/opencontainers/runc/issues/2603

I'm not quite sure which way to go -- let's discuss in there

Issue #2586, opened 2020-09-15
fifo blocks forever when parent runc process execs and then exits


robbmanes:
Related to [Red Hat Bugzilla #1872868](https://bugzilla.redhat.com/show_bug.cgi?id=1872868).

In the rare case that the parent runc process that rexecs `runc init` is killed or exits early, we are left blocking waiting for a fifo pipe that will never appear.  This leaves the `runc init` process forever waiting, and can hold container resources open indefinitely (such as the volumes and cgroups that have already been set aside for this process).

To confirm this, a coredump of the `runc init` process shows we're stuck waiting for an `open` syscall of the other side of a fifo:

```go
(dlv) bt
 0  0x000055b18499334a in syscall.Syscall6
    at /usr/lib/golang/src/syscall/asm_linux_amd64.s:53
 1  0x000055b184a09b2d in github.com/opencontainers/runc/vendor/golang.org/x/sys/unix.openat
    at /usr/src/debug/runc-1.0.0-67.rc10.rhaos4.3.el8.x86_64/GOPATH/src/github.com/opencontainers/runc/GOPATH/src/github.com/opencontainers/runc/vendor/golang.org/x/sys/unix/zsyscall_linux_amd64.go:89
 2  0x000055b184bd4821 in github.com/opencontainers/runc/vendor/golang.org/x/sys/unix.Open
    at /usr/src/debug/runc-1.0.0-67.rc10.rhaos4.3.el8.x86_64/GOPATH/src/github.com/opencontainers/runc/GOPATH/src/github.com/opencontainers/runc/vendor/golang.org/x/sys/unix/syscall_linux.go:150
 3  0x000055b184bd4821 in github.com/opencontainers/runc/libcontainer.(*linuxStandardInit).Init
    at /usr/src/debug/runc-1.0.0-67.rc10.rhaos4.3.el8.x86_64/GOPATH/src/github.com/opencontainers/runc/GOPATH/src/github.com/opencontainers/runc/libcontainer/standard_init_linux.go:188
 4  0x000055b184bc2ad6 in github.com/opencontainers/runc/libcontainer.(*LinuxFactory).StartInitialization
    at /usr/src/debug/runc-1.0.0-67.rc10.rhaos4.3.el8.x86_64/GOPATH/src/github.com/opencontainers/runc/GOPATH/src/github.com/opencontainers/runc/libcontainer/factory_linux.go:380
 5  0x000055b184c1b0bd in main.glob..func6
[...]

PR #2587, opened 2020-09-15
Protect against blocked fifo when parent exits early

robbmanes:
In the rare case the parent runc process that execs the "runc init"
exits early for any reason, make sure we do not block on opening
the fifo pipe forever.

Fixes https://github.com/opencontainers/runc/issues/2586

Signed-off-by: Robb Manes <robbmanes@protonmail.com>

### DIFF
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"time"
 
 	"github.com/opencontainers/runc/libcontainer/apparmor"
 	"github.com/opencontainers/runc/libcontainer/configs"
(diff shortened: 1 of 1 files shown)


cyphar:
NACK. `runc` has two separate stages -- "create" and "start". The FIFO is used to block the container from starting after `create` is done until `start` is called. The intention is to allow arbitrary user scripts to modify the container state before the container process starts. Adding a 10-second timeout would break this.

A nicer solution might be to have some way for the init process to indicate that it is done with setup and that the container should start waiting for the `start` signal (otherwise, exit).

robbmanes:
Understood, thanks for the review and explanation.

kolyshkin:
From `fifo(7)`:

>      When a process tries to write to a FIFO that is not opened for read
>       on the other side, the process is sent a SIGPIPE signal.

The default SIGPIPE handler aborts the program, and it doesn't look like it's changed, so apparently we're not getting SIGPIPE. Is this because we set up a custom handler?

Looking...

cyphar:
@kolyshkin We don't try to write to the FIFO -- the end-stage of `runc init` is trying to open the FIFO with `O_RDONLY`. This blocks until someone else opens it for writing. The issue is (as far as I understand) that the other half of `runc init` dies or has a non-zero exit code, leading to the process sticking around.

@robbmanes When this happens, does the container show up in `runc list`? Can you `runc kill` it?

robbmanes:
> @robbmanes When this happens, does the container show up in `runc list`? Can you `runc kill` it?

Yes on both counts; it appears in `runc list` and is killable with `runc kill`.  That's how we currently are working around this at the moment; scripts that check for long-lived `runc init` processes and kills them with `runc kill`.

Also noting we don't know exactly the scenario where the parent is removed early or why in this particular instance, but I don't think it necessarily matters for this scenario we're describing.  Thanks!

robbmanes:
This was explained to me a bit more by a peer.

So in this case, it appears the container engine in question is *likely* failing to issue `runc start` entirely (or it is prematurely exiting); specifically, we're using the `cri-o` runtime, so https://github.com/cri-o/cri-o/pull/4198 resolves a problem in `cri-o` where upon container creation failure we appropriately end the `runc init` processes as we do not issue `runc start` to them (specifically commit https://github.com/cri-o/cri-o/pull/4198/commits/b8dc2878a608a26145101d91678928ea742514ee).

I am not sure there is a way to make `runc init` more resilient here then; it is blocking on the pipe, yes, but the container engine in question isn't appropriately issuing start and just lets it linger, so would we call that a misuse/error in use of `runc`?  I'd assume yes, so I am not sure there is anything to fix here.  Feedback on this appreciated, and sorry for the bad PR here in either scenario.  Thanks!

PR #2590, opened 2020-09-16
allow bind mount w/o explicit "bind" opt but w/ explicit "bind" type

AkihiroSuda:
Previously, `{"type":"bind"}` without `{"options": ["bind"]}` was failing with ENODEV.

See https://github.com/containers/podman/issues/7652

### DIFF
--- a/tests/integration/mounts.bats
+++ b/tests/integration/mounts.bats
@@ -19,3 +19,12 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
 }
+
+@test "runc run [bind mount w/o explicit \"bind\" opt but w/ explicit \"bind\" type]" {
+	update_config 	' .mounts += [{"source": ".", "destination": "/tmp/bind", "type":"bind"}]
+			| .process.args |= ["ls", "/tmp/bind/config.json"]'
+
+	runc run test_bind_mount
+	[ "$status" -eq 0 ]
+	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
+}
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -329,6 +329,9 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 func createLibcontainerMount(cwd string, m specs.Mount) *configs.Mount {
 	flags, pgflags, data, ext := parseMountOptions(m.Options)
+	if m.Type == "bind" {
+		flags |= unix.MS_BIND
+	}
 	source := m.Source
 	device := m.Type
 	if flags&unix.MS_BIND != 0 {


tests/integration/mounts.bats
```diff
@@ -19,3 +19,12 @@ function teardown() {
...
 	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
 }
+
+@test "runc run [bind mount w/o explicit \"bind\" opt but w/ explicit \"bind\" type]" {
+	update_config 	' .mounts += [{"source": ".", "destination": "/tmp/bind", "type":"bind"}]
+			| .process.args |= ["ls", "/tmp/bind/config.json"]'
+
+	runc run test_bind_mount
+	[ "$status" -eq 0 ]
+	[[ "${lines[0]}" == *'/tmp/bind/config.json'* ]]
```
kolyshkin, line 29: this will also match something like "ls: /tmp/bind/config.json: no such file or directory" which is not good.

I think we should list the directory instead, and check the file is present in the output.

kolyshkin (changes requested):
test case needs fixing

tests/integration/mounts.bats
AkihiroSuda: We check the status code, so it won't match `ls: /tmp/bind/config.json: no such file or directory`

AkihiroSuda:
The test should be ok, please see https://github.com/opencontainers/runc/pull/2590#discussion_r489968559

tests/integration/mounts.bats
kolyshkin: well in this case it does not make sense to check the output, since `ls` will definitely return the error if the file is not there, so this check is redundant

kolyshkin: So I guess either
 - list the directory, check the exit code, check the output to contain file name
or
 - list the file, check the exit code

cyphar:
I would argue the config is wrong -- `"type": "bind"` doesn't mean anything in Linux (all that matters is the existence of `MS_BIND` in the mount options) -- though in fairness this was a bug in runc as well (#1753). Is there a reason that the config generator is not adding the `bind` option to mounts?

I do see why we'd want to add it, but it means that a hypothetical filesystem called `bind` couldn't be mounted with this change.

AkihiroSuda:
> Is there a reason that the config generator is not adding the bind option to mounts?
> a hypothetical filesystem called bind couldn't be mounted with this change.

This sounds too much hypothetical

cyphar:
That means crun is not implementing the runtime-spec correctly (or rather, is implementing an extension to the runtime-spec). To [quote the spec](https://github.com/opencontainers/runtime-spec/blob/master/config.md#mounts):

> A mount is a bind mount if it has either bind or rbind in the options.

I'm not against adding something as simple as this workaround, but just because crun (or runc for that matter) does something doesn't make it the correct behaviour.

giuseppe:
> As crun doesn't require this option, projects using crun as the primary runtime are likely to forget adding this option. [containers/podman#7652](https://github.com/containers/podman/issues/7652)

I'll fix it in crun to follow more closely the OCI specs.

PR #2591, opened 2020-09-16
vendor:  github.com/moby/sys/mountinfo v0.2.0

thaJeztah:
full diff: https://github.com/moby/sys/compare/mountinfo/v0.1.3...mountinfo/v0.2.0

Bug fixes:

- Fix path unescaping for paths with double quotes

Improvements:

- Mounted: speed up by adding fast paths using openat2 (Linux-only) and stat
- Mounted: relax path requirements (allow relative, non-cleaned paths, symlinks)
- Unescape fstype and source fields
- Documentation improvements

Testing/CI:

- Unit tests: exclude darwin
- CI: run tests under Fedora 32 to test openat2
- TestGetMounts: fix for Ubuntu build system
- Makefile: fix ignoring test failures
- CI: add cross build


go.mod
```diff
@@ -22,5 +22,5 @@ require (
 	// NOTE: urfave/cli must be <= v1.22.1 due to a regression: https://github.com/urfave/cli/issues/1092
 	github.com/urfave/cli v1.22.1
 	github.com/vishvananda/netlink v1.1.0
-	golang.org/x/sys v0.0.0-20200728102440-3e129f6d46b1
+	golang.org/x/sys v0.0.0-20200909081042-eff7692f9009
```
thaJeztah, line 25: note that the golang.org/x/sys updates are incompatible with the containerd/console package, and break compile on macOS. A fix is open for this in https://github.com/containerd/console/pull/40

I don't think anything in runc is compiled on macOS currently, so probably not a blocker, but it may cause issues for projects that use runc as a go module, as it will force them to update to a newer golang.org/x/sys version

thaJeztah:
@kolyshkin @cyphar ptal

AkihiroSuda (dismissed):

go.sum
```diff
@@ -34,6 +34,8 @@ github.com/konsorten/go-windows-terminal-sequences v1.0.3 h1:CE8S1cTafDpPvMhIxNJ
 github.com/konsorten/go-windows-terminal-sequences v1.0.3/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
 github.com/moby/sys/mountinfo v0.1.3 h1:KIrhRO14+AkwKvG/g2yIpNMOUVZ02xNhOw8KY1WsLOI=
 github.com/moby/sys/mountinfo v0.1.3/go.mod h1:w2t2Avltqx8vE7gX5l+QiBKxODu2TX0+Syr3h52Tw4o=
+github.com/moby/sys/mountinfo v0.2.0 h1:HgYSHMWCj8D7w7TE/cQJfWrY6W3TUxs3pwGFyC5qCvE=
+github.com/moby/sys/mountinfo v0.2.0/go.mod h1:rEr8tzG/lsIZHBtN/JjGG+LMYx9eXgW2JI+6q0qou+A=
```
kolyshkin, line 38: have you tried `go mod tidy`? I know it's not always works that is why I'm asking :)

thaJeztah: I think I ran `make vendor` (which should do all the steps, correct?)

thaJeztah: Let me jump to my computer and double check

thaJeztah: You were right; `go mod tidy` did give a diff. Perhaps the `make vendor` target needs to be updated to run `go mod tidy` afterwards (it now runs first)?

thaJeztah:
@AkihiroSuda I pushed again, so looks like you need to re-approve 😅

go.sum
kolyshkin: Interesting. From what I read at https://github.com/golang/go/wiki/Modules it seems that `go mod tidy` can add packages to go.mod, meaning it needs to be run before `go mod vendor`.

So, unless `go mod vendor` can modify `go.mod` and `go.sum` (looks like it does not, but I might be wrong here), there's no need to run `go mod tidy` after it.

thaJeztah: It's odd that CI was green, and didn't show a diff

kolyshkin: Took me a while to figure out. Apparently, this thing never worked :facepalm:  

See https://github.com/opencontainers/runc/pull/2593

AkihiroSuda (approved):

cyphar (approved):
LGTM, and :+1: for openat2 support. :smile_cat:

PR #2592, opened 2020-09-16
tests/integration: nits

kolyshkin:
Most notable one is spec.bats simplification and speedup.

The rest is nits and minor fixes.  Please see individual commits for details.

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -3,7 +3,6 @@
 load helpers
 
 function teardown() {
-    rm -f "$BATS_TMPDIR"/runc-cgroups-integration-test.json
     teardown_running_container test_cgroups_kmem
     teardown_running_container test_cgroups_permissions
     teardown_busybox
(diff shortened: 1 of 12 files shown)


kolyshkin:
Interesting... my changes resulting in spec generated by `runc spec --rootless` being validated (which was not performed before), and the validation fails with the following error:

```
The document is not valid. see errors :
 - linux.resources.pids: limit is required
```

Indeed, runtime-spec says pids.limit is required: 
https://github.com/opencontainers/runtime-spec/blob/80db136d65fc1a90b1bbbcc964a4b354b7aad0b2/schema/config-linux.json#L53

@AkihiroSuda mind to take a look? Or should we NOT run this for the rootless case?

AkihiroSuda:
Shouldn't be required even for rootful, I guess

kolyshkin:
OK this is caused by https://github.com/opencontainers/runc/blob/d636ad6256f9194b0f4c6ee181e75fb36e3446d8/tests/integration/helpers.bash#L108

kolyshkin:
@mrunalp @AkihiroSuda PTAL

AkihiroSuda (approved):

mrunalp (approved):

PR #2593, opened 2020-09-16
Makefile: fix vendor and verify-dependencies

kolyshkin:
Commit a08ab87fe added these targets. Alas, the `go mod tidy` never
worked. Here is the relevant part of the `vendor` target in `Makefile`:

https://github.com/opencontainers/runc/blob/892477ca26638b214dc79150a09ba3565b93137d/Makefile#L128-L130

which is essentially

```bash
export GO111MODULE=on $(GO) mod tidy && ...
```
which exports a bunch of variables, such as `go`, `mod`, and `tidy`
(or would export those if they existed), but does not run `go mod tidy`.

The fix would be to add a semicolon after the `export` statement,
but since `GO111MODULE` is not really needed here (maybe some older
golang versions needed it?), let's just drop it.

With this dropped, && does not make any sense, so drop it, too.

NOTE that if someone tries

        GO111MODULE=off make vendor

it will fail, but I guess it is expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,9 +126,8 @@ validate:
 ci: validate test release
 
 vendor:
-	export GO111MODULE=on \
-	$(GO) mod tidy && \
-	$(GO) mod vendor && \
+	$(GO) mod tidy
+	$(GO) mod vendor
 	$(GO) mod verify
 
 verify-dependencies: vendor


mrunalp (dismissed):

Makefile
```diff
@@ -126,7 +126,7 @@ validate:
 ci: validate test release
 
 vendor:
-	export GO111MODULE=on \
+	GO111MODULE=on \
```
thaJeztah, line 129: You still need to `export`, otherwise `mod vendor` and `mod verify` won't have `GO111MODULE=on` set;

```console
$ GO111MODULE=on go env GO111MODULE && go env GO111MODULE && go env GO111MODULE
on


$ export GO111MODULE=on; go env GO111MODULE && go env GO111MODULE && go env GO111MODULE
on
on
on
```

thaJeztah (changes requested):
Looks like you still need to export

Makefile
kolyshkin: You are quite right! But I think this setting is not needed now, at all.

Updated the commit and the PR description, PTAL

thaJeztah (approved):
ah, yes, that should probably work okay

LGTM

kolyshkin:
> ah, yes, that should probably work okay

In fact, moby/moby tries to build runc with `GO111MODULE=off`, but the `vendor` target is not called during the build.

AkihiroSuda (approved):

thaJeztah:
> In fact, moby/moby tries to build runc with GO111MODULE=off, but the vendor target is not called during the build.

Correct. I think that was at least from before run switched to go modules, but I also don't think it hurts as we git clone (not `go get`, which would ignore the vendor directory), and we want to build with the vendored dependencies, and make sure that `go build` doesn't try to modify anything.

kolyshkin:
> we want to build with the vendored dependencies, and make sure that go build doesn't try to modify anything

AFAIK `go -mod=vendor` can be used to this effect (and should be sufficient) since Go 1.12 or so, and since Go 1.14 it is the default if vendor subdir is present.

thaJeztah:
Yeah, current versions of go should be able to detect the `vendor` directory when using `go build` or `go test`. Can't do a `go get` and use the vendored dependencies (that still requires a manual git clone). Anyway, that all is a separate discussion.

mrunalp (approved):

PR #2594, opened 2020-09-18
Makefile: disable kernel memory accounting on RHEL 7 3.10 kernels

thaJeztah:
Kernel-memory accounting is known to be broken on RHEL 7 kernels. This patch automatically disables kernel-memory accounting when building on a 3.10 kernel.

The original behavior is preserved if a custom BUILDTAGS is passed;

    make BUILDTAGS='seccomp selinux' ...

I recalled we had a similar patch in Moby (https://github.com/moby/moby/commit/8972aa9350d52e4a7e58242447b7a9d2f0c27f37), and thought I'd might as well upstream this.

Also note that https://github.com/moby/moby/pull/41254#issue-456148870

> Kernel memory limit is not supported on cgroup v2. Even on cgroup v1, kernel memory limit (kmem.limit_in_bytes) has been deprecated since kernel 5.4. https://github.com/torvalds/linux/commit/0158115f702b0ba208ab0b5adf44cae99b3ebcc7

So perhaps either add a similar check for kernel 5.4+, or disabling kmem by default, and making it an opt-on could make sense (@AkihiroSuda wdyt?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,16 +61,19 @@ sudo make install
 with some of them enabled by default (see `BUILDTAGS` in top-level `Makefile`).
 
 To change build tags from the default, set the `BUILDTAGS` variable for make,
-e.g.
+e.g. to disable `seccomp` and enable `nokmem`, run:
 
 ```bash
-make BUILDTAGS='seccomp'
+make BUILDTAGS="nokmem"
 ```
 
 | Build Tag | Feature                            | Enabled by default | Dependency |
 |-----------|------------------------------------|--------------------|------------|
 | seccomp   | Syscall filtering                  | yes                | libseccomp |
-| nokmem    | disable kernel memory accounting   | no                 | <none>     |
+| nokmem    | disable kernel memory accounting   | usually not        | _none_     |
+
+**Note** `nokmem` build tag is now auto-set by the Makefile in case
+the running kernel version looks like one from RHEL7 (3.10.0-\*.el7.)
 
 The following build tags were used earlier, but are now obsoleted:
  - **apparmor** (since runc v1.0.0-rc93 the feature is always enabled)
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,12 @@ ifeq ($(shell $(GO) env GOOS),linux)
 		endif
 	endif
 endif
+
+# Disable kmem if building on RHEL7 (kernel 3.10.0 el7)
+ifneq ($(shell uname -r | grep '^3\.10\.0.*\.el7\.'),)
+	BUILDTAGS += nokmem
+endif
+
 GO_BUILD := $(GO) build -trimpath $(GO_BUILDMODE) $(EXTRA_FLAGS) -tags "$(BUILDTAGS)" \
 	-ldflags "-X main.gitCommit=$(COMMIT) -X main.version=$(VERSION) $(EXTRA_LDFLAGS)"
 GO_BUILD_STATIC := CGO_ENABLED=1 $(GO) build -trimpath $(EXTRA_FLAGS) -tags "$(BUILDTAGS) netgo osusergo" \
--- a/libcontainer/cgroups/fs/kmem.go
+++ b/libcontainer/cgroups/fs/kmem.go
@@ -13,7 +13,10 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-const cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
+const (
+	kmemDisabled            = false
+	cgroupKernelMemoryLimit = "memory.kmem.limit_in_bytes"
+)
 
 func EnableKernelMemoryAccounting(path string) error {
 	// Ensure that kernel memory is available in this kernel build. If it
--- a/libcontainer/cgroups/fs/kmem_disabled.go
+++ b/libcontainer/cgroups/fs/kmem_disabled.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 )
 
+const kmemDisabled = true
+
 func EnableKernelMemoryAccounting(path string) error {
 	return nil
 }
--- a/libcontainer/cgroups/fs/memory_test.go
+++ b/libcontainer/cgroups/fs/memory_test.go
@@ -191,6 +191,9 @@ func TestMemorySetSwapSmallerThanMemory(t *testing.T) {
 }
 
 func TestMemorySetKernelMemory(t *testing.T) {
+	if kmemDisabled {
+		t.Skip("kernel memory limits are disabled")
+	}
 	helper := NewCgroupTestUtil("memory", t)
 	defer helper.cleanup()
 
@@ -219,6 +222,9 @@ func TestMemorySetKernelMemory(t *testing.T) {
 }
 
 func TestMemorySetKernelMemoryTCP(t *testing.T) {
+	if kmemDisabled {
+		t.Skip("kernel memory limits are disabled")
+	}
 	helper := NewCgroupTestUtil("memory", t)
 	defer helper.cleanup()
 
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -685,6 +685,9 @@ func testRunWithKernelMemory(t *testing.T, systemd bool) {
 	if cgroups.IsCgroup2UnifiedMode() {
 		t.Skip("cgroup v2 does not support kernel memory limit")
 	}
+	if kmemDisabled {
+		t.Skip("kernel memory is disabled in this runc build")
+	}
 
 	rootfs, err := newRootfs()
 	ok(t, err)
--- a/libcontainer/integration/kmem_disabled_test.go
+++ b/libcontainer/integration/kmem_disabled_test.go
@@ -1,0 +1,5 @@
+// +build linux,nokmem
+
+package integration
+
+const kmemDisabled = true
--- a/libcontainer/integration/kmem_test.go
+++ b/libcontainer/integration/kmem_test.go
@@ -1,0 +1,5 @@
+// +build linux,!nokmem
+
+package integration
+
+const kmemDisabled = false
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -318,6 +318,10 @@ function requires() {
 			if [ ! -e "${CGROUP_MEMORY_BASE_PATH}/memory.kmem.limit_in_bytes" ]; then
 				skip_me=1
 			fi
+			if uname -r | grep -q '^3\.10\.0.*\.el7\.'; then
+				# In RHEL7 kernel, kmem is broken so runc disables it.
+				skip_me=1
+			fi
 			;;
 		cgroups_rt)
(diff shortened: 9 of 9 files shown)


thaJeztah:
@kolyshkin ptal

cyphar:
TIL that kmemcg is now deprecated, neither the kernel containers ML nor the cgroup ML was on the Cc list for that change.

thaJeztah:
Thanks for Akihiro, who noticed that. I guess it has been somewhat problematic overall, but I was surprised as well.

Given its deprecation, perhaps a decision should be made before runc 1.0.0 wether or not it should be enabled by default (or supported at all; it could be (silently) "ignored" to not break compatibility)

kolyshkin:
This is very similar to how I did this initially in #1920. Someone [raised the concern](https://github.com/opencontainers/runc/pull/1920#issuecomment-434897881) that it should be done by packagers rather than during runtime, and we eventually ended up with https://github.com/opencontainers/runc/pull/1921.

I think we're good, meaning whoever packages runc for RHEL7 should set `nokmem`.

kolyshkin:
> TIL that kmemcg is now deprecated, neither the kernel containers ML nor the cgroup ML was on the Cc list for that change.

I learned it recently, too. As much as I like it to be a separate limit, I don't think the majority of users are able to figure out their kmem settings. Currently (at least in cgroupv2) kmem is still accounted and [indirectly] limited by cgroup.memory settings, it's just there are no more direct controls.

kolyshkin:
Back in the day in OpenVZ we used to have 20+ limits (https://wiki.openvz.org/Proc/user_beancounters) and users were confused. Ended up having just two primary knobs: ram and swap (https://wiki.openvz.org/VSwap).

cyphar:
We will need to remove it from the runtime-spec too...

kolyshkin:
This CI failure (on CentOS 7) needs to be fixed if we want to proceed:
```
=== RUN   TestMemorySetKernelMemory
1290    memory_test.go:207: kernel memory accounting disabled in this runc build
1291--- FAIL: TestMemorySetKernelMemory (0.00s)
```

thaJeztah:
Added a `t.Skip()` for those tests if kernel-memory is disabled

kolyshkin:
I'm not sure what to do about this one. RHEL7 is old, most probably no one is using runc installed from source on it, and whoever is packaging runc for RHEL7 is already taking care of adding `nokmem` build tag (or so I hope).

thaJeztah:
> whoever is packaging runc for RHEL7 is already taking care of adding nokmem build tag (or so I hope).

Yeah, it's a bit tricky. Agreed that packagers should take this into account, OTOH, it's nice to have `make XX` "do the right thing" out-of-the-box, and this one is easy to miss (things don't fail directly if you missed setting this option).

Even more so if more recent kernels are used (and users may not be aware of it being deprecated);

> So perhaps either add a similar check for kernel 5.4+, or disabling kmem by default, and making it an opt-on could make

kolyshkin:
The only problem I see it's now impossible to build runc on RHEL7 with kmem. But maybe it's for good, and I'm pretty sure kmem accounting won't be fixed there.

@thaJeztah can you please re-base to bring in recent CI?

cyphar:
I think it doesn't hurt to include this, especially since we have other (actual code) workarounds for RHEL7 (the first thing that comes to mind is the `last_cap` capabilities thing).

kolyshkin:
I took a liberty to 
- rebase;
- simplify the kernel version check a bit;
- fix adding the nokmem tag (it has to be set before `GO_BUILD :=`, not after).

It was not working before:
```console
[kir@centos70-ne runc]$ make
go build "-mod=vendor" "-buildmode=pie"  -tags "seccomp selinux apparmor" -ldflags "-X main.gitCommit="5ec32f266c90ecb8da36282a63a6340edf718a18" -X main.version=1.0.0-rc92+dev " -o runc .
[kir@centos70-ne runc]$ uname -a
Linux centos70-ne 3.10.0-1160.6.1.el7.x86_64 #1 SMP Tue Nov 17 13:59:11 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
[kir@centos70-ne runc]$ if uname -r | grep -q '^3\.10\.0.*\.el7\.'; then echo "yes"; fi
yes
```

It is working now:
```console
[kir@centos70-ne runc]$ make
go build -trimpath "-mod=vendor" "-buildmode=pie"  -tags "seccomp nokmem" -ldflags "-X main.gitCommit="fd7f880166e3b57393e42a905d953ead07683866" -X main.version=1.0.0-rc93+dev " -o runc .
```

The changes are
```diff
diff --git a/Makefile b/Makefile
index c283b9a6..5ff44f35 100644
--- a/Makefile
+++ b/Makefile
@@ -25,16 +25,17 @@ ifeq ($(shell $(GO) env GOOS),linux)
                endif
        endif
 endif
+
+# Disable kmem if building on RHEL7 (kernel 3.10.0 el7)
+ifneq ($(shell uname -r | grep '^3\.10\.0.*\.el7\.'),)
+       BUILDTAGS += nokmem
+endif
+
 GO_BUILD := $(GO) build -trimpath $(MOD_VENDOR) $(GO_BUILDMODE) $(EXTRA_FLAGS) -tags "$(BUILDTAGS)" \
        -ldflags "-X main.gitCommit=$(COMMIT) -X main.version=$(VERSION) $(EXTRA_LDFLAGS)"
 GO_BUILD_STATIC := CGO_ENABLED=1 $(GO) build -trimpath $(MOD_VENDOR) $(EXTRA_FLAGS) -tags "$(BUILDTAGS) netgo osusergo" \
        -ldflags "-w -extldflags -static -X main.gitCommit=$(COMMIT) -X main.version=$(VERSION) $(EXTRA_LDFLAGS)"
 
-ifeq ($(shell if uname -r | grep -q '^3\.10\.0.*\.el7\.'; then echo "yes"; fi),yes)
-       # Disable kmem accounting/limiting on RHEL7 kernels (3.10.0 el7)
-       BUILDTAGS += nokmem
-endif
-
 .DEFAULT: runc
 
 runc:
```
```

**Comment by kolyshkin** [comment]
```
    exec_test.go:701: runContainer failed with kernel memory limit: container_linux.go:367: starting container process caused: process_linux.go:520: container init caused: process_linux.go:483: setting cgroup config for procHooks process caused: kernel memory accounting disabled in this runc build
--- FAIL: TestRunWithKernelMemory (0.29s)
=== RUN   TestRunWithKernelMemorySystemd
    exec_test.go:701: runContainer failed with kernel memory limit: container_linux.go:367: starting container process caused: process_linux.go:520: container init caused: process_linux.go:483: setting cgroup config for procHooks process caused: kernel memory accounting disabled in this runc build

--- FAIL: TestRunWithKernelMemorySystemd (0.29s)
```

More tests need to be disabled...

**Comment by kolyshkin** [comment]
Added a commit to disable libct/int kmem-related tests if nokmem build tag is set.

I have also changed `kmem_disabled` to `kmemDisabled` to follow the Go naming guidelines.

@thaJeztah PTAL

**Comment by kolyshkin** [comment]
> (it has to be set before GO_BUILD :=, not after)

This roots in a difference between make's `=` and `:=`. The latter substitutes all variables in place (i.e. right here right now), while the former keeps the `$(VARIABLE)` as is and they are only expanded when used. For more info, see https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors

**Comment by kolyshkin** [comment]
Finally, I think we need to amend the README to say nokmem is auto-set when compiling on 3.0.0-el7 kernel.

**Comment by thaJeztah** [comment]
Thanks Kir, sorry, I forgot about this one 🤗

**Comment by kolyshkin** [comment]
Make is fun -- depending on whether the variable is coming from the environment or CLI arguments, the effect is different:

```console
[kir@centos70-ne runc]$  make
go build -trimpath "-mod=vendor" "-buildmode=pie"  -tags "seccomp nokmem" -ldflags "-X main.gitCommit="0b2de072c311e8095e3810a9efb69171d06974c2" -X main.version=1.0.0-rc93+dev " -o runc .

[kir@centos70-ne runc]$ BUILDTAGS="" make
go build -trimpath "-mod=vendor" "-buildmode=pie"  -tags " nokmem" -ldflags "-X main.gitCommit="0b2de072c311e8095e3810a9efb69171d06974c2" -X main.version=1.0.0-rc93+dev " -o runc .

[kir@centos70-ne runc]$ make BUILDTAGS=""
go build -trimpath "-mod=vendor" "-buildmode=pie"  -tags "" -ldflags "-X main.gitCommit="0b2de072c311e8095e3810a9efb69171d06974c2" -X main.version=1.0.0-rc93+dev " -o runc .
```

kolyshkin:
Seeing this for the first time (on CentOS 7):

```
--- PASS: TestFreeze (0.50s)
=== RUN   TestSystemdFreeze
    utils_test.go:86: exec_test.go:539: unexpected error: unable to freeze

--- FAIL: TestSystemdFreeze (1.03s)
```

kolyshkin:
Added a commit with README.md changes.

kolyshkin:
Yet another test to fix :)

```
not ok 1 runc update --kernel-memory{,-tcp} (initialized)
# (in test file tests/integration/cgroups.bats, line 24)
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-7663/runc.OeJiKe/tty/sock test_cgroups_kmem (status=1):
# time="2021-03-04T20:20:24Z" level=error msg="container_linux.go:367: starting container process caused: process_linux.go:520: container init caused: process_linux.go:483: setting cgroup config for procHooks process caused: kernel memory accounting disabled in this runc build"
```

kolyshkin:
Rebased, added a skip to int tests.

kolyshkin:
CI failure

```
not ok 27 global --debug to --log --log-format 'text'
# (in test file tests/integration/debug.bats, line 58)
```

is a well-known flake (see https://github.com/opencontainers/runc/issues/2756). Restarted.

README.md
```diff
@@ -61,16 +61,19 @@ sudo make install
 with some of them enabled by default (see `BUILDTAGS` in top-level `Makefile`).
 
 To change build tags from the default, set the `BUILDTAGS` variable for make,
-e.g.
+e.g. to disable `seccomp` and enable `nokmem`, run:
 
 ```bash
-make BUILDTAGS='seccomp'
+make BUILDTAGS="nokmem"
```
thaJeztah, line 67: I think we should encourage keeping seccomp enabled; should we change to (e.g.);

```suggestion
e.g. to disable kernel memory limiting, but keep `seccomp`, run:
 
  ```bash
  make BUILDTAGS="nokmem seccomp"
```

kolyshkin: Doesn't matter much to me, it's just a usage example, not a recommendation about how to compile runc, so can be anything (and I sometimes find myself disabling seccomp so I can compile runc without having to install libseccomp-devel).

But yes, maybe yours is a better/fuller example.

thaJeztah: Yeah, I agree it's just an example. People love to copy/paste though, and hate to read 😓

PR #2595, opened 2020-09-19
Libcontainer: refactor capabilities code and remove RHEL6 workaround

thaJeztah:
See individual commit messages for details

### DIFF
--- a/libcontainer/capabilities_linux.go
+++ b/libcontainer/capabilities_linux.go
@@ -21,12 +21,12 @@ func init() {
 	if last == capability.Cap(63) {
 		last = capability.CAP_BLOCK_SUSPEND
 	}
-	for _, cap := range capability.List() {
-		if cap > last {
+	for _, c := range capability.List() {
+		if c > last {
 			continue
 		}
-		capKey := fmt.Sprintf("CAP_%s", strings.ToUpper(cap.String()))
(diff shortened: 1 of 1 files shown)


thaJeztah:
@kolyshkin @mrunalp ptal

AkihiroSuda:
Can we update the README to clarify the current minimum kernel requirement? 3.10? (Does it work without Red Hat patches?)

thaJeztah:
I don't think we test 3.10.x on anything other than RHEL 7 / CentOS 7 kernels. Not sure if there's still uses of runc on Debian Jessie (3.16); I know that kernel was somewhat problematic at some point.

Perhaps we should be explicit on the intent; something along the lines of "Runc support kernel 3.10 to facilitate RHEL7 and CentOS7. For other distributions, we recommend/support kernel 4.0 and up"

thaJeztah:
FWIW; I'd be ok with keeping the RHEL6 hack if it's a concern (removing it is just a minor cleanup, so I'm not strongly attached to removing it)

kolyshkin (approved):
LGTM, and thanks!

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda (approved):

PR #2597, opened 2020-09-22
Fix/improve GetHugePageSize

kolyshkin:
### libct/cgroups/GetHugePageSize: use Readdirnames

ioutil.ReadFile does a stat() on every entry and returns a slice of
os.Stat structures. What we need here is just a file name.
    
This change both simplifies and speeds up the code a bit
(30% on my laptop).

### libct/cgroups: rewrite getHugePageSizeFromFilenames

This is a function to convert huge page sizes (obtained by reading
/sys/kernel/mm/hugepages directory entries) to strings user for hugetlb
cgroup controller resource files. Those strings are when used to get the
hugetlb resource statistics.

This function used external library, floating point numbers, and can
(theoretically) produce invalid values, since the kernel only uses KB,
MB, and GB suffixes.

Rewrite it to produce the same strings as used in the kernel (see [1]).
As a result, it's also faster (about 11x on my laptop -- see below), more
future-proof (entries that do not start with "hugepages-" are skipped),
and does more input sanity checks.

As a side effect, libcontainer no longer depends on docker/go-units.

While at it, improve the test logic and add more test cases.

NOTE on removing `HugePageSizeUnitList` -- this was added by commit
6f77e35da and was used by kubernetes code in [2], which was later
superceded by [3], so there are (hopefully) no external users.
If there are any, they should not be doing that.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/hugetlb_cgroup.c?id=eff48ddeab782e35e58ccc8853f7386bbae9dec4#n574
[2] https://github.com/kubernetes/kubernetes/pull/78495
[3] https://github.com/kubernetes/kubernetes/pull/84154

### Benchmarks

Before:
```
BenchmarkGetHugePageSize-8                115213              9400 ns/op
BenchmarkGetHugePageSizeImpl-8            397873              2971 ns/op
```

After:
```
BenchmarkGetHugePageSize-8                222898              4554 ns/op
BenchmarkGetHugePageSizeImpl-8           4738924               241 ns/op
```

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -15,7 +15,6 @@ import (
 	"sync"
 	"time"
 
-	units "github.com/docker/go-units"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
@@ -29,13 +28,6 @@ var (
 	isUnifiedOnce sync.Once
 	isUnified     bool
 )
-
-// HugePageSizeUnitList is a list of the units used by the linux kernel when
-// naming the HugePage control files.
-// https://www.kernel.org/doc/Documentation/cgroup-v1/hugetlb.txt
-// TODO Since the kernel only use KB, MB and GB; TB and PB should be removed,
-// depends on https://github.com/docker/go-units/commit/a09cd47f892041a4fac473133d181f5aea6fa393
-var HugePageSizeUnitList = []string{"B", "KB", "MB", "GB", "TB", "PB"}
 
 // IsCgroup2UnifiedMode returns whether we are running in cgroup v2 unified mode.
 func IsCgroup2UnifiedMode() bool {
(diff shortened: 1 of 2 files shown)


mrunalp (approved):

libcontainer/cgroups/utils.go
```diff
@@ -285,27 +277,48 @@ func RemovePaths(paths map[string]string) (err error) {
...
+		// example: hugepages-1048576kB (always in kB)
+		val := strings.TrimPrefix(file, "hugepages-")
+		if len(val) == len(file) {
+			// unexpected file name: no prefix found
+			continue
+		}
+		eLen := len(val) - 2
+		val = strings.TrimSuffix(val, "kB")
+		if len(val) != eLen {
+			return nil, fmt.Errorf("%s: invalid suffix (expected \"kB\")", file)
```
AkihiroSuda, line 306: Should this be a warning rather than a critical error?

kolyshkin: Changed to a warning; modified the test case to check for warnings.

@AkihiroSuda PTAL

AkihiroSuda (approved):

mrunalp (approved):

PR #2598, opened 2020-09-23
libct/cgroups/fscommon: add openat2() support

kolyshkin:
This is a naive initial attempt to add openat2 support to libct/cgroups/fscommon, and switch libct/cgroups to use it.

This is now being split into a few more digestible PRs:

 - [x] #2604 libct/cgroups/fscommon: nits
 - [x] #2605 libct/cgroups: switch to fscommon.{Read,Write}File
 - [x] #2635 Introduce and use OpenFile
 - [ ] #2668 Add openat2() support to OpenFile

### DIFF
--- a/libcontainer/cgroups/fs/blkio.go
+++ b/libcontainer/cgroups/fs/blkio.go
@@ -6,7 +6,6 @@ import (
 	"bufio"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -105,9 +104,9 @@ func splitBlkioStatLine(r rune) bool {
 	return r == ' ' || r == ':'
 }
 
-func getBlkioStat(path string) ([]cgroups.BlkioStatEntry, error) {
+func getBlkioStat(dir, file string) ([]cgroups.BlkioStatEntry, error) {
 	var blkioStats []cgroups.BlkioStatEntry
-	f, err := os.Open(path)
+	f, err := fscommon.OpenFile(dir, file, os.O_RDONLY)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return blkioStats, nil
@@ -125,7 +124,7 @@ func getBlkioStat(path string) ([]cgroups.BlkioStatEntry, error) {
 				// skip total line
 				continue
 			} else {
-				return nil, fmt.Errorf("Invalid line found while parsing %s: %s", path, sc.Text())
+				return nil, fmt.Errorf("Invalid line found while parsing %s/%s: %s", dir, file, sc.Text())
 			}
 		}
 
@@ -159,7 +158,7 @@ func getBlkioStat(path string) ([]cgroups.BlkioStatEntry, error) {
 
 func (s *BlkioGroup) GetStats(path string, stats *cgroups.Stats) error {
 	// Try to read CFQ stats available on all CFQ enabled kernels first
-	if blkioStats, err := getBlkioStat(filepath.Join(path, "blkio.io_serviced_recursive")); err == nil && blkioStats != nil {
+	if blkioStats, err := getBlkioStat(path, "blkio.io_serviced_recursive"); err == nil && blkioStats != nil {
 		return getCFQStats(path, stats)
 	}
 	return getStats(path, stats) // Use generic stats as fallback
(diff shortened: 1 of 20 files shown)


kolyshkin:
Funny, `libcontainer/integration` tests fail because now `WriteFile` do not use `O_CREAT` (which is obviously not needed for real cgroupfs). Need to think of a way to handle this.

libcontainer/cgroups/fscommon/openat2.go
```diff
@@ -0,0 +1,73 @@
...
+var (
+	cgroupFd     int = -1
+	prepOnce     sync.Once
+	prepErr      error
+	resolveFlags uint64
+)
+
+func prepareOpenat2() error {
+	prepOnce.Do(func() {
+		fd, err := unix.Openat2(-1, cgroupfsDir, &unix.OpenHow{Flags: unix.O_DIRECTORY | unix.O_PATH})
```
cyphar, line 26: You never assign `cgroupFd` to this fd, so it gets leaked and the later logic isn't right (you'll always get `EBADF`).

libcontainer/cgroups/fscommon/openat2.go
```diff
@@ -0,0 +1,73 @@
...
+		var st unix.Statfs_t
+		if err = unix.Fstatfs(fd, &st); err != nil {
+			prepErr = &os.PathError{Op: "statfs", Path: cgroupfsDir, Err: err}
+			return
+		}
+		if st.Type == unix.CGROUP2_SUPER_MAGIC {
+			resolveFlags = unix.RESOLVE_NO_XDEV | unix.RESOLVE_NO_SYMLINKS
+		} else {
+			// cgroupfs is a set of separate mounts
+			resolveFlags = unix.RESOLVE_BENEATH
```
cyphar, line 40: You could cache an fd for each subsystem, but I get why you wouldn't want to do that.

cyphar:
Just wanted to mentioned that my long-term plan is to make it easier to use openat2 for this use-case from userspace with libpathrs -- `securejoin.SecureJoin` is actually quite insecure, and ideally we would use the more complicated userspace path lookup emulation. But I'm totally fine with doing some piece-meal things like this -- since we can easily port it to `libpathrs` in the future. The nice thing about libpathrs is that we don't need any of the fallback code to live in runc -- it's all handled transparently by libpathrs.

libcontainer/cgroups/fscommon/openat2.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	})
+
+	return prepErr
+}
+
+func OpenFile(dir, file string, flags int) (*os.File, error) {
```
cyphar, line 47: I do get why you take a `dir` here but the interface just looks a little bit weird, maybe it should be clearer that this is only intended to open cgroup paths...

libcontainer/cgroups/fscommon/openat2.go
kolyshkin: lost while rebasing :)

libcontainer/cgroups/fscommon/openat2.go
kolyshkin: I haven't decided yet whether we should use it for v1 ...

libcontainer/cgroups/fscommon/openat2.go
kolyshkin: The inital name was `OpenCgroupFile()` but when I saw the other functions in the package are `ReadFile` and `WriteFile` so I renamed accordingly.

Docs is still todo.

cyphar: Fair enough.

kolyshkin: Oh, and I forgot to mention that existing `ReadFile` and `WriteFile` also have `dir` and `file` arguments (which seems reasonable as all the callers have a cgroup path and want to deal with various files under it) so I'm just modelling after.

cyphar: Yeah since it's common code annoyingly we have to deal with the directory path (including the subsystem name). :/

kolyshkin:
Pushed a new patchset, slightly better organized/splitted, and with more users converted.

At this point I'm thinking about maybe changing the whole libct/cg to use per-cgroup fd (rather than fd opened to /sys/fs/cgroup) and `openat()` (when `openat2` is not available) -- this should speed things up because the kernel will have to do less dentry lookups.

kolyshkin:
> At this point I'm thinking about maybe changing the whole libct/cg to use per-cgroup fd (rather than fd opened to /sys/fs/cgroup) and openat() (when openat2 is not available) -- this should speed things up because the kernel will have to do less dentry lookups.

This can easily be done on top of this one, so later.

For now, I am going to split this into smaller more digestible PRs.

AkihiroSuda:
needs rebase

kolyshkin:
Closed as ￼this is now splitted into
- #2604 libct/cgroups/fscommon: nits
- #2605 libct/cgroups: switch to fscommon.{Read,Write}File
- ￼#2635 Introduce and use OpenFile
- ￼#2668 Add openat2() support to OpenFile

and the only last one is not yet merged

PR #1049, opened 2016-09-15
Add flag to allow getting all mounts for cgroups subsystems

mrunalp:
This requirement comes from cadvisor so it can pick the right mounts that it needs when run in a container. It shouldn't affect internal libcontainer/runc code.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

PR #2599, opened 2020-09-23
Another nail to mountinfo coffin

kolyshkin:
This is a set of patches for libcontainer/cgroups that

 - implements mountinfo cache for cgroup filesystem and reuses it;
 - drops two custom mountinfo parsers (one in favor of `moby/sys/mountinfo`, another to reuse `cgroups/utils`);
 - removes mountinfo dependency in `cgroups/fs/cpuset` code.

The idea is roughly the same as https://github.com/opencontainers/runc/pull/2265 but the implementation is way simpler.
The end result is smaller code (`98 insertions(+), 147 deletions(-)`), less bugs (I hope) and overall higher performance.

Tested it with strace shows that the number of times `runc run` opens `/proc/self/mountinfo` has decreased from 19 to 6 times (note it was 108 half a year ago). Of which 5 times must be out of cgroups code (haven't checked), so it's 14 -> 1 improvement. In absolute terms, it can be up to 0.023*14 ~= 0.3 seconds savings on a container start (on a system very busy with mounts).

**v2 changes:**
 - unexport `readCgroupMountinfo` (by removing its two external users, see below);
 - switch cgroups/fs/getCgroupRoot to use `cgroups.GetCgroupMounts`;
 - don't rely on mountinfo in cgroups/fs/cpuset code (switch to `statfs`);
 - improve commit messages.

NOTE this currently depends on (and includes) #2689 -- will rebase once that one is merged.

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -3,11 +3,9 @@
 package fs
 
 import (
-	"bufio"
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 	"sync"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -133,45 +131,18 @@ func getCgroupRoot() (string, error) {
 		return cgroupRoot, nil
 	}
 
-	// slow path: parse mountinfo, find the first mount where fs=cgroup
-	// (e.g. "/sys/fs/cgroup/memory"), use its parent.
-	f, err := os.Open("/proc/self/mountinfo")
+	// slow path: parse mountinfo
+	mi, err := cgroups.GetCgroupMounts(false)
 	if err != nil {
 		return "", err
 	}
-	defer f.Close()
-
-	var root string
-	scanner := bufio.NewScanner(f)
-	for scanner.Scan() {
-		text := scanner.Text()
-		fields := strings.Split(text, " ")
-		// Safe as mountinfo encodes mountpoints with spaces as \040.
-		index := strings.Index(text, " - ")
-		postSeparatorFields := strings.Fields(text[index+3:])
-		numPostFields := len(postSeparatorFields)
-
-		// This is an error as we can't detect if the mount is for "cgroup"
-		if numPostFields == 0 {
-			return "", fmt.Errorf("mountinfo: found no fields post '-' in %q", text)
-		}
-
-		if postSeparatorFields[0] == "cgroup" {
-			// Check that the mount is properly formatted.
-			if numPostFields < 3 {
-				return "", fmt.Errorf("Error found less than 3 fields post '-' in %q", text)
-			}
-
-			root = filepath.Dir(fields[4])
-			break
-		}
-	}
-	if err := scanner.Err(); err != nil {
-		return "", err
-	}
-	if root == "" {
+	if len(mi) < 1 {
 		return "", errors.New("no cgroup mount found in mountinfo")
 	}
+
+	// Get the first cgroup mount (e.g. "/sys/fs/cgroup/memory"),
+	// use its parent directory.
+	root := filepath.Dir(mi[0].Mountpoint)
 
 	if _, err := os.Stat(root); err != nil {
 		return "", err
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/moby/sys/mountinfo"
 	"github.com/sirupsen/logrus"
 )
 
@@ -244,7 +245,13 @@ func TestGetCgroupMounts(t *testing.T) {
 		},
 	}
 	for _, td := range testTable {
-		mi := bytes.NewBufferString(td.mountInfo)
+		mi, err := mountinfo.GetMountsFromReader(
+			bytes.NewBufferString(td.mountInfo),
+			mountinfo.FSTypeFilter("cgroup"),
+		)
+		if err != nil {
+			t.Fatal(err)
+		}
 		cgMounts, err := getCgroupMountsHelper(td.subsystems, mi, false)
 		if err != nil {
 			t.Fatal(err)
@@ -281,7 +288,6 @@ func TestGetCgroupMounts(t *testing.T) {
 		// Test the all=true case.
 
 		// Reset the test input.
-		mi = bytes.NewBufferString(td.mountInfo)
 		for k := range td.subsystems {
 			td.subsystems[k] = false
 		}
@@ -317,11 +323,15 @@ func BenchmarkGetCgroupMounts(b *testing.B) {
 		"perf_event": false,
 		"hugetlb":    false,
 	}
+	mi, err := mountinfo.GetMountsFromReader(
+		bytes.NewBufferString(fedoraMountinfo),
+		mountinfo.FSTypeFilter("cgroup"),
+	)
+	if err != nil {
+		b.Fatal(err)
+	}
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		b.StopTimer()
-		mi := bytes.NewBufferString(fedoraMountinfo)
-		b.StartTimer()
 		if _, err := getCgroupMountsHelper(subsystems, mi, false); err != nil {
 			b.Fatal(err)
 		}
@@ -396,7 +406,13 @@ func TestIgnoreCgroup2Mount(t *testing.T) {
 		"name=systemd": false,
 	}
 
-	mi := bytes.NewBufferString(cgroup2Mountinfo)
+	mi, err := mountinfo.GetMountsFromReader(
+		bytes.NewBufferString(cgroup2Mountinfo),
+		mountinfo.FSTypeFilter("cgroup"),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
 	cgMounts, err := getCgroupMountsHelper(subsystems, mi, false)
 	if err != nil {
 		t.Fatal(err)
@@ -409,10 +425,8 @@ func TestIgnoreCgroup2Mount(t *testing.T) {
 }
 
 func TestFindCgroupMountpointAndRoot(t *testing.T) {
-	fakeMountInfo := `
-35 27 0:29 / /foo rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices
-35 27 0:29 / /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices
-`
+	fakeMountInfo := `35 27 0:29 / /foo rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices
+35 27 0:29 / /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices`
 	testCases := []struct {
 		cgroupPath string
 		output     string
@@ -421,8 +435,16 @@ func TestFindCgroupMountpointAndRoot(t *testing.T) {
 		{cgroupPath: "", output: "/foo"},
 	}
 
+	mi, err := mountinfo.GetMountsFromReader(
+		bytes.NewBufferString(fakeMountInfo),
+		mountinfo.FSTypeFilter("cgroup"),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	for _, c := range testCases {
-		mountpoint, _, _ := findCgroupMountpointAndRootFromReader(strings.NewReader(fakeMountInfo), c.cgroupPath, "devices")
+		mountpoint, _, _ := findCgroupMountpointAndRootFromMI(mi, c.cgroupPath, "devices")
 		if mountpoint != c.output {
 			t.Errorf("expected %s, got %s", c.output, mountpoint)
 		}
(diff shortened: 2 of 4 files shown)


libcontainer/cgroups/utils_test.go
```diff
@@ -381,10 +396,8 @@ func TestIgnoreCgroup2Mount(t *testing.T) {
 }
 
 func TestFindCgroupMountpointAndRoot(t *testing.T) {
-	fakeMountInfo := `
-35 27 0:29 / /foo rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices
-35 27 0:29 / /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices
-`
+	fakeMountInfo := `35 27 0:29 / /foo rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices
+35 27 0:29 / /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,devices`
```
kolyshkin, line 400: this is fixed to not contain any extra empty lines (which is an issue with the mountinfo parser).

libcontainer/cgroups/v1_utils.go
```diff
@@ -90,6 +94,16 @@ func tryDefaultPath(cgroupPath, subsystem string) string {
 	return path
 }
 
+func ReadCgroupMountinfo() ([]*mountinfo.Info, error) {
+	readMountinfoOnce.Do(func() {
```
kolyshkin, line 98: Here we assume that no new cgroup mounts appear after runc (or anyone who uses libcontainer) start, iow cgroup mounts are performed early during system start, earlier than any container orchestration daemons are started.

libcontainer/cgroups/fs/fs.go
```diff
@@ -133,46 +131,19 @@ func getCgroupRoot() (string, error) {
...
-		return "", err
-	}
-	if root == "" {
+	if len(mi) < 1 {
 		return "", errors.New("no cgroup mount found in mountinfo")
 	}
 
+	// Get the first cgroup mount (e.g. "/sys/fs/cgroup/memory"),
+	// use its parent directory.
+	root := filepath.Dir(mi[0].Root)
```
thaJeztah, line 145: Should we be using `path` instead of `filepath`? I think everything here assumes Linux separators, in which case `path` would be a better fit.

kolyshkin: This question can be applied to lots of code :)

I'd rather be careful and use `filepath.Dir`. I understand `path.Dir` should be somewhat faster, but we're not calling it 10.000 times here.

I think it's a separate issue and can be addressed globally (replace all `filepath.XXX` with `path.XXX` or so).

kolyshkin: Ahh, cpuset.go already uses both `path.Dir` and `filepath.Dir`. Surely this can be unified; let me see...

thaJeztah: Yes, makes sense to do separate. My eye fell on it here, so sorry for the stray comment ☺️

kolyshkin:
Rebased; @AkihiroSuda @mrunalp PTAL

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda (dismissed):

AkihiroSuda:
@cyphar PTAL

kolyshkin:
Rebased on top of #2666 just in case

AkihiroSuda:
@cyphar WDYT?

libcontainer/cgroups/utils_test.go
thaJeztah: @kolyshkin this still a problem? Can we fix that in the parser? (trim leading/trailing whitespace?)

kolyshkin: This was a problem with the test data, which is fixed.

I do not want to add extra trims/checks to the parser as those are entirely useless for real kernel-provided data, and I am trying to keep the parser fast and clean.

AkihiroSuda (approved):
Ping @cyphar

libcontainer/cgroups/utils_test.go
thaJeztah: Thx! I wondered if leading/trailing whitespace could occur in actual situations as well, but it that's not a possibility, then I agree with not stripping

libcontainer/cgroups/v1_utils.go
```diff
@@ -91,6 +95,16 @@ func tryDefaultPath(cgroupPath, subsystem string) string {
 	return path
 }
 
+func ReadCgroupMountinfo() ([]*mountinfo.Info, error) {
```
thaJeztah, line 98:
```suggestion
// ReadCgroupMountinfo returns a list of cgroup mounts for the current running
// process. ReadCgroupMountinfo caches the results in memory, and assumes that
// no new cgroup mounts appear after it was first called.
func ReadCgroupMountinfo() ([]*mountinfo.Info, error) {
```

(perhaps "a list of cgroup v1 mounts"?)

libcontainer/cgroups/v1_utils.go
thaJeztah: Going through this PR again; perhaps put that information in a GoDoc for the function (I notice various exported functions here are not documented; should we have CI checking for that?

(I left a suggestion above)

libcontainer/cgroups/v1_utils.go
thaJeztah: Actually; the name `ReadCgroupMountinfo` is now a bit ambiguous; `Read....` would indicate it actually _reads_ (not caches) `mountinfo`.

Looking a bit further; instead of introducing a new, exported, function;

- This same file has  a `getCgroupMountsV1()`, which is called by `GetCgroupMounts()`
- `getCgroupMountsV1()` performs additional conversion (mapping `mountinfo.Info` to a `libcontainer/cgroups.Mount` using `getCgroupMountsHelper()`)
    - this conversion will be done _every time_ `GetCgroupMounts()` is called
- *side-note*: I see `` and `getCgroupMountsV1()` have an `all` argument, that seems to be always `false` (perhaps it's no longer needed)
- All current uses of `ReadCgroupMountinfo()` outside of this package (`libcontainer/cgroups/fs/cpuset.go` and `libcontainer/cgroups/fs/fs.go` are only interested in `Mountpoint` and `Root`, _both_ of which are also preserved in `libcontainer/cgroups.Mount` (returned by `getCgroupMountsV1 ()`


Perhaps:

- un-export `ReadCgroupMountinfo()`
- remove the `sync.Once()` from `ReadCgroupMountinfo()`
- _export_ `getCgroupMountsV1` (perhaps rename to `GetCgroupV1Mounts()`
    - This makes it a nice parallel to `GetCgroupMounts()` (but only returning `v1` mounts, and thus clearer on what it does)
- remove the `all` argument (if it's indeed no longer needed)
- add the `sync.Once` to `GetCgroupV1Mounts()`, so that the conversions are only done once (if no new cgroups are expected to be added, there should be no need to perform the conversion each time either?)

This also makes the interface somewhat cleaner, as we're not returning an "external" type (`mountinfo.Info`, but instead a type that's defined in this package (`cgroups.Mount`)

thaJeztah (changes requested):
Left a suggestion (let me know what you think)

libcontainer/cgroups/v1_utils.go
kolyshkin: 1. `all` was introduced by https://github.com/opencontainers/runc/pull/1049, and it very vividly demonstrates the major problem with runc: it's a tool as well as a set of packages with lots of users. I have changed runc/libcontainer public API in the past and in some cases the users had to modify their code, and in some cases I needed to revert back.

I will look into cadvisor code wrt using `all` argument.

2. The idea of reusing GetCgroupMounts (and adding sync.Once to it) looks interesting. The problem is, it's somewhat harder to do with `all` argument as it is now. I tried it in the past, and the result was lots of code. What I like about this PR is it is simple, it removes lots of code making things faster at the same time.

3. I don't like yet another exported function either. I think the solution to that is to use `internal`.

kolyshkin: > I will look into cadvisor code wrt using all argument.

https://github.com/google/cadvisor/issues/1461
https://github.com/google/cadvisor/pull/1476

I read all that and thought for some time and I am still not sure how to handle this :(

kolyshkin: So, I have added a comment describing `ReadCgroupMountinfo` as per your earlier suggestion, and I'll take a look at internalizing it later.

libcontainer/cgroups/v1_utils.go
cyphar: This is a fairly reasonable assumption to make, and to be honest the previous code didn't handle cgroup mounts changing in a race-free way either (not to mention that the median lifetime of a runc process is measured in milliseconds) so practically speaking this semantic change doesn't matter.

libcontainer/cgroups/v1_utils.go
cyphar: I agree with doing this API cleanup later.

libcontainer/cgroups/v1_utils.go
```diff
@@ -200,19 +181,18 @@ func getCgroupMountsHelper(ss map[string]bool, mi io.Reader, all bool) ([]Mount,
 		if len(m.Subsystems) > 0 || all {
 			res = append(res, m)
 		}
-	}
-	if err := scanner.Err(); err != nil {
-		return nil, err
+		if numFound >= len(ss) {
+			break
+		}
```
cyphar, line 186: Is this new check necessary? It's also not correct if `all` is `true`.

libcontainer/cgroups/fs/fs.go
cyphar: Yeah we generally use `filepath` for our own sanity, but we can do a switch in a separate PR if there is a strong reason to do so.

cyphar:
Aside from my nit, this LGTM.

libcontainer/cgroups/v1_utils.go
kolyshkin: I haven't changed the original logic here, which was
```go
for scanner.Scan() && numFound < len(ss) {
```

but it's good that you asked, as now I see that this function is not tested with `all` set to true.

kolyshkin: Ah, OK, the `all` logic was added by https://github.com/opencontainers/runc/pull/1049/commits/f55799640143fea4d6ca699ddc2aac72e3e56d7a (done for https://github.com/google/cadvisor/pull/1476) and it correctly avoided incrementing `numFound` in case `all == true`....

.... until https://github.com/opencontainers/runc/pull/1817/commits/5ee0648bfbe037bcf4b36b2d53e59cc91bf42b55 (PR #1817) broke it (since v1.0.0-rc6 and until now).

I am not sure why e.g. google/cadvisor#2309 (that bumps runc to a version with this bug) haven't reintroduced the https://github.com/google/cadvisor/issues/1461. There is _still_ an open issue that looks very similar to it though: https://github.com/google/cadvisor/issues/1843.

Anyway, I am fixing this in https://github.com/opencontainers/runc/pull/2689,
and rebasing this PR on top of that one.

libcontainer/cgroups/v1_utils.go
kolyshkin: OK, I took a fresh look and found a way to unexport `ReadCgroupMountinfo` by not using in the other two commits. The core of this PR stays the same, but the last two commits (getCgroupRoot and cpusets ones) changed a lot. PR description updated with "v2 changes", please re-review @thaJeztah @cyphar

AkihiroSuda:
CI failing

kolyshkin:
Hmm, failure on rootless + idmap on Fedora 33:

```
4034not ok 90 runc exec [tty consolesize]
4035# (in test file /vagrant/tests/integration/tty.bats, line 183)
4036#   `timeout 1 tail --pid="$(head -n 1 pid.txt)" -f /dev/null' failed with status 124
4037# runc spec --rootless (status=0):
4038# 
4039# runc run -d --console-socket /tmp/console.sock test_busybox (status=0):
4040# 
4041# runc state test_busybox (status=0):
4042# {
4043#   "ociVersion": "1.0.2-dev",
4044#   "id": "test_busybox",
4045#   "pid": 72435,
4046#   "status": "running",
4047#   "bundle": "/tmp/busyboxtest",
4048#   "rootfs": "/tmp/busyboxtest/rootfs",
4049#   "created": "2020-11-27T11:03:43.589028179Z",
4050#   "owner": ""
4051# }
4052# runc exec -t --pid-file pid.txt -d --console-socket /tmp/console.sock -p /dev/fd/63 test_busybox (status=0):
4053# 
```
This is the timeout being hit. Does not look related, and I haven't ever seen this before.

Maybe CI is exceptionally slow today? :)

AkihiroSuda:
CI seems to have got stuck, could you push to retrigger CI?

kolyshkin:
> CI seems to have got stuck, could you push to retrigger CI?

Yes, basically all old PRs (that don't have #2690) needs to be rebased :-\

Alternatively, I can unmark those checks as required.

kolyshkin:
Rebased to include #2690

AkihiroSuda:
needs rebase again sorry

kolyshkin:
rebased (for new CI)

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

thaJeztah:
@cyphar still LGTY?

AkihiroSuda:
@cyphar PTAL

cyphar (approved):
LGTM, thanks Kir!

Issue #2473, opened 2020-06-16
CentOS 7 CI: analyze failure of  systemd unit tests and rootless integration tests


AkihiroSuda:
systemd unit tests (not integration tests) and rootless integration tests are failing ong CentOS 7.
We should analyze these failures and fix the failing tests: (see `FIXME` comment lines)

https://github.com/opencontainers/runc/blob/82d2fa4eb069b98af7c34ae525f85fcfb0e40a22/.travis.yml#L38-L53

**EDIT**: The systemd unit tests are analyzed and being fixed in #2501

JFHwang:
Hi, I would like to try looking into this.

AkihiroSuda:
The systemd unit tests are analyzed and being fixed in https://github.com/opencontainers/runc/pull/2501

JFHwang:
I've looked into the cause of failure of the rootless integration tests, and based on what I've seen, it seems to source from a bug in the Vagrant centos7 image. The image actually doesn't allow rootless writes to /sys/fs/cgroup/devices/.../cgroup,procs even though other cgroups are fine. Here is an example I did on a fresh VM.

// I created my own directories
[vagrant@localhost ~]$ sudo su
[root@localhost vagrant]# mkdir /sys/fs/cgroup/blkio/mycgroup/
[root@localhost vagrant]# mkdir /sys/fs/cgroup/devices/mycgroup
[root@localhost vagrant]# chown -R rootless:rootless /sys/fs/cgroup/blkio/mycgroup/
[root@localhost vagrant]# chown -R rootless:rootless /sys/fs/cgroup/devices/mycgroup
[root@localhost vagrant]# exit
exit

// Ran a process for which to use its pid
[vagrant@localhost ~]$ sudo su rootless
[rootless@localhost vagrant]$ cd ~
[rootless@localhost ~]$ touch tmp
[rootless@localhost ~]$ tail -f tmp &
[1] 2512

// Only writes to /sys/fs/cgroup/devices/.../cgroup.procs fail
[rootless@localhost vagrant]$ echo "2512" > /sys/fs/cgroup/blkio/mycgroup/cgroup.procs
[rootless@localhost ~]$ echo "2512" > /sys/fs/cgroup/devices/mycgroup/cgroup.procs
bash: echo: write error: Operation not permitted

I've tried looking into using Vagrant's Centos8 image and this bug does not occur there, so changing the tests to use Centos8 could be a direction to look towards. Any input would be appreciated.

[...]

PR #2501, opened 2020-07-04
fix TestPidsSystemd and TestRunWithKernelMemorySystemd test error

XiaodongLoong:
Test command : `make localunittest`
Test logs:
 ```
FAIL: TestPidsSystemd (0.08s)
    utils_test.go:55: exec_test.go:630: unexpected error: container_linux.go:353: starting container process caused: process_linux.go:326: applying cgroup configuration for process caused: mountpoint for devices not found
FAIL: TestRunWithKernelMemorySystemd (0.07s)
    exec_test.go:713: runContainer failed with kernel memory limit: container_linux.go:353: starting container process caused: process_linux.go:326: applying cgroup configuration for process caused: mkdir : no such file or directory
```
Root cause:
TestPidsSystemd: The  cgroup root path is  `integration/test` when the test container is stating. The devices subsystem can not find under this root path, then casue apply cgroup error. The code show as following,
https://github.com/opencontainers/runc/blob/3cb1909c70f16c36fcbc88f48fab7907eed7b1fc/libcontainer/cgroups/systemd/v1.go#L274

TestRunWithKernelMemorySystemd: The root cause is same as the TestPidsSystemd. But it leads to `mkdir ""`, when system want to set "KernelMemory" property. See the code as  following,
https://github.com/opencontainers/runc/blob/3cb1909c70f16c36fcbc88f48fab7907eed7b1fc/libcontainer/cgroups/systemd/v1.go#L431

Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>

PR #2600, opened 2020-09-24
libcontainer/integration: fix cgroupv1 + systemd tests

kolyshkin:
Commit f57bb2fe3db set it to /sys/fs/cgroup which is somewhat confusing
as now tests create cgroups that look like this on the filesystem:

 - /sys/fs/cgroup/memory/sys/fs/cgroup
 - /sys/fs/cgroup/devices/sys/fs/cgroup
 ... etc.

but it worked!

It seems that some(?) tests were still not working properly with systemd because of invalid configuration.

Similar to what libcontainer/specconv does, add the functionality to set proper cgroup parameters for systemd.
This fixes all the tests.

Fixes: #2473

----

This PR also fixes the root cause of the breakage, which calls for
a separate story to be told. Gather around, kids, let me tell it to you.

Commit a1d5398afa91b0cb10 ("Respect container's cgroup path") added a
`cgroupPath` argument to `FindCgroupMountpoint` to make runc/libcontainer
work in a custom multitenant environment with multiple cgroup mount
points.

It also added passing `c.Path` as an argument to `FindCgroupMountpoint`
for systemd (v1) controller:

https://github.com/opencontainers/runc/blob/0a9af3949f66117a66ee0edf121b0c132b223128/libcontainer/cgroups/systemd/v1.go#L270

This change is wrong for two reasons:

1. systemd controller do not use `c.Path` at all (and `c.Path` is never set
   by `specconv`) -- instead, it uses `Name` and `Parent` fields.

2. `c.Path`, if set, is not absolute -- it is relative to `/sys/fs/cgroup`
   -- but it is used as an absolute path here.

Now, as `c.Path` is never set, the change did not result in any breakage,
and this code sit quietly for some time. The issue might not have been
discovered -- until we started running libcontainer/integration tests
in a CentOS 7 VM, which resulted in a following weird error:

> FAIL: TestPidsSystemd: utils_test.go:55: exec_test.go:630: unexpected error: container_linux.go:353: starting container process caused: process_linux.go:326: applying cgroup configuration for process caused: mountpoint for devices not found

The error was "fixed" in commit f57bb2fe3dbe8 by changing the tests'
cgroups Path to be "/sys/fs/cgroup/". This indeed fixed the issue, but
resulted in creation of cgroup directories like `/sys/fs/cgroup/memory/sys/fs/cgroup`,
`/sys/fs/cgroup/devices/sys/fs/cgroup` and so on.

So, revert the change to avoid confusion.

----

A bonus story for you.

Commit 27d3dd3df35 ("don't fail when subsystem not mounted") added
ignoring "not found" error to enableKmem, and as a result the function
now tries to call Mkdir with an empty path, which results in a weird
error message. For example, this is a failure from a
libcontainer/integration test:

> === RUN   TestRunWithKernelMemorySystemd
>    exec_test.go:704: runContainer failed with kernel memory limit: container_linux.go:370: starting container process caused: process_linux.go:327: applying cgroup configuration for process caused: mkdir : no such file or directory

I am not entirely sure if it is a good idea to silently ignore set
limits, but at least let's fix the error handling.

### DIFF
--- a/libcontainer/integration/template_test.go
+++ b/libcontainer/integration/template_test.go
@@ -1,6 +1,9 @@
 package integration
 
 import (
+	"math/rand"
+	"strconv"
+
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/specconv"
 
@@ -17,8 +20,9 @@ var standardEnvironment = []string{
 const defaultMountFlags = unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NODEV
 
 type tParam struct {
-	rootfs string
-	userns bool
+	rootfs  string
+	userns  bool
+	systemd bool
 }
 
 // newTemplateConfig returns a base template for running a container
@@ -122,7 +126,6 @@ func newTemplateConfig(p *tParam) *configs.Config {
 			{Type: configs.NEWNET},
 		}),
 		Cgroups: &configs.Cgroup{
-			Path: "/sys/fs/cgroup/",
 			Resources: &configs.Resources{
 				MemorySwappiness: nil,
 				Devices:          allowedDevices,
@@ -207,6 +210,16 @@ func newTemplateConfig(p *tParam) *configs.Config {
 			Device:      "cgroup",
 			Flags:       defaultMountFlags | unix.MS_RDONLY,
 		})
+	}
+
+	if p.systemd {
+		id := strconv.FormatUint(rand.Uint64(), 36)
+		config.Cgroups.Name = "test" + id
+		// do not change Parent (see newContainerWithName)
+		config.Cgroups.Parent = "system.slice"
+		config.Cgroups.ScopePrefix = "runc-test"
+	} else {
+		config.Cgroups.Path = "/test/integration"
 	}
 
 	return config
--- a/libcontainer/integration/template_test.go
+++ b/libcontainer/integration/template_test.go
@@ -18,6 +18,7 @@ const defaultMountFlags = unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NODEV
 
 type tParam struct {
 	rootfs string
+	userns bool
 }
 
 // newTemplateConfig returns a base template for running a container
@@ -29,7 +30,7 @@ func newTemplateConfig(p *tParam) *configs.Config {
 	for _, device := range specconv.AllowedDevices {
 		allowedDevices = append(allowedDevices, &device.DeviceRule)
 	}
-	return &configs.Config{
+	config := &configs.Config{
 		Rootfs: p.rootfs,
 		Capabilities: &configs.Capabilities{
 			Bounding: []string{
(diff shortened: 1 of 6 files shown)


kolyshkin:
This is interesting and need to be looked at

```
=== RUN   TestPidsSystemd
time="2020-09-24T02:34:46Z" level=warning msg="signal: terminated"
    utils_test.go:55: exec_test.go:621: unexpected error: container_linux.go:370: starting container process caused: process_linux.go:327: applying cgroup configuration for process caused: mountpoint for devices not found
        
--- FAIL: TestPidsSystemd (0.30s)
--- PASS: TestRunWithKernelMemory (0.28s)
=== RUN   TestRunWithKernelMemorySystemd
time="2020-09-24T02:34:46Z" level=warning msg="signal: killed"
    exec_test.go:704: runContainer failed with kernel memory limit: container_linux.go:370: starting container process caused: process_linux.go:327: applying cgroup configuration for process caused: mkdir : no such file or directory
--- FAIL: TestRunWithKernelMemorySystemd (0.06s)
```

kolyshkin:
pushed a half-finished PR to test intermediate commits

kolyshkin:
@AkihiroSuda @mrunalp PTAL

libcontainer/integration/template_test.go
```diff
@@ -191,4 +199,28 @@ func newTemplateConfig(rootfs string) *configs.Config {
 			},
 		},
 	}
+
+	if p.userns {
+		config.UidMappings = []configs.IDMap{{HostID: 0, ContainerID: 0, Size: 1000}}
+		config.GidMappings = []configs.IDMap{{HostID: 0, ContainerID: 0, Size: 1000}}
+		config.Namespaces = append(config.Namespaces, configs.Namespace{Type: configs.NEWUSER})
+	} else {
```
thaJeztah, line 207: This function will be a long, long list of `ifs` if more options are added; wondering, would it be an idea to use functional arguments for this? `newTemplateConfig(withRootfs(foo), withSystemd())`?

kolyshkin: Frankly speaking I just don't like that paradigm much. I find it harder to read as you have to jump back and forth too much to read the code. It might be useful for APIs with external users, to make the code using it less ugly, but in this case these are internal tests, so the only user is runc.

Actually, let me try rewriting it with functional args, and let's see which one looks better...

kolyshkin: OK, I tried, but I am not sure how to handle "else" cases. Such as this one:

```go
        if p.userns {
                config.UidMappings = []configs.IDMap{{HostID: 0, ContainerID: 0, Size: 1000}}
                config.GidMappings = []configs.IDMap{{HostID: 0, ContainerID: 0, Size: 1000}}
                config.Namespaces = append(config.Namespaces, configs.Namespace{Type: configs.NEWUSER})
        } else {
                config.Mounts = append(config.Mounts, &configs.Mount{
                        Destination: "/sys/fs/cgroup",
                        Device:      "cgroup",
                        Flags:       defaultMountFlags | unix.MS_RDONLY,
                })
        }
```

Surely I can add this mount unconditionally, and make `withUserNS` remove it, but it does not look perfect :-\

So, I'd rather let it stay as is for now.

kolyshkin: Feel free to code the alternative though if you like @thaJeztah

kolyshkin:
Added two more commit fixing the underlying code issues found by tests and hidden by #2501.

I find it quite interesting that bad tests revealed a couple of actual (albeit small) bugs.

AkihiroSuda (approved):

AkihiroSuda:
ping @cyphar @mrunalp

mrunalp (approved):

PR #2606, opened 2020-09-29
libcontainer/intelrdt: use moby/sys/mountinfo

kolyshkin:
It might be a tad slower but it surely more correct and well maintained,
so it's better to use it than rely on a custom implementation which is
kind of hard to get entirely right.

This (together with / after) #2599 removes the last custom mountinfo
parser from the codebase (we had five of those).

### DIFF
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/moby/sys/mountinfo"
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
@Creatone @xiaochenshen PTAL

libcontainer/intelrdt/intelrdt.go
```diff
@@ -241,45 +242,26 @@ func init() {
...
 	}
-	defer f.Close()
-
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		text := s.Text()
-		fields := strings.Split(text, " ")
-		// Safe as mountinfo encodes mountpoints with spaces as \040.
-		index := strings.Index(text, " - ")
-		postSeparatorFields := strings.Fields(text[index+3:])
```
Creatone, line 256: I was working with this code because there was a possibility of accessing elements that not exist and that could cause a panic error.
Thanks, @kolyshkin for fixing it and improve code readability.

I adjusted the unit tests to your code. Could you consider using them?

https://github.com/kolyshkin/runc/compare/intel-rdt...Creatone:creatone/intelrdt-moby-sys-mountinfo

Creatone:
LGTM

libcontainer/intelrdt/intelrdt.go
kolyshkin: Left some comments then decided it's easier to rewrite the code. Hope it's OK with you.

Creatone: I'm fine with this :+1:

AkihiroSuda (approved):

xiaochenshen (commented):
LGTM.

I have run some mount/umount test cases today. It works well and no functional change found.
Thank you @kolyshkin @Creatone for fixing and enhancement for Intel RDT.

thaJeztah (approved):
LGTM

mrunalp (approved):

PR #2607, opened 2020-09-29
libcontainer: move capabilities to separate package

thaJeztah:
I had these changes open in the branch I used for https://github.com/opencontainers/runc/pull/2595. TBH, I don't exactly recall why I made these changes, but I guess because it's slightly cleaner to have it in a separate package (similar to cgroups, devices, apparmor, etc all being separate).

I'm not super-attached to it, so feel free to close if you don't like this change, but thought I'd push it to give the opportunity to review 😅 

@kolyshkin @AkihiroSuda

### DIFF
--- a/libcontainer/capabilities/capabilities.go
+++ b/libcontainer/capabilities/capabilities.go
@@ -1,6 +1,6 @@
 // +build linux
 
-package libcontainer
+package capabilities
 
 import (
 	"fmt"
@@ -24,10 +24,11 @@ func init() {
 	}
 }
 
-func newContainerCapList(capConfig *configs.Capabilities) (*containerCapabilities, error) {
+// New creates a new Caps from the given Capabilities config.
+func New(capConfig *configs.Capabilities) (*Caps, error) {
 	var (
 		err  error
-		caps containerCapabilities
+		caps Caps
 	)
 
 	if caps.bounding, err = capSlice(capConfig.Bounding); err != nil {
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):

kolyshkin:
While at it, we might start using `internal` to make sure no other package (except runc) starts using this package.

I am split between the top-level `internal` and `libcontainer/internal`. In this very case I'm in favor of the latter.

kolyshkin:
> I don't exactly recall why I made these changes

Probably because of `func init()` in libcontainer/capabilities_linux.go, so it won't be always run.

I just found myself creating an alternative to this one: https://github.com/opencontainers/runc/pull/2646

thaJeztah:
Ah, yes, avoiding init is definitely good. I still think that having it in a separate package is clean (not critical of course); slightly on the fence on using "internal" (which would hard restrict external users); guess "libcontainer" is now a bit of a confusing name, as it's no longer meant to use as a library (by others)

AkihiroSuda:
What's current status?

crosbymichael (approved):

AkihiroSuda (dismissed):

thaJeztah:
@kolyshkin good to go?

AkihiroSuda:
Could you rebase to retrigger CI?

thaJeztah:
rebased 👍

AkihiroSuda (approved):

thaJeztah:
@kolyshkin @crosbymichael PTAL; this one good to go?

thaJeztah:
rebased.

@AkihiroSuda @kolyshkin @crosbymichael PTAL; this one good to go?

AkihiroSuda (approved):

AkihiroSuda:
@cyphar @kolyshkin Can we have this in rc93?

cyphar (approved):
LGTM. Regarding `internal` I think we should move basically all of `libcontainer` into `internal` at some point and move the bits of code people actually use to be outside `internal`, so that the default state of `libcontainer` is private.

PR #2610, opened 2020-09-29
checkpoint: setPageServer: use net.SplitHostPort instead of strings.Split

thaJeztah:
strings.Split() isn't the best solution for this, as it will not handle (e.g.) IPv6 addresses.


AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #2611, opened 2020-09-29
libcontainer/cgroups: use const for templates

thaJeztah:
The names of these templates overlapped with some local variables, which made reading the code somewhat confusing.

Changing them to a const, given that these were not updated anywere.


AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #2612, opened 2020-09-29
use string-concatenation instead of sprintf for simple cases

thaJeztah:

### DIFF
--- a/libcontainer/apparmor/apparmor.go
+++ b/libcontainer/apparmor/apparmor.go
@@ -23,9 +23,7 @@ func IsEnabled() bool {
 func setProcAttr(attr, value string) error {
 	// Under AppArmor you can only change your own attr, so use /proc/self/
 	// instead of /proc/<tid>/ like libapparmor does
-	path := fmt.Sprintf("/proc/self/attr/%s", attr)
-
-	f, err := os.OpenFile(path, os.O_WRONLY, 0)
+	f, err := os.OpenFile("/proc/self/attr/"+attr, os.O_WRONLY, 0)
 	if err != nil {
 		return err
 	}
@@ -35,14 +33,13 @@ func setProcAttr(attr, value string) error {
 		return err
 	}
 
-	_, err = fmt.Fprintf(f, "%s", value)
+	_, err = f.WriteString(value)
 	return err
 }
 
 // changeOnExec reimplements aa_change_onexec from libapparmor in Go
 func changeOnExec(name string) error {
-	value := "exec " + name
-	if err := setProcAttr("exec", value); err != nil {
+	if err := setProcAttr("exec", "exec "+name); err != nil {
 		return fmt.Errorf("apparmor failed to apply profile: %s", err)
 	}
 	return nil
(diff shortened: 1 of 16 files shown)


libcontainer/apparmor/apparmor.go
```diff
@@ -35,14 +33,13 @@ func setProcAttr(attr, value string) error {
 		return err
 	}
 
-	_, err = fmt.Fprintf(f, "%s", value)
+	_, err = fmt.Fprint(f, value)
```
kolyshkin, line 36: It's better to use
```go
_, err := `f.WriteString(value)`
```
instead I guess.

libcontainer/cgroups/systemd/common.go
```diff
@@ -307,7 +307,7 @@ func newProp(name string, units interface{}) systemdDbus.Property {
 func getUnitName(c *configs.Cgroup) string {
 	// by default, we create a scope unless the user explicitly asks for a slice.
 	if !strings.HasSuffix(c.Name, ".slice") {
-		return fmt.Sprintf("%s-%s.scope", c.ScopePrefix, c.Name)
+		return c.ScopePrefix + "-" + c.Name
```
kolyshkin, line 310: You lost the `.scope` part; I guess CI failures are because if this

utils_linux.go
```diff
@@ -214,13 +214,13 @@ func createPidFile(path string, process *libcontainer.Process) error {
...
 		tmpDir  = filepath.Dir(path)
-		tmpName = filepath.Join(tmpDir, fmt.Sprintf(".%s", filepath.Base(path)))
+		tmpName = filepath.Join(tmpDir, "."+filepath.Base(path))
 	)
 	f, err := os.OpenFile(tmpName, os.O_RDWR|os.O_CREATE|os.O_EXCL|os.O_SYNC, 0666)
 	if err != nil {
 		return err
 	}
-	_, err = fmt.Fprintf(f, "%d", pid)
+	_, err = fmt.Fprint(f, strconv.Itoa(pid))
```
kolyshkin, line 223: `f.WriteString`

libcontainer/apparmor/apparmor.go
thaJeztah: Ah, yes, didn't think of that 👍, will update

libcontainer/cgroups/systemd/common.go
thaJeztah: 🤦 thanks; good that we have reviews and CI

thaJeztah:
Updated; PTAL 🤗

libcontainer/system/proc.go
```diff
@@ -78,7 +78,7 @@ func GetProcessStartTime(pid int) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("%d", stat.StartTime), nil
+	return strconv.FormatUint(stat.StartTime, 10), nil
```
kolyshkin, line 81: Wonder if `GetProcessStartTime` has any users left? I looked around, can't find any. Maybe we can just remove it.

kolyshkin: Can and should be addressed separately of course.

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

PR #2621, opened 2020-10-01
libcontainer: remove loadConfig which is the unused function

KentaTada:
loadConfig() is the unused function now.

### DIFF
--- a/libcontainer/configs/config_linux_test.go
+++ b/libcontainer/configs/config_linux_test.go
@@ -1,59 +1,12 @@
 package configs
 
 import (
-	"encoding/json"
-	"fmt"
-	"os"
-	"path/filepath"
 	"testing"
 )
 
 var (
 	HookNameList = []HookName{Prestart, CreateRuntime, CreateContainer, StartContainer, Poststart, Poststop}
 )
-
-func loadConfig(name string) (*Config, error) {
-	f, err := os.Open(filepath.Join("../sample_configs", name))
-	if err != nil {
-		return nil, err
-	}
-	defer f.Close()
-
-	var container *Config
-	if err := json.NewDecoder(f).Decode(&container); err != nil {
-		return nil, err
-	}
-
-	// Check that a config doesn't contain extra fields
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

libcontainer/configs/config_linux_test.go
```diff
@@ -1,60 +1,13 @@
...
-	"os"
-	"path/filepath"
 	"testing"
 )
 
 var (
 	HookNameList = []HookName{Prestart, CreateRuntime, CreateContainer, StartContainer, Poststart, Poststop}
 )
 
-func loadConfig(name string) (*Config, error) {
```
thaJeztah, line 15: Looks like this was added a long time ago in https://github.com/opencontainers/runc/commit/43e877e1db2b31abb6c3850fd4e5f3006a705e0f

Only think I'm curious about is if we have a test somewhere that tests loading configs? At a quick glance, the only JSON in the repository is https://github.com/opencontainers/runc/blob/ffe5cdc4f1c87f5346ca69d26ac2fc57e331cf73/tests/integration/config.json, which is an empty file 🤔

KentaTada: loadConfig() uses test configs from the hard-coded directory of `sample_configs` https://github.com/opencontainers/runc/pull/2621/files#diff-096825b702b0fa9613f172dcec5f9c66L16 which has already been deleted.
`sample_configs` was actually removed in 18a2ca375865157091e4fd7a8be49f3a976f1ac9
That test was for nsinit and it is no longer used.

I think we should rewrite the test if we need it after this unused test is deleted.

thaJeztah: Thanks! Yes, if the code it was covering is no longer used, it's fine to remove. I thought I'd left a comment, just in case there's still coverage needed for something like this (but not very familiar with all tests in this repository 😅) 👍

kolyshkin (approved):
Nice! LGTM

PR #2628, opened 2020-10-02
fix some linting issues

thaJeztah:
mostly a test to see if golang-ci-lint picks up these changes in CI as well (not sure if the "nolint" tags work everywhere; they can be somewhat hit and miss

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -277,6 +277,7 @@ func RemovePaths(paths map[string]string) (err error) {
 			}
 		}
 		if len(paths) == 0 {
+			//nolint:ineffassign // done to help garbage collecting: opencontainers/runc#2506
 			paths = make(map[string]string)
 			return nil
 		}
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -432,6 +432,7 @@ func setupRlimits(limits []configs.Rlimit, pid int) error {
 
 const _P_PID = 1
 
+//nolint:structcheck,unused
 type siginfo struct {
 	si_signo int32
 	si_errno int32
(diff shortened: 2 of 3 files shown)


libcontainer/cgroups/utils.go
```diff
@@ -277,6 +277,7 @@ func RemovePaths(paths map[string]string) (err error) {
 			}
 		}
 		if len(paths) == 0 {
+			//nolint:ineffassign // done to help garbage collecting: opencontainers/runc#2506
```
thaJeztah, line 280: Looks like it at least finds them 😂 had a typo here;

```
level=warning msg="[runner/nolint] Found unknown linters in //nolint directives: ineffasign"
```

AkihiroSuda: Can we change the function to return a separate map?

kolyshkin, line 280: I don't remember if it's the recommended practice but the comment inside the commend looks a bit weird...

kolyshkin: Apparently it is... https://golangci-lint.run/usage/false-positives/

kolyshkin: > Can we change the function to return a separate map?

I don't think this makes any sense. No caller of RemovePaths is interested in this map; it is cleaned here in place only to help garbage collector.

If you don't like it we can just drop the `paths = make(map[string]string)` line, but I think it makes sense to keep it.

thaJeztah: Yes, it looks a bit weird

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

PR #2629, opened 2020-10-03
Convert root path to absolute on create command

### LINKED ISSUE #2613: inconsistent relative --root behavior
A relative path can be specified for `--root`, but it is sometimes interpreted as relative to current directory and sometimes as relative to bundle directory:
```
> sudo find .
.
./mybundle
./mybundle/config.json
./myroot
> sudo runc --root myroot create --bundle mybundle a
> sudo find .
.
./mybundle
./mybundle/config.json
./mybundle/myroot
./mybundle/myroot/a
./mybundle/myroot/a/exec.fifo
./mybundle/myroot/a/state.json
./myroot
> sudo runc --root myroot state a
container "a" does not exist
> sudo runc --root mybundle/myroot state a
{
  "ociVersion": "1.0.0",
  "id": "a",
  "pid": 241,
  "status": "created",
...
```

This seems to be because commands like `create` change directory to the bundle before getting the absolute path to the root. And commands like `state` do _not_ change directory to the bundle before getting the absolute path to the root.

Possibly, there should be a `reviseRootDir` similar to the existing `revisePidFile`?

knabben:
Converting relative root path to absolute, so flag is relative to the current directory.

```
runc --root myroot state a
{
  "ociVersion": "1.0.2-dev",
  "id": "a",
  "pid": 403514,
  "status": "created",
  "bundle": "/root/container/mybundle",
  "rootfs": "/root/container/mybundle/rootfs",
  "created": "2020-10-03T02:04:11.259347723Z",
  "owner": ""
}
````

Fixes #2613

### DIFF
--- a/main.go
+++ b/main.go
@@ -141,6 +141,9 @@ func main() {
 				fatal(err)
 			}
 		}
+		if err := reviseRootDir(context); err != nil {
+			return err
+		}
 		return logs.ConfigureLogging(createLogConfig(context))
 	}
 
--- a/utils.go
+++ b/utils.go
@@ -92,6 +92,21 @@ func revisePidFile(context *cli.Context) error {
 	return context.Set("pid-file", pidFile)
 }
 
+// reviseRootDir convert the root to absolute path
+func reviseRootDir(context *cli.Context) error {
+	root := context.GlobalString("root")
+	if root == "" {
+		return nil
+	}
+
+	root, err := filepath.Abs(root)
+	if err != nil {
+		return err
+	}
+
+	return context.GlobalSet("root", root)
+}
+
 // parseBoolOrAuto returns (nil, nil) if s is empty or "auto"
 func parseBoolOrAuto(s string) (*bool, error) {
 	if s == "" || strings.ToLower(s) == "auto" {

### CODE REVIEW

**Comment by knabben** [comment]
@cyphar @kolyshkin PTAL

File: create.go [hunk 1]
```diff
@@ -58,6 +58,9 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 		if err := revisePidFile(context); err != nil {
 			return err
 		}
+		if err := reviseRootDir(context); err != nil {
+			return err
+		}
```
cyphar, line 63: This should go inside the `app.Before` function closure in `main.go` because `--root` is a global flag and we should sanitise it for all subcommands.

cyphar (changes requested):

utils.go
```diff
@@ -92,6 +92,21 @@ func revisePidFile(context *cli.Context) error {
...
+func reviseRootDir(context *cli.Context) error {
+	rootFile := context.GlobalString("root")
+	if rootFile == "" {
+		return nil
+	}
+
+	rootFile, err := filepath.Abs(rootFile)
+	if err != nil {
+		return err
+	}
```
cyphar, line 105: `--root` isn't a file, the variable should be called `root`.

utils.go
```diff
@@ -92,6 +92,21 @@ func revisePidFile(context *cli.Context) error {
 	return context.Set("pid-file", pidFile)
 }
 
+// reviseRootDir convert the root to absolute path
+func reviseRootDir(context *cli.Context) error {
+	rootFile := context.GlobalString("root")
```
cyphar, line 97: `--root` isn't a file, the variable should be called `root`.

knabben: Should *rootFolder* be more appropriated here?

cyphar: Just `root` (or `rootDir` if you really want to use two words).

knabben: ok, kept root in the last commit.

cyphar (approved):
LGTM.

main.go
```diff
@@ -141,6 +141,9 @@ func main() {
 				fatal(err)
 			}
 		}
+		if err := reviseRootDir(context); err != nil {
```
kolyshkin, line 144: We have `ResolveRootfs` in `libcontainer/utils` so maybe use that one here?

kolyshkin: Or maybe that's excessive...

kolyshkin (approved):
LGTM

PR #2633, opened 2020-10-05
Kill processes in cgroup even if process Wait fails

### LINKED ISSUE #2632: Process leak if cgroup init process shares process namespace
The cgroup init process wait function uses Cmd.Wait from os/exec. That Wait function can return a non-nil error if the init process does not complete successfully say if it exits with a non-zero exit code or is terminated.

If the cgroup's init process shares process namespace, all the other processes, if any, in the cgroup will not be killed, resulting in a process leak.

From libcontainer/process_linux.go:
```go
func (p *initProcess) wait() (*os.ProcessState, error) {
	err := p.cmd.Wait()
	if err != nil {
		return p.cmd.ProcessState, err
	}
	// we should kill all processes in cgroup when init is died if we use host PID namespace
	if p.sharePidns {
		signalAllProcesses(p.manager, unix.SIGKILL)
	}
	return p.cmd.ProcessState, nil
}
```

chaitanyabandi:
If the cgroup's init process doesn't complete successfully, Wait returns a
non-nil error. We should still kill all the processes in the cgroup in that
case, if process namespace is shared. Otherwise, it may result in process leak.

Fixes #2632

Signed-off-by: Chaitanya Bandi <kbandi@cs.stonybrook.edu>

### DIFF
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer"
@@ -1657,6 +1658,59 @@ func TestPIDHost(t *testing.T) {
 
 	if actual := strings.Trim(buffers.Stdout.String(), "\n"); actual != l {
 		t.Fatalf("ipc link not equal to host link %q %q", actual, l)
+	}
+}
+
+func TestPIDHostInitProcessWait(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
+	rootfs, err := newRootfs()
+	ok(t, err)
+	defer remove(rootfs)
+
+	pidns := "/proc/1/ns/pid"
+
+	// Run a container with two long-running processes.
+	config := newTemplateConfig(rootfs)
+	config.Namespaces.Add(configs.NEWPID, pidns)
+	container, err := newContainerWithName("test", config)
+	ok(t, err)
+	defer func() {
+		_ = container.Destroy()
+	}()
+
+	process1 := &libcontainer.Process{
+		Cwd:  "/",
+		Args: []string{"sleep", "100"},
+		Env:  standardEnvironment,
+		Init: true,
+	}
+	err = container.Run(process1)
+	ok(t, err)
+
+	process2 := &libcontainer.Process{
+		Cwd:  "/",
+		Args: []string{"sleep", "100"},
+		Env:  standardEnvironment,
+		Init: false,
+	}
+	err = container.Run(process2)
+	ok(t, err)
+
+	// Kill the init process and Wait for it.
+	err = process1.Signal(syscall.SIGKILL)
+	ok(t, err)
+	_, err = process1.Wait()
+	if err == nil {
+		t.Fatal("expected Wait to indicate failure")
+	}
+
+	// The non-init process must've been killed.
+	err = process2.Signal(syscall.Signal(0))
+	if err == nil || err.Error() != "no such process" {
+		t.Fatalf("expected process to have been killed: %v", err)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


kolyshkin:
@chaitanyabandi thanks! I suppose you have a repro... care to add a test case?

chaitanyabandi:
> @chaitanyabandi thanks! I suppose you have a repro... care to add a test case?

Added a test in libcontainer/integration. Thanks for the feedback.

chaitanyabandi:
@mrunalp , @kolyshkin PTAL, thank you.

libcontainer/integration/exec_test.go
```diff
@@ -1660,6 +1661,70 @@ func TestPIDHost(t *testing.T) {
...
+
+	pidns := "/proc/1/ns/pid"
+
+	config := newTemplateConfig(rootfs)
+	config.Namespaces.Add(configs.NEWPID, pidns)
+	container, err := newContainerWithName("test", config)
+	ok(t, err)
+	defer func() {
+		err := container.Destroy()
+		ok(t, err)
```
kolyshkin, line 1681: You can't make any checks in defer, as the results of those will be discarded.

So, I guess, just
```go
defer container.Destroy()
```

libcontainer/integration/exec_test.go
```diff
@@ -1660,6 +1661,70 @@ func TestPIDHost(t *testing.T) {
...
+	}()
+
+	// Run a container with two long-running processes.
+	stdinR1, stdinW1, err := os.Pipe()
+	ok(t, err)
+	process1 := &libcontainer.Process{
+		Cwd:   "/",
+		Args:  []string{"cat"},
```
kolyshkin, line 1689: Easier to do with something like `sleep 100`, this way you don't need those pipes.

libcontainer/integration/exec_test.go
chaitanyabandi: Using defer container.Destroy() caused the go linter to fail saying the return value of Destroy() was ignored. Hence I added the check. Should that linter error be ignored?

libcontainer/integration/exec_test.go
chaitanyabandi: I agree. Will post another change.

libcontainer/integration/exec_test.go
```diff
@@ -1660,6 +1661,70 @@ func TestPIDHost(t *testing.T) {
...
+	ok(t, err)
+	_, err = process1.Wait()
+	if err == nil {
+		t.Fatal("expected Wait to indicate failure")
+	}
+
+	// The non-init process must've been killed.
+	err = process2.Signal(syscall.Signal(0))
+	if err == nil || err.Error() != "no such process" {
+		t.Fatalf("expected process to have been killed: %q", err)
```
kolyshkin, line 1724:
```suggestion
		t.Fatalf("expected process to have been killed: %v", err)
```
As otherwise when it fails it prints
```
    exec_test.go:1721: expected process to have been killed: %!q(<nil>)
```

kolyshkin:
Looks good except a few nits, and I have checked the test case works (i.e. fails without the fix).

libcontainer/integration/exec_test.go
```diff
@@ -1660,6 +1661,57 @@ func TestPIDHost(t *testing.T) {
...
+	defer remove(rootfs)
+
+	pidns := "/proc/1/ns/pid"
+
+	// Run a container with two long-running processes.
+	config := newTemplateConfig(rootfs)
+	config.Namespaces.Add(configs.NEWPID, pidns)
+	container, err := newContainerWithName("test", config)
+	ok(t, err)
+	defer container.Destroy()
```
kolyshkin, line 1680: I think
```suggestion
	defer _ = container.Destroy()
```

mrunalp (approved):

kolyshkin (approved):
LGTM

Issue #2573, opened 2020-09-03
rootless runc exec causes panic: cannot statfs cgroup root (When /sys is not mounted)


coryb:
Seeing a panic from `runc exec` when run with rootlesskit.  

Here is the full stack, runc built from latest [master](https://github.com/opencontainers/runc/commit/09ddc63afdde16d5fb859a1d3ab010bd45f08497)
```
panic: cannot statfs cgroup root

goroutine 1 [running, locked to thread]:
github.com/opencontainers/runc/libcontainer/cgroups.IsCgroup2UnifiedMode.func1()
        /go/src/github.com/opencontainers/runc/libcontainer/cgroups/utils.go:45 +0xc5
sync.(*Once).doSlow(0x56030c2f2dd8, 0x56030c00ccc0)
        /usr/lib/go-1.15/src/sync/once.go:66 +0xee
sync.(*Once).Do(...)
        /usr/lib/go-1.15/src/sync/once.go:57
github.com/opencontainers/runc/libcontainer/cgroups.IsCgroup2UnifiedMode(0x20)
        /go/src/github.com/opencontainers/runc/libcontainer/cgroups/utils.go:42 +0x5a
github.com/opencontainers/runc/libcontainer.cgroupfs(0xc0001ae120, 0xc00000f000, 0x56030c021f00, 0x56030c29da60)
        /go/src/github.com/opencontainers/runc/libcontainer/factory_linux.go:121 +0x28
github.com/opencontainers/runc/libcontainer.Cgroupfs(...)
        /go/src/github.com/opencontainers/runc/libcontainer/factory_linux.go:134
github.com/opencontainers/runc/libcontainer.New(0x0, 0x0, 0x0, 0x0, 0x0, 0xc0001aad88, 0x56030c021f60, 0xc000253618, 0x0)
        /go/src/github.com/opencontainers/runc/libcontainer/factory_linux.go:199 +0x165
main.glob..func6(0xc00012e840, 0x0, 0xc000059b60)
        /go/src/github.com/opencontainers/runc/init.go:42 +0x39
github.com/urfave/cli.HandleAction(0x56030bf6e900, 0x56030c00dd18, 0xc00012e840, 0xc00012e840, 0x0)
        /go/src/github.com/opencontainers/runc/vendor/github.com/urfave/cli/app.go:523 +0xfd
[...]

PR #2634, opened 2020-10-05
don't panic when /sys/fs/cgroup is missing for rootless

coryb:
Fixes issue #2573

Downgrade the panic to a debug message when /sys/fs/cgroup is missing.  This issue is preventing runc exec from working within [buildkit](https://github.com/moby/buildkit/issues/1716) when running rootless (where /sys is not mounted at all).

Steps to reproduce:
```
$ mkdir ~/mycontainer
$ cd ~/mycontainer
$ mkdir rootfs
$ docker export $(docker create busybox) | tar -C rootfs -xvf -
$ runc spec --rootless
<edit config.json to remove /sys mount>
$ runc --root /tmp/runc run mycontainerid
# 
```
now in separate shell:
```
cd ~/mycontainer
runc --root /tmp/runc exec --tty mycontainerid /bin/sh
# 
```

A potentially better solution is to change the signature from:
```
func IsCgroup2UnifiedMode() bool
```
to:
```
func IsCgroup2UnifiedMode() (bool, error)
```
so that we can manage the error more directly in rootless mode, but the diff for this change would be much larger, since the function is used quite a lot.  I can update is this direction is preferred.

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
+	"github.com/opencontainers/runc/libcontainer/system"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
@@ -34,8 +35,15 @@ var (
 func IsCgroup2UnifiedMode() bool {
 	isUnifiedOnce.Do(func() {
 		var st unix.Statfs_t
-		if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
-			panic("cannot statfs cgroup root")
+		err := unix.Statfs(unifiedMountpoint, &st)
+		if err != nil {
+			if os.IsNotExist(err) && system.RunningInUserNS() {
+				// ignore the "not found" error if running in userns
+				logrus.WithError(err).Debugf("%s missing, assuming cgroup v1", unifiedMountpoint)
+				isUnified = false
+				return
+			}
+			panic(fmt.Sprintf("cannot statfs cgroup root: %s", err))
 		}
 		isUnified = st.Type == unix.CGROUP2_SUPER_MAGIC
 	})


libcontainer/cgroups/utils.go
```diff
@@ -34,7 +34,9 @@ func IsCgroup2UnifiedMode() bool {
 	isUnifiedOnce.Do(func() {
 		var st unix.Statfs_t
 		if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
-			panic("cannot statfs cgroup root")
+			logrus.WithError(err).Debug("cannot statfs cgroup root")
```
kolyshkin, line 37: 1. It should be `Info` not `Debug` I think as it's pretty major.
2. Amend the message to say "assuming cgroup v1".

kolyshkin: 3. Maybe add a TODO item to change the default to cgroupv2 somewhere in 2024 :)

kolyshkin:
@AkihiroSuda PTAL

libcontainer/cgroups/utils.go
coryb: yeah, this is partially why I am leaning towards changing the signature to return an `error`, otherwise users will get this message 100% of the time exec'ing to a container via buildkit with rootless.

AkihiroSuda (approved):
Thanks

libcontainer/cgroups/utils.go
coryb: Does something like this seem more reasonable to you?
```
		err := unix.Statfs(unifiedMountpoint, &st)
		if err != nil && os.IsNotExist(err) {
			// /sys/fs/cgroup not found, likely rootless
			logrus.WithError(err).Debugf("%s missing, assuming cgroup v1", unifiedMountpoint)
			isUnified = false
			return
		} else if err != nil {
			panic(fmt.Sprintf("cannot statfs cgroup root: %s", err))
		}
```
This would preserve the panic behavior for all error scenarios except where the `/sys/fs/cgroup` path simply does not exist (which is a normal situation under rootless)

AkihiroSuda: Also better to panic when `os.IsNotExist && !system.RunningInUserNS()`

coryb: updated the check to only ignore (w/ debug msg) errors for NotExists while RunningInUserNS

AkihiroSuda (approved):

libcontainer/cgroups/utils.go
```diff
@@ -34,8 +35,14 @@ var (
 func IsCgroup2UnifiedMode() bool {
 	isUnifiedOnce.Do(func() {
 		var st unix.Statfs_t
-		if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
-			panic("cannot statfs cgroup root")
+		err := unix.Statfs(unifiedMountpoint, &st)
+		if err != nil && os.IsNotExist(err) && system.RunningInUserNS() {
+			// /sys/fs/cgroup not found, likely rootless running in user NS
```
kolyshkin, line 40: I would not say "likely" as at this time we're pretty sure about it.

libcontainer/cgroups/utils.go
```diff
@@ -34,8 +35,14 @@ var (
...
+			logrus.WithError(err).Debugf("%s missing, assuming cgroup v1", unifiedMountpoint)
+			isUnified = false
+			return
+		} else if err != nil {
+			panic(fmt.Sprintf("cannot statfs cgroup root: %s", err))
```
kolyshkin, line 45: Can I suggest changing the error handling structure slightly, for better readability? Something like

```suggestion
		if err != nil {
		  if os.IsNotExist(err) && system.RunningInUserNS() {
		    // ignore the "not found" error if running in userns
		  	logrus.WithError(err).Debugf("%s missing, assuming cgroup v1", unifiedMountpoint)
  			isUnified = false
  			return
      }
      panic(fmt.Sprintf("cannot statfs cgroup root: %s", err))
```

kolyshkin (changes requested):
couple of nits

libcontainer/cgroups/utils.go
coryb: Yeah, thanks, that is better.  Updated.

coryb:
The test failure ([timeout](https://travis-ci.org/github/opencontainers/runc/jobs/733525597#L3007)) seems unrelated to my change, and Travis is not allowing me to retry the build.

coryb:
I squashed and forced push, tests passed after the push this time.  I think I have addressed all the feedback, so hopefully we can get this merged in.

AkihiroSuda (approved):

AkihiroSuda:
ping @kolyshkin @cyphar @mrunalp

AkihiroSuda:
@kolyshkin @cyphar @mrunalp PTAL

cyphar (approved):
LGTM.

PR #2635, opened 2020-10-06
libct/cg: introduce and use fscommon.OpenFile

kolyshkin:
_This is part III from https://github.com/opencontainers/runc/pull/2598, preparing the code to use `openat2()`_

Move the functionality of opening a cgroup file into a separate
function, `OpenFile`, which, similar to `ReadFile` and `WriteFile`,
use separate dir and file arguments.
    
Change `ReadFile` and `WriteFile` to rely on `OpenFile`, and use lower-level
read and write instead of ones from ioutil.
    
It changes the semantics of `WriteFile` a bit -- it no longer uses
`O_CREAT` flag. This is good for real cgroup as there is no need to try
creating the files in there, but can potentially break WriteFile users
-- previously, `EPERM` error was returned for non-existing files, and
now it's `ENOENT`.
    
This also breaks the fs/fs2 unit tests since they write to pseudo-cgroup
files inside a test directory (not to a real cgroup fs), and now
`fscommon.WriteFile` do not create or truncate files, so we have to add a
variable that is set by the unit tests to emulate the old behavior.

Use the new interface where possible.

### DIFF
--- a/libcontainer/cgroups/fscommon/open.go
+++ b/libcontainer/cgroups/fscommon/open.go
@@ -1,0 +1,33 @@
+package fscommon
+
+import (
+	"os"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+	"github.com/pkg/errors"
+)
+
+var (
+	// Set to true by fs unit tests
+	TestMode bool
+)
+
+// OpenFile opens a cgroup file in a given dir with given flags.
+// It is supposed to be used for cgroup files only.
+func OpenFile(dir, file string, flags int) (*os.File, error) {
+	if dir == "" {
+		return nil, errors.Errorf("no directory specified for %s", file)
+	}
+	mode := os.FileMode(0)
+	if TestMode && flags&os.O_WRONLY != 0 {
+		// "emulate" cgroup fs for unit tests
+		flags |= os.O_TRUNC | os.O_CREATE
+		mode = 0o600
+	}
+	path, err := securejoin.SecureJoin(dir, file)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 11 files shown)


libcontainer/cgroups/fscommon/open.go
```diff
@@ -0,0 +1,33 @@
...
+
+// OpenFile opens a cgroup file in a given dir with given flags.
+// It is supposed to be used for cgroup files only.
+func OpenFile(dir, file string, flags int) (*os.File, error) {
+	if dir == "" {
+		return nil, errors.Errorf("no directory specified for %s", file)
+	}
+	mode := os.FileMode(0)
+	if TestMode && flags&os.O_WRONLY != 0 {
+		// "emulate" cgroup fs for unit tests
```
AkihiroSuda, line 23: Shouldn't this create a clean dir for unit tests?

kolyshkin: I am not sure what you mean. A test dir is created by unit tests, and then they use WriteFile to populate the mock-cgroupfs. Here's a copy/paste from the commit message:

> It changes the semantics of WriteFile a bit -- it no longer uses
O_CREAT flag. This is good for real cgroup as there is no need to try
creating the files in there, but can potentially break WriteFile users
-- previously, EPERM error was returned for non-existing files, and
now it's ENOENT.

> This also breaks the fs/fs2 unit tests since they write to pseudo-cgroup
files inside a test directory (not to a real cgroup fs), and now
fscommon.WriteFile do not create or truncate files, so we have to add a
variable that is set by the unit tests.

AkihiroSuda (approved):

mrunalp (approved):

PR #2625, opened 2020-10-01
libcontainer/intelrdt: modify the incorrect file mode

KentaTada:
This file mode is unintentional apparently.

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

PR #2636, opened 2020-10-06
libcontainer/intelrdt: adjust the file mode

KentaTada:
and also changes the file mode value in accordance with default.

This commit is related to https://github.com/opencontainers/runc/pull/2625

### DIFF
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -413,7 +413,7 @@ func writeFile(dir, file, data string) error {
 	if dir == "" {
 		return fmt.Errorf("no such directory for %s", file)
 	}
-	if err := ioutil.WriteFile(filepath.Join(dir, file), []byte(data+"\n"), 0700); err != nil {
+	if err := ioutil.WriteFile(filepath.Join(dir, file), []byte(data+"\n"), 0o600); err != nil {
 		return fmt.Errorf("failed to write %v to %v: %v", data, file, err)
 	}
 	return nil
@@ -520,7 +520,7 @@ func WriteIntelRdtTasks(dir string, pid int) error {
 
 	// Don't attach any pid if -1 is specified as a pid
 	if pid != -1 {
-		if err := ioutil.WriteFile(filepath.Join(dir, IntelRdtTasks), []byte(strconv.Itoa(pid)), 0700); err != nil {
+		if err := ioutil.WriteFile(filepath.Join(dir, IntelRdtTasks), []byte(strconv.Itoa(pid)), 0o600); err != nil {
 			return fmt.Errorf("failed to write %v to %v: %v", pid, IntelRdtTasks, err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

libcontainer/intelrdt/intelrdt.go
```diff
@@ -397,7 +397,7 @@ func writeFile(dir, file, data string) error {
 	if dir == "" {
 		return fmt.Errorf("no such directory for %s", file)
 	}
-	if err := ioutil.WriteFile(filepath.Join(dir, file), []byte(data+"\n"), 0700); err != nil {
+	if err := ioutil.WriteFile(filepath.Join(dir, file), []byte(data+"\n"), 0o644); err != nil {
```
kolyshkin, line 400: Strictly speaking, file mode is not used here as the file is *not* being created, it's already there.

But I'd prefer `0o600` here.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -504,7 +504,7 @@ func WriteIntelRdtTasks(dir string, pid int) error {
 
 	// Don't attach any pid if -1 is specified as a pid
 	if pid != -1 {
-		if err := ioutil.WriteFile(filepath.Join(dir, IntelRdtTasks), []byte(strconv.Itoa(pid)), 0700); err != nil {
+		if err := ioutil.WriteFile(filepath.Join(dir, IntelRdtTasks), []byte(strconv.Itoa(pid)), 0o644); err != nil {
```
kolyshkin, line 507: ditto

kolyshkin (dismissed):
a nit

libcontainer/intelrdt/intelrdt.go
cyphar: `0o` isn't commonly used in Go (or C) for octal values, just use `0644` -- same goes for the other changes.

KentaTada: @kolyshkin 

Thank you for reviewing the commits.

>Strictly speaking, file mode is not used here as the file is not being created, it's already there.

Yes.

>But I'd prefer 0o600 here.

As you know, This file is automatically created as `0o644` and it is not changed even if we specify `0o600`.
So I think we don't need to change the default `0o644` .

And is it ok to change the format from `0o` to `0`? If yes, I will modify the format again.
Either is fine for me. But I just want to unify the format as we discussed in below
https://github.com/opencontainers/runc/pull/2625#issuecomment-702627945

kolyshkin: > 0o isn't commonly used in Go (or C) for octal values, just use 0644 -- same goes for the other changes.

It's a new thing as of Go 1.13, and I think it makes sense to use it as it is more explicit.

AkihiroSuda:
ping @KentaTada

KentaTada:
I'm sorry for the late reply.
The remaining discussion item is whether we should adopt the format of "0o".
I think the format of "0o" is easy to read and reduces the bugs like https://github.com/opencontainers/runc/pull/2625

kolyshkin:
@KentaTada can you please rebase to kick ci?

KentaTada:
@kolyshkin I rebased it.

kolyshkin:
Needs a rebase since we moved away from travis-ci

AkihiroSuda (approved):

AkihiroSuda:
@kolyshkin PTAL

cyphar (approved):
LGTM.

Issue #2627, opened 2020-10-02
Review and address golang-ci linting errors 


thaJeztah:
We recently added golang-ci-lint, which runs on pull requests and on master.

Currently, looks like the linter on PR's only checks the changed lines, but found various issues in the existing code on master (see https://github.com/opencontainers/runc/pull/2618#issuecomment-702531748); the linter picked up the problem in https://github.com/opencontainers/runc/runs/1190924966

Looks like that output is a good list of things to work on;

```
Error: `userFromOS` is unused (deadcode)
Error: `groupFromOS` is unused (deadcode)
Error: `loadConfig` is unused (deadcode)
Error: `removePath` is unused (deadcode)
Error: `wildcard` is unused (deadcode)
Error: Error return value of `os.Mkdir` is not checked (errcheck)
Error: Error return value of `os.Symlink` is not checked (errcheck)
Error: Error return value of `console.ClearONLCR` is not checked (errcheck)
Error: Error return value of `io.Copy` is not checked (errcheck)
Error: Error return value of `io.Copy` is not checked (errcheck)
Error: Error return value of `io.Copy` is not checked (errcheck)
Error: Error return value of `cmd.Process.Kill` is not checked (errcheck)
Error: Error return value of `cmd.Wait` is not checked (errcheck)
Error: Error return value of `fHook.Run` is not checked (errcheck)
Error: Error return value of `join` is not checked (errcheck)
Error: Error return value of `os.Symlink` is not checked (errcheck)
Error: Error return value of `dbusConnection.ResetFailedUnit` is not checked (errcheck)
Error: Error return value of `dbusConnection.ResetFailedUnit` is not checked (errcheck)
Error: Error return value of `unix.Unmount` is not checked (errcheck)
Error: Error return value of `criuClientCon.CloseWrite` is not checked (errcheck)
Error: Error return value of `Cgroupfs` is not checked (errcheck)
Error: Error return value of `unix.Unmount` is not checked (errcheck)
Error: Error return value of `m.Freeze` is not checked (errcheck)
Error: Error return value of `p.wait` is not checked (errcheck)
Error: Error return value of `p.cmd.Wait` is not checked (errcheck)
Error: Error return value of `p.cmd.Wait` is not checked (errcheck)
Error: Error return value of `p.cmd.Wait` is not checked (errcheck)
Error: Error return value of `p.manager.Destroy` is not checked (errcheck)
[...]

PR #2641, opened 2020-10-10
2627 fix "some of" golangci-lint problems

ehsundar:
this request is related to [this issue](https://github.com/opencontainers/runc/issues/2627)

### DIFF
--- a/contrib/cmd/recvtty/recvtty.go
+++ b/contrib/cmd/recvtty/recvtty.go
@@ -65,19 +65,19 @@ func bail(err error) {
 	os.Exit(1)
 }
 
-func handleSingle(path string, noStdin bool) error {
+func handleSingle(path string, noStdin bool) (retErr error) {
 	// Open a socket.
-	ln, err := net.Listen("unix", path)
-	if err != nil {
-		return err
+	ln, retErr := net.Listen("unix", path)
+	if retErr != nil {
+		return retErr
 	}
 	defer ln.Close()
 
 	// We only accept a single connection, since we can only really have
 	// one reader for os.Stdin. Plus this is all a PoC.
-	conn, err := ln.Accept()
-	if err != nil {
-		return err
+	conn, retErr := ln.Accept()
+	if retErr != nil {
+		return
 	}
 	defer conn.Close()
 
@@ -90,20 +90,20 @@ func handleSingle(path string, noStdin bool) error {
 		return fmt.Errorf("failed to cast to unixconn")
 	}
 
-	socket, err := unixconn.File()
-	if err != nil {
-		return err
+	socket, retErr := unixconn.File()
+	if retErr != nil {
+		return
 	}
 	defer socket.Close()
 
 	// Get the master file descriptor from runC.
-	master, err := utils.RecvFd(socket)
-	if err != nil {
-		return err
+	master, retErr := utils.RecvFd(socket)
+	if retErr != nil {
+		return
 	}
-	c, err := console.ConsoleFromFile(master)
-	if err != nil {
-		return err
+	c, retErr := console.ConsoleFromFile(master)
+	if retErr != nil {
+		return
 	}
 	if err := console.ClearONLCR(c.Fd()); err != nil {
 		return err
@@ -111,21 +111,32 @@ func handleSingle(path string, noStdin bool) error {
 
 	// Copy from our stdio to the master fd.
 	quitChan := make(chan struct{})
+	errChan := make(chan error)
 	go func() {
-		io.Copy(os.Stdout, c)
+		_, err := io.Copy(os.Stdout, c)
+		if err != nil {
+			errChan <- err
+		}
 		quitChan <- struct{}{}
 	}()
 	if !noStdin {
 		go func() {
-			io.Copy(c, os.Stdin)
+			_, err := io.Copy(c, os.Stdin)
+			if err != nil {
+				errChan <- err
+			}
 			quitChan <- struct{}{}
 		}()
 	}
 
 	// Only close the master fd once we've stopped copying.
-	<-quitChan
-	c.Close()
-	return nil
+	select {
+	case retErr = <-errChan:
+		return retErr
+	case <-quitChan:
+		retErr = c.Close()
+		return retErr
+	}
 }
 
 func handleNull(path string) error {
--- a/contrib/cmd/recvtty/recvtty.go
+++ b/contrib/cmd/recvtty/recvtty.go
@@ -61,7 +61,7 @@ terminals:
 )
 
 func bail(err error) {
-	fmt.Fprintf(os.Stderr, "[recvtty] fatal error: %v\n", err)
+	_, _ = fmt.Fprintf(os.Stderr, "[recvtty] fatal error: %v\n", err)
 	os.Exit(1)
 }
 
@@ -71,7 +71,9 @@ func handleSingle(path string, noStdin bool) (retErr error) {
 	if retErr != nil {
 		return retErr
 	}
-	defer ln.Close()
+	defer func() {
+		_ = ln.Close()
+	}()
 
 	// We only accept a single connection, since we can only really have
 	// one reader for os.Stdin. Plus this is all a PoC.
@@ -79,10 +81,12 @@ func handleSingle(path string, noStdin bool) (retErr error) {
 	if retErr != nil {
 		return
 	}
-	defer conn.Close()
+	defer func() {
+		_ = conn.Close()
+	}()
 
 	// Close ln, to allow for other instances to take over.
-	ln.Close()
+	_ = ln.Close()
 
 	// Get the fd of the connection.
 	unixconn, ok := conn.(*net.UnixConn)
@@ -94,7 +98,9 @@ func handleSingle(path string, noStdin bool) (retErr error) {
 	if retErr != nil {
 		return
 	}
-	defer socket.Close()
+	defer func() {
+		_ = socket.Close()
+	}()
 
 	// Get the master file descriptor from runC.
 	master, retErr := utils.RecvFd(socket)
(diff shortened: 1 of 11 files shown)


contrib/cmd/recvtty/recvtty.go
```diff
@@ -61,28 +61,35 @@ terminals:
 )
 
 func bail(err error) {
-	fmt.Fprintf(os.Stderr, "[recvtty] fatal error: %v\n", err)
+	_, _ = fmt.Fprintf(os.Stderr, "[recvtty] fatal error: %v\n", err)
 	os.Exit(1)
 }
 
-func handleSingle(path string, noStdin bool) error {
+func handleSingle(path string, noStdin bool) (err error) {
```
kolyshkin, line 68: I'd suggest calling it `retErr` or so, to distinguish between errors.

contrib/cmd/recvtty/recvtty.go
```diff
@@ -61,28 +61,35 @@ terminals:
...
 	// Open a socket.
 	ln, err := net.Listen("unix", path)
 	if err != nil {
 		return err
 	}
-	defer ln.Close()
+	defer func() {
+		err = ln.Close()
```
kolyshkin, line 75: So, what would happen if `net.Listen` fails? Please be careful.

contrib/cmd/recvtty/recvtty.go
```diff
@@ -61,28 +61,35 @@ terminals:
...
 
 	// We only accept a single connection, since we can only really have
 	// one reader for os.Stdin. Plus this is all a PoC.
 	conn, err := ln.Accept()
 	if err != nil {
 		return err
 	}
-	defer conn.Close()
+	defer func() {
+		err = conn.Close()
```
kolyshkin, line 85: same here

contrib/cmd/recvtty/recvtty.go
```diff
@@ -61,28 +61,35 @@ terminals:
 )
 
 func bail(err error) {
-	fmt.Fprintf(os.Stderr, "[recvtty] fatal error: %v\n", err)
+	_, _ = fmt.Fprintf(os.Stderr, "[recvtty] fatal error: %v\n", err)
```
kolyshkin, line 64: This change is not inline with what the commit message says. Please split your commits carefully.

libcontainer/configs/config_test.go
```diff
@@ -149,7 +149,10 @@ func TestFuncHookRun(t *testing.T) {
 		return nil
 	})
 
-	fHook.Run(state)
+	err := fHook.Run(state)
+	if err != nil {
+		t.Errorf("Unexpected error %+v", err)
```
kolyshkin, line 154:
```suggestion
		t.Errorf("Unexpected error running hook: %+v", err)
```

kolyshkin:
Thank you for PR. I left a few suggestions.

In general, I would like it to be split into smaller focused patches (I see you already did that), and make sure your code is correct.

contrib/cmd/recvtty/recvtty.go
cyphar: `Err` is what I usually use.

contrib/cmd/recvtty/recvtty.go
cyphar: There are several changes of this form, and while `Close` *can* fail very few Go programs handle this case successfully all the time and you would need to write the same boilerplate for each case:

```
defer func() {
  if Err != nil {
    Err = foo.Close()
  }
}()
```

Aside from adding a bunch of boilerplate to lots of functions, the second question is when will `close(2)` fail and does it make sense to deal with this? Most close failures are actually when the filesystem had an error writing to disk, but this mechanism in Linux is unpredictable and ultimately there's not much to do if that's why `close(2)` failed (very often the failure happens during a later `fsync(2)`). Most other `close(2)` failures don't apply to most Go programs. Not to mention that the worst thing that happens is that we keep a file descriptor open in a program which takes about 100ms to run -- and we can't do anything about it anyway because the semantics of `close(2)` are that you cannot safely attempt to re-close a file after calling `close(2)`.

We can do it, but I don't see the benefit and if the only reason for this change is to make a CI lint happy, we should turn off the CI lint IMHO.

contrib/cmd/recvtty/recvtty.go
kolyshkin: I was wrong here (except that you rewriting the error).

contrib/cmd/recvtty/recvtty.go
kolyshkin: Yeah, the alternative that is not too much fuss and makes static checkers happy is same as in `(void)write();` in C, i.e.
```go
defer func() {
  _ := foo.Close()
}
```

contrib/cmd/recvtty/recvtty.go
ehsundar: No need to metion the defer part will run after func returns. so if any error happens in closing, the `retErr` will be refilled. the question is: is it important to you to handle this kind of errors in the caller function?

ehsundar: There is also another problem: if the func returns because of another error, and at the same time, the close also fails, the former error will be lost! while the closing error is bound (in my opinion) I believe the original error is more significant ... tell me what to do?

ehsundar: I decided to ignore closing error...

contrib/cmd/recvtty/recvtty.go
ehsundar: Thanks for your expressive description. As a boilerplate, I may suggest this one: 

```go
defer func() {
  if err = foo.Close(); err != nil && retErr != nil {
    retErr = err
  }
}()
```

because we should try to close() anyways. but its much more than its needed (I think), and as @kolyshkin said, I'll try to simply ignore the closing error across my MR

ehsundar:
Sorry for being late, I didn't receive any email or notif. I should have configured github before. thanks

ehsundar:
I tried to apply your suggestions. and I tried to split my commits a little more. sorry for making mistakes. It's the first time I contribute to an open source repo. I really didn't expect this precise review. thank you.

contrib/cmd/recvtty/recvtty.go
```diff
@@ -90,42 +90,53 @@ func handleSingle(path string, noStdin bool) error {
...
+	if retErr != nil {
+		return
 	}
 	if err := console.ClearONLCR(c.Fd()); err != nil {
 		return err
 	}
 
 	// Copy from our stdio to the master fd.
 	quitChan := make(chan struct{})
+	errChan := make(chan error)
```
kolyshkin, line 120: OK it seems that in this case a single channel is enough, there is no need to have two.

contrib/cmd/recvtty/recvtty.go
```diff
@@ -65,19 +65,19 @@ func bail(err error) {
...
 
-func handleSingle(path string, noStdin bool) error {
+func handleSingle(path string, noStdin bool) (retErr error) {
 	// Open a socket.
-	ln, err := net.Listen("unix", path)
-	if err != nil {
-		return err
+	ln, retErr := net.Listen("unix", path)
+	if retErr != nil {
+		return retErr
```
kolyshkin, line 72: You did't have to change any of this.

contrib/cmd/recvtty/recvtty.go
```diff
@@ -65,19 +65,19 @@ func bail(err error) {
...
 	defer ln.Close()
 
 	// We only accept a single connection, since we can only really have
 	// one reader for os.Stdin. Plus this is all a PoC.
-	conn, err := ln.Accept()
-	if err != nil {
-		return err
+	conn, retErr := ln.Accept()
+	if retErr != nil {
+		return
```
kolyshkin, line 82: same here

contrib/cmd/recvtty/recvtty.go
```diff
@@ -90,42 +90,53 @@ func handleSingle(path string, noStdin bool) error {
 		return fmt.Errorf("failed to cast to unixconn")
 	}
 
-	socket, err := unixconn.File()
-	if err != nil {
-		return err
+	socket, retErr := unixconn.File()
+	if retErr != nil {
+		return
```
kolyshkin, line 99: same here

contrib/cmd/recvtty/recvtty.go
```diff
@@ -90,42 +90,53 @@ func handleSingle(path string, noStdin bool) error {
...
+	master, retErr := utils.RecvFd(socket)
+	if retErr != nil {
+		return
 	}
-	c, err := console.ConsoleFromFile(master)
-	if err != nil {
-		return err
+	c, retErr := console.ConsoleFromFile(master)
+	if retErr != nil {
+		return
```
kolyshkin, line 112: same here

contrib/cmd/recvtty/recvtty.go
```diff
@@ -90,42 +90,53 @@ func handleSingle(path string, noStdin bool) error {
...
 	// Copy from our stdio to the master fd.
 	quitChan := make(chan struct{})
+	errChan := make(chan error)
 	go func() {
-		io.Copy(os.Stdout, c)
+		_, err := io.Copy(os.Stdout, c)
+		if err != nil {
+			errChan <- err
+		}
 		quitChan <- struct{}{}
```
kolyshkin, line 126: IOW this could be simple
```go
_, err := io.Copy(os.Stdout, c)
chan <- err
```

and the code that's receiving this will check for nil

contrib/cmd/recvtty/recvtty.go
kolyshkin: IOW this can stay `err` -- the only place to use `retErr` is the defer.

kolyshkin:
@ehsundar please let us know if you're going to keep working on that or not

AkihiroSuda:
@kolyshkin Would you be interested in carrying this PR? 🙏

kolyshkin:
> @kolyshkin Would you be interested in carrying this PR?

Looks like I have no choice :) so yes, I will :)

AkihiroSuda:
Thanks!

kolyshkin:
* https://github.com/opencontainers/runc/pull/2694
* https://github.com/opencontainers/runc/pull/2695

kolyshkin:
Not using any patches from this one since it's easier to fix from scratch, thus closing. The issue #2627 is still open though.

kolyshkin:
superseded by the above PRs

PR #2646, opened 2020-10-13
libcontainer: rm init() from capabilities_linux.go

kolyshkin:
Instead, use sync.Once to init the map when needed.

This should give a small speedup to operations which do not require
filling in the capabilities.

### DIFF
--- a/libcontainer/capabilities_linux.go
+++ b/libcontainer/capabilities_linux.go
@@ -5,6 +5,7 @@ package libcontainer
 import (
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/syndtr/gocapability/capability"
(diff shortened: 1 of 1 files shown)


dims:
+1 to this new pattern for initialization.

cyphar:
To be honest, I don't like this because `func init()` is kind of the recommended way of doing these kinds of initialisations in Go. And every container setup will end up calling `initCapMap` anyway.

You could argue that in principle operations other than `runc create` *could* be *sightly* faster -- but how much of a difference are we talking about? Not to mention that `sync.Once` actually takes locks internally, so in the realm of micro-optimisations this actually makes container setup ever so slightly slower.

kolyshkin:
My concern was mostly about runc init which should be a small binary but it's not, partially because it includes libcontainer which does things like this.

PR #2647, opened 2020-10-13
rootfs: handle nested procfs mounts for MS_MOVE

### LINKED ISSUE #2639: process_linux.go:297: applying cgroup configuration for process: read-only file system
I'm trying to run `busybox` container in Colab, however I've got the following error:

```
WARN[0000] signal: killed                               
ERRO[0000] container_linux.go:349: starting container process caused "process_linux.go:297: applying cgroup configuration for process caused \"mkdir /sys/fs/cgroup/cpuset/container1: read-only file system\"" 
container_linux.go:349: starting container process caused "process_linux.go:297: applying cgroup configuration for process caused \"mkdir /sys/fs/cgroup/cpuset/container1: read-only file system\""
```

Here are the steps:

1. Install Docker and runc:

```
%%shell
curl -s https://download.docker.com/linux/static/stable/x86_64/docker-19.03.9.tgz | tar vxz --strip=1 -C /usr/local/bin/
wget -cqO /usr/local/bin/runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc92/runc.amd64 && chmod +x /usr/local/bin/runc
docker --version
runc --version
```

2. Extract Busybox container into `busybox/rootfs`:

```
%%shell
dockerd -b none --iptables=0 -l warn &
sleep 1
mkdir -pv busybox/rootfs
docker export $(docker create busybox) | tar -C busybox/rootfs -xf -
kill $(jobs -p)
```

3. Run:

```
%%shell
cd busybox
runc spec --rootless
runc run --no-new-keyring --no-pivot container1
```

Demo: https://colab.research.google.com/drive/19hVpEODrL8kb7KvyWrA9vE6Pd7ZKMA4G#scrollTo=VhgKc1a6zMTq

Is there any way to run the container having read-only access to cgroup configuration?

cyphar:
In a case where the host /proc mount has already been overmounted, the
MS_MOVE handling would get ENOENT when trying to hide (for instance)
"/proc/bus" because it had already hidden away "/proc". This revealed
two issues in the previous implementation of this hardening feaure:

1. No checks were done to make sure the mount was a "full" mount (it is
   a mount of the root of the filesystem), but the kernel doesn't permit
   a non-full mount to be converted to a full mount (for reference, see
   mnt_already_visible). This just removes extra busy-work during setup.

2. ENOENT was treated as a critical error, even though it actually
   indicates the mount doesn't exist and thus isn't a problem. A more
   theoretically pure solution would be to store the set of mountpoints
   to be hidden and only ignore the error if an ancestor directory of
   the current mountpoint was already hidden, but that would just add
   complexity with little justification.

In addition, better document the reasoning behind this logic so that
folks aren't confused when looking at it.

Fixes #2639
Fixes: 28a697cce3e4 ("rootfs: umount all procfs and sysfs with --no-pivot")

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -809,24 +809,46 @@ func pivotRoot(rootfs string) error {
 }
 
 func msMoveRoot(rootfs string) error {
+	// Before we move the root and chroot we have to mask all "full" sysfs and
+	// procfs mounts which exist on the host. This is because while the kernel
+	// has protections against mounting procfs if it has masks, when using
+	// chroot(2) the *host* procfs mount is still reachable in the mount
+	// namespace and the kernel permits procfs mounts inside --no-pivot
+	// containers.
+	//
+	// Users shouldn't be using --no-pivot except in exceptional circumstances,
+	// but to avoid such a trivial security flaw we apply a best-effort
+	// protection here. The kernel only allows a mount of a pseudo-filesystem
+	// like procfs or sysfs if there is a *full* mount (the root of the
+	// filesystem is mounted) without any other locked mount points covering a
+	// subtree of the mount.
+	//
+	// So we try to unmount (or mount tmpfs on top of) any mountpoint which is
+	// a full mount of either sysfs or procfs (since those are the most
+	// concerning filesystems to us).
 	mountinfos, err := mountinfo.GetMounts(func(info *mountinfo.Info) (skip, stop bool) {
-		skip = false
-		stop = false
-		// Collect every sysfs and proc file systems, except those under the container rootfs
-		if (info.FSType != "proc" && info.FSType != "sysfs") || strings.HasPrefix(info.Mountpoint, rootfs) {
+		// Collect every sysfs and procfs filesystem, except for those which
+		// are non-full mounts or are inside the rootfs of the container.
+		if info.Root != "/" ||
+			(info.FSType != "proc" && info.FSType != "sysfs") ||
+			strings.HasPrefix(info.Mountpoint, rootfs) {
 			skip = true
-			return
 		}
 		return
 	})
 	if err != nil {
 		return err
 	}
-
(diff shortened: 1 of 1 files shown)


giuseppe (dismissed):
just a nit, otherwise the change makes sense to me.  Thanks!

libcontainer/rootfs_linux.go
```diff
@@ -809,24 +809,45 @@ func pivotRoot(rootfs string) error {
...
+	// concerning filesystems to us).
 	mountinfos, err := mountinfo.GetMounts(func(info *mountinfo.Info) (skip, stop bool) {
-		skip = false
-		stop = false
-		// Collect every sysfs and proc file systems, except those under the container rootfs
-		if (info.FSType != "proc" && info.FSType != "sysfs") || strings.HasPrefix(info.Mountpoint, rootfs) {
+		// Collect every sysfs and procfs filesystem, except for those which
+		// are non-full mounts or are inside the rootfs of the container.
+		if (info.FSType != "proc" && info.FSType != "sysfs") ||
+			info.Root != "/" || strings.HasPrefix(info.Mountpoint, rootfs) {
```
kolyshkin, line 833: nit: I'd place `info.Root != "/"` first as this is the easiest/fastest check. It will also fit the description better this way.

```suggestion
		if info.Root != "/" ||
		  (info.FSType != "proc" && info.FSType != "sysfs") ||
			strings.HasPrefix(info.Mountpoint, rootfs) {
```

kolyshkin:
LGTM except for a single nit.

libcontainer/rootfs_linux.go
cyphar: Fixed.

kolyshkin (approved):
LGTM

kolyshkin:
@AkihiroSuda PTAL

mrunalp (approved):

PR #2653, opened 2020-10-17
libcontainer/intelrdt: IntelRdtManager refactoring

xiaochenshen:
This is a set of patches for IntelRdtManager refactoring. 
It is similar to part of the cgroup refactoring in #2387.

No functional change.
Please review commits one by one.

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -149,11 +149,7 @@ func RootlessCgroupfs(l *LinuxFactory) error {
 // create and manage Intel RDT resources (e.g., L3 cache, memory bandwidth).
 func IntelRdtFs(l *LinuxFactory) error {
 	l.NewIntelRdtManager = func(config *configs.Config, id string, path string) intelrdt.Manager {
-		return &intelrdt.IntelRdtManager{
-			Config: config,
-			Id:     id,
-			Path:   path,
-		}
+		return intelrdt.NewManager(config, id, path)
 	}
 	return nil
 }
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -165,18 +165,18 @@ type Manager interface {
 }
 
 // This implements interface Manager
-type IntelRdtManager struct {
+type intelRdtManager struct {
 	mu     sync.Mutex
-	Config *configs.Config
-	Id     string
-	Path   string
+	config *configs.Config
+	id     string
+	path   string
 }
 
 func NewManager(config *configs.Config, id string, path string) Manager {
-	return &IntelRdtManager{
-		Config: config,
-		Id:     id,
-		Path:   path,
+	return &intelRdtManager{
+		config: config,
+		id:     id,
+		path:   path,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


xiaochenshen:
@kolyshkin @AkihiroSuda @Creatone 
Could you help review this PR at your convenience? Thank you!

Creatone:
@xiaochenshen Can it wait till Monday? I'm on holidays now.

xiaochenshen:
> 
> 
> @xiaochenshen Can it wait till Monday? I'm on holidays now.

@Creatone No problem. Thank you for help! 😃

Creatone (approved):
LGTM

kolyshkin (approved):
LGTM

Creatone:
@AkihiroSuda

AkihiroSuda (approved):

kolyshkin:
This broke cadvisor but the fix is trivial (https://github.com/google/cadvisor/pull/2712/commits/48e7375af579bd966fb60ee9116a85cbfd1afa1f#diff-c1ad83c3e8484f4f134e3fb0eff9666f7b19db23d0c20bd5476920ff1bcd9709)

By the way @xiaochenshen @Creatone why do we have to specify empty config, like this:

```go
mgr := intelrdt.NewManager(&configs.Config{IntelRdt: &configs.IntelRdt{}}, id, resctrlPath)
```

Can we do something like
```go
func NewManager(config *configs.Config, id string, path string) Manager {
        if config == nil {
                config = &configs.Config{IntelRdt: &configs.IntelRdt{}}
        }
        return &intelRdtManager{
                config: config,
                id:     id,
                path:   path,
        }
}
```

or it does not make any sense?

xiaochenshen:
> 
> This broke cadvisor but the fix is trivial ([google/cadvisor@48e7375#diff-c1ad83c3e8484f4f134e3fb0eff9666f7b19db23d0c20bd5476920ff1bcd9709](https://github.com/google/cadvisor/commit/48e7375af579bd966fb60ee9116a85cbfd1afa1f#diff-c1ad83c3e8484f4f134e3fb0eff9666f7b19db23d0c20bd5476920ff1bcd9709))
> 

@kolyshkin Thank you for fixing this.

> By the way @xiaochenshen @Creatone why do we have to specify empty config, like this:
> 
> ```go
> mgr := intelrdt.NewManager(&configs.Config{IntelRdt: &configs.IntelRdt{}}, id, resctrlPath)
> ```
> 

@kolyshkin The reason why we have to specify empty config:
In #1615 fixing, we add check `if Config.IntelRdt == nil` to make sure that GetStats() doesn't return error unexpectedly in this
https://github.com/opencontainers/runc/blob/27227a9358b54c253e3dad85cfe532a256b88e00/libcontainer/intelrdt/intelrdt.go#L587-L591

special case in `runc update`:
If `intelRdt` is not specified in original configuration when starting the container, we could update `L3CacheSchema` and `MemBwSchema` in `config.IntelRdt` through `runc update` command later in runtime.

See the code and comment here:
https://github.com/opencontainers/runc/blob/27227a9358b54c253e3dad85cfe532a256b88e00/update.go#L310-L328

> Can we do something like

Yes. I think so.

> 
> ```go
> func NewManager(config *configs.Config, id string, path string) Manager {
>         if config == nil {
> ```
Maybe:
`if config.IntelRdt == nil {` ?

> ```
>                 config = &configs.Config{IntelRdt: &configs.IntelRdt{}}
>         }
>         return &intelRdtManager{
>                 config: config,
>                 id:     id,
>                 path:   path,
>         }
> }
>```
> or it does not make any sense?

PR #2654, opened 2020-10-20
Revert "rootfs: handle nested procfs mounts for MS_MOVE" + add a test

AkihiroSuda:
Reverts opencontainers/runc#2647 due to a security issue
Reopens https://github.com/opencontainers/runc/issues/2639

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -809,46 +809,24 @@ func pivotRoot(rootfs string) error {
 }
 
 func msMoveRoot(rootfs string) error {
-	// Before we move the root and chroot we have to mask all "full" sysfs and
-	// procfs mounts which exist on the host. This is because while the kernel
-	// has protections against mounting procfs if it has masks, when using
-	// chroot(2) the *host* procfs mount is still reachable in the mount
-	// namespace and the kernel permits procfs mounts inside --no-pivot
-	// containers.
-	//
-	// Users shouldn't be using --no-pivot except in exceptional circumstances,
-	// but to avoid such a trivial security flaw we apply a best-effort
-	// protection here. The kernel only allows a mount of a pseudo-filesystem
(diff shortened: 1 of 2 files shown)


cyphar:
This is actually a bug in `github.com/moby/sys/mountinfo` -- `info.Root` is always `""`. I've already submitted a fix (https://github.com/moby/sys/pull/50).

libcontainer/rootfs_linux.go
```diff
@@ -809,46 +809,24 @@ func pivotRoot(rootfs string) error {
...
-	// filesystem is mounted) without any other locked mount points covering a
-	// subtree of the mount.
-	//
-	// So we try to unmount (or mount tmpfs on top of) any mountpoint which is
-	// a full mount of either sysfs or procfs (since those are the most
-	// concerning filesystems to us).
 	mountinfos, err := mountinfo.GetMounts(func(info *mountinfo.Info) (skip, stop bool) {
-		// Collect every sysfs and procfs filesystem, except for those which
-		// are non-full mounts or are inside the rootfs of the container.
-		if info.Root != "/" ||
```
cyphar, line 832: It's overkill to revert it -- a simpler patch is to just remove this `info.Root != "/"` clause which won't cause us to revert the fix (the `ENOENT` check later in the patch is sufficient to fix that bug).

AkihiroSuda: Let's revert it first and then revendor https://github.com/moby/sys/pull/50 when it is merged

AkihiroSuda:
Added a new commit: `test: add "runc run --no-pivot must not expose bare /proc"`

libcontainer/rootfs_linux.go
kolyshkin: I agree with @cyphar here -- for the reason of having a slightly less dramatic git history.

cyphar:
#2655 removes just the `info.Root` check and cherry-picks 729e786.

PR #2655, opened 2020-10-21
rootfs: handle github.com/moby/sys/mountinfo deficiency

cyphar:
github.com/moby/sys/mountinfo doesn't fill the Root field of
mountinfo.Info when calling filter functions, meaning that the change in
commit b8bf57281298 ("rootfs: handle nested procfs mounts for MS_MOVE")
regressed the security hardening.

While this is being fixed upstream in [1], just remove this info.Root
optimisation because the ENOENT check is actually sufficient to fix the
original issue.

[1]: https://github.com/moby/sys/pull/50

Closes #2654
Reported-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

### DIFF
--- a/tests/integration/no_pivot.bats
+++ b/tests/integration/no_pivot.bats
@@ -1,0 +1,22 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+	teardown_busybox
+	setup_busybox
+}
+
+function teardown() {
+	teardown_busybox
+}
+
+@test "runc run --no-pivot must not expose bare /proc" {
+	requires root
+
+	update_config '.process.args |= ["unshare", "-mrpf", "sh", "-euxc", "mount -t proc none /proc && echo h > /proc/sysrq-trigger"]'
+
+	runc run --no-pivot test_no_pivot
+	[ "$status" -eq 1 ]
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

tests/integration/no_pivot.bats
```diff
@@ -0,0 +1,22 @@
...
+}
+
+@test "runc run --no-pivot must not expose bare /proc" {
+	requires root
+
+	update_config '.process.args |= ["unshare", "-mrpf", "sh", "-euxc", "mount -t proc none /proc && echo h > /proc/sysrq-trigger"]'
+
+	runc run --no-pivot test_no_pivot
+	[ "$status" -eq 1 ]
+	[[ ${lines[*]} == *"mount: permission denied"* ]]
```
kolyshkin, line 21: nit:
```suggestion
	[[ "$output" == *"mount: permission denied"* ]]
```

kolyshkin (changes requested):
a nit in the test case, otherwise lgtm

AkihiroSuda:
closing as suggsted in https://github.com/opencontainers/runc/pull/2657#issuecomment-714022906

PR #2658, opened 2020-10-22
tests/integration: rm excessive run use; fix exec --preserve-fds test

kolyshkin:
### 1. rm excessive `run` use

Bats' run should only be used when we want to check both the command
output and its non-zero exit status.

Otherwise, we can rely on implicit exit code check (as the tests are
run with set -e), or use if, etc.

### 2. fix "runc exec --preserve-fds" test case

1. Fix the check, mostly by changing `cat hello` to `echo hello`,
   and checking for `"hello"` (whole string) rather than `*"hello"*` (substring).

   Previously, `cat hello` generated `cat: hello: no such file or directory`
   error message, which `run` added to `$output` and
   so the check for $output containing `hello` worked!

2. Simplify the test by not using the subshell and the `run`.
   The only catch is, fd 3 is used by bats itself, so we have to use
   fd 4 and thus --preserve-fds 2.

### 3. simplify `teardown_running_container`

In its current form, it is complicated, unreliable, and error prone (TOCTOU etc)

Using `runc delete -f` will kill and remove any container, running or not,
and it won't error if a container with a given name does not exist.

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -43,8 +43,7 @@ function check_pipes() {
 	echo Ping >&${in_w}
 	exec {in_w}>&-
 	exec {out_w}>&-
-	run cat <&${out_r}
-	[ "$status" -eq 0 ]
+	output=$(cat <&${out_r})
 	[[ "${output}" == *"ponG Ping"* ]]
 }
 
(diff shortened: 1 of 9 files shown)


giuseppe (dismissed):
LGTM.  Not a maintainer but I've introduced the issue fixed by the second commit

mrunalp (approved):

kolyshkin:
Rebased on top of #2666 to fix CI

kolyshkin:
```
=== RUN   TestExecInTTY
2820    execin_test.go:349: unexpected carriage-return in output
2821--- FAIL: TestExecInTTY (0.10s)
```

Hmm I think I've seen it before...

kolyshkin:
restarted CI

kolyshkin:
@AkihiroSuda @mrunalp PTAL

mrunalp (approved):

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda (approved):

PR #2660, opened 2020-10-23
libcontainer/configs: improve unit tests

mauriciovasquezbernal:
Test that CommandHook actually executes a new process with the given env
variables, parameters and json state.

This commit also solves an issue with the previous approach that was calling
'os.Exit(0)' failing to signal test failures.

cc @alban 

Fixes: #2765

### DIFF
--- a/libcontainer/configs/config_test.go
+++ b/libcontainer/configs/config_test.go
@@ -3,6 +3,7 @@ package configs_test
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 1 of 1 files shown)


kolyshkin:
Can you please rebase this one since #2661 is merged?

mauriciovasquezbernal:
While studying these tests in more detail I found it was possible to improve them. 
I updated the PR to do so, it also fixes the original problem I opened this PR for.

kolyshkin:
Need to be rebased again once https://github.com/opencontainers/runc/pull/2666 is merged

kolyshkin:
> Need to be rebased again once #2666 is merged

@mauriciovasquezbernal please rebase

mauriciovasquezbernal:
@kolyshkin rebased!

libcontainer/configs/config_test.go
```diff
@@ -160,18 +161,53 @@ func TestCommandHookRun(t *testing.T) {
...
+
+	tmpFile, err := ioutil.TempFile("", "runcTest")
+	if err != nil {
+		t.Fatalf("Failed to create tmp file: %v", err)
+	}
+	defer os.Remove(tmpFile.Name())
+	if _, err := tmpFile.Write([]byte(verifyCommand)); err != nil {
+		t.Fatalf("Failed to write tmp file: %v", err)
+	}
+	if err := tmpFile.Chmod(0777); err != nil {
```
kolyshkin, line 198: `0777` looks overly permissive. `0755` should be good enough and not overly restrictive, although even `0700` should work just fine.

kolyshkin: Also, there is no need to create a temp file, you can just run `sh -` and provide the script to stdin. Maybe this will simplify the code a bit

mauriciovasquezbernal: I'm not sure how I could pass the script to stdin. The hook already receives the state using stdin. I could also use `sh -c 'script'` to avoid creating a temp file but passing the parameters gets tricky. I think it's more clear to create a temp script and directly execute it. 

I update the code to use 0700 and simplified it a bit.

mauriciovasquezbernal:
@kolyshkin ping?

kolyshkin (dismissed):
LGTM but needs to be rebased (again) to kick new CI

mauriciovasquezbernal:
@kolyshkin rebased and CI is green! :)

kolyshkin (approved):
LGTM

AkihiroSuda (approved):
LGTM, but it might be better to have `set -eu`

PR #2666, opened 2020-10-29
shfmt mounts.bats to pass `make validate`

AkihiroSuda:
CI on the master has been broken due to this.

https://travis-ci.org/github/opencontainers/runc/builds/739817745

Not sure why we couldn't catch this before merging it.

PR #2667, opened 2020-10-29
tests: fix shfmt

mauriciovasquezbernal:
Fix format to avoid "make validate" failing. Ref https://github.com/opencontainers/runc/pull/2570.

It's not my intention to make a lot of noise with this lint related fixes but we're trying to make travis pass on a PR we're preparing and we prefer to summit these changes separately. 

cc @alban


mauriciovasquezbernal:
Forget that, there is already a PR open for it https://github.com/opencontainers/runc/pull/2666.

PR #2605, opened 2020-09-28
libct/cgroups: switch to fscommon.{Read,Write}File

kolyshkin:
This switches all code under libcontainer/cgroups that reads or writes cgroup files to use appropriate functions from `fscommon`. This used to be part of #2598.

Besides, it slightly refactors `fs/cpuset.go` code.

PR #2604, opened 2020-09-28
libct/cgroups/fscommon: nits

kolyshkin:
This is mostly error message fixes, doc improvements, and other minor cleanups.

This used to be a part of #2598

PR #2668, opened 2020-10-29
libcontainer/cgroups/fscommon: add openat2 support

kolyshkin:
_(based on previous work in PRs #2604, #2605, #2635; originally part of #2598)_

In case `openat2` is available, it will be used to guarantee
that we're not accessing anything other than cgroupfs[2] files.

In cases when `openat2` is not available, or when cgroup has a
non-standard prefix (not `"/sys/fs/cgroup"`, which might theoretically
be the case on some very old installs and/or very custom systems),
fall back to using `securejoin` + `os.Open` like we did before.

### DIFF
--- a/libcontainer/cgroups/fscommon/open.go
+++ b/libcontainer/cgroups/fscommon/open.go
@@ -2,15 +2,55 @@ package fscommon
 
 import (
 	"os"
+	"strings"
+	"sync"
 
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	cgroupfsDir    = "/sys/fs/cgroup"
+	cgroupfsPrefix = cgroupfsDir + "/"
 )
 
 var (
 	// Set to true by fs unit tests
 	TestMode bool
+
+	cgroupFd     int = -1
+	prepOnce     sync.Once
+	prepErr      error
+	resolveFlags uint64
 )
+
+func prepareOpenat2() error {
+	prepOnce.Do(func() {
+		fd, err := unix.Openat2(-1, cgroupfsDir, &unix.OpenHow{
+			Flags: unix.O_DIRECTORY | unix.O_PATH})
+		if err != nil {
+			prepErr = &os.PathError{Op: "openat2", Path: cgroupfsDir, Err: err}
+			return
+		}
+		var st unix.Statfs_t
+		if err = unix.Fstatfs(fd, &st); err != nil {
+			prepErr = &os.PathError{Op: "statfs", Path: cgroupfsDir, Err: err}
+			return
+		}
+
+		cgroupFd = fd
+
+		resolveFlags = unix.RESOLVE_BENEATH | unix.RESOLVE_NO_MAGICLINKS
+		if st.Type == unix.CGROUP2_SUPER_MAGIC {
+			// cgroupv2 has a single mountpoint and no "cpu,cpuacct" symlinks
+			resolveFlags |= unix.RESOLVE_NO_XDEV | unix.RESOLVE_NO_SYMLINKS
+		}
+
+	})
+
+	return prepErr
+}
 
 // OpenFile opens a cgroup file in a given dir with given flags.
 // It is supposed to be used for cgroup files only.
@@ -24,6 +64,29 @@ func OpenFile(dir, file string, flags int) (*os.File, error) {
 		flags |= os.O_TRUNC | os.O_CREATE
 		mode = 0o600
 	}
+	reldir := strings.TrimPrefix(dir, cgroupfsPrefix)
+	if len(reldir) == len(dir) { // non-standard path, old system?
+		return openWithSecureJoin(dir, file, flags, mode)
+	}
+	if prepareOpenat2() != nil {
+		return openWithSecureJoin(dir, file, flags, mode)
+	}
+
+	relname := reldir + "/" + file
+	fd, err := unix.Openat2(cgroupFd, relname,
+		&unix.OpenHow{
+			Resolve: resolveFlags,
+			Flags:   uint64(flags) | unix.O_CLOEXEC,
+			Mode:    uint64(mode),
+		})
+	if err != nil {
+		return nil, &os.PathError{Op: "openat2", Path: dir + "/" + file, Err: err}
+	}
+
+	return os.NewFile(uintptr(fd), cgroupfsPrefix+relname), nil
+}
+
+func openWithSecureJoin(dir, file string, flags int, mode os.FileMode) (*os.File, error) {
 	path, err := securejoin.SecureJoin(dir, file)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


kolyshkin:
Rebased on top of #2666 to fix CI

libcontainer/cgroups/fscommon/open.go
```diff
@@ -24,6 +64,29 @@ func OpenFile(dir, file string, flags int) (*os.File, error) {
 		flags |= os.O_TRUNC | os.O_CREATE
 		mode = 0o600
 	}
+	reldir := strings.TrimPrefix(dir, cgroupfsPrefix)
```
AkihiroSuda, line 67: Seems to require dir = filepath.Clean(filepath.Abs(dir))

AkihiroSuda (changes requested):

libcontainer/cgroups/fscommon/open.go
kolyshkin: I would not do it here. This function is called from cgroup managers (with one exception of WriteCgroupFile from runc/libcontainer/utils), and I think we can validate the path in there.

But that is a valid concern, thanks, I'll audit the callers.

kolyshkin: In fact what happens here is we fall back to old implementation of using securejoin in case `dir` does not start with `/sys/fs/cgroup/`. In case it starts with that, and open2 is available, we use open2 which is guaranteed to not open anything out of /sys/fs/cgroup, so I think we're good here.

libcontainer/cgroups/fscommon/open.go
```diff
@@ -2,16 +2,56 @@ package fscommon
...
+		if err = unix.Fstatfs(fd, &st); err != nil {
+			// XXX: make unix.ENOSYS a special case?
+			prepErr = &os.PathError{Op: "statfs", Path: cgroupfsDir, Err: err}
+			return
+		}
+
+		cgroupFd = fd
+
+		if st.Type == unix.CGROUP2_SUPER_MAGIC {
+			resolveFlags = unix.RESOLVE_NO_XDEV | unix.RESOLVE_NO_SYMLINKS
```
cyphar, line 45: We could add `RESOLVE_BENEATH` here too.

libcontainer/cgroups/fscommon/open.go
cyphar: I think it would be nice if we made this a /little/ less cryptic -- maybe we should make the common interface actually pass the subsystem, cgroup path, and target path -- but aside from that (which we discussed in the other PR and decided against) I don't think this is a problem.

libcontainer/cgroups/fscommon/open.go
kolyshkin: makes sense; done

libcontainer/cgroups/fscommon/open.go
kolyshkin: @AkihiroSuda so I looked at it one more time -- with the exception of runc/libcontainer/utils.WriteCgroupFile, this function is only used from cgroup managers, and all of them already make sure the path is absolute and resolved (and, in case of cgroupv2, always starts with /sys/fs/cgroup/ which is what we check for here).

Even if it's not, the worst case scenario is we fall back to using securejoin.

So, I don't see it's an issue.

kolyshkin:
CI failure is shfmt that went crazy; caused by https://github.com/mvdan/sh/issues/628.

Restarted travis.

kolyshkin:
> CI failure is shfmt that went crazy; caused by mvdan/sh#628.

Should be fixed by https://github.com/opencontainers/runc/pull/2671

libcontainer/cgroups/fscommon/open.go
cyphar: Yeah, agreed.

libcontainer/cgroups/fscommon/open.go
```diff
@@ -2,16 +2,57 @@ package fscommon
...
+
+		cgroupFd = fd
+
+		if st.Type == unix.CGROUP2_SUPER_MAGIC {
+			resolveFlags = unix.RESOLVE_BENEATH |
+				unix.RESOLVE_NO_XDEV | unix.RESOLVE_NO_SYMLINKS
+		} else {
+			// cgroupfs v1 is a set of separate mounts
+			resolveFlags = unix.RESOLVE_BENEATH
+		}
```
cyphar, line 50:
```go
resolveFlags = unix.RESOLVE_BENEATH
if st.Type == unix.CGROUP2_SUPER_MAGIC {
	// cgroupv2 has a single mountpoint and no "cpu,cpuacct" symlinks
	resolveFlags |= unix.RESOLVE_NO_XDEV | unix.RESOLVE_NO_SYMLINKS
}
```

kolyshkin: That was my initial idea, but at the time I thought it hurts readability.

Now, looking at your code, it doesn't seem like it.

libcontainer/cgroups/fscommon/open.go
```diff
@@ -2,16 +2,56 @@ package fscommon
...
+	resolveFlags uint64
 )
 
+func prepareOpenat2() error {
+	prepOnce.Do(func() {
+		fd, err := unix.Openat2(-1, cgroupfsDir, &unix.OpenHow{
+			Flags: unix.O_DIRECTORY | unix.O_PATH})
+		if err != nil {
+			prepErr = &os.PathError{Op: "openat2", Path: cgroupfsDir, Err: err}
+			return
```
kolyshkin, line 34: The only thing that concerns me is `prepErr` is actually never used (except for comparing it to `nil`), and I am on the verge about adding a warning here (which will be printed once), something like

```go
if err != unix.ENOTSUP {
        logrus.Warnf("prepareOpenat2: %s", prepErr)
}
```

libcontainer/cgroups/fscommon/open.go
```diff
@@ -2,16 +2,56 @@ package fscommon
...
+		}
+		var st unix.Statfs_t
+		if err = unix.Fstatfs(fd, &st); err != nil {
+			prepErr = &os.PathError{Op: "statfs", Path: cgroupfsDir, Err: err}
+			return
```
kolyshkin, line 39: ... and here:
```go
        logrus.Warnf("prepareOpenat2: %s", prepErr)
```

kolyshkin: Added a commit to log a warning/debug message.

libcontainer/cgroups/fscommon/open.go
kolyshkin: added the second commit

kolyshkin:
@AkihiroSuda @cyphar @mrunalp can you please re-review? I think I have addressed all concerns and review comments

kolyshkin:
@AkihiroSuda @cyphar @mrunalp PTAL

cyphar (approved):
LGTM, thanks!

AkihiroSuda (approved):

Issue #2603, opened 2020-09-28
Cgroups.Resources.Unified support wrt systemd driver


kolyshkin:
Initial support for cgroupv2 unified resources was merged in https://github.com/opencontainers/runc/pull/2584.

With that in place, one can set, say, `linux.resources.unified.pids.max = 42` which will make runc to write `42` to `pids.max`. The problem is, systemd `TasksMax` will still be unset/unlimited.

The issue can be solved by adding a map of some known "unified" parameters
to systemd properties.

Alternatively, we can forget about systemd properties and implement https://github.com/opencontainers/runc/issues/2602 instead (which AFAIK is what crun does).

AkihiroSuda:
>  implement #2602 instead (which AFAIK is what crun does).

:+1:

kolyshkin:
Copy/pasting my comment from https://github.com/opencontainers/runc/issues/2602#issuecomment-702269224

> OK so we have discussed it with @mrunalp and @giuseppe and looks like we can proceed with the plan outlined in #2603.

> The biggest concern about switching systemd drivers to create a sub-cgroup is that software that uses libcontainer (kubelet etc.) might break. We could do it as an option but I don't think there will be any benefit.

So that's the plan for now

PR #2669, opened 2020-10-30
Initial v2 resources.unified systemd support

kolyshkin:
In case systemd is used as cgroups manager, and a user sets some
resources using unified resource map (as per https://github.com/opencontainers/runtime-spec/pull/1040),
systemd is not aware of any parameters, so there will be a discrepancy between
the cgroupfs state and systemd unit state.

Let's try to fix that by converting known unified resources to systemd
properties.

Resources/properties currently supported:

 * pids.max / TasksMax
 * cpu.max / CPUQuota*
 * cpu.weight / CPUWeight
 * cpuset.cpus / AllowedCPUs
 * cpuset.mems / AllowedMemoryNodes
 * memory.{high,low,min,max,swap.max} / Memory{High,Low,Min,Max,SwapMap}

TODO:
 * io.weight, io.max, io.latency / IO* (not in this PR)

:information_source: While at it, fix `runc update` wrt unified resources which was missing from #2584.

Fixes: https://github.com/opencontainers/runc/issues/2603


kolyshkin:
This is just a POC for now, @AkihiroSuda @mrunalp @cyphar PTAL.

Going to add more resources later.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -34,6 +36,39 @@ func NewUnifiedManager(config *configs.Cgroup, path string, rootless bool) cgrou
...
+				num, err = strconv.ParseUint(v, 10, 64)
+				if err != nil {
+					return nil, fmt.Errorf("unified resource %q value conversion error: %w", k, err)
+				}
+			}
+			props = append(props,
+				newProp("TasksAccounting", true),
+				newProp("TasksMax", uint64(num)))
+		default:
+			logrus.Debugf("skipping unknown unified resource %q=%q", k, v)
```
AkihiroSuda, line 65: @giuseppe 

Should this be an error?

giuseppe: I believe yes.  Otherwise the runtime will ignore the user configuration.  Alternatively, if it is not handled by systemd, should it be written directly to the cgroupfs?

kolyshkin: I don't think so. This is a "cgroupfs v2 file -> systemd unit property" translation, done on a best effort basis. We can't possibly know about and translate all the properties.

The fs2 cgroup driver will be called later and it will return an error in case a parameter file does not exist.

kolyshkin: Still I left a debug message so we know the property is ignored (by the translation layer). Maybe the message should be amended to make it clear that this is just for the systemd property, not for the parameter per se.

kolyshkin: To reiterate, runc systemd driver always uses fs/fs2 driver as a backend -- first it creates everything using systemd means, when it calls fs/fs2's Set. This is ugly, but that's the way it is. One reason, I guess, is that systemd does not support all cgroupfs controllers/parameters.

AkihiroSuda: SGTM

kolyshkin: I have improved the debug message, and added a comment explaining the resource won't be ignored.

kolyshkin:
Update: added `cpu.max` (aka `CPUQuota*`), fix `runc update` wrt resources.unified.

AkihiroSuda (dismissed):

kolyshkin:
Fixed a stupid bug with cpu.max handling

mrunalp (dismissed):

kolyshkin:
OK, `cpuset.` conversion to DBus property needs more attention, I'll continue tomorrow.

kolyshkin:
> OK, cpuset. conversion to DBus property needs more attention, I'll continue tomorrow.

Made it working; PTAL @mrunalp @AkihiroSuda

AkihiroSuda (approved):

mrunalp (approved):

PR #2665, opened 2020-10-28
CI: update Fedora to 33

AkihiroSuda:
release note: https://docs.fedoraproject.org/en-US/fedora/f33/release-notes/sysadmin/Distribution/

The vagrant image seems to still use ext4 as the rootfs.

```
$ sudo ssh default 'sh -exc "uname -a && systemctl --version && df -T"'
+ uname -a
Linux localhost.localdomain 5.8.15-301.fc33.x86_64 #1 SMP Thu Oct 15 16:58:06 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
+ systemctl --version
systemd 246 (v246.6-3.fc33)
+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +ZSTD +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=unified
+ df -T
Filesystem     Type     1K-blocks    Used Available Use% Mounted on
devtmpfs       devtmpfs    995536       0    995536   0% /dev
tmpfs          tmpfs      1013572       0   1013572   0% /dev/shm
tmpfs          tmpfs       405432     540    404892   1% /run
/dev/vda1      ext4      41021664 1955980  36952200   6% /
tmpfs          tmpfs      1013572       4   1013568   1% /tmp
tmpfs          tmpfs       202712       0    202712   0% /run/user/1000
```

PR #2672, opened 2020-11-04
linux: fix remount readonly in a user namespace

giuseppe:
if we are remounting root read only when in a user namespace, make
sure the existing flags (e.g. MS_NOEXEC, MS_NODEV) are maintained
otherwise the mount fails with EPERM.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -741,7 +741,7 @@ func prepareRoot(config *configs.Config) error {
 }
 
 func setReadonly() error {
-	flags := uintptr(unix.MS_BIND | unix.MS_REMOUNT | unix.MS_RDONLY | unix.MS_REC)
+	flags := uintptr(unix.MS_BIND | unix.MS_REMOUNT | unix.MS_RDONLY)
 
 	err := unix.Mount("", "/", "", flags, "")
 	if err == nil {
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -741,7 +741,19 @@ func prepareRoot(config *configs.Config) error {
 }
 
 func setReadonly() error {
-	return unix.Mount("/", "/", "bind", unix.MS_BIND|unix.MS_REMOUNT|unix.MS_RDONLY|unix.MS_REC, "")
+	flags := uintptr(unix.MS_BIND | unix.MS_REMOUNT | unix.MS_RDONLY | unix.MS_REC)
+
+	err := unix.Mount("", "/", "", flags, "")
+	if err == nil {
+		return nil
+	}
+	var s unix.Statfs_t
+	if err := unix.Statfs("/", &s); err != nil {
+		return &os.PathError{Op: "statfs", Path: "/", Err: err}
+	}
+	flags |= uintptr(s.Flags)
+	return unix.Mount("", "/", "", flags, "")
+
 }
 
 func setupPtmx(config *configs.Config) error {


libcontainer/rootfs_linux.go
```diff
@@ -741,7 +741,19 @@ func prepareRoot(config *configs.Config) error {
 }
 
 func setReadonly() error {
-	return unix.Mount("/", "/", "bind", unix.MS_BIND|unix.MS_REMOUNT|unix.MS_RDONLY|unix.MS_REC, "")
+	flags := uintptr(unix.MS_BIND | unix.MS_REMOUNT | unix.MS_RDONLY | unix.MS_REC)
+
+	err := unix.Mount("", "/", "", flags, "")
```
kolyshkin, line 746: I don't understand why this mount is needed (or, in other words, why it has to be done using two mount(2) calls).

cyphar: The logic is that we only retain the old mount flags if it fails, but I'm confused why this code belongs in runc. I solved this problem in Docker some time ago with this code:

```go
// Get the set of mount flags that are set on the mount that contains the given
// path and are locked by CL_UNPRIVILEGED. This is necessary to ensure that
// bind-mounting "with options" will not fail with user namespaces, due to
// kernel restrictions that require user namespace mounts to preserve
// CL_UNPRIVILEGED locked flags.
func getUnprivilegedMountFlags(path string) ([]string, error) {
	var statfs unix.Statfs_t
	if err := unix.Statfs(path, &statfs); err != nil {
		return nil, err
	}

	// The set of keys come from https://github.com/torvalds/linux/blob/v4.13/fs/namespace.c#L1034-L1048.
	unprivilegedFlags := map[uint64]string{
		unix.MS_RDONLY:     "ro",
		unix.MS_NODEV:      "nodev",
		unix.MS_NOEXEC:     "noexec",
		unix.MS_NOSUID:     "nosuid",
		unix.MS_NOATIME:    "noatime",
		unix.MS_RELATIME:   "relatime",
		unix.MS_NODIRATIME: "nodiratime",
	}

	var flags []string
	for mask, flag := range unprivilegedFlags {
		if uint64(statfs.Flags)&mask == mask {
			flags = append(flags, flag)
		}
	}

	return flags, nil
}
```

The problem with doing it in runc is that we silently add mount flags which the user didn't specify. In theory this won't reduce security (since all of the `MNT_LOCKED` flags restrict the mount rather than permitting extra operations) but I still don't like the look of this.

@giuseppe did you hit this issue with podman? Maybe it'd be better to port the code from Docker to podman?

cyphar: As an aside, `MS_BIND | MS_REMOUNT | MS_RDONLY | MS_REC` doesn't actually make the mount recursively read-only but that's a separate issue...

giuseppe: @cyphar we are currently hitting it with CRI-O here: https://github.com/cri-o/cri-o/pull/4333

I am fine to move the check into CRI-O but it is currently affecting the rootfs remount, so we'd need a way to pass down these options in the OCI configuration.  AFAICS, we can set only the rootfs propagation

kolyshkin: Indeed, I don't see a way to specify mount options for container root mount.

AkihiroSuda (dismissed):

libcontainer/rootfs_linux.go
cyphar: Yeah if this is for root mounts this is a bit more of an issue -- but since this code only affects `/` remounting this might actually be alright since that means a user didn't specify it...

kolyshkin: Should we drop `MS_REC` while at it? It does not make sense anyway

libcontainer/rootfs_linux.go
```diff
@@ -741,7 +741,19 @@ func prepareRoot(config *configs.Config) error {
...
+	if err == nil {
+		return nil
+	}
+	var s unix.Statfs_t
+	if err := unix.Statfs("/", &s); err != nil {
```
kolyshkin, line 751: The errors from unix.* are bare (e.g. ENOENT) so it makes sense to wrap it

libcontainer/rootfs_linux.go
```diff
@@ -741,7 +741,19 @@ func prepareRoot(config *configs.Config) error {
...
+		return err
```
kolyshkin, line 752:
```suggestion
		return *os.PathError{Op: "statfs", Path: "/", Err: err}
```

libcontainer/rootfs_linux.go
kolyshkin: > Should we drop MS_REC while at it? It does not make sense anyway

can be addressed separately

kolyshkin (approved):
LGTM

kolyshkin:
Wonder if it's possible to add an integration test for this...

AkihiroSuda (approved):

AkihiroSuda:
https://github.com/opencontainers/runc/pull/2665 works as an integration test for this

kolyshkin:
> #2665 works as an integration test for this

ah, nice!

PR #2674, opened 2020-11-09
Move cgroup v2 out of experimental

### LINKED ISSUE #2663: Move cgroup v2 out of experimental
We have seen a lot of refactoring (thanks @kolyshkin ) for the cgroup drivers.

I think v1 and v2 now have same level of implementation quality, so I suggest moving cgroup v2 out of experimental.

`README.md` and `docs/cgroup-v2.md` need to be updated, but the code does not need to be updated.

AkihiroSuda:
After a lot of refactoring, our cgroup v1 and v2 drivers now have same level of implementation quality,
so we can move the v2 driver out of experimental.

Close #2663

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,6 @@ This means that `runc` 1.0.0 should implement the 1.0 version of the specificati
 
 You can find official releases of `runc` on the [release](https://github.com/opencontainers/runc/releases) page.
 
-Currently, the following features are not considered to be production-ready:
-
-* [Support for cgroup v2](./docs/cgroup-v2.md)
-
 ## Security
 
 The reporting process and disclosure communications are outlined [here](https://github.com/opencontainers/org/blob/master/SECURITY.md).
--- a/docs/cgroup-v2.md
+++ b/docs/cgroup-v2.md
@@ -1,6 +1,6 @@
 # cgroup v2
 
-runc supports cgroup v2 (unified mode) experimentally since v1.0.0-rc91.
+runc fully supports cgroup v2 (unified mode) since v1.0.0-rc93.
 
 To use cgroup v2, you might need to change the configuration of the host init system.
 Fedora (>= 31) uses cgroup v2 by default and no extra configuration is required.


docs/cgroup-v2.md
```diff
@@ -1,6 +1,6 @@
 # cgroup v2
 
-runc supports cgroup v2 (unified mode) experimentally since v1.0.0-rc91.
+runc supports cgroup v2 (unified mode) experimentally since v1.0.0-rc91, officially since v1.0.0-rc93.
```
kolyshkin, line 3: I don't think it makes much sense to mention experimental support and older release, IOW

```suggestion
runc fully supports cgroup v2 (unified mode) since v1.0.0-rc93.
```

or something like that

README.md
```diff
@@ -17,10 +17,6 @@ This means that `runc` 1.0.0 should implement the 1.0 version of the specificati
 
 You can find official releases of `runc` on the [release](https://github.com/opencontainers/runc/releases) page.
 
-Currently, the following features are not considered to be production-ready:
-
-* [Support for cgroup v2](./docs/cgroup-v2.md)
```
kolyshkin, line 22: Maybe it makes sense to have a section named "Additional documentation" and then link whatever we have in docs/ from there?

kolyshkin:
A fix to docs/cgroup-v2.md is needed, as it currently says "runc supports cgroup v2 (unified mode) experimentally since v1.0.0-rc91" and we want to say "fully supported since rc93" or so.

README.md
kolyshkin: Scratch that. It's better to just add a link to the doc (see the comment above)

docs/cgroup-v2.md
kolyshkin: yet better
```suggestion
runc fully supports cgroup v2 (unified mode) since v1.0.0-rc93. For more details,
see [docs/cgroup-v2.md](./docs/cgroup-v2.md).

AkihiroSuda: > see [docs/cgroup-v2.md](./docs/cgroup-v2.md)

This file itself is docs/cgroup-v2.md

README.md
AkihiroSuda: We already have a link in the "cgroup v2" section of the top-level README.md

AkihiroSuda:
updated

mrunalp (approved):

README.md
kolyshkin: yes we do; right

kolyshkin (approved):
LGTM

PR #2677, opened 2020-11-10
libcontainer/intelrdt: rename CAT and MBA enabled flags

xiaochenshen:
Rename CAT and MBA enabled flags to be consistent with others.
No functional change.

Signed-off-by: Xiaochen Shen <xiaochen.shen@intel.com>

PR #2643, opened 2020-10-12
libcontainer/intelrdt: fix CMT feature check

xiaochenshen:
Intel RDT sub-features can be selectively disabled or enabled by kernel
command line. See "rdt=" option details in kernel document:
https://www.kernel.org/doc/Documentation/admin-guide/kernel-parameters.txt

But Cache Monitoring Technology (CMT) feature is not correctly checked
in init() and getCMTNumaNodeStats() now. If CMT is disabled by kernel
command line (e.g., rdt=!cmt,mbmtotal,mbmlocal,l3cat,mba) while hardware
supports CMT, we may get following error when getting Intel RDT stats:
  runc run c1
  runc events c1
  ERRO[0005] container_linux.go:200: getting container's Intel RDT stats
  caused: open /sys/fs/resctrl/c1/mon_data/mon_L3_00/llc_occupancy: no
  such file or directory

Fix CMT feature check in init() and GetStats() call paths.

Signed-off-by: Xiaochen Shen <xiaochen.shen@intel.com>

PR #2678, opened 2020-11-10
libcontainer/intelrdt: rm init() from intelrdt.go

xiaochenshen:
Use sync.Once to init Intel RDT when needed for a small speedup to
operations which do not require Intel RDT.

Simplify IntelRdtManager initialization in LinuxFactory.

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -148,8 +148,12 @@ func RootlessCgroupfs(l *LinuxFactory) error {
 // containers that use the Intel RDT "resource control" filesystem to
 // create and manage Intel RDT resources (e.g., L3 cache, memory bandwidth).
 func IntelRdtFs(l *LinuxFactory) error {
-	l.NewIntelRdtManager = func(config *configs.Config, id string, path string) intelrdt.Manager {
-		return intelrdt.NewManager(config, id, path)
+	if !intelrdt.IsCATEnabled() && !intelrdt.IsMBAEnabled() {
+		l.NewIntelRdtManager = nil
+	} else {
+		l.NewIntelRdtManager = func(config *configs.Config, id string, path string) intelrdt.Manager {
+			return intelrdt.NewManager(config, id, path)
+		}
 	}
 	return nil
 }
@@ -272,7 +276,7 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 		newgidmapPath: l.NewgidmapPath,
 		cgroupManager: l.NewCgroupsManager(config.Cgroups, nil),
 	}
-	if intelrdt.IsCATEnabled() || intelrdt.IsMBAEnabled() {
+	if l.NewIntelRdtManager != nil {
 		c.intelRdtManager = l.NewIntelRdtManager(config, id, "")
 	}
 	c.state = &stoppedState{c: c}
@@ -314,12 +318,12 @@ func (l *LinuxFactory) Load(id string) (Container, error) {
 		root:                 containerRoot,
 		created:              state.Created,
 	}
+	if l.NewIntelRdtManager != nil {
+		c.intelRdtManager = l.NewIntelRdtManager(&state.Config, id, state.IntelRdtPath)
+	}
 	c.state = &loadedState{c: c}
 	if err := c.refreshState(); err != nil {
 		return nil, err
-	}
-	if intelrdt.IsCATEnabled() || intelrdt.IsMBAEnabled() {
-		c.intelRdtManager = l.NewIntelRdtManager(&state.Config, id, state.IntelRdtPath)
 	}
 	return c, nil
 }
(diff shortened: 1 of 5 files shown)


xiaochenshen:
This is the follow-up PR for #2643 :
@kolyshkin suggested and reviewed this change. But appreciate the input from @cyphar on the matter.

See details in https://github.com/opencontainers/runc/pull/2643#issuecomment-714237155.

crosbymichael (dismissed):
LGTM

xiaochenshen:
Rebased for dependency #2677 has been merged.
No code change.

Creatone:
LGTM

AkihiroSuda (approved):

xiaochenshen:
@crosbymichael @AkihiroSuda 
Thank you very much for code review and approval!
This PR has a dependent commit which is addressed in #2643 .
Could you also help review #2643 which one more LGTM needed to merge? Thank you!

xiaochenshen:
@crosbymichael @AkihiroSuda @Creatone 
Code rebased to merge with master branch: just removed the dependent commits, the patch for this PR keeps unchanged.
Please help re-approve.
Thank you!

xiaochenshen:
Travis-CI always fails on cases "Go: 1.15.x", "Go: 1.14.x", and "Go: tip" with the same error:
https://travis-ci.org/github/opencontainers/runc/jobs/744584826#L1922

Looks like something wrong with Travis-CI itself?

xiaochenshen:
The Travis-CI system issue seems to be fixed already.

@crosbymichael @AkihiroSuda @kolyshkin @Creatone 
Code rebased but the patch keeps unchanged.
Could you help re-approve at your convenience?
Thank you!

libcontainer/intelrdt/intelrdt.go
```diff
@@ -203,55 +206,56 @@ type intelRdtData struct {
...
-	if mbaScEnabled {
-		// We confirm MBA Software Controller is enabled in step 2,
-		// MBA should be enabled because MBA Software Controller
-		// depends on MBA
-		mbaEnabled = true
-	} else if flagsSet.MBA {
-		if _, err := os.Stat(filepath.Join(intelRdtRoot, "info", "MB")); err == nil {
-			mbaEnabled = true
+// Check if Intel RDT sub-features are enabled in InitIntelRdt()
+func InitIntelRdt() {
```
kolyshkin, line 210: since the package name is intelrdt, you can just use `Init` as a name (so the call to it will be `interldt.Init()`)

xiaochenshen: > since the package name is intelrdt, you can just use `Init` as a name (so the call to it will be `interldt.Init()`)

@kolyshkin Good suggestion. Thank you!
I have updated the code in commit https://github.com/opencontainers/runc/pull/2678/commits/c2fc85e1d802e3952899d429c10480f0ebfa778d

xiaochenshen:
@crosbymichael @AkihiroSuda @kolyshkin 
This PR is pending for a long time. I think just re-approvals are required for merge. Am I right?
The code change in the old version of the patch has been approved by you. But then this PR is rebased against master branch (without code change), the original approvals are outdated.
 
Could you help re-approve at your convenience?
Thank you very much for help!

kolyshkin (commented):
LGTM

kolyshkin (approved):
LGTM

libcontainer/factory_linux.go
```diff
@@ -148,8 +148,13 @@ func RootlessCgroupfs(l *LinuxFactory) error {
 // containers that use the Intel RDT "resource control" filesystem to
 // create and manage Intel RDT resources (e.g., L3 cache, memory bandwidth).
 func IntelRdtFs(l *LinuxFactory) error {
-	l.NewIntelRdtManager = func(config *configs.Config, id string, path string) intelrdt.Manager {
-		return intelrdt.NewManager(config, id, path)
+	intelrdt.Init()
+	if !intelrdt.IsCATEnabled() && !intelrdt.IsMBAEnabled() {
```
kolyshkin, line 152: So, I think by requiring to call Init() explicitly you change the public API. If the only user is runc itself, this is fine. If someone else (kubernetes etc) is using libcontainer/intelrtd, this is problematic.

What you should do, I guess, is to make Init private, and let some functions that need it (such as `IsCATEnabled`) etc. call it. That way, the initialization will be performed in a transparent manner, but will only be done when needed.

kolyshkin (changes requested):
On a second thought, I think it needs to be changed to not require user call Init explicitly (see details above).

xiaochenshen:
> 
> 
> On a second thought, I think it needs to be changed to not require user call Init explicitly (see details above).

@kolyshkin Thank you very much for code review and good suggestions. I will make change to prevent calling Init() explicitly by outside caller.

xiaochenshen:
@kolyshkin 
I have made the change according to your suggestion.

Change log:
1. Remove intelrdt.Init() call in IntelRdtFs().
2. Privatize and rename Init() to featuresInit().
3. Call featuresInit() in Is{CAT | MBA | MBASc | CMT | MBM}Enabled().

kolyshkin (approved):
LGTM, and thanks for working on this.

@AkihiroSuda PTAL

AkihiroSuda (approved):

Issue #2425, opened 2020-05-20
TestExecInTTY: execin_test.go:351: unexpected carriage-return in output


kolyshkin:
During the course of CI for #2411 I saw a single failure from Fedora32 on Vagrant ([logs here](https://travis-ci.org/github/opencontainers/runc/jobs/689070268)):
```
=== RUN   TestExecInTTY
    TestExecInTTY: execin_test.go:351: unexpected carriage-return in output
--- FAIL: TestExecInTTY (0.14s)
```

It only happened once, and so I suspect some race in the code or in the test. The test code comes from #1146. 

@cyphar maybe you could take a look?

XiaodongLoong:
I also meet this issue:
https://travis-ci.org/github/opencontainers/runc/jobs/708255866

kolyshkin:
Got it one more time here: https://travis-ci.org/github/opencontainers/runc/jobs/740066484 (https://github.com/opencontainers/runc/pull/2658#issuecomment-719884438)

kolyshkin:
Another occurrence: https://github.com/opencontainers/runc/pull/2682#issuecomment-738490882

kolyshkin:
Seems that github actions is faster so we'll see this one more and more now since #2690 is in

kolyshkin:
OK, finally got the output (from https://github.com/opencontainers/runc/runs/1659399228?check_suite_focus=true):

```
=== RUN   TestExecInTTY
832
    execin_test.go:349: unexpected carriage-return in output "PID   USER     TIME  COMMAND\r\n    1 root      0:00 cat\n    6 root      0:00 ps\n"
833
--- FAIL: TestExecInTTY (0.14s)
```

Note the first line ends with `\r\n`, while the subsequent ones only have `\n`. This looks like a race between running `ps` and setting terminal.

kolyshkin:
Easy to repro locally:

```console
$ cd libcontainer/integration
$ go test -c .
$ $ sudo ./integration.test -test.run TestExecInT -test.count 1000
--- FAIL: TestExecInTTY (0.02s)
    execin_test.go:332: unexpected carriage-return in output "PID   USER     TIME  COMMAND\r\n    1 root      0:00 cat\r\n    8 root      0:00 ps\r\n"
--- FAIL: TestExecInTTY (0.03s)
[...]

PR #2682, opened 2020-11-13
Implement seccomp notify support

mauriciovasquezbernal:
Add support for notify seccomp actions. Seccomp profiles can now specify the SCMP_ACT_NOTIFY action that notifies a user space application (seccomp agent) when such syscall is performed. The seccomp agent can, in turn, decide to perform the syscall on behalf of the container and set the syscall return values or even inject file descriptors to the container. This, for example, allows for unprivileged containers to run some privileged syscalls that the seccomp agent performs on its behalf.

The format and more details can be found in the OCI runtime specification, in the [seccomp section][oci-seccomp]. If you are a developer and want more information, check out the [linux kernel documentation][linux-doc], [libseccomp-golang][libseccomp] and [Christian Brauner's blog post][brauner-blog]. To use this feature you need Linux >= 5.7 and libseccomp >= 2.5.0.
 
[oci-seccomp]: https://github.com/opencontainers/runtime-spec/blob/20a2d9782986ec2a7e0812ebe1515f73736c6a0c/config-linux.md#seccomp 
[linux-doc]: https://www.kernel.org/doc/html/v5.7/userspace-api/seccomp_filter.html#userspace-notification 
[libseccomp]: https://github.com/seccomp/libseccomp-golang 
[brauner-blog]: https://brauner.github.io/2020/07/23/seccomp-notify.html


Dockerfile
```diff
@@ -27,21 +27,26 @@ RUN dpkg --add-architecture armel \
...
         python-minimal \
         sudo \
         uidmap \
+	astyle \
+	golint \
+	gperf \
     && apt-get clean \
     && rm -rf /var/cache/apt /var/lib/apt/lists/*;
 
+RUN wget https://github.com/seccomp/libseccomp/releases/download/v2.5.0/libseccomp-2.5.0.tar.gz \
```
AkihiroSuda, line 42: Can we `ADD` a helper script?

mauriciovasquezbernal: Sure. Done as part of https://github.com/opencontainers/runc/pull/2682/commits/f549f8989c9db32dec19ee6c4ad9bce5fa5abf32.

kolyshkin:
CI failure is a known (but rare) flake, see https://github.com/opencontainers/runc/issues/2425
```
=== RUN   TestExecInTTY
    execin_test.go:349: unexpected carriage-return in output
--- FAIL: TestExecInTTY (0.15s)
```

(restarted the job)

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -33,54 +34,92 @@ const (
 // Started in the container init process, and carried over to all child processes
 // Setns calls, however, require a separate invocation, as they are not children
 // of the init until they join the namespace
```
vbatts, line 36: please add docs on the returned fd

mauriciovasquezbernal: Done.

vbatts:
https://github.com/opencontainers/runtime-spec/pull/1074 merged now

rata:
Hi! I'm a coworker of @mauriciovasquezbernal, I'll coordinate with him to fix the conflicts in this PR and remove the draft when ready! :)

rata:
Hi!

I’ve rebased the PR and fixed all the conflicts, all tests should be green too.

The OCI runtime-spec change this patch implements [was recently merged][oci] and I see that [the spec change is already merged in runc][runc-spec] too. Should we mark this PR for 1.0 milestone? Currently it is in the post 1.0 milestone.

I’m keeping the PR as a draft for now, as the [libseccomp-golang PR][libseccomp] is not yet merged and we need it for this PR (not only for tests, we use some consts in core runc too).  However, changes on the libseccomp PR will have minimal (most likely none) consequences to this code. So, we keep it as draft until that PR is merged, but please review whenever you can :)


[oci]: https://github.com/opencontainers/runtime-spec/pull/1074
[runc-spec]: https://github.com/opencontainers/runc/commit/dd2caace166b1f1614c446f8bb32268cdd0b1c3c 
[libseccomp]: https://github.com/seccomp/libseccomp-golang/pull/59

rata:
Updated, fixed a conflict in go.sum too

rata:
Rebased on top of latest master

libcontainer/process_linux.go
```diff
@@ -178,6 +179,49 @@ func (p *setnsProcess) start() (retErr error) {
...
+			}
+			defer unix.Close(int(pidfd))
+
+			seccompFd, _, err := unix.Syscall(unix.SYS_PIDFD_GETFD, pidfd, uintptr(sync.Fd), 0)
+			if err != 0 {
+				return newSystemErrorWithCause(err, "performing SYS_PIDFD_GETFD syscall")
+			}
+			defer unix.Close(int(seccompFd))
+
+			if p.config.Config.Seccomp.ListenerPath == "" {
```
kolyshkin, line 196: Maybe makes sense to move this check up (no sense in calling PIDFD_OPEN and PIDFD_GETFD if the path is not set)?

libcontainer/process_linux.go
```diff
@@ -439,6 +483,47 @@ func (p *initProcess) start() (retErr error) {
...
+
+			seccompFd, _, errno := unix.Syscall(unix.SYS_PIDFD_GETFD, pidfd, uintptr(sync.Fd), 0)
+			if errno != 0 {
+				return newSystemErrorWithCause(errno, "performing SYS_PIDFD_GETFD syscall")
+			}
+			defer unix.Close(int(seccompFd))
+
+			if p.config.Config.Seccomp.ListenerPath == "" {
+				return newSystemError(errors.New("listenerPath is not set"))
+			}
```
kolyshkin, line 502: Perhaps it makes sense to consolidate this code since it's repeated twice. Something like

```go
seccompFd, err := recvSeccompFd(childPid, sync.Fd)
```

maybe?

libcontainer/process_linux.go
```diff
@@ -634,6 +719,31 @@ func (p *initProcess) forwardChildLogs() chan error {
 	return logs.ForwardLogs(p.logFilePair.parent)
 }
 
+func sendContainerProcessState(listenerPath string, state *specs.ContainerProcessState, fd int) error {
+	conn, err := net.Dial("unix", listenerPath)
+	if err != nil {
+		return fmt.Errorf("cannot connect to %q: %v\n", listenerPath, err)
```
kolyshkin, line 725: 1. `\n` should be removed.
2. Please to use %w for error.

kolyshkin: 3. I think `net.Dial` already includes path/address in the error message, as well as the error (`dial`), so there's no need to wrap it.

libcontainer/process_linux.go
```diff
@@ -634,6 +719,31 @@ func (p *initProcess) forwardChildLogs() chan error {
...
+	}
+
+	socket, err := conn.(*net.UnixConn).File()
+	if err != nil {
+		return fmt.Errorf("cannot get socket: %v\n", err)
```
kolyshkin, line 730: 1. `\n` should be removed.
2. use `%w` for errors.

libcontainer/process_linux.go
```diff
@@ -634,6 +719,31 @@ func (p *initProcess) forwardChildLogs() chan error {
...
+	}
+	defer socket.Close()
+
+	b, err := json.Marshal(state)
+	if err != nil {
+		return fmt.Errorf("cannot marshall seccomp state: %v\n", err)
```
kolyshkin, line 736: ditto

libcontainer/process_linux.go
```diff
@@ -634,6 +719,31 @@ func (p *initProcess) forwardChildLogs() chan error {
...
+	}
+
+	err = utils.SendFds(socket, b, fd)
+	if err != nil {
+		return fmt.Errorf("cannot send seccomp fd to %s: %v\n", listenerPath, err)
```
kolyshkin, line 741: ditto

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -14,67 +14,86 @@ import (
...
 )
 
-// Filters given syscalls in a container, preventing them from being used
-// Started in the container init process, and carried over to all child processes
-// Setns calls, however, require a separate invocation, as they are not children
-// of the init until they join the namespace
-func InitSeccomp(config *configs.Seccomp) error {
+// InitSeccomp installs the seccomp filters to be used in the container as
+// specified in config.
+// Returns the seccomp file descriptor is any of the filters include a
```
kolyshkin, line 33: typo? `s/ is / if /`?

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -14,67 +14,86 @@ import (
...
 	if err != nil {
-		return fmt.Errorf("error creating filter: %s", err)
+		return -1, fmt.Errorf("error creating filter: %s", err)
+	}
+
+	// TODO: config.Flags defines the options to pass to seccomp(2) but
+	// it's not taken into consideration.
+	for _, call := range config.Syscalls {
+		if call.Action == configs.Notify {
+			if call.Name == "write" {
```
kolyshkin, line 58: nit:
```suggestion
		if call.Action == configs.Notify && call.Name == "write" {
```

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -14,67 +14,86 @@ import (
...
+			}
+		}
 	}
 
 	// Add extra architectures
 	for _, arch := range config.Architectures {
 		scmpArch, err := libseccomp.GetArchFromString(arch)
 		if err != nil {
-			return fmt.Errorf("error validating Seccomp architecture: %s", err)
+			return -1, fmt.Errorf("error validating Seccomp architecture: %s", err)
```
kolyshkin, line 68: Since you're changing this, while at it, can you use `%w` instead of `%s`?

libcontainer/standard_init_linux.go
```diff
@@ -142,12 +142,6 @@ func (l *linuxStandardInit) Init() error {
 			return errors.Wrap(err, "set nonewprivileges")
 		}
 	}
-	// Tell our parent that we're ready to Execv. This must be done before the
-	// Seccomp rules have been applied, because we need to be able to read and
-	// write to a socket.
```
kolyshkin, line 147: Is this still a concern? If not, please amend the commit message to explain why.

libcontainer/standard_init_linux.go
```diff
@@ -156,9 +150,24 @@ func (l *linuxStandardInit) Init() error {
...
+		if seccompFd != -1 {
+			err := syncParentSeccomp(l.pipe, seccompFd)
+			if err != nil {
+				return errors.Wrap(err, "sync parent seccomp")
+			}
+			err = unix.Close(seccompFd)
+			if err != nil {
+				return errors.Wrap(err, "close seccomp fd")
+			}
+		}
```
kolyshkin, line 166: Maybe move this part out to a function, as this code is used twice.

libcontainer/process_linux.go
rata: Good point, moving it!

libcontainer/process_linux.go
rata: Yeap, good call. I'm moving only these bits and not the `if p.config.Config.Seccomp.ListenerPath == "" ` part as we need `p` for that and it would be called with `p` being different structs. Also, IMHO, semantically seems better to keep both things separate

libcontainer/process_linux.go
rata: Cool, returning err then

libcontainer/process_linux.go
rata: Fixed, thanks!

libcontainer/process_linux.go
rata: Fixed, thanks!

libcontainer/process_linux.go
rata: Fixed, thanks!

libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -565,9 +571,9 @@ func enosysPatchFilter(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (
 	return fprog, nil
 }
 
-func filterFlags(filter *libseccomp.ScmpFilter) (flags uint, noNewPrivs bool, err error) {
+func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags uint, noNewPrivs bool, err error) {
 	// Ignore the error since pre-2.4 libseccomp is treated as API level 0.
-	apiLevel, _ := libseccomp.GetApi()
+	apiLevel, _ := libseccomp.GetAPI()
```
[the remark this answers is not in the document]
rata: True, moving it fixes compilation at that commit too. Thanks!

libcontainer/seccomp/seccomp_linux.go
rata: Fixed, thanks

libcontainer/seccomp/seccomp_linux.go
rata: Thanks, adding exactly that locally. Will be in the next push

libcontainer/seccomp/seccomp_linux.go
rata: Sure!

libcontainer/standard_init_linux.go
rata: Fixed, thanks. Moved to `syncParentSeccomp()`

libcontainer/process_linux.go
alban: > I think net.Dial already includes path/address in the error message, as well as the error (dial), so there's no need to wrap it.

Correct:

```
$ cat main.go 
package main

import (
	"fmt"
	"net"
)

func main() {
	_, err := net.Dial("unix", "/fdsjkfhdjk")
	if err != nil {
		fmt.Printf("%s\n", err)
	}
}
$ go run main.go 
dial unix /fdsjkfhdjk: connect: no such file or directory
```

alban:
Thanks for the reviews!

* https://github.com/seccomp/libseccomp-golang/pull/59 has been merged so I updated the vendoring: we don't need to rely on the fork anymore.
* I rebased the branch
* I removed the 'draft' status of this PR.
* I added the fixes that @rata prepared.
* Still two concerns from @kolyshkin that I need to check:
  1. [ ] Explain possible restrictions on read/write in the commitmsg (syncParentReady after seccomp)
  2. [ ] Comment "seccomp as close to execve as possible"

libcontainer/standard_init_linux.go
alban: I don't remember why syncParentReady was moved up. During syncParentReady, the parent process just sets rlimits for the child and runs the PreStart OCI hook, so I don't think it matters if it is before or after the seccomp. I'm moving it back where it was and it seems to work fine.

<details>

I also ran some tests with blocking the write syscall: In the case noNewPrivileges==false, blocking the "write" syscall in the seccomp profile was already broken on master because we had `unix.Write(fd, []byte("0"))` after `seccomp.InitSeccomp` anyway. In practice, it breaks with the following symptoms:
```
$ sudo ../runc run c1
ERRO[0000] cannot start an already running container
```
(the parent process misinterpret the empty data on the pipe, so the error message is confusing)

I would argue that blocking the read/write syscalls in the seccomp profile is a bad idea because:
- The read/write syscalls are needed in all use cases I can think of
- They are part of the 4 syscalls initially accepted in the strict mode of seccomp (see manual page seccomp(2))
- runc was already broken when blocking "write" (at least with noNewPrivileges==false) so probably nobody is doing that.

So if we ever need to move it, that should be fine too.

</details>

libcontainer/standard_init_linux.go
```diff
@@ -181,6 +190,25 @@ func (l *linuxStandardInit) Init() error {
 	if err != nil {
 		return err
 	}
+	// Set seccomp as close to execve as possible, so as few syscalls take
+	// place afterward (reducing the amount of syscalls that users need to
+	// enable in their seccomp profiles).
```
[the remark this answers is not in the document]
alban: I'll append the following comment:

> However, this needs to be done before closing the pipe since we need it to pass the seccompFd to the parent.

alban:
I rebased the branch and addressed the 2 last comments. The tests pass fine now.

rata:
Rebased on top of latest master to fix some conflicts.

Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN echo 'deb https://download.opensuse.org/repositories/devel:/tools:/criu/Debi
     && dpkg --add-architecture ppc64el \
     && apt-get update \
     && apt-get install -y --no-install-recommends \
+        astyle \
```
AkihiroSuda, line 15: Why?

rata: Not really sure, I think maybe it was needed when running on travis (yeah, the PR was open so long ago :D), removed that and golint and tests pass fine. I need to keep gperf, though

rata:
@AkihiroSuda @kolyshkin I know you are probably busy with runc 1.0 release, but when you have time PTAL. I've rebased on master to fix conflicts and removed astyle and golint (unneeded dependencies)

Also, what do you think of moving the milestone to 1.1.0 maybe? (now it is in post-1.0) It will be great for us to merge soon :)

rata:
Rebased on top of latest master

rata:
I fixed the conflicts caused by the recent merge of https://github.com/opencontainers/runc/pull/3011. It took me some hours, though, and is getting painful to maintain such a big change out of tree.

@kolyshkin @AkihiroSuda any chance you can review this soon, so we hopefully merge this soon too? :)

Please note [as I mentioned previously][comment] that the spec change this patch implements is [already vendored in runc master][runc-spec] too. So, merging this PR is the only thing left.

[runc-spec]: https://github.com/opencontainers/runc/commit/dd2caace166b1f1614c446f8bb32268cdd0b1c3c 
[comment]: https://github.com/opencontainers/runc/pull/2682#issuecomment-802867420

rata:
@kolyshkin @AkihiroSuda friendly ping?

There are conflicts again, I'll rebase and fix them, but having a review will greatly help anyways :)

rata:
Rebased on top of latest master and fixed the conflicts. CI on my fork is green too :)

rata:
@kolyshkin @AkihiroSuda friendly ping?

I've just rebased and did a very minor change: improve the error msg to this as suggested by @thaJeztah [in the moby PR][link]. The change is this line:

https://github.com/opencontainers/runc/pull/2682/files#diff-f0214a0f16408fc7f168c6fc9837d189590025cc1813ebf7c1d751136936dfbfR727

Before if the seccomp agent was not running the error was:
```
$ sudo ../runc/runc run mycontainer
ERRO[0000] unable to start container process: dial unix /tmp/socket.sock: connect: no such file or directory 
```

Now it is:
```
$ sudo ../runc/runc run mycontainer
ERRO[0000] unable to start container process: failed to connect with seccomp agent specified in the seccomp profile: dial unix /tmp/socket.sock: connect: no such file or directory 
```

[link]: https://github.com/moby/moby/pull/42604#issuecomment-876763471

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	buffer[len(buffer)-1] = 0
+	s := buffer[:bytes.IndexByte(buffer, 0)]
+	return string(s), nil
+}
+
+func runMkdirForContainer(pid uint32, fileName string, mode uint32, metadata string) error {
+	if strings.HasPrefix(fileName, "/") {
+		return unix.Mkdir(fmt.Sprintf("/proc/%d/root%s-%s", pid, fileName, metadata), mode)
```
AkihiroSuda, line 149: nit: can we use https://github.com/cyphar/filepath-securejoin ?

contrib/cmd/seccompagent/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Run the seccomp agent in the background:
+```bash
+sudo ./contrib/cmd/seccompagent/seccompagent &
+```
+
+Start the container:
+```bash
+runc run mycontainerid
+```
```
AkihiroSuda, line 66: Can we explain what user should do after starting the container?

e.g., run mkdir, chmod, ...

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,265 @@
...
+			continue
+		}
+		newFd, metadata, err := handleNewMessage(int(socket.Fd()))
+		socket.Close()
+		if err != nil {
+			logrus.Errorf("Error receiving seccomp file descriptor: %v", err)
+			continue
+		}
+		logrus.Infof("Received new seccomp fd: %v\n", newFd)
+		go notifHandler(libseccomp.ScmpFd(newFd), metadata)
```
AkihiroSuda, line 263: verify metadata

libcontainer/configs/config.go
```diff
@@ -31,10 +31,12 @@ type IDMap struct {
...
-	DefaultAction   Action     `json:"default_action"`
-	Architectures   []string   `json:"architectures"`
-	Syscalls        []*Syscall `json:"syscalls"`
-	DefaultErrnoRet *uint      `json:"default_errno_ret"`
+	DefaultAction    Action     `json:"default_action"`
+	Architectures    []string   `json:"architectures"`
+	Syscalls         []*Syscall `json:"syscalls"`
+	DefaultErrnoRet  *uint      `json:"default_errno_ret"`
+	ListenerPath     string     `json:"listenerPath,omitempty"`
+	ListenerMetadata string     `json:"listenerMetadata,omitempty"`
```
AkihiroSuda, line 39: Seems mixing up snake_case and camelCase

libcontainer/init_linux.go
```diff
@@ -273,6 +273,36 @@ func syncParentHooks(pipe io.ReadWriter) error {
...
+		return err
+	}
+
+	// Wait for parent to give the all-clear.
+	if err := readSync(pipe, procSeccompDone); err != nil {
+		unix.Close(seccompFd)
+		return fmt.Errorf("sync parent seccomp: %w", err)
+	}
+
+	if err := unix.Close(seccompFd); err != nil {
```
AkihiroSuda, line 299: Do we need to retry on EINTR? (for every `unix.Close`)

script/install-libseccomp.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -eux -o pipefail
+
+VERSION=2.5.0
+MD5SUM=463b688bf7d227325b5a465b6bdc3ec4
```
AkihiroSuda, line 5: nit: SHA256SUM is considered to be more preferable today

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,209 @@
...
+function scmp_act_notify_template() {
+	# The agent intercepts mkdir syscalls and creates the folder appending
+	# "-bar" (listenerMetadata below) to the name.
+	update_config '.process.args = ["/bin/sh", "-c", "'"$1"'"] |
+	.process.noNewPrivileges = '"$2"' |
+	.linux.seccomp = {
+		"defaultAction":"SCMP_ACT_ALLOW",
+		"listenerPath": "'"$SECCCOMP_AGENT_SOCKET"'",
+		"listenerMetadata": "bar",
+		"architectures":["SCMP_ARCH_X86","SCMP_ARCH_X32"],
```
AkihiroSuda, line 30: Is `SCMP_ARCH_X86_64` intentionally excluded here?

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,209 @@
...
+
+# The call to seccomp is done at different places according to the value of
+# noNewPrivileges, for this reason many of the following cases are tested with
+# both values.
+
+# check that SCMP_ACT_NOTIFY on older kernels fails with a clear error message.
+@test "runc run [seccomp] (SCMP_ACT_NOTIFY old kernel)" {
+	if [[ "$KERNEL_MAJOR" -gt 5 || ("$KERNEL_MAJOR" -eq 5 && "$KERNEL_MINOR" -ge 6) ]]; then
+		skip "requires kernel less than 5.6"
+	fi
```
AkihiroSuda, line 47: I guess we also need to skip the test when the host is not x86_64

vendor/modules.txt
```diff
@@ -39,7 +39,7 @@ github.com/opencontainers/selinux/pkg/pwalk
 github.com/pkg/errors
 # github.com/russross/blackfriday/v2 v2.0.1
 github.com/russross/blackfriday/v2
-# github.com/seccomp/libseccomp-golang v0.9.1
+# github.com/seccomp/libseccomp-golang v0.9.2-0.20210429002308-3879420cc921
```
AkihiroSuda, line 42: Can we have v0.9.2 soon?

contrib/cmd/seccompagent/seccompagent.go
rata: Oh, sure

contrib/cmd/seccompagent/README.md
rata: Right, will clarify that. Thanks!

vendor/modules.txt
rata: It's really not up to us :)

contrib/cmd/seccompagent/seccompagent.go
rata: What do you mean with verify metadata? Verify that is something we can use as a suffix for a filename? Can you please elaborate?

libcontainer/configs/config.go
rata: Ouch, thanks! Will fix

contrib/cmd/seccompagent/seccompagent.go
AkihiroSuda: Yes, we should error out if metadata has "/"

libcontainer/init_linux.go
rata: I don't think so. Because:
 * If it fails with EINTR, we will exit, so it is not that it is not handled. The container might fail in the worst case, but not run with the seccomp fd open, that would be a big problem. If there are bug reports about this, of course, it can be changed to loop instead
 * IIUC, it should not return EINTR, as go runtime registers the signal handlers with SA_RESTART. That means that the close(2) syscall should be automatically restarted if it was interrupted, instead of returning EINTR. There are bugs in the wild, though, and some syscalls on some OS do return EINTR when they shouldn't. This should be only Linux and I haven't seen changes in the go runtime looping on close to handle those bugs (as they were done with others), so I think we are just fine.

But, as you prefer, really.

script/install-libseccomp.sh
rata: Oh, will update to check for that

contrib/cmd/seccompagent/README.md
rata: Simplified this. I added a config.json to the directory, referenced that to be used, simplified the command that is run and added here the expected output with a brief explanation.

Thanks!

contrib/cmd/seccompagent/seccompagent.go
rata: Added a validation, will be in the next push

I basically used `filepath.Base()` to get something reasonable and also check if it contains "/", if it does, use a fallback safe string instead.

script/install-libseccomp.sh
rata: libseccomp publishes the .SHA256SUM file on releases, so I've updated this to download the .tar.gz and the .SHA256SUM file and verify it matches.

If you prefer, I can store the SHA256 in the script instead, though.

And as a minot change, updated this to use v2.5.1 too.

rata: Well, I've updated it to have the expected SHA256SUM in the script. Even though we are running on github and downloading from github, is not much of overhead to have the SUM on the script and update it on updates. libseccomp is not something that is released super often :)

tests/integration/seccomp-notify.bats
rata: Added a "requires arch_x86_64", as we only care about one arch for now. But let me know if you prefer to add a “requires_arch x86_64”, similar to "requires_kernel" added here too.

tests/integration/seccomp-notify.bats
rata: No, added it. Thanks!

libcontainer/configs/config.go
rata: I’ve changed it to keep the style currently in master (came case for struct field names and underscore for json tags), thanks!

I think we probably can't change this, as it is in libcontainer and might break users. But it wasn't obvious why it was working for me, so looking at for example the DefaultErrnoRet field seems to be parsed correctly today on master because from [json unmarshall doc][link]:

> Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match”. So, it all works because the struct field name is as expected (and match is case insensitive if needed, and we need that as well as the struct field name starts with a capital letter but not the json field name), making basically the json tags quite pointless.

And then, I guess it works when the json has `"defaultErrnoRet": 6` because the struct field name is DefaultErrnoRet and matches case-insensitive (defaultErrnoRet), and the json tag is mostly unused (as probably most users use camel case instead of underscores, as that is how the field is defined in the runtime spec).

Did the same (so basically camel case and underscore works) for the ListenerPath and ListenerMetadata fields.

Thanks!

[link]: https://pkg.go.dev/encoding/json#Unmarshal

contrib/cmd/seccompagent/seccompagent.go
rata: Added!

rata:
@AkihiroSuda thanks! PTAL. I addressed all the review comments.

I've also added one more integration test (to test returning errors too, that was missing an integration test).

PTAL, thanks again for the review! :)

rata:
I've updated the PR with some simple changes:
* Fixed compilation when building without seccomp build tag (`make BUILDTAGS="" all`): Added `func main()` to `contrib/cmd/seccompagent/unsupported.go` so compilation works fine
* Removed `skopeo` dep from `.github/workflows/test.yml`: it was not needed.
* Rename `SCMP_ACT_IMAGINARY_ACTION` --> `SCMP_ACT_NON_EXIXTENT_ACTION` and `SCMP_ARCH_IMAGINARY_ARCH` --> `SCMP_ARCH_NON_EXISTENT_ARCH` in tests added by this PR
* Removed a redundant `TODO`: we were adding a TODO comment in `libcontainer/seccomp/seccomp_linux.go` saying seccomp flags are ignored. However, this is already mentioned [here][link], it is already the case without this PR and there is nothing in this PR that makes it relevant to add such a comment IMHO. So, just removed the TODO comment we were adding.

[link]: https://github.com/opencontainers/runc/blob/713748d4a8ff418bc8218ffb5f6c80ed396b8c1a/libcontainer/specconv/spec_linux.go#L850-L853

AkihiroSuda:
Sorry needs rebase

rata:
@AkihiroSuda thanks a lot! Rebased and the simple conflict is fixed :)

Also, I've added another simple test case: Check that starting when no seccomp agent is running fails with a clear error (the test was missing, no code change was needed).

CI should be green too (just run the CI on a fork) :tada:

AkihiroSuda (dismissed):

libcontainer/process_linux.go
rata: FYI, this was moved to the top. Just that github is smart enough to move the comment to the top too :)

rata:
Just rebased to solve the trivial conflicts with the merges done some hours ago.

@AkihiroSuda It lost your LGTM after the rebase. Can you please LGTM again? :)

@kolyshkin @cyphar ping? :)

libcontainer/process_linux.go
```diff
@@ -174,6 +175,42 @@ func (p *setnsProcess) start() (retErr error) {
...
+				},
+			}
+			if err := sendContainerProcessState(p.config.Config.Seccomp.ListenerPath,
+				containerProcessState, seccompFd); err != nil {
+				return err
+			}
+
+			// Sync with child.
+			if err := writeSync(p.messageSockPair.parent, procSeccompDone); err != nil {
+				return fmt.Errorf("writing syncT 'SeccompDone': %w", err)
```
kolyshkin, line 211: It looks like you can move this error wrapping to `writeSyncWithFd`.

.github/workflows/test.yml
```diff
@@ -31,7 +31,8 @@ jobs:
         # criu repo
         sudo add-apt-repository -y ppa:criu/ppa
         # apt-add-repository runs apt update so we don't have to
-        sudo apt -q install libseccomp-dev criu
+        sudo apt -q install criu gperf build-essential
+        sudo ./script/install-libseccomp.sh x86_64-linux-gnu
```
kolyshkin, line 35: It looks like ubuntu 20.04 have libseccomp 2.5.1 in updates now, so this commit may no longer be necessary.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,296 @@
...
+		unix.Close(recvFds[i])
+	}
+
+	return fd, nil
+}
+
+func handleNewMessage(sockfd int) (uintptr, string, error) {
+	MaxNameLen := 4096
+	oobSpace := unix.CmsgSpace(4)
+	stateBuf := make([]byte, 4096)
```
kolyshkin, line 85: Nit
```go
const maxNameLen = 4096
stateBuf := make([]byte, maxNameLen)
```

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,296 @@
...
+			if err != nil {
+				resp.Error = int32(unix.ENOSYS)
+				resp.Val = ^uint64(0) // -1
+			}
+			resp.Flags = 0
+		case "chmod":
+			fallthrough
+		case "fchmod":
+			fallthrough
+		case "fchmodat":
```
kolyshkin, line 225:
```suggestion
case "chmod", "fchmod", "fchmodat":
```

tests/integration/helpers.bash
```diff
@@ -490,3 +513,11 @@ function teardown_bundle() {
 	done
 	rm -rf "$ROOT"
 }
+
+function requires_kernel() {
+	MAJOR_REQUIRED=$(echo "$1" | cut -d. -f1)
+	MINOR_REQUIRED=$(echo "$1" | cut -d. -f2)
```
kolyshkin, line 519: Use `local` and lowercase for local variables.

IOW
```bash
local major_req minor_req
major_req=$(echo "$1" | cut -d. -f1)
minor_req=$(echo "$1" | cut -d. -f2)
```

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,249 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+	teardown_seccompagent
```
kolyshkin, line 6: This should not be necessary -- bats is reliably calling teardown.

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,249 @@
...
+
+	scmp_act_notify_template "/bin/true" false '"mkdir"'
+
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+}
+
+@test "runc run [seccomp] (SCMP_ACT_NOTIFY noNewPrivileges false)" {
+	requires_kernel 5.6
+	requires arch_x86_64
```
kolyshkin, line 59: If these are requirements for every tests (it seems so), then it's better to move this to `setup()` (so we can skip setup entirely).

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,249 @@
...
+
+	scmp_act_notify_template "touch /dev/shm/foo && chmod 777 /dev/shm/foo" false '"chmod", "fchmod", "fchmodat"'
+
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+	[[ "$output" == *"No medium found"* ]]
+}
+
+# check that trying to use SCMP_ACT_NOTIFY with write() gives a meaningfull error.
```
kolyshkin, line 208: typo: `meaningful`

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed, thanks!

.github/workflows/test.yml
rata: Yeap, and the updates is used in the github CI. So removing this chunk, thanks!

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed locally, thanks!

tests/integration/seccomp-notify.bats
rata: Fixed, thanks!

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,249 @@
...
+	[[ "$output" == *"SCMP_ACT_NOTIFY cannot be used for the write syscall"* ]]
+}
+
+# check that a startContainer hook doesn't get any extra file descriptor.
+@test "runc run [seccomp] (SCMP_ACT_NOTIFY startContainer hook)" {
+	requires_kernel 5.6
+	requires arch_x86_64
+
+	# shellcheck disable=SC2016
+	# For some (unknown) reason shellcheck doesn't like L199 together with the "if" in L210.
```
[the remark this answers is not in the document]
rata: Removing line numbers, thanks!

It seems the check ("SC2016: Expressions don't expand in single quotes, use double quotes for that.") is not really helpful when using `update_config()`. Is not just the startContainer command, there are several other issues that make this check and several good uses of `update_config()` incompatible:
* When using `update_config()` we use single quotes to properly delimit the $1 param to the function
* This shellcheck check checks that, basically, there are no `$` inside a single quote string. For example: it says that this part `"'"$SECCCOMP_AGENT_SOCKET"'"` will not escape the var, although it will as the opening single quote was in a previous line.
* If we are running commands in a subshell inside the single quoted string, like `$(cmd)` it also complains about that. I've tried to use a backticks for running commands, but it still fails the first complain in this list and even if I break the string to make shellcheck happy, it still complains with much useful clues about this line: `	update_config '.process.args = ["/bin/true"] | `. Which, honestly, couldn't find out why.

But I really think the SC2016 check is quite simple and not compatible with several valid uses of `update_config()`, like using single quotes to properly delimit the $1 param and if you grep the runc code for `disable=SC2016`, you will see that it is quite a lot out there, several times when using `update_config()`.

So, I'll add a comment saying that this checks fails for severa valid uses of update_config() where we use single quotes to delimit the $1 arg to `update_config()`. Let me know what you think.

tests/integration/seccomp-notify.bats
rata: Only the first test doesn't require this. I've moved this to the `setup()` function and created a new file for the first file (tests that we give a clear error msg on older kernels). Let me know what you think

tests/integration/seccomp-notify.bats
rata: Right, removed! When we opened the PR the teardown was usually called in other functions in the code, I think this is why this was here in the first place. But this has changed in runc since then :)

tests/integration/helpers.bash
rata: Fixed, thanks!

libcontainer/process_linux.go
rata: Hmm, you mean `writeSync()`? In any case, no, I don't see how to move the error wrapping there.

We can do error wrapping there and keep `writeSync()` returning an error, but when calling a function here we _need_ to check if it returns an error or not, and act accordingly.

Am I missing something?

rata:
@kolyshkin thanks, I've addressed all the review comments (except [this one][link] that not sure I follow). PTAL :)

[link]: https://github.com/opencontainers/runc/pull/2682#discussion_r682821640

rata:
Oh, it seems the cirrus checks are not run on my fork so that's why it doesn't work. Will see that, but please have a look. The runc code, except for some added things to the cirrus CI, won't need to change :)

rata:
Ok, the issue should be fixed now :)

It was failing because, when splitting the e2e test for old kernels in a different file (for [this comment][link]), I added a check for the actual error msg (not just the return code as it was before). The error msg was not what I expected, therefore it failed. However, the error runc show before was not clear. So, I've kept the expected error (clear error msg output) in the test and fixed the runc code to show a clear error. See the test in `tests/integration/seccomp-notify-compat.bats` for the error msg.

@kolyshkin PTAL, thanks!

[link]: https://github.com/opencontainers/runc/pull/2682#discussion_r683560945

script/install-libseccomp.sh
```diff
@@ -0,0 +1,34 @@
...
+if [ -z ${@+x} ]; then
+	HOSTS=(
+		"x86_64-linux-gnu"
+		"arm-linux-gnueabi"
+		"arm-linux-gnueabihf"
+		"aarch64-linux-gnu"
+		"powerpc64le-linux-gnu"
+	)
+else
+	HOSTS="$@"
```
rata, line 16: Can probably remove the else now, no one is calling with a parameter this script

rata: Removed it

Dockerfile
```diff
@@ -21,22 +21,20 @@ RUN echo 'deb https://download.opensuse.org/repositories/devel:/tools:/criu/Debi
...
-        libseccomp2 \
         pkg-config \
         python-minimal \
         sudo \
         uidmap \
     && apt-get clean \
     && rm -rf /var/cache/apt /var/lib/apt/lists/* /etc/apt/sources.list.d/*.list
 
+COPY script/install-libseccomp.sh /
+RUN /install-libseccomp.sh
```
rata, line 36: We will be able to simplify this when:
 * debian bullseye is released, estimated for Aug 14 (date has been moving, not sure if that is expected again or not) 
 * _And_ if golang releases a new image using bullseye as base

In that case, we can just revert this chunk and update to bullseye. But not sure if there are other things to consider when changing the dockerfile to bullseye.

I can add a comment with that here, so we don't forget

rata: Added a comment for this in the same commit, next to the `FROM` line that currently uses buster

libcontainer/sync.go
```diff
@@ -47,7 +53,13 @@ func (i initError) Error() string {
...
 // if there was a problem writing the payload.
 func writeSync(pipe io.Writer, sync syncType) error {
-	return utils.WriteJSON(pipe, syncT{sync})
+	return utils.WriteJSON(pipe, syncT{sync, -1})
+}
+
+// writeSyncWithFd is used to write to a synchronisation pipe. An error is
+// returned if there was a problem writing the payload.
+func writeSyncWithFd(pipe io.Writer, sync syncType, fd int) error {
+	return utils.WriteJSON(pipe, syncT{sync, fd})
```
kolyshkin, line 62: So,

1. `writeSync` can call `writeSyncWithFd(pipe, sync, -1)`.

2. `writeSyncWithFd` can wrap the error from `WriteJSON` so it's clear as to what is going on, e.g.

```go
func writeSyncWithFd(pipe io.Writer, sync syncType, fd int) error {
	err := utils.WriteJSON(pipe, syncT{sync, fd})
	if err == nil {
		return nil
	}
	return fmt.Errorf("writing syncT %q: %w", string(sync), err)
}
```

libcontainer/process_linux.go
kolyshkin: Please see https://github.com/opencontainers/runc/pull/2682/commits/91ade3538df28dd920b5dc27cc9310d8c07f1922#r686185793

tests/integration/seccomp-notify.bats
kolyshkin: Yes, this was cleaned up some time ago :)

libcontainer/sync.go
rata: Ohh, sorry, I didn't get that. I've changed it to that and made sure that all callers just return the error (only paths in this PR needed to be adjusted to avoid errors like: `writing syncT SeccompDone: writing syncT SeccompDone: xx` and have just `writing syncT SeccompDone: xx`)

Thanks!

rata:
@kolyshkin Improved that too, thanks! PTAL :)

Dockerfile
```diff
@@ -1,6 +1,13 @@
 ARG GO_VERSION=1.16
 ARG BATS_VERSION=v1.3.0
 
+# XXX: Debian buster includes libseccomp < 2.5.0 and we need >= 2.5.0. For that
+# reason, we use the script "script/install-libseccomp.sh" to install a newer
+# libseccomp here. However, Debian bullseye will include libseccomp 2.5.1.
+# Therefore, whenever bullseye is released and the go team publishes an image on
+# that, we can use it and remove the script/install-libseccomp.sh script
```
kolyshkin, line 8: bullseye golang images are already available -- can we switch to bullseye?

Dockerfile
kolyshkin: Debian Bullseye is now available -- see https://github.com/opencontainers/runc/pull/3171 -- so maybe it's better to use it.

contrib/cmd/seccompagent/README.md
```diff
@@ -0,0 +1,57 @@
...
+mkdir container-seccomp-notify
+cd container-seccomp-notify
+mkdir rootfs
+docker export $(docker create busybox) | tar -C rootfs -xvf -
+```
+
+Run the seccomp agent in the background:
+```bash
+sudo ./contrib/cmd/seccompagent/seccompagent &
+```
```
kolyshkin, line 29: I think it makes sense to put "Run the seccomp agent ..." before "Prepare a container" as otherwise the command to run the agent won't work (since we earlier changed the directory).

contrib/cmd/seccompagent/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Copy the config.json in this directory to use for the runc container and start it:
```
kolyshkin, line 31: Not sure what you mean by "copy the config.json". A leftover from the older version?

kolyshkin: Ah, now I see. Maybe be more specific, like

```suggestion
Copy the example `config.json` file from the directory where this README.md is to the container
directory you prepared earlier (`container-seccomp-notify`), then start a container:```

File: contrib/cmd/seccompagent/config.json [hunk 41]
```diff
@@ -0,0 +1,197 @@
...
+	"process": {
+		"terminal": true,
+		"user": {
+			"uid": 0,
+			"gid": 0
+		},
+		"args": [
+			"sh",
+			"-c",
+			"set -x; cd /dev/shm; mkdir test-dir; touch test-file; chmod 777 test-file; ls -l /dev/shm; echo \"Note the agent added a suffix for the directory name and chmod fails\" "
```
kolyshkin, line 12: Perhaps it makes sense to add comments to at least mark those parts of the file that are specific to this demo. This includes this `args` section, and `seccomp` below. Anything else?

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+var (
+	socketFile string
+	pidFile    string
+)
+
+func init() {
+	flag.StringVar(&socketFile, "socketfile", "/run/seccomp-agent.socket", "Socket file")
+	flag.StringVar(&pidFile, "pid-file", "", "Pid file")
+	logrus.SetLevel(logrus.DebugLevel)
+}
```
kolyshkin, line 32: Any reason to not do all this in `main()`?

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		// We found the seccompFdName twice. Error out!
+		if name == specs.SeccompFdName && idx != -1 {
+			err = true
+		}
+	}
+
+	if idx == -1 || err {
+		return 0, fmt.Errorf("seccomp fd not found or malformed containerProcessState.Fds")
```
kolyshkin, line 62: nit: `errors.New`

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	if idx >= len(recvFds) || idx < 0 {
+		return 0, fmt.Errorf("seccomp fd index out of range")
```
kolyshkin, line 66: nit: `errors.New`

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		unix.Close(i)
+	}
+}
+
+// parseStateFds returns the seccomp-fd and closes the rest of the fds in recvFds.
+// In case of error, no fd is closed.
+// StateFds is assumed to be formated as specs.ContainerProcessState.Fds and
+// recvFds the corresponding list of received fds in the same SCM_RIGHT message.
+func parseStateFds(stateFds []string, recvFds []int) (uintptr, error) {
+	// Lets find the index in stateFds of the seccomp-fd.
```
kolyshkin, line 45: Nit: `Let's`

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		case "mkdir":
+			fileName, err := readArgString(req.Pid, int64(req.Data.Args[0]))
+			if err != nil {
+				logrus.Errorf("Cannot read argument: %s", err)
+				resp.Error = int32(unix.ENOSYS)
+				resp.Val = ^uint64(0) // -1
+				goto sendResponse
+			}
+
+			logrus.Debugf("mkdir: %q\n", fileName)
```
kolyshkin, line 207: `\n` is not needed for logrus calls -- a newline is added by logrus.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		if err != nil {
+			logrus.Errorf("Error in NotifReceive(): %s", err)
+			continue
+		}
+		syscallName, err := req.Data.Syscall.GetName()
+		if err != nil {
+			logrus.Errorf("Error decoding syscall %v(): %s", req.Data.Syscall, err)
+			continue
+		}
+		logrus.Debugf("Received syscall %q, pid %v, arch %q, args %+v\n", syscallName, req.Pid, req.Data.Arch, req.Data.Args)
```
kolyshkin, line 182: `\n` is not needed for logrus calls -- a newline is added by logrus.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+
+func main() {
+	// Parse arguments
+	flag.Parse()
+	if flag.NArg() > 0 {
+		flag.PrintDefaults()
+		logrus.Fatal("Invalid command")
+	}
+
+	if err := os.RemoveAll(socketFile); err != nil {
```
kolyshkin, line 243: `RemoveAll` is a recursive function, there's no sense to use it here.

kolyshkin: IOW
```suggestion
	if err := os.Remove(socketFile); err != nil && !errors.Is(err, os.ErrNotExist) {
```

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		logrus.Fatalf("Cannot cleanup socket file %s: %v", socketFile, err)
```
kolyshkin, line 244: The error message from `os.Remove` (or `os.RemoveAll`) already contains the file name, no need to have it twice.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	if pidFile != "" {
+		pid := fmt.Sprintf("%d\n", os.Getpid())
+		if err := ioutil.WriteFile(pidFile, []byte(pid), 0o644); err != nil {
+			logrus.Fatalf("Cannot write pid file %s: %v", pidFile, err)
```
kolyshkin, line 250: The error message from `ioutil.WriteFile` already contains the file name; no need to have it twice.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+	}
+
+	logrus.Info("Waiting for seccomp file descriptors")
+	l, err := net.Listen("unix", socketFile)
+	if err != nil {
+		logrus.Fatalf("Cannot listen on %s: %s", socketFile, err)
```
kolyshkin, line 257: The error from `net.Listen` should already contains the path; no need to have it twice.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+	for {
+		conn, err := l.Accept()
+		if err != nil {
+			logrus.Errorf("Cannot accept connection: %s", err)
+			continue
+		}
+		socket, err := conn.(*net.UnixConn).File()
+		conn.Close()
+		if err != nil {
+			logrus.Errorf("Cannot get socket: %v\n", err)
```
kolyshkin, line 270: No need to have `\n` is logrus messages.

contrib/cmd/seccompagent/seccompagent.go
```diff
@@ -0,0 +1,292 @@
...
+		// Make sure we don't allow strings like "/../p", as that means
+		// a file in a different location than expected. We just want
+		// safe things to use as a suffix for a file name.
+		metadata = filepath.Base(metadata)
+		if strings.Contains(metadata, "/") {
+			// Fallback to a safe string.
+			metadata = "agent-generated-suffix"
+		}
+
+		logrus.Infof("Received new seccomp fd: %v\n", newFd)
```
kolyshkin, line 289: No need to have `\n` is logrus messages.

contrib/cmd/seccompagent/config.json
kolyshkin: Or, add a short script instead that would modify config.json to achieve what you need. You can see lots of examples of how to do it with `jq` in `tests/integration`

tests/integration/seccomp.bats
```diff
@@ -33,3 +33,71 @@ function teardown() {
...
+		.process.noNewPrivileges = false |
+		.linux.seccomp = {
+			"defaultAction":"SCMP_ACT_ALLOW",
+			"architectures":["SCMP_ARCH_X86","SCMP_ARCH_X32"],
+			"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO"}]
+		}'
+
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+	[[ "$output" == *"Operation not permitted"* ]]
```
kolyshkin, line 52: Maybe make the test more strict and look for something like
`*"mkdir:"*"/dev/shm/foo"*"Operation not permitted"*`
in order to not have a false pass when runc merely prints "Operation not permitted" (for whatever reason).

tests/integration/seccomp.bats
```diff
@@ -33,3 +33,71 @@ function teardown() {
...
 }
+
+# TODO:
+# - Test other actions like SCMP_ACT_TRAP, SCMP_ACT_TRACE, SCMP_ACT_LOG.
+# - Test args (index, value, valueTwo, etc).
+
+@test "runc run [seccomp] (SCMP_ACT_ERRNO default)" {
+	update_config '.process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"] |
+		.process.noNewPrivileges = false |
+		.linux.seccomp = {
```
kolyshkin, line 44: Not matters much, but elsewhere in tests/integration we usually use slightly different format of multiline `update_config` statements. See e.g.
https://github.com/opencontainers/runc/blob/9835e9c6b2959ccf40c23c215202f9a9b1d9644c/tests/integration/cgroups.bats#L169-L173

(the first line is tab and two spaces, the rest of them just tabs and `| ` etc.)

tests/integration/helpers.bash
```diff
@@ -16,6 +16,7 @@ unset IMAGES
 
 RUNC="${INTEGRATION_ROOT}/../../runc"
 RECVTTY="${INTEGRATION_ROOT}/../../contrib/cmd/recvtty/recvtty"
+SECCOMPAGENT="${INTEGRATION_ROOT}/../../contrib/cmd/seccompagent/seccompagent"
```
kolyshkin, line 19: Can we please have it as `SECCOMP_AGENT` (slightly easier to read)?

contrib/cmd/seccompagent/config.json
kolyshkin: Either way is fine with me -- I just want to be able to find out in 15+ years which parts are specific to the test, and which are whatever `runc spec` generated back in 2021 :)

tests/integration/seccomp.bats
```diff
@@ -33,3 +33,71 @@ function teardown() {
...
+			"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO"}]
+		}'
+
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+	[[ "$output" == *"Operation not permitted"* ]]
+}
+
+@test "runc run [seccomp] (SCMP_ACT_ERRNO explicit errno)" {
+	update_config '.process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"] |
```
kolyshkin, line 56: Same as https://github.com/opencontainers/runc/pull/2682/commits/fdb3c255232171af99e91590cc2b3daf7c0c4691#r691717884

tests/integration/seccomp.bats
```diff
@@ -33,3 +33,71 @@ function teardown() {
...
+			"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO", "errnoRet": 100}]
+		}'
+
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+	[[ "$output" == *"Network is down"* ]]
+}
+
+@test "runc run [seccomp] (SCMP_ACT_KILL)" {
+	update_config '.process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"] |
```
kolyshkin, line 70: Same as https://github.com/opencontainers/runc/pull/2682/commits/fdb3c255232171af99e91590cc2b3daf7c0c4691#r691717884

tests/integration/seccomp.bats
```diff
@@ -33,3 +33,71 @@ function teardown() {
...
+			"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_KILL"}]
+		}'
+
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+}
+
+# check that a startContainer hook is run with the seccomp filters applied
+@test "runc run [seccomp] (startContainer hook)" {
+	update_config '.process.args = ["/bin/true"] |
```
kolyshkin, line 84: Same as https://github.com/opencontainers/runc/pull/2682/commits/fdb3c255232171af99e91590cc2b3daf7c0c4691#r691717884

tests/integration/seccomp-notify.bats
```diff
@@ -0,0 +1,202 @@
...
+}
+
+# Create config.json template with SCMP_ACT_NOTIFY actions
+# $1: command to run
+# $2: noNewPrivileges (false/true)
+# $3: list of syscalls
+function scmp_act_notify_template() {
+	# The agent intercepts mkdir syscalls and creates the folder appending
+	# "-bar" (listenerMetadata below) to the name.
+	update_config '.process.args = ["/bin/sh", "-c", "'"$1"'"] |
```
kolyshkin, line 30: Same as https://github.com/opencontainers/runc/pull/2682/commits/fdb3c255232171af99e91590cc2b3daf7c0c4691#r691717884

tests/integration/seccomp-notify-compat.bats
```diff
@@ -0,0 +1,35 @@
...
+
+# Support for seccomp notify requires Linux > 5.6, check that on older kernels
+# return an error.
+@test "runc run [seccomp] (SCMP_ACT_NOTIFY old kernel)" {
+	if [[ "$KERNEL_MAJOR" -gt 5 || ("$KERNEL_MAJOR" -eq 5 && "$KERNEL_MINOR" -ge 6) ]]; then
+		skip "requires kernel less than 5.6"
+	fi
+
+	# Use just any seccomp profile with a notify action.
+	update_config '.linux.seccomp = {
```
kolyshkin, line 25: Same as https://github.com/opencontainers/runc/pull/2682/commits/fdb3c255232171af99e91590cc2b3daf7c0c4691#r691717884

tests/integration/seccomp-notify-compat.bats
```diff
@@ -0,0 +1,35 @@
...
+	teardown_seccompagent
+	teardown_bundle
+}
+
+# Support for seccomp notify requires Linux > 5.6, check that on older kernels
+# return an error.
+@test "runc run [seccomp] (SCMP_ACT_NOTIFY old kernel)" {
+	if [[ "$KERNEL_MAJOR" -gt 5 || ("$KERNEL_MAJOR" -eq 5 && "$KERNEL_MINOR" -ge 6) ]]; then
+		skip "requires kernel less than 5.6"
+	fi
```
kolyshkin, line 22: Maybe move this to `setup()` so it won't do any expensive setup if the kernel is not suitable?

kolyshkin (changes requested):
We're getting there! The only major change that would be nice to have is switching to bullseye to avoid installing libseccomp from Dockerfile. Everything else is just me nitpicking.

contrib/cmd/seccompagent/README.md
rata: Right, moving it

contrib/cmd/seccompagent/README.md
rata: Added, thanks!

contrib/cmd/seccompagent/config.json
rata: yes, it is a config as generated by runc spec, with only the args and seccomp section modified. I stated that in the README so it is easy for our future selfs :-) (json can't include comments AFAIK).

contrib/cmd/seccompagent/seccompagent.go
rata: Nope, moving it to main.

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed (and checked all the occurrences in all commits, but these two were all :))

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed. I think when I rebase the CI might find more typos now, will fix anything that catches too :)

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed, thanks

contrib/cmd/seccompagent/seccompagent.go
rata: Fixed, thanks

contrib/cmd/seccompagent/seccompagent.go
rata: Right, fixed!

contrib/cmd/seccompagent/seccompagent.go
rata: Right, fixed!

contrib/cmd/seccompagent/seccompagent.go
rata: Right, fixed!

contrib/cmd/seccompagent/seccompagent.go
rata: Right, fixed!

contrib/cmd/seccompagent/seccompagent.go
rata: Right, fixed!

contrib/cmd/seccompagent/seccompagent.go
rata: Right, fixed!

(You can just say to do the same on all the rest of the lines that do this and that is enough. No need to spot them all. Assuming doing that is more comfortable for you :))

tests/integration/seccomp.bats
rata: Oh, good idea. Changed it for that now, thanks!

rata: Also, did the analogous changes to all that can be done the same: `@test "runc run [seccomp] (SCMP_ACT_NOTIFY kill seccompagent)`, `@test "runc run [seccomp] (SCMP_ACT_NOTIFY error chmod)`

tests/integration/seccomp.bats
rata: Fixed the style, thanks!

tests/integration/helpers.bash
rata: Sure, fixed

tests/integration/seccomp.bats
rata: Fixed

tests/integration/seccomp.bats
rata: Fixed

tests/integration/seccomp.bats
rata: Fixed

tests/integration/seccomp-notify.bats
rata: Fixed

tests/integration/seccomp-notify-compat.bats
rata: Fixed

tests/integration/seccomp-notify-compat.bats
rata: Sure, moved!

Dockerfile
rata: Ok, chery-picked your commit here!

rata:
@kolyshkin cool, thanks for the review! I addressed all the comments. I've also cherry-picked your commit from #3171 and dropped the commit to download it manually.

All tests are green too. If you want to merge PR #3171 first, I can drop the commit here (or maybe git is smart enough that we don't even need to drop it?)

Dockerfile
rata: Ok, the commit was super simple and I see you just did it. I cherry-picked it here, as you asked [here][link] then :D


[link]: https://github.com/opencontainers/runc/pull/2682#discussion_r691697814

vendor/modules.txt
kolyshkin: I took a look at what it takes to have 0.9.2 and its milestone have quite a few opened issues (see https://github.com/seccomp/libseccomp-golang/milestone/3)

I tried to help a bit, creating a couple of PRs (https://github.com/seccomp/libseccomp-golang/pull/65, https://github.com/seccomp/libseccomp-golang/pull/66) and asked a question in https://github.com/seccomp/libseccomp-golang/issues/37#issuecomment-891488113. Alas, no reaction for the last 2 weeks :(

kolyshkin: In general, I think, looking into https://github.com/seccomp/libseccomp-golang/milestone/3 and trying to address some issues may help releasing 0.9.2.

rata: @kolyshkin nice, thanks! But AFAIK the lack of 0.9.2 doesn't limit this PR in any way, right? It will be nice to have 0.9.2, but AFAIK it won't happen anytime soon.

@AkihiroSuda also asked in this PR: https://github.com/seccomp/libseccomp-golang/pull/59#issuecomment-885756065, the response was to help with remaining items as you did :)

rata:
PR #3171 was just merged, so I rebased this. I hope this is ready to go in now :crossed_fingers:

kolyshkin:
@rata can you modify the PR description to include a proposed changelog entry for runc 1.1.0? A few sentences telling about the feature, it's requirements and possible uses. A draft is fine -- maintainers may edit it later.

rata:
@kolyshkin sure!

kolyshkin (dismissed):
LGTM

rata:
@kolyshkin added the changelog description with all those items!

@AkihiroSuda maybe you want to review again now? :)

The centos-stream-8 failure seems unrelated to this PR (the same code did past the tests on my fork). It was stuck trying to installing dependencies and it seems there are several retries already in place to work-around some issues. It might be related to that

kolyshkin:
centos-stream-8 CI hit a snag; restarted

AkihiroSuda (approved):

AkihiroSuda:
@cyphar PTAL

rata:
Friendly ping?

Hi! The 1.1.0 milestone is due in ~2 weeks and it will be nice to see this feature there (this is planned for that milestone so far). Is there anything else I can do?

vbatts:
dang. This has had 2 LGTMs, but hoping for a @cyphar review.
Now it needs to be rebased again. 😵 
@opencontainers/runc-maintainers can we get eyes on this?

rata:
I rebased to fix the conflicts, tests are green. But the LGTM are lost after the rebase. @kolyshkin @AkihiroSuda can you please LGTM again? :)

kolyshkin (dismissed):
still LGTM :)

dqminh (approved):

kolyshkin:
@AkihiroSuda this needs your still-LGTM after a rebase.

@cyphar PTAL

AkihiroSuda (dismissed):

contrib/cmd/seccompagent/config.json
cyphar: I think he meant that it would be nice if there was a simple script that did:

```sh
% runc spec
% jq '.process.args = [...]' <config.json | sponge config.json
```

Another option would be to put the test commands in a script and execute the script (copying it into the rootfs in the integration test) which would be even cleaner.

But we can always fix that separately (and having an explanation in the readme is probably sufficient).

rata: Sure, I can do that in a follow-up PR :)

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -15,67 +15,90 @@ import (
...
+	apiLevel, _ := libseccomp.GetAPI()
+	for _, call := range config.Syscalls {
+		if call.Action == configs.Notify {
+			if apiLevel < 6 {
+				return -1, fmt.Errorf("seccomp notify unsupported: API level: got %d, want at least 6. Please try with libseccomp >= 2.5.0 and Linux >= 5.7", apiLevel)
+			}
+
+			if call.Name == "write" {
+				return -1, errors.New("SCMP_ACT_NOTIFY cannot be used for the write syscall")
+			}
```
cyphar, line 56: Is there a reason for this restriction? I can imagine how it might be unsafe (or unwise) to do this on paper, but is it really necessary to disallow this within runc?

cyphar (commented):
Looked through it and it all seems very reasonable -- kolyshkin has probably reviewed this several times more thoroughly than I have. :wink:

I only have one question about the seemingly arbitrary write restriction.

libcontainer/seccomp/seccomp_linux.go
rata: Yes, basically it will hang for ever if we allow it. We need the write syscall to send the seccomp fd over the pipe to the parent (we write to the pipe [here][link]), then the parent gets the fd with pifd_getfd and sends it over the unix socket to the seccomp agent.

If the write syscall has the notify action, we can't never send a msg to the parent, which will mean the seccomp agent won't get the seccomp fd and the write will be blocked forever.

However read/close (functions that we also use to sync with parent via the pipe) are not restricted as it is possible to send the seccomp fd to the parent, the seccomp agent gets it and the first few notifications it will get if read/close are handled are for the ones done in `syncParentSeccomp()` (and quite trivially allow to proceed syscalls on fds that are not known by the agent, for example).

I will add a comment to explain this in the code, if that sounds good to you @cyphar 

[link]: https://github.com/opencontainers/runc/pull/2682/commits/10ad86bd09910557e2c3680c97971fbfd7b1c60d#diff-84fffb149f8b86c903e09613cbebad12de596f4139d54e288388b1458e08b8ecR288

rata: @cyphar Added a comment explaining this [here][link]


[link]: https://github.com/opencontainers/runc/pull/2682/files#diff-3693bdb9f47092da6d6e35007a1689f74ddc39be8e017d475a593481c437ac3bR55-R73

rata:
Rebased to fix conflicts (a simple conflict with https://github.com/opencontainers/runc/commit/64358c4de9258471b5540c16d104460747d28fed that moved some code we are touching) and added a comment to explain the write restriction @cyphar commented on (see https://github.com/opencontainers/runc/pull/2682#discussion_r703331476).

@kolyshkin @AkihiroSuda @dqminh can you please re-LGTM this? :)

@cyphar thanks for the comment! I guess this is ready to merge? Or is there anything else I can do?

libcontainer/seccomp/seccomp_linux.go
cyphar: Ah of course. Yeah that's a bit sad :/ -- but we can always figure out some workaround for this in the future. (I wonder how the LXC folks worked around this problem -- though I think they don't have a similar write-based synchronisation model since they're written in C which means you don't need to do so much busy work.)

kolyshkin (approved):
(I wish github had a way to see the difference between old and new commits, e.g. similar to what gerrit does)

Trusting the changes are mostly cosmetical, still LGTM

cyphar (approved):
LGTM, thanks for working on this for so long!

rata:
@cyphar @kolyshkin @dqminh @vbatts thanks for your time and reviews! :)

libcontainer/seccomp/seccomp_linux.go
rata: @cyphar I've asked Christian Brauner [in the past](https://lore.kernel.org/lkml/20210301110907.2qoxmiy55gpkgwnq@wittgenstein/). It seems they use clone3 sharing the fds table and use a fixed number for the seccomp notify fd. Then, they get it with pidfd_getfd().

Not sure how LXC uses a fixed fd number for seccomp notify, though.

cyphar: You could just pick an arbitrarily high fd and use `dup3` to clone it to the right fd number. It might be interesting to look into whether we could do it with `pidfd_getfd` though since we'd need to have a fallback for quite a while, it'd probably complicate the code a bit too much...

rata: Oh, right, that was easy. Yeah, for runc I think is simpler to use the pipe, that is the main sync mechanism used tree-wide. We are using `pidfd_getfd` here, though: we sync with the parent to let them know when the fd is ready and send the fd plain number, then the parent gets it with `pidfd_getfd`

cyphar: Ah, I forgot that we are requiring Linux 5.7 or newer for the feature -- pidfd_getfd isn't available for the earlier kernels with seccomp notify support.

rata: And 5.7 is also enforced by libseccomp-golang, that only supports notification with tsync (requires libseccomp api level 6). This is as the vast majority of go programs will need tsync too. [Here](https://github.com/seccomp/libseccomp-golang/pull/59#discussion_r615949270) is the conversation we had with upstream about this when writing the patches for libseccomp-golang

Before 5.7, notify and tsync where mutually exclusive (torvalds/linux@7a0df7f). From Linux 5.7 introduced a new API to enable seccomp notification with tsync together in this commit torvalds/linux@5189149.

rata:
@kolyshkin I've updated the release notes for this. I realized I forgot to mention the seccomp agent example in the runc source code! Updated to mention that now.

PR #2685, opened 2020-11-19
libctr/init_linux: reorder chdir

haircommander:
commit https://github.com/opencontainers/runc/commit/5e0e67d76cc99d76c8228d48f38f37034503f315 moved the chdir to be one of the
first steps of finalizing the namespace of the container.

However, this causes issues when the cwd is not accessible by the user running runc, but rather
as the container user.

Thus, setupUser has to happen before we call chdir. setupUser still happens before setting the caps,
so the user should be privileged enough to mitigate the issues fixed in https://github.com/opencontainers/runc/commit/5e0e67d76cc99d76c8228d48f38f37034503f315 (I've tested it on my machine, so I believe it does not regress)

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -129,12 +129,6 @@ func finalizeNamespace(config *initConfig) error {
 		return errors.Wrap(err, "close exec fds")
 	}
 
-	if config.Cwd != "" {
-		if err := unix.Chdir(config.Cwd); err != nil {
-			return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
-		}
-	}
-
 	capabilities := &configs.Capabilities{}
 	if config.Capabilities != nil {
 		capabilities = config.Capabilities
@@ -155,6 +149,14 @@ func finalizeNamespace(config *initConfig) error {
 	}
 	if err := setupUser(config); err != nil {
 		return errors.Wrap(err, "setup user")
+	}
+	// Change working directory AFTER the user has been set up.
+	// Otherwise, if the cwd is also a volume that's been chowned to the container user (and not the user running runc),
+	// this command will EPERM.
+	if config.Cwd != "" {
+		if err := unix.Chdir(config.Cwd); err != nil {
+			return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
+		}
 	}
 	if err := system.ClearKeepCaps(); err != nil {
 		return errors.Wrap(err, "clear keep caps")
--- a/tests/integration/cwd-priv.bats
+++ b/tests/integration/cwd-priv.bats
@@ -1,0 +1,44 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# This test must be particular with how it's run. The user that runs it must have the privileges
+# to chown a directory it creates away from itself, but not have CAP_DAC_OVERRIDE to override
+# the fact it was chowned away. However, it must also be able to clean up.
+# Thus, this test must be skipped if the UID running the test is root, or the user doesn't have sudo privileges.
+function setup() {
+	if [ "$EUID" -eq 0 ]; then
+		skip "This test must be run as a non-root user"
+	fi
+	if ! sudo -n -v; then
+		skip "This test must be run as a user with sudo privileges"
+	fi
+
+	teardown_busybox
+	setup_busybox
+
+	SOURCE=$(mktemp -d "$BATS_TMPDIR/cwd.XXXXXX")
+}
+
+function teardown() {
+	sudo -n rm -rf "$SOURCE"
+	teardown_busybox
+}
+
+# Verify a cwd owned by the container user can be chdir'd to,
+# even if runc doesn't have the privilege to do so.
+@test "runc create sets up user before chdir to cwd" {
+	USER=10000001
+	sudo -n chown $USER:$USER "$SOURCE"
+	ls -ld "$SOURCE" >&3
+	# shellcheck disable=SC2016
+	update_config ' .mounts += [{"source": "'"$SOURCE"'", "destination": "'"$SOURCE"'", "options": ["bind"]}]
+				| .process.user.uid = $USER
+				| .process.user.gid = $USER
+				| .process.cwd = "'"$SOURCE"'"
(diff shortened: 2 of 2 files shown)


thaJeztah:
I see there was some discussion around the original change in https://github.com/opencontainers/runc/pull/2086#issuecomment-591801550 as well


@cyphar @justincormack @crosbymichael ptal

haircommander:
```
Warning: apt-key output should not be parsed (stdout is not a terminal)

https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/Release.key:

2020-11-19 18:37:41 ERROR 404: Not Found.

gpg: no valid OpenPGP data found.

The command '/bin/sh -c echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_Unstable/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list     && wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_Unstable/Release.key -O- | sudo apt-key add -     && apt-get update     && apt-get install -y --no-install-recommends skopeo     && rm -rf /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list     && apt-get clean     && rm -rf /var/cache/apt /var/lib/apt/lists/*;' returned a non-zero code: 2
```

this doesn't seem related. can someon retrigger the tests?

also, PTAL @kolyshkin @mrunalp @AkihiroSuda

thaJeztah:
been running into that as well; /cc @cyphar

kolyshkin:
Proposed CI fix is in https://github.com/opencontainers/runc/pull/2686. Once merged, you'll need to rebase to fix CI.

haircommander:
rebased, thanks @kolyshkin

libcontainer/init_linux.go
```diff
@@ -156,6 +150,14 @@ func finalizeNamespace(config *initConfig) error {
 	if err := setupUser(config); err != nil {
 		return errors.Wrap(err, "setup user")
 	}
+	// Change workding directory AFTER the user has been set up.
```
mrunalp, line 153: typo: `working`

haircommander: fixed!

mrunalp (dismissed):

kolyshkin:
@haircommander this seems like a regression. Can we have a test case?

haircommander:
I am having trouble writing a test (I personally have not created a situation where this bug comes up, only had it come up in someone else's node in a pretty specific scenario). I'll keep trying, but ideally we could go forward with this without it for now

haircommander:
ok the test is written--it's not pretty but it exists. PTAL @kolyshkin

tests/integration/cwd-priv.bats
```diff
@@ -0,0 +1,45 @@
...
+	USER=10000001
+	sudo chown $USER:$USER "$SOURCE"
+	ls -ld "$SOURCE" >&3
+	update_config " .mounts += [{\"source\": \"$SOURCE\", \"destination\": \"$SOURCE\", \"options\": [\"bind\"]}]
+				| .process.user.uid = $USER
+				| .process.user.gid = $USER
+				| .process.cwd = \"$SOURCE\"
+				| .process.args |= [\"ls\", \"$SOURCE\"]"
+
+	output=$(runc run test_busybox 2>&1)
```
kolyshkin, line 43: In here `runc` is already a wrapper that uses `run`, so
```suggestion
	runc run test_busybox
```

tests/integration/cwd-priv.bats
```diff
@@ -0,0 +1,45 @@
...
+	teardown_busybox
+}
+
+# Verify a cwd owned by the container user can be chdir'd to,
+# even if runc doesn't have the privilege to do so.
+@test "runc create sets up user before chdir to cwd" {
+	USER=10000001
+	sudo chown $USER:$USER "$SOURCE"
+	ls -ld "$SOURCE" >&3
+	update_config " .mounts += [{\"source\": \"$SOURCE\", \"destination\": \"$SOURCE\", \"options\": [\"bind\"]}]
```
kolyshkin, line 37: if you can use single quotes around the whole jq statement, you won't have to escape the double quotes. Shell env vars can still be used, e.g.

```suggestion
	update_config ' .mounts += [{"source": "'"$SOURCE"'", "destination": "'"$SOURCE"'", "options": ["bind"]}]
...

File: tests/integration/cwd-priv.bats (hunk 2 above)
Comment (kolyshkin) [reply tests/integration/cwd-priv.bats:43]: and we should probably also check that `runc` succeeds, e.g.
```bash
[ "$status" -eq 0 ]
```

kolyshkin:
CI is failing on rootless tests
```
....
ok 21 runc create --pid-file with new CWD
[sudo] password for rootless: 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

Maybe `sudo` needs `-n`?

haircommander:
@kolyshkin @mrunalp @AkihiroSuda @cyphar @justincormack @crosbymichael PTAL

kolyshkin:
CI went south :(

> GitHub Actions has encountered an internal error when running your job.

tests/integration/cwd-priv.bats
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# This test must be particular with how it's run. The user that runs it must have the privileges
+# to chown a directory it creates away from itself, but not have CAP_DAC_OVERRIDE to override
+# the fact it was chowned away. However, it must also be able to clean up.
+# Thus, this test must be skipped if the UID running the test is root, or the user doesn't have sudo privileges.
```
kolyshkin, line 8: I don't quite get it. Can this test be run as root, and then drop the privs to a non-root user when it needs to?

Asking because the way it is right now, our CI never runs it.

haircommander: unfortunately, I wasn't able to figure out how to run the test as a non privileged user

kolyshkin:
Also related: https://github.com/containerd/containerd/pull/4669

kolyshkin:
Carrying this over in https://github.com/opencontainers/runc/pull/2712

PR #2689, opened 2020-11-24
libct/cgroup/utils: fix GetCgroupMounts(all=true)

kolyshkin:
The `all` argument was introduced by commit f5579964014 (PR #1049) specifically
for use by cAdvisor (see https://github.com/google/cadvisor/pull/1476), but there were no test cases added,
so it was later accidentally broken by 5ee0648bfbe037bcf4b (#1817) which started incrementing
`numFound` unconditionally.

Fix this (by not checking `numFound` when `all` is true), and add a
simple test case to avoid future regressions.

PS It appears to be that cadvisor is the only user of `all=true` functionality.
If we could eliminate it, we would remove `all=true` entirely, but it's not clear
how to do that. I sincerely hope one day cgroup v1 is going away entirely,
including all this code.

### DIFF
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -181,8 +181,11 @@ const cgroup2Mountinfo = `18 64 0:18 / /sys rw,nosuid,nodev,noexec,relatime shar
 
 func TestGetCgroupMounts(t *testing.T) {
 	type testData struct {
-		mountInfo  string
-		root       string
+		mountInfo string
+		root      string
+		// all is the total number of records expected with all=true,
+		// or 0 for no extra records expected (most cases).
+		all        int
 		subsystems map[string]bool
 	}
 	testTable := []testData{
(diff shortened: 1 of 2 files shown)


AkihiroSuda (dismissed):

AkihiroSuda:
@mrunalp PTAL

kolyshkin:
Rebased to include new CI checks; no changes to the code

AkihiroSuda (approved):

dqminh:
LGTM.

But i also don't see why cadvisor needs to use `all` and filter the cgroup out by itself, perhaps it can just use the current API and things will work (?)

dqminh (approved):

kolyshkin:
> But i also don't see why cadvisor needs to use all and filter the cgroup out by itself, perhaps it can just use the current API and things will work (?)

It's complicated :-\ I tried to get rid of it but I'm afraid of re-introducing https://github.com/google/cadvisor/issues/1461

Issue #2512, opened 2020-07-08
CI: move to GitHub Actions?


AkihiroSuda:
Travis often fails to report the CI status these days :cry:
GitHub Actions might be more stable.

cyphar:
I'm currently moving umoci's CI to GitHub Actions (despite my dislike for making our processes locked into GitHub). My current experience is that it's "fine", though quite a few aspects of setting up CI workflows are far more frustrating than necessary...

cyphar:
After spending a whole day to [port umoci's CI to GitHub Actions](https://github.com/cyphar/umoci-test/actions/runs/162612101) I found that it's the worst CI system I've ever used. I'm sure there are some that are worse, but every aspect of getting umoci's CI to work was incredibly painful (not to mention the end result is still incredibly slow even compared to Travis which doesn't have any of my caching optimisations). To be fair, umoci's CI is a bit more complicated than runc's but I suspect it wouldn't be much better. Even basic things like getting the logs of a currently-running action are dodgy. 

After complaining about this, a colleague of mine (who's in the Kubernetes world) mentioned that CircleCI is much better designed for this kind of thing -- so maybe we should look at them instead of GitHub actions?

AkihiroSuda:
@cyphar What are the actual problems of GH Actions?

> CircleCI

It doesn't support nested VM AFAIK.
Also, its free(-as-in-beer) edition has tight limitation: https://circleci.com/pricing/#comparison-table

cyphar:
> What are the actual problems of GH Actions?

Well, here's a short list which affect umoci:

 * The documentation is lacking. It took me a fair while to configure the CI, mostly because the documentation (in true GitHub fashion) doesn't provide a proper reference text for the features and semantics. You have to 
   - This is compounded by the fact that many **incredibly basic** features like "check out the code of this repository" are not extensively documented by the Workflows documentation, so you have to look through the https://github.com/actions/checkout README in order to understand what all the parameters do (and even then the README is missing information as well).
   - Lots of the documentation links take you to the GitHub API documentation which doesn't actually answer what the link says (for instance, try to figure out what each of the possible event types mean in the `on` block -- all of the links take you to the GitHub API which is written from the perspective of a GitHub App and the set of events in the API are different to the ones available in Workflows).
[...]

PR #2487, opened 2020-06-25
Improve nsexec logging

kolyshkin:
This is a carry of https://github.com/opencontainers/runc/pull/2460 with a few fixes on top:

 - rebased on top of current master (resolved conflicts due to #2481);
 - fixed C formatting in nsexec.c to satisfy `make validate`;
 - added JSON string escaping added (fixes debug.bats failures);
 - rebased on top of #2835
 
 Keeping this one a draft until #2835 is merged.
 
Closes: #2460

PR #2690, opened 2020-11-24
ci: add integration / unit tests to github actions

kolyshkin:
This is Ubuntu 20.04 so alas cgroup v1 testing only.

The upside is it's pretty fast :fast_forward: we have results of unit + integration tests
for both go 1.14 and 1.15 within 5 minutes. This seems useful now when travis-ci.org
is throttling jobs :disappointed: 

rootless tests take longer time so we run them in parallel
with non-rootless ones. Also, to balance the time, we run
unit tests when rootless is not set. IOW, 'rootless' in job
name means it's only rootless integration tests, and no rootless
means it's unit tests, integration tests with fs driver, and
integration tests with systemd driver.
    
Note, `script` is used to run integration tests to provide a tty
(which by default is not available on gh actions).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,82 @@
+name: ci
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version: [1.14.x, 1.15.x]
+        rootless: ["rootless", ""]
+
+    steps:
+
+    - name: checkout
+      uses: actions/checkout@v2
+
+    - name: install deps
+      run: |
+        # skopeo repo
+        echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/ /' | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+        wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_10/Release.key -O- | sudo apt-key add -
+        # criu repo
+        sudo add-apt-repository -y ppa:criu/ppa
+        # apt-add-repository runs apt update so we don't have to
+        sudo apt -q install libseccomp-dev criu skopeo
+
+    - name: install go ${{ matrix.go-version }}
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go-version }}
+
+    - name: build
+      run: sudo -E PATH="$PATH" make all
+
+    - name: install umoci
+      run: |
+        mkdir ~/bin
+        echo "PATH=$HOME/bin:$PATH" >> $GITHUB_ENV
+        curl -o ~/bin/umoci -fsSL https://github.com/opencontainers/umoci/releases/download/v0.4.6/umoci.amd64
+        chmod +x ~/bin/umoci
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: install deps
+      run: |
+        sudo apt update
+        sudo apt install libseccomp-dev
+    - name: install criu
+      env:
+        VERSION: 3.15
+      run: |
+        mkdir criu
+        curl -fsSL https://github.com/checkpoint-restore/criu/archive/${{ matrix.criu-version }}.tar.gz | tar -C criu -xz --strip-components=1
```
mauriciovasquezbernal, line 31: `v` is missing before `${{ matrix.criu-version }}`.

kolyshkin: thanks, fixed! this is still a WIP as I am trying to make it work

kolyshkin:
OK, half the tests are working now, the other half complains:

> time="2020-11-27T11:40:27Z" level=error msg="open /dev/tty: no such device or address"

One way to fix is to use run tests under docker.

There's a bug with lots of comments about it: https://github.com/actions/runner/issues/241

Will take a look later.

kolyshkin:
OK I think this is ready for prime time. I don't mean to replace travis (yet?), it's just in addition to get a faster test results.

@AkihiroSuda @mrunalp PTAL

kolyshkin:
Ahh, interesting...

```
not ok 23 global --debug to --log
# (in test file tests/integration/debug.bats, line 41)
```

kolyshkin:
> Ahh, interesting...
> 
> ```
> not ok 23 global --debug to --log
> # (in test file tests/integration/debug.bats, line 41)
> #   `[[ "${output}" == *"child process in init()"* ]]' failed
> # runc spec --rootless (status=0):
> ```

I barely remember something about this logging being a tad racy, and I've been trying to fix that in https://github.com/opencontainers/runc/pull/2487.

Oh yes, this ugly commit: https://github.com/opencontainers/runc/pull/2487/commits/f9b4cf01f697e101784bd695c05bff4f3a90bb9b

Alas, I haven't finished that PR and I don't even remember why now. Let me try to revive it...

kolyshkin:
> Let me try to revive it...

In the meantime, this PR is fine, the flake in debug logging is known problem (and a very minor one, since this is runc debug logging no one but a few runc devs care about).

kolyshkin:
@AkihiroSuda @mrunalp PTAL

AkihiroSuda (approved):

AkihiroSuda:
FYI previously there was a decision not to migrate to GHA (https://github.com/opencontainers/runc/issues/2512), but I guess the context has changed after Travis changed the free(-as-in-beer) plan. So LGTM.

cc @cyphar WDYT?

kolyshkin:
> FYI previously there was a decision not to migrate to GHA (#2512), but I guess the context has changed after Travis changed the free(-as-in-beer) plan. So LGTM.

Sorry I overlooked it. Indeed, the situation with Travis is unclear, and GH actions are (currently) both fast and free (and in case they won't be, we could ask LF to pay for it).

I am also thinking about using https://cirrus-ci.org/ as it supports Fedora so we can test cgroupv2.

mrunalp (approved):

AkihiroSuda:
> I am also thinking about using https://cirrus-ci.org/ as it supports Fedora so we can test cgroupv2.

Seems to need nested VM? (which is also supported by GHA on macOS instances)

PR #2691, opened 2020-11-27
Dockerfile: use binary criu release; update some ci deps

kolyshkin:
As suggested by @adrianreber in https://github.com/checkpoint-restore/criu/issues/1293#issuecomment-733531226
    
This "reverts" commit 9ff7b82f9. We use the repo provided and maintained
by the CRIU team instead of compiling it from source.
    
While at it,
1. Don't install packages that were solely needed to compile CRIU.
2. Combine all apt-related stuff into a single step, so we don't
       download and then remove apt metadata 3 times.
3. bats: update to 1.2.1
4. umoci: update to 0.4.6

~~TODO:~~
~~* have CRIU make a Debian repo~~
~~* remove packages that are needed to build CRIU~~


Dockerfile
```diff
@@ -58,14 +57,12 @@ RUN cd /tmp \
...
-ARG CRIU_VERSION
-RUN mkdir -p /usr/src/criu \
-    && curl -fsSL https://github.com/checkpoint-restore/criu/archive/${CRIU_VERSION}.tar.gz | tar -C /usr/src/criu/ -xz --strip-components=1 \
-    && cd /usr/src/criu \
-    && echo 1 > .gitid \
-    && make -j $(nproc) install-criu \
-    && cd - \
-    && rm -rf /usr/src/criu
+RUN echo 'deb http://ppa.launchpad.net/criu/ppa/ubuntu/ bionic main' > /etc/apt/sources.list.d/criu.list \
+    && wget -nv 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x4e2a48715c45aeec077b48169b29eec9246b6ce2' -O- | apt-key add - \
```
adrianreber, line 61: Why aren't you using something like 'add-apt-repository -y ppa:criu/ppa'?

adrianreber:
Looks good. The CentOS 7 and Fedora 33 tests were already using my packages and now also the Ubuntu based test. I had a look at the test logs and CRIU is actually installed in all cases from a package and the checkpoint/restore tests are being run everywhere.

Dockerfile
kolyshkin: I tried it and it did not work. Apparently this util is Ubuntu specific, and this Dockerfile is Debian based.

adrianreber: Right. These tests are running in a container. I expected this to be the travis environment but it is not. This means CRIU should also provide a binary for Debian. Because right now it is basically luck that the Ubuntu package runs on Debian.

mrunalp (approved):

Dockerfile
kolyshkin: > right now it is basically luck that the Ubuntu package runs on Debian

Well I assumed since Ubuntu releases are based on Debian releases, they are compatible.

I googled and apparently I was wrong, so yes, it works only by a coincidence.

Now, official golang docker images (https://hub.docker.com/_/golang) are either Debian or Alpine based. We can obviously switch to Ubuntu and install golang from the binaries, but I'm not 100% convinced it makes sense to do it only because of criu.

kolyshkin: @adrianreber OK I think having a Debian repo for CRIU would also benefit at least Moby (that currently also compiles CRIU from sources).

adrianreber: I am looking how to create and host Debian packages for CRIU. I will let you know. Right now I am looking at https://build.opensuse.org/ which could help.

thaJeztah: Are those repos x86 only, or also other arches?

adrianreber: > Are those repos x86 only, or also other arches?

The CRIU ppa provides packages for aarch64, ppc64le, s390x and x86_64.

Although Travis also provides all those arches CRIU only works on x86_64 as the other Travis arches are running in LXD containers which are too unprivileged for CRIU.

adrianreber: @kolyshkin please try the packages at https://download.opensuse.org/repositories/devel:/tools:/criu/Debian_10/

kolyshkin: @adrianreber thanks! Looks like opensuse.org is down ATM :(

kolyshkin: OK it's back, PR updated

Dockerfile
```diff
@@ -58,14 +50,13 @@ RUN cd /tmp \
...
-    && curl -fsSL https://github.com/checkpoint-restore/criu/archive/${CRIU_VERSION}.tar.gz | tar -C /usr/src/criu/ -xz --strip-components=1 \
-    && cd /usr/src/criu \
-    && echo 1 > .gitid \
-    && make -j $(nproc) install-criu \
-    && cd - \
-    && rm -rf /usr/src/criu
+RUN echo 'deb https://download.opensuse.org/repositories/devel:/tools:/criu/Debian_10/ /' > /etc/apt/sources.list.d/criu.list \
+    && wget -nv https://download.opensuse.org/repositories/devel:/tools:/criu/Debian_10/Release.key -O- | sudo apt-key add - \
+    && apt-get -q update \
+    && apt-get -q install -y --no-install-recommends criu \
```
AkihiroSuda, line 56: can we apt get `crui` and `skopeo` at once?

kolyshkin: Yeah, I was thinking about that. It will be faster to merge all the apt-related steps into one, iow add all repos and run `apt update && apt install`.

The biggest issue with the current Dockerfile is is downloads apt metadata 3 times and then removes it.

I took a look at moby/moby and in there they added a cache so the stages are still separate but the metadata is not downloaded multiple times. This needs a recent docker though, and afaik do not work with podman (it does not error out, but no caching is done either).

Given the fact that this Dockerfile is pretty simple, I think combining all the steps is the best solution.

kolyshkin: Updated.

AkihiroSuda (approved):

mrunalp (approved):

PR #2694, opened 2020-11-30
recttty: nits

kolyshkin:
Related to #2627.

#### 1. recvtty: use ioutil.Discard
    
    Saves us a few lines of code.

#### 2. recvtty: fix waiting for both goroutines
    
    It looks like we need to wait for the both copy goroutines to finish,
    not just the one that happens to finish first.

#### 3. recvtty: fix errcheck linter warnings
    
    Fixes the following errcheck linter warnings
    
    > contrib/cmd/recvtty/recvtty.go:115:10: Error return value of `io.Copy` is not checked (errcheck)
    >               io.Copy(os.Stdout, c)
    >                      ^
    > contrib/cmd/recvtty/recvtty.go:120:11: Error return value of `io.Copy` is not checked (errcheck)
    >                       io.Copy(c, os.Stdin)
    >                              ^
    > contrib/cmd/recvtty/recvtty.go:175:11: Error return value of `io.Copy` is not checked (errcheck)
    >                       io.Copy(devnull, master)
    >                              ^

PR #2695, opened 2020-11-30
Misc linter nits

kolyshkin:
This fixes some sensible linter warnings, and silence a few that do not make much sense. Please see individual commits for details.

With this other PRs (#2694, #2696, #2700), we are down to 3 warnings (plus a lot of `errcheck` ones which are to be addressed separately):

```console
$ golangci-lint run -D errcheck ./...
libcontainer/cgroups/ebpf/ebpf.go:35:12: SA1019: prog.Attach is deprecated: use link.RawAttachProgram instead.  (staticcheck)
	if err := prog.Attach(dirFD, ebpf.AttachCGroupDevice, unix.BPF_F_ALLOW_MULTI); err != nil {
	          ^
libcontainer/cgroups/ebpf/ebpf.go:39:13: SA1019: prog.Detach is deprecated: use link.RawDetachProgram instead.  (staticcheck)
		if err := prog.Detach(dirFD, ebpf.AttachCGroupDevice, unix.BPF_F_ALLOW_MULTI); err != nil {
		          ^
libcontainer/container_linux.go:32:2: SA1019: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (staticcheck)
	"github.com/golang/protobuf/proto"
	^
```

The protobuf one needs coordinated changes in go-criu (cc @adrianreber).

The ebfp one is related to https://github.com/cilium/ebpf/pull/108/commits/5727d650c09d79ac987b3c2d4e5420e6393cd5c7 and probably https://github.com/cilium/ebpf/pull/110

Related to #2627

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -278,7 +278,7 @@ func RemovePaths(paths map[string]string) (err error) {
 			}
 		}
 		if len(paths) == 0 {
-			//nolint:ineffassign // done to help garbage collecting: opencontainers/runc#2506
+			//nolint:ineffassign,staticcheck // done to help garbage collecting: opencontainers/runc#2506
 			paths = make(map[string]string)
 			return nil
 		}
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -763,11 +763,7 @@ func (c *linuxContainer) checkCriuVersion(minVersion int) error {
 const descriptorsFilename = "descriptors.json"
 
 func (c *linuxContainer) addCriuDumpMount(req *criurpc.CriuReq, m *configs.Mount) {
-	mountDest := m.Destination
-	if strings.HasPrefix(mountDest, c.config.Rootfs) {
-		mountDest = mountDest[len(c.config.Rootfs):]
-	}
-
+	mountDest := strings.TrimPrefix(m.Destination, c.config.Rootfs)
 	extMnt := &criurpc.ExtMountMap{
 		Key: proto.String(mountDest),
 		Val: proto.String(mountDest),
@@ -1145,11 +1141,7 @@ func (c *linuxContainer) Checkpoint(criuOpts *CriuOpts) error {
 }
 
 func (c *linuxContainer) addCriuRestoreMount(req *criurpc.CriuReq, m *configs.Mount) {
-	mountDest := m.Destination
-	if strings.HasPrefix(mountDest, c.config.Rootfs) {
-		mountDest = mountDest[len(c.config.Rootfs):]
-	}
-
+	mountDest := strings.TrimPrefix(m.Destination, c.config.Rootfs)
 	extMnt := &criurpc.ExtMountMap{
 		Key: proto.String(mountDest),
 		Val: proto.String(m.Source),
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -680,8 +680,7 @@ var criuFeatures *criurpc.CriuFeatures
 
 func (c *linuxContainer) checkCriuFeatures(criuOpts *CriuOpts, rpcOpts *criurpc.CriuOpts, criuFeat *criurpc.CriuFeatures) error {
 
-	var t criurpc.CriuReqType
-	t = criurpc.CriuReqType_FEATURE_CHECK
+	t := criurpc.CriuReqType_FEATURE_CHECK
 
 	// make sure the features we are looking for are really not from
 	// some previous check
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1810,10 +1810,6 @@ func (c *linuxContainer) saveState(s *State) (retErr error) {
 	return os.Rename(tmpFile.Name(), stateFilePath)
 }
 
-func (c *linuxContainer) deleteState() error {
-	return os.Remove(filepath.Join(c.root, stateFilename))
-}
-
 func (c *linuxContainer) currentStatus() (Status, error) {
 	if err := c.refreshState(); err != nil {
 		return -1, err
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -366,7 +366,7 @@ func TestProcessEmptyCaps(t *testing.T) {
 	// Wait for process
 	waitProcess(&pconfig, t)
 
-	outputStatus := string(stdout.Bytes())
+	outputStatus := stdout.String()
 
 	lines := strings.Split(outputStatus, "\n")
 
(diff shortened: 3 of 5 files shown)


libcontainer/user/lookup_windows.go
```diff
@@ -7,6 +7,28 @@ import (
...
 
+// userFromOS converts an os/user.(*User) to local User
+//
+// (This does not include Pass, Shell or Gecos)
+func userFromOS(u *user.User) (User, error) {
+	newUser := User{
+		Name: u.Username,
+		Home: u.HomeDir,
+	}
+	id, err := strconv.Atoi(u.Uid)
```
thaJeztah, line 18: I'm wondering if this even _works_ on Windows. Looking at the code; if user is "current" user, it is returned as-is; https://github.com/golang/go/blob/go1.15.5/src/os/user/lookup.go#L32-L37

In other cases, it would look-up the user through a syscall; https://github.com/golang/go/blob/go1.15.5/src/os/user/lookup_windows.go#L322-L331

And construct a user; https://github.com/golang/go/blob/go1.15.5/src/os/user/lookup_windows.go#L290-L297

However, both `Uid` and `Gid` on Windows are actually a SID (https://en.wikipedia.org/wiki/Security_Identifier), which would looks something like `S-1-5-21-3623811015-3361044348-30300820-1013`, or `S-1-5-21domain-513`

libcontainer/user/lookup_windows.go
```diff
@@ -23,6 +45,23 @@ func lookupUid(uid int) (User, error) {
...
 
+// groupFromOS converts an os/user.(*Group) to local Group
+//
+// (This does not include Pass or List)
+func groupFromOS(g *user.Group) (Group, error) {
+	newGroup := Group{
+		Name: g.Name,
+	}
+
+	id, err := strconv.Atoi(g.Gid)
```
thaJeztah, line 56: Same here

libcontainer/user/lookup_windows.go
thaJeztah: Looks like 

- in Moby, `libcontainer/user` is only used in unix files.
- In BuildKit it's used in a non-specific file https://github.com/moby/buildkit/blob/f26452a84763b9e13e7e157904cf28008cd58016/executor/oci/user.go
but BuildKit currently doesn't run on Windows; @TBBle may be able to tell if that file is actually used on Windows
- containerd also uses this package in a non-platform specific file https://github.com/containerd/containerd/blob/62083eeed275db8c154dd87a5f84444e97ddb507/oci/spec_opts.go#L539

thaJeztah (approved):
LGTM

left questions w.r.t. the Windows changes (change doesn't really affect current code, but perhaps we need a tracking issue for it and/or remove the windows code altogether?)

libcontainer/user/lookup_windows.go
TBBle: For BuildKit, `GetUser` (and that whole file you linked) is Linux-container specific, since it populates a [linux/solaris-specific fields](https://github.com/opencontainers/runtime-spec/blob/7413a7f753e1bd9a6a9c6dc7f96f55888cbbd476/specs-go/config.go#L88-L91).

Windows containers populate [a username field](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L96-L97) instead.

So that file would be used in LCOW mode, I guess, on Windows builds, so it can't be unix-specific. Right now BuildKit has a bunch of stuff we'd need for LCOW mode that's in platform-specific source files anyway, but I'm uninclined to make it that worse. Also, last time I tried, LCOW didn't support non-root UIDs anyway. But that was a long time ago, I was hoping WSL2 would fix that.

Looking slightly outwards, the containerd executor in BuildKit does not take care to only call these APIs with a Linux target platform, so attempting to use `USER` in a Dockerfile or the equivalent in LLB will fail for a Windows container, because we pass it through that `GetUser` code path, and a bunch of other unix-target-specific stuff. (Assuming there's not a higher-level rejection floating around). `USER 5:7` would pass the checks in BuildKit, but end up putting data in the linux/solaris specific fields I mentioned above, and _hopefully_ fail the container startup.

I hadn't noticed that, and apparently never tried to test that, so I'll add it to the BuildKit issue tracking WCOW.

Coming back to this code, based on the above, the whole lookup.go exported API only makes sense on unix host platforms, since (if I am skimming this correctly) it's for looking up users and groups on the _host_, I assume in order to resolve users/groups from outside so that the uid/gid can be passed into the spec.

So I'd suggest just dropping lookup.go and lookup_windows.go, and move those exported functions into lookup_unix.go, like other functions, e.g., `CurrentUser()`.

To address the actual linked file, I don't see any code in BuildKit that can call through to *these* APIs. The linked file is using the other part of the API (from user.go) which is about resolving users from inside the container. Its Unix container-specific (relies on Uid as `int`), but makes sense to execute on Windows for LCOW (as it's resolving from an /etc/passwd and /etc/group from inside the container's mounted filesystem).

I had a quick look at the containerd link you have, and similar to BuildKit, it's calling the user.go APIs that care about the container platform, not the host platform.

thaJeztah: > Coming back to this code, based on the above, the whole lookup.go exported API only makes sense on unix host platforms, since (if I am skimming this correctly) it's for looking up users and groups on the host, I assume in order to resolve users/groups from outside so that the uid/gid can be passed into the spec.
>
> So I'd suggest just dropping lookup.go and lookup_windows.go, and move those exported functions into lookup_unix.go, like other functions, e.g., CurrentUser().

Thanks for looking! I was thinking in that direction as well; better to explicitly not implement (or alternatively, add a "stub" implementation that produces an error when called) than to have something that gives the impression it's implemented.

TBBle: Yeah, I was thinking about an always-error stub, but since we don't seem to have any current callers in Moby, BuildKit, or containerd, I'm not sure it's worth trying to maintain the API.

But perhaps we do need stubs, _because_ we want to not regress the API, faulty as it may be. Anyone calling this on a Windows platform would have been unavoidably getting an error from `Atoi` anyway.

thaJeztah: There's been some discussions in this repository about libcontainer not being intended for external consumption (which is why I did some refactoring in other repositories to reduce it's usage); that said, `libcontainer/user` is used in quite some places. containerd is not at v1.0.0 yet, so breaking changes are (technically) still allowed, so from that perspective making them _now_ (before v1.0.0) and rip off the bandaid could be preferable (if libcontainer is always gonna be intended for _linux_ / _unix_ only, and not for Windows).

I'm open to both approaches though (heh, and not a maintainer on this repository 😅)

kolyshkin: So, do we gravitate towards removing this code? I am all for it! Removed (and a couple other windows things).

TBBle (approved):

AkihiroSuda (approved):

kolyshkin:
As per conversation in https://github.com/moby/moby/pull/41746#issuecomment-738148046, I separated out the windows part of this to #2700.

mrunalp (approved):

AkihiroSuda (approved):

PR #2696, opened 2020-11-30
libct/system.Stat: fix panic, speed up, add benchmarks

kolyshkin:
#### 1.    libct/system/proc_test: fix, improve, add benchmarks
    
1. Add a test case that tests parentheses in command.
    
2. Replace individual comparisons with `reflect.DeepEqual`.
   This also fixes wrong `%`-style types used in `Fatalf` statements.
    
3. Replace `Fatalf` with `Errorf` so we don't bail out on the first
   failure, and do not check result on error.
    
4. Add two benchmarks.

#### 2. libct/system.Stat: fix/improve/speedup

1. Remove PID field. It is totally useless -- we already know the PID.
    
2. Rewrite `parseStat()` to make it faster and more correct:
   - do not use `fmt.Scanf` as it is very slow;
   - avoid splitting data into 20+ fields, of which we only need 2 (should also help garbage collector);
   - make sure to not panic on short lines and other bad input;
   - add some bad input tests (some fail with old code);
   - use `LastIndexByte` instead of `LastIndex` where appropriate.
 
Benchmarks:

before (from the previous commit message):

```    
BenchmarkParseStat-4              116415             10804 ns/op
BenchmarkParseRealStat-4             240           4781769 ns/op
```
    
after:
```    
BenchmarkParseStat-4           1164948              1068 ns/op
BenchmarkParseRealStat-4           331           3458315 ns/op
```    

We are seeing 10x speedup in a synthetic benchmark, and about 1.4x
speedup in a real world benchmark. All this while being more scrupulous
about any possible errors. I also suspect much less allocations and thus
garbage to collect.
     
While at it, do not ignore any possible errors, and properly wrap those.

3. Add `I` and `P` process states (available since Linux 4.14).

## History
    
* v2: use pkg/errors more, remove t.Logf from test
* v3: rebased; drop pkg/errors; gofumpt'ed
* v4: rebased; improved description
* v5: rebased; mention bad input tests, added second benchmark results
* v6: remove PID field, do not use strings.Split, further speedup.
* v7: readability improvements, added new process states.
* v8: improved parser to bail earier on incorrect data, more doc, more tests.

### DIFF
--- a/libcontainer/system/proc.go
+++ b/libcontainer/system/proc.go
@@ -48,9 +48,6 @@ func (s State) String() string {
 // described in proc(5) with names based on the /proc/[pid]/status
 // fields.
 type Stat_t struct {
-	// PID is the process ID.
-	PID uint
-
 	// Name is the command run by the process.
 	Name string
 
@@ -72,32 +69,53 @@ func Stat(pid int) (stat Stat_t, err error) {
 }
 
 func parseStat(data string) (stat Stat_t, err error) {
-	// From proc(5), field 2 could contain space and is inside `(` and `)`.
-	// The following is an example:
+	// Example:
 	// 89653 (gunicorn: maste) S 89630 89653 89653 0 -1 4194560 29689 28896 0 3 146 32 76 19 20 0 1 0 2971844 52965376 3920 18446744073709551615 1 1 0 0 0 0 0 16781312 137447943 0 0 0 17 1 0 0 0 0 0 0 0 0 0 0 0 0 0
-	i := strings.LastIndex(data, ")")
-	if i <= 2 || i >= len(data)-1 {
-		return stat, fmt.Errorf("invalid stat data: %q", data)
+	// The fields are space-separated, see full description in proc(5).
+	//
+	// We are only interested in:
+	//  * field 2: process name. It is the only field enclosed into
+	//    parenthesis, as it can contain spaces (and parenthesis) inside.
+	//  * field 3: process state, a single character (%c)
+	//  * field 22: process start time, a long unsigned integer (%llu).
+
+	// 1. Look for the first '(' and the last ')' first, what's in between is Name.
+	//    We expect at least 20 fields and a space after the last one.
+
+	const minAfterName = 20*2 + 1 // the min field is '0 '.
+
+	first := strings.IndexByte(data, '(')
+	if first < 0 || first+minAfterName >= len(data) {
+		return stat, fmt.Errorf("invalid stat data (no comm or too short): %q", data)
 	}
 
-	parts := strings.SplitN(data[:i], "(", 2)
-	if len(parts) != 2 {
-		return stat, fmt.Errorf("invalid stat data: %q", data)
+	last := strings.LastIndexByte(data, ')')
+	if last <= first || last+minAfterName >= len(data) {
+		return stat, fmt.Errorf("invalid stat data (no comm or too short): %q", data)
 	}
 
-	stat.Name = parts[1]
-	_, err = fmt.Sscanf(parts[0], "%d", &stat.PID)
+	stat.Name = data[first+1 : last]
+
+	// 2. Remove fields 1 and 2 and a space after. State is right after.
+	data = data[last+2:]
+	stat.State = State(data[0])
+
+	// 3. StartTime is field 22, data is at field 3 now, so we need to skip 19 spaces.
+	skipSpaces := 22 - 3
+	for first = 0; skipSpaces > 0 && first < len(data); first++ {
+		if data[first] == ' ' {
+			skipSpaces--
+		}
+	}
+	// Now first points to StartTime; look for space right after.
+	i := strings.IndexByte(data[first:], ' ')
+	if i < 0 {
+		return stat, fmt.Errorf("invalid stat data (too short): %q", data)
+	}
+	stat.StartTime, err = strconv.ParseUint(data[first:first+i], 10, 64)
 	if err != nil {
-		return stat, err
+		return stat, fmt.Errorf("invalid stat data (bad start time): %w", err)
 	}
 
-	// parts indexes should be offset by 3 from the field number given
-	// proc(5), because parts is zero-indexed and we've removed fields
-	// one (PID) and two (Name) in the paren-split.
-	parts = strings.Split(data[i+2:], " ")
-	var state int
-	fmt.Sscanf(parts[3-3], "%c", &state) //nolint:staticcheck // "3-3" is more readable in this context.
-	stat.State = State(state)
-	fmt.Sscanf(parts[22-3], "%d", &stat.StartTime)
 	return stat, nil
 }
--- a/libcontainer/system/proc.go
+++ b/libcontainer/system/proc.go
@@ -19,6 +19,8 @@ const ( // Only values for Linux 3.14 and later are listed here
 	Stopped     State = 'T'
 	TracingStop State = 't'
 	Zombie      State = 'Z'
+	Parked      State = 'P'
+	Idle        State = 'I'
 )
 
 // String forms of the state from proc(5)'s documentation for
@@ -39,6 +41,10 @@ func (s State) String() string {
 		return "tracing stop"
 	case Zombie:
 		return "zombie"
+	case Parked:
+		return "parked"
+	case Idle:
+		return "idle" // kernel thread
 	default:
 		return fmt.Sprintf("unknown (%c)", s)
 	}
--- a/libcontainer/system/proc_test.go
+++ b/libcontainer/system/proc_test.go
@@ -1,45 +1,120 @@
 package system
 
-import "testing"
+import (
+	"errors"
+	"math/bits"
+	"os"
+	"reflect"
+	"strconv"
+	"testing"
+)
 
-func TestParseStartTime(t *testing.T) {
-	data := map[string]Stat_t{
-		"4902 (gunicorn: maste) S 4885 4902 4902 0 -1 4194560 29683 29929 61 83 78 16 96 17 20 0 1 0 9126532 52965376 1903 18446744073709551615 4194304 7461796 140733928751520 140733928698072 139816984959091 0 0 16781312 137447943 1 0 0 17 3 0 0 9 0 0 9559488 10071156 33050624 140733928758775 140733928758945 140733928758945 140733928759264 0": {
-			PID:       4902,
-			Name:      "gunicorn: maste",
-			State:     'S',
-			StartTime: 9126532,
-		},
-		"9534 (cat) R 9323 9534 9323 34828 9534 4194304 95 0 0 0 0 0 0 0 20 0 1 0 9214966 7626752 168 18446744073709551615 4194304 4240332 140732237651568 140732237650920 140570710391216 0 0 0 0 0 0 0 17 1 0 0 0 0 0 6340112 6341364 21553152 140732237653865 140732237653885 140732237653885 140732237656047 0": {
-			PID:       9534,
-			Name:      "cat",
-			State:     'R',
-			StartTime: 9214966,
-		},
+var procdata = map[string]Stat_t{
+	"4902 (gunicorn: maste) S 4885 4902 4902 0 -1 4194560 29683 29929 61 83 78 16 96 17 20 0 1 0 9126532 52965376 1903 18446744073709551615 4194304 7461796 140733928751520 140733928698072 139816984959091 0 0 16781312 137447943 1 0 0 17 3 0 0 9 0 0 9559488 10071156 33050624 140733928758775 140733928758945 140733928758945 140733928759264 0": {
+		PID:       4902,
+		Name:      "gunicorn: maste",
+		State:     'S',
+		StartTime: 9126532,
+	},
+	"9534 (cat) R 9323 9534 9323 34828 9534 4194304 95 0 0 0 0 0 0 0 20 0 1 0 9214966 7626752 168 18446744073709551615 4194304 4240332 140732237651568 140732237650920 140570710391216 0 0 0 0 0 0 0 17 1 0 0 0 0 0 6340112 6341364 21553152 140732237653865 140732237653885 140732237653885 140732237656047 0": {
+		PID:       9534,
+		Name:      "cat",
+		State:     'R',
+		StartTime: 9214966,
+	},
+	"12345 ((ugly )pr()cess() R 9323 9534 9323 34828 9534 4194304 95 0 0 0 0 0 0 0 20 0 1 0 9214966 7626752 168 18446744073709551615 4194304 4240332 140732237651568 140732237650920 140570710391216 0 0 0 0 0 0 0 17 1 0 0 0 0 0 6340112 6341364 21553152 140732237653865 140732237653885 140732237653885 140732237656047 0": {
+		PID:       12345,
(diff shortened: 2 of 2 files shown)


libcontainer/system/proc.go
```diff
@@ -75,29 +78,40 @@ func parseStat(data string) (stat Stat_t, err error) {
...
+	}
+	stat.PID = uint(pid)
+
+	// Remove the first two fields and a space after.
+	data = data[i+2:]
+	stat.State = State(data[0])
+
+	// StartTime is field 22, data is at field 3 now, so we need to skip 19 spaces.
+	skipSpaces := 22 - 3
+	for i = 0; skipSpaces > 0; i++ {
```
thaJeztah, line 104: I realize this is probably theoretical; should there be a safeguard in case no spaces are found and `i` becomes `> len(data)` ?

kolyshkin:
Rebased to benefit from #2690 (while Travis was not able to test this PR for 6+ hours :cry:)

thaJeztah (approved):
LGTM

libcontainer/system/proc.go
kolyshkin: You're quite right, thanks for catching this. Fixed, and added a test case checking all kinds of "bad input".

NOTE I am not striving at being absolutely perfect in input validation, but we should not panic in any case.

libcontainer/system/proc.go
```diff
@@ -75,29 +78,42 @@ func parseStat(data string) (stat Stat_t, err error) {
 	// From proc(5), field 2 could contain space and is inside `(` and `)`.
 	// The following is an example:
 	// 89653 (gunicorn: maste) S 89630 89653 89653 0 -1 4194560 29689 28896 0 3 146 32 76 19 20 0 1 0 2971844 52965376 3920 18446744073709551615 1 1 0 0 0 0 0 16781312 137447943 0 0 0 17 1 0 0 0 0 0 0 0 0 0 0 0 0 0
-	i := strings.LastIndex(data, ")")
+	i := strings.LastIndexByte(data, ')')
```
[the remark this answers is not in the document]
kolyshkin: > wondering if passing data to parseStat() as []byte (not converting it to a string) could optimise even more

I was thinking about that, too, but decided to not go down that road (thought it will decrease readability and result in a few smaller allocations instead of a single bigger one, which is not necessarily makes things better).

kolyshkin:
@AkihiroSuda @mrunalp PTAL

kolyshkin:
rebased (for new CI)

libcontainer/system/proc_test.go
```diff
@@ -1,45 +1,95 @@
...
 		st, err := parseStat(line)
 		if err != nil {
-			t.Fatal(err)
-		}
-		if st.PID != expected.PID {
-			t.Fatalf("expected PID %q but received %q", expected.PID, st.PID)
-		}
-		if st.State != expected.State {
-			t.Fatalf("expected state %q but received %q", expected.State, st.State)
+			t.Errorf("input %q, unexpected error %v", line, err)
```
thaJeztah, line 39: No need to change; could also use sub-tests for these cases (if each case has a useful name/description)

libcontainer/system/proc_test.go
```diff
@@ -1,45 +1,95 @@
...
+			"123 (cmd) S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 -1 "},
+		{"bad stime 2", // would be valid if not -1
+			"123 (cmd) S                   -1 "},
+	}
+	for _, c := range cases {
+		st, err := parseStat(c.input)
+		t.Logf("case %q: error: %v, stat: %v", c.desc, err, st)
+		if err == nil {
+			t.Errorf("case %q, expected error, got nil", c.desc)
 		}
```
thaJeztah, line 72: Given that each test has a description, could use sub-tests here;

```suggestion
	for _, c := range cases {
		c := c
		t.Run(c.desc, func(t *testing.T) {
			st, err := parseStat(c.input)
			t.Logf("error: %v, stat: %v", err, st)
			if err == nil {
				t.Fatalf("expected error, got nil", c.desc)
			}
		})
```

libcontainer/system/proc.go
```diff
@@ -75,29 +78,42 @@ func parseStat(data string) (stat Stat_t, err error) {
...
 	if i <= 2 || i >= len(data)-1 {
-		return stat, fmt.Errorf("invalid stat data: %q", data)
+		return stat, fmt.Errorf("invalid stat data (no comm): %q", data)
```
thaJeztah, line 83: While at it, perhaps;

```suggestion
		return stat, errors.Errorf("invalid stat data (no comm): %q", data)
```

libcontainer/system/proc.go
```diff
@@ -75,29 +78,42 @@ func parseStat(data string) (stat Stat_t, err error) {
...
 	}
 
-	parts := strings.SplitN(data[:i], "(", 2)
+	parts := strings.SplitN(data[:i], " (", 2)
 	if len(parts) != 2 {
-		return stat, fmt.Errorf("invalid stat data: %q", data)
+		return stat, fmt.Errorf("invalid stat data (no comm): %q", data)
```
thaJeztah, line 88: Same here

```suggestion
		return stat, errors.Errorf("invalid stat data (no comm): %q", data)
```

libcontainer/system/proc.go
```diff
@@ -75,29 +78,42 @@ func parseStat(data string) (stat Stat_t, err error) {
...
+	// StartTime is field 22, data is at field 3 now, so we need to skip 19 spaces.
+	skipSpaces := 22 - 3
+	for i = 0; skipSpaces > 0 && i < len(data); i++ {
+		if data[i] == ' ' {
+			skipSpaces--
+		}
+	}
+	j := strings.IndexByte(data[i:], ' ')
+	if j <= 0 {
+		return stat, fmt.Errorf("invalid stat data (too short): %q", data)
```
thaJeztah, line 111:
```suggestion
		return stat, errors.Errorf("invalid stat data (too short): %q", data)
```

thaJeztah (dismissed):
LGTM, but left some thoughts/suggestions for consideration

libcontainer/system/proc.go
kolyshkin: So the differences are
 - errors package is not standard
 - it encapsulates the calling stack (i.e. backtrace) into the error (which might be very handy, but I've never seen it used anywhere)
 - anything else that I miss?
 
As this code does not have many callers, I think the backtrace does not make much sense, does it?

libcontainer/system/proc_test.go
kolyshkin: I'm not sure it makes much sense here, for such trivial checks... the only differences I can think of are more lines of test output, and (apparently) more code (which IMHO decreases the readability). Or am I missing something?

libcontainer/system/proc_test.go
```diff
@@ -43,6 +43,36 @@ func TestParseStartTime(t *testing.T) {
...
+		{"short line (no space after stime)",
+			"123 (cmd) S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 42"},
+		{"bad stime",
+			"123 (cmd) S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 -1 "},
+		{"bad stime 2", // would be valid if not -1
+			"123 (cmd) S                   -1 "},
+	}
+	for _, c := range cases {
+		st, err := parseStat(c.input)
+		t.Logf("case %q: error: %v, stat: %v", c.desc, err, st)
```
kolyshkin, line 69: This t.Logf needs to be removed. We have enough output in our CI logs already :)

libcontainer/system/proc.go
cyphar: Backtrace is used for `--debug` I think? If it isn't, it should be (that's what I use in umoci).

libcontainer/system/proc_test.go
cyphar: I personally prefer `t.Run` because it makes it possible to run a single subtest and it better represents the idea of having the same test code for many cases. But yeah this test is simple enough that `t.Run` probably adds too little to be worth it.

cyphar:
Is this a bugfix or an improvement? If it's an improvement it belongs in post-1.0 as a milestone.

libcontainer/system/proc.go
kolyshkin: We did not do that, but it's trivial to fix: https://github.com/opencontainers/runc/pull/2789

kolyshkin:
> Is this a bugfix or an improvement? If it's an improvement it belongs in post-1.0 as a milestone.

Improvement (except eliminating a linter warning which might be classified as bugfix), and that's lots of code, so moving to post 1.0

libcontainer/system/proc.go
kolyshkin: fixed to use pkg/errors

libcontainer/system/proc_test.go
kolyshkin: fixed

cyphar:
Yeah agreed, this change looks a little scary tbh.

kolyshkin:
close/open to re-kick ci

thaJeztah:
linter is failing;

```
 Error: File is not `gofumpt`-ed (gofumpt)
```

kolyshkin:
v3: rebased; drop using pkg/errors; gofumpt'ed

kolyshkin:
CI failure is unrelated (https://github.com/opencontainers/runc/issues/3050)

kolyshkin:
Rebased to fix CI.

libcontainer/system/proc_test.go
```diff
@@ -1,45 +1,160 @@
...
 		},
-
-		"24767 (irq/44-mei_me) S 2 0 0 0 -1 2129984 0 0 0 0 0 0 0 0 -51 0 1 0 8722075 0 0 18446744073709551615 0 0 0 0 0 0 0 2147483647 0 0 0 0 17 1 50 1 0 0 0 0 0 0 0 0 0 0 0": {
-			PID:       24767,
-			Name:      "irq/44-mei_me",
-			State:     'S',
-			StartTime: 8722075,
+		{
+			"negative pid",
+			"-1 (cmd) S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0",
```
thaJeztah, line 64: Should there be a test-case for `0` as pid ?

kolyshkin: In fact, the more I look at it, the more I think we should drop the PID field entirely. We call `Stat(pid)` and we receive pid :)

Let me drop it; it's entirely useless. One use of it would be to do a sanity check (that the pid is the same as we expect it to be) but we're not doing that either and I don't think we should (not trust the kernel _that_ much).

kolyshkin: done!

Also, no one is using Name field either, but (unlike PID) it _might_ be useful, and the benchmarks after removing it do not show drastic difference, so I'm keeping it.

kolyshkin:
Some quite interesting observations:
* removing `strings.Split` (which result in ~20 fields slice) and a couple of `fmt.Scanf`s results in 7x improvement!
* removing `strings.Split` (for much shorter data, only the first two fields) results in further 2x improvement!

kolyshkin:
@thaJeztah PTAL

thaJeztah (commented):
overall looks good; I was looking for corner-cases, and possibly found some (may not be important to add, but thought I'd post them)

I'm fine with merging as-is though

libcontainer/system/proc_test.go
```diff
@@ -52,6 +48,55 @@ func TestParseStat(t *testing.T) {
...
+			"123 ( S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0",
+		},
+		{
+			"misplaced ()",
+			"123 )one( S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0",
+		},
+		{
+			"misplaced empty ()",
+			"123 )( S 2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0",
+		},
```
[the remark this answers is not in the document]
kolyshkin: I think that cases 1 (aka "( at start") and 3 (aka ") at start") will start failing if we return PID parsing that was removed in the last iteration. So, in fact, this is "invalid PID" or "garbage in field 1" cases, but we don't care about PID.

Another way this can be fixed is looking for ` (` instead of `(`, i.e. add a space, but that will only slow things down.

Case 2 (aka "( at end") seems legit from the current parser's POV as it only cares about the field within `()` (aka Name), the field right after (aka State), and the Start Time (22nd field).

Case 4 is seen as error as there are not enough fields after `)`.

Now, there are more situations in which the current parser won't detect an error. For example, if any field from 4 to 52 is not a number, or empty, or (for some of them) negative, etc. -- we don't detect any of that.

We do not want to validate the input here, as we trust the kernel. We only want to
 - extract the fields we're interested in;
 - be sure to not panic in any circumstances.
 
 If you can think of any cases where the current code would panic, that'd be great.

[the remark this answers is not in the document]
kolyshkin: aaaand I just did:

```
                {
                        "a tad short",
                        "1234 (cmd) ",
                },
```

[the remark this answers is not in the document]
kolyshkin: Improved the parser to bail early on an incorrect input, and documented it a bit more.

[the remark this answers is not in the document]
kolyshkin: PTAL @thaJeztah

thaJeztah (approved):
LGTM

libcontainer/system/proc.go
```diff
@@ -72,32 +75,53 @@ func Stat(pid int) (stat Stat_t, err error) {
...
+	// The fields are space-separated, see full description in proc(5).
+	//
+	// We are only interested in:
+	//  * field 2: process name. It is the only field enclosed into
+	//    parenthesis, as it can contain spaces (and parenthesis) inside.
+	//  * field 3: process state, a single character (%c)
+	//  * field 22: process start time, a long unsigned integer (%llu).
+
+	// 1. Look for the first '(' and the last ')' first, what's in between is Name.
+	//    We expect at least 20 fields and a space after the last one.
```
cyphar, line 89: Here's hoping that they don't add a new field in a few kernel versions that contains `(` *grimace*. Why on earth doesn't `/proc/self/status` contain StartTime? If it did we could just use that rather than this unparseable crap. `/rant`

cyphar (approved):
LGTM.

libcontainer/system/proc.go
thaJeztah: I heard they will switch to YAML, because it seems to be popular 🤪

PR #2698, opened 2020-12-02
nsenter: hard-code memfd_create(2) syscall numbers

cyphar:
By all rights this patch shouldn't be necessary, but we've had cases where a customer has had other weird issues (`-EINVAL` from the read-only mount setup as well as all of the writeable partitions set to `noexec`) which ends up hitting errors that could be avoided if `SYS_memfd_create` was defined.

But ultimately this is fixing a SLE problem, so I'd be fine with carrying it downstream if it looks ugly (which it does).

<hr>

Some libc versions still in use by distributions (such as SLE) do not
define SYS_memfd_create even though the kernel supports the feature.
Since the syscall numbers are fixed, we can just hard-code them if
__NR_memfd_create is not defined.

We only do this for a handful of architectures, since containers aren't
widely supported on every possible Linux architecture.

### DIFF
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -59,14 +59,38 @@
 #include <sys/syscall.h>
 
 /* Use our own wrapper for memfd_create. */
-#if !defined(SYS_memfd_create) && defined(__NR_memfd_create)
-#  define SYS_memfd_create __NR_memfd_create
+#ifndef SYS_memfd_create
+#  ifdef __NR_memfd_create
+#    define SYS_memfd_create __NR_memfd_create
+#  else
+/* These values come from <https://fedora.juszkiewicz.com.pl/syscalls.html>. */
+#    warning "libc is outdated -- using hard-coded SYS_memfd_create"
+#    if defined(__x86_64__) // x86_64
+#      define SYS_memfd_create 319
+#    elif defined(__i386__) // i386
+#      define SYS_memfd_create 356
+#    elif defined(__ia64__) // ia64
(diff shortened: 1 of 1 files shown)


libcontainer/nsenter/cloned_binary.c
```diff
@@ -59,14 +59,38 @@
...
-#  define SYS_memfd_create __NR_memfd_create
+#ifndef SYS_memfd_create
+#  ifdef __NR_memfd_create
+#    define SYS_memfd_create __NR_memfd_create
+#  else
+/* These values come from <https://fedora.juszkiewicz.com.pl/syscalls.html>. */
+#    warning "libc is outdated -- using hard-coded SYS_memfd_create"
+#    if defined(__x86_64__) // x86_64
+#      define SYS_memfd_create 319
+#    elif defined(__i386__) // x32
```
AkihiroSuda, line 70: i386 isn't x32
https://en.wikipedia.org/wiki/X32_ABI

AkihiroSuda (changes requested):
i386 isn't x32

libcontainer/nsenter/cloned_binary.c
cyphar: Ah, right. The `#ifdef` is right but the comment is wrong.

AkihiroSuda (approved):
LGTM, but I'm not sure runc really works on ia64

kolyshkin (approved):
LGTM, and yeah, this is ugly but understandably necessary.

Ideally, I'd move all this to a separate `.h` file so devs won't have to see this upon editing `cloned_binary.c` but I'm fine with the way it is.

PR #2706, opened 2020-12-10
libct/configs/validate.sysctl: fix host netns check

### LINKED ISSUE #2693: Use the --network=host option using net.* sysctls are allowed when docker run
This is described in the official Docker documentation：
If you use the --network=host option using these sysctls that beginning with net.* are not allowed.

But I succeeded when I ran the following command
```
$ docker run -tid --name=test_systcl --network=host --sysctl=net.ipv4.ip_forward=1 --rm busybox
e32a79053e1172135876f65fc9e635ca3bed0d3000cf76327ecea119cebc0483
$ docker ps | grep test_systcl
e32a79053e11    busybox    "sh"    8 seconds ago    Up 7 seconds    test_systcl
```

My Docker verison
```
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        99c5edce
 Built:             Thu Jan  9 08:27:14 2020
 Experimental:      false

 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       99c5edce
  Built:            Thu Jan  9 08:26:30 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```
I found that the code(libcontainer/configs/validate/validator.go) for runc does have a corresponding validation.
```golang
// sysctl validates that the specified sysctl keys are valid or not.
// /proc/sys isn't completely namespaced and depending on which namespaces
// are specified, a subset of sysctls are permitted.
func (v *ConfigValidator) sysctl(config *configs.Config) error {
	validSysctlMap := map[string]bool{
		"kernel.msgmax":          true,
		"kernel.msgmnb":          true,
		"kernel.msgmni":          true,
		"kernel.sem":             true,
		"kernel.shmall":          true,
		"kernel.shmmax":          true,
		"kernel.shmmni":          true,
		"kernel.shm_rmid_forced": true,
	}

	for s := range config.Sysctl {
		if validSysctlMap[s] || strings.HasPrefix(s, "fs.mqueue.") {
			if config.Namespaces.Contains(configs.NEWIPC) {
				continue
			} else {
				return fmt.Errorf("sysctl %q is not allowed in the hosts ipc namespace", s)
			}
		}
		if strings.HasPrefix(s, "net.") {
			if config.Namespaces.Contains(configs.NEWNET) {
				if path := config.Namespaces.PathOf(configs.NEWNET); path != "" {
					if err := checkHostNs(s, path); err != nil {
						return err
					}
				}
				continue
			} else {
				return fmt.Errorf("sysctl %q is not allowed in the hosts network namespace", s)
			}
		}
		if config.Namespaces.Contains(configs.NEWUTS) {
			switch s {
			case "kernel.domainname":
				// This is namespaced and there's no explicit OCI field for it.
				continue
			case "kernel.hostname":
				// This is namespaced but there's a conflicting (dedicated) OCI field for it.
				return fmt.Errorf("sysctl %q is not allowed as it conflicts with the OCI %q field", s, "hostname")
			}
		}
		return fmt.Errorf("sysctl %q is not in a separate kernel namespace", s)
	}

	return nil
}
```
When I found that call checkHostNs (s, path), the path is _/var/run/docker/netns/default, the path is not a  symbolic link_, so can't judge whether the network namespace of the container is the same as the host

kolyshkin:
In case nsfs mount (such as /run/docker/netns/xxxx) is provided as
the netns path, the current way of determining whether path is of
host netns or not is not working.
    
The proper way to check is to do stat(2) and compare dev_t and
inode fields, which is what this commit does.

A test case is added and I have checked that it fails before and passes
after the fix.

While at it, optimize repeated check in case more than one net.*
sysctl is given and there is no error.

Fixes #2693

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
@@ -145,6 +146,12 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 		"kernel.shm_rmid_forced": true,
 	}
 
+	var (
+		netOnce    sync.Once
+		hostnet    bool
+		hostnetErr error
+	)
+
 	for s := range config.Sysctl {
 		if validSysctlMap[s] || strings.HasPrefix(s, "fs.mqueue.") {
 			if config.Namespaces.Contains(configs.NEWIPC) {
@@ -154,16 +161,27 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 			}
 		}
 		if strings.HasPrefix(s, "net.") {
-			if config.Namespaces.Contains(configs.NEWNET) {
-				if path := config.Namespaces.PathOf(configs.NEWNET); path != "" {
-					if err := checkHostNs(s, path); err != nil {
-						return err
-					}
+			// Is container using host netns?
+			// Here "host" means "current", not "initial".
+			netOnce.Do(func() {
+				if !config.Namespaces.Contains(configs.NEWNET) {
+					hostnet = true
+					return
 				}
-				continue
-			} else {
-				return fmt.Errorf("sysctl %q is not allowed in the hosts network namespace", s)
+				path := config.Namespaces.PathOf(configs.NEWNET)
+				if path == "" {
+					// own netns, so hostnet = false
+					return
+				}
+				hostnet, hostnetErr = isHostNetNS(path)
+			})
+			if hostnetErr != nil {
+				return hostnetErr
 			}
+			if hostnet {
+				return fmt.Errorf("sysctl %q not allowed in host network namespace", s)
+			}
+			continue
 		}
 		if config.Namespaces.Contains(configs.NEWUTS) {
 			switch s {
@@ -218,16 +236,4 @@ func isHostNetNS(path string) (bool, error) {
 	}
 
 	return (st1.Dev == st2.Dev) && (st1.Ino == st2.Ino), nil
-}
-
-// checkHostNs checks whether network sysctl is used in host namespace.
-func checkHostNs(sysctlConfig string, path string) error {
-	host, err := isHostNetNS(path)
-	if err != nil {
-		return err
-	}
-	if host {
-		return fmt.Errorf("sysctl %q is not allowed in the hosts network namespace", sysctlConfig)
-	}
-	return nil
 }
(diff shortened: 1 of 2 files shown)


kolyshkin:
We run unit tests as root on GHA, so this was tested by ci / test jobs:
```
=== RUN   TestValidateSysctlWithBindHostNetNS
--- PASS: TestValidateSysctlWithBindHostNetNS (0.00s)
```

kolyshkin:
@AkihiroSuda @mrunalp @crosbymichael PTAL

libcontainer/configs/validate/validator.go
```diff
@@ -154,16 +161,26 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
...
+				if !config.Namespaces.Contains(configs.NEWNET) {
+					hostnet = true
+					return
 				}
-				continue
-			} else {
-				return fmt.Errorf("sysctl %q is not allowed in the hosts network namespace", s)
+				path := config.Namespaces.PathOf(configs.NEWNET)
+				if path == "" {
+					// hostnet = false
```
mrunalp, line 172: Shouldn't hostnet be true if path is empty?

mrunalp: Never mind. This is fine :+1:

kolyshkin: yeah, this is the case when configs.NEWNET is set but the path is not (which means "make a new netns").

libcontainer/configs/validate/validator.go
```diff
@@ -204,43 +222,18 @@ func (v *ConfigValidator) intelrdt(config *configs.Config) error {
 	return nil
 }
 
-func isSymbolicLink(path string) (bool, error) {
-	fi, err := os.Lstat(path)
-	if err != nil {
-		return false, err
-	}
+func isHostNetNS(path string) (bool, error) {
+	// assuming the current process is in initial netns
```
AkihiroSuda, line 226: s/initial/parent/

kolyshkin: I guess this depends on the definition of "host". I think of "host" as "the one that has initial namespaces".

Surely the current process is in the parent namespace, but is it in the host namespace? May be, may be not. What I meant by this is this function can give false positive (provided by definition of "host") in case runc is not in the host namespace.

Anyway, I think I was wrong, and in here by "host" we mean "current" or "parent to a container being created". In this case, it's better to just remove the comment to avoid confusion, or tell that here by "host" we mean "parent of a container being created".

libcontainer/configs/validate/validator.go
kolyshkin: I have added a tiny clarification to that comment

libcontainer/configs/validate/validator.go
kolyshkin: removed this comment, added another one elsewhere saying by "host" we mean "current" not "initial".

AkihiroSuda (approved):

mrunalp (approved):

PR #2709, opened 2020-12-15
remove "apparmor" and "selinux" build tags

### LINKED ISSUE #2704: Proposal: remove "apparmor" build tag
The `apparmor` tag was introduced in a01ed80babebf1db6f303a64c871c04914456b18  (2014)  to make cgo dependency on libapparmor optional.

However, the cgo dependency was removed in db093f621f4e6d40feafbe6d7b89a9bee6f6ff85 (2017), so it doesn't seem meaningful to keep `apparmor` build tag. I suggest removing the build tag.

AkihiroSuda:
## AppArmor
The "apparmor" tag was introduced in a01ed80 (2014) to make cgo dependency on libapparmor optional.

However, the cgo dependency was removed in db093f6 (2017), so it is no longer meaningful to keep apparmor build tag.

Close #2704 

## SELinux

The build tag was removed in go-selinux v1.8.0: https://github.com/opencontainers/selinux/pull/132


libcontainer/apparmor/apparmor_disabled.go
```diff
@@ -1,4 +1,4 @@
-// +build !apparmor !linux
+// +build !linux
```
thaJeztah, line 1: Perhaps this file should be renamed to `_unsupported.go` ?

thaJeztah: Somewhat orthogonal, but wondering if we should have a single implementation of these;

- rewrite `IsEnabled()` to use a `sync.Once` (could use `runtime.GOOS` to skip the actual check)
- rewrite `ApplyProfile()` to use `IsEnabled()`

Note that behavior is currently somewhat inconsistent;

- on non-linux, we produce an error if  `ApplyProfile()` is called (because `ErrApparmorNotEnabled`)
- on linux, `ApplyProfile()` does _NOT_ check if AppArmor is enabled, and always proceeds with `setProcAttr()` (not sure if that produces an error in that case, or is silently ignored)?

thaJeztah (changes requested):
Left some thoughts; @AkihiroSuda @kolyshkin wdyt?

README.md
```diff
@@ -63,16 +63,18 @@ To change build tags from the default, set the `BUILDTAGS` variable for make,
...
 | Build Tag | Feature                            | Enabled by default | Dependency |
 |-----------|------------------------------------|--------------------|------------|
 | seccomp   | Syscall filtering                  | yes                | libseccomp |
 | selinux   | selinux process and mount labeling | yes                | <none>     |
-| apparmor  | apparmor profile support           | yes                | <none>     |
 | nokmem    | disable kernel memory accounting   | no                 | <none>     |
 
+The following tags were used in the past versions, but no longer meaningful in the
+latest version:
+- apparmor
```
kolyshkin, line 76: I would be more specific here, as "latest version" is vague.

Something like 

```suggestion
The following build tags were used earlier, but are now obsoleted:
 - **apparmor** (since runc v1.0.0-rc93 the feature is always enabled)
```

(or something like that -- the main idea is to mention a specific version)

script/release.sh
```diff
@@ -27,7 +27,7 @@ function build_project() {
 	builddir="$(dirname "$1")"
 
 	# Build with all tags enabled.
-	make -C "$root" COMMIT_NO= BUILDTAGS="seccomp selinux apparmor" static
+	make -C "$root" COMMIT_NO= BUILDTAGS="seccomp selinux" static
```
kolyshkin: We can just drop `BUILDTAGS` entirely from here, because since #2254 they are already set by the Makefile.

libcontainer/apparmor/apparmor_disabled.go
kolyshkin: Agree WRT renaming to `_unsupported`, the rest of it can be addressed as a follow-up.

AkihiroSuda: changed to _unsupported. I agree that the rest can be a follow-up.

README.md
AkihiroSuda: fixed

script/release.sh
AkihiroSuda: fixed

kolyshkin (approved):
LGTM

mrunalp (approved):

PR #2712, opened 2020-12-18
libct/init_linux: reorder chdir to fix EPERM

kolyshkin:
_This is a carry of https://github.com/opencontainers/runc/pull/2685, adding test cases for it, as well for https://github.com/opencontainers/runc/pull/2086 to make sure there is no regression. Most work was done by @haircommander, I just tinkered with the tests. Original description from Peter follows._

----

commit 5e0e67d moved the chdir to be one of the
first steps of finalizing the namespace of the container.

However, this causes issues when the cwd is not accessible by the user running runc, but rather
as the container user.

Thus, setupUser has to happen before we call chdir. setupUser still happens before setting the caps,
so the user should be privileged enough to mitigate the issues fixed in 5e0e67d (I've tested it on my machine, so I believe it does not regress)

----

Closes: #2685

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -129,12 +129,6 @@ func finalizeNamespace(config *initConfig) error {
 		return errors.Wrap(err, "close exec fds")
 	}
 
-	if config.Cwd != "" {
-		if err := unix.Chdir(config.Cwd); err != nil {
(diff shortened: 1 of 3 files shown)


kolyshkin:
@AkihiroSuda @mrunalp @cyphar @crosbymichael PTAL

haircommander:
LGTM, thanks for picking this up :+1:

AkihiroSuda (approved):

mrunalp (approved):

PR #2717, opened 2021-01-05
libct/checkProcMounts: optimize

kolyshkin:
Commit 9c1242ecb ("Add white list for bind mount chec", Jan 6 2016)
added a set of entries under /proc which we allow to be mounted to,
for the benefit of lxcfs-like fuse-backed hack to have container's
own version of /proc/meminfo etc.

For some reason, the allow list check is performed at the very
beginning of the function, which is not optimal.

Move the check to the end -- at this point in the code we already
know we're under /proc, so it make sense to consult the allow list.

This makes the code slightly more logical and hopefully slightly faster.


kolyshkin:
rebased (for new CI)

kolyshkin:
CI failure is a flake (https://github.com/opencontainers/runc/issues/2425), restarted.

kolyshkin:
This one is easy to review in fact -- it just moves a block of code around.

mrunalp (approved):

AkihiroSuda (approved):

PR #2687, opened 2020-11-20
More GitHub actions

kolyshkin:
This moves a few checks from Travis to github actions.

PR #2618, opened 2020-09-30
CI: add golangci-lint via github actions

kolyshkin:
It is set to only check the new code, so this should help maintainers to see any new issues introduced by PRs.

To see the example how it works, check https://github.com/kolyshkin/runc/pull/1

Fixes: #2617

PR #2721, opened 2021-01-05
ci: replace travis with GHA for Fedora 33 and CentOS 7

kolyshkin:
This PR moves away from travis-ci.org entirely, replacing it with github actions based jobs.

This, together with some older PRs (#2618, #2687, #2690 etc.) completes the task of revamping the CI.

Brief description (see commit messages for more info):

### 1. tests/events.bats: unify duplicated code

This also fixes the "events --interval 100ms" test which was previously stalling on GHA.

### 2. ci: move commit length check from travis to gha

Retiring vbatts/git-validation.

### 3. ci: move cross compile check from travis to gha

This is now the only ci job that uses Docker (as Ubuntu
does not support all the arches we want).

### 4. ci: move misc validate tasks from travis to gha

These "misc" ones are actually `make validate` and `make release`
that somehow do not belong elsewhere.

### 5. ci: move Fedora 33 and CentOS 7 tests to gha
    
...and finally remove .travis.yml as GHA now covers everything it used to
(unless I overlooked something).
    
Inspired by https://github.com/containerd/containerd/pull/4279


kolyshkin:
Almost ready. TODO: check that commit length check works, otherwise looks ready for the prime time.

kolyshkin:
`validate / commit` fails if the subject is long.

re-checking when such commit is not the last one in the series.

kolyshkin:
> re-checking when such commit is not the last one in the series.

works either way (i.e. all commits are being checked).

This one is ready. @AkihiroSuda @mrunalp PTAL

.github/workflows/test.yml
```diff
@@ -80,3 +80,72 @@ jobs:
...
+      - name: system info
+        run: ssh default 'sh -exc "uname -a && systemctl --version && df -T"'
+
+      - name: unit tests
+        run: ssh default 'cd /vagrant && sudo make localunittest'
+
+      # The integration tests require tty which GH actions lack;
+      # wrap those in "script" to emulate tty.
+      - name: cgroupv2 with systemd
+        run: ssh default "script -e -c /bin/bash -c 'stty rows 40 cols 80; TERM=xterm sudo make -C /vagrant localintegration RUNC_USE_SYSTEMD=yes'"
```
AkihiroSuda, line 109: maybe `script` and `ssty` stuff should be moved to Makefile, but can be worked out later

AkihiroSuda (approved):

.github/workflows/test.yml
kolyshkin: Yes, this looks kinda ugly. I thought about a shell script or a program that checks if tty is available, and re-executes itself through script (and does other needed things) if not, and then execute whatever is passed on in arguments.

GHA is the only environment without a tty, so it does not make sense to have this in Makefile (at least unconditionally).

Let's do it later.

tests/integration/events.bats
```diff
@@ -27,25 +27,30 @@ function teardown() {
...
+
 	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
 	[ "$status" -eq 0 ]
 
-	# spawn two sub processes (shells)
-	# the first sub process is an event logger that sends stats events to events.log
-	# the second sub process waits for an event that includes test_busybox then
-	# kills the test_busybox container which causes the event logger to exit
-	(__runc events test_busybox >events.log) &
+	# Spawn two subshels:
```
mrunalp, line 44: nit: subshells

mrunalp (approved):

thaJeztah:
Nice, good to see it worked; thanks!

kolyshkin:
> Nice, good to see it worked; thanks!

Thank you so much for your help @thaJeztah!

PR #2723, opened 2021-01-07
Fix a race in tty code leading to \r in output

kolyshkin:
The TestExecInTTY test case is sometimes failing like this:
    
> execin_test.go:332: unexpected carriage-return in output "PID USER TIME COMMAND\r\n 1 root 0:00 cat\r\n 7 root 0:00 ps\r\n"
    
or this:
    
> execin_test.go:332: unexpected carriage-return in output "PID USER TIME COMMAND\r\n 1 root 0:00 cat\n 7 root 0:00 ps\n"
    
(this is easy to repro with `go test -run TestExecInTTY -count 1000`).
    
This is caused by a race between
    
 - an `Init()` (in this case it is is `(*linuxSetnsInit.Init()`, but
   `(*linuxStandardInit).Init()` is no different in this regard),
   which creates a pty pair, sends pty master to runc, and execs
   the container process,
    
and
    
 - a parent runc process, which receives the pty master fd and calls
   `ClearONLCR()` on it.

One way of fixing it would be to add a synchronization mechanism
between these two, so `Init()` won't exec the process until the parent
sets the flag. This seems excessive, though, as we can just move
the `ClearONLCR()` call to `Init()`, putting it right after `console.NewPty()`.

Note that bug was only seen in the `TestExecInTTY` test case, but
from looking at the code it seems like it can also happen during
`runc run` or `runc exec`.

While at it:
 - simplify/improve the test case;
 - make it repeat 300 times for better chances to catch the bug.

Fixes: https://github.com/opencontainers/runc/issues/2425

### DIFF
--- a/libcontainer/integration/execin_test.go
+++ b/libcontainer/integration/execin_test.go
@@ -291,46 +291,55 @@ func TestExecInTTY(t *testing.T) {
 		Args: []string{"ps"},
 		Env:  standardEnvironment,
 	}
-	parent, child, err := utils.NewSockPair("console")
-	if err != nil {
-		ok(t, err)
-	}
-	defer parent.Close()
-	defer child.Close()
-	ps.ConsoleSocket = child
-	done := make(chan (error))
-	go func() {
-		f, err := utils.RecvFd(parent)
-		if err != nil {
-			done <- fmt.Errorf("RecvFd: %w", err)
-			return
-		}
-		c, err := console.ConsoleFromFile(f)
-		if err != nil {
-			done <- fmt.Errorf("ConsoleFromFile: %w", err)
-			return
-		}
-		// An error from io.Copy is expected once the terminal
-		// is gone, so we deliberately ignore it.
-		_, _ = io.Copy(&stdout, c)
-		done <- nil
-	}()
-	err = container.Run(ps)
-	ok(t, err)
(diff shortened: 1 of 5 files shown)


libcontainer/integration/execin_test.go
```diff
@@ -291,46 +291,55 @@ func TestExecInTTY(t *testing.T) {
...
-		f, err := utils.RecvFd(parent)
-		if err != nil {
-			done <- fmt.Errorf("RecvFd: %w", err)
-			return
-		}
-		c, err := console.ConsoleFromFile(f)
+
+	// Repeat to increase chances to catch a race; see
+	// https://github.com/opencontainers/runc/issues/2425.
+	for i := 0; i < 300; i++ {
```
kolyshkin, line 297: Now, this commit is totally optional -- I am not sure myself I want it, since the bug is fixed and it takes about 5s on my laptop for this test now.

kolyshkin:
@cyphar PTAL

kolyshkin:
@AkihiroSuda @mrunalp PTAL

This fixes a (now) frequent test failure which I am tired of seeing.

AkihiroSuda (approved):

mrunalp (approved):

PR #2725, opened 2021-01-08
 systemd: fix rootful-in-userns regression

### LINKED ISSUE #2724: [Rootful in UserNS] Regression: systemd rootless attempted when cgroup v1 in use
Hello, I've noticed a runc regression by using Podman on a Chromebook's 'Linux beta' environment, which is furnished via LXC. This is the array of conditions at play:

* Running podman/runc as root
* Apparently within a userns (presumably because of LXC)
* Systemd is running
* cgroup v1 *is* mounted, cgroup v2 *not* mounted

Given all this, latest runc fails:

```
> git checkout v1.0.0-rc92
> make
> make install
> sudo podman run --rm busybox echo hi world
Error: OCI runtime error: cgroup v2 not enabled on this host, can't use systemd (rootless) as cgroups manager
```

Debian ships with this ancient runc which works:

```
> runc --version
runc version 1.0.0~rc6+dfsg1
commit: 1.0.0~rc6+dfsg1-3
spec: 1.0.1
> sudo podman run --rm busybox echo hi world
hi world
```

I've written a simple git-bisect script to find the exact cause, which pointed at https://github.com/opencontainers/runc/commit/bf15cc99b1e7b486dfbb394992fcc4b38157d56e .

Seems like systemd-rootless shouldn't be enabled unless cgroupv2 is enabled.

Initial investigative work happened here: https://github.com/containers/podman/issues/8881

AkihiroSuda:
`shouldUseRootlessCgroupManager()` was always returning true when we are inside a userns.

If the systemd OwnerUID is 0, we have rootful systemd inside the userns, and we can just use the rootful cgroup driver.

Fix #2724

### DIFF
--- a/rootless_linux.go
+++ b/rootless_linux.go
@@ -5,7 +5,9 @@ package main
 import (
 	"os"
 
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
 	"github.com/opencontainers/runc/libcontainer/system"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
@@ -28,6 +30,23 @@ func shouldUseRootlessCgroupManager(context *cli.Context) (bool, error) {
 		return false, nil
 	}
 	// euid = 0, in a userns.
+	//
+	// [systemd driver]
+	// We can call DetectUID() to parse the OwnerUID value from `busctl --user --no-pager status` result.
+	// The value corresponds to sd_bus_creds_get_owner_uid(3).
+	// If the value is 0, we have rootful systemd inside userns, so we do not need the rootless cgroup manager.
+	//
+	// On error, we assume we are root. An error may happen during shelling out to `busctl` CLI,
+	// mostly when $DBUS_SESSION_BUS_ADDRESS is unset.
+	if context.GlobalBool("systemd-cgroup") {
+		ownerUID, err := systemd.DetectUID()
+		if err != nil {
+			logrus.WithError(err).Debug("failed to get the OwnerUID value, assuming the value to be 0")
+			ownerUID = 0
+		}
+		return ownerUID != 0, nil
+	}
+	// [cgroupfs driver]
 	// As we are unaware of cgroups path, we can't determine whether we have the full
 	// access to the cgroups path.
 	// Either way, we can safely decide to use the rootless cgroups manager.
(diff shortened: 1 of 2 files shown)


danopia:
Running this branch, I saw the new warning logged repeatedly while running an image build, does that make sense?

```
WARN[0000] failed to get the OwnerUID value, assuming the value to be 0  error="could not execute `busctl --user --no-pager status`: \"Failed to set address: No such file or directory\\n\": exit status 1"
```

Screenshot 2021-01-08 at 22 04 09

The build still completed successfully.

AkihiroSuda:
Updated PR to silence the log by changing the log level from Warn to Debug.

> root doesn't have a user session on my machine (should it??) 

It may have, but shouldn't be a requirement for root.

AkihiroSuda:
CI is currently broken. Will be fixed in https://github.com/opencontainers/runc/pull/2731 .

AkihiroSuda:

rootless_linux.go
```diff
@@ -28,6 +30,23 @@ func shouldUseRootlessCgroupManager(context *cli.Context) (bool, error) {
...
+	// We can call DetectUID() to parse the OwnerUID value from `busctl --user --no-pager status` result.
+	// The value corresponds to sd_bus_creds_get_owner_uid(3).
+	// If the value is 0, we have rootful systemd inside userns, so we do not need the rootless cgroup manager.
+	//
+	// On error, we assume we are root. An error may happen during shelling out to `busctl` CLI,
+	// mostly when $DBUS_SESSION_BUS_ADDRESS is unset.
+	if context.GlobalBool("systemd-cgroup") {
+		ownerUID, err := systemd.DetectUID()
+		if err != nil {
+			logrus.WithError(err).Debug("failed to get the OwnerUID value, assuming the value to be 0")
```
kolyshkin, line 44: Maybe upgrade this from Debug to a Warning? Otherwise the user will never know why it's not working as rootless.

kolyshkin:
A single question, LGTM otherwise

rootless_linux.go
AkihiroSuda: Intentionally demoted to Warning: https://github.com/opencontainers/runc/pull/2725#issuecomment-756997560

kolyshkin (approved):
LGTM

mrunalp (approved):

PR #2727, opened 2021-01-08
libc/cg: convert r.CPU.Cpus/Mems to systemd props

kolyshkin:
Support for systemd properties AllowedCPUs and AllowedMemoryNodes
was added by commit 13afa58d0e46ceb0, but only for unified resources
of systemd v2 driver.

This adds support for Cpu.Cpus and Cpu.Mems resources to
both systemd v1 and v2 cgroup drivers. If systemd is not new
enough to support the properties, a  warning is logged.

An integration test is added to check that the settings work.

While at it:
 - log a warning in a different place checking systemd version;
 - add version check to systemd unified support handling.

### DIFF
--- a/tests/integration/update.bats
+++ b/tests/integration/update.bats
@@ -392,7 +392,63 @@ EOF
 	check_systemd_value "TasksMax" 10
 }
 
+@test "update cpuset parameters via resources.CPU" {
+	[[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
+	requires smp
+
+	local AllowedCPUs='AllowedCPUs' AllowedMemoryNodes='AllowedMemoryNodes'
+	# these properties require systemd >= v244
+	if [ "$(systemd_version)" -lt 244 ]; then
+		# a hack to skip checks, see check_systemd_value()
+		AllowedCPUs='unsupported'
+		AllowedMemoryNodes='unsupported'
+	fi
+
+	update_config ' .linux.resources.CPU |= {
+				"Cpus": "0",
+				"Mems": "0"
+			}' "${BUSYBOX_BUNDLE}"
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_update
+	[ "$status" -eq 0 ]
+
+	# check that initial values were properly set
+	check_systemd_value "$AllowedCPUs" 0
+	check_systemd_value "$AllowedMemoryNodes" 0
+
+	runc update -r - test_update <<EOF
+{
+  "CPU": {
+    "Cpus": "1"
+  }
+}
+EOF
+	[ "$status" -eq 0 ]
+
+	# check the updated systemd unit properties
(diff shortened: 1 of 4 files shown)


kolyshkin:
This one is ready

tests/integration/update.bats
```diff
@@ -392,6 +392,53 @@ EOF
...
+	file="/sys/fs/cgroup/cpuset.mems.effective"
+	if ! test -r $file || grep -q '^0$' $file; then
+		# skip the rest of it
+		return 0
+	fi
+
+	runc update -r - test_update <<EOF
+{
+  "CPU": {
+    "Cpus": "1"
```
mrunalp, line 432: This should be `Mems`, right?

kolyshkin: Good catch, thanks! Alas this part of test is skipped in CI as we need a NUMA machine for it.

Fixed.

kolyshkin:
> time="2021-01-14T01:13:19Z" level=error msg="container_linux.go:367: starting container process caused: process_linux.go:340: applying cgroup configuration for process caused: Cannot set property AllowedCPUs, or unknown property."
150

So we need systemd version check...

kolyshkin:
```
not ok 104 update cpuset parameters via resources.CPU
# (from function `check_systemd_value' in file tests/integration/helpers.bash, line 207,
#  in test file tests/integration/update.bats, line 407)
#   `check_systemd_value "AllowedCPUs" 0' failed
# runc spec (status=0):
# runc run -d --console-socket /tmp/console.sock test_update (status=0):
# time="2021-01-14T02:35:35Z" level=warning msg="systemd v219 is too old to support CPUQuotaPeriodSec  (setting will still be applied to cgroupfs)"
# time="2021-01-14T02:35:35Z" level=warning msg="systemd v219 is too old to support AllowedCPUs/AllowedMemoryNodes (settings will still be applied to cgroupfs)"
# time="2021-01-14T02:35:35Z" level=warning msg="systemd v219 is too old to support CPUQuotaPeriodSec  (setting will still be applied to cgroupfs)"
# time="2021-01-14T02:35:35Z" level=warning msg="systemd v219 is too old to support AllowedCPUs/AllowedMemoryNodes (settings will still be applied to cgroupfs)"
# systemd AllowedCPUs: current  !? 0 
ok 105 update cpuset parameters via v2 unified map # skip test requires cgroups_v2
ok 106 update rt period and runtime # skip test requires no_systemd
ok 107 update devices [minimal transition rules]
ok 108 update paused container
ok 109 runc version
make: *** [localintegration] Error 1
make: Leaving directory `/vagrant'
Script started, file is typescript
```

Two issues:
1. test is run on centos7 where it obviously fails, but it should be skipped. **fixed**
2. test has failed and test no failure is reported. **Investigating**

AkihiroSuda (approved):

mrunalp (approved):

PR #2729, opened 2021-01-11
script/check-config.sh: fixes for cgroupv2 and v5.x kernels

kolyshkin:
This fixes script/check-config.sh to
- not produce false failures on cgroupv2;
- not check removed options on recent 5.x kernels.


script/check-config.sh
```diff
@@ -164,17 +164,21 @@ echo
 echo 'Generally Necessary:'
 
 echo -n '- '
-cgroupSubsystemDir="$(awk '/[, ](cpu|cpuacct|cpuset|devices|freezer|memory)[, ]/ && $3 == "cgroup" { print $2 }' /proc/mounts | head -n1)"
-cgroupDir="$(dirname "$cgroupSubsystemDir")"
-if [ -d "$cgroupDir/cpu" -o -d "$cgroupDir/cpuacct" -o -d "$cgroupDir/cpuset" -o -d "$cgroupDir/devices" -o -d "$cgroupDir/freezer" -o -d "$cgroupDir/memory" ]; then
-	echo "$(wrap_good 'cgroup hierarchy' 'properly mounted') [$cgroupDir]"
+if [[ $(stat -f -c %t /sys/fs/cgroup 2>/dev/null) = "63677270" ]]; then
```
tianon, line 167: `[[` isn't necessary here; technically we should do `$()` in a separate command to catch the failure, but in this case we don't actually care about the failure (and the `else` should cover it) so it's fine:

```suggestion
if [ "$(stat -f -c %t /sys/fs/cgroup 2>/dev/null)" = '63677270' ]; then
```

tianon (commented):
Generally LGTM, just one minor nit :+1: :heart:

tianon:
(Sorry, saw `check-config.sh` and assumed this was `moby/moby` and didn't realize it was `runc` until I had already commented :see_no_evil:  Hopefully forgivable since I'm the original author of the first version of this script? :innocent: :heart:  Feel free to ignore me if maintainers here don't agree. :+1:)

script/check-config.sh
kolyshkin: Thanks, makes sense (I am unsure why I used `[[` in the first place).

Fixed.

kolyshkin:
CI failure is #2425, fixed by #2723. Restarting...

```
=== RUN   TestExecInTTY
836
    execin_test.go:349: unexpected carriage-return in output "PID   USER     TIME  COMMAND\r\n    1 root      0:00 cat\r\n    6 root      0:00 ps\r\n"
837
--- FAIL: TestExecInTTY (0.10s)
```

script/check-config.sh
```diff
@@ -164,17 +164,21 @@ echo
 echo 'Generally Necessary:'
 
 echo -n '- '
-cgroupSubsystemDir="$(awk '/[, ](cpu|cpuacct|cpuset|devices|freezer|memory)[, ]/ && $3 == "cgroup" { print $2 }' /proc/mounts | head -n1)"
-cgroupDir="$(dirname "$cgroupSubsystemDir")"
-if [ -d "$cgroupDir/cpu" -o -d "$cgroupDir/cpuacct" -o -d "$cgroupDir/cpuset" -o -d "$cgroupDir/devices" -o -d "$cgroupDir/freezer" -o -d "$cgroupDir/memory" ]; then
-	echo "$(wrap_good 'cgroup hierarchy' 'properly mounted') [$cgroupDir]"
+if [ "$(stat -f -c %t /sys/fs/cgroup 2>/dev/null)" = "63677270" ]; then
+	echo "$(wrap_good 'cgroup hierarchy' 'cgroupv2')"
```
AkihiroSuda, line 168: I think we need more validations for cgroup2, but it can be worked out later, so LGTM

AkihiroSuda (approved):

mrunalp (approved):

PR #2731, opened 2021-01-12
tests/int/multi-arch.bash: fix busybox URL

kolyshkin:
Due to https://github.com/docker-library/busybox/pull/94 the URLs
to get busybox has *just* changed, so the current URLs give HTTP 404.

Fix the URLs accordingly to bring CI back to :heavy_check_mark: 

NOTE there can be many improvements made to this script, but
the whole point of this PR is to fix CI, and we can work on the rest
of it separately.

### DIFF
--- a/tests/integration/multi-arch.bash
+++ b/tests/integration/multi-arch.bash
@@ -2,10 +2,10 @@
 get_busybox() {
 	case $(go env GOARCH) in
 	arm64)
-		echo 'https://github.com/docker-library/busybox/raw/dist-arm64v8/glibc/busybox.tar.xz'
+		echo 'https://github.com/docker-library/busybox/raw/dist-arm64v8/stable/glibc/busybox.tar.xz'
 		;;
 	*)
-		echo 'https://github.com/docker-library/busybox/raw/dist-amd64/glibc/busybox.tar.xz'
+		echo 'https://github.com/docker-library/busybox/raw/dist-amd64/stable/glibc/busybox.tar.xz'
 		;;
 	esac
 }


tests/integration/multi-arch.bash
```diff
@@ -2,10 +2,10 @@
 get_busybox() {
 	case $(go env GOARCH) in
 	arm64)
-		echo 'https://github.com/docker-library/busybox/raw/dist-arm64v8/glibc/busybox.tar.xz'
+		echo 'https://github.com/docker-library/busybox/raw/dist-arm64v8/stable/glibc/busybox.tar.xz'
```
tianon, line 5: Wouldn't it be more stable to use a specific commit here and document how/when to update it? :innocent: :grimacing:

kolyshkin: We always used the latest version, and I never saw an issue. Once we'll switch to a specific version, it will never be updated :(

Your point is valid though. Let me see.

kolyshkin: @tianon updated; PTAL

tests/integration/multi-arch.bash
```diff
@@ -1,13 +1,22 @@
...
+	local sha
+
 	case $(go env GOARCH) in
 	arm64)
-		echo 'https://github.com/docker-library/busybox/raw/dist-arm64v8/glibc/busybox.tar.xz'
+		# To update to the latest package, copy top commit id from
+		# https://github.com/docker-library/busybox/commits/dist-arm64v8?author=docker-library-bot
+		sha=385dcf62178190a3cc8ef5129a3236466d3abab5
 		;;
 	*)
```
AkihiroSuda, line 12: * -> amd64

kolyshkin: That's the way it was, I am not changing anything here. Seems that amd64 is the default.

We can certainly error out here on an unknown arch.

AkihiroSuda (dismissed):

AkihiroSuda (approved):

AkihiroSuda (changes requested):
`/home/runner/work/runc/runc/tests/integration/helpers.bash: line 50: syntax error near unexpected token `('`

tests/integration/multi-arch.bash
tianon: LGTM; busybox should be pretty stable (and limited scope here), so I doubt it really *needs* to be updated very often :sweat_smile:

tests/integration/helpers.bash
```diff
@@ -17,8 +17,7 @@ BUSYBOX_IMAGE="$BATS_TMPDIR/busybox.tar"
 BUSYBOX_BUNDLE="$BATS_TMPDIR/busyboxtest"
 
 # hello-world in tar format
-HELLO_FILE=$(get_hello)
-HELLO_IMAGE="$TESTDATA/$HELLO_FILE"
+HELLO_IMAGE="$TESTDATA/hello-world-$(go env GOARCH).tar
```
tianon, line 20:
```suggestion
HELLO_IMAGE="$TESTDATA/hello-world-$(go env GOARCH).tar"
```

:eyes:

kolyshkin: yup; fixed

AkihiroSuda:
`/bin/sh: 7: tests/integration/multi-arch.bash: Syntax error: "(" unexpected (expecting "}")`

kolyshkin:
OK now I added a bashism to multi-arch.sh script and need to modify how it's executed from Dockerfile... Overall, I want more changes to the script; at the same time we need to fix CI fast.

Therefore, let's split this PR into two:
1. Fix busybox URL (first commit). This is the PR.
2. All the fancy improvements (the rest of it). TODO

tests/integration/multi-arch.bash
kolyshkin: OK I am moving all the improvements to a separate PR, as this one we need to merge ASAP to fix CI.

AkihiroSuda (approved):

mrunalp (approved):

Issue #2735, opened 2021-01-14
seccomp specification has non-obvious semantics


cyphar:
Right now, our seccomp rule specifications have this strange behaviour where if two arguments touch the same argument index, the argument list is treated as a series of OR conditions rather than AND conditions. This was added by e9193ba6e612c13c064449ee4189b268d1a9c94c to fix a breaking change, but the net result is that our seccomp rule generation feels at least slightly unpredictable as a result.

We should really rework this alongside the longer-term fix for #2151.

cyphar:
This also means that you cannot create a seccomp rule which only permits a syscall if more than one conditional is met (such as `value > X && value < Y` or something similar. This appears to have been an oversight in the original implementation but I'm not clear how exactly we could change this now, several years later.

Issue #1847, opened 2018-07-19
seccomp rejects redundant rules with "requested action matches default action of filter"


alban:
I am testing a config.json with seccomp rules that defaults to "allow" and a syscall with the same action "allow".

<details>

```
		"seccomp": {
			"defaultAction": "SCMP_ACT_ALLOW",
			"architectures": [
				"SCMP_ARCH_X86_64",
				"SCMP_ARCH_X86",
				"SCMP_ARCH_X32"
			],
			"syscalls": [
				{
					"names": [
						"personality"
					],
					"action": "SCMP_ACT_ALLOW",
					"args": [
						{
							"index": 0,
							"value": 0,
							"op": "SCMP_CMP_EQ"
						},
						{
							"index": 0,
							"value": 8,
							"op": "SCMP_CMP_EQ"
						},
						{
							"index": 0,
							"value": 4294967295,
							"op": "SCMP_CMP_EQ"
						}
					]
				},
				{
					"names": [
						"clone"
					],
					"action": "SCMP_ACT_ALLOW",
					"args": [
						{
							"index": 0,
							"value": 2080505856,
							"op": "SCMP_CMP_MASKED_EQ"
						}
					]
				},
				{
					"names": [
						"getcwd"
					],
					"action": "SCMP_ACT_ERRNO"
				}
			]
		}
```

</details>

runc returns the error
```
error adding seccomp rule for syscall personality: requested action matches default action of filter
```

`config.json` could be programatically generated with this kind of redundant rules. Could runc accepts them instead and in this case drop the no-op rule about the `personality` syscall?

In the [spec about actions](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#seccomp), it says:
> action (string, REQUIRED) - the action for seccomp rules

But it does not say the action has to be different to the default action.

If you think runc's behaviour is correct, I could file an issue on the spec instead to say that the action has to be different than the default action.

This issue was found with [runtime-tools test linux_seccomp.go](https://github.com/opencontainers/runtime-tools/pull/514/files).

mrunalp:
@alban I think that runc's (and libseccomp) behavior is correct in this place and should get the spec updated for this.
cc: @giuseppe

kolyshkin:
It seems that I just fixed it in https://github.com/opencontainers/runc/pull/3109

PR #2737, opened 2021-01-14
seccomp: force -ENOSYS as new default action

### LINKED ISSUE #2151: seccomp filter should return ENOSYS for unknown syscalls

cyphar:
Having -EPERM is the default was a fairly significant mistake from a
future-proofing standpoint in that it makes any new syscall return a
non-ignorable error (from glibc's point of view). We need to correct
this now because faccessat2(2) is something glibc *critically* needs to
have support for, but they're blocked on container runtimes because we
return -EPERM unconditionally (leading to confusion in glibc).

Unfortunately there are several issues which stop us from having a clean
solution to this problem:

 1. libseccomp has several limitations which require us to emulate
	behaviour we want:

	a. We cannot do logic based on syscall number, meaning we cannot
	   specify a "largest known syscall number";
	b. libseccomp doesn't know in which kernel version a syscall was
	   added, and has no API for "minimum kernel version" so we cannot
	   simply ask libseccomp to generate sane -ENOSYS rules for us.
	c. Additional seccomp rules for the same syscall are not treated as
	   distinct rules -- if rules overlap, seccomp will merge them. This
	   means we cannot add per-syscall -EPERM fallbacks;
	d. There is no inverse operation for SCMP_CMP_MASKED_EQ;
	e. libseccomp does not allow you to specify multiple rules for a
	   single argument, making it impossible to invert OR rules for
	   arguments.

 2. The runtime-spec does not have any way of specifying:

	a. The errno for the default action;
	b. The minimum kernel version or "newest syscall at time of profile
	   creation"; nor
	c. Which syscalls were intentionally excluded from the allow list
	   (weird syscalls that are no longer used were excluded entirely,
	   but Docker et al expect those syscalls to get EPERM not ENOSYS).

 3. Certain syscalls should not return -ENOSYS (especially only for
	certain argument combinations) because this could also trigger glibc
	confusion. This means we have to return -EPERM for certain syscalls
	but not as a global default.

 4. There is not an obvious (and reasonable) upper limit to syscall
	numbers, so we cannot create a set of rules for each syscall above
	the largest syscall number in libseccomp. This means we must handle
	inverse rules as described below.

 5. Any syscall can be specified multiple times, which can make
	generation of hotfix rules much harder.

As a result, we have to work around all of these things by coming up
with a heuristic to stop the bleeding. In the future we could hopefully
improve the situation in the runtime-spec and libseccomp.

The solution is to switch the default action errno to -ENOSYS, and
simply pick a Linux kernel version which is treated as the universal
minimum version (Linux 3.0 was picked since basically no widely-used
distribution ships an older kernel, and it contains enough common
syscalls which would cause confusion if -ENOSYS was returned).

We then generate -EPERM rules for each syscall which:

 1. Existed in Linux 3.0 (read: has a syscall number smaller than
	"setns" on that platform) and does not have any explicit rules in
	the configuration. These syscalls get an unconditional -EPERM,
	because they are assumed to have been excluded intentionally.

 2. Have an explicit *conditional* rule. These syscalls get a set of
	-EPERM rules which invert the logic of the conditional, to ensure
	that syscalls which are only partially permitted do not return
	-ENOSYS.

	This is trivial for all operations *except* SCMP_CMP_MASKED_EQ,
	which has no inverse operatin. For MASKED_EQ we generate a series of
	MASKED_EQ rules except for the value which was checked in the
	original rule.

	There are additional complications, which are all treated as simply
	"too hard" and will thus fall back to -ENOSYS:

	a. Any syscall specified in more than one seccomp rule is treated as
	   "too hard to handle" and will be ignored. This is because they
	   are logical OR rules but things become far too complicated if the
	   action is non-trivial.
	b. Syscalls which have OR rules (the same argument specified
	   multiple times) because it would require AND support in
	   libseccomp, which doesn't exist.
	c. Syscalls which do not have allow rules (ACT_LOG or ACT_ALLOW),
	   because other actions have different logical meanings to allow
	   rules and it'd be too complicated to handle those as well for a
	   hotfix.

None of the above happens if the default action is not SCMP_RET_ERRNO.

Fixes #2151


cyphar:
I'm not sure if this works in all scenarios because I haven't tested this thoroughly yet, but I'm pushing this so folks can comment.

/cc @richfelker @jethrob @fweimer Does this seem reasonable-ish to you? This is more complicated than I'd like but it's a result of several legacy aspects of runc's seccomp code (one of which is #2735 -- and is probably a bug).

richfelker:
@cyphar: I still don't entirely understand the need for inverse rules rather than a well-defined fallback order, but if the latter is hard, I guess they work. At least the patch does not look too large/invasive and the high-level explanation of what it's doing seems to be reasonable and in alignment with what I expected.

cyphar:
@richfelker As far as I understand, libseccomp doesn't have per-syscall fallback order (or at least it's not guaranteed to do so) -- if you add an unconditional `-EPERM` rule for a syscall it will remove any other rules present for that syscall. I'm pretty sure that the code for generating this is imperfect so we could probably hack around it (by doing `MASKED_EQ` of `0` to `0`) but it seems a bit like playing with fire to do something like that.

Once seccomp/libseccomp#286 is resolved we can switch away from this workaround.

cyphar:
It appears that libseccomp doesn't support a single rule referencing the same argument more than once (this is enforced by `db_col_rule_add` but appears to be undocumented). This means that we cannot invert OR-ified rules properly (as per De-Morgan's laws, a negation of OR statements becomes an AND statement of negations).

Just to make sure, I've tried adding a basic fallback rule (even with the `MASKED_EQ`) trick and as far as I can tell it does override existing rules -- though strangely it only overrides conditional rules, and non-conditional rules are not overridden (my guess it's because they are treated separately internally).

I'm starting to get a creeping feeling that this might've been simpler to implement in BPF without any other workarounds.

/cc @pcmoore Am I misunderstanding libseccomp, or are these real restrictions on filters created with libseccomp?

cyphar:
I'm also starting to think it may be simplest to simply implement this with `EPERM` as the default action and explicit `ENOSYS` values until libseccomp allows us to fix this properly...

cyphar:
Okay this latest version handles things pretty well. The main thing which we cannot handle is OR rules for arguments (because of the above libseccomp restriction) but this doesn't impact the default Docker rules and thus is not a major issue -- besides I would argue the current implementation is actually a bug (see #2735).

All-in-all, I've tested this with a variety of cases and it does appear to work correctly. I'm marking this as ready for review -- PTAL @opencontainers/runc-maintainers. I'd also appreciate a review from @justincormack et al.

cyphar:
I have a test program I've been using, but I'm not sure what's the nicest way to convert it to an integration test:

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <assert.h>
#include <errno.h>
#include <fcntl.h>

#include <sys/syscall.h>
#include <linux/openat2.h>

int openat2(int dfd, const char *path, struct open_how *how, size_t usize)
{
	int ret = syscall(SYS_openat2, dfd, path, how, usize);
	if (ret < 0)
		ret = -errno;
	return ret;
}

int pidfd_open(pid_t pid, unsigned int flags)
{
	int ret = syscall(SYS_pidfd_open, pid, flags);
	if (ret < 0)
		ret = -errno;
	return ret;
}

int _dup3(int oldfd, int newfd, int flags)
{
	int ret = dup3(oldfd, newfd, flags);
	if (ret < 0)
		ret = -errno;
	return ret;
}

int fake_syscall(void)
{
	int ret = syscall(1000);
	if (ret < 0)
		ret = -errno;
	return ret;
}

#define debug_assert(lval, rval)											\
	do {																	\
		int L = (lval), R = (rval);											\
		if (L != R) {														\
			printf("debug_assert(%s == %s) failed: %d != %d\n", #lval, #rval, L, R); \
		}																	\
	} while (0)

int main(void)
{
	// Basic rule.

	// Multiple arguments with OR rules.
	debug_assert(openat2(100, NULL, NULL, 0), -EINVAL);
	debug_assert(openat2(9001, NULL, NULL, 0), -EINVAL);
	debug_assert(openat2(0, NULL, NULL, 0), -ENOSYS); // :(
	debug_assert(openat2(0, NULL, NULL, 0), -ENOSYS); // :(

	// Multiple arguments with AND rules.
	debug_assert(pidfd_open(1, 1337), -EINVAL);
	debug_assert(pidfd_open(1, 0), -EPERM);
	debug_assert(pidfd_open(500, 1337), -EPERM);
	debug_assert(pidfd_open(500, 500), -EPERM);

	// Multiple rules for the same syscall.
	debug_assert(_dup3(0, -100, 0xFFFF), -ENOSYS);
	debug_assert(_dup3(1, -100, 0xFFFF), -EINVAL);
	debug_assert(_dup3(2, -100, 0xFFFF), -ENOSYS);
	debug_assert(_dup3(3, -100, 0xFFFF), -EINVAL);

	// Out-of-bounds syscall.
	debug_assert(fake_syscall(), -ENOSYS);
}
```

giuseppe:
it seems to work well with the default Podman seccomp profile, that has some weird handling for `socket`: https://github.com/containers/common/blob/master/pkg/seccomp/seccomp.json#L791-L897

AkihiroSuda:
> I have a test program I've been using, but I'm not sure what's the nicest way to convert it to an integration test:

I think we can just compile and run the C code from bats

cyphar:
@giuseppe That example is actually ignored entirely by the workaround code -- you will get `-ENOSYS` as a fallback because we can't generate a proper rule-set for the fallback. Hopefully that doesn't break anything :crossed_fingers:.

cyphar:
I hit my first `-ENOSYS` issue -- `renameat2` in glibc hides `-ENOSYS` and will return a fake `-EINVAL` if you have non-zero flag arguments. I've worked around it but boy do I really hope there aren't many more regressions like this...

giuseppe:
> @giuseppe That example is actually ignored entirely by the workaround code -- you will get `-ENOSYS` as a fallback because we can't generate a proper rule-set for the fallback. Hopefully that doesn't break anything crossed_fingers.

I see.  I guess it works because `socket` doesn't rely on the default return value in this case, and it will be allowed by the next rule.

Differently, both `personality` and `clone` get -ENOSYS by default, but I think these syscalls are more difficult to cause issues

cyphar:
`clone` shouldn't be getting `ENOSYS` -- while the podman/docker profile has multiple references to it, only one rule is actually generated (at least according to my tests). `clone` should still be getting `EPERM` but yeah the `personality` configuration appears to trigger `ENOSYS` too...

fweimer-rh:
> I hit my first `-ENOSYS` issue -- `renameat2` in glibc hides `-ENOSYS` and will return a fake `-EINVAL` if you have non-zero flag arguments. I've worked around it but boy do I really hope there aren't many more regressions like this...

`mlock2` does this as well. The idea is to mirror what the kernel would do if it encounters an unknown/unsupported flag. If the system call does not have a flag argument, then all flags are unsupported.

I don't see how this could be a problem. Could you please elaborate?

You can make a direct system call using `syscall (SYS_renameat2, …)` to avoid this, but `SYS_renameat2` might not be defined if it's missing from the kernel headers (likewise the `renameat2` function is only available since glibc 2.28).

cyphar:
@fweimer-rh The (potential) issue is that programs which used to get `-EPERM` when a seccomp profile was violated (such as "no `clone(CLONE_NEWUSER)` for containers") may now start getting strange errors like `-EINVAL` or `-ENOENT` depending on the glibc fallback code. While it is still an error, it is possible that userspace programs have fallbacks for `-EPERM` but not for `-EINVAL` (or `-ENOSYS`). runc has fallback code that uses `os.IsPermission` which won't catch `-EINVAL` nor `-ENOENT`.

I don't think there is a nice solution for this outside of making `libseccomp` be able to handle this whole problem more nicely, but it is something that does worry me a fair bit...

fweimer-rh:
@cyphar The `EINVAL` behavior was specifically motivated by the existing kernel behavior. [`clone(2)`](https://www.man7.org/linux/man-pages/man2/clone.2.html) says this:

> **`EINVAL`**
>
> `CLONE_NEWUSER` was specified in the flags mask, but the kernel was not configured with the `CONFIG_USER_NS` option.

My mental model is that seccomp is essentially deconfiguring otherwise supported kernel interfaces, so returning `EINVAL` for filtered flags is not wrong per se.

(Note that glibc doesn't do anything like that with `clone3` currently, so this is hypothetical at this point, but `clone(2)` is fairly well-documented, so it's a good example.)

cyphar:
@fweimer-rh 

> My mental model is that seccomp is essentially deconfiguring otherwise supported kernel interfaces, so returning EINVAL for filtered flags is not wrong per se.

I appreciate this is a bit of a philosophical discussion, but my mental model is that we are acting like an ad-hoc MAC or LSM-like system which is explicitly restricting features. This isn't true for unknown syscalls, but it is definitely true for cases like `clone(CLONE_NEWUSER)`.

pcmoore:
> It appears that libseccomp doesn't support a single rule referencing the same argument more than once (this is enforced by `db_col_rule_add` but appears to be undocumented).

My hope was that the `seccomp_rule_add(3)` manpage made this clear (excerpt below), suggestions on how to improve the documentation are always welcome.

> When adding syscall argument comparisons to the filter it is important to remember that while it is possible to have multiple comparisons in a single rule, you can only compare each argument once in a single rule.  In other words, you can not have multiple comparisons of the 3rd syscall argument in a single rule.

* https://www.mankier.com/3/seccomp_rule_add

cyphar:
@pcmoore It seems that text is only available in `man '3+1' seccomp_rule_add` on my distribution, not `man 3 secomp_rule_add`. So it is documented, just not in the version of the documentation I was reading. :grimacing:

cyphar:
Seems like the `--debug` test is flaky, it's failed about 30% of the time I've re-run CI on this PR...

cyphar:
Hmm, now the test is failing on Fedora with no warnings or any output other than `runc run` exiting with a status of `218`...

cyphar:
Hmm, now the test is failing on CentOS 7 with no warnings or any output other than `runc run` exiting with a status of `218`. Oh and it looks like the GHA integration tests aren't catching test failures under Vagrant? :grimacing:

pcmoore:
> @pcmoore It seems that text is only available in `man '3+1' seccomp_rule_add` on my distribution, not `man 3 secomp_rule_add`. So it is documented, just not in the version of the documentation I was reading.

Sometimes it's easy to miss things in the docs too; I've done it more times than I care to remember.

The `man 3+1 <blah>` is new to me, and a quick search through man's manpage doesn't show anything jumping out.  Odd.  Out of curiosity, what distro are you using?

justincormack:
This seems totally the wrong place to do this, runc is (sigh) just doing what it is told and if you have to second guess its behaviour too thats even worse. There are only a few seccomp policies in the world, just fix them.

richfelker:
> > It appears that libseccomp doesn't support a single rule referencing the same argument more than once (this is enforced by db_col_rule_add but appears to be undocumented).
> 
> My hope was that the seccomp_rule_add(3) manpage made this clear (excerpt below), suggestions on how to improve the documentation are always welcome.

I don't think this is a documentation problem. The underlying problem is that the DSL/libseccomp API for this is just utterly misdesigned in a way that's anti-intuitive, harder to use than, and less powerful than, raw BPF. Is there no way it can be fixed to be somewhat less bad while maintaining compatibility with important caller expectations? If not, can the necessary fixups just be done on the resulting BPF rather than via the deficient libseccomp API?

pcmoore:
> I don't think this is a documentation problem. The underlying problem is that the DSL/libseccomp API for this is just utterly misdesigned in a way that's anti-intuitive, harder to use than, and less powerful than, raw BPF. Is there no way it can be fixed to be somewhat less bad while maintaining compatibility with important caller expectations? If not, can the necessary fixups just be done on the resulting BPF rather than via the deficient libseccomp API?

That libseccomp library is just awful isn't it?  As I understand it the libseccomp maintainers are open to accepting patches to improve the library; which is good, because they are clearly a bunch of idiots who don't know what they are doing.

Alternatively you could simply write your own seccomp filters in raw BPF, that's clearly what a *real* programmer would do in this case.  The libseccomp maintainers openly say that libseccomp is not for everyone or every use case; they say the goal is to lower the bar for the common seccomp use cases in hopes that more applications/developers will make use of seccomp.  However, what they should really be saying is that libseccomp is *always* worse than writing raw BPF and only idiots should use it.

cyphar:
@pcmoore 

> The man 3+1 <blah> is new to me, and a quick search through man's manpage doesn't show anything jumping out. Odd. Out of curiosity, what distro are you using?

openSUSE. The `seccomp_rule_add(3+1)` man page is an entirely separate doc and has several paragraphs that are missing from `seccomp_rule_add(3)`.

@justincormack 

> This seems totally the wrong place to do this, runc is (sigh) just doing what it is told and if you have to second guess its behaviour too thats even worse. There are only a few seccomp policies in the world, just fix them.

So switch the default errno to `-ENOSYS` and then add explicit `-EPERM` rules? The issue is that the current implementation of this is a workaround (I would hope it would only exist for a few months at most) -- the long-term goal would be that seccomp profiles indicate what their minimum kernel version is, so that the `-EPERM`/`-ENOSYS` split is not handled by the profile author explicitly.

The other (potentially simpler) solution is to temporarily implement this in BPF ourselves until libseccomp has the minimum kernel feature we need (there are some other features that would be nice like supporting multiple AND rules for the same argument, but they aren't mandatory).

Vagrantfile.centos7
```diff
@@ -12,7 +12,7 @@ Vagrant.configure("2") do |config|
     v.cpus = 2
   end
   config.vm.provision "shell", inline: <<-SHELL
-    set -e -u -o pipefail
+    set -Eeuxo pipefail
```
kolyshkin, line 15: why do you need `-E`?

kolyshkin: or `-x`? This script seems trivial, I don't see a need to have a trace

tests/integration/seccomp.bats
```diff
@@ -0,0 +1,33 @@
...
+	"seccomp_syscall_test1"
+)
+
+function setup() {
+	teardown_busybox
+	setup_busybox
+
+	# Compile the test cases...
+	for test_name in "${SECCOMP_TESTS[@]}"; do
+		gcc -static -o "$SECCOMP_TESTS_BINDIR/$test_name" "$TESTDATA/$test_name.c"
```
kolyshkin, line 16: note that `setup()` is run for _every_ `@test`. Once the second test case is added, this logic (of compiling all *.c files for every test) will look weird.

tests/integration/testdata/.gitignore
```diff
@@ -0,0 +1 @@
+/seccomp_syscall_test1
```
kolyshkin, line 1: as no one puts the binary in here, there's no need for this `.gitingore`.

tests/integration/testdata/seccomp_syscall_test1.c
```diff
@@ -0,0 +1,68 @@
...
+#include <errno.h>
+#include <fcntl.h>
+#include <sched.h>
+#include <sys/syscall.h>
+
+// We need to have a "raw" version because the glibc wrapper hides -ENOSYS from
+// us, which leads to confusing results...
+int raw_renameat2(int olddirfd, const char *oldpath, int newdirfd, const char *newpath, unsigned int flags)
+{
+	return syscall(SYS_renameat2, olddirfd, oldpath, newdirfd, newpath, flags);
```
kolyshkin, line 15: This will probably fail to compile on e.g. centos7.

tests/integration/seccomp.bats
```diff
@@ -0,0 +1,33 @@
...
+function teardown() {
+	teardown_busybox
+	rm -rf "$SECCOMP_TESTS_BINDIR"
+}
+
+@test "runc run [seccomp -ENOSYS handling]" {
+	update_config ".linux.seccomp = $(<"${TESTDATA}/seccomp_syscall_test1.json")"
+	update_config '.process.args = ["/seccomp_syscall_test1"]'
+
+	cp "$SECCOMP_TESTS_BINDIR/seccomp_syscall_test1" rootfs/
```
kolyshkin, line 29: So maybe just compile it right here with `gcc ... -o rootfs/test1` or so

tests/integration/testdata/seccomp_syscall_test1.c
kolyshkin: Indeed it fails:

```
not ok 70 runc run [seccomp -ENOSYS handling]
# (in test file tests/integration/seccomp.bats, line 32)
#   `[ "$status" -eq 0 ]' failed with status 218
# runc spec (status=0):
# 
# runc run test_busybox (status=218):
```

but the CI is green because of an issue fixed by https://github.com/opencontainers/runc/pull/2745 (cc @AkihiroSuda @mrunalp)

Vagrantfile.centos7
cyphar: All of the `yum` commands are done with `-q` so while testing I wasn't sure if the package I requested was actually being installed. I can drop it, but the provisioning commands being really quiet for `Vagrantfile.centos7` confused me when testing.

tests/integration/seccomp.bats
cyphar: Yeah, I'll move compilation to the `@test`.

tests/integration/testdata/.gitignore
cyphar: I did by accident, but okay.

tests/integration/seccomp.bats
cyphar: Will do.

tests/integration/testdata/seccomp_syscall_test1.c
cyphar: This isn't failing to compile -- this is failing during execution but without a loud warning. I mentioned this [yesterday](https://github.com/opencontainers/runc/pull/2737#issuecomment-760999681) but I'm still confused why there's no errors, I'll investigate this on Monday...

cyphar:
@richfelker 

>  If not, can the necessary fixups just be done on the resulting BPF rather than via the deficient libseccomp API?

Leaving aside the way you phrased your point, I just tried my hand at this and it's at least somewhat non-trivial to get right. Because libseccomp doesn't have per-syscall fallback paths everything goes through a final `reta` instruction at the end (except for per-syscall `SECCOMP_RET_ERRNO` rules maybe?) -- the result is that we would need to figure out how to disentangle the different failures when patching. We would also need to do this in an architecture-specific way and we'd still have issues with holes in the syscall table.

All of this is far more complicated than just making the default `-ENOSYS` and returning `-EPERM` for syscalls present in Linux 3.0 but not mentioned in the filter. For the cases where we are returning `-ENOSYS` but we previously returned `-EPERM`, this PR does handle more of those cases but not all of them -- and ideally this hotfix should be as simple as possible without breaking containers.

<hr>

So for now it may just be simplest to do the most basic `-ENOSYS` switch with trivial `-EPERM` handling (return `-EPERM` for any syscall not mentioned in the profile at all but was present in Linux 3.0). I think that this PR will lead to less breakages than the most trivial `-EPERM` handling, but maybe we should see what breaks before we commit to more complicated filter modifications.

After that I can work on cooking up our own BPF generation that handles these cases in the way we would like them to be handled. It is a bit sad we'll have to reinvent most of libseccomp's architecture handling, but we will still use the libseccomp syscall name lookup and architecture information to avoid having to handle all of that ourselves too... And we will need kernel version information anyway (which I will work on implementing in libseccomp).

I do think long-term we could switch back to libseccomp but there are a few other limitations we would ideally like to have fixed in libseccomp (the multiple-rule-for-a-single-argument restriction would be great to remove -- separately from this whole issue).

The slightly positive thing is that we can use the `golang.org/x/net/bpf` emulation code to do unit testing of our generated programs.

cyphar:
#2746 is a hotfix variant of this PR that only does `-EPERM` injection for the "easy" case. I am working on the custom BPF generation separately...

tests/integration/testdata/seccomp_syscall_test1.c
cyphar: I finally figured it out -- it's because I'm missing `return 0;`. The last register value is used instead (`-38` for `-ENOSYS`) which then gets wrapped around to `218`. I guess newer GCCs add an implicit `return 0;` but not the one in CentOS 7.

giuseppe:
> You also can't specify the return value in the default action, so we will need to change runc somehow.

do you think it makes sense to update the runtime specs now to allow overriding the default action?

cyphar:
@giuseppe I assume you meant the default errno for the default action? That's probably not a bad idea to do anyway, but that's not actually enough to solve this problem -- ideally the `ENOSYS` behaviour should be auto-generated by the container runtime because it'd be too complicated for users of runc to generate it (not to mention the whole thing is a footgun).

So we will also need to add either a "minimal kernel version" field to the spec or otherwise define the `-ENOSYS` behaviour we need.

There are also several aspects of runc's seccomp implementation that aren't really in line with the runtime-spec -- such as #2735 (though the runtime-spec's definition of seccomp rules is **woefully** vague so you could argue that it's compliant purely because it's so ill-defined). We need to fix those too.

giuseppe:
> @giuseppe I assume you meant the default errno for the default action? That's probably not a bad idea to do anyway, but that's not actually enough to solve this problem -- ideally the `ENOSYS` behaviour should be auto-generated by the container runtime because it'd be too complicated for users of runc to generate it (not to mention the whole thing is a footgun).

yes, I was referring to the errno for the default action, but perhaps it is not necessary with the approach in https://github.com/opencontainers/runc/pull/2746?  Would exposing `lastSyscall` (and default to `setns` if it is not specified) be better?

cyphar:
@giuseppe A better solution is something more like seccomp/libseccomp#11 -- you specify the minimum kernel version of the profile -- any syscalls newer than that kernel version get `-ENOSYS` rather than `-EPERM`. In principle you don't really need `lastSyscall` (or even a "minimum kernel version") because that information is already present in the profile (the newest syscall in the profile is the latest one) -- but in order for this to be robust we need kernel version information in libseccomp (syscalls aren't always added sequentially to the syscall table).

cyphar:
But I would prefer we don't touch the runtime-spec until libseccomp has the features we need to make a decision about what the right API should be. The current seccomp API was pretty much wholesale copied from the internal runc implementation, which is quite worrying because the runc implementation has weird footguns like #2735 and #1847.

fweimer-rh:
What is the impact of the Linux 3.0 cut-off point? I'm asking because modern glibc assumes at least kernel 3.2 (main differences are `prlimit64` and `accept4` on some architectures, I think). If I understand things correctly, it means that `getrlimit` fails with `EPERM` with the 3.2 limit, and with `ENOSYS` with the 3.0 limit because the implementation calls `prlimit64` internally, unless there is a rule for `prlimit64` in the policy. In theory, `getrlimit` and `pr64limit` with a zero PID and a null pointer for the new limit should be treated identically. But maybe this has to be punted to policy authors.

A complete different concern is who maintains the name-to-syscall mapping. @cyphar You said you wanted to get this from libseccomp. However, downstream, at least we don't treat libseccomp as a package that we need to keep updating continuously for each system call that is added to the upstream (!) kernel. Maybe this is something that we need to change downstream, and I will have to start a discussion about that.

cyphar:
@fweimer-rh Well this is intended as a hotfix, if you think 3.2 is a better cutoff point (the fact it's the minimum version required for glibc is probably the best reason to pick it) we can use that instead. Especially if you think a Linux 3.0 cut-off will break glibc installations. But we don't plan to have this workaround stick around for a very long time -- which is why I've hacked up #2746 as an even simpler implementation of this PR so that we can have *something* merged.

The longer-term goal is to get rid of the hard-coded hotfix and instead either:

 1. Get policy authors to specify what kernel version the profile was written for --effectively indicating what set of syscalls had been audited when creating the profile. Any syscall that existed in the specified kernel version will get the default action (`-EPERM`) but any newer syscalls (syscalls which did not exist in the syscall table in that kernel version) will get `-ENOSYS` to be future-compatible. This is the most "correct" solution but requires users to specify the kernel version.

 2. Auto-detect what kernel version the profile was written for by taking the newest syscall referenced in the profile and using its syscall version as the target kernel version. This requires no additional configuration, but will give the wrong results in some cases.

I think the way we may implement this eventually is to have (1) be the preferred method but if no kernel version is specified we fall back to (2). The reason we need to do this on a kernel version level rather than syscall number level is that the syscall table can have holes in it, and thus any simplistic "just check if the syscall number is larger than X" checks can be wrong -- this happened recently with `close_range` -- it was added after `openat2` despite `openat2` having a larger syscall number.

> A complete different concern is who maintains the name-to-syscall mapping. @cyphar You said you wanted to get this from libseccomp.

We already use libseccomp for this purpose (for two reasons -- there is no real kernel API for it, and we only care about syscall names in the context of seccomp). If we change sources of this information that'd be fine so long as the mapping didn't change for old syscalls.

I agree that this is less than ideal because it requires libseccomp to be updated fairly regularly, but right now these are the tools we have at our disposal (and in practice it hasn't been a huge problem -- the libseccomp maintainers are pretty reactive to new syscall additions). However it should also be noted that for libseccomp's usecase we also need to have the following information which probably won't exist in a hypothetical kernel API:

 1. Syscall numbers for different architectures because libseccomp supports generating a filter that works on more than one architecture. libseccomp already has this information.
 2. In what kernel version a syscall first appeared. This is not currently provided by libseccomp, but is something they are working on (seccomp/libseccomp#11) and will be useful for the long-term solution I described above.

cyphar:
I've thought of a nicer solution to this problem -- inspired by @richfelker, except we **prepend** our `-ENOSYS` handling -- which I'll implement in this PR tomorrow.

cyphar:
#2750 is a cleaner (though still not ideal) solution to this problem. Closing in favour of that implementation.

PR #2741, opened 2021-01-15
ci: untangle getting test images

kolyshkin:
**TL;DR:** this optimizes, simplifies and unifies the process of getting container images
used for tests, resulting in less code, less chances to have broken CI, easier way to
run libcontainer/integration tests, and easier way to support more architectures.

Currently, we have three different ways of getting images:

1. (for hello-world) the image is in this repo under tests/integration/testdata.

2. (for busybox) download it from github (the repo that is used for
   preparing official Docker image) using curl.

3. (for debian) download from Docker hub, using skopeo and umoci.

To further complicate things, we have to do this downloading in multiple
scenarios (at least 4): locally, in github CI, from Dockefile, inside a
Vagrant VM. For each scenario, we have to install skopeo and umoci, and
those two are not yet universally available for all the distros that we
use.

Yet another complication is those images are used for tests/integration
(bats-driven tests) as well as for libcontainer/integration (go tests).
The tests in libcontainer/integration rely on busybox being available
from /busybox, and the bats tests just download the images to a
temporary location during every run.

It is also hard to support CI for other architectures, because all
the machinery for preparing images is so complicated.

This commit is an attempt to simplify and optimize getting images,
mostly by getting rid of skopeo and umoci dependencies, but also
by moving the download logic into one small shell script, which
is used from all the places.

Benefits:

 - images (if not present) are only downloaded once;
 - same images are used for both kind of tests (go and bats);
 - same images are used for local and inside-docker tests
   (because source directory is mounted into container);
 - the download logic is located within 1 simple shell script.

Side benefits:
 - tests won't try to use amd64 image on an unknown architecture;
 - some other arches (like mips64le or s390) might already work
   (except for hello-world but it is trivial to fix);
 - I can finally remove `/busybox` directory from my laptop
   (`cd libcontainer/integration && go test` will download images if needed);
 - I can run any bats tests locally without having to install skopeo/umoci.
   
Caveats:
 - currently the script downloads all (busybox and debian) images at once. Unless we'll have more than 5 images I want to keep it that way (for simplicity);
 - the latest versions of images are downloaded (e.g. those are not pinned to a specific sha). Pinning would require a per-image per-arch sha map, and I'd like to keep it simple while it works.

While at it:
 - make rootless.sh more robust (`set -e -u -o pipefail`) and shellcheck-valid.
   This does not really belong here but I don't want to have too many small PRs.
- use "slim" debian variant (thanks to @tianon)

Previous work related to this:
 - https://github.com/opencontainers/runc/pull/2508
 - https://github.com/opencontainers/runc/pull/2486


libcontainer/integration/utils_test.go
```diff
@@ -19,6 +20,36 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
+var busyboxTar string
+
+// init makes sure the container images are downloaded,
+// and initializes busyboxTar. If images can't be downloaded,
+// we are unable to run any tests, so panic.
+func init() {
```
kolyshkin, line 28: This function looks a tad complicated, but all it does is calls `get-images.sh`
and parses its output to get the value of `BUSYBOX_IMAGE`. This way we
don't have to worry about doing some extra steps before running `go test`.

kolyshkin:
@tianon could you please review this (at least the shell part, i.e. get-images.sh)?

tests/integration/helpers.bash
```diff
@@ -3,25 +3,15 @@
 # Root directory of integration tests.
 INTEGRATION_ROOT=$(dirname "$(readlink -f "$BASH_SOURCE")")
 
-. ${INTEGRATION_ROOT}/multi-arch.bash
+# Download images, get *_IMAGE variables.
+eval $("${INTEGRATION_ROOT}"/get-images.sh)
```
tianon, line 7: `eval` on this output seems kind of weird - why isn't this sourced instead?
(if there's a good reason for doing so, perhaps a comment for posterity would make sense :innocent:)

Additionally, if `get-images.sh` fails, this likely won't fail properly (you'd want something like `GET_IMAGES="$(.../get-images.sh)"; eval "$GET_IMAGES"` instead). :eyes:

tests/integration/get-images.sh
```diff
@@ -0,0 +1,61 @@
...
+# busybox
+BUSYBOX_IMAGE="$TESTDATA/busybox-${arch}.tar.xz"
+get "$BUSYBOX_IMAGE" \
+	"https://github.com/docker-library/busybox/raw/dist-${arch}/stable/glibc/busybox.tar.xz"
+echo "BUSYBOX_IMAGE=$BUSYBOX_IMAGE"
+
+# debian
+DEBIAN_IMAGE="$TESTDATA/debian-${arch}.tar.xz"
+get "$DEBIAN_IMAGE" \
+	"https://github.com/debuerreotype/docker-debian-artifacts/raw/dist-${arch}/buster/rootfs.tar.xz"
```
tianon, line 56: I still think it's a bit dangerous to rely on me not changing these paths again (and would suggest instead embedding an explicit commit reference), but I'm not the one who'll have a failing CI if it changes again so I guess it's your choice? :innocent:

(I certainly can't promise these won't change again. :see_no_evil:)

You might also want to consider grabbing `buster/slim/rootfs.tar.xz` instead if you just need a "basic Debian environment" and aren't too picky about what's in it (since it'll be a smaller download: ~14.6 MB vs ~28.7 MB).

tests/integration/helpers.bash
kolyshkin: This script is also used from two other places (`libcontainer/integration/utils_test.go` and `tests/rootless.sh`) and thus I want it to be executed in the same manner from all the places, thus the eval.

Sourcing it would have a few side effects, like
 - setting shell options (`set -e -u -o pipefail`) -- fix for that is kinda ugly as it requires to use another sub-shell inside the script;
 - setting additional env vars -- fix would be to add `unset XXX` to the end of the script, but when we'll have to maintain that list for unset.
 
Besides, if we source it, someone would spot the same variables (`$TESTDATA` and `$INTEGRATION_ROOT`) being used and will try to optimize it out, which is wrong.

Anyway, it all boils down to -- this is a standalone script. I have emphasized it in the comment at the very beginning.

> Additionally, if get-images.sh fails, this likely won't fail properly (you'd want something like GET_IMAGES="$(.../get-images.sh)"; eval "$GET_IMAGES" instead). ￼

￼Good catch; fixed!

PTAL

tests/integration/get-images.sh
kolyshkin: Nice, let's try it

kolyshkin: works!

kolyshkin:
@AkihiroSuda @mrunalp PTAL

tests/rootless.sh
```diff
@@ -19,12 +19,16 @@
 # a new feature, please match the existing style. Add an entry to $ALL_FEATURES,
 # and add an enable_* and disable_* hook.
 
+set -e -u -o pipefail
+[ -v RUNC_USE_SYSTEMD ] || RUNC_USE_SYSTEMD=
```
AkihiroSuda, line 23: nit: `: ${RUNC_USE_SYSTEMD:=}` is slightly shorter

AkihiroSuda (dismissed):

tests/rootless.sh
kolyshkin: Fixed

kolyshkin:
Rebased; addressed review comment.

cyphar:
I would prefer that we kept using umoci+skopeo for this, since having our own scripts for fetching images (especially tying it to GitHub blob URLs) makes me a little bit nervous, but if this does fix CI issues then I guess it's a net positive. Hopefully umoci works its way into more distributions so we don't have to do this forever.

kolyshkin:
@cyphar note we are only using umoci+skopeo for fetching Debian, busybox is (and was always) done by curl.

I agree using proper tools (rather than relying on some URLs) is better in theory, but practically that this PR does simplifies things a lot from a perspective of CI maintainer, makes the code smaller, and solves a few dangling issues (like inability to run `go test` or `bats tests/integration/whatever.bats` locally without doing some complex setup first) along the way).

Once skopeo and umoci are universally available we'll modify the `tests/integration/get-images.sh` script to use them.

cyphar:
Yeah, agreed. Sorry the above was intended to be an LGTM. Lemme do that now.

cyphar (approved):
LGTM.

libcontainer/integration/utils_test.go
```diff
@@ -19,6 +20,36 @@ import (
...
+	// Figure out path to get-images.sh. Note it won't work
+	// in case the compiled test binary is moved elsewhere.
+	_, ex, _, _ := runtime.Caller(0)
+	getImages, err := filepath.Abs(filepath.Join(filepath.Dir(ex), "..", "..", "tests", "integration", "get-images.sh"))
+	if err != nil {
+		panic(err)
```
AkihiroSuda, line 34: Shouldn't this cause t.Skip rather than panic ?

cyphar: You can't cause `t.Skip` because it's done in `init()`.

AkihiroSuda: We can, by defining a global variable

cyphar: You would want it to be `t.Fatal` because the images are needed for tests (and skipping them would hide test failures), and `t.Fatal` will kill the test anyway. I guess it is slightly nicer to do that than `panic`, but this really shouldn't happen often enough to be an issue (not to mention you'd need to add `t.Fatal` to some helper function and I can't see an obvious candidate in `utils_test.go` -- none of the functions take `*testing.T`).

AkihiroSuda (approved):

PR #2743, opened 2021-01-15
gha: cache docker layers to speed up make runcimage

kolyshkin:
It currently takes close to 2 minutes to build docker image
(which is currently only used for cross compilation).

Let's see if caching docker layers will help.


kolyshkin:
This saves about 30s (from 3m) on `validate / cross` job run (1m50 to create image from scratch vs 1m20 to load layers from cache). I am not sure it's worth it.

kolyshkin:
From the other side, it will definitely worth it if we'll have to use docker image from multiple stages.

mrunalp (approved):

AkihiroSuda (approved):

PR #2744, opened 2021-01-15
gha: don't check commits on push

kolyshkin:
This check (for commit subject being short) only make sense for pull
requests, and it fails on pushes to master (i.e. on merges, see e.g.
https://github.com/opencontainers/runc/runs/1711154129?check_suite_focus=true)

Make sure it's only run for pull requests.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -85,6 +85,8 @@ jobs:
 
   commit:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


kolyshkin:
Alas there's no way to test this :) so let's just merge and see :)

AkihiroSuda (approved):

mrunalp (approved):

PR #2730, opened 2021-01-12
[WIP] ci/gha: move fake term to a separate script

kolyshkin:
Add test execution helper script for github actions. The purpose is to hide
the complexity of running the tests under GHA.

1. GHA execution environment lacks a terminal, and some tests require it.
  Fake the terminal by using script utility with some minimal setup
   (stty and TERM value).

2. Use `sudo -n -E PATH="$PATH"` to run tests as root with proper PATH set.

3. Change to the proper directory (needed for vagrant runs).

PR #2745, opened 2021-01-15
GHA: use ssh -tt instead of script tool (and fix CI failure reporting)

kolyshkin:
Github Actions execution environments lack tty, which is needed for
integration tests. We have used script utility before to provide a fake
tty, but it appears to be buggy on CentOS 7.
    
Since we use ssh anyway to access Vagant VM, add -tt to forcefully
allocate a tty.
    
Apparently, neither "TERM=xterm" nor "stty size" is needed for
the tests to work, so remove those, too.

The benefits of this are:
 - simpler code;
 - more robust error reporting (I saw failed tests but zero exit code on CentOS 7, presumably an issue with script).

(this also carries some minor cleanups from #2730)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,3 +1,9 @@
+# NOTE Github Actions execution environments lack a terminal, needed for
+# some integration tests. Two ways to get a terminal are used below:
+#
+# 1. script utility -- for "local" integration tests;
+# 2. ssh -tt -- for Vagrant VMs (script is buggy on CentOS 7).
+
 name: ci
 on:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -74,19 +74,19 @@ jobs:
         sudo chown -R rootless.rootless /home/rootless
 
     - name: integration test (fs driver)
-      run: sudo -E PATH="$PATH" script -e -c /bin/bash -c 'stty rows 40 cols 80; TERM=xterm make local${{ matrix.rootless }}integration'
+      run: sudo -E PATH="$PATH" script -e -c 'make local${{ matrix.rootless }}integration'
```
AkihiroSuda, line 77: Do we still need `script`?

kolyshkin: In this case yes -- there's no terminal otherwise, and there' no `ssh -tt` to make one (although I was thinking about ssh-ing to locahost for the sake of using `-tt`, but `script` looks more elegant).

Also, in this case script is working fine -- I think the only issue is with the one from CentOS 7.

AkihiroSuda: Please add that to comment lines? Then LGTM

kolyshkin: Yep, I had this comment and then it was removed.

Added.

AkihiroSuda (approved):

mrunalp (approved):

PR #2746, opened 2021-01-17
[hotfix] seccomp: default to -ENOSYS for SECCOMP_RET_ERRNO

### LINKED ISSUE #2151: seccomp filter should return ENOSYS for unknown syscalls

cyphar:
This is a hotfix to make -ENOSYS the default errno if the default action
is SECCOMP_RET_ERRNO. This is necessary because glibc cannot make use of
newer syscalls if we block all unknown syscalls with -EPERM (our old
behaviour).

Unfortunately this is not an ideal solution (syscalls with complicated
rules will now return -ENOSYS rather than -EPERM) but a complete
solution will require far more work -- most likely a reimplementation of
libseccomp to allow us to create custom BPF filters ourselves, or large
changes to libseccomp to better accommodate our requirements -- and thus
this hotfix was written to solve the immediate problem while we work on
a more complete solution.

Fixes #2151


cyphar:
@richfalker Maybe this PR (specifically the test program) better demonstrates why we need some way of either having inverse rules or another way of returning `-EPERM` for intended failures.

cyphar:
@richfelker Maybe this PR (specifically the test program) better demonstrates why we need some way of either having inverse rules or another way of returning `-EPERM` for intended failures.

cyphar:
I've thought of a nicer solution to this problem, which I'll implement in a separate PR tomorrow.

PR #2278, opened 2020-03-30
Exposing memory.numa_stats

iwankgb:
Making information on page usage by type and NUMA node available.

PR #2751, opened 2021-01-19
cgroup: fix panic in parse memory.numa_stat

Ace-Tang:
strings.SplitN not always return N fields if not staify, sometimes
cgroup interface add some custom fields make parse memory.numa_stat
fails, it will case panic

Signed-off-by: acetang <aceapril@126.com>

Fix panic in parse memory.numa_stat, since memory.numa_stat was add some custom fields that make splitN fail
image

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -294,6 +294,9 @@ func getPageUsageByNUMA(cgroupPath string) (cgroups.PageUsageByNUMA, error) {
 
 		for _, column := range columns {
 			pagesByNode := strings.SplitN(column, numaStatKeyValueSeparator, numaStatColumnSliceLength)
+			if len(pagesByNode) != numaStatColumnSliceLength {
+				continue
+			}
 
 			if strings.HasPrefix(pagesByNode[numaStatTypeIndex], numaNodeSymbol) {
 				nodeID, err := strconv.ParseUint(pagesByNode[numaStatTypeIndex][1:], 10, 8)


libcontainer/cgroups/fs/memory.go
```diff
@@ -294,6 +294,9 @@ func getPageUsageByNUMA(cgroupPath string) (cgroups.PageUsageByNUMA, error) {
 
 		for _, column := range columns {
 			pagesByNode := strings.SplitN(column, numaStatKeyValueSeparator, numaStatColumnSliceLength)
+			if len(pagesByNode) != 2 {
```
AkihiroSuda, line 297: nit: s/2/numaStatColumnSliceLength/

Ace-Tang: Update~

AkihiroSuda (approved):

kolyshkin:
@Ace-Tang can you show the contents of `memory.numa_stat` from that system?

For the record, here's kernel doc about memory.numa_stat (from https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt)

```
5.6 numa_stat

This is similar to numa_maps but operates on a per-memcg basis.  This is
useful for providing visibility into the numa locality information within
an memcg since the pages are allowed to be allocated from any physical
node.  One of the use cases is evaluating application performance by
combining this information with the application's CPU allocation.

Each memcg's numa_stat file includes "total", "file", "anon" and "unevictable"
per-node page counts including "hierarchical_<counter>" which sums up all
hierarchical children's values in addition to the memcg's own value.

The output format of memory.numa_stat is:

total=<total pages> N0=<node 0 pages> N1=<node 1 pages> ...
file=<total file pages> N0=<node 0 pages> N1=<node 1 pages> ...
anon=<total anon pages> N0=<node 0 pages> N1=<node 1 pages> ...
unevictable=<total anon pages> N0=<node 0 pages> N1=<node 1 pages> ...
hierarchical_<counter>=<counter pages> N0=<node 0 pages> N1=<node 1 pages> ...

The "total" count is sum of file + anon + unevictable.
```

The code was introduced by #2278, @iwankgb PTAL

kolyshkin:
On my (non-NUMA) system it's always 2 fields:
`find /sys/fs/cgroup/ -name memory.numa_stat | xargs awk '{print NF}' | grep -v ^2$`

dqminh (changes requested):
The check is reasonable. However, it would be nice if we can add a failing test.

Ace-Tang:
The memory.numa_stat interface changes was add by other guys, but it was used online, althrough the interface develop not follow  kernel doc, fields is 
image

@kolyshkin @dqminh

kolyshkin:
Ah, OK, I was not able to find these kernel patches in the wild -- the most similar one is "sched/numa: introduce per-cgroup NUMA locality info" from Alibaba's Michael Wang (which also adds some non-standard fields to `memory.numa_stat`). Another showcase of why text-based interfaces are not that good...

I took a look at the code and found that if someone ever adds some line that will contain `=` to `memory.numa_stat` file (e.g. `a=b`), the whole `GetStats` will return an error. Maybe it makes sense to make this code less fragile / more forgiving to unknown fields while we're at it. I'll look into it.

In the meantime, LGTM

kolyshkin (approved):
LGTM

kolyshkin:
@dqminh can we merge this hotfix as it is? I'm working on more robust parser and will add more tests.

kolyshkin:
> @dqminh can we merge this hotfix as it is? I'm working on more robust parser and will add more tests.

Alternatively, PTAL https://github.com/opencontainers/runc/pull/2755 which has this fix, and more fixes and tests.

dqminh:
@kolyshkin lets go with #2755 . Thanks @Ace-Tang very much for your fix !

Ace-Tang:
Thanks

PR #2755, opened 2021-01-20
libct/cg/fs: getPageUsageByNUMA: rewrite/optimize, fix panic, add more tests

kolyshkin:
This is a rewrite of getPageUsageByNUMA, fixing a number of issues.

0. Fix a panic for lines that do not contain `=` (first commit, by @Ace-Tang).

1. Be less strict to unknown contents, i.e. skip it. This makes the
       function more future-proof. Before this commit, if a line like
       "a=b" is encountered, the function returns an error, which is
       propagated all the way up to and returned by (CgroupManager).GetStats.
    
2. Be more strict to contents it recognizes, i.e. return an error.
       In case the first field in the line is recognized (e.g. "total=123",
       the rest of the line should be in format "N<id>=<value> ...".
    
3. Optimize. Before this commit, addNUMAStatsByType was called for every
       item in the line, which is excessive and might even be slow in case
       there are many NUMA nodes. It is enough to look up the field once.
       This probably helps to reduce the load on garbage collector, too.
    
4. Remove a bunch of global numaNode* and numaStat* constants. Those
       were used by only one function, and it does not make sense to have
       them defined globally. Some were moved to the function, some were
       eliminated entirely.
    
5. Improve readability and added code comments.
    
Finally, add some test cases for good and bad contents.

NOTE this PR includes #2751.

Closes: #2751

### DIFF
--- a/libcontainer/cgroups/fs/memory.go
+++ b/libcontainer/cgroups/fs/memory.go
@@ -17,16 +17,8 @@ import (
 )
 
 const (
-	numaNodeSymbol            = "N"
-	numaStatColumnSeparator   = " "
-	numaStatKeyValueSeparator = "="
-	numaStatMaxColumns        = math.MaxUint8 + 1
-	numaStatValueIndex        = 1
-	numaStatTypeIndex         = 0
-	numaStatColumnSliceLength = 2
-	cgroupMemorySwapLimit     = "memory.memsw.limit_in_bytes"
-	cgroupMemoryLimit         = "memory.limit_in_bytes"
-	cgroupMemoryPagesByNuma   = "memory.numa_stat"
+	cgroupMemorySwapLimit = "memory.memsw.limit_in_bytes"
+	cgroupMemoryLimit     = "memory.limit_in_bytes"
 )
 
 type MemoryGroup struct {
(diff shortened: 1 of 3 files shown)


kolyshkin:
CI failure is unrelated (flaky nsexec debug logging, see https://github.com/opencontainers/runc/issues/2756). Restarted.

libcontainer/cgroups/fs/memory.go
```diff
@@ -277,47 +269,81 @@ func getMemoryData(path, name string) (cgroups.MemoryData, error) {
...
+			//   numa_exectime 0
+			if len(byNode) < 2 {
+				if i == 0 {
+					// Ignore/skip those.
+					continue nextline
+				} else {
+					// first item was already validated,
+					// so be strict to the rest
+					return stats, fmt.Errorf("malformed line %q in %s",
+						scanner.Text(), filename)
```
dqminh, line 313: let's captured the line before and reuse here, scanner.Text() effectively convert []byte to string, which afaik is a little costly, even though it doesn't matter that much in practice because this is error case.

dqminh (approved):

libcontainer/cgroups/fs/memory.go
kolyshkin: Right, I thought about it and figured out optimizing the error case makes little sense.

Anyway, fixed, and I also made a few tiny improvements (better comments, use `break` instead of `continue nextline`). Please re-lgtm :)

dqminh (approved):

AkihiroSuda (approved):

PR #2758, opened 2021-01-21
libct/cg: demote "systemd is too old" to debug

kolyshkin:
A recent commit a35cad3b225f8d (PR #2727) added warnings about systemd being
too old. While those warnings are valid, they break some existing tests (e.g. https://github.com/cri-o/cri-o/issues/4509),
and also don't add much value to a user (IOW no one is going to upgrade
systemd because runc says it's old).

Demote those to warnings.


kolyshkin:
This is a hotfix for #2727. @opencontainers/runc-maintainers PTAL

AkihiroSuda (approved):

mrunalp (approved):

Issue #2760, opened 2021-01-22
[centos7] not ok 15 checkpoint --lazy-pages and restore


kolyshkin:
~~After _accidentally_ enabling checkpoint/restore tests on CentOS 7 (in https://github.com/opencontainers/runc/pull/2757/commits/a15ebe5a48fcead0ab050f08071c066c27652db9, PR #2757), I found out that~~
lazy-pages test **sometimes** fails without providing any decent error:

```
not ok 15 checkpoint --lazy-pages and restore
# (in test file tests/integration/checkpoint.bats, line 193)
#   `[ "$out" = "0000000 000000 0000001" ]' failed
# runc spec (status=0):
# 
# uffd-noncoop is supported
# runc state test_busybox (status=0):
# {
#   "ociVersion": "1.0.2-dev",
#   "id": "test_busybox",
#   "pid": 9727,
#   "status": "running",
#   "bundle": "/tmp/bats-run-7505/busyboxtest",
#   "rootfs": "/tmp/bats-run-7505/busyboxtest/rootfs",
#   "created": "2021-01-22T01:12:07.588049056Z",
#   "owner": ""
# }
# time="2021-01-22T01:12:09Z" level=error msg="container is not destroyed"
# time="2021-01-22T01:12:09Z" level=error msg="criu failed: type NOTIFY errno 0\nlog file: work-dir/dump.log"
```

The command that fails is:
```
        __runc --criu "$CRIU" checkpoint --lazy-pages --page-server 0.0.0.0:${port} --status-fd ${lazy_w} --work-path ./work-dir --image-path ./image-dir test_busybox &
```

kolyshkin:
TODO: make it run multiple times, show full log, cross my fingers and hope it will fail.

kolyshkin:
Got another failure ( https://github.com/opencontainers/runc/pull/2087/checks?check_run_id=1841731496), this time with slightly different output:

```
not ok 15 checkpoint --lazy-pages and restore
# (in test file tests/integration/checkpoint.bats, line 193)
#   `[ "$out" = "0000000 000000 0000001" ]' failed
# runc spec (status=0):
# 
# uffd-noncoop is supported
# runc state test_busybox (status=0):
# {
#   "ociVersion": "1.0.2-dev".
#   "id": "test_busybox",
#   "pid": 23602,
#   "status": "running",
#   "bundle": "/tmp/busyboxtest",
#   "rootfs": "/tmp/busyboxtest/rootfs",
[...]

PR #2761, opened 2021-01-22
[DNM] debugging criu --lazy-pages failure

kolyshkin:
Looking into https://github.com/opencontainers/runc/issues/2760


kolyshkin:
Hmm, it's working now even on CentOS. Restarting...

kolyshkin:
TODO: make it run multiple times, show full log, cross my fingers and hope it will fail.

kolyshkin:
OK it fails on CentOS at `-- iteration 47 of 400 --`

kolyshkin:
The analysis were inconclusive :( but we do not see these failures recently :) so closing

PR #2764, opened 2021-01-24
test: unify debian and busybox image setup

zhsj:
Like busybox image, download debian rootfs from GitHub.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,13 +29,10 @@ jobs:
 
     - name: install deps
       run: |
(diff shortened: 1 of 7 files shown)


AkihiroSuda (approved):
LGTM.

We should error out when `$(go env GOARCH) != amd64 && $(go env GOARCH) != amd64`, but the issue is not new in this PR and can be fixed later.

tianon:
Lots of overlap here with #2741 :sweat_smile:

zhsj:
close in favor of #2741

Issue #2753, opened 2021-01-19
Runc init process step on DISK Sleep Status When freezen cgroup


gaopeiliang:
we upgrade docker-ce version to 20:10 from 19.3,  And now docker component version:

```
Client: Docker Engine - Community
 Version:           20.10.1
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        831ebea
 Built:             Tue Dec 15 04:35:01 2020
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.1
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       f001486
  Built:            Tue Dec 15 04:32:57 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

when Kubernetes terminal POD , we found container process step into DISK Sleep status ,,,,  docker inspect container  hang forever ,,    kubelet pleg not healthy ,  kubelet not ready ......

when we ps container process .... we found  runc init process also DISK Sleep ....

```
root     15758 57587  0 16:39 ?        00:00:00 runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/452a153f1775d37f5813c1b71ff46df6e0da76ce5e8450d3df46d1de3cc97b2c/log.json --log-format json --systemd-cgroup exec --process /tmp/runc-process627966115 --detach --pid-file /run/containerd/io.containerd.runtime.v2.task/moby/452a153f1775d37f5813c1b71ff46df6e0da76ce5e8450d3df46d1de3cc97b2c/27e5850550ae06545bde2abb0470316e95938c54ad673bb9b1e0686129d10ae1.pid 452a153f1775d37f5813c1b71ff46df6e0da76ce5e8450d3df46d1de3cc97b2c
root     15795 15758  0 16:39 ?        00:00:00 runc init
[...]

PR #2767, opened 2021-01-26
add timeout set freeze to fix exec with update make set freeze step o…

gaopeiliang:
sometimes we found runc init  and container process step on Disk Sleep status when use runc-rc92 version

we found runc exec run with runc update concurrence，will make cgroup freezing incomplete,  we should retry set freezer status or cancel set  not block forever as kernel freezer-subsystem.txt said

even though kernel freezing status will not step on more time , but we found it forever on Ubuntu 16.04, kernel 4.19 

when unix execv(2) replace self and write cgroup freezen make cgroup step on freezing forever ,, maybe kernel something ,, but we should around it , not stop the world .....

### DIFF
--- a/libcontainer/cgroups/fs/freezer.go
+++ b/libcontainer/cgroups/fs/freezer.go
@@ -15,6 +15,8 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var FreezerTimeout = errors.New("set freezer status timeout")
+
 type FreezerGroup struct {
 }
 
@@ -29,6 +31,7 @@ func (s *FreezerGroup) Apply(path string, d *cgroupData) error {
 func (s *FreezerGroup) Set(path string, cgroup *configs.Cgroup) error {
 	switch cgroup.Resources.Freezer {
 	case configs.Frozen, configs.Thawed:
+		maxRetry := 5
 		for {
 			// In case this loop does not exit because it doesn't get the expected
 			// state, let's write again this state, hoping it's going to be properly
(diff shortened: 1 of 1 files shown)


kolyshkin:
Thank you for your contribution @gaopeiliang. Can you please describe the issue you're trying to fix in more details?

kolyshkin:
> Can you please describe the issue you're trying to fix in more details?

Ideally, a code / test case that demonstrates it.

thaJeztah:
related to / addresses https://github.com/opencontainers/runc/issues/2753

kolyshkin:
> we should retry set freezer status or cancel set not block forever as kernel freezer-subsystem.txt said

Here's the relevant portion:

<blockquote>
It's important to note that freezing can be incomplete. In that case we return
EBUSY. This means that some tasks in the cgroup are busy doing something that
prevents us from completely freezing the cgroup at this time. After EBUSY,
the cgroup will remain partially frozen -- reflected by freezer.state reporting
"FREEZING" when read. The state will remain "FREEZING" until one of these
things happens:

	1) Userspace cancels the freezing operation by writing "THAWED" to
		the freezer.state file
	2) Userspace retries the freezing operation by writing "FROZEN" to
		the freezer.state file (writing "FREEZING" is not legal
		and returns EINVAL)
	3) The tasks that blocked the cgroup from entering the "FROZEN"
		state disappear from the cgroup's set of tasks.
</blockquote>

kolyshkin:
@gaopeiliang I'm still working on that stuff, no need to replace your implementation with mine. Once ready I'll open an PR.

gaopeiliang:
> @gaopeiliang I'm still working on that stuff, no need to replace your implementation with mine. Once ready I'll open an PR.

OK ,,  maybe it will come soon ,,,   we have about 2k node in risk status ......

gaopeiliang:
https://github.com/opencontainers/runc/pull/2774

PR #2675, opened 2020-11-09
update vendor

AkihiroSuda:
- https://github.com/cilium/ebpf/compare/1c8d4c9ef775...v0.1.0
- https://github.com/golang/protobuf/compare/v1.4.2...v1.4.3
- https://github.com/opencontainers/runtime-spec/compare/f9c09b4ea1df...e6143ca7d51d
- https://github.com/golang/sys/compare/2334cc1a136f...4d91cf3a1aaf

PR #2769, opened 2021-01-26
ci: test with golang 1.16-rc1

kolyshkin:
Using beta version of go as per https://github.com/actions/setup-go/issues/92

[v2: move from beta1 to rc1]

~~Includes #2660 to fix #2675.~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

AkihiroSuda (approved):

cyphar (approved):
LGTM, though it would be nice if we added `1.x` and `tip` to our tests so we don't need to keep manually updating the GHA.

Issue #2728, opened 2021-01-08
ci/gha: make the artefacts from make release available


kolyshkin:
We currently have a ci job that builds a static binary. It would be beneficial to have the binary (tarball) published somewhere so that whoever looks at PR can quickly test the result.

tianon:
I've played with https://github.com/actions/upload-artifact and it's not awful, but IMO the Artifacts support in Actions leaves a lot to be desired; for example, downloading the artifacts requires you to be authenticated, which is fine when you click download from a browser, but makes it kind of annoying to try to download from GitHub Actions in a separate repo to do cross-repo testing (for which I ended up using https://github.com/dawidd6/action-download-artifact).

Issue #2756, opened 2021-01-20
nsexec logging is racy (aka "debug.bats is flaky")


kolyshkin:
Sometimes debug.bats test cases are failing like this:

```
not ok 27 global --debug to --log --log-format 'json'
# (in test file tests/integration/debug.bats, line 77)
#   `[[ "${output}" == *"child process in init()"* ]]' failed
```

This is caused by the fact that log forwarding from nsexec is not waiting for all logs to be written to disk ( before runc process ends, and in case of short-lived processes this result in missing debug from nsexec.

This happens when writing to disk, and may probably be made 100% reproducible by slowing down disk I/O.

This is probably not an issue for any real customers (no one except developers is interested in nsexec debug log), but it's at least an issue with the test (thus the subject).

~~I tried to fix it in https://github.com/opencontainers/runc/pull/2487 (last commit) but it needs more work.~~

PR #2770, opened 2021-01-27
ci: make static binary available for every PR

kolyshkin:
This uploads the results of `make release` step (static binary and a source tarball) so that they are available. I am not very sure if it's of any use, but at least one can download and play with a static binary from any PR.

The result is available at Checks -> validate (scroll to the bottom of the page), or from Actions -> Validate -> (any action)

Fixes: https://github.com/opencontainers/runc/issues/2728

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -156,3 +156,8 @@ jobs:
         sudo apt -qq install libseccomp-dev
     - name: make release
       run: make release
+    - name: upload artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: release-${{ github.run_id }}
+        path: release/*
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -153,7 +153,7 @@ jobs:
     - name: install deps
       run: |
         sudo apt -qq update
-        sudo apt -qq install libseccomp-dev
+        sudo apt -qq install gperf
     - name: make release
       run: make release
     - name: upload artifacts
--- a/script/release.sh
+++ b/script/release.sh
@@ -24,9 +24,46 @@ root="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
 # This function takes an output path as an argument, where the built
 # (preferably static) binary should be placed.
 function build_project() {
+	local builddir
 	builddir="$(dirname "$1")"
 
-	make -C "$root" COMMIT_NO= static
+	# Due to libseccomp being LGPL we must include its sources,
+	# so download, install and build against it.
+
+	# Sanity check: check that the build won't use distro-provided
+	# libseccomp.a. If this is the case, explain and abort.
+	#
+	# Note go is not used here as it caches the build heavily,
+	# including, apparently, the results of pkg-config.
+	# shellcheck disable=SC2046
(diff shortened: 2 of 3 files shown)


.github/workflows/validate.yml
```diff
@@ -138,3 +138,8 @@ jobs:
       run: make validate
     - name: make release
       run: make release
+    - name: upload release
```
AkihiroSuda, line 141: I would call this as "artifact" rather than "release"

kolyshkin: fixed

AkihiroSuda (dismissed):

cyphar:
My only concern is whether we're going to run into issues with libseccomp LGPL requirements -- at the moment I manually resolve them when doing releases but I was thinking of automating the procedure a little bit (by downloading, building, and installing libseccomp as part of `make release`).

kolyshkin:
I amended script/release.sh to download/install/use/bundle libseccomp. Despite my best efforts this does not look super neat, but it works.

An alternative would be to build a static binary without seccomp.

kolyshkin:
CI using go 1.16-rc1 is currently broken; fix is in #2808

kolyshkin:
Rebased to include #2808 to fix CI. @cyphar @AkihiroSuda @mrunalp PTAL

script/release.sh
```diff
@@ -19,14 +19,51 @@ set -e
...
+		echo "Unable to build a static binary against own libsecomp." >&2
+		echo "Please uninstall libseccomp-static to fix." >&2
+		exit 1
+	fi
+
+	local libseccomp_ver='2.5.1'
+	local tarball="libseccomp-${libseccomp_ver}.tar.gz"
+	local prefix
+	prefix="$(mktemp -d)"
+	curl -O -L "https://github.com/seccomp/libseccomp/releases/download/v${libseccomp_ver}/${tarball}"
```
cyphar, line 56: In order make my life even easier when I do releases, can you also download the `.asc` signature and then include both the tarball and the detached signature in the release directory so that they're part of the signed `sha256sum` and I can more easily upload the release artefacts?

AkihiroSuda (approved):

script/release.sh
kolyshkin: Thought about it but wasn't sure.

Added.

kolyshkin:
CI failure is a known flake (https://github.com/opencontainers/runc/issues/2756), CI restarted

kolyshkin:
@cyphar I have addressed your comment (https://github.com/opencontainers/runc/pull/2770#discussion_r579321045), PTAL

kolyshkin:
close/reopen to kick ci

AkihiroSuda (approved):

AkihiroSuda:
@cyphar Can we have this in rc94?

script/release.sh
```diff
@@ -121,12 +158,13 @@ git archive --format=tar --prefix="$project-$version/" "$commit" | xz >"$release
 )
```
cyphar, line 158: Nit: It would be nice if the checksum also included the libseccomp artefacts. But I'll do that in a separate PR.

cyphar (approved):
LGTM. This doesn't appear to break my local `make release` flow and it finally makes it so that we have the version of libseccomp used to build runc be part of the repo. Thanks!

PR #2773, opened 2021-01-29
Makefile: add -trimpath go build flag

aisk:
`-trimpath` is added in Go 1.13 which matched runc's minimal go version in readme, and this option will remove the filesystem's path in build binary's symbol, which will reduce the binary size a little, and can protect the privary for builders.

https://golang.org/doc/go1.13#go-command


AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #2775, opened 2021-02-01
cgroupfs: cpuset: fix broken build

cyphar:
The merge 6eed6e579596 broke the build because ab27e12cebf1 ("Implement
GetStat for cpuset cgroup.") dropped the errors import which was used by
c85cd2b3254a ("libct/cg/fs/cpuset: don't parse mountinfo") and the CI
wasn't retriggered.

Fix this by just importing "github.com/pkg/errors" again.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

PR #2776, opened 2021-02-01
libcontainer: the fstype argument is ignored for bind mount

Iceber:
`"bind"` is confusing as a `fstype` argument.
```go
    unix.Mount(node.Path, dest, "bind", unix.MS_BIND, "")
```

And the `fstype` argument is ignored for bind mount.
```go
 func Mount(source string, target string, fstype string, flags uintptr, data string) (err error) 
```

### DIFF
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -278,7 +278,7 @@ func TestValidateSysctlWithBindHostNetNS(t *testing.T) {
 	defer os.Remove(file)
 	fd.Close()
 
-	if err := unix.Mount(selfnet, file, "bind", unix.MS_BIND, ""); err != nil {
+	if err := unix.Mount(selfnet, file, "", unix.MS_BIND, ""); err != nil {
 		t.Fatalf("can't bind-mount %s to %s: %s", selfnet, file, err)
 	}
 	defer func() {
--- a/libcontainer/console_linux.go
+++ b/libcontainer/console_linux.go
@@ -18,7 +18,7 @@ func mountConsole(slavePath string) error {
 	if f != nil {
 		f.Close()
 	}
-	return unix.Mount(slavePath, "/dev/console", "bind", unix.MS_BIND, "")
+	return unix.Mount(slavePath, "/dev/console", "", unix.MS_BIND, "")
 }
 
 // dupStdio opens the slavePath for the console and dups the fds to the current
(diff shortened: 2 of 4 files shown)


Iceber:
need to merge https://github.com/opencontainers/runc/pull/2775

kolyshkin:
It is already `""` in some places -- and I agree we should convert the test.

Nit: typo in commit subject: `s/amount/mount/`

kolyshkin:
@Iceber can you please rebase to fix CI?

AkihiroSuda:
From my experience, Kubelet can panic when mountinfo has "" entry. (Bug on kubelet side)

Iceber:
@kolyshkin use `grep unix.Mount . -r | grep bind` to find and modify other files

kolyshkin (approved):
LGTM

mrunalp:
This one gives me pause. I think we had some edge cases in the past where this was needed maybe for mount propagation or remounts.

cyphar:
I don't think so @mrunalp  -- `fstype` has always been ignored for all of the key flags (`MS_BIND`, `MS_MOVE`, `MS_REMOUNT` ...). The issue we had in the past was that internally libcontainer requires you to specify `bind` or `rbind` as types (in other words, the `bind` option didn't actually do anything if you didn't also specify the type).

I'm fine with dropping it, though I don't personally think it's confusing to have it -- it makes `strace(1)` output slightly nicer to read.

cyphar (approved):
LGTM, but I'll give @mrunalp a chance to take another look.

AkihiroSuda:
Better to move to v1.1.0 milestone?

cyphar:
I think it's trivial enough it doesn't need to be blocked on 1.0, but I don't really mind (it's a no-op change anyway).

kolyshkin:
Agree we can do it post 1.0

kolyshkin:
This is indeed a no-op; the only difference is in strace output.

BTW I just checked and `mount` binary from util-linux 2.40.1 doesn't even use `mount(2)` system call. Instead, it does this:

```console
# strace -f mount --bind -o ro /tmp/src /tmp/dst
...
open_tree(AT_FDCWD, "/tmp/src", OPEN_TREE_CLONE|OPEN_TREE_CLOEXEC) = 3
...
mount_setattr(3, "", AT_EMPTY_PATH, {attr_set=MOUNT_ATTR_RDONLY, attr_clr=0, propagation=0 /* MS_??? */, userns_fd=0}, 32) = 0
move_mount(3, "", AT_FDCWD, "/tmp/dst", MOVE_MOUNT_F_EMPTY_PATH) = 0
close(3) = 0
```

PR #2777, opened 2021-02-01
fix missing import in libcontainer/cgroups/fs

thaJeztah:
probably caused by PR's being merged out of order


thaJeztah:
oh! dup of https://github.com/opencontainers/runc/pull/2775

Issue #2778, opened 2021-02-01
libcontainer/seccomp/patchbpf: build fails without cgo


thaJeztah:
Didn't look too closely, but noticed this while trying to run golang-ci-lint locally;

```bash
docker run --rm -v $(pwd):/go/src/github.com/opencontainers/runc -w /go/src/github.com/opencontainers/runc golangci/golangci-lint:v1.35.2 golangci-lint run -v
```

...

```
level=warning msg="[runner] Can't run linter goanalysis_metalinter: S1018: failed prerequisites: 
[(inspect@github.com/opencontainers/runc/libcontainer/seccomp/patchbpf, isgenerated@github.com/opencontainers/runc/libcontainer/seccomp/patchbpf): analysis skipped:
errors in package: [/go/src/github.com/opencontainers/runc/libcontainer/seccomp/patchbpf/enosys_unsupported.go:10:13:
could not import github.com/seccomp/libseccomp-golang (/go/src/github.com/opencontainers/runc/vendor/github.com/seccomp/libseccomp-golang/seccomp.go:26:8:
could not import C (cgo preprocessing failed))]]"
```

https://github.com/opencontainers/runc/blob/c531a6f26aa2ee5ccf89b88d12b59d6152127ce3/libcontainer/seccomp/patchbpf/enosys_unsupported.go#L10-L13

thaJeztah:
@cyphar @AkihiroSuda ptal

cyphar:
Yeah this is problem I solved in CI by requiring `libseccomp-devel`. It seems that golang-ci can't handle cgo projects? I'm not sure how we could work around this because we need cgo in order to implement `patchbpf`...

AkihiroSuda:
A quick workaround would be to just replace `*libseccomp.ScmpFilter` with `interface{}`, though it may look weird.

thaJeztah:
>  I'm not sure how we could work around this because we need cgo in order to implement patchbpf...

Yup; the issue in this case is that it's in the `_unsupported.go` file, so in situations where seccomp or cgo are not enabled / supported.

> A quick workaround would be to just replace *libseccomp.ScmpFilter with interface{}, though it may look weird.

Perhaps not that bad it we add a comment; this file is just a "stub" if its not supported, so probably not too horrible.

thaJeztah:
[...]

PR #2780, opened 2021-02-01
Remove "PatchAndLoad" stub as it's not used without seccomp enabled

thaJeztah:
This function is called by `InitSeccomp`, but only when compiled with seccomp (and cgo) enabled, so should not be needed for other situations.

fixes https://github.com/opencontainers/runc/issues/2778

### DIFF
--- a/libcontainer/seccomp/patchbpf/enosys_unsupported.go
+++ b/libcontainer/seccomp/patchbpf/enosys_unsupported.go
@@ -1,18 +1,3 @@
 // +build !linux !cgo !seccomp
 
 package patchbpf
-
-import (
-	"errors"
-
-	"github.com/opencontainers/runc/libcontainer/configs"
-
-	libseccomp "github.com/seccomp/libseccomp-golang"
-)
-
-func PatchAndLoad(config *configs.Seccomp, filter *libseccomp.ScmpFilter) error {
-	if config != nil {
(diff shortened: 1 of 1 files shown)


libcontainer/seccomp/patchbpf/enosys_unsupported.go
```diff
@@ -1,18 +1,3 @@
 // +build !linux !cgo !seccomp
 
 package patchbpf
```
thaJeztah, line 3: perhaps file can be removed altogether (trying to avoid an empty package for these build tags)

cyphar:
Needs to be rebased to fix CI.

cyphar:
We can drop it, but I would keep the file to avoid empty package errors.

thaJeztah:
> but I would keep the file to avoid empty package errors.

👍 agreed; kept it as-is then

rebased

kolyshkin (approved):
LGTM

libcontainer/seccomp/patchbpf/enosys_unsupported.go
kolyshkin: Maybe add a comment saying "we are keeping this file to avoid empty package errors"?

cyphar:
Ah, with this change you can drop the `libseccomp-devel` installation from the GHA for golang-ci!

cyphar (approved):
LGTM.

PR #2781, opened 2021-02-01
Errcheck linting fixes

thaJeztah:
Split it to to smaller commit for easier review, but happy to squash if wanted

addresses part of https://github.com/opencontainers/runc/issues/2627

### DIFF
--- a/libcontainer/configs/config_test.go
+++ b/libcontainer/configs/config_test.go
@@ -145,12 +145,15 @@ func TestFuncHookRun(t *testing.T) {
 
 	fHook := configs.NewFunctionHook(func(s *specs.State) error {
 		if !reflect.DeepEqual(state, s) {
-			t.Errorf("Expected state %+v to equal %+v", state, s)
+			return fmt.Errorf("expected state %+v to equal %+v", state, s)
 		}
 		return nil
 	})
 
-	fHook.Run(state)
+	err := fHook.Run(state)
+	if err != nil {
+		t.Fatal(err)
+	}
 }
 
 func TestCommandHookRun(t *testing.T) {
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1352,7 +1352,7 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 	if err != nil {
 		return err
 	}
-	defer unix.Unmount(root, unix.MNT_DETACH)
+	defer unix.Unmount(root, unix.MNT_DETACH) //nolint: errcheck
 	t := criurpc.CriuReqType_RESTORE
 	req := &criurpc.CriuReq{
 		Type: &t,
@@ -1665,7 +1665,7 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 		break
 	}
 
-	criuClientCon.CloseWrite()
+	_ = criuClientCon.CloseWrite()
 	// cmd.Wait() waits cmd.goroutines which are used for proxying file descriptors.
 	// Here we want to wait only the CRIU process.
 	criuProcessState, err = criuProcess.Wait()
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -196,7 +196,11 @@ func New(root string, options ...func(*LinuxFactory) error) (Factory, error) {
 		Validator: validate.New(),
 		CriuPath:  "criu",
 	}
-	Cgroupfs(l)
+
+	if err := Cgroupfs(l); err != nil {
+		return nil, err
+	}
+
 	for _, opt := range options {
 		if opt == nil {
 			continue
--- a/libcontainer/factory_linux_test.go
+++ b/libcontainer/factory_linux_test.go
@@ -122,7 +122,10 @@ func TestFactoryNewTmpfs(t *testing.T) {
 	if m.Source != "tmpfs" {
 		t.Fatalf("Source of root: %s, expected %s", m.Source, "tmpfs")
 	}
-	unix.Unmount(root, unix.MNT_DETACH)
+	err = unix.Unmount(root, unix.MNT_DETACH)
+	if err != nil {
+		t.Error("failed to unmount root:", err)
+	}
 }
 
 func TestFactoryLoadNotExists(t *testing.T) {
@@ -130,7 +133,7 @@ func TestFactoryLoadNotExists(t *testing.T) {
 	if rerr != nil {
 		t.Fatal(rerr)
 	}
-	defer os.RemoveAll(root)
+	defer os.RemoveAll(root) //nolint: errcheck
 	factory, err := New(root, Cgroupfs)
 	if err != nil {
 		t.Fatal(err)
@@ -153,7 +156,7 @@ func TestFactoryLoadContainer(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer os.RemoveAll(root)
+	defer os.RemoveAll(root) //nolint: errcheck
 	// setup default container config and state for mocking
 	var (
 		id            = "1"
@@ -219,7 +222,7 @@ func marshal(path string, v interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
+	defer f.Close() //nolint: errcheck
 	return utils.WriteJSON(f, v)
 }
 
--- a/libcontainer/integration/checkpoint_test.go
+++ b/libcontainer/integration/checkpoint_test.go
@@ -14,16 +14,16 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func showFile(t *testing.T, fname string) error {
+func showFile(t *testing.T, fname string) {
 	t.Helper()
 	t.Logf("=== %s ===\n", fname)
 
 	f, err := os.Open(fname)
 	if err != nil {
 		t.Log(err)
-		return err
+		return
 	}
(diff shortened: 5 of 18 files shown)


libcontainer/factory_linux_test.go
```diff
@@ -122,15 +122,18 @@ func TestFactoryNewTmpfs(t *testing.T) {
 	if m.Source != "tmpfs" {
 		t.Fatalf("Source of root: %s, expected %s", m.Source, "tmpfs")
 	}
-	unix.Unmount(root, unix.MNT_DETACH)
+	err = unix.Unmount(root, unix.MNT_DETACH)
+	if err != nil {
+		t.Error("failed to unmount root", err)
+	}
 }
```
thaJeztah, line 129: Could possibly ignore this one, unless we're concerned about things not being unmounted

tty.go
```diff
@@ -175,19 +175,19 @@ func (t *tty) ClosePostStart() error {
 func (t *tty) Close() error {
```
thaJeztah, line 175: Note that this function _never_ returns an error; wondering if it should either not have the error return, or if one (or more) of the errors should actually be returned (or a multi-error used)

thaJeztah:
@kolyshkin @AkihiroSuda ptal

libcontainer/integration/checkpoint_test.go
```diff
@@ -136,7 +136,7 @@ func testCheckpoint(t *testing.T, userns bool) {
 	preDumpLog := filepath.Join(preDumpOpts.WorkDirectory, "dump.log")
 
 	if err := container.Checkpoint(preDumpOpts); err != nil {
-		showFile(t, preDumpLog)
+		_ = showFile(t, preDumpLog)
```
kolyshkin, line 139: Since no caller of `showFile()` checks for any errors returned by it, I think it's better to change the function itself to return no error.

libcontainer/integration/exec_test.go
```diff
@@ -1446,18 +1446,18 @@ func TestRootfsPropagationSlaveMount(t *testing.T) {
...
 	// Create mnt1host/mnt2host and bind mount itself on top of it. This
 	// should be visible in container.
 	dir2host, err := ioutil.TempDir(dir1host, "mnt2host")
 	ok(t, err)
-	defer os.RemoveAll(dir2host)
+	defer remove(dir2host)
 
 	err = unix.Mount(dir2host, dir2host, "bind", unix.MS_BIND, "")
-	defer unmountOp(dir2host)
+	defer func() { _ = unmountOp(dir2host) }()
```
kolyshkin, line 1460: same here -- since no caller of unmountOp() cares about its return, it's better to change it to not return anything.

libcontainer/configs/config_test.go
```diff
@@ -145,12 +145,15 @@ func TestFuncHookRun(t *testing.T) {
...
-			t.Errorf("Expected state %+v to equal %+v", state, s)
+			return fmt.Errorf("expected state %+v to equal %+v", state, s)
 		}
 		return nil
 	})
 
-	fHook.Run(state)
+	err := fHook.Run(state)
+	if err != nil {
+		t.Fatal(err)
```
kolyshkin, line 155: Not important, but the code used to keep checking other cases even after one of them failed, and now it's not.

Fine either way for me, just noticing.

libcontainer/factory_linux.go
```diff
@@ -196,7 +196,7 @@ func New(root string, options ...func(*LinuxFactory) error) (Factory, error) {
 		Validator: validate.New(),
 		CriuPath:  "criu",
 	}
-	Cgroupfs(l)
+	_ = Cgroupfs(l)
```
kolyshkin, line 199: Maybe it's better to return an error from here rather than ignore it.

libcontainer/factory_linux_test.go
```diff
@@ -122,15 +122,18 @@ func TestFactoryNewTmpfs(t *testing.T) {
 	if m.Source != "tmpfs" {
 		t.Fatalf("Source of root: %s, expected %s", m.Source, "tmpfs")
 	}
-	unix.Unmount(root, unix.MNT_DETACH)
+	err = unix.Unmount(root, unix.MNT_DETACH)
+	if err != nil {
+		t.Error("failed to unmount root", err)
```
kolyshkin, line 127:
```suggestion
		t.Error("failed to unmount root:", err)
```

tty.go
kolyshkin: Currently no callers of this care about the error, so I'd make it not return anything.

kolyshkin:
Mostly LGTM; left some nits.

kolyshkin:
Overall, I don't think we should try to make master CI green again before rc93, so I'd rather have this one merged **after** the release.

libcontainer/container_linux.go
```diff
@@ -1299,7 +1299,7 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 	if err != nil {
 		return err
 	}
-	defer unix.Unmount(root, unix.MNT_DETACH)
+	defer func() { _ = unix.Unmount(root, unix.MNT_DETACH) }()
```
cyphar, line 1302: For the `go funcWhichReturnsErr()` cases I would go with `// nolint` because the alternative is quite a bit uglier (and I think in some cases may have strange behaviour with variable values -- because IIRC closures capture variables slightly differently to `go` statements).

tty.go
cyphar: Having `Close()` return an error is mostly so that you can map it to `io.Closer` as well as being "good Go practice". I think there is actually a bug here that we aren't returning `Close()` errors from any of the `Close`s we call in the function, but since very few projects check `Close()` errors I guess it isn't that bad...

cyphar (changes requested):
My main comment is that I'm not sure `defer func() { _ = ... }()` is worth the annoyance over `// nolint` -- especially since I'm pretty sure it captures variables differently to a naked `defer`.

tty.go
cyphar: Having `Close()` return an error is mostly so that you can map it to `io.Closer` as well as being "good Go practice". I think there is actually a bug here that we aren't returning `Close()` errors from any of the `Close`s we call in the function, but since very few projects check `Close()` errors I guess it isn't that bad... :man_shrugging:

libcontainer/container_linux.go
kolyshkin: I was thinking about suggesting something like that, too. Having to add a whole function call to satisfy a linter is borderline overkill.

libcontainer/integration/checkpoint_test.go
thaJeztah: Updated (not pushed yet)

libcontainer/configs/config_test.go
thaJeztah: Hmm... but there's only a single check in this test, and if that would fail it would already fail the test.

I _think_ it's correct with this change, so I'll keep it, but feel free to comment if I overlooked something 😅

libcontainer/factory_linux.go
thaJeztah: updated (not pushed yet)

libcontainer/integration/exec_test.go
thaJeztah: updated 👍

libcontainer/container_linux.go
thaJeztah: changed those to use `//nolint: errcheck`

libcontainer/nsenter/nsenter_test.go
```diff
@@ -173,8 +173,10 @@ func TestNsenterChildLogging(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create log pipe %v", err)
 	}
-	defer logread.Close()
-	defer logwrite.Close()
+	defer func() {
+		_ = logread.Close()
+		_ = logwrite.Close()
+	}()
```
thaJeztah, line 179: Oh! I need to reverse these to keep the old behaviour; defers are triggered in the reverse order that they were added
```suggestion
	defer func() {
		_ = logwrite.Close()
		_ = logread.Close()
	}()
```

thaJeztah: 👍 updated

thaJeztah:
rebased

thaJeztah:
@cyphar @kolyshkin ptal

AkihiroSuda (approved):

AkihiroSuda:
Needs rebase

thaJeztah:
rebased 👍

AkihiroSuda:
@kolyshkin @cyphar PTAL, v1.0.0 should be released with green CI

cyphar (approved):
LGTM, though we should probably also enable the golint tests for PRs if it's green now.

cyphar:
Seems that this didn't fix all the lints:

```
  Running [/home/runner/golangci-lint-1.36.0-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: SA4000: identical expressions on the left and right side of the '-' operator (staticcheck)
  Error: `archStringToNative` is unused (deadcode)
  Error: `memoryAssigned` is unused (deadcode)
  Error: Error return value of `container.Destroy` is not checked (errcheck)
  Error: Error return value of `container.Destroy` is not checked (errcheck)
  Error: Error return value of `container.Destroy` is not checked (errcheck)
  Error: ineffectual assignment to `bpfType` (ineffassign)
```

I'll send a PR to fix these today.

PR #2782, opened 2021-02-01
.golangci.yml: add

kolyshkin:
This is to enable seccomp build tag (some linters do use tags,
some don't).

Related to https://github.com/opencontainers/runc/issues/2778

~~NOTE i have temporary added a commit to test/report all the issues
(rather than only those brought by a PR) -- this is needed to see how
the newly added commit with a build tag works. I will remove the commit later.~~ Removed

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v2
         with:
           # must be specified without patch version
-          version: v1.31
+          version: v1.36
           # Only show new issues for a pull request.
           only-new-issues: true
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,5 @@
+# For documentation, see https://golangci-lint.run/usage/configuration/
+
+run:
+  build-tags:
+    - seccomp


cyphar:
But this requires us to keep the `libseccomp-devel` dependency, right?

kolyshkin:
OK it's working now, finding an issue that was previously missed:

>   Error: `archStringToNative` is unused (deadcode)

(see https://github.com/opencontainers/runc/pull/2782/checks?check_run_id=1810442799)

Reverting the last commit so this is ready to be merged.

kolyshkin:
> But this requires us to keep the libseccomp-devel dependency, right?

Yes. I guess it's inevitable (or we'll skip some linters on the code that's under `seccomp` build tag).

kolyshkin:
> But this requires us to keep the libseccomp-devel dependency, right?

Are you trying to say it's a problem?

cyphar:
No no, it's not a problem. You mentioned it when I added it to the GHA workflow, but I guess we will need it indefinitely until we stop directly importing `libseccomp` via CGO in runc.

AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #2757, opened 2021-01-21
tests/int: cleanups

kolyshkin:
The ultimate idea is to parallelize running the integration tests.

So far just some preliminary cleanups, mostly resulting in simpler setup/cleanup
and laying the ground for parallel tests (every test is run in its own set of dirs and cgroups).

Issue #2158, opened 2019-10-27
cgroup2: cannot mount /sys/fs/cgroup when running in UserNS


AkihiroSuda:
As of c4d8e1688c816a8cef632a3b44a38611511b7140, runc cannot mount `/sys/fs/cgroup` when launched in UserNS with cgroup2 unified-mode

```console
$ rootlesskit runc run foo
WARN[0000] signal: killed                               
ERRO[0000] container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"cgroup\\\" to rootfs \\\"/home/suda/tmp/runctest/rootfs\\\" at \\\"/sys/fs/cgroup\\\" caused \\\"operation not permitted\\\"\"" 
container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"cgroup\\\" to rootfs \\\"/home/suda/tmp/runctest/rootfs\\\" at \\\"/sys/fs/cgroup\\\" caused \\\"operation not permitted\\\"\""
[rootlesskit:child ] error: command [runc run foo] exited: exit status 1
[rootlesskit:parent] error: child exited: exit status 1
```

`config.json` contains:
```json
    {
      "destination": "/sys/fs/cgroup",
      "type": "cgroup",
      "source": "cgroup",
      "options": [
        "nosuid",
        "noexec",
        "nodev",
        "relatime",
        "ro"
      ]
    }
```

The same config works with crun v0.10.2.
runc can start up when the `/sys/fs/cgroup` entry is not specified in the config.

AkihiroSuda:
It seems `cgroup2` cannot be mounted without unsharing CgroupNS.

crun seems solving this issue by bind-mounting `/sys/fs/cgroup` from the host.

https://github.com/containers/crun/blob/4325a78320852aa5dacb4d403ae01b241413068c/src/libcrun/linux.c#L433-L459

PR #2786, opened 2021-02-03
Do not convert blkio weight value using blkio->io conversion scheme

dqminh:
bfq weight controller (i.e. io.bfq.weight if present) is still using the
same bfq weight scheme (i.e 1->1000, see [1].) Unfortunaly the
documentation for this was wrong, and only fixed recently [2].

Therefore, if we map blkio weight to io.bfq.weight, there's no need to
do any conversion.

[1] https://github.com/torvalds/linux/blob/master/Documentation/block/bfq-iosched.rst
[2] https://github.com/torvalds/linux/commit/65752aef0a407e1ef17ec78a7fc31ba4e0b360f9

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -190,6 +190,20 @@ function setup() {
 	[[ "$output" == *'invalid configuration'* ]]
 }
 
+@test "runc run (blkio weight)" {
+	requires root cgroups_v2
+
+	set_cgroups_path "$BUSYBOX_BUNDLE"
+	update_config '.linux.resources.blockIO |= {"weight": 750}' "${BUSYBOX_BUNDLE}"
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroups_unified
+	[ "$status" -eq 0 ]
+
+	runc exec test_cgroups_unified sh -c 'cat /sys/fs/cgroup/io.bfq.weight'
+	[ "$status" -eq 0 ]
+	[ "$output" = 'default 750' ]
+}
+
 @test "runc run (cgroup v2 resources.unified only)" {
 	requires root cgroups_v2
 
--- a/libcontainer/cgroups/fs2/io.go
+++ b/libcontainer/cgroups/fs2/io.go
@@ -29,7 +29,7 @@ func setIo(dirPath string, cgroup *configs.Cgroup) error {
 	if cgroup.Resources.BlkioWeight != 0 {
 		filename := "io.bfq.weight"
 		if err := fscommon.WriteFile(dirPath, filename,
-			strconv.FormatUint(cgroups.ConvertBlkIOToCgroupV2Value(cgroup.Resources.BlkioWeight), 10)); err != nil {
+			strconv.FormatUint(uint64(cgroup.Resources.BlkioWeight), 10)); err != nil {
 			return err
 		}
 	}
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -402,17 +402,6 @@ func WriteCgroupProc(dir string, pid int) error {
 
 // Since the OCI spec is designed for cgroup v1, in some cases
 // there is need to convert from the cgroup v1 configuration to cgroup v2
-// the formula for BlkIOWeight is y = (1 + (x - 10) * 9999 / 990)
-// convert linearly from [10-1000] to [1-10000]
-func ConvertBlkIOToCgroupV2Value(blkIoWeight uint16) uint64 {
-	if blkIoWeight == 0 {
-		return 0
-	}
-	return uint64(1 + (uint64(blkIoWeight)-10)*9999/990)
-}
-
-// Since the OCI spec is designed for cgroup v1, in some cases
-// there is need to convert from the cgroup v1 configuration to cgroup v2
 // the formula for cpuShares is y = (1 + ((x - 2) * 9999) / 262142)
 // convert from [2-262144] to [1-10000]
 // 262144 comes from Linux kernel definition "#define MAX_SHARES (1UL << 18)"
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -548,20 +548,6 @@ func TestGetHugePageSizeImpl(t *testing.T) {
 	}
 }
 
-func TestConvertBlkIOToCgroupV2Value(t *testing.T) {
-	cases := map[uint16]uint64{
-		0:    0,
-		10:   1,
-		1000: 10000,
-	}
-	for i, expected := range cases {
-		got := ConvertBlkIOToCgroupV2Value(i)
-		if got != expected {
-			t.Errorf("expected ConvertBlkIOToCgroupV2Value(%d) to be %d, got %d", i, expected, got)
-		}
-	}
-}
-
 func TestConvertCPUSharesToCgroupV2Value(t *testing.T) {
 	cases := map[uint64]uint64{
 		0:      0,


dqminh:
This will also need to be fixed in crun and systemd among other things afair @giuseppe

giuseppe:
> This will also need to be fixed in crun and systemd among other things afair @giuseppe

thanks for letting me know!  I've opened a PR for crun as well: https://github.com/containers/crun/pull/584

giuseppe:
@keszybz FYI

dqminh:
Updated to add a test so we can try to catch this in the future. I'm not sure if there's movement to merge io.weight and io.bfq.weight at the moment, at which point we need to fix the inconsistency between 2 interface.

kolyshkin (approved):
LGTM

kolyshkin:
CI failure on Fedora is a glitch (of download.fedoraproject.org) -- created https://github.com/opencontainers/runc/pull/2787 to address.

CI restared.

tests/integration/cgroups.bats
```diff
@@ -190,6 +190,22 @@ function setup() {
 	[[ "$output" == *'invalid configuration'* ]]
 }
 
+@test "runc run (blkio weight)" {
+	requires root cgroups_v2
```
AkihiroSuda, line 194: Shouldn't require root?

tests/integration/cgroups.bats
```diff
@@ -190,6 +190,22 @@ function setup() {
...
+
+	set_cgroups_path "$BUSYBOX_BUNDLE"
+	update_config '.linux.resources.blockIO |= {"weight": 750}' "${BUSYBOX_BUNDLE}"
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroups_v2_blkio
```
cyphar, line 199: If you add a new container name you need to add it to the `teardown()` hook at the start -- `teardown_running_container`.

tests/integration/cgroups.bats
```diff
@@ -190,6 +190,22 @@ function setup() {
...
+	[ "$status" -eq 0 ]
+
+	runc exec test_cgroups_v2_blkio sh -c 'cat /sys/fs/cgroup/io.bfq.weight'
+	[ "$status" -eq 0 ]
+	echo "$output"
```
cyphar, line 204: This shouldn't be necessary? `runc` is a function which prints the output to the bats log (I'm also pretty sure a naked `echo` doesn't work which is why `runc` redirects to `stderr`):

```bash
# Wrapper for runc.
function runc() {
	run __runc "$@"

	# Some debug information to make life easier. bats will only print it if the
	# test failed, in which case the output is useful.
	echo "runc $@ (status=$status):" >&2
	echo "$output" >&2
}

# Raw wrapper for runc.
function __runc() {
	"$RUNC" ${RUNC_USE_SYSTEMD+--systemd-cgroup} --root "$ROOT" "$@"
}
```

cyphar (changes requested):
Test needs to be cleaned up.

keszybz:
@giuseppe thanks for the ping.

> This will also need to be fixed in crun and systemd among other things afair

Do I get this right that `io.bfq.weight` wants a value that is in the range 1..1000, while `io.weight` wants a value in the range 1..10000? systemd treats the cgroupv2 range as the input value, so we'd need to add a conversion `io.bfq.weight = (io.weight + 9) / 10`.

keszybz:
^ @paolo-github

keszybz:
https://github.com/systemd/systemd/pull/18467

dqminh:
@keszybz yes, i think that is correct.

Which then also raises more questions:

1. Do we want to set both interface ( like what systemd is doing ) ? In our case, since the spec is saying that the allowed range is 0-1000, and i assume that we don't want breaking changes, we still have to do the conversion to write to `io.weight`
2. I don't think runc's systemd v2 is setting IOWeight, is that right ? I tried to do a quick search and can't find any.

In general, `io.weight` iocost interface is also more likely to be available everywhere (is it ?, see https://github.com/containers/podman/issues/8737), so maybe doing 1) makes sense.

tests/integration/cgroups.bats
kolyshkin: I noticed it, too, but did not comment because I am fixing all this soon (after https://github.com/opencontainers/runc/pull/2757 is merged).

kolyshkin (commented):
Indeed, the `io.bfq.weight` range is 1-1000, but `io.weight` range is 1-10000. runc is not currently setting `io.weight`.

paolo-github:
If useful for you, I'll be glad to change the range for BFQ too. The current BFQ range is due just to legacy.

keszybz:
> If useful for you, I'll be glad to change the range for BFQ too. The current BFQ range is due just to legacy.

That'd make an even bigger mess ;( Please just make bfq read `io.weight`...

paolo-github:
Ok, I don't want to add noise

Yes, making bfq read io.weight is the sensible thing. We fought for that, for months. Not only we pushed, but we also proposed three or four working solutions, following the always varying requests. At no avail, as I said in the other thread.

I don't think I can do more than that, sorry.

kolyshkin:
nit: a typo in commit message: `Unfortunaly -> Unfortunately`.

The only non-addressed review comment is https://github.com/opencontainers/runc/pull/2786#discussion_r569835244

AkihiroSuda (approved):
Nits can be fixed later. LGTM.

kolyshkin:
Took a liberty to rebase, fix a typo in commit message, and remove root requirement from the integration test case. Let's see if it passes CI.

kolyshkin (approved):
LGTM

kolyshkin:
@AkihiroSuda PTAL

kolyshkin:
The test case is not working under rootless on Fedora:
```
not ok 10 runc run (blkio weight)
# (in test file /vagrant/tests/integration/cgroups.bats, line 203)
# 
# runc run -d --console-socket /tmp/console.sock test_cgroups_unified (status=0):
# 
# runc exec test_cgroups_unified sh -c cat /sys/fs/cgroup/io.bfq.weight (status=1):
# cat: can't open '/sys/fs/cgroup/io.bfq.weight': No such file or directory
```

Apparently the problem is, rootless containers somehow do no honor cgroupns. Instead of "just this container" we see everything in container's view of /sys/fs/cgroup.

Looking...

kolyshkin:
Normal container's cgroup2 mount:
> [kir@kir-rhat runc]$ cat /proc/2474364/mountinfo  | grep cgroup
> 1956 1955 0:27 /user.slice/user-1000.slice/user@1000.service/x453 /sys/fs/cgroup ro,nosuid,nodev,noexec,relatime - cgroup2 cgroup rw,seclabel

rootless container cgroup2 mount:
> [kir@kir-rhat runc]$ cat /proc/2475909/mountinfo  | grep cgroup
> 30 23 0:27 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel

Ah, so it is caused by https://github.com/opencontainers/runc/issues/2158. Hmmm.

dqminh:
@kolyshkin sorry for the delay. yes i indeed hit this along the removal of root for the test so it wasn't done, and i forgot to put a comment on that.

kolyshkin:
@dqminh your patch is fine (note that I pushed into your branch though), the rootless cgroup mount is a separate issue I'm going to address.

tests/integration/cgroups.bats
kolyshkin: @AkihiroSuda once https://github.com/opencontainers/runc/pull/2818 is merged we will be able to enable it (rm root, add `[[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup`)

kolyshkin: ...because currently `runc exec test_cgroups_unified sh -c 'cat /sys/fs/cgroup/io.bfq.weight'` is not working

kolyshkin (approved):
LGTM

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda (approved):

kolyshkin:
@cyphar PTAL (you requested changes and thus merging is blocked until you approve). Seems your review comments are now addressed.

cyphar (approved):
LGTM. As an aside, feel free to dismiss my old reviews if they are clearly no longer applicable. (I wish GitHub did this already, because it treats approvals as something that can go stale but "request changes" reviews aren't treated as stale.)

PR #2788, opened 2021-02-03
libcontainer: fix bad conversion from syscall.Errno to error

mauriciovasquezbernal:
The correct way to do that conversion according to
https://pkg.go.dev/syscall#Errno is:

```
err = nil
if errno != 0 {
	err = errno
}
```

In this case the error check will always report a false positive in
unix.RawSyscall(unix.SYS_SECCOMP, ...), probably nobody has faced this
problem because the code takes the other path in most of the cases.

I found that while rebasing https://github.com/opencontainers/runc/pull/2682. 

cc @cyphar


cyphar (approved):
Huh, I could've sworn the other way worked as well. Yeah, LGTM.

mrunalp (approved):

PR #2785, opened 2021-02-03
thaw and retry again in freezing state

Vanient:
1. just return nil if freeze succeed, retry in any other case.
    2. freeze retry always fail in freezing state, but if we thaw it and
       retry again, more likely to succeed, so we'd better thaw and retry
       freeze again.
Signed-off-by: xiadanni <xiadanni1@huawei.com>

PR #2791, opened 2021-02-04
cgroupv1 freezer: thaw to increase freeze chances

kolyshkin:
_This is an alternative to #2785 and a continuation of #2774, addressing #2753_

It appears that briefly thawing the cgroup while freezing
greatly increases its chances to freeze successfully.

The test case I used is doing runc exec in a look parallel with runc
pause/resume in another loop, and the failure to freeze rate reduced
from 40 to 0 per minute (tested inside a VM using a busybox container
running sleep 1h, doing about 1500 pause/resumes and 650 execs per
minute), with max retries being 150 (of 1000).

This is still a game of chances, so failures are possible.

Thanks to @Vanient for idea and initial implementation.

Closes: #2785


cyphar (approved):
LGTM.

dqminh (approved):

PR #2793, opened 2021-02-05
cgroup2: devices: handle eBPF skipping more correctly

### LINKED ISSUE #2792: Setting cgroup config failed when nested inside LXC unprivileged container w/ pure cgroups v2 ("unified hierarchy")
I am trying to run Docker inside an unprivileged LXC container on a host with pure cgroups v2 (systemd "unified hierarchy"), but with limited success with `runc` runtime. Before `runc 1.0-rc93`, the runtime fails in an attempt to use `systemd --user` even when it is the root user inside the unprivileged LXC container, which I believe is the same regression as described in #2724 and now fixed in #2725. However, now, on `1.0-rc93`, another new error emerges from `runc`:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: process_linux.go:458: setting cgroup config for procHooks process caused: can't load program: operation not permitted: unknown.
```

For context, using `crun` as Docker's OCI runtime instead of `runc` works perfectly and I've had no issues with it so far.

cyphar:
In the past we incorrectly handled eBPF errors in two ways:

 1. We would only ignore errors if there was an allow rule in the list
    (this doesn't make sense because for security purposes we only care
    if a *deny* rule is being ignored). Arguably this is a security flaw
    but you would only get an error from bpf(2) in rare cases, and thus
    is not a big enough deal to go through security review.

 2. If we were in a rootless container we would still return an error
    even though bpf(2) is blocked for rootless containers.

Fixes #2792 

### DIFF
--- a/libcontainer/cgroups/fs2/devices.go
+++ b/libcontainer/cgroups/fs2/devices.go
@@ -7,6 +7,8 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups/ebpf/devicefilter"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/devices"
+	"github.com/opencontainers/runc/libcontainer/system"
+
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
@@ -26,11 +28,26 @@ func isRWM(perms devices.Permissions) bool {
 	return r && w && m
 }
 
-// the logic is from crun
-// https://github.com/containers/crun/blob/0.10.2/src/libcrun/cgroup.c#L1644-L1652
+// This is similar to the logic applied in crun for handling errors from bpf(2)
+// <https://github.com/containers/crun/blob/0.17/src/libcrun/cgroup.c#L2438-L2470>.
 func canSkipEBPFError(cgroup *configs.Cgroup) bool {
+	// If we're running in a user namespace we can ignore eBPF rules because we
+	// usually cannot use bpf(2), as well as rootless containers usually don't
+	// have the necessary privileges to mknod(2) device inodes or access
+	// host-level instances (though ideally we would be blocking device access
+	// for rootless containers anyway).
+	if system.RunningInUserNS() {
+		return true
+	}
+
+	// We cannot ignore an eBPF load error if any rule if is a block rule or it
+	// doesn't permit all access modes.
+	//
+	// NOTE: This will sometimes trigger in cases where access modes are split
+	//       between different rules but to handle this correctly would require
+	//       using ".../libcontainer/cgroup/devices".Emulator.
 	for _, dev := range cgroup.Resources.Devices {
-		if dev.Allow || !isRWM(dev.Permissions) {
+		if !dev.Allow || !isRWM(dev.Permissions) {
 			return false
 		}
 	}


libcontainer/cgroups/fs2/devices.go
```diff
@@ -26,11 +28,21 @@ func isRWM(perms devices.Permissions) bool {
...
+
+	// We cannot ignore an eBPF load error if any rule if is a block rule or it
+	// doesn't permit all access modes.
+	//
+	// NOTE: This will sometimes trigger in cases where access modes are split
+	//       between different rules but to handle this correctly would require
+	//       using ".../libcontainer/cgroup/devices".Emulator.
 	for _, dev := range cgroup.Resources.Devices {
-		if dev.Allow || !isRWM(dev.Permissions) {
+		if !dev.Allow || !isRWM(dev.Permissions) {
```
AkihiroSuda, line 45: Can we have an integration test for this? Also we need to make sure this is not a vulnerability.

libcontainer/cgroups/fs2/devices.go
```diff
@@ -26,11 +28,21 @@ func isRWM(perms devices.Permissions) bool {
 	return r && w && m
 }
 
-// the logic is from crun
-// https://github.com/containers/crun/blob/0.10.2/src/libcrun/cgroup.c#L1644-L1652
 func canSkipEBPFError(cgroup *configs.Cgroup) bool {
+	// If we're running in a user namespace we can ignore eBPF rules because
+	// we're already restricted to not be able to mknod(2) by default.
```
AkihiroSuda, line 35: s/mknod/bpf/ ?

cyphar: I did mean `mknod(2)` because that's one of the things the devices rules block, but now that you mention it `bpf(2)` being blocked is the more direct reason why we ignore this (and the devices rules block other things that are non-trivial to block outside of cgroups).

libcontainer/cgroups/fs2/devices.go
cyphar: If the error path can be hit by non-rootless containers then yeah this is probably a vulnerability.

cyphar: I'll need to think about the easiest way of integration testing this one -- we have to trigger a syscall failure.

cyphar:
@PeterCxy  can you check if this fixes the issue for you?

PeterCxy:
@cyphar It did get rid of the original error in #2792, but unfortunately there seems to be a new error now

```
docker: Error response from daemon: OCI runtime create failed: unable to retrieve OCI runtime error (open /var/run/docker/containerd/daemon/io.containerd.runtime.v2.task/moby/69c64ca8f339bc9a92aa1c3370075a32eacb7704647503bdbd75a838d7525cbb/log.json: no such file or directory): fork/exec /root/go/github.com/opencontainers/runc: permission denied: <nil>: unknown.
```

I guess I will open a new issue about the new error. This PR seems to work for what it was intended to fix.

PeterCxy:
@cyphar (Please ignore my last deleted reply, it was a mistake)

I can confirm this fixed #2792 and now Docker works fine with runc inside unprivileged containers with cgroups v2

giuseppe (approved):
LGTM

AkihiroSuda (approved):

libcontainer/cgroups/fs2/devices.go
cyphar: I tried to do this but it seems like it's not really simple to force a failure with `bpf(2)` -- the obvious way of doing it (making a program that is too big so it trips up the eBPF program size rules) appears to cause the test to stall endlessly. I'll add a unit test instead.

cyphar: I've opened #2797 to track the need for unit tests more generally. I will work on that separately.

AkihiroSuda (approved):

mrunalp (approved):

PR #2794, opened 2021-02-05
libcontainer: fix file descriptor leaks (exec fifo and log file pair)

cclerget:
When using the libcontainer package directly, an application can quickly raise the `Too many open files` error
due to some file descriptors left open, so far exec fifo and log file pair are not closed correctly.

Signed-off-by: Cédric Clerget <cedric.clerget@gmail.com>

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -245,7 +245,10 @@ func (p *setnsProcess) setExternalDescriptors(newFds []string) {
 }
 
 func (p *setnsProcess) forwardChildLogs() {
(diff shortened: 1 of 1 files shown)


kolyshkin:
We need a unit test for this I think.

kolyshkin:
@cclerget sorry for hijacking this, but I was looking into how we'd better handle this closing, as well as if there are any other leaked fds, and ended up with https://github.com/opencontainers/runc/pull/2802 -- can you please review that one?

kolyshkin:
Closing in favor of #2802

Issue #2748, opened 2021-01-17
container restore layering conundrum


adrianreber:
I am currently working on implementing checkpoint/restore on the Kubernetes level (https://github.com/kubernetes/kubernetes/pull/97194) and during restore I have a problem with missing directories for bind mounts. The following situation gives me currently problems

1. Kubernetes creates something like `/var/lib/kubelet/pods/36dfe704-5fd4-4ec5-aaf1-9d7375364de0/volumes/kubernetes.io~secret/default-token-d2cs7` and it tells CRI-O, in my case, to mount it at `/var/run/secrets/kubernetes.io/serviceaccount`.
2. CRI-O mounts an empty  `tmpfs` for `/run/secrets` and tells `runc` to bind mount it at `/run/secrets`.
3. `runc` does it exactly what it is expected to do. It bind mounts the external `tmpfs` at `/run/secrets` and bind mounts  `/var/lib/kubelet/pods/36dfe704-5fd4-4ec5-aaf1-9d7375364de0/volumes/kubernetes.io~secret/default-token-d2cs7` at `/var/run/secrets/kubernetes.io/serviceaccount`.
4. `runc` creates the directory for that in the bind mounted `tmpfs` at `/run/secrets`

At this point everything is done and it is working. The thing which I do not totally understand is that `runc` seems to move the mounts of the container rootfs after it is done. Not sure what is happening exactly but if I look, in one of my tests, at `/var/lib/containers/storage/overlay/89579e5eb2910e11b8e8d852b1ee022272c789aa97670e92be388613e6984fe6/merged/run/secrets/` I do not see the necessary directories, but after a lot of printf debugging I see that `runc` actually creates the directory at this location. So, the part I am not totally understanding is that the directory is created at one location, but I can see it, outside of the container at some other location, once the container is running. So it seems the mount is moved after/during container creation.

During restore I now have the following situation.

1. Kubernetes re-creates the same directory to be mounted in the container at `/var/run/secrets/kubernetes.io/serviceaccount`.
2. CRI-O still mounts an empty `tmpfs` for `/run/secrets`. This `tmpfs` is basically how and where the problem appears.
3. `runc` prepares restoring the container and recreates the directory structure just as during initial container creation so that CRIU can mount the previous directories at the same places they used to be. `runc`, during restore, does not bind mount directories to create mount points (maybe this is the main problem and the reason for the errors I am seeing)
[...]

PR #2715, opened 2021-01-05
runc run: resolve tmpfs mount dest in container scope

kolyshkin:
In case a tmpfs mount path contains absolute symlinks, runc errors out
because those symlinks are resolved in the host (rather than container)
filesystem scope.
    
The fix is similar to that for bind mounts -- resolve the destination
in container rootfs scope using securejoin, and use the resolved path.
    
A simple integration test case is added to prevent future regressions.
    
Fixes https://github.com/opencontainers/runc/issues/2683.

PR #2798, opened 2021-02-08
Correctly restore containers with nested bind mounts

adrianreber:
This tries to fix https://github.com/opencontainers/runc/issues/2748

This change is motivated by the work in https://github.com/kubernetes/kubernetes/pull/97194

During restore CRIU mounts all mounts as they were during checkpointing. Using nested bind mounts this can lead to situation where a mountpoint in the parent bind mount is missing.

During initial container creation 'runc' mounts a bind mount and then creates the mountpoints for nested mountpoints.

This PR changes `runc` to also mount all bind mounts before telling CRIU to restore the container to re-create missing mountpoints.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1235,10 +1235,37 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
 	// Now go through all mounts and create the mountpoints
 	// if the mountpoints are not on a tmpfs, as CRIU will
 	// restore the complete tmpfs content from its checkpoint.
+	umounts := []string{}
+	defer func() {
+		for _, u := range umounts {
+			if e := unix.Unmount(u, unix.MNT_DETACH); e != nil {
+				if e != unix.EINVAL {
+					// Ignore EINVAL as it means 'target is not a mount point.'
+					// It probably has already been unmounted.
+					logrus.Warnf("Error during cleanup unmounting of %q (%v)", u, e)
+				}
+			}
+		}
+	}()
 	for _, m := range mounts {
 		if !isPathInPrefixList(m.Destination, tmpfs) {
 			if err := c.makeCriuRestoreMountpoints(m); err != nil {
 				return err
+			}
+			// If the mount point is a bind mount, we need to mount
+			// it now so that runc can create the necessary mount
+			// points for mounts in bind mounts.
+			// This also happens during initial container creation.
+			// Without this CRIU restore will fail
+			// See: https://github.com/opencontainers/runc/issues/2748
+			// It is also not necessary to order the mount points
+			// because during initial container creation mounts are
+			// set up in the order they are configured.
+			if m.Device == "bind" {
+				if err := unix.Mount(m.Source, m.Destination, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+					return errorsf.Wrapf(err, "unable to bind mount %q to %q", m.Source, m.Destination)
+				}
+				umounts = append(umounts, m.Destination)
 			}
 		}
 	}
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -315,3 +315,45 @@ function simple_cr() {
 	unlink "$tmp"
 	test -f ./work-dir/"$tmplog2" && unlink ./work-dir/"$tmplog2"
 }
+
+@test "checkpoint and restore with nested bind mounts" {
+	bind1=$(mktemp -d -p .)
+	bind2=$(mktemp -d -p .)
+	update_config '	  .mounts += [{
+					type: "bind",
+					source: "'"$bind1"'",
+					destination: "/test",
+					options: ["rw", "bind"]
+				},
+	                        {
+					type: "bind",
+					source: "'"$bind2"'",
+					destination: "/test/for/nested",
+					options: ["rw", "bind"]
+				}]'
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 0 ]
+
+	testcontainer test_busybox running
+
+	# checkpoint the running container
+	runc --criu "$CRIU" checkpoint --work-path ./work-dir test_busybox
+	grep -B 5 Error ./work-dir/dump.log || true
+	[ "$status" -eq 0 ]
+
+	# after checkpoint busybox is no longer running
+	testcontainer test_busybox checkpointed
+
+	# cleanup mountpoints created by runc during creation
+	# the mountpoints should be recreated during restore - that is the actual thing tested here
+	rm -rf "${bind1:?}"/*
+
+	# restore from checkpoint
+	runc --criu "$CRIU" restore -d --work-path ./work-dir --console-socket "$CONSOLE_SOCKET" test_busybox
+	grep -B 5 Error ./work-dir/restore.log || true
+	[ "$status" -eq 0 ]
+
+	# busybox should be back up and running
+	testcontainer test_busybox running
+}


libcontainer/container_linux.go
```diff
@@ -1235,11 +1235,39 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
...
 	// if the mountpoints are not on a tmpfs, as CRIU will
 	// restore the complete tmpfs content from its checkpoint.
+	umounts := []string{}
+	defer func() {
+		for _, u := range umounts {
+			if e := unix.Unmount(u, unix.MNT_DETACH); e != nil {
+				if e != unix.EINVAL {
+					// Ignore EINVAL as it means 'EINVAL target is not a mount point.'
+					// It probably has already been unmounted.
+					logrus.Warnf("Error during cleanup unmounting of %q (%v)\n", u, e)
```
kolyshkin, line 1245: nit: `\n` is already added by logrus.

libcontainer/container_linux.go
```diff
@@ -1235,11 +1235,39 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
 	// Now go through all mounts and create the mountpoints
 	// if the mountpoints are not on a tmpfs, as CRIU will
 	// restore the complete tmpfs content from its checkpoint.
+	umounts := []string{}
+	defer func() {
+		for _, u := range umounts {
+			if e := unix.Unmount(u, unix.MNT_DETACH); e != nil {
+				if e != unix.EINVAL {
+					// Ignore EINVAL as it means 'EINVAL target is not a mount point.'
```
kolyshkin, line 1243: nit
```suggestion
					// Ignore EINVAL as it means 'target is not a mount point.'
					``

File: libcontainer/container_linux.go [hunk 3]
```diff
@@ -1235,11 +1235,39 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
...
+			// points for mounts in bind mounts.
+			// This also happens during initial container creation.
+			// Without this CRIU restore will fail
+			// See: https://github.com/opencontainers/runc/issues/2748
+			// It is also not necessary to order the mount points
+			// because during initial container creation mounts are
+			// set up in the order they are configured.
+			if m.Device == "bind" {
+				if err := unix.Mount(m.Source, m.Destination, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+					logrus.Warnf("Error during mounting of %q (%v)\n", m.Destination, err)
```
kolyshkin, line 1266: 1. nit: no need for `\n`.

kolyshkin: 2. Since we're returning an error anyway, maybe there's no need to log it (the caller will).

libcontainer/container_linux.go
```diff
@@ -1235,11 +1235,39 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
...
+					return err
```
kolyshkin, line 1267: Need to wrap an error here, as the error returned by `unix.Mount` is extremely non-descriptive (it's basically an errno wrapped in `syscall.Error`).

```suggestion
					return errors.Wrapf(err, "unable to bind mount %q to %q", m.Source, m.Destination).
```

libcontainer/container_linux.go
```diff
@@ -1235,11 +1235,39 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
...
+			// it now so that runc can create the necessary mount
+			// points for mounts in bind mounts.
+			// This also happens during initial container creation.
+			// Without this CRIU restore will fail
+			// See: https://github.com/opencontainers/runc/issues/2748
+			// It is also not necessary to order the mount points
+			// because during initial container creation mounts are
+			// set up in the order they are configured.
+			if m.Device == "bind" {
+				if err := unix.Mount(m.Source, m.Destination, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
```
kolyshkin, line 1265: We _might_ need to resolve m.Destination inside a container root (see e.g. https://github.com/opencontainers/runc/pull/2715/commits/637f82d6b8b45a99519b863f8cf7461760da7a97 or `prepareBindMount` at https://github.com/opencontainers/runc/blob/94ae7afa36cc3b8f551e0bc67cf83e5efdf2555f/libcontainer/rootfs_linux.go#L198-L212)

To check, you need to add a test case where `m.Destination` is a symlink (something like `/mnt/test2 -> ../test1` or `/test2 -> /test1`).

tests/integration/checkpoint.bats
```diff
@@ -315,3 +315,56 @@ function simple_cr() {
...
+
+	# restore from checkpoint
+	runc --criu "$CRIU" restore -d --work-path ./work-dir --console-socket "$CONSOLE_SOCKET" test_busybox
+	grep -B 5 Error ./work-dir/restore.log || true
+	[ "$status" -eq 0 ]
+
+	# busybox should be back up and running
+	testcontainer test_busybox running
+
+	# Clean up
```
kolyshkin, line 364: This is problematic -- if the test ever fails, it exits and this code won't be run, ending up with a dangling mount.

Looks like this is easy to solve by not using a tmpfs mount on the host, and instead of remounting $bind1 use `rm -rf $bind1/*`.

libcontainer/container_linux.go
```diff
@@ -1235,11 +1235,39 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
...
 			}
+			// If the mount point is a bind mount, we need to mount
+			// it now so that runc can create the necessary mount
+			// points for mounts in bind mounts.
+			// This also happens during initial container creation.
+			// Without this CRIU restore will fail
+			// See: https://github.com/opencontainers/runc/issues/2748
+			// It is also not necessary to order the mount points
+			// because during initial container creation mounts are
+			// set up in the order they are configured.
```
kolyshkin, line 1263: Just curious -- what happens if the order of (nested) mounts in config is wrong? Will runc run fail?

I wrote a quick test and it works for me:
```bats
@test "runc run [wrong order of tmpfs mounts]" {
        update_config '   .mounts += [{
                                        type: "tmpfs",
                                        source: "tmpfs",
                                        destination: "/mount/sub/dir",
                                        options: ["ro", "nodev", "nosuid"]
                                },
                                {
                                        type: "tmpfs",
                                        source: "tmpfs",
                                        destination: "/mount",
                                        options: ["rw", "nodev", "nosuid"]
                                }]
                        | .process.args |= ["true"]'
        runc run test_busybox
        [ "$status" -eq 0 ]
}
```

It works no matter what the order of added mounts is. Same with bind mounts:
```bats
@test "runc run [wrong order of bind mounts]" {
        mkdir bind1 bind2
        update_config '   .mounts += [{
                                        source: "bind2",
                                        destination: "/mount/sub/dir",
                                        options: ["bind"]
                                },
                                {
                                        source: "bind1",
                                        destination: "/mount",
                                        options: ["bind"]
                                }]
                        | .process.args |= ["true"]'
        runc run test_busybox
        [ "$status" -eq 0 ]
}
```
(the test succeeds).

tests/integration/checkpoint.bats
cyphar: Also the `teardown_*` commands aren't needed here because they're already present in the file-level `teardown` hook.

libcontainer/container_linux.go
adrianreber: fixed

libcontainer/container_linux.go
adrianreber: fixed

libcontainer/container_linux.go
adrianreber: done

libcontainer/container_linux.go
adrianreber: done

libcontainer/container_linux.go
adrianreber: I was thinking about the exactly the same thing and initially had the code copied from `prepareBindMount()`, but as  `prepareBindMount()` is updating `m.Destination` with the changed destination path (https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L217) there is no need to run it a second time here.

libcontainer/container_linux.go
adrianreber: No, `runc` will not fail.

In your first case the second tmpfs mount will just overmount the first tmpfs mount. Same for your bind mount test. The mountpoints for the first mounts will always be created, but they will not be visible in the container. You should see a failure if you do a `ls /mount/sub/dir` in `.process.args`.

I was first looking at the code base for some time to find the function that orders the mounts to be able to re-use that code. But, from what I saw, `runc` just goes through the mounts as they appear in `config.json`.

tests/integration/checkpoint.bats
adrianreber: I changed it as proposed by @kolyshkin. I initially repeated the code from the `teardown` because I wanted to make sure the directories are gone after the test. It would only have cleaned up after a successful test, but as that is the expected result I went for that.

I am now no longer using the tmpfs, but I am just deleting the created files before restore and I added a conditional cleanup block to existing file-level `teardown`.

libcontainer/container_linux.go
```diff
@@ -1235,11 +1235,38 @@ func (c *linuxContainer) prepareCriuRestoreMounts(mounts []*configs.Mount) error
...
+			if m.Device == "bind" {
+				if err := unix.Mount(m.Source, m.Destination, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+					return errorsf.Wrapf(err, "unable to bind mount %q to %q", m.Source, m.Destination)
```
kolyshkin, line 1266: typo? `s/errorsf/errors/`

kolyshkin: Nope; both "errors" and "github.org/pkg/errors" are used here, the second is aliased as `errorsf`.

tests/integration/checkpoint.bats
```diff
@@ -315,3 +323,45 @@ function simple_cr() {
 	unlink "$tmp"
 	test -f ./work-dir/"$tmplog2" && unlink ./work-dir/"$tmplog2"
 }
+
+@test "checkpoint and restore with nested bind mounts" {
+	CR_TMP=$(mktemp -d)
+	bind1=$(mktemp -d -p "$CR_TMP")
+	bind2=$(mktemp -d -p "$CR_TMP")
```
kolyshkin, line 330: Here every test case is run with cd to a temporary directory containing the bundle (i.e. `rootfs` and `config.json`). This directory can be reused for other things, too, and is removed by `teardown_busybox`, so if you use `mktemp -p .` you don't have to clean up anything.

tests/integration/checkpoint.bats
```diff
@@ -315,3 +323,45 @@ function simple_cr() {
...
+
+	# checkpoint the running container
+	runc --criu "$CRIU" checkpoint --work-path ./work-dir test_busybox
+	grep -B 5 Error ./work-dir/dump.log || true
+	[ "$status" -eq 0 ]
+
+	# after checkpoint busybox is no longer running
+	testcontainer test_busybox checkpointed
+
+	# mount it again so that runc re-creates missing mountpoints
```
kolyshkin, line 357: seems that the comment is a leftover from the older version

kolyshkin (changes requested):
Just a single nit about making temp dirs under CWD, LGTM otherwise, thanks @adrianreber!

I have also checked that the test case indeed fails before the fix.

tests/integration/checkpoint.bats
adrianreber: fixed

tests/integration/checkpoint.bats
adrianreber: removed

adrianreber:
Fixed mktemp to use `-p .` and removed old comment.

kolyshkin (approved):
LGTM, and thanks!

kolyshkin:

AkihiroSuda (approved):

AkihiroSuda:
> test (1.16.x) Expected — Waiting for status to be reported

CI seems to have got stuck, attempting to restart by closing & reopening this PR

PR #2802, opened 2021-02-10
libct: fix some container.Run fd leaks, add fd leak test

kolyshkin:
_this is an alternative for https://github.com/opencontainers/runc/pull/2794_

TL;DR: Fix two cases of fd leaks on run, added a test case to avoid more in the future.

1. **libct: close execFifo after start**
    
    Apparently, the parent never closes execFifo fd. Not a problem for runc
    per se, but can be an issue for a user of libcontainer.

2. **libct: close log pipe**
    
   ~~Otherwise we have one extra fd opened after container.Run.~~
   Commit removed (obsoleted by #2835)

3. **libct/int: add TestFdLeaks**
    
    This is a very simple test that checks that container.Run do not leak
    opened file descriptors.
    
    In fact it does, so we have to add two exclusions:
    
    1. `/sys/fs/cgroup` directory is opened once per lifetime in `prepareOpenat2()`,
        provided that cgroupv2 is used and openat2 is available. This
        works as intended ("it's not a bug, it's a feature").
    
    2. ebpf program fd is leaked every time we call `setDevices()` for
       cgroupv2 (iow, every `container.Run` or `container.Set` leaks 1 fd).
       This needs to be fixed in ebpf, thus FIXME (https://github.com/opencontainers/runc/issues/2366#issuecomment-776411392)

Closes: #2794

Thanks to @cclerget for discovery and the initial fix.

## Suggested changelog entry
* libcontainer: fixed some container.Run file descriptor leaks (#2794, #2802)

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -55,6 +55,7 @@ type linuxContainer struct {
 	criuVersion          int
 	state                containerState
 	created              time.Time
+	fifo                 *os.File
 }
 
 // State represents a running container's state
@@ -380,6 +381,7 @@ func (c *linuxContainer) start(process *Process) (retErr error) {
 	}
 
 	if process.Init {
+		c.fifo.Close()
 		if c.config.Hooks != nil {
 			s, err := c.currentOCIState()
 			if err != nil {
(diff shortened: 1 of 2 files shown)


kolyshkin:
Looks like by closing fifo fd I have introduced some kind of a race. Looking...

cclerget:
Looks good to me, thanks for tackling this one :+1:

kolyshkin:
CI failure in Fedora 33 seems unrelated (filed https://github.com/opencontainers/runc/issues/2805) -- CI restarted.

kolyshkin:
@AkihiroSuda @cyphar @mrunalp PTAL

AkihiroSuda (dismissed):

mrunalp:
Needs rebase. Looks fine otherwise.

kolyshkin:
Rebased (removed the first patch as it is no longer needed (#2835 implements closing the log pipe).

mrunalp (approved):

AkihiroSuda (approved):

PR #2806, opened 2021-02-10
README: add gha badges

kolyshkin:
Note that validate currently fails on master -- this is tracked
in https://github.com/opencontainers/runc/issues/2627.


AkihiroSuda:
I think we want to fix the lint before adding this badge

kolyshkin:
> I think we want to fix the lint before adding this badge

I'm on the verge -- from one side, yes, from another side, having a red badge gives more motivation to fix things. IDK

AkihiroSuda (approved):

mrunalp (approved):

PR #2808, opened 2021-02-16
ci/gha: bump go 1.16-rc1 -> 1.16.x; fix gha ci

kolyshkin:
As the final go 1.16 is released, rc1 is no longer available.

This fixes our CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.14.x, 1.15.x, 1.16.0-rc1]
+        go-version: [1.14.x, 1.15.x, 1.16.x]
         rootless: ["rootless", ""]
 
     steps:


kolyshkin:
Can we merge this one quick please? Currently the CI is broken as go-1.16-rc1 is removed from downloads.

mrunalp (approved):

mrunalp:
@AkihiroSuda ptal

AkihiroSuda (approved):

kolyshkin:
Marked go 1.16.x jobs as required

cyphar:
Can we switch to `1.x` and `tip` (or `beta`) rather than having to update the GHA with each Go release?

thaJeztah:
> We might want to keep some older versions in the GHA

Those older versions would have to be updated if a new Go version is released (assuming support is `"latest + previous"`), or did you want to run `"latest" + <some fixed minimal version>"` ?

I'm slightly concerned that `1.x` makes it somewhat unpredictable what version a pull request ran on (and adding support for a new Go version may still need to be a conscious decision)

cyphar:
IMHO we should be testing whether CI fails if you build HEAD with a new Go version, we don't really have the concept of a "supported Go version" in runc in the same way that exists in Docker. The issue is we probably want to have latest + previous (maybe + one fixed minimal version), but I think having `1.x` strictly improves the situation even if we end up regularly updating the GHA configuration -- at the very least we immediately know if a new Go version breaks runc.

PR #2810, opened 2021-02-18
Retry when write file busy

### LINKED ISSUE #2736: Creating a pod has a probability of failure
``` yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
  namespace: default
spec:
  replicas: 10
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - args:
        - -c
        - sleep inf
        command:
        - /bin/sh
        image: busybox
        imagePullPolicy: IfNotPresent
        name: test
        ports:
        - containerPort: 4444
          name: test
          protocol: TCP
        resources:
          limits:
            cpu: 10m
            memory: 10Mi
          requests:
            cpu: 10m
            memory: 10Mi
```

image
image

image

This will not happen if remove the memory limit or change it to 100MI

It's the fault here
https://github.com/opencontainers/runc/blob/dbbe7e60c7b8c18cdf8a6e87d2dddcbfb24d7c90/libcontainer/cgroups/fscommon/fscommon.go#L42-L51

Maybe it's missing a condition check? `unix.EBUSY`
``` golang
func retryingWriteFile(filename string, data []byte, perm os.FileMode) error {
	for {
		err := ioutil.WriteFile(filename, data, perm)
		if errors.Is(err, unix.EINTR) || errors.Is(err, unix.EBUSY) {
			logrus.Infof("interrupted while writing %s to %s", string(data), filename)
			continue
		}
		return err
	}
}
```

May I ask if this is a known problem? if not, I will try to fix it

wzshiming:
Fixes #2736

### DIFF
--- a/libcontainer/cgroups/fscommon/fscommon.go
+++ b/libcontainer/cgroups/fscommon/fscommon.go
@@ -42,7 +42,7 @@ func ReadFile(dir, file string) (string, error) {
 func retryingWriteFile(fd *os.File, data string) error {
 	for {
 		_, err := fd.Write([]byte(data))
-		if errors.Is(err, unix.EINTR) {
+		if errors.Is(err, unix.EINTR) || errors.Is(err, unix.EBUSY) {
 			logrus.Infof("interrupted while writing %s to %s", data, fd.Name())
 			continue
 		}


libcontainer/cgroups/fscommon/fscommon.go
```diff
@@ -42,7 +42,7 @@ func ReadFile(dir, file string) (string, error) {
 func retryingWriteFile(fd *os.File, data string) error {
 	for {
 		_, err := fd.Write([]byte(data))
-		if errors.Is(err, unix.EINTR) {
+		if errors.Is(err, unix.EINTR) || errors.Is(err, unix.EBUSY) {
 			logrus.Infof("interrupted while writing %s to %s", data, fd.Name())
```
kolyshkin, line 46: The log message is obviously not right for EBUSY case.

kolyshkin:
@wzshiming thank you for your PR.

As discussed in https://github.com/opencontainers/runc/issues/2736, we don't think retrying on EBUSY is the adequate solution. This is a configuration error (memory limit is too tight), and the best we can do in such a case is provide a better error message (looking into that).

Are you able to reproduce the issue with adequate memory limit set?

cyphar:
In addition, `-EBUSY` is not always a transient error (unlike `-EINTR`) -- it's entirely possible that this will loop infinitely for some files.

kolyshkin:
I am working on a more specific fix (to clarify the error message and provide more diagnostics in case of error -- WIP is in https://github.com/opencontainers/runc/pull/2812)

PR #2812, opened 2021-02-19
Better handle hitting the memory limit

kolyshkin:
This improves runc error reporting in cases where memory limit is set too low.

NOTE that while this is clearly an enhancement I want this in rc94 because
 - it does not change runc behavior other than error text;
 - it helps a lot with some issues reported by customers.

## 1. Improve EBUSY handling on setting cgroup memory limit ##

 On cgroup v1 + fs driver, setting memory.limit_in_bytes results in EBUSY. This is reported as

> ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: process_linux.go:458: setting cgroup config for procHooks process caused: failed to write "1000": write /sys/fs/cgroup/memory/user.slice/user-1000.slice/session-1011.scope/xe3/memory.limit_in_bytes: device or resource busy

To decipher the error a user needs to know that EBUSY is returned by the kernel when the limit being set is too low. In addition, it would be handy to know what the current usage is.

Handle EBUSY and report this:

> ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: process_linux.go:458: setting cgroup config for procHooks process caused: unable to set memory limit to 1000 (current usage: 1204224, peak usage: 2936832)

Related to https://github.com/opencontainers/runc/issues/2736

## 2. Check for OOM kill on container start ##

When a container fails to start due to memory limit set too low, and container init being OOM-killed,
error messages returned by runc are semi-random and rather cryptic. Here are a few examples
(truncated to remove the common prefix for clarity):

 - `process_linux.go:348: copying bootstrap data to pipe caused: write init-p: broken pipe` (cgroup v1 + systemd driver)
 - `process_linux.go:352: getting the final child's pid from pipe caused: EOF` (cgroup v1 + systemd driver)
 - `process_linux.go:495: container init caused: read init-p: connection reset by peer` (cgroup v2)
 - `process_linux.go:484: writing syncT 'resume' caused: write init-p: broken pipe` (cgroup v2)

On container start error path, add a check if OOM kill has happened, and report it instead of the original (cryptic) error:

> ERRO[0000] container_linux.go:367: starting container process caused: container init was OOM-killed (memory limit too low?)

(or, if `--debug` is set, also provide the original error):

> ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:343: container init was OOM-killed (memory limit too low?) caused: process_linux.go:520: container init caused: process_linux.go:509: writing syncT 'resume' caused: write init-p: broken pipe 

## 3. Check for OOM kill on container exec ##

Same as above, with some nuances:

1. The container is already running and OOM kill counter might not be
       zero.  This is why we have to read the counter before exec and after
       it failed.
    
2. An unrelated OOM kill event might occur in parallel with our exec
       (and I see no way to find out which process was killed, except to
       parse kernel logs which seems excessive and not very reliable).
       This is why we report _possible_ OOM kill.

The error message changed from 

> ERRO[0000] exec failed: container_linux.go:367: starting container process caused: read init-p: connection reset by peer

to

> ERRO[0000] exec failed: container_linux.go:367: starting container process caused: process_linux.go:105: possibly OOM-killed caused: read init-p: connection reset by peer 

## 4. Some minor improvements and refactoring along the way. ##

## TODO ##
 - look into why runc init memory current and max differs that much
 - ~~look into why systemd v1 driver does not fail to set the limits like fs driver does~~ see #2813 / #2814.

### DIFF
--- a/libcontainer/cgroups/cgroups.go
+++ b/libcontainer/cgroups/cgroups.go
@@ -48,4 +48,7 @@ type Manager interface {
 
 	// Whether the cgroup path exists or not
 	Exists() bool
+
+	// OOMKillCount reports OOM kill count for the cgroup.
+	OOMKillCount() (uint64, error)
 }
--- a/libcontainer/cgroups/fscommon/utils.go
+++ b/libcontainer/cgroups/fscommon/utils.go
@@ -35,22 +35,22 @@ func ParseUint(s string, base, bitSize int) (uint64, error) {
 	return value, nil
 }
 
-// GetCgroupParamKeyValue parses a space-separated "name value" kind of cgroup
-// parameter and returns its components. For example, "io_service_bytes 1234"
-// will return as "io_service_bytes", 1234.
-func GetCgroupParamKeyValue(t string) (string, uint64, error) {
-	parts := strings.Fields(t)
-	switch len(parts) {
-	case 2:
-		value, err := ParseUint(parts[1], 10, 64)
-		if err != nil {
-			return "", 0, fmt.Errorf("unable to convert to uint64: %v", err)
-		}
-
-		return parts[0], value, nil
-	default:
-		return "", 0, ErrNotValidFormat
+// ParseKeyValue parses a space-separated "name value" kind of cgroup
+// parameter and returns its key as a string, and its value as uint64
+// (ParseUint is used to convert the value). For example,
+// "io_service_bytes 1234" will be returned as "io_service_bytes", 1234.
+func ParseKeyValue(t string) (string, uint64, error) {
+	parts := strings.SplitN(t, " ", 3)
+	if len(parts) != 2 {
+		return "", 0, fmt.Errorf("line %q is not in key value format", t)
 	}
+
+	value, err := ParseUint(parts[1], 10, 64)
+	if err != nil {
+		return "", 0, fmt.Errorf("unable to convert to uint64: %v", err)
+	}
+
+	return parts[0], value, nil
 }
 
 // GetCgroupParamUint reads a single uint64 value from the specified cgroup file.
--- a/libcontainer/cgroups/fscommon/utils.go
+++ b/libcontainer/cgroups/fscommon/utils.go
@@ -53,6 +53,26 @@ func ParseKeyValue(t string) (string, uint64, error) {
 	return parts[0], value, nil
 }
 
+// GetValueByKey reads a key-value pairs from the specified cgroup file,
+// and returns a value of the specified key. ParseUint is used for value
+// conversion.
+func GetValueByKey(path, file, key string) (uint64, error) {
+	content, err := ReadFile(path, file)
+	if err != nil {
+		return 0, err
+	}
+
+	lines := strings.Split(string(content), "\n")
+	for _, line := range lines {
+		arr := strings.Split(line, " ")
+		if len(arr) == 2 && arr[0] == key {
+			return ParseUint(arr[1], 10, 64)
+		}
+	}
+
+	return 0, nil
+}
+
 // GetCgroupParamUint reads a single uint64 value from the specified cgroup file.
 // If the value read is "max", the math.MaxUint64 is returned.
 func GetCgroupParamUint(path, file string) (uint64, error) {
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -97,7 +97,7 @@ func (s *CpuGroup) GetStats(path string, stats *cgroups.Stats) error {
 
 	sc := bufio.NewScanner(f)
 	for sc.Scan() {
-		t, v, err := fscommon.GetCgroupParamKeyValue(sc.Text())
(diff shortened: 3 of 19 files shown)


kolyshkin:
Before:

> ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: process_linux.go:458: setting cgroup config for procHooks process caused: failed to write "1000": write /sys/fs/cgroup/memory/user.slice/user-1000.slice/session-1011.scope/xe3/memory.limit_in_bytes: device or resource busy

After:
> ERRO[0000] container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: process_linux.go:458: setting cgroup config for procHooks process caused: unable to set memory limit to 1000 (current usage: 1204224, peak usage: 2936832)

This at least gives some info about what's wrong as well as a hint about the lowest possible limit.

This also sheds some light about what is going on in #2736 -- apparently runc init has some peak memory use upon the start, this is why retrying on EBUSY helps @wzshiming.

Yet I am not convinced this should be handled by a retry.

kolyshkin:
Failure on centos 7, seems like a random one (some rare race in old systemd?):

```
not ok 68 ps -e -x
(in test file tests/integration/ps.bats, line 53)
  `[ "$status" -eq 0 ]' failed
 runc spec (status=0):
 runc run -d --console-socket /tmp/console.sock test_busybox (status=1):
 time="2021-02-19T11:44:35Z" level=error msg="container_linux.go:367: starting container process
    caused: process_linux.go:502: container init
    caused: process_linux.go:465: setting cgroup config for procHooks process
    caused: Unit runc-test_busybox.scope is not loaded."
 ```

**Comment by kolyshkin** [comment]
@wzshiming can you test this PR on your setup to reproduce https://github.com/opencontainers/runc/issues/2736 and copy-paste the runc errors here? It will help to understand what is going on in your case.

**Comment by kolyshkin** [comment]
> why systemd v1 driver does not fail to set the limits like fs driver does;
￼￼
This happens because we do
 - cgroup Manager.Apply()
 - start init
 - cgroup Manager.Set()

Now, the difference between fs and systemd managers are:
 - fs manager Apply() merely creates cgroups (it and does not set any limits)
 - systemd manager Apply() creates the systemd with all properties already set, i.e. it sets limits (those that systemd understands)

Because of this difference, when fs driver is used, we can actually start a container init (which gets OOM-killed a moment later). When systemd driver is used, we can not start a container as the limits are too tight.

I think we should somehow unify the behavior. The two options are
1. Call Set() before container is started.
2. Make systemd's Apply() create a unit with no limits, only adding those later during Set().

**Comment by kolyshkin** [comment]
No longer a wip. `runc init` memory usage analysis were not done, but it won't change anything in this code.

**Comment by kolyshkin** [comment]
Moved to draft to add OOM handling in runc exec. Added, tested, ready for the prime time!

File: libcontainer/cgroups/cgroups.go [hunk 1]
```diff
@@ -48,4 +48,7 @@ type Manager interface {
 
 	// Whether the cgroup path exists or not
 	Exists() bool
+
+	// OOMKill reports OOM kill count for the cgroup.
+	OOMKill() (uint64, error)
```
mrunalp, line 53: Maybe rename this to `OOMKillCount()` or similar?

kolyshkin: Right, otherwise it looks like an action :)

kolyshkin: fixed

mrunalp (approved):

libcontainer/cgroups/fscommon/utils.go
```diff
@@ -35,22 +35,42 @@ func ParseUint(s string, base, bitSize int) (uint64, error) {
...
-		value, err := ParseUint(parts[1], 10, 64)
-		if err != nil {
-			return "", 0, fmt.Errorf("unable to convert to uint64: %v", err)
-		}
+// ParseKeyValue parses a space-separated "name value" kind of cgroup
+// parameter and returns its key as a string, and its value as uint64
+// (ParseUint is used to convert the value). For example,
+// "io_service_bytes 1234" will be returned as "io_service_bytes", 1234.
+func ParseKeyValue(t string) (string, uint64, error) {
+	parts := strings.SplitN(t, " ", 3)
```
AkihiroSuda, line 43: SplitN should be called with 2, not 3?

kolyshkin: Specifying 3 instead of 2 here helps to catch a situation with an extra space (if there is one, we error out on `len(parts) != 2`.

Practically it does not matter much, as `ParseUint` will also fail given an extra a space, but I'd rather fail earlier.

kolyshkin: IOW
```go
parts := strings.SplitN(t, " ", 3)
if len(parts) != 2 {
   // return error
}
```

is a way to split t into two parts, while also checking there's no more than two parts.

AkihiroSuda (approved):

Issue #2813, opened 2021-02-20
Fix discrepancy between systemd and fs cgroup managers


kolyshkin:
All cgroup managers implement `Apply()` and `Set()` methods:
 - **Apply** is used to create a cgroup (and, in case of systemd, a systemd unit) and/or put a PID into the cgroup (and unit);
 - **Set** is used to set various cgroup resources and limits;
 
fs/fs2 cgroup managers implement the functionality as per the description above.

systemd v1/v2 managers are somewhat weird. They set most of cgroup limits (those that can be projected to systemd unit properties) in Apply(), and they set all cgroup limits in Set -- first indirectly via systemd properties -- same as in Apply, then via cgroupfs (actually backed by fs manager)).

> A bit of recent history -- before #2287/#2343 systemd managers used to set unit properties (incl. resources) in Apply, and cgroup "raw" properties in Set. After that PR, they set properties in both Apply and Set.

To reiterate, systemd managers are peculiar since:
 - they set _some_ properties in Apply() (fs managers do not);
 - they set _all_ properties again in set.

Since runc calls both `Apply()` and `Set()`, this is not really a problem, except for some curious side effects (such as https://github.com/opencontainers/runc/pull/2812#issuecomment-782487341). This might be worse for other users (need to be looked at).

The proposed solution is not set and resources in Apply (the actual fix is surprisingly tidy -- see PR #2813).

TODO:
 - look at how crun does it (maybe @giuseppe can shed some light)
 - look at how kubernetes uses cgroup manager's Set and Apply.

odinuge:
+1 on this one!

I have been working with this from the k8s side, so I can help with insight if needed. Anything special you need?

I also think we should strive to adhere to the systemd delegation a bit better as well (eg we don't need to write to `cpu.shares/cpu.weight` when systemd already will do that): https://systemd.io/CGROUP_DELEGATION/

I think crun only create the slice/scope, and has a separate sub-cgroup (for v2 at least) where they enforce limits. I do however the runc approach is better, since we actually use the systemd-api.

odinuge:
> look at how kubernetes uses cgroup manager's Set and Apply.

The changes you are talking about here make sense for k8s, since we always use Set after Apply in order to enforce limits. The systemd-part is however currently broken/has never really made sense, but that should be fixed with: https://github.com/kubernetes/kubernetes/pull/98374 (feel free to look at that PR and/or review if you have any thoughts).

kolyshkin:
[...]

PR #2814, opened 2021-02-20
libct/cgroups/systemd: don't set limits in Apply

kolyshkin:
All cgroup managers has Apply() and Set() methods:

 - Apply is used to create a cgroup (and, in case of systemd,
   a systemd unit) and/or put a PID into the cgroup (and unit);

 - Set is used to set various cgroup resources and limits.

The fs/fs2 cgroup manager implements the functionality as described above.

The systemd v1/v2 manager deviate -- it sets *most* of cgroup limits
(those that can be projected to systemd unit properties) in Apply(),
and then again *all* cgroup limits in Set (first indirectly via systemd
properties -- same as in Apply, then via cgroupfs).

This commit removes setting the cgroup limits from Apply,
so now the systemd manager behaves the same way as the fs manager.

Fixes: #2813

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -86,7 +86,6 @@ func genV1ResourcesProperties(c *configs.Cgroup, conn *systemdDbus.Conn) ([]syst
 
 	if r.PidsLimit > 0 || r.PidsLimit == -1 {
 		properties = append(properties,
-			newProp("TasksAccounting", true),
 			newProp("TasksMax", uint64(r.PidsLimit)))
 	}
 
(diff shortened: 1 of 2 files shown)


kolyshkin:
I wanted to fix it half a year ago but did not quite understand the purpose of Apply and Set at the time.

odinuge (approved):
/lgtm

I like this difference between Apply and Set! I think the reason for the previous behavior was that the slice/scope was created in Apply, and never updated in Set (only by writing to cgroup, aka. not correct). Now, i think this makes more sense.

kolyshkin:
@AkihiroSuda I would not call this a refactor -- this is a bug fix (#2813), although I agree the bug is kinda subtle.

dqminh (approved):

mrunalp (approved):

PR #2816, opened 2021-02-22
Fix memory stats for cache in fs2

odinuge:
In cgroup v2, the "cache" value from cgroup v1 is called "file" in v2.
There are no values called "cache" in v2.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/memcontrol.c?id=31caf8b2a847214be856f843e251fc2ed2cd1075#n1521


kolyshkin (approved):
Took a quick look into the kernel -- looks like this is it.

LGTM

mrunalp (approved):

PR #2818, opened 2021-02-24
Fix cgroup2 mount for rootless case

kolyshkin:
In case of rootless, cgroup2 mount is not possible (see https://github.com/opencontainers/runc/issues/2158
for more details), so since commit 9c81440fb5a7 runc bind-mounts the whole
/sys/fs/cgroup into container.

Problem is, if cgroupns is enabled, /sys/fs/cgroup inside the container
is supposed to show the cgroup files for this cgroup, not the root one.

The fix is to pass through and use the cgroup path in case cgroup2
mount failed, cgroupns is enabled, and the path is non-empty.
    
Before:
    
            $ ./runc run aaa
            # find /sys/fs/cgroup/ -type d
            /sys/fs/cgroup
            /sys/fs/cgroup/user.slice
            /sys/fs/cgroup/user.slice/user-1000.slice
            /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service
            ...
            # ls -l /sys/fs/cgroup/cgroup.controllers
            -r--r--r--    1 nobody   nogroup          0 Feb 24 02:22 /sys/fs/cgroup/cgroup.controllers
            # wc -w /sys/fs/cgroup/cgroup.procs
            142 /sys/fs/cgroup/cgroup.procs
            # cat /sys/fs/cgroup/memory.current
            cat: can't open '/sys/fs/cgroup/memory.current': No such file or directory
    
After:
    
            # find /sys/fs/cgroup/ -type d
            /sys/fs/cgroup/
            # ls -l /sys/fs/cgroup/cgroup.controllers
            -r--r--r--    1 root     root             0 Feb 24 02:43 /sys/fs/cgroup/cgroup.controllers
            # wc -w /sys/fs/cgroup/cgroup.procs
            2 /sys/fs/cgroup/cgroup.procs
            # cat /sys/fs/cgroup/memory.current
            577536

PS found while trying to enable rootless for the test case from https://github.com/opencontainers/runc/pull/2786

## Suggested changelog entry
 * fixed rootless containers to not expose host's /sys/fs/cgroup hierarchy

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -70,6 +70,7 @@ type initConfig struct {
 	RootlessEUID     bool                  `json:"rootless_euid,omitempty"`
 	RootlessCgroups  bool                  `json:"rootless_cgroups,omitempty"`
 	SpecState        *specs.State          `json:"spec_state,omitempty"`
+	Cgroup2Path      string                `json:"cgroup2_path,omitempty"`
 }
 
 type initer interface {
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -29,6 +29,12 @@ import (
 
 const defaultMountFlags = unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NODEV
 
+type mountConfig struct {
+	root     string
+	label    string
+	cgroupns bool
+}
+
 // needsSetupDev returns true if /dev needs to be set up.
 func needsSetupDev(config *configs.Config) bool {
 	for _, m := range config.Mounts {
@@ -48,7 +54,11 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 		return newSystemErrorWithCause(err, "preparing rootfs")
 	}
 
-	hasCgroupns := config.Namespaces.Contains(configs.NEWCGROUP)
+	mountConfig := &mountConfig{
+		root:     config.Rootfs,
+		label:    config.MountLabel,
+		cgroupns: config.Namespaces.Contains(configs.NEWCGROUP),
+	}
 	setupDev := needsSetupDev(config)
 	for _, m := range config.Mounts {
 		for _, precmd := range m.PremountCmds {
@@ -56,7 +66,7 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig) (err error) {
 				return newSystemErrorWithCause(err, "running premount command")
 			}
 		}
-		if err := mountToRootfs(m, config.Rootfs, config.MountLabel, hasCgroupns); err != nil {
+		if err := mountToRootfs(m, mountConfig); err != nil {
 			return newSystemErrorWithCausef(err, "mounting %q to rootfs at %q", m.Source, m.Destination)
 		}
 
@@ -222,7 +232,7 @@ func prepareBindMount(m *configs.Mount, rootfs string) error {
 	return nil
 }
 
-func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns bool) error {
+func mountCgroupV1(m *configs.Mount, c *mountConfig) error {
 	binds, err := getCgroupMounts(m)
 	if err != nil {
 		return err
@@ -242,12 +252,12 @@ func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 		Data:             "mode=755",
 		PropagationFlags: m.PropagationFlags,
 	}
-	if err := mountToRootfs(tmpfs, rootfs, mountLabel, enableCgroupns); err != nil {
+	if err := mountToRootfs(tmpfs, c); err != nil {
 		return err
 	}
 	for _, b := range binds {
-		if enableCgroupns {
-			subsystemPath := filepath.Join(rootfs, b.Destination)
+		if c.cgroupns {
+			subsystemPath := filepath.Join(c.root, b.Destination)
 			if err := os.MkdirAll(subsystemPath, 0755); err != nil {
 				return err
 			}
@@ -266,7 +276,7 @@ func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 				return err
 			}
 		} else {
-			if err := mountToRootfs(b, rootfs, mountLabel, enableCgroupns); err != nil {
+			if err := mountToRootfs(b, c); err != nil {
 				return err
 			}
 		}
@@ -276,7 +286,7 @@ func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 			// symlink(2) is very dumb, it will just shove the path into
 			// the link and doesn't do any checks or relative path
 			// conversion. Also, don't error out if the cgroup already exists.
-			if err := os.Symlink(mc, filepath.Join(rootfs, m.Destination, ss)); err != nil && !os.IsExist(err) {
+			if err := os.Symlink(mc, filepath.Join(c.root, m.Destination, ss)); err != nil && !os.IsExist(err) {
 				return err
 			}
 		}
@@ -284,8 +294,8 @@ func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 	return nil
 }
 
-func mountCgroupV2(m *configs.Mount, rootfs, mountLabel string, enableCgroupns bool) error {
-	cgroupPath, err := securejoin.SecureJoin(rootfs, m.Destination)
+func mountCgroupV2(m *configs.Mount, c *mountConfig) error {
+	cgroupPath, err := securejoin.SecureJoin(c.root, m.Destination)
 	if err != nil {
 		return err
 	}
@@ -302,10 +312,10 @@ func mountCgroupV2(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 	return nil
 }
 
-func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns bool) error {
-	var (
-		dest = m.Destination
-	)
+func mountToRootfs(m *configs.Mount, c *mountConfig) error {
+	rootfs := c.root
+	mountLabel := c.label
+	dest := m.Destination
 	if !strings.HasPrefix(dest, rootfs) {
 		dest = filepath.Join(rootfs, dest)
 	}
@@ -424,9 +434,9 @@ func mountToRootfs(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
 		}
 	case "cgroup":
 		if cgroups.IsCgroup2UnifiedMode() {
-			return mountCgroupV2(m, rootfs, mountLabel, enableCgroupns)
+			return mountCgroupV2(m, c)
 		}
-		return mountCgroupV1(m, rootfs, mountLabel, enableCgroupns)
+		return mountCgroupV1(m, c)
 	default:
 		// ensure that the destination of the mount is resolved of symlinks at mount time because
 		// any previous mounts can invalidate the next mount's destination.
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -17,6 +17,7 @@ import (
 	"github.com/moby/sys/mountinfo"
 	"github.com/mrunalp/fileutils"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/devices"
 	"github.com/opencontainers/runc/libcontainer/userns"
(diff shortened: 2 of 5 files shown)


kolyshkin:
Not sure if this is rc94 of post-1.0 material -- WDYT @cyphar @mrunalp @AkihiroSuda ?

AkihiroSuda:
Good fit for rc94 IMO

libcontainer/init_linux.go
```diff
@@ -70,6 +70,7 @@ type initConfig struct {
 	RootlessEUID     bool                  `json:"rootless_euid,omitempty"`
 	RootlessCgroups  bool                  `json:"rootless_cgroups,omitempty"`
 	SpecState        *specs.State          `json:"spec_state,omitempty"`
+	Cgroup2Path      string                `json:"cgroup2path,omitempty"`
```
AkihiroSuda, line 73: json name should be `cgroup2_path` ?

Also, this field needs a comment line to explain that this is the cg2 path of the runtime itself, not the cg2 path for the container being launched.

libcontainer/rootfs_linux.go
```diff
@@ -276,36 +289,43 @@ func mountCgroupV1(m *configs.Mount, rootfs, mountLabel string, enableCgroupns b
...
 				return err
 			}
 		}
 	}
 	return nil
 }
 
-func mountCgroupV2(m *configs.Mount, rootfs, mountLabel string, enableCgroupns bool) error {
-	cgroupPath, err := securejoin.SecureJoin(rootfs, m.Destination)
+func mountCgroupV2(m *configs.Mount, c *mountConfig) error {
```
AkihiroSuda, line 300: Can we have bats tests for this PR?

libcontainer/init_linux.go
kolyshkin: > json name should be cgroup2_path ?

fixed

> to explain that this is the cg2 path of the runtime itself, not the cg2 path for the container being launched

Can you explain in more details? To me, this is exactly the cgroup2 path for container being launched, taken from the container's cgroup manager which is already initialized with the proper config etc.

AkihiroSuda: Ah, sorry I misread the code..

libcontainer/rootfs_linux.go
kolyshkin: Added a simple case checking that there are no extra directories under /sys/fs/cgroup, and enabled the blkio test from https://github.com/opencontainers/runc/pull/2786 for rootless case.

kolyshkin:
Checked that the added test case is failing without the fix:

```console
sudo ROOTLESS_TESTPATH=/cgroups.bats ./tests/rootless.sh
...
not ok 13 runc run (cgroupv2 mount inside container)
# (in test file tests/integration/cgroups.bats, line 296)
#   `[ "$(wc -l <<<"$output")" -eq 1 ]' failed
# 
# runc run -d --console-socket /tmp/bats-run-3210676/runc.4mlpW2/tty/sock test_cgroups_unified (status=0):
# 
# runc exec test_cgroups_unified find /sys/fs/cgroup/ -type d (status=0):
# /sys/fs/cgroup/
# /sys/fs/cgroup/user.slice
# /sys/fs/cgroup/user.slice/user-1000.slice
# /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service
# /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/vte-spawn-98b64ba6-eca3-41f7-b7e4-2f94a0613230.scope
# /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/gvfs-goa-volume-monitor.service
# /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/vte-spawn-a5d79e49-117d-4728-b261-ccba710e136d.scope
...
```

kolyshkin:
OK, I have totally missed the fact that rootless spec removes the /sys/fs/cgroup mount and therefore my code doesn't work (it worked locally because I have /sys/fs/cgroup mount in my spec).

This needs more work.

kolyshkin:
Failure in centos 7 rootful fs test:

> not ok 9 runc run (cgroup v1 + unified resources should fail)
> (from function `teardown_bundle' in file tests/integration/helpers.bash, line 481,
>  from function `teardown' in test file tests/integration/cgroups.bats, line 6)
>   `teardown_bundle' failed
> runc spec (status=0):
> 
> runc run -d --console-socket /tmp/bats-run-7694/runc.Ujskgr/tty/sock test_cgroups_unified (status=1):
> time="2021-02-28T19:06:43Z" level=warning msg="unable to get oom kill count" error="no directory specified for memory.oom_control"
> time="2021-02-28T19:06:43Z" level=error msg="unlinkat /tmp/bats-run-7694/runc.Ujskgr/state/test_cgroups_unified/runc.JCDSLu: device or resource busy"
> time="2021-02-28T19:06:43Z" level=error msg="container_linux.go:367: starting container process caused: process_linux.go:365: applying cgroup configuration for process caused: invalid configuration: cannot use unified on cgroup v1"
> load container test_cgroups_unified: container "test_cgroups_unified" does not exist
> rm: cannot remove ‘/tmp/bats-run-7694/runc.Ujskgr/state/test_cgroups_unified/runc.JCDSLu’: Device or resource busy

Looks unrelated but worrisome.

kolyshkin:
Fedora VM fails to run dnf transaction:

>    default: Errors during downloading metadata for repository 'fedora-modular':
>    default:   - Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-33&arch=x86_64 (IP: 209.132.190.2)
>    default: Error: Failed to download metadata for repo 'fedora-modular': Cannot prepare internal mirrorlist: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-33&arch=x86_64 (IP: 209.132.190.2)

CI restarted

kolyshkin:

AkihiroSuda:
I can't repro the issue with the current master (497cd0c96).

`<ENGINE> run --runtime=runc --cgroupns=private --memory 42m alpine cat /sys/fs/cgroup/memory.max` shows `44040192` as expected. (`<ENGINE>` = rootless docker, podman, nerdctl)

AkihiroSuda:
@kolyshkin PTAL↑

kolyshkin:
Need more time to look into this. Removed rc94 milestone for now.

kolyshkin:
Need more time to have another look into this; converting to draft for now.

kolyshkin:
@AkihiroSuda I guess the upper layer tools (podman/docker/nerdctl) generate the correct `config.json` for this case, while `runc spec` does not, this is why I am seeing this with bare runc and you don't see it with podman/docker/nerdctl.

There are other similar cases (for example, in case we use host pidns, the /proc mounts needs to be a bind mount, not the fstype=proc one), and ultimately the question is -- do we want runc to handle this (modify/interpret the mounts accordingly), or do we want the upper layers to provide the correct OCI runtime spec for these cases (the third way, I guess, is to overload `runc spec` with kludges).

I think this was discussed before and @crosbymichael was for literal spec interpretation (i.e. runc should not adopt mounts), and yet in some cases runc already does that (don't have an example from the top of my mind, will look it up later).

kolyshkin:
@AkihiroSuda is it crucial that we use upper-level tools to repro? I don't know how to correctly make them add userns into OCI spec (so we can hit the codepath added by https://github.com/opencontainers/runc/commit/9c81440fb5a7b81408b3de281e9851bc332993ec).

I can easily reproduce this with bare runc (see last two commits).

Here is a full manual repro (Fedora 33, cgroup v2):
```console
[kir@kir-rhat runc]$ mkdir foo
[kir@kir-rhat runc]$ cd foo
[kir@kir-rhat foo]$ mkdir rootfs
[kir@kir-rhat foo]$ eval `../tests/integration/get-images.sh`
[kir@kir-rhat foo]$ tar -C rootfs -xf $BUSYBOX_IMAGE
[kir@kir-rhat foo]$ ../runc spec --rootless
[kir@kir-rhat foo]$ ../runc run xx4
/ # ls -l /sys/fs/cgroup
total 0
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 cgroup.controllers
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 cgroup.max.depth
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 cgroup.max.descendants
-rw-rw-r--    1 nobody   nobody           0 Apr  6 22:02 cgroup.procs
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 cgroup.stat
-rw-r--r--    1 nobody   nobody           0 Apr 15 00:13 cgroup.subtree_control
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 cgroup.threads
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 cpu.pressure
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 cpu.stat
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 cpuset.cpus.effective
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 cpuset.mems.effective
drwxr-xr-x    2 nobody   nobody           0 Mar  5 00:53 init.scope
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 io.cost.model
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 io.cost.qos
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 io.pressure
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 io.stat
drwxr-xr-x    2 nobody   nobody           0 Apr  6 20:35 machine.slice
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 memory.numa_stat
-rw-r--r--    1 nobody   nobody           0 Apr 19 23:56 memory.pressure
-r--r--r--    1 nobody   nobody           0 Apr 19 23:56 memory.stat
drwxr-xr-x    3 nobody   nobody           0 Mar 19 21:03 pod_123.slice
drwxr-xr-x    2 nobody   nobody           0 Mar 10 23:20 sdet35
drwxr-xr-x  138 nobody   nobody           0 Apr 19 23:51 system.slice
drwxr-xr-x    2 nobody   nobody           0 Apr 14 23:31 test
drwxr-xr-x    3 nobody   nobody           0 Apr 15 17:45 user.slice
drwxr-xr-x    3 nobody   nobody           0 Mar  9 03:04 woo
/ #
```

We see all the cgroups, while we should only see ours.

Another repro is via our integration tests -- just modify this test case (from https://github.com/opencontainers/runc/pull/2786/commits/c3ffd2ef81aa67398e06c798d53d7f98cef39e9e#diff-df7d610d8608d0748bfcbaab39894c9e189211cddf54df5a955811287d199d5d) so it can be run as rootless (same as in one of this PR commits):

```diff
diff --git a/tests/integration/cgroups.bats b/tests/integration/cgroups.bats
index c5a927bd..2c49dfef 100644
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -138,7 +138,8 @@ function setup() {
 }
 
 @test "runc run (blkio weight)" {
-       requires root cgroups_v2
+       requires cgroups_v2
+       [[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
 
        set_cgroups_path
        update_config '.linux.resources.blockIO |= {"weight": 750}'
```

and run it in a rootless mode:
```console
[kir@kir-rhat runc]$ sudo make localrootlessintegration ROOTLESS_TESTPATH=/cgroups.bats
....
[02] run rootless tests ... (cgroup)
++ cat /sys/fs/cgroup/cgroup.controllers
+ for f in $(cat "$CGROUP_MOUNT/cgroup.controllers")
+ echo +cpuset
+ for f in $(cat "$CGROUP_MOUNT/cgroup.controllers")
+ echo +cpu
+ for f in $(cat "$CGROUP_MOUNT/cgroup.controllers")
+ echo +io
+ for f in $(cat "$CGROUP_MOUNT/cgroup.controllers")
+ echo +memory
+ for f in $(cat "$CGROUP_MOUNT/cgroup.controllers")
+ echo +hugetlb
+ for f in $(cat "$CGROUP_MOUNT/cgroup.controllers")
+ echo +pids
+ set +x
path: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/lib/snapd/snap/bin
1..10
ok 1 runc create (no limits + no cgrouppath + no permission) succeeds
ok 2 runc create (rootless + no limits + cgrouppath + no permission) fails with permission error # skip test requires rootless_no_cgroup
ok 3 runc create (rootless + limits + no cgrouppath + no permission) fails with informative error # skip test requires rootless_no_cgroup
ok 4 runc create (limits + cgrouppath + permission on the cgroup dir) succeeds
ok 5 runc exec (limits + cgrouppath + permission on the cgroup dir) succeeds
ok 6 runc exec (cgroup v2 + init process in non-root cgroup) succeeds # skip test requires root
ok 7 runc run (cgroup v1 + unified resources should fail) # skip test requires root
not ok 8 runc run (blkio weight)
# (in test file tests/integration/cgroups.bats, line 155)
#   `[ "$output" = 'default 7475' ]' failed
# 
# runc run -d --console-socket /tmp/bats-run-1256838/runc.ZUIMwn/tty/sock test_cgroups_unified (status=0):
# 
# runc exec test_cgroups_unified sh -c cat /sys/fs/cgroup/io.bfq.weight (status=1):
# cat: can't open '/sys/fs/cgroup/io.bfq.weight': No such file or directory
# runc exec test_cgroups_unified sh -c cat /sys/fs/cgroup/io.weight (status=1):
# cat: can't open '/sys/fs/cgroup/io.weight': No such file or directory
```

This PR fixes the above test case, and also adds a separate test case to make sure we don't see the full cgroup hierarchy.

libcontainer/specconv/example.go
```diff
@@ -200,8 +200,14 @@ func ToRootless(spec *specs.Spec) {
 	// Fix up mounts.
 	var mounts []specs.Mount
 	for _, mount := range spec.Mounts {
-		// Ignore all mounts that are under /sys.
-		if strings.HasPrefix(mount.Destination, "/sys") {
+		// Replace the /sys mount with an rbind.
+		if mount.Destination == "/sys" {
```
AkihiroSuda, line 204: Needs filepath.Clean ?

kolyshkin: According to https://github.com/opencontainers/runtime-spec/blob/master/config.md, `mount.Destination` MUST be an absolute path.

We should validate that the destination is absolute elsewhere. Alas we do not do that. Will open a PR soon.

AkihiroSuda (approved):

libcontainer/specconv/example.go
kolyshkin: Filed https://github.com/opencontainers/runc/pull/2917 but realized we are talking about clean path, not absolute path here.

Fixed.

AkihiroSuda (approved):

mrunalp (approved):

PR #2820, opened 2021-02-24
fs2: fallback to setting io.weight if io.bfq.weight

dqminh:
if bfq is not loaded, then io.bfq.weight is not available. io.weight
should always be available and is the next best equivalent thing.

### DIFF
--- a/libcontainer/cgroups/fs2/io.go
+++ b/libcontainer/cgroups/fs2/io.go
@@ -30,7 +30,15 @@ func setIo(dirPath string, cgroup *configs.Cgroup) error {
 		filename := "io.bfq.weight"
 		if err := fscommon.WriteFile(dirPath, filename,
 			strconv.FormatUint(uint64(cgroup.Resources.BlkioWeight), 10)); err != nil {
-			return err
+			// if io.bfq.weight does not exist, then bfq module is not loaded.
+			// Fallback to use io.weight with a conversion scheme
+			if !os.IsNotExist(err) {
+				return err
+			}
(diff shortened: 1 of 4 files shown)


libcontainer/cgroups/fs2/io.go
```diff
@@ -30,7 +30,16 @@ func setIo(dirPath string, cgroup *configs.Cgroup) error {
 		filename := "io.bfq.weight"
 		if err := fscommon.WriteFile(dirPath, filename,
 			strconv.FormatUint(uint64(cgroup.Resources.BlkioWeight), 10)); err != nil {
-			return err
+			// if io.bfq.weight does not exist, then bfq module is not loaded.
+			// Fallback to use io.weight with a conversion scheme
+			if os.IsNotExist(err) {
```
kolyshkin, line 35: Wonder if it works in case openat2 is used; let me check...

Yes, it works. Good.

kolyshkin, line 35: I'd rather do
```go
if !os.IsNotExist(err) {
   return err
}
// Fallback to use io.weight.
...
```

just for readability

kolyshkin: @dqminh PTAL ^^^

dqminh: thanks, it's done.

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

PR #2825, opened 2021-02-26
proposal: add --keep to runc run

### LINKED ISSUE #2817: proposal: add --no-delete to runc run
`runc run` is much more simple to use than `create/start` async commands that are almost impossible to use safely without introducing a special shim-like binary to deal with `SIGCHLD` reaper issues.

One of the problems though is that when using `run` calling process can't get the full stats of the process because cgroups are already deleted when it returns. Could we consider adding `--no-delete` option that leaves cgroups and other resources in place so the caller can analyze them and then call `runc delete` to do the manual cleanup.

lifubang:
To close #2817

As described in #2817, for some times, we need to analyze the container's resource usage, so we expect runc does not delete the container after it exits.

## Changelog entry
 (by @kolyshkin)
```
New features:
* runc run: new `--keep` option to skip removal exited containers artefacts.
  This might be useful to check the state (e.g. of cgroup controllers) after
  the container has￼exited. (#2817)
```

### DIFF
--- a/run.go
+++ b/run.go
@@ -40,6 +40,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "detach, d",
 			Usage: "detach from the container's process",
 		},
+		cli.BoolFlag{
+			Name:  "keep",
+			Usage: "do not delete the container after it exits",
+		},
 		cli.StringFlag{
 			Name:  "pid-file",
 			Value: "",
--- a/tests/integration/run.bats
+++ b/tests/integration/run.bats
@@ -1,0 +1,59 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function setup() {
+	setup_hello
+}
+
+function teardown() {
+	teardown_bundle
+}
+
+@test "runc run" {
+	runc run test_hello
+	[ "$status" -eq 0 ]
+
+	runc state test_hello
+	[ "$status" -ne 0 ]
+}
+
+@test "runc run --keep" {
+	runc run --keep test_run_keep
+	[ "$status" -eq 0 ]
+
+	testcontainer test_run_keep stopped
+
+	runc state test_run_keep
+	[ "$status" -eq 0 ]
+
+	runc delete test_run_keep
+
+	runc state test_run_keep
+	[ "$status" -ne 0 ]
+}
+
+@test "runc run --keep (check cgroup exists)" {
+	# for systemd driver, the unit's cgroup path will be auto removed if container's all processes exited
+	requires no_systemd
+
+	[[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
+
+	set_cgroups_path
+
+	runc run --keep test_run_keep
+	[ "$status" -eq 0 ]
+
+	testcontainer test_run_keep stopped
+
+	runc state test_run_keep
+	[ "$status" -eq 0 ]
+
+	# check that cgroup exists
+	check_cgroup_value "pids.max" "max"
+
+	runc delete test_run_keep
+
+	runc state test_run_keep
+	[ "$status" -ne 0 ]
+}
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -439,7 +439,7 @@ func startContainer(context *cli.Context, spec *specs.Spec, action CtAct, criuOp
 
 	r := &runner{
 		enableSubreaper: !context.Bool("no-subreaper"),
-		shouldDestroy:   true,
+		shouldDestroy:   !context.Bool("keep"),
 		container:       container,
 		listenFDs:       listenFDs,
 		notifySocket:    notifySocket,
--- a/man/runc-run.8.md
+++ b/man/runc-run.8.md
@@ -42,6 +42,12 @@ container to inherit the calling processes session key.
 : Pass _N_ additional file descriptors to the container (**stdio** +
 **$LISTEN_FDS** + _N_ in total). Default is **0**.
 
+**--keep**
+: Keep container's state directory and cgroup. This can be helpful if a user
+wants to check the state (e.g. of cgroup controllers) after the container has
+exited. If this option is used, a manual **runc delete** is needed afterwards
+to clean an exited container's artefacts.
+
 # SEE ALSO
 
 **runc**(8).


run.go
```diff
@@ -40,6 +40,10 @@ command(s) that get executed on start, edit the args parameter of the spec. See
 			Name:  "detach, d",
 			Usage: "detach from the container's process",
 		},
+		cli.BoolFlag{
+			Name:  "no-delete",
+			Usage: "don't auto destory the container after it stopped",
```
thaJeztah, line 45: small typo (s/destory/destroy/)☺️

```suggestion
			Usage: "don't auto-destroy the container after it stopped",
```

Or perhaps

```suggestion
			Usage: "do not delete the container after it exits",
```

thaJeztah (changes requested):

utils_linux.go
```diff
@@ -441,7 +441,7 @@ func startContainer(context *cli.Context, spec *specs.Spec, action CtAct, criuOp
 
 	r := &runner{
 		enableSubreaper: !context.Bool("no-subreaper"),
-		shouldDestroy:   true,
+		shouldDestroy:   !context.Bool("no-delete"),
```
thaJeztah, line 444: Not something new in this PR (so don't change) but it would be good to use the "default" values match the default behavior. (e.g. change the field to `keep` or `disableAutoDestroy`), so that the default for "boolean" values (`false`) also is the expected "default"

lifubang: Yes, it dose look strange to treat the default value `false` of `--no-delete` as auto-delete the container.
May be we can change `--no-delete` to `--disable-autodestroy`?

kolyshkin: I have suggested `--keep-cgroup` in #2817, but as we also have to keep the runc state directory, maybe just `--keep`?

tests/integration/run.bats
```diff
@@ -0,0 +1,39 @@
...
+function setup() {
+	setup_busybox
+}
+
+function teardown() {
+	teardown_bundle
+}
+
+@test "runc run" {
+	update_config ' .process.args |= ["ls", "/", "-alh"]'
```
kolyshkin, line 14: perhaps just `true`?

tests/integration/run.bats
```diff
@@ -0,0 +1,39 @@
...
+
+@test "runc run --no-delete" {
+	update_config ' .process.args |= ["ls", "/", "-alh"]'
+
+	runc run --no-delete test_no_delete
+	[ "$status" -eq 0 ]
+
+	testcontainer test_no_delete stopped
+
+	runc state test_no_delete
```
kolyshkin, line 31: Maybe also check that the cgroup is there?

utils_linux.go
thaJeztah: Yes, I was considering "--keep" as well; so sgtm

lifubang: SGTM too.
Welcome more better suggestions...

tonistiigi: The reason I suggested `--no-delete` was because it is consistent with the three existing `--no-*` flags in `runc run` and `delete` is consistent with the existing `runc delete` command that has the matching behavior. As a user, I don't care of course, and can use any flag.

kolyshkin: The existing `--no-*` flags are
```
   --no-subreaper            disable the use of the subreaper used to reap reparented processes
   --no-pivot                do not use pivot root to jail process inside rootfs.  This should be used whenever the rootfs is on top of a ramdisk
   --no-new-keyring          do not create a new session keyring for the container.  This will cause the container to inherit the calling processes session key
```

and in all 3 cases it's a `--no-NOUN`. There's also "--preserve-fds`...

To me `--keep` still looks like the best choice, the only issue is it's too generic (keep everything?).

tests/integration/run.bats
```diff
@@ -0,0 +1,55 @@
...
+
+	runc state test_run_keep
+	[ "$status" -eq 0 ]
+
+	# check cgroup path
+	if [ "x$CGROUP_UNIFIED" = "xyes" ]; then
+		check_cgroup_value "cgroup.freeze" "0"
+	else
+		check_cgroup_value "freezer.state" "THAWED"
+	fi
```
kolyshkin, line 48: Maybe it's easier to check something like pids limit:
```
        check_cgroup_value "pids.max" max
```

tests/integration/run.bats
```diff
@@ -0,0 +1,55 @@
...
+
+	runc run --keep test_run_keep
+	[ "$status" -eq 0 ]
+
+	testcontainer test_run_keep stopped
+
+	runc state test_run_keep
+	[ "$status" -eq 0 ]
+
+	# check cgroup path
```
kolyshkin, line 43: nit: `# check that cgroup exists`

tests/integration/run.bats
```diff
@@ -0,0 +1,67 @@
...
+
+	runc run --keep test_run_keep
+	[ "$status" -eq 0 ]
+
+	testcontainer test_run_keep stopped
+
+	runc state test_run_keep
+	[ "$status" -eq 0 ]
+
+	# delete test_busybox
```
kolyshkin, line 34: please remove this comment, it is of no use

kolyshkin:
> Because of systemd's unit garbage collection policy, it seems that there is no way to keep container's cgroup paths after it exit when we use systemd driver.

Hmm, do we have a place in the documentation to add this fact? I think we need to create one (working on it)

tonistiigi:
@kolyshkin What's the blocker in here?

utils_linux.go
tonistiigi: I guess I wouldn't consider `pivot` noun in this context (short for "pivot root"). Any flag is ok for me though.

cyphar: The difference is that for all of those `--no` settings, you're disabling a security feature and so the default should be false. Since the default for this option should also be false, it should be named something that makes sense in that context.

AkihiroSuda:
The blocker was v1.0.0 GA, but it will be released _really_ soon, finally 😅

kolyshkin:
Did another review round. As for the name for the option, I don't like `--keep` much, nor do I like `--no-delete`.

Since there are no better options, I guess either `--keep` or `--no-delete` is fine with me.

Setting the milestone to 1.1.0.

kolyshkin (commented):
LGTM, except we need to describe the new option in https://github.com/opencontainers/runc/blob/master/man/runc-run.8.md

I propose
```markdown
**--keep**
: Keep container's state directory and cgroup. This can be helpful if a user wants to check
the state (of cgroup etc.) after the container has exited. Note if this option is used, a manual
**runc delete** is needed afterwards.
```

kolyshkin:
I took a liberty to add a commit with a man page, and rebase.

CI is expected to fail because of #3050.

AkihiroSuda (approved):
LGTM if green

kolyshkin:
Rebased to kick the latest CI. Added labels. Provided changelog entry. Let's merge this already.

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

AkihiroSuda:
@cyphar PTAL

AkihiroSuda:
@cyphar LGTY?

AkihiroSuda:
> Merging is blocked
> Merging can be performed automatically once the requested changes are addressed.

Seems to need LGTM from @thaJeztah

thaJeztah (approved):
LGTM

PR #2827, opened 2021-02-26
libct: fix namespaces handling in the currentState

Iceber:
It feels redundant to handle `c.config.Namespaces` separately.


libcontainer/container_linux_test.go
```diff
@@ -297,19 +297,24 @@ func TestGetContainerState(t *testing.T) {
...
 			}
 		} else {
-			file := ""
+			var file string
 			switch ns.Type {
-			case configs.NEWNET:
-				file = "net"
 			case configs.NEWNS:
 				file = "mnt"
 			case configs.NEWPID:
```
kolyshkin, line 320: All this can probably be replaced with a call to `ns.GetPath()` (which was added later than this test).

libcontainer/container_linux_test.go
```diff
@@ -297,19 +297,24 @@ func TestGetContainerState(t *testing.T) {
...
 	}
 	for _, ns := range container.config.Namespaces {
-		path := state.NamespacePaths[ns.Type]
+		path, ok := state.NamespacePaths[ns.Type]
+		if !configs.IsNamespaceSupported(ns.Type) {
+			if ok {
+				t.Fatalf("namespace %s is not supported", ns.Type)
+			}
+			continue
+		}
```
kolyshkin, line 306: I would do something like this instead
```go
path := state.NamespacePaths[ns.Type]
if path == "" {
	t.Fatalf("expected non nil namespace path for %s", ns.Type)
}
if !configs.IsNamespaceSupported(ns.Type) {
	t.Fatalf("namespace %s not supported but its path is non-empty: %s", ns.Type, path)
}
```

This is basically the same check as you added, but with less changes to the existing code -- easier to review.

libcontainer/container_linux.go
```diff
@@ -1889,17 +1889,12 @@ func (c *linuxContainer) currentState() (*State, error) {
...
 		for _, nsType := range configs.NamespaceTypes() {
 			if !configs.IsNamespaceSupported(nsType) {
 				continue
 			}
-			if _, ok := state.NamespacePaths[nsType]; !ok {
-				ns := configs.Namespace{Type: nsType}
-				state.NamespacePaths[ns.Type] = ns.GetPath(pid)
-			}
+			ns := configs.Namespace{Type: nsType}
+			state.NamespacePaths[ns.Type] = ns.GetPath(pid)
```
kolyshkin, line 1897: I am not sure why this code was written the way it was -- I checked it comes from commits 556f798a19ecf and 78f5148c67498b, and I'm sure even @crosbymichael does not remember anymore.

kolyshkin:
This is kind of hard to review without any detailed description of what was done and why. @Iceber can you provide us with some more details about why you did this change, what this PR solves etc.

I might be wrong here, but from what I see, before this PR, if the namespace is set in config but not supported by the running kernel, this results in an error. With this PR, a non-supported namespace is ignored. I am not sure this is a correct behavior (and even if it is, it surely requires an explanation and a changelog entry).

Iceber:
@kolyshkin Thanks. 
I looked at it again, and the current change filters out the unsupported namespaces in the `config.Namespaces`, which causes unsupported namespace to not be checked in the `(*linuxContainer).orderNamespacePaths`, and then reports an error.

It doesn't feel right to do so. I'm going to modify this pr

libcontainer/container_linux.go
```diff
@@ -1893,13 +1893,15 @@ func (c *linuxContainer) currentState() (*State, error) {
 			state.NamespacePaths[ns.Type] = ns.GetPath(pid)
 		}
 		for _, nsType := range configs.NamespaceTypes() {
-			if !configs.IsNamespaceSupported(nsType) {
+			if _, ok := state.NamespacePaths[nsType]; ok {
```
Iceber, line 1896: Determines whether the namespace already exists first, which might make it clearer that existing namespaces will not be filtered

kolyshkin: This comment ^^^ belongs to commit description.

kolyshkin:
LGTM overall. Can you please split the test improvement into a separate commit (as it has nothing to do with the rest of the changes), and describe both commits?

Iceber:
@kolyshkin Split into two commits

kolyshkin (approved):
LGTM, and thanks!

PR #2829, opened 2021-03-02
libcontainer/factory*: adjust the file mode

KentaTada:
Related to #2625

BTW, I think we should not only modify the file mode from the old style to the latest style but also check whether the current implementation sets proper file permissions.
So, I will create the PRs each subsystem or codes and I may create the new issue to summarize this activity.


kolyshkin:
CI failure on Fedora is unrelated (https://github.com/opencontainers/runc/issues/2805); restarted CI

kolyshkin (approved):
This is purely cosmetical, but LGTM

mrunalp (approved):

PR #2834, opened 2021-03-05
CI: cache ~/.vagrant.d/boxes

AkihiroSuda:
For deflaking CI failures during `vagrant up`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,6 +81,13 @@ jobs:
     needs: [test]
     steps:
       - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


kolyshkin (approved):
I'm not sure if the hash is chosen correctly (say we add a package to `dnf install` in Vagrantfile, this doesn't change the fact that the cache should be invalidated), but maybe I'm missing something.

LGTM

AkihiroSuda:
@cyphar @mrunalp PTAL, this one is easy to review

mrunalp (approved):

PR #2835, opened 2021-03-06
Fix init log forwarding race

kolyshkin:
_(originally developed in #2487; decided to split for clarity)_

Sometimes debug.bats test cases are failing like this (issue #2756):

```
not ok 27 global --debug to --log --log-format 'json'
# (in test file tests/integration/debug.bats, line 77)
```

It happens more often while writing to disk.

The issue is caused by the fact that runc spawns log forwarding goroutine
(`ForwardLogs`) but does not wait for it to finish, resulting in missing
debug lines from nsexec.

`ForwardLogs` itself, though, never finishes, because it reads from a
reading side of a pipe whose writing side is not closed. This is
especially true in case of `runc create`, which spawns `runc init` and
exits; meanwhile `runc init` waits on exec fifo for some undetermined long
time before doing execve.

So, to fix the issue described above, we need to:

 1. Make runc create/run/exec wait for ForwardLogs to finish;

 2. Make runc init close its log pipe file descriptor (i.e.
    the one which value is passed in _LIBCONTAINER_LOGPIPE
    environment variable).

This is exactly what this commit does. In addition, it

 - adds logrus debug to late stages of runc init, and amends the
   integration tests to check for those messages;

 - adds `runc --debug exec` test cases, similar to those in `debug.bats`
   but for `runc exec` rather than `runc run`;

 - refactors libcontainer/logs unit tests to simplify and update for
   the new `ForwardLogs`.

PS I have to admit I still do not understand why an explicit closing of log pipe
fd is required in e.g. `(*linuxSetnsInit).Init`, right before the execve
which (thanks to `CLOEXEC`) closes the fd anyway.

Fixes: #2756

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -357,12 +357,24 @@ type openResult struct {
 	err  error
 }
 
-func (c *linuxContainer) start(process *Process) error {
+func (c *linuxContainer) start(process *Process) (retErr error) {
 	parent, err := c.newParentProcess(process)
 	if err != nil {
 		return newSystemErrorWithCause(err, "creating new parent process")
 	}
-	parent.forwardChildLogs()
+
+	logsDone := parent.forwardChildLogs()
+	if logsDone != nil {
+		defer func() {
+			// Wait for log forwarder to finish. This depends on
+			// runc init closing the _LIBCONTAINER_LOGPIPE log fd.
+			err := <-logsDone
+			if err != nil && retErr == nil {
+				retErr = newSystemErrorWithCause(err, "forwarding init logs")
+			}
+		}()
+	}
+
 	if err := parent.start(); err != nil {
 		return newSystemErrorWithCause(err, "starting container process")
 	}
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -365,6 +365,12 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 		defer consoleSocket.Close()
 	}
 
+	logPipeFdStr := os.Getenv("_LIBCONTAINER_LOGPIPE")
+	logPipeFd, err := strconv.Atoi(logPipeFdStr)
+	if err != nil {
+		return fmt.Errorf("unable to convert _LIBCONTAINER_LOGPIPE=%s to int: %s", logPipeFdStr, err)
+	}
+
 	// clear the current process's environment to clean any libcontainer
 	// specific env vars.
 	os.Clearenv()
@@ -387,7 +393,7 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 		}
 	}()
 
-	i, err := newContainerInit(it, pipe, consoleSocket, fifofd)
+	i, err := newContainerInit(it, pipe, consoleSocket, fifofd, logPipeFd)
 	if err != nil {
 		return err
 	}
--- a/libcontainer/logs/logs_linux_test.go
+++ b/libcontainer/logs/logs_linux_test.go
@@ -72,7 +72,7 @@ func logToLogWriter(t *testing.T, l *log, message string) {
 type log struct {
 	w    io.WriteCloser
 	file string
-	done chan struct{}
+	done chan error
 }
 
 func runLogForwarding(t *testing.T) *log {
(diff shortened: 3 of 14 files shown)


kolyshkin:
Looks like the "checkpoint --lazy-pages and restore" test case has stuck on Fedora, resulting in a timeout. Haven't seen it before.

CI restarted.

kolyshkin:
> Looks like the "checkpoint --lazy-pages and restore" test case has stuck on Fedora, resulting in a timeout. Haven't seen it before.

It was probably the VM itself, not the test case that got stuck (seen it elsewhere the same day)

kolyshkin:
The last commit was kinda heavy and thus complex to review -- I have split it into three smaller and much more digestible ones.

mrunalp (approved):

kolyshkin:
@cyphar @AkihiroSuda PTAL

libcontainer/container_linux.go
```diff
@@ -357,12 +357,26 @@ type openResult struct {
...
 	}
-	parent.forwardChildLogs()
+
+	logsDone := parent.forwardChildLogs()
+	if logsDone != nil {
+		defer func() {
+			// Wait for log forwarder to finish. This depends
+			// on init closing the _LIBCONTAINER_LOGPIPE log fd:
+			// For standard init, this is done by closeLogFD.
+			// For setns init, the fd is closed upon execve.
```
kolyshkin, line 372: This comment is obsoleted and should be removed.

cyphar:
> I have to admit I still do not understand why an explicit closing of log pipe
fd is required in e.g. (*linuxSetnsInit).Init, right before the execve
which (thanks to CLOEXEC) closes the fd anyway.

It's to work around a kernel bug we found ([and fixed in 4.10](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=613cc2b6f272c1a8ad33aefa21cad77af23139f7)) during the resolution of CVE-2016-9962. Basically it turned out that `O_CLOEXEC` was applied after the non-dumpable flag was reset during `execve(2)` -- leading to a race where a container process could access host file descriptors that were marked `O_CLOEXEC`. So we just have a general policy in runc to close file descriptors even if `O_CLOEXEC` should be enough (and it is enough on newer kernels).

libcontainer/container_linux.go
kolyshkin: fixed

kolyshkin:
> It's to work around a kernel bug we found (and fixed in 4.10) during the resolution of CVE-2016-9962. Basically it turned out that O_CLOEXEC was applied after the non-dumpable flag was reset during execve(2) -- leading to a race where a container process could access host file descriptors that were marked O_CLOEXEC. So we just have a general policy in runc to close file descriptors even if O_CLOEXEC should be enough (and it is enough on newer kernels).

No, I was asking about the code that **I have added** in this PR, see `unix.Close(l.logFd)` in https://github.com/opencontainers/runc/pull/2835/commits/9def77aa82e46f1942f3a13dd0674455660a1fa4#diff-f8136d4f3312c98ab4c92c6227d5b7731fb1d10d419afc88ffa02b3018187dc0 or https://github.com/kolyshkin/runc/blob/fix-init-log-race/libcontainer/setns_init_linux.go#L91-L97).

Anyway, this PR is working correctly, so my interest is purely theoretical.

kolyshkin:
Added area/ci as it fixes a long-standing CI flake (#2756).

Once this is in, #2836 will be unblocked.

libcontainer/logs/logs_linux_test.go
```diff
@@ -1,116 +1,100 @@
...
+	t.Cleanup(func() {
+		logR.Close()
+		logW.Close()
+	})
 
 	tempFile, err := ioutil.TempFile("", "")
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer tempFile.Close()
```
kolyshkin, line 93: can be just
```suggestion
	tempFile.Close()
```

libcontainer/factory_linux.go
```diff
@@ -365,6 +365,12 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 		defer consoleSocket.Close()
 	}
 
+	logPipeFdStr := os.Getenv("_LIBCONTAINER_LOGPIPE")
+	logPipeFd, err := strconv.Atoi(logPipeFdStr)
+	if err != nil {
+		return fmt.Errorf("unable to convert _LIBCONTAINER_LOGPIPE=%s to int: %s", logPipeFdStr, err)
+	}
+
```
kolyshkin, line 373: Alternatively, we can pass logFd as a parameter to `StartInitalization` as we just converted it from env in `init.go`...

Decided to do it this way since all the other parameters are obtained from environment right here.

libcontainer/logs/logs_linux_test.go
kolyshkin: fixed

AkihiroSuda (approved):

mrunalp (approved):

PR #2837, opened 2021-03-08
libcontainer/notify_linux*: adjust the file mode

KentaTada:
In addition to that, I changed those modes from 0700 to 0600
as same as #2636

Related to #2625

### DIFF
--- a/libcontainer/notify_linux.go
+++ b/libcontainer/notify_linux.go
@@ -5,10 +5,10 @@ package libcontainer
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"golang.org/x/sys/unix"
 )
 
@@ -33,13 +33,14 @@ func registerMemoryEvent(cgDir string, evName string, arg string) (<-chan struct
 
 	eventfd := os.NewFile(uintptr(fd), "eventfd")
 
-	eventControlPath := filepath.Join(cgDir, "cgroup.event_control")
+	const cgEvCtlFile = "cgroup.event_control"
 	data := fmt.Sprintf("%d %d %s", eventfd.Fd(), evFile.Fd(), arg)
-	if err := ioutil.WriteFile(eventControlPath, []byte(data), 0o700); err != nil {
+	if err := cgroups.WriteFile(cgDir, cgEvCtlFile, data); err != nil {
 		eventfd.Close()
 		evFile.Close()
 		return nil, err
 	}
+	eventControlPath := filepath.Join(cgDir, cgEvCtlFile)
 	ch := make(chan struct{})
 	go func() {
 		defer func() {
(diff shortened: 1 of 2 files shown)


libcontainer/notify_linux.go
```diff
@@ -35,7 +35,7 @@ func registerMemoryEvent(cgDir string, evName string, arg string) (<-chan struct
 
 	eventControlPath := filepath.Join(cgDir, "cgroup.event_control")
 	data := fmt.Sprintf("%d %d %s", eventfd.Fd(), evFile.Fd(), arg)
-	if err := ioutil.WriteFile(eventControlPath, []byte(data), 0700); err != nil {
+	if err := ioutil.WriteFile(eventControlPath, []byte(data), 0o600); err != nil {
```
kolyshkin, line 38: 1. In fact it does not matter what the mode here is, as the file exists (and it can't possibly be created as this is cgroupfs).

2. Might make sense to switch to `fscommon.WriteFile` here.

kolyshkin:
Would be nice to switch to `fscommon.WriteFile` here; it will eliminate the filemode arg.

libcontainer/notify_linux.go
KentaTada: I'll try to switch to `fscommon.WriteFile`.
It's a great way.

thaJeztah:
Wondering if these should be moved into `libcontainer/cgroups` (not as part of this PR)

libcontainer/notify_linux.go
```diff
@@ -33,13 +33,14 @@ func registerMemoryEvent(cgDir string, evName string, arg string) (<-chan struct
 
 	eventfd := os.NewFile(uintptr(fd), "eventfd")
 
-	eventControlPath := filepath.Join(cgDir, "cgroup.event_control")
+	const cgEvCtlFile = "cgroup.event_control"
 	data := fmt.Sprintf("%d %d %s", eventfd.Fd(), evFile.Fd(), arg)
-	if err := ioutil.WriteFile(eventControlPath, []byte(data), 0700); err != nil {
+	if err := fscommon.WriteFile(cgDir, "cgroup.event_control", data); err != nil {
```
kolyshkin, line 38: This should be
```suggestion
if err := fscommon.WriteFile(cgDir, cgEvCtlFile, data); err != nil {
```

KentaTada:
I'm not familiar with the current implementations but I think openFile() checks whether the actual cgropfs is used.
And so this change could not pass the test.
I should modify the test.

libcontainer/notify_linux_test.go
```diff
@@ -23,10 +23,10 @@ func testMemoryNotification(t *testing.T, evName string, notify notifyFunc, targ
 	}
 	evFile := filepath.Join(memoryPath, evName)
 	eventPath := filepath.Join(memoryPath, "cgroup.event_control")
```
thaJeztah, line 25: Should this also use the const? (`cgEvCtlFile`)

kolyshkin:
Since this file already exists, the `mode` argument is ignored, so there is nothing to fix here.

PR #2840, opened 2021-03-08
Ignore kernel memory settings

kolyshkin:
This is somewhat radical approach to deal with kernel memory.

Per-cgroup kernel memory limiting was always problematic. A few
examples:

 - older kernels had bugs and were even oopsing sometimes (best example
   is RHEL7 kernel);
 - kernel is unable to reclaim the kernel memory so once the limit is
   hit a cgroup is toasted;
 - some kernel memory allocations don't allow failing.

In addition to that,

 - users don't have a clue about how to set kernel memory limits
   (as the concept is much more complicated than e.g. [user] memory);
 - different kernels might have different kernel memory usage,
   which is sort of unexpected;
 - cgroup v2 do not have a [dedicated] kmem limit knob, and thus
   runc silently ignores kernel memory limits for v2;
 - kernel v5.4 made cgroup v1 kmem.limit obsoleted (see
   https://github.com/torvalds/linux/commit/0158115f702b).

In view of all this, and as the runtime-spec lists memory.kernel
and memory.kernelTCP as OPTIONAL (and NOT RECOMMENDED
since https://github.com/opencontainers/runtime-spec/pull/1093), let's silently ignore kernel memory
limits (for cgroup v1, same as we're already doing for v2).

This should result in less bugs and better user experience.

The only bad side effect from it might be that stat can show kernel
memory usage as 0 (since the accounting is not enabled).

See also:
 - https://github.com/opencontainers/runtime-spec/pull/1093
 - https://github.com/torvalds/linux/commit/0158115f702b
 - https://github.com/opencontainers/runc/pull/1921

Closes: #2594


thaJeztah:
I think this makes sense to do; even if runc would continue supporting it, it would depend on the host/kernel if it would even work, and with it being deprecated in the kernel, it would be kinda "living on borrowed time" for any user using this feature

README.md
```diff
@@ -61,18 +61,18 @@ sudo make install
 with some of them enabled by default (see `BUILDTAGS` in top-level `Makefile`).
 
 To change build tags from the default, set the `BUILDTAGS` variable for make,
-e.g.
+e.g. to disable selinux:
```
AkihiroSuda, line 64: to disable seccomp, right?

thaJeztah: guess it's too late to change to a `noseccomp` build-tag instead of `seccomp` ? Would me more intuitive (and make the logic easier in the makefile etc)

kolyshkin: Ah yeah, seccomp. Fixed!

kolyshkin: So, `seccomp` vs `noseccomp` is out of scope for this PR, and is probably post-1.0 material. Should be trivial to implement though, as no third-party packages use it.

kolyshkin:
@opencontainers/runc-maintainers we need to decide whether this makes sense, and if it is rc94 or post-1.0 material.

I don't think it can break any users that SET kernel memory. The only breakage I can think of is when someone is expecting kmem usage > 0 and with not enabling the acct it will be 0 which might disappoint some monitoring tools.

From the POV of disappointment, seems it's better to do before 1.0 (so in the worst case scenario we can revert it).

From the feature freeze POV, this looks like a feature and thus should not be included.

Any opinions welcome :pray:

kolyshkin:
The (less radical) alternative to this would be to enable kmem accounting (if limits are set) but don't apply any kmem limits. This incurs some runtime overhead in the kernel, but as there are no limits there are no bad consequences described above.

thaJeztah:
I'm still +1 to this; better to rip off the bandaid sooner  than later, as it will need to happen _at some point_

- AFAIK there's no replacement, so it will be a hit/miss if a host supports it
- Runtime spec accepted https://github.com/opencontainers/runtime-spec/pull/1093 to mark this as not-recommended (and memory as a whole is  `OPTIONAL`)
- If runc is a reference implementation of the spec; following the spec to "not use non-recommended" options, feels ok

kolyshkin:
close/reopen to kick CI

@opencontainers/runc-maintainers PTAL (I think this is suitable for rc94).

mrunalp (approved):

libcontainer/cgroups/fs/memory.go
```diff
@@ -140,23 +115,14 @@ func (s *MemoryGroup) Set(path string, cgroup *configs.Cgroup) error {
 		return err
 	}
 
-	if cgroup.Resources.KernelMemory != 0 {
-		if err := setKernelMemory(path, cgroup.Resources.KernelMemory); err != nil {
-			return err
-		}
-	}
+	// ignore KernelMemory and KernelMemoryTCP
```
hqhq, line 118: Can we at least print something if kernel memory limit is set so users can be aware of this?

kolyshkin: This is already done in specconv.

kolyshkin: https://github.com/opencontainers/runc/pull/2840/commits/2f54a31e3ec5461bf11d955ddb6927ab7a328952#diff-93ae3e89134c5b7d0b37869c5b950405c369d60559b300ed579b85b9f89f3884R256

hqhq: Isn't that only for update command?

kolyshkin: Yep I got it all mixed up; sorry about that.

Please see updated patch. In short, we
 - ignore kmem settings in cgroup v1 driver (same as we were doing for cgroup v2);
 - warn about settings being ignored in `specconv` and `runc update`.

mrunalp (approved):

hqhq (approved):

PR #2841, opened 2021-03-08
tests: Move fuzzers upstream

AdamKorcz:
This PR moves the fuzzers and oss-fuzz build script upstream from: https://github.com/google/oss-fuzz/tree/master/projects/runc


tests/fuzzing/oss_fuzz_build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+mv $SRC/id_map_fuzzer.go $SRC/runc/libcontainer/system/
+compile_go_fuzzer ./libcontainer/system FuzzUIDMap id_map_fuzzer linux
```
kolyshkin, line 4: Where is this binary (`compile_go_fuzzer`) coming from?

tests/fuzzing/oss_fuzz_build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+mv $SRC/id_map_fuzzer.go $SRC/runc/libcontainer/system/
```
kolyshkin, line 3: 1. Who calls this script (and sets `$SRC`)?

kolyshkin:
Overall, I'd rather have it the way it is laid out in the official example (linked from https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/):
https://github.com/golang/go/blob/4ad13555184eb0697c2e92c64c1b0bdb287ccc10/src/html/fuzz.go#L13

IOW, add files with Fuzz functions to their appropriate directories, guarded by the `gofuzz` build tag.

tests/fuzzing/oss_fuzz_build.sh
AdamKorcz: The script is only meant to be run by OSS-fuzz who also sets `$SRC`. I have added a bit of documentation inside the build file itself.

tests/fuzzing/oss_fuzz_build.sh
AdamKorcz: This is defined by OSS-fuzz. Documentation about this has been added inside the build file.

AdamKorcz:
I have moved the fuzzers to their respective folders. All comments should be addressed at this point.

cyphar (approved):
LGTM.

kolyshkin (approved):
LGTM

PR #2842, opened 2021-03-08
tests/checkpoint: show full log lazy pages cpt

kolyshkin:
Sometimes the test fails without any clear diagnostics:

> not ok 16 checkpoint --lazy-pages and restore
> (in test file tests/integration/checkpoint.bats, line 191)
>  `[ "$out" = "0000000 000000 0000001" ]' failed
> ...
> criu failed: type NOTIFY errno 3\nlog file: work-dir/dump.log

We look for and print errors via grep, but in the above case
there are nothing that is denoted error in the log.

So, let's show the damn log in its entirely (note it is only shown
if test fails).

Hope it will help with #2760

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -185,8 +185,8 @@ function simple_cr() {
 	exec {lazy_w}>&-
 	# shellcheck disable=SC2116,SC2086
 	out=$(echo $out) # rm newlines
-	# show errors if there are any before we fail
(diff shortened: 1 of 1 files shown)


mrunalp (approved):
LGTM

kolyshkin:
Looks like VMs on GHA are not particularly reliable -- sometimes they just time out.

CI restarted.

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda (approved):

PR #2847, opened 2021-03-11
refactor: refactor code structure

huweihuang:
The original code structure of RUNC was messy and not easy to read, so the code structure was restructured in accordance with the K8S code style to facilitate development and reading, and hopefully can be merged


AkihiroSuda:
This will break installation scripts of a bunch of the downstream projects without clear benefit. So NOTLGTM.

kolyshkin (changes requested):
NACK.

Generally, we should weight the benefits of a particular change vs kind of disruption it makes.

In here, the disruption is massive (think about git history, and the benefits are not too clear to me.

If you want to proceed, please explain the benefits better.

kolyshkin:
No feedback from the author, thus closing

PR #2848, opened 2021-03-11
Fuzzing: Add new fuzzer

AdamKorcz:
This PR adds a fuzzer that creates a container with a randomized configuration and then calls `State()`.

Furthermore a seed and a dictionary is added to help the fuzzer mutate relevant input data.

Lastly the OSS-fuzz build file has been updated to include this new fuzzer.


AdamKorcz:
Failing build looks unrelated

tests/fuzzing/corpus/state_fuzzer_seed_1
```diff
@@ -0,0 +1,180 @@
...
+		},
+	},
+	Rlimits: []configs.Rlimit{
+		{
+			Type: unix.RLIMIT_NOFILE,
+			Hard: uint64(1025),
+			Soft: uint64(1025),
+		},
+	},
+}
```
kolyshkin, line 180: nit: no newline at EOF

AkihiroSuda (changes requested):
The fuzzer should be ensured to run `FuzzStateApi()` only with a valid JSON bytes. I'm not sure whether it is possible with go-fuzz, but without ensuring that, fuzzing is not really meaningful. (Assuming that `json.Unmarshal` is stable enough and not worth fuzzing.)

AdamKorcz:
> The fuzzer should be ensured to run FuzzStateApi() only with a valid JSON bytes.

You are right, and the fuzzer will take care of that. It will eventually start creating json to get a valid container.

The way the fuzzer creates the container allows it to be validated:

https://github.com/opencontainers/runc/blob/f973238cb71d948fd8ba5fe06121a0516b5c7ccc/libcontainer/factory_linux.go#L192-L198

... and if the fuzzer does not generate a valid container, it will return and try over:

```
container, err := newContainer(root, config)
if err != nil {
	return 0
}
```

State() is therefore not called on invalid conainers.

> (Assuming that json.Unmarshal is stable enough and not worth fuzzing.)

`json.Unmarshal` is being fuzzed elsewhere continuously.

AdamKorcz:
@AkihiroSuda Can you check what I wrote? It would be helpful to get this up and running on OSS-fuzz to see how it performs continuously. I will modify the fuzzer over the next couple of weeks, and getting feedback from OSS-fuzz helps tremendously in that process.

libcontainer/libcontainer_fuzzer.go
```diff
@@ -0,0 +1,85 @@
...
+		return -1
+	}
+
+	// The path to our root dir MUST be found
+	// in our input
+	err = checkRootInData(data, root)
+	if err != nil {
+		return -1
+	}
+	var config *configs.Config
```
AkihiroSuda, line 45: Why pointer?

libcontainer/libcontainer_fuzzer.go
```diff
@@ -0,0 +1,85 @@
...
+	reader := bytes.NewReader(data)
+	if err := json.NewDecoder(reader).Decode(&config); err != nil {
```
AkihiroSuda, line 47: I still don't understand how `data` is guaranteed to be unmarshallable JSON

AdamKorcz: It is not guaranteed. But in the case of this fuzzer that doesn't matter, since we are using the validator. Over time the fuzzer will generate data that is unmarshallable JSON, but every test case will at no time be guaranteed to be useful.
If this has consequences for the fuzzer, then I will adjust it.

libcontainer/libcontainer_fuzzer.go
AdamKorcz: in the helper `newContainerWithName` in the fuzzer we call into f.Create (https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L243) which expects a pointer.

AdamKorcz:
@AkihiroSuda Please check the updated version of the fuzzer. We have aborted using json to create the config and instead pass random values to the struct.

AkihiroSuda (approved):
Looks better, thanks

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/opencontainers/runc
 go 1.14
 
 require (
+	github.com/AdaLogics/go-fuzz-headers v0.0.0-20210323220609-b81652571c3f
```
AkihiroSuda, line 6: Please add LICENSE file to go-fuzz-headers repo

AkihiroSuda (changes requested):

AdamKorcz:
@kolyshkin @AkihiroSuda How is this one looking now?

AkihiroSuda (dismissed):

AdamKorcz:
I ran the tests again, and this PR now passes

AdamKorcz:
@AkihiroSuda @kolyshkin How is this one looking now?

kolyshkin:
@AdamKorcz can you please fix the commit message? Now it merely says "Updated" which is not a good description of what the commit does.

kolyshkin:
Also needs a rebase.

kolyshkin (dismissed):
LGTM

(but this will probably need a rebase after #2925 is merged)

AdamKorcz:
@kolyshkin Can this be merged?

caniszczyk:
it has branch conflicts now

but cc: @opencontainers/runc-maintainers

AdamKorcz:
rebased

cyphar (approved):
LGTM.

kolyshkin:
I'm inclined to close this one (and other similar PRs) as we were unable to merge this for the last few years.

PR #2849, opened 2021-03-13
fix: format the import package

huweihuang:
format the import package.
The order of packages: system library, third party library, current project library

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,14 @@ run:
   build-tags:
     - seccomp
 
+# all available settings of specific linters
+linters-settings:
+  goimports:
+    # put imports beginning with prefix after 3rd-party packages;
+    # it's a comma-separated list of prefixes
+    local-prefixes: github.com/opencontainers/runc
+
 linters:
   enable:
     - gofmt
+    - goimports
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -10,12 +10,13 @@ import (
 	"strconv"
 
 	criu "github.com/checkpoint-restore/go-criu/v4/rpc"
-	"github.com/opencontainers/runc/libcontainer"
-	"github.com/opencontainers/runc/libcontainer/system"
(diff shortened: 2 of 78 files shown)


thaJeztah:
If we switch to using the "local" format (i.e., use 3 groups: "stdlib", "external imports", "local imports"), we should probably add a check in CI (and possibly a makefile target) to verify;

```bash
goimports -w --local=github.com/opencontainers/runc .
```

(excluding `vendor/`)

huweihuang:
It might be clearer to separate the reference packages.

thaJeztah:
> It might be clearer to separate the reference packages.

I don't disagree, but if we make this change, we should consider having a check in CI, and a make target, so that contributors can verify that their code is formatted correctly; a standard `gofmt -s -w` won't group imports in three groups, and to do that, you'd need to use `goimports` (which the `--local` option); `goimports` is not installed by default, so having a `make` target that (installs, and) runs `goimports` with the right options would be useful for that.

@kolyshkin @AkihiroSuda wdyt?

kolyshkin:
I'd rather leave this as is, following the official gofmt/goimports with default settings (and if in the future it will switch to the same formatting as this PR suggests, so be it).

kolyshkin:
In any case, this is not for rc94.

kolyshkin:
@huweihuang if you want to proceed with this, you need to add CI infra to make sure these imports won't be broken again. One way to do this is to amend golangci-lint config to use an appropriate linter (or a linter option).

kolyshkin:
No response from @huweihuang -- closing

kolyshkin:
I was wrong about it -- in fact, @huweihuang added the second commit with the proper linter configuration.

Reopening. I suggest we merge this after 1.1 release.

kolyshkin:
@huweihuang do you want to rebase it?

kolyshkin:
Let's close this one as it's too old to be rebased. Also, we now have some tools / linters in place and are gradually improving the order of imports.

PR #2854, opened 2021-03-15
capabilities: WARN, not ERROR, for unknown / unavailable capabilities

thaJeztah:
### go.mod: OCI runtime-spec v1.0.3-0.20210326190908-1c3f411f0417

full diff: https://github.com/opencontainers/runtime-spec/compare/a8c4a9ee0f6b...1c3f411f0417

- Fix seccomp notify inconsistencies
    - seccomp: Add missing const for seccomp notify action
    - schema/defs-linux: Fix inconsistencies with seccomp notify
- Proposal: runtime should ignore capabilities that cannot be granted

###  capabilities: WARN, not ERROR, for unknown / unavailable capabilities

This updates handling of capabilities to match the updated runtime specification,
in https://github.com/opencontainers/runtime-spec/pull/1094.

Prior to that change, the specification required runtimes to produce a (fatal)
error if a container configuration requested capabilities that could not be
granted (either the capability is "unknown" to the runtime, not supported by the
kernel version in use, or not available in the environment that the runtime
operates in).

This caused problems in situations where the runtime was running in a restricted
environment (for example, docker-in-docker), or if there is a mismatch between
the list of capabilities known by higher-level runtimes and the OCI runtime.

Some examples:

- Kernel 5.8 introduced CAP_PERFMON, CAP_BPF, and CAP_CHECKPOINT_RESTORE
  capabilities. Docker 20.10.0 ("higher level runtime") shipped with
  an updated list of capabilities, and when creating a "privileged" container,
  would determine what capabilities are known by the kernel in use, and request
  all those capabilities (by including them in the container config).
  However, runc did not yet have an updated list of capabilities, and therefore
  reject the container specification, producing an error because the new
  capabilities were "unknown".
- When running nested containers, for example, when running docker-in-docker,
  the "inner" container may be using a more recent version of docker than the
  "outer" container. In this situation, the "outer" container may be missing
  capabilities that the inner container expects to be supported (based on
  kernel version). However, starting the container would fail, because the OCI
  runtime could not grant those capabilities (them not being available in the
  environment it's running in).

WARN (but otherwise ignore) capabilities that cannot be granted
--------------------------------------------------------------------------------

This patch changes the handling to WARN (but otherwise ignore) capabilities that
are requested in the container config, but cannot be granted, alleviating higher
level runtimes to detect what capabilities are supported (by the kernel, and
in the current environment), as well as avoiding failures in situations where
the higher-level runtime is aware of capabilities that are not (yet) supported
by runc.

Impact on security
--------------------------------------------------------------------------------

Given that `capabilities` is an "allow-list", ignoring unknown capabilities does
not impose a security risk; worst case, a container does not get all requested
capabilities granted and, as a result, some actions may fail.

Backward-compatibility
--------------------------------------------------------------------------------

This change should be fully backward compatible. Higher-level runtimes that
already dynamically adjust the list of requested capabilities can continue to do
so. Runtimes that do not adjust will see an improvement (containers can start
even if some of the requested capabilities are not granted). Container processes
MAY fail (as described in "impact on security"), but users can debug this
situation either by looking at the warnings produces by the OCI runtime, or using
tools such as `capsh` / `libcap` to get the list of actual capabilities in the
container.

### DIFF
--- a/libcontainer/capabilities/capabilities.go
+++ b/libcontainer/capabilities/capabilities.go
@@ -3,10 +3,11 @@
 package capabilities
 
 import (
-	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/sirupsen/logrus"
 	"github.com/syndtr/gocapability/capability"
 )
 
@@ -34,27 +35,21 @@ func init() {
 }
 
 // New creates a new Caps from the given Capabilities config. Unknown Capabilities
-// or Capabilities that are unavailable in the current environment produce an error.
+// or Capabilities that are unavailable in the current environment are ignored,
+// printing a warning instead.
 func New(capConfig *configs.Capabilities) (*Caps, error) {
 	var (
 		err error
-		c   = Caps{caps: make(map[capability.CapType][]capability.Cap, len(capTypes))}
+		c   Caps
 	)
 
-	if c.caps[capability.BOUNDING], err = capSlice(capConfig.Bounding); err != nil {
-		return nil, err
-	}
-	if c.caps[capability.EFFECTIVE], err = capSlice(capConfig.Effective); err != nil {
-		return nil, err
-	}
-	if c.caps[capability.INHERITABLE], err = capSlice(capConfig.Inheritable); err != nil {
-		return nil, err
-	}
-	if c.caps[capability.PERMITTED], err = capSlice(capConfig.Permitted); err != nil {
-		return nil, err
-	}
-	if c.caps[capability.AMBIENT], err = capSlice(capConfig.Ambient); err != nil {
-		return nil, err
+	unknownCaps := make(map[string]struct{})
+	c.caps = map[capability.CapType][]capability.Cap{
+		capability.BOUNDING:    capSlice(capConfig.Bounding, unknownCaps),
+		capability.EFFECTIVE:   capSlice(capConfig.Effective, unknownCaps),
+		capability.INHERITABLE: capSlice(capConfig.Inheritable, unknownCaps),
+		capability.PERMITTED:   capSlice(capConfig.Permitted, unknownCaps),
+		capability.AMBIENT:     capSlice(capConfig.Ambient, unknownCaps),
 	}
 	if c.pid, err = capability.NewPid2(0); err != nil {
 		return nil, err
@@ -62,22 +57,35 @@ func New(capConfig *configs.Capabilities) (*Caps, error) {
 	if err = c.pid.Load(); err != nil {
 		return nil, err
 	}
+	if len(unknownCaps) > 0 {
+		logrus.Warn("ignoring unknown or unavailable capabilities: ", mapKeys(unknownCaps))
+	}
 	return &c, nil
 }
 
 // capSlice converts the slice of capability names in caps, to their numeric
 // equivalent, and returns them as a slice. Unknown or unavailable capabilities
-// produce an error.
-func capSlice(caps []string) ([]capability.Cap, error) {
-	out := make([]capability.Cap, len(caps))
-	for i, c := range caps {
-		v, ok := capabilityMap[c]
-		if !ok {
-			return nil, fmt.Errorf("unknown capability %q", c)
+// are not returned, but appended to unknownCaps.
+func capSlice(caps []string, unknownCaps map[string]struct{}) []capability.Cap {
+	var out []capability.Cap
+	for _, c := range caps {
+		if v, ok := capabilityMap[c]; !ok {
+			unknownCaps[c] = struct{}{}
+		} else {
+			out = append(out, v)
 		}
-		out[i] = v
 	}
-	return out, nil
+	return out
+}
+
+// mapKeys returns the keys of input in sorted order
+func mapKeys(input map[string]struct{}) []string {
+	var keys []string
+	for c := range input {
+		keys = append(keys, c)
+	}
+	sort.Strings(keys)
+	return keys
 }
 
 // Caps holds the capabilities for a container.
--- a/libcontainer/capabilities/capabilities_linux_test.go
+++ b/libcontainer/capabilities/capabilities_linux_test.go
@@ -1,14 +1,18 @@
 package capabilities
 
 import (
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/syndtr/gocapability/capability"
 )
 
 func TestNew(t *testing.T) {
-	cs := []string{"CAP_CHOWN"}
+	cs := []string{"CAP_CHOWN", "CAP_UNKNOWN", "CAP_UNKNOWN2"}
 	conf := configs.Capabilities{
 		Bounding:    cs,
 		Effective:   cs,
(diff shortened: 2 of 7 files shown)


thaJeztah:
@kolyshkin @AkihiroSuda ptal; upstream change wasn't accepted yet, but we can probably already discuss this (possible) implementation.

libcontainer/capabilities/capabilities.go
```diff
@@ -24,47 +24,69 @@ func init() {
...
 
-func capSlice(caps []string) ([]capability.Cap, error) {
+func printWarnings(warnings []string) {
+	if len(warnings) == 0 {
+		return
+	}
+	var unknownCaps = map[string]struct{}{}
+	for _, c := range warnings {
+		if _, ok := unknownCaps[c]; !ok {
+			logrus.WithField("capability", c).Warn("ignoring unknown or unavailable capability")
```
thaJeztah, line 70: Only consideration I had here was if we wanted to pass a logger to `New()`, instead of always using the default logrus logger (in case this is used elsewhere as library)

thaJeztah: Also not sure if it must be a WARN level, or if we think INFO (possibly with "warning" in the log message) to reduce noise

kolyshkin: I'd rather have a single message listing all the capabilities, maybe something like `Warning: unknown capabilities [CAP_ONE CAP_TWO ...] are ignored`.

libcontainer/capabilities/capabilities.go
```diff
@@ -24,47 +24,69 @@ func init() {
...
+		err      error
+		caps     Caps
+		ignored  []string
+		warnings = make([]string, 0)
 	)
 
-	if caps.bounding, err = capSlice(capConfig.Bounding); err != nil {
-		return nil, err
+	if caps.bounding, ignored = capSlice(capConfig.Bounding); len(ignored) > 0 {
+		warnings = append(warnings, ignored...)
```
kolyshkin, line 39: IMO the best way to de-dup the list of unknown caps is to put them into a `map[string]struct{}{}` right away.

Say `capSlice` may have a second argument with the `ignored` map, to which it would add more entries.

```go
// ...
// Names of unknown or unavailable capabilities are added to the ignored map.
func capSlice(caps []string, ignored map[string]struct{}{}) []capability.Cap {
```

libcontainer/capabilities/capabilities.go
```diff
@@ -24,47 +24,69 @@ func init() {
...
+	var unknownCaps = map[string]struct{}{}
+	for _, c := range warnings {
+		if _, ok := unknownCaps[c]; !ok {
+			logrus.WithField("capability", c).Warn("ignoring unknown or unavailable capability")
+			unknownCaps[c] = struct{}{}
+		}
+	}
+}
+
+// capSlice converts the given slice of capability names, and converts them
```
kolyshkin, line 76: nit: converts ... and converts ?

vendor/modules.txt
```diff
@@ -57,6 +57,7 @@ github.com/shurcooL/sanitized_anchor_name
 # github.com/sirupsen/logrus v1.7.0
 ## explicit
 github.com/sirupsen/logrus
+github.com/sirupsen/logrus/hooks/test
```
kolyshkin, line 60: Not sure how this ended up here ...

kolyshkin: Ah... got it

thaJeztah:
@kolyshkin I played around with some approaches; did some refactoring (first two commits); let me know if you like this approach

libcontainer/capabilities/capabilities.go
```diff
@@ -34,50 +35,58 @@ func init() {
...
 	}
 	if err = c.pid.Load(); err != nil {
 		return nil, err
 	}
+	printWarnings(warnings)
 	return &c, nil
 }
 
-// capSlice converts the slice of capability names in caps, to their numeric
+// capSlice converts the slice of capability names in newCaps, to their numeric
```
kolyshkin, line 64: nit: the argument is still called `caps` below.

libcontainer/capabilities/capabilities.go
```diff
@@ -34,50 +35,58 @@ func init() {
...
-// produce an error.
-func capSlice(caps []string) ([]capability.Cap, error) {
-	out := make([]capability.Cap, len(caps))
-	for i, c := range caps {
-		v, ok := capabilityMap[c]
-		if !ok {
-			return nil, fmt.Errorf("unknown capability %q", c)
+// are not returned, but appended to warnings, which can be used by the caller to
+// print.
+func capSlice(caps []string, warnings map[string]struct{}) []capability.Cap {
```
kolyshkin, line 68: I'd rather call it "unknownCaps" not "warnings".

libcontainer/capabilities/capabilities.go
```diff
@@ -34,50 +35,58 @@ func init() {
...
+// capSlice converts the slice of capability names in newCaps, to their numeric
 // equivalent, and returns them as a slice. Unknown or unavailable capabilities
-// produce an error.
-func capSlice(caps []string) ([]capability.Cap, error) {
-	out := make([]capability.Cap, len(caps))
-	for i, c := range caps {
-		v, ok := capabilityMap[c]
-		if !ok {
-			return nil, fmt.Errorf("unknown capability %q", c)
+// are not returned, but appended to warnings, which can be used by the caller to
```
kolyshkin, line 66: I'd say "but added to warnings map" and stop here (it's up to the caller what to do with them -- print a warning, an error, or set the world on fire).

libcontainer/capabilities/capabilities.go
```diff
@@ -34,50 +35,58 @@ func init() {
...
+		} else {
+			out = append(out, v)
 		}
-		out[i] = v
 	}
-	return out, nil
+	return out
+}
+
+func printWarnings(warnings map[string]struct{}) {
```
kolyshkin, line 80: Maybe `func warnUnknownCaps(caps map[string]struct{})`?

libcontainer/capabilities/capabilities.go
```diff
@@ -34,50 +35,58 @@ func init() {
...
+		capability.PERMITTED:   capSlice(capConfig.Permitted, warnings),
+		capability.AMBIENT:     capSlice(capConfig.Ambient, warnings),
 	}
 	if c.pid, err = capability.NewPid2(0); err != nil {
 		return nil, err
 	}
 	if err = c.pid.Load(); err != nil {
 		return nil, err
 	}
+	printWarnings(warnings)
```
kolyshkin, line 60: I would just do something like
```go
logrus.Warnf("ignoring unknown or unavailable capabilities: %s", reflect.ValueOf(warnings).MapKeys())
```

for the sake of saving the LOC. Example: https://play.golang.org/p/Hl7JERnmtxY

It may be slower than your code (but maybe not) and not super readable, but in this place I guess no one cares.

kolyshkin:
> I played around with some approaches; did some refactoring (first two commits); let me know if you like this approach

Thanks, I like the refactoring and the map approach; left a few nits wrt naming and a way to easier print the unknown caps.

thaJeztah:
@kolyshkin updated; PTAL

Let me also open the refactor (first two commits) as a separate PR

thaJeztah:
opened https://github.com/opencontainers/runc/pull/2863

thaJeztah:
Rebased to get rid of the commits of https://github.com/opencontainers/runc/pull/2863

kolyshkin (approved):
LGTM

thaJeztah:
Proposal was accepted in the runtime-spec; I updated the runtime spec in `go.mod` so that this change matches the version of the spec we implement, rebase, and moved this out of draft

@AkihiroSuda @kolyshkin @mrunalp PTAL

AkihiroSuda (approved):

libcontainer/capabilities/capabilities_linux_test.go
```diff
@@ -17,10 +21,36 @@ func TestNew(t *testing.T) {
 		Ambient:     cs,
 	}
 
+	hook := test.NewGlobal()
```
cpuguy83, line 24: Oh clever.

cpuguy83 (approved):
LGTM

image

kolyshkin (approved):
LGTM

kolyshkin:
close/reopen to kick ci

PR #2843, opened 2021-03-10
umount all mount points in runc root dir

lifubang:
Sometimes, `umount` with `MNT_DETACH` flag can't work in a small probability.
So, maybe we should unmount all mount points in runc root dir when we remove this dir.

```
root@f7c63618dde3:/opt/busybox# runc delete test2
remove /run/runc/test2: unlinkat /run/runc/test2/runc.W24K2t: device or resource busy
ERRO[0000] container "test2" does not exist             
```
```
root@f7c63618dde3:/opt/busybox# ls /run/runc/test2 -alh
total 14M
drwx--x--x 2 root root  60 Mar  9 14:24 .
drwx------ 5 root root 100 Mar  9 15:15 ..
-rwxr-xr-x 1 root root 14M Mar  9 15:15 runc.W24K2t
```

Signed-off-by: lifubang <lifubang@acmcoder.com>

PR #2855, opened 2021-03-16
Avoid killing runc init early

kolyshkin:
Two commits, each one solving a problem of a stale bind mount
left by runc init after unsuccessful container start.

See https://github.com/opencontainers/runc/pull/2843 for more details
and the initial investigation.

Closes: #2843 

### start: don't kill runc init too early

The stars can be aligned in a way that results in runc to leave a stale
bind mount in container's state directory, which manifests itself later,
while trying to remove the container, in an error like this:

> remove /run/runc/test2: unlinkat /run/runc/test2/runc.W24K2t: device or resource busy

The stale mount happens because runc start/run/exec kills runc init
while it is inside ensure_cloned_binary(). One such scenario is when
a unified cgroup resource is specified for cgroup v1, a cgroup manager's
Apply returns an error (as of commit b006f4a18047), and when
(*initProcess).start() kills runc init just after it was started.

One solution is NOT to kill runc init too early. To achieve that,
amend the libcontainer/nsenter code to send a \0 byte to signal
that it is past the initial setup, and make start() (for both
run/start and exec) wait for this byte before proceeding with
kill on an error path.

While at it, improve some error messages.

### libct/configs/validator: add some cgroup support

Add some minimal validation for cgroups. The following checks
are implemented:

 - cgroup name and/or prefix (or path) is set;
 - for cgroup v1, unified resources are not set;
 - for cgroup v2, if memorySwap is set, memory is also set,
   and memorySwap > memory.

This makes some invalid configurations fail earlier (before runc init
is started), which is better, as this should prevent killing runc init
in the middle of ensure_cloned_binary().

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	selinux "github.com/opencontainers/selinux/go-selinux"
@@ -55,6 +56,10 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 			return err
 		}
 	}
+	if err := v.cgroups(config); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


kolyshkin:
Some repo (as configured by github) appears to be broken:

> Err:51 https://dl.bintray.com/sbt/debian  Release
>  502  Bad Gateway [IP: 52.43.227.140 443]

Restarted; got a different but similar error:

> E: The repository 'https://dl.bintray.com/sbt/debian  Release' is no longer signed.

kolyshkin:
(validate CI is failing here -- fixed in https://github.com/opencontainers/runc/pull/2860)

@AkihiroSuda @cyphar @lifubang PTAL

kolyshkin:
@AkihiroSuda @cyphar @lifubang PTAL (please ignore failed CI -- it is fixed in #2860)

AkihiroSuda:
Could you rebase, LGTM then

kolyshkin:
> Could you rebase, LGTM then

Rebased to include #2860

AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #2836, opened 2021-03-06
Improve nsexec logging

kolyshkin:
~~Draft as it's blocked by #2835~~

In order to make `runc --debug` actually useful for debugging nsexec
bugs, provide information about all the internal operations when in
debug mode.

Before:
```console
# runc --debug run x123
DEBU[0000] nsexec:601 nsexec started                    
DEBU[0000] child process in init()                      
DEBU[0000] logging has already been configured          
DEBU[0000] log pipe has been closed: EOF                
DEBU[0001] process exited                                pid=1472529 status=0
```

After:
```console
# runc --debug run x123
DEBU[0000] nsexec[1472380]: => nsexec container setup   
DEBU[0000] nsexec-0[1472380]: ~> nsexec stage-0         
DEBU[0000] nsexec-0[1472380]: spawn stage-1             
DEBU[0000] nsexec-0[1472380]: -> stage-1 synchronisation loop 
DEBU[0000] nsexec-1[1472384]: ~> nsexec stage-1         
DEBU[0000] nsexec-1[1472384]: unshare remaining namespace (except cgroupns) 
DEBU[0000] nsexec-1[1472384]: spawn stage-2             
DEBU[0000] nsexec-1[1472384]: request stage-0 to forward stage-2 pid (1472385) 
DEBU[0000] nsexec-0[1472380]: stage-1 requested pid to be forwarded 
DEBU[0000] nsexec-0[1472380]: forward stage-1 (1472384) and stage-2 (1472385) pids to runc 
DEBU[0000] nsexec-2[1]: ~> nsexec stage-2               
DEBU[0000] nsexec-1[1472384]: signal completion to stage-0 
DEBU[0000] nsexec-1[1472384]: <~ nsexec stage-1         
DEBU[0000] nsexec-0[1472380]: stage-1 complete          
DEBU[0000] nsexec-0[1472380]: <- stage-1 synchronisation loop 
DEBU[0000] nsexec-0[1472380]: -> stage-2 synchronisation loop 
DEBU[0000] nsexec-0[1472380]: signalling stage-2 to run 
DEBU[0000] nsexec-2[1]: unshare cgroup namespace        
DEBU[0000] nsexec-2[1]: signal completion to stage-0    
DEBU[0000] nsexec-0[1472380]: stage-2 complete          
DEBU[0000] nsexec-0[1472380]: <- stage-2 synchronisation loop 
DEBU[0000] nsexec-0[1472380]: <~ nsexec stage-0         
[...]

PR #2860, opened 2021-03-16
fix/simplify scripts/validate-c, fix *.c formatting

kolyshkin:
Apparently, scripts/validate-c is not working in CI (or maybe
maintainers ignored the failures from it?) -- current C code
gets some changes if we run recommended `indent` command on it.

This commit fixes this, simplifying things along the way.

In particular:

1. Remove `validate` make target, add `cfmt` target that just runs
   `indent` on all `*.c` files in the repository (NOTE that `*.h` files
   are not included, as before).

   This may help a contributor to fix their code -- they just need
   to run "make cfmt" now instead of running "make validate" and
   copy-pasting the indent command and options from the hint.

2. Split GHA `validate/misc` job into `validate/release` and `validate/cfmt`.
   The latter checks that the sources are not changed after `make cfmt`.

3. Adds a few more options to `indent`. This was mostly motivated by
   trying to save the existing formatting as much as possible, minimizing
   the amount of changes indent produces.

   The added options are:

   * `-il0`: sets the offset for goto labels to 0 (currently all labels
     but one are not indented -- let's keep it that way);

   * `-ppi2`: sets the indentation for nested preprocessor directives
     to 2 spaces (same as in current code defining `SYS_memfd_create`);

   * `-cp1`: sets the indentation between `#else` / `#endif` and the
     following comment to 1 space.

4. Reformat the code using the new indent options.

5. Remove the now-unused `script/{.validate,validate-c}`.

_v2: removed scripts/validate-c in favor of `make cfmt` and a GHA job; added a few indent options_
_v3: removed redundant comments from SYS_memfd_create defines_


AkihiroSuda (dismissed):

libcontainer/nsenter/cloned_binary.c
```diff
@@ -60,35 +60,35 @@
...
 /* Use our own wrapper for memfd_create. */
 #ifndef SYS_memfd_create
-#  ifdef __NR_memfd_create
-#    define SYS_memfd_create __NR_memfd_create
-#  else
+#ifdef __NR_memfd_create
+#define SYS_memfd_create __NR_memfd_create
+#else
 /* These values come from <https://fedora.juszkiewicz.com.pl/syscalls.html>. */
-#    warning "libc is outdated -- using hard-coded SYS_memfd_create"
```
thaJeztah, line 67: curious; that indentation looks like it was intentional; is there some option / other formatting that assumes those to be indented? (I'm not a c-coder, but with indentation looks somewhat more readable to me)

kolyshkin: I don't like most of these changes either, but don't know about any tools better than `indent` (took a quick look into `clang-format`, need to play more).

In C world it's usually a combination of CodingStyle document, a bit of help from an editor, and some common sense.

kolyshkin: I'm going to play a bit more with `indent` and `clang-format`

kolyshkin:
I'm going to play a bit more with `indent` and `clang-format` -- moving to draft for now.

libcontainer/nsenter/cloned_binary.c
thaJeztah: Thanks! Glad it's not just me 😂

kolyshkin: OK, what I've just pushed should be somewhat better.

kolyshkin:
Adding a test commit to check the newly added `validate / cfmt` job (extra vertical space, extra space between #endif and its comment, too long comment line):

```diff
@@ -132,7 +132,12 @@ int setns(int fd, int nstype)
 {
        return syscall(SYS_setns, fd, nstype);
 }
-#endif
+#endif      /* GLIBC < 2.14 */
+
+
+
+
+/* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. */
 
 static void write_log_with_info(const char *level, const char *function, int line, const char *format, ...)
 {
```

kolyshkin:
The `validate / cfmt` job fails as it should with the badly formatted input, so it works:

```
indent -linux -l120 -il0 -ppi2 -cp1 -T size_t -T jmp_buf libcontainer/nsenter/cloned_binary.c libcontainer/nsenter/nsexec.c tests/integration/testdata/dev_access_test.c tests/integration/testdata/seccomp_syscall_test1.c
diff --git a/libcontainer/nsenter/nsexec.c b/libcontainer/nsenter/nsexec.c
index 01ae8489..7d52e287 100644
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -132,10 +132,7 @@ int setns(int fd, int nstype)
 {
 	return syscall(SYS_setns, fd, nstype);
 }
-#endif      /* GLIBC < 2.14 */
-
-
-
+#endif /* GLIBC < 2.14 */
 
 /* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. */


Error: Process completed with exit code 1.
```

Test commit removed.

kolyshkin:
No longer a draft. Note that `validate / misc` job is gone, replaced by `validate / release` and `validate / cfmt`, so it's not showing.

Once this PR is merged, I'll update the GH list of required checks.

libcontainer/nsenter/cloned_binary.c
cyphar: Yeah the indentation was intentional, and the fixed version looks better.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -65,19 +65,20 @@
...
 #      define SYS_memfd_create 385
-#    elif defined(__aarch64__) // arm64
+#    elif defined(__aarch64__)	// arm64
 #      define SYS_memfd_create 279
-#    elif defined(__ppc__) || defined(__ppc64__) // ppc + ppc64
+#    elif defined(__ppc__) || defined(__ppc64__)// ppc + ppc64
 #      define SYS_memfd_create 360
-#    elif defined(__s390__) || defined(__s390x__) // s390(x)
+#    elif defined(__s390__) || defined(__s390x__)
+						// s390(x)
```
thaJeztah, line 81: This one doesn't look right? I'd expected the old one here (also notice that it's using tabs to indent the _comment_, whereas lines above use spaces for indentation (but tabs for the last comments? (I'll leave a separate comment about that)

Screenshot 2021-03-18 at 10 00 24

So there's _six_ tabs in front of the comment; is it trying to fit it into columns?

Anticipating something like;

- line 78, being _slightly_ shorter; comment starts at position 49 (so don't add tabs to align)
- line 80: comment started at position 50/51; 
    - -> "just beyond the cut"
    - -> add 6 tabs (24 positions?) to put it in the next column
    - -> now line-width is at 50+24+10 (comment-length)
    - -> which is > 80 characters, so wrap the line


That said, the comment at line 376 did _not_ get reformatted/wrapped, and is at (assuming tab=4) 100 positions wide, so not sure.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -65,19 +65,20 @@
...
+#    elif defined(__ia64__)	// ia64
 #      define SYS_memfd_create 1340
-#    elif defined(__arm__) // arm
+#    elif defined(__arm__)	// arm
 #      define SYS_memfd_create 385
-#    elif defined(__aarch64__) // arm64
+#    elif defined(__aarch64__)	// arm64
 #      define SYS_memfd_create 279
-#    elif defined(__ppc__) || defined(__ppc64__) // ppc + ppc64
+#    elif defined(__ppc__) || defined(__ppc64__)// ppc + ppc64
```
thaJeztah, line 78: This one doesn't look right; why is it removing the whitespace?

libcontainer/nsenter/cloned_binary.c
```diff
@@ -65,19 +65,20 @@
...
+#    elif defined(__i386__)	// i386
 #      define SYS_memfd_create 356
-#    elif defined(__ia64__) // ia64
+#    elif defined(__ia64__)	// ia64
 #      define SYS_memfd_create 1340
-#    elif defined(__arm__) // arm
+#    elif defined(__arm__)	// arm
 #      define SYS_memfd_create 385
-#    elif defined(__aarch64__) // arm64
+#    elif defined(__aarch64__)	// arm64
```
thaJeztah, line 76: I should probbably point this out as well, as it's difficult to see on GitHub;

```c
#    elif defined(__aarch64__)	// arm64
  ^                           ^
  spaces                      tab
```

Perhaps that's standard, but it felt a bit odd that the code uses tabs (both for indentation _and_ for aligning?), but in these blocks, _spaces_ are used for indentation, which felt a bit odd (perhaps it's convention/standard?). Of course, I'm also used to Go's formatting now (tabs for indentation, but spaces for aligning).

libcontainer/nsenter/cloned_binary.c
```diff
@@ -372,35 +372,35 @@ static int seal_execfd(int *fd, int fdtype)
 	switch (fdtype) {
 	case EFD_MEMFD:
 		return fcntl(*fd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
-	case EFD_FILE: {
-		/* Need to re-open our pseudo-memfd as an O_PATH to avoid execve(2) giving -ETXTBSY. */
-		int newfd;
-		char fdpath[PATH_MAX] = {0};
+	case EFD_FILE:{
+			/* Need to re-open our pseudo-memfd as an O_PATH to avoid execve(2) giving -ETXTBSY. */
```
thaJeztah, line 376: This is the only line > 80-chars; looks like it does not reformat comments, so perhaps manually wrap this one;

```suggestion
			/*
			 * Need to re-open our pseudo-memfd as an O_PATH to avoid
			 * execve(2) giving -ETXTBSY.
			 */
```

thaJeztah (changes requested):

libcontainer/nsenter/cloned_binary.c
```diff
@@ -372,35 +372,35 @@ static int seal_execfd(int *fd, int fdtype)
 	switch (fdtype) {
 	case EFD_MEMFD:
 		return fcntl(*fd, F_ADD_SEALS, RUNC_MEMFD_SEALS);
-	case EFD_FILE: {
```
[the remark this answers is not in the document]
kolyshkin: It is removing the whitespace here as I have removed the `{ }` block (see description in the first commit).

libcontainer/nsenter/cloned_binary.c
kolyshkin: Spaces are used to indent the nested preprocessor directives as described above; let me quote myself

> Adds a few more options to indent. This was mostly motivated by
trying to save the existing formatting as much as possible, minimizing
the amount of changes indent produces.
> ...
> * -ppi2: sets the indentation for nested preprocessor directives
to 2 spaces (same as in current code defining SYS_memfd_create);

Current code use ~three~ four different ways to indent the nested cpp directives
(no indentation, tabs, two spaces, spaces and tabs). I have settled on two spaces.

As I said the idea was to minimize the amount of changes, and yet keep the formatting
consistent. I will be happy to reconsider (there's also an option to ignore cpp indentation).

libcontainer/nsenter/cloned_binary.c
kolyshkin: Currently the max width is set to 120 for indent (and it was that way since indent was introduced).

Oh, and yes, indent seem to not try to break very long comments (see my comments in PR while I tested whether the check works).

As this comments fits into 120 columns I'm not going to touch it.

libcontainer/nsenter/cloned_binary.c
kolyshkin: I don't like it either, but this is the way indent works (trying to bring the adjacent comments to the same column, and, I suspect, have some predefined preferred column position for such one-line comments).

Overall it makes sense to me -- this is how it looks like for me:

Screenshot from 2021-03-18 16-43-41

(note that ppc and and s390 are aligned -- I suspect they are not for you as you have tabstop != 8)

libcontainer/nsenter/cloned_binary.c
kolyshkin: See above

libcontainer/nsenter/cloned_binary.c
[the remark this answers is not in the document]
kolyshkin: Sorry I was talking about a different hunk.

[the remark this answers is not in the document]
kolyshkin: I guess `indent` has a flag for this one, too, but I'm too lazy to look for one, since this change only happens once, so adding that flag won't result in drastic changes in the patch size, and I don't want to have too many flags.

libcontainer/nsenter/cloned_binary.c
kolyshkin: BTW one way to fix this is remove those comments (they are kinda redundant anyway)

kolyshkin: Removed `// $ARCH` comments

AkihiroSuda (approved):

libcontainer/nsenter/nsexec.c
```diff
@@ -102,16 +102,16 @@ static int logfd = -1;
...
-#define UIDMAP_ATTR			27283
-#define GIDMAP_ATTR			27284
+#define UIDMAP_ATTR		27283
+#define GIDMAP_ATTR		27284
 #define SETGROUP_ATTR		27285
 #define OOM_SCORE_ADJ_ATTR	27286
 #define ROOTLESS_EUID_ATTR	27287
-#define UIDMAPPATH_ATTR	    27288
-#define GIDMAPPATH_ATTR	    27289
+#define UIDMAPPATH_ATTR		27288
```
mrunalp, line 113: Is the alignment still off for the last two?

kolyshkin: It was a mix of tabs and spaces, now it's just tabs and it looks aligned for tabstep=8.

GH does not use tabstep=8 so probably why it looks skewed here.

thaJeztah: Looks like you can adjust with `?ts=<width>` when viewing; https://github.com/tiimgreen/github-cheat-sheet#adjust-tab-space

kolyshkin: @thaJeztah does not seem to work when viewing diffs :-\

thaJeztah: 😕😕😞😞😩

mrunalp (approved):

thaJeztah (approved):
LGTM

I did see one line still having some alignment issues in `nsexec.c`; probably not a showstopper, but leaving screenshots from GitHub, my IDE, and a command-line editor;

Screenshot 2021-03-27 at 19 19 27
Screenshot 2021-03-27 at 19 19 41
Screenshot 2021-03-27 at 19 21 07

kolyshkin:
@thaJeztah fixing this one in https://github.com/opencontainers/runc/pull/2836

PR #2799, opened 2021-02-08
Fix data races, enable testing with -race

kolyshkin:
1. As reported by go test -race ./libcontainer/configs:
```
=== RUN   TestCommandHookRunTimeout
==================
WARNING: DATA RACE
Read at 0x00c000202230 by goroutine 23:
  os/exec.(*Cmd).Wait()
      /usr/lib/golang/src/os/exec/exec.go:502 +0x91
  github.com/opencontainers/runc/libcontainer/configs.Command.Run()
      /home/kir/go/src/github.com/opencontainers/runc/libcontainer/configs/config.go:390 +0x58c
  github.com/opencontainers/runc/libcontainer/configs_test.TestCommandHookRunTimeout()
      /home/kir/go/src/github.com/opencontainers/runc/libcontainer/configs/config_test.go:223 +0x3ed
  testing.tRunner()
      /usr/lib/golang/src/testing/testing.go:1123 +0x202

Previous write at 0x00c000202230 by goroutine 27:
  os/exec.(*Cmd).Wait()
      /usr/lib/golang/src/os/exec/exec.go:505 +0xb4
  github.com/opencontainers/runc/libcontainer/configs.Command.Run.func1()
      /home/kir/go/src/github.com/opencontainers/runc/libcontainer/configs/config.go:373 +0x55

Goroutine 23 (running) created at:
  testing.(*T).Run()
      /usr/lib/golang/src/testing/testing.go:1168 +0x5bb
  testing.runTests.func1()
      /usr/lib/golang/src/testing/testing.go:1439 +0xa6
  testing.tRunner()
      /usr/lib/golang/src/testing/testing.go:1123 +0x202
  testing.runTests()
      /usr/lib/golang/src/testing/testing.go:1437 +0x612
  testing.(*M).Run()
      /usr/lib/golang/src/testing/testing.go:1345 +0x3b3
  main.main()
      _testmain.go:69 +0x236

Goroutine 27 (running) created at:
  github.com/opencontainers/runc/libcontainer/configs.Command.Run()
      /home/kir/go/src/github.com/opencontainers/runc/libcontainer/configs/config.go:372 +0x415
  github.com/opencontainers/runc/libcontainer/configs_test.TestCommandHookRunTimeout()
      /home/kir/go/src/github.com/opencontainers/runc/libcontainer/configs/config_test.go:223 +0x3ed
  testing.tRunner()
      /usr/lib/golang/src/testing/testing.go:1123 +0x202
==================
[...]

PR #2862, opened 2021-03-18
Renew dbus connection

wzshiming:
After the current dbus restarts, legacyManager will always return an error 'dbus: connection closed by user'
This PR fixes when dbus returns this error, try to reconnect to ensure that subsequent calls are normal
xref https://github.com/kubernetes/kubernetes/issues/100328

Please review the changes, if there is no problem, I will to add a test.

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -2,7 +2,6 @@ package systemd
 
 import (
 	"bufio"
-	"context"
 	"fmt"
 	"math"
 	"os"
@@ -29,10 +28,6 @@ const (
 )
 
 var (
-	connOnce sync.Once
-	connDbus *systemdDbus.Conn
-	connErr  error
-
 	versionOnce sync.Once
 	version     int
 
@@ -290,19 +285,6 @@ func generateDeviceProperties(rules []*devices.Rule) ([]systemdDbus.Property, er
 
 	properties = append(properties, newProp("DeviceAllow", deviceAllowList))
 	return properties, nil
-}
-
-// getDbusConnection lazy initializes systemd dbus connection
-// and returns it
-func getDbusConnection(rootless bool) (*systemdDbus.Conn, error) {
-	connOnce.Do(func() {
-		if rootless {
-			connDbus, connErr = NewUserSystemdDbus()
-		} else {
-			connDbus, connErr = systemdDbus.NewWithContext(context.TODO())
-		}
-	})
-	return connDbus, connErr
 }
 
 func newProp(name string, units interface{}) systemdDbus.Property {
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -312,6 +312,16 @@ func isUnitExists(err error) bool {
 	return false
 }
 
+// isDbusClosed returns true if the error is that connection closed.
+func isDbusClosed(err error) bool {
+	if err != nil {
+		if dbusError, ok := err.(dbus.Error); ok {
+			return strings.Contains(dbusError.Name, "connection closed by user")
+		}
+	}
+	return false
+}
+
 func startUnit(dbusConnection *systemdDbus.Conn, unitName string, properties []systemdDbus.Property) error {
 	statusChan := make(chan string, 1)
 	if _, err := dbusConnection.StartTransientUnit(unitName, "replace", properties, statusChan); err == nil {
--- a/libcontainer/cgroups/systemd/dbus.go
+++ b/libcontainer/cgroups/systemd/dbus.go
@@ -54,7 +54,7 @@ func (d *dbusConnManager) getConnection() (*systemdDbus.Conn, error) {
 
 func (d *dbusConnManager) newConnection() (*systemdDbus.Conn, error) {
 	if d.rootless {
-		return NewUserSystemdDbus()
+		return newUserSystemdDbus()
 	}
 	return systemdDbus.NewWithContext(context.TODO())
 }
--- a/libcontainer/cgroups/systemd/dbus.go
+++ b/libcontainer/cgroups/systemd/dbus.go
@@ -1,0 +1,59 @@
+// +build linux
+
+package systemd
+
+import (
+	"context"
+	"sync"
+
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+)
+
+type dbusConnManager struct {
+	conn     *systemdDbus.Conn
+	rootless bool
+	sync.RWMutex
+}
+
+// newDbusConnManager initializes systemd dbus connection
+func newDbusConnManager(rootless bool) *dbusConnManager {
+	return &dbusConnManager{
+		rootless: rootless,
+	}
+}
+
+// getConnection lazy initializes systemd dbus connection and returns it
+func (d *dbusConnManager) getConnection() (*systemdDbus.Conn, error) {
+	// In the case where d.conn != nil
+	// Use the read lock the first time to ensure
+	// that Conn can be acquired at the same time.
+	d.RLock()
+	if conn := d.conn; conn != nil {
+		d.RUnlock()
+		return conn, nil
+	}
+	d.RUnlock()
+
+	// In the case where d.conn == nil
+	// Use write lock to ensure that only one
+	// will be created
+	d.Lock()
+	defer d.Unlock()
+	if conn := d.conn; conn != nil {
+		return conn, nil
+	}
+
+	conn, err := d.newConnection()
+	if err != nil {
+		return nil, err
+	}
+	d.conn = conn
+	return conn, nil
+}
+
+func (d *dbusConnManager) newConnection() (*systemdDbus.Conn, error) {
+	if d.rootless {
+		return NewUserSystemdDbus()
+	}
+	return systemdDbus.NewWithContext(context.TODO())
+}
--- a/libcontainer/cgroups/systemd/dbus.go
+++ b/libcontainer/cgroups/systemd/dbus.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/sirupsen/logrus"
 )
 
 type dbusConnManager struct {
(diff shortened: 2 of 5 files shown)


libcontainer/cgroups/systemd/common.go
```diff
@@ -295,15 +296,23 @@ func generateDeviceProperties(rules []*devices.Rule) ([]systemdDbus.Property, er
...
-			connDbus, connErr = systemdDbus.New()
-		}
+		isRootless = rootless
+		connDbus, connErr = newDbusConnection(rootless)
 	})
 	return connDbus, connErr
 }
 
+func renewDbusConnection() {
+	connDbus, connErr = newDbusConnection(isRootless)
```
cyphar, line 306: This is a data race -- two threads could call `renewDbusConnection` simultaneously. If we need to set this conditionally (potentially multiple times) we probably need a mutex.

cyphar: (Another option would be https://pkg.go.dev/golang.org/x/sync/singleflight but that's probably overkill.)

wzshiming: Updated, what do you think of this.

libcontainer/cgroups/systemd/common.go
```diff
@@ -367,6 +394,8 @@ func stopUnit(dbusConnection *systemdDbus.Conn, unitName string) error {
 		case <-time.After(time.Second):
 			logrus.Warnf("Timed out while waiting for StopUnit(%s) completion signal from dbus. Continuing...", unitName)
 		}
+	} else if isDbusClosed(err) {
+		renewDbusConnection()
```
kolyshkin, line 398: ditto

libcontainer/cgroups/systemd/v1.go
```diff
@@ -383,6 +383,9 @@ func (m *legacyManager) Set(container *configs.Config) error {
 
 	if err := dbusConnection.SetUnitProperties(getUnitName(container.Cgroups), true, properties...); err != nil {
 		_ = m.Freeze(targetFreezerState)
+		if isDbusClosed(err) {
+			renewDbusConnection()
```
kolyshkin, line 387: ditto

libcontainer/cgroups/systemd/common.go
```diff
@@ -295,15 +298,26 @@ func generateDeviceProperties(rules []*devices.Rule) ([]systemdDbus.Property, er
...
-			connDbus, connErr = systemdDbus.New()
-		}
+		isRootless = rootless
+		connDbus, connErr = newDbusConnection(rootless)
 	})
 	return connDbus, connErr
 }
 
+func renewDbusConnection() {
+	if atomic.CompareAndSwapInt32(&isRenewing, 0, 1) {
```
kolyshkin, line 308: Too bad there's no higher-level primitive in golang for this.

Makes sense at least add a comment.

libcontainer/cgroups/systemd/common.go
```diff
@@ -28,9 +29,11 @@ const (
...
 
 var (
-	connOnce sync.Once
-	connDbus *systemdDbus.Conn
-	connErr  error
+	isRootless bool
+	isRenewing int32
+	connOnce   sync.Once
+	connDbus   *systemdDbus.Conn
+	connErr    error
```
cyphar, line 36: Honestly, the longer I look at this the less convinced I am that we really need to make this global. How much performance are we really getting from stashing this in a global, and if we add locks I imagine this will only get less performant (not to mention more complicated).

I looked at the git history and it seems like this has always been global, so my guess is that it was written that way because it seemed right at the time -- I would suggest we remove this and make it local and not worry about creating the connection a few more times.

cyphar (changes requested):

libcontainer/cgroups/systemd/common.go
cyphar: Yeah this logic came from pre-libcontainer docker (3ce347c35fd45aab9da0c3f94c5c4b4e0bc49d44) and I'm not sure it's worth looking into it much further.

cyphar: Okay, it has always been implemented this way (moby/moby@6c7835050e53b733181ddfca6152c358fd625400). @alexlarsson, I know it's been a while but is there a particular reason you implemented it this way -- would it be okay to make it non-global?

libcontainer/cgroups/systemd/common.go
```diff
@@ -329,6 +343,16 @@ func isUnitExists(err error) bool {
 	return false
 }
 
+// isDbusClosed returns true if the error is that connection closed.
+func isDbusClosed(err error) bool {
+	if err != nil {
+		if dbusError, ok := err.(dbus.Error); ok {
```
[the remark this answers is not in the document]
wzshiming: This is following https://github.com/opencontainers/runc/blob/0ae14750669f42209a1f974943408df754148e9d/libcontainer/cgroups/systemd/common.go#L322-L330

wzshiming:
@cyphar @kolyshkin
Thanks, updated.

libcontainer/cgroups/systemd/common.go
kolyshkin: > If you run this code with go build -race it almost certainly will trigger race warnings.

As of https://github.com/opencontainers/runc/pull/2799 we already build runc and unit tests with -race enabled (and run all tests). I guess it makes sense to add a test case to test the renew functionality (not quite sure how to do that though, will take a look later).

libcontainer/cgroups/systemd/dbus.go
```diff
@@ -0,0 +1,66 @@
...
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+)
+
+type dbusConnManager struct {
+	conn     *systemdDbus.Conn
+	rootless bool
+	mut      sync.RWMutex
+}
+
+// newDbusConn initializes systemd dbus connection
```
kolyshkin, line 15: nit: `s/newDbusConn/newDbusConnManager/`

Also please end a sentence with a period.

libcontainer/cgroups/systemd/dbus.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"sync"
+
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+)
+
+type dbusConnManager struct {
+	conn     *systemdDbus.Conn
+	rootless bool
+	mut      sync.RWMutex
```
kolyshkin, line 12: You can include it anonymously, i.e.
```suggestion
	sync.RWMutex
```

and use `d.RLock()` etc below.

libcontainer/cgroups/systemd/dbus.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, err
+	}
+	d.conn = conn
+	return conn, nil
+}
+
+func (d *dbusConnManager) newDbusConn() (*systemdDbus.Conn, error) {
+	if d.rootless {
+		return NewUserSystemdDbus()
```
kolyshkin, line 53: Hmm @AkihiroSuda do you think we should privatize this function? I don't see it used anywhere outside of runc/libcontainer/cgroups/systemd

libcontainer/cgroups/systemd/v1.go
```diff
@@ -164,7 +166,7 @@ func (m *legacyManager) Apply(pid int) error {
 	properties = append(properties,
 		newProp("DefaultDependencies", false))
 
-	dbusConnection, err := getDbusConnection(false)
+	dbusConnection, err := m.dbus.getDbusConnection()
```
kolyshkin, line 169: I'd rename '`getDbusConnection` to `getConnection`, to avoid tautology here.

libcontainer/cgroups/systemd/dbus.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return systemdDbus.New()
+}
+
+// resetDbusConn reset to initial state
+func (d *dbusConnManager) resetDbusConn() {
```
kolyshkin, line 59: I don't see this method used anywhere. Perhaps it should be introduced in the next commit?

libcontainer/cgroups/systemd/common.go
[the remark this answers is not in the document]
kolyshkin: One way to fix this is to add a commit (before your main changes) that changes the code here to use `errors.As()`, then follow the new style.

Or leave it as is, I am fine with that.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -225,6 +229,10 @@ func (m *legacyManager) Destroy() error {
 	unitName := getUnitName(m.cgroups)
 
 	stopErr := stopUnit(dbusConnection, unitName)
+	// reset it if Dbus is closed
+	if isDbusClosed(stopErr) {
+		m.dbus.resetDbusConn()
```
kolyshkin, line 234: So, as I asked before -- do we want to retry in such case?

kolyshkin: Same question to similar cases below

wzshiming: Call to `getConnection` will retry, i think this should be enough.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -216,13 +222,17 @@ func (m *legacyManager) Destroy() error {
...
 
-	dbusConnection, err := getDbusConnection(false)
+	dbusConnection, err := m.dbus.getConnection()
 	if err != nil {
 		return err
 	}
 	unitName := getUnitName(m.cgroups)
 
 	stopErr := stopUnit(dbusConnection, unitName)
+	// reset it if Dbus is closed
```
kolyshkin, line 232: So, `stopUnit` do not ever return a dbus error.

I think it would be better to make it (as well as startUnit) accept `m.dbus` as the first argument, and do the close (and maybe retry) logic within.

libcontainer/cgroups/systemd/v1.go
kolyshkin: So yes, this is definitely an improvement as we're no longer stuck in "dbus: connection closed by user" state.

I think this may be improved further if any operation that requires dbus and receives this error will not only close the connection, but also retry (i.e. get a new connection, retry the operation).

Maybe it makes sense to add a method to dbusConnManager which will execute a function, check if the error it returns is "dbus closed" and if it is, close, reconnect and retry. This way we can wrap all the calls into this (rather than modify the code in a bunch of places to check for this particular error and call close).

wzshiming: Thanks. updated.

wzshiming:
@kolyshkin 
Maybe I can move Dbus-related functions to `dbusConnManager` as its method?

wzshiming:
Hi @kolyshkin, could you have time to take a look.

wzshiming:
PING @kolyshkin

libcontainer/cgroups/systemd/common.go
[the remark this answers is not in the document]
cyphar: Yeah we can change this some other time.

libcontainer/cgroups/systemd/dbus.go
cyphar: Probably not a bad idea, especially since people have a tendency to use functions we export because we don't use `internal`...

wzshiming:
How can I continue to advance this PR?

kolyshkin:
@wzshiming I described the idea of retry in place above (https://github.com/opencontainers/runc/pull/2862#discussion_r600977482 and https://github.com/opencontainers/runc/pull/2862#discussion_r600975567), and I guess my description was not clear enough.

I have took your patches and added retry-in-place on top, please review https://github.com/opencontainers/runc/pull/2923

wzshiming:
@kolyshkin 
Thank you for taking over

PR #2864, opened 2021-03-19
Specconv: Add fuzzer

AdamKorcz:


AdamKorcz:
@cyphar Could I ask about this one as well. `make vendor` should be sufficient to add a new vendor, right?

libcontainer/specconv/specconv_fuzzer.go
```diff
@@ -0,0 +1,94 @@
+package specconv
```
kolyshkin, line 1: missing `// +build gofuzz`?

vendor/modules.txt
```diff
@@ -1,82 +1,63 @@
+# github.com/AdamKorcz/go-fuzz-headers v0.0.0-20210319161527-f761c2329661
+github.com/AdamKorcz/go-fuzz-headers
 # github.com/checkpoint-restore/go-criu/v4 v4.1.0
-## explicit
 github.com/checkpoint-restore/go-criu/v4
 github.com/checkpoint-restore/go-criu/v4/rpc
 # github.com/cilium/ebpf v0.2.0
-## explicit
```
kolyshkin, line 6: Not sure what you did with vendor/modules.txt but it seems wrong

go.sum
```diff
@@ -1,3 +1,5 @@
+github.com/AdamKorcz/go-fuzz-headers v0.0.0-20210319161527-f761c2329661 h1:LxxqfxscKXL1kv7QNh4nggNf4Ais8B0ME8zWMCAsttY=
```
kolyshkin, line 1: Can this package live in some org repo, rather than under your personal account?

vendor/modules.txt
AdamKorcz: This was the result of running `make vendor`, which I understood will take care of everything when introducing a new dependency.

AdamKorcz: That part is the result of running `make vendor` after having introduced a new dependency. Am I missing something or shouldn't `make vendor` handle new dependencies?

AdamKorcz:
@kolyshkin The errors have been resolved.

kolyshkin:
1. Needs a rebase
2. The second commit does not make sense.

kolyshkin:
I'm inclined to close this one (and other similar PRs) as we were unable to merge this for the last few years.

PR #2866, opened 2021-03-21
seccomp: fix the inconsistent check of seccomp mode

KentaTada:
When it reads from /proc/self/status,
runc cannot check the seccomp mode until seccomp is actually set up.
And the current code just checks the existence of key which is "Seccomp".
On the other hand, Prctl() shows the supported mode corretly.
So I just removed the code which parses /proc/self/status.


kolyshkin:
LGTM, except I think you forgot to `git rm libcontainer/seccomp/fixtures/proc_self_status`

kolyshkin:
I think the reason why parsing of /proc/self/status was added by commit 41edbeb25e73d0d is to avoid a chance of runc being killed. Quoting prctl(2) man page:

```
       PR_GET_SECCOMP (since Linux 2.6.23)
              Return (as the function result) the secure computing mode of the
              calling  thread.  If the caller is not in secure computing mode,
              this operation returns 0; if the caller is in strict secure com‐
              puting  mode,  then the prctl() call will cause a SIGKILL signal
              to be sent to the process.  If the caller is in filter mode, and
              this  system  call is allowed by the seccomp filters, it returns
              2; otherwise, the process is killed with a SIGKILL signal.  This
              operation  is  available  only  if the kernel is configured with
              CONFIG_SECCOMP enabled.

              Since Linux 3.8, the Seccomp  field  of  the  /proc/[pid]/status
              file  provides a method of obtaining the same information, with‐
              out the risk that the process is killed; see proc(5).
```

Now the question is, what is the chance of `prctl(2)` being denied for `runc`, and should we keep the workaround?

KentaTada:
>LGTM, except I think you forgot to git rm libcontainer/seccomp/fixtures/proc_self_status

I'm sorry. I removed it at first.

>Now the question is, what is the chance of prctl(2) being denied for runc, and should we keep the workaround?

Thanks!
Hmm... 
The feature of seccomp is not always enabled.
So I think we should not remove the code of parsing of /proc/self/status.
But strictly speaking, the existence of the key of "Seccomp" is not meaning enabling CONFIG_SECCOMP_FILTER.

kolyshkin:
I don't see any use of `seccomp.IsEnabled()` except for printing seccomp version in `main.go` (which I believe is wrong -- we should check whether seccomp is compiled in, not enabled).

Which user(s) (and/or use cases) did you have in mind when fixing this, @KentaTada ?

kolyshkin:
I saw that the code from runc was copied (without much changes) to
 - containerd/containerd (https://github.com/containerd/containerd/blob/master/pkg/seccomp/seccomp_linux.go)
 - containers/common (https://github.com/containers/common/blob/master/pkg/seccomp/supported.go)
 
All the implementations are as questionable as the one in runc, and I am not sure about their usage, too (runc does not have users, apparently, except for the main that prints seccomp version, see above).

kolyshkin:
> I don't see any use of seccomp.IsEnabled() except for printing seccomp version in main.go (which I believe is wrong -- we should check whether seccomp is compiled in, not enabled).

Created https://github.com/opencontainers/runc/pull/2870

kolyshkin:
As for `IsEnabled`, maybe we should just drop it as we do not use it (nor can I find any external users -- but maybe I overlooked it).

kolyshkin:
> As for IsEnabled, maybe we should just drop it as we do not use it (nor can I find any external users -- but maybe I overlooked it).

Updated #2870

kolyshkin:
> I saw that the code from runc was copied (without much changes) to
> - containerd/containerd (https://github.com/containerd/containerd/blob/master/pkg/seccomp/seccomp_linux.go)

Opened https://github.com/containerd/containerd/pull/5256

> - containers/common (https://github.com/containers/common/blob/master/pkg/seccomp/supported.go)

Opened https://github.com/containers/common/pull/483

KentaTada:
>Which user(s) (and/or use cases) did you have in mind when fixing this, @KentaTada ?

We need the Golang function to check the facilities of kernel features for seccomp.
I thought _libcontainer/seccomp_ was suitable for that.

> > I saw that the code from runc was copied (without much changes) to
> > 
> > * containerd/containerd (https://github.com/containerd/containerd/blob/master/pkg/seccomp/seccomp_linux.go)
> 
> Opened [containerd/containerd#5256](https://github.com/containerd/containerd/pull/5256)
> 
> > * containers/common (https://github.com/containers/common/blob/master/pkg/seccomp/supported.go)
> 
> Opened [containers/common#483](https://github.com/containers/common/pull/483)

So, in my opinion, libcontainer provides **seccomp.IsEnabled()** and above software just use libcontainer.

kolyshkin:
> So, in my opinion, libcontainer provides seccomp.IsEnabled() and above software just use libcontainer.

Well, it does not. They copy/pasted the code instead, and I guess it's OK. If you need this, I suggest you use `github.com/containers/common/pkg/seccomp` (where https://github.com/containers/common/pull/483 is already merged) or `github.com/containerd/containerd/pkg/seccomp`.

kolyshkin:
I suggest we close this one in favor of https://github.com/opencontainers/runc/pull/2870

kolyshkin:
Closing in favor of (merged) #2870

PR #2869, opened 2021-03-23
Fix oss-fuzz build

dqminh:
THis fixes incorrect module path and also add proper tags for
FuzzUIDMap.

Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32109#c1

### DIFF
--- a/tests/fuzzing/oss_fuzz_build.sh
+++ b/tests/fuzzing/oss_fuzz_build.sh
@@ -4,10 +4,10 @@
 # if run outside of it.
 # The api, compile_go_fuzzer() is provided by the OSS-fuzz
 # environment and is a high level helper function for a series
-# of compilation and linking steps to build the fuzzers in the 
+# of compilation and linking steps to build the fuzzers in the
 # OSS-fuzz environment.
 # More info about compile_go_fuzzer() can be found here:
 #     https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/#buildsh
-compile_go_fuzzer ./libcontainer/system FuzzUIDMap id_map_fuzzer linux
-compile_go_fuzzer ./libcontainer/user FuzzUser user_fuzzer
-compile_go_fuzzer ./libcontainer/configs FuzzUnmarshalJSON configs_fuzzer
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/system FuzzUIDMap id_map_fuzzer linux,gofuzz
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/user FuzzUser user_fuzzer
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/configs FuzzUnmarshalJSON configs_fuzzer


kolyshkin:
@AdamKorcz PTAL

kolyshkin (approved):
LGTM, and thanks!

tests/fuzzing/oss_fuzz_build.sh
```diff
@@ -4,10 +4,10 @@
...
+# of compilation and linking steps to build the fuzzers in the
 # OSS-fuzz environment.
 # More info about compile_go_fuzzer() can be found here:
 #     https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/#buildsh
-compile_go_fuzzer ./libcontainer/system FuzzUIDMap id_map_fuzzer linux
-compile_go_fuzzer ./libcontainer/user FuzzUser user_fuzzer
-compile_go_fuzzer ./libcontainer/configs FuzzUnmarshalJSON configs_fuzzer
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/system FuzzUIDMap id_map_fuzzer linux,gofuzz
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/user FuzzUser user_fuzzer
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/configs FuzzUnmarshalJSON configs_fuzzer
```
kolyshkin, line 13: Wonder if we need `gofuzz` tag for these lines as well

tests/fuzzing/oss_fuzz_build.sh
```diff
@@ -4,10 +4,10 @@
...
 # environment and is a high level helper function for a series
-# of compilation and linking steps to build the fuzzers in the 
+# of compilation and linking steps to build the fuzzers in the
 # OSS-fuzz environment.
 # More info about compile_go_fuzzer() can be found here:
 #     https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/#buildsh
-compile_go_fuzzer ./libcontainer/system FuzzUIDMap id_map_fuzzer linux
-compile_go_fuzzer ./libcontainer/user FuzzUser user_fuzzer
-compile_go_fuzzer ./libcontainer/configs FuzzUnmarshalJSON configs_fuzzer
+compile_go_fuzzer github.com/opencontainers/runc/libcontainer/system FuzzUIDMap id_map_fuzzer linux,gofuzz
```
kolyshkin, line 11: The `linux` tag is probably redundant -- as long as fuzzing is done on Linux.

AdamKorcz:
@dqminh Thank you for fixing this! 

The broken build is the coverage build. To test that build specifically locally you need to run:
`infra/helper.py build_fuzzers --sanitizer=coverage runc`

The specific part of the documentation is here: https://google.github.io/oss-fuzz/advanced-topics/code-coverage/#establish-access-to-gcs

When I test the updated build file locally I get the error:
`./fuzzuser_test.go:56:3: undefined: FuzzUser`

AdamKorcz:
I tested this one again, and it works fine with both address sanitizer and coverage sanitizer which is exactly what we need. 

My previous test might have been with an OSS-fuzz image that was not the latest.

LGTM

tests/fuzzing/oss_fuzz_build.sh
AdamKorcz: It works fine as is locally for me.

dqminh: gofuzz tag is automatically appended by compile_go_fuzzer when we don't specify any build tags

kolyshkin:
close/reopen to kick the latest ci

kolyshkin:
@AkihiroSuda PTAL

tests/fuzzing/oss_fuzz_build.sh
kolyshkin: > OSS-fuzz only fuzz on Linux.

This means that the linux tag is there implicitly (set by the platform), and so it is redundant.

kolyshkin: Anyway, this can be addressed separately

mrunalp (approved):

dqminh:
Build was fixed in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32109#c6 :partying_face:

PR #471, opened 2016-01-13
add seccomp.IsEnabled() function

jfrazelle:
This is much like apparmor.IsEnabled() function and a nice helper.

Signed-off-by: Jessica Frazelle <acidburn@docker.com>

PR #2870, opened 2021-03-23
libct/seccomp: remove IsEnabled

kolyshkin:
### 1. runc version: don't use seccomp.IsEnabled()

As pointed out to in https://github.com/opencontainers/runc/pull/2866,
seccomp.IsEnabled does some runtime checks,
while here (when printing libseccomp version) we should merely print
the version number of libseccomp, if compiled in.

Change the code to check if the version is non-zero (as seccomp.Version
returns 0, 0, 0 in case seccomp is not compiled in).

### 2. libct/seccomp: rm IsEnabled

seccomp.IsEnabled is not well defined (the presence of Seccomp: field
in /proc/self/status does not tell us whether `CONFIG_SECCOMP_FILTER`
is enabled in the kernel; parsing all keys in `/proc/self/status` is a
moderate waste of resources, etc).

I traced its addition back to [#471][1] and even in there it is not clear
what for it was added. There were never an internal user (except
for the recently added one, removed by the previous commit), and
can't find any external users (but found two copy-pastes of this
code, suffering from the same problems, see [containerd/containerd][2] and [containers/common][3]).

Since it is broken and has no users, remove it.

[1]: https://github.com/opencontainers/runc/pull/471
[2]: https://github.com/containerd/containerd/blob/master/pkg/seccomp/seccomp_linux.go
[3]: https://github.com/containers/common/blob/master/pkg/seccomp/supported.go


KentaTada:
In addition to the existence of libseccomp,
Maybe I think runc should check the facilities of seccomp(CONFIG_SECCOMP and CONFIG_SECCOMP_FILTER) in Linux Kernel like seccomp.IsEnabled.

kolyshkin:
> In addition to the existence of libseccomp, maybe I think runc should check the facilities of seccomp(CONFIG_SECCOMP and CONFIG_SECCOMP_FILTER) in Linux Kernel like seccomp.IsEnabled.

Are you talking about `runc version`? Sounds like too much for me.

kolyshkin:
Added to rc94 milestone as it's part of public API and if we won't remove this before 1.0, we'll have to keep it.

AkihiroSuda (approved):

mrunalp (approved):

PR #2872, opened 2021-03-24
Libcontainer: Add fuzzer

AdamKorcz:
This PR adds a fuzzer that passes the pseudo-random data into `state.json` from which the state of the factory is loaded when `Load()` is called.

This one should get merged after fuzzers in existing PRs get merged to keep all fuzzers of each package in a single file.


AdamKorcz:
We will wait with this one for now.

PR #2873, opened 2021-03-26
libct/cg/fs2.GetStats() improvements

kolyshkin:
Various `GetStats()` improvements, mostly to address cgroup v2 compatibility issues.
NOTE that this functionality is not used by runc itself.

## 1. libct/cg/fs2/memory: fix swap reporting
    
cgroup v1 reports combined mem+swap in stats.MemoryStats.SwapUsage.
In cgroup v2, swap is separate.
    
For the sake of compatibility, make v2 report mem+swap as well.
This also includes Limit.

## 2. libct/cg/fs[2]/getMemoryData[V2]: optimize
    
Existing code ignores `ENOENT` in case we're reading data from
controls that might not be enabled. While this is correct, the code
can be improved:
    
1. Check `name != ""` instead of `moduleName != "memory"`, as these checks
       are equivalent but the new one is faster.
    
2. It does not make sense to ignore subsequent errors -- if the control
       is not available, we won't hit this code path.
    
3. Add a comment explaining why we ignore the error.

## 3. libct/cg/fs2/memory.Stat: add usage for root cgroup

There is no memory.{current,max} for the root node in cgroup v2, and
thus stats for "/sys/fs/cgroup" return an error.

The same thing works with cgroup v1 (as there are
memory.{usage,limit}_in_bytes files in the root node).

Emulate stats for /sys/fs/cgroup by getting numbers from
/proc/self/meminfo.

[Initially, I wanted to avoid parsing yet another /proc file and
instead mock some numbers using data from memory.stat but was
unable to come up with formulae that make sense.]

## 4. libct/cg/fs2.Stat: always call statCpu

Giuseppe found out that cpu.stat for a cgroup is available even if
the cpu controller is not enabled for it. So, let's call statCpu
regradress, and ignore ENOENT.

## 5. libct/cg/fs2/getPidsWithoutController: optimize
    
It is inefficient to create an associative map for the whole purpose of
counting the number of elements in it, especially if the elements are
all unique. It uses more CPU than necessary and creates some work for
the garbage collector.
    
The file we read contains PIDs and newlines, and the easiest/fastest way
to get the number of PIDs is just to count the newlines.

## 6. libct/cg/fs2.statPids: fall back directly
    
When getting pids stats, instead of checking whether the pids controller
is available, let's use a fall back function in case pids.current does
not exist. This simplifies the logic in fs2.GetStats.

## 7. libct/cg/fs2.Stat: don't look for available controllers

Some controllers might still have stats available even if they are
disabled (this is definitely so for cpu.stat -- see earlier commit).

Some stat methods might implement sensible fallbacks (see previous
commit for statPids.

In the view of all that, it makes sense to not check if a particular
controller is available, but rather ignore ENOENT from it.

### DIFF
--- a/libcontainer/cgroups/fs2/memory.go
+++ b/libcontainer/cgroups/fs2/memory.go
@@ -4,6 +4,7 @@ package fs2
 
 import (
 	"bufio"
+	"math"
 	"os"
 	"strconv"
 
@@ -89,6 +90,9 @@ func statMemory(dirPath string, stats *cgroups.Stats) error {
 		stats.MemoryStats.Stats[t] = v
 	}
 	stats.MemoryStats.Cache = stats.MemoryStats.Stats["file"]
+	// Unlike cgroup v1 which has memory.use_hierarchy binary knob,
+	// cgroup v2 is always hierarchical.
+	stats.MemoryStats.UseHierarchy = true
 
 	memoryUsage, err := getMemoryDataV2(dirPath, "")
 	if err != nil {
@@ -99,9 +103,15 @@ func statMemory(dirPath string, stats *cgroups.Stats) error {
 	if err != nil {
 		return err
 	}
+	// As cgroup v1 reports SwapUsage values as mem+swap combined,
+	// while in cgroup v2 swap values do not include memory,
+	// report combined mem+swap for v1 compatibility.
+	swapUsage.Usage += memoryUsage.Usage
+	if swapUsage.Limit != math.MaxUint64 {
+		swapUsage.Limit += memoryUsage.Limit
+	}
 	stats.MemoryStats.SwapUsage = swapUsage
 
-	stats.MemoryStats.UseHierarchy = true
 	return nil
 }
 
--- a/libcontainer/cgroups/fs2/memory.go
+++ b/libcontainer/cgroups/fs2/memory.go
@@ -127,7 +127,10 @@ func getMemoryDataV2(path, name string) (cgroups.MemoryData, error) {
 
 	value, err := fscommon.GetCgroupParamUint(path, usage)
 	if err != nil {
-		if moduleName != "memory" && os.IsNotExist(err) {
+		if name != "" && os.IsNotExist(err) {
+			// Ignore EEXIST as there's no swap accounting
+			// if kernel CONFIG_MEMCG_SWAP is not set or
+			// swapaccount=0 kernel boot parameter is given.
 			return cgroups.MemoryData{}, nil
 		}
 		return cgroups.MemoryData{}, errors.Wrapf(err, "failed to parse %s", usage)
@@ -136,9 +139,6 @@ func getMemoryDataV2(path, name string) (cgroups.MemoryData, error) {
 
 	value, err = fscommon.GetCgroupParamUint(path, limit)
 	if err != nil {
-		if moduleName != "memory" && os.IsNotExist(err) {
-			return cgroups.MemoryData{}, nil
-		}
 		return cgroups.MemoryData{}, errors.Wrapf(err, "failed to parse %s", limit)
 	}
 	memoryData.Limit = value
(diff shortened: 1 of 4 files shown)


kolyshkin:
@giuseppe PTAL

giuseppe (approved):
thanks,

LGTM

mrunalp:
I thought that kernel was adding support for the root level memory stats for cgroups v2 cc: @giuseppe

kolyshkin:
> I thought that kernel was adding support for the root level memory stats for cgroups v2 cc: @giuseppe

If you're talking about `memory.stat`, it is present for the root cgroup, but I was not able to come up with a decent way to deduce usage from there.

If you mean `memory.current`, I have modified the last patch to check for it first, even for the root cgroup.

kolyshkin:
CI failure is a known flake (https://github.com/opencontainers/runc/issues/2805). CI restarted.

libcontainer/cgroups/fs2/memory.go
```diff
@@ -126,12 +146,93 @@ func getMemoryDataV2(path, name string) (cgroups.MemoryData, error) {
...
+	main_cached := page_cache + slab_reclaimable
+
+	// If main_available is greater than main_total or our calculation of
+	// mem_used overflows, that's symptomatic of running within a lxc
+	// container where such values will be dramatically distorted over
+	// those of the host.
+	if main_available > main_total {
+		main_available = main_free
+	}
+	mem_used := int64(main_total - main_free - main_cached - main_buffers)
```
giuseppe, line 230: cAdvisor calculates the memory working set as `MemoryUsed - stats["inactive_files"]` (https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L827-L835).  If we already drop the cached files size here, the result given by cAdvisor is always negative.  Should we just use `mem_used = int64(main_total - main_free)`?

kolyshkin: Yeah I was just mimicking whatever `free` is showing, not whatever cgroup v1's `memory.usage_in_bytes` is showing (and yes, this was a wrong thing to do). cgroup v1's and v2's (`memory.current` for non-root cgroups) DO include the page cache (which is questionable but that's the way it is).

Thanks for pointing out, fixed.

libcontainer/cgroups/fs2/memory.go
```diff
@@ -99,9 +103,15 @@ func statMemory(dirPath string, stats *cgroups.Stats) error {
 	if err != nil {
 		return err
 	}
+	// As cgroup v1 reports SwapUsage values as mem+swap combined,
```
mrunalp, line 113: Should we expose a separate field that returns just the swap for consumers? (Not a blocker for this PR)

kolyshkin: Alas, the field that holds swap+mem usage is inconveniently called "Swap".

So, we could add something like "SwapOnly" (for both v1 and v2) or "SwapV2" (for v2 only), or something entirely different -- I am very open to suggestions.

kolyshkin:
This helps cadvisor with cgroupv2 (https://github.com/kubernetes/kubernetes/pull/100631) so I'd like this to go in rc94. This won't cause any regressions in runc per se as it does not use GetStats.

mrunalp (approved):

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda (approved):

Zheaoli:
@kolyshkin Hello, I got some problem here

1. Seems `statsFromMeminfo` will never triggered by runc it self right?
2. How should I test this `statsFromMeminfo` function, seems I need to set up the `/sys/fs/cgroup` right?

Zheaoli:
> @kolyshkin Hello, I got some problem here
> 
> 1. Seems `statsFromMeminfo` will never triggered by runc it self right?
> 2. How should I test this `statsFromMeminfo` function, seems I need to set up the `/sys/fs/cgroup` right?

@kolyshkin ping QAQ

PR #2874, opened 2021-03-26
script/check-config.sh: add SELinux and AppArmor

KentaTada:
SELinux and AppArmor are always enabled since runc v1.0.0-rc93

### DIFF
--- a/script/check-config.sh
+++ b/script/check-config.sh
@@ -271,5 +271,7 @@ flags=(
 	IP_VS_PROTO_TCP
 	IP_VS_PROTO_UDP
 	IP_VS_RR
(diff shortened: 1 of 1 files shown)


kolyshkin (changes requested):
NACK. These should go in optional features. The fact that these are always enabled does not mean these are required -- it means these are supported.

KentaTada:
Thanks a lot.
I changed these options from necessary to optional features.

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

kolyshkin:
CI failure is a flake (#2805); CI restarted

PR #2875, opened 2021-03-26
go.mod: github.com/cilium/ebpf v0.4.0

thaJeztah:
full diff: https://github.com/cilium/ebpf/compare/v0.2.0...v0.4.0


kolyshkin:
Seems they've renamed `readme.md` to `README.md`
```
make: *** [Makefile:134: verify-dependencies] Error 1
 HEAD detached at pull/2875/merge
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    vendor/github.com/cilium/ebpf/readme.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	vendor/github.com/cilium/ebpf/README.md
```

thaJeztah:
Argh; will probably have to redo this from my case-sensitive partition.

I saw they mentioned breaking changes; doesn't look to be in the parts we use, but thought it would be good to at least test in some repositories if we can update

thaJeztah:
ok, looks like `git config core.ignorecase false` does the trick as well (as macOS is case-preserving, so with that enabled it also picks up the change in case only 🎉 ) 👍

thaJeztah:
Of course; that fixed it locally for me, but not in CI; added a commit that adds a `.gitconfig` to this repository, and hope that that's picked up by CI

AkihiroSuda (dismissed):

.gitconfig
```diff
@@ -0,0 +1,2 @@
+[core]
+	ignorecase = false
```
kolyshkin, line 2: Hmm... wonder why this needs to be set to `false` explicitly, if this is the default; git-config(1) says:

> The default is false, except git-clone(1) or git-init(1) will probe and set core.ignoreCase true if appropriate when the repository is created.

thaJeztah: Could the "where appropriate" be the issue when it detects a case-insensitive platform?

thaJeztah: I could try reverting the commit to see if it was just coincidence that it passed after I added it? idk

kolyshkin: If this helps devs using Mac OS X I'm fine with the change as it is.

kolyshkin:
Needs a rebase @thaJeztah

.gitconfig
thaJeztah: No it didn't change my local configuration, and I'm even in doubt now if GitHub actually uses the file, or that it was sheer luck that pushing again solved the Windows linter; I rebased, and removed the commit to confirm that suspicion

AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #2880, opened 2021-03-31
libcontainer/cgroups/systemd: don't call stopUnit

kolyshkin:
A cgroup driver manages cgroups. A systemd cgroup driver calls startUnit
to create a new systemd unit. Systemd units are auto-removed ("garbage
collected") as long as systemd sees they are finished (the definition of
"finished" depends on the unit type -- sometimes it's "no more processes
left", sometimes it's "init process exited").

A systemd cgroup driver should not call stopUnit, because case there are
processes inside cgroup, will cause systemd to TERM/KILL those, and this
is not what is expected from a cgroup driver, especially since  runc
already takes care about terminating container's processes. This is
currently manifested in https://github.com/opencontainers/runc/issues/2805 (see https://github.com/opencontainers/runc/issues/2805#issuecomment-810718209 for details).

Note we still try to remove cgroup dir(s) from Destroy, and return an
error if they are still there.

To reiterate:
 - stopUnit is not needed because systemd auto-removes finished units;
 - stopUnit can be dangerous as we do not expect that a cgroup driver
   can kill processes.

So, remove stopUnit implementation and all calls to it.

Fixes: #2805

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -354,23 +354,6 @@ func startUnit(dbusConnection *systemdDbus.Conn, unitName string, properties []s
 	return nil
 }
 
-func stopUnit(dbusConnection *systemdDbus.Conn, unitName string) error {
-	statusChan := make(chan string, 1)
-	if _, err := dbusConnection.StopUnit(unitName, "replace", statusChan); err == nil {
-		select {
-		case s := <-statusChan:
-			close(statusChan)
-			// Please refer to https://godoc.org/github.com/coreos/go-systemd/dbus#Conn.StartUnit
-			if s != "done" {
-				logrus.Warnf("error removing unit `%s`: got `%s`. Continuing...", unitName, s)
-			}
-		case <-time.After(time.Second):
-			logrus.Warnf("Timed out while waiting for StopUnit(%s) completion signal from dbus. Continuing...", unitName)
-		}
-	}
-	return nil
-}
-
 func systemdVersion(conn *systemdDbus.Conn) int {
 	versionOnce.Do(func() {
 		version = -1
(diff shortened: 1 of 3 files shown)


kolyshkin:
OK, so it seems wrong. Despite what systemd.scope(5) says about scope lifetime

> The lifecycle of the scope unit is thus not bound to the lifetime of
> one specific process, but to the existence of at least one process in the scope. 

... the unit is still active long after all its processes are gone:

```console
[kir@kir-rhat crun]$ sudo systemctl status runc-test-test16ddfwutxgjte.scope
Warning: The unit file, source configuration file or drop-ins of runc-test-test16ddfwutxgjte.scope changed on disk. Run 'systemctl daemon-reload' to reload units.
● runc-test-test16ddfwutxgjte.scope - libcontainer container test16ddfwutxgjte
     Loaded: loaded (/run/systemd/transient/runc-test-test16ddfwutxgjte.scope; transient)
  Transient: yes
    Drop-In: /run/systemd/transient/runc-test-test16ddfwutxgjte.scope.d
             └─50-DeviceAllow.conf, 50-DevicePolicy.conf, 50-MemoryMax.conf, 50-MemoryMin.conf
     Active: active (running) since Wed 2021-03-31 17:29:41 PDT; 55s ago
         IO: 0B read, 0B written
      Tasks: 0 (limit: 18933)
     Memory: 0B (min: 128.0K max: 512.0M)
        CPU: 0
     CGroup: /system.slice/runc-test-test16ddfwutxgjte.scope

Mar 31 17:29:41 kir-rhat systemd[1]: Started libcontainer container test16ddfwutxgjte.
```

So I am removing the second commit.

Yet we're doing something wrong here. Maybe we need to
 - set `KillMode=none` (if we're `runc` and we want to kill the container ourselves)
 - set `StopCommand=runc delete -f ...` for the unit, so that systemctl stop will do the right thing
 
 Just thinking out loud

kolyshkin:
The fact that only one test has failed because of this shows that our test suite has some room for improvement (to say at least).

PR #2881, opened 2021-04-01
tests/int: some refactoring, fix a flake

kolyshkin:
Fixes https://github.com/opencontainers/runc/issues/2805

See all commits for descriptions. Most important ones are:

## tests/int/cpt: fix lazy-pages flakiness

"checkpoint --lazy-pages and restore" test sometimes fails on restore
in our CI on Fedora 33 when systemd cgroup driver is used:

> (00.076104) Error (compel/src/lib/infect.c:1513): Task 48521 is in unexpected state: f7f
> (00.076122) Error (compel/src/lib/infect.c:1520): Task stopped with 15: Terminated
> ...
> (00.078246) Error (criu/cr-restore.c:2483): Restoring FAILED.

I think what happens is

1. The test runs runc checkpoint in lazy-pages mode in background.
2. The test runs criu lazy-pages in background.
3. The test runs runc restore.

Now, all three are working in together: criu restore restores, criu
lazy-pages listens for page faults on a uffd and fetch missing pages
from runc checkpoint, who serves those pages.

At some point criu lazy-pages decides to fetch the rest of the pages,
and once it's done it exits, and runc checkpoint, as there are no more
pages to serve, exits too.

At the end of runc checkpoint the container is removed (see "defer
destroy(container)" in checkpoint.go. This involves a call to
cgroupManager.Destroy, which, in case systemd manager is used,
calls stopUnit, which makes systemd to not just remove the unit,
but also send SIGTERM to its processes, if there are any.

As the container is being restored into the same systemd unit,
sometimes this results in sending SIGTERM to a process which
criu restores, and thus restoring fails.

:spiral_notepad: for slightly more detailed description of the above, see https://github.com/opencontainers/runc/issues/2805#issuecomment-810718209)

The remedy here is to change the name of systemd unit to which the
container is restored.

##     tests/int: really randomize cgroup/unit names
    
Commit 41670e21f0 added some randomization to cgroup paths
and (if systemd cgroup driver is used) systemd unit names,
but
    
 - the randomization was done only if set_cgroups_path is called,
    which is not done by every test;
    
 - the randomization was per bats instance, not per test.
    
Fix both issues by refactoring init_cgroups_path/set_cgroups_path
(moving variable part to set_cgroups_path), and calling the latter
from runc_spec, so it is now applicable to every container.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -717,7 +717,6 @@ func (c *linuxContainer) checkCriuFeatures(criuOpts *CriuOpts, rpcOpts *criurpc.
 		return errors.New("CRIU feature check failed")
 	}
 
-	logrus.Debugf("Feature check says: %s", criuFeatures)
 	missingFeatures := false
 
 	// The outer if checks if the fields actually exist
@@ -1518,7 +1517,6 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 		// the initial CRIU run to detect the version. Skip it.
 		logrus.Debugf("Using CRIU %d at: %s", c.criuVersion, c.criuPath)
 	}
-	logrus.Debugf("Using CRIU with following args: %s", args)
 	cmd := exec.Command(c.criuPath, args...)
 	if process != nil {
 		cmd.Stdin = process.Stdin
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -189,8 +189,8 @@ function setup() {
 @test "runc run (blkio weight)" {
 	requires root cgroups_v2
 
-	set_cgroups_path "$BUSYBOX_BUNDLE"
-	update_config '.linux.resources.blockIO |= {"weight": 750}' "${BUSYBOX_BUNDLE}"
+	set_cgroups_path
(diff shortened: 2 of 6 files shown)


kolyshkin:
@adrianreber PTAL (last commit only)

kolyshkin:
Need to figure out why rootless fails after the second commit (force randomized cgroup paths). For tomorrow

adrianreber:
> @adrianreber PTAL (last commit only)

Uh... I like your analysis and the description of the solution. Sounds plausible to me. Not sure I can say anything about the actual code changes. You remove more lines than you add. That is usually a good sign :smile:

kolyshkin:
@adrianreber sorry, this is not yet ready, still working on it and I got the patch I wanted you to look at removed. Basically I just change the cgroupsPath, so the container is restored into a different cgroup. I'll let you know once it's ready.

kolyshkin:
This fixes a flake, and all the changes (except couple of removed debug prints in checkpoint code) are in tests/integration, so adding rc94 milestone.

mrunalp (approved):

kolyshkin:
@cyphar @AkihiroSuda PTAL

AkihiroSuda (approved):

PR #2883, opened 2021-04-01
Add support for rdma cgroup introduced in Linux Kernel 4.11

flouthoc:
RDMA controller/cgroup support is added with Linux kernel 4.11.
Similar format is used by cgroup-v2

Reference
https://www.kernel.org/doc/Documentation/cgroup-v1/rdma.txt

### DIFF
--- a/libcontainer/cgroups/fs/rdma.go
+++ b/libcontainer/cgroups/fs/rdma.go
@@ -1,0 +1,25 @@
+package fs
+
+import (
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+type RdmaGroup struct{}
+
+func (s *RdmaGroup) Name() string {
+	return "rdma"
+}
+
+func (s *RdmaGroup) Apply(path string, d *cgroupData) error {
+	return join(path, d.pid)
+}
+
+func (s *RdmaGroup) Set(path string, r *configs.Resources) error {
+	return fscommon.RdmaSet(path, r)
+}
+
+func (s *RdmaGroup) GetStats(path string, stats *cgroups.Stats) error {
+	return fscommon.RdmaGetStats(path, stats)
+}
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -125,6 +125,10 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 	if err := statHugeTlb(m.dirPath, st); err != nil && !os.IsNotExist(err) {
 		errs = append(errs, err)
 	}
+	// rdma (since kernel 4.11)
+	if err := fscommon.RdmaGetStats(m.dirPath, st); err != nil && !os.IsNotExist(err) {
+		errs = append(errs, err)
+	}
 	if len(errs) > 0 && !m.rootless {
 		return st, fmt.Errorf("error while statting cgroup v2: %+v", errs)
 	}
@@ -181,6 +185,10 @@ func (m *manager) Set(r *configs.Resources) error {
 	}
 	// hugetlb (since kernel 5.6)
 	if err := setHugeTlb(m.dirPath, r); err != nil {
+		return err
+	}
+	// rdma (since kernel 4.11)
+	if err := fscommon.RdmaSet(m.dirPath, r); err != nil {
 		return err
 	}
 	// freezer (since kernel 5.2, pseudo-controller)
--- a/libcontainer/cgroups/fscommon/rdma.go
+++ b/libcontainer/cgroups/fscommon/rdma.go
@@ -1,0 +1,121 @@
+package fscommon
+
+import (
+	"bufio"
+	"errors"
+	"math"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"golang.org/x/sys/unix"
+)
+
+// parseRdmaKV parses raw string to RdmaEntry.
+func parseRdmaKV(raw string, entry *cgroups.RdmaEntry) error {
+	var value uint32
+
+	parts := strings.SplitN(raw, "=", 3)
+
+	if len(parts) != 2 {
+		return errors.New("Unable to parse RDMA entry")
+	}
+
+	k, v := parts[0], parts[1]
+
+	if v == "max" {
+		value = math.MaxUint32
+	} else {
+		val64, err := strconv.ParseUint(v, 10, 32)
+		if err != nil {
+			return err
+		}
+		value = uint32(val64)
+	}
+	if k == "hca_handle" {
+		entry.HcaHandles = value
+	} else if k == "hca_object" {
+		entry.HcaObjects = value
+	}
+
+	return nil
+}
+
+// readRdmaEntries reads and converts array of rawstrings to RdmaEntries from file.
+// example entry: mlx4_0 hca_handle=2 hca_object=2000
+func readRdmaEntries(dir, file string) ([]cgroups.RdmaEntry, error) {
+	rdmaEntries := make([]cgroups.RdmaEntry, 0)
+	fd, err := cgroups.OpenFile(dir, file, unix.O_RDONLY)
+	if err != nil {
+		return nil, err
+	}
+	defer fd.Close() //nolint:errorlint
+	scanner := bufio.NewScanner(fd)
+	for scanner.Scan() {
+		parts := strings.SplitN(scanner.Text(), " ", 4)
+		if len(parts) == 3 {
+			entry := new(cgroups.RdmaEntry)
+			entry.Device = parts[0]
+			err = parseRdmaKV(parts[1], entry)
+			if err != nil {
+				continue
+			}
+			err = parseRdmaKV(parts[2], entry)
+			if err != nil {
+				continue
+			}
+
+			rdmaEntries = append(rdmaEntries, *entry)
+		}
+	}
+	return rdmaEntries, scanner.Err()
+}
+
+// RdmaGetStats returns rdma stats such as totalLimit and current entries.
+func RdmaGetStats(path string, stats *cgroups.Stats) error {
+	currentEntries, err := readRdmaEntries(path, "rdma.current")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			err = nil
+		}
+		return err
+	}
+	maxEntries, err := readRdmaEntries(path, "rdma.max")
+	if err != nil {
+		return err
+	}
+	// If device got removed between reading two files, ignore returning stats.
+	if len(currentEntries) != len(maxEntries) {
+		return nil
+	}
+
+	stats.RdmaStats = cgroups.RdmaStats{
+		RdmaLimit:   maxEntries,
+		RdmaCurrent: currentEntries,
+	}
+
+	return nil
+}
+
+func createCmdString(device string, limits configs.LinuxRdma) string {
+	cmdString := device
+	if limits.HcaHandles != nil {
+		cmdString += " hca_handle=" + strconv.FormatUint(uint64(*limits.HcaHandles), 10)
+	}
+	if limits.HcaObjects != nil {
+		cmdString += " hca_object=" + strconv.FormatUint(uint64(*limits.HcaObjects), 10)
+	}
+	return cmdString
+}
+
+// RdmaSet sets RDMA resources.
+func RdmaSet(path string, r *configs.Resources) error {
+	for device, limits := range r.Rdma {
+		if err := cgroups.WriteFile(path, "rdma.max", createCmdString(device, limits)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/libcontainer/cgroups/fscommon/rdma_test.go
+++ b/libcontainer/cgroups/fscommon/rdma_test.go
@@ -1,0 +1,57 @@
+package fscommon
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+/* Roadmap for future */
+// (Low-priority)  TODO: Check if it is possible to virtually mimic an actual RDMA device.
+// TODO: Think of more edge-cases to add.
+
+// TestRdmaSet performs an E2E test of RdmaSet(), parseRdmaKV() using dummy device and a dummy cgroup file-system.
+// Note: Following test does not guarantees that your host supports RDMA since this mocks underlying infrastructure.
+func TestRdmaSet(t *testing.T) {
+	testCgroupPath := filepath.Join(t.TempDir(), "rdma")
+
+	// Ensure the full mock cgroup path exists.
+	err := os.Mkdir(testCgroupPath, 0o755)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rdmaDevice := "mlx5_1"
+	maxHandles := uint32(100)
+	maxObjects := uint32(300)
+
+	rdmaStubResource := &configs.Resources{
+		Rdma: map[string]configs.LinuxRdma{
+			rdmaDevice: {
+				HcaHandles: &maxHandles,
+				HcaObjects: &maxObjects,
+			},
+		},
+	}
+
+	if err := RdmaSet(testCgroupPath, rdmaStubResource); err != nil {
+		t.Fatal(err)
+	}
+
+	// The default rdma.max must be written.
+	rdmaEntries, err := readRdmaEntries(testCgroupPath, "rdma.max")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(rdmaEntries) != 1 {
+		t.Fatal("rdma_test: Got the wrong values while parsing entries from rdma.max")
+	}
+	if rdmaEntries[0].HcaHandles != maxHandles {
+		t.Fatalf("rdma_test: Got the wrong value for hca_handles")
+	}
+	if rdmaEntries[0].HcaObjects != maxObjects {
+		t.Fatalf("rdma_test: Got the wrong value for hca_Objects")
+	}
+}
--- a/libcontainer/cgroups/stats.go
+++ b/libcontainer/cgroups/stats.go
@@ -146,6 +146,17 @@ type HugetlbStats struct {
 	Failcnt uint64 `json:"failcnt"`
 }
 
+type RdmaEntry struct {
+	Device     string `json:"device,omitempty"`
+	HcaHandles uint32 `json:"hca_handles,omitempty"`
+	HcaObjects uint32 `json:"hca_objects,omitempty"`
+}
+
+type RdmaStats struct {
+	RdmaLimit   []RdmaEntry `json:"rdma_limit,omitempty"`
+	RdmaCurrent []RdmaEntry `json:"rdma_current,omitempty"`
+}
+
 type Stats struct {
 	CpuStats    CpuStats    `json:"cpu_stats,omitempty"`
 	CPUSetStats CPUSetStats `json:"cpuset_stats,omitempty"`
@@ -154,6 +165,7 @@ type Stats struct {
 	BlkioStats  BlkioStats  `json:"blkio_stats,omitempty"`
 	// the map is in the format "size of hugepage: stats of the hugepage"
 	HugetlbStats map[string]HugetlbStats `json:"hugetlb_stats,omitempty"`
+	RdmaStats    RdmaStats               `json:"rdma_stats,omitempty"`
 }
 
 func NewStats() *Stats {
--- a/libcontainer/configs/rdma.go
+++ b/libcontainer/configs/rdma.go
@@ -1,0 +1,9 @@
+package configs
+
+// LinuxRdma for Linux cgroup 'rdma' resource management (Linux 4.11)
+type LinuxRdma struct {
+	// Maximum number of HCA handles that can be opened. Default is "no limit".
+	HcaHandles *uint32 `json:"hca_handles,omitempty"`
+	// Maximum number of HCA objects that can be created. Default is "no limit".
+	HcaObjects *uint32 `json:"hca_objects,omitempty"`
+}
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -607,6 +607,15 @@ func CreateCgroupConfig(opts *CreateOpts, defaultDevs []*devices.Device) (*confi
 					Limit:    l.Limit,
 				})
 			}
+			if len(r.Rdma) > 0 {
+				c.Resources.Rdma = make(map[string]configs.LinuxRdma, len(r.Rdma))
+				for k, v := range r.Rdma {
+					c.Resources.Rdma[k] = configs.LinuxRdma{
+						HcaHandles: v.HcaHandles,
+						HcaObjects: v.HcaObjects,
+					}
+				}
+			}
 			if r.Network != nil {
 				if r.Network.ClassID != nil {
 					c.Resources.NetClsClassid = *r.Network.ClassID
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -31,6 +31,7 @@ var (
 		&NetPrioGroup{},
 		&PerfEventGroup{},
 		&FreezerGroup{},
+		&RdmaGroup{},
 		&NameGroup{GroupName: "name=systemd", Join: true},
 	}
 	HugePageSizes, _ = cgroups.GetHugePageSize()
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -58,6 +58,7 @@ var legacySubsystems = []subsystem{
 	&fs.NetPrioGroup{},
 	&fs.NetClsGroup{},
 	&fs.NameGroup{GroupName: "name=systemd"},
+	&fs.RdmaGroup{},
 }
 
 func genV1ResourcesProperties(r *configs.Resources, cm *dbusConnManager) ([]systemdDbus.Property, error) {
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -117,6 +117,9 @@ type Resources struct {
 	// Set class identifier for container's network packets
 	NetClsClassid uint32 `json:"net_cls_classid_u"`
 
+	// Rdma resource restriction configuration
+	Rdma map[string]LinuxRdma `json:"rdma"`
+
 	// Used on cgroups v2:
 
 	// CpuWeight sets a proportional bandwidth limit.


libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	return "rdma"
+}
+
+func (s *RdmaGroup) Apply(path string, d *cgroupData) error {
+	return join(path, d.pid)
+}
+
+func createCmdString(device string, limits configs.LinuxRdma) string {
+	var cmdString string
+	cmdString = device
```
kolyshkin, line 27: why not
```suggestion
	cmdString := device
```

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	if limits.HcaHandles != nil {
+		cmdString = cmdString + " " + "hca_handle=" + strconv.FormatUint(uint64(*limits.HcaHandles), 10)
```
kolyshkin, line 29:
```suggestion
cmdString += " hca_handle=" + strconv.FormatUint(uint64(*limits.HcaHandles), 10)
```

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if limits.HcaObjects != nil {
+		cmdString = cmdString + " " + "hca_object=" + strconv.FormatUint(uint64(*limits.HcaObjects), 10)
```
kolyshkin, line 32: same as above

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (s *RdmaGroup) GetStats(path string, stats *cgroups.Stats) error {
+	if !cgroups.PathExists(path) {
+		return nil
+	}
+	currentData, err := fscommon.GetCgroupParamString(path, "rdma.current")
+	if err != nil {
+		return err
+	}
+	currentPerDevices := strings.Split(string(currentData), "\n")
```
kolyshkin, line 54: 1. No need to convert `currentData` to `string` type -- it is already a string.
2. You'd better use `fscommon.ReadFile` here -- `GetCgroupParamString` is tailored more for single-value files.

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	maxData, err := fscommon.GetCgroupParamString(path, "rdma.max")
```
kolyshkin, line 55: ditto

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+	maxPerDevices := strings.Split(string(maxData), "\n")
```
kolyshkin, line 59: ditto

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	stats.RdmaStats = cgroups.RdmaStats{
+		RdmaLimit:   maxEntries,
+		RdmaCurrent: currentEntries,
+	}
+
+	return nil
+}
+
+// Parse raw string to RdmaEntry
+func ParseRdmaKV(raw string, entry *cgroups.RdmaEntry) {
```
kolyshkin, line 77: Why this function is public?

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+
+	stats.RdmaStats = cgroups.RdmaStats{
+		RdmaLimit:   maxEntries,
+		RdmaCurrent: currentEntries,
+	}
+
+	return nil
+}
+
+// Parse raw string to RdmaEntry
```
kolyshkin, line 76: In Go, a comment should start with a function name, i.e. something like
```go
// ParseRdmaKV parses a raw string to RdmaEntry.
func ParseRdmaKV ....
```

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+			entry.HcaObjects = uint32(value)
+		}
+	}
+}
+
+// Converts array of rawstrings to RdmaEntries
+func ConvertRdmaEntry(strEntries []string) []cgroups.RdmaEntry {
+	var rdmaEntries []cgroups.RdmaEntry
+	for i := range strEntries {
+		parts := strings.Fields(strEntries[i])
```
kolyshkin, line 104: `strings.Fields` is 2x slower than `strings.Split`. Any justification for using it here?

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return nil
+}
+
+// Parse raw string to RdmaEntry
+func ParseRdmaKV(raw string, entry *cgroups.RdmaEntry) {
+	var value uint64
+	var err error
+
+	parts := strings.Split(raw, "=")
```
kolyshkin, line 81: As we're not interested in more than 2 parts, I'd do
```suggestion
	parts := strings.SplitN(raw, "=",3)
```

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	switch len(parts) {
+	case 2:
```
kolyshkin, line 83: I'd do
```go
if len(parts) != 2 {
   continue
}
```

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+			entry.HcaHandles = uint32(value)
+		} else if parts[0] == "hca_object" {
+			entry.HcaObjects = uint32(value)
+		}
+	}
+}
+
+// Converts array of rawstrings to RdmaEntries
+func ConvertRdmaEntry(strEntries []string) []cgroups.RdmaEntry {
+	var rdmaEntries []cgroups.RdmaEntry
```
kolyshkin, line 102: If the expected number of elements in `rdmaEntries` equals to number of elements in `strEntries`, you'd better preallocate those.

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+	for i := range strEntries {
+		parts := strings.Fields(strEntries[i])
+		switch len(parts) {
+		case 3:
```
kolyshkin, line 106: Same here -- I'd rather use
```go
if len(parts) != 3 {
   continue
}
```
as this would improve readability.

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+		if parts[0] == "hca_handle" {
+			entry.HcaHandles = uint32(value)
+		} else if parts[0] == "hca_object" {
+			entry.HcaObjects = uint32(value)
+		}
+	}
+}
+
+// Converts array of rawstrings to RdmaEntries
+func ConvertRdmaEntry(strEntries []string) []cgroups.RdmaEntry {
```
kolyshkin, line 101: Why this function is public?

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		if parts[0] == "hca_handle" {
+			entry.HcaHandles = uint32(value)
+		} else if parts[0] == "hca_object" {
+			entry.HcaObjects = uint32(value)
+		}
+	}
+}
+
+// Converts array of rawstrings to RdmaEntries
```
kolyshkin, line 100: same as https://github.com/opencontainers/runc/pull/2883/commits/c357a29009e21028486f20bbec2fdeb098733141#r605967909

libcontainer/cgroups/fs2/rdma.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func isRdmaSet(cgroup *configs.Cgroup) bool {
+	return len(cgroup.Resources.Rdma) > 0
+}
+
+func createCmdString(device string, limits configs.LinuxRdma) string {
+	var cmdString string
+
+	cmdString = device
```
kolyshkin, line 21: same as above:
```suggestion
	cmdString := device
```

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,118 @@
...
+
+// Parse raw string to RdmaEntry
+func ParseRdmaKV(raw string, entry *cgroups.RdmaEntry) {
+	var value uint64
+	var err error
+
+	parts := strings.Split(raw, "=")
+	switch len(parts) {
+	case 2:
+		if parts[1] == "max" {
```
kolyshkin, line 84: Does cgroup v1 ever reports `"max"`? I thought only cgroup v2 does that

kolyshkin:
Looks good in general, left some comments. 

We also need some integration tests for that.

libcontainer/cgroups/fs/rdma.go
flouthoc: Fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: Fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: Fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: Fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: @kolyshkin Following function is intentionally public as being used by rdma controller in `fs2`. I didnt wanted to have redundant code in both implementations of `fs` and `fs2` and these functions are exclusive to Rdma so i thought i'd rather keep them with rdma and not move to `fscommon`

libcontainer/cgroups/fs/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: @kolyshkin `string.Fields` handles whitespaces very well with appropriate trimming where necessary. With `strings.Split` we would still handle those cases for safety. So i thought since input is a very small constant `Fields` would rather keep code clean without taking any significant hit on performance. But sure if you think `strings.Split` makes more sense i'd rather switch to that. Not a big issue.

libcontainer/cgroups/fs/rdma.go
flouthoc: Fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs2/rdma.go
flouthoc: fixed.

libcontainer/cgroups/fs/rdma.go
flouthoc: @kolyshkin It does as per man pages afaik, let me know if you think i misunderstood something. https://www.kernel.org/doc/Documentation/cgroup-v1/rdma.txt

libcontainer/cgroups/fs/rdma.go
flouthoc: @kolyshkin Following function is intentionally public as being used by rdma controller in fs2. I didnt wanted to have redundant code in both implementations of fs and fs2 and these functions are exclusive to Rdma so i thought i'd rather keep them with rdma and not move to fscommon

flouthoc:
@kolyshkin Thanks for the quick review, i have fixed almost everything. 3 comments are still unresolved but i have commented the reason on the review itself. Could you please take a look.

libcontainer/cgroups/fs/rdma.go
flouthoc: resolved.

thaJeztah:
I see there's a typo in the commit message ("kernal" instead of "kernel"); perhaps good to fix that in case we want to find it back later if someone searches for `kernel` related commits

flouthoc:
@thaJeztah Thanks for pointing this out. Fixed commit message `kernal` -> `kernel` :smiley:

AkihiroSuda:
Please sign the commit with real name

flouthoc:
@AkihiroSuda Done.

libcontainer/cgroups/fs/rdma.go
kolyshkin: If the input is coming from the kernel (rather than e.g. a user-written config file), it does not contain any extra spaces (and if it does, something is really wrong). So I'd use strings.Split here.

flouthoc: @kolyshkin Sure makes sense. I have upated this and converted to `strings.SplitN`. This is resolved.

flouthoc:
@kolyshkin Fixed requested changes. Let me know if anything else needs to be done.

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,115 @@
+// +build linux
+
+package fs
+
+import (
+	"github.com/opencontainers/runc/libcontainer/cgroups"
```
fuweid, line 6: Should be aligned with other files, like 

```
import (
        "math"
	"strconv"
	"strings"

        "github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
)
```

libcontainer/cgroups/fs2/rdma.go
```diff
@@ -0,0 +1,69 @@
+// +build linux
+
+package fs2
+
+import (
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fs"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"strconv"
```
fuweid, line 10: ditto

libcontainer/cgroups/fs/rdma.go
flouthoc: @fuweid thanks, this is fixed.

libcontainer/cgroups/fs2/rdma.go
flouthoc: @fuweid fixed :smile:

libcontainer/cgroups/fs/rdma.go
fuweid: fscommon can hold the helper for RDMA. fs2 imports fscommon better than fs package. just my two cents.

libcontainer/configs/rdma.go
```diff
@@ -0,0 +1,9 @@
+package configs
+
+// LinuxRdma for Linux cgroup 'rdma' resource management (Linux 4.11)
+type LinuxRdma struct {
+	// Maximum number of HCA handles that can be opened. Default is "no limit".
+	HcaHandles *uint32 `json:"hcaHandles,omitempty"`
```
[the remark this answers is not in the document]
flouthoc: @fuweid Sure looks cool to me. Did the change :smile:

libcontainer/cgroups/fs/rdma.go
flouthoc: @fuweid Oh so idea was since `ParseRdmaKV` and `ConvertRdmaEntry` uses types from package `cgroups` having them in `fscommon` will introduce `import cycles`. Since both of these functions are only going to be used only by `rdma` therefore i thought its better to keep them with `rdma` implementation only. But let me know if you guys think otherwise we can figure out a way. :smile:

flouthoc:
@kolyshkin @fuweid Were you guys able to take look at reviews, are they good to be marked as resolved ? I think we have to anyways park this PR till `rc94`

libcontainer/cgroups/fs/rdma.go
fuweid: I prefer to keep the duplicate code in separate packages (non-binding). :)

fuweid:
@flouthoc the cgroup v1 uses pseudo cgroup fs to test the function. I think it should be easier to add the unit test in this pr.

flouthoc:
@fuweid Sure thanks a lot.

kolyshkin:
I took a fresh look, at it seems that

1. If you can move common rmda code into `libcontainer/cgroups` (instead of putting it to `libcontainer/cgroups/fs` and re-using from `fs2`), it will look better. I

2. It might make more sense to use bufio's `Scan` rather than reading/returning the whole file and doing `strings.Split(contents, "\n")` on the result.

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -141,6 +141,12 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 			errs = append(errs, err)
 		}
 	}
+	// rdma (since kernel 4.11)
+	if _, ok := m.controllers["rdma"]; ok {
```
kolyshkin, line 145: Since commit 0f8d2b6b we no longer check if the controller is available.

flouthoc: Sure i'll do the required changes.

flouthoc:
@kolyshkin Sure having common chunk in `libcontainer/cgroups` makes sense i'll do the required changes.

libcontainer/cgroups/fs/rdma.go
kolyshkin: I think this should go to `fscommon`, which naturally contains the stuff that's common between fs and fs2.

kolyshkin: Which is a problem, as described above (https://github.com/opencontainers/runc/pull/2883/commits/02eb7236c86c4b700c7465cf4779ae12bf1a19a3#r610598989).

To solve it, I think we should move OpenFile/ReadFile etc from fscommon to `cgroups/utils`.

libcontainer/cgroups/stats.go
```diff
@@ -146,6 +146,17 @@ type HugetlbStats struct {
 	Failcnt uint64 `json:"failcnt"`
 }
 
+type RdmaEntry struct {
+	Device     string `json:"device,omitempty"`
+	HcaHandles uint32 `json:"hca_handles,omitempty"`
```
thaJeztah, line 151: Should these all be `RDMAEntry` and `HCAHandles` (capital RDMA and HCA)?

libcontainer/cgroups/fs/rdma.go
kolyshkin: https://github.com/opencontainers/runc/pull/3024

libcontainer/cgroups/stats.go
kolyshkin: We have a linter for that now, don't we? Let me try to run it.

kolyshkin:
Close/reopen to use latest linters in CI.

libcontainer/cgroups/stats.go
thaJeztah: Not sure it the linter understands all acronyms of course.

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,116 @@
...
+// Following is intentionally public as it going to be used by Rdma controller in fs2
+// ParseRdmaKV: parse raw string to RdmaEntry.
+func ParseRdmaKV(raw string, entry *cgroups.RdmaEntry) {
+	var value uint64
+	var err error
+
+	parts := strings.SplitN(raw, "=", 3)
+	if len(parts) == 2 {
+		if parts[1] == "max" {
+			value = math.MaxUint32
```
kolyshkin, line 85: `value` is declared as `uint64` but here you use `MaxUint32`. Why not `MaxUint64` (or declare `value` as `uint32`)?

libcontainer/cgroups/stats.go
kolyshkin: Right. I agree those should be named as in https://github.com/containerd/console/pull/54

libcontainer/cgroups/fs/rdma.go
flouthoc: @kolyshkin oh makes sense i'll re-visit this entire PR once this gets merged as well https://github.com/opencontainers/runc/pull/3024 WDYT ?

kolyshkin:
@flouthoc #3024 is now merged, so you can put the common RDMA code into fscommon (or its own package if you want).

flouthoc:
@kolyshkin Sure thanks for notifying. I am on it.

libcontainer/cgroups/fs/rdma.go
flouthoc: @kolyshkin So i remember reason for this was `strconv.ParseUInt` always returns uint64 event when `bitsize` is capped to `32`. So i had to cast back using `uint32`.

flouthoc:
@kolyshkin @thaJeztah @fuweid Could you guys please trigger workflow on this.

libcontainer/cgroups/fs2/fs2.go
kolyshkin: This ^^^ is still a valid concern.

libcontainer/cgroups/fscommon/utils.go
```diff
@@ -129,3 +129,43 @@ func GetCgroupParamString(path, file string) (string, error) {
...
+
+// ParseRdmaKV: parse raw string to RdmaEntry.
+func parseRdmaKV(raw string, entry *cgroups.RdmaEntry) {
+	var value uint64
+	var err error
+
+	parts := strings.SplitN(raw, "=", 3)
+	if len(parts) == 2 {
+		if parts[1] == "max" {
+			value = math.MaxUint32
```
kolyshkin, line 141: `MaxUint64`?

libcontainer/cgroups/fscommon/utils.go
```diff
@@ -129,3 +129,43 @@ func GetCgroupParamString(path, file string) (string, error) {
 
 	return strings.TrimSpace(contents), nil
 }
+
+// ParseRdmaKV: parse raw string to RdmaEntry.
```
kolyshkin, line 133: `s/ParseRdmaKV: parse /parseRdmaKV parses /`

libcontainer/cgroups/fscommon/utils.go
```diff
@@ -129,3 +129,43 @@ func GetCgroupParamString(path, file string) (string, error) {
...
+func parseRdmaKV(raw string, entry *cgroups.RdmaEntry) {
+	var value uint64
+	var err error
+
+	parts := strings.SplitN(raw, "=", 3)
+	if len(parts) == 2 {
```
kolyshkin, line 139:
```go
if len(parts) != 2 {
    continue
```

libcontainer/cgroups/stats.go
kolyshkin: This ^^^  concern is still valid.

libcontainer/cgroups/fs2/rdma.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	currentData, err := cgroups.ReadFile(path, "rdma.current")
+	if err != nil {
+		return err
+	}
+	currentPerDevices := strings.Split(currentData, "\n")
+	maxData, err := cgroups.ReadFile(path, "rdma.max")
+	if err != nil {
+		return err
+	}
```
kolyshkin, line 53: This string manipulation is probably not very efficient (also in terms of garbage collection).

It's probably easier and better to offload the reading/parsing to `ConvertRDMAEntry` (which I'd rename to `ReadRDMAEntries`).

Something like this
```go
currentEntries, err := fscommon.ReadRDMAEntries(path, "rdma.current")
...
```

and then `ReadRDMAEntries` should look like

```go
func ReadRDMAEntries(dir, file string) ([]cgroups.RdmaEntry, error) {
   fd, err := OpenFile(dir, file)
   if err != nil {
      return nil, err
   }
   defer fd.Close() //nolint:errorlint
   s := bufio.NewScanner(fd)
   for s.Scan() {
   }
   return entries, s.Err()
}
```

This will decrease the amount of temporary objects, unify the processing between v1 and v2 more.

kolyshkin (changes requested):
Left a few comments.

Can you please use `rdma.go` file name for the common RDMA code, rather than `uitls.go`?

libcontainer/cgroups/stats.go
flouthoc: @thaJeztah @kolyshkin Following convention is analogous to upstream `opencontainers/runtime-spec` https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L347 . My Intention is to keep everything consistent across the ecosystem but i am cool with anything

libcontainer/cgroups/fs/rdma.go
flouthoc: Resolved in latest commit.

libcontainer/cgroups/fs2/fs2.go
flouthoc: Resolved in latest commit.

libcontainer/cgroups/fs/rdma.go
flouthoc: Resolved.

libcontainer/cgroups/fscommon/utils.go
flouthoc: Resolved.

libcontainer/cgroups/fs2/rdma.go
flouthoc: @kolyshkin Thanks for this suggestion , added `ReadRDMAEntries` in latest commit.

libcontainer/cgroups/fscommon/utils.go
flouthoc: Resolved.

libcontainer/cgroups/fscommon/utils.go
flouthoc: Following is handled in calling loop by throwing custom error.

flouthoc:
@kolyshkin @thaJeztah  Added all the suggested changes. Just had doubt on feedback raised by @thaJeztah about naming scheme i have written reason for that.

flouthoc:
@kolyshkin Could we re-trigger tests.

kolyshkin (changes requested):
I took a fresh look, and it seems that the code can be unified more between v1 and v2.

I'd like to see something similar to what you did in fs2, but with the fs2/rdma.go merged into fscommon/rdma.go (and renamed `Rdma*`), so that cgroups/fs will look like a mere wrapper:
```go
func (s *RdmaGroup) Set(path string, r *configs.Resources) error {
   return fscommon.RdmaSet(path, r)
}

func (s *RdmaGroup) GetStats(path string, stats *cgroups.Stats) error {
   return fscommon.RdmaGetStats(path, stats)
}
```

libcontainer/cgroups/fs2/rdma.go
```diff
@@ -0,0 +1,64 @@
...
+	if limits.HcaObjects != nil {
+		cmdString += " hca_object=" + strconv.FormatUint(uint64(*limits.HcaObjects), 10)
+	}
+	return cmdString
+}
+
+func setRdma(path string, r *configs.Resources) error {
+	if !isRdmaSet(r) {
+		return nil
+	}
```
kolyshkin, line 31: There's no need to have this, as the following `for` loop will have 0 iterations when `len(r.Rdma) == 0` anyway.

flouthoc: @kolyshkin Sure let me add relevant changes.

flouthoc:
@kolyshkin Applied all the requested changes, could you please take a look again.

libcontainer/cgroups/fs2/rdma.go
flouthoc: @kolyshkin added this change in latest commit.

flouthoc:
@kolyshkin Not sure why tests are failing but CI doesn't really states any error seems like they flaked but not sure.

kolyshkin:
CI failures were expected (fixed in #3051).

CI restarted.

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	} else {
+		return errors.New("Unable to parse RDMA entry")
+	}
+
+	return nil
+}
+
+// ReadRDMAEntries: Reads and converts array of rawstrings to RdmaEntries from file
+func ReadRDMAEntries(dir, file string) ([]cgroups.RdmaEntry, error) {
```
kolyshkin, line 46: This doesn't need to be public anymore.

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+	if limits.HcaHandles != nil {
+		cmdString += " hca_handle=" + strconv.FormatUint(uint64(*limits.HcaHandles), 10)
+	}
+	if limits.HcaObjects != nil {
+		cmdString += " hca_object=" + strconv.FormatUint(uint64(*limits.HcaObjects), 10)
+	}
+	return cmdString
+}
+
+// RdmaSet: sets rdma subsystem with configured values.
```
kolyshkin, line 113: Nit:
```suggestion
// RdmaSet sets RDMA resources.
```

libcontainer/specconv/spec_linux.go
```diff
@@ -616,6 +616,15 @@ func CreateCgroupConfig(opts *CreateOpts, defaultDevs []*devices.Device) (*confi
 					Limit:    l.Limit,
 				})
 			}
+			if len(r.Rdma) > 0 {
```
kolyshkin, line 619: It looks like the check for `len` here is redundant -- the `for` below will be a no-op when `len == 0`.

kolyshkin: With the current code it is, but the correct code would be

```go
if len(r.Rdma) > 0 {
    c.Resources.Rdma = make(map[string]configs.LinuxRdma, len(r.Rdma))
    for k, v := range r.Rdma {
        c.Resources.Rdma[k] = configs.LinuxRdma{
            ......
```

libcontainer/specconv/spec_linux.go
```diff
@@ -616,6 +616,15 @@ func CreateCgroupConfig(opts *CreateOpts, defaultDevs []*devices.Device) (*confi
...
+				for k, v := range r.Rdma {
+					c.Resources.Rdma = make(map[string]configs.LinuxRdma, len(r.Rdma))
```
kolyshkin, line 621: This is plain wrong -- you recreate a whole map for every iteration. This should be outside of the loop.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: Resolved.

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+				continue
+			}
+
+			rdmaEntries = append(rdmaEntries, *entry)
+		}
+	}
+	return rdmaEntries, scanner.Err()
+}
+
+// RdmaGetStats: Returns rdma stats such as totalLimit and current entries
```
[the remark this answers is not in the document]
flouthoc: Resolved.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: Resolved.

libcontainer/specconv/spec_linux.go
flouthoc: Resolved.

libcontainer/specconv/spec_linux.go
flouthoc: resolved.

flouthoc:
@kolyshkin Added fixes.

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+	rdmaEntries := make([]cgroups.RdmaEntry, 0)
+	fd, err := cgroups.OpenFile(dir, file, unix.O_RDONLY)
+	if err != nil {
+		return nil, err
+	}
+	defer fd.Close() //nolint:errorlint
+	scanner := bufio.NewScanner(fd)
+	for scanner.Scan() {
+		strEntry := scanner.Text()
+		parts := strings.SplitN(strEntry, " ", 4)
```
fuweid, line 56: SplitN 3 here?

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+			if err != nil {
+				return err
+			}
+		}
+		if parts[0] == "hca_handle" {
+			entry.HcaHandles = uint32(value)
+		} else if parts[0] == "hca_object" {
+			entry.HcaObjects = uint32(value)
+		}
+	} else {
```
fuweid, line 38: maybe use short-circuiting here?

```
if len(parts) != 2 {
    return errors.New("Unable to parse RDMA entry")
}

k, v := parts[0], parts[1]
...
}
```

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"golang.org/x/sys/unix"
+)
+
+// parseRdmaKV: parses raw string to RdmaEntry.
+func parseRdmaKV(raw string, entry *cgroups.RdmaEntry) error {
+	var value uint64
+	var err error
+
+	parts := strings.SplitN(raw, "=", 3)
```
fuweid, line 22: SplitN 2 here?

thaJeztah: Better to use Splitn 3, so that the code is able to catch len(parts) > 2

libcontainer/cgroups/fscommon/rdma.go
thaJeztah: See above; doing so wouldn't allow detecting len(parts) > 3

thaJeztah (changes requested):

libcontainer/cgroups/fscommon/rdma.go
fuweid: Oh, Good. Thanks!

flouthoc: @fuweid @thaJeztah Can i mark this as resolved ? As SplitN 3 is needed here.

fuweid: Yes, please. :)

libcontainer/cgroups/fscommon/rdma.go
flouthoc: @fuweid This is needed resolving this,

libcontainer/cgroups/fscommon/rdma.go
flouthoc: @fuweid Resolved in latest commit.

flouthoc:
@fuweid @thaJeztah @kolyshkin Resolved feedback points in latest commit.

libcontainer/cgroups/fscommon/rdma.go
fuweid: Do you push the commit for this?

flouthoc: @fuweid oh i forgot `git add` on the file. Just added now.

flouthoc:
@fuweid @kolyshkin could you guys PTAL.

fuweid (commented):
lgtm(nb)

libcontainer/cgroups/fs2/rdma.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+func setRdma(path string, r *configs.Resources) error {
+	return fscommon.RdmaSet(path, r)
+}
+
+func statRdma(path string, stats *cgroups.Stats) error {
+	return fscommon.RdmaGetStats(path, stats)
+}
```
kolyshkin, line 17: It seems this whole file is useless -- you can just call `fscommon.Rdma*` functions directly in `fs2/fs2.go`

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"golang.org/x/sys/unix"
+)
+
+// parseRdmaKV: parses raw string to RdmaEntry.
+func parseRdmaKV(raw string, entry *cgroups.RdmaEntry) error {
+	var value uint64
```
kolyshkin, line 19: As you are casting the value to uint32 below, maybe it makes sense to have it as uint32.

```go
	var value uint32
...
	if v == "max" {
		value = math.MaxUint32
	} else {
		val64, err := strconv.ParseUint(v, 10, 32)
		if err != nil {
			return err
		}
		value  = uint32(val64)
	}
```

kolyshkin:
Looks good except for couple of nits.

What worries me though is the absence of tests. How can we be sure this thing works (except for doing a review)?

I think there should be `fscommon/rdma_test.go` which does something similar to what fs and fs2 tests are doing. Can be added as a separate commit.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: @kolyshkin Resolved thanks for spotting this.

libcontainer/cgroups/fs2/rdma.go
flouthoc: @kolyshkin Resolved thanks for spotting this.

flouthoc:
@kolyshkin @fuweid  Resolved all feedback points. Also added tests `fscommon/rdma_test.go` which tries writing rdma entry into a mock cgroup filesystem and verifies it. Also wrote future roadmap of tests which can be figured out later in a passive manner i.e how to verify test against an actual RdmaDevice or mimic one adding actual device to CI might be too much.

flouthoc:
@kolyshkin @fuweid PTAL.

fuweid:
@flouthoc would you please fix the linter issue? Thanks!

flouthoc:
@fuweid Could you run checks again also would be helpful if you could make these checks trigger on every push. I think no manual intervention should be needed to run checks even for first time contributor also .

flouthoc:
@fuweid @kolyshkin lint errors are fixed and resolved in latest commit. :)

flouthoc:
@kolyshkin @fuweid please let me know if anything else needs to be done.

kolyshkin:
@flouthoc can you please rebase? We have tweaked CI again...

Other than that I think this one is ready to go in.

libcontainer/cgroups/fscommon/utils_test.go
```diff
@@ -101,3 +107,53 @@ func TestGetCgroupParamsInt(t *testing.T) {
...
+
+// Write the specified contents on the mock of the specified cgroup files.
+func (c *cgroupTestUtil) writeFileContents(fileContents map[string]string) {
+	for file, contents := range fileContents {
+		err := cgroups.WriteFile(c.CgroupPath, file, contents)
+		if err != nil {
+			c.t.Fatal(err)
+		}
+	}
+}
```
kolyshkin, line 159: OK I haven't realized you were copying all this from `fs` package. This is not good. For one thing, the implementation in `fs` has already improved. That does not mean that you need to copy those improvements -- this is not going to work in the long term.

I think since both fs and fscommon tests need these helpers, they need to be moved to cgroups package (somewhat similar to what I did in #3024).

kolyshkin (changes requested):

libcontainer/cgroups/fscommon/utils_test.go
flouthoc: @kolyshkin Sure i can move this to `cgroup` I'll make relevant changes and re-base

flouthoc: @kolyshkin Should i create a separate PR for refactor of TestUtils or seperate commit in same PR cause it contains changes which are out-of-scope wrt to this pull request.

flouthoc: @kolyshkin i think separate PR would be better for discussions https://github.com/opencontainers/runc/pull/3122 since too many changes are there.

kolyshkin: OK, this was a very bad idea of mine.

I took a look at the test you are adding -- in practice, it seems that having something like

```go
root := t.TempDir()
rdmaDir := filepath.Join(root, "rdma")
 if err := os.MkdirAll(testCgroupPath, 0o755); err != nil {
                t.Fatal(err)
        }
```

should be enough for the test case you're adding.

flouthoc: @kolyshkin Should i drop changes  in `libcontainer/cgroups/fscommon/utils_test.go` and have this minimal block in `libcontainer/cgroups/fscommon/rdma_test.go` itself only, WDYT ? however this might be inconsistent with how tests for other subsystems are written.

flouthoc: @kolyshkin Done made changes and re-based with upstream please take a look again.

flouthoc: @kolyshkin dropped changes for `libcontainer/cgroups/fscommon/utils_test.go` everything stays with RDMA tests itself. PTAL.

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,83 @@
...
+func TestRdmaSet(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "cgroup_test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	root := tempDir
+	testCgroupPath := filepath.Join(root, "rdma")
+	if err != nil {
+		t.Fatal(err)
+	}
```
kolyshkin, line 29: what error are you trying to check here?

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+/* Roadmap for future */
+// (Low-priority)  TODO: Check if it is possible to virtually mimic an actual RDMA device.
+// TODO: Think of more edge-cases to add.
+
+// TestRdmaSet performs an E2E test of RdmaSet(), parseRdmaKV() using dummy device and a dummy cgroup file-system.
+// Note: Following test does not gurantees that your host supports RDMA since this mocks underlying infrastructure.
+func TestRdmaSet(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "cgroup_test")
```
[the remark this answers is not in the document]
kolyshkin: See the example I provided earlier in https://github.com/opencontainers/runc/pull/2883#discussion_r680143156

kolyshkin (changes requested):
some nits here and there

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,123 @@
...
+	rdmaEntries := make([]cgroups.RdmaEntry, 0)
+	fd, err := cgroups.OpenFile(dir, file, unix.O_RDONLY)
+	if err != nil {
+		return nil, err
+	}
+	defer fd.Close() //nolint:errorlint
+	scanner := bufio.NewScanner(fd)
+	for scanner.Scan() {
+		strEntry := scanner.Text()
+		parts := strings.SplitN(strEntry, " ", 4)
```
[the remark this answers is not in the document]
flouthoc: @kolyshkin Thanks, resolved in latest commit.

libcontainer/cgroups/fscommon/rdma_test.go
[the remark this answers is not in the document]
flouthoc: resolved in latest commit.

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	err = parseRdmaKV(z[1], &rdmaEntry)
+	if err != nil {
+		t.Fatalf("Failed while parsing RDMA entry: %s", err)
+	}
+	err = parseRdmaKV(z[2], &rdmaEntry)
+	if err != nil {
+		t.Fatalf("Failed while parsing RDMA entry: %s", err)
+	}
```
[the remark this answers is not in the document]
flouthoc: @kolyshkin my bad here, switched to `rdmaReadEntries` in latest commit. Thanks for spotting this.

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,83 @@
...
+	// Ensure the full mock cgroup path exists.
+	err = os.MkdirAll(testCgroupPath, 0o755)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = cgroups.WriteFile(testCgroupPath, "rdma.max", "")
+	if err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
flouthoc: @kolyshkin RdmaSet will automatically create this file, this is not needed.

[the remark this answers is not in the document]
flouthoc: Resolved in latest commit.

libcontainer/cgroups/fscommon/rdma_test.go
flouthoc: @kolyshkin this is not needed. I don't remember why did i write it. :sob: . Resolved in latest commit.

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	rdmaDevice := "mlx5_1"
+	maxHandles := uint32(100)
+	maxObjects := uint32(300)
+
+	rdmaStubResource := &configs.Resources{}
+	rdmaStubResource.Rdma = map[string]configs.LinuxRdma{rdmaDevice: {HcaHandles: &maxHandles, HcaObjects: &maxObjects}}
```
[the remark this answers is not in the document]
flouthoc: @kolyshkin fixed in latest commit.

flouthoc:
@kolyshkin Could you please give a quick eye again. Resolved all the nits :sweat_smile:

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,122 @@
...
+	defer fd.Close() //nolint:errorlint
+	scanner := bufio.NewScanner(fd)
+	for scanner.Scan() {
+		parts := strings.SplitN(scanner.Text(), " ", 4)
+		if len(parts) == 3 {
+			entry := new(cgroups.RdmaEntry)
+			entry.Device = parts[0]
+			err = parseRdmaKV(parts[1], entry)
+			if err != nil {
+				continue
```
kolyshkin, line 63: @flouthoc can you remind me why we chose to continue rather than return an error here?

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+			err = parseRdmaKV(parts[2], entry)
+			if err != nil {
+				continue
```
kolyshkin, line 67: ditto

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+/* Roadmap for future */
+// (Low-priority)  TODO: Check if it is possible to virtually mimic an actual RDMA device.
+// TODO: Think of more edge-cases to add.
+
+// TestRdmaSet performs an E2E test of RdmaSet(), parseRdmaKV() using dummy device and a dummy cgroup file-system.
+// Note: Following test does not gurantees that your host supports RDMA since this mocks underlying infrastructure.
+func TestRdmaSet(t *testing.T) {
+	tempDir := t.TempDir()
+	root := tempDir
```
kolyshkin, line 19: So, why not
```go
root := t.TempDir()
```
?

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,59 @@
...
+	testCgroupPath := filepath.Join(root, "rdma")
+
+	// Ensure the full mock cgroup path exists.
+	err := os.MkdirAll(testCgroupPath, 0o755)
```
kolyshkin, line 23: as `testCgroupPath` is a direct subdirectory of `root`, MkdirAll is not needed, it's plain old `Mkdir`.

libcontainer/cgroups/fscommon/rdma_test.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rdmaDevice := "mlx5_1"
+	maxHandles := uint32(100)
+	maxObjects := uint32(300)
```
kolyshkin, line 30: maybe `const`?

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,122 @@
...
+
+// RdmaGetStats: Returns rdma stats such as totalLimit and current entries.
+func RdmaGetStats(path string, stats *cgroups.Stats) error {
+	if !cgroups.PathExists(path) {
+		return nil
+	}
+	currentEntries, err := readRdmaEntries(path, "rdma.current")
+	if err != nil {
+		return err
+	}
```
kolyshkin, line 84: It seems that in here we should return all errors except for os.IsNotExist. So, something like

```go
cur, err := readRdmaEntries(path, "rdma.current")
if err != nil && !errors.Is(err, os.IsNotExist) {
    return err
}
```

Now, with this in place, you don't have to check that the path exists beforehand, so you can drop the `if !cgroups.PathExists(path)`.

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// readRdmaEntries: Reads and converts array of rawstrings to RdmaEntries from file.
+func readRdmaEntries(dir, file string) ([]cgroups.RdmaEntry, error) {
+	rdmaEntries := make([]cgroups.RdmaEntry, 0)
+	fd, err := cgroups.OpenFile(dir, file, unix.O_RDONLY)
+	if err != nil {
+		return nil, err
+	}
+	defer fd.Close() //nolint:errorlint
```
kolyshkin, line 54: Would be nice to have a comment showing how the file looks like. It usually helps a lot when reading such parser code.

libcontainer/cgroups/fscommon/rdma.go
kolyshkin: ...and if you choose to not ignore errors from here, please use `parseError` type (see examples of using it elsewhere in fscommon).

flouthoc: @kolyshkin Idea is that we want to parse other entries even if some entries fail which is highly unlikely as writing wrong entries should throw `EINVAL` from subsystem itself. I think we can just drop error checking here. WDYT ?

libcontainer/cgroups/fscommon/rdma_test.go
flouthoc: Struct takes `pointer` to  values from spec with `const` we cant do that.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: @kolyshkin I think this should be 
```go
	currentEntries, err := readRdmaEntries(path, "rdma.current")
	if err != nil && !errors.Is(err, os.ErrNotExist) {
		return err
	}
```
switching `os.IsNotExist` with `os.ErrNotExist` as `os.IsNotExist` is a method which returns `bool`.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: Resolved in latest commit.

libcontainer/cgroups/fscommon/rdma_test.go
flouthoc: this is resolved.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: being discussed in above thread.

flouthoc:
@kolyshkin Resolved all of them just a small doubt in first one i have tried clarifying in review itself.

libcontainer/cgroups/fscommon/rdma.go
kolyshkin: Yes, that's right, I mistyped.

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return rdmaEntries, scanner.Err()
+}
+
+// RdmaGetStats: Returns rdma stats such as totalLimit and current entries.
+func RdmaGetStats(path string, stats *cgroups.Stats) error {
+	currentEntries, err := readRdmaEntries(path, "rdma.current")
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
```
kolyshkin, line 83: OK, this code is trying to parse empty entries.

What I meant and should have suggested earlier is to return `nil` if the file is not found.
```go
currentEntries, err := readRdmaEntries(path, "rdma.current")
if err != nil {
	if errors.Is(err, os.ErrNotExist) {
		err = nil
	}
	return err
}
```

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,121 @@
...
+	maxEntries, err := readRdmaEntries(path, "rdma.max")
+	if err != nil {
+		return err
+	}
+	// If device got removed between reading two files, ignore returning
+	// stats.
```
kolyshkin, line 89: Can you fit this in one line please?

kolyshkin (changes requested):
One fix (to the code that I suggested), one nit, and a note about documentation about functions.

You write it as
```go
// FuncName: does this and that.
```

It should be

```go
// FuncName does this and that.
```

libcontainer/cgroups/fscommon/rdma.go
flouthoc: @kolyshkin resolved in latest commit.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: @kolyshkin resolved in latest commit.

flouthoc:
@kolyshkin I have made requested changes could you PTAL.

flouthoc:
@kolyshkin Could you please let me know if anything is pending here from my side.

kolyshkin (dismissed):
LGTM

@thaJeztah @AkihiroSuda PTAL

libcontainer/cgroups/fscommon/rdma.go
```diff
@@ -0,0 +1,123 @@
+// +build linux
```
AkihiroSuda, line 1: We do not need build tags if we use filenames like `rdma_linux.go`

libcontainer/cgroups/fs/rdma.go
```diff
@@ -0,0 +1,27 @@
+// +build linux
```
AkihiroSuda, line 1: We do not need build tags if we use filenames like rdma_linux.go

flouthoc:
@AkihiroSuda thanks :) resolved in latest commit.

libcontainer/cgroups/fscommon/rdma.go
flouthoc: resolved in latest commit.

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #2886, opened 2021-04-02
libcontainer/configs/validate: make Validate() less DRY

thaJeztah:
Noticed this when looking through the changes in https://github.com/opencontainers/runc/pull/2855

(this one will need to be updated if that one is merged first)


cyphar (approved):
LGTM, though I think technically this makes it more DRY not less?

AkihiroSuda (approved):

PR #2894, opened 2021-04-06
libct/init_linux: retry chdir to fix EPERM (regression in rc93)

haircommander:
Alas, the EPERM on chdir saga continues...

Unfortunately, the there were two releases between when https://github.com/opencontainers/runc/commit/5e0e67d76cc99d76c8228d48f38f37034503f315  was released
and when the workaround https://github.com/opencontainers/runc/pull/2712 was added.

Between this, folks started relying on the ability to have a workdir that the container user doesn't have access to.

Since this case was previously valid, we should continue support for it.

Now, we retry the chdir:
Once at the top of the function (to catch cases where the runc user has access, but container user does not)
and once after we setup user (to catch cases where the container user has access, and the runc user does not)

Add a test case for this as well.

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -131,6 +131,26 @@ func finalizeNamespace(config *initConfig) error {
 		return errors.Wrap(err, "close exec fds")
 	}
 
+	// we only do chdir if it's specified
+	doChdir := config.Cwd != ""
+	if doChdir {
+		// First, attempt the chdir before setting up the user.
+		// This could allow us to access a directory that the user running runc can access
+		// but the container user cannot.
+		err := unix.Chdir(config.Cwd)
+		switch {
+		case err == nil:
+			doChdir = false
+		case os.IsPermission(err):
+			// If we hit an EPERM, we should attempt again after setting up user.
+			// This will allow us to successfully chdir if the container user has access
+			// to the directory, but the user running runc does not.
+			// This is useful in cases where the cwd is also a volume that's been chowned to the container user.
+		default:
+			return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
+		}
+	}
+
 	caps := &configs.Capabilities{}
 	if config.Capabilities != nil {
 		caps = config.Capabilities
@@ -152,10 +172,8 @@ func finalizeNamespace(config *initConfig) error {
 	if err := setupUser(config); err != nil {
 		return errors.Wrap(err, "setup user")
 	}
-	// Change working directory AFTER the user has been set up.
-	// Otherwise, if the cwd is also a volume that's been chowned to the container user (and not the user running runc),
-	// this command will EPERM.
-	if config.Cwd != "" {
+	// Change working directory AFTER the user has been set up, if we haven't done it yet.
+	if doChdir {
 		if err := unix.Chdir(config.Cwd); err != nil {
 			return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
 		}
--- a/tests/integration/cwd.bats
+++ b/tests/integration/cwd.bats
@@ -30,7 +30,7 @@ function teardown() {
 
 # Verify a cwd owned by the container user can be chdir'd to,
 # even if runc doesn't have the privilege to do so.
-@test "runc create sets up user before chdir to cwd" {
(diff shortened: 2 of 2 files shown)


libcontainer/init_linux.go
```diff
@@ -131,6 +131,25 @@ func finalizeNamespace(config *initConfig) error {
...
 
+	// we only do chdir if it's specified
+	doChdir := config.Cwd != ""
+	if doChdir {
+		doChdir = false
+		// First, attempt the chdir before setting up the user.
+		// This could allow us to access a directory that the user running runc can access
+		// but the container user cannot.
+		if err := unix.Chdir(config.Cwd); err != nil {
+			if !os.IsPermission(err) {
```
kolyshkin, line 142: Since this is `unix.`, the errors are bare (unwrapped), and it's _probably_ better to compare directly against `unix.EPERM`.

haircommander: for some reason I thought that wasn't working as expected (though I suspect now i pushed to the wrong branch before...)

libcontainer/init_linux.go
```diff
@@ -131,6 +131,25 @@ func finalizeNamespace(config *initConfig) error {
 		return errors.Wrap(err, "close exec fds")
 	}
 
+	// we only do chdir if it's specified
+	doChdir := config.Cwd != ""
+	if doChdir {
+		doChdir = false
```
kolyshkin, line 137: I'd rather set doChdir to false after `unix.Chdir` has succeeded (and modify the comment below saying we ignore EPERM for now because we'll retry below).

Not too important though, it is fine as it is.

kolyshkin (approved):
LGTM

libcontainer/init_linux.go
haircommander: I like the code flow here better than the alternative of :
```
	if doChdir {
		// First, attempt the chdir before setting up the user.
		// This could allow us to access a directory that the user running runc can access
		// but the container user cannot.
		if err := unix.Chdir(config.Cwd); err != nil {
			if err != unix.EPERM {
				return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
			}
			// If we hit an EPERM, we should attempt again after setting up user.
			// This will allow us to successfully chdir if the container user has access
			// to the directory, but the user running runc does not.
			// This is useful in cases where the cwd is also a volume that's been chowned to the container user.
		} else {
			doChdir = false
		}
	}
```
(elses are kinda ugly imo)

haircommander:
I think IsPermission fixes the failures

libcontainer/init_linux.go
kolyshkin: You can do switch/case, looks decent to me:

```go
switch err := unix.Chdir(config.Cwd); err {
case nil: 
    doChdir = false
case EPERM:
    // will retry below
default:
    return fmt.Errorf("chdir to cwd (%q) set in config.json failed: %v", config.Cwd, err)
}
```

haircommander: agreed, that's cleaner. I've adopted it (but can't switch on err because of the failures caused by not using IsPermission)

kolyshkin:
Ah, the `EPERM` check fails because the error returned by chdir is `EACCES`. As per chdir(2) man page:

>        EACCES Search permission is denied for one of the components of path.  (See also path_resolution(7).)

and there's no EPERM in the list of errors returned by chdir.

So we could use `case EACCES:` but I think using os.IsPermission is actually a bit cleaner and more readable.

kolyshkin:
CI on CentOS hit a snag; restarted.

libcontainer/init_linux.go
```diff
@@ -131,6 +131,26 @@ func finalizeNamespace(config *initConfig) error {
...
+	if doChdir {
+		// First, attempt the chdir before setting up the user.
+		// This could allow us to access a directory that the user running runc can access
+		// but the container user cannot.
+		err := unix.Chdir(config.Cwd)
+		switch {
+		case err == nil:
+			doChdir = false
+		case os.IsPermission(err):
+			// If we hit an EPERM, we should attempt again after setting up user.
```
kolyshkin, line 145: this needs to mention EACCES not EPERM.

kolyshkin (approved):
LGTM except for a nit in a comment (not that important)

mrunalp (approved):

PR #2899, opened 2021-04-09
 libcontainer: read root pids.current from /proc 

giuseppe:
more fixes on top of https://github.com/opencontainers/runc/pull/2873

when running on cgroup v2, if pids.current is not present in the
root (as of Linux 5.11 it is not) then read the number of current
processes from /proc.

Also revert a patch that causes a regression in the Kubelet when
vendored in Kubernetes: https://github.com/kubernetes/kubernetes/pull/100631#issuecomment-810001808

### DIFF
--- a/libcontainer/cgroups/fs2/pids.go
+++ b/libcontainer/cgroups/fs2/pids.go
@@ -3,7 +3,10 @@
 package fs2
 
 import (
+	"io/ioutil"
+	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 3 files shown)


cyphar:
Yeah `pids` hasn't ever accounted the root cgroup (on v1 nor v2) -- see torvalds/linux@67cde9c4938945b9510730c64e68d2f1dd7bc0aa.

giuseppe:
> Yeah `pids` hasn't ever accounted the root cgroup (on v1 nor v2) -- see [torvalds/linux@67cde9c](https://github.com/torvalds/linux/commit/67cde9c4938945b9510730c64e68d2f1dd7bc0aa).

Thanks!  So I am going to close this PR to not add more complexity given it works in the same way on cgroup v1.  If there is anything failing in cAdvisor/Kubernetes then I'll address it there

cyphar:
Sorry, I didn't mean to say "this PR is wrong", just that we'd need to adjust this for both v1 and v2 if we need special handling -- though are there real scenarios where folks run containers in the root cgroup (in the unified hierarchy no less -- where you wouldn't be able to even apply a devices policy)?

PR #2902, opened 2021-04-13
checkpoint: resolve symlink for external bind mount

liusdu:
runc resolves symlink before doing bind mount. So
we should save original path while formatting CriuReq for
checkpoint.

Signed-off-by: Liu Hua <weldonliu@tencent.com>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -779,6 +779,9 @@ const descriptorsFilename = "descriptors.json"
 
 func (c *linuxContainer) addCriuDumpMount(req *criurpc.CriuReq, m *configs.Mount) {
 	mountDest := strings.TrimPrefix(m.Destination, c.config.Rootfs)
+	if dest, err := securejoin.SecureJoin(c.config.Rootfs, mountDest); err == nil {
+		mountDest = dest[len(c.config.Rootfs):]
+	}
 	extMnt := &criurpc.ExtMountMap{
 		Key: proto.String(mountDest),
 		Val: proto.String(mountDest),
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -126,7 +126,18 @@ function simple_cr() {
 	done
 }
 
-@test "checkpoint and restore " {
+@test "checkpoint and restore" {
+	simple_cr
+}
+
+@test "checkpoint and restore (bind mount, destination is symlink)" {
+	mkdir -p rootfs/real/conf
+	ln -s /real/conf rootfs/conf
+	update_config '	  .mounts += [{
+					source: ".",
+					destination: "/conf",
+					options: ["bind"]
+				}]'
 	simple_cr
 }
 


kolyshkin:
@liusdu thank you for the patch! Nice that you're using checkpoint/restore!

Can you add a test case (to e.g. `tests/integration/checkpoint.bats`?)

kolyshkin:
@liusdu ^^^

liusdu:
> @liusdu ^^^

@kolyshkin  Sorry for this late,  checkpoint/restore  of runc works very well for me.

tests/integration/checkpoint.bats
```diff
@@ -129,6 +129,17 @@ function simple_cr() {
 @test "checkpoint and restore " {
 	simple_cr
 }
+@test "checkpoint and restore(bind mount, destination is symlink)" {
```
kolyshkin, line 132: Not: missing space before `(`

liusdu: Thans.. Fixed..

liusdu:
@kolyshkin

tests/integration/checkpoint.bats
```diff
@@ -129,6 +129,17 @@ function simple_cr() {
...
+@test "checkpoint and restore (bind mount, destination is symlink)" {
+	mkdir -p rootfs/real/conf
+	ln -s /real/conf rootfs/conf
+	update_config '	  .mounts += [{
+					source: ".",
+					destination: "/conf",
+					options: ["bind"]
+				}]'
+	simple_cr
+
```
kolyshkin, line 141: nit: extra empty line

libcontainer/container_linux.go
```diff
@@ -779,6 +779,10 @@ const descriptorsFilename = "descriptors.json"
 
 func (c *linuxContainer) addCriuDumpMount(req *criurpc.CriuReq, m *configs.Mount) {
 	mountDest := strings.TrimPrefix(m.Destination, c.config.Rootfs)
+
```
kolyshkin, line 782: nit: extra empty line

kolyshkin:
LGTM except for a couple of nits. @adrianreber PTAL

adrianreber:
@liusdu Can you give an example where you saw this error. It looks correct, but I am curious when this can happen.

liusdu:
@adrianreber  Look at following Dockerfile

```
FROM openjdk:8-jre
RUN wget  -qO-  'https://www.apache.org/dyn/closer.cgi?action=download&filename=flink/flink-1.12.2/flink-1.12.2-bin-scala_2.11.tgz' |tar -zvx -C /opt
RUN  ln -s  /opt/flink-1.12.2  /opt/flink
```

If we pass configuration as volume to container, we can meet this type of error.

`docker run -d -v  ${FLINK_CONF}`:/opt/flink/conf. ${flink_image}`

liusdu:
@kolyshkin thanks for you review~

tests/integration/checkpoint.bats
```diff
@@ -126,7 +126,17 @@ function simple_cr() {
 	done
 }
 
-@test "checkpoint and restore " {
+@test "checkpoint and restore" {
+	simple_cr
+}
+@test "checkpoint and restore (bind mount, destination is symlink)" {
```
kolyshkin, line 132: please add an empty line before @test

liusdu: Fixed

liusdu:
ping @kolyshkin @adrianreber

kolyshkin (approved):
LGTM

kolyshkin:
close/reopen to re-kick CI

tests/integration/checkpoint.bats
```diff
@@ -126,7 +126,18 @@ function simple_cr() {
...
 }
 
-@test "checkpoint and restore " {
+@test "checkpoint and restore" {
+	simple_cr
+}
+
+@test "checkpoint and restore (bind mount, destination is symlink)" {
+	mkdir -p rootfs/real/conf
+	ln -s /real/conf rootfs/conf
```
haircommander, line 135: This is not blocking, but just curious:
why an absolute path for `/real/conf`? I see we're using the same path in `mounts.bats`, so it works, but it's not totally clear to me why we're using this path rather than somewhere in a tmpfs (or at least something test-specific)

liusdu: Hi @haircommander 
Commit [runc run: resolve tmpfs mount dest in container scope](https://github.com/opencontainers/runc/commit/637f82d6b8b45a99519b863f8cf7461760da7a97)  changes behavior of tmpfs mount and bind mount, when the destination is symlink.
 -  changes of bind mount breaks the functionality of runc checkpoint. This is why we need this path.
 -  changes of tmpfs mount does not break anything. Since tmpfs mount is fully owned by container's mnt namespace.

haircommander: thanks for the explanation, makes sense

libcontainer/container_linux.go
```diff
@@ -779,6 +779,9 @@ const descriptorsFilename = "descriptors.json"
 
 func (c *linuxContainer) addCriuDumpMount(req *criurpc.CriuReq, m *configs.Mount) {
 	mountDest := strings.TrimPrefix(m.Destination, c.config.Rootfs)
+	if dest, err := securejoin.SecureJoin(c.config.Rootfs, mountDest); err == nil {
+		mountDest = dest[len(c.config.Rootfs):]
```
cyphar, line 783: I do wonder -- what happens if the mount is moved from inside the container (with a simple `rename`) after the filesystem has been set up? Is this also something CRIU doesn't support / really doesn't like?

liusdu: Yes @cyphar  Since CRIU does not know how to restore bind mounts(master in host namespace),  CRIU need  this type of mount marked as external(via `--external`). 
  runc formats `--external` parameter according to spec((link)[https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L1112]). So for situation you metioned,  criu does not work properly.  btw, should we export `--external` for runc. Then users can mark external resource themselves?

adrianreber: >> I do wonder -- what happens if the mount is moved from inside the container (with a simple rename) after the filesystem has been set up? Is this also something CRIU doesn't support / really doesn't like?

I do not understand this question. But initially I would have said that CRIU can handle it. Now @liusdu says it is not possible. I probably misunderstood the question.

> So for situation you mentioned, CRIU does not work properly. btw, should we export `--external` for runc. Then users can mark external resource themselves?

For Kubernetes, restoring containers in different Pods, I will need CRIU's `--external` to be available at the runc level.

Currently runc informs CRIU about all external mounts but it is not possible to change the location of the external mount between checkpoint and restore. For Kubernetes I will need to change the location of an external mount. The most obvious bind mount currently is the location of the secrets directory which has the Pod UID in its container external path and I need to tell CRIU/runc that that directory should be mounted from some other location.

liusdu: ping @cyphar  ..

liusdu:
@haircommander can you approve this pr?

haircommander:
It LGTM, but I'm not a maintainer so my approval wouldn't help unfortunately
@kolyshkin @mrunalp @AkihiroSuda PTAL

kolyshkin (approved):
LGTM

kolyshkin:
We're in the middle of 1.0.0 GA release. Once it's done, I think we can merge this.

mrunalp (approved):
LGTM

kolyshkin:
There's something wrong with CI :(

kolyshkin:
Might be caused by https://github.com/opencontainers/runc/pull/3029. @liusdu can you take a look? The failed CI can be seen from the Actions tab, https://github.com/opencontainers/runc/actions

Issue #2699, opened 2020-12-03
cgroups/ebpf: switch to new cilium/ebpf API


kolyshkin:
`cilium/ebpf`'s `prog.{Attach,Detach}` API is now deprecated in favor of new link package (see  https://github.com/cilium/ebpf/pull/108/commits/5727d650c09d79ac987b3c2d4e5420e6393cd5c7 and https://github.com/cilium/ebpf/pull/110).

runc need to adopt.

```console
libcontainer/cgroups/ebpf/ebpf.go:35:12: SA1019: prog.Attach is deprecated: use link.RawAttachProgram instead.  (staticcheck)
	if err := prog.Attach(dirFD, ebpf.AttachCGroupDevice, unix.BPF_F_ALLOW_MULTI); err != nil {
	          ^
libcontainer/cgroups/ebpf/ebpf.go:39:13: SA1019: prog.Detach is deprecated: use link.RawDetachProgram instead.  (staticcheck)
		if err := prog.Detach(dirFD, ebpf.AttachCGroupDevice, unix.BPF_F_ALLOW_MULTI); err != nil {
		          ^
```

I don't know much about these matters; @cyphar mind to take a look?

thaJeztah:
missed there was a tracking issue for this; I opened https://github.com/opencontainers/runc/pull/2903, which also revealed a potential bug in the closer (pls double-check that one)

PR #2145, opened 2019-10-17
cgroup2: port over eBPF device controller from crun

AkihiroSuda:
The implementation is based on https://github.com/containers/crun/blob/0.10.2/src/libcrun/ebpf.c

Although `ebpf.c` is originally licensed under LGPL-3.0-or-later, the author Giuseppe Scrivano agreed to relicense the file in Apache License 2.0: https://github.com/opencontainers/runc/issues/2144#issuecomment-543116397

See `libcontainer/cgroups/ebpf/devicefilter/devicefilter_test.go` for tested configurations.

Close #2144

PR #2903, opened 2021-04-13
ebpf: replace deprecated prog.Attach/prog.Detach

thaJeztah:
fixes https://github.com/opencontainers/runc/issues/2699

Caught by golangci-lint when enabling golint:

    libcontainer/cgroups/ebpf/ebpf.go:35:12: SA1019: prog.Attach is deprecated: use link.RawAttachProgram instead. (staticcheck)
        if err := prog.Attach(dirFD, ebpf.AttachCGroupDevice, unix.BPF_F_ALLOW_MULTI); err != nil {
                  ^
    libcontainer/cgroups/ebpf/ebpf.go:39:13: SA1019: prog.Detach is deprecated: use link.RawDetachProgram instead. (staticcheck)
            if err := prog.Detach(dirFD, ebpf.AttachCGroupDevice, unix.BPF_F_ALLOW_MULTI); err != nil {
                      ^

Worth noting that we currently call prog.Detach() with unix.BPF_F_ALLOW_MULTI;
https://github.com/golang/sys/blob/22da62e12c0cd9c1da93581e1113ca4d82a5be14/unix/zerrors_linux.go#L178

    BPF_F_ALLOW_MULTI = 0x2

⚠️ Looking at the source code for prog.Detach(); https://github.com/cilium/ebpf/blob/v0.4.0/prog.go#L579-L581,
this would _always_ produce an error:

    if flags != 0 {
        return errors.New("flags must be zero")
    }

Note that the flags parameter is not used (except for that validation)

### DIFF
--- a/libcontainer/cgroups/ebpf/ebpf.go
+++ b/libcontainer/cgroups/ebpf/ebpf.go
@@ -3,6 +3,7 @@ package ebpf
 import (
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/link"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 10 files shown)


thaJeztah:
@AkihiroSuda @giuseppe @kolyshkin PTAL; please double check if I did the conversion correctly, and if my reading of the `prog.Detach()` issue is correct, as I'm really not familiar with this.

I see the `prog.Detach()` was added in https://github.com/opencontainers/runc/commit/faf673ee45bae8420f27d4e040579e6969cc009b (https://github.com/opencontainers/runc/pull/2145), based on the c implementation in crun

thaJeztah:
also /cc (I see there was discussion about this code in https://github.com/opencontainers/runc/issues/2366)

thaJeztah:
also /cc @cyphar (I see there was discussion about this code in https://github.com/opencontainers/runc/issues/2366)

AkihiroSuda (approved):

kolyshkin (approved):
LGTM, but please DNM until we get LGTM from @cyphar

kolyshkin:
Also, I agree with your analysis (`Detach` should not be called with `BPF_F_ALLOW_MULTI`).

cyphar (approved):
Oh, good -- looking at `RawDetachProgram` it looks like this will mean we only detach the program we just inserted (though more broadly we still do the wrong thing on cgroup updates). This was not the case with our incorrect proper usage of `Detach` AFAICS.

LGTM.

PR #2909, opened 2021-04-14
libct/int: misc cleanups and fixes

kolyshkin:
Please see individual commits for details. Nothing serious, but improved container names and cgroups randomization might help to reduce flakes.

### DIFF
--- a/libcontainer/integration/template_test.go
+++ b/libcontainer/integration/template_test.go
@@ -1,8 +1,9 @@
 package integration
 
 import (
-	"math/rand"
 	"strconv"
+	"testing"
+	"time"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/devices"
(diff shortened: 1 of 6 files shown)


AkihiroSuda (approved):

libcontainer/integration/template_test.go
```diff
@@ -213,9 +214,9 @@ func newTemplateConfig(p *tParam) *configs.Config {
 	}
 
 	if p.systemd {
-		id := strconv.FormatUint(rand.Uint64(), 36)
-		config.Cgroups.Name = "test" + id
-		// do not change Parent (see newContainerWithName)
+		id := strconv.FormatInt(-int64(time.Now().Nanosecond()), 36)
```
fuweid, line 217: minus~ good one

fuweid: Why not just use `rand.Uint64` to prevent from using the same id

kolyshkin: Because it needs to be seeded somewhere

mrunalp (approved):

kolyshkin:
this is CI only stuff, so merging

PR #2911, opened 2021-04-15
logging: enable file/line info if --debug is set

kolyshkin:
This helps a lot to find out where the errors come from.

Before:
> level=warning msg="lstat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/test_hello: no such file or directory"

After:
> level=warning msg="lstat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/test_hello: no such file or directory" func=github.com/opencontainers/runc/libcontainer.destroy file="github.com/opencontainers/runc/libcontainer/state_linux.go:44"

Presumably this comes with an overhead, but I guess no one is using
`--debug` by default anyway.

PS I'd love to cut the 'github.com/opencontainers/runc/` prefix but don't see an easy way to :-\

### DIFF
--- a/libcontainer/logs/logs.go
+++ b/libcontainer/logs/logs.go
@@ -24,6 +24,7 @@ type Config struct {
 	LogFormat   string
 	LogFilePath string
 	LogPipeFd   int
+	LogCaller   bool
 }
 
 func ForwardLogs(logPipe io.ReadCloser) chan error {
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

fuweid (approved):
LGTM(nb)

mrunalp (approved):

PR #2912, opened 2021-04-16
Add cfs throttle stats to cgroup v2

odinuge:
This adds cfs throttle stats to cgroup v2.

More info here: https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html

Tried adding a test, but cgroup v2 have no tests, and the test infrastructure is tightly coupled with cgroup v1. Adding test infra to v2 is outside the scope of this PR.

### DIFF
--- a/libcontainer/cgroups/fs2/cpu.go
+++ b/libcontainer/cgroups/fs2/cpu.go
@@ -70,6 +70,15 @@ func statCpu(dirPath string, stats *cgroups.Stats) error {
 
 		case "system_usec":
 			stats.CpuStats.CpuUsage.UsageInKernelmode = v * 1000
+
+		case "nr_periods":
+			stats.CpuStats.ThrottlingData.Periods = v
+
+		case "nr_throttled":
+			stats.CpuStats.ThrottlingData.ThrottledPeriods = v
+
+		case "throttled_usec":
+			stats.CpuStats.ThrottlingData.ThrottledTime = v * 1000
 		}
 	}
 	return nil


kolyshkin (approved):
LGTM

AkihiroSuda:
> Tried adding a test, but cgroup v2 have no tests, and the test infrastructure is tightly coupled with cgroup v1. Adding test infra to v2 is outside the scope of this PR.

We have v2 infra https://github.com/opencontainers/runc/blob/fce58ab2d5c488bc573d02712db476a6daa9a60c/.github/workflows/test.yml#L75

odinuge:
> We have v2 infra

Ahh, should probably been more specific and said unit tests instead. Thought about adding a test like this one for `cgroups/fs`, 
https://github.com/opencontainers/runc/blob/fce58ab2d5c488bc573d02712db476a6daa9a60c/libcontainer/cgroups/fs/cpu_test.go#L105-L134

but `cgroups/fs2` have no unit tests like these, and the tooling used in `cgroups/fs` tests has to be copied/changed to be used in `cgroups/fs2`.

Can take a look at adding an integration test of `runc events --stats` that checks throttling, but there is currently no tests about it yet. Should I look at that in this PR, or in another?

mrunalp (approved):

AkihiroSuda (approved):

AkihiroSuda:
> Can take a look at adding an integration test of runc events --stats that checks throttling, but there is currently no tests about it yet. Should I look at that in this PR, or in another?

Can be another PR, thanks

PR #2913, opened 2021-04-19
tiny fix iterative checkpoint test case

liusdu:
Criu creates symlink named "parent" under image-path while doing
interactive checkpoint, without checking wether symlink points to
right place. This patch corrects related test case.

Signed-off-by: Liu Hua <weldonliu@tencent.com>

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"strconv"
 
 	criu "github.com/checkpoint-restore/go-criu/v5/rpc"
@@ -78,12 +79,37 @@ checkpointed.`,
 	},
 }
 
-func getCheckpointImagePath(context *cli.Context) string {
+func prepareImagePaths(context *cli.Context) (string, string, error) {
 	imagePath := context.String("image-path")
 	if imagePath == "" {
 		imagePath = getDefaultImagePath(context)
 	}
-	return imagePath
+
+	if err := os.MkdirAll(imagePath, 0600); err != nil {
+		return "", "", err
+	}
+
+	parentPath := context.String("parent-path")
+	if parentPath == "" {
+		return imagePath, parentPath, nil
+	}
+
+	if filepath.IsAbs(parentPath) {
+		return "", "", errors.New("--parent-path must be relative")
+	}
+
+	realParent := filepath.Join(imagePath, parentPath)
+	fi, err := os.Stat(realParent)
+	if err == nil && !fi.IsDir() {
+		err = &os.PathError{Path: realParent, Err: unix.ENOTDIR}
+	}
+
+	if err != nil {
+		return "", "", fmt.Errorf("invalid --parent-path: %w", err)
+	}
+
+	return imagePath, parentPath, nil
+
 }
 
 func setPageServer(context *cli.Context, options *libcontainer.CriuOpts) {
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -159,9 +159,12 @@ function simple_cr() {
 	# checkpoint the running container
 	mkdir image-dir
 	mkdir work-dir
-	runc --criu "$CRIU" checkpoint --parent-path ./parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	runc --criu "$CRIU" checkpoint --parent-path ../parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
 	grep -B 5 Error ./work-dir/dump.log || true
 	[ "$status" -eq 0 ]
+
+	# check parent path is valid
+	[ -e ./image-dir/parent ]
 
 	# after checkpoint busybox is no longer running
 	testcontainer test_busybox checkpointed
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -144,6 +144,23 @@ function simple_cr() {
 	simple_cr
 }
 
+@test "checkpoint --pre-dump (bad --parent-path)" {
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 0 ]
+
+	testcontainer test_busybox running
+
+	# runc should fail with absolute parent image path.
+	runc --criu "$CRIU" checkpoint --parent-path "$(pwd)"/parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	[[ "${output}" == *"--parent-path"* ]]
+	[ "$status" -ne 0 ]
+
+	# runc should fail with invalid parent image path.
+	runc --criu "$CRIU" checkpoint --parent-path ./parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	[[ "${output}" == *"--parent-path"* ]]
+	[ "$status" -ne 0 ]
+}
+
 @test "checkpoint --pre-dump and restore" {
 	setup_pipes
 	runc_run_with_pipes test_busybox
--- a/restore.go
+++ b/restore.go
@@ -121,14 +121,15 @@ using the runc checkpoint command.`,
 }
 
 func criuOptions(context *cli.Context) *libcontainer.CriuOpts {
-	imagePath := getCheckpointImagePath(context)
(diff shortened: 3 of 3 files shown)


kolyshkin:
CI got stuck in Fedora -- restarted.

kolyshkin:
@adrianreber PTAL 🙏🏻

adrianreber:
The change looks correct. `--parent-path` needs to point to the previous checkpoint directory from the current checkpoint's directory point of view.

@liusdu have you seen any errors? Where did you seen errors. Just curious.

liusdu:
> The change looks correct. `--parent-path` needs to point to the previous checkpoint directory from the current checkpoint's directory point of view.
> 
> @liusdu have you seen any errors? Where did you seen errors. Just curious.

 @kolyshkin  I  repushed  the commit. And all tests passed.

@adrianreber  I did iterative checkpoint test and got several types of error  with wrong `--parent-path` value:
1.  invalid `--parent-path` : Last dump has everything and pre-dump is useless at all ( what runc test does)
2. right `--parent-path`  but with absolute path: all dumps go well but restore will hang(criu just ignore partent images)

So I went into runc code to debug. And I found this wrong test.

Bte, should I add some piece of code to validate  this?

liusdu:
@adrianreber @calavera  I add a patch to validate parent path. 
Please take a look. I can revert it if not necessary.

checkpoint.go
```diff
@@ -78,12 +79,30 @@ checkpointed.`,
 	},
 }
 
-func getCheckpointImagePath(context *cli.Context) string {
+func prepareImagePaths(context *cli.Context) (string, string, error) {
 	imagePath := context.String("image-path")
+
```
kolyshkin, line 84: no need to add a newline here

checkpoint.go
```diff
@@ -78,12 +79,30 @@ checkpointed.`,
...
 		imagePath = getDefaultImagePath(context)
 	}
-	return imagePath
+
+	if err := os.MkdirAll(imagePath, 0755); err != nil {
+		return "", "", err
+	}
+
+	parentPath := context.String("parent-path")
+
```
kolyshkin, line 94: Please remove this newline

checkpoint.go
```diff
@@ -78,12 +79,30 @@ checkpointed.`,
...
+	if parentPath == "" {
+		return imagePath, parentPath, nil
+	}
+
+	realParent := filepath.Join(imagePath, parentPath)
```
kolyshkin, line 99: I guess you should check first if `parentPath` is absolute, and error out if it's not
```go
if !filepath.IsAbs(parentPath) {
   return errors.New("--parent-path must be relative")
}
```
or so.

tests/integration/checkpoint.bats
```diff
@@ -159,10 +159,13 @@ function simple_cr() {
 	# checkpoint the running container
 	mkdir image-dir
 	mkdir work-dir
-	runc --criu "$CRIU" checkpoint --parent-path ./parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	runc --criu "$CRIU" checkpoint --parent-path ../parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
```
kolyshkin, line 173: So what I still don't understand is why the test was not failing.

checkpoint.go
```diff
@@ -78,12 +79,30 @@ checkpointed.`,
...
+	if fi, err := os.Stat(realParent); err != nil || !fi.IsDir() {
+		return "", "", fmt.Errorf("--parent-path needs to be relative path of --image-path, and points to previous checkpoint directory")
```
kolyshkin, line 101: I think this belongs to option description, in here you can just return `fmt.Errorf("invalid --parent-path %s: %w", parentPath, err)` or something like that.

kolyshkin: ...with
```go
if err == nil && !fi.IsDir {
   err := &os.PathError(Path: realParent, Err: unix.ENOTDIR)
}
```

tests/integration/checkpoint.bats
```diff
@@ -159,6 +159,17 @@ function simple_cr() {
...
+	# runc will fail whith absolute parent image path.
+	runc --criu "$CRIU" checkpoint --parent-path "$(pwd)"/parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	[[ "${output}" == *"--parent-path needs to be relative path of --image-path, and points to previous checkpoint directory"* ]]
+	[ "$status" -ne 0 ]
+
+	# runc will fail whith invalid parent image path.
+	runc --criu "$CRIU" checkpoint --parent-path ./parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	[[ "${output}" == *"--parent-path needs to be relative path of --image-path, and points to previous checkpoint directory"* ]]
+	[ "$status" -ne 0 ]
+
```
kolyshkin: simple_cr() is called from many places, I think this deserves a separate test case. Also, I guess it's enough to check that runc fails, and the error contains "--parent-path", rather than checking for a full specific error message (which might change).

tests/integration/checkpoint.bats
adrianreber: I think that is CRIU's fault and confuses many people when using pre-dump. CRIU silently switched to non pre-dump if `--parent-path` is not correct. https://github.com/checkpoint-restore/criu/pull/1403

Restore also works without any message, but it is not using the pre-dump. This will probably be an error in the next CRIU release. Having it fixed in runc already know makes sense.

tests/integration/checkpoint.bats
liusdu: @kolyshkin Thanks very much for you review messages～.  I add a test case  to  test  `--parent-path`. Am I missing something？

tests/integration/checkpoint.bats
```diff
@@ -144,6 +144,35 @@ function simple_cr() {
 	simple_cr
 }
 
+@test "checkpoint --pre-dump (with invalid parent path)" {
+
```
kolyshkin, line 148: extra empty line

tests/integration/checkpoint.bats
```diff
@@ -144,6 +144,35 @@ function simple_cr() {
...
+	setup_pipes
+	runc_run_with_pipes test_busybox
```
kolyshkin, line 150: As you don't really checkpoint or restore anything, there's no need to do `setup_pipes` / `runc_run_with_pipes` / `check_pipes`. A simple
```bats
        [ "$status" -eq 0 ]
```
should be enough.

tests/integration/checkpoint.bats
```diff
@@ -144,6 +144,35 @@ function simple_cr() {
...
+	runc --criu "$CRIU" checkpoint --parent-path "$(pwd)"/parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	[[ "${output}" == *"--parent-path"* ]]
+	[ "$status" -ne 0 ]
+
+	# runc should fail with invalid parent image path.
+	runc --criu "$CRIU" checkpoint --parent-path ./parent-dir --work-path ./work-dir --image-path ./image-dir test_busybox
+	[[ "${output}" == *"--parent-path"* ]]
+	[ "$status" -ne 0 ]
+
+	check_pipes
```
kolyshkin, line 173: As per the above comment, remove this. This is to check that the restored container works, we don't restore the container so ...

tests/integration/checkpoint.bats
```diff
@@ -144,6 +144,35 @@ function simple_cr() {
...
+	runc_run_with_pipes test_busybox
+
+	mkdir parent-dir
+	runc --criu "$CRIU" checkpoint --pre-dump --image-path ./parent-dir test_busybox
+	[ "$status" -eq 0 ]
+
+	# busybox should still be running
+	testcontainer test_busybox running
+
+	# checkpoint the running container
```
kolyshkin, line 159: this comment is misleading, please remove

kolyshkin:
A few comments about simplification of "bad --parent-path" test case. Otherwise LGTM

kolyshkin (approved):
LGTM

liusdu:
ping @adrianreber

liusdu:
@kolyshkin all test passed

kolyshkin:
@liusdu I have approved these changes a while ago, waiting for the second LGTM. @opencontainers/runc-maintainers PTAL

haircommander:
LGTM

PR #2915, opened 2021-04-20
Update utils_linux.go

xiaoding945:
does func destory is  duplicate? i think whatever it will be destroyed，it clould just appear once in last


xiaoding945:
if any err occours , need destory

PR #2917, opened 2021-04-20
Validate mounts

kolyshkin:
According to OCI spec, mount destination MUST be an absolute path. We never validate it, I think we should.

Add validation to both specconv and configs/validate, and error out on non-abs paths.


cyphar (approved):
LGTM.

AkihiroSuda (approved):

PR #2918, opened 2021-04-20
libct/cg/fs/freezer: make sure to thaw on failure

kolyshkin:
Function `(*FreezerGroup).Set` has a few paths where in can return an
error. In any case, if an error is returned, we failed to freeze,
and we need to thaw to avoid leaving the cgroup in a stuck state.

This is a continuation of #2791 and #2774, addressing #2753 and similar bugs.


mrunalp (approved):

AkihiroSuda (approved):
nit: `s/Err/retErr/` would be more Go-ish, but not a huge deal, so merging

PR #2919, opened 2021-04-26
cloned_binary: switch from #error to #warning for SYS_memfd_create

cyphar:
We shouldn't refuse to build on architectures just because we don't know
what the syscall number of memfd_create(2) is. In addition, use the
correct defined(...) macros for ppc64 (these are the ones glibc uses).

Fixes: 3aead32ea246 ("nsenter: hard-code memfd_create(2) syscall numbers")


AkihiroSuda (approved):

cyphar:
Since this regressed builds on ppc64 and is such a trivial patch, I'm going to add this to the rc milestone.

PR #2920, opened 2021-04-26
libcontainer/intelrdt: support ClosID parameter

marquiz:
- Handle ClosID parameter of IntelRdt. Makes it possible to use
pre-configured classes/ClosIDs and avoid running out of available IDs
which easily happens with per-container classes.

- libcontainer/intelrdt: verify ClosID existence
Check that the ClosID directory pre-exists if no L3 or MB schema has
been specified. Conform with the following line from runtime-spec
(config-linux):

  *If closID is set, and neither of l3CacheSchema and memBwSchema are
  set, runtime MUST check if corresponding pre-configured directory
  closID is present in mounted resctrl. If such pre-configured directory
  closID exists, runtime MUST assign container to this closID and
  generate an error if directory does not exist.*

- Add a TODO note for verifying existing schemata against L3/MB
parameters.

- Run unit tests irrespective of the underlying system configuration, i.e.
even if RDT has not been enabled or is not supported. The tests do not
depend on real kernel interfaces.

### DIFF
--- a/libcontainer/configs/intelrdt.go
+++ b/libcontainer/configs/intelrdt.go
@@ -1,6 +1,9 @@
 package configs
 
 type IntelRdt struct {
+	// The identity for RDT Class of Service
+	ClosID string `json:"closID,omitempty"`
+
 	// The schema for L3 cache id and capacity bitmask (CBM)
 	// Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
 	L3CacheSchema string `json:"l3_cache_schema,omitempty"`
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -205,18 +205,15 @@ func (v *ConfigValidator) intelrdt(config *configs.Config) error {
 			return errors.New("intelRdt is specified in config, but Intel RDT is not supported or enabled")
 		}
 
+		if config.IntelRdt.ClosID == "." || config.IntelRdt.ClosID == ".." || strings.Contains(config.IntelRdt.ClosID, "/") {
+			return fmt.Errorf("invalid intelRdt.ClosID %q", config.IntelRdt.ClosID)
+		}
+
 		if !intelrdt.IsCATEnabled() && config.IntelRdt.L3CacheSchema != "" {
 			return errors.New("intelRdt.l3CacheSchema is specified in config, but Intel RDT/CAT is not enabled")
 		}
 		if !intelrdt.IsMBAEnabled() && config.IntelRdt.MemBwSchema != "" {
 			return errors.New("intelRdt.memBwSchema is specified in config, but Intel RDT/MBA is not enabled")
-		}
-
-		if intelrdt.IsCATEnabled() && config.IntelRdt.L3CacheSchema == "" {
-			return errors.New("Intel RDT/CAT is enabled and intelRdt is specified in config, but intelRdt.l3CacheSchema is empty")
-		}
-		if intelrdt.IsMBAEnabled() && config.IntelRdt.MemBwSchema == "" {
-			return errors.New("Intel RDT/MBA is enabled and intelRdt is specified in config, but intelRdt.memBwSchema is empty")
 		}
 	}
 
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -205,7 +205,6 @@ var (
 )
 
 type intelRdtData struct {
-	root   string
 	config *configs.Config
 }
 
@@ -548,6 +547,14 @@ func (m *intelRdtManager) Apply(pid int) (err error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if m.config.IntelRdt.ClosID != "" && m.config.IntelRdt.L3CacheSchema == "" && m.config.IntelRdt.MemBwSchema == "" {
+		// Check that the CLOS exists, i.e. it has been pre-configured to
+		// conform with the runtime spec
+		if _, err := os.Stat(path); err != nil {
+			return fmt.Errorf("clos dir not accessible (must be pre-created when l3CacheSchema and memBwSchema are empty): %w", err)
+		}
+	}
+
 	if err := os.MkdirAll(path, 0o755); err != nil {
 		return newLastCmdError(err)
 	}
@@ -722,6 +729,12 @@ func (m *intelRdtManager) Set(container *configs.Config) error {
 		path := m.GetPath()
 		l3CacheSchema := container.IntelRdt.L3CacheSchema
 		memBwSchema := container.IntelRdt.MemBwSchema
+
+		// TODO: verify that l3CacheSchema and/or memBwSchema match the
+		// existing schemata if ClosID has been specified. This is a more
+		// involved than reading the file and doing plain string comparison as
+		// the value written in does not necessarily match what gets read out
+		// (leading zeros, cache id ordering etc).
 
 		// Write a single joint schema string to schemata file
 		if l3CacheSchema != "" && memBwSchema != "" {
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -525,7 +525,12 @@ func (m *intelRdtManager) getIntelRdtPath() (string, error) {
 		return "", err
 	}
 
-	return filepath.Join(rootPath, m.id), nil
+	clos := m.id
+	if m.config.IntelRdt != nil && m.config.IntelRdt.ClosID != "" {
+		clos = m.config.IntelRdt.ClosID
+	}
+
+	return filepath.Join(rootPath, clos), nil
 }
 
 // Applies Intel RDT configuration to the process with the specified pid
@@ -557,12 +562,17 @@ func (m *intelRdtManager) Apply(pid int) (err error) {
 
 // Destroys the Intel RDT 'container_id' group
 func (m *intelRdtManager) Destroy() error {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-	if err := os.RemoveAll(m.GetPath()); err != nil {
-		return err
+	// Don't remove resctrl group if closid has been explicitly specified. The
+	// group is likely externally managed, i.e. by some other entity than us.
+	// There are probably other containers/tasks sharing the same group.
+	if m.config.IntelRdt == nil || m.config.IntelRdt.ClosID == "" {
+		m.mu.Lock()
+		defer m.mu.Unlock()
+		if err := os.RemoveAll(m.GetPath()); err != nil {
+			return err
+		}
+		m.path = ""
 	}
-	m.path = ""
 	return nil
 }
 
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -72,7 +72,7 @@ import (
  * |-- ...
  * |-- schemata
  * |-- tasks
- * |-- <container_id>
+ * |-- <clos>
  *     |-- ...
  *     |-- schemata
  *     |-- tasks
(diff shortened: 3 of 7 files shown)


libcontainer/intelrdt/intelrdt.go
```diff
@@ -563,6 +562,14 @@ func (m *intelRdtManager) Apply(pid int) (err error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if m.config.IntelRdt.ClosID != "" && m.config.IntelRdt.L3CacheSchema == "" && m.config.IntelRdt.MemBwSchema == "" {
+		// Check that the CLOS exists, i.e. it has been pre-configured to
+		// conform with the runtime spec
+		if _, err := os.Stat(path); err != nil {
+			return fmt.Errorf("unable to access %q (must be pre-created when l3CacheSchema and memBwSchema are empty): %v", path, err)
```
kolyshkin, line 569: 1. an error from os.Stat already contains path, no need to have it twice in the error message.
2. use `%w` for error.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -738,6 +745,12 @@ func (m *intelRdtManager) Set(container *configs.Config) error {
 		l3CacheSchema := container.IntelRdt.L3CacheSchema
 		memBwSchema := container.IntelRdt.MemBwSchema
 
+		// TODO: verify hat l3CacheSchema and/or memBwSchema match with
```
kolyshkin, line 748: typo: `s/ hat / that /`

kolyshkin:
Looks good overall, added some notes.

Also, the first commit performs a nice refactoring (which I like), but it is kind of hard to review because it also introduces ClosID support. Can you maybe split it into (1) refactoring (2) adding ClosID?

libcontainer/intelrdt/intelrdt.go
marquiz: Fixed

libcontainer/intelrdt/intelrdt.go
marquiz: Valid point, changed now to `"clos dir not accessible (must be pre-created when l3CacheSchema and memBwSchema are empty): %w"`

marquiz:
Thanks for the review @kolyshkin!

> Also, the first commit performs a nice refactoring (which I like), but it is kind of hard to review because it also introduces ClosID support. Can you maybe split it into (1) refactoring (2) adding ClosID?

Yeah, I see 😊 I now split the first commit into two

marquiz:
BTW, related, I submitted a PR against runtime-spec in an attempt to make it easier read:
https://github.com/opencontainers/runtime-spec/pull/1104

marquiz:
Ping @kolyshkin. Could you enable testing?

libcontainer/intelrdt/intelrdt.go
```diff
@@ -738,6 +745,12 @@ func (m *intelRdtManager) Set(container *configs.Config) error {
 		l3CacheSchema := container.IntelRdt.L3CacheSchema
 		memBwSchema := container.IntelRdt.MemBwSchema
 
+		// TODO: verify that l3CacheSchema and/or memBwSchema match with
```
kolyshkin, line 748: nit: `s/match with/match the/`

libcontainer/intelrdt/intelrdt.go
```diff
@@ -738,6 +745,12 @@ func (m *intelRdtManager) Set(container *configs.Config) error {
...
+		// existing schemata if ClosID has been specified. This is a more
+		// involved than reading the file and doing plain string comparison as
+		// the value written in does not necessarily match what gets read out
+		// (leading zeros, cache id ordering etc).
```
kolyshkin, line 752: Maybe it makes sense to relax the runtime spec and say "runtime MAY check" rather than "must check", because it if is a must and you're not doing it, you're not conformant.

libcontainer/intelrdt/intelrdt.go
marquiz: Fixed

libcontainer/intelrdt/intelrdt.go
marquiz: Might be, but I'm not entirely sure as it leads to inconsistency. User asking for `ClosID=foo,  L3Schema=bar` might get what she asks for or then something totally different.

Anyways, I think that should be debated in runtime-spec repo instead of runc.

libcontainer/configs/validate/validator.go
```diff
@@ -200,19 +200,16 @@ func (v *ConfigValidator) intelrdt(config *configs.Config) error {
 			return errors.New("intelRdt is specified in config, but Intel RDT is not supported or enabled")
 		}
 
+		if config.IntelRdt.ClosID == "." || config.IntelRdt.ClosID == ".." || strings.ContainsRune(config.IntelRdt.ClosID, '/') {
```
kolyshkin, line 203: I think `strings.ContainsRune` is overkill here, as we don't care about unicode, and `strings.Contains` should be sufficient (and probably faster).

libcontainer/configs/intelrdt.go
```diff
@@ -1,6 +1,9 @@
 package configs
 
 type IntelRdt struct {
+	// The identity for RDT Class of Service
```
kolyshkin, line 4: Perhaps you need to document it more, for example, say that
 * if not set, it equals the container ID;
 * if explicitly set, it is assumed that the corresponding resctrl directory is managed externally, and it is not removed by runc when he container is destroyed.

kolyshkin: Or, if this is sufficiently documented in runtime-spec, please disregard the above comment.

marquiz: K

libcontainer/configs/validate/validator.go
marquiz: Hmm, good point! I was thinking that `ContainsRune` would be better than substring matching but now that I look at the source code of the strings package that's not the case. Changed to `strings.Contains()`

marquiz:
@kolyshkin anything I can do to advance this? The latest review comment(s) should be addressed, now

marquiz:
ping @kolyshkin

marquiz:
How to re-run tests??
/retest

marquiz:
ping @kolyshkin. How to get this forward?

cyphar:
We need to manually retrigger them, there isn't a bot to control them unfortunately.

kolyshkin:
> you can retrigger it by force pushing a new commit

One other way is to close and reopen a PR.

marquiz:
> (I've retriggered it.)

Thanks @cyphar! Everything's green now 😎

marquiz:
ping @crosbymichael @mrunalp @dqminh @hqhq @cyphar @AkihiroSuda @kolyshkin

Any comments, advice?

libcontainer/specconv/spec_linux.go
```diff
@@ -301,6 +301,9 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 		}
 		if spec.Linux.IntelRdt != nil {
 			config.IntelRdt = &configs.IntelRdt{}
+			if spec.Linux.IntelRdt.ClosID != "" {
+				config.IntelRdt.ClosID = spec.Linux.IntelRdt.ClosID
+			}
```
kolyshkin, line 306: So, this makes no sense, as both variables are strings, the default value is `""` and so `if` is totally redundant.

I just found this by looking into intelrtd code; will open a small fix PR soon.

kolyshkin: Resulted in not-so-small PR although I think it makes sense to have that before this one.

@marquiz PTAL  https://github.com/opencontainers/runc/pull/3041 🙏🏻

marquiz: Resolved by rebase

marquiz:
Rebased

libcontainer/intelrdt/intelrdt.go
```diff
@@ -579,7 +571,7 @@ func (m *intelRdtManager) Destroy() error {
 // restore the object later
 func (m *intelRdtManager) GetPath() string {
 	if m.path == "" {
-		m.path, _ = GetIntelRdtPath(m.id)
+		m.path, _ = m.getIntelRdtPath()
 	}
 	return m.path
 }
```
kolyshkin, line 594: Why not merge `getIntelRdtPath` into here, and use it everywhere?

libcontainer/intelrdt/intelrdt.go
```diff
@@ -548,16 +535,21 @@ func (m *intelRdtManager) Apply(pid int) (err error) {
 	if m.config.IntelRdt == nil {
 		return nil
 	}
-	d, err := getIntelRdtData(m.config, pid)
+
+	path, err := m.getIntelRdtPath()
```
kolyshkin, line 543: Why not m.GetPath()?

kolyshkin: Ah, I see, getIntelRdtPath returns an error, while GetPath does not.

libcontainer/intelrdt/intelrdt.go
kolyshkin: Nevermind.

libcontainer/configs/validate/validator.go
```diff
@@ -205,19 +205,16 @@ func (v *ConfigValidator) intelrdt(config *configs.Config) error {
...
 		if !intelrdt.IsMBAEnabled() && config.IntelRdt.MemBwSchema != "" {
 			return errors.New("intelRdt.memBwSchema is specified in config, but Intel RDT/MBA is not enabled")
 		}
-
-		if intelrdt.IsCATEnabled() && config.IntelRdt.L3CacheSchema == "" {
-			return errors.New("Intel RDT/CAT is enabled and intelRdt is specified in config, but intelRdt.l3CacheSchema is empty")
-		}
-		if intelrdt.IsMBAEnabled() && config.IntelRdt.MemBwSchema == "" {
-			return errors.New("Intel RDT/MBA is enabled and intelRdt is specified in config, but intelRdt.memBwSchema is empty")
-		}
```
kolyshkin, line 220: Took me some time to figure out why it was removed. It's a duplicate of the lines above!

Added by commit 27560ace2fcc0fdaa5af1 (PR #1632).

I'd put it into a separate commit, for the ease of review, but it's fine as it is.

kolyshkin:
A linter warning that needs to be fixed; otherwise LGTM.

libcontainer/configs/validate/validator.go
marquiz: They were not duplicate but need to drop those checks in order to specify only ClosID and leave these fields empty. Clarified this in the commit message

marquiz:
> A linter warning that needs to be fixed; otherwise LGTM.

Woops, fixed

kolyshkin (approved):
LGTM

marquiz:
Rebased

marquiz:
ping#2 @crosbymichael @mrunalp @dqminh @hqhq @cyphar @AkihiroSuda @kolyshkin

marquiz:
ping @crosbymichael @mrunalp @dqminh @hqhq @cyphar @AkihiroSuda @kolyshkin ❗️

This starts to be critical as RDT support was merged in CRI-O and we really need the runc to support it, too.
https://github.com/cri-o/cri-o/pull/4830

kolyshkin (approved):
Still LGTM :)

@Akihiro @thaJeztah PTAL

libcontainer/intelrdt/intelrdt.go
```diff
@@ -532,14 +518,18 @@ func IsMBAScEnabled() bool {
 }
 
 // Get the 'container_id' path in Intel RDT "resource control" filesystem
```
AkihiroSuda, line 520: Perhaps container_id -> ClosID ?

marquiz: Good catch, thanks. I addressed this in a separate patch, modifying some other code comments as well. I used the term *clos group*

AkihiroSuda (approved):

marquiz:
Signed-off-by was missing from the last patch 😭

AkihiroSuda (approved):

marquiz:
ping @kolyshkin

kolyshkin (approved):
LGTM

PR #2169, opened 2019-11-07
cgroup2: split fs2 from fs

AkihiroSuda:
split fs2 package from fs, as mixing up fs and fs2 is very likely to result in
unmaintainable code.

Inspired by containerd/cgroups#109

Fix #2157

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

---

This PR is large, but almost all changes are in `libcontainer/cgroups/fs2` and `libcontainer/cgroups/systemd/unified_hierarchy.go`.
Other changes are just for deduplicating files across `cgroups/fs` and `cgroups/fs2`.

PR #2921, opened 2021-04-27
Fix fsmanager.GetStats performance (regression in rc93)

kolyshkin:
Commit 88e8350de28 (PR #2169), among the other things, replaced filepath.Join with
securejoin.SecureJoin for both reads and writes to cgroupfs.
    
Commits e76ac1c054 and 31f0f5b7e03a (PR #2604) switched more code to use
`fscommon.ReadFile` (and thus `securejoin`). Commit 0228226e6d (PR #2635) introduced
`fscommon.OpenFile` (which uses `securejoin` as a fallback if `openat2(2)` is not available,
which is the case for older kernels), and commit c95e69007c4 (PR #2635) switched
most of cgroup/fs[2] code to use it.
    
As a result, `fs.GetStats()` method became noticeable slower, mostly due
to securejoin calling `os.Lstat` and `filepath.Clean`.
    
Using securejoin as a security measure for cgroupfs files is
not well justified, as cgroupfs do not contain symlinks, and none of the
code using it have uncleaned paths. In particular, fs/fs2/systemd
managers do check and sanitize their paths.

This PR modifies the code to not use securejoin for cgroupfs. Instead,
it checks that the opened file is indeed on cgroupfs.

Using a BenchmarkGetStats that this PR introduces, I see the following
improvement (on a CentOS 8 VM):
    
Before:
```
BenchmarkGetStats-8               8376            625135 ns/op
```    
After:
```
BenchmarkGetStats-8              12226            485015 ns/op
```    
An intermediate version, with no fstatfs to check fstype:
```
BenchmarkGetStats-8              13162            452281 ns/op
```

### DIFF
--- a/libcontainer/cgroups/fscommon/open.go
+++ b/libcontainer/cgroups/fscommon/open.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 	"sync"
 
-	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
@@ -17,7 +16,7 @@ const (
 )
 
 var (
-	// Set to true by fs unit tests
+	// TestMode is set to true by unit tests that need "fake" cgroupfs.
 	TestMode bool
 
 	cgroupFd     int = -1
@@ -72,11 +71,11 @@ func OpenFile(dir, file string, flags int) (*os.File, error) {
 		mode = 0o600
 	}
 	if prepareOpenat2() != nil {
-		return openWithSecureJoin(dir, file, flags, mode)
+		return openFallback(dir, file, flags, mode)
 	}
 	reldir := strings.TrimPrefix(dir, cgroupfsPrefix)
 	if len(reldir) == len(dir) { // non-standard path, old system?
-		return openWithSecureJoin(dir, file, flags, mode)
+		return openFallback(dir, file, flags, mode)
 	}
 
 	relname := reldir + "/" + file
@@ -93,11 +92,29 @@ func OpenFile(dir, file string, flags int) (*os.File, error) {
 	return os.NewFile(uintptr(fd), cgroupfsPrefix+relname), nil
 }
 
-func openWithSecureJoin(dir, file string, flags int, mode os.FileMode) (*os.File, error) {
-	path, err := securejoin.SecureJoin(dir, file)
+var errNotCgroupfs = errors.New("not a cgroup file")
+
+// openFallback is used when openat2(2) is not available. It checks the opened
+// file is on cgroupfs, returning an error otherwise.
+func openFallback(dir, file string, flags int, mode os.FileMode) (*os.File, error) {
+	path := dir + "/" + file
+	fd, err := os.OpenFile(path, flags, mode)
 	if err != nil {
 		return nil, err
 	}
+	if TestMode {
+		return fd, nil
+	}
+	// Check this is a cgroupfs file.
+	var st unix.Statfs_t
+	if err := unix.Fstatfs(int(fd.Fd()), &st); err != nil {
+		_ = fd.Close()
+		return nil, &os.PathError{Op: "statfs", Path: path, Err: err}
+	}
+	if st.Type != unix.CGROUP_SUPER_MAGIC && st.Type != unix.CGROUP2_SUPER_MAGIC {
+		_ = fd.Close()
+		return nil, &os.PathError{Op: "open", Path: path, Err: errNotCgroupfs}
+	}
 
-	return os.OpenFile(path, flags, mode)
+	return fd, nil
 }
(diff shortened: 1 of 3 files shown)


libcontainer/cgroups/fscommon/open.go
```diff
@@ -93,10 +93,18 @@ func OpenFile(dir, file string, flags int) (*os.File, error) {
...
-		return nil, err
+func openFallback(dir, file string, flags int, mode os.FileMode) (*os.File, error) {
+	var path string
+	// Do not use securejoin for files opened read-only,
+	// to speed up operations like GetStats.
+	if flags&unix.O_ACCMODE == unix.O_RDONLY {
+		path = dir + "/" + file
+	} else {
+		var err error
+		path, err = securejoin.SecureJoin(dir, file)
```
mrunalp, line 104: Are we certain that we need this for writes, too? Another option may be to check for the caller being root and taking the fast path for that, too.

kolyshkin: I thought about writes, too. Frankly, I don't know if we should use securejoin or not, just being cautious here as write is potentially more harmful.

Maybe @AkihiroSuda can chime in wrt why it was added in PR #2169. Cc @cyphar.

cyphar: There is a potential for attack along the lines of the `/proc` attacks we've seen in the past. Unfortunately there isn't a nice solution that doesn't involve a performance penalty -- even with libpathrs. However there are two counterpoints:

 1. We could make a slightly faster slow path for host paths -- just open the target path as `O_PATH` and then check the target (`statfs`, `readlink == intended path`, and so on). This is probably a more practical defense anyway.
 2. `SecureJoin` doesn't defend against the more powerful bind-mount attack, so arguably using it isn't protecting us as much as (1) would anyway.

But yes, it is important to note that just because something is a host path, it isn't necessarily safe because it's possible that we access it before `pivot_root(2)` but after the container has done their mounts (now, I'm not sure this actually happens for `/sys/fs/cgroup` but I'd rather be safe than sorry).

kolyshkin: @cyphar is your comment about further optimizing the slow (write) path? If yes, we can address it later, as currently I am more worried about the open-for-read performance -- we have frequent calls to `GetStats` calls from k8s/cadvisor, resulting in lots of open-for-read calls. Writes are relatively seldom compared to that -- they mostly come from `Set` which I guess is not called that frequently.

I was thinking about using `fstatfs(2)` right after open -- merely to check that we have just opened cgroupfs/cgroupfs2 file. Perhaps this check is enough, but I am not sure what's the overhead of added fstatfs (will measure it).

kolyshkin: > I was thinking about using `fstatfs(2)` right after open

I am not sure if this adds any security, but adding a fstatfs after open to check that we're still on cgroupfs do not affect performance much (seeing ~5% degradation in `BenchmarkGetStats`), so I have added this commit.

AFAIK neither v1 or v2 cgroupfs have symlinks so no room for /proc-like exploits.

Given this, maybe we should not use SecureJoin for the write case... 🤔

kolyshkin: Also, from what I see, we always audit/clean cgroup paths so we can't possibly have a path with `../` in it.

thaJeztah:
Looks like CI is failing @kolyshkin (haven't checked details)

libcontainer/cgroups/fscommon/open.go
kolyshkin: Updated the PR to
 - not have a distinction between read and write;
 - never use securejoin;
 - always check that the opened file is on cgroupfs.

kolyshkin:
>> Looks like CI is failing

The error message is quite funny :)
>    utils_test.go:36: open /tmp/cgroup_utils_test065049446/cgroup.file: not a cgroup file

I added a check that opened files are on cgroupfs, and forgot to enable TestMode flag in unit tests.

libcontainer/cgroups/fscommon/open.go
cyphar: > is your comment about further optimizing the slow (write) path? I

I meant to say that we should do what you ended up doing -- having a much less expensive check (`fstatfs`) which is probably more effective when `openat2` isn't available -- for both read and write. (I call non-`openat2` the slow path because `openat2` does exactly what we need here, and `SecureJoin` or `fstatfs` are both less than ideal in comparison.)

> AFAIK neither v1 or v2 cgroupfs have symlinks so no room for /proc-like exploits.

They don't have symlinks, but there *is* room for `/proc`-like exploits because you could have a mount configuration that messes around with `/sys/fs/cgroup` and thus results in runc not actually writing to the path you expect (there was a CVE like this in `/proc` but it wasn't `/proc`-specific -- it was just targeting the process labels).

*However* the only real solution for this is `openat2(RESOLVE_NO_XDEV)` which we use if it's available -- so there's probably not much else you can do.

cyphar (approved):
LGTM．

AkihiroSuda (approved):

PR #2203, opened 2020-01-11
Separate systemd dbus connection initialization from running check

mrunalp:
This allows us to catch and report any failures in initializing the systemd dbus connection.

PR #2923, opened 2021-04-27
libct/cg/sd: reconnect and retry on dbus connection error

kolyshkin:
~~PR #2203 started to reuse the same dbus connection. While this improves the performance,~~
In case the dbus daemon is ever restarted, the connection is no longer valid and every operation
fails. This is a minor concern for short-lived runc, but much more of a issue in case there is
a long-running daemon (e.g. `cri-o`) is using runc's libcontainer, as the connection is never
retried and the only remedy is to restart the daemon.

The solution to the above is to check the errors returned for `dbus: connection closed by user`
error, and try to re-connect on that. This is what PR #2862 does.

This is a carry of #2862, implementing the idea of retry-in-place (first described
at https://github.com/opencontainers/runc/pull/2862#discussion_r600977482 and https://github.com/opencontainers/runc/pull/2862#discussion_r600975567) on top of what it does.

For more info, see commit messages as well as #2862.

Fixes:
* https://github.com/kubernetes/kubernetes/issues/100328
* https://bugzilla.redhat.com/show_bug.cgi?id=1941456
* https://bugzilla.redhat.com/show_bug.cgi?id=1634092

### Changelog entry
 * libcontainer/cgroups/systemd: reconnect and retry in case dbus connection is closed (after dbus restart)

### DIFF
--- a/libcontainer/cgroups/systemd/dbus.go
+++ b/libcontainer/cgroups/systemd/dbus.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+	dbus "github.com/godbus/dbus/v5"
+	"github.com/sirupsen/logrus"
 )
 
 type dbusConnManager struct {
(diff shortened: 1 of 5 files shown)


kolyshkin:
@wzshiming PTAL

mrunalp (dismissed):

kolyshkin:
I was wrong in my initial analysis that this is a regression caused by PR #2203. In fact it's not, it has always been working that way (see e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1634092 from 2018).

mrunalp (approved):

libcontainer/cgroups/systemd/dbus.go
```diff
@@ -0,0 +1,90 @@
...
+func (d *dbusConnManager) getConnection() (*systemdDbus.Conn, error) {
+	// In the case where d.conn != nil
+	// Use the read lock the first time to ensure
+	// that Conn can be acquired at the same time.
+	d.RLock()
+	if conn := d.conn; conn != nil {
+		d.RUnlock()
+		return conn, nil
+	}
+	d.RUnlock()
```
cyphar, line 36: This could be

```go
d.RLock()
conn := d.conn
d.RUnlock()
if conn != nil { ... }
```

But it doesn't really matter.

cyphar (approved):
LGTM.

PR #2926, opened 2021-04-29
cgroups/systemd: replace deprecated dbus functions

thaJeztah:
These are deprecated, but pkg.go.dev does not make that very clear; https://pkg.go.dev/github.com/coreos/go-systemd/v22/dbus#Conn.StopUnitContext

It also does a really lame job at pointing out that you're reading the "latest version" of an "ancient" release

Screenshot 2021-04-29 at 22 05 26

relates to https://github.com/opencontainers/runc/pull/2923

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -2,6 +2,7 @@ package systemd
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"math"
 	"os"
@@ -321,7 +322,7 @@ func isUnitExists(err error) bool {
 func startUnit(cm *dbusConnManager, unitName string, properties []systemdDbus.Property) error {
 	statusChan := make(chan string, 1)
 	err := cm.retryOnDisconnect(func(c *systemdDbus.Conn) error {
-		_, err := c.StartTransientUnit(unitName, "replace", properties, statusChan)
+		_, err := c.StartTransientUnitContext(context.TODO(), unitName, "replace", properties, statusChan)
 		return err
 	})
 	if err == nil {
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/systemd/common.go
```diff
@@ -321,7 +322,7 @@ func isUnitExists(err error) bool {
 func startUnit(cm *dbusConnManager, unitName string, properties []systemdDbus.Property) error {
 	statusChan := make(chan string, 1)
 	err := cm.retryOnDisconnect(func(c *systemdDbus.Conn) error {
-		_, err := c.StartTransientUnit(unitName, "replace", properties, statusChan)
+		_, err := c.StartTransientUnitContext(context.TODO(), unitName, "replace", properties, statusChan)
```
thaJeztah, line 325: Possibly we can use a `context.WithTimeout()` to replace the handling below, but I didn't want to make larger changes, because the last changes were fixing a bug, so didn't want to jinx it.

/cc @kolyshkin

AkihiroSuda (approved):

cyphar (approved):
`context.TODO()` is a little bit :grimacing: but we'd need to change all of the callers to pass `context` if we wanted to change it, so we can do that some other time.

thaJeztah:
> context.TODO() is a little bit 😬 but we'd need to change all of the callers to pass context if we wanted to change it

Agreed; I picked `context.TODO()` for that reason; it should be an indicator that something needs to be done 😉 

Do you want me to open a tracking issue for that effort?

kolyshkin:
I was preparing a similar PR but dropped it, as the changes introduced basically do two things
* fix linter warnings (replacing them with TODO);
* make backports harder.

> Do you want me to open a tracking issue for that effort?

Yes please (I guess it's inevitable we add context to libcontainer's public API).

libcontainer/cgroups/systemd/common.go
kolyshkin: I was taking a look at it earlier this week, and I am not sure if the timeout can be replaced with a `context.WithTimeout()`. I read the code and don't quite understand it.

PR #2930, opened 2021-05-03
release: clean go cache

cyphar:
Turns out that Go caches pkg-config results fairly heavily, which
results in builds failing if you call "script/release.sh" multiple times
in a row.

### DIFF
--- a/script/release.sh
+++ b/script/release.sh
@@ -52,7 +52,7 @@ EOF
 	local libseccomp_ver='2.5.1'
 	local tarball="libseccomp-${libseccomp_ver}.tar.gz"
 	local prefix
-	prefix="$(mktemp -d)"
+	prefix="$(mktemp -dt runc-release-staticlibs.XXXXXX)"
 	wget "https://github.com/seccomp/libseccomp/releases/download/v${libseccomp_ver}/${tarball}"{,.asc}
 	tar xf "$tarball"
 	(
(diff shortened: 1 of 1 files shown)


cyphar:
@kolyshkin I expect this means we don't need the "no `libseccomp-static-devel`" restriction anymore? But I'm fine keeping it as-is.

script/release.sh
```diff
@@ -62,6 +62,10 @@ EOF
 	)
 	mv "$tarball"{,.asc} "$builddir"
 
+	# Go caches pkg-config pretty heavily (including from previous
+	# script/release.sh builds) so clean the cache before building.
+	go clean -cache
+
 	make -C "$root" PKG_CONFIG_PATH="${prefix}/lib/pkgconfig" COMMIT_NO= static
```
kolyshkin, line 69: It seems that adding `EXTRA_FLAGS="-a"` to `make` should solve the issue of pkg-config caching, and is more subtle/focused than cleaning the cache. Source: https://github.com/golang/go/issues/24544#issuecomment-410127446

kolyshkin:
> I expect this means we don't need the "no libseccomp-static-devel" restriction anymore?

You're right, it is useless now. PTAL https://github.com/opencontainers/runc/pull/2931

script/release.sh
kolyshkin: implemented in #2931

cyphar:
Sure, `-a` is fine too.

PR #2933, opened 2021-05-04
Makefile: allow overriding go command by environment

metux:
This is required for environments/build systems where a specific
go version / command needs to be used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 CONTAINER_ENGINE := docker
-GO := go
+GO ?= go
 
 PREFIX ?= /usr/local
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

thaJeztah (approved):
LGTM

cyphar (approved):
LGTM.

kolyshkin:
`make GO=$GO ...` worked before. Surely we can pick it up from the environment, too.

cyphar:
Yeah you're right that it worked before (both with `GO=... make` and `make GO=...`) but it's a no-op change so it doesn't really matter.

PR #2934, opened 2021-05-04
Makefile: install recvtty and make install depend on build

metux:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,9 @@ shell: runcimage
 		-v $(CURDIR):/go/src/$(PROJECT) \
 		$(RUNC_IMAGE) bash
 
-install:
+install: runc recvtty
 	install -D -m0755 runc $(DESTDIR)$(BINDIR)/runc
+	install -D -m0755 contrib/cmd/recvtty/recvtty $(DESTDIR)$(BINDIR)/recvtty
 
 install-bash:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -102,8 +102,9 @@ shell: runcimage
 		-v $(CURDIR):/go/src/$(PROJECT) \
 		$(RUNC_IMAGE) bash
 
-install:
+install: runc recvtty
 	install -D -m0755 runc $(DESTDIR)$(BINDIR)/runc
+	install -D -m0755 contrib/cmd/recvtty/recvtty $(DESTDIR)$(BINDIR)/recvtty
```
kolyshkin, line 107: Not sure if we should install `recvtty`. It is not meant to be a standalone utility, but rather an example of how to write something like that.

cyphar:
NACK, `recvtty` is not a real utility people should use, it's a PoC so people can implement `--console-socket` handling in their own container managers.

metux:
So, higher layers (eg. moby) don't use it all ? I've just seen that the precompiled docker-ce deb's ship that file, so I added it.

cyphar:
No none of the higher layers use it at all (Docker/containerd use containerd-shim, and cri-o/podman use conmon), I'm a bit surprised that docker-ce would ship it.

metux:
> No none of the higher layers use it at all (Docker/containerd use containerd-shim, and cri-o/podman use conmon), I'm a bit surprised that docker-ce would ship it.

At least the one I've found in Debian.

But if it's not neded at all, I probably won't need to package it.

PR #2935, opened 2021-05-04
[WIP] libct/int: don't skip cgroup mount

kolyshkin:
This is not necessary since commit ff692f289b60e1 (PR #2818).


kolyshkin:
Alas, this makes the container non-checkpointable.
```
    checkpoint_test.go:28: (00.026747) Error (criu/mount.c:938): mnt: Mount 427 ./sys/fs/cgroup (master_id: 4 shared_id: 0) has unreachable sharing. Try --enable-external-masters.
```

Issue #2907, opened 2021-04-14
[centos7] Test{,Systemd}Freeze FAIL: unexpected error: unable to freeze


kolyshkin:
Hmm, we are getting this a lot lately :-\

Last example is from https://github.com/opencontainers/runc/pull/2868/checks?check_run_id=2299195550

```
=== RUN   TestSystemdFreeze
    utils_test.go:86: exec_test.go:539: unexpected error: unable to freeze
        
--- FAIL: TestSystemdFreeze (1.17s)
```

kolyshkin:
A few more past failures
* 8 days old: https://github.com/opencontainers/runc/runs/2272440180?check_suite_focus=true
* 11 days old: https://github.com/opencontainers/runc/runs/2254259869?check_suite_focus=true
* 12 days old: https://github.com/opencontainers/runc/runs/2252814238?check_suite_focus=true

kolyshkin:
* 14 days old: https://github.com/opencontainers/runc/runs/2230249911?check_suite_focus=true
* 14 days old **this one is without systemd**: https://github.com/opencontainers/runc/runs/2228626065?check_suite_focus=true
* Mar 9: https://github.com/opencontainers/runc/runs/2073456610?check_suite_focus=true
* Mar 4: https://github.com/opencontainers/runc/runs/2034186723?check_suite_focus=true

this ^^^ seems to be the first one; no failures found before that (checked up to Jan 26).

kolyshkin:
Might be a regression caused by either https://github.com/opencontainers/runc/pull/2774 (merged Feb 1) https://github.com/opencontainers/runc/pull/2791 (merged Feb 7). Not sure what to do with it though...

kolyshkin:
Another one from May 2: https://github.com/opencontainers/runc/pull/2929/checks?check_run_id=2488750446
```
=== RUN   TestFreeze
    utils_test.go:86: exec_test.go:539: unexpected error: unable to freeze
        
--- FAIL: TestFreeze (0.62s)
```

PR #2936, opened 2021-05-04
libct/cg/sd: fix dbus connection leak

kolyshkin:
_TL;DR: fixes a regression (open fd leak) caused by PR #2923. This is an alternative to #2937._

When running libcontainer/integration/TestSystemdFreeze 2000 times,
I got "too many open files" error after a while:

```console
[vagrant@localhost integration]$ sudo -E PATH=$PATH ./integration.test -test.v -test.run Freeze -test.count 2000
<...>
=== RUN   TestFreeze
--- PASS: TestFreeze (0.28s)
=== RUN   TestSystemdFreeze
--- PASS: TestSystemdFreeze (0.42s)
=== RUN   TestFreeze
    utils_test.go:86: exec_test.go:536: unexpected error: container_linux.go:380: starting container process caused: process_linux.go:338: starting init process command caused: fork/exec /proc/self/exe: too many open files
        
--- FAIL: TestFreeze (0.16s)
=== RUN   TestSystemdFreeze
    utils_test.go:86: exec_test.go:536: unexpected error: container_linux.go:380: starting container process caused: process_linux.go:338: starting init process command caused: fork/exec /proc/self/exe: too many open files

<...>

--- FAIL: TestFreeze (0.00s)
=== RUN   TestSystemdFreeze
    utils_test.go:86: exec_test.go:536: unexpected error: container_linux.go:380: starting container process caused: process_linux.go:338: starting init process command caused: open /dev/null: too many open files
        
--- FAIL: TestSystemdFreeze (0.16s)
=== RUN   TestFreeze
    utils_test.go:86: exec_test.go:512: unexpected error: untar error "fork/exec /bin/sh: too many open files": ""
        
--- FAIL: TestFreeze (0.00s)
=== RUN   TestSystemdFreeze
    utils_test.go:86: exec_test.go:536: unexpected error: container_linux.go:364: creating new parent process caused: container_linux.go:496: including execfifo in cmd.Exec setup caused: open /tmp/libcontainer548959440/test/exec.fifo: too many open files
       
 --- FAIL: TestSystemdFreeze (0.17s)
=== RUN   TestFreeze
    utils_test.go:86: exec_test.go:512: unexpected error: untar error "open /dev/null: too many open files": ""
        
<...>
```

Indeed, systemd cgroup manager never closes its dbus connection.
    
This was not a problem before PR #2923 because we only had a single connection
for the whole runtime. Now it is per manager instance, so we leak a
connection (apparently two sockets) per cgroup manager instance.
One possible way to fix this is to go back to using a single global dbus connection. That is implemented by https://github.com/opencontainers/runc/pull/2937. The alternative, that this PR implements, is to 
    
1. Close the connection from cgroup manager's `Destroy` (assuming the
       cgroup manager won't be used after `Destroy` -- even in case it will be,
       a new connection will be initiated). This helps for cases when
       the container (and its cgroup) are actually removed.
    
2. Set a finalizer in `newDbusConnManager()`, so once it is garbage
       collected, the `closeConnection()` method is called. This helps for
       cases of a long-running runtime (such as kubernetes) which uses
       a libcontainer/cgroup/systemd manager to create (but not remove)
       a cgroup.
    
A test case is added in a separate commit. Before the fix, it always fails
like this:
```
        exec_test.go:2030: extra fd 8 -> socket:[659703]
        exec_test.go:2030: extra fd 11 -> socket:[658715]
        exec_test.go:2033: found 2 extra fds after container.Run
    --- FAIL: TestFdLeaksSystemd (0.20s)

```

### DIFF
--- a/libcontainer/cgroups/systemd/dbus.go
+++ b/libcontainer/cgroups/systemd/dbus.go
@@ -4,6 +4,7 @@ package systemd
 
 import (
 	"context"
+	"runtime"
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
@@ -18,9 +19,15 @@ type dbusConnManager struct {
 
 // newDbusConnManager initializes systemd dbus connection manager.
 func newDbusConnManager(rootless bool) *dbusConnManager {
-	return &dbusConnManager{
+	m := &dbusConnManager{
 		rootless: rootless,
 	}
+
+	runtime.SetFinalizer(m, func(d *dbusConnManager) {
+		d.closeConnection()
+	})
+
+	return m
 }
 
 // getConnection lazily initializes and returns systemd dbus connection.
@@ -57,6 +64,15 @@ func (d *dbusConnManager) newConnection() (*systemdDbus.Conn, error) {
 		return newUserSystemdDbus()
 	}
 	return systemdDbus.NewWithContext(context.TODO())
+}
+
+func (d *dbusConnManager) closeConnection() {
+	d.Lock()
+	defer d.Unlock()
+	if d.conn != nil {
+		d.conn.Close()
+		d.conn = nil
+	}
 }
 
 // resetConnection resets the connection to its initial state
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -337,16 +337,8 @@ func (m *unifiedManager) getSliceFull() (string, error) {
 	}
 
 	if m.rootless {
-		dbusConnection, err := m.dbus.getConnection()
-		if err != nil {
-			return "", err
-		}
-		// managerCGQuoted is typically "/user.slice/user-${uid}.slice/user@${uid}.service" including the quote symbols
-		managerCGQuoted, err := dbusConnection.GetManagerProperty("ControlGroup")
-		if err != nil {
-			return "", err
-		}
-		managerCG, err := strconv.Unquote(managerCGQuoted)
+		// managerCG is typically "/user.slice/user-${uid}.slice/user@${uid}.service".
+		managerCG, err := getManagerProperty(m.dbus, "ControlGroup")
 		if err != nil {
 			return "", err
 		}
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -302,6 +302,7 @@ func (m *unifiedManager) Destroy() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	defer m.dbus.closeConnection()
 	unitName := getUnitName(m.cgroups)
 	if err := stopUnit(m.dbus, unitName); err != nil {
 		return err
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -384,15 +384,23 @@ func setUnitProperties(cm *dbusConnManager, name string, properties ...systemdDb
 	})
 }
 
+func getManagerProperty(cm *dbusConnManager, name string) (string, error) {
+	str := ""
+	err := cm.retryOnDisconnect(func(c *systemdDbus.Conn) error {
+		var err error
+		str, err = c.GetManagerProperty(name)
+		return err
+	})
+	if err != nil {
+		return "", err
+	}
+	return strconv.Unquote(str)
+}
+
 func systemdVersion(cm *dbusConnManager) int {
 	versionOnce.Do(func() {
 		version = -1
-		var verStr string
-		err := cm.retryOnDisconnect(func(c *systemdDbus.Conn) error {
-			var err error
-			verStr, err = c.GetManagerProperty("Version")
-			return err
-		})
+		verStr, err := getManagerProperty(cm, "Version")
 		if err == nil {
 			version, err = systemdVersionAtoi(verStr)
 		}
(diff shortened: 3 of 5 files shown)


kolyshkin:
The alternative is to revert back to a single per-runtime connection, rather than per-cgroup manager. Let me work on that.

libcontainer/cgroups/systemd/dbus.go
```diff
@@ -18,9 +19,15 @@ type dbusConnManager struct {
...
 // newDbusConnManager initializes systemd dbus connection manager.
 func newDbusConnManager(rootless bool) *dbusConnManager {
-	return &dbusConnManager{
+	m := &dbusConnManager{
 		rootless: rootless,
 	}
+
+	runtime.SetFinalizer(m, func(d *dbusConnManager) {
+		d.closeConnection()
+	})
```
cyphar, line 28: I think this is a better solution than switching back to a single-connection setup.

cyphar (approved):
LGTM.

libcontainer/cgroups/systemd/dbus.go
kolyshkin: The only issue with this PR I can think of is, if a runtime creates a lot of containers and keeps references to all of them (and thus to cgroup manager for every container), there will be N*2 opened sockets rather than just 2.

The issue with the other PR is it mixes root and rootless connection into the same variable.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -337,12 +338,12 @@ func (m *unifiedManager) getSliceFull() (string, error) {
 	}
 
 	if m.rootless {
-		dbusConnection, err := m.dbus.getConnection()
-		if err != nil {
-			return "", err
-		}
-		// managerCGQuoted is typically "/user.slice/user-${uid}.slice/user@${uid}.service" including the quote symbols
```
AkihiroSuda, line 344: Can we retain this comment

kolyshkin: right; fixed

kolyshkin:
Rebased (minor conflicts in libct/int), addressed a review nit.

libcontainer/cgroups/systemd/v2.go
AkihiroSuda: Doesn’t seem fixed

kolyshkin:
CI failure is a known flake (#2907) which is appearing a lot lately.

libcontainer/cgroups/systemd/v2.go
kolyshkin: `ENOTENOUGHCOFFEE`

Let me try again :^)

kolyshkin: Yay, it worked this time!

libcontainer/cgroups/systemd/dbus.go
mrunalp: kubelet could be operating on up to ~200 pods (typical per node limit) on a single node so there is the overhead of having to create/destroy connections to dbus frequently. We'll have to measure the impact of that.

AkihiroSuda (dismissed):

libcontainer/cgroups/systemd/dbus.go
kolyshkin: Takes about 1 ms to connect on my laptop.

Here is a a benchmark:
```go
package systemd

import (
	"os"
	"testing"

	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
)

func benchmarkDbusConn(b *testing.B, closeConn bool) {
	cm := newDbusConnManager(os.Getuid() != 0)

	var c *systemdDbus.Conn
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		var err error
		if closeConn {
			cm.closeConnection()
		}
		c, err = cm.getConnection()
		if err != nil {
			b.Fatal(err)
		}
	}
	b.Log(c.GetManagerProperty("Version"))
	cm.closeConnection()
}

func BenchmarkDbusConnReuse(b *testing.B) {
	benchmarkDbusConn(b, false)
}

func BenchmarkDbusConnReopen(b *testing.B) {
	benchmarkDbusConn(b, true)
}
```

and the results:
```console
kir@kir-rhat systemd]$ go test -v -run 345 -bench .
goos: linux
goarch: amd64
pkg: github.com/opencontainers/runc/libcontainer/cgroups/systemd
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
BenchmarkDbusConnReuse
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
BenchmarkDbusConnReuse-4    	78536199	        13.88 ns/op
BenchmarkDbusConnReopen
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
    dbus_test.go:25: "v246.10-1.fc33" <nil>
BenchmarkDbusConnReopen-4   	    1520	    804216 ns/op
PASS
ok  	github.com/opencontainers/runc/libcontainer/cgroups/systemd	3.421s
```

kolyshkin: Surely, there's also some other overhead, most notably opened fds.

AkihiroSuda:
Merged #2937

PR #2939, opened 2021-05-05
[DNM] Debug freeze on CentOS 7 CI

kolyshkin:
GHA CI almost always fails on CentOS 7 (https://github.com/opencontainers/runc/issues/2907):

```
=== RUN   TestFreeze
    utils_test.go:85: exec_test.go:539: unexpected error: unable to freeze
        
--- FAIL: TestFreeze (0.71s)
```

Trying to find out what to do about it.

This is complicated because the kind of mac os x host GHA gives for the test is a lottery. In most cases it's good, and sometimes it's slow and buggy.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,114 +14,11 @@ on:
   pull_request:
 
 jobs:
-  test:
-    runs-on: ubuntu-20.04
-    strategy:
-      fail-fast: false
-      matrix:
-        # Dockre/Moby still builds runc with Go 1.13, so we should still support Go 1.13.
-        go-version: [1.13.x, 1.15.x, 1.16.x]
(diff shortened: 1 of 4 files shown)


kolyshkin:
```
--- PASS: TestAdditionalGroups (0.30s)
=== RUN   TestFreeze
    utils_test.go:85: exec_test.go:539: unexpected error: unable to freeze (1000 retries, 20 thaws, last state: FREEZING)
        
--- FAIL: TestFreeze (0.63s)
```

Not sure what to do here. Add more iterations? Increase the timeout?

kolyshkin:
OK, the conclusion is adding an occasional short delay between writing "frozen" and reading the status back helps for this case (very slow system).

kolyshkin:
The fix is #2941

PR #2941, opened 2021-05-06
freezer: add delay after freeze

kolyshkin:
I hate to keep adding those kludges (for earlier ones, see #2918, #2791, #2774)
but lately TestFreeze (and TestSystemdFreeze) from libcontainer/integration
fails a lot (see #2907).

The failure comes and goes, and is probably this is caused by a slow host
allocated for the test, and a slow VM on top of it.

To remediate, add a small sleep on every 25th iteration in between
asking the kernel to freeze and checking its status.

In the worst case scenario (failure to freeze), this adds about 0.4 ms
(40 x 10 us) to the duration of the call.

It is hard to measure how this affects CI as GHA plays a roulette when
allocating a node to run the test on, but it seems to help. With
additional debug info, I saw somewhat frequent "frozen after 24 retries"
or "frozen after 49 retries", meaning it succeeded right after the added
sleep.

While at it, rewrite/improve the comments.

Fixes: #2907.

### DIFF
--- a/libcontainer/cgroups/fs/freezer.go
+++ b/libcontainer/cgroups/fs/freezer.go
@@ -43,20 +43,31 @@ func (s *FreezerGroup) Set(path string, r *configs.Resources) (Err error) {
 		// kernel commit ef9fe980c6fcc1821), if FREEZING is seen,
 		// userspace should either retry or thaw. While current
 		// kernel cgroup v1 docs no longer mention a need to retry,
-		// the kernel (tested on v5.4, Ubuntu 20.04) can't reliably
-		// freeze a cgroup while new processes keep appearing in it
+		// even a recent kernel (v5.4, Ubuntu 20.04) can't reliably
+		// freeze a cgroup v1 while new processes keep appearing in it
 		// (either via fork/clone or by writing new PIDs to
 		// cgroup.procs).
 		//
-		// The numbers below are chosen to have a decent chance to
-		// succeed even in the worst case scenario (runc pause/unpause
-		// with parallel runc exec).
+		// The numbers below are empirically chosen to have a decent
+		// chance to succeed in various scenarios ("runc pause/unpause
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/fs/freezer.go
```diff
@@ -65,6 +65,10 @@ func (s *FreezerGroup) Set(path string, r *configs.Resources) (Err error) {
 				return err
 			}
 
+			if i%25 == 24 {
+				// A short sleep before reading back also helps.
```
AkihiroSuda, line 69: Could you update the comment to clarify this helps what

kolyshkin:
This is from a "good" VM (once you start testing, gha gives you a good VM...).

Out of 400 runs, only 147 needed more than 1 retry.

Out of those that need a retry, there are peaks at 24 and 49, meaning sleep helps.

```
[kir@kir-rhat runc]$ grep frozen xx | awk '{print $5}' | sort -n | uniq -c
      8 2
      7 3
      4 4
      2 5
      1 6
      2 7
      2 8
      1 10
      1 13
      1 14
      1 19
      4 22
      1 23
     85 24
      1 25
      1 29
      1 30
      1 31
      1 36
      1 39
      1 41
      1 45
     15 49
      1 50
      1 52
      1 54
      1 123
```

libcontainer/cgroups/fs/freezer.go
kolyshkin: There's a big comment above telling the whole story...

Ah, I just found it now contradicts what I say here. Will fix.

dqminh:
LGTM

libcontainer/cgroups/fs/freezer.go
kolyshkin: Comments rewritten; PTAL @AkihiroSuda

mrunalp (approved):

AkihiroSuda (approved):

Issue #2942, opened 2021-05-06
CI seems skipping `cgroupv2 with rootless (systemd)`


AkihiroSuda:
![image](https://user-images.githubusercontent.com/9248427/117261107-bb897600-ae8a-11eb-8093-52b1b5312fa8.png)
https://github.com/opencontainers/runc/runs/2512206353
https://github.com/opencontainers/runc/blob/2d38476c96e3407eba876e77b301880c6d7e5ccc/tests/integration/helpers.bash#L374

Most tests in `cgroupv2 with rootless (systemd)` seem skipped.

This is not a recent regression. The log from February has the same `skip` output. https://github.com/opencontainers/runc/runs/1867231647?check_suite_focus=true

AkihiroSuda:
cc @kolyshkin

kolyshkin:
Good catch, working on a fix

kolyshkin:
Hmm, we're not in a very bad shape, the previously skipped tests only found 1 bug. PTAL https://github.com/opencontainers/runc/pull/2944.

AkihiroSuda:
Thanks for fixing! 👍

PR #2944, opened 2021-05-07
Enable rootless cgroup v2 tests + some fixes

kolyshkin:
_TL;DR: this is mostly tests/integration simplifications and improvements for systemd + cgroup v2 + rootless case, addressing #2942. The only exception is the second commit, which fixes the real bug uncovered by tests that are no longer skipped thanks to this PR._

1. ### tests/int/helpers: rm old code
    
    This was added by commit ca4f427af, together with `set_resources_limit`,
    and was needed for the latter, since it used `sed` to update resources.
    
    Now when we switched from `sed` to `jq` in commit 79fe41d3c16b, this kludge
    is no longer needed.

2. ### libct/cg/sd/v2: call initPath from Path
    
    Sometimes `Path()` is called before m.path is initialized (in particular,
    this happens from `(*linuxContainer).newInitConfig)`, so we do need to
    make sure to call `initPath`.
    
    This fixes the following integration tests (for cgroup v2 + systemd case,
    currently not enabled -- to be enabled by further commits):
    
     * `runc run (blkio weight)`
     * `runc run (cgroupv2 mount inside container)`
    
    Fixes: ff692f289b6 ("Fix cgroup2 mount for rootless case") aka PR #2818.

3. ### tests/int/update.bats: don't set cpuset in setup
    
    Setting `cpuset.cpus` requires cpuset controller, which might
    not be available in case of cgroup v2 + systemd < 244.
    
    Rather than require cpuset support from every test case, do not set the initial
    cpuset value from `setup()`, and do not check it.

4. ### tests/int/helpers: generalize require cgroups_freezer
    
    With this, something like cgroups_pids can be used (to be added
    by the next commit).

5. ### tests/int: enable/use `requires cgroups_<ctrl>`
    
    1. In case of cgroup v2 + systemd + rootless, get the list of controllers from
       the own cgroup, rather than from the root one (as systemd cgroup delegation
       is required in this case, and it might not be set or fully working).
    
    2. Use `requires cgroups_<controller>` in tests that need those
       controllers.
    
    Tested on Fedora 31 (which does not support delegation of cpuset controller).

6. ### tests/int/cgroups: don't check for hugetlb
    
    Systemd is not able to delegate hugetlb controller, and it is needed for
    cgroup v2 + systemd + rootless case (which is currently skipped because
    of "requires rootless_cgroup", but will be enabled by a later commit).
    
    The failure being fixed looks like this:
    
    > not ok 4 runc create (limits + cgrouppath + permission on the cgroup dir) succeeds
    >  (from function `check_cgroup_value' in file /vagrant/tests/integration/helpers.bash, line 188,
    >   in test file /vagrant/tests/integration/cgroups.bats, line 53)
    >    `check_cgroup_value "cgroup.controllers" "$(cat /sys/fs/cgroup/user.slice/user-$(id -u).slice/cgroup.controllers)"' failed
    >  <....>
    >  /sys/fs/cgroup/user.slice/user-2000.slice/user@2000.service/machine.slice/runc-cgroups-integration-test-20341.scope/cgroup.controllers
    >  current cpuset cpu io memory pids !? cpuset cpu io memory hugetlb pids

7. ### tests/int: run rootless_cgroup tests for v2+systemd
    
    Before this commit, `require rootless_cgroup` feature check required `cgroup`
    to be present in `$ROOTLESS_FEATURES`. The idea of the requirement,
    though, is to _ensure that rootless runc can manage cgroups_.
    
    In case of systemd + cgroup v2, rootless runc can manage cgroups,
    thanks to systemd delegation, so modify the feature check accordingly.
    
    Next, convert (simplify) some of the existing users to the modified check,
    essentially enabling some tests that were disabled for systemd + cgroup v2 +
    rootless case before.

    Fixes: #2942. 

8. ### Vagrantfile.fedora: set Delegate=yes
    
    Instead of listing all individual controllers we want to be delegated,
    just say "yes" which means "all the controllers that are available".

### DIFF
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -103,13 +103,21 @@ function init_cgroup_paths() {
 
 	if stat -f -c %t /sys/fs/cgroup | grep -qFw 63677270; then
 		CGROUP_UNIFIED=yes
+		local controllers="/sys/fs/cgroup/cgroup.controllers"
+		# For rootless + systemd case, controllers delegation is required,
+		# so check the controllers that the current user has, not the top one.
+		# NOTE: delegation of cpuset requires systemd >= 244 (Fedora >= 32, Ubuntu >= 20.04).
+		if [[ "$ROOTLESS" -ne 0 && -n "$RUNC_USE_SYSTEMD" ]]; then
+			controllers="/sys/fs/cgroup/user.slice/user-$(id -u).slice/user@$(id -u).service/cgroup.controllers"
+		fi
+
 		# "pseudo" controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
 		# - devices (since kernel 4.15) we must assume to be supported because
 		#   it's quite hard to test.
 		# - freezer (since kernel 5.2) we can auto-detect by looking for the
 		#   "cgroup.freeze" file a *non-root* cgroup.
 		CGROUP_SUBSYSTEMS=$(
-			cat /sys/fs/cgroup/cgroup.controllers
+			cat "$controllers"
 			echo devices
 		)
 		CGROUP_BASE_PATH=/sys/fs/cgroup
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -304,12 +304,6 @@ function requires() {
 				skip_me=1
 			fi
 			;;
-		cgroups_freezer)
-			init_cgroup_paths
-			if [[ "$CGROUP_SUBSYSTEMS" != *"freezer"* ]]; then
-				skip_me=1
-			fi
-			;;
 		cgroups_rt)
 			init_cgroup_paths
 			if [ ! -e "${CGROUP_CPU_BASE_PATH}/cpu.rt_period_us" ]; then
@@ -336,6 +330,13 @@ function requires() {
 		cgroups_v2)
 			init_cgroup_paths
 			if [ "$CGROUP_UNIFIED" != "yes" ]; then
+				skip_me=1
+			fi
+			;;
+		cgroups_*)
+			init_cgroup_paths
+			var=${var#cgroups_}
+			if [[ "$CGROUP_SUBSYSTEMS" != *"$var"* ]]; then
 				skip_me=1
 			fi
 			;;
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -68,11 +68,6 @@ function runc_spec() {
 	if [[ "$ROOTLESS" -ne 0 ]] && [[ "$ROOTLESS_FEATURES" == *"idmap"* ]]; then
 		runc_rootless_idmap "$bundle"
 	fi
-
-	# Ensure config.json contains linux.resources
-	if [[ "$ROOTLESS" -ne 0 ]] && [[ "$ROOTLESS_FEATURES" == *"cgroup"* ]]; then
-		runc_rootless_cgroup "$bundle"
-	fi
 }
 
 # Helper function to reformat config.json file. Input uses jq syntax.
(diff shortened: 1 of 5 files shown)


kolyshkin:
Failure in Fedora is caused by the fact systemd do not support delegation of `hugetlb` controller.

```
not ok 4 runc create (limits + cgrouppath + permission on the cgroup dir) succeeds
# (from function `check_cgroup_value' in file /vagrant/tests/integration/helpers.bash, line 188,
#  in test file /vagrant/tests/integration/cgroups.bats, line 53)
#   `check_cgroup_value "cgroup.controllers" "$(cat /sys/fs/cgroup/user.slice/user-$(id -u).slice/cgroup.controllers)"' failed
# 
# runc run -d --console-socket /tmp/bats-run-49422/runc.Jkng1u/tty/sock test_cgroups_permissions (status=0):
# 
# /sys/fs/cgroup/user.slice/user-2000.slice/user@2000.service/machine.slice/runc-cgroups-integration-test-20341.scope/cgroup.controllers
# current cpuset cpu io memory pids !? cpuset cpu io memory hugetlb pids
```

Will probably disable the check.

kolyshkin:
CentOS failure is a known flake (https://github.com/opencontainers/runc/issues/2760). I have a half-ready fix but no time to finish it :(

tests/integration/helpers.bash
```diff
@@ -103,13 +103,21 @@ function init_cgroup_paths() {
 
 	if stat -f -c %t /sys/fs/cgroup | grep -qFw 63677270; then
 		CGROUP_UNIFIED=yes
+		local controllers="/sys/fs/cgroup/cgroup.controllers"
+		# For rootless + systemd case, controllers delegation is required,
+		# so check the controllers that the current user has, not the top one.
+		# NOTE: delegation of cpuset requires systemd >= 244 (Fedora >= 32, Ubuntu >= 20.04).
+		if [[ "$ROOTLESS" -ne 0 && -n "$RUNC_USE_SYSTEMD" ]]; then
+			controllers="/sys/fs/cgroup/user.slice/user-$(id -u).slice/user@$(id -u).service/cgroup.controllers"
```
AkihiroSuda, line 111: Better to check cgroup.subtree_control?

kolyshkin: Theoretically, if systemd did the job of delegation right, it does not matter.

Fixed nevertheless.

mrunalp (dismissed):

AkihiroSuda:
We have Fedora 34, but cpuset tests seems skipped?

kolyshkin:
> We have Fedora 34, but cpuset tests seems skipped?

That was a typo in `tests/int: enable/use requires cgroups_<ctrl>` commit 🥵

Thanks for catching this, should be fixed now.

kolyshkin:
CentOS 7 failure:

```
ok 100 runc exec [tty consolesize]
ok 101 runc create [terminal=false]
not ok 102 runc run [terminal=false]
# (from function `retry' in file tests/integration/helpers.bash, line 399,
#  from function `wait_for_container' in file tests/integration/helpers.bash, line 407,
#  in test file tests/integration/tty.bats, line 238)
#   `wait_for_container 15 1 test_busybox running' failed
# runc spec (status=0):
# 
# time="2021-05-07T19:51:27Z" level=warning msg="unable to get oom kill count" error="no directory specified for memory.oom_control"
# time="2021-05-07T19:51:27Z" level=error msg="container_linux.go:380: starting container process caused: process_linux.go:385: applying cgroup configuration for process caused: Unit runc-test_busybox.scope already exists."
# Command "eval __runc state test_busybox | grep -qw running" failed 15 times. Output: time="2021-05-07T19:51:51Z" level=error msg="container \"test_busybox\" does not exist"
```

This is probably caused by the fact that some earlier container was not removed, which in turn is probably caused by a very slow VM, a small (1 second) timeout in `stopUnit`, and the fact that it does not return an error if the timeout is hit. This was fixed before for `startUnit` (commit 38447895a54daf52e, PR #2614) but not for `stopUnit`.

In any case look unrelated to the PR. I'll look into it.

CI restarted.

kolyshkin:
> In any case look unrelated to the PR. I'll look into it.

Proposed fix (unrelated to this PR): https://github.com/opencontainers/runc/pull/2946

kolyshkin:
Weird, I still see this in fedora + rootless + systemd:

> ok 105 update cgroup v1/v2 common limits # skip test requires cpuset

Looking.

kolyshkin:
No longer a draft. The last issue that was fixed is described at https://github.com/opencontainers/runc/pull/2944#discussion_r628705203

AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #2946, opened 2021-05-07
libct/cg/sd: return error from stopUnit

kolyshkin:
_Fixes the source of the error seen at https://github.com/opencontainers/runc/pull/2944#issuecomment-834819690_

Historically, we never returned an error from failed `startUnit`
or `stopUnit`. The `startUnit` case was fixed by commit 3844789 (PR #2614).

It is time to fix `stopUnit`, too. The reasons are:

1. Ignoring an error from `stopUnit` means an unexpected trouble down the
   road, for example a failure to create a container with the same name:

   > time="2021-05-07T19:51:27Z" level=error msg="container_linux.go:380: starting container process caused: process_linux.go:385: applying cgroup configuration for process caused: Unit runc-test_busybox.scope already exists."

2. A somewhat short timeout of 1 second means the cgroup might
   actually be removed a few seconds later but we might have a
   race between removing the cgroup and creating another one
   with the same name, resulting in the same error as above.

So, return an error if removal failed, and increase the timeout.

Now, modify the systemd cgroup v1 manager to not mask the error from
`stopUnit` (`stopErr`) with the subsequent one from `cgroups.RemovePath`,
as `stopErr` is most probably the reason why `RemovePath` failed.

Note that for v1 we do want to remove the paths even in case
of a failure from `stopUnit`, as some were not created by systemd.
There's no need to do that for v2, thanks to unified hierarchy,
so no changes there.

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -355,6 +355,9 @@ func stopUnit(cm *dbusConnManager, unitName string) error {
 		return err
 	})
 	if err == nil {
+		timeout := time.NewTimer(30 * time.Second)
+		defer timeout.Stop()
+
 		select {
 		case s := <-statusChan:
 			close(statusChan)
@@ -362,8 +365,8 @@ func stopUnit(cm *dbusConnManager, unitName string) error {
 			if s != "done" {
 				logrus.Warnf("error removing unit `%s`: got `%s`. Continuing...", unitName, s)
 			}
-		case <-time.After(time.Second):
-			logrus.Warnf("Timed out while waiting for StopUnit(%s) completion signal from dbus. Continuing...", unitName)
+		case <-timeout.C:
+			return errors.New("Timed out while waiting for systemd to remove " + unitName)
 		}
 	}
 	return nil
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -207,9 +207,10 @@ func (m *legacyManager) Destroy() error {
 
 	stopErr := stopUnit(m.dbus, getUnitName(m.cgroups))
 
-	// Both on success and on error, cleanup all the cgroups we are aware of.
-	// Some of them were created directly by Apply() and are not managed by systemd.
-	if err := cgroups.RemovePaths(m.paths); err != nil {
+	// Both on success and on error, cleanup all the cgroups
+	// we are aware of, as some of them were created directly
+	// by Apply() and are not managed by systemd.
+	if err := cgroups.RemovePaths(m.paths); err != nil && stopErr == nil {
 		return err
 	}
 


kolyshkin:
Failure on Fedora:

```
not ok 80 runc start
# (in test file tests/integration/start.bats, line 15)
# runc spec (status=0):
# 
# runc create --console-socket /tmp/bats-run-23697/runc.8iGh0j/tty/sock test_busybox (status=1):
# time="2021-05-07T23:29:00Z" level=error msg="dial unix /tmp/bats-run-23697/runc.8iGh0j/tty/sock: connect: no such file or directory"
```

I have no explanation for this (other than recvtty suddenly died?)

kolyshkin:
close/reopen to kick CI

kolyshkin:
Tentatively added this to 1.0 milestone as I think it makes sense to have it (as said earlier, ignoring an error means more trouble later).

libcontainer/cgroups/systemd/common.go
```diff
@@ -355,15 +355,18 @@ func stopUnit(cm *dbusConnManager, unitName string) error {
 		return err
 	})
 	if err == nil {
+		timeout := time.NewTimer(30 * time.Second)
```
cyphar, line 358: Why did you switch away from `<-time.After(...)`? `<-time.After(30*time.Second)` is more idiomatic IMHO.

cyphar: Ah never mind, it's because you would need to wait for 30 seconds for the GC to free it. Fair point.

cyphar (approved):
LGTM.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -207,9 +207,10 @@ func (m *legacyManager) Destroy() error {
 
 	stopErr := stopUnit(m.dbus, getUnitName(m.cgroups))
 
-	// Both on success and on error, cleanup all the cgroups we are aware of.
-	// Some of them were created directly by Apply() and are not managed by systemd.
-	if err := cgroups.RemovePaths(m.paths); err != nil {
+	// Both on success and on error, cleanup all the cgroups
+	// we are aware of, as some of them were created directly
+	// by Apply() and are not managed by systemd.
+	if err := cgroups.RemovePaths(m.paths); err != nil && stopErr == nil {
```
cyphar, line 213: I feel this change could've been a bit neater but it works so w/e.

AkihiroSuda (approved):

libcontainer/cgroups/systemd/v1.go
kolyshkin: You mean something like this?
```go
rmErr := cgroups.RemovePaths(m.paths)
// stopErr should prevail as it might be the reason for rmErr.
if stopErr != nil {
     return stopErr
}
return rmErr

Issue #2797, opened 2021-02-08
cgroupv2 devices needs unit tests and a minor rework


cyphar:
Based on #2796 and #2793, it looks like the eBPF devices code needs some more tests to make sure it's acting properly. If we can do emulation of the BPF program (like I did for seccomp), that would be ideal.

cyphar:
I also have a feeling the wildcard behaviour is wrong -- maybe we should use the devices cgroup emulator to get the computed minimal ruleset and then generate a program based on that?

cyphar:
I'm working on this now. I think it's quite critical for devices cgroup security on cgroupv2.

h-vetinari:
Any update on this effort?

cyphar:
It seems that we would need to implement the BPF generation ourselves. Now, I do now have experience with doing this, but I'm a little bit apprehensive about doing it entirely ourselves. Unless it turns out to be simpler than it looks, it's probably going to be a post-1.0 thing. (Especially since the behaviour with updating also needs to be taken into consideration -- see #2366.)

cyphar:
Actually, looking at this again, this might not be too hard to do since cilium supports testing eBPF programs (though rather than doing emulation -- which is what `golang.org/x/net` does -- eBPF testing is actually done in-kernel).

cyphar:
Ah, the kernel doesn't support `BPF_PROG_TEST_RUN` for `prog_type=BPF_PROG_TYPE_CGROUP_DEVICE`. That sucks. In that case we can't really add unit tests right now, but I'll work on sending a patch upstream to support this.

AkihiroSuda:
Merged https://github.com/opencontainers/runc/pull/2951

PR #2951, opened 2021-05-12
cgroup2: devices filtering cleanup

cyphar:
This is my attempt at solving both #2797 and #2366. I'm not entirely sure if the behaviour I've implemented if a cgroup has more than one filter installed (detach all old filters, so that only the new one is effective) is quite correct, but it's necessary to fix the program leak problem (outside of switching to BPF pins entirely).

NOTE: I wanted to add some unit tests for our devicefilter code, but unfortunately `BPF_PROG_TEST_RUN` doesn't support cgroup programs at the moment. I would have to push some code upstream for that to happen (though I'll look into that next week).

<hr>

Changelog entry:

```
cgroup2: devices: rework the filter generation to produce consistent results with cgroupv1, and always clobber any existing eBPF program(s) to fix `runc update` and avoid leaking eBPF programs (resulting in errors when managing containers). 
```


libcontainer/cgroups/ebpf/devicefilter/devicefilter.go
```diff
@@ -10,8 +10,10 @@ import (
 	"math"
 	"strconv"
 
-	"github.com/cilium/ebpf/asm"
+	devicesemulator "github.com/opencontainers/runc/libcontainer/cgroups/devices"
 	"github.com/opencontainers/runc/libcontainer/devices"
+
```
kolyshkin, line 15: Remove the extra newline.

Or, if you want to switch to 3 groups of imports, the order should be
1. standard
2. external third-party
3. "local" (as in `goimports --local`)

IOW `github.com/opencontainers/runc/` should be the last group.

libcontainer/cgroups/ebpf/devicefilter/devicefilter.go
```diff
@@ -68,38 +103,32 @@ func (p *program) init() {
 }
 
 // appendDevice needs to be called from the last element of OCI linux.resources.devices to the head element.
```
kolyshkin, line 105: `s/appendDevice/appendRule/`

kolyshkin: Also seems that this entire comment is now obsoleted.

kolyshkin:
A typo in the second commit message: `s/speciifc/specific/`.

libcontainer/cgroups/ebpf/ebpf.go
```diff
@@ -1,29 +1,93 @@
...
+		QueryType   uint32
+		AttachFlags uint32
+		ProgIds     uint64 // __aligned_u64
+		ProgCnt     uint32
+	}
+
+	// Currently you can only have 64 eBPF programs attached to a cgroup.
+	size := 64
+	allowRetry := true
+	for allowRetry {
```
kolyshkin, line 32: It seems that `allowRetry` is redundant; you can just do

```go
for {
	...
	if errno != 0 {
		// On ENOSPC we get the correct number of programs.
		if errno == unix.ENOSPC {
			size = int(query.ProgCnt)
			continue
		}
		return nil, fmt.Errorf("bpf_prog_query: %w", errno)
	}
	...
	return ...
}
```

libcontainer/cgroups/ebpf/ebpf.go
```diff
@@ -1,29 +1,93 @@
...
+			// On ENOSPC we get the correct number of programs.
+			if errno == unix.ENOSPC {
+				allowRetry = false
+				continue
+			}
+			return nil, errno
+		}
+
+		// Convert the ids to program handles.
+		programs := make([]*ebpf.Program, len(progIds))
```
kolyshkin, line 58: should this be `size` rather than `len(progIds)`?

libcontainer/cgroups/ebpf/ebpf.go
```diff
@@ -1,29 +1,93 @@
...
+		for idx, progId := range progIds {
```
kolyshkin, line 59: same here: `range progIds[:size]`. Or have `progIds = progIds[:size]` above (line 47).

libcontainer/cgroups/ebpf/ebpf.go
```diff
@@ -1,29 +1,93 @@
...
+			if err != nil {
+				return nil, errors.Wrap(err, "cannot fetch program from id")
+			}
+			programs[idx] = program
+		}
+		runtime.KeepAlive(progIds)
+		return programs, nil
+	}
+
+	return nil, errors.Errorf("could not get list of CGROUP_DEVICE programs")
```
kolyshkin, line 70: nit: use `errors.New`

libcontainer/cgroups/ebpf/ebpf.go
```diff
@@ -1,29 +1,93 @@
...
+			}
+			return nil, errno
+		}
+
+		// Convert the ids to program handles.
+		programs := make([]*ebpf.Program, len(progIds))
+		for idx, progId := range progIds {
+			program, err := ebpf.NewProgramFromID(ebpf.ProgramID(progId))
+			if err != nil {
+				return nil, errors.Wrap(err, "cannot fetch program from id")
```
kolyshkin, line 62: The new golang style to wrap an error is `fmt.Errorf("cannot fetch program from id: %w", err)`, but we still have a lot of `errors.Wrap()` so whatever.

libcontainer/cgroups/ebpf/devicefilter/devicefilter.go
cyphar: Interesting, I've always done external third-party last (so it's in order of "locality"). But sure, I can drop the newline.

libcontainer/cgroups/ebpf/ebpf.go
cyphar: I don't want to loop endlessly ... but you're quite right that the `allowRetry` logic was wrong (oops). I'll fix it up.

libcontainer/cgroups/ebpf/ebpf.go
cyphar: Yup, you're quite right.

libcontainer/cgroups/ebpf/ebpf.go
cyphar: I'll switch them all to `Errorf("%w")`.

AkihiroSuda: What is the problem with `errors.Wrap`?

cyphar: There's nothing wrong with `errors.Wrap` specifically, but the newer `fmt.Errorf("%w")` setup is in the stdlib and the stdlib `errors.As` and `errors.Is` are probably going to end up more widely used (`pkg/errors` does still have the upside that you get a stack trace though...).

cyphar:
We also can't test the `runc update` behaviour because it turns out `runc update` doesn't support cgroup device limits... I tried adding support for it, but the behaviour is a bit weird because of the default allowed devices... Hmmm....

kolyshkin:
> We also can't test the runc update behaviour because it turns out runc update doesn't support cgroup device limits... I tried adding support for it, but the behaviour is a bit weird because of the default allowed devices... Hmmm....

Yeah, `update.go` basically repeats specconv, a year ago I tried to reuse the specconv in update but got nowhere...

Maybe we can at least have a simple test case added to libcontainer/integration, which removes and re-adds a device allow rule say 100 times (and check that device access works as expected). Without this PR we should fail on 64th iteration due to eBPF progs limit reached.

Same for runc update -- we could have a stupid test calling `runc update` 100 times (without actually changing anything) but I am not sure if such test makes sense if we implement what's described in the previous paragraph.

kolyshkin (approved):
LGTM except for the missing integration test(s).

cyphar:
Yeah looking at it I think we would need to rethink the "default allow" device list behaviour if we ever want to add devices support to `runc update` (otherwise users will accidentally brick their containers when using it). I'll cook up a unit test on Monday.

cyphar:
Seems as though the Microsoft PPA is down?

tests/integration/update.bats
```diff
@@ -648,3 +648,30 @@ EOF
...
+	# In the past runc would not delete old cgroupv2 eBPF programs, so this
+	# test ensures that once we go past the program limit (64 stacked programs
+	# at time of writing) you can still run "runc" update.
+
+	# Run the container in the background.
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_update
+	[ "$status" -eq 0 ]
+
+	for new_limit in $(seq 300)
+	do
```
kolyshkin, line 670: `make shfmt` suggests:
```diff
-	for new_limit in $(seq 300)
-	do
+	for new_limit in $(seq 300); do
```

I like it either way but having a single coding style is good.

kolyshkin:
CI restarted (looks like it's working today).

kolyshkin:
CI restarted; GHA still has some issues :(

> E: Failed to fetch https://packages.microsoft.com/ubuntu/20.04/prod/dists/focal/main/binary-amd64/Packages.bz2  File has unexpected size (73402 != 73160). Mirror sync in progress? [IP: 104.42.185.173 443]

cyphar:
CI passes again now. :smile_cat:

kolyshkin (approved):
LGTM

kolyshkin:
@AkihiroSuda @mrunalp PTAL

tests/integration/update.bats
```diff
@@ -648,3 +648,29 @@ EOF
...
+	#
+	# In the past runc would not delete old cgroupv2 eBPF programs, so this
+	# test ensures that once we go past the program limit (64 stacked programs
+	# at time of writing) you can still run "runc" update.
+
+	# Run the container in the background.
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_update
+	[ "$status" -eq 0 ]
+
+	for new_limit in $(seq 300); do
```
AkihiroSuda, line 669: seq might cause extra alloc compared to `for ((i = 0; i < 300; i++))`, but negligible

tests/integration/update.bats
```diff
@@ -648,3 +648,29 @@ EOF
 	runc resume test_update
 	[ "$status" -eq 0 ]
 }
+
+@test "runc update replaces devices cgroup program" {
+	[[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
```
AkihiroSuda, line 653: The test is NOP in rootless, right? Can we have a comment line to explain the expected behavior?

libcontainer/cgroups/ebpf/ebpf.go
```diff
@@ -1,29 +1,95 @@
...
+			unsafe.Sizeof(query))
+		size = int(query.ProgCnt)
+		runtime.KeepAlive(query)
+		if errno != 0 {
+			// On ENOSPC we get the correct number of programs.
+			if errno == unix.ENOSPC {
+				retries++
+				continue
+			}
+			return nil, fmt.Errorf("bpf_prog_query(BPF_CGROUP_DEVICE) failed: %w", errno)
```
AkihiroSuda, line 55: We may get EINTR wand want to retry?

libcontainer/integration/exec_test.go
```diff
@@ -1921,6 +1921,9 @@ func TestFdLeaks(t *testing.T) {
 }
 
 func TestFdLeaksSystemd(t *testing.T) {
+	if !systemd.IsRunningSystemd() {
+		t.Skip("Systemd is unsupported")
```
AkihiroSuda, line 1925: "test requires systemd" might be less confusing.

kolyshkin: @AkihiroSuda this is what we already have in other cases:

```
--
libcontainer/integration/exec_test.go:  if !systemd.IsRunningSystemd() {
libcontainer/integration/exec_test.go-          t.Skip("Systemd is unsupported")
--
libcontainer/integration/exec_test.go:  if !systemd.IsRunningSystemd() {
libcontainer/integration/exec_test.go-          t.Skip("Systemd is unsupported")
--
libcontainer/integration/exec_test.go:  if !systemd.IsRunningSystemd() {
libcontainer/integration/exec_test.go-          t.Skip("Systemd is unsupported")
--
libcontainer/integration/exec_test.go:  if !systemd.IsRunningSystemd() {
libcontainer/integration/exec_test.go-          t.Skip("Systemd is unsupported")
--
libcontainer/integration/exec_test.go:  if !systemd.IsRunningSystemd() {
libcontainer/integration/exec_test.go-          t.Skip("Systemd is unsupported")
--
```

libcontainer/cgroups/ebpf/ebpf.go
kolyshkin: bpf(2) is considered EINTR-safe. Can not quote any sources though :(

cyphar: Seems that you can get EINTR from `bpf` (especially in Go thanks to green thread preemption). cilium/ebpf#24

libcontainer/integration/exec_test.go
cyphar: Yeah I copied this from the other integration functions. If we really want to change it we can but we should change it for all of them in one commit.

tests/integration/update.bats
cyphar: It's also no-op for cgroupv1...

libcontainer/cgroups/ebpf/devicefilter/devicefilter.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strconv"
 
 	"github.com/cilium/ebpf/asm"
+	devicesemulator "github.com/opencontainers/runc/libcontainer/cgroups/devices"
```
kolyshkin, line 14: nit: `devicesEmulator`

libcontainer/cgroups/ebpf/ebpf.go
kolyshkin: Found another case -- they retry on `bpf(BPF_PROG_LOAD)` but on `EAGAIN` not `EINTR`.

So I guess we don't need a EINTR handler here.

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

PR #2953, opened 2021-05-13
[WIP] simplify cgroup manager; fix lazy migration failures

kolyshkin:
WIP


kolyshkin:
Still working on that; apparently CRIU needs fixing, too (it is not able to restore into a different cgroup v2 ATM, I need at least to file a bug about it).

Issue #2943, opened 2021-05-06
[RFE] support for seccomp default errno return code


giuseppe:
Allow to override the default errno for seccomp actions.

It was added to the OCI runtime specs with: https://github.com/opencontainers/runtime-spec/pull/1087

PR #2954, opened 2021-05-14
libcontainer: honor seccomp defaultErrnoRet

giuseppe:
https://github.com/opencontainers/runtime-spec/pull/1087 added support
for defaultErrnoRet to the OCI runtime specs.

If a defaultErrnoRet is specified, disable patching the generated
libseccomp cBPF.

Closes: https://github.com/opencontainers/runc/issues/2943


rhatdan:
@kolyshkin @AkihiroSuda @cyphar PTAL

kolyshkin (dismissed):
LGTM, thanks!

libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -523,6 +523,10 @@ func assemble(program []bpf.Instruction) ([]unix.SockFilter, error) {
 }
 
 func generatePatch(config *configs.Seccomp) ([]bpf.Instruction, error) {
+	// Path the generated cBPF only when there is not a defaultErrnoRet set
+	if config.DefaultErrnoRet != nil {
+		return nil, nil
+	}
```
cyphar, line 529: This behaviour is incorrect -- and is what I meant when I said a while ago in the runtime-spec that `defaultErrnoRet` is *useful* but doesn't solve the problem that the patching does.

We must *always* patch the cBPF program if the action is a "deny" action. The fact that the errno is not `-EINVAL` isn't really a good enough reason to re-enable the seccomp footgun (someone is going to set it to `-EACCES` and that'll be a ticking timebomb for no good reason). Maybe we could make it so that `-ENOSYS` is special-cased but that that's the only exception.

cyphar (changes requested):

libcontainer/seccomp/patchbpf/enosys_linux.go
giuseppe: fair.  I'd expect `ENOSYS` to be the only reasonable value for `DefaultErrnoRet` anyway.  I've adjusted the patch to skip patching only when `DefaultErrnoRet == ENOSYS`

cyphar: Thanks! Yeah I would hope *most* people wouldn't set it to anything else, but you can never be too careful.

kolyshkin (approved):

cyphar (approved):
LGTM.

h-vetinari:
CI for https://github.com/opencontainers/runc/commit/c01a56034f5ab0c1aa314377a499fe60a9c26b36 failed pretty comprehensively on master.

cyphar:
@h-vetinari This was because of some issue with Azure (or whatever is backing GitHub actions). The Microsoft Debian repos were broken for the past day or two, and are fixed now (this PR passed before I merged it). I've re-run the CI.

PR #2955, opened 2021-05-14
libct/cg/fs2/freezer: optimize and fix

kolyshkin:
`fs2.setFreezer` improvements:

1. Optimize it to do 1 open instead of 3 (or 0 instead of 1 in case `config.Undefined` is passed).
2. Fix it to actually wait for freeze to be finished.

Note: I chose to use simple wait/retry for re-reading `cgroup.events`. An implementation based on `poll(2)` or `inotify(7)` is possible, but it makes things more complicated. If needed, a poll- or inotify-based waiting can be add later.

### DIFF
--- a/libcontainer/cgroups/fs2/freezer.go
+++ b/libcontainer/cgroups/fs2/freezer.go
@@ -3,8 +3,12 @@
 package fs2
 
 import (
+	"bufio"
 	stdErrors "errors"
+	"fmt"
 	"os"
+	"strings"
+	"time"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
 	"github.com/opencontainers/runc/libcontainer/configs"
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
> Note: I chose to use simple wait/retry for re-reading cgroup.events. An implementation based on poll(2) or inotify(7) is possible, but it makes things more complicated. If needed, a poll- or inotify-based waiting can be add later.

Could you mention inotify in the code comments as TODO (or just open an issue for inotify), then LGTM

cyphar:
LGTM aside from Akihiro's nit.

kolyshkin:
> Could you mention inotify in the code comments as TODO (or just open an issue for inotify), then LGTM

Added a `XXX` comment (as this is "maybe TODO" rather than a hard TODO).

(I actually did a poll(2)-based implementation but it looked so horrible I ditched it. Haven't tried inotify yet (although there's already an implementation in `libcontainter/notify_linux_v2.go` that can possibly be reused.)

cyphar (approved):
LGTM.

kolyshkin:
Fedora CI failed during `vagrant up`; restarted

AkihiroSuda (approved):

PR #2925, opened 2021-04-28
Relax the requirement of Go version: 1.14 -> 1.13

AkihiroSuda:
Dockre/Moby still builds runc with Go 1.13, so we should still support Go 1.13.

PR #2956, opened 2021-05-20
Fix/improve runc -v

kolyshkin:
Two things:

1. `runc -v`: set default for, always show `main.version`
    
    Apparently not everyone compiles runc via the provided Makefile. For
    example, one can just run `go build`, in which case `version` variable
    is left empty, which leads to:
    
            $ ./runc -v
            runc version spec: 1.0.2-dev
            go: go1.16.3
    
    Surely, the main problem here is runc was built in a wrong way, but the
    second problem is such output is very confusing -- it may seem that we
    have runc 1.0.2.
    
    To solve, make sure to _always_ add version (even if empty), plus set a
    default value in case someone compiles runc in a wrong way.
    
            $ ./runc -v
            runc version unknown
            spec: 1.0.2-dev
            go: go1.16.3

    NOTE this does not change anything in case runc is compiled normally
    (via the Makefile).
       
2. `Makefile`: use `git describe` for `COMMIT`
    
    Use `git describe --dirty --long` instead of `git rev-parse`. As a
    result, the commit ID will contain the closest tag, the number of commits
    since the tag, and the (abbreviated) git commit sha (see the example below).
    
    NOTE that this tag is still unique and can be used instead of bare sha
    for all git commands.
    
    Example output of `runc -v | grep commit`.
    
    Before:

            commit: 4d875738716862d08a84c00211b1983c3044f711

    After:

            commit: v1.0.0-rc95-9-g6f55d074
    
    This means that
     - the closest tag is v1.0.0-rc95
     - there were 9 commits after the tag
     - the abbreviated sha is 6f55d074

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,9 +15,9 @@ import (
 	"github.com/urfave/cli"
 )
 
-// version will be populated by the Makefile, read from
-// VERSION file of the source code.
-var version = ""
+// version must be set from the contents of VERSION file by go build's
+// -X main.version= option in the Makefile.
+var version = "unknown"
 
 // gitCommit will be the hash that the binary was built from
 // and will be populated by the Makefile
@@ -55,10 +55,8 @@ func main() {
 	app.Name = "runc"
 	app.Usage = usage
 
-	var v []string
-	if version != "" {
-		v = append(v, version)
-	}
+	v := []string{version}
+
 	if gitCommit != "" {
 		v = append(v, "commit: "+gitCommit)
 	}
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,7 @@ GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 RUNC_IMAGE := runc_dev$(if $(GIT_BRANCH_CLEAN),:$(GIT_BRANCH_CLEAN))
 PROJECT := github.com/opencontainers/runc
 BUILDTAGS ?= seccomp
-COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
-COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),"$(COMMIT_NO)-dirty","$(COMMIT_NO)")
+COMMIT ?= $(shell git describe --dirty --long --always)
 VERSION := $(shell cat ./VERSION)
 
 # TODO: rm -mod=vendor once go 1.13 is unsupported


AkihiroSuda (approved):

AkihiroSuda:
I'm feeling that printing the spec version is almost meaningless because it is actually the version of the `runtime-spec` Go module, not the version of the spec itself. But if we omit the `spec:` line, it may break some downstream projects... So LGTM.

main.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // version will be populated by the Makefile, read from
 // VERSION file of the source code.
-var version = ""
+var version = "(unset)"
```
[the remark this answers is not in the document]
kolyshkin: done

kolyshkin:
> Funnily enough on go1.16 we can finally get rid of the need to set VERSION through -ldflags -X because we could use //go:embed VERSION. But probably best left for another PR.

Yes, using `go:embed` for this would be nice! Still, given the fact that we still have to support 1.13 now (https://github.com/opencontainers/runc/pull/2925), the day when we can drop go < 1.16 compatibility is yet to come...

kolyshkin:
> While I realise this is already broken, will (unset) break Docker's parsing of the version information?

I guess it's no more broken than having to parse "spec 1.0.2-dev" as a runc version (which, if I am not mistaken, the docker version bundled in Ubuntu 20.04 currently shows -- so it won't be a problem to show "unknown").

kolyshkin:
Yes, this is from Ubuntu 20.04:
```console
kir@ubu2004:~$ docker version
Client:
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.8
 Git commit:        20.10.2-0ubuntu1~20.04.2
 Built:             Tue Mar 30 21:24:57 2021
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.2
  Go version:       go1.13.8
  Git commit:       20.10.2-0ubuntu1~20.04.2
  Built:            Mon Mar 29 19:10:09 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.3.3-0ubuntu2.3
  GitCommit:        
 runc:
  Version:          spec: 1.0.2-dev
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        
kir@ubu2004:~$ dpkg -l docker.io containerd runc
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version                     Architecture Description
+++-==============-===========================-============-=================================
ii  containerd     1.3.3-0ubuntu2.3            amd64        daemon to control runC
ii  docker.io      20.10.2-0ubuntu1~20.04.2    amd64        Linux container runtime
ii  runc           1.0.0~rc93-0ubuntu1~20.04.2 amd64        Open Container Project - runtime
kir@ubu2004:~$ /usr/sbin/runc -v
runc version spec: 1.0.2-dev
go: go1.13.8
libseccomp: 2.5.1

```

thaJeztah:
Wondering if instead it would make sense to default to "same as is in `VERSION`"? (currently `1.0.0-rc95+dev`), or (containerd approach); `1.0.0-rc94+unknown`, that way the version is "slightly" more representative for what's built.

(Would probably be good to then add a check in CI to make sure we don't forget to update the version in Go together with `VERSION` when doing a release)

kolyshkin:
@thaJeztah not a bad idea, please see updated commit.

thaJeztah (dismissed):
yup, I like this variant

LGTM (non binding)

kolyshkin:
> I'm feeling that printing the spec version is almost meaningless because it is actually the version of the runtime-spec Go module, not the version of the spec itself. But if we omit the spec: line, it may break some downstream projects... So LGTM.

@AkihiroSuda we can still remove it while we're before 1.0 -- let me know if you want me to do it.

We can also replace it with git commit from go.mod, with some additional sorcery in the Makefile, but I'd rather not.

main.go
```diff
@@ -15,9 +15,10 @@ import (
 	"github.com/urfave/cli"
 )
 
-// version will be populated by the Makefile, read from
-// VERSION file of the source code.
-var version = ""
+// version is normally set from the contents of VERSION file by go build's
+// -X main.version= option in the Makefile. In case it's not set, the value
+// here provides a sane default (matching the VERSION modulo +prefix).
+var version = "1.0.0-rc95+unknown"
```
cyphar, line 21: If we do it this way then we technically don't need to have `VERSION` at all (after all the idea, of `VERSION` was specifically so we didn't need to have the version string in the source code). You can just grab the version by just `grep`-ing it out.

(I also am not sure I like that the `+dev` prefix is dropped and tagged releases will have the same version as untagged commits. If we're going to put the version in the source we should just get rid of `VERSION` -- at least until we can switch to `go:embed` and go back.)

cyphar:
> I'm feeling that printing the spec version is almost meaningless because it is actually the version of the `runtime-spec` Go module, not the version of the spec itself. But if we omit the `spec:` line, it may break some downstream projects... So LGTM.

The `runtime-spec` tags are the same as the spec tags. It is a bit dodgy to have `+dev` versions, but it is technically useful to know what version of the spec structures are supported (even though it doesn't actually guarantee runc implements them properly...)

main.go
kolyshkin: In this case I'd rather go back to the original version of this PR -- having `unknown` as the default.

kolyshkin:
Reverted to the original version of this PR -- having `unknown` as the default for version string.

@cyphar @AkihiroSuda PTAL

cyphar (approved):
LGTM. Yeah grepping the version out of .go files is a bit hairy. At least once we can update to 1.16+ we can make this fully consistent.

AkihiroSuda (approved):

PR #2958, opened 2021-05-22
libct/cg/sd: fix SkipDevices for systemd

kolyshkin:
1. **libct/cg/sd: fix SkipDevices for systemd**
    
    Commit 108ee85b8271e (PR #2490) adds `SkipDevices` flag, which is used by kubernetes
    to create cgroups for pods.
    
    Unfortunately the above commit falls short, and systemd `DevicePolicy` and
    `DeviceAllow` properties are still set, which requires kubernetes to set
    "allow everything" rule.
    
    This commit fixes this: if `SkipDevices` flag is set, we return
    `Device*` properties to allow all devices.

2. **libct/cg/sd: add SkipDevices unit test**
    
    The idea is to mimic what kubelet is doing, with minimum amount of code.
    
    First, create a slice with `SkipDevices=true`. It should have access to
    all devices.
    
    Next, create a scope within the above slice, allowing access to `/dev/full`
    only.
    
    Check that within that scope we can only access `/dev/full` and not other
    devices (such as `/dev/null`).
    
    Repeat the test with `SkipDevices=false`, make sure we can **not** access any
    devices (as they are disallowed by a parent cgroup). This is done only
    to assess the test correctness.
    
    NOTE that cgroup v1 and v2 behave differently for `SkipDevices=false`
    case, and thus the check is different. Cgroup v1 returns `EPERM` on
    writing to `devices.allow`, so cgroup manager's `Set()` fails, and we check
    for a particular error from `m.Set()`. Cgroup v2 allows to create a child
    cgroup, but denies access to any device (despite access being enabled)
    -- so we check the error from the shell script running in that cgroup.
    Again, this is only about `SkipDevices=false` case.

Previous discussions on topic: https://github.com/opencontainers/runc/pull/2490#discussion_r448702204, https://github.com/kubernetes/kubernetes/pull/92862#discussion_r451147445

Fixes: 108ee85b8271e

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -158,14 +158,27 @@ func findDeviceGroup(ruleType devices.Type, ruleMajor int64) (string, error) {
 	return "", nil
 }
 
+// DeviceAllow is the dbus type "a(ss)" which means we need a struct
+// to represent it in Go.
+type deviceAllowEntry struct {
+	Path  string
+	Perms string
+}
+
+func allowAllDevices() []systemdDbus.Property {
+	// Setting mode to auto and removing all DeviceAllow rules
+	// results in allowing access to all devices.
+	return []systemdDbus.Property{
+		newProp("DevicePolicy", "auto"),
+		newProp("DeviceAllow", []deviceAllowEntry{}),
+	}
+}
+
 // generateDeviceProperties takes the configured device rules and generates a
 // corresponding set of systemd properties to configure the devices correctly.
-func generateDeviceProperties(rules []*devices.Rule) ([]systemdDbus.Property, error) {
-	// DeviceAllow is the type "a(ss)" which means we need a temporary struct
-	// to represent it in Go.
-	type deviceAllowEntry struct {
-		Path  string
-		Perms string
+func generateDeviceProperties(r *configs.Resources) ([]systemdDbus.Property, error) {
+	if r.SkipDevices {
+		return allowAllDevices(), nil
 	}
 
 	properties := []systemdDbus.Property{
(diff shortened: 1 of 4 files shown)


libcontainer/cgroups/systemd/common.go
```diff
@@ -160,7 +160,15 @@ func findDeviceGroup(ruleType devices.Type, ruleMajor int64) (string, error) {
...
 // generateDeviceProperties takes the configured device rules and generates a
 // corresponding set of systemd properties to configure the devices correctly.
-func generateDeviceProperties(rules []*devices.Rule) ([]systemdDbus.Property, error) {
+func generateDeviceProperties(r *configs.Resources) ([]systemdDbus.Property, error) {
+	if r.SkipDevices {
+		// Not adding any Device* properties means the default of
+		// DevicePolicy=auto will be used by systemd, which allows
+		// access to all devices (if no explicit DeviceAllow= is
+		// present).
+		return []systemdDbus.Property{}, nil
```
cyphar, line 169: Does this actually reset the policy or does it keep the old one? What I did was set `DevicePolicy=auto` explicitly later in this function. If that's really not necessary then we can also drop it from the later "this rule is allow-all" `return ...`.

kolyshkin: Initially I did the same thing but then thought it is not needed. Now I'm not sure -- will test this tomorrow.

kolyshkin: Ah, OK, I did not have the update scenario in mind, only the initial one-time configuration (assuming no `Device*` properties were ever set before for the given unit).

Modified the commit to explicitly return `DevicePolicy=auto` and an empty list of `DeviceAllow` rules, reusing the code you added earlier.

cyphar (approved):
LGTM.

AkihiroSuda:
Would it be possible to have some tests?

kolyshkin:
> Would it be possible to have some tests?

I started working on something that mimics the kubelet behavior...

cyphar:
CentOS fails -- SELinux?

cyphar:
Once this is merged I will send out the v1.0.0 vote (it would be nice to do the release -- as @h-vetinari suggested -- on the 3rd of June since that's the 5-year anniversary of 1.0.0-rc1).

kolyshkin:
Test case fixed, PR description updated.

mrunalp (approved):

AkihiroSuda (approved):

cyphar (approved):
LGTM. Will send the v1.0.0 vote out today.

PR #2961, opened 2021-05-24
tests/int/dev: only call lsblk once

kolyshkin:
The "runc run [device cgroup allow rm block device]" test calls lsblk
three times to get device name, minor and major number. This creates a
potential problem when the devices are changed between the calls.

Simplify the code by using bash read together with IFS (as there's no
way to have lsblk output MAJOR:MINOR pair without a semicolon).

Note that head -n 1 is not needed as read already reads a single line.


cyphar (approved):
LGTM.

AkihiroSuda (approved):

PR #2962, opened 2021-05-25
*: clean up remaining golangci-lint failures

cyphar:
Most of these were false positives or cases where we want to ignore the
lint, but the change to the BPF generation is actually useful.

Follow-up to #2781

### DIFF
--- a/libcontainer/system/proc.go
+++ b/libcontainer/system/proc.go
@@ -96,7 +96,7 @@ func parseStat(data string) (stat Stat_t, err error) {
 	// one (PID) and two (Name) in the paren-split.
 	parts = strings.Split(data[i+2:], " ")
 	var state int
-	fmt.Sscanf(parts[3-3], "%c", &state)
+	fmt.Sscanf(parts[3-3], "%c", &state) //nolint:staticcheck // "3-3" is more readable in this context.
 	stat.State = State(state)
 	fmt.Sscanf(parts[22-3], "%d", &stat.StartTime)
 	return stat, nil
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -21,8 +21,6 @@ jobs:
         with:
           # must be specified without patch version
           version: v1.40
-          # Only show new issues for a pull request.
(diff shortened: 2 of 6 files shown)


cyphar:
Should we give warnings for all lint errors? I can disable that flag again, but I noticed that some of the new errors were added in PRs that I made *after* we enabled the CI check -- so I wonder if the `golangci-lint` struggles to deal with giving new warnings if there are existing warnings? If we just make sure we only merge code that passes `golangci-lint` it probably doesn't hurt to just show all errors for every PR run...

AkihiroSuda (approved):

hqhq (approved):

libcontainer/system/proc.go
```diff
@@ -96,7 +96,7 @@ func parseStat(data string) (stat Stat_t, err error) {
 	// one (PID) and two (Name) in the paren-split.
 	parts = strings.Split(data[i+2:], " ")
 	var state int
-	fmt.Sscanf(parts[3-3], "%c", &state)
+	fmt.Sscanf(parts[3-3], "%c", &state) //nolint:staticcheck // "3-3" is more readable in this context.
```
kolyshkin, line 99: I have a fix for this https://github.com/opencontainers/runc/pull/2696 (which started as a fix for this linter issue but ended up as a rewrite of `parseStat` to be 7x faster and more error-prone).

kolyshkin:
> Should we give warnings for all lint errors? I can disable that flag again, but I noticed that some of the new errors were added in PRs that I made _after_ we enabled the CI check -- so I wonder if the `golangci-lint` struggles to deal with giving new warnings if there are existing warnings? If we just make sure we only merge code that passes `golangci-lint` it probably doesn't hurt to just show all errors for every PR run...

The idea was to drop this golangci-lint flag as long as the master branch is clean of any existing issues, which is what this PR does, so yes, sure.

And yes indeed, I occasionally saw it skip some issues, probably because their "limit the scope to what this PR brings" algo is not perfect for some reason.

Issue #2345, opened 2020-04-22
Support time namespaces


djboris9:
Hi there

Since Kernel 5.6 there is support for time namespaces using `CLONE_NEWTIME` (see `setns(2)`).
Support for it in `runc` would allow people to have different times in containers, which is useful for testing of time based services/jobs.

Currently we are doing it with [libfaketime](https://github.com/wolfcw/libfaketime) but the software which is being tested needs to be compatible with it.

It looks like `golang.org/x/sys/unix` already supports the `CLONE_NEWTIME` argument.

I've opened this issue to track it. If I get some time I will try to implement it and open a pull request.

Boris

P.S. The challenge seams to be that as soon as the first child is spawned after the unshare, the time offsets cannot be changed. An example is shown in [time_namespaces(7)](http://man7.org/linux/man-pages/man7/time_namespaces.7.html)

kolyshkin:
runc should not support timens directly, this feature is aimed mostly for stuff like checkpoint/restore (so that the system time won't jump forward in the restored container). Maybe @avagin can chime in, too.

Feel free to discuss it further, but I'm closing this for now

AkihiroSuda:
I think runc can support timens directly. Having different clocks across containers would be very useful for Jepsen(-ish)  testing.

cyphar:
I agree, though the only annoyance is that we'd need to have a different configuration mechanism for `timens` than any of the other namespaces -- because the canonical way to configure `CLONE_NEWTIME` is with the `clone_args` passed to `clone3`. So we'd also need to port `nsexec.c` to `clone3` but that's something we were going to have to do anyway (and `clone3` is objectively much nicer than `clone2`).

In any case, this issue should be an issue/PR against the runtime-spec.

tianon:
FWIW, there's an initial PR on the runtime-spec now: https://github.com/opencontainers/runtime-spec/pull/1062

thaJeztah:
- This should be addressed by https://github.com/opencontainers/runc/pull/3876

PR #2963, opened 2021-05-25
libcontainer: configs: initial support for timens

metux:
Allow explicit unsharing of time namespace.


cyphar:
This isn't really sufficient, in order to really make use of the timens you need to configure the time offsets when the container first starts (and that would require switching `nsenter` to use `clone3`, as well as figure out a reasonable way to configure it -- which will probably require runtime-spec changes).

kolyshkin:
@metux if you want to work on that, the first step is working on the proposal in https://github.com/opencontainers/runtime-spec/pull/1062, which is still pending. Once merged, we can look into implementing timens.

Closing for now as much more changes are needed than adding a few constants (for more details, see https://github.com/opencontainers/runc/issues/2345).

PR #2965, opened 2021-05-26
cgroup2: capitalize io stats read and write Op values

### LINKED ISSUE #2964: Inconsistent IO stats Op values between cgroup v1 and v2
We use libcontainer cgroup library [GetStats](https://github.com/opencontainers/runc/blob/e005fee9d37f3ee35cfc6d4cfee687043656928f/libcontainer/cgroups/cgroups.go#L23) api in [cadvisor](https://github.com/google/cadvisor/) to fetch the usage stats for a given cgroup set. But it turns out that cgroup v2 [implementation](https://github.com/opencontainers/runc/blob/e005fee9d37f3ee35cfc6d4cfee687043656928f/libcontainer/cgroups/fs2/io.go#L122-L124) of cgroup manager GetStats api provides blkio stats with lower case `Op` keys under `BlkioStatEntry` as compare to cgroup v1. For instance, in cgroup v1, we get Op = `Write` for `IoServiceBytesRecursive` under `BlkioStats`, while in cgroup v2, we get Op = `write` (lower w case) for same stat. See [this](https://github.com/opencontainers/runc/blob/c061bc78aae9b39b8ac5d3011e55736b20288cb3/libcontainer/cgroups/fs/blkio.go#L72-L101 ) to know about how BlkioStatEntry get populated in cg1 as compare to [cg2](https://github.com/opencontainers/runc/blob/e005fee9d37f3ee35cfc6d4cfee687043656928f/libcontainer/cgroups/fs2/io.go#L122-L124).

Does it make sense to change Op value to capitalized case `Read` and `Write` similar to cgroup v1? If yes, I can create a pull request with the fix.

Original cAdvisor Issue - https://github.com/google/cadvisor/issues/2881

iyashu:
Fixes #2964

### DIFF
--- a/libcontainer/cgroups/fs2/io.go
+++ b/libcontainer/cgroups/fs2/io.go
@@ -118,9 +118,9 @@ func statIo(dirPath string, stats *cgroups.Stats) error {
 			// Accommodate the cgroup v1 naming
 			switch op {
 			case "rbytes":
-				op = "read"
+				op = "Read"
 			case "wbytes":
-				op = "write"
+				op = "Write"
 			}
 
 			value, err := strconv.ParseUint(d[1], 10, 0)


AkihiroSuda (approved):

dims:
yes please! LGTM

dims (approved):

cyphar (approved):
LGTM. We should probably add some kind of test to make sure we don't have divergences like this in the future.

PR #2966, opened 2021-05-27
golangci: enable GoDoc linting

thaJeztah:
relates to https://github.com/opencontainers/runc/issues/2627

golint is scoring pretty low currently, because many exported
functions and variables are missing a well-formatted GoDoc comment:

https://goreportcard.com/report/github.com/opencontainers/runc

PR #2974, opened 2021-06-01
Fix various linting issues

thaJeztah:
splitting this from https://github.com/opencontainers/runc/pull/2966
relates to https://github.com/opencontainers/runc/issues/2627


thaJeztah:
@kolyshkin @AkihiroSuda PTAL

still some to address after this (see https://github.com/opencontainers/runc/pull/2966#issuecomment-852049218), but cleaning up the other PR

libcontainer/apparmor/apparmor_unsupported.go
```diff
@@ -6,13 +6,14 @@ import (
 	"errors"
 )
 
+// ErrApparmorNotEnabled indicates that AppArmor is not enabled or not supported.
 var ErrApparmorNotEnabled = errors.New("apparmor: config provided but apparmor not supported")
```
AkihiroSuda, line 9: This looks defined twice (but passing CI?)

thaJeztah: Good catch; looks like I didn't remove it here.

I guess CI doesn't catch it as it only runs on Linux?

thaJeztah: Fixed (removed this one)

kzys (approved):

libcontainer/configs/cgroup_linux.go
```diff
@@ -13,12 +13,14 @@ const (
 	Thawed    FreezerState = "THAWED"
 )
 
+// Cgroup holds properties of a cgroup on Linux
```
kolyshkin, line 16: nit: period missing

libcontainer/configs/cgroup_linux.go
```diff
@@ -13,12 +13,14 @@ const (
...
 type Cgroup struct {
-	// Deprecated, use Path instead
+	// Name specifies the name of the cgroup
+	// Deprecated: use Path instead
```
kolyshkin, line 19: I recently found out that Name and Parent are not deprecated at least for systemd cgroup manager case, because Path can not be used in that case. This might have to be fixed, but for now, I guess, it's better to remove the deprecation notice.

libcontainer/configs/mount.go
```diff
@@ -3,7 +3,7 @@ package configs
 const (
 	// EXT_COPYUP is a directive to copy up the contents of a directory when
 	// a tmpfs is mounted over it.
-	EXT_COPYUP = 1 << iota
+	EXT_COPYUP = 1 << iota //nolint:golint // ignore don't use ALL_CAPS in Go names; use CamelCase (golint)
```
kolyshkin, line 6: Perhaps
```suggestion
	EXT_COPYUP = 1 << iota //nolint:golint // ignore "don't use ALL_CAPS in Go names; use CamelCase (golint)"
```

or just
```suggestion
	EXT_COPYUP = 1 << iota //nolint:golint // ignore "don't use ALL_CAPS" warning
```

libcontainer/configs/network.go
```diff
@@ -50,23 +50,26 @@ type Network struct {
 	HairpinMode bool `json:"hairpin_mode"`
 }
 
-// Routes can be specified to create entries in the route table as the container is started
+// Route defines a route table entry.
```
kolyshkin, line 53: I think the proper term is "routing table" not "route table".

So
```suggestion
// Route defines a routing table entry.
```

libcontainer/configs/network.go
```diff
@@ -50,23 +50,26 @@ type Network struct {
...
+//
+// Routes can be specified to create entries in the route table as the container
```
kolyshkin, line 55: same: `s/route table/routing table/`

libcontainer/configs/network.go
```diff
@@ -50,23 +50,26 @@ type Network struct {
...
+// is started.
 //
 // All of destination, source, and gateway should be either IPv4 or IPv6.
 // One of the three options must be present, and omitted entries will use their
 // IP family default for the route table.  For IPv4 for example, setting the
 // gateway to 1.2.3.4 and the interface to eth0 will set up a standard
 // destination of 0.0.0.0(or *) when viewed in the route table.
 type Route struct {
-	// Sets the destination and mask, should be a CIDR.  Accepts IPv4 and IPv6
+	// Destination specifies the destination and mask, should be a CIDR.  Accepts IPv4 and IPv6
```
kolyshkin, line 64:
```suggestion
// Destination specifies the destination IP address and mask in the CIDR form, either IPv4 or IPv6.
```

libcontainer/configs/network.go
```diff
@@ -50,23 +50,26 @@ type Network struct {
...
 	Destination string `json:"destination"`
 
-	// Sets the source and mask, should be a CIDR.  Accepts IPv4 and IPv6
+	// Source specifies the source and mask, should be a CIDR.  Accepts IPv4 and IPv6
```
kolyshkin, line 67: ditto

libcontainer/configs/network.go
```diff
@@ -50,23 +50,26 @@ type Network struct {
...
 	Source string `json:"source"`
 
-	// Sets the gateway.  Accepts IPv4 and IPv6
+	// Gateway specifies the gateway.  Accepts IPv4 and IPv6
```
kolyshkin, line 70:
```suggestion
// Gateway specifies the gateway IP address.
```

libcontainer/configs/network.go
```diff
@@ -50,23 +50,26 @@ type Network struct {
...
 	Gateway string `json:"gateway"`
 
-	// The device to set this route up for, for example: eth0
+	// InterfaceName specifies the device to set this route up for, for example: eth0
```
kolyshkin, line 73:
```suggestion
	// InterfaceName specifies the device to set this route up for, for example eth0.
``````

File: libcontainer/user/user.go [hunk 11]
```diff
@@ -11,16 +11,17 @@ import (
 )
 
 const (
-	minId = 0
-	maxId = 1<<31 - 1 //for 32-bit systems compatibility
+	minID = 0
+	maxID = 1<<31 - 1 // for 32-bit systems compatibility
```
kolyshkin, line 15: I think I have a similar change in https://github.com/opencontainers/runc/pull/2975

libcontainer/user/user.go
```diff
@@ -11,16 +11,17 @@ import (
...
 )
 
 var (
-	// No matching entries found in file.
+	// ErrNoPasswdEntries is generated if no matching entries were found in /etc/group.
```
kolyshkin, line 19: nit: `s/generated/returned/`

libcontainer/user/user.go
```diff
@@ -11,16 +11,17 @@ import (
...
 	ErrNoPasswdEntries = errors.New("no matching entries in passwd file")
-	ErrNoGroupEntries  = errors.New("no matching entries in group file")
-
-	ErrRange = fmt.Errorf("uids and gids must be in range %d-%d", minId, maxId)
+	// ErrNoGroupEntries is generated if no matching entries were found in /etc/passwd.
```
kolyshkin, line 21: ditto

libcontainer/user/user.go
```diff
@@ -11,16 +11,17 @@ import (
...
+	ErrNoGroupEntries = errors.New("no matching entries in group file")
+	// ErrRange is generated if a UID or GID is outside of the valid range.
```
kolyshkin, line 23: ditto

libcontainer/devices/device_unix.go
```diff
@@ -29,8 +29,9 @@ func mkDev(d *Rule) (uint64, error) {
 	return unix.Mkdev(uint32(d.Major), uint32(d.Minor)), nil
 }
 
-// Given the path to a device and its cgroup_permissions(which cannot be easily queried) look up the
-// information about a linux device and return that information as a Device struct.
+// DeviceFromPath takes the path to a device and its cgroup_permissions(which
```
kolyshkin, line 32: nit: missing space before `(which`

kolyshkin (changes requested):
Mostly LGTM except for a few nits.

I wonder where these errors are coming from? I don't see it running `golangci-lint run ./...` locally, nor in GHA CI. Are you using a different golangci config or version?

kolyshkin:
Tentatively set 1.0.0 milestone as this seems like a good fit.

cyphar:
I'm not sure we need to tag this for 1.0.0. If it makes it in, great -- but I think we should go for a release if #2986 fixes the Moby CI regression.

thaJeztah:
Thanks for review! I'll try to update this PR later tonight

> I wonder where these errors are coming from? I don't see it running golangci-lint run ./... locally, nor in GHA CI. Are you using a different golangci config or version?

Yes; enabled "golint" and `EXC002`, which is a default exception in golang-ci-lint ("because nobody writes docs the right way 🙄) see the last commit on https://github.com/opencontainers/runc/pull/2966

> I'm not sure we need to tag this for 1.0.0. If it makes it in, great -- but I think we should go for a release if #2986 fixes the Moby CI regression.

I agree for the changes in _this_ PR (they're "nice to have", but most can be merged at any time). That said; the linter also picked up some incorrectly named _exported_ variables and properties; see https://github.com/opencontainers/runc/pull/2966#issuecomment-852049218

Changing those after v1.0.0 would be a breaking change, so perhaps it's good to have a look at those, and decide wether or not they should be renamed, or "ignored" ("won't fix").

thaJeztah:
@kolyshkin @cyphar @AkihiroSuda updated; PTAL

AkihiroSuda (approved):

kolyshkin (approved):
LGTM (some missing periods at end of sentences but it's OK).

Thank you @thaJeztah!

thaJeztah:
> some missing periods at end of sentences but it's OK

Ah, sorry. yes, that's me. I have a tendency of omitting punctuation for single-line comments 😓 😅

PR #2977, opened 2021-06-01
Enable dependabot

kolyshkin:
This should enable a bot that auto-creates PRs to update dependencies.
For more info, see https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically


kolyshkin:
Once enabled, it should be seen from https://github.com/opencontainers/runc/network/updates (as well as new PRs).

PR #2987, opened 2021-06-02
make systemd and its dependencies optional via 'no_systemd' build tag

metux:
Running under systemd requires lots of special code that contributes
to ca. 10 percent (ca. 1 MB) to the binary size. This is only needed on
targets that might run systemd - there're dozens of distros, let alone
embedded/edge devices or special images (eg. cluster worker nodes) that
do not and never will run systemd, thus do not need that code at all.

It's not just about reducing memory consumption, but also having over
10.000 lines of code less to audit.

In order not to change default behaviour, introducing an inverse build tag,
'no_systemd', for explicitly opting out from systemd special handlings.

### DIFF
--- a/libcontainer/configs/cgroup_systemd_linux.go
+++ b/libcontainer/configs/cgroup_systemd_linux.go
@@ -1,0 +1,15 @@
+// +build linux,!no_systemd
+
+package configs
+
+import (
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+)
+
+// Systemd specific parts of Cgroup properties
+type CgroupSD struct {
+	// SystemdProps are any additional properties for systemd,
+	// derived from org.systemd.property.xxx annotations.
+	// Ignored unless systemd is used for managing cgroups.
+	SystemdProps []systemdDbus.Property `json:"-"`
+}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,7 @@ jobs:
         go-version: [1.13.x, 1.15.x, 1.16.x]
         rootless: ["rootless", ""]
         race: ["-race", ""]
+        buildtags: ["seccomp", "seccomp,no_systemd", "no_systemd", ""]
 
     steps:
 
@@ -42,8 +43,8 @@ jobs:
         stable: '!contains(${{ matrix.go-version }}, "beta") && !contains(${{ matrix.go-version }}, "rc")'
         go-version: ${{ matrix.go-version }}
 
-    - name: build
-      run: sudo -E PATH="$PATH" make EXTRA_FLAGS="${{ matrix.race }}" all
+    - name: build with tags ${{ matrix.buildtags }}
+      run: sudo -E PATH="$PATH" make EXTRA_FLAGS="${{ matrix.race }}" BUILDTAGS="${{ matrix.buildtags }}" all
 
     - name: install bats
       uses: mig4/setup-bats@v1
--- a/README.md
+++ b/README.md
@@ -67,9 +67,10 @@ e.g. to disable seccomp:
 make BUILDTAGS=""
 ```
 
-| Build Tag | Feature                            | Enabled by default | Dependency |
-|-----------|------------------------------------|--------------------|------------|
-| seccomp   | Syscall filtering                  | yes                | libseccomp |
+| Build Tag  | Feature                            | Enabled by default | Dependency |
+|------------|------------------------------------|--------------------|------------|
+| seccomp    | Syscall filtering                  | yes                | libseccomp |
+| no_systemd | disable systemd dependencies       | no                 | systemd    |
 
 The following build tags were used earlier, but are now obsoleted:
  - **nokmem** (since runc v1.0.0-rc94 kernel memory settings are ignored)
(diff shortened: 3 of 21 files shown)


libcontainer/configs/cgroup_systemd_linux.go
```diff
@@ -0,0 +1,35 @@
...
+	// This takes precedence over Path.
+	Paths map[string]string
+
+	// Resources contains various cgroups settings to apply
+	*Resources
+
+	// SystemdProps are any additional properties for systemd,
+	// derived from org.systemd.property.xxx annotations.
+	// Ignored unless systemd is used for managing cgroups.
+	SystemdProps []systemdDbus.Property `json:"-"`
```
kolyshkin, line 34: Instead of duplicating the whole structure, you need to embed a sub-structure which can be defined to be empty in case `no_systemd` build tag is set.

cyphar:
@AkihiroSuda @kolyshkin Does cgroupv2 still require systemd, or is systemd support only required if you're using cgroupv2 on a systemd system? (From memory the early design required it, but I haven't closely followed how it's evolved since. Since we have `fs2` I guess you can technically run cgroupv2 without systemd -- as long as you're not running with systemd, because of the obvious conflicts.)

This needs new CI jobs to make sure that `no_systemd` builds actually work and pass the unit/integration tests.

AkihiroSuda:
> Does cgroupv2 still require systemd, or is systemd support only required if you're using cgroupv2 on a systemd system?

Never required on non-systemd hosts.

Even on systemd hosts, the systemd driver was never strictly required in practice, but future version of systemd may strictly require it, in theory.

AkihiroSuda:
> This needs new CI jobs to make sure that no_systemd builds actually work and pass the unit/integration tests.

Yes, and README (the build tags section) needs to be updated too.

cyphar:
> Even on systemd hosts, the systemd driver was never strictly required in practice, but future version of systemd may strictly require it, in theory.

kolyshkin:
> Though you can't set `Delegate=yes` if you don't have systemd support I guess?

I think we can implement "minimal" systemd support, adding something like
```go
if IsRunningSystemd() {
    err := os.Exec("systemctl", "set-property", unit, "Delegate=yes").Run()
    ...
}
```

This code won't add any new dependencies or impact the binary size much. In fact, ideally this should be done with 

Thinking about this, an "alternative" systemd driver can be added to runc, which will call `systemctl` upon container start/stop/update. This has its own pros and cons, of course.

metux:
> > Though you can't set `Delegate=yes` if you don't have systemd support I guess?
> 
> I think we can implement "minimal" systemd support, adding something like
> 
> ```go
> if IsRunningSystemd() {
>     err := os.Exec("systemctl", "set-property", unit, "Delegate=yes").Run()
>     ...
> }
> ```
> 
> This code won't add any new dependencies or impact the binary size much. In fact, ideally this should be done with

Where does IsRunningSystemd() come from when speaking to systemd and thats dependencies are off ?
I wonder what the purpose if that shall be when systemd support is switched off anyways.

OTOH, I feel very unhappy with such arbitrary command calls, especially in case when systemd stuff had been disabled explicitly. Could end up calling a completely different program. Actually, I once used to write one with that name, long before systemd even existed.

libcontainer/configs/cgroup_systemd_linux.go
metux: fixed that in recent push

kolyshkin:
I think we should have it in 1.1. Set the milestone accordingly.

kolyshkin:
> > > Though you can't set `Delegate=yes` if you don't have systemd support I guess?
> > 
> > 
> > I think we can implement "minimal" systemd support, adding something like
> > ```go
> > if IsRunningSystemd() {
> >     err := os.Exec("systemctl", "set-property", unit, "Delegate=yes").Run()
> >     ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This code won't add any new dependencies or impact the binary size much. In fact, ideally this should be done with
> 
> Where does IsRunningSystemd() come from when speaking to systemd and thats dependencies are off ?
> I wonder what the purpose if that shall be when systemd support is switched off anyways.
> 
> OTOH, I feel very unhappy with such arbitrary command calls, especially in case when systemd stuff had been disabled explicitly. Could end up calling a completely different program. Actually, I once used to write one with that name, long before systemd even existed.

The idea is to make it possible to use `runc` compiled without systemd support on a systemd-enabled system.

The `IsRunningSystemd` function does a simple `stat(2)`, and keeping it even when building with `no_systemd` won't affect the binary size much, nor will it bring any new dependencies.

But I don't think that calling `systemctl` will work as I thought it would -- first of all, we don't have a `unit` to set properties for.

So we can just assume that system won't screw up whatever we create in `/sys/fs/cgroup`.

kolyshkin:
Can we somehow limit the number of jobs? 60 is too much :)

Say, only test with non-default tags on 1 go version. Or something similar.

kolyshkin:
@metux can you please address https://github.com/opencontainers/runc/pull/2987#issuecomment-862837908, and also do a rebase?

kolyshkin:
@metux please let us know if you want to keep working on this. This needs a rebase, and https://github.com/opencontainers/runc/pull/2987#issuecomment-862837908 also needs to be addressed.

metux:
obsoleted by rewrite: #3959

PR #2988, opened 2021-06-02
Don't modify devices on update

crosbymichael:
You cannot modify devices with the update command so we should skip this update
as well.  It can conflict with other device management, mainly NVIDIA GPUS, when
update is called.

This does not need to be in the 1.0, its a long standing issue and can be in a `x.x.1`.  

Signed-off-by: Michael Crosby <michael@thepasture.io>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -225,6 +225,10 @@ func (c *linuxContainer) Set(config configs.Config) error {
 	if status == Stopped {
 		return newGenericError(errors.New("container not running"), ContainerNotRunning)
 	}
+	// Set is only called via the update calls and the update CLI does not support setting
+	// any device information.  It's better to skip devices on the update to not overwrite
+	// any out of band changes like NVIDIA gpu plugins.
+	config.Cgroups.SkipDevices = true
 	if err := c.cgroupManager.Set(config.Cgroups.Resources); err != nil {
 		// Set configs back
 		if err2 := c.cgroupManager.Set(c.config.Cgroups.Resources); err2 != nil {
@@ -244,6 +248,9 @@ func (c *linuxContainer) Set(config configs.Config) error {
 			return err
 		}
 	}
+	// reset SkipDevices before saving the state as this will cause major issues when
+	// trying to exec into the container or other process related actions.
+	config.Cgroups.SkipDevices = false
 	// After config setting succeed, update config and states
 	c.config = &config
 	_, err = c.updateState(nil)


libcontainer/container_linux.go
```diff
@@ -225,6 +225,10 @@ func (c *linuxContainer) Set(config configs.Config) error {
 	if status == Stopped {
 		return newGenericError(errors.New("container not running"), ContainerNotRunning)
 	}
+	// Set is only called via the update calls and the update CLI does not support setting
+	// any device information.  It's better to skip devices on the update to not overwrite
+	// any out of band changes like NVIDIA gpu plugins.
+	config.Cgroups.SkipDevices = true
```
kolyshkin, line 231: Looks way too hacky. Perhaps if we rename Set to Update, this will look less hacky.

Ideally, though, we should add something like a check that devices are already as they were set before, and skip the update in such a case.

libcontainer/container_linux.go
```diff
@@ -244,6 +248,9 @@ func (c *linuxContainer) Set(config configs.Config) error {
 			return err
 		}
 	}
+	// reset SkipDevices before saving the state as this will cause major issues when
+	// trying to exec into the container or other process related actions.
+	config.Cgroups.SkipDevices = false
```
kolyshkin, line 253: You're not resetting this on error.

I think, if it has to be done this way, using a `defer` will be more shatterproof:

```go
	config.Cgroups.SkipDevices = true
	defer func() {config.Cgroups.SkipDevices = false}
```

kolyshkin: Oops, I was wrong about `defer` 😊

libcontainer/container_linux.go
crosbymichael: well that is the problem.  With the nvidia device plugin or runtime, it modifies the devices outside of the spec.  Then if you call an update, it wipes it out.  That wouldn't be too big of an issue but devices are not specified or can be changed on `runc update`

kolyshkin: So maybe set SkipDevices right there in `update.go` (looks less hacky to me), and do not save it to the state file (as a "real" container started by runc can never has it set to `true`.

This is what I mean
```diff
diff --git a/libcontainer/configs/cgroup_linux.go b/libcontainer/configs/cgroup_linux.go
index 715ec1cb..e2472fe7 100644
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -127,8 +127,8 @@ type Resources struct {
 
        // SkipDevices allows to skip configuring device permissions.
        // Used by e.g. kubelet while creating a parent cgroup (kubepods)
-       // common for many containers.
+       // common for many containers, and also by runc update.
        //
        // NOTE it is impossible to start a container which has this flag set.
-       SkipDevices bool `json:"skip_devices"`
+       SkipDevices bool `json:"-"`
 }
diff --git a/update.go b/update.go
index ce89de08..7d5fba34 100644
--- a/update.go
+++ b/update.go
@@ -329,6 +329,11 @@ other options are ignored.
                        config.IntelRdt.MemBwSchema = memBwSchema
                }
 
+               // XXX(kolyshkin@): currently "runc update" is unable to change
+               // device configuration, so add this to skip device update.
+               // Note this field is not saved into container's state.json.
+               config.Cgroups.SkipDevices = true
+
                return container.Set(config)
        },
 }
```

Still looks like a kludge, but at least it is in more proper place.

// Cc @cyphar

h-vetinari: > With the nvidia device plugin or runtime, it modifies the devices outside of the spec.

Would it make sense to accommodate this usecase in the spec?

crosbymichael: @kolyshkin ok, that will work since we remove it from the state on disk.  I could have refactored more but didn't want to cause a big mess for such a small change at this point with the 1.0 being finalized.  I can make this change if you are good with it.  Then after we have the 1.0 out the door, we can look at a better refactor without just hacking it in.  1.0 is more important at this point and keeping the code from churning.

crosbymichael: @h-vetinari I think we will have to wait until the CDI work in finalized.  Doing these side band operations is hacky but we don't have a better solution at the moment that works across various runtimes.

kolyshkin: Opened https://github.com/opencontainers/runc/pull/2994 based on the diff above.

PR #2990, opened 2021-06-02
[WIP] ci/gha: redo the commit subject length checker

kolyshkin:
We were using https://github.com/tim-actions/commit-message-checker-with-regex
before, but it seems not very well maintained (v1 tag is older than
v0.3.1, and there's no way to file an issue).

Switch to gs-commit-message-checker which gained the ability to check
all commits in v1.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -89,19 +89,16 @@ jobs:
     # Only check commits on pull requests.
     if: github.event_name == 'pull_request'
     steps:
-      - name: get pr commits
-        id: 'get-pr-commits'
-        uses: tim-actions/get-pr-commits@v1.0.0
-        with:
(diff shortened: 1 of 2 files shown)


thaJeztah:
There's also https://github.com/vbatts/git-validation (not a github action in itself, but it's in Go, and I know the containerd projects use it in their actions; https://github.com/containerd/project/blob/main/.github/workflows/ci.yml

kolyshkin:
Not working :(

kolyshkin:
> There's also https://github.com/vbatts/git-validation

This is what we used before [we went to GHA], but it seems a tad too heavy to be used solely to check the commit subject length.

The current solution works, the only problem is `v1` in their repo, and the inability to open an issue about it.

PR #2992, opened 2021-06-03
tests: nits

kolyshkin:
I was cleaning up some of the old work and found some assorted test bits that make sense to have.

See commits for details.


AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #2994, opened 2021-06-03
runc update: skip devices

kolyshkin:
Since `runc update` CLI is not able to modify devices (and it is not clear
how to implement that properly), let's set `SkipDevices=true`,
so that a cgroup controller won't try to update devices cgroup.

This helps use cases when some other device management (NVIDIA GPUs)
applies its configuration on top of what runc does.

Make sure we do not save `SkipDevices` into `state.json`.

This is an alternative to https://github.com/opencontainers/runc/pull/2988.

## Proposed changelog entry
* `runc update` no longer updates cgroup devices configuration.

### DIFF
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -127,8 +127,8 @@ type Resources struct {
 
 	// SkipDevices allows to skip configuring device permissions.
 	// Used by e.g. kubelet while creating a parent cgroup (kubepods)
-	// common for many containers.
+	// common for many containers, and by runc update.
 	//
 	// NOTE it is impossible to start a container which has this flag set.
-	SkipDevices bool `json:"skip_devices"`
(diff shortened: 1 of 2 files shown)


cyphar (approved):
LGTM, I think dropping `SkipDevices` from the state is a good idea regardless of the original issue (and I prefer this to the other PR).

cyphar:
Though it should be noted this pretty much invalidates the `runc update` test I added in #2951 -- now `runc update` no longer triggers the cgroupv2 re-apply. So maybe it'd be better to remove the test since it's a complete no-op now (we can add a Go-based integration test later).

crosbymichael (approved):

kolyshkin:
Yes, I am going to remove the test and maybe try to add a unit test.

I'd rather do it separately from this PR though. Let's discuss it in there: https://github.com/opencontainers/runc/pull/3000

PR #2998, opened 2021-06-03
More linting fixes (with golint/revive enabled)

thaJeztah:
extracted from https://github.com/opencontainers/runc/pull/2966

See individual commits for details


thaJeztah:
Remaining issues are outlined in https://github.com/opencontainers/runc/pull/2966#issuecomment-854193373

thaJeztah:
@kolyshkin @AkihiroSuda ptal

libcontainer/cgroups/systemd/common.go
```diff
@@ -158,8 +160,7 @@ func findDeviceGroup(ruleType devices.Type, ruleMajor int64) (string, error) {
 	return "", nil
 }
 
-// DeviceAllow is the dbus type "a(ss)" which means we need a struct
-// to represent it in Go.
+// deviceAllowEntry represents the DeviceAllow dbus type "a(ss)" in Go.
```
kolyshkin, line 163: Hmm, your comment makes less sense than the original.

Perhaps
```suggestion
// deviceAllowEntry is a type representing an entry for DeviceAllow
// dbus property value, which has dbus type of "a(ss)".
```

libcontainer/cgroups/utils.go
```diff
@@ -24,6 +24,10 @@ import (
 const (
 	CgroupProcesses   = "cgroup.procs"
 	unifiedMountpoint = "/sys/fs/cgroup"
+
+	// maxShares corresponds with MAX_SHARES in the Linux kernel
```
kolyshkin, line 28: `s/with/to/`?

Also missing period at the end of the sentense.

libcontainer/cgroups/utils.go
```diff
@@ -399,16 +403,19 @@ func WriteCgroupProc(dir string, pid int) error {
...
 }
 
-// Since the OCI spec is designed for cgroup v1, in some cases
-// there is need to convert from the cgroup v1 configuration to cgroup v2
-// the formula for cpuShares is y = (1 + ((x - 2) * 9999) / 262142)
+// ConvertCPUSharesToCgroupV2Value converts a CPU shares cgroup v1 configuration
+// to cgroup v2.
+//
+// The OCI spec is designed for cgroup v1, in some cases there is need to
+// convert from the cgroup v1 configuration to cgroup v2 the formula for cpuShares
```
kolyshkin, line 410: The "in some cases" was a generic comment (as in "in some cases we need to convert from cgroup v1 to cgroup v2, and cpu shares is one such case").

The comment was talking about generic stuff first, when went to specifics of CPU shares.

I am not sure how to rephrase this in a good way if we must start with specifics (`ConvertCPUSharesToCgroupV2Value`). Maybe just skip the generic part?

```suggestion
// ConvertCPUSharesToCgroupV2Value converts a CPU shares value
// from cgroup v1 to cgroup v2.
//
// The OCI spec is designed for cgroup v1, in some cases there is need to
// convert from the cgroup v1 configuration to cgroup v2. The formula for cpuShares
// is y = (1 + ((x - 2) * 9999) / 262142), which is a conversion from
// [2-262144] to [1-10000].
```

or some such.

(note you did not separated the two sentences, the second one starts with "the formula for")

libcontainer/cgroups/utils.go
```diff
@@ -399,16 +403,19 @@ func WriteCgroupProc(dir string, pid int) error {
...
+// is y = (1 + ((x - 2) * 9999) / 262142)
 // convert from [2-262144] to [1-10000]
 // 262144 comes from Linux kernel definition "#define MAX_SHARES (1UL << 18)"
```
kolyshkin, line 413: This line needs to be removed as this is already explained earlier.

And again, the way it was written makes more sense -- we explain
that we convert from the range [A-B] to [C-D], and then explan
where the B comes from.

After the commit, the explanation for B is separated out, which
makes it harder to understand.

libcontainer/cgroups/utils.go
kolyshkin: OK I think the generic part (in some cases we need to convert) should be its own comment,
e.g.

```go
// The OCI spec is designed for cgroup v1, in some cases there is need to
// convert from the cgroup v1 configuration to cgroup v2. A few functions
// below are doing such conversion.

// ConvertCPUSharesToCgroupV2Value converts a CPU shares value
// from cgroup v1 to cgroup v2.
//
// The formula for CPU shares is ....
func ConvertCPUSharesToCgroupV2Value(cpuShares uint64) uint64 {
...

//  ConvertMemorySwapToCgroupV2Value converts MemorySwap value from OCI spec
// ...
func ConvertMemorySwapToCgroupV2Value(memorySwap, memory int64) (int64, error) {
...

// ConvertBlkIOToIOWeightValue converts a BlkIOWeight cgroup v1 value
// to cgroup v2.
//
// The formula for BlkIOWeight to IOWeight is y = (1 + (x - 10) * 9999 / 990),
// as we convert linearly from [10-1000] to [1-10000].
func ....
```
and so on

libcontainer/cgroups/v1_utils.go
```diff
@@ -23,7 +23,8 @@ const (
 )
 
 var (
-	errUnified     = errors.New("not implemented for cgroup v2 unified hierarchy")
+	errUnified = errors.New("not implemented for cgroup v2 unified hierarchy")
+	// ErrV1NoUnified is returned when trying to use unified cgroups.
```
kolyshkin, line 27:
```suggestion
// ErrV1NoUnified is returned when a "unified" config value is used in cgroup v1.
```

libcontainer/cgroups/v1_utils.go
```diff
@@ -110,7 +111,9 @@ func readCgroupMountinfo() ([]*mountinfo.Info, error) {
 	return cgroupMountinfo, readMountinfoErr
 }
 
-// https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
+// FindCgroupMountpoint returns the mountpoint for the given subsystem.
+//
+// see https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt.
```
kolyshkin, line 116: nit: `s/see/See/`

libcontainer/configs/cgroup_linux.go
```diff
@@ -5,8 +5,10 @@ import (
 	"github.com/opencontainers/runc/libcontainer/devices"
 )
 
+// FreezerState is a state of the freezer cgroup.
 type FreezerState string
 
+// FreezerState state constants.
```
kolyshkin, line 11:
```suggestion
// FreezerState values.
```

kolyshkin: or "Possible values for FreezerState".

libcontainer/configs/config.go
```diff
@@ -40,6 +40,7 @@ type Seccomp struct {
 // Action is taken upon rule match in Seccomp
 type Action int
 
+// Seccomp rule actions.
```
kolyshkin, line 43: Possible values for seccomp Action.

libcontainer/configs/devices.go
```diff
@@ -3,15 +3,19 @@ package configs
 import "github.com/opencontainers/runc/libcontainer/devices"
 
 type (
+	// Device is deprecated: use libcontainer/devices.Device instead.
 	// Deprecated: use libcontainer/devices.Device
```
kolyshkin, line 7: Hmm. We are saying the same thing twice, in a slightly different way.

I'd rather not add new documentation for whatever is deprecated.

libcontainer/system/proc.go
```diff
@@ -11,7 +11,9 @@ import (
 // State is the status of a process.
 type State rune
 
-const ( // Only values for Linux 3.14 and later are listed here
+// Process state values.
+// Only values for Linux 3.14 and later are listed here
```
kolyshkin, line 15: Nit: missing period.

libcontainer/system/proc.go
```diff
@@ -47,7 +49,7 @@ func (s State) String() string {
 // Stat_t represents the information from /proc/[pid]/stat, as
 // described in proc(5) with names based on the /proc/[pid]/status
 // fields.
-type Stat_t struct {
+type Stat_t struct { //nolint:revive // ignore "type Stat_t should be named StatT"
```
kolyshkin, line 52: Do we need to export this type at all? Looks like we don't. I have renamed it to `statT` and everything works.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -206,11 +204,11 @@ type intelRdtData struct {
 	pid    int
 }
 
-// Check if Intel RDT sub-features are enabled in featuresInit()
+// featuresInit checks if Intel RDT sub-features are enabled in.
```
kolyshkin, line 207: Hmm... unfinished sentence.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -315,17 +314,17 @@ func isIntelRdtMounted() bool {
 }
 
 type cpuInfoFlags struct {
-	CAT bool // Cache Allocation Technology
-	MBA bool // Memory Bandwidth Allocation
+	CAT bool // CAT indicates that Cache Allocation Technology is supported.
+	MBA bool // MBA indicates that Memory Bandwidth Allocation.
```
kolyshkin, line 318: missing "is supported"?

libcontainer/intelrdt/intelrdt.go
```diff
@@ -315,17 +314,17 @@ func isIntelRdtMounted() bool {
 }
 
 type cpuInfoFlags struct {
-	CAT bool // Cache Allocation Technology
-	MBA bool // Memory Bandwidth Allocation
+	CAT bool // CAT indicates that Cache Allocation Technology is supported.
```
kolyshkin, line 317: I'd rather say
```suggestion
	CAT bool // CAT indicates whether Cache Allocation Technology is supported.
```

libcontainer/intelrdt/intelrdt.go
```diff
@@ -315,17 +314,17 @@ func isIntelRdtMounted() bool {
...
+	MBA bool // MBA indicates that Memory Bandwidth Allocation.
 
 	// Memory Bandwidth Monitoring related.
 	MBMTotal bool
 	MBMLocal bool
 
-	CMT bool // Cache Monitoring Technology
+	CMT bool // CMT indicates that Cache Monitoring Technology
```
kolyshkin, line 324: missing "is supported"?

libcontainer/intelrdt/intelrdt.go
```diff
@@ -258,7 +256,8 @@ func featuresInit() {
 	})
 }
 
-// Return the mount point path of Intel RDT "resource control" filesysem
+// findIntelRdtMountpointDir returns the mount point path of Intel RDT "resource
```
kolyshkin, line 259: mount point is a path, so "mount point path" is probably a tautology.

kolyshkin:
I have only reviewed the first two commits.

My gut feeling is we should be more conservative about how much changes do we want. Not just making linters warnings go away, but think about readability and maintainability first. IOW, let's take linter warnings as hints, rather than be linter slaves.

Say, if changing a comment makes a linter silent, but reduces the comment readability, this is not a good change.

Another example -- it does not make sense to fix comments describing some variables or struct fields that are only used for one particular test.

It does not make sense to try to clean up documentation for something that is deprecated.

OTOH if the changes are like "Sets the values" -> "Set sets the values", I am totally for it. I am not against renames like Cpu -> CPU either.

libcontainer/cgroups/utils.go
thaJeztah: > `s/with/to/`?

Good question, I alway use "with", but "to" is also possible; https://english.stackexchange.com/questions/142495/correspond-to-vs-correspond-with

I'm fine with either, so let me update

libcontainer/configs/devices.go
thaJeztah: Perhaps we should remove these stubs now; they were included in the v1.0.0-rc95 tag, so _possible_ consumers of this would've (ideally) seen the deprecation.

libcontainer/system/proc.go
thaJeztah: The type is returned by an exported function; it's good practice to export the type if it's returned by an exported function.

I assumed we chose `Stat_t` as name here to match `syscall.Stat_t` / `unix.Stat_t`, but we could rename it to `StatT` to use Go conventions; looks like that's what we did in Moby

thaJeztah:
> It does not make sense to try to clean up documentation for something that is deprecated.

Addressed by https://github.com/opencontainers/runc/pull/2999, and rebased this one because that was merged.

libcontainer/system/proc.go
thaJeztah: Renamed to `StatT`

thaJeztah:
Rebased, PTAL

libcontainer/configs/namespaces_linux.go
```diff
@@ -6,6 +6,7 @@ import (
 	"sync"
 )
 
+// NamespaceType constants
```
kolyshkin, line 9: nit: missing period.

libcontainer/system/proc.go
```diff
@@ -44,10 +46,10 @@ func (s State) String() string {
 	}
 }
 
-// Stat_t represents the information from /proc/[pid]/stat, as
+// StatT represents the information from /proc/[pid]/stat, as
```
kolyshkin, line 49: I guess the name `Stat_t` was used as it resembles C / Unix `stat_t` type (which has `_t` suffix to distinguish it from `stat()` function), and is used here for the same reason (to distingush from Stat() function).

I like Stat_t much better, despite its non-conformance to Go standards. Note that golang.org/x/sys/unix (as well as stdlib's syscall package) also have names like `Stat_t` and `Statfs_t`.

kolyshkin: Surely we can just rename it to e.g. ProcStat or something.

libcontainer/intelrdt/cmt_test.go
```diff
@@ -13,10 +13,10 @@ func TestGetCMTNumaNodeStats(t *testing.T) {
 		"llc_occupancy": 9123911,
 	}
 
-	mockedL3_MON, err := mockResctrlL3_MON(mocksNUMANodesToCreate, mocksFilesToCreate)
```
kolyshkin, line 16: I think it's called L3_MON because there is a directory with the same name under /sys/fs/resctrl (`L3_MON`), and thus rename will make things harder to find. Not super important of course but I'd rather keep this as is.

https://github.com/opencontainers/runc/blob/3a041e9654169a301c19cfcaa52df2dc16115db2/libcontainer/intelrdt/intelrdt.go#L63

libcontainer/intelrdt/intelrdt.go
```diff
@@ -149,24 +149,24 @@ import (
  */
 
 type Manager interface {
-	// Applies Intel RDT configuration to the process with the specified pid
+	// Apply applies Intel RDT configuration to the process with the specified pid.
 	Apply(pid int) error
 
-	// Returns statistics for Intel RDT
+	// GetStats returns statistics for Intel RDT/
```
kolyshkin, line 155: nit: `s|/$|.|`

libcontainer/intelrdt/intelrdt.go
```diff
@@ -149,24 +149,24 @@ import (
...
+	// restore the object later.
 	GetPath() string
 
-	// Set Intel RDT "resource control" filesystem as configured.
+	// Set sets Intel RDT "resource control" filesystem as configured.
 	Set(container *configs.Config) error
 }
 
-// This implements interface Manager
+// intelRdtManager implements interface Manager
```
kolyshkin, line 169: nit: missing period.

kolyshkin:
Not quite sure we should rename all `ThisId` and `ThatId` to *ID. After all, `Id` is not an acronym, but short for Identifier.

Hmm, Wikipedia says it is. https://en.wikipedia.org/wiki/Identifier, so let it be.

thaJeztah:
I need to have a look at rebasing this one, to see which commits are still relevant

libcontainer/configs/namespaces_linux.go
kolyshkin: I am finally addressing this in https://github.com/opencontainers/runc/pull/3304, yay!

thaJeztah:
@kolyshkin I rebased this one

PR #2999, opened 2021-06-04
libcontainer: remove aliases for deprecated functions

thaJeztah:
These were deprecated and moved; the stubs were included in (at least) the last two (rc94, rc95) releases, so external consumers would have the chance to update their code.

Removing this so that this doesn't get into v1.0.0 GA


thaJeztah:
@kolyshkin @AkihiroSuda ptal

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #3000, opened 2021-06-04
libct/int: add device update test

kolyshkin:
... and remove the one from tests/integration.

The idea is similar to the one for the test case being removed -- try
updating device rules many times to make sure we are not leaking eBPF
programs after every update/Set(). This is better though as we can
really change the device rules every time (which "runc update" can't)
and check that the rule is applied.

The test works, but I see the following warning (on Fedora 34, cgroup v2):
```
WARN[0000] found more than one filter (2) attached to a cgroup -- removing extra filters! 
```

### DIFF
--- a/libcontainer/integration/update_test.go
+++ b/libcontainer/integration/update_test.go
@@ -1,0 +1,104 @@
+package integration
+
+import (
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/opencontainers/runc/libcontainer"
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
+	"github.com/opencontainers/runc/libcontainer/devices"
+)
+
+func testUpdateDevices(t *testing.T, systemd bool) {
+	if testing.Short() {
+		return
+	}
+	rootfs, err := newRootfs()
+	ok(t, err)
+	defer remove(rootfs)
+	config := newTemplateConfig(t, &tParam{
+		rootfs:  rootfs,
+		systemd: systemd,
+	})
+	container, err := newContainer(t, config)
+	ok(t, err)
+	defer destroyContainer(container)
+
+	// Execute a first process in the container
+	stdinR, stdinW, err := os.Pipe()
+	ok(t, err)
+	process := &libcontainer.Process{
+		Cwd:   "/",
+		Args:  []string{"cat"},
+		Env:   standardEnvironment,
+		Stdin: stdinR,
+		Init:  true,
+	}
+	err = container.Run(process)
+	_ = stdinR.Close()
+	defer func() {
+		_ = stdinW.Close()
+		if _, err := process.Wait(); err != nil {
+			t.Log(err)
+		}
+	}()
+	ok(t, err)
+
+	var buf bytes.Buffer
+	devCheck := &libcontainer.Process{
+		Cwd:    "/",
+		Args:   []string{"/bin/sh", "-c", "echo > /dev/full; cat /dev/null; true"},
+		Env:    standardEnvironment,
+		Stderr: &buf,
+	}
+	isAllowed := true
+	expected := map[bool][]string{
+		true: {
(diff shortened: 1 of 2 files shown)


kolyshkin:
Same in Fedora CI:
```
=== RUN   TestUpdateDevices
time="2021-06-04T22:28:40Z" level=warning msg="found more than one filter (2) attached to a cgroup -- removing extra filters!"
time="2021-06-04T22:28:40Z" level=warning msg="found more than one filter (2) attached to a cgroup -- removing extra filters!"
...
time="2021-06-04T22:28:54Z" level=warning msg="found more than one filter (2) attached to a cgroup -- removing extra filters!"
--- PASS: TestUpdateDevices (14.49s)
PASS
```

I changed the test like this:

```diff
diff --git a/libcontainer/integration/update_test.go b/libcontainer/integration/update_test.go
index 4a83ed41..0ec1dca1 100644
--- a/libcontainer/integration/update_test.go
+++ b/libcontainer/integration/update_test.go
@@ -69,14 +69,18 @@ func TestUpdateDevices(t *testing.T) {
 
                // Now flip the access permission
                isAllowed = !isAllowed
-               config.Cgroups.Resources.Devices = []*devices.Rule{
-                       {
-                               Type:        devices.CharDevice,
-                               Major:       1,
-                               Minor:       7,
-                               Permissions: "rwm",
-                               Allow:       isAllowed,
-                       },
+               if isAllowed {
+                       config.Cgroups.Resources.Devices = []*devices.Rule{
+                               {
+                                       Type:        devices.CharDevice,
+                                       Major:       1,
+                                       Minor:       7,
+                                       Permissions: "rwm",
+                                       Allow:       true,
+                               },
+                       }
+               } else {
+                       config.Cgroups.Resources.Devices = []*devices.Rule{}
                }
                if err := container.Set(*config); err != nil {
                        t.Fatal(err)
```

and it's still emitting the same warning.

kolyshkin:
Any way I modify the devices list, the warning is here.

Modified the test to
* add systemd
* add a second check (for /dev/null)
* instead of grant/remove access to just one device, alternate between default and empty lists

kolyshkin:
So I have tried the following ways of modifying device access:

1. Start from the default list, then use the list with a single /dev/full entry, alternating between `Allow: true` and `Allow: false`.
2. Start from the default list, then alternate between the empty list and the list with a single /dev/full entry.
2. Start from the default list, then alternate between empty and the default list.

The only method I haven't tried yet is modifying device permissions (say remove/re-add `w` from /dev/full`).

cyphar:
Can you cherry-pick the debugging info I added in #2986? Or maybe we should just merge that first since it does fix *a* bug (even if it doesn't fix the Docker CI issue that's blocking the 1.0.0 release)?

kolyshkin:
I think it's easier for you to cherry-pick this test to PR #2986 (or just merge this and then expand this test to cover more cases).

libcontainer/integration/update_test.go
```diff
@@ -0,0 +1,106 @@
...
+	defaultDevices := config.Cgroups.Resources.Devices
+
+	for i := 0; i < 300; i++ {
+		// Check the access
+		buf.Reset()
+		err = container.Run(devCheck)
+		ok(t, err)
+		waitProcess(devCheck, t)
+
+		t.Logf("[%d] allowed: %v output: %s", i, isAllowed, buf.String())
```
kolyshkin, line 76: This needs to be removed

kolyshkin: done

kolyshkin:
@cyphar let me know if you're OK to merge this as is, or is this test totally useless, or there is something else I should address.

cyphar (approved):
LGTM. This is a pretty good smoke test and actually exercises the "update" aspect as well, so device rule updates can't be a no-op either.

mrunalp (approved):

PR #3001, opened 2021-06-05
parseCgroupFromReader(): Ignore unified paths with no subsystem

thaJeztah:
When working on https://github.com/containerd/cgroups/pull/197, I noticed this change, which was not in the libcontainer variant of the same function. I thought I'd open a pull request in case this is relevant in this repository as well.

This ports the changes from containerd's cgroup package:
https://github.com/containerd/cgroups/commit/fe1947308d8f3fc3a7ef8996ba1da7b82de4e053 (https://github.com/containerd/cgroups/pull/21)

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -181,7 +181,9 @@ func parseCgroupFromReader(r io.Reader) (map[string]string, error) {
 		}
 
 		for _, subs := range strings.Split(parts[1], ",") {
-			cgroups[subs] = parts[2]
+			if subs != "" {
+				cgroups[subs] = parts[2]
+			}
 		}
 	}
 	if err := s.Err(); err != nil {
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -390,6 +390,30 @@ func TestParseCgroupString(t *testing.T) {
 	}
 }
 
+func TestEmptySubsystem(t *testing.T) {
+	const data = `10:devices:/user.slice
+ 	9:net_cls,net_prio:/
+ 	8:blkio:/
+ 	7:freezer:/
+ 	6:perf_event:/
+ 	5:cpuset:/
+ 	4:memory:/
+ 	3:pids:/user.slice/user-1000.slice/user@1000.service
+ 	2:cpu,cpuacct:/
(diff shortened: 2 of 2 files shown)


thaJeztah:
@AkihiroSuda @kolyshkin @crosbymichael ptal

libcontainer/cgroups/utils.go
```diff
@@ -181,7 +181,9 @@ func parseCgroupFromReader(r io.Reader) (map[string]string, error) {
 		}
 
 		for _, subs := range strings.Split(parts[1], ",") {
-			cgroups[subs] = parts[2]
+			if subs != "" {
+				cgroups[subs] = parts[2]
+			}
```
kolyshkin, line 186: Actually we rely on this functionality in cgroup v2, where the subsystem is empty.

kolyshkin: By "we" I mean not only runc itself, but also kubernetes. I was going to introduce a v2-specific function but this one does the job just fine, the only overhead is it returns a map with a a single element with empty key, rather than just a string.

kolyshkin: IOW, NACK

kolyshkin:
> This ports the changes from containerd's cgroup package:
containerd/cgroups@fe19473 (containerd/cgroups#21)

More to say, this change ^^^ needs to be reverted, if you rely on ParseCgroupsFile in your v2 code.

thaJeztah:
> More to say, this change ^^^ needs to be reverted, if you rely on ParseCgroupsFile in your v2 code.

Yes, I'm trying to make sense of it, so the containerd/cgroups package has a `v1` and `v2` packages in it.
Similar to libcontainer here, both also have a `parseCgroupFromReader()` one variant returns a single path, and one variant returns multiple (from comma-separated);

https://github.com/opencontainers/runc/blob/02b4b5f4daaf6ed2918fddec1dea0dea7e9a948f/libcontainer/cgroups/fs2/defaultpath.go#L84
https://github.com/opencontainers/runc/blob/02b4b5f4daaf6ed2918fddec1dea0dea7e9a948f/libcontainer/cgroups/utils.go#L166


From your comment, I understand that both v1 and v2 cgroups can have the comma-separated format? If v2 never has a comma-separated list; should all v2 code use the variant that returns a string instead?

PR #3002, opened 2021-06-07
Make DevicesGroup's "TestingSkipFinalCheck" attribute public

iholder-redhat:
Users would like to have the possibility to skip checks for their tests the same way they are skipped within the tests in runc.

Not exposing this variable makes it very hard to test components that use this library. To avoid copying-and-pasting the code into outside projects I suggest to expose this variable to the users.

### DIFF
--- a/libcontainer/cgroups/fs/devices.go
+++ b/libcontainer/cgroups/fs/devices.go
@@ -13,7 +13,7 @@ import (
 )
 
 type DevicesGroup struct {
-	testingSkipFinalCheck bool
+	TestingSkipFinalCheck bool
 }
 
 func (s *DevicesGroup) Name() string {
@@ -90,7 +90,7 @@ func (s *DevicesGroup) Set(path string, r *configs.Resources) error {
 	//
 	// This safety-check is skipped for the unit tests because we cannot
 	// currently mock devices.list correctly.
-	if !s.testingSkipFinalCheck {
+	if !s.TestingSkipFinalCheck {
 		currentAfter, err := loadEmulator(path)
 		if err != nil {
 			return err
--- a/libcontainer/cgroups/fs/devices_test.go
+++ b/libcontainer/cgroups/fs/devices_test.go
@@ -29,7 +29,7 @@ func TestDevicesSetAllow(t *testing.T) {
 		},
 	}
 
-	d := &DevicesGroup{testingSkipFinalCheck: true}
+	d := &DevicesGroup{TestingSkipFinalCheck: true}
 	if err := d.Set(path, r); err != nil {
 		t.Fatal(err)
 	}


cyphar:
Do you have an example of such users? My first worry is that this flag basically disables a security feature, so making it public makes a me little bit concerned (also the libcontainer interface in runc is not really intended to be used by downstream users -- though I guess this is a Kubernetes thing?).

iholder-redhat:
Hey @cyphar !

I'm such a user :)
We use your library in [Kubevirt](https://github.com/kubevirt/kubevirt).

We have certain components to deal with cgroups permissions etc'. To do so we use runc. The problem is that it's nearly impossible to have unit test for our components without copying your code into our repository.

What I would like to do is create a fake (mocked) path for our tests. My first problem was that the path is considered to be `not a cgroup file` at first. This makes sense in production but not for tests. I overcame this by using `chroot` so that the path will appear as if it's a regular valid cgroup path.

But now I'm stuck again.
Right after you [apply the new rule](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/devices.go#L84) (by adding it to `devices.allow` / `devices.deny`), [you check](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/devices.go#L100) that that the target emulator and current emulator are deeply equal, which [technically means](https://github.com/opencontainers/runc/blob/2f8e8e9d9775003bf76730a8a1c10f29b0000edf/libcontainer/cgroups/fs/devices.go#L38) that `devices.list` file is changed as expected.

Changing `devices.allow` won't effect `devices.list` since they are both mocked files, and therefore I fail over there. In order to being able to test this you made a `testingSkipFinalCheck` which is private. You even [wrote in comments](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/devices.go#L93):
```
// This safety-check is skipped for the unit tests because we cannot
// currently mock devices.list correctly.
```

Today we copy-paste some of your code because we have no other choice, but I think it's much better if we can simply avoid the checks when testing, the same we you do.

kolyshkin (dismissed):
LGTM

kolyshkin:
> My first worry is that this flag basically disables a security feature, so making it public makes a me little bit concerned

@cyphar there is no access to instance(s) of DeviceCgroup used by cgroup manager. So, unless one uses DeviceCgroup directly, they can't set this flag.

cyphar:
> > My first worry is that this flag basically disables a security feature, so making it public makes a me little bit concerned
> 
> @cyphar there is no access to instance(s) of DeviceCgroup used by cgroup manager. So, unless one uses DeviceCgroup directly, they can't set this flag.

Yeah that's a fair point. I guess it's okay if folks are using it for testing purposes, the only problem is it's non-trivial to enforce that it's only used for tests...

iholder-redhat:
> @iholder-redhat can you please rebase this (needed to run latest CI)

Done :)

kolyshkin (approved):
LGTM

Issue #2928, opened 2021-05-01
runc@master failing with Moby CI: TestTemplatedConfig: Options:[rbind ro]}: mount destination templated_config not absolute: unknown


AkihiroSuda:
```
--- FAIL: TestTemplatedConfig (61.70s)

[2021-05-01T17:09:25.901Z]     config_test.go:307: timeout hit after 1m0s: waiting for tasks to enter run state. task failed with error: starting container failed: failed to create shim: OCI runtime create failed: invalid mount {Destination:templated_config Type:bind Source:/go/src/github.com/docker/docker/bundles/test-integration/TestTemplatedConfig/d1fa52d447d48/root/containers/301097d0bae895342d137fd9217923474fddc25e6f06af4e3c3bd5a17cb5283c/mounts/secrets/l447q2mi60ayqntjg5bajt1hd Options:[rbind ro]}: mount destination templated_config not absolute: unknown
```

https://github.com/moby/moby/pull/42308
(moby/moby@59751bb23395a137253ca0b4fb99058ecec5a569 , runc @ d279ebd97d8832020e2c6f50cc3a11d0499a4690)

Regression in https://github.com/opencontainers/runc/pull/2917/commits/2192670a2430054c1e539ca7a3bf0ddc43b0bc3d  ( #2917 )
```
libct/configs/validate: validate mounts
    
Add a check that mount destination is absolute (as per OCI spec).
```

OCI spec: https://github.com/opencontainers/runtime-spec/blob/master/config.md#mounts

AkihiroSuda:
- Option 1: revert #2917 and violate OCI spec
- Option 2: revert #2917, and modify OCI spec to allow slash-less relative path
- Option 3: modify Moby to avoid using relative paths

Option 3 is the most ideal, but may face compatibility issues.

cc @thaJeztah @tiborvass

cyphar:
Yeah I agree option 3 is the best solution.

AkihiroSuda:
Let's close this and proceed the release process.

thaJeztah:
Generally, option 3 sounds good to me (actually surprised that this worked).

If we fix this in Moby/Docker, the only concern would be for existing versions of docker that would upgrade to a newer version of RunC.

Having a quick look, and it looks like regular bind-mounts already produce an error (on docker 20.10, runc rc93);

```bash
$ docker run -it --rm -w /app -v $(pwd):somewhere alpine sh
[...]

PR #3004, opened 2021-06-07
libcontainer: relax validation for absolute paths

thaJeztah:
Commits 1f1e91b1a09b5b22eb414ba7e5ad8c3534526216 and 2192670a2430054c1e539ca7a3bf0ddc43b0bc3d (https://github.com/opencontainers/runc/pull/2917) added validation for mountpoints to be an absolute path, to match the OCI specs.

Unfortunately, the old behavior (accepting the path to be a relative path) has been around for a long time, and although "not according to the spec", various higher level runtimes rely on this behavior.

While higher level runtime have been updated to address this requirement, there will be a transition period before all runtimes are updated to carry these fixes.

This patch relaxes the validation, to generate a WARNING instead of failing, allowing runtimes to update (but allowing them to update runc to the current version, which includes security fixes).

We can remove this exception in a future patch release.

relates to https://github.com/opencontainers/runc/issues/2928, https://github.com/containerd/containerd/issues/5547

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -29,6 +29,7 @@ type check func(config *configs.Config) error
 
 func (v *ConfigValidator) Validate(config *configs.Config) error {
 	checks := []check{
+		v.cgroups,
 		v.rootfs,
 		v.network,
 		v.hostname,
@@ -45,10 +46,6 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 			return err
 		}
 	}
-	if err := v.cgroups(config); err != nil {
-		return err
-	}
-
 	return nil
 }
 
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	selinux "github.com/opencontainers/selinux/go-selinux"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
@@ -39,11 +40,19 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 		v.sysctl,
 		v.intelrdt,
 		v.rootlessEUID,
-		v.mounts,
 	}
 	for _, c := range checks {
 		if err := c(config); err != nil {
 			return err
+		}
+	}
+	// Relaxed validation rules for backward compatibility
+	warns := []check{
+		v.mounts, // TODO (runc v1.x.x): make this an error instead of a warning
+	}
+	for _, c := range warns {
+		if err := c(config); err != nil {
+			logrus.WithError(err).Warnf("invalid configuration")
 		}
 	}
 	return nil
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -333,7 +333,10 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error) {
 	if !filepath.IsAbs(m.Destination) {
-		return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+		// Relax validation for backward compatibility
+		// TODO (runc v1.x.x): change warning to an error
+		// return nil, fmt.Errorf("mount destination %s is not absolute", m.Destination)
+		logrus.Warnf("mount destination %s is not absolute. Support for non-absolute mount destinations will be removed in a future release.", m.Destination)
 	}
 	flags, pgflags, data, ext := parseMountOptions(m.Options)
 	source := m.Source
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -323,10 +323,12 @@ func TestValidateMounts(t *testing.T) {
 		isErr bool
 		dest  string
 	}{
-		{isErr: true, dest: "not/an/abs/path"},
-		{isErr: true, dest: "./rel/path"},
-		{isErr: true, dest: "./rel/path"},
-		{isErr: true, dest: "../../path"},
+		// TODO (runc v1.x.x): make these relative paths an error. See https://github.com/opencontainers/runc/pull/3004
+		{isErr: false, dest: "not/an/abs/path"},
+		{isErr: false, dest: "./rel/path"},
+		{isErr: false, dest: "./rel/path"},
+		{isErr: false, dest: "../../path"},
+
 		{isErr: false, dest: "/abs/path"},
 		{isErr: false, dest: "/abs/but/../unclean"},
 	}


thaJeztah:
@kolyshkin @AkihiroSuda PTAL. Let me know if this is acceptable for v1.0.0 (we can remove it in a patch release after that)

AkihiroSuda:
AFAIK, only Docker Swarm-mode was known to have depended on the old behavior, and that was fixed in Docker 20.10.7: https://github.com/moby/moby/commit/afbb1277a3b74a9e6852c9cb222775f70b4a2d03

Is there any other project that depends on the old behavior?

thaJeztah:
BuildKit also needed changes, and (I think) for the swarm kit case at least, it's possible that existing services will break after updating (as they would have to be recreated for the new paths to be set); I'm wondering if there would be other situations where this could happen.

AkihiroSuda:
> BuildKit also needed changes,

Do we have an issue ticket?
BuildKit CI does not seem hitting the issue (for the master branch, at least) https://github.com/moby/buildkit/pull/2143

thaJeztah:
A fix was merged on master https://github.com/moby/buildkit/pull/2123, and back ported to the v0.8 branch (but no tagged release exists with that fix)

libcontainer/specconv/spec_linux.go
```diff
@@ -333,7 +333,14 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error) {
 	if !filepath.IsAbs(m.Destination) {
-		return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+		// Relax validation for backward compatibility
+		oldDest := m.Destination
+		var err error
+		m.Destination, err = filepath.Abs(m.Destination)
```
thaJeztah, line 339: can probably simplify this to `filepath.Join("/", ...)` (like on BuildKit); let me do so

libcontainer/specconv/spec_linux.go
```diff
@@ -333,7 +333,10 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error) {
 	if !filepath.IsAbs(m.Destination) {
-		return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+		// Relax validation for backward compatibility
+		oldDest := m.Destination
+		m.Destination = filepath.Join("/", m.Destination)
```
thaJeztah, line 338: Hm... actually; did it previously take `cwd` into account? i.e. does this have to be;

```suggestion
		m.Destination = filepath.Join("/", cwd, m.Destination)
```

kolyshkin: Yes, non-absolute paths are relative to CWD (and this is what makes them kinda dangerous -- thus the requirement in OCI spec for paths to be absolute I guess), and so `filepath.Join("/", ...)` is the same as `filepath.Abs()` with CWD=="/").

I like filepath.Abs() better despite it being dependent on cwd -- this is how all relative paths should be treated.

thaJeztah: I can change it back to `filepath.Abs()`

Hmmm.... but checking the old behaviour; looks like it does not take the working directory of the container into account, so paths were treated as "relative to `/` (inside the container)"

```bash
docker run --rm -w /somepath -v hello busybox ls -la
total 8
drwxr-xr-x    2 root     root          4096 Jun  7 21:56 .
drwxr-xr-x    1 root     root          4096 Jun  7 21:56 ..

docker run --rm -w /somepath -v hello busybox ls -la /
total 52
drwxr-xr-x    1 root     root          4096 Jun  7 21:56 .
drwxr-xr-x    1 root     root          4096 Jun  7 21:56 ..
-rwxr-xr-x    1 root     root             0 Jun  7 21:56 .dockerenv
drwxr-xr-x    2 root     root         12288 Apr  7 19:47 bin
drwxr-xr-x    5 root     root           340 Jun  7 21:56 dev
drwxr-xr-x    1 root     root          4096 Jun  7 21:56 etc
drwxr-xr-x    2 root     root          4096 Jun  7 21:56 hello
drwxr-xr-x    2 nobody   nobody        4096 Apr  7 19:47 home
dr-xr-xr-x  201 root     root             0 Jun  7 21:56 proc
drwx------    2 root     root          4096 Apr  7 19:47 root
drwxr-xr-x    2 root     root          4096 Jun  7 21:56 somepath
dr-xr-xr-x   13 root     root             0 Jun  7 21:56 sys
drwxrwxrwt    2 root     root          4096 Apr  7 19:47 tmp
drwxr-xr-x    3 root     root          4096 Apr  7 19:47 usr
drwxr-xr-x    4 root     root          4096 Apr  7 19:47 var
```

thaJeztah: Tried a build with `filepath.Abs`, and looks like that's giving the same result; let me know if you like that more (and if you want me to squash)

kolyshkin: You are right -- the non-absolute paths are treated as they are relative to the container root (as in `securejoin.SecureJoin(rootfs, m.Destination)`, where `rootfs` is container root, and `m.Destination` is what supposed to be an absolute path).

What I think we should do is
```diff
-                return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+               // TODO (runc 1.2.0): replace a warning below with
+               // return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+               logrus.Warnf("mount destination %s not absolute (this won't be supported past runc 1.2)", m.Destination)
```

and use the m.Destination as-is.

kolyshkin: In general it's a dark territory as to when we'll release runc 1.2.0, but I guess at least now for a year from now.

thaJeztah: > What I think we should do is

👍 on adding the comment, and yes, removing in 1.1.0 or 1.2.0 SGTM (hoping that users by then have upgraded to more current runtimes)

> and use the m.Destination as-is.

SGTM. I was thinking that migrating it in code would help future upgrades, but I guess the actual config for these containers would be managed by higher level runtimes, so not an issue.

thaJeztah: Updated with your suggestion

dims:
I like this approach. thanks @thaJeztah

libcontainer/specconv/spec_linux.go
```diff
@@ -333,7 +333,10 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 
 func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error) {
 	if !filepath.IsAbs(m.Destination) {
-		return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+		// Relax validation for backward compatibility
+		// TODO (runc 1.2.0): replace a warning below with
+		// return nil, fmt.Errorf("mount destination %s not absolute", m.Destination)
+		logrus.Warnf("mount destination %s not absolute (this won't be supported past runc 1.2.0)", m.Destination)
```
AkihiroSuda, line 339: We are still not sure how our post-v1 version scheme will look like.  
So "late 2021" might be better than "1.2.0"?

libcontainer/configs/validate/validator.go
```diff
@@ -29,6 +30,7 @@ type check func(config *configs.Config) error
 
 func (v *ConfigValidator) Validate(config *configs.Config) error {
 	checks := []check{
```
AkihiroSuda, line 32: `v.mounts` has to be removed from this slice

libcontainer/specconv/spec_linux.go
thaJeztah: Think we should make it more "generic"?

```suggestion
		logrus.Warnf("mount destination %s not absolute (this will no longer be supported in a future release)", m.Destination)
```

libcontainer/configs/validate/validator.go
thaJeztah: arf... 🤦 good catch

libcontainer/specconv/spec_linux.go
thaJeztah: Updated the description and TODO

libcontainer/configs/validate/validator.go
thaJeztah: Fixed

cyphar (changes requested):
You need to disable (`t.Skip` is fine, with a link to this PR or the linked issue) the `TestValidateMounts` test.

kolyshkin (approved):
LGTM

cyphar (approved):
LGTM.

PR #3005, opened 2021-06-07
Allow restoring with a different LSM profile

adrianreber:
Restoring an SELinux enabled container with Podman will result in a container with the exactly same SELinux process labels as during checkpointing. CRIU takes care of all the process labels.

Restoring multiple copies of a checkpointed container will result in all containers having the same SELinux process labels, which might be undesired.

When looking at Pods all container in a Pod share the process label of the infrastructure container. To restore a container into and existing Pod it is necessary to tell CRIU to restore the container with the infrastructure container process label.

CRIU supports setting different process labels using `--lsm-profile` for a long time and this just passes the process label information from runc to CRIU.

Unfortunately CRIU has a bug as no one was using the `--lsm-profile` option so this changes requires the upcoming CRIU version 3.16.

See: https://github.com/checkpoint-restore/criu/pull/1486

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -1375,6 +1375,15 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 		},
 	}
 
+	if criuOpts.LsmProfile != "" {
+		// CRIU older than 3.16 has a bug which breaks the possibility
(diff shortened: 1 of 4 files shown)


kolyshkin (approved):
LGTM

AkihiroSuda (approved):

PR #2407, opened 2020-05-15
Update blkio GetStats

piowag:
Updates blkio GetStats, as proposed in [2365](https://github.com/opencontainers/runc/issues/2365)

Signed-off-by: Piotr Wagner <piotr.wagner@intel.com>

PR #3010, opened 2021-06-08
libcontainer/cgroups/fs/blkio: support BFQ weight[_device]

askervin:
- Update the blkio cgroup to support the BFQ I/O Scheduler, that has
  replaced CFQ in the Linux kernel.
- BFQ is controlled through blkio.bfq.weight[_device] instead of
  CFQ's blkio.weight[_device] in cgroups v1.
- BFQ does not support blkio.leaf_weight[_device], so that behavior
  remains untouched.
- Do not change behavior on legacy CFQ systems.
- Enable using blkio weights on BFQ systems.

Signed-off-by: Antti Kervinen <antti.kervinen@intel.com>

### DIFF
--- a/libcontainer/cgroups/fs/blkio_test.go
+++ b/libcontainer/cgroups/fs/blkio_test.go
@@ -171,62 +171,80 @@ func appendBlkioStatEntry(blkioStatEntries *[]cgroups.BlkioStatEntry, major, min
 }
 
 func TestBlkioSetWeight(t *testing.T) {
-	helper := NewCgroupTestUtil("blkio", t)
-	defer helper.cleanup()
-
 	const (
 		weightBefore = 100
 		weightAfter  = 200
 	)
 
-	helper.writeFileContents(map[string]string{
-		"blkio.weight": strconv.Itoa(weightBefore),
-	})
-
-	helper.CgroupData.config.Resources.BlkioWeight = weightAfter
-	blkio := &BlkioGroup{}
-	if err := blkio.Set(helper.CgroupPath, helper.CgroupData.config.Resources); err != nil {
-		t.Fatal(err)
-	}
-
-	value, err := fscommon.GetCgroupParamUint(helper.CgroupPath, "blkio.weight")
-	if err != nil {
-		t.Fatalf("Failed to parse blkio.weight - %s", err)
-	}
-
-	if value != weightAfter {
-		t.Fatal("Got the wrong value, set blkio.weight failed.")
+	for _, legacyIOScheduler := range []bool{false, true} {
+		// Populate cgroup
+		helper := NewCgroupTestUtil("blkio", t)
+		defer helper.cleanup()
+		weightFilename := "blkio.bfq.weight"
+		if legacyIOScheduler {
+			weightFilename = "blkio.weight"
+		}
+		helper.writeFileContents(map[string]string{
+			weightFilename: strconv.Itoa(weightBefore),
+		})
+		// Apply new configuration
+		helper.CgroupData.config.Resources.BlkioWeight = weightAfter
+		blkio := &BlkioGroup{}
+		if err := blkio.Set(helper.CgroupPath, helper.CgroupData.config.Resources); err != nil {
+			t.Fatal(err)
+		}
+		// Verify results
+		if weightFilename != blkio.weightFilename {
+			t.Fatalf("weight filename detection failed: expected %q, detected %q", weightFilename, blkio.weightFilename)
+		}
+		value, err := fscommon.GetCgroupParamUint(helper.CgroupPath, weightFilename)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if value != weightAfter {
+			t.Fatalf("Got the wrong value, set %s failed.", weightFilename)
+		}
 	}
 }
 
 func TestBlkioSetWeightDevice(t *testing.T) {
-	helper := NewCgroupTestUtil("blkio", t)
-	defer helper.cleanup()
-
 	const (
 		weightDeviceBefore = "8:0 400"
 	)
 
-	wd := configs.NewWeightDevice(8, 0, 500, 0)
-	weightDeviceAfter := wd.WeightString()
-
-	helper.writeFileContents(map[string]string{
-		"blkio.weight_device": weightDeviceBefore,
-	})
-
-	helper.CgroupData.config.Resources.BlkioWeightDevice = []*configs.WeightDevice{wd}
-	blkio := &BlkioGroup{}
-	if err := blkio.Set(helper.CgroupPath, helper.CgroupData.config.Resources); err != nil {
-		t.Fatal(err)
-	}
-
-	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.weight_device")
-	if err != nil {
-		t.Fatalf("Failed to parse blkio.weight_device - %s", err)
-	}
-
-	if value != weightDeviceAfter {
-		t.Fatal("Got the wrong value, set blkio.weight_device failed.")
+	for _, legacyIOScheduler := range []bool{false, true} {
+		// Populate cgroup
+		helper := NewCgroupTestUtil("blkio", t)
+		defer helper.cleanup()
+		weightFilename := "blkio.bfq.weight"
+		weightDeviceFilename := "blkio.bfq.weight_device"
+		if legacyIOScheduler {
+			weightFilename = "blkio.weight"
+			weightDeviceFilename = "blkio.weight_device"
+		}
+		helper.writeFileContents(map[string]string{
+			weightFilename:       "",
+			weightDeviceFilename: weightDeviceBefore,
+		})
+		// Apply new configuration
+		wd := configs.NewWeightDevice(8, 0, 500, 0)
+		weightDeviceAfter := wd.WeightString()
+		helper.CgroupData.config.Resources.BlkioWeightDevice = []*configs.WeightDevice{wd}
+		blkio := &BlkioGroup{}
+		if err := blkio.Set(helper.CgroupPath, helper.CgroupData.config.Resources); err != nil {
+			t.Fatal(err)
+		}
+		// Verify results
+		if weightDeviceFilename != blkio.weightDeviceFilename {
+			t.Fatalf("weight_device filename detection failed: expected %q, detected %q", weightDeviceFilename, blkio.weightDeviceFilename)
+		}
+		value, err := fscommon.GetCgroupParamString(helper.CgroupPath, weightDeviceFilename)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if value != weightDeviceAfter {
+			t.Fatalf("Got the wrong value, set %s failed.", weightDeviceFilename)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


askervin:
Hi @AkihiroSuda,

There's something essential that I should have brought up earlier:

1. CFQ was removed from Linux kernel (commit f382fb0bcef4, "block: remove legacy IO schedulers") 10/2018.
2. BFQ replaced CFQ in RHEL 8, released 5/2019.
3. BFQ replaced CFQ in Ubuntu 19.10, released 10/2019.

That is, this patch will fix `runc` on Linux running less than 3-years-old kernel. Therefore I'm hoping this patch would be considered to be *a fix* rather than *an enhancement*.

Without this patch adjusting `weight/weight_device` parameters in blockio PRs for [CRI-O](https://github.com/cri-o/cri-o/pull/4873) and [containerd](https://github.com/containerd/containerd/pull/5490) will prevent creating containers on any up-to-date Linux:

```
# kubectl describe pod pod1
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
   ...
  Normal   Created    61s               kubelet            Created container pod1c1
  Normal   Pulled     61s               kubelet            Container image "busybox" already present on machine                                                        
  Warning  Failed     60s               kubelet            Error: container create failed: time="2021-06-09T11:25:57Z" level=error msg="container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: process_linux.go:458: setting cgroup config for procHooks process caused: open /sys/fs/cgroup/blkio/kubepods.slice/kubepodspodc83ca3da_5552_4fd0_8e17_e62c417b3018.slice/crio-f28b255e00529312c9bd177b3186778b2b8f1369533c9f80a8d48df1455ba7bf.scope/blkio.weight: no such file or directory"
```

libcontainer/cgroups/fscommon/fscommon.go
```diff
@@ -39,6 +40,14 @@ func ReadFile(dir, file string) (string, error) {
 	return buf.String(), err
 }
 
+// FileExists returns true if a file exists in dir, otherwise false.
+func FileExists(dir, file string) bool {
+	if _, err := os.Stat(path.Join(dir, file)); err == nil {
+		return true
+	}
+	return false
```
AkihiroSuda, line 48: Should return true if !errors.Is(err, os.ErrNotExist)

askervin: Thanks, definitely so! Fixed.

AkihiroSuda (dismissed):

libcontainer/cgroups/fscommon/fscommon.go
```diff
@@ -39,6 +40,14 @@ func ReadFile(dir, file string) (string, error) {
 	return buf.String(), err
 }
 
+// FileExists returns true if a file exists in dir, otherwise false.
+func FileExists(dir, file string) bool {
```
kolyshkin, line 44: We have used `cgroups.PathExists` for similar cases before, see e.g. https://github.com/opencontainers/runc/blob/dcdf6b631da266153abba1e48493ffc218cf554b/libcontainer/cgroups/fs/memory.go#L74-L75

libcontainer/cgroups/fs/blkio_test.go
```diff
@@ -210,23 +211,24 @@ func TestBlkioSetWeightDevice(t *testing.T) {
...
-	blkio := &BlkioGroup{}
 	if err := blkio.Set(helper.CgroupPath, helper.CgroupData.config.Resources); err != nil {
 		t.Fatal(err)
 	}
 
-	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.weight_device")
+	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, blkio.weightDeviceFilename)
 	if err != nil {
-		t.Fatalf("Failed to parse blkio.weight_device - %s", err)
+		t.Fatalf("Failed to parse %s - %s", blkio.weightDeviceFilename, err)
```
kolyshkin, line 227: So, I am working on that in https://github.com/opencontainers/runc/pull/3011 but if you want to modify this, I'd rather do something like
```suggestion
		t.Fatal(err)
```
because the error message from `fscommon.Get*` functions already contains enough context (filename).

In fact you can borrow the whole or a part of https://github.com/opencontainers/runc/pull/3011/commits/ac5ca790f10974c0dcb2563fea071abdf7fce3e3, put it before your commit so it becomes shorter.

kolyshkin:
Interesting... we support getting stats from BFQ since PR #2407, but never got to set the parameters for it :)

libcontainer/cgroups/fscommon/fscommon.go
askervin: Thanks! Switched to use `cgroups.PathExists` and removed `fscommon.FileExists`. Did not touch PathExists contents, yet @AkihiroSuda's [previous comment](https://github.com/opencontainers/runc/pull/3010#discussion_r648216865) seemed perfectly valid.

libcontainer/cgroups/fs/blkio_test.go
askervin: Switched to `t.Fatal(err)`'s in both cases that were touched to align with your work. Tested them both with deliberately wrong filenames. Error messages look good, as you say.
```
--- FAIL: TestBlkioSetWeightDevice (0.00s)
blkio_test.go:227: open /tmp/cgroup_test459287435/blkio/blkio.bfq.weight_devicey: no such file or directory
```

askervin:
> I am looking at a similar code for cgroup v2. In there, we also have a fallback (implemented in commit [8c7ece1](https://github.com/opencontainers/runc/commit/8c7ece1e6d1aba12c85cffa78f814042de262b8a)). There's a weight conversion needed though. Do we need some kind of conversion here as well?

Good question. The situation is:
- CFQ blkio.weight has range [10..1000], the default is 500.
- BFQ blkio.bfq.weight has range [1..1000], the default is 100.
- We do not have information on which I/O scheduler the user/admin thinks is under the hood.

I considered two options:

1. A full-range conversion (linear, or non-linear due to different defaults). I believe that converting values on our code only adds complexity and is likely to result in bug reports like _I tried to set weight "100", but got "500" or "101"_.

2. Convert values < 10 to 10 if running CFQ. After all, even in this case I'd prefer `Numerical result out of range` error. It reveals that the system is running an unexpected I/O scheduler. Either the system needs reconfiguration or weights that the user/admin is using should be recalibrated. I think errors like this should not be hidden.

> It seems that for BFQ we need kernel 5.x, but since this is cgroup v1, our Fedora is not good. Do we need to add a recent Debian or Ubuntu-based test? Or reboot the same Fedora image with cgroup v1 (modify kernel parameters with something like `grubby --args="systemd.unified_cgroup_hierarchy=0" --update-kernel $(grubby --default-kernel)`, reboot, wait, repeat all the tests with cgroup v1).

Maybe adding a new platform to run all tests would be a good topic for another PR?

kolyshkin:
@askervin thank you for a detailed reply! I agree that in this case no conversion is probably the best thing to do.

I also agree that testing can be discussed separately -- but without the test we can't be sure if the whole thing works or not (even if there is not too much code).

libcontainer/cgroups/fs/blkio_test.go
```diff
@@ -210,23 +211,24 @@ func TestBlkioSetWeightDevice(t *testing.T) {
 	wd := configs.NewWeightDevice(8, 0, 500, 0)
 	weightDeviceAfter := wd.WeightString()
 
+	blkio := &BlkioGroup{}
+	blkio.detectWeightFilenames(helper.CgroupPath)
```
kolyshkin, line 215: Please note we are not using real cgroupfs here, but a mock of it, so this code won't work the same way (instead, if you pre-create `blkio.weight`, it will be used, otherwise BFQ). Maybe it makes sense to test both scenarios (with and without `blkio.weight` pre-created by the test).

libcontainer/cgroups/fs/blkio_test.go
```diff
@@ -179,23 +179,24 @@ func TestBlkioSetWeight(t *testing.T) {
 		weightAfter  = 200
 	)
 
+	blkio := &BlkioGroup{}
+	blkio.detectWeightFilenames(helper.CgroupPath)
```
kolyshkin, line 183: Same as https://github.com/opencontainers/runc/pull/3010/commits/feceacc10800376cfc9e5bb9630134cfb6e9d6a7#r649557576

libcontainer/cgroups/fs/blkio_test.go
```diff
@@ -247,23 +249,24 @@ func TestBlkioSetMultipleWeightDevice(t *testing.T) {
 	// is present will suffice for the test to ensure multiple writes are done.
 	weightDeviceAfter := wd2.WeightString()
 
+	blkio := &BlkioGroup{}
+	blkio.detectWeightFilenames(helper.CgroupPath)
```
kolyshkin, line 253: Same as https://github.com/opencontainers/runc/pull/3010/commits/feceacc10800376cfc9e5bb9630134cfb6e9d6a7#r649557576

libcontainer/cgroups/fs/blkio_test.go
askervin: Fixed.

libcontainer/cgroups/fs/blkio_test.go
askervin: Fixed.

libcontainer/cgroups/fs/blkio_test.go
askervin: Fixed.

askervin:
@kolyshkin , Thanks for your feedback on unit tests, too!

I updated these tests so that they cover
1. detecting the `weight` file with BFQ and CFQ cases (`legacyIOScheduler` `false` and `true`, respectively).
2. detecting the `weight_device` file with BFQ and CFQ cases
3. detecting files when `blkio` controller is disabled: no weight files in the cgroup. Should try to use (and in real case fail using) BFQ files in this case.

AkihiroSuda (approved):

kolyshkin (approved):
LGTM, and thanks!

kolyshkin:
Ah, this is why I wanted a test case:

```
not ok 9 runc run (per-device io weight)
# (in test file tests/integration/cgroups.bats, line 173)
# runc spec (status=0):
# 
# 16+0 records in
# 16+0 records out
# 65536 bytes (66 kB, 64 KiB) copied, 0.000114909 s, 570 MB/s
# runc run -d --console-socket /tmp/bats-run-14673/runc.axD6tf/tty/sock test_dev_weight (status=1):
# time="2021-06-14T00:54:57Z" level=error msg="container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: process_linux.go:508: setting cgroup config for procHooks process caused: openat2 /sys/fs/cgroup/blkio/system.slice/runner-provisioner.service/test_dev_weight/blkio.leaf_weight_device: no such file or directory"
```

Fix is coming

kolyshkin:
The fix is part of #3022 https://github.com/opencontainers/runc/pull/3022/commits/00b8e313cd925ed6be106106c3d814dc22bb519e

PR #3011, opened 2021-06-09
Improve error reporting, add errorlint, fix its warnings; drop pkg/errors

kolyshkin:
This commit does a few things related to error reporting and handling.

1. Fix and improve a bunch of error messages and associated error handling,
using features that appear in Go 1.13 (`fmt.Errorf` with `%w` to wrap errors,
`errors.Is` and `errors.As` for unwrapping) and some other methods
(such as using `os.PathError`, introducing and using `ParseError` etc.)

2. Fix all `errorlint` linter warnings, enable errorlint it in CI.

3. Remove all uses of pkg/errors, switching to native Go error (un)wrapping.

~~TODO (not in this PR, as it's already too big):~~
- modernize and simplify `libcontainer/error` (done in #3033)

### DIFF
--- a/libcontainer/cgroups/fs/blkio_test.go
+++ b/libcontainer/cgroups/fs/blkio_test.go
@@ -283,7 +283,6 @@ func TestBlkioSetMultipleWeightDevice(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
 	if value != weightDeviceAfter {
 		t.Fatalf("Got the wrong value, set %s failed.", blkio.weightDeviceFilename)
 	}
@@ -772,9 +771,8 @@ func TestBlkioSetThrottleReadBpsDevice(t *testing.T) {
 
 	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.throttle.read_bps_device")
 	if err != nil {
-		t.Fatalf("Failed to parse blkio.throttle.read_bps_device - %s", err)
+		t.Fatal(err)
 	}
-
 	if value != throttleAfter {
 		t.Fatal("Got the wrong value, set blkio.throttle.read_bps_device failed.")
 	}
@@ -803,9 +801,8 @@ func TestBlkioSetThrottleWriteBpsDevice(t *testing.T) {
 
 	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.throttle.write_bps_device")
 	if err != nil {
-		t.Fatalf("Failed to parse blkio.throttle.write_bps_device - %s", err)
+		t.Fatal(err)
 	}
-
 	if value != throttleAfter {
 		t.Fatal("Got the wrong value, set blkio.throttle.write_bps_device failed.")
 	}
@@ -834,9 +831,8 @@ func TestBlkioSetThrottleReadIOpsDevice(t *testing.T) {
 
 	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.throttle.read_iops_device")
 	if err != nil {
-		t.Fatalf("Failed to parse blkio.throttle.read_iops_device - %s", err)
+		t.Fatal(err)
 	}
-
 	if value != throttleAfter {
 		t.Fatal("Got the wrong value, set blkio.throttle.read_iops_device failed.")
 	}
@@ -865,9 +861,8 @@ func TestBlkioSetThrottleWriteIOpsDevice(t *testing.T) {
 
 	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.throttle.write_iops_device")
 	if err != nil {
-		t.Fatalf("Failed to parse blkio.throttle.write_iops_device - %s", err)
+		t.Fatal(err)
 	}
-
 	if value != throttleAfter {
 		t.Fatal("Got the wrong value, set blkio.throttle.write_iops_device failed.")
 	}
--- a/libcontainer/cgroups/fs/error.go
+++ b/libcontainer/cgroups/fs/error.go
@@ -1,0 +1,15 @@
+package fs
+
+import (
+	"fmt"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
+)
+
+type parseError = fscommon.ParseError
+
+// malformedLine is used by all cgroupfs file parsers that expect a line
+// in a particular format but get some garbage instead.
+func malformedLine(path, file, line string) error {
+	return &parseError{Path: path, File: file, Err: fmt.Errorf("malformed line: %s", line)}
+}
--- a/libcontainer/cgroups/fs/pids.go
+++ b/libcontainer/cgroups/fs/pids.go
@@ -3,8 +3,7 @@
 package fs
 
 import (
-	"fmt"
-	"path/filepath"
+	"math"
 	"strconv"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -45,21 +44,18 @@ func (s *PidsGroup) GetStats(path string, stats *cgroups.Stats) error {
 	}
 	current, err := fscommon.GetCgroupParamUint(path, "pids.current")
 	if err != nil {
-		return fmt.Errorf("failed to parse pids.current - %s", err)
+		return err
 	}
 
-	maxString, err := fscommon.GetCgroupParamString(path, "pids.max")
+	max, err := fscommon.GetCgroupParamUint(path, "pids.max")
 	if err != nil {
-		return fmt.Errorf("failed to parse pids.max - %s", err)
+		return err
 	}
-
-	// Default if pids.max == "max" is 0 -- which represents "no limit".
-	var max uint64
-	if maxString != "max" {
-		max, err = fscommon.ParseUint(maxString, 10, 64)
-		if err != nil {
-			return fmt.Errorf("failed to parse pids.max - unable to parse %q as a uint from Cgroup file %q", maxString, filepath.Join(path, "pids.max"))
-		}
+	// If no limit is set, read from pids.max returns "max", which is
+	// converted to MaxUint64 by GetCgroupParamUint. Historically, we
+	// represent "no limit" for pids as 0, thus this conversion.
+	if max == math.MaxUint64 {
+		max = 0
 	}
 
 	stats.PidsStats.Current = current
--- a/libcontainer/cgroups/fs2/pids.go
+++ b/libcontainer/cgroups/fs2/pids.go
@@ -3,8 +3,8 @@
 package fs2
 
 import (
+	"math"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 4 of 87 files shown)


kolyshkin:
Hmm, weird failure in `update devices [minimal transition rules]` test; filed https://github.com/opencontainers/runc/issues/3014; CI restarted.

libcontainer/restored_process.go
```diff
@@ -89,19 +90,19 @@ type nonChildProcess struct {
...
+	return newGenericError(errors.New("restored process cannot be started"), SystemError)
 }
 
 func (p *nonChildProcess) pid() int {
 	return p.processPid
 }
 
 func (p *nonChildProcess) terminate() error {
-	return newGenericError(fmt.Errorf("restored process cannot be terminated"), SystemError)
+	return newGenericError(errors.New("restored process cannot be terminated"), SystemError)
```
thaJeztah, line 101: It would be great if we got rid of these `newGenericError()` etc. Afaics, these were created to provide the stack as part of the error, but `github.com/pkg/errors` provides the same.

kolyshkin: Yes, I agree, see my TODO item in this PR description above.

I took a quick look as to how dismantle all this and it seems it's easier to do it in a separate PR (once this one is merged).

kolyshkin:
> Hmm, weird failure in update devices [minimal transition rules] test; filed #3014; CI restarted.

OK it seems that this error is repeatable. Kicked CI again, created https://github.com/opencontainers/runc/pull/3015 to investigate if this is indeed caused by this PR.

kolyshkin:
Rebased; set milestone to 1.1

kolyshkin:
Added commits that switch from `pkg/error` to Go native error (un)wrapping.

libcontainer/restored_process.go
cyphar: Yeah `newGenericError` came from a time before `pkg/error` existed and to be honest I've always hated the `genericError` error messages. But yeah, let's do it in a separate PR.

kolyshkin:
Rebased on top of merged #3024

libcontainer/cgroups/fscommon/utils.go
```diff
@@ -23,6 +23,19 @@ var (
...
 
+// ParseError records a parse error details, including the file path.
+type ParseError struct {
+	Path string
+	File string
+	Err  error
+}
+
+func (e *ParseError) Error() string {
+	return "unable to parse " + e.Path + "/" + e.File + ": " + e.Err.Error()
```
kolyshkin, line 31: Maybe use path.Join() here.

libcontainer/cgroups/fs/error.go
```diff
@@ -0,0 +1,15 @@
+package fs
+
+import (
+	"fmt"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
+)
+
+type ParseError = fscommon.ParseError
```
kolyshkin, line 9: Doesn't have to be public, i.e. `parseError` should be used.

libcontainer/cgroups/fs/error.go
```diff
@@ -0,0 +1,15 @@
...
+
+// malformedLine is used by all cgroupfs file parsers that expect a line
+// in a particular format but get some garbage instead.
+func malformedLine(path, file, line string) error {
```
kolyshkin, line 13: Shouldn't this be in fscommon?

kolyshkin: For now this is specific to cgroups/fs only -- no cgroups/fs2 parses returns an error like this. For some reason fs2 parsers all more forgiving -- they skip, rather than error on, unknown lines.

If there will be a need we can move this to fscommon later.

libcontainer/rootfs_linux.go
```diff
@@ -188,10 +189,10 @@ func prepareTmp(topTmpDir string) (string, error) {
 		return "", err
 	}
 	if err := unix.Mount(tmpdir, tmpdir, "bind", unix.MS_BIND, ""); err != nil {
-		return "", err
+		return "", &os.PathError{Op: "bind-mount", Path: tmpdir, Err: err}
```
kolyshkin, line 192: Perhaps it makes sense to introduce a special `mountError`, similar to one from github.com/moby/sys/mount (unfortunately we can't use moby/sys/mount directly as it's too high-level).

kolyshkin:
Working on addressing my own review comments :)

kolyshkin:
(my own) review comments addressed; no longer WIP.

AkihiroSuda (approved):

libcontainer/cgroups/fscommon/utils.go
kolyshkin: Done

libcontainer/cgroups/fs/error.go
kolyshkin: Fixed

libcontainer/rootfs_linux.go
kolyshkin: Done

libcontainer/cgroups/fs/pids.go
```diff
@@ -45,21 +44,18 @@ func (s *PidsGroup) GetStats(path string, stats *cgroups.Stats) error {
...
-	if maxString != "max" {
-		max, err = fscommon.ParseUint(maxString, 10, 64)
-		if err != nil {
-			return fmt.Errorf("failed to parse pids.max - unable to parse %q as a uint from Cgroup file %q", maxString, filepath.Join(path, "pids.max"))
-		}
+	// If no limit is set, read from pids.max returns "max", which is
+	// converted to MaxUint64 by GetCgroupParamUint. Historically, we
+	// represent "no limit" for pids as 0, thus this conversion.
+	if max == math.MaxUint64 {
+		max = 0
```
cyphar, line 58: This hunk isn't right -- the `pids.max` file contains the string `max` if its in unlimited mode and such a string will fail to parse here (the `math.MaxUint64` fallback also isn't necessary I believe).

libcontainer/cgroups/fs2/pids.go
```diff
@@ -53,22 +54,18 @@ func statPids(dirPath string, stats *cgroups.Stats) error {
...
-		if err != nil {
-			return errors.Wrapf(err, "failed to parse pids.max - unable to parse %q as a uint from Cgroup file %q",
-				maxString, filepath.Join(dirPath, "pids.max"))
-		}
+	// If no limit is set, read from pids.max returns "max", which is
+	// converted to MaxUint64 by GetCgroupParamUint. Historically, we
+	// represent "no limit" for pids as 0, thus this conversion.
+	if max == math.MaxUint64 {
+		max = 0
 	}
```
cyphar, line 69: Same comment with regards to `pids.max`.

libcontainer/container_linux.go
```diff
@@ -666,7 +665,7 @@ func (c *linuxContainer) Resume() error {
 		return err
 	}
 	if status != Paused {
-		return newGenericError(fmt.Errorf("container not paused"), ContainerNotPaused)
+		return newGenericError(errors.New("container not paused"), ContainerNotPaused)
```
cyphar, line 668: I'd also like to get rid of `genericError` as well but I guess that'll need to wait for the next PR?

cyphar (changes requested):
One nit, two things need changing, and one question.

libcontainer/container_linux.go
cyphar: Ah that's what #3033 is for -- I got the two PRs confused.

libcontainer/cgroups/fs/blkio_test.go
```diff
@@ -772,9 +771,8 @@ func TestBlkioSetThrottleReadBpsDevice(t *testing.T) {
 
 	value, err := fscommon.GetCgroupParamString(helper.CgroupPath, "blkio.throttle.read_bps_device")
 	if err != nil {
-		t.Fatalf("Failed to parse blkio.throttle.read_bps_device - %s", err)
+		t.Fatal(err)
```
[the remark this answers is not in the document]
kolyshkin: Adding more context to an error here is redundant on two levels:
1. this is test, and `t.Fatal` shows the line number which is enough to find out what exactly has failed.
2. `fscommon.GetCgroupParamString` adds enough context (full file name) to its error.

libcontainer/cgroups/fs/pids.go
kolyshkin: Please re-review. 

The logic is explained in the commit message and in the comment above. We use `fscommon.GetCgroupParamUint` which returns `max.MaxUint64` in case the `max` string is read from the file:
```go
  // GetCgroupParamUint reads a single uint64 value from the specified cgroup file.
  // If the value read is "max", the math.MaxUint64 is returned.
  func GetCgroupParamUint(path, file string) (uint64, error) {
          contents, err := GetCgroupParamString(path, file)
          if err != nil {
                  return 0, err
          }
          contents = strings.TrimSpace(contents)
          if contents == "max" {
                  return math.MaxUint64, nil
          }
 ```
 
 ...and we convert `math.MaxUint64` to `0` as we historically return `0` for "no limit set".

libcontainer/cgroups/fs2/pids.go
kolyshkin: See https://github.com/opencontainers/runc/pull/3011#discussion_r657344259

kolyshkin:
> One nit, two things need changing, and one question.

Sorry, what was the question? I can't find it.

libcontainer/cgroups/fs/blkio_test.go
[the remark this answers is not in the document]
kolyshkin: It's also described in the commit message
> The error from fscommon.GetCgroup* already contains the file name and so
> on, so there's no need to wrap it.

libcontainer/cgroups/fs/pids.go
kolyshkin: For the commit message, see https://github.com/opencontainers/runc/pull/3011/commits/adcd3b4451303fe2061177189be74f491671cd41
(it's sad that github PR review UI downplays individual commits in favor of overall PR changes)

libcontainer/cgroups/fs/blkio_test.go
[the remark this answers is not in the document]
cyphar: Ah you're right. Yup, thanks.

libcontainer/cgroups/fs/pids.go
cyphar: Yeah sorry, for PRs like this I do a review on the entire changeset because doing commit-by-commit reviews isn't as convenient as it is when you're doing them in a mail client. Downside is I miss explanations in commit messages. Sorry for the noise (and I didn't know that the `max` handling already existed).

cyphar:
@kolyshkin It was about removing genericError but after posting it I noticed that there was a separate PR for that. https://github.com/opencontainers/runc/pull/3011#discussion_r656862680

cyphar (approved):
LGTM, my suggestions missed that the `pids.max` handling already existed in the fscommon utilities.

libcontainer/cgroups/fs/pids.go
cyphar: I also missed that the tests pass, which they shouldn't if this was a real bug.

PR #3012, opened 2021-06-09
libct/cg/fs: don't forget to close a file

kolyshkin:
Fixes: 7fe0a98e79fc87f54f4d

(`os.File` has a runtime finalizer set which closes it when it's garbage collected, but we don't know when it is going to be called)


AkihiroSuda (approved):

cyphar (approved):
LGTM.

Issue #3014, opened 2021-06-09
(regression in master) not ok 117 update devices [minimal transition rules]


kolyshkin:
Got an error from CI for PR #3011, during test (1.16.x) integration test with systemd. Also reproducible on master and other recent PRs.

Quoting https://github.com/opencontainers/runc/runs/2787163520?check_suite_focus=true:

```
not ok 117 update devices [minimal transition rules]
# (in test file tests/integration/update.bats, line 625)
#   `[ -z "$(<"$CONTAINER_OUTPUT")" ]' failed
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-25456/runc-tmp-recvtty.9wbHaU/console.sock test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# runc update --pids-limit 30 test_update (status=0):
# 
# sh: can't create /dev/null: Operation not permitted
# /tmp/bats-run-25456/bats.37480.src: line 603: 39193 Killed                  ( "$RECVTTY" --no-stdin --pid-file "$TMP_RECVTTY_PID" --mode single "$TMP_CONSOLE_SOCKET" &> "$CONTAINER_OUTPUT" )
```

The test case was added in commit 4438eaa5e45aa8d7f1bc109ae5d43760153776ea, PR #2391.

This `sh: can't create /dev/null: Operation not permitted` means somewhere during `runc update` the device access list might become empty for some time.

May be related to recent PR #2994 (although I do not see how). Definitely not related to PR #2986 as we have cgroup v1 here.

kolyshkin:
Got it again on the same PR, the only difference is go version (1.15 this time). Hmm, maybe it is related to the changes I made in #3011.

From https://github.com/opencontainers/runc/runs/2787301089?check_suite_focus=true:
```
[...]

PR #3015, opened 2021-06-09
[DNM] [test] update devices [minimal transition rules]

kolyshkin:
Testing issue https://github.com/opencontainers/runc/issues/3014

 - add more iterations for better chances to catch this
 - remove some useless logging


kolyshkin:
Apparently this is caused by #2994 (commit bf7492e), and the reason is we no longer freeze the cgroup during update, and systemd update is buggy (I remember @cyphar pointed this weird systemd behavior recently, but can't find it now).

I think the best possible solution is to always try to freeze the cgroup.

PR #3016, opened 2021-06-09
Revert "runc update: skip devices"

kolyshkin:
We can not and should not use SkipDevices flag on a running container.

This reverts PR #2994 (commit bf7492ee5d022cd99a9dbe71c5c4f965041552e9)

Fixes: https://github.com/opencontainers/runc/issues/3014


cyphar:
I think it would be better to fix the bug I convinced you to introduce in #2958 (looking at it again, I think that code was wrong even at the time when I told you to add it -- sorry about that).

kolyshkin:
Closing in favor of #3019

PR #3017, opened 2021-06-10
Returns clearer error message for Setenv

wzshiming:
> container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"setenv: invalid argument\"": unknown

xref kubernetes/kubernetes#102732

Pick the judgment logic from `os.Setenv`, and added clear error message 
https://github.com/golang/go/blob/27f83723e98d8e3795a07bdca2b3a8155b0d72b3/src/syscall/env_unix.go#L97-L130

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -112,9 +112,19 @@ func populateProcessEnvironment(env []string) error {
 	for _, pair := range env {
 		p := strings.SplitN(pair, "=", 2)
 		if len(p) < 2 {
-			return fmt.Errorf("invalid environment '%v'", pair)
+			return fmt.Errorf("invalid environment variable: %q", pair)
 		}
-		if err := os.Setenv(p[0], p[1]); err != nil {
+		name, val := p[0], p[1]
+		if name == "" {
+			return fmt.Errorf("environment variable name can't be empty: %q", pair)
+		}
+		if strings.IndexByte(name, 0) >= 0 {
+			return fmt.Errorf("environment variable name can't contain null(\\x00): %q", pair)
+		}
+		if strings.IndexByte(val, 0) >= 0 {
+			return fmt.Errorf("environment variable value can't contain null(\\x00): %q", pair)
+		}
+		if err := os.Setenv(name, val); err != nil {
 			return err
 		}
 	}


libcontainer/init_linux.go
```diff
@@ -112,9 +112,19 @@ func populateProcessEnvironment(env []string) error {
...
+		key := p[0]
+		val := p[1]
+		if key == "" {
+			return fmt.Errorf("environment key can't be empty %q", pair)
+		}
+		for i := 0; i < len(val); i++ {
+			if val[i] == 0 {
+				return fmt.Errorf("environment value can't contain null(\\x00) %q", pair)
+			}
+		}
```
kolyshkin, line 126: You can simplify this by using something like
```go
if strings.IndexByte(val, 0) >= 0 {
   return fmt.Errorf("environment variable value can't contain null(\\x00): %q", pair)
}

File: libcontainer/init_linux.go [hunk 2]
```diff
@@ -112,9 +112,19 @@ func populateProcessEnvironment(env []string) error {
...
 		p := strings.SplitN(pair, "=", 2)
 		if len(p) < 2 {
-			return fmt.Errorf("invalid environment '%v'", pair)
+			return fmt.Errorf("invalid environment %q", pair)
 		}
-		if err := os.Setenv(p[0], p[1]); err != nil {
+		key := p[0]
+		val := p[1]
+		if key == "" {
+			return fmt.Errorf("environment key can't be empty %q", pair)
```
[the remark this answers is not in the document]
kolyshkin: I.e.

```go
return fmt.Errorf("environment variable name can't be empty: %q", pair)
```

libcontainer/init_linux.go
```diff
@@ -112,9 +112,19 @@ func populateProcessEnvironment(env []string) error {
 	for _, pair := range env {
 		p := strings.SplitN(pair, "=", 2)
 		if len(p) < 2 {
-			return fmt.Errorf("invalid environment '%v'", pair)
+			return fmt.Errorf("invalid environment %q", pair)
 		}
-		if err := os.Setenv(p[0], p[1]); err != nil {
+		key := p[0]
+		val := p[1]
```
kolyshkin, line 118:
```suggestion
name, val := p[0], p[1]
```

kolyshkin (approved):
LGTM

libcontainer/init_linux.go
```diff
@@ -112,9 +112,16 @@ func populateProcessEnvironment(env []string) error {
...
-			return fmt.Errorf("invalid environment '%v'", pair)
+			return fmt.Errorf("invalid environment variable: %q", pair)
 		}
-		if err := os.Setenv(p[0], p[1]); err != nil {
+		name, val := p[0], p[1]
+		if name == "" {
+			return fmt.Errorf("environment variable name can't be empty: %q", pair)
+		}
+		if strings.IndexByte(val, 0) >= 0 {
+			return fmt.Errorf("environment variable value can't contain null(\\x00): %q", pair)
```
AkihiroSuda, line 122: Can we do the same check for `name`?

wzshiming: Thanks, updated

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #3019, opened 2021-06-10
libct/cg/sd: fix "SkipDevices" handling

kolyshkin:
1. Fix handling of `SkipDevices` option -- it should be what it literally means,
   i.e. do not set any device-related options.

2. Revert the part of commit 108ee85 which skipped the freeze
   when the `SkipDevices` is set. Apparently, the freeze is needed on
   update even if no `Device*` properties are being set.

3. Add `runc update` to "runc run [device cgroup deny]" test.

Fixes: 752e7a8
Fixes: 108ee85
Fixes: #3014


kolyshkin:
Also testing it in https://github.com/kubernetes/kubernetes/pull/102508

cyphar (approved):
LGTM.

AkihiroSuda (approved):

PR #3022, opened 2021-06-11
libct/cg/fs2: set per-device io weight if available

kolyshkin:
1. A **bugfix** to be able to set per-device weights on cgroup v1 with BFQ.
2. Set per-device weights for v2 with BFQ.
3. A test case for the above two.

I accidentally found out that per-device weight is supported
in BFQ scheduler since kernel v5.4 (kernel commit 795fe54c2a8),
so let's set those if supplied.

Next, I added a test case for it and decided to not make is v2-specific,
as the test for cgroup v1 would look very much the same. This test found a bug
in v1 (see https://github.com/opencontainers/runc/pull/3010#issuecomment-860302432), so I fixed it as well.

This closes a small gap in cgroup v2 feature parity with cgroup v1.

## Proposed changelog entry
* cgroup v2: set per-device io weights if BFQ IO scheduler is available.

### DIFF
--- a/libcontainer/cgroups/fs/blkio.go
+++ b/libcontainer/cgroups/fs/blkio.go
@@ -41,11 +41,15 @@ func (s *BlkioGroup) Set(path string, r *configs.Resources) error {
 		}
 	}
 	for _, wd := range r.BlkioWeightDevice {
-		if err := cgroups.WriteFile(path, s.weightDeviceFilename, wd.WeightString()); err != nil {
-			return err
+		if wd.Weight != 0 {
+			if err := cgroups.WriteFile(path, s.weightDeviceFilename, wd.WeightString()); err != nil {
+				return err
+			}
 		}
-		if err := cgroups.WriteFile(path, "blkio.leaf_weight_device", wd.LeafWeightString()); err != nil {
-			return err
+		if wd.LeafWeight != 0 {
+			if err := cgroups.WriteFile(path, "blkio.leaf_weight_device", wd.LeafWeightString()); err != nil {
+				return err
+			}
 		}
 	}
 	for _, td := range r.BlkioThrottleReadBpsDevice {
--- a/libcontainer/cgroups/fs2/io.go
+++ b/libcontainer/cgroups/fs2/io.go
@@ -4,6 +4,8 @@ package fs2
 
 import (
 	"bufio"
+	"bytes"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -16,10 +18,25 @@ import (
 
 func isIoSet(r *configs.Resources) bool {
 	return r.BlkioWeight != 0 ||
+		len(r.BlkioWeightDevice) > 0 ||
 		len(r.BlkioThrottleReadBpsDevice) > 0 ||
 		len(r.BlkioThrottleWriteBpsDevice) > 0 ||
 		len(r.BlkioThrottleReadIOPSDevice) > 0 ||
 		len(r.BlkioThrottleWriteIOPSDevice) > 0
+}
+
+// bfqDeviceWeightSupported checks for per-device BFQ weight support (added
+// in kernel v5.4, commit 795fe54c2a8) by reading from "io.bfq.weight".
+func bfqDeviceWeightSupported(bfq *os.File) bool {
+	if bfq == nil {
+		return false
+	}
+	_, _ = bfq.Seek(0, 0)
+	buf := make([]byte, 32)
+	_, _ = bfq.Read(buf)
+	// If only a single number (default weight) if read back, we have older kernel.
+	_, err := strconv.ParseInt(string(bytes.TrimSpace(buf)), 10, 64)
+	return err != nil
 }
 
 func setIo(dirPath string, r *configs.Resources) error {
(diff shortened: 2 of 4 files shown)


kolyshkin:
@askervin PTAL

kolyshkin:
This is a feature, not a bug (as I see it), so adding to 1.1 milestone

libcontainer/cgroups/fs2/io.go
```diff
@@ -43,6 +45,16 @@ func setIo(dirPath string, r *configs.Resources) error {
 			}
 		}
 	}
+	for _, wd := range r.BlkioWeightDevice {
+		// per-device weight is supported since kernel v5.4, commit 795fe54c2a8.
+		if err := fscommon.WriteFile(dirPath, bfqWeight, wd.WeightString()); err != nil {
```
cyphar, line 50: This is unsafe to do unconditionally because old kernels called `kstrtoull` on the input string unconditionally -- and since `WeightString` is of the form `major:minor weight` we will end up setting the weight to the last `major` value in `BlkioWeightDevice`. We should check whether `bfq.weight_device` exists before writing these entries.

cyphar, line 50: This is unsafe to do unconditionally because old kernels called `kstrtoull` on the input string unconditionally -- and since `WeightString` is of the form `major:minor weight` we will end up setting the weight to the last `major` value in `BlkioWeightDevice`. We should check whether `bfq.weight_device` exists before writing these entries.

cyphar (changes requested):

libcontainer/cgroups/fs2/io.go
cyphar: We should also be returning an error if the user has explicitly set `BlkioWeightDevice` when the kernel doesn't support it.

kolyshkin: > This is unsafe to do unconditionally because old kernels called `kstrtoull` on the input string unconditionally -- and since `WeightString` is of the form `major:minor weight` we will end up setting the weight to the last `major` value in `BlkioWeightDevice`.

Not quite -- it chokes on the first non-numeric character which is not a newline, in our case `:`

```console
root@ubuntu1910:/sys/fs/cgroup/init.scope# echo "100:101 222" > io.bfq.weight 
bash: echo: write error: Invalid argument
root@ubuntu1910:/sys/fs/cgroup/init.scope# echo "100:100 222" > io.bfq.weight 
bash: echo: write error: Invalid argument
root@ubuntu1910:/sys/fs/cgroup/init.scope# echo "10:0 20" > io.bfq.weight 
bash: echo: write error: Invalid argument
root@ubuntu1910:/sys/fs/cgroup/init.scope# echo "10" > io.bfq.weight 
root@ubuntu1910:/sys/fs/cgroup/init.scope# echo "10:" > io.bfq.weight 
bash: echo: write error: Invalid argument
root@ubuntu1910:/sys/fs/cgroup/init.scope# uname -a
Linux ubuntu1910 5.3.0-64-generic #58-Ubuntu SMP Fri Jul 10 19:33:51 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
root@ubuntu1910:/sys/fs/cgroup/init.scope# 
```

> We should check whether `bfq.weight_device` exists before writing these entries.

This only exists on cgroup v1, and this is v2 code. We can read `io.bfq.weight` and if it says `default ...` rather than a number, the per-device weights are available. Or, add `default` when writing a default weight, if it's accepted we have per-device weights.

kolyshkin: Updated to check if per-device weights are supported. Also added a test case.

kolyshkin:
The test case for cgroup v2 reveal an issue with v1 blkio, so I am adding a bug fix for it, and changing the milestone to 1.0.0 (as it is now a bug fix).

I can split this PR into two -- a bug fix (for cgroup v1) and a new feature (for cgroup v2), but as the test case is the same I'd rather keep this together.

kolyshkin:
Need more time to finish the test case -- hope to be done tomorrow.

libcontainer/cgroups/fs/blkio.go
```diff
@@ -42,11 +42,15 @@ func (s *BlkioGroup) Set(path string, r *configs.Resources) error {
...
-		if err := fscommon.WriteFile(path, s.weightDeviceFilename, wd.WeightString()); err != nil {
-			return err
+		if wd.Weight != 0 {
+			if err := fscommon.WriteFile(path, s.weightDeviceFilename, wd.WeightString()); err != nil {
+				return err
+			}
 		}
-		if err := fscommon.WriteFile(path, "blkio.leaf_weight_device", wd.LeafWeightString()); err != nil {
-			return err
+		if wd.LeafWeight != 0 {
```
kolyshkin, line 50: The question is, CFQ (removed from kernel v5.0) supported weight and leaf_weight, while BFQ only supports weight. Should we ignore leaf weight for BFQ case or error out?

kolyshkin: I think erroring out is fine, as leaf weights are obsoleted and no one should set them. If they do, they get an error.

kolyshkin:
Rebased on top of merged #3024.

askervin (commented):
@kolyshkin , great test & finding that `leaf_device_weight` was being written in cgroups v1 every time!

I did some manual testing with this cgroups v2 code, just to make sure that sequences like
- open-write-seek-read-write
- open-seek-read-write
on `io.bfq.weight` indeed do what we expect. And it does.

This looks good and works fine. Only a super-nit: a typo in a comment.

libcontainer/cgroups/fs2/io.go
```diff
@@ -16,32 +18,63 @@ import (
...
+	_, _ = bfq.Read(got)
+	return bytes.Equal(exp, got)
+}
+
 func setIo(dirPath string, r *configs.Resources) error {
 	if !isIoSet(r) {
 		return nil
 	}
 
+	// If BFQ IO scheduler if available, use it.
```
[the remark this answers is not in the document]
kolyshkin: Thanks for catching that, fixed!

cyphar (dismissed):
LGTM.

askervin (dismissed):
LGTM. Special thanks for fixing the issue in PR #3010! :)

libcontainer/cgroups/fs2/io.go
```diff
@@ -16,32 +18,63 @@ import (
...
 }
 
+// bfqDeviceWeightSupported checks for per-device BFQ weight support (added
+// in kernel v5.4, commit 795fe54c2a8) by reading from "io.bfq.weight".
+func bfqDeviceWeightSupported(bfq *os.File) bool {
+	if bfq == nil {
+		return false
+	}
+	_, _ = bfq.Seek(0, 0)
+	exp := []byte("default ")
```
AkihiroSuda, line 35: Is this really guaranteed to be in the first line?

kolyshkin: Yes; see `bfq_io_show_weight` in `block/bfq-cgroup.c`, e.g. 

https://github.com/torvalds/linux/blob/009c9aa5be652675a06d5211e1640e02bbb1c33d/block/bfq-cgroup.c#L977-L986

kolyshkin: And before https://github.com/torvalds/linux/commit/795fe54c2a828099e461e8c36b04210b2df462ed it was a number in there.

AkihiroSuda: Future version of the kernel may have another line before "default 100".
So I guess we have to scan all the lines?

cyphar: *In theory* that would be an ABI breakage, but yeah it seems a bit risky to depend on the exact order of lines in a kernel file when the alternative is pretty inexpensive. (And in the past cgroup ABIs haven't exactly been completely free of ABI breakages.)

kolyshkin: I disagree.

So far we have to distinguish between two versions of the `bfq.io.weight` API. Let's call them version 1 and version 2. Version 1 only allows to get/set the default weight. Version 2 also allows to get/set per-device weights.

Now, if the **first** line of `bfq.io.weight` does read back as `default XXX`, we do have version 2. Otherwise it's something else. Either version 1 which does not support setting per-device weights, or perhaps a version 3 we don't know anything about.

But let's assume that write format for version 3 is backward-compatible with version 2 (same as version 2 write is backward-compatible with version 1, although the read is not). In this case we need to check not for version 2, but for anything other than version 1.

Such a check would be to read the entire file and see that it's not a version 1 file, i.e. not a single number followed by a newline.

I will update the commit.

kolyshkin: Updated. The new check specifically checks for v1 file, not v2 file, and is thus more future-proof (assuming that in v3 file format will expand and not go back to same as it was in v1).

AkihiroSuda (approved):

cyphar (approved):
LGTM. With this I'll prep a release.

PR #3024, opened 2021-06-11
libct/cg: mv fscommon.{Open,Read,Write}File to cgroups

kolyshkin:
This is a better place as cgroups itself is using these.
Should help with moving more stuff common in between fs and fs2 to
fscommon. For example, it should help #2883 to have RDMA code
common between fs and fs2 in fscommon (see discussion at 
https://github.com/opencontainers/runc/pull/2883#discussion_r605967485)

Looks big, but this is just moving the implementation of `fscommon.{Open,Read,Write}File`
and its test case to `cgroups` pkg:

    fscommon/{fscommon,open}.go -> cgroups/file.go
    fscommon/fscommon_test.go   -> cgroups/file_test.go

There's no functional change.


kolyshkin:
From one thing, this is a public API that we'd rather not change after 1.0, so it's better to change in before, and there is no functional change here, so it's relatively safe to merge.

OTOH I don't like moving things around right before a release.

kolyshkin:
Found two existing users of the functionality that is being moved:

* https://github.com/hashicorp/nomad/blob/main/client/lib/cgutil/cgutil_linux.go
* https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/hotplug-disk/mount.go

So I think we should do the change before 1.0 and I will help these two packages to adopt accordingly.

kolyshkin:
(split into two commits; testing the first one thus draft for now)

kolyshkin:
> So I think we should do the change before 1.0 and I will help these two packages to adopt accordingly.

Yet better, I have added the backward-compatible aliases and marked them as deprecated, so there's no breakage plus a hint from a linter to switch to new names.

kolyshkin:
> Yet better, I have added the backward-compatible aliases and marked them as deprecated, so there's no breakage plus a hint from a linter to switch to new names.

OTOH that means we don't have to change before 1.0, although it's still better to do it before.

AkihiroSuda (approved):

mrunalp (approved):

Issue #3026, opened 2021-06-13
v1.0.0rc95 ENOENT when launching container looking for cpuset.cpus with kernel 5.8


MikeDombo:
Hi,
I'm using libcontainer to create my container and I found a regression when Ubuntu updated from kernel 5.4 to kernel 5.8 on Thursday/Friday. I'm currently seeing `process_linux.go:385: applying cgroup configuration for process caused: openat2 /sys/fs/cgroup/cpuset/<container id>/cpuset.cpus: no such file or directory`. On kernel 5.4 using the exact same code it work fine. I edited libcontainer to do a directory listing and I can verify that the `cpuset.cpus` "file" does exist as it should, however it is unable to be opened by libcontainer. I am able to run `cat` on the path without error from outside the container.

Any pointers would be incredibly helpful, thank you very much.

cyphar:
It seems that the issue is that `openat2` is available after the update (`openat2` was added in Linux 5.6). However this code definitely works on both old and new kernels, so it's a bit puzzling why it's failing for you. Does your code mess around with open files at all? With openat2, we cache a file descriptor to `/sys/fs/cgroup` -- so if some of your code is dup-ing over that file descriptor with another file descriptor, you're going to end up with the cgroup2 code trying to open `/some/random/file/cpuset/container-id/cpuset.cpus` which will either fail or cause security issues.

(As an aside @kolyshkin I think that the error message in the openat2 code should be changed to better describe the path being opened -- `readlink` of the fd would be a good idea. I'll write a patch for that.)

(I also want to point out that in case you weren't aware -- we don't support using libcontainer directly. I understand some people do this, but we don't provide any guarantees that it will continue working . Please use runc or [the go-runc library which wraps runc](https://github.com/containerd/go-runc) if you can. If there is some reason you cannot use runc, please open an issue to describe why you cannot use runc so we can see what we can do.)

cyphar:
Hi, it seems that you've opened an issue that is in relation to using `github.com/opencontainers/runc/libcontainer` directly from your own Go code. While we may do our best to help you solve your issue, **we do not support this.** In addition, our Semantic Versioning version **does not** include the Go API in its scope -- meaning we are free to break the Go APIs in point releases.

[...]

PR #3030, opened 2021-06-15
libct/cg/OpenFile: fix/improve openat2 handling

kolyshkin:
1. While looking into https://github.com/opencontainers/runc/issues/3026, I started writing a test case and found out that 
    `OpenFile("/sys/fs/cgroup", name)` falls back to using open and fstatfs.
    This is caused by a peculiarities of its `dir` and `name` arguments handling,
    and was never discovered before because of the fallback.

    Fix the bug, add the test cases.

2. Improve the error message for case like in the issue #3026, based on @cyphar suggestion.

See commit messages for more details.

### DIFF
--- a/libcontainer/cgroups/file.go
+++ b/libcontainer/cgroups/file.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 	"sync"
 
@@ -13,7 +14,11 @@ import (
 )
 
 // OpenFile opens a cgroup file in a given dir with given flags.
-// It is supposed to be used for cgroup files only.
+// It is supposed to be used for cgroup files only, and returns
+// an error if the file is not a cgroup file.
+//
+// Arguments dir and file are joined together to form an absolute path
+// to a file being opened.
 func OpenFile(dir, file string, flags int) (*os.File, error) {
 	if dir == "" {
 		return nil, fmt.Errorf("no directory specified for %s", file)
@@ -109,8 +114,6 @@ func prepareOpenat2() error {
 	return prepErr
 }
 
-// OpenFile opens a cgroup file in a given dir with given flags.
-// It is supposed to be used for cgroup files only.
 func openFile(dir, file string, flags int) (*os.File, error) {
 	mode := os.FileMode(0)
 	if TestMode && flags&os.O_WRONLY != 0 {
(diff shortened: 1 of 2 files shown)


kolyshkin:
No need to backport it to 1.0 -- this is a bug, but the fallback works so the impact is pretty low

kolyshkin:
This is a real bug fix, can @opencontainers/runc-maintainers PTAL 🙏🏻 ?

libcontainer/cgroups/file.go
```diff
@@ -109,43 +114,43 @@ func prepareOpenat2() error {
...
 	if prepareOpenat2() != nil {
-		return openFallback(dir, file, flags, mode)
+		return openFallback(path, flags, mode)
 	}
-	reldir := strings.TrimPrefix(dir, cgroupfsPrefix)
-	if len(reldir) == len(dir) { // non-standard path, old system?
-		return openFallback(dir, file, flags, mode)
+	relPath := strings.TrimPrefix(path, cgroupfsPrefix)
+	if len(relPath) == len(path) { // non-standard path, old system?
+		return openFallback(path, flags, mode)
```
kolyshkin, line 131: To help reviewers.

Basically, the code previously assumed that `dir` starts with `/sys/fs/cgroup/` (including the ending slash), which is not always the case (such as when calling `OpenFile("/sys/fs/cgroup", "cgroup.controllers")`). So, the fallback was used for this case, which is not correct.

This commit removes such assumption,  using `path.Join(dir, file)` as a full path. This simplifies the code, and helps the case above, as well as something like `OpenFile("/sys", "fs/cgroup/cgroup.controllers")`.

hqhq (dismissed):
LGTM

libcontainer/cgroups/file.go
```diff
@@ -137,6 +138,17 @@ func openFile(dir, file string, flags int) (*os.File, error) {
...
 	if err != nil {
+		// Check if cgroupFd is still opened to cgroupfsDir.
+		fdStr := strconv.Itoa(cgroupFd)
+		fdDest, _ := os.Readlink("/proc/self/fd/" + fdStr)
+		if fdDest != cgroupfsDir {
+			// TODO: reopen cgroupFd and retry openat2.
+
+			// Enhance the Path in the error to contain the
+			// cgroupFd value and the directory it is opened to,
+			// for example: "@[fd 7:/!=/sys/fs/cgroup]/cpu.stat".
```
AkihiroSuda, line 149: Is this convention already widely used, or newly invented here?

Rather than defining a new microformat, can we just wrap `os.PathError` with our own struct?

kolyshkin: > Is this convention already widely used, or newly invented here?

Newly invented (originally comes from https://github.com/opencontainers/runc/issues/3026#issuecomment-860148103)

We are already using "fd N" in a few places where we operate on a file descriptor:
```console
[kir@kir-rhat runc]$ git grep PathError | grep -w fd
libcontainer/rootfs_linux.go:		return &os.PathError{Op: "fchdir", Path: "fd " + strconv.Itoa(newroot), Err: err}
libcontainer/rootfs_linux.go:		return &os.PathError{Op: "fchdir", Path: "fd " + strconv.Itoa(oldroot), Err: err}
libcontainer/setns_init_linux.go:		return &os.PathError{Op: "close log pipe", Path: "fd " + strconv.Itoa(l.logFd), Err: err}
libcontainer/standard_init_linux.go:		return &os.PathError{Op: "close log pipe", Path: "fd " + strconv.Itoa(l.logFd), Err: err}
vendor/github.com/opencontainers/selinux/go-selinux/selinux_linux.go:		return "", &os.PathError{Op: "getsockopt", Path: "fd " + strconv.Itoa(int(fd)), Err: err}
```

but in this case something more useful is needed.

> Rather than defining a new microformat, can we just wrap os.PathError with our own struct?

Done

kolyshkin: @AkihiroSuda PTAL ^^^

AkihiroSuda (approved):

kolyshkin:
@cyphar @thaJeztah @mrunalp  PTAL

thaJeztah (approved):
LGTM, thanks!

PR #3033, opened 2021-06-22
libcontainer: rm own error system

kolyshkin:
This removes libcontainer's own error wrapping system, consisting of a
few types and functions, aimed at typization, wrapping and unwrapping
of errors, as well as saving error stack traces.
    
Since Go 1.13 now provides its own error wrapping mechanism and a few
related functions, it makes sense to switch to it.
    
A few things worth mentioning:
    
1. We lose stack traces (which were never shown anyway).
    
2. Users of libcontainer that relied on particular errors (like
       `ContainerNotExists`) need to switch to using `errors.Is` with
       the new errors defined in libcontainer/error.go.
       An example of such conversion:
```diff
-		if err != nil {
-			var lerr libcontainer.Error
-			if errors.As(err, &lerr) && lerr.Code() == libcontainer.ContainerNotExists {
+			if errors.Is(err, libcontainer.ErrNotExist) {
```
3. (_Implementation detail_) `encoding/json` is unable to unmarshal the built-in
    error type, so we have to introduce `initError` and wrap the errors into it
   (basically passing the error as a string). This is the same
   as it was before, just a tad simpler (actually the `initError`
   itself is a type that got removed in commit afa8443; also suddenly
   `ierr` variable name makes sense now).

## Proposed changelog entry
* `libcontainer.Error` type, together with error codes, is removed in favor of new errors. Users relying on removed types and codes should adopt to use `errors.Is` with new errors (see `libcontainer/errors.go`). (#3033)


kolyshkin:
CI failure is probably a flake (https://github.com/opencontainers/runc/issues/3035); CI restarted

kolyshkin:
Rebased on top of #3011

zhsj:
With removal of  generic_error.go, probably remove `libcontainer/stacktrace` together?

libcontainer/container.go
```diff
@@ -74,71 +74,39 @@ type BaseContainer interface {
 	ID() string
 
 	// Returns the current status of the container.
-	//
```
kolyshkin, line 77: This documentation (about which errors can be returned) is removed because
 - libcontainer/error.go already list possible errors;
 - it's not easy to keep these lists up to date.

kolyshkin:
> With removal of generic_error.go, probably remove libcontainer/stacktrace together?

Thanks, @zhsj! Removed.

cyphar (approved):
LGTM. Let's merge this before #3049 so you don't need to rebase this one.

libcontainer/factory_linux.go
```diff
@@ -249,28 +251,28 @@ type LinuxFactory struct {
...
 	if l.Root == "" {
-		return nil, newGenericError(errors.New("invalid root"), ConfigInvalid)
+		return nil, &ConfigError{"invalid root"}
 	}
 	if err := l.validateID(id); err != nil {
 		return nil, err
 	}
 	if err := l.Validator.Validate(config); err != nil {
-		return nil, newGenericError(err, ConfigInvalid)
+		return nil, &ConfigError{err.Error()}
```
AkihiroSuda, line 260: Can we use %w here too?

libcontainer/factory_linux.go
```diff
@@ -249,28 +251,28 @@ type LinuxFactory struct {
 
 func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, error) {
 	if l.Root == "" {
-		return nil, newGenericError(errors.New("invalid root"), ConfigInvalid)
+		return nil, &ConfigError{"invalid root"}
```
AkihiroSuda, line 254: Do we need `&ConfigError{}`?

AkihiroSuda (approved):

libcontainer/factory_linux.go
kolyshkin: Yeah, this looks a tad weird. I might change this to using an error later.

libcontainer/factory_linux.go
kolyshkin: It's a replacement for "ConfigInvalid" error code, so a caller can distinguish configuration errors from everything else.

libcontainer/factory_linux.go
kolyshkin: Please see https://github.com/opencontainers/runc/pull/3056

libcontainer/factory_linux.go
kolyshkin: You're right @AkihiroSuda, it's better not to have it, but I was not able to come with a good solution before. Now I am; please see https://github.com/opencontainers/runc/pull/3056

Issue #3028, opened 2021-06-13
move most packages into `internal`


cyphar:
At the moment all of our internal packages are importable from anywhere. There are several historical reasons for this:

 * Docker originally used LXC, but then created a Go container runtime as an internal library (libcontainer) which was used by Docker directly. This meant that containers were created by forking Docker and doing the namespace setup immediately, which turns out to not be a great idea. This library was moved to a separate repo, but was still imported by Docker. So most of the library code was being imported by another project. The OCI was created and libcontainer was wrapped in a binary called "runc", which was then moved to the OCI repo. But Docker still used libcontainer for a little while before switching. So there was a long time when we had to have our APIs be external.
 * The `internal` package system didn't exist until Go 1.4, and it wasn't really widely publicised at the time (I only heard about it a few years ago).

However, there are still modern users of our APIs:

 * Kubernetes (specifically cAdvisor and Kubelet) make use of our cgroup libraries fairly heavily and were the first external users from memory. Quite a few other projects also use our cgroup libraries.
 * Docker still imports a handful of our APIs, some of which only really exist for Docker's sake and aren't even used within runc (`HostDevices` comes to mind).

So we can't just move everything into an `internal` package but we really should move most of it. We do have some users using `libcontainer` as a library but it is fairly scary because it is fairly easy to misconfigure containers if you use the `Process` APIs for instance. We should move as many libraries as possible behind an `internal` package.

This will also make it easier to explain our SemVer policy -- because right now I would argue that SemVer doesn't cover our Go APIs.

dims:
taking stock, as of today here's what kubernetes is importing:

```
[dims@dims-a01 19:14] ~/go/src/k8s.io/kubernetes ⟩ rg opencontainers/runc | grep -v vendor | grep "\.go:" | grep -oh "\".*\"" | sort | uniq
"github.com/opencontainers/runc/libcontainer/apparmor"
"github.com/opencontainers/runc/libcontainer/cgroups"
"github.com/opencontainers/runc/libcontainer/cgroups/fs"
"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
"github.com/opencontainers/runc/libcontainer/configs"
"github.com/opencontainers/runc/libcontainer/devices"
"github.com/opencontainers/runc/libcontainer/utils"
```

dims:
[...]

PR #3049, opened 2021-06-29
libct: internalize some packages

kolyshkin:
This moves a few packages from `libcontainer` to `internal`, so that they are no longer available to external users. This is the first step in implementing #3028.

I chose `internal` over, say, `libcontainer/internal` or `pkg/internal` merely because of a short path.


kolyshkin:
Hope this won't conflict with #3033 as it will be a nightmare to rebase that one.

AkihiroSuda (dismissed):

kolyshkin:
Are there any objections or comments about this PR, @opencontainers/runc-maintainers? I have another PR (https://github.com/opencontainers/runc/pull/3056) that is based on this one and want to move forward with it.

hqhq:
I personally prefer `libcontainer/internal`, because libcontainer is supposed to be the library of container, even these files are moved to internal, they should still be part of the "library of container".

kolyshkin:
> I personally prefer `libcontainer/internal`, because libcontainer is supposed to be the library of container, even these files are moved to internal, they should still be part of the "library of container".

First, I was hoping to make those import paths shorter (or at least not longer).

Second, runc/libcontainer is used by other projects, while runc/internal clearly contains a set of packages aimed for runc itself. Are those packages about containers? Yes, they are -- the whole runc is, and no, they aren't -- say, `capabilities` package is about process caps, and is not really about containers per se.

I'd like to keep it as is (mostly because it's shorter).

hqhq:
> > I personally prefer `libcontainer/internal`, because libcontainer is supposed to be the library of container, even these files are moved to internal, they should still be part of the "library of container".
> 
> First, I was hoping to make those import paths shorter (or at least not longer).
> 
> Second, runc/libcontainer is used by other projects, while runc/internal clearly contains a set of packages aimed for runc itself. Are those packages about containers? Yes, they are -- the whole runc is, and no, they aren't -- say, `capabilities` package is about process caps, and is not really about containers per se.
> 

Yes, everything in runc repo is about container, but the phrase I used was "library of container",
it's different with just "container".

Back to the old days, we only have libcontainer, then we moved it to a separate repo to build runc based
on it, runc only added very limited things, the most important part is the cli, other than that, just minor
things like scripts, test cases, docs, man pages, etc.

According to #3028 , we are about to move most of libcontainer packages to internal, which means we are
removing the ability to create containers just by importing libcontainer and using its APIs, it's fine to me if
nobody is doing so.  But I think we can still keep libcontainer as independent, it's kind of the `core` of container
runtime.

I'm happy to change my mind if there are other concerns, just shorter path isn't a compelling reason for me.
@cyphar @crosbymichael WDYT?

> I'd like to keep it as is (mostly because it's shorter).

kolyshkin:
draft until we decide on the path.

kolyshkin:
No consensus on the path for internal packages, so I'm closing this one.

PR #3054, opened 2021-06-30
makefile: update ldflags and add strip for static builds

kailun-qin:
This patch
* drops the default `-w` flag for `make static`, which helps with
  debugging the static runc binary;
* adds `EXTRA_LDFLAGS="-w -s"` to `script/release.sh` to disable DWARF
  generation and symbol table for the release runc binary;
* adds strip in `script/release.sh` for a further size-optimized release
  runc binary.

### DIFF
--- a/script/release.sh
+++ b/script/release.sh
@@ -33,6 +33,7 @@ function build_project() {
 	local libseccomp_ver='2.5.1'
 	local tarball="libseccomp-${libseccomp_ver}.tar.gz"
 	local prefix
+	local ldflags="-w -s"
 	prefix="$(mktemp -d)"
 	wget "https://github.com/seccomp/libseccomp/releases/download/v${libseccomp_ver}/${tarball}"{,.asc}
 	tar xf "$tarball"
@@ -46,8 +47,9 @@ function build_project() {
 	# Add -a to go build flags to make sure it links against
 	# the provided libseccomp, not the system one (otherwise
 	# it can reuse cached pkg-config results).
-	make -C "$root" PKG_CONFIG_PATH="${prefix}/lib/pkgconfig" COMMIT_NO= EXTRA_FLAGS="-a" static
+	make -C "$root" PKG_CONFIG_PATH="${prefix}/lib/pkgconfig" COMMIT_NO= EXTRA_FLAGS="-a" EXTRA_LDFLAGS="${ldflags}" static
 	rm -rf "$prefix"
+	strip "$root/$project"
 	mv "$root/$project" "$1"
 }
 
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ endif
 GO_BUILD := $(GO) build -trimpath $(MOD_VENDOR) $(GO_BUILDMODE) $(EXTRA_FLAGS) -tags "$(BUILDTAGS)" \
 	-ldflags "-X main.gitCommit=$(COMMIT) -X main.version=$(VERSION) $(EXTRA_LDFLAGS)"
 GO_BUILD_STATIC := CGO_ENABLED=1 $(GO) build -trimpath $(MOD_VENDOR) $(EXTRA_FLAGS) -tags "$(BUILDTAGS) netgo osusergo" \
-	-ldflags "-w -extldflags -static -X main.gitCommit=$(COMMIT) -X main.version=$(VERSION) $(EXTRA_LDFLAGS)"
+	-ldflags "-extldflags -static -X main.gitCommit=$(COMMIT) -X main.version=$(VERSION) $(EXTRA_LDFLAGS)"
 
 .DEFAULT: runc
 


kolyshkin:
The flag was added by commit ecd64631019d0, not sure why though. The `-w` flag is to disable DWARF generation (which, as far as I can see, can be stripped using `strip`).

I think it's better to drop the flag entirely, and instead add `EXTRA_LDFLAGS="-w -s"` to `script/release.sh` instead. Here `-s` is 
to disable symbol table (which further reduces the binary size).

@kailun-qin can you do that? While at it, please don't use markdown in commit subject.

Interestingly, even the binary build with `-w -s` is strippable further; here is a comparison:

```
[kir@kir-rhat runc]$ ls -l *runc-static*
-rwxrwxr-x. 1 kir kir 15527744 Jun 30 10:52 runc-static
-rwxrwxr-x. 1 kir kir 11599152 Jun 30 11:01 runc-static-s
-rwxrwxr-x. 1 kir kir 13137864 Jun 30 10:52 runc-static-w
-rwxrwxr-x. 1 kir kir 11599152 Jun 30 10:51 runc-static-w-s
-rwxrwxr-x. 1 kir kir 10169624 Jun 30 10:53 stripped-runc-static
-rwxrwxr-x. 1 kir kir 10169624 Jun 30 11:01 stripped-runc-static-s
-rwxrwxr-x. 1 kir kir 10169624 Jun 30 10:53 stripped-runc-static-w
-rwxrwxr-x. 1 kir kir 10169624 Jun 30 10:53 stripped-runc-static-w-s
```

(in the listing above the first four items are builds with and without -w and -w, and the last for are stripped versions of those).

cyphar:
@kolyshkin GNU `strip` used to break stack backtraces in Go programs, while `-w` would reduce the size of the binary without that problem (in fact there are some go-nuts threads from a long time ago where folks were told to not use GNU strip on Go binaries). Now, it seems that this is no longer the case so maybe we can just use `strip` directly?

kailun-qin:
@kolyshkin @cyphar Thanks for the comments, updated! I intended to propose to add `-s` (for release) by default as well.
See if you would like to use `strip` v.s. `-w -s`. The `-w -s` approach would not introduce any dependency, for example `strip` may not be available on Windows systems.

script/release.sh
```diff
@@ -33,7 +33,9 @@ function build_project() {
 	local libseccomp_ver='2.5.1'
 	local tarball="libseccomp-${libseccomp_ver}.tar.gz"
 	local prefix
+	local ldflags
 	prefix="$(mktemp -d)"
+	ldflags="-w -s"
```
kolyshkin, line 38: nit:
```suggestion
local ldflags="-w -s"
local prefix
prefix=$(mktemp -d)"
```

The `prefix` assignment is separated from declaration as otherwise it would hide an error from mktemp. There can be no error in assigning a string, so no need to do the same for `ldflags` assignment.

kolyshkin:
> Now, it seems that this is no longer the case so maybe we can just use strip directly?

To me it makes sense to use both `-w -s` and `strip`. The linker flags are merely to save some CPU and disk I/O.

As a fun experiment, I just tried to upx the resulting binary, slashing its size from 9.7M down to 3.5M. Apparently everything still works, the only downsides are:
 - the startup is 0.1s slower;
 - `go version -m <binary>` no longer works.

script/release.sh
```diff
@@ -46,7 +48,7 @@ function build_project() {
 	# Add -a to go build flags to make sure it links against
 	# the provided libseccomp, not the system one (otherwise
 	# it can reuse cached pkg-config results).
-	make -C "$root" PKG_CONFIG_PATH="${prefix}/lib/pkgconfig" COMMIT_NO= EXTRA_FLAGS="-a" static
+	make -C "$root" PKG_CONFIG_PATH="${prefix}/lib/pkgconfig" COMMIT_NO= EXTRA_FLAGS="-a" EXTRA_LDFLAGS="${ldflags}" static
 	rm -rf "$prefix"
 	mv "$root/$project" "$1"
```
kolyshkin, line 53: @kailun-qin as discussed above, can you also add `strip` before this line here?

kailun-qin: @kolyshkin Added, thanks! PTAL.

kailun-qin:
> > Now, it seems that this is no longer the case so maybe we can just use strip directly?
> 
> To me it makes sense to use both `-w -s` and `strip`. The linker flags are merely to save some CPU and disk I/O.
> 
> As a fun experiment, I just tried to upx the resulting binary, slashing its size from 9.7M down to 3.5M. Apparently everything still works, the only downsides are:
> 
> * the startup is 0.1s slower;
> * `go version -m <binary>` no longer works.

Exactly, upx is usually considered as the ultimate approach for binary size optimization.

kailun-qin:
@kolyshkin @cyphar Would you please kindly take another look when you get a chance? Thanks!

script/release.sh
kolyshkin: @kailun-qin please address this one ^^^^

kolyshkin: Ah, I see, you already did that, but in a separate commit.

kolyshkin:
@kailun-qin please squash your commits, it's easier to review and in general makes more sense in this case.

The only other nit I see is the word "proposes" in the commit description. Once merged, it's no longer a proposal but a factual change, so please change the description accordingly ("This patch does this and that because ...", or, yet better, "Do this and that
because...").

kailun-qin:
> @kailun-qin please squash your commits, it's easier to review and in general makes more sense in this case.
> 
> The only other nit I see is the word "proposes" in the commit description. Once merged, it's no longer a proposal but a factual change, so please change the description accordingly ("This patch does this and that because ...", or, yet better, "Do this and that
> because...").

Squashed and updated the commit message. PTAL, thanks!

kolyshkin (approved):
LGTM

kolyshkin:
Needs a rebase. I am carrying this in #3099 and have rebased it in there, so I guess we can close this one.

PR #3055, opened 2021-07-01
cgroupv2: ebpf: ignore inaccessible existing programs

cyphar:
This is necessary in order for runc to be able to configure device
cgroups with --systemd-cgroup on distributions that have very strict
SELinux policies such as [openSUSE MicroOS][1].

The core issue here is that systemd is adding its own BPF policy that
has an SELinux label such that runc cannot interact with it. In order to
work around this, we can just ignore the policy -- in theory this
behaviour is not correct but given that the most obvious case
(--systemd-cgroup) will still handle updates correctly, this logic is
reasonable.

[1]: https://bugzilla.suse.com/show_bug.cgi?id=1182428

<hr>

### Changelog Entry ###
```
 * cgroupv2: bpf: Ignore inaccessible existing programs in case of
   permission error when handling replacement of existing bpf cgroup
   programs. This fixes a regression in 1.0.0, where some SELinux
   policies would block runc from being able to run entirely. #3055
```


cyphar:
This was a regression in 1.0.0.

cyphar:
This doesn't actually solve the issue -- cilium's eBPF library is explicitly hiding the underlying error here:

```go
func wrapObjError(err error) error {
	if err == nil {
		return nil
	}
	if errors.Is(err, unix.ENOENT) {
		return fmt.Errorf("%w", ErrNotExist)
	}
	return errors.New(err.Error())
}
```

cyphar:
We're waiting for cilium/ebpf#336.

kolyshkin:
Nice!

Should we mark this one as a draft, waiting for 0.6.2 (or whatever the next version is) of cilium/ebpf to be released?

cyphar:
Sure, I marked it as a draft (though not sure when they'll do a release).

kolyshkin:
> Sure, I marked it as a draft (though not sure when they'll do a release).

Filed https://github.com/cilium/ebpf/issues/340, hope they will cut a release soon.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/bits-and-blooms/bitset v1.2.0
 	github.com/checkpoint-restore/go-criu/v5 v5.0.0
-	github.com/cilium/ebpf v0.6.1
+	github.com/cilium/ebpf v0.6.2-0.20210702091250-ca492085341e
```
AkihiroSuda, line 8: https://github.com/cilium/ebpf/releases/tag/v0.6.2 🎉

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

thaJeztah:
Got a report that the fix may not be sufficient, and errors still pop up (I asked them to open a ticket in this repository); https://github.com/moby/moby/issues/42677#issuecomment-889719961

PR #3056, opened 2021-07-01
libct/errors.go: remove ConfigError in favor of ErrInvalidConfig

kolyshkin:
Inspired by https://github.com/opencontainers/runc/pull/3033#discussion_r661966694.

Replace `ConfigError` type (introduced in #3033) with an `ErrInvalidConfig`
variable, for better and more uniform errors returned by libcontainer.
    
This is somewhat complicated, as we need to make all errors returned
from config validator to be `ErrInvalidConfig`, while keeping them
unwrappable.

Implemented by introducing a type which holds an underlying error,
and has the `Is` method that answers that the error is
`ErrInvalidConfig`.

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -16,6 +16,8 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var ErrInvalidConfig = errors.New("invalid configuration")
+
 type Validator interface {
 	Validate(*configs.Config) error
 }
@@ -25,6 +27,24 @@ func New() Validator {
 }
 
 type ConfigValidator struct{}
+
+// invalidConfig type is needed to make any error returned from Validator
+// to be ErrInvalidConfig.
+type invalidConfig struct {
+	err error
+}
+
+func (e *invalidConfig) Error() string {
+	return ErrInvalidConfig.Error() + ": " + e.err.Error()
+}
+
+func (e *invalidConfig) Is(target error) bool {
+	return target == ErrInvalidConfig //nolint:errorlint // this method assumes direct comparison
+}
+
+func (e *invalidConfig) Unwrap() error {
+	return e.err
+}
 
 type check func(config *configs.Config) error
 
@@ -43,7 +63,7 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 	}
 	for _, c := range checks {
 		if err := c(config); err != nil {
-			return err
+			return &invalidConfig{err: err}
 		}
 	}
 	// Relaxed validation rules for backward compatibility
@@ -52,7 +72,7 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 	}
 	for _, c := range warns {
 		if err := c(config); err != nil {
-			logrus.WithError(err).Warnf("invalid configuration")
+			logrus.WithError(err).Warn(ErrInvalidConfig)
 		}
 	}
 	return nil
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -62,20 +62,17 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 // to the container's root filesystem.
 func (v *ConfigValidator) rootfs(config *configs.Config) error {
 	if _, err := os.Stat(config.Rootfs); err != nil {
-		if os.IsNotExist(err) {
-			return fmt.Errorf("rootfs (%s) does not exist", config.Rootfs)
-		}
-		return err
+		return fmt.Errorf("invalid rootfs: %w", err)
 	}
 	cleaned, err := filepath.Abs(config.Rootfs)
 	if err != nil {
-		return err
+		return fmt.Errorf("invalid rootfs: %w", err)
 	}
 	if cleaned, err = filepath.EvalSymlinks(cleaned); err != nil {
-		return err
+		return fmt.Errorf("invalid rootfs: %w", err)
 	}
 	if filepath.Clean(config.Rootfs) != cleaned {
-		return fmt.Errorf("%s is not an absolute path or is a symlink", config.Rootfs)
+		return fmt.Errorf("invalid rootfs: not an absolute path, or a symlink")
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


kolyshkin:
ci failure in centos7 is a epel repo glitch. Will retry later.

internal/configs/validate/validator.go
```diff
@@ -26,6 +28,24 @@ func New() Validator {
...
+type invalidConfig struct {
+	err error
+}
+
+func (e *invalidConfig) Error() string {
+	return ErrInvalidConfig.Error() + ": " + e.err.Error()
+}
+
+func (e *invalidConfig) Is(err error) bool {
+	return err == ErrInvalidConfig
```
kolyshkin, line 42: Linter complains

```
comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error (errorlint)
```

but as it *is* already `Is` method, following the recommendation will result in ∞ recursion ☺

kolyshkin: I was probably wrong about infinite recursion, but the direct comparison here is entirely correct, as the argument is not supposed to be unwrapped (as far as I understand).

Added a nolint annotation, filed https://github.com/polyfloyd/go-errorlint/issues/11

kolyshkin:
Draft until #3049 is merged

kolyshkin:
OK, it looks like #3049 is not going anywhere, so I'll redo this without it.

kolyshkin:
Removed dependency on #3049, rebased, ready for review.

thaJeztah:
> as we need to make all errors returned from config validator to be ErrInvalidConfig

Can you provide more details on why it needs an `ErrInvalidConfig` ?

> while keeping them unwrappable

This confuses me somewhat; if the validator invalidated the config, applying / wrapping it in `ErrInvalidConfig` means the error was handled. Is there code that needs to know the underlying error to handle it?

I feel like it should be either one of the above; don't wrap the error in an `ErrInvalidConfig` (if underlying errors must be handled), or wrap it, and consider `ErrInvalidConfig` to be "it was invalid for (whatever reason), and we're done".

Looking at how it's used;

https://github.com/opencontainers/runc/blob/e918d021399e62442711cc006f99beeae5e3ea1b/libcontainer/factory_linux.go#L259-L261

https://github.com/opencontainers/runc/blob/e918d021399e62442711cc006f99beeae5e3ea1b/utils_linux.go#L246

https://github.com/opencontainers/runc/blob/e918d021399e62442711cc006f99beeae5e3ea1b/utils_linux.go#L413-L416

kolyshkin:
> > as we need to make all errors returned from config validator to be ErrInvalidConfig
> 
> Can you provide more details on why it needs an `ErrInvalidConfig` ?

So, we want to have some predefined errors returned by libcontainer. Those are listed in libcontainer/error.go.

PR #3033 did a good job of dismantling the libcontainer's own error system (in favor of Go 1.13+ errors), while keeping these predefined errors (see PR description).

So, a libcontainer user can check for a particular error using errors.Is, for example:

```go
if errors.Is(err, libcontainer.ErrNotExist)
```

One thing, though, where the above PR falls short, is "invalid config" error. There used to be `ConfigInvalid` error, which is replaced by a `ConfigError` type. Since this is a type, the check for a particular error (to answer a question "do we have an invalid config?") can't be in the same manner as for all other such errors (using `errors.Is`) -- instead you have to use something like `errors.As()`. 

This PR is solving this problem, so you can now use `errors.Is(err, libcontainer.ErrInvalidConfig)`, same as for other "generic" errors.

> > while keeping them unwrappable
> 
> This confuses me somewhat; if the validator invalidated the config, applying / wrapping it in `ErrInvalidConfig` means the error was handled. Is there code that needs to know the underlying error to handle it?

Wrapping it in `ErrInvalidConfig` makes it possible to use `errors.Is(err, ErrInvalidConfig)`, that's it.

> I feel like it should be either one of the above; don't wrap the error in an `ErrInvalidConfig` (if underlying errors must be handled), or wrap it, and consider `ErrInvalidConfig` to be "it was invalid for (whatever reason), and we're done".
> 
> Looking at how it's used;
> 
> https://github.com/opencontainers/runc/blob/e918d021399e62442711cc006f99beeae5e3ea1b/libcontainer/factory_linux.go#L259-L261

This PR removes `ConfigError` type.

> https://github.com/opencontainers/runc/blob/e918d021399e62442711cc006f99beeae5e3ea1b/utils_linux.go#L246
> 
> https://github.com/opencontainers/runc/blob/e918d021399e62442711cc006f99beeae5e3ea1b/utils_linux.go#L413-L416

thaJeztah:
> The "keeping then unwrappable" part means that you can still get the exact underlying error.

Right, so question is: is it needed? Looking at the code in this repository, the only relevant error that's checked is `ErrNotExist`; https://github.com/opencontainers/runc/blob/5547b5774f71f75a088e7432fa961778750a0fbd/delete.go#L58

Which is not on "create", but on "delete" (container already gone situation).

Outside of runc itself, it also looks to be the only error that some users check for (but using the "old" errors); https://grep.app/search?q=libcontainer.E

Mostly wondering; are there practical situations where the caller can implement logic to handle these errors, other than "it's invalid"?

kolyshkin:
@thaJeztah I don't know if anyone needs it. I did it mostly to not lose any functionality compared to what we had before #3033, and generally thinking it's good to have more structured errors.

If that's not needed, I can simplify the whole thing, removing both `ConfigError` type and `ErrInvalidConfig`.

@AkihiroSuda WDYT?

thaJeztah:
> generally thinking it's good to have more structured errors.

Bit on the fence if errors should be wrapped multiple times. I think that's what `github.com/pkg/errors` provided, wasn't it? But the Go maintainers explicitly considered "wrapping an error, means it's handled, and is the end".

Happy to hear others input though.

kolyshkin:
> But the Go maintainers explicitly considered "wrapping an error, means it's handled, and is the end".

I don't quite understand this. Can you share a link or something?

Wrapping an error, in many cases, means adding more context to it. For example, `os` returns many errors wrapped into `os.PathError` or `os.SyscallError`, to add context like file name and/or the failed operation or syscall. This certainly does not mean the error is handled.

In case of this PR, though, we wrap the error into a newly introduced type for the sole reason of making `errors.Is(err, libcontainer.ErrInvalidConfig)` work. I agree that maybe no one is using this, and it is not needed at all, and this is why I suggested to drop both `ConfigError` and `ErrInvalidConfig` in https://github.com/opencontainers/runc/pull/3056#issuecomment-902030069.

But I totally don't understand how wrapping an error means handling it.

kolyshkin:
> I agree that maybe no one is using this, and it is not needed at all, and this is why I suggested to drop both `ConfigError` and `ErrInvalidConfig` in [#3056 (comment)](https://github.com/opencontainers/runc/pull/3056#issuecomment-902030069).

Here's the alternative: https://github.com/opencontainers/runc/pull/3176
PTAL if it makes more sense.

kolyshkin:
Closing in favor of #3176 which is simpler.

PR #3057, opened 2021-07-02
chown cgroup to process uid in container namespace

frasertweedale:
Delegating cgroups to the container enables more complex workloads,
including systemd-based workloads.  The OCI runtime-spec was
recently updated to explicitly admit such delegation, through
specification of cgroup ownership semantics:

  https://github.com/opencontainers/runtime-spec/pull/1123

Pursuant to the updated OCI runtime-spec, change the ownership of
the container's cgroup directory and particular files therein, when
using cgroups v2 and when the cgroupfs is to be mounted read/write.

As a result of this change, systemd workloads can run in isolated
user namespaces on OpenShift when the sandbox's cgroupfs is mounted
read/write.

It might be possible to implement this feature in other cgroup
managers, but that work is deferred.

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -1,8 +1,10 @@
 package systemd
 
 import (
+	"bufio"
 	"fmt"
 	"math"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -288,7 +290,44 @@ func (m *unifiedManager) Apply(pid int) error {
 	if err := fs2.CreateCgroupPath(m.path, m.cgroups); err != nil {
 		return err
 	}
+
+	if c.OwnerUID != nil {
+		filesToChown, err := cgroupFilesToChown()
+		if err != nil {
+			return err
+		}
+
+		for _, v := range filesToChown {
+			err := os.Chown(m.path+"/"+v, *c.OwnerUID, -1)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
 	return nil
+}
+
+// The kernel exposes a list of files that should be chowned to the delegate
+// uid in /sys/kernel/cgroup/delegate.  If the file is not present
+// (Linux < 4.15), use the initial values mentioned in cgroups(7).
+func cgroupFilesToChown() ([]string, error) {
+	filesToChown := []string{"."} // the directory itself must be chowned
+	const cgroupDelegateFile = "/sys/kernel/cgroup/delegate"
+	f, err := os.Open(cgroupDelegateFile)
+	if err == nil {
+		defer f.Close()
+		scanner := bufio.NewScanner(f)
+		for scanner.Scan() {
+			filesToChown = append(filesToChown, scanner.Text())
+		}
+		if err := scanner.Err(); err != nil {
+			return nil, fmt.Errorf("error reading %s: %w", cgroupDelegateFile, err)
+		}
+	} else {
+		filesToChown = append(filesToChown, "cgroup.procs", "cgroup.subtree_control", "cgroup.threads")
+	}
+	return filesToChown, nil
 }
 
 func (m *unifiedManager) Destroy() error {
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -41,6 +41,13 @@ type Cgroup struct {
 
 	// Rootless tells if rootless cgroups should be used.
 	Rootless bool
+
+	// The host UID that should own the cgroup, or nil to accept
+	// the default ownership.  This should only be set when the
+	// cgroupfs is to be mounted read/write.
+	// Not all cgroup manager implementations support changing
+	// the ownership.
+	OwnerUID *int `json:"owner_uid,omitempty"`
 }
 
 type Resources struct {
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -366,6 +366,49 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 			}
 		}
 	}
+
+	// Set the host UID that should own the container's cgroup.
+	// This must be performed after setupUserNamespace, so that
+	// config.HostRootUID() returns the correct result.
+	//
+	// Only set it if the container will have its own cgroup
+	// namespace and the cgroupfs will be mounted read/write.
+	//
+	hasCgroupNS := config.Namespaces.Contains(configs.NEWCGROUP) && config.Namespaces.PathOf(configs.NEWCGROUP) == ""
+	hasRwCgroupfs := false
+	if hasCgroupNS {
+		for _, m := range config.Mounts {
+			if m.Source == "cgroup" && filepath.Clean(m.Destination) == "/sys/fs/cgroup" && (m.Flags&unix.MS_RDONLY) == 0 {
+				hasRwCgroupfs = true
+				break
+			}
+		}
+	}
+	processUid := 0
+	if spec.Process != nil {
+		// Chown the cgroup to the UID running the process,
+		// which is not necessarily UID 0 in the container
+		// namespace (e.g., an unprivileged UID in the host
+		// user namespace).
+		processUid = int(spec.Process.User.UID)
+	}
+	if hasCgroupNS && hasRwCgroupfs {
+		ownerUid, err := config.HostUID(processUid)
+		// There are two error cases; we can ignore both.
+		//
+		// 1. uidMappings is unset.  Either there is no user
+		//    namespace (fine), or it is an error (which is
+		//    checked elsewhere).
+		//
+		// 2. The user is unmapped in the user namespace.  This is an
+		//    unusual configuration and might be an error.  But it too
+		//    will be checked elsewhere, so we can ignore it here.
+		//
+		if err == nil {
+			config.Cgroups.OwnerUID = &ownerUid
+		}
+	}
+
 	if spec.Process != nil {
 		config.OomScoreAdj = spec.Process.OOMScoreAdj
 		config.NoNewPrivileges = spec.Process.NoNewPrivileges
--- a/tests/integration/cgroup_delegation.bats
+++ b/tests/integration/cgroup_delegation.bats
@@ -1,0 +1,61 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function teardown() {
+	teardown_bundle
+}
+
+function setup() {
+	requires root cgroups_v2 systemd
+
+	setup_busybox
+
+	# chown test temp dir to allow host user to read it
+	chown 100000 "$ROOT"
+
+	# chown rootfs to allow host user to mkdir mount points
+	chown 100000 "$ROOT"/bundle/rootfs
+
+	set_cgroups_path
+
+	# configure a user namespace
+	update_config '   .linux.namespaces += [{"type": "user"}]
+			| .linux.uidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]
+			| .linux.gidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]
+			'
+}
+
+@test "runc exec (cgroup v2, ro cgroupfs, new cgroupns) does not chown cgroup" {
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroup_chown
+	[ "$status" -eq 0 ]
+
+	runc exec test_cgroup_chown sh -c "stat -c %U /sys/fs/cgroup"
+	[ "$status" -eq 0 ]
+	[ "$output" = "nobody" ] # /sys/fs/cgroup owned by unmapped user
+}
+
+@test "runc exec (cgroup v2, rw cgroupfs, inh cgroupns) does not chown cgroup" {
+	set_cgroup_mount_writable
+
+	# inherit cgroup namespace (remove cgroup from namespaces list)
+	update_config '.linux.namespaces |= map(select(.type != "cgroup"))'
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroup_chown
+	[ "$status" -eq 0 ]
+
+	runc exec test_cgroup_chown sh -c "stat -c %U /sys/fs/cgroup"
+	[ "$status" -eq 0 ]
+	[ "$output" = "nobody" ] # /sys/fs/cgroup owned by unmapped user
+}
+
+@test "runc exec (cgroup v2, rw cgroupfs, new cgroupns) does chown cgroup" {
+	set_cgroup_mount_writable
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroup_chown
+	[ "$status" -eq 0 ]
+
+	runc exec test_cgroup_chown sh -c "stat -c %U /sys/fs/cgroup"
+	[ "$status" -eq 0 ]
+	[ "$output" = "root" ] # /sys/fs/cgroup owned by root (of user namespace)
+}


libcontainer/cgroups/systemd/v2.go
```diff
@@ -292,6 +296,32 @@ func (m *unifiedManager) Apply(pid int) error {
...
 	}
+
+	goodToChown := []string{"cgroup.procs", "cgroup.subtree_control", "cgroup.threads"}
+	member := func(haystack []string, needle string) bool {
+		for _, v := range haystack {
+			if v == needle {
+				return true
+			}
+		}
+		return false
```
kolyshkin, line 307: Perhaps it's easier to use `map[string]struct{}` here.

Or just be direct
```
if name == "cgroup.procs" || name == "cgroup.subtree_control" || name == ....
```

libcontainer/cgroups/systemd/v2.go
```diff
@@ -292,6 +296,32 @@ func (m *unifiedManager) Apply(pid int) error {
...
+	}
+	if m.uid >= 0 {
+		err := filepath.Walk(m.path, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if path == m.path || member(goodToChown, info.Name()) {
+				return os.Chown(path, m.uid, -1)
+			} else {
```
kolyshkin, line 316: `else` is redundant here

libcontainer/cgroups/systemd/v2.go
```diff
@@ -27,14 +27,18 @@ type unifiedManager struct {
 	path     string
 	rootless bool
 	dbus     *dbusConnManager
+	// host UID corresponding to UID 0 in container namespace;
+	// or -1 if cgroup should not be chown'd
+	uid int
 }
 
-func NewUnifiedManager(config *configs.Cgroup, path string, rootless bool) cgroups.Manager {
+func NewUnifiedManager(config *configs.Cgroup, path string, uid int, rootless bool) cgroups.Manager {
```
kolyshkin, line 35: I'm afraid we can't change this function signature, as other software (kubernetes, cri-o, possibly cAdvisor etc.) depend on it (unfortunately).

Maybe you can add a separate method to set the uid?

libcontainer/factory_linux.go
```diff
@@ -243,7 +243,7 @@ type LinuxFactory struct {
 	Validator validate.Validator
 
 	// NewCgroupsManager returns an initialized cgroups manager for a single container.
-	NewCgroupsManager func(config *configs.Cgroup, paths map[string]string) cgroups.Manager
+	NewCgroupsManager func(config *configs.Cgroup, paths map[string]string, pid int) cgroups.Manager
```
kolyshkin, line 246: typo (`s/pid/uid/`)?

libcontainer/cgroups/systemd/v2.go
kolyshkin: or a `WithUID()` modifier method.

kolyshkin:
In general this kinda makes sense, although I'm afraid systemd might overwrite that at some random moment.

This also needs test(s).

libcontainer/factory_linux.go
frasertweedale: yep, thanks.

libcontainer/cgroups/systemd/v2.go
frasertweedale: Thanks, I will do as you suggest.

libcontainer/cgroups/systemd/v2.go
frasertweedale: OK, I'll use a map.

libcontainer/cgroups/systemd/v2.go
frasertweedale: I prefer if / else to be balanced.  Does it violate a formal style guide?

frasertweedale:
> In general this kinda makes sense, although I'm afraid systemd might overwrite that at some random moment.
> 
> This also needs test(s).

You can use systemd transient unit API to set a different user.  Unfortunately, the uid needs to
have a corresponding passwd entity or else systemd refuses to proceed.

I filed an issue against systemd to relax this requirement.  It was provisionally rejected:

libcontainer/cgroups/systemd/v2.go
```diff
@@ -27,14 +27,18 @@ type unifiedManager struct {
 	path     string
 	rootless bool
 	dbus     *dbusConnManager
+	// host UID corresponding to UID 0 in container namespace;
+	// or -1 if cgroup should not be chown'd
+	uid int
```
AkihiroSuda, line 32: `-1` -> `nil` , `int` -> `*int`

libcontainer/cgroups/systemd/v2.go
```diff
@@ -292,6 +296,32 @@ func (m *unifiedManager) Apply(pid int) error {
 	if err := fs2.CreateCgroupPath(m.path, m.cgroups); err != nil {
 		return err
 	}
+
+	goodToChown := []string{"cgroup.procs", "cgroup.subtree_control", "cgroup.threads"}
```
AkihiroSuda, line 300: Needs to be documented

AkihiroSuda (changes requested):
For security reason, the default behavior should not be changed.
We can probably have a feature gate as an annotation 

https://github.com/containers/crun/blob/main/crun.1.md#extensions-to-oci

https://github.com/opencontainers/runc/blob/master/docs/systemd.md#auxiliary-properties

frasertweedale:
> For security reason, the default behavior should not be changed.
> We can probably have a feature gate as an annotation
> 
> https://github.com/containers/crun/blob/main/crun.1.md#extensions-to-oci
> 
> https://github.com/opencontainers/runc/blob/master/docs/systemd.md#auxiliary-properties

User namespaces is already behind an annotation (`io.kubernetes.cri-o.userns-mode`).  And user namespaces are somewhat hobbled without the ability to manage cgroups (for the workloads that I care about anyway).  So, do we really need a separate annotation to gate this feature?  I don't mind if the answer is yes, but something to consider...

AkihiroSuda:
I'm talking about OCI Runtime Spec annotations, not about Kubernetes annotations.

CRI-O may set the suggest OCI Runtime Spec annotation by default at their own risk.

frasertweedale:
@AkihiroSuda thanks for the clarification.  Fair enough.  I'll gate it.  I'll try to make all the suggested changes next week.

frasertweedale:
@AkihiroSuda what would be an appropriate key/value for the annotation?  `"org.opencontainers.runc.chown-cgroup": "{true,false}"`

AkihiroSuda:
>  "org.opencontainers.runc.chown-cgroup": "{true,false}"

SGTM, but in the long term we should have proper JSON entry in OCI Runtime Spec https://github.com/opencontainers/runtime-spec

frasertweedale:
> > "org.opencontainers.runc.chown-cgroup": "{true,false}"
> 
> SGTM, but in the long term we should have proper JSON entry in OCI Runtime Spec https://github.com/opencontainers/runtime-spec

Better would be to adjust the semantics such that the container process can create scopes/slices in its own cgroup, and manage its child processes/threads.  But that is a discussion to be had over there, not here :)

libcontainer/cgroups/systemd/v2.go
kolyshkin: Not really, but some linters do complain, and personally I find it redundant, too -- because of `return`.

Feel free to keep though.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -292,6 +296,32 @@ func (m *unifiedManager) Apply(pid int) error {
...
+	member := func(haystack []string, needle string) bool {
+		for _, v := range haystack {
+			if v == needle {
+				return true
+			}
+		}
+		return false
+	}
+	if m.uid >= 0 {
+		err := filepath.Walk(m.path, func(path string, info os.FileInfo, err error) error {
```
kolyshkin, line 310: Since you're only using info.Name(), it probably make sense to switch to filepath.WalkDir. Problem is, it is only added in go 1.16. So maybe add a TODO item:
```go
// TODO: switch to filepath.WalkDir once go < 1.16 is not supported.
```

frasertweedale: Yep, I originally used WalkDir then had to change for that reason.  Happy to add the comment.

frasertweedale:
Thanks @kolyshkin and @AkihiroSuda for your reviews.  I implemented all of your requested changes.  I still need to add test and document the feature.

frasertweedale:
Ping @kolyshkin and @AkihiroSuda - are you able to re-review this PR soon?

frasertweedale:
I published a video demonstration of using this feature to run systemd-based workloads in isolated user namespaces on OpenShift: https://www.youtube.com/watch?v=OGAVvIJwmd0

Also a blog post that follows the same procedure: https://frasertweedale.github.io/blog-redhat/posts/2021-07-22-openshift-systemd-workload-demo.html

frasertweedale:
@kolyshkin, @AkihiroSuda: weekly nudge.  Anything I can do to help with review or testing?

libcontainer/cgroups/systemd/v2.go
```diff
@@ -292,6 +304,38 @@ func (m *unifiedManager) Apply(pid int) error {
...
+	//
+	// We follow systemd's lead and chown only those same files.
+	goodToChown := map[string]struct{}{
+		"cgroup.procs":           struct{}{},
+		"cgroup.subtree_control": struct{}{},
+		"cgroup.threads":         struct{}{},
+	}
+	if m.uid != nil {
+		// TODO: when go < 1.16 is no longer supported, switch to filepath.WalkDir
+		err := filepath.Walk(m.path, func(path string, info os.FileInfo, err error) error {
```
kolyshkin, line 322: So, are we expecting to have sub-directories here? AFAICS there can be none, or am I missing something?

libcontainer/factory_linux.go
```diff
@@ -327,6 +339,16 @@ func (l *LinuxFactory) Load(id string) (Container, error) {
...
+	if state.Config.ChownCgroup {
+		uid, err := state.Config.HostRootUID()
+		if err != nil {
+			return nil, err
+		}
+		err = c.cgroupManager.SetUID(&uid)
+		if err != nil {
+			return nil, err
+		}
+	}
```
kolyshkin, line 351: 1. Why this code needs to be in `Load`? I can't think of any code paths that can use it...

2. If this is indeed required here, maybe it's better to have a function to be called from both `Load` and `Create`.

kolyshkin:
Overall I think it's a great idea to have the option of giving the container user an ability to create sub-cgroups.

I'd rather call the method `Chown` than `SetUID` as the latter has a special meaning. Using a pointer to int as its argument does not make sense to me, too (storing at as a pointer, though, to distinguish between "0" and "not set", kinda does). These things are not too important, of course, and are trivial to fix.

It bothers me that this is a new public method of cgroupManager, and out of 4 implementations that we have in libcontainer, only one implements it. Leaving cgroup v1 aside, is there anything that prevents fs2 manager to have it?

In general, I think, the API provided by the cgroup manager is not adequate for this task, and this is why this looks strange. I was thinking about changing that but alas don't have a clear idea how.

If we could do the chown in `Set` instead, it'd straighten things up a lot -- instead of a special method, we'd just add a flag to container Resources (like we did for, say, SkipDevices).

libcontainer/cgroups/systemd/v2.go
frasertweedale: Just the one level at this time.  But I thought it better not to couple tightly to this assumption.  Do you prefer to explicitly chown exactly the 4 files needed at this time (the cgroup dir, and the 3 files within it).

frasertweedale:
@kolyshkin thanks for your feedback.  I'll try and redesign it to use an extra field in `Resources`, which as you point out will avoid any API changes.

frasertweedale:
@kolyshkin OK, now I have a better comprehension of where `Apply` and `Set` get invoked.  I feel strongly that the setting the cgroup ownership should be a behaviour of `Apply`, not `Set`.  This is because there are alternative ways - besides `runc` performing `chown(2)` on relevant files - that the cgroup ownership could be set.  In particular, you can specify an alternative cgroup ownership via the systemd transient unit API.  In fact, this would be a nice way to do it except for one aspect of systemd's behaviour related to `passwd` entities (and they are resisting changing this behaviour; see https://github.com/systemd/systemd/issues/19781).

Nevertheless, to better support such an approach if systemd should relax its checks, or to support alternative approaches, I think `Apply` rather than `Set(Resources)` is the better approach.

To avoid the API changes and confusing propagation of the UID setting, perhaps we should add the cgroup owner UID as a field of `configs.Cgroup` rather than `configs.Resources`.  For now, I will continue along that trajectory.

kolyshkin:
@frasertweedale I was also thinking about adding `Create`.

There are two reasons for that:

1. In many cases (including kubernetes using runc's libcontainer/cgroups) `Apply(-1)` is used as a surrogate for `Create`.
2. In many cases `Apply(-1)` is followed by `Set`, as the cgroup is initially created with no limits set.

So, perhaps we can add `Create()` with the same argument as `Set` (so it will create a cgroup and set its resources). Now, `Apply` can be renamed to e.g. `AddProcess`, and `Set` can become `Update`.

I understand this is mostly out of scope for this PR, and is a major change of the API, but this PR creates a challenge that the existing API is seemingly failing to solve.

frasertweedale:
Note well: the only CI job that currently executes (i.e. does not skip) the test for this feature is the Cirrus CI Fedora 34 job.

frasertweedale:
@kolyshkin I updated the PR.  Using a new field in the `configs.Cgroup` struct simplified the implementation a lot.  Now there are no new methods or other API changes (besides the new struct field).

I also changed the chown behaviour to directly chown the 4 files of interest, instead of using `os.Walk`.

I don't see any reason this can't be implemented for fs2.  But it is not my use case (OpenShift, which uses systemd manager), so I'd rather leave it out of this commit.

libcontainer/cgroups/systemd/v2.go
frasertweedale: I changed it to explicitly mention (and chown) the dir and 3 files needed; no more `Walk`.

libcontainer/factory_linux.go
frasertweedale: Latest version of patch conveys the owner UID in the `configs.Cgroup` struct.  No changes needed in `factory_linux.go` anymore.

frasertweedale:
@kolyshkin *nudge*.  I look forward to review of updated patch, when you have the capacity.  Thanks!

frasertweedale:
*nudge*: periodic request for review.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -292,6 +292,24 @@ func (m *unifiedManager) Apply(pid int) error {
...
 	}
+
+	// When systemd creates a unit owned by a user other than the
+	// user running the service manager (usually root), it chowns
+	// the cgroup directory and the three files mentioned below.
+	// This enables the service to create new scopes/slices and set
+	// the cgroup for its child processes/threads.
+	//
+	// We follow systemd's lead and chown only those same files.
+	filesToChown := []string{".", "cgroup.procs", "cgroup.subtree_control", "cgroup.threads"}
```
kolyshkin, line 303: I remembered there's a cgroup file which lists these three, but can't find it now. Maybe I remember wrong :)

frasertweedale: @kolyshkin I am not aware of it.

frasertweedale:
CRI-O doesn't set the annotation, but it does propagate Pod annotations to the OCI runtime.  So for the time being you have to set two annotations.  Feel free to propose the CRI-O change, but it probably doesn't make sense to do so until this is merged, in case the annotation name changes.

m-yosefpor:
@frasertweedale thank you for your explanation. So for now setting both annotations will work for us.

tests/integration/cgroups.bats
```diff
@@ -119,6 +119,59 @@ function setup() {
...
+	# chown test temp dir to allow host user to read it
+	chown 100000 "$ROOT"
+
+	# chown rootfs to allow host user to mkdir mount points
+	chown 100000 "$ROOT"/bundle/rootfs
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_cgroup_chown
+	[ "$status" -eq 0 ]
+
+	runc exec test_cgroup_chown sh -c "ls -ld /sys/fs/cgroup | awk '{print \$3}'"
```
kolyshkin, line 142: nit: no need to escape `$` since it's inside single quotes.

libcontainer/specconv/spec_linux.go
```diff
@@ -306,6 +306,18 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 			}
 		}
 	}
+
+	// Set the host UID that should own the container's cgroup.
+	// This must be performed after setupUserNamespace, so that
+	// config.HostRootUID() returns the correct result.
+	if v := spec.Annotations["org.opencontainers.runc.chown-cgroup"]; v == "true" {
```
kolyshkin, line 313: What if crun wants to add support for this? Can we use, say, `run.oci.cgroup.chown` or something like this?

@giuseppe PTAL

frasertweedale: @kolyshkin actually, `crun` already has this behaviour:

```
% jq '.linux.uidMappings, .linux.cgroupsPath, .process.user' < config.json
[
  {
    "containerID": 0,
    "hostID": 1000,
    "size": 1
  },
  {
    "containerID": 1,
    "hostID": 100000,
    "size": 65536
  }
]
"user.slice:runc:bar"
{
  "uid": 0,
  "gid": 0
}
% sudo crun --systemd-cgroup run test 
... # in separate shell
% ls -alin /sys/fs/cgroup/user.slice/test-test.scope/container |grep 1000
40139 drwxr-xr-x. 2 1000 1000 0 Sep 10 16:09 .
40141 -rw-r--r--. 1 1000 1000 0 Sep 10 16:09 cgroup.procs
40144 -rw-r--r--. 1 1000 1000 0 Sep 10 16:09 cgroup.subtree_control
40142 -rw-r--r--. 1 1000 1000 0 Sep 10 16:09 cgroup.threads
40176 -rw-r--r--. 1 1000 1000 0 Sep 10 16:09 memory.oom.group
```

giuseppe: yes, crun always chowns the cgroup to root in the user namespace

frasertweedale: @kolyshkin with this in mind, what do you think about removing the annotation and just having this as the default behaviour in `runc` also?

thaJeztah: I'm definitely not a fan of using annotations for this; it feels like a band-aid / hack to work around missing functionality in the runtime spec. If there _would be_ an annotation, it should be one that's defined in the runtime spec (and image-spec?), which should describe how runtimes should process these annotations.

What are the consequences of setting this as default? Does it affect non-systemd-in-container situations?

thaJeztah: (apologies for going a bit orthogonal here)

I understand the "systemd in a container" use-case, but recall discussions from the past where some of those configuration changes needed were _beneficial_ for that use-case, but could have a _negative_ impact on other scenarios. I also recall other configuration changes (e.g. the `CONTAINER` env-var, which so far hasn't been described in the runtime spec) are needed to make systemd work properly inside a container.
Thinking out loud here, but would it be an idea of defining (e.g.) a concept of "profiles" in the runtime spec, that (when set) instructs the runtime to make the required changes to facilitate systemd? (`profile=systemd`)

frasertweedale: @thaJeztah as for non-systemd workloads, well, it means they can manage their own cgroups too.  I do not think it would break anything (never say never...)

As for changes to support systemd that break or impair other use cases, do you have examples?
In my case, this experimental version of runc, combined with the CRI-O user namespace support and worker nodes configured with cgroupsv2, is sufficient to run systemd-based container on OpenShift.

I am wrestling with some other problems with more advanced systemd features (e.g. *systemd-tmpfiles* not playing well with `emptyDir` mounts) but these can be worked around.

frasertweedale: > I'm definitely not a fan of using annotations for this; it feels like a band-aid / hack to work around missing functionality in the runtime spec. If there _would be_ an annotation, it should be one that's defined in the runtime spec (and image-spec?), which should describe how runtimes should process these annotations.

I will propose a clarification to the OCI runtime spec to specify that cgroups created by the OCI runtime should be owned by the container's uid 0.  At the very least some implementation notes/advice should be adopted into the OCI documents.

frasertweedale: runtime-spec PR: https://github.com/opencontainers/runtime-spec/pull/1123

thaJeztah: Thanks!

> As for changes to support systemd that break or impair other use cases, do you have examples?


Don't have direct examples, and I would have to dig into history for that.


> In my case, this experimental version of runc, combined with the CRI-O user namespace support and worker nodes configured with cgroupsv2, is sufficient to run systemd-based container on OpenShift.


@giuseppe may know more about the implementation, but it's possible that CRI-O and/or OpenShift automatically apply some configs and/or hooks to facilitate systemd inside a container.

There's a long history on systemd and containers with two camps (_"every container should have systemd as PID1"_ vs _"systemd doesn't belong in a container"_), which unfortunately led to a situation where either the user has to go through the (sometimes complicated) steps to make systemd work, or runtimes had to do preparations for this.

My personal opinion is that neither of those camps are "right", as it really depends on the use-case. In general, the default for _any_ container should be to have as little privileges as possible, and no _assumptions_ / _preparations_ should be made for a specific init system / tooling inside the container _by default_. _BUT_ I also think that it should be as easy as possible to be _able_ to use systemd (or other tools) in a container, and to be able to do so in a "portable" / "standard" way.

The OCI specifications are there to enhance portability; this is where I think "profiles" (or give it any other name) would be able to help. The OCI could define a "well known" profile for `systemd`, which would describe what preparations and/or changes to the container's spec runtimes are expected to make for containers using such profile (which could include permissions, environment variables, additional mounts and/or capabilities, etc.).

giuseppe: I'd say this feature is needed more in general for enabling cgroup delegation when using a user namespace.

I don't think there should be any emphasis on systemd, since it is just an user of such feature.

For example on a cgroup v2 system:
```
$ podman --runtime /usr/bin/runc run --privileged --uidmap 0:1:1000 --rm fedora sh -c 'ls -lnd /sys/fs/cgroup; mkdir /sys/fs/cgroup/foo && echo created'
drwxr-xr-x. 2 65534 65534 0 Sep 13 08:02 /sys/fs/cgroup
mkdir: cannot create directory '/sys/fs/cgroup/foo': Permission denied
```

crun automatically chowns the cgroup to root in the userns:
```
$ podman --runtime /usr/bin/crun run --privileged --uidmap 0:1:1000 --rm fedora sh -c 'ls -lnd /sys/fs/cgroup; mkdir /sys/fs/cgroup/foo && echo created'
drwxr-xr-x. 2 0 0 0 Sep 13 08:02 /sys/fs/cgroup
created
```

frasertweedale: @giuseppe indeed, systemd workload is just one notable use case.

Separately, your input over at https://github.com/opencontainers/runtime-spec/pull/1123 would be appreciated.

frasertweedale:
OK, what is the next step for this PR.

I don't see any more objections to the overall implementation approach.
There is some debate/question over whether we need the annotation or not.

I also created a [runtime-spec PR](https://github.com/opencontainers/runtime-spec/pull/1123) to propose a formal semantics for cgroup ownership.
It looks like there will be a debate about whether to have an explicit config.  It may take
some time to resolve that matter.

Could we merge this PR in its current form, then remove or rename the annotation after
the runtime-spec discussion has been resolved?  Or, should we just remove the
annotation now, and try and convince runtime-spec to come to the party?  If there is a a doubt
let's be conservative and keep the annotation for now, so that this can be merged.

libcontainer/specconv/spec_linux.go
thaJeztah: If this should be set unconditionally (and if it can be done without risk of introducing breaking changes), I'm definitely +1 on doing so; I agree that  https://github.com/opencontainers/runtime-spec/pull/1123 would be good for that (to define that in the spec).

frasertweedale:
Could maintainers with authority to merge (I'm not sure who those are) please decide: do we remove the annotation now, or can we merge it as-is now, and potentially follow-up to remove the annotation later, as the runtime-spec outcome becomes clear?

This PR has dragged on a long time.  We need a clear decision from the maintainers so that it can be merged, even if that
decision is "we'll decide later whether to keep the annotation or make this behaviour unconditional".

ping @giuseppe @kolyshkin @AkihiroSuda @mrunalp

libcontainer/specconv/spec_linux.go
AkihiroSuda: If we are going to enable chowning by default, we will need to have a way to opt-out for that.

I guess skipping chowning when the mount is ro (https://github.com/opencontainers/runtime-spec/pull/1123#discussion_r707393997 , suggested by @giuseppe) would be fine.

frasertweedale: I have updated the PR to remove the annotation, and only chown the cgroup when the cgroupfs is mounted `rw`.

tests/integration/cgroups.bats
frasertweedale: It is inside single quotes... inside double quotes :)  So the escape is necessary.  For example:
```shell
% echo "'$0'"
'zsh'
```

cyphar:
On paper I'd be okay with having an annotation, though unlike crun we've historically not really used annotations for extensions (mainly because runc is meant to be a reference implemetnation of the spec so out-of-spec stuff is in a bit of a grey area and extensions based on annotations could very quickly become an out-of-spec defacto standard like most other runc behaviour has become over the years). Unfortunately I don't see how we could add any features without this happening (and the horses have long-since bolted and ships long-since set sail on this topic -- so we're not going to solve it here).

As for making this behaviour the default, imho this entirely depends on whether it will break some existing programs that have faulty handling of cgroups. The other issue is that practically using annotations is somewhat difficult because image labels don't get translated to runtime labels by Docker AFAIK (podman/crio do though).

frasertweedale:
> On paper I'd be okay with having an annotation, though unlike crun we've historically not really used annotations for extensions (mainly because runc is meant to be a reference implementation of the spec so out-of-spec stuff is in a bit of a grey area and extensions based on annotations could very quickly become an out-of-spec defacto standard like most other runc behaviour has become over the years). Unfortunately I don't see how we could add any features without this happening (and the horses have long-since bolted and ships long-since set sail on this topic -- so we're not going to solve it here).
> 
> As for making this behaviour the default, imho this entirely depends on whether it will break some existing programs that have faulty handling of cgroups. The other issue is that practically using annotations is somewhat difficult because image labels don't get translated to runtime labels by Docker AFAIK (podman/crio do though).

@cyphar thanks for your feedback.  How to judge "whether it will break some existing programs"?  In the absence of a specific counterexample, it is an academic exercise in estimating the likelihood of breakage.  Or else, there is the "scream test" - merge it, see who screams, and be prepared to roll back or gate with an annotation if it's worse than anticipated.  Realistically, what do you think the risks of breakage are?

Personally I don't see much risk because this change strictly allows the container greater access on parts of its cgroupfs than it would otherwise have.  Container processes would have to be specifically inspecting the cgroupfs ownership or attempting/evaluating access to behave differently.  Or else, perhaps obscure UID maps could cause breakage (e.g. no uid 0 in the container's user namespace).

cyphar:
>  How to judge "whether it will break some existing programs"? In the absence of a specific counterexample, it is an academic exercise in estimating the likelihood of breakage. Or else, there is the "scream test" - merge it, see who screams, and be prepared to roll back or gate with an annotation if it's worse than anticipated. Realistically, what do you think the risks of breakage are?

This is a good question and aside from testing with a bunch of different programs (trying to get a representative sample) there isn't much more we can do than the scream test. But I think we should just merge this with the annotation-based flag and then when it is used by some downstreams we can see whether it caused any breakages. I hope it won't, and it seems somewhat unlikely but it's quite hard to predict what containerised programs try to do.

frasertweedale:
I pushed another update in response to https://github.com/opencontainers/runtime-spec/pull/1123#issuecomment-925464201.  Now the chown behaviour is also restricted to when runc will create a new cgroup namespace for the container.

frasertweedale:
In relation to @cyphar's recent comment - should I put the annotation guard back in?  I would like to hear from others on this - there seem to be people on both sides, within the last week.

frasertweedale:
I updated this PR to reflect the recent changes to the runtime-spec cgroup ownership
semantics proposed in PR https://github.com/opencontainers/runtime-spec/pull/1123. In particular:

- chown cgroup to host uid that maps to process.user.uid in container namespace (instead of uid 0 unconditionally)
- read list of files to chown from /sys/kernel/cgroup/delegate (with fallback if not exist)

I added these as separate commits to make them easier to review and, if necessary, modify or revert.
If desired, they can be squashed before this PR gets merged - just let me know.

Thanks, I look forward to your feedback.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -288,6 +290,36 @@ func (m *unifiedManager) Apply(pid int) error {
 	if err := fs2.CreateCgroupPath(m.path, m.cgroups); err != nil {
 		return err
 	}
+
+	// The kernel exposes a list of files that should chowned to
```
jfroy, line 294: Perhaps move the code to compute `filesToChown` inside the `if c.OwnerUID != nil {` scope?

libcontainer/configs/cgroup_linux.go
```diff
@@ -41,6 +41,13 @@ type Cgroup struct {
 
 	// Rootless tells if rootless cgroups should be used.
 	Rootless bool
+
+	// The host UID that should own the cgroup, or nil to accept
+	// the default ownership.  This should only be set when the
+	// cgroupfs is to be mounted read/write.
+	// Not all cgroup manager implementations support changing
+	// the ownership.
+	OwnerUID *int `json:"owner_uid,omitempty"`
```
jfroy, line 50: Is there a need for group as well? crun sets both.

tests/integration/cgroups.bats
jfroy: You could also use `stat`. Perhaps a bit more simple than ls + awk.

```shell
stat -c "%u %g" "/sys/fs/cgroup"
[ "$output" = "0 0" ]
```

libcontainer/configs/cgroup_linux.go
frasertweedale: I'm not sure.  I don't need it for my use case.  It would be best to discuss it in the OCI runtime-spec PR (https://github.com/opencontainers/runtime-spec/pull/1123) where we are formalising a semantics for container cgroup ownership.

libcontainer/cgroups/systemd/v2.go
frasertweedale: I shall implement this change in the next push.

frasertweedale:
@jfroy thanks for your review.  I have implemented your proposed changes.  Discussion about whether group ownership should also be changed, it would be better to have that discussion over in https://github.com/opencontainers/runtime-spec/pull/1123.

frasertweedale:
The failing CI job (vagrant DISTRO:fedora) is failing when the commit tree is equal to that of master @ cbd725e6ef8a8c09ceea92b08d2530fc5af9a8ef.  See https://github.com/opencontainers/runc/runs/4149528511 for an
example.  Perhaps there is some defect in the CI environment?

The tests for the behaviour introduced by this PR are passing.  See https://cirrus-ci.com/task/5113759288000512:
```
...
ok 7 runc exec (cgroup v2, ro cgroupfs, new cgroupns) does not chown cgroup
ok 8 runc exec (cgroup v2, rw cgroupfs, inh cgroupns) does not chown cgroup
ok 9 runc exec (cgroup v2, rw cgroupfs, new cgroupns) does chown cgroup
...
```

AkihiroSuda:
The issues of Fedora CI are tracked in https://github.com/opencontainers/runc/issues/3266

frasertweedale:
Rebased, to absorb the CI fixes from main branch.

libcontainer/cgroups/systemd/v2.go
kolyshkin: This is also documented in the kernel docs: 
https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#model-of-delegation

(no need to add a link I guess, just noticing)

libcontainer/specconv/spec_linux.go
```diff
@@ -366,6 +366,39 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
...
+	// namespace and the cgroupfs will be mounted read/write.
+	//
+	hasCgroupNS := config.Namespaces.Contains(configs.NEWCGROUP) && config.Namespaces.PathOf(configs.NEWCGROUP) == ""
+	hasRwCgroupfs := false
+	for _, m := range config.Mounts {
+		if m.Source == "cgroup" && m.Destination == "/sys/fs/cgroup" && (m.Flags&unix.MS_RDONLY) == 0 {
+			hasRwCgroupfs = true
+			break
+		}
+	}
```
kolyshkin, line 384: In case `hasCgroupNS` is false, the results of this check is useless. This path can be quite common (for cgroup v1) so it probably makes sense to optimize it.

tests/integration/cgroups.bats
```diff
@@ -119,6 +119,86 @@ function setup() {
...
 
+@test "runc exec (cgroup v2, ro cgroupfs, new cgroupns) does not chown cgroup" {
+	requires root cgroups_v2 systemd
+
+	set_cgroups_path
+
+	# configure a user namespace
+	update_config '.linux.namespaces += [{"type": "user"}]'
+	update_config '.linux.uidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]'
+	update_config '.linux.gidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]'
```
kolyshkin, line 130: A single update_config statement is enough. Not important, but we tend to use a single one elsewhere.

tests/integration/cgroups.bats
```diff
@@ -119,6 +119,86 @@ function setup() {
...
+@test "runc exec (cgroup v2, rw cgroupfs, inh cgroupns) does not chown cgroup" {
+	requires root cgroups_v2 systemd
+
+	set_cgroups_path
+	set_cgroup_mount_writable
+
+	# configure a user namespace
+	update_config '.linux.namespaces += [{"type": "user"}]'
+	update_config '.linux.uidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]'
+	update_config '.linux.gidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]'
```
kolyshkin, line 155: ditto

tests/integration/cgroups.bats
```diff
@@ -119,6 +119,86 @@ function setup() {
...
+
+	# make cgroupfs mount read/write
+	update_config '.mounts |= map(select(.source == "cgroup").options |= map(select(. != "ro")) + ["rw"])'
```
kolyshkin, line 186: Isn't it identical to `set_cgroup_mount_writable` above (and thus redundant)?

tests/integration/cgroups.bats
```diff
@@ -119,6 +119,86 @@ function setup() {
...
+	# configure a user namespace
+	update_config '.linux.namespaces += [{"type": "user"}]'
+	update_config '.linux.uidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]'
+	update_config '.linux.gidMappings += [{"hostID": 100000, "containerID": 0, "size": 65536}]'
+
+	# chown test temp dir to allow host user to read it
+	chown 100000 "$ROOT"
+
+	# chown rootfs to allow host user to mkdir mount points
+	chown 100000 "$ROOT"/bundle/rootfs
```
kolyshkin, line 136: As you're doing this chowns and update_config in every test, maybe it make sense to move these to a separate file (e.g. `cgroup_delegate.bats`) and move the common part to `setup()` function.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -290,15 +291,27 @@ func (m *unifiedManager) Apply(pid int) error {
...
-	//
-	// We follow systemd's lead and chown only those same files.
-	filesToChown := []string{".", "cgroup.procs", "cgroup.subtree_control", "cgroup.threads"}
 	if c.OwnerUID != nil {
+		// The kernel exposes a list of files that should chowned to
+		// the delegate uid in /sys/kernel/cgroup/delegate.  If the
+		// file is not present (Linux < 4.15), use the initial
+		// values mentioned in in cgroups(7).
+		filesToChown := []string{"."} // the directory itself must be chowned
+		cgroupDelegateFile := "/sys/kernel/cgroup/delegate"
```
kolyshkin, line 300: Ah! This is what I was talking about in https://github.com/opencontainers/runc/pull/3057/commits/56a353b13f93e60752ac52290df13e704db5d5d0#r700644360

kolyshkin: So maybe merge this one into the first commit

kolyshkin (changes requested):
Mostly LGTM; I'd like for the last commit to be merged into the first one for clarity.

libcontainer/specconv/spec_linux.go
frasertweedale: OK, I'll guard this loop behind `if hasCgroupNS`.

libcontainer/cgroups/systemd/v2.go
frasertweedale: Sure, I'm going to squash all commits.  Now that https://github.com/opencontainers/runtime-spec/pull/1123 was finally merged it is preferable to treat this whole PR as a single atomic commit.

tests/integration/cgroups.bats
frasertweedale: Thanks, it's vestigal, before I noticed `set_cgroup_mount_writable` was a thing.  Removing :)

tests/integration/cgroups.bats
frasertweedale: Moving the tests to a separate file, sure.

Extracting the config manipulation to `setup()`, sure.

Unfortunately, these `chown`s can't be moved, because they need to be guarded by `requires root`.  Maybe that could also be moved to `setup()` but that feels wrong and likely to confuse.

frasertweedale: But I did extract those `chown`s to a helper subroutine.

frasertweedale:
@kolyshkin thanks for your review.  I have implemented all of your suggestions.

frasertweedale:
ping @kolyshkin @AkihiroSuda @jfroy are there any further things to address for this PR?  Note that the runtime-spec PR https://github.com/opencontainers/runtime-spec/pull/1123 which proposed these cgroup ownership semantics has been merged.

libcontainer/specconv/spec_linux.go
```diff
@@ -366,6 +366,49 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
...
+	// config.HostRootUID() returns the correct result.
+	//
+	// Only set it if the container will have its own cgroup
+	// namespace and the cgroupfs will be mounted read/write.
+	//
+	hasCgroupNS := config.Namespaces.Contains(configs.NEWCGROUP) && config.Namespaces.PathOf(configs.NEWCGROUP) == ""
+	hasRwCgroupfs := false
+	if hasCgroupNS {
+		for _, m := range config.Mounts {
+			if m.Source == "cgroup" && m.Destination == "/sys/fs/cgroup" && (m.Flags&unix.MS_RDONLY) == 0 {
```
AkihiroSuda, line 381:
```suggestion
			if m.Source == "cgroup" && filepath.Clean(m.Destination) == "/sys/fs/cgroup" && (m.Flags&unix.MS_RDONLY) == 0 {
```

frasertweedale: @AkihiroSuda thank you.  I updated to the PR to implement this change, which you requested.

AkihiroSuda (dismissed):

libcontainer/cgroups/systemd/v2.go
```diff
@@ -288,6 +290,36 @@ func (m *unifiedManager) Apply(pid int) error {
 	if err := fs2.CreateCgroupPath(m.path, m.cgroups); err != nil {
 		return err
 	}
+
+	if c.OwnerUID != nil {
+		// The kernel exposes a list of files that should chowned to
+		// the delegate uid in /sys/kernel/cgroup/delegate.  If the
+		// file is not present (Linux < 4.15), use the initial
+		// values mentioned in in cgroups(7).
```
kolyshkin, line 298: typo: `in in`.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -288,6 +290,36 @@ func (m *unifiedManager) Apply(pid int) error {
 	if err := fs2.CreateCgroupPath(m.path, m.cgroups); err != nil {
 		return err
 	}
+
+	if c.OwnerUID != nil {
+		// The kernel exposes a list of files that should chowned to
```
kolyshkin, line 295: typo: should **be** chowned.

tests/integration/cgroup_delegation.bats
```diff
@@ -0,0 +1,76 @@
...
+function chown_mounts() {
+	# chown test temp dir to allow host user to read it
+	chown 100000 "$ROOT"
+
+	# chown rootfs to allow host user to mkdir mount points
+	chown 100000 "$ROOT"/bundle/rootfs
+}
+
+@test "runc exec (cgroup v2, ro cgroupfs, new cgroupns) does not chown cgroup" {
+	requires root cgroups_v2 systemd
```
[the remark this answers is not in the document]
frasertweedale: done, thanks.

libcontainer/cgroups/systemd/v2.go
frasertweedale: thanks!  fixed in the latest push.

libcontainer/cgroups/systemd/v2.go
frasertweedale: thanks!  fixed in the latest push.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -288,6 +290,36 @@ func (m *unifiedManager) Apply(pid int) error {
...
+				filesToChown = append(filesToChown, scanner.Text())
+			}
+			if err := scanner.Err(); err != nil {
+				return fmt.Errorf("error reading %s: %w", cgroupDelegateFile, err)
+			}
+		} else {
+			filesToChown = append(filesToChown, "cgroup.procs", "cgroup.subtree_control", "cgroup.threads")
+		}
+
+		for _, v := range filesToChown {
```
kolyshkin, line 315: Perhaps it's better to separate out the code that finds files to be chowned to a separate `filesToChown()` function, and just call it from below, e.g.

```go
for _, v := range filesToChown() {
...
}
```

This should increase readability and decrease Apply() complexity.

The only thing is, to simplify, you would not return any errors from filesToChown(), instead maybe logging it as warnings and falling back to built-in list.

kolyshkin (changes requested):
Overall looks good and ready; only left a single suggestion.

libcontainer/cgroups/systemd/v2.go
frasertweedale: I extracted the behaviour to another function.  But I kept the error return.  I feel that `/sys/kernel/cgroup/delegate` existing but being unreadable is something we should fail hard on, because its values may (and currently do) differ from the original list.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -288,9 +290,46 @@ func (m *unifiedManager) Apply(pid int) error {
...
+
 	return nil
 }
 
+// The kernel exposes a list of files that should be chowned to the delegate
+// uid in /sys/kernel/cgroup/delegate.  If the file is not present
+// (Linux < 4.15), use the initial values mentioned in cgroups(7).
+func cgroupFilesToChown() ([]string, error) {
+	filesToChown := []string{"."} // the directory itself must be chowned
+	cgroupDelegateFile := "/sys/kernel/cgroup/delegate"
```
kolyshkin, line 316: Nit: this can be `const`.

frasertweedale: Thanks! Done.

libcontainer/cgroups/systemd/v2.go
kolyshkin: Thanks, this looks better and more reviewable!

kolyshkin (approved):
LGTM

libcontainer/specconv/spec_linux.go
```diff
@@ -366,6 +366,49 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
...
+	// config.HostRootUID() returns the correct result.
+	//
+	// Only set it if the container will have its own cgroup
+	// namespace and the cgroupfs will be mounted read/write.
+	//
+	hasCgroupNS := config.Namespaces.Contains(configs.NEWCGROUP) && config.Namespaces.PathOf(configs.NEWCGROUP) == ""
+	hasRwCgroupfs := false
+	if hasCgroupNS {
+		for _, m := range config.Mounts {
+			if m.Source == "cgroup" && filepath.Clean(m.Destination) == "/sys/fs/cgroup" && (m.Flags&unix.MS_RDONLY) == 0 {
```
kolyshkin, line 381: Ideally, the simplest check (m.Flags is RO) should be the first one, to make things slightly faster, but it's not a big deal, plus the way it's written is more readable.

frasertweedale: Yes, I think we can spend a few CPU cycles to save some brain cycles :)

kolyshkin (approved):
LGTM

frasertweedale:
Thanks to all who have reviewed this PR and given feedback.  We need more approvals and then someone can press the merge button.  Ping @cyphar, @AkihiroSuda, @giuseppe, @jfroy.

frasertweedale:
Any takers?  @mrunalp?

cyphar (approved):
LGTM.

frasertweedale:
Many thanks, @cyphar and all reviewers.

frasertweedale:
I created https://github.com/opencontainers/runc/pull/3311 to backport this feature to the `release-1.0` branch.

PR #2576, opened 2020-09-07
Open bind mount sources from the host userns

alban:
The source of the bind mount might not be accessible in a different usernamespace because a component of the source path might not be traversed under the users and groups mapped inside the user namespace. This caused errors such as the following:
```
  # time="2020-06-22T13:48:26Z" level=error msg="container_linux.go:367:
  starting container process caused: process_linux.go:459:
  container init caused: rootfs_linux.go:58:
  mounting \"/tmp/busyboxtest/source-inaccessible/dir\"
  to rootfs at \"/tmp/inaccessible\" caused:
  stat /tmp/busyboxtest/source-inaccessible/dir: permission denied"
```
To solve this problem, this patch performs the following:

1. in nsexec.c, it opens the source path in the host userns (so we have the right permissions to open it) but in the container mntns (so the kernel cross mntns mount check let us mount it later: https://github.com/torvalds/linux/blob/v5.8/fs/namespace.c#L2312).
2. in nsexec.c, it passes the file descriptors of the source to the child process with SCM_RIGHTS.
3. In runc-init in Golang, it finishes the mounts while inside the userns even without access to the some components of the source paths.

Passing the fds with SCM_RIGHTS is necessary because once the child process is in the container mntns, it is already in the container userns so it cannot temporarily join the host mntns.

This patch uses the existing mechanism with _LIBCONTAINER_* environment variables to pass the file descriptors from runc to runc init.

This patch uses the existing mechanism with the Netlink-style bootstrap to pass information about the list of source mounts to nsexec.c.

-----

Fixes: https://github.com/opencontainers/runc/issues/2484

TODO:
- [x] It does not work yet when the bind mount is configured as read-only in `config.json`.
- [x] Unit tests fail.
- [x] a single env var to pass on all the mount fds

PR #3058, opened 2021-07-02
libcontainer: Don't close already closed fds and bail on close(2) failures

rata:
Hi!

We've found these bugs while rebasing https://github.com/opencontainers/runc/pull/2576. The bug is that we are closing an fd that is already closed, and as we ignore close(2) failures we never realized.

It was not a big deal in runc master, it seems, as that close fails and is otherwise ignored. However, when working on PR #2576 we are opening a new fd, sometimes between the two close(2) calls, the fd number is reused (allocates the lower available fd) and the second close(2) was closing this new open fd that reused the number.

I've create two patches: one to remove the bogus close, and another one to check failures on close(2) so this bug does not happen again (as long as we continue to check for close failures :)). I've also verified that if we check for close failures on the close I remove, it indeed always returns an error on master.

You can find a more in detail explanation in the commit messages :)

cc @alban

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -749,7 +749,8 @@ void nsexec(void)
 				bail("unable to spawn stage-1");
 
 			syncfd = sync_child_pipe[1];
-			close(sync_child_pipe[0]);
+			if (close(sync_child_pipe[0]) < 0)
+				bail("failed to close sync_child_pipe[0] fd");
 
 			/*
 			 * State machine for synchronisation with the children. We only
@@ -839,7 +840,9 @@ void nsexec(void)
 
 			/* Now sync with grandchild. */
 			syncfd = sync_grandchild_pipe[1];
-			close(sync_grandchild_pipe[0]);
+			if (close(sync_grandchild_pipe[0]) < 0)
+				bail("failed to close sync_grandchild_pipe[0] fd");
+
 			write_log(DEBUG, "-> stage-2 synchronisation loop");
 			stage2_complete = false;
 			while (!stage2_complete) {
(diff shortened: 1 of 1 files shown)


libcontainer/nsenter/nsexec.c
```diff
@@ -749,7 +749,9 @@ void nsexec(void)
 				bail("unable to spawn stage-1");
 
 			syncfd = sync_child_pipe[1];
-			close(sync_child_pipe[0]);
+			int r = close(sync_child_pipe[0]);
+			if (r == -1)
+				bail("Failed to close fd sync_child_pipe[0]");
```
cyphar, line 754: Please do

```
if (close(...) < 0)
  bail("failed to close <description> fd"); // note lowercase string
```

for all of these cases.

cyphar (changes requested):

libcontainer/nsenter/nsexec.c
rata: Fixed and pushed, thanks!

rata:
@cyphar fixed, PTAL

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

kolyshkin:
As this is not a bug (without #2576) I think we don't need to backport it to 1.0.

rata:
@kolyshkin agree, it was there for 3 years and doesn't seem to hurt anyone.

It should be safe to backport just the removal of the extra close "libcontainer: Don't close fds already closed" (7d479e6beb7d48a4c6dd9aaaeafd80cf888acfff).

But as you think it is best, backporting it won't fix any visible regressions AFAIK.

Issue #3040, opened 2021-06-25
[RFC] runc exec --cgroup


kolyshkin:
Any container can have sub-cgroups
* created by the container itself, if access is provided (cgroupns is enabled and cgroup mount is not readonly -- see examples for cgroup [v1](https://github.com/opencontainers/runc/blob/51beb5c436b159ae2d483b219c37ecfde13b006a/tests/integration/delete.bats#L53) and [v2](https://github.com/opencontainers/runc/blob/51beb5c436b159ae2d483b219c37ecfde13b006a/tests/integration/delete.bats#L103));
* created by some other tools (higher-level orchestrators, systemd, etc);
* created by runc itself (if a future runtime-spec will add an ability to define sub-cgroups and their resources).

This is a proposal to add a feature to have `runc exec` executed in a sub-cgroup of a container, rather than its top-level cgroup as it happens now.

For example:

```
runc exec -d --cgroup /foo/bar CID cmd args...
```

will run `cmd args` inside container `CID`, putting `cmd` in container's `foo/bar` cgroup, relative to container's top-level cgroup.

Obviously, the default value for `--cgroup` is `/`, which is how it's working now.

In a similar manner, runtime-spec's `Process` structure need to add a `Cgroup` field with the same meaning as runc exec's `--cgroup`. For container init it doesn't make sense to have Cgroup set to any value other than `/`. For other execs, it can be changed.

One other implementation details is, I guess `runc exec --cgroup` should NOT create the sub-cgroup if it does not exist, but return an error.

mrunalp:
The use case for this is dpdk applications that use a subset of the configured cpuset in the main container entirely and they can't tolerate k8s probes (runc exec) to run on those cpus.

AkihiroSuda:
How will this relate to subtree_controllers?

kolyshkin:
> How will this relate to subtree_controllers?

I think it should be orthogonal (except for the "runtime-spec will add an ability to define sub-cgroups" item which is actually not part of this proposal, and only given as an example).

The `runc exec --cgroup` merely uses the existing in-container cgroup, and errors out if it's not available.

giuseppe:
how will this work with cgroup v2?  The processes can be only in the leaf nodes, do we need to ensure the existing processes are moved to a sibling cgroup first?

kolyshkin:
> do we need to ensure the existing processes are moved to a sibling cgroup first?

[...]

PR #3059, opened 2021-07-02
runc exec --cgroup

kolyshkin:
This does some minor cgroup cleanups, and implements https://github.com/opencontainers/runc/issues/3040.

TODO: more tests (thus the draft status).


kolyshkin:
CI went mad:
```
Run mig4/setup-bats@v1
Setting up BATS 1.3.0 ...
Object prototype may only be an Object or null: undefined
Waiting 11 seconds before trying again
Object prototype may only be an Object or null: undefined
Waiting 10 seconds before trying again
Error: Object prototype may only be an Object or null: undefined
```

Restarted.

kolyshkin:
```
# runc exec test_busybox sh -euc ! grep -v ':/$' /proc/self/cgroup (status=1):
# 0::/../../system.slice/runner-provisioner.service
```

This must be a unified path for hybrid cgroup v1+v2, which we do not handle well. Looks like I see what needs to be fixed (not a problem in this PR).

kolyshkin:
> This must be a unified path for hybrid cgroup v1+v2, which we do not handle well. Looks like I see what needs to be fixed (not a problem in this PR).

This is presumably fixed by https://github.com/opencontainers/runc/pull/2087. Let me check...

kolyshkin:
> This is presumably fixed by #2087. Let me check...

Yes, but it requires criu 3.16+, so draft until it's released.

tests/integration/exec.bats
```diff
@@ -167,3 +167,115 @@ function check_exec_debug() {
...
+	[ "$status" -eq 0 ]
+
+	# Check that explicit --cgroup works.
+	runc exec --cgroup memory:submem --cgroup cpu,cpuacct:subcpu test_busybox cat /proc/self/cgroup
+	[ "$status" -eq 0 ]
+	[[ "$output" == *":memory:/submem"* ]]
+	[[ "$output" == *":cpu"*":/subcpu"* ]]
+}
+
+@test "runc exec --cgroup subcgroup [v2]" {
```
AkihiroSuda, line 218: Can we have a test and doc for host cgroupns mode?

kolyshkin: It should be all the same (IOW nothing special about it). As noted in the comment above, the cgroupns is enabled only to simplify the test code (as we don't have to figure out in-container paths to sub-cgroups).

I can certainly rewrite the test to not be dependent on cgroupns. Let me try that.

kolyshkin: Made it working without cgroupns; PTAL @AkihiroSuda.

The major differences are:
1. Instead of looking for `/` in /proc/self/cgroup, we look for `$REL_CGROUPS_PATH`.
2. The rdma controller has to be filtered out of /proc/cgroup/path in two cases, as we do not support for it yet (see #2883). It works as is with cgroupns since if enabled, /proc/<pid>cgroup does not (and can not) distinguish between a process being in host root cgroup and container root cgroup -- showing `/` as the path to the cgroup in both cases.

IOW it works the same as without cgroupns.

kolyshkin:
Rebased; made the cgroup v1 test case work without cgroupns.

kolyshkin:
Looks like criu 3.16 won't be in time for runc 1.1.0 release, so moving the milestone to 1.2.0.

tests/integration/exec.bats
kolyshkin: Now that #2883 is merged, I removed filtering out rdma.

```diff
diff --git a/tests/integration/exec.bats b/tests/integration/exec.bats
index c605aec2..24444d1e 100644
--- a/tests/integration/exec.bats
+++ b/tests/integration/exec.bats
@@ -213,14 +213,12 @@ function check_exec_debug() {
        [[ "$output" == *"unknown controller "* ]]
 
        # Check we can join top-level cgroup (implicit).
-       # XXX: rdma is not yet supported so we filter it out.
-       runc exec test_busybox sh -euc "grep -vF :rdma: /proc/self/cgroup"
+       runc exec test_busybox cat /proc/self/cgroup
        [ "$status" -eq 0 ]
        ! grep -v ":$REL_CGROUPS_PATH\$" <<<"$output"
 
        # Check we can join top-level cgroup (explicit).
-       # XXX: rdma is not yet supported so we filter it out.
-       runc exec --cgroup / test_busybox sh -euc "grep -vF :rdma: /proc/self/cgroup"
+       runc exec --cgroup / test_busybox cat /proc/self/cgroup
        [ "$status" -eq 0 ]
        ! grep -v ":$REL_CGROUPS_PATH\$" <<<"$output"
 
```

Strictly speaking, in this test we have to only check the controllers that we know about, but I'm too lazy to implement it right now (let's wait for a new controller to appear first -- maybe we will add it faster if CI will fail).

kolyshkin:
> Looks like criu 3.16 won't be in time for runc 1.1.0 release, so moving the milestone to 1.2.0.

Or maybe it will -- if it is, I'll move the milestone back to 1.1.0, as this is quite important (not the feature per se, but the hybrid cgroup entering code from #2087).

kolyshkin:
Rebased; this is supposed to fail until criu PPA will have CRIU 3.16 (I hope soon).

h-vetinari:
Should this have a changelog entry?

kolyshkin:
> Should this have a changelog entry?

@h-vetinari added, PTAL

tests/integration/cgroups.bats
```diff
@@ -282,3 +282,25 @@ function setup() {
 	[ "$status" -eq 0 ]
 	[ "$(wc -l <<<"$output")" -eq 1 ]
 }
+
+@test "runc exec (cgroup v2 hybrid joins correct cgroup)" {
+	requires root cgroups_hybrid
+
+	set_cgroups_path
+	set_cgroup_mount_writable
```
kolyshkin, line 290: Not sure why this is needed in here; looks like it's not needed at all.

kolyshkin: removed

cyphar (approved):
LGTM. My only minor concern is that someone will think that `--cgroup` applies to the host cgroup paths, but since it's well documented hopefully that won't happen.

AkihiroSuda (approved):

Issue #3036, opened 2021-06-24
starting container process caused: setup user: unable to find <user>: bufio.Scanner: token too long: unknown.


erict-square:
### Problem

I'm running into the following problem when I run `docker run --network=host --hostname=<hostname> --user <user> -v /etc/group:/etc/group:ro <image>`.

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: setup user: unable to find groups for spec <user>: bufio.Scanner: token too long: unknown.
```

**My current finding shows me that this error is only thrown if a line in `/etc/group` exceeds 65536 characters.**

### Diagnosis 

I'm not familiar with runc code or golang, but i tried my best to find what's going on. I was able to find where this error message is thrown `unable to find groups for spec` in https://github.com/opencontainers/runc/blob/48d76adf7b8d074296edbc8c4475ebede1062116/libcontainer/user/user.go#L359

And from there, I find that the error likely came from running `ParseGroupFilter` function in https://github.com/opencontainers/runc/blob/48d76adf7b8d074296edbc8c4475ebede1062116/libcontainer/user/user.go#L339

Reading `ParseGroupFilter` function, https://github.com/opencontainers/runc/blob/48d76adf7b8d074296edbc8c4475ebede1062116/libcontainer/user/user.go#L178-L211, I suspect that we're hitting a token limit in `bufio.NewScanner` (which happens to be 65536 according to https://golang.org/pkg/bufio/#pkg-constants). This is likely given that another project https://github.com/sirupsen/logrus/issues/564  has hit a similar issue as well.

kolyshkin:
There was a same bug in golang stdlib os/user (https://github.com/golang/go/issues/43636), which was fixed by https://github.com/golang/go/issues/43636.

I was looking at this bug and ended up implementing `GroupIds` for non-cgo/osusergo case in upstream (see https://go-review.googlesource.com/c/go/+/330753). Unfortunately it's a feature freeze for golang now, but once my changes will be merged, my plan is to
 - use stdlib as much as possible;
 - for GroupIds, copy the implementation, conditionally using it, depending on the go version, so that eventually we can get rid of the copy;
 - maybe eventually we can remove our user package entirely.

I will start working on it next week. It's a big job though.

Before that, though, let me try to cook a short fix that we can backport.

erict-square:
@kolyshkin Thanks for the update and taking on this work! I'm really thrilled!

kolyshkin:
[...]

PR #3062, opened 2021-07-02
libct/user: fix parsing long /etc/group lines

kolyshkin:
Lines in /etc/group longer than 64 characters breaks the current
implementation of group parser. This is caused by bufio.Scanner
buffer limit.
    
Fix by re-using the fix for a similar problem in golang os/user,
namely https://go-review.googlesource.com/c/go/+/283601
(fixing a similar bug in golang os/user, https://github.com/golang/go/issues/43636).
    
Add some tests.

Fixes: #3036 
   
Reported-by: @erict-square 
Co-authored-by: Andrey Bokhanko <andreybokhanko@gmail.com>

## Proposed changelog entry
* Fixed `unable to find groups ... token too long` error when /etc/group containing lines longer than 64K characters (#3036)

### DIFF
--- a/libcontainer/user/user.go
+++ b/libcontainer/user/user.go
@@ -187,7 +187,7 @@ func ParseGroupFilter(r io.Reader, filter func(Group) bool) ([]Group, error) {
 	)
 
 	for s.Scan() {
-		text := s.Bytes()
+		text := bytes.TrimSpace(s.Bytes())
 		if len(text) == 0 {
 			continue
 		}
--- a/libcontainer/user/user.go
+++ b/libcontainer/user/user.go
@@ -180,15 +180,53 @@ func ParseGroupFilter(r io.Reader, filter func(Group) bool) ([]Group, error) {
 	if r == nil {
 		return nil, errors.New("nil source for group-formatted data")
 	}
+	rd := bufio.NewReader(r)
+	out := []Group{}
 
-	var (
-		s   = bufio.NewScanner(r)
-		out = []Group{}
-	)
+	// Read the file line-by-line.
+	for {
+		var (
+			isPrefix  bool
+			wholeLine []byte
+			err       error
+		)
 
-	for s.Scan() {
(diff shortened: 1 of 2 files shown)


libcontainer/user/user.go
```diff
@@ -179,15 +180,53 @@ func ParseGroupFilter(r io.Reader, filter func(Group) bool) ([]Group, error) {
...
+		// buffer is able to keep), check if we read enough columns
+		// already on each step and store final result in wholeLine.
+		for {
+			var line []byte
+			line, isPrefix, err = rd.ReadLine()
+
+			if err != nil {
+				// We should return no error if EOF is reached
+				// without a match.
+				if err == io.EOF { //nolint:errorlint // comparison with io.EOF is legit, https://github.com/polyfloyd/go-errorlint/pull/12
```
AkihiroSuda, line 204: Why not use errors.Is ?

kolyshkin: Many functions from https://golang.org/pkg/bufio/ (including `ReadLine` method) legitimately return `io.EOF` error as a way to notify about EOF. Surely it is possible to use `errors.Is` here as well but for cases like this (when we do know we have a moderate to high chances to get this particular error, and it will never be wrapped) this is redundant and a direct comparison is fine.

errorlint has a set of exclusions configured for such direct io.EOF comparisons, and since https://github.com/polyfloyd/go-errorlint/pull/12 this is not even reported by the linter.

golang stdlib itself uses direct comparison with io.EOF quite often, and never uses `errors.Is` with `io.EOF`.

```
[kir@kir-rhat src]$ grep -r 'err == io.EOF' | grep -v ^vendor/ | wc -l
186
[kir@kir-rhat src]$ grep -r 'errors.Is' | grep io.EOF | wc -l
0
```

(this is in Go 1.17beta1 sources)

AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #3067, opened 2021-07-07
Don't freeze cgroup on update for systemd cgroup v2

odinuge:


tests/integration/update.bats
```diff
@@ -584,10 +584,52 @@ EOF
...
 	[[ "$ROOTLESS" -ne 0 ]] && requires rootless_cgroup
-	# This test currently only makes sense on cgroupv1.
-	requires cgroups_v1
+
+	requires root
+
+	# Run a basic shell script that tries to read from to /dev/private,
+	# a dummy device it should not have access to. During "runc update",
+	# it should still not be able to read from the file/device.
+	update_config '.process.args |= ["sh", "-c", "while true; do head -c 10 /dev/private 2>&1 | grep -v \"head: /dev/private: Operation not permitted\"; done"]'
```
odinuge, line 595: TODO: Switch to /dev/kmsg instead. Didn't do that initially due to permission issues; see https://github.com/opencontainers/runc/pull/3069.

kolyshkin:
Indeed, I reinstated the freeze because of https://github.com/opencontainers/runc/issues/3014, which is v1 only bug.

tests/integration/update.bats
```diff
@@ -584,10 +584,52 @@ EOF
...
+	done
+
+	# Kill recvtty.
+	kill -9 "$(<"$TMP_RECVTTY_PID")"
+
+	# There should've been no output from the container.
+	cat "$CONTAINER_OUTPUT"
+	[ -z "$(<"$CONTAINER_OUTPUT")" ]
+}
+@test "update devices does not deny access [minimal transition rules]" {
```
kolyshkin, line 629: The test you're adding very much duplicates the existing one, and copies a lot of code around.

Perhaps it's better to just amend the existing test with the reading from a denied device case, to avoid code duplication.

kolyshkin: Or, if you'd like to keep them separate, rename `@test "update devices [minimal transition rules]" {` to a function, add some variable arguments, and have two test cases call it with different args.

kolyshkin (changes requested):
Second commit looks good; I don't like code duplication in the first one.

tests/integration/update.bats
odinuge: Agree! Squashed them into one now! :+1: Thanks!

kolyshkin (approved):
LGTM

kolyshkin:
Thanks @odinuge for the update.

I treat this as an improvement (rather than a bug fix), and so this is not a candidate for 1.0 backport. Let me know if you disagree.

kolyshkin:

kolyshkin:
OK, I found out this is actually a bug fix -- it fixes the inability to freeze the container/cgroup via cgroup manager's `Set` method for cgroup v2.

While I find using Set for freezing cgroup questionable (as the freezer is kind of special -- and so we have `Freeze` and `GetFreezerState` cgroup manager methods, as well as `Pause` and `Resume` container methods), the functionality is there.

This PR fixes the ability to freeze systemd/v2 cgroup via `Set()`. I am adding a test case for that in #3082.

So, this is a bug, but I still don't think it calls for 1.0 backport.

tests/integration/update.bats
```diff
@@ -586,13 +586,25 @@ EOF
...
+	# "runc update" makes use of minimal transition rules, updates should not cause
+	# writes to fail at any point. For systemd cgroup driver on cgroup v1, the cgroup
+	# is frozen to ensure this.
+	update_config ' .linux.resources.devices = [{"allow": false, "access": "rwm"}, {"allow": false, "type": "c", "major": 1, "minor": 11, "access": "rwa"}]
+			| .linux.devices = [{"path": "/dev/kmsg", "type": "c", "major": 1, "minor": 11}]
+			| .process.capabilities.bounding += ["CAP_SYSLOG"]
+			| .process.capabilities.effective += ["CAP_SYSLOG"]
+			| .process.capabilities.inheritable += ["CAP_SYSLOG"]
+			| .process.capabilities.permitted += ["CAP_SYSLOG"]
+			| .process.args |= ["sh", "-c", "while true; do head -c 100 /dev/kmsg 2> /dev/null; done"]'
```
kolyshkin, line 607: @odinuge I just tried using `head -c 100 /dev/kmsg` and it fails:

```console
[kir@kir-rhat systemd]$ head -2 /dev/kmsg 
6,0,0,-;microcode: microcode updated early to revision 0xea, date = 2021-01-05
5,1,0,-;Linux version 5.12.12-300.fc34.x86_64 (mockbuild@bkernel01.iad2.fedoraproject.org) (gcc (GCC) 11.1.1 20210531 (Red Hat 11.1.1-3), GNU ld version 2.35.1-41.fc34) #1 SMP Fri Jun 18 14:30:51 UTC 2021

[kir@kir-rhat systemd]$ head -c 100 /dev/kmsg 
6,0,0,-;microcode: microcode updated early to revision 0xea, date = 2021-01-05
head: error reading '/dev/kmsg': Invalid argument

[kir@kir-rhat systemd]$ LANG=C strace head -c 100 /dev/kmsg 
execve("/usr/bin/head", ["head", "-c", "100", "/dev/kmsg"], 0x7fff61bd5808 /* 70 vars */) = 0
...
openat(AT_FDCWD, "/dev/kmsg", O_RDONLY) = 3
read(3, "6,0,0,-;microcode: microcode upd"..., 100) = 79
newfstatat(1, "", {st_mode=S_IFCHR|0620, st_rdev=makedev(0x88, 0x1), ...}, AT_EMPTY_PATH) = 0
write(1, "6,0,0,-;microcode: microcode upd"..., 796,0,0,-;microcode: microcode updated early to revision 0xea, date = 2021-01-05
) = 79
read(3, 0x7ffc5a54bfa0, 21)             = -1 EINVAL (Invalid argument)
write(2, "head: ", 6head: )                   = 6
write(2, "error reading '/dev/kmsg'", 25error reading '/dev/kmsg') = 25
write(2, ": Invalid argument", 18: Invalid argument)      = 18
write(2, "\n", 1
...
```

Seems like it happens because the kernel can only write full lines to `/dev/kmsg`.

The fix is easy: use `head -n 1` instead.

kolyshkin: Nah, this is OK as we're not expected to read anything from `/dev/kmsg` (if we will, it's an error).

Nevermind.

kolyshkin: ... also, using `head -c 100` is preferable as it reads only 100 bytes, while `head -n 1` reads 8K (on my system).

mrunalp (approved):

cyphar (approved):
LGTM.

PR #3068, opened 2021-07-07
support changing of lsm mount context on restore

adrianreber:
Wire through CRIU's support to change the mount context on restore.

This is especially useful if restoring a container in a different pod.

Single container restore uses the same SELinux process label and same mount context as during checkpointing. If a container is being restored into an existing pod the process label and the mount context needs to be changed to the context of the pod.

Changing process label on restore is already supported by runc. This patch adds the possibility to change the mount context.

This requires yet unreleased version of CRIU 3.16.


man/runc-restore.8.md
```diff
@@ -75,6 +75,14 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
 **apparamor** or **selinux**, and _label_ is a valid LSM label. For example,
 **--lsm-profile "selinux:system_u:system_r:container_t:s0:c82,c137"**.
 
+**--mount-context** _context_
+: Specify an LSM mount context to be used during restore. Only mounts with an
+existing context will have their context replaced. With this option it is
+possible to change SELinux mount options. Instead of mounting with the
+checkpointed context like `context="system_u:object_r:container_file_t:s0:c31,c779"`
```
kolyshkin, line 82: I don't think the example is needed here.

man/runc-restore.8.md
```diff
@@ -75,6 +75,14 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
 **apparamor** or **selinux**, and _label_ is a valid LSM label. For example,
 **--lsm-profile "selinux:system_u:system_r:container_t:s0:c82,c137"**.
 
+**--mount-context** _context_
```
kolyshkin, line 78: typo: `s/--mount-context/--lsm-mount-context/`

man/runc-restore.8.md
```diff
@@ -75,6 +75,14 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
...
+: Specify an LSM mount context to be used during restore. Only mounts with an
+existing context will have their context replaced. With this option it is
+possible to change SELinux mount options. Instead of mounting with the
+checkpointed context like `context="system_u:object_r:container_file_t:s0:c31,c779"`
+the mount context will be replaced with the value specified using **--mount-context**.
+For example, **--mount-context "system_u:object_r:container_file_t:s0:c82,c137"**.
```
kolyshkin, line 84: typo: `s/--mount-context/--lsm-mount-context/`

libcontainer/container_linux.go
```diff
@@ -1344,6 +1344,12 @@ func (c *linuxContainer) Restore(process *Process, criuOpts *CriuOpts) error {
 		}
 		req.Opts.LsmProfile = proto.String(criuOpts.LsmProfile)
 	}
+	if criuOpts.LsmMountContext != "" {
+		if err := c.checkCriuVersion(31600); err != nil {
+			return errors.New("--mount-context requires at least CRIU 3.16")
```
kolyshkin, line 1349: typo: `s/--mount-context/--lsm-mount-context/`

man/runc-restore.8.md
```diff
@@ -75,6 +75,14 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
 **apparamor** or **selinux**, and _label_ is a valid LSM label. For example,
 **--lsm-profile "selinux:system_u:system_r:container_t:s0:c82,c137"**.
```
kolyshkin, line 76: Not really related, but should we amend the doc to say something like "By default, the checkpointed LSM profile is used upon restore"?

adrianreber:
Made all requested changes and force pushed.

man/runc-restore.8.md
```diff
@@ -74,6 +74,15 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
 : Specify an LSM profile to be used during restore. Here _type_ can either be
 **apparamor** or **selinux**, and _label_ is a valid LSM label. For example,
 **--lsm-profile "selinux:system_u:system_r:container_t:s0:c82,c137"**.
+By default, the checkpointed LSM profile is used upon restore.
+
+**--lsm-mount-context** _context_
+: Specify an LSM mount context to be used during restore. Only mounts with an
+existing context will have their context replaced. With this option it is
+possible to change SELinux mount options. Instead of mounting with the
+checkpointed context the mount context will be replaced with the value specified
```
kolyshkin, line 83: A comma is missing?
```suggestion
checkpointed context, the mount context will be replaced with the value specified
```

man/runc-restore.8.md
```diff
@@ -74,6 +74,15 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
...
+using **--lsm-mount-context**.
```
kolyshkin, line 84: It's weird that the option being described refers to itself.

I think it makes more sense to say something like
```suggestion
checkpointed context, the _context_ will be used for mounting.

**Comment by adrianreber** [comment]
Both man page changes applied.

**Comment by kolyshkin** [comment]
@adrianreber could you please rebase (needed to run latest CI)?

**Review by kolyshkin — DISMISSED** [review]
LGTM

**Comment by kolyshkin** [comment]
@adrianreber needs a rebase 🙏🏻

**Comment by adrianreber** [comment]
> @adrianreber needs a rebase 🙏🏻

rebased

**Review by kolyshkin — APPROVED** [review]
LGTM

**Comment by kolyshkin** [comment]
@thaJeztah PTAL

**Comment by thaJeztah** [comment]
Overall looks good; I'm not a very SELinux person, but I noticed that we also have

```
--process-label label : Set the asm process label for the process commonly used with selinux(7).
```

Which uses _label_ as term. Is it clear that both apply a label?


Just asking; I'm fine with this if this term is clearer to more SELinux-savvy users 😅

**Comment by kolyshkin** [comment]
1. mount(8) man page clearly talks about contexts (although it has no less than 4 options for that), so I guess "context" is the right term here.

2. Option `--process-label` sets SELinux values via /proc using go-selinux pkg. The package uses term "label", the proc(5) (when describing /proc/[pid]/attr files such as attr/exec) uses the terms "attribute", "label", and "[security] context", with "attribute" being the most generic term.

3. From selinux(7) it seems that "security context" and "label" terms are interchangeable. 

Overall, it seems OK as it is (process -- label, mount -- context). We also have `--lsm-profile` which argument is called label.

**Comment by kolyshkin** [comment]
@rhatdan can you please review this from the POV of terminology used?

File: man/runc-restore.8.md [hunk 8]
```diff
@@ -74,6 +74,14 @@ daemon. See [criu --lazy-pages option](https://criu.org/CLI/opt/--lazy-pages).
 : Specify an LSM profile to be used during restore. Here _type_ can either be
 **apparamor** or **selinux**, and _label_ is a valid LSM label. For example,
 **--lsm-profile "selinux:system_u:system_r:container_t:s0:c82,c137"**.
+By default, the checkpointed LSM profile is used upon restore.
+
+**--lsm-mount-context** _context_
```
rhatdan, line 79: I hate typing extra letters, why not just --mount-context?  Or --mount-label, since context is not a common user facing term.

rhatdan: As a matter of fact, I think all through runc, it is referred to as the MountLabel.

kolyshkin: @adrianreber WDYT about `--mount-context` or `--mount-label`?

adrianreber: The only reason not to rename it is to keep the options consistent between criu and runc.

Another reason not to rename it is that Podman already uses this option if it finds support for it in runc.

kolyshkin: OK I'll leave this decision to @rhatdan 🦆

cyphar: Yeah I'm with @kolyshkin. I think it would be nice to use something a bit less cryptic -- the fact that podman already expects this (presumably because crun already supports this?) is a bit of a shame. Maybe we can have a hidden option called `--lsm-mount-context` in order to allow podman to use it but users are pushed towards using the more friendly `--mount-label` (or `--selinux-label` as @thaJeztah suggested)?

kolyshkin: It can't be `--selinux-label` as the format is _type:label_, and the _type_ can currently be apparmor or selinux.

rst0git: > I hate typing extra letters, why not just --mount-context? Or --mount-label, since context is not a common user facing term.

@rhatdan I think this option is not likely to be used very often directly by users from command-line and using an explicit name clarifies the meaning of "context".

adrianreber: > Yeah I'm with @kolyshkin. I think it would be nice to use something a bit less cryptic -- the fact that podman already expects this (presumably because crun already supports this?) is a bit of a shame.

I thought a bit more about the Podman problem and it is not really a problem. Podman currently checks for the existence of `--lsm-mount-context` and tells the user early if runc does not support it. So we can still call it whatever we want in runc and adapt Podman.

Should I change it to `--mount-label`? Would that be a good name for everyone?

cyphar: @kolyshkin Are you sure? The example given in the docs is `--lsm-mount-context "system_u:object_r:container_file_t:s0:c82,c137"`. `--lsm-profile` takes `type:profile`.

However, since we already have `--lsm-profile` maybe `--lsm-mount-context` is more consistent. Since it looks like we've copied the CRIU argument names in the past we should probably just stick with it...

adrianreber:
Rebased.

adrianreber:
So, did we agree on not changing the name of the option? Is this ready to be merged?

man/runc-restore.8.md
kolyshkin: @cyphar right, I mixed it with `--lsm-profile` which does have `apparmor` or `selinux` prefix.

thaJeztah: > Since it looks like we've copied the CRIU argument names in the past we should probably just stick with it...

Fair point. Adding `--lsm-mount-label` to the list of possible names (naming is hard!)

cyphar (approved):
LGTM. Yeah, let's keep it in line with the other `--lsm-*` options.

adrianreber:
I had to rebase again for go.mod conflicts.

kolyshkin (approved):
LGTM

thaJeztah (approved):
LGTM

PR #3069, opened 2021-07-07
tests/int/dev: add CAP_SYSLOG to /dev/kmsg tests

odinuge:
Add CAP_SYSLOG to ensure that /dev/kmsg can be accesses on systems where
the sysctl kernel.dmesg_restrict = 1.


kolyshkin (approved):
LGTM

mrunalp (approved):

PR #3072, opened 2021-07-07
[WIP] libct/cg/sd/v1: Set: don't leave cgroup frozen

kolyshkin:
_This is the alternative to https://github.com/opencontainers/runc/pull/3065 with a different approach and the same test case._

I. In case a parent cgroup is frozen, the current code reads the cgroup
state (into `targetFreezerState`) as `FROZEN`, and sets it explicitly after
setting unit properties. This is obviously wrong, as we should not
explicitly freeze the cgroup because its parent was frozen.
    
The issue happens because:
    
1. The `m.GetFreezerState` method does not distinguish between
     the "self frozen" and "parent frozen" states (i.e. it does not
      consult `freezer.self_freezing`).
    
2. The `m.Freeze` method changes `m.cgroups.Resources.Frozen` field.

II. The current code does freeze/thaw unconditionally, meaning it
        freezes the already frozen cgroup (which is a no-op except it
        still requires some reads and writes to `freezer.state`), and
        thaws the cgroup which is about to be frozen.
    
Solve all this by figuring out whether we need to freeze and thaw,
and introducing and using a method which does not change value of
`m.cgroups.Resources.Frozen`.

Add a test case (initially developed in https://github.com/opencontainers/runc/pull/3065#issuecomment-875217491, and tested to fail without the fix in https://github.com/opencontainers/runc/pull/3066).
    
NOTE the current code assume that the container that is currently frozen
will remain frozen for the duration of `SetUnitProperties`. This might not
be true but there's no way to guarantee it (even in case we freeze the
cgroup ourselves).

### DIFF
--- a/libcontainer/cgroups/systemd/systemd_test.go
+++ b/libcontainer/cgroups/systemd/systemd_test.go
@@ -1,6 +1,7 @@
 package systemd
 
 import (
+	"bufio"
 	"bytes"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


libcontainer/cgroups/systemd/v1.go
```diff
@@ -340,14 +345,21 @@ func (m *legacyManager) Set(r *configs.Resources) error {
...
+	// Figure out if cgroup needs to be frozen before applying systemd properties
+	// and thawed after, while trying to avoid unnecessary state changes.
+	freezerState, err := m.GetFreezerState()
 	if err != nil {
 		return err
 	}
-	if targetFreezerState == configs.Undefined {
-		targetFreezerState = configs.Thawed
+	needsFreeze := true
+	if freezerState == configs.Frozen {
```
odinuge, line 355: This could result in a race between the freezing of the current cgroup and one of its ancestors.

If one ancestor is FROZEN, this will not freeze the control group. If the parent is thowed before systemd does its device update sequence, things will not work as expected.

odinuge: This does also not fix the difference between cgroup v1 and v2.

If the parent control group is frozen, and the control group itself is not, `GetFreezerState` will return `FROZEN` on v1 and `THAWED` on v2. I don't think the naming or definition is important, other than the fact that it works the same way on both.

libcontainer/cgroups/systemd/systemd_test.go
```diff
@@ -193,3 +194,140 @@ func TestUnitExistsIgnored(t *testing.T) {
...
+	if err := cm.Set(containerConfig.Resources); err != nil {
+		t.Fatal(err)
+	}
+	t.Log("container pid added")
+	// Check that we put the "container" into the "pod" cgroup.
+	if !strings.HasPrefix(cm.Path("freezer"), pm.Path("freezer")) {
+		t.Fatalf("expected container cgroup path %q to be under pod cgroup path %q",
+			cm.Path("freezer"), pm.Path("freezer"))
+	}
+
```
odinuge, line 304: Depending on the definition of `GetFreezerState`, we should do a check of its output here as well, to make sure it works the same on cgroup v1 and v2.

```
	cf, err := cm.GetFreezerState()
	if err != nil {
		t.Fatal(err)
	}
	if cf != configs.Thawed {
		t.Fatalf("expected container to be thawed, got %v", cf)
	}
```

libcontainer/cgroups/systemd/v1.go
odinuge: If that make sense to you @kolyshkin and @cyphar?

odinuge: But overall, I agree that always thawing the cgroup is a valid way to do it. If people want to freeze it.

However, doing:

```
runc pause container
runc update container
# Contianer will now still be paused
runc resume container
```

now works as expected, but with your pr it will just resume the container on update (or)?

kolyshkin: > This could result in a race between the freezing of the current cgroup and one of its ancestors.

You are right, and I have anticipated that -- this is noted in the commit and PR description.

libcontainer/cgroups/systemd/systemd_test.go
kolyshkin: Good catch; will add.

kolyshkin:
Closing in favor of #3080 which carries this one.

PR #3073, opened 2021-07-08
runc exec: fail with exit code of 255

kolyshkin:
Noticing that `runc exec` exit status is not documented, I ended up with this.

Currently there's no way to distinguish between the two cases:
 - runc exec failed;
 - the command executed returned 1.

This was possible before commit 8477638aab5a, as runc exec exited with
the code of 255 if exec itself has failed. The code of 255 is the same
convention as used by e.g. ssh.

Re-introduce the feature, document it, and add some tests so it won't be
broken again.

### Changelog entry
```
New functionality:
* runc exec now produces exit code of 255 when the exec failed.
  This may help in distinguishing between runc exec failures
  (such as invalid options, non-running container or non-existent
  binary etc.) and failures of the command being executed. (#3073)

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -101,7 +101,8 @@ following will output a list of processes running in the container:
 		if err == nil {
 			os.Exit(status)
 		}
-		return fmt.Errorf("exec failed: %w", err)
+		fatalWithCode(fmt.Errorf("exec failed: %w", err), 255)
+		return nil // to satisfy the linter
 	},
 	SkipArgReorder: true,
 }
--- a/man/runc-exec.8.md
+++ b/man/runc-exec.8.md
@@ -59,6 +59,11 @@ multiple times.
 : Pass _N_ additional file descriptors to the container (**stdio** +
 **$LISTEN_FDS** + _N_ in total). Default is **0**.
 
+# EXIT STATUS
+
+Exits with a status of _command_ (unless **-d** is used), or **255** if
+an error occurred.
+
 # EXAMPLES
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by kolyshkin** [comment]
CI failure in centos7 is a glitch; restarted.

**Review by fuweid — COMMENTED** [review]
LGTM(nb)

File: exec.go [hunk 1]
```diff
@@ -101,7 +101,8 @@ following will output a list of processes running in the container:
 		if err == nil {
 			os.Exit(status)
 		}
-		return fmt.Errorf("exec failed: %w", err)
+		fatalWithCode(fmt.Errorf("exec failed: %w", err), -1)
```
AkihiroSuda, line 104: > -1

Can't we use `255` here?

AkihiroSuda (approved):

exec.go
kolyshkin: Yes, it will probably look a tad better.

Fixed, PTAL @AkihiroSuda

AkihiroSuda (dismissed):

kolyshkin:
close/reopen to kick new cirrus ci

kolyshkin:
OK, it's not working that way for cirrus-ci apparently, need to rebase.

exec.go
AkihiroSuda: Doesn't seem fixed

kolyshkin: My bad. Let's try again

AkihiroSuda (approved):

mrunalp (approved):

PR #3080, opened 2021-07-09
systemd cgroup v1 freeze fixes

kolyshkin:
_This is a combo of #3065 and #3072, fixing issues resulting in kubernetes revert of runc 1.0.0 vendoring._

1. Fix cgroup v1 `GetFreezerState` to not report cgroup state as frozen if a parent cgroup is frozen.

   After the fix, we only report the cgroup is frozen if it is frozen directly (and not via an ancestor).
   This is how cgroup v2 GetFreezerState works, so now there's no discrepancy between v1 and v2 behavior.

2. Fix cgroup/systemd v1 Set to not leave sub-cgroup frozen in case a parent cgroup is frozen.

   This issue mostly stems from the `GetFreezerState` reporting incorrect state (fixed above).

3. Cleanup cgroup/systemd v1 freeze/thaw code (do not overwrite cgroup resources freezer).

4. Avoid unnecessary freeze/thaw.

5. Tests for all the above issues.

Co-authored-by: @odinuge

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -279,18 +279,23 @@ func getSubsystemPath(c *configs.Cgroup, subsystem string) (string, error) {
 }
 
 func (m *legacyManager) Freeze(state configs.FreezerState) error {
+	err := m.doFreeze(state)
+	if err == nil {
+		m.cgroups.Resources.Freezer = state
+	}
+	return err
+}
+
+// doFreeze is the same as Freeze but without
+// changing the m.cgroups.Resources.Frozen field.
+func (m *legacyManager) doFreeze(state configs.FreezerState) error {
 	path, ok := m.paths["freezer"]
 	if !ok {
 		return errSubsystemDoesNotExist
 	}
-	prevState := m.cgroups.Resources.Freezer
-	m.cgroups.Resources.Freezer = state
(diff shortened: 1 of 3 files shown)


libcontainer/cgroups/systemd/v1.go
```diff
@@ -279,18 +279,23 @@ func getSubsystemPath(c *configs.Cgroup, subsystem string) (string, error) {
 }
 
 func (m *legacyManager) Freeze(state configs.FreezerState) error {
+	err := m.doFreeze(state)
+	if err == nil {
+		m.cgroups.Resources.Freezer = state
+	}
```
kolyshkin, line 285: I don't know if this is necessary, but am inclined to keep this just in case.

kolyshkin:
As suggested in https://github.com/opencontainers/runc/pull/3065#issuecomment-876650769, this can be further improved to avoid the freeze entirely if we're sure systemd won't set the deny-all device rule.

Looking at
https://github.com/systemd/systemd/blob/dd376574fd62c6fcf446ee500ca85558a71d645e/src/core/cgroup.c#L1130-L1133
it seems that freeze can be skipped if `DevicePolicy` is `auto` (which is the default) AND `DeviceAllow` list is empty.

kolyshkin:
Split into #3081 and #3082

PR #3081, opened 2021-07-09
Make cgroup freezer only care about current control group (carry #3065)

kolyshkin:
This is the same as #3065 with a few nitpicks and minor fixes on top.

Fixes the issue (frozen containers) that led to runc bump to 1.0.0 reverted in kubernetes.

Diff from #3065 (not including commit messages):

```diff
diff --git a/libcontainer/cgroups/fs/freezer.go b/libcontainer/cgroups/fs/freezer.go
index 734fc46a..4baa2798 100644
--- a/libcontainer/cgroups/fs/freezer.go
+++ b/libcontainer/cgroups/fs/freezer.go
@@ -129,8 +129,8 @@ func (s *FreezerGroup) GetState(path string) (configs.FreezerState, error) {
                case "THAWED":
                        return configs.Thawed, nil
                case "FROZEN":
-                       // Check freezer state for current control group to check
-                       // if it is frozen, or if it is one of its ancestors.
+                       // Find out whether the cgroup is frozen directly,
+                       // or indirectly via an ancestor.
                        self, err := cgroups.ReadFile(path, "freezer.self_freezing")
                        if err != nil {
                                // If the kernel is too old, then we just treat
@@ -146,7 +146,7 @@ func (s *FreezerGroup) GetState(path string) (configs.FreezerState, error) {
                        case "1\n":
                                return configs.Frozen, nil
                        default:
-                               return configs.Undefined, fmt.Errorf(`unknown "freezer.self_freezing" state: %q`, state)
+                               return configs.Undefined, fmt.Errorf(`unknown "freezer.self_freezing" state: %q`, self)
                        }
                case "FREEZING":
                        // Make sure we get a stable freezer state, so retry if the cgroup
diff --git a/libcontainer/cgroups/systemd/systemd_test.go b/libcontainer/cgroups/systemd/systemd_test.go
index 30957cdf..a444ff61 100644
--- a/libcontainer/cgroups/systemd/systemd_test.go
+++ b/libcontainer/cgroups/systemd/systemd_test.go
@@ -234,7 +234,6 @@ func TestFreezePodCgroup(t *testing.T) {
        if pf != configs.Frozen {
                t.Fatalf("expected pod to be frozen, got %v", pf)
        }
-       t.Log("pod frozen")
 
        // Create a "container" within the "pod" cgroup.
        // This is not a real container, just a process in the cgroup.
@@ -277,7 +276,6 @@ func TestFreezePodCgroup(t *testing.T) {
        if err != nil {
[...]

PR #3082, opened 2021-07-09
cgroups: Set: fix freeze, avoid unnecessary freeze from systemd v1

kolyshkin:
_~~based on (and currently includes) #3081 and #3067. Keeping this a draft before those two are merged.~~_

1. Fix the inability to freeze the container using its `Set()` method (with `r.Freezer` set to `Frozen`). Add a test.
2. Avoid unnecessary freeze/thaw from system v1 driver. Add a test.

Review commit-by-commit. Please see individual commits for details.

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -279,18 +279,23 @@ func getSubsystemPath(c *configs.Cgroup, subsystem string) (string, error) {
 }
 
 func (m *legacyManager) Freeze(state configs.FreezerState) error {
+	err := m.doFreeze(state)
+	if err == nil {
+		m.cgroups.Resources.Freezer = state
+	}
+	return err
+}
+
+// doFreeze is the same as Freeze but without
+// changing the m.cgroups.Resources.Frozen field.
+func (m *legacyManager) doFreeze(state configs.FreezerState) error {
 	path, ok := m.paths["freezer"]
 	if !ok {
 		return errSubsystemDoesNotExist
 	}
-	prevState := m.cgroups.Resources.Freezer
-	m.cgroups.Resources.Freezer = state
 	freezer := &fs.FreezerGroup{}
-	if err := freezer.Set(path, m.cgroups.Resources); err != nil {
-		m.cgroups.Resources.Freezer = prevState
-		return err
-	}
-	return nil
+	resources := &configs.Resources{Freezer: state}
+	return freezer.Set(path, resources)
 }
 
 func (m *legacyManager) GetPids() ([]int, error) {
@@ -356,18 +361,16 @@ func (m *legacyManager) Set(r *configs.Resources) error {
 	// (unlike our fs driver, they will happily write deny-all rules to running
 	// containers). So we freeze the container to avoid them hitting the cgroup
 	// error. But if the freezer cgroup isn't supported, we just warn about it.
-	if err := m.Freeze(configs.Frozen); err != nil {
+	if err := m.doFreeze(configs.Frozen); err != nil {
 		logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
 	}
 
-	if err := setUnitProperties(m.dbus, getUnitName(m.cgroups), properties...); err != nil {
-		_ = m.Freeze(targetFreezerState)
-		return err
-	}
+	setErr := setUnitProperties(m.dbus, getUnitName(m.cgroups), properties...)
 
-	// Reset freezer state before we apply the configuration, to avoid clashing
-	// with the freezer setting in the configuration.
-	_ = m.Freeze(targetFreezerState)
+	_ = m.doFreeze(targetFreezerState)
+	if setErr != nil {
+		return setErr
+	}
 
 	for _, sys := range legacySubsystems {
 		// Get the subsystem path, but don't error out for not found cgroups.
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/godbus/dbus/v5"
 	"github.com/sirupsen/logrus"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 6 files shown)


kolyshkin:
As suggested in https://github.com/opencontainers/runc/pull/3065#issuecomment-876650769, this can be further improved to avoid the freeze entirely if we're sure systemd won't set the deny-all device rule.

Looking at
https://github.com/systemd/systemd/blob/dd376574fd62c6fcf446ee500ca85558a71d645e/src/core/cgroup.c#L1130-L1133
it seems that freeze can be skipped if `DevicePolicy` is `auto` (which is the default) AND `DeviceAllow` list is empty.

Problem is, because of `SkipDevices` use in update.go, we have to query the current/existing properties first, and only skip freezing if all of the following is true:
1. Current properties have `DevicePolicy=auto`
2. Current properties have empty `DeviceAllow` list
3. `SkipDevices` is true.

odinuge:
Together with https://github.com/opencontainers/runc/pull/3081#pullrequestreview-703009467, I think this PR is pretty neat and good.

@kolyshkin in https://github.com/opencontainers/runc/pull/3065#issuecomment-876650769
> So, one way to fix this would be to have a code that tells whether systemd is going to do deny-all on SetUnitProperties, and skip the freeze it it won't.

If we want to do that for container as well, yeah. Although I don't know how often containers are allowed access to all devices?

For the k8s usage for managing a control group, a simple flag is ok though, but that is up to you and the other runc maintainers.

> This is a kludge to a kludge and it's dependent upon systemd internals (which may change) but in these circumstances this may be the best way to proceed.

Yup, it becomes quite a mess... :upside_down_face: 

The best fix is probably to just start using cgroup v2 instead. `crun` has opted to just skip dealing with systemd mostly, and use a delegated scope for cgroup v2, and that works kinda ok I guess, but its not perfect. Not sure how they deal with cgroup v1 tho.

kolyshkin:
> If we want to do that for container as well, yeah. Although I don't know how often containers are allowed access to all devices?

Container run via `docker run --privileged` allows access to all devices. I guess `podman` and some other runtimes have the same or similar thing.

> For the k8s usage for managing a control group, a simple flag is ok though, but that is up to you and the other runc maintainers.

Problem is, from the libcontainer/cgroup we do not know if it's a container or a pod cgroup. Even if it's a pod, it can still have some device access rules (kubernetes doesn't currently set any, but this may change), so having something like `Pod: true` or `SkipFreeze: true` is a no go.

The only thing why I don't like the current solution (as implemented by `libct/cg/sd/v1: Set: avoid unnecessary freeze/thaw` commit in this PR) is it does two calls to dbus to find out whether to skip the freeze, but as we don't have/know the current state, this is necessary.

cyphar:
@kolyshkin 

> Looking at
https://github.com/systemd/systemd/blob/dd376574fd62c6fcf446ee500ca85558a71d645e/src/core/cgroup.c#L1130-L1133
it seems that freeze can be skipped if DevicePolicy is auto (which is the default) AND DeviceAllow list is empty.

In that case the device rules would be allow-all, wouldn't they?

```c
                if (c->device_allow || policy != CGROUP_DEVICE_POLICY_AUTO)
                        r = cg_set_attribute("devices", path, "devices.deny", "a");
                else
                        r = cg_set_attribute("devices", path, "devices.allow", "a");
```

If you have no device_allow list, and the device policy is auto then `devices.allow` gets set to `a` which will mean that a container could access devices it shouldn't during the race window (which is worse than the `-EACCES` inconvenience).

kolyshkin:
> EDIT: Ah, your patch only does the skip if we have SkipDevices and the slice has been configured to be allow-all. Is that a common scenario?

@cyphar Yes, at least for kubernetes that uses libcontainter/cgroup to configure pod cgroups. As pod cgroup is a parent for a few containers cgroups, we do unnecessarily freeze all those containers on update, which is definitely not the way to do.

Those *pod* tests that we're adding recently are trying to emulate those kubernetes usage scenarios.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -326,6 +332,57 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
...
+// applying its systemd unit properties, and thaw after, while avoiding
+// unnecessary freezer state changes.
+//
+// The reason why we have to freeze is that systemd's application of device
+// rules is done disruptively, resulting in spurious errors to common devices
+// (unlike our fs driver, they will happily write deny-all rules to running
+// containers). So we have to freeze the container to avoid the container get
+// an occasional "permission denied" error.
+func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, Err error) {
+	// Special case for SkipDevices.
```
cyphar, line 345: I would mention in this comment that this is also a special case where the systemd cgroup has been configured with an allow-all device policy (and that this situation is pretty much Kubernetes-specific), to make sure we don't actually expand this behaviour in the future.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -326,6 +332,57 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
...
+// freezeBeforeSet answers whether there is a need to freeze the cgroup before
+// applying its systemd unit properties, and thaw after, while avoiding
+// unnecessary freezer state changes.
+//
+// The reason why we have to freeze is that systemd's application of device
+// rules is done disruptively, resulting in spurious errors to common devices
+// (unlike our fs driver, they will happily write deny-all rules to running
+// containers). So we have to freeze the container to avoid the container get
+// an occasional "permission denied" error.
+func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, Err error) {
```
cyphar, line 344: I think it would be nicer to call the returned error `Err` and have a special name for the error we intentionally skip returning in `SkipDevices` rather than the other way around.

kolyshkin: Yes, it is intentional -- I have decided to play it safe and not treat an error while getting a property value as fatal for Set, but rather just fall back to do the freeze. Not sure if that makes sense or not, just did not wanted to introduce another way for Set to fail.

kolyshkin:
Addressed @cyphar review comments, here are the changes I've made:
```diff
diff --git a/libcontainer/cgroups/systemd/v1.go b/libcontainer/cgroups/systemd/v1.go
index 2a567bb8..1a8e1e3c 100644
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -341,7 +341,7 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
 // (unlike our fs driver, they will happily write deny-all rules to running
 // containers). So we have to freeze the container to avoid the container get
 // an occasional "permission denied" error.
-func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, Err error) {
+func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, err error) {
        // Special case for SkipDevices, as used by Kubernetes to create pod
        // cgroups with allow-all device policy).
        if r.SkipDevices {
@@ -352,10 +352,13 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
                // Interestingly, (1) and (2) are the same here because
                // a non-existent unit returns default properties,
                // and settings in (2) are the defaults.
-               devPolicy, err := getUnitProperty(m.dbus, unitName, "DevicePolicy")
-               if err == nil && devPolicy.Value == dbus.MakeVariant("auto") {
-                       devAllow, err := getUnitProperty(m.dbus, unitName, "DeviceAllow")
-                       if err == nil && devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) {
+               //
+               // Do not return errors from getUnitProperty, as they alone
+               // should not prevent Set from working.
+               devPolicy, e := getUnitProperty(m.dbus, unitName, "DevicePolicy")
+               if e == nil && devPolicy.Value == dbus.MakeVariant("auto") {
+                       devAllow, e := getUnitProperty(m.dbus, unitName, "DeviceAllow")
+                       if e == nil && devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) {
                                needsFreeze = false
                                needsThaw = false
                                return
@@ -367,8 +370,8 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
        needsThaw = true
 
        // Check the current freezer state.
-       freezerState, Err := m.GetFreezerState()
-       if Err != nil {
+       freezerState, err := m.GetFreezerState()
+       if err != nil {
                return
        }
        if freezerState == configs.Frozen {
```

libcontainer/cgroups/systemd/v1.go
kolyshkin: fixed `Err -> err`, added a comment about not returning an error from `getUnitProperty`.

libcontainer/cgroups/systemd/v1.go
kolyshkin: amended the comment.

kolyshkin:
CI went south :(
>  Could not connect to azure.archive.ubuntu.com:80 (52.252.75.106), connection timed out

cyphar (approved):
LGTM.

mrunalp (approved):

PR #3085, opened 2021-07-12
[1.0] Make cgroup freezer only care about current control group

kolyshkin:
_This is a backport of PR #3081 to 1.0._

Fixes the issue (frozen containers) that led to runc bump to 1.0.0 reverted in kubernetes.

See also: #3065.

## Changelog entry
```
Bugfixes:
* cgroupv1/systemd: fix freezing cgroup after Set if a parent cgroup is frozen (#3085)
```


AkihiroSuda (approved):

cyphar:
Ah, we need to add the 2-LGTM require rule to `release-1.0` since it's not a protected branch...

cyphar (approved):
LGTM.

PR #3087, opened 2021-07-14
[1.0] cgroupv2: ebpf: ignore inaccessible existing programs

kolyshkin:
_This is a backport of PR #3055 to 1.0 branch. ~~Draft until that one is merged.~~ Original description follows._

This is necessary in order for runc to be able to configure device
cgroups with --systemd-cgroup on distributions that have very strict
SELinux policies such as [openSUSE MicroOS][1].

The core issue here is that systemd is adding its own BPF policy that
has an SELinux label such that runc cannot interact with it. In order to
work around this, we can just ignore the policy -- in theory this
behaviour is not correct but given that the most obvious case
(--systemd-cgroup) will still handle updates correctly, this logic is
reasonable.

[1]: https://bugzilla.suse.com/show_bug.cgi?id=1182428

<hr>

### Changelog Entry ###
```
 * cgroupv2: bpf: Ignore inaccessible existing programs in case of
   permission error when handling replacement of existing bpf cgroup
   programs. This fixes a regression in 1.0.0, where some SELinux
   policies would block runc from being able to run entirely. #3055
```


AkihiroSuda (approved):

cyphar (approved):
LGTM.

cyphar (changes requested):
Ah wait, this PR was made against the wrong branch -- should be against `release-1.0`.

cyphar:
(Good thing I do merges manually -- `git mm` screamed at me when I tried to merge this.)

kolyshkin:
> Ah wait, this PR was made against the wrong branch -- should be against release-1.0.

My bad 🤦🏻

Fixed.

mrunalp (approved):

cyphar (approved):
LGTM.

Issue #3084, opened 2021-07-09
Reducing CPU period fails for subsystems if existing parent has quota>0 with systemd driver


hk-vmg:
When setting a lower CFS CPU period, creating a runc container in Kubernetes using systemd driver fails with:
`write /sys/fs/cgroup/cpu,cpuacct/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod<HASH>.slice/cri-containerd-<HASH>.scope/cpu.cfs_period_us: invalid argument: unknown`

cgroupfs driver works fine.

It appears that order matters, when both period and quota files are changed. I believe the root cause is that the one-at-a-time updates to the child files cause the parent limits to be exceeded. This is seen in Kubernetes - if all containers in a pod have limits, then the parent "pod" slice will set a quota and the child "container" settings can't be updated.

https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/fs/cpu.go#L74

Tested with latest `master` branch runc. 

systemd is version 219 but I believe I have all the required OS/systemd patches:
```
$ rpm -q systemd --changelog | grep -i cpu
core: add CPUQuotaPeriodSec= (#1770379)
core: downgrade CPUQuotaPeriodSec= clamping logs to debug (#1770379) 
```

@sureshvis

hk-vmg:
@sureshvis

kolyshkin:
1. The error actually comes from the kernel.
2. systemd driver sets the property/limit for systemd unit (those that are possible to represent for systemd, see https://github.com/opencontainers/runc/blob/master/docs/systemd.md#resource-limits), and then uses fs driver as a backup.
3. The translation of OCI runtime spec properties to systemd properties for CPU quota and period happens here: https://github.com/opencontainers/runc/blob/fceadf23869f5936db3a90eb62178cc03dd3fad7/libcontainer/cgroups/systemd/common.go#L445-L477

Note that systemd >= 242 is required to set systemd `CPUQuotaPeriod`, meaning that in your case runc only sets the quota and do not set the period for systemd. This is probably why setting the period fails later.

I see a few ways to fix this:

1. For systemd, if the period is not default, either set the quota and period together, or do not set at all (it will be set anyway by the fs driver). This should be an easy fix.
2. Instead of relying on systemd version to figure out whether to set `CPUQuotaPeriod`, query the current value -- if parameter exists, we can set it. This will slow things down a bit.
[...]

PR #3090, opened 2021-07-14
libct/cg/v1: work around CPU quota period set failure

kolyshkin:
_This is based on (and currently includes) #3175._

I. As reported in issue #3084, sometimes setting CPU quota period fails
when a new period is lower and a parent cgroup has CPU quota limit set.

This happens as in cgroup v1 the quota and the period can not be set
together (this is fixed in v2), and since the period is being set first,
`new_limit = old_quota/new_period` may be higher than the parent
cgroup limit.

The fix is to retry setting the period after the quota, to cover all
possible scenarios.

Add a test case to cover a regression caused by an earlier version of
this patch (ignoring a failure of setting invalid period when quota is
not set).

Fixes: #3084
Closes: #2044

II. Add a test case for an issue fixed by the previous commit.
    
Unfortunately, this is somewhat complicated as there's no easy way to
create a transient unit, so a binary, sd-helper, had to be added. On top
of that, an ability to create a parent/pod cgroup is added to
helpers.bash, which might be useful for future integration tests.

#### History
* v2: rearranged, cleanups split to #3175, simplified sd-helper to not depend on urlfave/cli.

### DIFF
--- a/contrib/cmd/sd-helper/helper.go
+++ b/contrib/cmd/sd-helper/helper.go
@@ -1,0 +1,83 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/systemd"
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+func usage() {
+	fmt.Print(`Open Container Initiative contrib/cmd/sd-helper
+
+sd-helper is a tool that uses runc/libcontainer/cgroups/systemd package
+functionality to communicate to systemd in order to perform various operations.
+Currently this is limited to starting and stopping systemd transient slice
+units.
+
+Usage:
+	sd-helper [-debug] [-parent <pname>] {start|stop} <name>
+
+Example:
+	sd-helper -parent system.slice start system-pod123.slice
+`)
+	os.Exit(1)
+}
+
+var (
+	debug  = flag.Bool("debug", false, "enable debug output")
+	parent = flag.String("parent", "", "parent unit name")
+)
+
+func main() {
+	if !systemd.IsRunningSystemd() {
+		logrus.Fatal("systemd is required")
+	}
+
+	// Set the flags.
+	flag.Parse()
+	if *debug {
+		logrus.SetLevel(logrus.DebugLevel)
+	}
+	if flag.NArg() != 2 {
+		usage()
+	}
+
+	cmd := flag.Arg(0)
+	unit := flag.Arg(1)
+
+	err := unitCommand(cmd, unit, *parent)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+}
+
+func newManager(config *configs.Cgroup) cgroups.Manager {
+	if cgroups.IsCgroup2UnifiedMode() {
+		return systemd.NewUnifiedManager(config, "", false)
+	}
+	return systemd.NewLegacyManager(config, nil)
+}
+
+func unitCommand(cmd, name, parent string) error {
+	podConfig := &configs.Cgroup{
+		Name:      name,
+		Parent:    parent,
+		Resources: &configs.Resources{},
+	}
+	pm := newManager(podConfig)
+
+	switch cmd {
+	case "start":
+		return pm.Apply(-1)
+	case "stop":
+		return pm.Destroy()
+	}
+
+	return fmt.Errorf("unknown command: %s", cmd)
+}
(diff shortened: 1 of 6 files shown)


kolyshkin:
A repro (which took me a while to write) fails as expected! 

> time="2021-07-15T01:24:25Z" level=error msg="failed to write \"10000\": write /sys/fs/cgroup/cpu,cpuacct/runc-cgroups-integration-test/pod_31883/test-cgroup-14305/cpu.cfs_period_us: invalid argument"
152

Adding the fix now.

kolyshkin:
Don't know how to create a transient scope from bash :( and this is required for centos 7's systemd.

kolyshkin:
> Don't know how to create a transient scope from bash :( and this is required for centos 7's systemd.

In theory, I should be able to use `dbus-send` to call `org.freedesktop.systemd1.Manager.StartTransientUnit` but I can't figure out a way to.

Another way is to use libcontainer and have a small go binary to do that.

kolyshkin:
> Another way is to use libcontainer and have a small go binary to do that.

Trying it as I can't think of any better way right now.

kolyshkin:
Might make sense to backport to 1.0 (maybe only the fix itself, as it's short and clear, unlike the test case).

libcontainer/cgroups/fs/cpu.go
```diff
@@ -71,15 +73,32 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
...
+		period = strconv.FormatUint(r.CpuPeriod, 10)
+		if err := cgroups.WriteFile(path, "cpu.cfs_period_us", period); err != nil {
+			// Sometimes when the period to be set is smaller
+			// than the current one, it is rejected by the kernel
+			// (EINVAL) as old_quota/new_period exceeds the parent
+			// cgroup quota limit. If this happens, ignore the
+			// error for now and retry after setting the quota.
+			if !errors.Is(err, unix.EINVAL) {
+				return err
+			}
```
hk-vmg, line 88: If Quota is not being set (r.CpuQuota == 0), this will silently fail Period change as well. Any idea whether it's possible to get EINVAL in cases other than the original bug report?

kolyshkin: Good catch. Funny that I added quite a few corner cases for cpu quota (e.g. in #2458) and it's still not enough :)

kolyshkin: Fixed, and added yet another corner test case ("setting invalid period without quota should fail").

kolyshkin:
Ughm; the backport (#3115) is already merged, and this one has no LGTMs :-\ @opencontainers/runc-maintainers PTAL 🙏🏻

kolyshkin:

contrib/cmd/sd-helper/helper.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+// version will be populated by the Makefile, read from
+// VERSION file of the source code.
+var version = ""
+
+// gitCommit will be the hash that the binary was built from
+// and will be populated by the Makefile.
+var gitCommit = ""
```
thaJeztah, line 22: Will this utility only be used for tests? Wondering if adding version information is a bit too much for that (or even using `urfave/cli` for this).

Reason I'm looking at the `version` variable, is also that if we want to switch to `//go:embed` (https://github.com/opencontainers/runc/pull/3163), the path to files to embed cannot traverse to parent directories (so `//go:embed ../../VERSION` won't work) 😅

thaJeztah (commented):
Didn't do a full review yet, but left a comment. 

Also looks like the last commit is somewhat unrelated (could make sense separately)

kolyshkin:
> Also looks like the last commit is somewhat unrelated (could make sense separately)

That's right, but it's harder to work on it if it's separate (since other commits in this PR also change the same file), and I don't want to forget about it.

Let me try to rework it so the shellcheck fixes are separate.

contrib/cmd/sd-helper/helper.go
kolyshkin: I just copied this bit (as well as using `urlfave/cli`) from `recvtty`.

Let me try to drop both.

kolyshkin: Fixed, PTAL @thaJeztah

kolyshkin:
> Let me try to rework it so the shellcheck fixes are separate.

Those fixes separated out to #3175 but this one depends on (and currently includes) it.

We can either review/merge #3175 first, and then I'll rebase, or merge this one as is and close #3175.

kolyshkin:
This fixes a real bug, and the backport to 1.0 was already merged and released in 1.0.2. @opencontainers/runc-maintainers PTAL

AkihiroSuda (approved):

kolyshkin:
For the reference, the same fix for crun: https://github.com/containers/crun/pull/1188

PR #3103, opened 2021-07-21
README.md: remove abandoned versioning policy

AkihiroSuda:
"`runc` X.Y.Z should implement the X.Y version of the specification." is no longer correct.

runc v1.1 will be probably released ahead of OCI Runtime Spec v1.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,10 +12,6 @@
 
 ## Releases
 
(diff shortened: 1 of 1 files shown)


cyphar (approved):
LGTM, though maybe we should add an explanation of our new versioning policy here somewhere (after we've done a few releases and have gotten the hang of things).

kolyshkin (approved):
LGTM

Issue #2924, opened 2021-04-28
[fedora] not ok 16 checkpoint --lazy-pages and restore


kolyshkin:
This is from a recent CI run (https://github.com/opencontainers/runc/pull/2923/checks?check_run_id=2452887046)

```
not ok 16 checkpoint --lazy-pages and restore
# (from function `fail' in file tests/integration/helpers.bash, line 274,
#  from function `runc_restore_with_pipes' in file tests/integration/checkpoint.bats, line 94,
#  in test file tests/integration/checkpoint.bats, line 225)
#   `runc_restore_with_pipes ./image-dir test_busybox_restore --lazy-pages' failed
# runc spec (status=0):
# 
# uffd-noncoop is supported
# runc state test_busybox (status=0):
# {
#   "ociVersion": "1.0.2-dev",
#   "id": "test_busybox",
#   "pid": 27036,
#   "status": "running",
#   "bundle": "/tmp/bats-run-24758/runc.cXN8Qe/bundle",
#   "rootfs": "/tmp/bats-run-24758/runc.cXN8Qe/bundle/rootfs",
#   "created": "2021-04-27T23:34:48.770811211Z",
#   "owner": ""
# }
# __runc restore test_busybox_restore failed (status: 1)
# time="2021-04-27T23:34:49Z" level=error msg="criu failed: type NOTIFY errno 0\nlog file: image-dir/restore.log"
# CRIU log errors (if any):
# ./image-dir/restore.log-(00.079105) 27107 (native) is going to execute the syscall 11, required is 15
# ./image-dir/restore.log-(00.079116) 27107 was trapped
# ./image-dir/restore.log-(00.079118) `- Expecting exit
# ./image-dir/restore.log-(00.079126) 27107 was trapped
# ./image-dir/restore.log-(00.079129) 27107 (native) is going to execute the syscall 18446744073709551615, required is 15
# ./image-dir/restore.log:(00.079136) Error (compel/src/lib/infect.c:1513): Task 27107 is in unexpected state: f7f
# ./image-dir/restore.log:(00.079742) Error (compel/src/lib/infect.c:1520): Task stopped with 15: Terminated
# ./image-dir/restore.log:(00.079748) Error (criu/cr-restore.c:2416): Can't stop all tasks on rt_sigreturn
# ./image-dir/restore.log:(00.079750) Error (criu/cr-restore.c:2454): Killing processes because of failure on restore.
[...]

PR #3104, opened 2021-07-21
Do not use Vagrant for CentOS 7/8

adrianreber:
As Cirrus CI does not provide a real terminal this uses the same 'ssh -tt' workaround as the Vagrant setup. This sets up the CentOS 7 and 8 to allow SSH as root to localhost so that we can run all the tests via 'ssh -tt'.

Not going through vagrant reduces CI times for CentOS 7 and 8 from 6 minutes to 4 minutes.


.cirrus.yml
```diff
@@ -65,3 +66,130 @@ vagrant_task:
...
+  integration_systemd_rootless_script: |
+    echo "SKIP: integration_systemd_rootless_script requires cgroup v2"
+  integration_fs_rootless_script: |
+    echo "SKIP: FIXME: integration_fs_rootless_script is skipped because of EPERM on writing cgroup.procs"
+
+task:
+  name: ci / centos 8
+  compute_engine_instance:
+    image_project: centos-cloud
+    image: family/centos-8
```
AkihiroSuda, line 139: I'm not sure we want CentOS 8 CI.
CentOS 8 is going to die soon.

.cirrus.yml
```diff
@@ -65,3 +66,130 @@ vagrant_task:
...
+    memory: 8
+
+  timeout_in: 30m
+  env:
+    HOME: /root
+    CIRRUS_WORKING_DIR: /home/runc
+    GO_VERSION: "1.16.4"
+    BATS_VERSION: "v1.3.0"
+
+  install_dependencies_script: |
```
AkihiroSuda, line 86: Can we now remove `Vagrantfile.centos7`?

.cirrus.yml
adrianreber: replaced centos-8 with centos-stream-8. I can remove the complete centos 8 section if unnecessary. Just thought I add it.

.cirrus.yml
adrianreber: removed

.cirrus.yml
```diff
@@ -65,3 +66,130 @@ vagrant_task:
...
+  unit_tests_script: |
+    ssh -tt localhost "make -C /home/runc localunittest"
+  integration_systemd_script: |
+    ssh -tt localhost "make -C /home/runc localintegration RUNC_USE_SYSTEMD=yes"
+  integration_fs_script: |
+    ssh -tt localhost "make -C /home/runc localintegration"
+  integration_systemd_rootless_script: |
+    echo "SKIP: integration_systemd_rootless_script requires cgroup v2"
+  integration_fs_rootless_script: |
+    echo "SKIP: FIXME: integration_fs_rootless_script is skipped because of EPERM on writing cgroup.procs"
```
AkihiroSuda, line 195: I guess this should not be skipped for CentOS 8.

Also, can we use `matrix` to reduce code clone?

adrianreber: > I guess this should not be skipped for CentOS Stream 8.

I tried to make it work for a couple of hours but it always fails with:
```
not ok 117 update rt period and runtime
# (in test file tests/integration/update.bats, line 548)
#   `echo "$root_period" >"$target_period"' failed
# 
# Writing 1000000 to /sys/fs/cgroup/cpu,cpuacct//cpu.rt_period_us
# /tmp/bats-run-106184/bats.115768.src: line 548: /sys/fs/cgroup/cpu,cpuacct//cpu.rt_period_us: Permission denied
```

> Also, can we use `matrix` to reduce code clone?

Yes. I thought about it, but was unsure what would be better. I will update the PR.

.cirrus.yml
kolyshkin: I think it makes sense to have it (and I always wanted to add it).

CentOS stream 8 makes sense 👍🏻

.cirrus.yml
```diff
@@ -65,3 +66,79 @@ vagrant_task:
     else
       ssh -tt default "sudo -i make -C /vagrant localrootlessintegration"
     fi
+
+task:
+  timeout_in: 30m
+  env:
+    HOME: /root
+    CIRRUS_WORKING_DIR: /home/runc
+    GO_VERSION: "1.16.4"
```
kolyshkin, line 75: Can switch to 1.16.6 while at it.

.cirrus.yml
```diff
@@ -65,3 +66,79 @@ vagrant_task:
...
+    git clone https://github.com/bats-core/bats-core
+    cd bats-core
+    git checkout $BATS_VERSION
+    ./install.sh /usr/local
+    cd -
+    if [ $DISTRO == centos-7 ]; then
+      # sysctl
+      echo "user.max_user_namespaces=15076" > /etc/sysctl.d/userns.conf
+      sysctl --system
+    fi
```
kolyshkin, line 112: Move it to under the previous if / switch

.cirrus.yml
kolyshkin: > /tmp/bats-run-106184/bats.115768.src: line 548: /sys/fs/cgroup/cpu,cpuacct//cpu.rt_period_us: Permission denied

Mea culpa. Fix is in https://github.com/opencontainers/runc/pull/3106 -- @adrianreber you can cherry-pick it to this PR.

.cirrus.yml
```diff
@@ -65,3 +66,79 @@ vagrant_task:
...
+      DISTRO: centos-7
+      DISTRO: centos-stream-8
+
+  name: ci / $DISTRO
+  compute_engine_instance:
+    image_project: centos-cloud
+    image: family/$DISTRO
+    platform: linux
+    cpu: 4
+    memory: 8
```
kolyshkin, line 88: Can you use a suffix here? I.e. `8G`.

.cirrus.yml
```diff
@@ -65,3 +66,79 @@ vagrant_task:
...
+
+  install_dependencies_script: |
+    if [ $DISTRO == centos-7 ]; then
```
kolyshkin, line 91:
```bash
case $DISTRO in
centos-7)
    .....
    ;;
centos-stream-8)
    ....
    ;;
esac
```

.cirrus.yml
```diff
@@ -65,3 +66,79 @@ vagrant_task:
...
+      yum install -y -q epel-release
+      (cd /etc/yum.repos.d && curl -O https://copr.fedorainfracloud.org/coprs/adrian/criu-el7/repo/epel-7/adrian-criu-el7-epel-7.repo)
+    elif [ $DISTRO == centos-stream-8 ]; then
+      yum install -y -q https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm dnf-plugins-core
```
kolyshkin, line 95: It seems that epel-release is available in centos-stream-8 repo (in extras): http://mirror.centos.org/centos/8-stream/extras/x86_64/os/Packages/

So maybe
```bash
yum --enablerepo extras install -y -q epel-release dnf-plugins-core
```

kolyshkin:
CI failure in Fedora 34 ("not ok 18 checkpoint --lazy-pages and restore") may be a flake I am fighting for quite some time (see https://github.com/opencontainers/runc/issues/2924). It is rare so for now the workaround is to restart CI. It might have been fixed with criu 1.16 (https://github.com/checkpoint-restore/criu/issues/1524) but more changes are needed from runc side (I have a WIP PR, https://github.com/opencontainers/runc/pull/2761, that needs a lot of cleaning), and I have yet to re-check that.

.cirrus.yml
```diff
@@ -1,29 +1,30 @@
...
-  image_project: cirrus-images
-  image: family/docker-kvm
-  platform: linux
-  nested_virtualization: true
-  # CPU limit: `16 / NTASK`: see https://cirrus-ci.org/faq/#are-there-any-limits
-  cpu: 8
-  # Memory limit: `4GB * NCPU`
-  memory: 32G
+task:
+  name: ci / fedora 34 (vagrant)
```
kolyshkin, line 10: Can be not rename the test (it is currently called `vagrant DISTRO:fedora34`)? It is marked as required in this repo configuration, and we'll need to change that.

Or, if you're renaming it, maybe call it vagrant/fedora or something, so we would not have to change it when we switch to f35.

.cirrus.yml
adrianreber: With your patch from #3106 it fails with:
```
# (in test file tests/integration/update.bats, line 548)
# Writing 1000000 to /sys/fs/cgroup/cpu,cpuacct/runc-cgroups-integration-test/cpu.rt_period_us
# /tmp/bats-run-106836/bats.116418.src: line 548: /sys/fs/cgroup/cpu,cpuacct/runc-cgroups-integration-test/cpu.rt_period_us: Permission denied
```

.cirrus.yml
adrianreber: done

.cirrus.yml
adrianreber: done

.cirrus.yml
adrianreber: done

.cirrus.yml
adrianreber: Changed the name back to `vagrant DISTRO:fedora34`. I still kept the new names of the CentOS tests as they are no longer vagrant based.

.cirrus.yml
kolyshkin: Yeah, this makes more sense. I'm afraid we need to have a hack for this in `tests/rootless.sh`'s `enable_cgroup`. Something like this:

```diff
diff --git a/tests/rootless.sh b/tests/rootless.sh
index bacea49d..36db5340 100755
--- a/tests/rootless.sh
+++ b/tests/rootless.sh
@@ -114,6 +114,8 @@ function enable_cgroup() {
                # necessary, and might actually be a bug in our impl of cgroup
                # handling.
                [[ "$cg" == "cpuset" ]] && chown rootless:rootless "$CGROUP_MOUNT/$cg$CGROUP_PATH/cpuset."{cpus,mems}
+               # The following is required by "update rt period and runtime".
+               [[ "$cg" == "cpu" ]] && chown rootless:rootless "$CGROUP_MOUNT/$cg$CGROUP_PATH/cpu.rt_"{period,quota}us
        done
        # cgroup v2
        if [[ -e "$CGROUP_MOUNT/cgroup.controllers" ]]; then
```

@AkihiroSuda do you maybe have some input on this?

kolyshkin: I am not sure this test ever worked. On GHA (ubuntu, rootless_cgroups) it says

> ok 118 update rt period and runtime # skip test requires cgroups_rt

.cirrus.yml
kolyshkin: Looks like not.

adrianreber: I was sure I did that change. Second try pushed just no.w

.cirrus.yml
kolyshkin: I have added the above fix to https://github.com/opencontainers/runc/pull/3106 -- again, feel free to cherry-pick @adrianreber

adrianreber: The added fix to #3106 makes other CI runs fail. As it can be seen in #3106

adrianreber: I adapted your second patch and now everything seems happy.

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

mrunalp (approved):

PR #3106, opened 2021-07-21
tests/int: fix "update rt period and runtime" for rootless

kolyshkin:
Since commit f09a3e1b8db33768, the value passed on to read starts with
a slash, resulting in the first element of the array to be empty.

As a result, the test tries to write to the top-level cgroup, which
fails when rootless:

>  Writing 1000000 to /sys/fs/cgroup/cpu,cpuacct//cpu.rt_period_us
>  /tmp/bats-run-106184/bats.115768.src: line 548: /sys/fs/cgroup/cpu,cpuacct//cpu.rt_period_us: Permission denied

To fix, remove the leading slash.

An alternative fix would be to do `for ((i = 1;` instead of `i = 0`, but
that seems less readable.

Fixes: f09a3e1b8db33768


kolyshkin:
Closing as this is included in #3104 (the second patch in a fixed form)

PR #3107, opened 2021-07-22
[WIP] [carry #2087] cgroups/systemd: add cgroup-v2 path to the list when using hybrid mode 

kolyshkin:
Carry #2087, which I need for #3059. This also demonstrates a bug in criu v3.15 (fixed in criu-dev).

IOW we need criu-3.16 to merge this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,12 @@ jobs:
         go-version: [1.15.x, 1.16.x]
         rootless: ["rootless", ""]
         race: ["-race", ""]
+        include:
+          # Also test against latest criu-dev
+          - go-version: 1.16.x
+            rootless: ""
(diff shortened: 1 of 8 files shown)


kolyshkin:
Seems like unified path is not being restored properly.

kolyshkin:
The exact error is (after restore with lazy):
> runc exec --cwd /bin test_busybox_restore echo ok (status=1):
> exec failed: unable to start container process: error adding pid 33554 to cgroups: failed to write 33554 to cgroup.procs: openat2 /sys/fs/cgroup/unified/machine.slice/runc-cgroups-integration-test-10279.scope/cgroup.procs: no such file or directory

kolyshkin:
Carrying this in #3059; this one was just for testing

PR #3109, opened 2021-07-22
seccomp: skip redundant rules

kolyshkin:
This fixes using runc with podman on my system (Fedora 34).

> $ podman --runtime $(pwd)/runc run --rm --memory 4M fedora echo it works
> Error: unable to start container process: error adding seccomp filter rule for syscall bdflush: permission denied: OCI permission denied

The problem is, libseccomp returns EPERM when a redundant rule (i.e. the
rule with the same action as the default one) is added, and podman (on
my machine) sets the following rules in config.json:
```json
    ....
    "seccomp": {
      "defaultAction": "SCMP_ACT_ERRNO",
      "architectures": [
        "SCMP_ARCH_X86_64",
        "SCMP_ARCH_X86",
        "SCMP_ARCH_X32"
      ],
      "syscalls": [
        {
          "names": [
            "bdflush",
            "io_pgetevents",
            ....
          ],
          "action": "SCMP_ACT_ERRNO",
          "errnoRet": 1
        },
        ....
```
(Note that defaultErrnoRet is not set, but it defaults to 1).

With this commit, it works:

> $ podman --runtime $(pwd)/runc run --memory 4M fedora echo it works
> it works

Similar crun commit:
 * https://github.com/containers/crun/commit/08229f3fb904c5ea19a7d9

Fixes: #1847
See also: https://github.com/containers/podman/issues/11031

## Changelog entry
```
Bugfixes:
* Redundant seccomp rules (i.e. those that has action equal to the default
  one) are now skipped. This fixes the inability to start a container with
  the "adding seccomp filter rule for syscall ..." error.
```

### DIFF
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -68,7 +68,7 @@ func InitSeccomp(config *configs.Seccomp) error {
 		if call == nil {
 			return errors.New("encountered nil syscall while initializing Seccomp")
 		}
-		if err := matchCall(filter, call); err != nil {
+		if err := matchCall(filter, call, defaultAction); err != nil {
 			return err
 		}
 	}
@@ -143,7 +143,7 @@ func getCondition(arg *configs.Arg) (libseccomp.ScmpCondition, error) {
 }
 
 // Add a rule to match a single syscall
-func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
+func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall, defAct libseccomp.ScmpAction) error {
 	if call == nil || filter == nil {
 		return errors.New("cannot use nil as syscall to block")
 	}
@@ -152,18 +152,23 @@ func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
 		return errors.New("empty string is not a valid syscall")
 	}
 
+	// Convert the call's action to the libseccomp equivalent
+	callAct, err := getAction(call.Action, call.ErrnoRet)
+	if err != nil {
+		return fmt.Errorf("action in seccomp profile is invalid: %w", err)
+	}
+	if callAct == defAct {
+		// This rule is redundant, silently skip it
+		// to avoid error from AddRule.
+		return nil
+	}
+
 	// If we can't resolve the syscall, assume it is not supported
 	// by this kernel. Warn about it, don't error out.
 	callNum, err := libseccomp.GetSyscallFromName(call.Name)
 	if err != nil {
 		logrus.Debugf("unknown seccomp syscall %q ignored", call.Name)
 		return nil
-	}
-
-	// Convert the call's action to the libseccomp equivalent
-	callAct, err := getAction(call.Action, call.ErrnoRet)
-	if err != nil {
-		return fmt.Errorf("action in seccomp profile is invalid: %w", err)
 	}
 
 	// Unconditional match - just add the rule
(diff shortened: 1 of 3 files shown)


libcontainer/seccomp/seccomp_linux.go
```diff
@@ -151,17 +151,20 @@ func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
...
 	}
 
-	// Convert the call's action to the libseccomp equivalent
-	callAct, err := getAction(call.Action, call.ErrnoRet)
+	// If we can't resolve the syscall, assume it's not supported on this kernel
+	// Ignore it, don't error out
+	callNum, err := libseccomp.GetSyscallFromName(call.Name)
 	if err != nil {
-		return fmt.Errorf("action in seccomp profile is invalid: %w", err)
+		return nil
```
chenk008, line 167: How about add some warning log?

kolyshkin: What you request is a new functionality, not really related to the fix, but yes, it makes sense.

Added as a separate commit, PTAL.

kolyshkin:
@opencontainers/runc-maintainers PTAL; this is a P0 as runc is rendered totally useless.

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -151,17 +152,23 @@ func matchCall(filter *libseccomp.ScmpFilter, call *configs.Syscall) error {
...
+	// Convert the call's action to the libseccomp equivalent
+	callAct, err := getAction(call.Action, call.ErrnoRet)
 	if err != nil {
+		return fmt.Errorf("action in seccomp profile is invalid: %w", err)
+	}
+	if callAct == defAct {
+		// This rule is redundant, silently skip it
+		// to avoid error from AddRule.
 		return nil
 	}
```
AkihiroSuda, line 159: We also have to compare errnoRet?

kolyshkin: No. The `callAct` and the `defAct` is a combo of the action (lower two bytes) and the error code (upper two bytes) -- see getAction which calls libseccomp-golang's `(ScmpAction).SetReturnCode`.

libcontainer/seccomp/seccomp_linux.go
chenk008: Thanks

cyphar: This shouldn't be a warning, because then you'll end up with a bunch of warnings on old kernels that serve no purpose. At best this should be a debug message.

cyphar (dismissed):
Please change the warning for unknown syscalls to a debug message. As for the rest of the patch, LGTM (though this seems like it's a bug in podman not in runc, but since the old behaviour is kind brittle and is not defined in the spec, I'm okay with working around it anyway).

kolyshkin:
> this seems like it's a bug in podman

Yes, it is (and it is being fixed).

But there were also older bug reports with the same diagnostics (not sure how they

>  since the old behaviour is kind brittle and is not defined in the spec

I think it is defined in spec (see the last paragraph in https://github.com/opencontainers/runc/issues/1847 description ... and yes, we need a CI job to run those tools, I'll get around to it later).

libcontainer/seccomp/seccomp_linux.go
kolyshkin: Done.

cyphar (approved):
LGTM.

cyphar:
Probably needs a changelog entry too.

kolyshkin:
> Probably needs a changelog entry too.

Added

mrunalp (approved):

PR #3110, opened 2021-07-24
libct/cg/devices: stop using regex

kolyshkin:
Looking into data generated by setting

	GODEBUG="inittrace=1"

I have noticed this line:

	init github.com/opencontainers/runc/libcontainer/cgroups/devices @1.2 ms, 0.020 ms clock, 10512 bytes, 133 allocs

This is the leader for both bytes and allocs among the packages from
this repo, and all of it is caused by a single regex:

	var devicesListRegexp = regexp.MustCompile(`^([abc])\s+(\d+|\*):(\d+|\*)\s+([rwm]+)$`)

It seems that the same parsing can be done without relying on
a regular expression, no decrease in readability or correctness
(including catching all invalid input), and 2x faster (according to
the benchmark added), and also hopefully makes runc start
slightly faster and leaner.

Before:

	BenchmarkParseLine-4   	  176240	      6768 ns/op	    6576 B/op	      64 allocs/op

After:

	BenchmarkParseLine-4   	  322441	      3535 ns/op	    5520 B/op	      53 allocs/op

[v2: single split with SplitFunc; fix a typo in an error message]
[v3: rebase after #3159 merge; re-ran benchmarks (results are similar)]

### DIFF
--- a/libcontainer/cgroups/devices/devices_emulator.go
+++ b/libcontainer/cgroups/devices/devices_emulator.go
@@ -22,12 +22,11 @@ package devices
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"io"
-	"regexp"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/devices"
 )
@@ -79,19 +78,21 @@ func (e *Emulator) IsAllowAll() bool {
 	return e.IsBlacklist() && len(e.rules) == 0
 }
 
-var devicesListRegexp = regexp.MustCompile(`^([abc])\s+(\d+|\*):(\d+|\*)\s+([rwm]+)$`)
-
 func parseLine(line string) (*deviceRule, error) {
-	matches := devicesListRegexp.FindStringSubmatch(line)
-	if matches == nil {
-		return nil, errors.New("line doesn't match devices.list format")
+	// Input: node major:minor perms.
+	fields := strings.FieldsFunc(line, func(r rune) bool {
+		return r == ' ' || r == ':'
+	})
+	if len(fields) != 4 {
+		return nil, fmt.Errorf("malformed devices.list rule %s", line)
 	}
+
 	var (
 		rule  deviceRule
-		node  = devices.Type(matches[1])
-		major = matches[2]
-		minor = matches[3]
-		perms = matches[4]
+		node  = devices.Type(fields[0])
+		major = fields[1]
+		minor = fields[2]
+		perms = fields[3]
 	)
 
 	// Parse the node type.
@@ -105,7 +106,6 @@ func parseLine(line string) (*deviceRule, error) {
 	case devices.BlockDevice, devices.CharDevice:
 		rule.meta.node = node
 	default:
-		// Should never happen!
 		return nil, fmt.Errorf("unknown device type %q", node)
 	}
 
@@ -134,7 +134,6 @@ func parseLine(line string) (*deviceRule, error) {
 	// Parse the access permissions.
 	rule.perms = devices.Permissions(perms)
 	if !rule.perms.IsValid() || rule.perms.IsEmpty() {
-		// Should never happen!
 		return nil, fmt.Errorf("parse access mode: contained unknown modes or is empty: %q", perms)
 	}
 
--- a/libcontainer/cgroups/devices/devices_emulator_test.go
+++ b/libcontainer/cgroups/devices/devices_emulator_test.go
@@ -19,8 +19,10 @@
 package devices
 
 import (
+	"bufio"
 	"bytes"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/opencontainers/runc/libcontainer/devices"
@@ -1111,4 +1113,32 @@ func TestDeviceEmulatorTransitionFromBlacklist(t *testing.T) {
 
 func TestDeviceEmulatorTransitionFromWhitelist(t *testing.T) {
 	testDeviceEmulatorTransition(t, false)
+}
+
+func BenchmarkParseLine(b *testing.B) {
+	list := `c *:* m
+b *:* m
+c 1:3 rwm
+c 1:5 rwm
+c 1:7 rwm
+c 1:8 rwm
+c 1:9 rwm
+c 5:0 rwm
+c 5:2 rwm
+c 136:* rwm
+c 10:200 rwm`
+
+	var r *deviceRule
+	var err error
+	for i := 0; i < b.N; i++ {
+		s := bufio.NewScanner(strings.NewReader(list))
+		for s.Scan() {
+			line := s.Text()
+			r, err = parseLine(line)
+		}
+		if err := s.Err(); err != nil {
+			b.Fatal(err)
+		}
+	}
+	b.Logf("rule: %v, err: %v", r, err)
 }


libcontainer/cgroups/devices/devices_emulator.go
```diff
@@ -79,19 +78,24 @@ func (e *Emulator) IsAllowAll() bool {
...
 }
 
-var devicesListRegexp = regexp.MustCompile(`^([abc])\s+(\d+|\*):(\d+|\*)\s+([rwm]+)$`)
-
 func parseLine(line string) (*deviceRule, error) {
-	matches := devicesListRegexp.FindStringSubmatch(line)
-	if matches == nil {
-		return nil, errors.New("line doesn't match devices.list format")
+	// The three fields are: node major:minor perms.
+	fields := strings.Fields(line)
```
cyphar, line 83: This will no longer fail to match `[abc]` and `[rwm]`. Does this matter in practice? Probably not, but it is a difference to the regex version.

kolyshkin: There are checks later in the code for both.

For the first field ("node"), in case it's not one of `a`, `b`, or `c`, the error is returned (line 113).

For the third field ("perms"), `rule.Perms.IsValid()` and `rule.Perms.IsEmpty()` are called to ensure there's nothing but `rwm` (line 142).

So, we're not losing any functionality (it's just the errors are more focused now).

kolyshkin: Glad that you asked, I looked at the code and found I forgot to remove one `// Should never happen` comment (because now it can happen).

cyphar (dismissed):
LGTM

kolyshkin:
@AkihiroSuda @thaJeztah PTAL

thaJeztah (dismissed):
LGTM, but left some suggestions (no show-stoppers).


I was also curious if changing the `string.Fields` etc functions to `bytes.Fields` would improve performance further, but that actually performed worse (🤷‍♂️), so 👍 on the current approach.

libcontainer/cgroups/devices/devices_emulator.go
```diff
@@ -79,19 +78,24 @@ func (e *Emulator) IsAllowAll() bool {
...
+	if len(fields) != 3 {
+		return nil, fmt.Errorf("malformed device.list rule %s", line)
+	}
+	// Parse the major:minor pair.
+	mm := strings.Split(fields[1], ":")
```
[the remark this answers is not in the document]
kolyshkin: So, I think since this is not user-provided input, so we don't have to be extra suspicious when parsing it.

`SplitN` is nice as it prevents bad input attacks (e.g. something like a few million `:` characters), but here we do not ever expect it, and so I lean towards better code readability (which SplitN decreases a bit).

libcontainer/cgroups/devices/devices_emulator.go
```diff
@@ -79,19 +78,24 @@ func (e *Emulator) IsAllowAll() bool {
...
 	var (
 		rule  deviceRule
-		node  = matches[1]
-		major = matches[2]
-		minor = matches[3]
-		perms = matches[4]
+		node  = fields[0]
+		major = mm[0]
+		minor = mm[1]
+		perms = fields[2]
```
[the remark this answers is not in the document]
kolyshkin: I was trying to balance between making the changes and keeping the patch easy to review. These changes will make
* the patch a bit more cluttered, and
* the resulting code a tad harder to read:
  * we parse in one place, check in another (can be fixed by moving the check next to parsing);
  * the parse order is no longer the same as the text.
  
For these reasons, I'd like to keep this as is.

libcontainer/cgroups/devices/devices_emulator_test.go
```diff
@@ -1112,3 +1114,31 @@ func TestDeviceEmulatorTransitionFromBlacklist(t *testing.T) {
 func TestDeviceEmulatorTransitionFromWhitelist(t *testing.T) {
 	testDeviceEmulatorTransition(t, false)
 }
+
+func BenchmarkParseLine(b *testing.B) {
+	list := `c *:* m
```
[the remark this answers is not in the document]
kolyshkin: I don't see a reason why `-benchmem` would not work if we want to see allocs stats.

libcontainer/cgroups/devices/devices_emulator_test.go
```diff
@@ -1112,3 +1114,31 @@ func TestDeviceEmulatorTransitionFromBlacklist(t *testing.T) {
...
+		s := bufio.NewScanner(strings.NewReader(list))
+		for s.Scan() {
+			line := s.Text()
+			r, err = parseLine(line)
+		}
+		if err := s.Err(); err != nil {
+			b.Fatal(err)
+		}
+	}
+	b.Logf("rule: %v, err: %v", r, err)
```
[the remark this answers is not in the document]
kolyshkin: Yes, this is what I usually do to keep variables used -- this also helps a visually assess that the code being benchmarked is doing something and working OK.

This log is out of the loop to not affect the performance (much).

As we don't run benchmarks in CI, this clutter is only shown to whoever is running it.

libcontainer/cgroups/devices/devices_emulator.go
kolyshkin, line 83: Ideally, I'd use `strings.Split(line, " ")` here -- it works just fine as contents from the kernel do not have any extra spaces, and makes is 1.5x faster. OTOH `strings.Fields` is safer, as it will eat extra whitespace. I'm on the verge here -- WDYT @thaJeztah ?

libcontainer/cgroups/devices/devices_emulator.go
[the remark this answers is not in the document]
thaJeztah: > So, I think since this is not user-provided input, so we don't have to be extra suspicious when parsing it.

Ack, yes, definitely not a blocker. I was also looking at the code, and specifying a limit also skips a `bytes.Count()` to be done; https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/strings/strings.go#L243-L245, but agreed that this is not user input, so malformed / bad values _should_ not be a thing here.

libcontainer/cgroups/devices/devices_emulator_test.go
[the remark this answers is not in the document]
thaJeztah: Ah, gotcha. I got a bit confused because you were actually printing it.

libcontainer/cgroups/devices/devices_emulator_test.go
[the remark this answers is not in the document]
thaJeztah: 😅 I always forget the name for the flag (`-benchmem`), so am lazy and just set it in the benchmark to enable it by default.

Definitely "just a suggestion". Let me mark this as "resolved" to not clutter the PR

libcontainer/cgroups/devices/devices_emulator.go
thaJeztah: I'm trying to Benchmark the difference, but for some reason, I don't seem to be doing things right (it shows `strings.Fields()` to be actually _faster_ ? I was considering `strings.FieldsFunc()`, as it would allow splitting on spaces _only_ (not tabs, newlines, etc), and still keeping the advantage of `strings.Fields()` (trimming whitespace)

I tried `strings.Fields()`, `strings.FieldsFunc()`, and `strings.Split()`

<details>
    <summary>test_test.go</summary>
    
    ```go
    package main
    
    import (
    	"bufio"
    	"strings"
    	"testing"
    )
    
    
    func BenchmarkParseFields(b *testing.B) {
    	list := `c *:* m
    b *:* m
    c 1:3 rwm
    c 1:5 rwm
    c 1:7 rwm
    c 1:8 rwm
    c 1:9 rwm
    c 5:0 rwm
    c 5:2 rwm
    c 136:* rwm
    c 10:200 rwm`
    
    	var out []string
    	for i := 0; i < b.N; i++ {
    		s := bufio.NewScanner(strings.NewReader(list))
    		for s.Scan() {
    			line := s.Text()
    			out = strings.Fields(line)
    		}
    		if err := s.Err(); err != nil {
    			b.Fatal(err)
    		}
    	}
    	b.Logf("rule: %v", out)
    }
    
    func isSpace(r rune) bool {
    	return r == ' '
    }
    
    func BenchmarkParseFieldFunc(b *testing.B) {
    	list := `c *:* m
    b *:* m
    c 1:3 rwm
    c 1:5 rwm
    c 1:7 rwm
    c 1:8 rwm
    c 1:9 rwm
    c 5:0 rwm
    c 5:2 rwm
    c 136:* rwm
    c 10:200 rwm`
    
    	var out []string
    	for i := 0; i < b.N; i++ {
    		s := bufio.NewScanner(strings.NewReader(list))
    		for s.Scan() {
    			line := s.Text()
    			out = strings.FieldsFunc(line, isSpace)
    		}
    		if err := s.Err(); err != nil {
    			b.Fatal(err)
    		}
    	}
    	b.Logf("rule: %v", out)
    }
    
    func BenchmarkParseSplit(b *testing.B) {
    	list := `c *:* m
    b *:* m
    c 1:3 rwm
    c 1:5 rwm
    c 1:7 rwm
    c 1:8 rwm
    c 1:9 rwm
    c 5:0 rwm
    c 5:2 rwm
    c 136:* rwm
    c 10:200 rwm`
    
    	var out []string
    	for i := 0; i < b.N; i++ {
    		s := bufio.NewScanner(strings.NewReader(list))
    		for s.Scan() {
    			line := s.Text()
    			out = strings.Split(line, " ")
    		}
    		if err := s.Err(); err != nil {
    			b.Fatal(err)
    		}
    	}
    	b.Logf("rule: %v", out)
    }
    ```

</details>

But for some reason, `strings.Fields()` is... fastest?

```console
go test -bench=. -benchmem -run XXX -v
goos: darwin
goarch: amd64
pkg: foo
cpu: Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
BenchmarkParseFields-8      	  510610	      1994 ns/op	    4800 B/op	      24 allocs/op
BenchmarkParseFieldFunc-8   	  506881	      2286 ns/op	    4800 B/op	      24 allocs/op
BenchmarkParseSplit-8       	  465390	      2182 ns/op	    4800 B/op	      24 allocs/op
PASS
ok  	foo	3.277s
```


That said, `strings.FieldsFunc()` _WOULD_ allow us to do all splitting at once (including `major:minor`); https://play.golang.org/p/klnqzHEJchG

```go
package main

import (
	"bufio"
	"fmt"
	"strings"
)

func main() {
	list := `c *:* m
b *:* m
c 1:3      rwm
c     1:5 rwm
c 1:7 rwm
c 1:8 rwm
c 1:9 rwm
c 5:0 rwm
c 5:2 rwm
c 136:* rwm
c 10:200 rwm`

	var out []string
	s := bufio.NewScanner(strings.NewReader(list))
	for s.Scan() {
		line := s.Text()
		out = strings.FieldsFunc(line, splitFn)
		fmt.Printf("%+v\n", out)
	}
	if err := s.Err(); err != nil {
		fmt.Println(err)
	}
}

func splitFn(r rune) bool {
	return r == ' ' || r == ':'
}
```

Which produces:

```
[c * * m]
[b * * m]
[c 1 3 rwm]
[c 1 5 rwm]
[c 1 7 rwm]
[c 1 8 rwm]
[c 1 9 rwm]
[c 5 0 rwm]
[c 5 2 rwm]
[c 136 * rwm]
[c 10 200 rwm]
```

thaJeztah: Giving that idea a try (patch below);

<details>

```patch
diff --git a/libcontainer/cgroups/devices/devices_emulator.go b/libcontainer/cgroups/devices/devices_emulator.go
index aea0f85e..6345e83b 100644
--- a/libcontainer/cgroups/devices/devices_emulator.go
+++ b/libcontainer/cgroups/devices/devices_emulator.go
@@ -78,24 +78,23 @@ func (e *Emulator) IsAllowAll() bool {
 	return e.IsBlacklist() && len(e.rules) == 0
 }
 
+func splitFn(r rune) bool {
+	return r == ' ' || r == ':'
+}
+
 func parseLine(line string) (*deviceRule, error) {
-	// The three fields are: node major:minor perms.
-	fields := strings.Fields(line)
-	if len(fields) != 3 {
+	// The four fields are: node major minor perms.
+	fields := strings.FieldsFunc(line, splitFn)
+	if len(fields) != 4 {
 		return nil, fmt.Errorf("malformed device.list rule %s", line)
 	}
-	// Parse the major:minor pair.
-	mm := strings.Split(fields[1], ":")
-	if len(mm) != 2 {
-		return nil, fmt.Errorf("unexpected minor:major pair %s", fields[1])
-	}
 
 	var (
 		rule  deviceRule
 		node  = fields[0]
-		major = mm[0]
-		minor = mm[1]
-		perms = fields[2]
+		major = fields[1]
+		minor = fields[2]
+		perms = fields[3]
 	)
 
 	// Parse the node type.
```

</details>

Before:

    BenchmarkParseLine-4   	  207354	      5622 ns/op	    5696 B/op	      64 allocs/op

After

    BenchmarkParseLine-4   	  201634	      5288 ns/op	    5520 B/op	      53 allocs/op

kolyshkin: Indeed, I was mistaken and in this scenario `strings.Split` is faster (not sure why I thought otherwise yesterday, but I discount that to the ongoing heat wave and the lack of AC at home 😓).

I like the idea of using strings.SplitFunc to avoid a second split. Implemented.

thaJeztah: > I like the idea of using strings.SplitFunc to avoid a second split. Implemented.

Yes! I was drafting the change, and suddenly thought; "hey, we can do it all at once!"

thaJeztah (approved):
LGTM, thanks!

kolyshkin:
Rebased after #3159 merge; PTAL @thaJeztah @cyphar

cyphar (approved):
LGTM.

thaJeztah (approved):
LGTM

PR #3112, opened 2021-07-27
libct/int/checkpoint_test: fix ParentImage (and bogus failure with criu 3.16)

kolyshkin:
The ParentImage set by the test should be a path relative to
ImagesDirectory, pointing to a parent images directory (created
by pre-dump).

The parent directory is created by TempDir and so its name is not
constant but has a variable suffix. So, the config was pointing to
a non-existent directory.

This left unnoticed by criu as it assumed the parent image does not
exist, and performed a full dump (meaning the pre-dump was not used).

Since criu PR https://github.com/checkpoint-restore/criu/pull/1403 (which will
be a part of criu 3.16) that is no longer the case -- the invalid parent
path is treated as an error, and so our test fails like this:

```
== RUN   TestCheckpoint
    checkpoint_test.go:145: === /tmp/criu070876105/dump.log ===
    checkpoint_test.go:145: open /tmp/criu070876105/dump.log: no such file or directory
    checkpoint_test.go:146: criu failed: type DUMP errno 56
        log file: /tmp/criu070876105/dump.log
--- FAIL: TestCheckpoint (0.26s)
```

Fix this by using the actual name of the parent image dir.

Fixes: 98f004182b9a23245

### DIFF
--- a/libcontainer/integration/checkpoint_test.go
+++ b/libcontainer/integration/checkpoint_test.go
@@ -133,10 +133,12 @@ func testCheckpoint(t *testing.T, userns bool) {
 	ok(t, err)
 	defer remove(imagesDir)
 
+	relParentDir, err := filepath.Rel(imagesDir, parentDir)
+	ok(t, err)
 	checkpointOpts := &libcontainer.CriuOpts{
 		ImagesDirectory: imagesDir,
 		WorkDirectory:   imagesDir,
-		ParentImage:     "../criu-parent",
+		ParentImage:     relParentDir,
 	}
 	dumpLog := filepath.Join(checkpointOpts.WorkDirectory, "dump.log")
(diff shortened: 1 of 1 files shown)


kolyshkin:
Somewhat related: #2913

@adrianreber @liusdu PTAL

adrianreber:
Looks correct. Just recently I did a similar fix in Podman.

avagin:
LGTM

kolyshkin:
OK it seems that the same bug is implicitly fixed by commit 3bc606e9d30e5e616e4602bd5b1b1610cc3b8fbe (which switches to a single upper temp dir and fixed `criu-parent` and `criu` subdirs, so `../criu-parent` is now valid.

IOW, this is no longer needed.

PR #3113, opened 2021-07-27
runc init: remove some code

kolyshkin:
Remove some code from runc init which is not needed. Please review commit-by-commit.

1. Remove option parsing from runc init.
2. Remove cgroupns sync mechanism (as pointed out in #2446 and in TODO).

### DIFF
--- a/init.go
+++ b/init.go
@@ -10,11 +10,12 @@ import (
 	"github.com/opencontainers/runc/libcontainer/logs"
 	_ "github.com/opencontainers/runc/libcontainer/nsenter"
 	"github.com/sirupsen/logrus"
-	"github.com/urfave/cli"
 )
 
 func init() {
 	if len(os.Args) > 1 && os.Args[1] == "init" {
+		// This is the golang entry point for runc init, executed
+		// before main() but after libcontainer/nsenter's nsexec().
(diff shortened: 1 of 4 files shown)


cyphar (approved):
LGTM.

init.go
```diff
@@ -10,11 +10,12 @@ import (
 	"github.com/opencontainers/runc/libcontainer/logs"
 	_ "github.com/opencontainers/runc/libcontainer/nsenter"
 	"github.com/sirupsen/logrus"
-	"github.com/urfave/cli"
 )
 
 func init() {
 	if len(os.Args) > 1 && os.Args[1] == "init" {
+		// This is the golang entry point for runc init, executed
+		// before main() but after libcontainer/nsenter's nsexec().
```
cyphar, line 18: Might not hurt to mention the function never returns (since this all ends in `execve`) so `main` never actually runs.

mrunalp (approved):

init.go
kolyshkin: I had something like that but eventually removed it as this is somewhat clear (for one thing, the code within the block ends with a `panic`).

I will try to improve this further in #3114 (which is a followup to this PR).

Issue #3119, opened 2021-07-29
nsexec logging race


kolyshkin:
As reported in https://kubernetes.slack.com/archives/CAZH62UR1/p1627584431224800
```console
[20:46] nb3:apps% kubectl logs nextcloud-nextcloud-846d6bbcf-9mmp8 nextcloud
time="2021-07-29T20:45:35+02:00" level=error msg="failed to decode \"{\\\"level\\\":\\\"debug\\\", \\\"msg\\\": \\\"nsexec-0[7972]: forward stage-1 (7973) and stage-2 (7974) pids to runc{\\\"level\\\":\\\"debug\\\", \\\"msg\\\": \\\"nsexec-1[7973]: signal completion to stage-0\\\"}\" to json: invalid character 'l' after object key:value pair"
time="2021-07-29T20:45:35+02:00" level=error msg="failed to decode \"\\\"}\" to json: unexpected end of JSON input"
Initializing nextcloud 20.0.11.1 ...
Initializing finished
New nextcloud instance
[29-Jul-2021 18:45:40] NOTICE: fpm is running, pid 1
[29-Jul-2021 18:45:40] NOTICE: ready to handle connections
```

What we see here is runc failed to decode JSON log message from runc init's nsexec.c. The messages are:

```
{"level":"debug", "msg": "nsexec-0[7972]: forward stage-1 (7973) and stage-2 (7974) pids to runc{"level":"debug", "msg": "nsexec-1[7973]: signal completion to stage-0"}
}
```

Looks like the code hit a jackpot of a debug log race between the parent at

https://github.com/opencontainers/runc/blob/d962bb0cb0060ec5bd54878b2308452243018d98/libcontainer/nsenter/nsexec.c#L820-L821

and the first child at

https://github.com/opencontainers/runc/blob/d962bb0cb0060ec5bd54878b2308452243018d98/libcontainer/nsenter/nsexec.c#L1021

Cc @cyphar

kolyshkin:
As the production rarely uses debug, in practice this is fixed by https://github.com/opencontainers/runc/pull/3114/commits/b1bc76248c642660269a0764b02bf7ed2aa77605

kolyshkin:
Another possible solution is something like this
```diff
diff --git a/libcontainer/nsenter/nsexec.c b/libcontainer/nsenter/nsexec.c
index 8199faf3..0c9eca00 100644
--- a/libcontainer/nsenter/nsexec.c
[...]

PR #3114, opened 2021-07-27
[DNM] runc init log refactor

kolyshkin:
This PR is mostly about runc init log refactoring. Makes the code more readable, improves its tests, fixes some issues.

1. Remove `logs.ConfigureLogging` whose existence always bugged me.
    - init.go, main.go: don't use logs.ConfigureLogging
    - libct/logs: remove ConfigureLogging

2. Improve libct/logs and its tests.
    - libct/logs: test: make more robust
    - libct/logs: parse log level implicitly
    - libct/logs: do not show caller in nsexec logs

3. Fix/improve libct/nsenter tests.
    - libct/nsenter: test: logging nits
    - libct/nsenter: test: improve newPipe
    - libct/nsenter: test: fix TestNsenterValidPaths
    - libct/nsenter: test: improve TestNsenterChildLogging
    - libct/nsenter: test: rm misleading comments

4. Make nsexec aware of the log level (i.e. do not emit logs that are to be discarded).
    - utils_linux: simplify newProcess
    - runc init: pass _LIBCONTAINER_LOGLEVEL as int
    - libct/nsenter/nsexec.c: improve bail
    - libct/nsenter/nsexec.c: factor out getenv_int
    - libct/nsenter/nsexec.c: honor _LIBCONTAINER_LOGLEVEL

5. Small improvement (not related to init logs)
    - runc --debug: shorter caller info

Update: split into a few smaller PRS:
 - [x] Items 1, 2, and 5 from the list above go to #3157
 - [x] Item 3 goes to #3158
 - [x] item 4 goes to #3201

PR #3120, opened 2021-07-29
libct/nsenter: fix logging race in nsexec (regression in rc94)

kolyshkin:
As reported in issue #3119, there is a race in nsexec logging
that can lead to garbled json received by log forwarder, which
complains about it with a "failed to decode" error.

This happens because dprintf (used since the very beginning of nsexec
logging introduced in commit ba3cabf932943) relies on multiple write(2)
calls, and with additional logging added by 64bb59f5920b15d (appeared in rc94)
a race is possible between runc init parent and its children.

The fix is to prepare a string and write it using a single call to
write(2).

Fixes: #3119

Practically, though, debug logging is not needed in 99.9% of all cases,
and so PR #3114 (commit https://github.com/opencontainers/runc/pull/3114/commits/b1bc76248c642660269a0764b02bf7ed2aa77605) eliminates the possibility
of it having this race in the first place. Yet I think this fix makes sense.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -136,7 +136,7 @@ int setns(int fd, int nstype)
 
 static void write_log(const char *level, const char *format, ...)
 {
-	char *message = NULL, *stage = NULL;
+	char *message = NULL, *stage = NULL, *json = NULL;
 	va_list args;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


kolyshkin:
I think this is a candidate for backporting to 1.0 -- while the race is rare, the consequences are fatal.

libcontainer/nsenter/nsexec.c
```diff
@@ -164,11 +164,16 @@ static void write_log(const char *level, const char *format, ...)
 	if (ret < 0)
 		goto out;
 
-	dprintf(logfd, "{\"level\":\"%s\", \"msg\": \"%s[%d]: %s\"}\n", level, stage, getpid(), message);
+	ret = asprintf(&json, "{\"level\":\"%s\", \"msg\": \"%s[%d]: %s\"}\n", level, stage, getpid(), message);
+	if (ret < 0)
```
AkihiroSuda, line 168: `json` has to be NULLified on error (or avoid `free` on error)

https://man7.org/linux/man-pages/man3/asprintf.3.html

kolyshkin: Indeed! Fixed.

AkihiroSuda (approved):

mrunalp (approved):

PR #3126, opened 2021-08-03
`linux`: Support setting execution domain via linux `personality`.

flouthoc:
Hi Team,

Following PR allows users to configure execution domain via `syscall.SYS_PERSONALITY`.
Valid inputs supported via spec are limited to `LINUX` and `LINUX32`.

References:
https://man7.org/linux/man-pages/man2/personality.2.html
https://raw.githubusercontent.com/torvalds/linux/master/include/uapi/linux/personality.h

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -206,6 +206,9 @@ type Config struct {
 	// RootlessCgroups is set when unlikely to have the full access to cgroups.
 	// When RootlessCgroups is set, cgroups errors are ignored.
 	RootlessCgroups bool `json:"rootless_cgroups,omitempty"`
+
+	// Personality contains configuration for the Linux personality syscall.
+	Personality *LinuxPersonality `json:"personality,omitempty"`
 }
 
 type (
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -230,6 +230,13 @@ func (c *linuxContainer) Set(config configs.Config) error {
 func (c *linuxContainer) Start(process *Process) error {
 	c.m.Lock()
 	defer c.m.Unlock()
+	// configure linux personality before starting the process
+	if c.config.Personality != nil {
+		err := system.SetLinuxPersonality(c.config.Personality.Domain)
+		if err != nil {
+			return err
+		}
+	}
 	if c.config.Cgroups.Resources.SkipDevices {
 		return &ConfigError{"can't start container with SkipDevices set"}
 	}
--- a/libcontainer/integration/exec_test.go
+++ b/libcontainer/integration/exec_test.go
@@ -137,6 +137,26 @@ func TestIPCBadPath(t *testing.T) {
 	}
 }
 
+func TestPersonality(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
+	config := newTemplateConfig(t, &tParam{})
+
+	// change execution domain to i686
+	config.Personality = &configs.LinuxPersonality{
+		Domain: configs.PER_LINUX32,
+	}
+
+	out, _, err := runContainer(t, config, "", "/bin/sh", "-c", "uname -a")
+	ok(t, err)
+	// output must contain kernel architecture configured as i686
+	if !strings.Contains(out.Stdout.String(), "i686") {
+		t.Fatalf("expected kernel architecture i686 configured via personality")
+	}
+}
+
 func TestRlimit(t *testing.T) {
 	testRlimit(t, false)
 }
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -305,6 +305,11 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 				MemBwSchema:   spec.Linux.IntelRdt.MemBwSchema,
 			}
 		}
+		if spec.Linux.Personality != nil {
+			config.Personality = &configs.LinuxPersonality{
+				Domain: getLinuxPersonalityFromStr(string(spec.Linux.Personality.Domain)),
+			}
+		}
 	}
 	if spec.Process != nil {
 		config.OomScoreAdj = spec.Process.OOMScoreAdj
@@ -389,6 +394,21 @@ func convertSecToUSec(value dbus.Variant) (dbus.Variant, error) {
 		return value, errors.New("not a number")
 	}
 	return dbus.MakeVariant(sec), nil
+}
+
+// getLinuxPersonalityFromStr converts the string domain received from spec to equivalent integer.
+// check libcontainer/configs/config_linux.go
+// check https://raw.githubusercontent.com/torvalds/linux/master/include/uapi/linux/personality.h
+// valid arguments
+// 	LINUX i.e PER_LINUX (since Linux 1.2.0).
+//   	LINUX32 i.e PER_LINUX32 (since Linux 2.2) LINUX32 will set the uname system call to show a 32 bit CPU type, such as i686.
+// 	everything unknown default to PER_LINUX.
+func getLinuxPersonalityFromStr(domain string) int {
+	// defaults to PER_LINUX
+	if domain == "LINUX32" {
+		return configs.PER_LINUX32
+	}
+	return configs.PER_LINUX
 }
 
 func initSystemdProps(spec *specs.Spec) ([]systemdDbus.Property, error) {
--- a/libcontainer/system/linux.go
+++ b/libcontainer/system/linux.go
@@ -3,7 +3,9 @@
 package system
 
 import (
+	"os"
 	"os/exec"
+	"syscall"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
@@ -107,4 +109,14 @@ func GetSubreaper() (int, error) {
 	}
 
 	return int(i), nil
+}
+
+// SetLinuxPersonality sets the Linux execution personality. For more information see the personality syscall documentation.
+// checkout getLinuxPersonalityFromStr() from libcontainer/specconv/spec_linux.go for type conversion.
+func SetLinuxPersonality(persona int) error {
+	_, _, errno := syscall.Syscall(syscall.SYS_PERSONALITY, uintptr(persona), 0, 0)
+	if errno != 0 {
+		return &os.SyscallError{Syscall: "set_personality", Err: errno}
+	}
+	return nil
 }
--- a/libcontainer/configs/config_linux.go
+++ b/libcontainer/configs/config_linux.go
@@ -9,6 +9,19 @@ var (
 	errNoGroupMap = errors.New("User namespaces enabled, but no group mapping found.")
 )
 
+// Please check https://man7.org/linux/man-pages/man2/personality.2.html for const details.
+// https://raw.githubusercontent.com/torvalds/linux/master/include/uapi/linux/personality.h
+const (
+	PER_LINUX   = 0x0000
+	PER_LINUX32 = 0x0008
+)
+
+type LinuxPersonality struct {
+	// Domain for the personality
+	// can only contain values "LINUX" and "LINUX32"
+	Domain int `json:"domain"`
+}
+
 // HostUID gets the translated uid for the process on host which could be
 // different when user namespaces are enabled.
 func (c Config) HostUID(containerId int) (int, error) {


utils_linux.go
```diff
@@ -272,6 +273,13 @@ func (r *runner) run(config *specs.Process) (int, error) {
 	if err = r.checkTerminal(config); err != nil {
 		return -1, err
 	}
+	// configure linux personality
+	if r.container.Config().Personality != nil {
+		err = setLinuxPersonality(r.container.Config().Personality.Domain)
```
flouthoc, line 278: Should we `fail` for this err ?  I low-key want to ignore this err if `syscall.SYS_PERSONALITY` fails  proceed as if nothing happened. I'll probably wait for mods to suggest here.

AkihiroSuda: Failing seems better for security.

I guess we should also fail when unknown `Personality.Flags` are set

flouthoc: @AkihiroSuda Cool i agree it makes sense from security perspective so i will let the check be there. For `Personality.Flags` i did not implement it as `runtime-spec` suggests that no flag values are supported as well as domain is limited to `LINUX` and `LINUX32`. So anything unknown defaults to `LINUX` i.e `0x0000`.

libcontainer/configs/config.go
```diff
@@ -206,6 +206,9 @@ type Config struct {
 	// RootlessCgroups is set when unlikely to have the full access to cgroups.
 	// When RootlessCgroups is set, cgroups errors are ignored.
 	RootlessCgroups bool `json:"rootless_cgroups,omitempty"`
+
+	// Personality contains configuration for the Linux personality syscall
```
kolyshkin, line 210: nit: missing period at EOL.

kolyshkin:
@flouthoc can you describe the use case? I thought it's for starting i386 binaries but they work as is.

I think internally it's better to represent the personality value as a number rather than a string.

kolyshkin:
This should shed some light: https://github.com/opencontainers/runtime-spec/commit/5cc25d0a579261273809ff9db503a49af4771aaa

crun support: https://github.com/containers/crun/commit/afc183b85518a14828d264fb983392f1a65009e0

libcontainer/configs/config.go
flouthoc: @kolyshkin resolved in latest commit.

flouthoc:
@kolyshkin the commit you shared describes the edge-case but I also think that build for any `i386` or `32-bit` binary will fail if current kernel is on `64-bit` also certain `32-bit` userspace apps expect `uname -m` to report `i686`. Following cases could be resolved by setting `LINUX32` execution domain.

> I think internally it's better to represent the personality value as a number rather than a string.

For this personalty  is defined as number `const` internally in runc but accepted as `string` via spec since that is how it is defined in upstream runtime-spec this would required change in upstream. Please correct me if i am wrong.
```go 
const (
	PER_LINUX   = 0x0000
	PER_LINUX32 = 0x0008
)
```

kolyshkin:
It's kind of weird that
1. this is still not implemented
2. this is not implemented by checking the architecture of container init binary (so, say, if we're running i386 binary, runc calls `personality(PER_LINUX32)` for us.

But I guess it is what it is.

kolyshkin:
> For this personalty is defined as number const internally in runc but accepted as string via spec since that is how it is defined in upstream runtime-spec this would required change in upstream. Please correct me if i am wrong.

I didn't mean to change the runtime spec. What I meant is we're converting the spec into our internal presentation anyway (in libcontainer/specconv), and this is where, I think, we should do a string -> number translation.

flouthoc:
@kolyshkin Applied requested changes could you please review again.

flouthoc:
@AkihiroSuda @kolyshkin We would also need downstream pr on managers / shims. https://github.com/containers/podman/pull/11141

flouthoc:
@kolyshkin anything else which needs to be done from my side.

utils_linux.go
```diff
@@ -454,3 +462,14 @@ func startContainer(context *cli.Context, spec *specs.Spec, action CtAct, criuOp
...
 	return r.run(spec.Process)
 }
+
+/* setLinuxPersonality sets the Linux execution personality. For more information see the personality syscall documentation.
+   checkout getLinuxPersonalityFromStr() from libcontainer/specconv/spec_linux.go for type conversion.
+*/
+func setLinuxPersonality(persona int) error {
+	_, _, errno := syscall.Syscall(syscall.SYS_PERSONALITY, uintptr(persona), 0, 0)
+	if errno != 0 {
+		return fmt.Errorf("syscall: SYS_PERSONALITY failed with error code: %d", errno)
```
kolyshkin, line 472: please wrap errno using `%w` (so it could be unwrapped).

utils_linux.go
```diff
@@ -454,3 +462,14 @@ func startContainer(context *cli.Context, spec *specs.Spec, action CtAct, criuOp
 	}
 	return r.run(spec.Process)
 }
+
+/* setLinuxPersonality sets the Linux execution personality. For more information see the personality syscall documentation.
+   checkout getLinuxPersonalityFromStr() from libcontainer/specconv/spec_linux.go for type conversion.
+*/
```
kolyshkin, line 468: Please use `//` comments.

kolyshkin (dismissed):
Couple of nits, and we need a test case.

utils_linux.go
flouthoc: @kolyshkin this is resolved in latest commit.

utils_linux.go
flouthoc: @kolyshkin This is resolved in latest commit.

flouthoc:
@kolyshkin Could you please take a look again ? resolved the feedback comments and added integration tests.  Also changed the approach a little bit.

* Instead of invoking syscall from `runner` we are now invoking it from `linuxcontainer.Start` , since integration tests are invoking `Start` directly instead of using **CLI** also new approach seems much better to me since `personality` is limited only to linux containers.

libcontainer/configs/config.go
flouthoc: Resolved this in latest commit.

flouthoc:
@kolyshkin Squeezed personality tests with exec integration tests, let me know if separate file needs to created.

flouthoc:
@kolyshkin were you able to take a look at this, is there anything else with needs to be done from my side

libcontainer/container_linux.go
```diff
@@ -228,6 +228,14 @@ func (c *linuxContainer) Set(config configs.Config) error {
 }
 
 func (c *linuxContainer) Start(process *Process) error {
+	// configure linux personality before starting the process
```
kolyshkin, line 231: why is this done before the lock? (right below)

libcontainer/container_linux.go
```diff
@@ -228,6 +228,14 @@ func (c *linuxContainer) Set(config configs.Config) error {
...
+	if c.Config().Personality != nil {
```
kolyshkin, line 232: Why are you using `c.Config().` rather than `c.config.`?

libcontainer/container_linux.go
flouthoc: I can move this below the lock that should do no harm but idea was to change execution domain before anything happens for the process.

libcontainer/container_linux.go
flouthoc: @kolyshkin this is my bad. I'll fix this.

flouthoc: @kolyshkin resolved in latest commit.

flouthoc:
@kolyshkin resolved all the feedback points, could you please take a look again.

libcontainer/system/linux.go
```diff
@@ -108,3 +110,13 @@ func GetSubreaper() (int, error) {
 
 	return int(i), nil
 }
+
+// SetLinuxPersonality sets the Linux execution personality. For more information see the personality syscall documentation.
+// checkout getLinuxPersonalityFromStr() from libcontainer/specconv/spec_linux.go for type conversion.
+func SetLinuxPersonality(persona int) error {
+	_, _, errno := syscall.Syscall(syscall.SYS_PERSONALITY, uintptr(persona), 0, 0)
+	if errno != 0 {
+		return fmt.Errorf("syscall: SYS_PERSONALITY failed with error code: %w", errno)
```
kolyshkin, line 119: You can do something like `return &os.SyscallError{Syscall: "set_personality", Err: errno}`, but it's not really important.

libcontainer/integration/template_test.go
```diff
@@ -22,8 +22,9 @@ var standardEnvironment = []string{
 const defaultMountFlags = unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NODEV
 
 type tParam struct {
-	userns  bool
-	systemd bool
+	userns      bool
+	systemd     bool
+	personality int
```
kolyshkin, line 27: Unless it's used in a few tests, I wouldn't add this as a parameter.

As it is used in 1 test only, better to set it in the test itself.

flouthoc: @kolyshkin Thanks this is resolved in latest commit

libcontainer/system/linux.go
flouthoc: resolved in latest commit.

libcontainer/container_linux.go
```diff
@@ -230,6 +230,13 @@ func (c *linuxContainer) Set(config configs.Config) error {
 func (c *linuxContainer) Start(process *Process) error {
 	c.m.Lock()
 	defer c.m.Unlock()
+	// configure linux personality before starting the process
+	if c.config.Personality != nil {
+		err := system.SetLinuxPersonality(c.config.Personality.Domain)
```
kolyshkin, line 235: I'm still unsure this is the best place. @cyphar can you please take a look?

flouthoc: Since we want to set personality as soon as process starts so this seems the best place to me.

kolyshkin: Well, here you set it _before_ the process starts, not _as soon as_.

flouthoc: sorry for confusion i meant this should happen just before the process start or as soon as the process start so execution happens in configured execution domain. But i am cool if you think there is any better place for this call.

cyphar: This should probably happen as close to execve as possible -- that would mean in `standard_init_linux.go` and `setns_init_linux.go` -- right before we apply seccomp filters probably.

libcontainer/integration/exec_test.go
```diff
@@ -137,6 +137,26 @@ func TestIPCBadPath(t *testing.T) {
...
+	}
+
+	out, _, err := runContainer(t, config, "", "/bin/sh", "-c", "uname -a")
+	ok(t, err)
+	// output must contain kernel architecture configured as i686
+	if !strings.Contains(out.Stdout.String(), "i686") {
+		t.Fatalf("expected kernel architecture i686 configured via personality")
+	}
+}
+
```
cyphar, line 159: Just for future reference -- I would prefer a bats-based integration test, but since you've already written this one, don't worry about it.

libcontainer/specconv/spec_linux.go
```diff
@@ -391,6 +396,21 @@ func convertSecToUSec(value dbus.Variant) (dbus.Variant, error) {
 	return dbus.MakeVariant(sec), nil
 }
 
+// getLinuxPersonalityFromStr converts the string domain received from spec to equivalent integer.
+// check libcontainer/configs/config_linux.go
+// check https://raw.githubusercontent.com/torvalds/linux/master/include/uapi/linux/personality.h
```
cyphar, line 401: You don't need to add these references each time.

cyphar, line 401: You don't need to add these references each time.

libcontainer/specconv/spec_linux.go
```diff
@@ -391,6 +396,21 @@ func convertSecToUSec(value dbus.Variant) (dbus.Variant, error) {
...
+// 	LINUX i.e PER_LINUX (since Linux 1.2.0).
+//   	LINUX32 i.e PER_LINUX32 (since Linux 2.2) LINUX32 will set the uname system call to show a 32 bit CPU type, such as i686.
+// 	everything unknown default to PER_LINUX.
+func getLinuxPersonalityFromStr(domain string) int {
+	// defaults to PER_LINUX
+	if domain == "LINUX32" {
+		return configs.PER_LINUX32
+	}
+	return configs.PER_LINUX
+}
```
cyphar, line 412: I don't think we should ignore strange settings -- this is just going to cause nightmares for backwards compatibility. Since `getLinuxPersonalityFromStr` will only be called if a user has *explicitly specified* a personality, we should definitely return an error if the personality name is meaningless.

libcontainer/integration/exec_test.go
cyphar, line 159: Just for future reference -- I would prefer a bats-based integration test, but since you've already written this one, don't worry about it.

libcontainer/container_linux.go
cyphar: I'm not super familiar with personalities to be honest, my guess is that we would want to do this as close to execve as possible since I don't know what effect setting the personality will have to runc's internal code. As such, I would suggest this should probably happen as close to execve as possible -- that would mean in `standard_init_linux.go` and `setns_init_linux.go` -- right before we apply seccomp filters probably.

@flouthoc When talking about the "process start" -- this code is being run very early in container setup. Do you mean that we need to set this way before the users' code has even started executing (I would've thought we only need to set it right before we execve the users' binary)?

cyphar (changes requested):

libcontainer/container_linux.go
flouthoc: @cyphar I actually tried it in `standard_init_linux.go` a while ago for some reasons I don't exactly recall i was having some issues, something something on the lines of goroutines and because of it personality tests were failing. I'll try it again and post insights thanks.

libcontainer/integration/exec_test.go
tianon: As-is, shouldn't this also verify the host architecture before assuming that `linux32` leads to `i686`?  (Perhaps an explicit `t.Skip` for non-x86 architectures?)

AkihiroSuda:
@flouthoc Could you update the PR?

kolyshkin:
@flouthoc do you still want to work on this one? If yes, can you please rebase?

Zheaoli:
@kolyshkin @AkihiroSuda @cyphar Hello guys, this PR seems been deprecated for a while. may I take handle of it and carry it?

AkihiroSuda:
> @kolyshkin @AkihiroSuda @cyphar Hello guys, this PR seems been deprecated for a while. may I take handle of it and carry it?

Yes, please

flouthoc:
Closing this PR in favor of above discussion. @Zheaoli Could you please open a new PR for this. Thanks.

Zheaoli:
> Closing this PR in favor of above discussion. @Zheaoli Could you please open a new PR for this, feel free to copy whatever code is needed as it is or change parts as per needs i'm not sure if its code in this PR is updated as per recent runc version but hope it works.

Of course, thanks for your work!

AkihiroSuda:
@Zheaoli Do you plan to open a new PR?

Zheaoli:
> @Zheaoli Do you plan to open a new PR?
> 
> 

Yes I will open a new PR this weekend

kolyshkin:
The carry of this PR (#4082) was included into runc v1.2.0-rc.1. I'm not sure why this one had v1.3.0 milestone but it was obviously wrong, so I removed it.

PR #3133, opened 2021-08-05
libct/cg: GetAllPids: optimize for go 1.16+ (3x to 3.5x speed improvement)

kolyshkin:
### I. libct/cg: improve GetAllPids and readProcsFile

* GetAllPids: skip non-directories, remove filename comparison.
* readProcsFile: accept directory, use own OpenFile.

### II. libct/cg: GetAllPids: optimize for go 1.16+

The `filepath.WalkDir` function, introduced in Go 1.16, doesn't do `stat(2)`
on every entry, and is therefore somewhat faster (see below).

Since we have to support Go 1.15, keep the old version for backward
compatibility.

Add a quick benchmark, which shows approximately 3x improvement:

        $ go1.15.15 test -bench AllPid -run xxx .
        BenchmarkGetAllPids-4                 48          23528839 ns/op

        $ go version
        go version go1.16.6 linux/amd64
        $ go test -bench AllPid -run xxx .
        BenchmarkGetAllPids-4                147           7700170 ns/op

(Unrelated but worth noting -- go 1.17rc2 is pushing it even further)

        $ go1.17rc2 test -bench AllPid -run xxx .
        BenchmarkGetAllPids-4                164           6820994 ns/op

So, 3.5x faster with go 1.17 but still compatible with go 1.15.

### DIFF
--- a/libcontainer/cgroups/getallpids.go
+++ b/libcontainer/cgroups/getallpids.go
@@ -1,0 +1,30 @@
+// +build linux
+
+package cgroups
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// GetAllPids returns all pids, that were added to cgroup at path and to all its
+// subcgroups.
+func GetAllPids(path string) ([]int, error) {
+	var pids []int
+	// collect pids from all sub-cgroups
+	err := filepath.Walk(path, func(p string, info os.FileInfo, iErr error) error {
+		if iErr != nil {
+			return iErr
+		}
+		if info.IsDir() || info.Name() != CgroupProcesses {
+			return nil
+		}
+		cPids, err := readProcsFile(p)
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):

cyphar (approved):
LGTM.

cyphar:
Ah it seems that "closes" in merge commits now causes its own issues. I guess I'll need to stop doing that in my manual merge commits...

kolyshkin:
> Ah it seems that "closes" in merge commits now causes its own issues. I guess I'll need to stop doing that in my manual merge commits...

Yeah, this is something new -- previously thse "closes: " lines did not affect the PR status (which was "merged"), now it's "merged and closed" which is a bit weird.

Issue #3132, opened 2021-08-05
Multiple containers in the same cgroup


kolyshkin:
Currently runc allows multiple containers to share the same cgroup. While such shared configuration might be OK, there are some issues.

1. When each container has its own resource limits, the order of start determines which limits will be effectively applied.
2. When one of containers is paused, all others are paused, too.
3. When a container is paused, any attempt to do runc create/run/exec end up with `runc init` stuck inside a frozen cgroup.
4. When a systemd cgroup manager is used, this is not working at all -- such as, stop (or even failed start) of any container results in stop unit being sent to systemd, and so (depending on unit properties) other containers can receive SIGTERM, be killed after a timeout etc.

All this may lead to various hard-to-debug situations in production (runc init stuck, cgroup removal error, wrong resource limits etc).

I originally found issue 3 from the list above and tried to solve it in https://github.com/opencontainers/runc/pull/3131, and this is how I found issue 4.

What can be done to avoid these bad situations?
1. Require that the cgroup (systemd unit) of a container being created is not present.
2. Require that the cgroup (systemd unit) of a container being created is either not present or empty (no processes).
3. Require that the cgroup is not frozen before running `runc init` (this is what #3131 does).

Surely, these measures might break some existing usage scenarios. This is why:
* runc 1.1 will warn about non-empty cgroup for a new container (runc run/create);
* runc 1.2 will make this warning an error.

kolyshkin:
I am going to implement
* solution 2 for runc create/run (i.e. require that the cgroup of a container to be created is empty (no processes), and
* solution 3 for runc exec (i.e. require that the cgroup is not frozen).
as the least intrusive way to handle this issue.

kolyshkin:
It seems that crun is equally affected (filed https://github.com/containers/crun/issues/716).

PR #3136, opened 2021-08-08
libct/cg: rm dead code to improve clarity

kolyshkin:
This was initially added by commits 41d9d2651301a2 and 4a8f0b4db492,
apparently to implement `docker run --cgroup container:ID` (https://github.com/moby/moby/pull/19244),
which was never merged. Therefore, this code is not and was never used.

It needs to be removed mainly because having it makes it much harder to
understand how cgroup manager works (because with this in place we have
not one or two but three sets of cgroup paths to think about).

Note if the paths are known and there is a need to add a PID to existing
cgroup, cgroup manager is not needed at all -- something like
cgroups.WriteCgroupProc or cgroups.EnterPid is sufficient (and the
latter is what runc exec uses in (*setnsProcess).start). If, for some reason,
the cgroup manager is still needed, it's better to write a separate shallow
implementation that would only do WriteCgroupProc/EnterPid in Apply and
have other methods empty.

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -196,26 +196,12 @@ func (m *manager) Apply(pid int) (err error) {
 		return cgroups.ErrV1NoUnified
 	}
 
-	m.paths = make(map[string]string)
-	if c.Paths != nil {
-		cgMap, err := cgroups.ParseCgroupFile("/proc/self/cgroup")
-		if err != nil {
-			return err
-		}
-		for name, path := range c.Paths {
-			// XXX(kolyshkin@): why this check is needed?
-			if _, ok := cgMap[name]; ok {
-				m.paths[name] = path
-			}
-		}
-		return cgroups.EnterPid(m.paths, pid)
-	}
-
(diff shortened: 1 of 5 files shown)


mrunalp (approved):
Makes sense to remove :+1:

kolyshkin:

kolyshkin:
@opencontainers/runc-maintainers PTAL. This removes the code that's not used but makes a huge roadblock preventing cgroup manager code understanding (at least this is the way it was for me), so its removal is quite important (if we want more people to work on runc).

hqhq (approved):
LGTM.

I remembered `--cgroup:container:<id>` was a useful feature needed by some projects, I'm surprised it's never merged. 😂

kolyshkin:
> I remembered `--cgroup:container:<id>` was a useful feature needed by some projects, I'm surprised it's never merged.

@hqhq  In terms of runc and/or OCI spec, specifying `cgroupPath` is sufficient for a container to share a cgroup with another container -- no other changes are needed.

OTOH doing this creates some other problems -- see e.g. https://github.com/opencontainers/runc/issues/3132

PR #3143, opened 2021-08-10
libct/cg/sd/v1: Fix unnecessary freeze/thaw

odinuge:
My testing shows that https://github.com/opencontainers/runc/pull/3082 (commit https://github.com/opencontainers/runc/pull/3082/commits/f2db87986cab5a7c98b8d9b875bb54e0e7a44219) doesn't work as expected. This causes big issues in k8s.

PTAL @kolyshkin :smile: 

Finally back again after having some time off, so haven't had time to look at it before now.

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -311,6 +311,14 @@ func getUnitName(c *configs.Cgroup) string {
 	return c.Name
 }
 
+// This code should be in sync with getUnitName.
+func getUnitType(unitName string) string {
+	if strings.HasSuffix(unitName, ".slice") {
+		return "Slice"
+	}
+	return "Scope"
+}
+
 // isDbusError returns true if the error is a specific dbus error.
 func isDbusError(err error, name string) bool {
 	if err != nil {
@@ -389,10 +397,10 @@ func resetFailedUnit(cm *dbusConnManager, name string) {
 	}
 }
 
-func getUnitProperty(cm *dbusConnManager, unitName string, propertyName string) (*systemdDbus.Property, error) {
+func getUnitTypeProperty(cm *dbusConnManager, unitName string, unitType string, propertyName string) (*systemdDbus.Property, error) {
 	var prop *systemdDbus.Property
 	err := cm.retryOnDisconnect(func(c *systemdDbus.Conn) (Err error) {
-		prop, Err = c.GetUnitPropertyContext(context.TODO(), unitName, propertyName)
+		prop, Err = c.GetUnitTypePropertyContext(context.TODO(), unitName, unitType, propertyName)
 		return Err
 	})
 	return prop, err
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"sync"
 
@@ -353,15 +354,20 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
 		// a non-existent unit returns default properties,
 		// and settings in (2) are the defaults.
 		//
-		// Do not return errors from getUnitProperty, as they alone
+		// Do not return errors from getUnitTypeProperty, as they alone
 		// should not prevent Set from working.
-		devPolicy, e := getUnitProperty(m.dbus, unitName, "DevicePolicy")
+
+		unitType := getUnitType(unitName)
+
+		devPolicy, e := getUnitTypeProperty(m.dbus, unitName, unitType, "DevicePolicy")
 		if e == nil && devPolicy.Value == dbus.MakeVariant("auto") {
-			devAllow, e := getUnitProperty(m.dbus, unitName, "DeviceAllow")
-			if e == nil && devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) {
-				needsFreeze = false
-				needsThaw = false
-				return
+			devAllow, e := getUnitTypeProperty(m.dbus, unitName, unitType, "DeviceAllow")
+			if e == nil {
+				if rv := reflect.ValueOf(devAllow.Value.Value()); rv.Kind() == reflect.Slice && rv.Len() == 0 {
+					needsFreeze = false
+					needsThaw = false
+					return
+				}
 			}
 		}
 	}
--- a/libcontainer/cgroups/systemd/systemd_test.go
+++ b/libcontainer/cgroups/systemd/systemd_test.go
@@ -40,6 +40,23 @@ func TestSystemdVersion(t *testing.T) {
 	}
 }
 
+func TestValidUnitTypes(t *testing.T) {
+	testCases := []struct {
+		unitName         string
+		expectedUnitType string
+	}{
+		{"system.slice", "Slice"},
+		{"kubepods.slice", "Slice"},
+		{"testing-container:ab.scope", "Scope"},
+	}
+	for _, sdTest := range testCases {
+		unitType := getUnitType(sdTest.unitName)
+		if unitType != sdTest.expectedUnitType {
+			t.Errorf("getUnitType(%s); want %q; got %q", sdTest.unitName, sdTest.expectedUnitType, unitType)
+		}
+	}
+}
+
 func newManager(config *configs.Cgroup) cgroups.Manager {
 	if cgroups.IsCgroup2UnifiedMode() {
 		return NewUnifiedManager(config, "", false)
--- a/libcontainer/cgroups/systemd/v1_test.go
+++ b/libcontainer/cgroups/systemd/v1_test.go
@@ -1,0 +1,217 @@
+package systemd
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+func TestFreezeBeforeSet(t *testing.T) {
+	requireV1(t)
+
+	testCases := []struct {
+		desc string
+		// Test input.
+		cg        *configs.Cgroup
+		preFreeze bool
+		// Expected values.
+		// Before unit creation (Apply).
+		freeze0, thaw0 bool
+		// After unit creation.
+		freeze1, thaw1 bool
+	}{
+		{
+			// A slice with SkipDevices.
+			desc: "slice,skip-devices",
+			cg: &configs.Cgroup{
+				Name:   "system-runc_test_freeze_1.slice",
+				Parent: "system.slice",
+				Resources: &configs.Resources{
+					SkipDevices: true,
+				},
+			},
+			// Expected.
+			freeze0: false,
+			thaw0:   false,
+			freeze1: false,
+			thaw1:   false,
+		},
+		{
+			// A scope with SkipDevices. Not a realistic scenario with runc
+			// (as container can't have SkipDevices == true), but possible
+			// for a standalone cgroup manager.
+			desc: "scope,skip-devices",
+			cg: &configs.Cgroup{
+				ScopePrefix: "test",
+				Name:        "testFreeze2",
+				Parent:      "system.slice",
+				Resources: &configs.Resources{
+					SkipDevices: true,
+				},
+			},
+			// Expected.
+			freeze0: false,
+			thaw0:   false,
+			freeze1: false,
+			thaw1:   false,
+		},
+		{
+			// A slice that is about to be frozen in Set.
+			desc: "slice,will-freeze",
+			cg: &configs.Cgroup{
+				Name:   "system-runc_test_freeze_3.slice",
+				Parent: "system.slice",
+				Resources: &configs.Resources{
+					Freezer: configs.Frozen,
+				},
+			},
+			// Expected.
+			freeze0: true,
+			thaw0:   false,
+			freeze1: true,
+			thaw1:   false,
+		},
+		{
+			// A pre-frozen slice that should stay frozen.
+			desc: "slice,pre-frozen,will-freeze",
+			cg: &configs.Cgroup{
+				Name:   "system-runc_test_freeze_4.slice",
+				Parent: "system.slice",
+				Resources: &configs.Resources{
+					Freezer: configs.Frozen,
+				},
+			},
+			preFreeze: true,
+			// Expected.
+			freeze0: true, // not actually frozen yet.
+			thaw0:   false,
+			freeze1: false,
+			thaw1:   false,
+		},
+		{
+			// A pre-frozen scope with skip devices set.
+			desc: "scope,pre-frozen,skip-devices",
+			cg: &configs.Cgroup{
+				ScopePrefix: "test",
+				Name:        "testFreeze5",
+				Parent:      "system.slice",
+				Resources: &configs.Resources{
+					SkipDevices: true,
+				},
+			},
+			preFreeze: true,
+			// Expected.
+			freeze0: false,
+			thaw0:   false,
+			freeze1: false,
+			thaw1:   false,
+		},
+		{
+			// A pre-frozen scope which will be thawed.
+			desc: "scope,pre-frozen",
+			cg: &configs.Cgroup{
+				ScopePrefix: "test",
+				Name:        "testFreeze6",
+				Parent:      "system.slice",
+				Resources:   &configs.Resources{},
+			},
+			preFreeze: true,
+			// Expected.
+			freeze0: true, // not actually frozen yet.
+			thaw0:   true,
+			freeze1: false,
+			thaw1:   false,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.desc, func(t *testing.T) {
+			m := NewLegacyManager(tc.cg, nil)
+			defer m.Destroy() //nolint:errcheck
+			lm := m.(*legacyManager)
+
+			// Checks for a non-existent unit.
+			freeze, thaw, err := lm.freezeBeforeSet(getUnitName(tc.cg), tc.cg.Resources)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if freeze != tc.freeze0 || thaw != tc.thaw0 {
+				t.Errorf("before Apply (non-existent unit): expected freeze: %v, thaw: %v, got freeze: %v, thaw: %v",
+					tc.freeze0, tc.thaw0, freeze, thaw)
+			}
+
+			// Create systemd unit.
+			pid := -1
+			if strings.HasSuffix(getUnitName(tc.cg), ".scope") {
+				// Scopes require a process inside.
+				cmd := exec.Command("bash", "-c", "sleep 1m")
+				if err := cmd.Start(); err != nil {
+					t.Fatal(err)
+				}
+				pid = cmd.Process.Pid
+				// Make sure to not leave a zombie.
+				defer func() {
+					// These may fail, we don't care.
+					_ = cmd.Process.Kill()
+					_ = cmd.Wait()
+				}()
+			}
+			if err := m.Apply(pid); err != nil {
+				t.Fatal(err)
+			}
+			if tc.preFreeze {
+				if err := m.Freeze(configs.Frozen); err != nil {
+					t.Error(err)
+					return // no more checks
+				}
+			}
+			freeze, thaw, err = lm.freezeBeforeSet(getUnitName(tc.cg), tc.cg.Resources)
+			if err != nil {
+				t.Error(err)
+				return // no more checks
+			}
+			if freeze != tc.freeze1 || thaw != tc.thaw1 {
+				t.Errorf("expected freeze: %v, thaw: %v, got freeze: %v, thaw: %v",
+					tc.freeze1, tc.thaw1, freeze, thaw)
+			}
+			// Destroy() timeouts on a frozen container, so we need to thaw it.
+			if tc.preFreeze {
+				if err := m.Freeze(configs.Thawed); err != nil {
+					t.Error(err)
+				}
+			}
+			// Destroy() does not kill processes in cgroup, so we should.
+			if pid != -1 {
+				if err = unix.Kill(pid, unix.SIGKILL); err != nil {
+					t.Errorf("unable to kill pid %d: %s", pid, err)
+				}
+			}
+			// Not really needed, but may help catch some bugs.
+			if err := m.Destroy(); err != nil {
+				t.Errorf("destroy: %s", err)
+			}
+		})
+	}
+}
+
+// requireV1 skips the test unless a set of requirements (cgroup v1,
+// systemd, root) is met.
+func requireV1(t *testing.T) {
+	t.Helper()
+	if cgroups.IsCgroup2UnifiedMode() {
+		t.Skip("Test requires cgroup v1.")
+	}
+	if !IsRunningSystemd() {
+		t.Skip("Test requires systemd.")
+	}
+	if os.Geteuid() != 0 {
+		t.Skip("Test requires root.")
+	}
+}


libcontainer/cgroups/systemd/common.go
```diff
@@ -518,3 +522,19 @@ func addCpuset(cm *dbusConnManager, props *[]systemdDbus.Property, cpus, mems st
 	}
 	return nil
 }
+
+func getUnitType(unitName string) string {
```
odinuge, line 526: Placeholder. Can probably do something more smooth. This is just so we can start discussing the best way to solve it.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -355,13 +356,20 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
...
+		devPolicy, e := getUnitTypeProperty(m.dbus, unitName, getUnitType(unitName), "DevicePolicy")
 		if e == nil && devPolicy.Value == dbus.MakeVariant("auto") {
-			devAllow, e := getUnitProperty(m.dbus, unitName, "DeviceAllow")
-			if e == nil && devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) {
-				needsFreeze = false
-				needsThaw = false
-				return
+			devAllow, e := getUnitTypeProperty(m.dbus, unitName, getUnitType(unitName), "DeviceAllow")
+			if e == nil {
+				rv := reflect.ValueOf(devAllow.Value.Value())
```
kolyshkin, line 363: Do you mean that `devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{})` did not work? Or that your version is more readable? (or both?)

libcontainer/cgroups/systemd/common.go
kolyshkin: We can probably just use `strings.ToTitle()` here.

libcontainer/cgroups/systemd/v1.go
odinuge: > Do you mean that devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) did not work? Or that your version is more readable? (or both?)

Well, it doesn't seen to work for me. Does that for you, in that case there is something wrong with my setup? :thinking:

libcontainer/cgroups/systemd/common.go
odinuge: Yeah, was thinking about that. Or should we have an allowlist like this, to ensure we don't forward "bad" strings? Doesn't matter that much to me

odinuge:
```go
validStuff := map[string]bool {
     "Slice": true
     "Scope": true
 }
  
 if validStuff[strings.ToTitle(type)] {
     return type
 }
 return err
```

Can probably do something like this to validate it.

kolyshkin:
I guess what we need first is a unit test that would create a unit (of various types) with a particular property, get the property, and check it's as expected.

Last commit of #3082 is an integration test (`TestPodSkipDevicesUpdate`) that checks that the freeze is not done. I look at all this and I still can't understand why that test does not catch the issue this PR is fixing...

odinuge:
>I guess what we need first is a unit test that would create a unit (of various types) with a particular property, get the property, and check it's as expected.

Yeah, some tests would be ok I guess. It is fairly hard to test tho. But keep reading `freezer.state` on the slice while constantly updating, and looking for only THAWED would work. I can look into doing it in go, but the bats stuff is a bit complicated tho., so that would require some boiler plate.

> Last commit of #3082 is an integration test (TestPodSkipDevicesUpdate) that checks that the freeze is not done. I look at all this and I still can't understand why that test does not catch the issue this PR is fixing...

Well, it only verifies that there is no errors with the devices. It does however not check if the slice is frozen or not.


To test this, one can simply verify like this - keep the loop running while running tests with `sudo go test -run=TestPodSkipDevicesUpdate -v -test.count=5`:

```
# in this branch
$ while true ; do grep FROZEN -R /sys/fs/cgroup/freezer/system.slice/; done
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN
# in master 
$ while true ; do grep FROZEN -R /sys/fs/cgroup/freezer/system.slice/; done             
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/freezer.state:FROZEN                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                                                     
/sys/fs/cgroup/freezer/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/freezer.state:FROZEN                        
```

As seen here, only the scope/container is frozen with this branch, while in master both the pod/slice and the container/scope is frozen. Does that make sense to you @kolyshkin ?

odinuge:
Hmm. Is something wrong with CI, or?

kolyshkin:
> Hmm. Is something wrong with CI, or?

Yup; see https://githubstatus.com/.

Seems over now, do a force push to trigger CI.

kolyshkin:
or we can close/open to re-kick CI (usually it works)

AkihiroSuda:
Does this fix https://github.com/kubernetes/kubernetes/issues/104280 ?

odinuge:
> @AkihiroSuda: Does this fix kubernetes/kubernetes#104280 ?

Yes.

But as I have said a few times already, for k8s a flag for disabling this would be better and simpler. I guess we (in k8s, not here in runc) have to discuss if we should avoid using `runc/libcontainer` for setting cgroup things, since things like this will increase the overhead waaay more than necessary.

libcontainer/cgroups/systemd/common.go
kolyshkin: We can either have `.slice` or `.scope` (see `getUnitName()`).

I suggest we add `getUnitType` which is very similar to `getUnitName`.

libcontainer/cgroups/systemd/v1.go
kolyshkin: I think I improved the readability a notch in https://github.com/opencontainers/runc/pull/3148:

```go
                        devAllow, e := getUnitTypeProperty(m.dbus, unitName, unitType, "DeviceAllow")
                          if e == nil {
                                  if rv := reflect.ValueOf(devAllow.Value.Value()); rv.Kind() == reflect.Slice && rv.Len() == 0 {
                                          needsFreeze = false
                                          needsThaw = false
                                          return
                                  }
                          }
```

kolyshkin:
@odinuge I was playing with the test case and ended up doing https://github.com/opencontainers/runc/pull/3148.

The other thing I did is slightly simplified the deviceAllow check (https://github.com/opencontainers/runc/pull/3143#discussion_r687222469)

We can either use #3148, or cherry-pick the test case from it to here.

libcontainer/cgroups/systemd/v1.go
odinuge: Guess I can do that instead if the others agree, even tho. the line is a bit too long. I'll update the PR.

libcontainer/cgroups/systemd/common.go
```diff
@@ -518,3 +518,10 @@ func addCpuset(cm *dbusConnManager, props *[]systemdDbus.Property, cpus, mems st
...
```
kolyshkin, line 522: Can you please add something like
```go
	// This code should be in sync with getUnitName.
```

odinuge:
> We can either use #3148, or cherry-pick the test case from it to here.

I can cherry-pick the test commit here if they should be in the same PR. I prefer to keep this PR, otherwise I cannot update and change the PR (you created based of this), making it a bit cumbersome to work on it. Having a separate PR with the test works as well.

libcontainer/cgroups/systemd/common.go
odinuge: Ack 👍

kolyshkin:
> I can cherry-pick the test commit here if they should be in the same PR

Please do (otherwise the CI is semi-useless).

> making it a bit cumbersome to work on it

Something like `git-rebase -i origin/master` is a life saver when working on PRs with multiple commits. In most complicated cases I do `git stash; git format-patch -o <dir> origin/master; git reset --hard origin/master`, when `git am` those commits one by one, making changes and doing  `git commit --amend` when necessary.

libcontainer/cgroups/systemd/v1.go
kolyshkin: What about this:
```go
                        if e == nil { 
                                rv := reflect.ValueOf(devAllow.Value.Value())
                                if rv.Kind() == reflect.Slice && rv.Len() == 0 { 
```

kolyshkin: To be it reads slightly better than the switch/case (although I agree that either version we came with so far is not super readable).

kolyshkin:
Once ready and merged, we'll have a backport to 1.0 and will try 1.0.2 released ASAP.

kolyshkin:
> things like this will increase the overhead waaay more than necessary

@odinuge your concern about this code path being slow is valid, it is also quite complicated. I was looking into how to make it fast and simple -- PTAL https://github.com/opencontainers/runc/pull/3151.

Surely a simple "SkipFreeze" flag will be faster, but this seems lightweight enough to not cause performance concerns.

odinuge:
> @odinuge your concern about this code path being slow is valid, it is also quite complicated. I was looking into how to make it fast and simple -- PTAL #3151.
>
> Surely a simple "SkipFreeze" flag will be faster, but this seems lightweight enough to not cause performance concerns.

Yeah, #3151 would probably work, but I don't like going around systemd like that approach does. Systemd should be the source of the truth in this case, as it is now.

(agian, I am mostly a fan of a flag for k8s, otherwise we have to look for alternatives. k8s do these calls very often, and with an increasing amount of pods, this will increase further. Same for shortlived containers. A flag as `SkipDevices` was previously is definitely the best approach for k8s; whether we have to use runc or implement it ourselves.)

> Something like git-rebase -i origin/master is a life saver when working on PRs with multiple commits. In most complicated cases I do git stash; git format-patch -o <dir> origin/master; git reset --hard origin/master, when git am those commits one by one, making changes and doing git commit --amend when necessary.

`git` is always the easy part. GitHub is the hard one, since I don't have push or edit access on your PR, only mine.

harche:
/hold 

I am running node e2e + systemd + crio tests locally using the runc binary built from this PR. The first run was successful, but I will have more runs to verify that this PR indeed fixes https://github.com/kubernetes/kubernetes/issues/104280#issuecomment-897093892

odinuge:
> I am running node e2e + systemd + crio tests locally using the runc binary built from this PR. The first run was successful, but I will have more runs to verify that this PR indeed fixes kubernetes/kubernetes#104280 (comment)

It is the vendoring of runc into k8s that is important for fixing kubernetes/kubernetes#104280, not only the runc binary used by the CRI runtime.

harche:
> > I am running node e2e + systemd + crio tests locally using the runc binary built from this PR. The first run was successful, but I will have more runs to verify that this PR indeed fixes [kubernetes/kubernetes#104280](https://github.com/kubernetes/kubernetes/issues/104280) (comment)
> 
> It is the vendoring of runc into k8s that is important for fixing [kubernetes/kubernetes#104280](https://github.com/kubernetes/kubernetes/issues/104280), not only the runc binary used by the CRI runtime.

True, but the failures I was seeing in CI were specifically coming from runc binary indicating that CRIO was able to start the container but it ended up in paused state. 

/hold cancel

odinuge:
> True, but the failures I was seeing in CI were specifically coming from runc binary indicating that CRIO was able to start the container but it ended up in paused state.

Yes, but that is due to kubelet (via this code) freezing `kubepods.slice`, `burstable.slice`, and `pod-xyz.slice`. Most of the issues in regards to the binary should already be fixed in runc `v1.0.1`.

harche:
> > True, but the failures I was seeing in CI were specifically coming from runc binary indicating that CRIO was able to start the container but it ended up in paused state.
> 
> Yes, but that is due to kubelet (via this code) freezing `kubepods.slice`, `burstable.slice`, and `pod-xyz.slice`. Most of the issues in regards to the binary should already be fixed in runc `v1.0.1`.

Oh ok, that means the success I am seeing in node e2e (2 out of 2 runs so far) is due to me updating runc from rc95 to your latest PR (which has v1.0.1).

libcontainer/cgroups/systemd/v1.go
```diff
@@ -353,15 +354,20 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
 		// a non-existent unit returns default properties,
 		// and settings in (2) are the defaults.
```
kolyshkin, line 355: The tests in this PR do not check whether this is true. I have added those checks last night, plus another test case -- @odinuge can you please re-cherry-pick the last commit from https://github.com/kolyshkin/runc/tree/fix-freeze-before-set

kolyshkin:
> Yeah, #3151 would probably work, but I don't like going around systemd like that approach does. Systemd should be the source of the truth in this case, as it is now.

So, we have to admit that this is a kludge already. We need a way to figure out whether systemd is going do a disruptive device rules update. Any way to figure that out is fine as long as it works. Checking unit properties is one way, checking devices.list is another. Sure, #3151 has an assumption that devices.list in the kernel and systemd Device* properties are in sync. I think such assumption is valid (if not, this would mean systemd is not doing its job, ignoring device properties updates or some such).

> (agian, I am mostly a fan of a flag for k8s, otherwise we have to look for alternatives. k8s do these calls very often, and with an increasing amount of pods, this will increase further. Same for shortlived containers. A flag as `SkipDevices` was previously is definitely the best approach for k8s; whether we have to use runc or implement it ourselves.)

Yes, this is also a way to figure out whether a freeze is needed or not, and it's fine as long as it works. The only thing I don't like about it is it requires an additional knob to set. Generally speaking, if there is a way to achieve a goal without adding an explicit knob, and it works reliably and fast, I think such way is better. Currently I see #3151 as such a way.

odinuge:
> So, we have to admit that this is a kludge already. We need a way to figure out whether systemd is going do a disruptive device rules update. Any way to figure that out is fine as long as it works. Checking unit properties is one way, checking devices.list is another. Sure, #3151 has an assumption that devices.list in the kernel and systemd Device* properties are in sync. I think such assumption is valid (if not, this would mean systemd is not doing its job, ignoring device properties updates or some such).

Yeah, that solution does also work, and yeah, this is all pretty kludgy, I agree about that :sweat_smile: k8s and the runc binary also work differently, with different purposes, making it even more kludgy. But yeah, then we instead can just look at the provided resources, and if the devices allow access to all, then drop the freeze. If we don't rely on systemd, and don't want a separate knob, I think checking devices would be better and simpler to understand, together with better performance.

Eg something like:

```go
oldDevices :=  []specs.LinuxDeviceCgroup{
  {
    Allow:  true,
    Access: "rwm",
  },
}

newDevices :=  []specs.LinuxDeviceCgroup{
  {
    Allow:  true,
    Access: "rwm",
  },
}


if allowsAll(oldDevices) && allowsAll(newDevices) {
  // Skip freezing
}
```

(although we might need to poke systemd or cgropufs in `Apply` to get state in the beginning, in slice/scope is already alive)

Any thoughts about that instead? Wouldn't that work for `runc update` as well, as long as people only manage the slice/scope via runc and not set devices directly in systemd? Or do you think that is too fragile?

> Yes, this is also a way to figure out whether a freeze is needed or not, and it's fine as long as it works. The only thing I don't like about it is it requires an additional knob to set. Generally speaking, if there is a way to achieve a goal without adding an explicit knob, and it works reliably and fast, I think such way is better. Currently I see #3151 as such a way.

The `SkipDevices` is _already_ a knob, and it worked properly for k8s until it was adapted for `runc update`. I don't really see an issue with it being a knob in the same way as `SkipDevices` is, but I guess that is another discussion. :sweat_smile:

libcontainer/cgroups/systemd/v1.go
odinuge: Ack :smile: Test looks good!

odinuge:
We can also change this to a single dbus call by using `GetUnitTypePropertiesContext` on the slice/scope, since both `DeviceAllow` and `DevicePolicy` are on the same interface.

kolyshkin:
> We can also change this to a single dbus call by using GetUnitTypePropertiesContext on the slice/scope, since both DeviceAllow and DevicePolicy are on the same interface.

It's an interesting idea, which I also considered but gave up on that since this returns all (tens of, I guess) properties in a `map[string]interface{}`, of which we only need two. This results in more garbage to collect by the GC.

OTOH if the call itself is relatively heavy, the cost of the saved call may outweigh the GC overhead, and this might be worth doing.

I guess the way to check which way is better is to benchmark both implementations, using runtime.GC() inside the benchmarked code to make sure GC overhead is accounted for. IOW something like

```go
func BenchmarkXXX(b *testing.B) {
     // ... any prep code
     runtime.GC()
     b.ResetTimer()
     for i := 0; i < b.N, i++ {
          // ... whatever we test
     }
     runtime.GC()
}
```

kolyshkin:
> ```go
> if allowsAll(oldDevices) && allowsAll(newDevices) {
>   // Skip freezing
> }
> ```
>
> (although we might need to poke systemd or cgropufs in `Apply` to get state in the beginning, in slice/scope is already alive)
> 
> Any thoughts about that instead? Wouldn't that work for `runc update` as well, as long as people only manage the slice/scope via runc and not set devices directly in systemd? Or do you think that is too fragile?

Well, this is conceptually the same that we do here and/or in #3151. In terms of #3151, `allowsAll(oldDevices)` is equivalent to `!hasDeviceRules()`, and `allowsAll(newDevices)` is equivalent to `r.SkipDevices`.

kolyshkin:
> `git` is always the easy part. GitHub is the hard one, since I don't have push or edit access on your PR, only mine.

Ha! This is easy peasy  -- become a runc maintainer, and you'll be able to shove your commits into contributors' PRs 🕺🏻 

On a serious note, though, I think I misunderstood you, thought you were against cherry-picking a commit. 😀

kolyshkin:
OK I ended up writing this benchmark for `freezeBeforeSet()` (based on the test code):

Here are the results (first line is this PR, second is #3151):
```
BenchmarkFreezeBeforeSet-8   	    3757	    286971 ns/op	   16632 B/op	     680 allocs/op
BenchmarkFreezeBeforeSet-8   	  131257	      9116 ns/op	    2344 B/op	       7 allocs/op
```

To have some reference points, I did a few quick benchmarks

1. start/stop a systemd scope:
```
BenchmarkApplyDestroy-8      	     808	   1390820 ns/op	  172188 B/op	    2130 allocs/op
``` 

2. Set() (with no freeze checking or freeze code -- as if SkipFreeze flag is provided)
```
BenchmarkSet-8               	    6412	    183464 ns/op	    7266 B/op	     283 allocs/op
```

3. Set() (with unconditional freeze, no processes in cgroup!):
```
BenchmarkSet-8               	    5127	    231167 ns/op	   10224 B/op	     298 allocs/op
```

From that we can estimate Set() overhead for kubelet case

* with current approach (#3082 (commit f2db879) + #3143): slowdown from 200 to 500 us (150% worse)
* with devices.list approach (#3151): slowdown from 200 to 210 us (5% worse)

kolyshkin:
Ah, here's the benchmarking code: https://gist.github.com/kolyshkin/20e8f43fedde4989033941be8bef9fc6

It needs some tweaks to measure Set(), I was using these placeholders for Set measurements:

```go
func (m *legacyManager) freezeBeforeSetNo(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, err error) {
        return false, false, nil
}

func (m *legacyManager) freezeBeforeSetYes(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, err error) {
        return true, true, nil
}
```

odinuge:
> With current approach (#3082 (commit f2db879) + #3143): slowdown from 200 to 500 us (150% worse).
>
> With devices.list approach (#3151): slowdown from 200 to 210 us (5% worse)

Nice. I guess that is more or less predicted. Freezing with a lot of procs, eg. 110 pods on 100 logical Linux CPUs, I assume the overhead will be quite noticeable, and probably quite huge (so people don't infer that unconditional freeze works ok).

> Well, this is conceptually the same that we do here and/or in #3151. In terms of #3151, allowsAll(oldDevices) is equivalent to !hasDeviceRules(), and allowsAll(newDevices) is equivalent to r.SkipDevices.

Well, looking at it that way all the solutions are conceptually the same. But if we only look at the oci-data to see if we can avoid freezing, without any external calls. `allowsAll(oldDevices)` would then just look at the old value, instead of checking with systemd or looking at cgroupfs to verify it.

kolyshkin:
> Well, looking at it that way all the solutions are conceptually the same. But if we only look at the oci-data to see if we can avoid freezing, without any external calls. `allowsAll(oldDevices)` would then just look at the old value, instead of checking with systemd or looking at cgroupfs to verify it.

Problem is, runc cgroup manager do not have persistent storage, and kubelet instantiates a new cgroup manager to call Set():

https://github.com/kubernetes/kubernetes/blob/b6d83f0ba3f155a4f6af9b37bf5511f66327cd5b/pkg/kubelet/cm/cgroup_manager_linux.go#L487

odinuge:
> Problem is, runc cgroup manager do not have persistent storage, and kubelet instantiates a new cgroup manager to call Set():
>
> https://github.com/kubernetes/kubernetes/blob/b6d83f0ba3f155a4f6af9b37bf5511f66327cd5b/pkg/kubelet/cm/cgroup_manager_linux.go#L487

Yeah, but we do supply cgroup config when creating the manager. But yeah, if the only way to avoid this "freeze dance" is for k8s to keep the manager around, I guess we can do that; but that will require some more code.

(not familiar with how the runc binary works in detail btw.)

kolyshkin:
> not familiar with how the runc binary works in detail btw.

It's (usually) a short-lived binary, so we also instantiate a new manager for e.g. `runc update` or `runc destroy`. runc has a state file though (which I plan to split into two parts), so it "knows" the current state (unless someone uses other tools to change it, that is).

kolyshkin:
> Yeah, but we do supply cgroup config when creating the manager.

You can supply anything, meaning this solution would be a skip-freeze flag in disguise.

Oh well, computers are tough. 😩 

https://github.com/opencontainers/runc/pull/3161

odinuge:
> You can supply anything, meaning this solution would be a skip-freeze flag in disguise.

And that is exactly what I want! :smile: 

> Oh well, computers are tough.

Yeee, they are...

> #3161

Ack. +1 for that one

kolyshkin:
Superceded by (already merged) https://github.com/opencontainers/runc/pull/3166 (which includes commits from this PR), thus closing. Thanks @odinuge for figuring this out!

PR #3144, opened 2021-08-11
Fix codespell warnings, add codespell to ci

kolyshkin:
Fix a bunch of warnings from codespell, and add a CI job to validate to automate it.

Note that
1. I had to use pip to install codespell, as the version provided with Ubuntu 20.04 is way old.
2. The two exceptions I had to add to codespellrc are:
 - `CLOS` (a term used by intelrtd);
 - `creat `(syscall name used in tests/integration/testdata/seccomp_*.json).

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -30,6 +30,16 @@ jobs:
       - name: compile with no build tags
         run: make BUILDTAGS=""
 
+  codespell:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: install deps
+      # Version of codespell bundled with Ubuntu is way old, so use pip.
+      run: pip install codespell
+    - name: run codespell
+      run: codespell
+
   shfmt:
     runs-on: ubuntu-20.04
     steps:
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = ./vendor,./.git
+ignore-words-list = clos,creat
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -3,7 +3,7 @@
 By default, runc creates cgroups and sets cgroup limits on its own (this mode
 is known as fs cgroup driver). When `--systemd-cgroup` global option is given
 (as in e.g. `runc --systemd-cgroup run ...`), runc switches to systemd cgroup
-driver. This document describes its features and pecularities.
+driver. This document describes its features and peculiarities.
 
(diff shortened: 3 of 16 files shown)


kolyshkin:
```
Run codespell
./.codespellrc:3: clos  ==> close
./.codespellrc:3: creat  ==> create
```

🤣

AkihiroSuda (approved):

.github/workflows/validate.yml
```diff
@@ -23,6 +23,16 @@ jobs:
           # must be specified without patch version
           version: v1.40
 
+  codespell:
```
thaJeztah, line 26: Curious; have you also tried enabling `misspell` in golang-ci-lint? Mostly wondering if it would catch the same errors (it would save us from having to run a separate linter for this); https://golangci-lint.run/usage/linters

thaJeztah: Oh, I guess `misspell` may only be looking for `.go` files 🤔

kolyshkin: That's right; as part of golangci-lint it will _probably_ only check `.go` files.

It is also
- apparently unmaintained since March 2018 (judging by its git repo at https://github.com/client9/misspell)
- can only find two misspellings in the very same codebase as this PR, when run as a standalone binary (checks all files);

```console
[kir@kir-rhat ~]$ go1.17 install github.com/client9/misspell/cmd/misspell@latest

[kir@kir-rhat runc]$ misspell * | grep -v ^vendor/
libcontainer/cgroups/ebpf/ebpf_linux.go:137:23: "succeded" is a misspelling of "succeeded"
libcontainer/cgroups/systemd/systemd_test.go:344:22: "contianer" is a misspelling of "container"
```

(the last typo it found is the one codespell can't see, but I've added the fix to this PR)

kolyshkin: In addition to the above, there's something that makes me think you'd favor codespell over it 🙃 

Once I've rebased:
```
[kir@kir-rhat runc]$ misspell * | grep -v ^vendor/
MAINTAINERS:8:0: "Sebastiaan" is a misspelling of "Sebastian"
```

kolyshkin:
Rebased to resolve a conflict after #3164 merge; PTAL @thaJeztah @AkihiroSuda

.github/workflows/validate.yml
thaJeztah: LOL!  😂

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

PR #3148, opened 2021-08-11
libct/cg/sd/v1: fix freezeBeforeSet

kolyshkin:
_this is an alternative to #3143; the biggest difference is the test case._

Commit f2db87986cab5a7c9 (PR #3082) did not do the right job getting unit
properties, and the deviceAllow property comparison was not working
either. Fix both issues.

The test case added by #3082 was not adequate either -- it only checked
that the freeze is performed if necessary, but did not check that the freeze
is avoided if possible.

Add a test case specific to freezeBeforeSet.

Fixes: f2db879
Co-authored-by: Odin Ugedal <odin@uged.al>

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -6,11 +6,11 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
-	"github.com/godbus/dbus/v5"
 	"github.com/sirupsen/logrus"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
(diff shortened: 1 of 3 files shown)


kolyshkin:
Ah, pushed the old version. Updated now.

libcontainer/cgroups/systemd/v1.go
```diff
@@ -353,15 +353,18 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
...
-		devPolicy, e := getUnitProperty(m.dbus, unitName, "DevicePolicy")
-		if e == nil && devPolicy.Value == dbus.MakeVariant("auto") {
-			devAllow, e := getUnitProperty(m.dbus, unitName, "DeviceAllow")
-			if e == nil && devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) {
-				needsFreeze = false
-				needsThaw = false
-				return
+		unitType := getUnitType(unitName)
+		devPolicy, e := getUnitTypeProperty(m.dbus, unitName, unitType, "DevicePolicy")
+		if e == nil && devPolicy.Value.String() == `"auto"` {
```
dims, line 360: interesting ... comparing with `"auto"` instead of `auto`

libcontainer/cgroups/systemd/v1_test.go
```diff
@@ -0,0 +1,172 @@
...
+				t.Errorf("expected freeze: %v, thaw: %v, got freeze: %v, thaw: %v",
+					tc.freeze, tc.thaw, freeze, thaw)
+			}
+			// Destroy() timeouts on a frozen container, so we need to thaw it.
+			if tc.preFreeze {
+				if err := m.Freeze(configs.Thawed); err != nil {
+					t.Error(err)
+				}
+			}
+			// Destroy() do not kill processes in cgroup, so we should.
```
mrunalp, line 145: nit: does not..

mrunalp (commented):
Should we move the check of the current freezer state before we do the SkipDevices check since we do that check always anyways and sometimes we could exit early avoiding the calls in SkipDevices?

mrunalp:
Left a couple of comments, otherwise looks fine.

libcontainer/cgroups/systemd/v1.go
kolyshkin: This is because `(dbus.Variant).String` uses `strconv.Quote` to format strings.

I am not quite sure whether the old or the new way to compare is better... Old one is working fine (for `devPolicy`, that is), so maybe leave it as is.

kolyshkin:
> Should we move the check of the current freezer state before we do the SkipDevices check since we do that check always anyways

Nope, we don't do the freezer state check in case SkipDevices is set and the device rules are "allow all".

> and sometimes we could exit early avoiding the calls in SkipDevices?

I was thinking about that while working on #3082 but decided against it, since a frozen cgroup is a rare case.

Wait a minute, I got a better idea!

kolyshkin:
> Wait a minute, I got a better idea!

PTAL https://github.com/opencontainers/runc/pull/3151 (faster/simpler check, even more test cases)

PR #3151, opened 2021-08-12
libct/cg/sd/v1: fix freezeBeforeSet (alt)

kolyshkin:
This is an alternative to #3148/#3143, switching to a faster/easier check in freezeBeforeSet.

Commit f2db87986cab5a7c9  (PR #3082) was not working:

1. It did not do the right job getting unit properties;
2. The check that deviceAllow is empty was not working.
    
While there is a way to fix both issues, let's employ a simpler check --
if r.SkipDevices is set, and devices.list only contains "allow all" rule,
we can safely skip freezing.
    
The only issue with the patch, in case we call freezeBeforeSet before
Apply(), the m.paths are not yet initialized, so we have to manually
figure the path to devices subsystem in haveDeviceRules. This is being
fixed elsewhere (currently in #3131 but I will split it out).
    
The test case added by #3082 was not adequate either -- it only checked
that the freeze is performed if necessary, but did not check that the freeze
is avoided if possible.

Add a test case specific to freezeBeforeSet.

Fixes: f2db879
Co-authored-by: Odin Ugedal <odin@uged.al>

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
-	"github.com/godbus/dbus/v5"
 	"github.com/sirupsen/logrus"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -332,6 +331,34 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
 	return stats, nil
 }
 
+// hasDeviceRules checks cgroup device rules. Returns true if any rules
+// other than "allow all" exist. Used by freezeBeforeSet.
+func (m *legacyManager) hasDeviceRules() bool {
+	const allowAll = "a *:* rwm\n"
+
+	path := m.Path("devices")
+	// XXX: if Apply has not been called, m.paths are not initialized, so
+	// we have to figure the path. TODO: remove once fixed.
+	if path == "" {
+		var err error
+		path, err = getSubsystemPath(m.cgroups, "devices")
+		if err != nil {
+			// Should not happen. Assume we have rules.
+			return true
+		}
+	}
+	list, err := cgroups.ReadFile(path, "devices.list")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false
+		}
+		// Should not happen. Log the error, assume we do have rules.
+		logrus.Debug(err)
+		return true
+	}
+	return list != allowAll
+}
+
 // freezeBeforeSet answers whether there is a need to freeze the cgroup before
 // applying its systemd unit properties, and thaw after, while avoiding
 // unnecessary freezer state changes.
@@ -341,29 +368,17 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
 // (unlike our fs driver, they will happily write deny-all rules to running
 // containers). So we have to freeze the container to avoid the container get
 // an occasional "permission denied" error.
-func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (needsFreeze, needsThaw bool, err error) {
-	// Special case for SkipDevices, as used by Kubernetes to create pod
-	// cgroups with allow-all device policy).
-	if r.SkipDevices {
-		// No need to freeze if SkipDevices is set, and either
-		// (1) systemd unit does not (yet) exist, or
-		// (2) it has DevicePolicy=auto and empty DeviceAllow list.
-		//
-		// Interestingly, (1) and (2) are the same here because
-		// a non-existent unit returns default properties,
-		// and settings in (2) are the defaults.
-		//
-		// Do not return errors from getUnitProperty, as they alone
-		// should not prevent Set from working.
-		devPolicy, e := getUnitProperty(m.dbus, unitName, "DevicePolicy")
-		if e == nil && devPolicy.Value == dbus.MakeVariant("auto") {
-			devAllow, e := getUnitProperty(m.dbus, unitName, "DeviceAllow")
-			if e == nil && devAllow.Value == dbus.MakeVariant([]deviceAllowEntry{}) {
-				needsFreeze = false
-				needsThaw = false
-				return
-			}
-		}
+func (m *legacyManager) freezeBeforeSet(r *configs.Resources) (needsFreeze, needsThaw bool, err error) {
+	// Special case for SkipDevices, as used by Kubernetes to create
+	// pod cgroup with allow-all device policy.
+	//
+	// SkipDevices means we won't set any Device*= properties,
+	// but systemd may still choose to re-apply existing ones,
+	// so we check that there are none.
+	if r.SkipDevices && !m.hasDeviceRules() {
+		needsFreeze = false
+		needsThaw = false
+		return
 	}
 
 	needsFreeze = true
@@ -401,8 +416,7 @@ func (m *legacyManager) Set(r *configs.Resources) error {
 		return err
 	}
 
-	unitName := getUnitName(m.cgroups)
-	needsFreeze, needsThaw, err := m.freezeBeforeSet(unitName, r)
+	needsFreeze, needsThaw, err := m.freezeBeforeSet(r)
 	if err != nil {
 		return err
 	}
@@ -413,6 +427,7 @@ func (m *legacyManager) Set(r *configs.Resources) error {
 			logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
 		}
 	}
+	unitName := getUnitName(m.cgroups)
 	setErr := setUnitProperties(m.dbus, unitName, properties...)
 	if needsThaw {
 		if err := m.doFreeze(configs.Thawed); err != nil {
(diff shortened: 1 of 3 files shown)


libcontainer/cgroups/systemd/v1.go
```diff
@@ -332,6 +331,32 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
 	return stats, nil
 }
 
+// haveDeviceRules checks cgroup device rules. Returns true if any rules
+// other than "allow all" exist. Used by freezeBeforeSet.
+func (m *legacyManager) haveDeviceRules() bool {
```
mrunalp, line 336: nit: Maybe name it `hasDeviceRules` ?

libcontainer/cgroups/systemd/v1.go
```diff
@@ -332,6 +331,32 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
...
+	list, err := cgroups.ReadFile(path, "devices.list")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false
+		}
+		// Should not happen. Log the error, assume we do have rules.
+		logrus.Debug(err)
+		return true
+	}
+	return list != "a *:* rwm\n"
```
mrunalp, line 357: nit: use a constant for the string

libcontainer/cgroups/systemd/v1.go
kolyshkin: done

libcontainer/cgroups/systemd/v1.go
kolyshkin: done

kolyshkin:
cirrus-ci failure on centos-stream-8 is a flake:
> yum install -y -q epel-release
> Error: Failed to download metadata for repo 'appstream': Yum repo downloading error: Downloading error(s): repodata/b90651eed1e1de7698def2538c508b142db41a174016665c7bffaa24e1fa0785-primary.xml.gz - Cannot download, all mirrors were already tried without success; repodata/256a38c4e18cce032edb33bba6fff7a239af1e130e2f6f0c572857748aa47ae8-filelists.xml.gz - Cannot download, all mirrors were already tried without success; repodata/86483bcfc6d853140a71ad58a16fa260b16e10334ae48153aa3e9acf96be6c46-modules.yaml.xz - Cannot download, all mirrors were already tried without success

CI restarted. If this will happen frequently, we need a workaround (something similar to e.g. commit c27b8e7fe78404ae7ad217e61510c1f0be3843c7)

libcontainer/cgroups/systemd/v1.go
```diff
@@ -332,6 +331,34 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
...
+	list, err := cgroups.ReadFile(path, "devices.list")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false
+		}
+		// Should not happen. Log the error, assume we do have rules.
+		logrus.Debug(err)
+		return true
+	}
+	return list != allowAll
```
dims, line 359: about the `\n` in `allowAll` should we use string.TrimSpace or string.Contains to essentially ignore the newline? (or is this `\n` an essential part of the check?)

kolyshkin: We can use `strings.TrimSpace` for sure, but in this particular case we're not parsing the string, we're merely checking that the entirety of file contents is this string plus a newline. As this is a kernel file and not some user-provided contents we can be sure there are no extra whitespace or some such.

mrunalp:
@odinuge @AkihiroSuda ptal

libcontainer/cgroups/systemd/v1.go
```diff
@@ -332,6 +331,34 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
...
+	// we have to figure the path. TODO: remove once fixed.
+	if path == "" {
+		var err error
+		path, err = getSubsystemPath(m.cgroups, "devices")
+		if err != nil {
+			// Should not happen. Assume we have rules.
+			return true
+		}
+	}
+	list, err := cgroups.ReadFile(path, "devices.list")
```
kolyshkin, line 350: So, this will incur some additional overhead in case devices.list contains more rules. Such case is quite common, e.g. on `runc update` (but not in k8s use case).

A way to optimize this is
```diff
@@ -347,7 +349,7 @@ func (m *legacyManager) hasDeviceRules() bool {
                        return true
                }
        }
-       list, err := cgroups.ReadFile(path, "devices.list")
+       fd, err := cgroups.OpenFile(path, "devices.list", os.O_RDONLY)
        if err != nil {
                if errors.Is(err, os.ErrNotExist) {
                        return false
@@ -356,7 +358,16 @@ func (m *legacyManager) hasDeviceRules() bool {
                logrus.Debug(err)
                return true
        }
-       return list != allowAll
+       defer fd.Close()
+
+       // Check contents matches allowAll, without reading more than needed.
+       buf := make([]byte, len(allowAll)+1) // 1 extra byte for io.EOF.
+       n, err := fd.Read(buf)
+       if n != len(allowAll) || err != io.EOF {
+               return true
+       }
+
+       return bytes.Equal(buf, []byte(allowAll))
 }
 
 // freezeBeforeSet answers whether there is a need to freeze the cgroup before
```

but it does not look very elegant. WDYT @mrunalp ?

odinuge: If we add some tests to `hasDeviceRules` where one can mock the file/input i think that would work. Would be interesting to run a benchmark to see how big the change in performance really is. :+1:

kolyshkin: I did a benchmark (with real `devices.list` files as I guess the kernel has some work to do to convert the internal rules representation to the text), and with a few files having "allow all" rules and a few files with 10-15 rules, the optimized code is 15% faster. I think this do not justify the decreased readability.

I also found 2 bugs in my proposed code (one is Read do not return io.EOF unless it's at EOF, the other is last check should be negated🤦🏻).

Based on all that, I am dropping this idea of optimizing the hell out of it.

For the reference, here's a benchmark results:
```
BenchmarkHasDeviceRulesOld-8   	   38041	     28395 ns/op
BenchmarkHasDeviceRulesNew-8   	   45121	     23767 ns/op
```

... and the benchmark itself in its entirety (you have to populate the paths yourself!):

```go
package x

import (
	"bytes"
	"errors"
	"os"
//	"fmt"
	"testing"
)

func hasDeviceRulesOld(file string) bool {
	const allowAll = "a *:* rwm\n"

	list, err := os.ReadFile(file)
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			return false
		}
		return true
	}
	return string(list) != allowAll
}

func hasDeviceRulesNew(file string) bool {
	const allowAll = "a *:* rwm\n"

	fd, err := os.Open(file)
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			return false
		}
		return true
	}
	defer fd.Close()

	// Check contents matches allowAll, without reading more than needed.
	buf := make([]byte, len(allowAll))
	n, _ := fd.Read(buf)
//	fmt.Printf("read n: %v, exp: %v, buf: %s\n", n, len(buf), buf)
	if n != len(buf) {
		return true
	}

	return !bytes.Equal(buf, []byte(allowAll))
}

func benchmarkHasDeviceRules(b *testing.B, fn func(string) bool) {
	tc := []struct {
		path string
		has  bool
	}{
		{"/sys/fs/cgroup/devices/s1/devices.list", true},
		{"/sys/fs/cgroup/devices/system.slice/systemd-logind.service/devices.list", true},
		{"/sys/fs/cgroup/devices/user.slice/devices.list", false},
	}

	for _, c := range tc {
		for i := 0; i < b.N; i++ {
			has := fn(c.path)
			if has != c.has {
				b.Fatalf("path %q: exp %v, got %v", c.path, c.has, has)
			}
		}
	}
}

func BenchmarkHasDeviceRulesOld(b *testing.B) {
	benchmarkHasDeviceRules(b, hasDeviceRulesOld)
}

func BenchmarkHasDeviceRulesNew(b *testing.B) {
	benchmarkHasDeviceRules(b, hasDeviceRulesNew)
}
```

kolyshkin: TL;DR: the complexity is not worth the savings.

kolyshkin:
Some performance analysis: https://github.com/opencontainers/runc/pull/3143#issuecomment-898069187

odinuge:
Still wrapping my head around this, and if this is the correct way to do it; both for k8s and for the runc binary.

It is possible that a container has a parent without full device access. In that case, the scope/slice for the container will have `DevicesPolicy=auto`, `DeviceAllow=`, without `devices.list` containing `a *:* rwm\n`. In that case, systemd will not do the "device dance", while with this PR will still freeze.

As is now, I am not 110% comfortable with backporting this to old k8s releases, as this will potentially have a bigger impact. (Again, a simple flag would be best for k8s; both for performance and for safety). I am more comfortable with back porting https://github.com/opencontainers/runc/pull/3143, as that does not have the potential issue from above..

kolyshkin:
Right, this one is not good enough. Querying unit properties seems too expensive to me. I can't think of any other option right now...

Heck, let's do a flag (as long as other maintainers agree) as a stop-gap measure. 

Long term solution is for k8s from using libcontainer/cgroups (https://github.com/kubernetes/kubernetes/issues/104325).

kolyshkin:
Closing in favor of #3161.

PR #3153, opened 2021-08-12
.cirrus.yml: simplify for centos, retry yum

kolyshkin:
The reason for this PR is CI fails on yum sometimes (seen twice today, last time at https://github.com/opencontainers/runc/pull/3151#issuecomment-897377886).

1. Remove redundant yum commands.
2. Add a sleep+retry on fail loop to yum.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -85,7 +85,6 @@ task:
     memory: 8G
 
   install_dependencies_script: |
-    yum install -y -q epel-release
     case $DISTRO in
     centos-7)
       (cd /etc/yum.repos.d && curl -O https://copr.fedorainfracloud.org/coprs/adrian/criu-el7/repo/epel-7/adrian-criu-el7-epel-7.repo)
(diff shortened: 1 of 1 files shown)


kolyshkin:
Ahhh
> Error: Unable to find a match: glibc-static
Seems that powertools repo is not enabled 👎🏻

kolyshkin:
OK, works as intended, hope to bring down flakes.

.cirrus.yml
```diff
@@ -94,11 +93,14 @@ task:
...
-      yum install -y -q dnf-plugins-core
-      yum config-manager --set-enabled powertools
+      yum config-manager --set-enabled powertools # for glibc-static
       ;;
     esac
-    yum install -y -q gcc git iptables jq glibc-static libseccomp-devel make criu
+    # Work around dnf mirror failures by retrying a few times.
+    for i in $(seq 0 2); do
+      sleep $i
+      yum install -y -q gcc git iptables jq glibc-static libseccomp-devel make criu && break
```
AkihiroSuda, line 102: We should make sure to fail here when yum failed on i==2

kolyshkin: Assuming the script is run with `set -e`, it will fail if yum fails on the last iteration. In case `-e` is not set, well, we have to add it to all scripts.

kolyshkin: Yes, it's run with `set -e`. Yes we need to check exit code after `for` 🤦🏻 

Thanks for catching this! Fixed, PTAL.

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

PR #3154, opened 2021-08-12
libct/nsenter: ignore unused return value of write

kailun-qin:
A warning (warn_unused_result) may show up during compilation if a
function returns a result that is never used.
```
nsexec.c: In function ‘write_log’:
nsexec.c:165:2: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
  165 |  write(logfd, json, ret);
      |  ^~~~~~~~~~~~~~~~~~~~~~~
```

This patch explicitly disables/ignores this "-Wunused-result" warning
for write() by using gcc diagnostic pragmas.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -163,8 +163,10 @@ static void write_log(const char *level, const char *format, ...)
 		json = NULL;
 		goto out;
 	}
-
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-result"
 	write(logfd, json, ret);
+#pragma GCC diagnostic pop
 
 out:
 	free(message);


libcontainer/nsenter/nsexec.c
```diff
@@ -164,7 +164,8 @@ static void write_log(const char *level, const char *format, ...)
 		goto out;
 	}
 
-	write(logfd, json, ret);
+	if (write(logfd, json, ret) != ret)
+		goto out;
```
kolyshkin, line 168: this is useless...

A way to suppress a warning is `(void)write(...)`

kailun-qin: Simply casting to (void) won't work. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=25509 for detailed info.

Alternatives below if we really don't want to check the return value, which may however lose kind of readability:
1) (void)!write(...)
2) if (write(...));

What's the preference? Thanks! @kolyshkin

kolyshkin: Oh my...

Frankly, I don't know. What's the best way from the readability point of view?

Perhaps something like

```c
if (write(logfd, json, ret) < 0)
	; // nothing we can do here
```

kailun-qin: Updated. PTAL, thanks! @kolyshkin

kolyshkin (approved):
LGTM

kailun-qin:
@AkihiroSuda @cyphar Would you please kindly take a look when you get a chance? Thanks!

libcontainer/nsenter/nsexec.c
AkihiroSuda: I guess that `if` stmt may break another linter.

Isn't there an gcc `__attribute__` to ignore the warning? (I'm googling around...)

AkihiroSuda (changes requested):
https://github.com/opencontainers/runc/pull/3154#discussion_r689907140

libcontainer/nsenter/nsexec.c
AkihiroSuda: `#pragma GCC diagnostic ignored "-Wunused-result"` may work (untested)

kailun-qin: @AkihiroSuda Thanks for the comment!

emm yes, I agree this may break another linter. But `#pragma GCC diagnostic ignored "-Wunused-result"` won't work for older gcc versions before `Diagnostic Pragmas` was introduced.

Do you think adding braces to avoid linters checking empty bodies for if statements work for you?
```
if (write(logfd, json, ret) < 0) {
	; // nothing we can do here
}
```

kailun-qin: Updated. PTAL thanks! @kolyshkin @AkihiroSuda

kolyshkin:
The alternative is #3165.

libcontainer/nsenter/nsexec.c
AkihiroSuda: Even gcc4 seems supporting that pragma https://stackoverflow.com/questions/63229590/enabling-long-long-in-c89-at-gcc-3-2-4-4-and-5-4

So we can safely use the pragma

kailun-qin: @AkihiroSuda Thanks for checking this!

Updated using pragmas, PTAL. Thanks!

AkihiroSuda (approved):

libcontainer/nsenter/nsexec.c
```diff
@@ -163,8 +163,10 @@ static void write_log(const char *level, const char *format, ...)
 		json = NULL;
 		goto out;
 	}
-
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-result"
 	write(logfd, json, ret);
+#pragma GCC diagnostic pop
```
kolyshkin, line 169: This looks way too heavy to silent a single warning.

I think something like #3168 looks more elegant; PTAL.

AkihiroSuda:
Fixed in https://github.com/opencontainers/runc/pull/3168

Thanks anyway for contribution

PR #3155, opened 2021-08-12
libct/nsenter: fix memory leak on asprintf() error

kailun-qin:
On asprintf() error, setting strp to NULL before free() may cause
memory leak as the contents of strp are undefined.

Setting the unused local pointer json to NULL is not a must as it
won't be accessed anymore after being freed.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -159,10 +159,8 @@ static void write_log(const char *level, const char *format, ...)
 		goto out;
 
 	ret = asprintf(&json, "{\"level\":\"%s\", \"msg\": \"%s[%d]: %s\"}\n", level, stage, getpid(), message);
-	if (ret < 0) {
-		json = NULL;
+	if (ret < 0)
 		goto out;
-	}
 
 	write(logfd, json, ret);
 


kolyshkin (changes requested):
NACK.

"Undefined" does not mean "points to something that needs to be free()d", it means it can be anything in there. We set it to NULL on error so that `free(json)` below won't get a garbage pointer.

Only a successful call to `asprintf()` stores a pointer that can be free()d.

kolyshkin:
> BTW, should we anyway make the logic consistent for message and stage as well on error?

Looks like it. @AkihiroSuda WDYT?

AkihiroSuda (changes requested):
@kolyshkin 

yes, we should nullify message and stage as well

AkihiroSuda:
(In actual glibc implementation we won’t get a garbage pointer on an error, though https://github.com/bminor/glibc/blob/glibc-2.34.9000/libio/vasprintf.c )

kailun-qin:
> (In actual glibc implementation we won’t get a garbage pointer on an error, though https://github.com/bminor/glibc/blob/glibc-2.34.9000/libio/vasprintf.c )

Thanks for checking this!

While in musl, it looks like there is possibility that the underneath [vasprintf](https://github.com/ifduyue/musl/blob/cfdfd5ea3ce14c6abf7fb22a531f3d99518b5a1b/src/stdio/vsnprintf.c#L50) and [vfprintf](https://github.com/ifduyue/musl/blob/cfdfd5ea3ce14c6abf7fb22a531f3d99518b5a1b/src/stdio/vfprintf.c#L657) of `asprintf()` may return `-1` without free after successfully allocated `strp`. I.e., manual handling of it is needed.

Would you please kindly help double check? Thanks! @kolyshkin @AkihiroSuda

kolyshkin:
> Would you please kindly help double check?

Does not look like it. There's a single malloc which can either return NULL or a free()able pointer.

https://git.musl-libc.org/cgit/musl/tree/src/stdio/vasprintf.c

kailun-qin:
> > Would you please kindly help double check?
> 
> Does not look like it. There's a single malloc which can either return NULL or a free()able pointer.
> 
> https://git.musl-libc.org/cgit/musl/tree/src/stdio/vasprintf.c

@kolyshkin 
Right, when malloc succeeds, a free()able pointer will be further passed into `vsnprintf()`. While in `vsnprintf()`, these seems no free of that free()able pointer on error and they return with `-1` directly. I.e., that free()able pointer needs to be handled afterwards when checking `asprintf()` results.

Correct me if I'm wrong. Thanks!

kolyshkin:
> Right, when malloc succeeds, a free()able pointer will be further passed into `vsnprintf()`. While in `vsnprintf()`, there seems no free of that free()able pointer on error and they return with `-1` directly. I.e., that free()able pointer needs to be handled afterwards when checking `asprintf()` results.

This looks correct. The only question is, can `vsnprintf()` return `-1`? It's really hard to say (I can't wrap my head around the entirety of https://git.musl-libc.org/cgit/musl/tree/src/stdio/vfprintf.c) but I assume that if the first vsnprintf call (the one to get the needed size) has not failed, the second one won't.

kailun-qin:
> > Right, when malloc succeeds, a free()able pointer will be further passed into `vsnprintf()`. While in `vsnprintf()`, there seems no free of that free()able pointer on error and they return with `-1` directly. I.e., that free()able pointer needs to be handled afterwards when checking `asprintf()` results.
> 
> This looks correct. The only question is, can `vsnprintf()` return `-1`? It's really hard to say (I can't wrap my head around the entirety of https://git.musl-libc.org/cgit/musl/tree/src/stdio/vfprintf.c) but I assume that if the first vsnprintf call (the one to get the needed size) has not failed, the second one won't.

@kolyshkin Seems when the 1st `vsnprintf()` call gets INT_MAX (not failed), then following by `(INT_MAX+1U)` to malloc (to accommodate `'\0'`) and the 2nd call, there is a chance (in the 2nd call) that `vsnprintf()` can error with `-1` caused by overflow: https://git.musl-libc.org/cgit/musl/tree/src/stdio/vsnprintf.c#n48.

kolyshkin:
> @kolyshkin Seems when the 1st `vsnprintf()` call gets INT_MAX (not failed), then following by `(INT_MAX+1U)` to malloc (to accommodate `'\0'`) and the 2nd call, there is a chance (in the 2nd call) that `vsnprintf()` can error with `-1` caused by overflow: https://git.musl-libc.org/cgit/musl/tree/src/stdio/vsnprintf.c#n48.

This might be correct -- I'd share your finding to the musl mailing list.

In case of runc, however, we never try to print lines long enough for the first vsnprintf to be able to return INT_MAX.

kailun-qin:
> > @kolyshkin Seems when the 1st `vsnprintf()` call gets INT_MAX (not failed), then following by `(INT_MAX+1U)` to malloc (to accommodate `'\0'`) and the 2nd call, there is a chance (in the 2nd call) that `vsnprintf()` can error with `-1` caused by overflow: https://git.musl-libc.org/cgit/musl/tree/src/stdio/vsnprintf.c#n48.
> 
> This might be correct -- I'd share your finding to the musl mailing list.
> 
> In case of runc, however, we never try to print lines long enough for the first vsnprintf to be able to return INT_MAX.

@kolyshkin Thanks a lot!

Updated to nullify message and stage as well to make the handling consistent. PTAL, thanks!

AkihiroSuda (dismissed):

kolyshkin:
Now the PR subject description doesn't match the commit. Perhaps it's better to close this PR and open a new one.

kailun-qin:
> Now the PR subject description doesn't match the commit. Perhaps it's better to close this PR and open a new one.

@kolyshkin @AkihiroSuda 

Changed the PR title and description back to the original one, so that it can match our discussions here.

Reopened a new PR: https://github.com/opencontainers/runc/pull/3162 to track the update.

PTAL, thanks!

kailun-qin:
Revert the code changes back to the original one as well.

PR #3158, opened 2021-08-12
libct/nsenter/nsenter_test.go: fix and improve

kolyshkin:
Fix and improve nsenter_test.go (originally part of #3114).

### DIFF
--- a/libcontainer/nsenter/nsenter_test.go
+++ b/libcontainer/nsenter/nsenter_test.go
@@ -28,10 +28,7 @@ type logentry struct {
 
 func TestNsenterValidPaths(t *testing.T) {
 	args := []string{"nsenter-exec"}
-	parent, child, err := newPipe()
-	if err != nil {
-		t.Fatalf("failed to create pipe %v", err)
-	}
+	parent, child := newPipe(t)
 
 	namespaces := []string{
 		// join pid ns of the current process
@@ -49,6 +46,7 @@ func TestNsenterValidPaths(t *testing.T) {
 	if err := cmd.Start(); err != nil {
 		t.Fatalf("nsenter failed to start %v", err)
 	}
+	child.Close()
 
 	// write cloneFlags
 	r := nl.NewNetlinkRequest(int(libcontainer.InitMsg), 0)
@@ -89,10 +87,7 @@ func TestNsenterValidPaths(t *testing.T) {
 
 func TestNsenterInvalidPaths(t *testing.T) {
 	args := []string{"nsenter-exec"}
-	parent, child, err := newPipe()
-	if err != nil {
-		t.Fatalf("failed to create pipe %v", err)
-	}
+	parent, child := newPipe(t)
 
 	namespaces := []string{
 		// join pid ns of the current process
@@ -108,6 +103,8 @@ func TestNsenterInvalidPaths(t *testing.T) {
 	if err := cmd.Start(); err != nil {
 		t.Fatal(err)
 	}
+	child.Close()
+
 	// write cloneFlags
 	r := nl.NewNetlinkRequest(int(libcontainer.InitMsg), 0)
 	r.AddData(&libcontainer.Int32msg{
@@ -130,10 +127,7 @@ func TestNsenterInvalidPaths(t *testing.T) {
 
 func TestNsenterIncorrectPathType(t *testing.T) {
 	args := []string{"nsenter-exec"}
-	parent, child, err := newPipe()
-	if err != nil {
-		t.Fatalf("failed to create pipe %v", err)
-	}
+	parent, child := newPipe(t)
 
 	namespaces := []string{
 		// join pid ns of the current process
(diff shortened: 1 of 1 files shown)


libcontainer/nsenter/nsenter_test.go
```diff
@@ -49,6 +46,7 @@ func TestNsenterValidPaths(t *testing.T) {
 	if err := cmd.Start(); err != nil {
 		t.Fatal("nsenter failed to start:", err)
 	}
+	child.Close()
```
thaJeztah, line 39: should we use a `defer` so that we can set that directly after creating it (with newpipe)?

I guess this is the "Make sure we close child fds after the start." case? Is the Cleanup() not enough for this? (curious)

libcontainer/nsenter/nsenter_test.go
```diff
@@ -257,3 +238,23 @@ func initWaiter(t *testing.T, r io.Reader) {
...
+	decoder := json.NewDecoder(parent)
+	decoder.DisallowUnknownFields()
+	var pid struct {
+		Pid2 int `json:"stage2_pid"`
+		Pid1 int `json:"stage1_pid"`
+	}
+	if err := decoder.Decode(&pid); err != nil {
+		t.Fatal(err)
+	}
+	if pid.Pid1 == 0 || pid.Pid2 == 0 {
```
thaJeztah, line 238: Does this print the pids as `0 0`? (wondering if `%+v` would be useful here)

libcontainer/nsenter/nsenter_test.go
```diff
@@ -257,3 +238,23 @@ func initWaiter(t *testing.T, r io.Reader) {
...
+		t.Fatal("got pids:", pid)
+	}
+
+	// Reap children.
+	_, _ = unix.Wait4(pid.Pid1, nil, 0, nil)
```
thaJeztah, line 243: Would reaping still be needed if either Pid1 or Pid2 is `!= 0`?

thaJeztah (approved):
LGTM

left some comments/questions, but don't think they're blockers

libcontainer/nsenter/nsenter_test.go
kolyshkin: Not sure I understand the question.

We create a pipe, one end is for the parent, the other is for the child. Once we have forked, the child end needs to be closed. This explicit close is needed because otherwise we won't get EOF in the parent once the child is gone (as the pipe will still be open).

Cleanup() is only executed after the test has finished.

libcontainer/nsenter/nsenter_test.go
kolyshkin: This will print pids as `{0 0}` (I guess the order matches the order of elements in the `struct`).

Since this should never happen normally, I guess it doesn't matter.

libcontainer/nsenter/nsenter_test.go
kolyshkin: In this case the test will fail, the test process will end, and these processes will be reparented to pid 1 (init) who'll reap them. So, technically, this reaping is not needed at all (until one day in a very distant future we'll have so many cases that the number of zombie processes will hit any of the PID limits, and then we'll have all sorts of mysterious failures).

Yet it makes sense to reap before we check for `0` (as we do it on a "best effort" basis anyway).

kolyshkin: done

kolyshkin:
Rebased; renamed `reapChild` to `reapChildren`; moved Wait4 to before pid checks.

libcontainer/nsenter/nsenter_test.go
kolyshkin: The part about EOF is irrelevant as we don't need an EOF from the child. In this case we close the child end of pipe right after the fork merely because we no longer need it, and it's always good to do it explicitly as at the very least it improves code readability ("we passed this end of the pipe to the child and closed it right after, as we have no use for it").

thaJeztah: > In this case we close the child end of pipe right after the fork merely because we no longer need it,

gotcha, yes, makes sense (was merely looking; do we need to do so, because `t.Cleanup()` also does so

thaJeztah (dismissed):
LGTM, thanks!

kolyshkin:
@AkihiroSuda PTAL

kolyshkin:
Had to rebase after #3144 merge. @thaJeztah @AkihiroSuda PTAL

thaJeztah (approved):
LGTM

libcontainer/nsenter/nsenter_test.go
```diff
@@ -47,7 +47,7 @@ func TestNsenterValidPaths(t *testing.T) {
 	}
 
 	if err := cmd.Start(); err != nil {
-		t.Fatalf("nsenter failed to start %v", err)
+		t.Fatal("nsenter failed to start:", err)
```
AkihiroSuda, line 37: Why? The previous form seems completely legit.

kolyshkin: A semicolon was missing. While adding it, I guess I decided to simplify it, that's it.

kolyshkin: The other reason is, I was using the same form in the subsequent patches, so I added it here, too, for uniformity.

Nevermind though, I re-did this commit so it uses `%v` in this and similar cases, and moved the patch to be the last one.

PTAL @AkihiroSuda

AkihiroSuda (approved):

thaJeztah (approved):
still LGTM

PR #3159, opened 2021-08-12
libct/devices: change devices.Type to be a string

thaJeztah:
> I noticed this while reviewing https://github.com/opencontainers/runc/pull/3110. Opening as draft to see if there's any failures in CI, and because this will conflict with that PR

Possibly there was a specific reason to use a rune for this, but I noticed
that there's various parts in the code that has to convert values from a
string to this type. Using a string as type for this can simplify some of
that code.

### DIFF
--- a/libcontainer/cgroups/devices/devices_emulator.go
+++ b/libcontainer/cgroups/devices/devices_emulator.go
@@ -88,7 +88,7 @@ func parseLine(line string) (*deviceRule, error) {
 	}
 	var (
 		rule  deviceRule
-		node  = matches[1]
+		node  = devices.Type(matches[1])
 		major = matches[2]
 		minor = matches[3]
 		perms = matches[4]
@@ -96,16 +96,14 @@ func parseLine(line string) (*deviceRule, error) {
 
 	// Parse the node type.
 	switch node {
-	case "a":
+	case devices.WildcardDevice:
 		// Super-special case -- "a" always means every device with every
 		// access mode. In fact, for devices.list this actually indicates that
 		// the cgroup is in black-list mode.
 		// TODO: Double-check that the entire file is "a *:* rwm".
 		return nil, nil
-	case "b":
-		rule.meta.node = devices.BlockDevice
-	case "c":
-		rule.meta.node = devices.CharDevice
+	case devices.BlockDevice, devices.CharDevice:
+		rule.meta.node = node
 	default:
 		// Should never happen!
 		return nil, fmt.Errorf("unknown device type %q", node)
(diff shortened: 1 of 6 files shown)


thaJeztah:
One failure; something I need to look into

```
=== RUN   TestDeviceFilter_Weird
299
    devicefilter_test.go:26: TestDeviceFilter_Weird: unable to remove 'allow' exception: requested rule [{b 8 2} rwm] not supported by devices cgroupv1 (cannot punch hole in existing wildcard rule [{b -1 -1} rwm]) (devices: [0xc000012e40 0xc000012e80 0xc000012ec0])
300
--- FAIL: TestDeviceFilter_Weird (0.00s)
```

kolyshkin:
I would use `byte` as it's a single character. String is OKish

thaJeztah:
> I would use `byte` as it's a single character

Doing so would bring back the same/similar situation as `rune`. The OCI type uses `string` for the `Type` field (and I guess JSON also doesn't have a concept of `byte` or `rune` values, other than using a `Number` for them).

kolyshkin (changes requested):
Can you split the removal of `stringToCgroupDeviceRune` to a separate commit please?

thaJeztah:
> Can you split the removal of `stringToCgroupDeviceRune` to a separate commit please?

Just the removal, or also calling the function?

I'll have a look tomorrow if I think of doing so 😁

kolyshkin:
> Just the removal, or also calling the function?

I was thinking that a separate first commit in which `stringToCgroupDeviceRune` (together with its use) is eliminated would make sense, but now I see doing this would require a temporary use of `stringToDeviceType` with an added error check, so maybe it's not worth it.

kolyshkin (approved):
LGTM

@cyphar PTAL

thaJeztah:
I think this one may conflict with https://github.com/opencontainers/runc/pull/3110, do you want me to keep it in draft until after that's merged? Otherwise I can move this out of draft now

cyphar (approved):
LGTM. I'm not sure but I think the usage of runes came from the idea that we should use the equivalent of `char` because "that's what you'd do in C", even though in Go `string`s are far more convenient than `rune` or `byte`.

thaJeztah:
> I think the usage of runes came from the idea that we should use the equivalent of char because "that's what you'd do in C", even though in Go strings are far more convenient than rune or byte.

Yes, makes sense! (I was wondering if there was (e.g.) some performance reason behind it that I overlooked)

I moved this one out of draft, but happy to rebase if https://github.com/opencontainers/runc/pull/3110 is gonna be merged first.

cyphar:
@kolyshkin Feel free to merge this if you don't think the rebase of #3110 wouldn't be too bad.

PR #3160, opened 2021-08-13
libct/nsenter: no need to check size_t less than 0

kailun-qin:
According to C standards, `size_t` is always an unsigned integer type.
Thus, checking unsigned expressions to be less than zero is not needed.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -311,7 +311,7 @@ static int try_mapping_tool(const char *app, int pid, char *map, size_t map_len)
 
 static void update_uidmap(const char *path, int pid, char *map, size_t map_len)
 {
-	if (map == NULL || map_len <= 0)
+	if (map == NULL || map_len == 0)
 		return;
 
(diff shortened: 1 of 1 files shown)


libcontainer/nsenter/nsexec.c
```diff
@@ -311,7 +311,7 @@ static int try_mapping_tool(const char *app, int pid, char *map, size_t map_len)
 
 static void update_uidmap(const char *path, int pid, char *map, size_t map_len)
 {
-	if (map == NULL || map_len <= 0)
+	if (map == NULL || !map_len)
```
kolyshkin, line 314: I would prefer `map_len == 0` as it's more readable and is in harmony with the `map == NULL` check.

kailun-qin: Thanks! Make sense to me.

Updated, PTAL! @kolyshkin

kolyshkin (approved):
LGTM

@cyphar @AkihiroSuda PTAL

AkihiroSuda (approved):

PR #3162, opened 2021-08-15
libct/nsenter: nullify pointer on asprintf error

kailun-qin:
The contents of the pointer returned on asprintf() error are undefined
i.e., it can be anything there. We set it to NULL on error so that
free() afterwards won't get a garbage pointer.

This patch applies the above to message and stage as well to be
consistent with what we do for json.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -146,8 +146,10 @@ static void write_log(const char *level, const char *format, ...)
 	va_start(args, format);
 	ret = vasprintf(&message, format, args);
 	va_end(args);
-	if (ret < 0)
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

thaJeztah (approved):
SGTM, but would like a review from @kolyshkin or @cyphar, who are more familiar with this code

kolyshkin:
The subject says

> libct/nsenter: nullify pointer on asprintf error

but this is not what the commit does.

Other than that, LGTM.

kolyshkin (approved):
I guess it better be saying

```
 libct/nsenter: nullify pointer on [v]asprintf error
```

but it's OK as it is...

PR #3164, opened 2021-08-16
CI: Validate compilation without buildtags

rata:
Today we support the seccomp build tag only that is used by default.
However, we are not testing that compiling without any build tag works.

I found the CI didn't catch this when working on #2682, that the CI was
green but compilation without build tags was broken.

We test compilation without build tags only, compilation with the only
build tag supported is done extensively in other actions.

Signed-off-by: Rodrigo Campos <rodrigo@kinvolk.io>

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -23,6 +23,13 @@ jobs:
           # must be specified without patch version
           version: v1.40
 
+  compile-buildtags:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: compile with no build tags
+        run: make BUILDTAGS=""
+
   shfmt:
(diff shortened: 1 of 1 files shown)


rata:
Please note that the action added is included in the checks for this PR

kolyshkin (approved):
Theoretically, we can add BUILDTAGS as another matrix dimension in test.yml. Practically, there will be too much jobs, so this small addition is better.

LGTM

thaJeztah:
> Theoretically, we can add BUILDTAGS as another matrix dimension in test.yml

Yes, I was thinking if we needed other configurations, but we can always look at that if needed.

thaJeztah:
We should also consider changing to `noseccomp` (as the default is / should be "enabled")

thaJeztah (approved):
LGTM

PR #3166, opened 2021-08-16
libct/cg/sd/v1: fix freezeBeforeSet (alt 2)

kolyshkin:
This is a combo of #3143, ~~#3151,~~ and #3161, which aims to be both fast and correct for all use cases.

From #3143 -- fixes to getting and checking device properties;
~~From #3151 -- fast(er) path in freezeBeforeSet (in case devices.list is empty we can skip getting unit properties);~~
From #3161 -- SkipFreezeOnSet knob for kubernetes.
From #3148 -- test cases (those are also cherry-picked in #3143)

Aimed to fix kubernetes/kubernetes#104280 (once we backport, release 1.0.2, and vendor it to kubernetes)

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -346,6 +346,11 @@ func (m *legacyManager) freezeBeforeSet(unitName string, r *configs.Resources) (
 	// Special case for SkipDevices, as used by Kubernetes to create pod
 	// cgroups with allow-all device policy).
 	if r.SkipDevices {
+		if r.SkipFreezeOnSet {
+			// Both needsFreeze and needsThaw are false.
+			return
+		}
+
 		// No need to freeze if SkipDevices is set, and either
 		// (1) systemd unit does not (yet) exist, or
 		// (2) it has DevicePolicy=auto and empty DeviceAllow list.
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


mrunalp:
Can we also add the knob on top of this? :)

kolyshkin:
> Can we also add the knob on top of this? :)

Done.

I can't say I am the biggest fan of this complex monster we have created, but this seems to cover all the use cases in the best known way.

mrunalp:
I am a :+1: on this approach. @odinuge @AkihiroSuda ptal

odinuge:
Yeah, I think this approach is ok overall. :+1:

mrunalp (dismissed):

libcontainer/cgroups/systemd/v1.go
```diff
@@ -332,6 +333,34 @@ func (m *legacyManager) GetStats() (*cgroups.Stats, error) {
 	return stats, nil
 }
 
+// hasDeviceRules checks cgroup device rules. Returns true if any rules
+// other than "allow all" exist. Used by freezeBeforeSet.
+func (m *legacyManager) hasDeviceRules() bool {
+	const allowAll = "a *:* rwm\n"
```
cyphar, line 339: If you have a deny-list rule set, this won't be able to detect it. Unfortunately there isn't a nice solution for this really (there's no real way to figure out what rules are denied if a cgroup is in default-allow mode). If this shortcut really is required, then okay (though this behaviour does at least deserve some kind of comment) but I'd prefer if we didn't have to do this if at all possible.

cyphar (dismissed):
meant to make the last review a "request changes" one

libcontainer/cgroups/systemd/v1.go
kolyshkin: > If you have a deny-list rule set, this won't be able to detect it.

Are you talking about rules from the spec? Note that this code is only used when we have SkipDevices set.

A long explanation of this commit is below.

If all three conditions below is true
1. we are setting properties of an existing systemd unit,
2. we have `SkipDevices` set (e.g. no device rules to be applied)
3. `devices.list` file of unit's cgroup contains nothing but `a *:* rwm` plus a newline,

then systemd will not be writing any device rules in a manner resulting in spurious "access denied" errors.

This is so because (2) means we are not setting any device rules, and (1) and (3) means no rules were ever set before for this systemd unit. This also relies on assumption that systemd properties are in sync with the cgroup state.

Now, if the above (1) && (2) && (3) is not true, we fall back to the slow path of checking systemd unit properties. This happens e.g. if a parent cgroup has some device access restrictions.

Now, I am more than ready to drop this shortcut, especially since we have SkipFreezeOnSet flag which is a "shorter shortcut" (if I can say so), but I am just trying to understand what am I missing here.

kolyshkin: Commit removed; PTAL @cyphar

mrunalp (approved):

libcontainer/cgroups/systemd/v1.go
cyphar: > This is so because (2) means we are not setting any device rules, and (1) and (3) means no rules were ever set before for this systemd unit. This also relies on assumption that systemd properties are in sync with the cgroup state.

The issue is that (3) doesn't necessarily mean that no rules were set.

If you put a cgroup into "deny-list mode" (in other words, you write `a *:* rwm` to `devices.allow` and then rules to `devices.deny`) then `devices.list` always shows `a *:* rwm` **even if you add deny rules to the cgroup**. This is a pretty big anti-feature in cgroupv1 and is one of the reasons there's quite a few hacks in the `devicesemulator` code to deal with the fact that you can't tell whether a cgroup actually has no rules set if `devices.list` says the rule set is `a *:* rwm`.

AkihiroSuda (approved):

cyphar:
(Delayed LGTM.)

PR #3173, opened 2021-08-20
*: release 1.0.2

cyphar:
```
This is a second stable release in 1.0 branch, fixing a few medium and high
priority issues, including one that affect Kubernetes using runc's libcontainer.

Bugfixes:

 - Fixed a failure to set CPU quota period in some cases on cgroup v1. (#3115)
 - Fixed the inability to start a container with the "adding seccomp filter
   rule for syscall ..." error, caused by redundant seccomp rules (i.e. those
   that has action equal to the default one). Such redundant rules are now
   skipped. (#3129)
 - Made release builds reproducible from now on. (#3142)
 - Fixed a rare debug log race in runc init, which can result in occasional
   harmful "failed to decode ..." errors from runc run or exec. (#3130)
 - Fixed the check in cgroup v1 systemd manager if a container needs to be
   frozen before Set, and add a setting to skip such freeze unconditionally.
   The previous fix for that issue, done in  runc 1.0.1, was not working.
   (#3167)
```


AkihiroSuda (approved):
LGTM

(Triggered Moby CI and it’s green)

thaJeztah (approved):
LGTM

For others; diff is https://github.com/opencontainers/runc/compare/v1.0.1...52b36a2dd837e8462de8e01458bf02cf9eea47dd

kolyshkin (approved):
LGTM

PR #3175, opened 2021-08-23
tests/int: cleanups

kolyshkin:
1. tests/int/helpers: rm $bundle handling
    
    It is not used since PR 2757, as all tests are run with cd to bundle
    directory.
    
    runc_spec argument count checking is removed since otherwise shellcheck
    complains:
    
    > SC2120: runc_spec references arguments, but none are ever passed.

2. tests/int/helpers: fix shellcheck warnings

     ... and add the file to be checked by shellcheck.


thaJeztah (approved):
LGTM

PR #3176, opened 2021-08-24
libct/error.go: rm ConfigError (alt)

kolyshkin:
This is the alternative to #3056, addressing review comment from https://github.com/opencontainers/runc/pull/3033#discussion_r661966694.

`ConfigError` type was added by commit e918d021399e62, while removing runc's
own error system, to preserve a way for a libcontainer user to distinguish
between a configuration error and something else.
    
The way ConfigError is implemented requires a different type of check
(compared to all other errors defined by error.go). An attempt was made
to rectify this, but the resulting code became even more complicated.
    
As no one is using this functionality (of differentiating a "bad config"
type of error from other errors), let's just drop the ConfigError type.

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -52,7 +52,7 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 	}
 	for _, c := range warns {
 		if err := c(config); err != nil {
-			logrus.WithError(err).Warnf("invalid configuration")
+			logrus.WithError(err).Warn("invalid configuration")
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


libcontainer/configs/validate/validator.go
```diff
@@ -52,7 +52,7 @@ func (v *ConfigValidator) Validate(config *configs.Config) error {
 	}
 	for _, c := range warns {
 		if err := c(config); err != nil {
-			logrus.WithError(err).Warnf("invalid configuration")
+			logrus.WithError(err).Warn("invalid configuration")
```
kolyshkin, line 55: Technically it doesn't belong here; let me remove it.

kolyshkin: moved to the first commit

kolyshkin:
Setting 1.1.0 milestone as I'd like this refactoring (which changes libcontainer API) to be over before 1.1.0 one way or another.

thaJeztah (approved):
LGTM

thanks!

AkihiroSuda (approved):

PR #3182, opened 2021-08-25
Revert "libct/devices: change devices.Type to be a string"

### LINKED ISSUE #3179: Unable to load containers after devices.Type change
Just found this today:

[root@kir-rhat runc-tst]# ../runc/runc list
load container cc6: json: cannot unmarshal number into Go struct field Device.config.devices.type of type devices.Type

@thaJeztah not sure if there's a way to work around that -- if not, we may need to revert  https://github.com/opencontainers/runc/pull/3159

cyphar:
This reverts commit 814f3ae1d93d35280eefd359be73f473519e7036. This
changed the on-disk state which breaks runc when it has to operate on
containers started with an older runc version. Working around this is
far more complicated than just reverting it.

Fixes #3179
Ref #3159

### DIFF
--- a/libcontainer/cgroups/devices/devices_emulator.go
+++ b/libcontainer/cgroups/devices/devices_emulator.go
@@ -89,7 +89,7 @@ func parseLine(line string) (*deviceRule, error) {
 
 	var (
 		rule  deviceRule
-		node  = devices.Type(fields[0])
+		node  = fields[0]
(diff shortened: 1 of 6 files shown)


AkihiroSuda (approved):

thaJeztah (approved):
LGTM 😢

PR #3185, opened 2021-08-27
Add go:build tags

kolyshkin:
1. *: rm redundant linux build tag
    
    For files that end with _linux.go or _linux_test.go, there is no need to
    specify linux build tag, as it is assumed from the file name.
    
    In addition, rename libcontainer/notify_linux_v2.go -> libcontainer/notify_v2_linux.go
    for the file name to make sense.
    
2. Rm build tags from some packages

    The `linux` build tag to main pkg was added by commit 5aa82c9
    back in the day when we thought
    runc is going to be cross-platform. It's very clear now it's Linux-only
    package. 

   Also, there are some packages that can't currently be compiled on
   non-linux anyway. Remove linux build tag (and _unsupported.go files)
   from those as well.

    While at it, further clarify it in README that we're Linux only.

3. *: add go-1.17+ go:build tags
    
    Go 1.17 introduce this new (and better) way to specify build tags.
    For more info, see https://golang.org/design/draft-gobuild.
    
    As a way to seamlessly switch from old to new build tags, gofmt (and
    gopls) from go 1.17 adds the new tags along with the old ones.
    
    Later, when go < 1.17 is no longer supported, the old build tags
    can be removed.
    
    Now, as I started to use latest gopls (v0.7.1), it adds these tags
    while I edit. Rather than to randomly add new build tags, I guess
    it is better to do it once for all files.
    
    Brought to you by
    
            go1.17 fmt ./...


thaJeztah (dismissed):
LGTM

kolyshkin:
OK, it seems we can remove `+build linux` from packages which currently can not be compiled on non-linux platforms.

For example, `libcontainer/cgroups` can't be compiled now:
```console
[kir@kir-rhat cgroups]$ GOOS=freebsd go build ./...
go build github.com/opencontainers/runc/libcontainer/cgroups/devices: build constraints exclude all Go files in /home/kir/go/src/github.com/opencontainers/runc/libcontainer/cgroups/devices
# github.com/opencontainers/runc/libcontainer/cgroups
./file.go:81:14: undefined: unix.Openat2
./file.go:81:45: undefined: unix.OpenHow
./file.go:102:18: undefined: unix.RESOLVE_BENEATH
./file.go:102:41: undefined: unix.RESOLVE_NO_MAGICLINKS
./file.go:103:17: undefined: unix.CGROUP2_SUPER_MAGIC
./file.go:105:20: undefined: unix.RESOLVE_NO_XDEV
./file.go:105:43: undefined: unix.RESOLVE_NO_SYMLINKS
./v1_utils.go:152:9: undefined: Mount
./v1_utils.go:160:87: undefined: Mount
./v1_utils.go:188:37: undefined: Mount
./file.go:105:43: too many errors
# github.com/opencontainers/runc/libcontainer/utils
../utils/utils.go:127:31: undefined: unix.O_PATH
../utils/utils_unix.go:19:17: undefined: unix.PROC_SUPER_MAGIC
# github.com/godbus/dbus/v5
../../vendor/github.com/godbus/dbus/v5/transport_unix.go:52:3: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)
../../vendor/github.com/godbus/dbus/v5/transport_unix.go:58:3: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)
```

so we can just strip all `// +build linux` tags from it.

Let me rework it so we can remove tags from most packages.

kolyshkin:
OK, more commits but hopefully it makes more sense. I have removed `// +build linux` from those packages that can not be built on non-linux anyway (so removal does not really break anything).

The only packages that can be compiled (tried GOOS=solaris and GOOS=freebsd) are (under `github.com/opencontainers/runc/libcontainer`):
```
system
userns
cgroups/devices
user
configs
devices
integration
nsenter
seccomp
seccomp/patchbpf
apparmor
logs
capabilities
```

These may or may not be used by some software built not just for linux, so let's keep them that way (doesn't make much sense but at least we're not breaking anything).

From now on, I guess, using `linux` build tag (or `_linux.go` suffix) should be discouraged, as runc is linux-only software.

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

PR #3192, opened 2021-08-31
CI: Mark CGO warnings as errors

rata:
This is a follow-up of: https://github.com/opencontainers/runc/pull/3165. In particular, https://github.com/opencontainers/runc/pull/3165#issuecomment-901390858.

As pointed out there by @kolyshkin and @tianon, this adds the -Werror flag to the CI only.

Note that I kept the "-g -O2" that is the default from "go env CGO_CFLAGS". The reason is: the warning that we fixed recently (https://github.com/opencontainers/runc/commit/db8330c9e516ac359e199b999c5321c73aaffabf) is not seen if we don't use those flags. If no env var CGO_CFLAGS is defined, it uses "-g -O2", but if it is defined, it uses what we specify. Therefore, if we add the env var CGO_CFLAGS to "-Werror" only, then it doesn't use "-g -O2", the warning is not produced and therefore not catched. So, I kept those things, but it is hardcoded (instead of append -Werror to whatever the output of "go env CGO_CFLAGS" is) because github actions need to know it statically if we apply them to all jobs

We can probably append "-Werror" to the output of "go env CGO_CFLAGS", but we need to do that in the "run:" section and, therefore, can't change it for all jobs (we can manually change all jobs if we want that, but we need to also remember in the future).

I prefer hardcoding "-g -O2" rather than changing all jobs or creating a new target that only compiles with -Werror, as we might have more buildtags in the future and IMHO it seems simpler to have all tests in the test workflow compile the same way. But let me know if you prefer otherwise :)


.github/workflows/test.yml
```diff
@@ -11,6 +11,10 @@ on:
       - release-*
   pull_request:
 
+env:
+  # Don't ignore warnings. Note that the output of "go env CGO_CFLAGS" by default is "-g -O2", so we keep them.
+  CGO_CFLAGS: -g -O2 -Werror
```
kolyshkin, line 16: perhaps `Don't ignore C warnings`

.github/workflows/validate.yml
```diff
@@ -25,6 +25,9 @@ jobs:
 
   compile-buildtags:
     runs-on: ubuntu-20.04
+    env:
+      # Don't ignore warnings. Note that the output of "go env CGO_CFLAGS" by default is "-g -O2", so we keep them.
```
kolyshkin, line 29: ditto

kolyshkin:
@rata thanks! Now, can you deliberately add a temporary second commit that deliberately introduces an warning to nsexec.c so we can see the CI jobs fail? (can just be a revert of commit db8330c9e516ac359e)

.github/workflows/test.yml
rata: Fixed, thanks!

.github/workflows/validate.yml
rata: Fixed, thanks!

rata:
@kolyshkin yes, verified that in a fork before opening the PR already. I even tried with that same commit you mentioned, it works as expected :)

kolyshkin:
Oh I see.
* https://github.com/kinvolk/runc/runs/3474525294?check_suite_focus=true
* https://github.com/kinvolk/runc/runs/3474525091?check_suite_focus=true

kolyshkin (approved):
LGTM

PR #3194, opened 2021-09-01
libcontainer: add support for Landlock

### LINKED ISSUE #2859: Seccomp vs. Landlock?
With the discussions about some of the limitations of (lib)seccomp (#2151, #2735 etc.), especially in the context of determining which error to return for syscalls that may or may not be around at runtime, the following caught my eye.

I presume most people here are aware of the series of "landlock" [patches](https://lore.kernel.org/lkml/20210316204252.427806-1-mic@digikod.net/) that have been through 30 iterations over the last 5 years. The cover letter states:
> Landlock is inspired by seccomp-bpf but instead of filtering syscalls and their raw arguments, a Landlock rule can restrict the use of kernel objects like file hierarchies, according to the kernel semantic.

I don't pretend to understand well enough if this could - in principle - solve some of the problems around the returned errors or other tricky bits in that regard, but it sounded intriguing - i.e. if that whole error dance could be avoided by letting landlock rather than seccomp handle things.

If landlock's approach is fundamentally unsuited to solving the problems here, I apologise. If OTOH there are only a few things missing for it to be usable, perhaps it would be a good time to reach out on LKML to see if they can be closed (or at least roadmapped) before the series is merged (looks like it's finalising).

kailun-qin:
This patch introduces Landlock Linux Security Module (LSM) support in
runc, which was landed in Linux kernel 5.13.

This allows unprivileged processes to create safe security sandboxes
that can securely restrict the ambient rights (e.g. global filesystem
access) for themselves.

runtime-spec: https://github.com/opencontainers/runtime-spec/pull/1111

Fixes https://github.com/opencontainers/runc/issues/2859


libcontainer/standard_init_linux.go
```diff
@@ -211,6 +212,14 @@ func (l *linuxStandardInit) Init() error {
 	// since been resolved.
 	// https://github.com/torvalds/linux/blob/v4.9/fs/exec.c#L1290-L1318
 	_ = unix.Close(l.fifoFd)
+
+	// `noNewPrivileges` must be enabled to use Landlock.
```
kolyshkin, line 216: Please do not use github markdown in comments.

libcontainer/configs/config.go
```diff
@@ -78,6 +78,53 @@ type Syscall struct {
...
+	HandledAccessFS AccessFS `json:"handledAccessFS"`
+}
+
+// Rules represents the security policies (i.e., actions allowed on objects) in Landlock.
+type Rules struct {
+	PathBeneath []*RulePathBeneath `json:"pathBeneath"`
+}
+
+// RulePathBeneath defines the file-hierarchy typed rule that grants the access rights specified by
+// `AllowedAccess` to the file hierarchies under the given `Paths` in Landlock.
```
kolyshkin, line 99: Please do not use github markdown in comments.

libcontainer/configs/config.go
```diff
@@ -78,6 +78,53 @@ type Syscall struct {
...
+	RemoveDir  AccessFS = (1 << 4)
+	RemoveFile AccessFS = (1 << 5)
+	MakeChar   AccessFS = (1 << 6)
+	MakeDir    AccessFS = (1 << 7)
+	MakeReg    AccessFS = (1 << 8)
+	MakeSock   AccessFS = (1 << 9)
+	MakeFifo   AccessFS = (1 << 10)
+	MakeBlock  AccessFS = (1 << 11)
+	MakeSym    AccessFS = (1 << 12)
+)
```
kolyshkin, line 126: Rather than a copy/paste from https://github.com/landlock-lsm/go-landlock/blob/main/landlock/syscall/landlock.go maybe it can be imported, and the constants are used directly?

libcontainer/landlock/config.go
```diff
@@ -0,0 +1,31 @@
...
+	"make_sym":    configs.MakeSym,
+}
+
+// ConvertStringToAccessFS converts a string into a Landlock access right.
+func ConvertStringToAccessFS(in string) (configs.AccessFS, error) {
+	if access, ok := accessFSs[in]; ok {
+		return access, nil
+	}
+	return 0, fmt.Errorf("string %s is not a valid access right for landlock", in)
+}
```
kolyshkin, line 31: Would be nice to have this functionality in https://github.com/landlock-lsm/go-landlock/blob/main/landlock/accessfs.go and use it.

libcontainer/landlock/landlock_linux.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return nil
+}
+
+// Convert Libcontainer AccessFS to go-landlock AccessFSSet.
+func getAccess(access configs.AccessFS) landlock.AccessFSSet {
+	return landlock.AccessFSSet(access)
+}
+
```
kolyshkin, line 49: Why not use `landlock.AccessFSSet` type directly?

libcontainer/landlock/landlock_unsupported.go
```diff
@@ -0,0 +1,19 @@
...
+
+var ErrLandlockNotSupported = errors.New("land: config provided but Landlock not supported")
+
+// InitLandlock does nothing because Landlock is not supported.
+func InitSLandlock(config *configs.Landlock) error {
+	if config != nil {
+		return ErrLandlockNotSupported
+	}
+	return nil
+}
```
kolyshkin, line 19: runc is linux-only, I guess you can drop this entirely (as well as `_linux` suffix from all the other files).

libcontainer/configs/config.go
```diff
@@ -78,6 +78,53 @@ type Syscall struct {
...
+
+// RulePathBeneath defines the file-hierarchy typed rule that grants the access rights specified by
+// `AllowedAccess` to the file hierarchies under the given `Paths` in Landlock.
+type RulePathBeneath struct {
+	AllowedAccess AccessFS `json:"allowedAccess"`
+	Paths         []string `json:"paths"`
+}
+
+// AccessFS is taken upon ruleset and rule setup in Landlock.
+type AccessFS uint64
```
kolyshkin, line 106: Why not use `landlock.AccessFS` type directly (or, if you're so inclined to have a locally defined type, use something like `type AccessFS=landlock.AccessFS` so you won't need to convert).

libcontainer/landlock/landlock_unsupported.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"errors"
+
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+var ErrLandlockNotSupported = errors.New("land: config provided but Landlock not supported")
+
+// InitLandlock does nothing because Landlock is not supported.
+func InitSLandlock(config *configs.Landlock) error {
```
kolyshkin, line 14: typo? (extra `S`)

libcontainer/landlock/landlock_linux.go
```diff
@@ -0,0 +1,67 @@
...
+	if config.DisableBestEffort {
+		llConfig = landlock.MustConfig(ruleset)
+	} else {
+		llConfig = landlock.MustConfig(ruleset).BestEffort()
+	}
+
+	if err := llConfig.RestrictPaths(
+		getPathAccesses(config.Rules)...,
+	); err != nil {
+		return fmt.Errorf("Could not restrict paths: %v", err)
```
kolyshkin, line 39: nit: error messages should not be Capitalized.

libcontainer/landlock/config.go
l0kod: Related to https://github.com/landlock-lsm/go-landlock/issues/14

libcontainer/landlock/landlock_unsupported.go
```diff
@@ -0,0 +1,19 @@
...
+
+package landlock
+
+import (
+	"errors"
+
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+var ErrLandlockNotSupported = errors.New("land: config provided but Landlock not supported")
```
l0kod, line 11: What does "land:" mean?

kailun-qin:
Address comments and add unit tests. PTAL, thanks!

libcontainer/landlock/config.go
l0kod: Feel free to create a PR for go-landlock :wink:

gnoack: As I mentioned also in https://github.com/landlock-lsm/go-landlock/issues/14, please be aware that when this string-to-enum mapping is part of go-landlock, the set of supported values might grow at some point in the future when you update the go-landlock library, and this might change the permitted values in your configuration language implicitly.

If this is the approach you want to take, and since @l0kod seems to also be on board, we could make that mapping part of the library, in the lower-case, underscore-delimited variant ("write_file")

l0kod: It may help to have an helper to check that an access right is supported by a specific ABI version, or specify the version as an argument. I'm not sure if the string conversion belongs to go-landlock or libcontainer though. Is libcontainer used by other container runtimes?

libcontainer/landlock/landlock_linux.go
```diff
@@ -0,0 +1,67 @@
...
+// security approach should be applied for Landlock access rights.
+func InitLandlock(config *configs.Landlock) error {
+	if config == nil {
+		return errors.New("cannot initialize Landlock - nil config passed")
+	}
+
+	var llConfig landlock.Config
+
+	ruleset := getAccess(config.Ruleset.HandledAccessFS)
+	// Panic on error when constructing the Landlock configuration using invalid config values.
```
gnoack: You can use landlock.NewConfig() instead to handle this more gracefully. (It returns the config and an error)

libcontainer/landlock/landlock_linux.go
```diff
@@ -0,0 +1,67 @@
...
+	if config.DisableBestEffort {
+		llConfig = landlock.MustConfig(ruleset)
+	} else {
+		llConfig = landlock.MustConfig(ruleset).BestEffort()
+	}
+
+	if err := llConfig.RestrictPaths(
```
gnoack: I have today changed the semantics here *slightly*, so that the AccessFSSet passed as PathAccess() parameter needs to be a subset of the one configured in MustConfig(). The intention being that it should be forbidden to say "A and B are forbidden, but C is permitted on /tmp". That makes no sense to say, as C was never forbidden, so it's now treated as an error. You might want to catch that case at the configuration level already, otherwise users will get an error from RestrictPaths() if they misconfigure it in their config file.

Commit is here:
https://github.com/landlock-lsm/go-landlock/commit/c56710719da463053cfbab001de126082d769f47

See bug https://github.com/landlock-lsm/go-landlock/issues/18 for details.

libcontainer/landlock/landlock_linux.go
kailun-qin: Thanks! More make sense to use NewConfig instead.

libcontainer/landlock/config.go
kailun-qin: It should be usable for other container runtimes but I'm not sure here. @kolyshkin @AkihiroSuda @cyphar Any insights? Thanks!

AkihiroSuda:
@kailun-qin 

Could you update this PR with the latest revision of https://github.com/opencontainers/runtime-spec/pull/1111 ?
Then we can confirm the design of the PRs toward merging.

Thanks!

AkihiroSuda:
ping @kailun-qin 🙏

Zheaoli:
@AkihiroSuda May I carry this PR?

rata (commented):
Grr, I wanted to post on the runtime-spec, sorry :joy:. Ignore this

AkihiroSuda:
Needs rebase

kolyshkin:
Needs rebase; I guess we have to remove this from 1.6 milestone.

PR #3197, opened 2021-09-03
*: add cross-build, CI job, update to libseccomp 2.5.2

kolyshkin:
This implements cross-build for `make release`, moving the build into a
container. This way we can support arm, arm64, ppc64, and whatnot.

* `script/seccomp.sh`: separate out of `script/release.sh`, amend to support
  cross-compile and save needed environment variables to a file.

* `Dockerfile`: add installing libseccomp from source, as this is needed
  for release builds.

* `script/release.sh`: amend to support more architectures in addition to
  the native build. Additional arches can be added by specifying
  `-a <arch>` argument (can be specified multiple times), or
  `make RELEASE_ARGS="-a arm64 -a ppc64le" release` if called via `make`.

* `Makefile`: move `release` target to `localrelease`, add `release`
  target to build via the Dockerfile. This is done because most distros
  (including Fedora and openSUSE) lack cross-glibc, which is needed to
  cross-compile libseccomp.

* `Makefile`: remove `cross` and `localcross` targets, as this is now done
  by the release script.

* `.github/workflows/validate.yum`: amend the release CI job to cross-build
  for supported architectures, remove `cross` job.


script/release.sh
```diff
@@ -51,10 +69,15 @@ function build_project() {
 	# Add -a to go build flags to make sure it links against
 	# the provided libseccomp, not the system one (otherwise
 	# it can reuse cached pkg-config results).
-	make -C "$root" PKG_CONFIG_PATH="${prefix}/lib/pkgconfig" COMMIT_NO= EXTRA_FLAGS="-a" EXTRA_LDFLAGS="${ldflags}" static
-	rm -rf "$prefix"
+	make -C "$root" PKG_CONFIG_PATH="${LIBSECCOMP_PREFIX}/lib/pkgconfig" COMMIT_NO= EXTRA_FLAGS="-a" EXTRA_LDFLAGS="${ldflags}" static
```
cyphar, line 72: Just wanted to point out that the hardcoding of `/lib/pkgconfig` as the path means this script hasn't worked on my machine for several releases now (I manually patch it to use `/lib64/...` which is where the pkgconfig descriptions get installed under openSUSE). I've been meaning to fix it properly (adding `--libdir` to `configure` should work) but since you said you're going to rework `release.sh` can you please fix this as well? Thanks.

kolyshkin: That's weird; `./configure --help says`:

>   --libdir=DIR            object code libraries [EPREFIX/lib]

so I can't understand how this can be changed to lib64. Guess it's time to add opensuse to my VM collection.

(half an hour later)

I guess one does not know autotools good enough, ever.

Maybe you can just can do `unset SITE_CONFIG`...

Anyway, let me patch it.

kolyshkin:
OK I made it working...

...alas not for opensuse (or Fedora, for that matter). Problem is, only Debian (and Ubuntu, to some extent) provide good cross-compilation support. For opensuse in particular, this is nicely summarized by this statement (from https://software.opensuse.org/package/cross-aarch64-gcc11):

> Note this is only useful for building freestanding things like the kernel since it fails to include target libraries and headers.

It's the same for Fedora (from `rpm -qi gcc-aarch64-linux-gnu` output):

> Only building kernels is currently supported.  Support for cross-building
> user space programs is not currently provided as that would massively multiply
> the number of packages.

Ubuntu seems to fully support at least arm64 (as shown by validate/release CI job in this PR), but does not support some other architectures (so we have to use Docker for validate/cross CI job).

To sum it up -- building runc.amd64 it's ready and working (in fact you can download it from this PR's CI -- go to Checks tab, choose validate, and scroll down looking for Artifacts, in there you can find a zip with the binaries), but only for Ubuntu/Debian.

We can change the release process to use the binaries built in CI, but it's a big change.

@cyphar @AkihiroSuda WDYT?

cyphar:
@kolyshkin Since we only have one dependency (libseccomp) which we build ourselves, I think it might be possible for us to do it -- we should only require an architecture-appropriate compiler and glibc. But if that isn't possible (or is too fragile / annoying) we can switch to building cross-architecture binaries in CI.

The only issue is that I wouldn't be able to automatically sign them as part of the release (I can obviously manually sign them afterwards but it'll be a bit more manual work than the release process currently uses). I wonder if we could just script the whole release process with the GitHub API (push the tag, wait for the CI to build the cross-arch artefacts, download, sign, then upload to the release). Then again, how much time would we be saving over the manual method? :sweat_smile:

AkihiroSuda:
> The only issue is that I wouldn't be able to automatically sign them as part of the release (I can obviously manually sign them afterwards but it'll be a bit more manual work than the release process currently uses). I wonder if we could just script the whole release process with the GitHub API (push the tag, wait for the CI to build the cross-arch artefacts, download, sign, then upload to the release). Then again, how much time would we be saving over the manual method? 😅

SGTM, but I don't think signing has to be automated.

If we want to sign the binary on GHA, at least we will have to create a new GPG key,  because I don't want you to leak your existing GPG key to GHA secrets 😄

script/release.sh
kolyshkin: Fixed by #3200

kolyshkin:
> Since we only have one dependency (libseccomp) which we build ourselves, I think it might be possible for us to do it -- we should only require an architecture-appropriate compiler and **glibc**.

which, sadly, neither Fedora nor openSUSE has -- they only provide gcc and binutils.

aarch64-glibc is available for both Debian and Ubuntu, though.

Automating the release process in CI seems like a bigger project that I can take on at the moment. If anyone can help that'd be awesome.

This PR can go through, though, once I'll add an option to `script/release.sh` to specify additional architectures.

cyphar:
> > The only issue is that I wouldn't be able to automatically sign them as part of the release (I can obviously manually sign them afterwards but it'll be a bit more manual work than the release process currently uses). I wonder if we could just script the whole release process with the GitHub API (push the tag, wait for the CI to build the cross-arch artefacts, download, sign, then upload to the release). Then again, how much time would we be saving over the manual method? sweat_smile
> 
> SGTM, but I don't think signing has to be automated.
> 
> If we want to sign the binary on GHA, at least we will have to create a new GPG key, because I don't want you to leak your existing GPG key to GHA secrets smile

Haha, I meant to say that we have a hypothetical "release.sh" script I run locally which:

 1. Pushes the release tag and waits for the CI to run and build the release artefacts.
 2. Downloads the release artefacts.
 3. (Optionally) verifies that the match the locally built version (for amd64).
 4. Signs everything.
 5. Creates a new draft release and uploads the artefacts and signatures via the GitHub API.
   - This would also allow us to automate the annoying copy-pasting I have to do for the LGPL license notice.
 6. (Optionally) Opens my web browser to the draft release.

For what it's worth we probably should have a shared runc maintainers key (or maybe a `runc.keyring` stored in the repository would be sufficient) so I'm not the only one who is able to sign releases. I'm also not a fan of the idea of having signing be automated with a CI action -- it somewhat defeats the purpose of signing (if someone has the ability to publish a release they'd have access to the signing keys too).

cyphar:
> > Since we only have one dependency (libseccomp) which we build ourselves, I think it might be possible for us to do it -- we should only require an architecture-appropriate compiler and **glibc**.
> 
> which, sadly, neither Fedora nor openSUSE has -- they only provide gcc and binutils.

Damn. Yeah guess we'll need to figure out some other solution for signing -- but this PR should be enough for the moment (I can manually add the artefacts for the next few releases until we have a better release script).

kolyshkin:
This is no longer a draft and ready to go in. Used by CI, and can be used locally as well (`make RELEASE_ARGS="-a arm64" release`) as long as prerequisites are met. @cyphar PTAL

kolyshkin:
> (Optionally) verifies that the match the locally built version (for amd64).

This won't work I guess, since the distro used in CI is different from yours, and thus gcc, glibc, binutils, and golang (and maybe something else) all differ.

BTW this reminds me we probably need to use go 1.17 for validate/release CI job

kolyshkin:
> This won't work I guess, since the distro used in CI is different from yours, and thus gcc, glibc, binutils, and golang (and maybe something else) all differ.

In fact, even the distro should be the same to have a repro binary...

This is an excerpt from diffoscope output, comparing 1.0.2 binaries from the release job (available from https://github.com/opencontainers/runc/actions/runs/1157854520) with the one we have released:

```
├── readelf --wide --decompress --string-dump=.comment {}
│ @@ -1,4 +1,4 @@
│  
│  String dump of section '.comment':
│ -  [     0]  GCC: (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
│ +  [     0]  GCC: (SUSE Linux) 11.1.1 20210721 [revision 076930b9690ac3564638636f6b13bbb6bc608aea]
```

AkihiroSuda:
> > (Optionally) verifies that the match the locally built version (for amd64).
> 
> This won't work I guess, since the distro used in CI is different from yours, and thus gcc, glibc, binutils, and golang (and maybe something else) all differ.
> 
> BTW this reminds me we probably need to use go 1.17 for validate/release CI job

Would it be possible to have CGO-less golang binding for seccomp to avoid non-reproducible builds? Has anyone attempted?

cyphar:
> Would it be possible to have CGO-less golang binding for seccomp to avoid non-reproducible builds? Has anyone attempted?

I don't see how we could -- libseccomp-golang depends on it. libseccomp does the filter generation and it's all a userspace API so we'd need to replace libseccomp (which I'm not against for the record -- it would be nice to have more control over the filter generation so we don't need the enosys stub). And obviously we can't ship a runc version without seccomp support, since it's one of the primary security mechanisms we have in the container world...

> This won't work I guess, since the distro used in CI is different from yours, and thus gcc, glibc, binutils, and golang (and maybe something else) all differ.

Ah you're quite right, the glibc issue alone makes this untenable. I mean, my laptop probably isn't strictly more secure than a GHA build environment -- it would just be nice to have some kind of sanity check before we ship binaries signed with our keys. Alternatively we could just always build the binaries inside a container, that way they'd be reproducible and we could use an Ubuntu image to allow cross-compiling locally? (Though I still might want to play around with automating the release procedure...)

kolyshkin:
> Alternatively (and I'm surprised it took me this long to think of this) -- we could just always build the binaries inside a container

You and me both 😆

We already have a Debian-based image that can be used for testing and cross-compilation. It's easy to reuse it for building release binaries... let me see...

h-vetinari:
> > Alternatively (and I'm surprised it took me this long to think of this) -- we could just always build the binaries inside a container
> 
> You and me both 😆

_(with nothing but deep respect for y'all's work)_ this is hilarious. 😄

kolyshkin:
OK, this is finally ready, you can use
```
make RELEASE_ARGS="-a arm64 -a armel -a armhf -a ppc64le" release
```
locally. It takes 5 minutes on CI (with cache, or 10 minutes without), and about 1 minute on my laptop (using podman, with all layers already cached).

You can still use script/release.sh directly as before (or `make localrelease` via Makefile). I am not sure if I should remove building libseccomp from it, as currently it relies on seccomp being pre-built).

Overall I think this is a big ugly now, and can be simplified and improved (I am pretty sure @thaJeztah can think of some Dockerfile improvements, for one thing), but I've already spent much more than I wanted on it.

kolyshkin:
Note that I have removed `cross` CI job as the `release` job supersedes it. Will remove `cross` from the list of required jobs if/after this is merged.

kolyshkin:
@cyphar @AkihiroSuda PTAL

script/seccomp.sh
```diff
@@ -0,0 +1,63 @@
...
+	shift
+	local arches=("$@")
+	local tar="libseccomp-${ver}.tar.gz"
+
+	# Download.
+	wget "https://github.com/seccomp/libseccomp/releases/download/v${ver}/${tar}"{,.asc}
+
+	# Build and install natively.
+	tar xf "$tar"
+	pushd "libseccomp-${ver}" || return
```
cyphar, line 29: It would be nice if we didn't put this directory in the current directory and instead used a temporary directory, but this is already an existing issue with the current script so it doesn't hurt to fix it some other time.

cyphar (approved):
LGTM. Now to automate the rest of the release process. (I'll work on that sometime soon. :wink:)

.github/workflows/validate.yml
```diff
@@ -169,12 +148,21 @@ jobs:
...
+      # under Docker will emerge, it will be good to have a separate make
+      # runcimage job and share its result (the docker image) with whoever
+      # needs it.
+    - uses: satackey/action-docker-layer-caching@v0.0.11
+      continue-on-error: true
+    - name: build docker image
+      run: make runcimage
     - name: make release
-      run: make release
+      run: make RELEASE_ARGS="-a arm64 -a armel -a armhf -a ppc64le" release
```
cyphar, line 165: It would be nice if there was a `make releaseall` or something target that built for every architecture we support.

script/seccomp.sh
kolyshkin: done

.github/workflows/validate.yml
kolyshkin: done

kolyshkin:
Addressed review comments from @cyphar; PTAL

cyphar (approved):
LGTM -- thanks for working on this!

cyphar:
I removed `cross` from the list of required checks since this PR removes the job entirely.

mrunalp (approved):

PR #3198, opened 2021-09-04
optimize log: move WriteJSON defer as early as possible

Vanient:
if function returns error before WriteJSON defer, error will not be
printed out, so move this defer as early as possible and use logrus to
print out error if returns before it.

Signed-off-by: xiadanni <xiadanni1@huawei.com>

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -22,6 +22,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs/validate"
 	"github.com/opencontainers/runc/libcontainer/intelrdt"
 	"github.com/opencontainers/runc/libcontainer/utils"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


libcontainer/factory_linux.go
```diff
@@ -346,11 +347,25 @@ func (l *LinuxFactory) StartInitialization() (err error) {
 	envInitPipe := os.Getenv("_LIBCONTAINER_INITPIPE")
 	pipefd, err := strconv.Atoi(envInitPipe)
 	if err != nil {
+		logrus.Errorf("unable to convert _LIBCONTAINER_INITPIPE: %v", err)
 		return fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE: %w", err)
```
kolyshkin, line 351:
```go
err = fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE: %w", err)
logrus.Error(err)
return err
```

(or something similar -- the idea is to not duplicate the error wrapping).

Vanient: Thanks for your feedback~ fixed.

kolyshkin (approved):
LGTM, and thanks.

This PR finally makes runc create/run/exec actually log all possible errors from runc init 👍🏻

AkihiroSuda (approved):

Issue #3135, opened 2021-08-06
Make use of cgroup.kill


kolyshkin:
~~Cgroup v2~~ Kernel 5.14+ implements `cgroup.kill` which does the same thing that we try to do in `signalAllProcesses` https://github.com/opencontainers/runc/blob/8772c4dd2fd36b1cc1b31d1ec9e38debd0c786a9/libcontainer/init_linux.go#L499
but in the kernel and reliably.

Since we're using `signalAllProcesses` mostly to send SIGKILL to container (with the exception of `runc kill --all`), this needs to be moved to cgroup managers to benefit from the new feature.

kolyshkin:
So, now we have this:

https://github.com/opencontainers/runc/blob/b114862c537017d9dbf10999a6efb5fc181672b5/libcontainer/state_linux.go#L40-L47

and this:

func killContainer(container libcontainer.Container) error {
	_ = container.Signal(unix.SIGKILL, false)
	for i := 0; i < 100; i++ {
		time.Sleep(100 * time.Millisecond)
		if err := container.Signal(unix.Signal(0), false); err != nil {
			destroy(container)
			return nil
		}
	}
	return errors.New("container init still running")

and it would be nice to use `cgroup.kill` in both cases (possibly by adding `Kill()` method to cgroupManager, implementing it in a traditional way for v1, and use `cgroup.kill` for v2).

kolyshkin:
This is a relatively new addition and might not be available for older kernels: https://lwn.net/Articles/855924/

PR #3199, opened 2021-09-05
cgroups: Add support for `cgroup.kill` added in Linux Kernel 5.14

flouthoc:
Following PR adds support for `cgroup.kill` to kill all the processes in
a cgroup instead of iterating over each process to save cost if
possible.

Linux kernel 5.14 adds support for `cgroup.kill`.

Ref
* https://lwn.net/Articles/855049/
* https://lwn.net/Articles/855924/

### DIFF
--- a/libcontainer/cgroups/cgroups.go
+++ b/libcontainer/cgroups/cgroups.go
@@ -41,6 +41,10 @@ type Manager interface {
 	// Destroy removes cgroup.
 	Destroy() error
 
+	// Leverages "cgroup.kill" added in kernel 5.14.
+	// Kills all the process listed in cgroup.
+	Kill() error
+
 	// Path returns a cgroup path to the specified controller/subsystem.
 	// For cgroupv2, the argument is unused and can be empty.
 	Path(string) string
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -150,6 +150,16 @@ func (m *Manager) Path(subsys string) string {
 	return m.paths[subsys]
 }
 
+func (m *Manager) Kill() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	// Since "cgroup.kill" is not available for v1 check for cgroup mode hybrid+v1
+	if cgroups.IsCgroup2HybridMode() {
+		return fscommon.Kill(m.paths[""])
+	}
+	return errors.New("cgroup.kill not supported")
+}
+
 func (m *Manager) GetStats() (*cgroups.Stats, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
 	"github.com/sirupsen/logrus"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -235,6 +236,16 @@ func (m *LegacyManager) Path(subsys string) string {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	return m.paths[subsys]
+}
+
+func (m *LegacyManager) Kill() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	// Since "cgroup.kill" is not available for v1 check for cgroup mode hybrid+v1
+	if cgroups.IsCgroup2HybridMode() {
+		return fscommon.Kill(m.paths[""])
+	}
+	return errors.New("cgroup.kill not supported")
 }
 
 func (m *LegacyManager) joinCgroups(pid int) error {
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -528,9 +528,18 @@ func isWaitable(pid int) (bool, error) {
 // exit status and only if it is will a wait be performed.
 func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	var procs []*os.Process
+	// If signal is SIGKILL try manager.Kill()
+	// in order to leverage "cgroup.kill" instead
+	// of iterating over each process. "cgroup.kill"
+	// was added in Linux kernel 5.14 onwards.
+	if s == unix.SIGKILL && m.Kill() == nil {
+		return nil
+	}
+	// Fallback to slow method
 	if err := m.Freeze(configs.Frozen); err != nil {
 		logrus.Warn(err)
 	}
+
 	pids, err := m.GetAllPids()
 	if err != nil {
 		if err := m.Freeze(configs.Thawed); err != nil {
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -143,6 +143,10 @@ func (m *Manager) Destroy() error {
 	return cgroups.RemovePath(m.dirPath)
 }
 
+func (m *Manager) Kill() error {
+	return fscommon.Kill(m.dirPath)
+}
+
 func (m *Manager) Path(_ string) string {
 	return m.dirPath
 }
--- a/libcontainer/cgroups/fscommon/kill.go
+++ b/libcontainer/cgroups/fscommon/kill.go
@@ -1,0 +1,11 @@
+package fscommon
+
+import (
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+)
+
+// Linux 5.14 supports the file "cgroup.kill" to kill all the processes
+// in a cgroup. Check https://lwn.net/Articles/855924/
+func Kill(path string) error {
+	return cgroups.WriteFile(path, "cgroup.kill", "1")
+}
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
@@ -368,6 +369,12 @@ func (m *UnifiedManager) Destroy() error {
 
 func (m *UnifiedManager) Path(_ string) string {
 	return m.path
+}
+
+func (m *UnifiedManager) Kill() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return fscommon.Kill(m.path)
 }
 
 // getSliceFull value is used in initPath.
--- a/libcontainer/container_linux_test.go
+++ b/libcontainer/container_linux_test.go
@@ -1,6 +1,7 @@
 package libcontainer
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"testing"
@@ -38,6 +39,10 @@ func (m *mockCgroupManager) Set(_ *configs.Resources) error {
 
 func (m *mockCgroupManager) Destroy() error {
 	return nil
+}
+
+func (m *mockCgroupManager) Kill() error {
+	return errors.New("Not supported for mock devices")
 }
 
 func (m *mockCgroupManager) Exists() bool {


flouthoc:
@giuseppe @AkihiroSuda @kolyshkin PTAL

flouthoc:
PS: this is already added in `crun` by @giuseppe via https://github.com/containers/crun/pull/724

libcontainer/init_linux.go
```diff
@@ -499,6 +499,26 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
...
+	if s == unix.SIGKILL {
+		// If manager.Kill() failed
+		// fallback to legacy method
+		err := m.Kill()
+		// If err is nil
+		// Means "cgroup.kill" was successful
+		// return early from function
+		if err == nil {
+			if err := m.Freeze(configs.Thawed); err != nil {
+				logrus.Warn(err)
```
AkihiroSuda, line 516: logrus.WithError(err).Warnf(“some human-readable message here”

flouthoc: Sure :)

libcontainer/cgroups/cgroups.go
```diff
@@ -26,6 +26,10 @@ type Manager interface {
 	// Destroy removes cgroup.
 	Destroy() error
 
+	// Leverages "cgroup.kill" added in kernel 5.14 onwards
+	// Kills all the process listed in cgroup
```
kolyshkin, line 30: nit
```suggestion
	// Leverages "cgroup.kill" added in kernel 5.14 onwards.
	// Kills all the process listed in cgroup.
```

libcontainer/cgroups/fs/fs.go
```diff
@@ -241,6 +241,15 @@ func (m *manager) Path(subsys string) string {
 	return m.paths[subsys]
 }
 
+func (m *manager) Kill() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	// Since its not a subsystem and we dont want to regenerate path using pid
```
kolyshkin, line 247: nits:
* `its` -> `it's`
* `dont' -> `don't`

libcontainer/cgroups/fs/fs.go
```diff
@@ -241,6 +241,15 @@ func (m *manager) Path(subsys string) string {
...
+	// strip dir path from any available subsystem, mostlikely cpu since its always there for v1
```
kolyshkin, line 248: 1. `mostlikely` -> `most likely`

2. Period is missing at the end of the sentence.

3. I think (but not 100% sure) that we usually depend on `devices` controller. Check the code.

libcontainer/init_linux.go
```diff
@@ -499,6 +499,26 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	if err := m.Freeze(configs.Frozen); err != nil {
 		logrus.Warn(err)
 	}
+	// If signal is SIGKILL try manager.Kill()
+	// inorder to leverage "cgroup.kill" instead
```
kolyshkin, line 503: nit: `in order`

libcontainer/init_linux.go
```diff
@@ -499,6 +499,26 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
...
+	// of iterating over each process. "cgroup.kill"
+	// was added in Linux kernel 5.14 onwards read
+	// more about it https://lwn.net/Articles/855924/
+	if s == unix.SIGKILL {
+		// If manager.Kill() failed
+		// fallback to legacy method
+		err := m.Kill()
+		// If err is nil
+		// Means "cgroup.kill" was successful
```
kolyshkin, line 512: Isn't that obvious?

libcontainer/init_linux.go
```diff
@@ -499,6 +499,26 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
...
+		// return early from function
+		if err == nil {
+			if err := m.Freeze(configs.Thawed); err != nil {
```
kolyshkin, line 515: The whole point of introducing cgroup.kill is you don't have to freeze/thaw.

kolyshkin:
This would fix https://github.com/opencontainers/runc/issues/3135

libcontainer/cgroups/fs/fs.go
cyphar: Yeah this should be using `devices`.

libcontainer/init_linux.go
```diff
@@ -499,6 +499,26 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	if err := m.Freeze(configs.Frozen); err != nil {
 		logrus.Warn(err)
 	}
+	// If signal is SIGKILL try manager.Kill()
+	// inorder to leverage "cgroup.kill" instead
+	// of iterating over each process. "cgroup.kill"
+	// was added in Linux kernel 5.14 onwards read
+	// more about it https://lwn.net/Articles/855924/
```
cyphar, line 506: You don't need to add a reference to LWN every time the feature is used (in fact the reference isn't really necessary at all -- `cgroup.kill` is a fairly easy-to-understand part of the kernel ABI that is documented in the actual kernel docs now).

libcontainer/init_linux.go
flouthoc: Freeze is removed in latest commit. Please check https://github.com/opencontainers/runc/pull/3199#discussion_r703047099

libcontainer/cgroups/cgroups.go
flouthoc: Fixed

libcontainer/cgroups/fs/fs.go
flouthoc: Fixed

libcontainer/cgroups/fs/fs.go
flouthoc: Fixed

libcontainer/init_linux.go
flouthoc: Fixed

libcontainer/init_linux.go
flouthoc: Fixed in latest commit

libcontainer/init_linux.go
flouthoc: Removed in latest commit.

libcontainer/init_linux.go
flouthoc: @cyphar thanks removed in latest commit

flouthoc:
@cyphar @AkihiroSuda @kolyshkin Addressed all the feedback points in latest commit. Could you please take a look again.

libcontainer/init_linux.go
```diff
@@ -499,6 +499,21 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
 	if err := m.Freeze(configs.Frozen); err != nil {
 		logrus.Warn(err)
 	}
```
kolyshkin, line 501: Again, the whole point of introducing cgroup.kill is you don't have to freeze/thaw, and yet you do that.

libcontainer/init_linux.go
```diff
@@ -499,6 +499,21 @@ func signalAllProcesses(m cgroups.Manager, s os.Signal) error {
...
+		// If manager.Kill() failed
+		// fallback to legacy method
+		err := m.Kill()
+		// If err is nil
+		// return early from function
+		if err == nil {
+			return nil
+		}
+	}
+
```
kolyshkin, line 516: All this is too verbose, and could just be
```suggestion
if s == unix.SIGKILL && m.Kill() == nil {
    return nil
}
// Fallback to slow method.
```

libcontainer/cgroups/systemd/v1.go
```diff
@@ -207,6 +208,15 @@ func (m *legacyManager) Path(subsys string) string {
 	return m.paths[subsys]
 }
 
+func (m *legacyManager) Kill() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	// Since "cgroup.kill" is not a subsystem and we don't want to regenerate path using pid
+	// strip dir path from any available subsystem, most likely devices since its always there for v1
+	dirPath, _ := filepath.Split(m.paths["devices"])
```
kolyshkin, line 216: So, why are you using `filepath.Split` here?

kolyshkin:
@flouthoc on what distro are you testing this? Problem is, we don't have 5.14+ kernel in CI yet.

libcontainer/init_linux.go
flouthoc: @kolyshkin ah i moved it after this in latest commit but as per man-pages it says

```
+	Killing a cgroup tree will deal with concurrent forks appropriately and
+	is protected against migrations. If callers require strict guarantees
+	they can issue the cgroup.kill request after a freezing the cgroup via
+	cgroup.freeze.```

libcontainer/cgroups/systemd/v1.go
flouthoc: @kolyshkin We want path to tree where we are going to write  "cgroup.kill" for v1 , example `/sys/fs/cgroup/containerID/cgroup.kill`

flouthoc:
@kolyshkin I am using ubuntu 21.04 but `5.14` is not shipped by default but its available on kernel.ubuntu.com mainline.

libcontainer/init_linux.go
kolyshkin: Please see the discussion between @brauner and @rgushchin, which stars [here]( https://lwn.net/ml/cgroups/YIt32/aQJfkw53ic@carbon.dhcp.thefacebook.com/).

In short, the sentence you quote was only in v1 of the patchset and it was dropped later. IOW freezing is not required, and it's one of the "selling points" of this feature.

kolyshkin:
I have booted 5.14 myself and, as I was suspecting, the feature is only available for cgroup v2 (or when using v1 + hybrid hierarchy). 

So, it does not make sense to add to v1 managers (unless they gain hybrid support), and the code you're using to get paths to cgroup.kill in v1 managers is probably not working.

Given the fact that this is not ready, and we don't have kernel 5.14 in CI, I think we have to postpone it to 1.2.0.

libcontainer/cgroups/systemd/v1.go
kolyshkin: 1. The path you provided as an example does not make sense for cgroup v1. 
2. The path that will be produced as a result of `filepath.Split` would be different from your example (it will actually be pointing to the parent cgroup, and I don't think it's a good idea to use it).

So, it looks like you're not testing this code at all (at least for v1). Do you?

libcontainer/init_linux.go
flouthoc: @kolyshkin thanks for pointing out this discussion this has been fixed in last commit.

libcontainer/cgroups/systemd/v1.go
flouthoc: @kolyshkin ah my bad i didn't spent time testing this on `v1` but I followed the path construction which was being created on `crun` let me harden this for `v1` and if its not supported i will remove it. But if its not available for `v1` i think man-pages should be updated.

flouthoc:
@kolyshkin fair points i think we can hold on to this till kernel `5.14` and I'll confirm if this works for `v1` at all , if not i will remove the case for `v1`. Will add a `.bats` test for `v2`  does that sounds good ?

libcontainer/cgroups/systemd/v1.go
flouthoc: @kolyshkin I am unable to write file `cgroup.kill` when working with `v1` under any of the paths even the base path but works when `cgroup_hierarchy=1` also tried with root.  But since its returning error for `v1` so code fallbacks to original slow process killing. PS: did not test this with `using v1 + hybrid`

flouthoc: Now i am thinking should `v1` managers just return `error("Not supported")` but i am not sure if it works with  `v1+hybrid` or not.

kolyshkin:
> and I'll confirm if this works for v1 at all , if not i will remove the case for v1.

There's no cgroup.kill for v1, but it is there in case hybrid hierarchy is used. Support for hybrid hierarchy is added by various PRs. In this particular case, we need https://github.com/opencontainers/runc/pull/2087, and once it's in, we can use `Path("")` to get path to the hybrid hierarchy, if available.

flouthoc:
@kolyshkin Sure makes sense lets wait for hybrid hierarchy.

libcontainer/cgroups/systemd/v1.go
flouthoc: Todo: switch this with `m.paths[""]` to get the path for `v1+hybrid`. Wait for : #2087

kolyshkin:
Marked as draft as we need
 - #2087 (or #3059 which carries it) for v1 + hybrid unified path support;
 - kernel 5.14 in CI;

Problem is, #3059 needs criu 3.16 which is not yet released.

flouthoc:
@kolyshkin @cyphar Just curious how would `m.paths[""]` would look like for `hybrid+v1` could you please state with an example. I'll try setting up a `hybrid+v1` in spare time for myself to work on this till other PR gets merged.

flouthoc:
Just putting mock code so its easier to revisit, please don't review yet.

kolyshkin:
@flouthoc can you revisit this? I almost ended up rewriting this from scratch (which would be a colossal waste of time).

flouthoc:
@kolyshkin Sure I'll revisit this. Thanks for the reminder.

cyphar:
I completely forgot about this patch. Ping me to review it when you're done rebasing. :D

kolyshkin:
@flouthoc can you take a look and refresh this PR? This would be nice to have in 1.2.0

kolyshkin:
I am working on an alternative, implemented without adding a new public method to cgroup managers, and modelled after cgroup.kill kernel tests. Still a WIP.

flouthoc:
@kolyshkin Thanks a lot for taking this over, i am closing this one.

PR #3201, opened 2021-09-07
nsexec.c: don't write logs that are to be discarded

kolyshkin:
~~_(this is based on and includes https://github.com/opencontainers/runc/pull/3157; only review the last 5 commits)_~~

Currently, if the log level is not set to e.g. "debug", runc init sends
some debug logs to the parent, which parses and discards it.
    
It is better to not send those in the first place.

Amend nsexec.c to honor `_LIBCONTAINER_LOGLEVEL`.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -380,39 +380,48 @@ static int clone_parent(jmp_buf *env, int jmpval)
 }
 
 /*
- * Gets the init pipe fd from the environment, which is used to read the
- * bootstrap data and tell the parent what the new pid is after we finish
- * setting up the environment.
+ * Returns an environment variable value as a non-negative integer, or -ENOENT
+ * if the variable was not found or has an empty value.
+ *
+ * If the value can not be converted to an integer, or the result is out of
+ * range, the function bails out.
  */
-static int initpipe(void)
+static int getenv_int(const char *name)
 {
-	int pipenum;
-	char *initpipe, *endptr;
+	char *val, *endptr;
+	int ret;
 
-	initpipe = getenv("_LIBCONTAINER_INITPIPE");
-	if (initpipe == NULL || *initpipe == '\0')
-		return -1;
+	val = getenv(name);
+	/* Treat empty value as unset variable. */
+	if (val == NULL || *val == '\0')
+		return -ENOENT;
 
-	pipenum = strtol(initpipe, &endptr, 10);
-	if (*endptr != '\0')
-		bail("unable to parse _LIBCONTAINER_INITPIPE");
+	ret = strtol(val, &endptr, 10);
+	if (val == endptr || *endptr != '\0')
+		bail("unable to parse %s=%s", name, val);
+	/*
+	 * Sanity check: this must be a non-negative number.
+	 */
+	if (ret < 0)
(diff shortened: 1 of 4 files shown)


cyphar:
#3157 merged.

kolyshkin:
Rebased, no longer a draft. @cyphar @AkihiroSuda @thaJeztah PTAL

libcontainer/nsenter/nsexec.c
```diff
@@ -89,14 +89,21 @@ struct nlconfig_t {
...
+#define ERROR   2
+#define WARNING 3
+#define INFO    4
+#define DEBUG   5
+#define TRACE   6
+
+static const char *level_str[] = { "panic", "fatal", "error", "warning", "info", "debug", "trace" };
 
 static int logfd = -1;
+static int loglevel = INFO;
```
cyphar, line 106: This should default to `DEBUG` -- if there's hypothetically an error very early on in parsing, we wouldn't be able to get debug logs even if we passed `--debug` to `runc`.

kolyshkin: Makes sense. Done.

cyphar (approved):
LGTM.

libcontainer/nsenter/nsexec.c
```diff
@@ -422,6 +432,11 @@ static void setup_logpipe(void)
 		return;
 	}
 	logfd = i;
+
+	i = getenv_int("_LIBCONTAINER_LOGLEVEL");
+	if (i < 0)
+		return;
```
thaJeztah, line 438: Do we have a convention on wether or not to use curly braces for these? I realise it's common in C to omit them, but (e.g.) a couple of lines up we _do_ use curly braces (given, in that case because of the comment).

Wondering if we need to have a convention (also as a safeguard), at least for "new" code.

thaJeztah (approved):
changes in go code LGTM

I reviewed the C code as well, which looks good afaics, but I'm not a C-coder, so would appreciate someone else to also review in case I missed something.

libcontainer/nsenter/nsexec.c
kolyshkin: Good point.

In Go, obligatory curly braces (and, ultimately, gofmt) were introduced in order to minimize such discussions.

In C, curly braces are optional if the body has a single statement, which sometimes leads to bugs like this

```c
zero();

if (val > max) /* Fixup val because something something */
	val = max;
	one();

two();
```

It's kinda hard to spot that `one()` is called unconditionally here (especially if the code is lengthier than in this example).

Fortunately, here we use `indent` (which is a poor man's gofmt for C), which will show a warning about wrong indentation in CI. It won't prevent a bug, but at least it will be obvious.

This is why I think both styles (with and without braces) are OK in here.

thaJeztah: Yes, I'm ok with this (especially because it's already for quite some code), just have to think about Apple's infamous "goto fail" security issue every time I see code without the curly braces 😂 https://www.imperialviolet.org/2014/02/22/applebug.html

kolyshkin: I wish we have some linter to take care of it (I mean real linter, not the indentation checker, although it seems to alleviate the issue so it is visible to the naked eye).

PR #2564, opened 2020-08-20
Add support for seccomp actions `ActKillThread` and `ActKillProcess`

saschagrunert:
Two new seccomp actions have been added to the libseccomp-golang
dependency, which can be now supported by runc, too.

ActKillThread kills the thread that violated the rule. It is the same as
ActKill. All other threads from the same thread group will continue to
execute.

ActKillProcess kills the process that violated the rule. All threads in
the thread group are also terminated. This action is only usable when
libseccomp API level 3 or higher is supported.

Xref for requesting a new release in libseccomp-golang: https://github.com/seccomp/libseccomp-golang/issues/55

PR #3204, opened 2021-09-09
Add support for seccomp actions ActKillThread and ActKillProcess

cyphar:
Carry of #2564.

Two new seccomp actions have been added to the libseccomp-golang
dependency, which can be now supported by runc, too.

ActKillThread kills the thread that violated the rule. It is the same as
ActKill. All other threads from the same thread group will continue to
execute.

ActKillProcess kills the process that violated the rule. All threads in
the thread group are also terminated. This action is only usable when
libseccomp API level 3 or higher is supported.

Signed-off-by: Sascha Grunert <sgrunert@redhat.com>

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -50,6 +50,8 @@ const (
 	Trace
 	Log
 	Notify
+	KillThread
+	KillProcess
 )
 
 // Operator is a comparison operator to be used when matching syscall arguments in Seccomp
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

kolyshkin (approved):
I would structure this the other way around (drop unnecessary defines first, add ActKill{Thread,Process} next) -- will be cleaner this way, and less git history to dig through.

But it's a teeny tiny change and thus it's fine with it as it is.

LGTM

cyphar:
Ah you're quite right @kolyshkin, fixed.

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

Issue #3145, opened 2021-08-11
Support CFS bandwidth burst


kailun-qin:
Burstable CFS controller is landed in Linux 5.14 [1], though the documentation is still WIP [2]. This helps with parallel workloads that might be bursty. They can get throttled even when their average utilization is under quota. And they may be latency sensitive at the same time so that throttling them is undesired.

This feature borrows time now against the future underrun, at the cost of increased interference against the other system users, by introducing cfs_burst_us into CFS bandwidth control to enact the cap on unused bandwidth accumulation, which will then used additionally for burst.

In general, it brings potential improvements on the efficiency of the CPU resource usage which may eventually help reduce the TCO and provide better QoS, which should be quite useful for many Cloud scenarios. It's great to have it supported in runc.

Drafted runtime-spec update: https://github.com/opencontainers/runtime-spec/pull/1120.

[1] torvalds/linux@f418371#diff-cc1a82129952a910fdc4292448c2a097a2ba538bebefcf3c06381e45639ae73e
[2] https://lore.kernel.org/lkml/20210730070956.44019-3-changhuaixin@linux.alibaba.com/T/

m-yosefpor:
just adding runc PR link: https://github.com/opencontainers/runc/pull/3205

AkihiroSuda:
Implemented in:
- https://github.com/opencontainers/runc/pull/3749

PR #3205, opened 2021-09-09
libct/cg: add CFS bandwidth burst for CPU

kailun-qin:
Burstable CFS controller is introduced in Linux 5.14. This helps with
parallel workloads that might be bursty. They can get throttled even
when their average utilization is under quota. And they may be latency
sensitive at the same time so that throttling them is undesired.

This feature borrows time now against the future underrun, at the cost
of increased interference against the other system users, by introducing
cfs_burst_us into CFS bandwidth control to enact the cap on unused
bandwidth accumulation, which will then used additionally for burst.

The patch adds the support/control for CFS bandwidth burst.

runtime-spec: https://github.com/opencontainers/runtime-spec/pull/1120

### DIFF
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -84,6 +84,23 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 			period = ""
 		}
 	}
+
+	var burst string
+	burst = strconv.FormatUint(r.CpuBurst, 10)
+	if err := cgroups.WriteFile(path, "cpu.cfs_burst_us", burst); err != nil {
+		// Sometimes when the burst to be set is larger
+		// than the current one, it is rejected by the kernel
+		// (EINVAL) as old_quota/new_burst exceeds the parent
+		// cgroup quota limit. If this happens and the quota is
+		// going to be set, ignore the error for now and retry
+		// after setting the quota.
+		if !errors.Is(err, unix.EINVAL) || r.CpuQuota == 0 {
+			return err
+		}
+	} else {
+		burst = ""
+	}
+
 	if r.CpuQuota != 0 {
 		if err := cgroups.WriteFile(path, "cpu.cfs_quota_us", strconv.FormatInt(r.CpuQuota, 10)); err != nil {
 			return err
@@ -93,7 +110,13 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 				return err
 			}
 		}
+		if burst != "" {
+			if err := cgroups.WriteFile(path, "cpu.cfs_burst_us", burst); err != nil {
+				return err
+			}
+		}
 	}
+
 	return s.SetRtSched(path, r)
 }
 
--- a/update.go
+++ b/update.go
@@ -44,6 +44,7 @@ The accepted format is as follow (unchanged values can be omitted):
   "cpu": {
     "shares": 0,
     "quota": 0,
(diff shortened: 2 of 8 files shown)


libcontainer/cgroups/fs/cpu.go
```diff
@@ -89,6 +89,25 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 			period = ""
 		}
 	}
+
+	var burst string
+	if r.CpuBurst != 0 {
```
kolyshkin, line 94: the spec (in https://github.com/opencontainers/runtime-spec/pull/1120) says

>    A value of `0` for indicates that the cgroup can not accumulate any unused bandwidth. It makes the traditional bandwidth control behavior unchanged.

but since we ignore `0` here, this means we can't disable burst during `runc update`.

update.go
```diff
@@ -258,30 +264,34 @@ other options are ignored.
 		// Update the values
 		config.Cgroups.Resources.BlkioWeight = *r.BlockIO.Weight
 
-		// Setting CPU quota and period independently does not make much sense,
-		// but historically runc allowed it and this needs to be supported
-		// to not break compatibility.
+		// Setting CPU quota, period and burst independently does not make much
+		// sense, but historically runc allowed it and this needs to be
+		// supported to not break compatibility.
```
kolyshkin, line 269: This is historically true for period and quota, but not for burst. I think we should require period to be set when setting burst.

libcontainer/cgroups/fs/cpu.go
kailun-qin: Good catch, thanks! Yes, I just realized after submitting this so I updated the spec accordingly to elaborate.

update.go
kailun-qin: Thanks, make sense to me. JFYI, underneath in kernel implementation, it looks like quota, period and burst are all set together. As for systemd, it's not yet supported for burst.

Zheaoli:
Is there any new update about this PR？

I think maybe I can help about this PR

kailun-qin:
> Is there any new update about this PR？
> 
> I think maybe I can help about this PR

@Zheaoli Thanks for raising this!

I'd love to move the PR forward but actually we need to have the spec (https://github.com/opencontainers/runtime-spec/pull/1120) discussed and merged first.

Besides I also received the same feature request from @litianxi who aims to apply this in their production and we discussed further offline.

@kolyshkin would you please kindly help review the runtime-spec? Thanks a lot.

Zheaoli:
After https://github.com/opencontainers/runtime-spec/pull/1120 has been completed, I think this PR is ready to go

kailun-qin:
Rebased on top of the latest main.

AkihiroSuda:
Can't be compiled
```
# github.com/opencontainers/runc/libcontainer/specconv
Error: libcontainer/specconv/spec_linux.go:740:14: r.CPU.Burst undefined (type *specs.LinuxCPU has no field or method Burst)
Error: libcontainer/specconv/spec_linux.go:741:36: r.CPU.Burst undefined (type *specs.LinuxCPU has no field or method Burst)
```

AkihiroSuda:
@kailun-qin  Could you update the PR?

Zheaoli:
@AkihiroSuda, I will take handle of this PR...

Zheaoli:
@AkihiroSuda After #3749 has been merged, I think this PR and #3145 should be closed.

PR #3206, opened 2021-09-09
Add tests for seccomp agent example

rata:
This is a follow-up for: https://github.com/opencontainers/runc/pull/2682#discussion_r703306375

Here I add a simple script to generate the config.json, use that script from tests so we verify that we don't break the user-facing example. I first modified the command the example runs (added one stat and use `&&` instead of `;` between different command), so the return code is not 0 on failures. This is useful to later use it on tests.

If you look at the config.json currently on master and the one generated by the script, be aware of using "git diff -w". Most changes are whitespaces and the seccomp section now is inserted at the end of the linux section by jq. Other than that, no real changes.


rata:
Wait, there are simple typos (x89-64 not in the update_config rule). I missed them as it was sunny and was outside :see_no_evil: . Will fix them and update here

contrib/cmd/seccompagent/gen-seccomp-example-cfg.sh
```diff
@@ -0,0 +1,35 @@
...
+# update_config() seems clearer. We don't even really need to keep them in sync.
+function update_config() {
+        jq "$1" "./config.json" | awk 'BEGIN{RS="";getline<"-";print>ARGV[1]}' "./config.json"
+}
+
+update_config '.linux.seccomp = {
+                        "defaultAction": "SCMP_ACT_ALLOW",
+                        "listenerPath": "/run/seccomp-agent.socket",
+                        "listenerMetadata": "foo",
+                        "architectures": [ "SCMP_ARCH_X86", "SCMP_ARCH_X32" ],
```
kolyshkin, line 22: So, your first commit adds `SCMP_ARCH_X86_64` to config.json, and this one removes config.json and adds a way to generate it, but it omits `SCMP_ARCH_X86_64`.

What is the purpose of the first commit then?

tests/integration/seccomp-notify.bats
```diff
@@ -198,3 +198,22 @@ function scmp_act_notify_template() {
...
+	# This takes a bare config.json and modifies to run an example.
+	"${INTEGRATION_ROOT}/../../contrib/cmd/seccompagent/gen-seccomp-example-cfg.sh"
+
+	# The listenerPath the previous command uses is the default used by the
+	# seccomp agent. However, inside bats the socket is in a bats tmp dir.
+	update_config '.linux.seccomp.listenerPath = "'"$SECCCOMP_AGENT_SOCKET"'"'
+
+	runc run test_busybox
+
+	# Exit status should be 0
```
kolyshkin, line 214: this is obvious :)

contrib/cmd/seccompagent/gen-seccomp-example-cfg.sh
rata: heh, you missed this comment: https://github.com/opencontainers/runc/pull/3206#issuecomment-916297244. It was a typo, fixed now

tests/integration/seccomp-notify.bats
rata: Right, removed it!

rata:
@kolyshkin fixed, PTAL

AkihiroSuda (approved):

contrib/cmd/seccompagent/gen-seccomp-example-cfg.sh
kolyshkin: Huh, I thought that the typo was "x89-64", but it was another typo :))))

kolyshkin (approved):
LGTM

PR #3208, opened 2021-09-09
ci/gha: update golangci-lint to 1.42.1

kolyshkin:
Update the linters, fix a minor issue found. See individual commits for details.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v2
         with:
           # must be specified without patch version
-          version: v1.40
+          version: v1.42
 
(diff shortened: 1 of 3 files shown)


kolyshkin:
It seems that new golangci-lint added newer gofumpt which now sets new style (go 1.17+) go:build tags, and thus the linter complains:

> contrib/cmd/seccompagent/seccompagent.go:1: File is not `gofumpt`-ed (gofumpt)

Fix by a separate commit (also removing `cgo` tag as it's not needed).

kolyshkin:
Added to 1.1 as this only affects CI, and it's good to have newer linters.

AkihiroSuda (approved):

cyphar (approved):
LGTM

Issue #3178, opened 2021-08-24
Limited cgroup manager usage pattern (and how to fix it)


kolyshkin:
Working on fixing #3132, I tried to use a new instance of cgroup manager (as a part of `runc run`) to check if cgroup already exists (and get all its PIDs). To my suprise, it did not work for most of cgroup managers, because `Exists()` (and `GetAllPids()`) rely on cgroup paths being known, and those paths are only initialized during `Apply()`.

In general, it's impossible to use a new instance of a cgroup manager created without known paths (as it is usually done except for `runc exec/restore/delete` for anything but `Apply()`!

This issue can be workarounded in many cases by providing an initialized set of paths to `New*Manager`, but it is not always possible.

There were attempts to partially solve it before on a case-by-case basis (see e.g. commit 4f8ccc5ff5828ec5e68eebc4ce747742f572e73c), but that was spotty/incomplete.

The fix is _relatively_ easy (but lengthy)-- we need to move paths initialization from Apply() to `New*Manager`. This also requires an ability to return an error from New*Manager, as sometimes the cgroup Name/Parent/Path provided is incorrect and we can't initialize paths.

Issue #3177, opened 2021-08-24
Complicated cgroup manager instantiation (and how to fix it)


kolyshkin:
Currently, libcontainer has 4 different cgroup managers:
* cgroup v1 cgroupfs-based (aka fs)
* cgroup v2 cgroupfs-based (aka fs2)
* cgroup v1 systemd-based (aka systemd v1)
* cgroup v2 systemd-based (aka systemd v2)

In addition, 3 of the above also implement optional "rootless" functionality,
which can be enabled during manager's instantiation.

Because of the above, there is no easy way to "get a cgroup manager",
and the code that instantiates one is usually complicated. For example,
these almost 100 lines of code:
https://github.com/opencontainers/runc/blob/f67f1efc70332b2bcea7c47c30ff6fdf7cdd6757/libcontainer/factory_linux.go#L58-L150

Also, every user of libcontainer/cgroups has to reimplement something like the above. This includes kubernetes, cadvisor, and our tests.

I propose to solve this by
1. Moving `rootless` boolean property from the cgroup manager to cgroup config.
2. Adding `Systemd` boolean property to cgroup config.
3. Introducing a new package, `libcontainer/cgroups/manager`, with the following functions:
```go 
// New returns the instance of a cgroup manager, which is chosen
// based on the local environment (whether cgroup v1 or v2 is used)
// and the config (whether config.Systemd is set or not).
func New(config *configs.Cgroup) (cgroups.Manager, error) {
       return NewWithPaths(config, nil)
}
 
// NewWithPaths is similar to New, and can be used in case cgroup paths
// are already well known, which can save some resources.
//
// For cgroup v1, the keys are controller/subsystem name, and the values
// are absolute filesystem paths to the appropriate cgroups.
//
// For cgroup v2, the only key allowed is "" (empty string), and the value
// is the unified cgroup path.
func NewWithPaths(config *configs.Cgroup, paths map[string]string) (cgroups.Manager, error)
```

With this in place, the above code (first example) is replaced with something like
```go
          cm, err := manager.New(config.Cgroups)
          if err != nil {
                  return err
          }
```

In a similar way, external users (kubernetes, cadvisor) can simplify cgroup manager instantiation.

kolyshkin:
The fix is implemented as part of PR #3131.

PR #3216, opened 2021-09-15
cgroup: make paths available, simplify getting manager

kolyshkin:
_This ~~is based on and currently includes #3215.~~ used to be part of #3131._

High level overview:

 * allow cgroup instantiation (`NewManager`) to return errors;
 * initialize cgroup paths during instantiation, so paths are always available
   (and now we can use a new instance for e.g. `Destroy` or `Exists`,
    which was not possible before);
 * simplifies cgroup manager instantiation, replacing the complicated
    logic of choosing a cgroup manager with a simple `manager.New()` call.
 * fixes some panics in cgroup manager (with `Resources==nil`).

This helps further commits, and makes cgroup manager easier to use
from e.g. Kubernetes.

Fixes: https://github.com/opencontainers/runc/issues/3177
Fixes: https://github.com/opencontainers/runc/issues/3178

## Proposed changelog entry
```
libcontainer API changes:
* New configs.Cgroup structure fields (#3177):
  * Systemd (whether to use systemd cgroup manager);
  * Rootless (whether to use rootless cgroups).
* New cgroups/manager package aiming to simplify cgroup manager instantiation. (#3177)
* All cgroup managers' instantiation methods now initialize cgroup paths and can return errors.
  This allows to use any cgroup manager method (e.g. Exists, Destroy, Set, GetStats) right after
  instantiation, which was not possible before (as paths were initialized in Apply only). (#3178)
```

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -118,8 +118,13 @@ func (m *manager) Apply(pid int) (err error) {
 		if err := sys.Apply(p, c.Resources, pid); err != nil {
 			// In the case of rootless (including euid=0 in userns), where an
 			// explicit cgroup path hasn't been set, we don't bail on error in
-			// case of permission problems. Cases where limits have been set
-			// (and we couldn't create our own cgroup) are handled by Set.
+			// case of permission problems here, but do delete the path from
+			// the m.paths map, since it is either non-existent and could not
+			// be created, or the pid could not be added to it.
+			//
+			// Cases where limits for the subsystem have been set are handled
+			// later by Set, which fails with a friendly error (see
+			// if path == "" in Set).
 			if isIgnorableError(c.Rootless, err) && c.Path == "" {
 				delete(m.paths, name)
 				continue
@@ -173,10 +178,11 @@ func (m *manager) Set(r *configs.Resources) error {
 	for _, sys := range subsystems {
 		path := m.paths[sys.Name()]
 		if err := sys.Set(path, r); err != nil {
+			// When rootless is true, errors from the device subsystem
+			// are ignored, as it is really not expected to work.
 			if m.cgroups.Rootless && sys.Name() == "devices" {
 				continue
 			}
-			// When rootless is true, errors from the device subsystem are ignored because it is really not expected to work.
 			// However, errors from other subsystems are not ignored.
 			// see @test "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error"
 			if path == "" {
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -49,18 +49,16 @@ type subsystem interface {
 }
 
 type manager struct {
-	mu       sync.Mutex
-	cgroups  *configs.Cgroup
-	rootless bool // ignore permission-related errors
-	paths    map[string]string
+	mu      sync.Mutex
+	cgroups *configs.Cgroup
+	paths   map[string]string
 }
 
-func NewManager(cg *configs.Cgroup, paths map[string]string, rootless bool) cgroups.Manager {
+func NewManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Manager, error) {
 	return &manager{
-		cgroups:  cg,
-		paths:    paths,
-		rootless: rootless,
-	}
+		cgroups: cg,
+		paths:   paths,
+	}, nil
 }
 
 // isIgnorableError returns whether err is a permission error (in the loose
@@ -121,7 +119,7 @@ func (m *manager) Apply(pid int) (err error) {
 			// explicit cgroup path hasn't been set, we don't bail on error in
 			// case of permission problems. Cases where limits have been set
 			// (and we couldn't create our own cgroup) are handled by Set.
-			if isIgnorableError(m.rootless, err) && m.cgroups.Path == "" {
+			if isIgnorableError(c.Rootless, err) && m.cgroups.Path == "" {
 				delete(m.paths, sys.Name())
 				continue
 			}
@@ -174,10 +172,10 @@ func (m *manager) Set(r *configs.Resources) error {
 	for _, sys := range subsystems {
 		path := m.paths[sys.Name()]
 		if err := sys.Set(path, r); err != nil {
-			if m.rootless && sys.Name() == "devices" {
+			if m.cgroups.Rootless && sys.Name() == "devices" {
 				continue
 			}
-			// When m.rootless is true, errors from the device subsystem are ignored because it is really not expected to work.
+			// When rootless is true, errors from the device subsystem are ignored because it is really not expected to work.
 			// However, errors from other subsystems are not ignored.
 			// see @test "runc create (rootless + limits + no cgrouppath + no permission) fails with informative error"
 			if path == "" {
@@ -249,7 +247,7 @@ func OOMKillCount(path string) (uint64, error) {
 func (m *manager) OOMKillCount() (uint64, error) {
 	c, err := OOMKillCount(m.Path("memory"))
 	// Ignore ENOENT when rootless as it couldn't create cgroup.
-	if err != nil && m.rootless && os.IsNotExist(err) {
+	if err != nil && m.cgroups.Rootless && os.IsNotExist(err) {
 		err = nil
 	}
 
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -55,6 +55,23 @@ type manager struct {
 }
 
 func NewManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Manager, error) {
+	// Some v1 controllers (cpu, cpuset, and devices) expect
+	// cgroups.Resources to not be nil in Apply.
+	if cg.Resources == nil {
+		return nil, errors.New("cgroup v1 manager needs configs.Resources to be set during manager creation")
+	}
+	if cg.Resources.Unified != nil {
+		return nil, cgroups.ErrV1NoUnified
+	}
+
+	if paths == nil {
+		var err error
+		paths, err = initPaths(cg)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &manager{
 		cgroups: cg,
 		paths:   paths,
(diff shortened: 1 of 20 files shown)


kolyshkin:
We can either merge #3215 then this, or just merge this one and close #3215. Up to you @opencontainers/runc-maintainers

kolyshkin:
Draft, pending #3215 merge

cyphar:
How many PRs until the cgroup handling is completely rewritten, it feels like this code is runc's [Ship of Theseus](https://en.wikipedia.org/wiki/Ship_of_Theseus). :wink: I'll review this once #3215 is done.

libcontainer/cgroups/fs/fs.go
```diff
@@ -121,6 +121,11 @@ func (m *manager) Apply(pid int) (err error) {
 			// case of permission problems. Cases where limits have been set
 			// (and we couldn't create our own cgroup) are handled by Set.
 			if isIgnorableError(c.Rootless, err) && c.Path == "" {
+				// Remove the path since it is either non-existent, or
+				// the pid could not be added to it. Later, Set() will
+				// fail if there are any limits specified for the
+				// subsystem, and a friendly error message is printe
```
kolyshkin, line 127: typo: `printed`

kolyshkin: Rewrote the whole comment, merging with the existing one.

libcontainer/cgroups/fs/fs.go
```diff
@@ -55,6 +55,23 @@ type manager struct {
 }
 
 func NewManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Manager, error) {
+	// Some v1 controllers (cpu, cpuset, and devices) expect
+	// cgroups.Resources to not be nil in Apply.
```
kolyshkin, line 59: I'm not 100% sure this is the right thing to do. Maybe we should not require it.

It is required because
1. cpu needs to set RT params, see https://github.com/opencontainers/runc/blob/79185bc80666a519867747dac398546da7902fda/libcontainer/cgroups/fs/cpu.go#L26-L29
2. cpuset needs to set cpus and mems in the parent cgroup (yes this is super ugly), see https://github.com/opencontainers/runc/blob/79185bc80666a519867747dac398546da7902fda/libcontainer/cgroups/fs/cpuset.go#L150-L166
3. devices need to check SkipDevices flag, see https://github.com/opencontainers/runc/blob/79185bc80666a519867747dac398546da7902fda/libcontainer/cgroups/fs/devices.go#L24

Now, we can modify all the three places to allow `nil` (and assume RT, cpus/mems, and SkipDevices are not set), and drop this requirement. After all, it is only for some specific cases, and not always required.

Honestly I don't know which way is better.

libcontainer/cgroups/manager/manager_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/opencontainers/runc/libcontainer/configs"
+)
+
+// TestNilResources checks that a cgroup manager do not panic when
+// config.Resources is nil. While it does not make sense to use a
+// manager with no resources, it should not result in a panic.
+//
+// This tests either v1 or v2 managers (both fs and systemd)
+// depending on what cgroup version is available on the host.
+
```
kolyshkin, line 15: extra vertical whitespace needs to be removed

libcontainer/cgroups/manager/new.go
```diff
@@ -0,0 +1,78 @@
...
+	return NewWithPaths(config, nil)
+}
+
+// NewWithPaths is similar to New, and can be used in case cgroup paths
+// are already well known, which can save some resources.
+//
+// For cgroup v1, the keys are controller/subsystem name, and the values
+// are absolute filesystem paths to the appropriate cgroups.
+//
+// For cgroup v2, the only key allowed is "" (empty string), and the value
```
mrunalp, line 28: nit: Instead of `""`, could we use a predefined string like `"unified"`?

libcontainer/cgroups/fs/fs.go
mrunalp: I think that creating a cgroup should be separated from the applying values, so my preference is to allow nil at those 3 places and remove this check.

libcontainer/cgroups/manager/new.go
kolyshkin: I seriously considered doing it (about a year ago) and ultimately decided against it, mostly because changing it requires adding some backward-compatible hacks (this key and value is saved into `state.json`). I am not even taking external users such as kubernetes and cadvisor into account -- with those it will be a bloody mess.

Note that this is the convention that is universally used in many places in runc, I am not introducing it here, merely documenting the fact (which is also documented in some other places, for example https://github.com/opencontainers/runc/blob/147ad561e8be8335723f6b7a1c2e2b74f69613b0/libcontainer/cgroups/cgroups.go#L38-L46)

libcontainer/cgroups/fs/fs.go
kolyshkin: > creating a cgroup should be separated from the applying values

Alas, with cgroup v1 it is not possible in some cases (outlined above), so as much as we want to separate create and set, we can not.

In this case, though, we can allow `nil` (the price to pay is to fail later on Apply in case either of CPU RT priority, cpus/mems, SkipDevices is set).

libcontainer/cgroups/systemd/v1.go
```diff
@@ -28,6 +28,16 @@ func NewLegacyManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Mana
 	if cg.Rootless {
 		return nil, errors.New("can't use rootless systemd cgroups manager on cgroup v1")
 	}
+	if cg.Resources != nil && cg.Resources.Unified != nil {
+		return nil, cgroups.ErrV1NoUnified
+	}
+	if paths == nil {
```
thaJeztah, line 34: Looking for potential corner-cases: is an empty `map[string]string` to be considered different for this code path, or should it be treated the same? (if "the same", we'd have to check for `len(paths)`)

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -129,6 +129,9 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 }
 
 func (m *manager) Freeze(state configs.FreezerState) error {
+	if m.config.Resources == nil {
```
thaJeztah, line 132: See my comment on one of the earlier commits (should we guard against `m.config` being able to be `nil`?)

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -20,16 +20,12 @@ type manager struct {
...
 	// excludes pseudo-controllers ("devices" and "freezer").
 	controllers map[string]struct{}
-	rootless    bool
 }
 
 // NewManager creates a manager for cgroup v2 unified hierarchy.
 // dirPath is like "/sys/fs/cgroup/user.slice/user-1001.slice/session-1.scope".
 // If dirPath is empty, it is automatically set using config.
-func NewManager(config *configs.Cgroup, dirPath string, rootless bool) (cgroups.Manager, error) {
-	if config == nil {
```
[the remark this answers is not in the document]
kolyshkin: One "interesting" side effect of such embedding is `Lock()`/`Unlock()` methods and `Cgroup` property becomes public. I don't think this is OK for Cgroup, and it's a bit questionable for mutex as well.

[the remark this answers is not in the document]
thaJeztah: Yeah, I mentioned the "implicitly export the config (likely not desirable" bit, and agree that (at least that bit) we likely don't want. The mutex is a bit more "grey area" (but looking at my changes, it's not cleaning up things _very much_, so don't think it's worth doing).

libcontainer/cgroups/systemd/v1.go
```diff
@@ -24,12 +24,15 @@ type legacyManager struct {
 	dbus    *dbusConnManager
 }
 
-func NewLegacyManager(cg *configs.Cgroup, paths map[string]string) cgroups.Manager {
+func NewLegacyManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Manager, error) {
+	if cg.Rootless {
+		return nil, errors.New("can't use rootless systemd cgroups manager on cgroup v1")
```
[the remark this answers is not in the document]
kolyshkin: I think we use `unable to` in many such cases. `can not` or `cannot` is fine, too.

Current usage in runc codebase is (excluding vendor, including comments):
* cannot: 98
* can not: 2
* can't: 61
* unable to: 93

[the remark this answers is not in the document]
thaJeztah: I usually pick "unable to" for either errors on "our" side and/or "non fatal" situations (for more "fatal" errors, I pick `failed to`).

In this case  it looks like a "fatal" error, but not because "we (runc)" made a whoopsie?

I'm fine with either; thanks for searching those; looks like some low-hanging fruit to make it more consistent for a better user experience.

libcontainer/cgroups/fs2/fs2.go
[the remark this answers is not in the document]
kolyshkin: As for adding the check for `nil` (and erroring out) here, and removing the `nil` check from other places, this makes sense (but will increase the patchset size a bit more).

Ah! I see this is already done in `cgroups/manager` package!  I am not quite sure we need to repeat this in every cgroup manager though, as users are supposed to switch to cgroups/manager.New as the entry point to create a manager instance.

So, what's left is removing `m.config == nil` checks for individual managers. Done.

libcontainer/cgroups/fs/fs.go
kolyshkin: OTOH it's easier to require non-nil Resources here, otherwise it's hard to explain later why cpu/cpuset/devices configuration has unexpectedly failed.

libcontainer/cgroups/systemd/v1.go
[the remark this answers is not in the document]
kolyshkin: Changed `can't` to `cannot` in two cases (first commit in this PR).

libcontainer/cgroups/fs2/fs2.go
kolyshkin: I have removed checks for `m.config` being nil (see the last commit), as this is checked in `cgroups/manager.New`, and all the other code assumes that `m.config` (called `m.cgroups` in all the other managers except fs2) is non-nil.

kolyshkin:
Implemented some very minor changes (`can't` -> `cannot` plus the last commit) based on @thaJeztah review.

libcontainer/cgroups/fs/fs.go
kolyshkin: Yes, let's leave it as is, unless there are other opinions.

thaJeztah (approved):
LGTM, thanks!

mrunalp (approved):

kolyshkin:
@AkihiroSuda PTAL (this is merged but if there's anything you feel is wrong here, we can address in a followup)

libcontainer/cgroups/fs/fs.go
cyphar: I think it's better to leave it as-is for the moment. The fact that we need to configure other cgroups for cpu makes me think that allowing nil Resources will make this even more ugly than it is today.

cyphar:
FWIW, LGTM.

libcontainer/cgroups/systemd/v1.go
kolyshkin: Forget to reply -- I thought about that and I guess the non-nil map (even when empty) should be handled as it is right now, IOW it means "paths are set, no need to guess any". Not sure if/when this can be useful but theoretically it kind of makes sense.

Issue #3134, opened 2021-08-06
runc delete -f fails to remove a paused container on cgroup v1


kolyshkin:
Cgroup v2:

```console
[root@kir-rhat runc-tst]# ../runc/runc run -d 345
[root@kir-rhat runc-tst]# ../runc/runc pause 345
[root@kir-rhat runc-tst]# time ../runc/runc delete -f 345

real	0m0.129s
user	0m0.010s
sys	0m0.008s
[root@kir-rhat runc-tst]# ../runc/runc list
ID          PID         STATUS      BUNDLE      CREATED     OWNER
```

Cgroup v1:

```console
root@ubu2004:/home/kir/git/runc/tst# ../runc run -d 345
root@ubu2004:/home/kir/git/runc/tst# ../runc pause 345
root@ubu2004:/home/kir/git/runc/tst# time ../runc delete -f 345
ERRO[0010] container init still running                 

real	0m10.090s
user	0m0.009s
sys	0m0.089s
root@ubu2004:/home/kir/git/runc/tst# ../runc list
ID          PID         STATUS      BUNDLE                   CREATED                          OWNER
345         229339      paused      /home/kir/git/runc/tst   2021-08-06T18:57:03.953021519Z   root
```

This happens because in cgroup v2, "[p]rocesses in the frozen cgroup can be killed by a fatal signal" (https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html), and it's not so for cgroup v1.

Most interesting thing is, `*linuxContainer) Destroy()` thaws a container if it is paused, but we don't call it until we kill init process in the `killContainer` (in delete.go) only calls it when the process is gone, which never happens.

kolyshkin:
Just checked that crun is not affected.

kolyshkin:
Fixed by #3217

PR #3217, opened 2021-09-15
runc delete -f: fix for cg v1 + paused container

kolyshkin:
runc delete -f is not working for a paused container, since in cgroup v1
SIGKILL does nothing if a process is frozen (unlike cgroup v2, in which
you can kill a frozen process with a fatal signal).

Theoretically, we only need this for v1, but doing it for v2 as well is
OK.

## Proposed changelog entry
```
Bugfixes:
* runc delete -f now succeeds (rather than timeouts) on a paused container (#3134)
```

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -393,6 +393,14 @@ func (c *linuxContainer) Signal(s os.Signal, all bool) error {
 		if err := c.initProcess.signal(s); err != nil {
 			return fmt.Errorf("unable to signal init: %w", err)
 		}
+		if status == Paused {
+			// For cgroup v1, killing a process in a frozen cgroup
+			// does nothing until it's thawed. Only thaw the cgroup
+			// for SIGKILL.
+			if s, ok := s.(unix.Signal); ok && s == unix.SIGKILL {
+				_ = c.cgroupManager.Freeze(configs.Thawed)
+			}
+		}
 		return nil
 	}
 	return ErrNotRunning
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -50,6 +50,16 @@ function teardown() {
 	[ "$status" -eq 0 ]
 }
 
+@test "runc delete --force [paused container]" {
+	runc run -d --console-socket "$CONSOLE_SOCKET" ct1
+	[ "$status" -eq 0 ]
+	testcontainer ct1 running
+
+	runc pause ct1
+	runc delete --force ct1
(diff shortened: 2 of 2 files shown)


thaJeztah (approved):
LGTM

libcontainer/container_linux.go
```diff
@@ -393,6 +393,14 @@ func (c *linuxContainer) Signal(s os.Signal, all bool) error {
...
+			// For cgroup v1, killing a process in a frozen cgroup
+			// does nothing until it's thawed. Only thaw the cgroup
+			// for SIGKILL.
+			if s, ok := s.(unix.Signal); ok && s == unix.SIGKILL {
+				_ = c.cgroupManager.Freeze(configs.Thawed)
+			}
+		}
 		return nil
 	}
 	return ErrNotRunning
```
thaJeztah, line 406: Haven't tried, but what happens currently if `-f` is used and it's not running? (thinking: `-f` should ignore the case and just proceed?) I see we have a special case for `all`

cyphar: The logic for not allowing `kill` even if you force it is that on paper you could trick runc to kill an arbitrary process. I think an actual attack that uses this is a bit theoretical but *shrug*.

cyphar (approved):
LGTM.

libcontainer/container_linux.go
```diff
@@ -393,6 +393,14 @@ func (c *linuxContainer) Signal(s os.Signal, all bool) error {
 		if err := c.initProcess.signal(s); err != nil {
 			return fmt.Errorf("unable to signal init: %w", err)
 		}
+		if status == Paused {
+			// For cgroup v1, killing a process in a frozen cgroup
+			// does nothing until it's thawed. Only thaw the cgroup
+			// for SIGKILL.
+			if s, ok := s.(unix.Signal); ok && s == unix.SIGKILL {
```
cyphar, line 400: Would've been nice to check if the cgroup manager is v1 or v2 here, but still LGTM.

kolyshkin: We are going to use cgroup.kill for v2 and hybrid hierarchies, this code is just to fix the issue

kolyshkin:
Fixes: #3134

libcontainer/container_linux.go
cyphar: Right, I get that -- my point is more that it would've been nice if we didn't thaw the cgroup unless we really had to.

PR #3222, opened 2021-09-23
CI/GHA: switch to OBS criu repo, bring in criu 3.16

kolyshkin:
Apparently (https://criu.org/index.php?title=Packages&diff=5197&oldid=5196) criu PPA repo, used for GHA CI, is replaced with the OBS repo (already used by Dockerfile), so let's use it.

While at it, modify the Dockefile to
 * update the repo URL to use Debian_11 (which was not available before);
 * fix the apt-key deprecation warning from Dockerfile.

See individual commits for more details.

This brings criu 3.16 to the table! 🎉

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,11 +37,6 @@ jobs:
     - name: checkout
       uses: actions/checkout@v2
 
-    - name: Dump matrix context
-      env:
-        MATRIX_CONTEXT: ${{ toJSON(matrix) }}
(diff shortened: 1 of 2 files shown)


kolyshkin:
This is CI only (i.e. no changes other than CI) and is needed
 - to reduce the number of criu-related failures in CI
 - for https://github.com/opencontainers/runc/pull/3059 (which, in turn, is needed for #3199)

mrunalp (approved):

cyphar (approved):
LGTM.

PR #3224, opened 2021-09-26
Allow /proc/sys/net to be mounted read-write if container is in not-host netns

### LINKED ISSUE #2826: checkProcMount() is too strict
TL;DR I'm trying to run a libvirt VM inside of a docker container without using `--privileged`.

The specific error I get is:
```
libvirt.libvirtError: cannot write to /proc/sys/net/ipv6/conf/virbr2/disable_ipv6 to enable/disable IPv6 on bridge virbr2: Read-only file system
```

What I would like to do is allow `/proc/sys/net` to be mounted read-write inside of the container. My understanding is that this is safe because everything that a container could touch there is net-namespaced, and won't affect the host net ns. (I would have to audit some kernel code to be sure, but it's certainly better than `--privileged`).

The problem is that `checkProcMount()` won't let me:

```
$ docker version
 Version:           20.10.3
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        48d30b5
 Built:             Fri Jan 29 14:33:25 2021
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.3
  Go version:       go1.13.15
  Git commit:       46229ca
  Built:            Fri Jan 29 14:31:38 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0


$ docker run --rm -it -v "/proc/sys/net:/proc/sys/net:rw" debian:10
docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:59: mounting "/proc/sys/net" to rootfs at "/proc/sys/net" caused: "/var/lib/docker/overlay2/9fd477a20091dd5d9babf5ce2bddb8d517349c89ba9a4e6c7f74f275f0c370c9/merged/proc/sys/net" cannot be mounted because it is inside /proc: unknown.
```

JonathonReinhart:
WIP: This branch has only been build-tested.

Closes #2826


libcontainer/container_linux.go
```diff
@@ -1190,7 +1190,7 @@ func (c *linuxContainer) makeCriuRestoreMountpoints(m *configs.Mount) error {
 	case "bind":
 		// The prepareBindMount() function checks if source
 		// exists. So it cannot be used for other filesystem types.
-		if err := prepareBindMount(m, c.config.Rootfs); err != nil {
+		if err := prepareBindMount(m, c.config.Rootfs, config); err != nil {
```
kolyshkin, line 1193: I'd rather provide it `c.config` instead.

libcontainer/container_linux.go
```diff
@@ -1199,7 +1199,7 @@ func (c *linuxContainer) makeCriuRestoreMountpoints(m *configs.Mount) error {
 		if err != nil {
 			return err
 		}
-		if err := checkProcMount(c.config.Rootfs, dest, ""); err != nil {
+		if err := checkProcMount(c.config.Rootfs, dest, "", c.config); err != nil {
```
kolyshkin, line 1202: ditto (maybe use singe `c.config` instead of two arguments).

kolyshkin: or make checkProcMount a method of `linuxContainer`

libcontainer/rootfs_linux.go
```diff
@@ -518,7 +518,7 @@ func getCgroupMounts(m *configs.Mount) ([]*configs.Mount, error) {
 // dest is required to be an abs path and have any symlinks resolved before calling this function.
 //
 // if source is nil, don't stat the filesystem.  This is used for restore of a checkpoint.
-func checkProcMount(rootfs, dest, source string) error {
+func checkProcMount(rootfs, dest, source string, config *configs.Config) error {
```
kolyshkin, line 521: Again, this can be
```go
func (c *linucContainer) checkProcMount(rootfs, dest, source string) error {
config := c.config
```
or something similar

kolyshkin:
@JonathonReinhart I left some review comments above; PTAL

libcontainer/container_linux.go
```diff
@@ -1176,7 +1176,7 @@ func (c *linuxContainer) restoreNetwork(req *criurpc.CriuReq, criuOpts *CriuOpts
 // makeCriuRestoreMountpoints makes the actual mountpoints for the
 // restore using CRIU. This function is inspired from the code in
 // rootfs_linux.go
-func (c *linuxContainer) makeCriuRestoreMountpoints(m *configs.Mount) error {
+func (c *linuxContainer) makeCriuRestoreMountpoints(m *configs.Mount, config *configs.Config) error {
```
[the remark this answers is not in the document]
JonathonReinhart: D'oh -- I entirely missed that `linuxContainer` already has `confg`. This will greatly simplify the diff. I'll fix this everywhere.

libcontainer/container_linux.go
JonathonReinhart: I considered moving `checkProcMount` to `container_linux.go` and making it a method of `linuxContainer` but that doesn't seem to be feasible without moving a bunch of `rootfs_linux.go`, too:

[](https://mermaid.live/edit#eyJjb2RlIjoiJSUgaHR0cHM6Ly9naXRodWIuY29tL29wZW5jb250YWluZXJzL3J1bmMvcHVsbC8zMjI0XG4lJSBodHRwczovL2dpdGh1Yi5jb20vb3BlbmNvbnRhaW5lcnMvcnVuYy90cmVlL2QxYzliNDNlL2xpYmNvbnRhaW5lclxuZ3JhcGggVERcblxuICBzdWJncmFwaCByb290ZnNfbGludXguZ29cbiAgICBwcmVwYXJlQmluZE1vdW50XG4gICAgY2hlY2tQcm9jTW91bnRcbiAgICBtb3VudFRvUm9vdGZzXG5cbiAgICBwcmVwYXJlQmluZE1vdW50IC0tPiBjaGVja1Byb2NNb3VudFxuICAgIG1vdW50VG9Sb290ZnMgLS0-fGNhc2UgYmluZHwgcHJlcGFyZUJpbmRNb3VudFxuICAgIG1vdW50VG9Sb290ZnMgLS0-fGNhc2UgY2dyb3VwfCBtb3VudENncm91cFYxXG4gICAgbW91bnRUb1Jvb3RmcyAtLT58ZGVmYXVsdCBjYXNlfCBjaGVja1Byb2NNb3VudFxuXG4gICAgcHJlcGFyZVJvb3RmcyAtLT58Zm9yIGVhY2ggbW91bnR8IG1vdW50VG9Sb290ZnNcbiAgICBtb3VudENncm91cFYxIC0tPiBtb3VudFRvUm9vdGZzXG4gICAgbW91bnRDZ3JvdXBWMSAtLT58Zm9yIGVhY2ggYmluZHwgbW91bnRUb1Jvb3Rmc1xuICBlbmRcblxuICBzdWJncmFwaCBjb250YWluZXJfbGludXguZ29cbiAgICBtYWtlQ3JpdVJlc3RvcmVNb3VudHBvaW50c1xuXG4gICAgbWFrZUNyaXVSZXN0b3JlTW91bnRwb2ludHMgLS0-fGNhc2UgYmluZHwgcHJlcGFyZUJpbmRNb3VudFxuICAgIG1ha2VDcml1UmVzdG9yZU1vdW50cG9pbnRzIC0tPnxkZWZhdWx0IGNhc2V8IGNoZWNrUHJvY01vdW50XG4gIGVuZCIsIm1lcm1haWQiOiJ7XG4gIFwidGhlbWVcIjogXCJkZWZhdWx0XCJcbn0iLCJ1cGRhdGVFZGl0b3IiOmZhbHNlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6ZmFsc2V9)

libcontainer/rootfs_linux.go
JonathonReinhart: See https://github.com/opencontainers/runc/pull/3224#discussion_r720606929

libcontainer/container_linux.go
kolyshkin: Yeah, it's a mess :(

If you can figure out a good way to refactor it, I'm all for it.

Issue #3225, opened 2021-09-27
When rootless runc exits, failed to remove cgroupv2 with sub-cgroup


chenk008:
In case of rootless, cgroup2 mount into container `/sys/fs/cgroup`. After I created sub-cgroup `test`, and then exited. A error message shown
```
ERRO[0029] remove /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/user.slice/runc-foo.scope: device or resource busy
```

Reproduce:

```
runc spec --rootless
// add cgroupfs mount into config.json
{
      "destination": "/sys/fs/cgroup",
      "type": "cgroup",
      "source": "cgroup",
      "options": [
        "nosuid",
        "noexec",
        "nodev",
        "relatime"
      ]
}
runc --systemd-cgroup run foo
// in the container
mkdir /sys/fs/cgroup/test
exit
```

systemd version: 239 (239-41.el8_3)
kernel version: 4.18.0-305.12.1.el8_4.x86_64
linux distro: centos 8

chenk008:
I think it relates to https://github.com/opencontainers/runc/blob/e999e29a288630b5dfbfd8bc789975e5b94e6102/libcontainer/cgroups/systemd/v2.go#L305

kolyshkin:
Ah, so you're using cgroup v2 on CentOS 8. We do not test this configuration, this is why we do not catch this bug.

PR #3226, opened 2021-09-27
Remove sub cgroup when container exits

chenk008:
Delete all child cgroup in container cgroup, otherwise it will failed to delete the container cgroup.

Close https://github.com/opencontainers/runc/issues/3225

Signed-off-by: Kang Chen <kongchen28@gmail.com>

### Proposed changelog entry

(by @kolyshkin)

```
Bugfixes:
* cgroup v2 systemd manager: delete sub-cgroups on cgroup delete
```

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -3,7 +3,6 @@ package systemd
 import (
 	"fmt"
 	"math"
-	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -301,9 +300,10 @@ func (m *unifiedManager) Destroy() error {
 		return err
 	}
 
-	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
-	if err != nil && !os.IsNotExist(err) {
+	// systemd 239 do not remove sub-cgroups.
+	err := m.fsMgr.Destroy()
+	// fsMgr.Destroy has handled ErrNotExist
+	if err != nil {
 		return err
 	}
 


libcontainer/cgroups/systemd/v2.go
```diff
@@ -302,7 +302,8 @@ func (m *unifiedManager) Destroy() error {
 	}
 
 	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
+	logrus.Warnf("try to delete %s", m.path)
```
AkihiroSuda, line 305: Debugf

chenk008: Sorry for my mistake, I have modified it.

AkihiroSuda:
Please squash commits and use real name for signing.
LGTM then.

chenk008:
@AkihiroSuda Thanks! I have squashed commits and signed with real name.

AkihiroSuda (dismissed):

kolyshkin (changes requested):
NACK.

1. `os.RemoveAll` tries to remove all the files, too, which is definitely not what's needed in this case. For such specific case (need to remove all directories recursively but not files) we have `cgroups.RemovePath()` function, which fs2 driver uses. I'd suggest to merely call `m.fsMgr.Destroy()` here, it does exactly what's needed, and other methods similarly rely on fsMgr.

2. We already have tests for cgroup v1 and v2 removal with sub-cgroups (see `tests/integration/delete.bats`). Need to explain why aren't they failing with the existing code.

chenk008:
@kolyshkin Thanks for your suggestion. I have modified it, PLTA.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -302,8 +301,10 @@ func (m *unifiedManager) Destroy() error {
 	}
 
 	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
-	if err != nil && !os.IsNotExist(err) {
+	logrus.Warnf("try to destroy cgroup: %s", m.path)
```
kolyshkin, line 304: Why is this a warning? Why log here at all?

kolyshkin:
> We already have tests for cgroup v1 and v2 removal with sub-cgroups (see `tests/integration/delete.bats`). Need to explain why aren't they failing with the existing code.

The question still remains -- why this test is not failing without this patch?

https://github.com/opencontainers/runc/blob/e999e29a288630b5dfbfd8bc789975e5b94e6102/tests/integration/delete.bats#L113-L158

Or, to rephrase the question -- can you provide us with a repro to see what issue this PR fixes?

kolyshkin:
> > We already have tests for cgroup v1 and v2 removal with sub-cgroups (see `tests/integration/delete.bats`). Need to explain why aren't they failing with the existing code.
> 
> The question still remains -- why this test is not failing without this patch?

In my case (Fedora 34, systemd 248 (v248.7-1.fc34)) systemd is removing the cgroup together with the children.

Maybe there's something that prevents system from removing the sub-cgroups. I wonder what is it.

libcontainer/cgroups/systemd/v2.go
chenk008: Sorry for commit mistake. I changed it to Debugf

chenk008:
@kolyshkin  This is my issue: https://github.com/opencontainers/runc/issues/3225

chenk008:
I think it is related to kernel code https://elixir.bootlin.com/linux/v4.18.20/source/kernel/cgroup/cgroup.c#L5125

chenk008:
@kolyshkin I have mocked the test on my computer, it seems the error message still shown, but the cgroup will be deleted.

terminal A:
```
[vagrant@localhost rootless]$ ./runc.amd64 --systemd-cgroup  run foo
/ # mkdir /sys/fs/cgroup/test
```

terminal B:
```
runc delete --force foo
```

After deleted, terminal A shows:
```
/ # ERRO[0044] remove /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/user.slice/runc-foo.scope: device or resource busy
[vagrant@localhost rootless]$ ll /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/user.slice/runc-foo.scope
ls: cannot access '/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/user.slice/runc-foo.scope': No such file or directory
```

And the bpf shows:
```
Attaching 6 probes...
Tracing cgroup remove events... Hit Ctrl-C to end.
TIME     Action
14:36:41 delete: runc-foo.scope cg: 0xffff9b4cb859d000
14:36:41 delete: 0xffff9b4cb859d000
14:36:41 check: 0xffff9b4cb859d000
14:36:41 css_has_online_children returned: 1
14:36:41 cgroup_destroy_locked returned: -16
14:36:42 delete: runc-foo.scope cg: 0xffff9b4cb859d000
14:36:42 delete: 0xffff9b4cb859d000
14:36:42 check: 0xffff9b4cb859d000
14:36:42 css_has_online_children returned: 1
14:36:42 cgroup_destroy_locked returned: -16
14:36:42 delete: test cg: 0xffff9b4d05fd3000
14:36:42 delete: 0xffff9b4d05fd3000
14:36:42 check: 0xffff9b4d05fd3000
14:36:42 css_has_online_children returned: 0
14:36:42 cgroup_destroy_locked returned: 0
14:36:42 delete: runc-foo.scope cg: 0xffff9b4cb859d000
14:36:42 delete: 0xffff9b4cb859d000
14:36:42 check: 0xffff9b4cb859d000
14:36:42 css_has_online_children returned: 0
14:36:42 cgroup_destroy_locked returned: 0
```

And I hooked the `fs2.(*manager).Destroy`, got the backtrace. The `linuxContainer` has a cgroupv2 manager which remove all directories recursively.
```
goroutine 1 [running]:
github.com/opencontainers/runc/libcontainer/cgroups/fs2.(*manager).Destroy(0xc0002b0720, 0x555b85e3b086, 0x6)
	github.com/opencontainers/runc/libcontainer/cgroups/fs2/fs2.go:141 +0x3b
github.com/opencontainers/runc/libcontainer.destroy(0xc000210400, 0xc000206700, 0x9)
	github.com/opencontainers/runc/libcontainer/state_linux.go:47 +0x83
github.com/opencontainers/runc/libcontainer.(*stoppedState).destroy(0xc000206780, 0x555b861ada40, 0xc000279c00)
	github.com/opencontainers/runc/libcontainer/state_linux.go:104 +0x30
github.com/opencontainers/runc/libcontainer.(*linuxContainer).Destroy(0xc000210400, 0x0, 0x0)
	github.com/opencontainers/runc/libcontainer/container_linux.go:639 +0x88
main.destroy(0x555b8626aec0, 0xc000210400)
	github.com/opencontainers/runc/utils_linux.go:143 +0x37
main.killContainer(0x555b8626aec0, 0xc000210400, 0x0, 0x0)
	github.com/opencontainers/runc/delete.go:23 +0xd0
main.glob..func3(0xc000238160, 0x0, 0xc00020ccf0)
	github.com/opencontainers/runc/delete.go:82 +0x3a6
github.com/urfave/cli.HandleAction(0x555b86179880, 0x555b86248f50, 0xc000238160, 0xc000238160, 0x0)
	github.com/urfave/cli@v1.22.1/app.go:523 +0x11c
github.com/urfave/cli.Command.Run(0x555b85e3b518, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x555b85e5bced, 0x4d, 0x0, ...)
	github.com/urfave/cli@v1.22.1/command.go:174 +0x570
github.com/urfave/cli.(*App).Run(0xc00027a000, 0xc000020080, 0x4, 0x4, 0x0, 0x0)
	github.com/urfave/cli@v1.22.1/app.go:276 +0x7b0
main.main()
	github.com/opencontainers/runc/main.go:163 +0xbda
```

kolyshkin:
Ah! So you're using rootless, and the test I referred to requires root (for some reason -- I think this req can be dropped, see #3229).

We still can't catch this bug as you are using centos 8 + cgroup v2, a combo we do not test.

The patch makes sense though. Can you please improve the commit subject and add the description?

kolyshkin:
The commit message subject should be something like `libct/cg/sd/v2: Destroy: remove cgroups recursively`

The commit message itself should link to the bug report, and briefly describe what are you fixing.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -302,8 +301,10 @@ func (m *unifiedManager) Destroy() error {
 	}
 
 	// XXX this is probably not needed, systemd should handle it
```
kolyshkin, line 303: This comment needs to be removed. Instead, maybe add that systemd vX.X.X (the one you used on CentOS 8) do not remove sub-cgroups.

chenk008:
@AkihiroSuda I have modified the commit message and comment, PTAL. Thanks!

kolyshkin (dismissed):
LGTM, and thanks!

(not sure why DCO check complains -- seems the commit is properly signed)

kolyshkin:

AkihiroSuda:
```
Commit sha: 9c5fe74, Author: wuhua.ck, Committer: wuhua.ck; Expected "wuhua.ck wuhua.ck@alibaba-inc.com", but got "Kang Chen kongchen28@gmail.com".
```

https://github.com/opencontainers/runc/pull/3226/checks?check_run_id=3739410743

You need `git config user.name Kang Chen`

libcontainer/cgroups/systemd/v2.go
```diff
@@ -301,9 +300,11 @@ func (m *unifiedManager) Destroy() error {
 		return err
 	}
 
-	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
-	if err != nil && !os.IsNotExist(err) {
+	// systemd 239 do not remove sub-cgroups.
+	logrus.Debugf("try to destroy cgroup: %s", m.path)
```
thaJeztah, line 304: I saw there was a comment about this earlier; mostly wondering how useful this debug log is, as it would be logging the "happy" path as well (successfully handled this); the "failure" path is already returned as an error (which may be more interesting in case things don't work as expected).

Is there a strong motivation to log this case?

chenk008: Thanks for your suggestion. When I met the error message, I want to make sure if this code is the root cause. I have deleted the debug log.

thaJeztah: Thanks! If the error message itself is not providing enough information, we could add a (debug) log _inside_ the `if err != nil`. I _suspect_ the error will contain information about the path that failed to be removed though (so perhaps not needed to add additional logs for that)

thaJeztah (approved):
LGTM

kolyshkin (approved):
LGTM

PR #3228, opened 2021-09-27
contrib: rm init from bash completion

kolyshkin:
Commit 7a0302f0d79 already removed "runc init" from runc help output,
as this is an internal option not supposed to be used by the end user.

Let's remove runc init completion, too.


AkihiroSuda (approved):

cyphar (approved):
LGTM.

PR #3229, opened 2021-09-28
tests/int/delete: drop root req from v2 subcgroup test

kolyshkin:
"runc delete --force in cgroupv2 with subcgroups" can be run as
non-root, too, and might reveal a bug in systemd + centos 8.

Related to #3225/#3226.

### DIFF
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -111,7 +111,8 @@ EOF
 }
 
 @test "runc delete --force in cgroupv2 with subcgroups" {
-	requires cgroups_v2 root
+	requires cgroups_v2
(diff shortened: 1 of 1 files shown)


kolyshkin:
CI failures are infra flakes such as:

> Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libs/libseccomp/libseccomp-dev_2.5.1-1ubuntu1~20.04.1_amd64.deb  Could not connect to azure.archive.ubuntu.com:80 (52.147.219.192), connection timed out

Alas we still can't catch the #3225 bug as it's using CentOS 8 + cgroup v2, a combo we do not test.

PR #3230, opened 2021-09-29
Dockerfile: fix for seccomp

kolyshkin:
Commit f30244ee1b22223cf (PR #3197) broke the scenario of using Dockefile for
anything but making a release. This happened because it installed
native libseccomp build to a temporary directory, and so linking against
libseccomp required setting a few environment variables.

Let's fix this, and simplify libseccomp installation. Instead of using
temporary directories, let's install native libseccomp to a specified
directory, all the cross-builds to its subdirectories, and set
PKG_CONFIG_PATH and LD_LIBRARY_PATH in Dockerfile so that the built
library will found by pkg-config and the dynamic linker (without setting
LD_LIBRARY_PATH, ld picks up distro-provided libseccomp.so).

While at it, fix some nit picks introduced by the abovementioned commit.

This fixes building runc in  make targets like shell, dbuild,
integration, unittest -- i.e. those that depend on runcimage.

Fixes: f30244ee1b22223cf38b505b42

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN KEYFILE=/usr/share/keyrings/criu-repo-keyring.gpg; \
         crossbuild-essential-armel \
         crossbuild-essential-armhf \
         crossbuild-essential-ppc64el \
+        crossbuild-essential-s390x \
         curl \
         gawk \
         gcc \
@@ -54,7 +55,7 @@ RUN cd /tmp \
 ARG LIBSECCOMP_VERSION
 COPY script/* /tmp/script/
 RUN mkdir -p /opt/libseccomp \
-    && /tmp/script/seccomp.sh "$LIBSECCOMP_VERSION" /opt/libseccomp arm64 armel armhf ppc64le
+    && /tmp/script/seccomp.sh "$LIBSECCOMP_VERSION" /opt/libseccomp arm64 armel armhf ppc64le s390x
 ENV LIBSECCOMP_VERSION=$LIBSECCOMP_VERSION
 ENV LD_LIBRARY_PATH=/opt/libseccomp/lib
 ENV PKG_CONFIG_PATH=/opt/libseccomp/lib/pkgconfig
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -53,7 +53,9 @@ RUN cd /tmp \
 # install libseccomp
 ARG LIBSECCOMP_VERSION
 COPY script/* /tmp/script/
-RUN mkdir -p /usr/local/src/libseccomp \
-    && /tmp/script/seccomp.sh "$LIBSECCOMP_VERSION" /usr/local/src/libseccomp /usr/local/src/libseccomp/.env-file arm64 armel armhf ppc64le
+RUN mkdir -p /opt/libseccomp \
+    && /tmp/script/seccomp.sh "$LIBSECCOMP_VERSION" /opt/libseccomp arm64 armel armhf ppc64le
```
thaJeztah, line 57: Not new in this PR, but wondering if it should have s390x as well?

Dockerfile
```diff
@@ -53,7 +53,9 @@ RUN cd /tmp \
 # install libseccomp
 ARG LIBSECCOMP_VERSION
```
thaJeztah, line 54: It's a bit unfortunate that both the Dockerfile and script/release.sh now contain the version to use (was wondering if we had to persist this ARG as an ENV (`ENV LIBSECCOMP_VERSION=$LIBSECCOMP_VERSION`), and update the script to take that into account as a way to override the version? 🤔 


(Perhaps same for the source path), e.g. something like (haven't checked if I have the syntax correct)

```bash
local libseccomp_version="${LIBSECCOMP_VERSION-2.5.2)}"
local seccompdir ="${LIBSECCOMP_SRC-$(mktemp -d)}"
```

Dockerfile
kolyshkin: To be addressed separately.

kolyshkin: Added; let me know if you want a separate PR for that.

kolyshkin:
@thaJeztah PTAL

AkihiroSuda (approved):

thaJeztah (approved):
LGTM, thanks!

PR #3231, opened 2021-10-04
fix createDevices when no Linux section

najohnsn:
In `spec_linux.go`, `createDevices` may fail to validate `spec.Linux` is not `nil` before using it. It _does_ validate `spec.Linux` before the 2nd phase of processing. But when processing a device in `AllowedDevices` that has a `Path` in the first phase, `spec.Linux` is not validated. This causes a nil pointer reference.

Fixed by validating `spec.Linux` before attempting to use `spec.Linux.Devices`.

Created a `TestSpecconvNoLinuxSection` to demonstrate/validate this fix. Without the fix to `spec_linux.go`, I get the following results of the test:
```
--- FAIL: TestSpecconvNoLinuxSection (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0x64898d]

goroutine 34 [running]:
testing.tRunner.func1(0xc0000e9000)
        /usr/lib/go-1.13/src/testing/testing.go:874 +0x3a3
panic(0x6a1f80, 0x905be0)
        /usr/lib/go-1.13/src/runtime/panic.go:679 +0x1b2
github.com/opencontainers/runc/libcontainer/specconv.createDevices(0xc0000e4680, 0xc0001041e0, 0x6f70dd, 0xe, 0x6f4d89, 0x6, 0x6f4d89)
        /home/najohnsn/go/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go:683 +0x1ed
github.com/opencontainers/runc/libcontainer/specconv.CreateLibcontainerConfig(0xc0000dff38, 0x0, 0x1, 0xc00009c5c0)
        /home/najohnsn/go/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux.go:246 +0x5bb
github.com/opencontainers/runc/libcontainer/specconv.TestSpecconvNoLinuxSection(0xc0000e9000)
        /home/najohnsn/go/src/github.com/opencontainers/runc/libcontainer/specconv/spec_linux_test.go:561 +0xc3
testing.tRunner(0xc0000e9000, 0x704cb0)
        /usr/lib/go-1.13/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
        /usr/lib/go-1.13/src/testing/testing.go:960 +0x350
FAIL    github.com/opencontainers/runc/libcontainer/specconv    0.133s
```

The new test passes with the fix:
```
=== RUN   TestSpecconvNoLinuxSection
--- PASS: TestSpecconvNoLinuxSection (0.00s)
```

Signed-off-by: Neil Johnson <najohnsn@us.ibm.com>


libcontainer/specconv/spec_linux_test.go
```diff
@@ -546,6 +546,29 @@ func TestSpecconvExampleValidate(t *testing.T) {
...
 
+func TestSpecconvNoLinuxSection(t *testing.T) {
+	spec := Example()
+	spec.Root.Path = "/"
+	spec.Linux = nil
+	spec.Hostname = ""
+
+	opts := &CreateOpts{
+		CgroupName:       "ContainerID",
+		UseSystemdCgroup: false,
```
kolyshkin, line 557: no need to specify that

libcontainer/specconv/spec_linux_test.go
```diff
@@ -546,6 +546,29 @@ func TestSpecconvExampleValidate(t *testing.T) {
...
+	}
+
+	config, err := CreateLibcontainerConfig(opts)
+	if err != nil {
+		t.Errorf("Couldn't create libcontainer config: %v", err)
+	}
+
+	validator := validate.New()
+	if err := validator.Validate(config); err != nil {
+		t.Errorf("Expected specconv to produce valid container config: %v", err)
```
kolyshkin, line 568:
```go
t.Errorf("Expected no error from validate, got %v", err)
```

kolyshkin (commented):
Couple of nits, otherwise LGTM

libcontainer/specconv/spec_linux_test.go
najohnsn: Removed UseSystemdCgroup

kolyshkin (commented):
please squash your commits.

libcontainer/specconv/spec_linux_test.go
najohnsn: The existing code has consistent wording with the same checking in `TestSpecconvExampleValidate` and `TestNonZeroEUIDCompatibleSpecconvValidate`. You this this new case deserves different wording?

libcontainer/specconv/spec_linux_test.go
kolyshkin: I can still see it.

libcontainer/specconv/spec_linux_test.go
kolyshkin: Ah, OK then

libcontainer/specconv/spec_linux_test.go
najohnsn: Sorry. I think it's really gone this time.

kolyshkin (commented):
1. Looking into `func createDevices` as a whole, it seems that it can be simplified for the case `spec.Linux == nil` like this:

```go
func createDevices(spec *specs.Spec, config *configs.Config) ([]*devices.Device, error) {
    if spec.Linux == nil {
        return AllowedDevices, nil
     }
     .....
```

2. Can you please describe how did you come across this issue?

kolyshkin:
Related: #1559, #1551, #1449 (the last one explains why we should tolerate spec.Linux == nil case).

najohnsn:
@kolyshkin 
1. That simplification does not seem right. That function not only returns a value, it also modifies the input config. Or is there something subtle I'm missing?

2. Came across this accidentally when a _normal_ spec with a Linux section was used in a prototype version of createDevices for z/OS (so, it had a nil spec.Zos). Inspecting the code, I could see that the spec.Linux == nil case wasn't being handled as I understood it should be.

kolyshkin:
> 1. That simplification does not seem right. That function not only returns a value, it also modifies the input config. Or is there something subtle I'm missing?

You're right, I missed that part. 

> 2. Came across this accidentally when a _normal_ spec with a Linux section was used in a prototype version of createDevices for z/OS (so, it had a nil spec.Zos). Inspecting the code, I could see that the spec.Linux == nil case wasn't being handled as I understood it should be.

Nice!

Issue #3232, opened 2021-10-05
Missing hugetlb support leads to many log messages


PKizzle:
System infos:
- Raspberry Pi 4
- Raspberry Pi OS / Debian (bullseye) arm64
- kernel (5.10.63-v8+)

The kernel shipped with Raspberry Pi OS currently does not have hugetlb support enabled by default, which leads to the following error messages. I have seen that it possible to deactivate the hugetlb metric in cadvisor, however I don't see any way to do this when using k3s. Since the error message "failed to fetch hugetlb info" is thrown by runc I guess that the problem has to be fixed in this project.

```
W1005 10:16:56.888449  789311 prometheus.go:1856] Couldn't get containers: partial failures: ["/kubepods/burstable/pod6a8f6862-e3b0-4fa0-b794-7b70f456706b/bc454ffcfd880777139aeb3e4e79e82da3702c6fb4a52069d0b23108f4fe87e4": containerDataToContainerInfo: unable to find data in memory cache], ["/kubepods/burstable/pod6a8f6862-e3b0-4fa0-b794-7b70f456706b/b270f0938dfb2e2eb9894b3b343b8a11ec442563e3244304c7f7c91bcee5cba7": containerDataToContainerI
[...]

PR #3233, opened 2021-10-06
libct/cg/fs2: fix GetStats for unsupported hugetlb

kolyshkin:
In case hugetlb is not supported, GetStats() should not error out,
and yet it does.

Assume that if `GetHugePageSize` return an error, hugetlb is
not supported (this is what cgroup v1 manager do).

NOTE this code is not used by the runc itself, only by external users
like kubernetes and cadvisor that import libcontainer/cgroup/...

Fixes: 89a87adb
Fixes: #3232

### DIFF
--- a/libcontainer/cgroups/fs2/hugetlb.go
+++ b/libcontainer/cgroups/fs2/hugetlb.go
@@ -1,7 +1,6 @@
 package fs2
 
 import (
-	"fmt"
 	"strconv"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -27,10 +26,7 @@ func setHugeTlb(dirPath string, r *configs.Resources) error {
 }
 
 func statHugeTlb(dirPath string, stats *cgroups.Stats) error {
-	hugePageSizes, err := cgroups.GetHugePageSize()
-	if err != nil {
-		return fmt.Errorf("failed to fetch hugetlb info: %w", err)
-	}
+	hugePageSizes, _ := cgroups.GetHugePageSize()
 	hugetlbStats := cgroups.HugetlbStats{}
 
 	for _, pagesize := range hugePageSizes {


libcontainer/cgroups/fs2/hugetlb.go
```diff
@@ -27,10 +26,7 @@ func setHugeTlb(dirPath string, r *configs.Resources) error {
 }
 
 func statHugeTlb(dirPath string, stats *cgroups.Stats) error {
-	hugePageSizes, err := cgroups.GetHugePageSize()
-	if err != nil {
-		return fmt.Errorf("failed to fetch hugetlb info: %w", err)
-	}
+	hugePageSizes, _ := cgroups.GetHugePageSize()
```
AkihiroSuda, line 29: Could you add a comment to clarify that we are ignoring err on purpose?

kolyshkin: This is kind of an intermediate commit, merely fixing the issue. Please see #3234 for a complete solution -- in where I remove the returned error altogether (as no one is using it), so there's no need to ignore it (and explain why).

ctrahey:
Thanks! What is the release timeline?

luispabon:
Sorry to be a nag, but each of my boxes is generating daily ~4GB in logs. Any chance for a release soon?

nodesocket:
👍🏻 also would love this released as I seeing my logs spammed with these errors.

AkihiroSuda:
Cherry-picking to v1.0.3
https://github.com/opencontainers/runc/pull/3295

nodesocket:
Once v1.0.3 is released, what's the upgrade process? I am running Debian 11 (Bullseye) on `aarch64`. Can I simply run `sudo apt-get install runc`?

kolyshkin:
> Once v1.0.3 is released, what's the upgrade process? I am running Debian 11 _(Bullseye)_ on `aarch64`. Can I simply run `sudo apt-get -y install runc`?

This depends on your distro and its maintainers. I am using Fedora, the runc package is being updated regularly, following the upstream (i.e. this repo).

In this case, though, it does not matter which version of runc binary is installed on the system. What matters is which version of runc's libcontainer k8s (or k3s, not sure which one are you using) is built against. It seems that both k8s and k3s are tracking upstream (this repo) closely, so chances are high that after we'll release 1.0.3 they will rebase.

nodesocket:
@kolyshkin thanks for the detailed reply.

I am using Kubernetes version `1.22.4`. Does this mean I'll have to upgrade the Kubernetes version once they rebase runc `v1.0.3` in?

nodesocket:
@kolyshkin any response to my previous question? To get this fix, will I have to upgrade my Kubernetes version from `1.22.4` and wait for Kubernetes to rollout a new release?

luispabon:
100% dependent on how you're deploying your kubernetes.

k3s for instance bundles the runc binary within itself, so you do need them to release on their end. Who knows how GKE or EKS nodes are built internally. Etc.

nodesocket:
@luispabon deployed Kubernetes using kubeadm. Not usingi GKE or EKS for this deployment. On-prem deploy with standard kubeadm.

```bash
cat << EOF > /tmp/kubernetes-master-config.yaml >/dev/null
apiVersion: kubeadm.k8s.io/v1beta3
kind: ClusterConfiguration
kubernetesVersion: 1.23.0
networking:
  podSubnet: "10.244.0.0/16"
EOF
```

```bash
sudo kubeadm init --config /tmp/kubernetes-master-config.yaml
```

thaJeztah:
> 100% dependent on how you're deploying your kubernetes.
> 
> k3s for instance bundles the runc binary within itself, so you do need them to release on their end. Who knows how GKE or EKS nodes are built internally. Etc.

Per earlier comment of @kolyshkin, it's likely not the `runc` binary that makes a difference here, but the go module (build-time dependency) used to build kubernetes; looks like the master branch of kubernetes is still on v1.0.2 currently; https://github.com/kubernetes/kubernetes/blob/a8e06cf2bfd2ecdde0f2f49af4df29f70e07f20e/go.mod#L63

```
github.com/opencontainers/runc v1.0.2
```

PR #3234, opened 2021-10-06
libct/cg: refactor/improve/rename GetHugePageSize -> HugePageSizes

kolyshkin:
~~_(this is based on and currently includes PR #3233)_~~

1. Since GetHugePageSize do not have any external users (checked by
       sourcegraph), and no internal user ever uses its second return value
       (the error), let's drop it.
    
2. Rename GetHugePageSize -> HugePageSizes (drop the Get prefix as per
       Go guidelines, add suffix since we return many sizes).

3.  Implement lazy init for `HugePageSizes`
    
    I have noticed that libct/cg/fs allocates 8K during init on every runc
    execution:
    
    > init github.com/opencontainers/runc/libcontainer/cgroups/fs @1.5 ms, 0.028 ms clock, **8512 bytes, 13 allocs**
    
    Apparently this is caused by the global `HugePageSizes` variable initialization,
    which is only used from `GetStat`s (i.e. it is never used by the runc binary itself).
    
    Remove it, and use `HugePageSizes()` directly instead. Make it init-once,
    so that `GetStats` (which, I guess, is periodically called by kubernetes)
    does not re-read huge page sizes over and over.
    
    This also removes 12 allocs and 8K from libct/cg/fs init section:
    
    > $ time GODEBUG=inittrace=1 ./runc --help 2>&1 | grep cgroups/fs
    > init github.com/opencontainers/runc/libcontainer/cgroups/fs @1.5 ms, 0.003 ms clock, **16 bytes, 1 allocs**

4. libct/cg: TestGetHugePageSizeImpl: use t.Run
    
    Move test case comments to doc strings, and use t.Run.

### DIFF
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -448,15 +448,12 @@ func TestFindCgroupMountpointAndRoot(t *testing.T) {
 	}
 }
 
-func BenchmarkGetHugePageSize(b *testing.B) {
-	var (
-		output []string
-		err    error
-	)
+func BenchmarkHugePageSizes(b *testing.B) {
+	var output []string
 	for i := 0; i < b.N; i++ {
-		output, err = GetHugePageSize()
+		output = HugePageSizes()
 	}
-	if err != nil || len(output) == 0 {
+	if len(output) == 0 {
 		b.Fatal("unexpected results")
 	}
 }
--- a/libcontainer/cgroups/utils_test.go
+++ b/libcontainer/cgroups/utils_test.go
@@ -448,16 +448,6 @@ func TestFindCgroupMountpointAndRoot(t *testing.T) {
 	}
 }
 
-func BenchmarkHugePageSizes(b *testing.B) {
-	var output []string
-	for i := 0; i < b.N; i++ {
-		output = HugePageSizes()
-	}
-	if len(output) == 0 {
-		b.Fatal("unexpected results")
-	}
-}
-
 func BenchmarkGetHugePageSizeImpl(b *testing.B) {
 	var (
 		input  = []string{"hugepages-1048576kB", "hugepages-2048kB", "hugepages-32768kB", "hugepages-64kB"}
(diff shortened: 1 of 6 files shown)


kolyshkin:
This is definitely post-1.1 material

kolyshkin:
Rebased on top of current git tip to repro https://github.com/opencontainers/runc/issues/3266

libcontainer/cgroups/utils_test.go
```diff
@@ -487,25 +485,29 @@ func TestGetHugePageSizeImpl(t *testing.T) {
...
 		},
-		{ // no suffix (skipped with a warning)
-			input:  []string{"hugepages-1024"},
-			isWarn: true,
+		{ // mixed valid and invalid entries
+			input:  []string{"hugepages-4194304kB", "hugepages-2048kB", "hugepages-akB", "hugepages-64kB"},
+			output: []string{"4GB", "2MB", "64KB"},
+			isErr:  true,
+		},
+		{ // mixed valid and invalid entries
```
thaJeztah, line 501: No need to change, but could include the comment as a `doc: "mixed valid and invalid entries",`, and then use subtests;

```
for _, c := range testCases {
    c :=c
    t.Run(c.doc, func(t *testing.T) {
...
```

That way the test output is documenting what test fails (easier to find back which test is failing), (obvious disclaimer that I like subtests for that reason 😅)

thaJeztah (approved):
code LGTM

is the failure still the thing we're debugging?

libcontainer/cgroups/utils_test.go
kolyshkin: Makes sense, added (as a separate commit for clarity and so you don't have to re-review the whole set).

kolyshkin:
> is the failure still the thing we're debugging?

No, it was a one time thing, fixed by https://github.com/opencontainers/runc/pull/3273

kolyshkin:
Since 1.1 is delayed I think this can be included.

thaJeztah (approved):
LGTM!

kolyshkin:
@AkihiroSuda PTAL

AkihiroSuda (approved):

PR #3237, opened 2021-10-07
[WIP] Better error messages on failed to exec

rhatdan:
We often get Podman error messages where container images fail to exec.

standard_init_linux.go:228: exec user process caused: no such file or directory

It would be much better to know which executable was not found.


thaJeztah:
> I figure this can only happen then the binary is deleted in between runc create and runc start.

There may be some other cases where this error is produced (that often confuses users)

- line endings in a shell script (executable not found)
- dynamically linked binary missing a library (also shows as executable not found)


see https://github.com/moby/moby/issues/29420 (and slightly related https://github.com/moby/moby/issues/30617)

PR #3238, opened 2021-10-07
libct/system: Exec: wrap the error

kolyshkin:
If the container binary to be run is removed in between runc create
and runc start, the latter spits the following error:

    can't exec user process: no such file or directory

This is a bit confusing since we don't see what file is missing.

Wrap the unix.Exec error into os.PathError, like in many other cases,
to provide some context. Remove the error wrapping from
(*linuxStandardInit).Init as it is now redundant.

With this patch, the error is now:

    exec /bin/false: no such file or directory

Closes: #3237


kolyshkin:
Repro:
```console
# runc create xxx
# mv rootfs/bin/sh rootfs/bin/sh.x
# runc start xxx
```

PR #3240, opened 2021-10-08
libct/capabilities: create capabilities map based on current environment

thaJeztah:
Relates to:

- https://github.com/moby/moby/issues/42906 docker container is privileged tries to assume more capabilities than available
- https://github.com/moby/buildkit/pull/2394 fix: provide only available capabilities to insecure environment
- https://github.com/moby/moby/pull/42911 fix: don't attempt to set unavailable capabilities in privileged mode
- https://github.com/moby/moby/issues/42892 Recent test failures in CI (apply caps: operation not permitted)
- https://github.com/opencontainers/runc/pull/2854 capabilities: WARN, not ERROR, for unknown / unavailable capabilities
- https://github.com/opencontainers/runtime-spec/pull/1094 runtime should WARN / ignore capabilities that cannot be granted

Commit 5fb831a0fa68578e8020dc9c59b90336cfd95198 (https://github.com/opencontainers/runc/pull/2854) changed the behavior of runc
to match the OCI runtime spec, which now describes that unknown or unavailable
capabilities should be ignored.

While this change addressed situations where a capability was requested that's
not supported by the current kernel ("unknown capabilities"), it did not take
into account situations where the kernel *supports* a capability, but that
capability is not *available* in the current environment.

This causes issues if, for example, runc is running in a docker-in-docker setup,
and the outer container does not have all known capabilities enabled, either
on purpose (for example, Talos version 0.13 drops two capabilities (kexec + module
loading) from all processes but PID 1), or because the outer container was created
by an older version of docker or runc, which did not yet support newer capabilities.

This patch attempts to address this problem by limiting the list of "known" capa-
bilities on the set of effective capabilties for the current process. This code
is based on the code in containerd's "caps" package, with some modifications:

- the full list of capabilities uses github.com/syndtr/gocapability, instead of
  a self-defined list. Containerd removed the use of github.com/syndtr/gocapability,
  but this dependency is still in use in runc, so this change makes it a closer
  match to the current code.
- functions where un-exported, as we don't intend them to be used externally.
- a sync.Once was added to the .current() function, so that /proc/self/status
  is only parsed once. This assumes effective capabilities do not change during
  runc's lifecycle.

There are some things left to be looked at:

1. current() may return an error when failing to parse /proc/self/status, but this
   error is currently ignored. If an error occurs in this code, it will mean that
   *no* capabilities are known. While this will be logged as warning when attempting
   to apply capabilities, it's not a very desirable situation. We'll have to decide
   what to do in that situation, which could be "panic" (runc unable to run success-
   fully), or "fall back to a safe/default list".
2. the current code applies the same list (effective caps) to every "type" (ambient,
   inheritable, bounding, ...). When applying capabilities, should each of those
   types in the container's spec be limited to the _corresponding_ type in the
   current processes' capabilities?
3. integration test: we may want an integration test for this.
4. do we want to upstream this functionality to github.com/syndtr/gocapability ?

### DIFF
--- a/libcontainer/capabilities/capabilities.go
+++ b/libcontainer/capabilities/capabilities.go
@@ -5,7 +5,6 @@ package capabilities
 
 import (
 	"sort"
-	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/sirupsen/logrus"
@@ -14,25 +13,12 @@ import (
 
 const allCapabilityTypes = capability.CAPS | capability.BOUNDING | capability.AMBIENT
 
-var (
-	capabilityMap map[string]capability.Cap
-	capTypes      = []capability.CapType{
-		capability.BOUNDING,
-		capability.PERMITTED,
-		capability.INHERITABLE,
-		capability.EFFECTIVE,
-		capability.AMBIENT,
-	}
-)
-
-func init() {
-	capabilityMap = make(map[string]capability.Cap, capability.CAP_LAST_CAP+1)
-	for _, c := range capability.List() {
-		if c > capability.CAP_LAST_CAP {
-			continue
-		}
-		capabilityMap["CAP_"+strings.ToUpper(c.String())] = c
-	}
+var capTypes = []capability.CapType{
+	capability.BOUNDING,
+	capability.PERMITTED,
+	capability.INHERITABLE,
+	capability.EFFECTIVE,
+	capability.AMBIENT,
 }
 
 // New creates a new Caps from the given Capabilities config. Unknown Capabilities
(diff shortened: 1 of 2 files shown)


tonistiigi:
@smira While `--privileged`  has historically meant to add a fixed list of caps hardcoded in docker, one could make an argument that expected behavior is to add all caps that docker is able to add. (There is a counterargument that we should prioritize reproducibility between systems instead but that is a discussion for a different repo.) But with user writing `--cap-add SYS_BOOT` there isn't really two different ways to think about it other than `SYS_BOOT` is required for this container.

tonistiigi:
@smira Sorry, I meant `dockerd` in here, not client. The Docker API already only deals with added caps, dropped caps  and special modes like `privileged`. Not full list of caps like runc. Dockerd runs in same environment as `runc` so can use `caps.current()` to figure out what is the correct list for the runc spec for specific case.

kolyshkin:
So, are we implementing this in higher levels instead?

kolyshkin:
> So, are we implementing this in higher levels instead?

@thaJeztah ^^^

thaJeztah:
So, we implemented this in Moby through https://github.com/moby/moby/pull/42933, but I (personally) still think it would be nice to have a way to indicate "give me whatever you have"; perhaps that would be more something like https://github.com/opencontainers/runtime-spec/issues/1071 (adding `ALL_CAPS` to the spec)? Although perhaps https://github.com/opencontainers/runtime-spec/pull/1130 could be considered for that instead.

PR #3245, opened 2021-10-13
Drop Go 1.15 support

kolyshkin:
Drop Go 1.15 support, requiring Go >= 1.16. Motivated by https://github.com/opencontainers/runc/pull/3242#issuecomment-939672253, should unblock that PR.

The last commit drops use of `io/ioutil`, which is not really required and can be postponed for later time (say after Go 1.18 release, which will _probably_ mark ioutil package as deprecated).

### DIFF
--- a/contrib/cmd/recvtty/recvtty.go
+++ b/contrib/cmd/recvtty/recvtty.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"os"
 	"strings"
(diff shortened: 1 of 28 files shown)


kolyshkin:
> Reading package lists...
> W: GPG error: https://download.opensuse.org/repositories/devel:/tools:/criu/xUbuntu_20.04  InRelease: The following signatures were invalid: EXPKEYSIG 30A8343A498D5A23 devel:tools OBS Project <devel:tools@build.opensuse.org>

PR #3254, opened 2021-10-29
libct/configs/validate: allow / in sysctl names

kolyshkin:
Runtime spec [says](https://github.com/opencontainers/runtime-spec/blob/0d6cc581aeeaacee38ce707ae651a0e122f76852/config-linux.md#sysctl):

> sysctl (object, OPTIONAL) allows kernel parameters to be modified at
> runtime for the container. For more information, see the sysctl(8)
> man page.

and sysctl(8) says:

> variable
>    The name of a key to read from. An example is
>    kernel.ostype. The '/' separator is also accepted in place of a '.'.

[Apparently](https://github.com/kubernetes/kubernetes/pull/102393#issuecomment-939047989), runc config validator do not support sysctls with `/` as a
separator. Fortunately this is a one-line fix.

Add some more test data where `/` is used as a separator.

Related to: https://github.com/kubernetes/kubernetes/pull/102393

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -150,6 +150,7 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 	)
 
 	for s := range config.Sysctl {
+		s := strings.Replace(s, "/", ".", -1)
 		if validSysctlMap[s] || strings.HasPrefix(s, "fs.mqueue.") {
 			if config.Namespaces.Contains(configs.NEWIPC) {
 				continue
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -188,8 +188,11 @@ func TestValidateUsernamespaceWithoutUserNS(t *testing.T) {
 func TestValidateSysctl(t *testing.T) {
 	sysctl := map[string]string{
 		"fs.mqueue.ctl": "ctl",
+		"fs/mqueue/ctl": "ctl",
 		"net.ctl":       "ctl",
+		"net/ctl":       "ctl",
 		"kernel.ctl":    "ctl",
+		"kernel/ctl":    "ctl",
 	}
 
 	for k, v := range sysctl {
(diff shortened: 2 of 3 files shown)


mengjiao-liu:
I think there is a problem with your conversion:

If the ssyctl value is net/pv4/conf/eno2.100/rp_filter
The correct conversion result should be net.ipv4.conf.eno2/100.rp_filter
 
In the unit test, the sysctl value with existing dots and slashes is also added.

libcontainer/configs/validate/validator_test.go
```diff
@@ -188,8 +188,11 @@ func TestValidateUsernamespaceWithoutUserNS(t *testing.T) {
 func TestValidateSysctl(t *testing.T) {
 	sysctl := map[string]string{
 		"fs.mqueue.ctl": "ctl",
+		"fs/mqueue/ctl": "ctl",
 		"net.ctl":       "ctl",
+		"net/ctl":       "ctl",
 		"kernel.ctl":    "ctl",
+		"kernel/ctl":    "ctl",
```
mengjiao-liu, line 195: In the unit test, the sysctl value with existing dots and slashes is also added.

libcontainer/configs/validate/validator.go
```diff
@@ -150,6 +150,7 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 	)
 
 	for s := range config.Sysctl {
+		s := strings.Replace(s, "/", ".", -1)
```
mengjiao-liu, line 153: I think there is a problem with your conversion:
If the ssyctl value is net/pv4/conf/eno2.100/rp_filter
The correct conversion result should be net.ipv4.conf.eno2/100.rp_filter

see kubernetes/kubernetes#102393 (comment)

AkihiroSuda: @mengjiao-liu Could you open a github issue?

mengjiao-liu: Ok, I will open a issue.

PR #3257, opened 2021-11-01
Fix the conversion of sysctl variable dots and slashes

### LINKED ISSUE #3256: '/' is allowed as a separator in sysctl name
[Runtime spec](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#sysctl) says:

> sysctl (object, OPTIONAL) allows kernel parameters to be modified at
runtime for the container. For more information, see the sysctl(8)
man page.

and linux sysctl says:

```
variable
The name of a key to read from. An example is
kernel.ostype. The '/' separator is also accepted in place of a '.'.
```
```
       Note that either "/" or "."  may be used as separators within
       sysctl variable names. If the first separator is a slash,
       remaining slashes and dots are left intact. If the first
       separator is a dot, dots and slashes are interchanged.
       "kernel.domainname=foo" and "kernel/domainname=foo" are
       equivalent and will cause "foo" to be written to
       /proc/sys/kernel/domainname. Either
       "net.ipv4.conf.enp3s0/200.forwarding" or
       "net/ipv4/conf/enp3s0.200/forwarding" may be used to refer to
       /proc/sys/net/ipv4/conf/enp3s0.200/forwarding. A glob glob(7)
       pattern may be used to write the same value to all matching keys.
       Keys for which an explicit pattern exists will be excluded from
       any glob matching. In addition, a key may be explicitly excluded
       from being set by any matching glob patterns by specifying the
       key name prefixed with a "-" character and not followed by "=",
       see SYNOPSIS.
```

https://man7.org/linux/man-pages/man8/sysctl.8.html
https://man7.org/linux/man-pages/man5/sysctl.d.5.html

However, in fact, the sysctl name in runc does not support a slash as a separator.

This PR  https://github.com/opencontainers/runc/pull/3254  wants to solve this problem, but I think the conversion method does not match the linux sysctl definition.

see https://github.com/opencontainers/runc/pull/3254/files#r739936786 for more details.

So I opened this issue to track the fix of this issue andI will open a new PR to fix this.

Related to: 
- kubernetes/kubernetes#102393
- https://github.com/opencontainers/runc/pull/3254

/assign

mengjiao-liu:
Fixes #3256 

see  
- https://github.com/opencontainers/runc/pull/3254/files#r739936786 
- https://github.com/kubernetes/kubernetes/pull/102393#discussion_r661550379

Fix the sysctl conversion function according to the linux ssyctl conversion definition.
```
       Note that either "/" or "."  may be used as separators within
       sysctl variable names. If the first separator is a slash,
       remaining slashes and dots are left intact. If the first
       separator is a dot, dots and slashes are interchanged.
       "kernel.domainname=foo" and "kernel/domainname=foo" are
       equivalent and will cause "foo" to be written to
       /proc/sys/kernel/domainname. Either
       "net.ipv4.conf.enp3s0/200.forwarding" or
       "net/ipv4/conf/enp3s0.200/forwarding" may be used to refer to
       /proc/sys/net/ipv4/conf/enp3s0.200/forwarding. A glob glob(7)
       pattern may be used to write the same value to all matching keys.
       Keys for which an explicit pattern exists will be excluded from
       any glob matching. In addition, a key may be explicitly excluded
       from being set by any matching glob patterns by specifying the
       key name prefixed with a "-" character and not followed by "=",
       see SYNOPSIS.
```
see following webpages for more details:
https://man7.org/linux/man-pages/man8/sysctl.8.html
https://man7.org/linux/man-pages/man5/sysctl.d.5.html 

Related to: 
- kubernetes/kubernetes#102393
- https://github.com/opencontainers/runc/pull/3254

### DIFF
--- a/libcontainer/configs/validate/validator.go
+++ b/libcontainer/configs/validate/validator.go
@@ -128,6 +128,36 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
 	return nil
 }
 
+// convertSysctlVariableToDotsSeparator can return sysctl variables in dots separator format.
+// The '/' separator is also accepted in place of a '.'.
+// Convert the sysctl variables to dots separator format for validation.
+// More info:
+//   https://man7.org/linux/man-pages/man8/sysctl.8.html
+//   https://man7.org/linux/man-pages/man5/sysctl.d.5.html
+// For example:
+// Input sysctl variable "net/ipv4/conf/eno2.100.rp_filter"
+// will return the converted value "net.ipv4.conf.eno2/100.rp_filter"
+func convertSysctlVariableToDotsSeparator(val string) string {
+	if val == "" {
+		return val
+	}
+	firstSepIndex := strings.IndexAny(val, "./")
+	if firstSepIndex == -1 || val[firstSepIndex] == '.' {
+		return val
+	}
+
+	f := func(r rune) rune {
+		switch r {
+		case '.':
+			return '/'
+		case '/':
+			return '.'
+		}
+		return r
+	}
+	return strings.Map(f, val)
+}
+
 // sysctl validates that the specified sysctl keys are valid or not.
 // /proc/sys isn't completely namespaced and depending on which namespaces
 // are specified, a subset of sysctls are permitted.
@@ -150,7 +180,7 @@ func (v *ConfigValidator) sysctl(config *configs.Config) error {
 	)
 
 	for s := range config.Sysctl {
-		s := strings.Replace(s, "/", ".", -1)
+		s := convertSysctlVariableToDotsSeparator(s)
 		if validSysctlMap[s] || strings.HasPrefix(s, "fs.mqueue.") {
 			if config.Namespaces.Contains(configs.NEWIPC) {
 				continue
--- a/libcontainer/configs/validate/validator_test.go
+++ b/libcontainer/configs/validate/validator_test.go
@@ -184,14 +184,40 @@ func TestValidateUsernamespaceWithoutUserNS(t *testing.T) {
 	}
 }
 
+// TestConvertSysctlVariableToDotsSeparator tests whether the sysctl variable
+// can be correctly converted to a dot as a separator.
+func TestConvertSysctlVariableToDotsSeparator(t *testing.T) {
+	type testCase struct {
+		in  string
+		out string
+	}
+	valid := []testCase{
+		{in: "kernel.shm_rmid_forced", out: "kernel.shm_rmid_forced"},
+		{in: "kernel/shm_rmid_forced", out: "kernel.shm_rmid_forced"},
+		{in: "net.ipv4.conf.eno2/100.rp_filter", out: "net.ipv4.conf.eno2/100.rp_filter"},
+		{in: "net/ipv4/conf/eno2.100/rp_filter", out: "net.ipv4.conf.eno2/100.rp_filter"},
+		{in: "net/ipv4/ip_local_port_range", out: "net.ipv4.ip_local_port_range"},
+		{in: "kernel/msgmax", out: "kernel.msgmax"},
+		{in: "kernel/sem", out: "kernel.sem"},
+	}
+
+	for _, test := range valid {
+		convertSysctlVal := convertSysctlVariableToDotsSeparator(test.in)
+		if convertSysctlVal != test.out {
+			t.Errorf("The sysctl variable was not converted correctly. got: %s, want: %s", convertSysctlVal, test.out)
+		}
+	}
+}
+
 func TestValidateSysctl(t *testing.T) {
 	sysctl := map[string]string{
-		"fs.mqueue.ctl": "ctl",
-		"fs/mqueue/ctl": "ctl",
-		"net.ctl":       "ctl",
-		"net/ctl":       "ctl",
-		"kernel.ctl":    "ctl",
-		"kernel/ctl":    "ctl",
+		"fs.mqueue.ctl":                    "ctl",
+		"fs/mqueue/ctl":                    "ctl",
+		"net.ctl":                          "ctl",
+		"net/ctl":                          "ctl",
+		"net.ipv4.conf.eno2/100.rp_filter": "ctl",
+		"kernel.ctl":                       "ctl",
+		"kernel/ctl":                       "ctl",
 	}
 
 	for k, v := range sysctl {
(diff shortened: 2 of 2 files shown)


mengjiao-liu:
PTAL @AkihiroSuda @mrunalp

libcontainer/configs/validate/validator.go
```diff
@@ -128,6 +128,41 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
 	return nil
 }
 
+// convertSysctlVariableToDotsSeparator can return sysctl variables in dots separator format.
+// The '/' separator is also accepted in place of a '.'.
+// Convert the sysctl variables to dots separator format for validation.
+// More info:
+//   https://man7.org/linux/man-pages/man8/sysctl.8.html
+//   https://man7.org/linux/man-pages/man5/sysctl.d.5.html
+func convertSysctlVariableToDotsSeparator(val string) string {
```
AkihiroSuda, line 137: Could you add example inputs and outputs the code comment?
Also could you consider adding unit tests

mengjiao-liu: I can add unit tests, but this package test uses `package validate_test`.
Then this function must be exported, otherwise it cannot be referenced. I am not sure whether to export this function, after all, this method is only used in this package.

AkihiroSuda: The package name should be renamed to `validate`, then you can add unit tests without exporting the function.

mengjiao-liu: Okay, I have renamed the test package name to `validate`.

AkihiroSuda (dismissed):

libcontainer/configs/validate/validator.go
```diff
@@ -128,6 +128,44 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
...
+// Input sysctl variable "net/ipv4/conf/eno2.100.rp_filter"
+// will return the converted value "net.ipv4.conf.eno2/100.rp_filter"
+func convertSysctlVariableToDotsSeparator(val string) string {
+	if val == "" {
+		return val
+	}
+	sepFunc := func(r rune) bool {
+		return r == '.' || r == '/'
+	}
+	firsSepIndex := strings.IndexFunc(val, sepFunc)
```
kolyshkin, line 147: Why not use `strings.IndexAny`?

libcontainer/configs/validate/validator.go
```diff
@@ -128,6 +128,44 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
...
+		} else {
+			f := func(r rune) rune {
+				switch r {
+				case '.':
+					return '/'
+				case '/':
+					return '.'
+				default:
+					return r
+				}
```
kolyshkin, line 160:
```suggestion
				return r
```

libcontainer/configs/validate/validator.go
```diff
@@ -128,6 +128,44 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
...
+			}
+			convertedSysctlVal := strings.Map(f, val)
+			return convertedSysctlVal
```
kolyshkin, line 163:
```suggestion
			return strings.Map(f, val)
```

libcontainer/configs/validate/validator.go
```diff
@@ -128,6 +128,44 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
...
+// will return the converted value "net.ipv4.conf.eno2/100.rp_filter"
+func convertSysctlVariableToDotsSeparator(val string) string {
+	if val == "" {
+		return val
+	}
+	sepFunc := func(r rune) bool {
+		return r == '.' || r == '/'
+	}
+	firsSepIndex := strings.IndexFunc(val, sepFunc)
+	if firsSepIndex != -1 {
```
kolyshkin, line 148: This can be rewritten to simplify things a bit. Less indentation, avoiding else block etc.

For example:

```go
if firstSepIndex == -1 || val[idx] == '.' {
	return val
}
// Replace . with / and vice versa.
f := func(r rune) rune {
.....
return strings.Map(f, val)
```

libcontainer/configs/validate/validator.go
kolyshkin: There's also a typo: `firs` -> `first`

libcontainer/configs/validate/validator_test.go
```diff
@@ -178,21 +177,47 @@ func TestValidateUsernamespaceWithoutUserNS(t *testing.T) {
...
 	err := validator.Validate(config)
 	if err == nil {
 		t.Error("Expected error to occur but it was nil")
 	}
 }
 
+// TestConvertSysctlVariableToDotsSeparator tests whether the sysctl variable
+// can be correctly converted to a dot as a separator.
+func TestConvertSysctlVariableToDotsSeparator(t *testing.T) {
+	type Test struct {
```
kolyshkin, line 190: `s/Test/testCase/g`
or something

libcontainer/configs/validate/validator_test.go
```diff
@@ -178,21 +177,47 @@ func TestValidateUsernamespaceWithoutUserNS(t *testing.T) {
...
+		sysctl string
+		except string
```
kolyshkin, line 192: you probably mean "expect" here.

You can also use "in" and "out" (as the test input and output).

kolyshkin (changes requested):
Left some nits.

Can you please separate out the `package validate_test -> package validate` change to a separate (first) commit?

libcontainer/configs/validate/validator.go
mengjiao-liu: Yes, this function `strings.IndexAny` is more convenient, I have modified it.

`firs` -> `first`： I didn’t pay attention, it has been modified.

libcontainer/configs/validate/validator.go
mengjiao-liu: Updated.

libcontainer/configs/validate/validator.go
mengjiao-liu: Thanks for advice. Updated.

libcontainer/configs/validate/validator_test.go
mengjiao-liu: Updated `Test` -> `testCase`.

libcontainer/configs/validate/validator_test.go
mengjiao-liu: It might be better to understand this way. Updated.

libcontainer/configs/validate/validator.go
mengjiao-liu: default:
					return r

You mean to delete these two lines?
But I think it’s not necessary to delete it, and it’s okay to write it in the switch segment.

libcontainer/configs/validate/validator.go
```diff
@@ -128,6 +128,37 @@ func (v *ConfigValidator) cgroupnamespace(config *configs.Config) error {
...
+
+	f := func(r rune) rune {
+		switch r {
+		case '.':
+			return '/'
+		case '/':
+			return '.'
+		default:
+			return r
+		}
```
kolyshkin, line 157:
```suggestion
		return r
```

(i.e. remove the `default` case)

kolyshkin:
LGTM except for a single nit, thank you @mengjiao-liu!

Also, this needs to be rebased due to #3261

kolyshkin:
Ah, and the signed-off-by needs to be added to the commits (see https://github.com/opencontainers/runc/pull/3257/checks?check_run_id=4100751081)

mengjiao-liu:
Rebased and signed-off-by. Thanks for reminding!

AkihiroSuda (approved):

kolyshkin (approved):
LGTM, thank you!

PR #3267, opened 2021-11-09
Run gosec against runc

SaschaSchwarze0:
I extended the GitHub action to run gosec as part of a build. I mitigated the one finding:

* In checkpoint.go there was a `Atoi` followed by a cast to `int32`. I changed this to `ParseInt` with the correct bit size. Mitigates *G109: Potential Integer overflow made by strconv.Atoi result conversion to int16/32*

And I added ignores to those places where it complains about file write permissions, and commands using variables.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,5 +7,11 @@ run:
 linters:
   enable:
     - gofumpt
+    - gosec
     - errorlint
     - unconvert
+
+linters-settings:
+  gosec:
+    confidence: medium
+    severity: medium
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -118,7 +118,7 @@ func setPageServer(context *cli.Context, options *libcontainer.CriuOpts) {
 		if err != nil || address == "" || port == "" {
 			fatal(errors.New("Use --page-server ADDRESS:PORT to specify page server"))
 		}
-		portInt, err := strconv.Atoi(port)
+		portInt, err := strconv.ParseInt(port, 10, 32)
 		if err != nil {
 			fatal(errors.New("Invalid port number"))
 		}
--- a/contrib/cmd/recvtty/recvtty.go
+++ b/contrib/cmd/recvtty/recvtty.go
@@ -224,7 +224,7 @@ func main() {
 		pidPath := ctx.String("pid-file")
 		if pidPath != "" {
 			pid := fmt.Sprintf("%d\n", os.Getpid())
-			if err := os.WriteFile(pidPath, []byte(pid), 0o644); err != nil {
+			if err := os.WriteFile(pidPath, []byte(pid), 0o644); err != nil { //nolint:gosec
 				return err
 			}
 		}
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -482,7 +482,7 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 }
 
 func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, childLogPipe *os.File) *exec.Cmd {
-	cmd := exec.Command(c.initPath, c.initArgs[1:]...)
+	cmd := exec.Command(c.initPath, c.initArgs[1:]...) //nolint:gosec // G204 (Subprocess launched with a potential tainted input or cmd arguments)
 	cmd.Args[0] = c.initArgs[0]
 	cmd.Stdin = p.Stdin
 	cmd.Stdout = p.Stdout
@@ -1608,7 +1608,7 @@ func (c *linuxContainer) criuSwrk(process *Process, req *criurpc.CriuReq, opts *
 		// the initial CRIU run to detect the version. Skip it.
 		logrus.Debugf("Using CRIU %d at: %s", c.criuVersion, c.criuPath)
 	}
-	cmd := exec.Command(c.criuPath, args...)
+	cmd := exec.Command(c.criuPath, args...) //nolint:gosec // G204 (Subprocess launched with a potential tainted input or cmd arguments)
 	if process != nil {
 		cmd.Stdin = process.Stdin
 		cmd.Stdout = process.Stdout
--- a/libcontainer/integration/utils_test.go
+++ b/libcontainer/integration/utils_test.go
@@ -32,7 +32,7 @@ func init() {
 		panic(err)
 	}
 	// Call it to make sure images are downloaded, and to get the paths.
-	out, err := exec.Command(getImages).CombinedOutput()
+	out, err := exec.Command(getImages).CombinedOutput() //nolint:gosec // G204 (Subprocess launched with variable)
 	if err != nil {
 		panic(fmt.Errorf("getImages error %w (output: %s)", err, out))
 	}
@@ -158,7 +158,7 @@ func remove(dir string) {
 // copyBusybox copies the rootfs for a busybox container created for the test image
 // into the new directory for the specific test
 func copyBusybox(dest string) error {
-	out, err := exec.Command("sh", "-c", fmt.Sprintf("tar --exclude './dev/*' -C %q -xf %q", dest, busyboxTar)).CombinedOutput()
+	out, err := exec.Command("sh", "-c", fmt.Sprintf("tar --exclude './dev/*' -C %q -xf %q", dest, busyboxTar)).CombinedOutput() //nolint:gosec // G204 (Subprocess launched with variable)
 	if err != nil {
 		return fmt.Errorf("untar error %w: %q", err, out)
 	}
--- a/libcontainer/user/user_test.go
+++ b/libcontainer/user/user_test.go
@@ -99,6 +99,7 @@ this is just some garbage data
 }
 
 func TestValidGetExecUser(t *testing.T) {
+	// #nosec G101 (Potentially hardcoded credentials)
 	const passwdContent = `
 root:x:0:0:root user:/root:/bin/bash
 adm:x:42:43:adm:/var/adm:/bin/false
@@ -253,6 +254,7 @@ this is just some garbage data
 }
 
 func TestInvalidGetExecUser(t *testing.T) {
+	// #nosec G101 (Potentially hardcoded credentials)
 	const passwdContent = `
 root:x:0:0:root user:/root:/bin/bash
 adm:x:42:43:adm:/var/adm:/bin/false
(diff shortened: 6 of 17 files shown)


kolyshkin (commented):
Yes, it would be nice to add this linter and fix bugs found by it, so this work is welcome.

.github/workflows/validate.yml
```diff
@@ -148,6 +148,17 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+    - name: checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Run Gosec Security Scanner
+      uses: securego/gosec@master
+      with:
+        args: -confidence medium -severity high -exclude-dir libcontainer/nsenter ./...
```
kolyshkin, line 161: Since we already run golangci-lint, which supports gosec, perhaps it's better to implement these changes in `.golangci.yml` instead of adding a separate CI job. I'm sure it's possible to add a linter-specific configuration in there.

.github/workflows/validate.yml
```diff
@@ -148,6 +148,17 @@ jobs:
         make cfmt
         git diff --exit-code
 
+  gosec:
```
thaJeztah, line 151: Give. that we already run golang-ci linter, which also provides gosec linter, would we instead enable it in the golangci linter config?

checkpoint.go
```diff
@@ -118,7 +118,7 @@ func setPageServer(context *cli.Context, options *libcontainer.CriuOpts) {
 		if err != nil || address == "" || port == "" {
 			fatal(errors.New("Use --page-server ADDRESS:PORT to specify page server"))
 		}
-		portInt, err := strconv.Atoi(port)
+		portInt, err := strconv.ParseInt(port, 10, 32)
```
kolyshkin, line 121: (I can't comment on the commit message, so I'm commenting here instead).

It would be nice to describe what the linter warning was, and the nature of the fix, in the commit message.

The better commit subject is perhaps something like `checkpoint.go: use ParseInt`

kolyshkin:
Finally, it order for this PR to be mergeable:
- all the issue need to be fixed;
- the order of commits reverted (i.e. fix the bugs first, then enable the linter) -- this way we won't break git-bisect.

For these reasons, I am converting it to draft for now.

.github/workflows/validate.yml
```diff
@@ -22,6 +23,7 @@ jobs:
         with:
           # must be specified without patch version
           version: v1.42
+          args: --enable gosec
```
kolyshkin, line 26: I _guess_ this is done by adding the `gosec` to  `linters:` / `enable:` section of `.golang.yml` file.

thaJeztah: Note: If configuring through `args:` becomes too complicated, we could consider adding a `.golangci-lint.yaml` (I personally find that more convenient, as it also allows documenting the options we use through comments)

thaJeztah: oh, lol, our comments crossed (and forgot we already have the config file)

SaschaSchwarze0: Yep, found the yaml file already. Somehow the action finds more than the command but as [it does not tell where](https://github.com/golangci/golangci-lint-action/issues/119) ...

SaschaSchwarze0: Hm, it finds issues with confidence: HIGH and severity: MEDIUM, though I filter the other way around. I'll set both to medium and fix or nosec the findings.

thaJeztah: Yes, it doesn't show the line numbers in the logs (which is definitely a pity), but it _does_ annotate the code in the PR;

Screenshot 2021-11-12 at 10 34 37

SaschaSchwarze0: Nice, that helps.

thaJeztah: (I would definitely love to see that linked issue fixes as well though, as I think they should not be mutually exclusive 😅, but that's an issue we already have with the linter, so not something new)

libcontainer/integration/utils_test.go
```diff
@@ -32,6 +32,7 @@ func init() {
 		panic(err)
 	}
 	// Call it to make sure images are downloaded, and to get the paths.
+	// #nosec: G204 (Subprocess launched with variable)
```
thaJeztah, line 35: Thinking if we should just  disable this one in the configuration, because most (all??) would be false positives (or at least: to be ignored)

SaschaSchwarze0:
@thaJeztah @kolyshkin I am through with all your suggestions. [We](https://github.com/shipwright-io/build/blob/main/.github/workflows/verify.yaml) should also be using this linter instead of the individual tools that we run atm.

.golangci.yml
```diff
@@ -7,5 +7,17 @@ run:
...
+    - gosec
     - errorlint
     - unconvert
+
+linters-settings:
+  gosec:
+    confidence: medium
+    severity: medium
+    excludes:
+      - G301
```
thaJeztah, line 19: Bit on the fence if we should disable this one, or instead add `nolint` comments. Admitted, this check _does_ produce many "false positives", but I also know of some advisories that were related to incorrect permissions set (don't recall from the top of my head if those were in runc, but  at least in moby/docker and containerd).

libcontainer/container_linux.go
```diff
@@ -482,6 +482,7 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 }
 
 func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, childLogPipe *os.File) *exec.Cmd {
+	// #nosec: G204 (Subprocess launched with a potential tainted input or cmd arguments)
 	cmd := exec.Command(c.initPath, c.initArgs[1:]...)
```
thaJeztah, line 486: I should probably add here that, when using golangci-lint, the comments are formatted different; something like

```
//nolint:gosec
```

or 

```
//nolint:gosec // optional comment describing why it's excluded
```

If possible (doesn't work in _all_ cases), I generally prefer the comment to be on the same line as the actual code, e.g.;

```suggestion
	cmd := exec.Command(c.initPath, c.initArgs[1:]...) //nolint:gosec // ignore G204 (Subprocess launched with a potential tainted input or cmd arguments)
```

SaschaSchwarze0: Imo the nosec comments should not go away, because otherwise if you run `gosec` (or `make gosec`), they would not apply. Also, are you sure that they have to be different? Because they seem to work, otherwise the action would not be green.

libcontainer/integration/utils_test.go
SaschaSchwarze0: I leave that at the end to you to decide. For me it is also a kind of situation where one better thinks about if it's okay. Especially if not just a value of an arg is coming from a variable, but rather a whole array is passed in. Final call from you?

.golangci.yml
SaschaSchwarze0: Okay, I'll add nosecs. But, be aware that I will not be able to evaluate if they are correct.

checkpoint.go
kolyshkin: @SaschaSchwarze0 ^^^^

Makefile
```diff
@@ -137,6 +137,9 @@ shfmt:
 	shfmt -ln bats -d -w tests/integration/*.bats
 	shfmt -ln bash -d -w man/*.sh script/* tests/*.sh tests/integration/*.bash
 
+gosec:
+	gosec -confidence medium -severity medium -exclude G301,G302,G304,G306,G307 -exclude-dir libcontainer/nsenter ./...
+
```
kolyshkin, line 142: So, it seems that you add `exclude-dir` here but not to golangci-lint gosec configuration. Why is that?

libcontainer/container_linux.go
kolyshkin: I'd rather see `//nolint:gosec` annotations, if they work.

Makefile
thaJeztah: Perhaps we should not have a separate make target for `gosec`, given that we already have a target to run golangci-lint (and we don't have separate targets for each of the linters it provides)

.golangci.yml
```diff
@@ -7,5 +7,17 @@ run:
...
   enable:
     - gofumpt
+    - gosec
     - errorlint
     - unconvert
+
+linters-settings:
+  gosec:
+    confidence: medium
+    severity: medium
```
kolyshkin, line 17: In the PR description you said "we need it pass severity: high" and yet you put it to medium here.

SaschaSchwarze0: See above:

> Hm, it finds issues with confidence: HIGH and severity: MEDIUM, though I filter the other way around. I'll set both to medium and fix or nosec the findings.

Makefile
SaschaSchwarze0: Done

libcontainer/container_linux.go
SaschaSchwarze0: Done

kolyshkin:
fedora CI failure is a bug in systemd; we have since worked around it, I guess you need to rebase @SaschaSchwarze0

SaschaSchwarze0:
> fedora CI failure is a bug in systemd; we have since worked around it, I guess you need to rebase @SaschaSchwarze0

It is green now @kolyshkin

libcontainer/user/user_test.go
```diff
@@ -99,6 +99,7 @@ this is just some garbage data
 }
 
 func TestValidGetExecUser(t *testing.T) {
+	// #nosec G101 (Potentially hardcoded credentials)
```
kolyshkin, line 102: `nolint:gosec`?

libcontainer/user/user_test.go
```diff
@@ -253,6 +254,7 @@ this is just some garbage data
 }
 
 func TestInvalidGetExecUser(t *testing.T) {
+	// #nosec G101 (Potentially hardcoded credentials)
```
kolyshkin, line 257: ditto

libcontainer/user/user_test.go
```diff
@@ -297,6 +299,7 @@ this is just some garbage data
 }
 
 func TestGetExecUserNilSources(t *testing.T) {
+	// #nosec G101 (Potentially hardcoded credentials)
```
kolyshkin, line 302: ditto

contrib/cmd/recvtty/recvtty.go
```diff
@@ -224,7 +224,7 @@ func main() {
 		pidPath := ctx.String("pid-file")
 		if pidPath != "" {
 			pid := fmt.Sprintf("%d\n", os.Getpid())
-			if err := os.WriteFile(pidPath, []byte(pid), 0o644); err != nil {
+			if err := os.WriteFile(pidPath, []byte(pid), 0o644); err != nil { //nolint:gosec
```
AkihiroSuda, line 227: In the comment line please explain the reason for nolint

AkihiroSuda: (Same applies for all the //nolint lines in this PR)

Issue #3266, opened 2021-11-08
`integration_systemd_rootless` on Fedora began to fail


AkihiroSuda:
`integration_systemd_rootless` on Fedora began to fail after commit https://github.com/opencontainers/runc/commit/86829836480f36429403ff4c0d52414fb7493d23 (`Merge pull request #3264 from thaJeztah/cirrus_go_version`, Nov 8, 2021).

https://github.com/opencontainers/runc/runs/4135786850

```
# runc spec --rootless (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-ReiKHb/runc.4aWdGN/tty/sock test_update (status=1):
# time="2021-11-08T07:22:53Z" level=error msg="runc run failed: unable to start container process: unable to apply cgroup configuration: unable to start unit \"runc-cgroups-integration-test-28687.scope\" (properties [{Name:Description Value:\"libcontainer container integration-test-28687\"} {Name:Slice Value:\"machine.slice\"} {Name:Delegate Value:true} {Name:PIDs Value:@au [77707]} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:TasksAccounting Value:true} {Name:DefaultDependencies Value:false}]): error creating systemd unit `runc-cgroups-integration-test-28687.scope`: got `failed`"
not ok 148 update cpuset parameters via resources.CPU
# (in test file /vagrant/tests/integration/update.bats, line 476)
#   `[ "$status" -eq 0 ]' failed
# runc spec --rootless (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-ReiKHb/runc.tIcuDO/tty/sock test_update (status=1):
# time="2021-11-08T07:22:53Z" level=error msg="runc run failed: unable to start container process: unable to apply cgroup configuration: unable to start unit \"runc-cgroups-integration-test-11196.scope\" (properties [{Name:Description Value:\"libcontainer container integration-test-11196\"} {Name:Slice Value:\"machine.slice\"} {Name:Delegate Value:true} {Name:PIDs Value:@au [77798]} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:TasksAccounting Value:true} {Name:DefaultDependencies Value:false}]): error creating systemd unit `runc-cgroups-integration-test-11196.scope`: got `failed`"
not ok 149 update cpuset parameters via v2 unified map
[...]

PR #3273, opened 2021-11-12
Vagrantfile.fedora: exclude systemd from upgrade

kolyshkin:
A bug in systemd-249.6-2.fc35.x86_64 prevents rootless containers from
start when systemd manager is used (see https://bugzilla.redhat.com/show_bug.cgi?id=2022571).

A possible workaround is to exclude the package from the update list;
unfortunately this revealed another bug: "config exclude" is not working
in dnf shell (see https://bugzilla.redhat.com/show_bug.cgi?id=2022041).
A workaround to that is specify excludes as an option.

This should fix our CI for the time being.

Fixes: #3266

### DIFF
--- a/Vagrantfile.fedora
+++ b/Vagrantfile.fedora
@@ -17,8 +17,14 @@ Vagrant.configure("2") do |config|
     # Work around dnf mirror failures by retrying a few times
     for i in $(seq 0 2); do
       sleep $i
-      cat << EOF | dnf -y shell && break
-config exclude kernel,kernel-core
+      # 1. "config exclude" dnf shell command is not working in Fedora 35
+      # (see https://bugzilla.redhat.com/show_bug.cgi?id=2022571);
+      # the workaround is to specify it as an option.
+      # 2. systemd 249.6-2.fc35 has a bug preventing rootless containers
+      # from starting when --systemd-cgroup is used for runc run
+      # (see https://bugzilla.redhat.com/show_bug.cgi?id=2022041),
+      # the workaround is not to upgrade systemd.
+      cat << EOF | dnf -y --exclude=kernel,kernel-core,systemd,systemd-* shell && break
 config install_weak_deps false
 update
 install iptables gcc make golang-go glibc-static libseccomp-devel bats jq git-core criu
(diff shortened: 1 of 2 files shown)


kolyshkin:
aaaand it works! @AkihiroSuda @cyphar @thaJeztah PTAL

AkihiroSuda (approved):
Thanks

cyphar (approved):
LGTM! Thanks.

Vagrantfile.fedora
```diff
@@ -17,8 +17,14 @@ Vagrant.configure("2") do |config|
...
-      cat << EOF | dnf -y shell && break
-config exclude kernel,kernel-core
+      # 1. "config exclude" dnf shell command is not working in Fedora 35
+      # (see https://bugzilla.redhat.com/show_bug.cgi?id=2022571);
+      # the workaround is to specify it as an option.
+      # 2. systemd 249.6-2.fc35 has a bug preventing rootless containers
+      # from starting when --systemd-cgroup is used for runc run
+      # (see https://bugzilla.redhat.com/show_bug.cgi?id=2022041),
+      # the workaround is not to upgrade systemd.
+      cat << EOF | dnf -y --exclude=kernel,kernel-core,systemd,systemd-* shell && break
```
thaJeztah, line 27: (OT) I'm curious; what's the advantage of using `dnf shell` in a script over just `dnf update, dnf install..` ? (IIUC, `dnf shell` is more for interactive use?)

thaJeztah (commented):
post-merge LGTM, thanks for fixing!

Vagrantfile.fedora
kolyshkin: `dnf shell` is for transactional use. IOW, if you want to, say, install, remove, and update something in a single transaction (rather than 3 separate ones), `dnf shell` is your friend. Surely, it can be done in 3 transactions, but 1 is slightly faster as dnf does some things (reading rpmdb, calculating dependencies, writing rpmdb, and somesuch) only once rather than thrice (or twice, in the code we discuss).

thaJeztah: Ah, right, thanks for explaining; it felt a bit odd

PR #3275, opened 2021-11-12
libcontainer/system: rm Prlimit

kolyshkin:
It is now available from golang.org/x/sys/unix
(https://go-review.googlesource.com/c/sys/+/332029)


AkihiroSuda (approved):

thaJeztah (approved):
LGTM

doesn't look like there's external consumers (outside of vendored code) https://grep.app/search?q=.Prlimit&case=true

Issue #3248, opened 2021-10-19
"ro" in /dev mount options in rc93 and later


djs55:
While updating https://github.com/linuxkit/linuxkit to a more modern runc, I noticed that some containers would fail to start.

I tracked this down to linuxkit adding "ro" to the mount options for `/dev` e.g.

```
   "mounts": [
        {
            "destination": "/dev",
            "type": "tmpfs",
            "source": "tmpfs",
            "options": [
                "nosuid",
                "strictatime",
                "mode=755",
                "size=65536k",
                "ro"
            ]
        },
```

I think this works in rc92 and earlier, but now fails with an error trying to write to `/dev`.

In linuxkit I saw
```
linuxkit-0e465ebdb34b:~# runc create --bundle /containers/onboot/000-binfmt foobar
[   27.379103] IPVS: ftp: loaded support on port[0] = 21
ERRO[0000] container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:88: creating device nodes caused: read-only file system 
```

I can reproduce the same type of error (slightly different message) in 1.0.2 from the `runc spec --help` example by adding the "ro" argument:

```
$ diff -u config.json.orig config.json
--- config.json.orig	2021-10-19 07:55:59.037628359 +0100
+++ config.json	2021-10-19 07:55:43.337924319 +0100
@@ -69,7 +69,7 @@
 				"nosuid",
 				"strictatime",
 				"mode=755",
-				"size=65536k"
+				"size=65536k", "ro"
 			]
 		},
 		{

$ sudo runc run foobar
[sudo] password for dave: 
ERRO[0000] container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "devpts" to rootfs at "/dev/pts" caused: mkdir /tmp/runc/rootfs/dev/pts: read-only file system 

$ runc --version
runc version 1.0.2
commit: v1.0.2-0-g52b36a2
spec: 1.0.2-dev
go: go1.16.8
libseccomp: 2.5.1
```
If I remove the "ro" mount option it works correctly.

I propose to remove the "ro" mount option from linuxkit to fix the issue, see https://github.com/linuxkit/linuxkit/pull/3718#discussion_r731279539

The containers involved all have
```
        "root": {
                "path": "rootfs",
                "readonly": true
        },

```
which is perhaps enough "readonly-ness".

Was it a mistake for us to add "ro" in the first place?

[...]

PR #3277, opened 2021-11-12
[1.0] Fix failure with read-only /dev in spec

kolyshkin:
_This is a backport of #3276 to release-1.0 branch. Original description follows._

_NOTE: draft pending merge of ~~#3282 and~~ #3276._

----

Commit fb4c27c4b79c7d (PR #2570, went into v1.0.0-rc93) fixed a bug with
read-only tmpfs, but introduced a bug with read-only /dev (#3248).

This happens because /dev is a tmpfs mount and is therefore remounted
read-only a bit earlier than before.

To fix,

1. Revert the part of the above commit which remounts all tmpfs mounts
   as read-only in mountToRootfs.

2. Reuse finalizeRootfs (which is already used to remount /dev
   read-only) to also remount all ro tmpfs mounts that were previously
   mounted rw in mountPropagate.

Add a test case to validate the fix and prevent future regressions;
make sure it fails before the fix:
```
 ✗ runc run [ro /dev mount]
   (in test file tests/integration/mounts.bats, line 45)
     `[ "$status" -eq 0 ]' failed
   runc spec (status=0):

   runc run test_busybox (status=1):
   time="2021-11-12T12:19:48-08:00" level=error msg="runc run failed: unable to start container process: error during container init: error mounting \"devpts\" to rootfs at \"/dev/pts\": mkdir /tmp/bats-run-VJXQk7/runc.0Fj70w/bundle/rootfs/dev/pts: read-only file system"
```
Fixes: #3248

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -22,7 +22,6 @@ import (
 	"github.com/opencontainers/runc/libcontainer/devices"
 	"github.com/opencontainers/runc/libcontainer/userns"
 	"github.com/opencontainers/runc/libcontainer/utils"
-	libcontainerUtils "github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


kolyshkin:
All checkpointing tests fail. I wonder if this is something related to criu 3.16.

kolyshkin:
> All checkpointing tests fail. I wonder if this is something related to criu 3.16.

This is unrelated but we need to fix it for 1.0.3; here's the issue: https://github.com/opencontainers/runc/issues/3278

cyphar:
#3276 merged

cyphar (approved):
LGTM

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

guess this can be moved out of draft, @kolyshkin

PR #3281, opened 2021-11-12
libct/specconv: reduce init allocations, use global maps, refactor

kolyshkin:
This is a set of improvements for libct/specconv. See individual commits for details.

The refactoring should help #3227

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -414,8 +414,6 @@ func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error) 
 // systemd property name check: latin letters only, at least 3 of them
 var isValidName = regexp.MustCompile(`^[a-zA-Z]{3,}$`).MatchString
 
-var isSecSuffix = regexp.MustCompile(`[a-z]Sec$`).MatchString
-
 // Some systemd properties are documented as having "Sec" suffix
 // (e.g. TimeoutStopSec) but are expected to have "USec" suffix
 // here, so let's provide conversion to improve compatibility.
@@ -462,11 +460,16 @@ func initSystemdProps(spec *specs.Spec) ([]systemdDbus.Property, error) {
 		if err != nil {
 			return nil, fmt.Errorf("Annotation %s=%s value parse error: %w", k, v, err)
 		}
-		if isSecSuffix(name) {
-			name = strings.TrimSuffix(name, "Sec") + "USec"
-			value, err = convertSecToUSec(value)
-			if err != nil {
-				return nil, fmt.Errorf("Annotation %s=%s value parse error: %w", k, v, err)
+		// Check for Sec suffix.
+		if trimName := strings.TrimSuffix(name, "Sec"); len(trimName) < len(name) {
+			// Check for a lowercase ascii a-z just before Sec.
+			if ch := trimName[len(trimName)-1]; ch >= 'a' && ch <= 'z' {
+				// Convert from Sec to USec.
+				name = trimName + "USec"
+				value, err = convertSecToUSec(value)
+				if err != nil {
+					return nil, fmt.Errorf("Annotation %s=%s value parse error: %w", k, v, err)
+				}
 			}
 		}
 		sp = append(sp, systemdDbus.Property{Name: name, Value: value})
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

kolyshkin:
@thaJeztah @mrunalp PTAL

libcontainer/specconv/spec_linux.go
```diff
@@ -462,11 +460,16 @@ func initSystemdProps(spec *specs.Spec) ([]systemdDbus.Property, error) {
...
-				return nil, fmt.Errorf("Annotation %s=%s value parse error: %w", k, v, err)
+		// Check for Sec suffix.
+		if trimName := strings.TrimSuffix(name, "Sec"); len(trimName) < len(name) {
+			// Check for a lowercase ascii a-z just before Sec.
+			if ch := trimName[len(trimName)-1]; ch >= 'a' && ch <= 'z' {
+				// Convert from Sec to USec.
+				name = trimName + "USec"
+				value, err = convertSecToUSec(value)
+				if err != nil {
+					return nil, fmt.Errorf("Annotation %s=%s value parse error: %w", k, v, err)
```
thaJeztah, line 483: Not something new, so probably better to fix separately, but this should've been lowercase

libcontainer/specconv/spec_linux.go
```diff
@@ -411,8 +410,21 @@ func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error)
 	return mnt, nil
 }
 
-// systemd property name check: latin letters only, at least 3 of them
-var isValidName = regexp.MustCompile(`^[a-zA-Z]{3,}$`).MatchString
+// isValidName checks if systemd property name is valid. It should consists
+// of latin letters only, and have least 3 of them.
+func isValidName(s string) bool {
```
thaJeztah, line 415: Related to my other comment (no need to fix in this PR); perhaps would be nice to have this return an error with the reason for it being invalid (too short or invalid character)

libcontainer/specconv/spec_linux.go
```diff
@@ -256,6 +267,8 @@ func CreateLibcontainerConfig(opts *CreateOpts) (*configs.Config, error) {
 	config.Cgroups = c
 	// set linux-specific config
 	if spec.Linux != nil {
+		initMaps()
```
thaJeztah, line 318: I like the improvement that's brought by this, but (as a follow-up) perhaps we should consider having utilities / accessor functions for these variables, so that the initialisation happens more 'organically' (reducing the cognitive overload of having to thing about _first_ calling `initMap()` whenever we need to use one of the variables.

In this case it could be a function to check if `foo` exists in `mountPropagationMapping`

thaJeztah (approved):
LGTM, nice!

left some suggestions for follow-up improvements

libcontainer/specconv/spec_linux.go
kolyshkin: Are you aware of any linter for such things? Would be nice to have

thaJeztah: I expect the `revive` linter to be warning about it; https://github.com/opencontainers/runc/pull/2998/commits/2cca1149cd9af24f3535d38365c0e48c93bbc6e6

thaJeztah: (that's a commit from https://github.com/opencontainers/runc/pull/2998)

libcontainer/specconv/spec_linux.go
kolyshkin: Makes sense; opened #3285

libcontainer/specconv/spec_linux.go
kolyshkin: Do you mean something like
```
func getNamespaceMapping() map[specs.LinuxNamespaceType]configs.NamespaceType {
    initMaps()
    return namespaceMapping
}

// ... and so on for every other map we lazy init.
....

nsMap := getNamespaceMapping()
```

To me it looks like a bloat, since the `nsMap := getNamespaceMapping()` is just a way to sugar-coat `initMaps()`.

A runtime assertion would be nice. If only accessing a nil map would panic....

kolyshkin: I can't think of anything better than
```go
if namespaceMapping == nil {
   panic("uninitialied map") // we have a bug in the code
}
```

but it's still inferior to
```go
initMaps()
```

kolyshkin: Some kind of a linter to which we can add a rule (in the source code) like "X should be called before using Y" would also be nice to have.

kolyshkin: Ideally, a lazy way to initialize variables. IOW same as this PR does, but automatically (perhaps via an annotation).

kolyshkin: Someone worked on something like this -- alas it does not look good without generics. https://github.com/zetamatta/go-lazy

I guess, let's postpone it will Go 2.0 ))))

PR #3283, opened 2021-11-17
Fix failure with rw bind mount of a ro fuse

kolyshkin:
As reported in https://github.com/containers/podman/issues/12205, in a case where read-only fuse (sshfs)
mount is used as a volume without specifying ro flag, the kernel fails
to remount it (when adding various flags such as nosuid and nodev),
returning EPERM.

Here's the relevant strace line:

> [pid 333966] mount("/tmp/bats-run-PRVfWc/runc.RbNv8g/bundle/mnt", "/proc/self/fd/7", 0xc0001e9164, MS_NOSUID|MS_NODEV|MS_REMOUNT|MS_BIND|MS_REC, NULL) = -1 EPERM (Operation not permitted)

I was not able to reproduce it with other read-only mounts as the source
(tried tmpfs, read-only bind mount, and an ext2 mount), so somehow this
might be specific to fuse.

The fix is to check whether the source has RDONLY flag, and retry the
remount with this flag added.

A test case (which was kind of hard to write) is added, and it fails
without the fix:

```
not ok 80 runc run [bind mount of a read-only sshfs mount]
# (in test file tests/integration/mounts.bats, line 78)
# 
# runc run test_busybox (status=1):
# time="2021-11-17T20:58:16Z" level=warning msg="unable to get oom kill count" error="no directory specified for memory.oom_control"
# time="2021-11-17T20:58:16Z" level=error msg="runc run failed: unable to start container process: error during container init: error mounting \"/tmp/bats-run-5470/runc.OkQNok/bundle/mnt\" to rootfs at \"/mnt\": mount /tmp/bats-run-5470/runc.OkQNok/bundle/mnt:/mnt (via /proc/self/fd/7), flags: 0x5026: operation not permitted"
```

Note that rootless user need to be able to ssh to rootless@localhost
in order to sshfs to work -- amend setup scripts to make it work.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -1065,7 +1065,22 @@ func remount(m *configs.Mount, rootfs string, mountFd *int) error {
 	}
 
 	return utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
-		return mount(source, m.Destination, procfd, m.Device, uintptr(m.Flags|unix.MS_REMOUNT), "")
+		flags := uintptr(m.Flags | unix.MS_REMOUNT)
+		err := mount(source, m.Destination, procfd, m.Device, flags, "")
+		if err == nil {
+			return nil
+		}
+		// Check if the source has ro flag...
+		var s unix.Statfs_t
+		if err := unix.Statfs(source, &s); err != nil {
+			return &os.PathError{Op: "statfs", Path: source, Err: err}
+		}
+		if s.Flags&unix.MS_RDONLY != unix.MS_RDONLY {
+			return err
+		}
+		// ... and retry the mount with ro flag set.
+		flags |= unix.MS_RDONLY
+		return mount(source, m.Destination, procfd, m.Device, flags, "")
 	})
 }
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -99,7 +99,7 @@ task:
     # Work around dnf mirror failures by retrying a few times.
     for i in $(seq 0 2); do
       sleep $i
-      yum install -y -q gcc git iptables jq glibc-static libseccomp-devel make criu && break
+      yum install -y -q gcc git iptables jq glibc-static libseccomp-devel make criu fuse-sshfs && break
     done
     [ $? -eq 0 ] # fail if yum failed
     # install Go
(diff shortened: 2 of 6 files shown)


tests/integration/mounts.bats
```diff
@@ -52,3 +57,28 @@ function teardown() {
 	runc run test_busybox
 	[ "$status" -eq 0 ]
 }
+
+@test "runc run [rw bind mount of a ro fuse sshfs mount]" {
```
AkihiroSuda, line 61: Can we move this into `mounts_sshfs.bats`, as this has huge extra depenedncy?

Or maybe just skip the test by default and require some env var to opt-in to this test.

libcontainer/rootfs_linux.go
```diff
@@ -1065,7 +1065,22 @@ func remount(m *configs.Mount, rootfs string, mountFd *int) error {
 	}
 
 	return utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
-		return mount(source, m.Destination, procfd, m.Device, uintptr(m.Flags|unix.MS_REMOUNT), "")
+		flags := uintptr(m.Flags | unix.MS_REMOUNT)
+		err := mount(source, m.Destination, procfd, m.Device, flags, "")
+		if err == nil {
```
AkihiroSuda, line 1070: Can we rather check `errors.Is(err, syscall.EPERM)`?

kolyshkin: A different kernel _might_ return a different error, and if the original mount is not read-only, we still return that error, and if it is, a retry with added RO flag will most probably return the very same error.

In other words, the error path overhead of a statfs() and (maybe) a mount() is pretty small, and the error returned is most probably the same, so it makes sense to retry on any error.

tests/integration/mounts.bats
kolyshkin: Will do.

libcontainer/rootfs_linux.go
kolyshkin: In short, it doesn't hurt to retry even if there was an error other than EPERM.

AkihiroSuda (approved):

kolyshkin:
@thaJeztah PTAL

tests/integration/mounts.bats
kolyshkin: Done. If sshfs is not installed or not working for some reason, the test is skipped.

kolyshkin:
This fixes a real bug and we need to backport it to 1.0.3. Can this be merged? PTAL @opencontainers/runc-maintainers

thaJeztah:
Ahm, sorry, I started looking, but got pulled away

thaJeztah (approved):
LGTM

libcontainer/rootfs_linux.go
```diff
@@ -1065,7 +1065,22 @@ func remount(m *configs.Mount, rootfs string, mountFd *int) error {
...
+			return nil
+		}
+		// Check if the source has ro flag...
```
mrunalp, line 1073: Meta question: Can this be solved by requiring the entity configuring the mounts to set the necessary flags?

kolyshkin: Yes, but it's not clear who should handle this (e.g. podman thinks it shouldn't, see https://github.com/containers/podman/issues/12205) , and handling it in runc is kind of trivial. Besides, crun already does this, and this makes it hard to convince upper layers should do this.

PR #3287, opened 2021-11-18
specconv: do not permit null bytes in mount fields

cyphar:
Using null bytes as control characters for sending strings via netlink
opens us up to a user explicitly putting a null byte in a mount string
(which JSON will happily let you do) and then causing us to open a mount
path different to the one expected.

In practice this is more of an issue in an environment such as
Kubernetes where you may have path-based access control policies (which
are more susceptible to these kinds of flaws).

Found by Google Project Zero.

Fixes: 9c444070ec7b ("Open bind mount sources from the host userns")
Reported-by: Felix Wilhelm <fwilhelm@google.com>

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -2195,6 +2195,9 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
 		var mounts []byte
 		for _, m := range c.config.Mounts {
 			if m.IsBind() {
+				if strings.IndexByte(m.Source, 0) >= 0 {
+					return nil, fmt.Errorf("mount source string contains null byte: %q", m.Source)
+				}
 				mounts = append(mounts, []byte(m.Source)...)
 			}
 			mounts = append(mounts, byte(0))
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -407,6 +407,18 @@ func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error) 
 			mnt.Source = filepath.Join(cwd, m.Source)
 		}
 	}
+
+	// None of the mount arguments can contain a null byte. Normally such
+	// strings would either cause some other failure or would just be truncated
+	// when we hit the null byte, but because we serialise these strings as
+	// netlink messages (which don't have special null-byte handling) we need
+	// to block this as early as possible.
+	if strings.IndexByte(mnt.Source, 0) >= 0 ||
+		strings.IndexByte(mnt.Destination, 0) >= 0 ||
+		strings.IndexByte(mnt.Device, 0) >= 0 {
+		return nil, errors.New("mount field contains null byte")
+	}
+
 	return mnt, nil
 }
 


libcontainer/specconv/spec_linux.go
```diff
@@ -407,6 +407,19 @@ func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error)
...
+
+	// None of the mount arguments can contain a null byte. Normally such
+	// strings would either cause some other failure or would just be truncated
+	// when we hit the null byte, but because we serialise these strings as
+	// netlink messages (which don't have special null-byte handling) we need
+	// to block this as early as possible.
+	if strings.Contains(mnt.Destination, "\x00") ||
+		strings.Contains(mnt.Source, "\x00") ||
+		strings.Contains(mnt.Device, "\x00") ||
+		strings.Contains(mnt.Data, "\x00") {
```
cyphar, line 419: It might be overkill to check *all* of these. We only strictly need to check `Source` but it seems reasonable to me to check at least `Destination` and `Device` as well.

kolyshkin (dismissed):
LGTM

libcontainer/specconv/spec_linux.go
kolyshkin: I was thinking if `Data` could contain a `NULL` byte. In mount(2) it is defined as `void *data` (rather than `char *data`), and it's said to be interpreted on a per-filesystem basis, so there's a possibility that it can be something with null bytes in the middle.

Practically, though, we always treat `Data` as a string with comma-separated options, and thus it makes sense to check it for `\0` as well.

libcontainer/specconv/spec_linux.go
```diff
@@ -407,6 +407,19 @@ func createLibcontainerMount(cwd string, m specs.Mount) (*configs.Mount, error)
...
+		return nil, errors.New("mount field contains null byte")
```
thaJeztah, line 420: would it be useful to include _which_ field (for users to debug what's wrong)?

In we think it's useful, we could do a loop over the fields, and check each one

thaJeztah: That said; it would still return the first error (unless we always check all of them), so usefulness may be limited

thaJeztah (approved):
LGTM, thanks!

one suggestion, but probably ok to take this as-is (feel free to merge if we're fine with the current implementation)

libcontainer/specconv/spec_linux.go
kolyshkin: This is more of a security measure, and less a mechanism to tell the user to correct something. If there's a null byte in there, something is very wrong.

libcontainer/container_linux.go
```diff
@@ -2195,6 +2195,9 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
 		var mounts []byte
 		for _, m := range c.config.Mounts {
 			if m.IsBind() {
+				if strings.Contains(m.Source, "\x00") {
```
kolyshkin, line 2198: In other places (e.g. https://github.com/opencontainers/runc/blob/0d5ac13200f0b12198f9a84d3c64942c3cd1637d/libcontainer/init_linux.go#L126-L131) we use `strings.IndexByte(str, 0) >= 0` which is slightly more straightforward/explicit.

Maybe it's better to use it here, too.

libcontainer/specconv/spec_linux.go
cyphar: Yeah I was thinking of that too. Might be better to only check the definitely-string fields.

libcontainer/container_linux.go
cyphar: :+1:

libcontainer/specconv/spec_linux.go
cyphar: Yeah the main reason for doing this in specconv is to make sure we always reject such configurations, rather than only rejecting them when we hit the actually attack-prevention check in `bootstrapData`.

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

libcontainer/specconv/spec_linux.go
thaJeztah: Yes, it was mostly a nit; I don't think "security measure" and "informative/detailed" error is mutually exclusive, but this situation should be a corner-case, so not really worth spending too much time on to make it "nice".

PR #3286, opened 2021-11-18
Avoid non-op when the list of `Hooks` is empty

chendave:
There is no need to run hooks when `Config.Hooks` is just an empty map,

(dlv) p p.config.Config.Hooks
github.com/opencontainers/runc/libcontainer/configs.Hooks []

Signed-off-by: Dave Chen <dave.chen@arm.com>

PR #3289, opened 2021-11-18
libct/standard_init: fix linter warning

kolyshkin:
The staticcheck linter points out that the `err != nil` comparison
after `system.Exec` is always true (see https://github.com/opencontainers/runc/runs/4248411930?check_suite_focus=true; for some reason it is only shown in there):

> libcontainer/standard_init_linux.go#L253
> SA4023: this comparison is always true (staticcheck)
> libcontainer/system/linux.go#L43
> SA4023(related information): github.com/opencontainers/runc/libcontainer/system.Exec never returns a nil interface value (staticcheck)

Indeed, Exec either returns an error or does not return at all.

Remove the (useless) check.

This should unblock #3286

### DIFF
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -250,8 +250,5 @@ func (l *linuxStandardInit) Init() error {
 		return err
 	}
 
-	if err := system.Exec(name, l.config.Args[0:], os.Environ()); err != nil {
-		return err
-	}
-	return nil
+	return system.Exec(name, l.config.Args[0:], os.Environ())
 }


thaJeztah (approved):
LGTM

cyphar (approved):
LGTM (since `if err != nil` is not necessary anyway), though IMHO the linter is being dumb here -- having an `err != nil` check is not unreasonable for a function which currently does not return an error. Having the check is better than not having it, then when you later start returning an error you forget to update every caller and you start missing error returns.

kolyshkin:
Let's not backport it to 1.0 unless we'll see the error from linter.

PR #3296, opened 2021-11-29
Add `runc features` command

### LINKED ISSUE #3274: Add `runc --version-json` and print the implemented mount option strings
A high-level container application may want to obtain the list of the mount option strings supported by runc, because runc just ignores unimplemented mount option strings.

However, there is no way to obtain the list of the implemented mount option strings other than manually inspecting the code:
https://github.com/opencontainers/runc/blob/c1103d986faa82e9e6af7868b15e9904bbfd6fa6/libcontainer/specconv/spec_linux.go#L771-L824

I suggest adding `runc --version-json` command to print the machine-readable list of the implemented mount option strings.
We can put other info in this JSON eventually, too.

```console
$ runc --version-json
{
  "version": "X.Y.Z+dev",
  "commit": "vX.Y.Z-XXX-gXXXXXXXX",
  "spec": "X.Y.Z",
  "go": "goX.Y.Z",
  "libseccomp": "X.Y.Z",
  "supportedMountOptions": ["ro", "rw", "private", "rprivate", ..., "tmpcopyup"]
}
```

Relates to : 
- https://github.com/opencontainers/runc/pull/3272

AkihiroSuda:
The `runc features` command prints available features in JSON.

Fix #3274

See `types/features/features.go` for the type definition.

```json
{
    "ociVersionMin": "1.0.0",
    "ociVersionMax": "1.0.2-dev",
    "hooks": [
        "prestart",
        "createRuntime",
        "createContainer",
        "startContainer",
        "poststart",
        "poststop"
    ],
    "mountOptions": [
        "acl",
        "async",
        "atime",
        "bind",
        "defaults",
        "dev",
        "diratime",
        "dirsync",
        "exec",
        "iversion",
        "lazytime",
        "loud",
        "mand",
        "noacl",
        "noatime",
        "nodev",
        "nodiratime",
        "noexec",
        "noiversion",
        "nolazytime",
        "nomand",
        "norelatime",
        "nostrictatime",
        "nosuid",
        "private",
        "rbind",
        "relatime",
        "remount",
        "ro",
        "rprivate",
        "rshared",
        "rslave",
        "runbindable",
        "rw",
        "shared",
        "silent",
        "slave",
        "strictatime",
        "suid",
        "sync",
        "tmpcopyup",
        "unbindable"
    ],
    "linux": {
        "namespaces": [
            "cgroup",
            "ipc",
            "mount",
            "network",
            "pid",
            "user",
            "uts"
        ],
        "capabilities": [
            "CAP_CHOWN",
            "CAP_DAC_OVERRIDE",
            "CAP_DAC_READ_SEARCH",
            "CAP_FOWNER",
            "CAP_FSETID",
            "CAP_KILL",
            "CAP_SETGID",
            "CAP_SETUID",
            "CAP_SETPCAP",
            "CAP_LINUX_IMMUTABLE",
            "CAP_NET_BIND_SERVICE",
            "CAP_NET_BROADCAST",
            "CAP_NET_ADMIN",
            "CAP_NET_RAW",
            "CAP_IPC_LOCK",
            "CAP_IPC_OWNER",
            "CAP_SYS_MODULE",
            "CAP_SYS_RAWIO",
            "CAP_SYS_CHROOT",
            "CAP_SYS_PTRACE",
            "CAP_SYS_PACCT",
            "CAP_SYS_ADMIN",
            "CAP_SYS_BOOT",
            "CAP_SYS_NICE",
            "CAP_SYS_RESOURCE",
            "CAP_SYS_TIME",
            "CAP_SYS_TTY_CONFIG",
            "CAP_MKNOD",
            "CAP_LEASE",
            "CAP_AUDIT_WRITE",
            "CAP_AUDIT_CONTROL",
            "CAP_SETFCAP",
            "CAP_MAC_OVERRIDE",
            "CAP_MAC_ADMIN",
            "CAP_SYSLOG",
            "CAP_WAKE_ALARM",
            "CAP_BLOCK_SUSPEND",
            "CAP_AUDIT_READ",
            "CAP_PERFMON",
            "CAP_BPF",
            "CAP_CHECKPOINT_RESTORE"
        ],
        "cgroup": {
            "v1": true,
            "v2": true,
            "systemd": true,
            "systemdUser": true
        },
        "seccomp": {
            "enabled": true,
            "actions": [
                "SCMP_ACT_ALLOW",
                "SCMP_ACT_ERRNO",
                "SCMP_ACT_KILL",
                "SCMP_ACT_LOG",
                "SCMP_ACT_NOTIFY",
                "SCMP_ACT_TRACE",
                "SCMP_ACT_TRAP"
            ],
            "operators": [
                "SCMP_CMP_EQ",
                "SCMP_CMP_GE",
                "SCMP_CMP_GT",
                "SCMP_CMP_LE",
                "SCMP_CMP_LT",
                "SCMP_CMP_MASKED_EQ",
                "SCMP_CMP_NE"
            ],
            "archs": [
                "SCMP_ARCH_AARCH64",
                "SCMP_ARCH_ARM",
                "SCMP_ARCH_MIPS",
                "SCMP_ARCH_MIPS64",
                "SCMP_ARCH_MIPS64N32",
                "SCMP_ARCH_MIPSEL",
                "SCMP_ARCH_MIPSEL64",
                "SCMP_ARCH_MIPSEL64N32",
                "SCMP_ARCH_PPC",
                "SCMP_ARCH_PPC64",
                "SCMP_ARCH_PPC64LE",
                "SCMP_ARCH_S390",
                "SCMP_ARCH_S390X",
                "SCMP_ARCH_X32",
                "SCMP_ARCH_X86",
                "SCMP_ARCH_X86_64"
            ]
        },
        "apparmor": {
            "enabled": true
        },
        "selinux": {
            "enabled": true
        }
    },
    "annotations": {
        "io.github.seccomp.libseccomp.version": "2.5.1",
        "org.opencontainers.runc.checkpoint.enabled": "true",
        "org.opencontainers.runc.commit": "v1.0.0-402-g520702da",
        "org.opencontainers.runc.version": "1.0.0+dev"
    }
}

```

### DIFF
--- a/features.go
+++ b/features.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/opencontainers/runc/libcontainer/capabilities"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/seccomp"
+	"github.com/opencontainers/runc/libcontainer/specconv"
+	"github.com/opencontainers/runc/types/features"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/urfave/cli"
+)
+
+var featuresCommand = cli.Command{
+	Name:      "features",
+	Usage:     "show the enabled features",
+	ArgsUsage: "",
+	Description: `Show the enabled features.
+   The result is parsable as a JSON.
+   See https://pkg.go.dev/github.com/opencontainers/runc/types/features for the type definition.
+`,
+	Action: func(context *cli.Context) error {
+		if err := checkArgs(context, 0, exactArgs); err != nil {
+			return err
+		}
+
+		tru := true
+
+		feat := features.Features{
+			OCIVersionMin: "1.0.0",
+			OCIVersionMax: specs.Version,
+			Annotations: map[string]string{
+				features.AnnotationRuncVersion:           version,
+				features.AnnotationRuncCommit:            gitCommit,
+				features.AnnotationRuncCheckpointEnabled: "true",
+			},
+			Hooks:        configs.KnownHookNames(),
+			MountOptions: specconv.KnownMountOptions(),
+			Linux: &features.Linux{
+				Namespaces:   specconv.KnownNamespaces(),
+				Capabilities: capabilities.KnownCapabilities(),
+				Cgroup: &features.Cgroup{
+					V1:          &tru,
+					V2:          &tru,
+					Systemd:     &tru,
+					SystemdUser: &tru,
+				},
+				Apparmor: &features.Apparmor{
+					Enabled: &tru,
+				},
+				Selinux: &features.Selinux{
+					Enabled: &tru,
+				},
+			},
+		}
+
+		if seccomp.Enabled {
+			feat.Linux.Seccomp = &features.Seccomp{
+				Enabled:   &tru,
+				Actions:   seccomp.KnownActions(),
+				Operators: seccomp.KnownOperators(),
+				Archs:     seccomp.KnownArchs(),
+			}
+			major, minor, patch := seccomp.Version()
+			feat.Annotations[features.AnnotationLibseccompVersion] = fmt.Sprintf("%d.%d.%d", major, minor, patch)
+		}
+
+		enc := json.NewEncoder(context.App.Writer)
+		enc.SetIndent("", "    ")
+		return enc.Encode(feat)
+	},
+}
--- a/types/features/features.go
+++ b/types/features/features.go
@@ -1,0 +1,129 @@
+// Package features provides the JSON structure that is printed by `runc features` (since runc v1.1.0).
+package features
+
+// Features represents the supported features of the runtime.
+type Features struct {
+	// OCIVersionMin is the minimum OCI Runtime Spec version recognized by the runtime, e.g., "1.0.0".
+	OCIVersionMin string `json:"ociVersionMin,omitempty"`
+
+	// OCIVersionMax is the maximum OCI Runtime Spec version recognized by the runtime, e.g., "1.0.2-dev".
+	OCIVersionMax string `json:"ociVersionMax,omitempty"`
+
+	// Hooks is the list of the recognized hook names, e.g., "createRuntime".
+	// Nil value means "unknown", not "no support for any hook".
+	Hooks []string `json:"hooks,omitempty"`
+
+	// MountOptions is the list of the recognized mount options, e.g., "ro".
+	// Nil value means "unknown", not "no support for any mount option".
+	MountOptions []string `json:"mountOptions,omitempty"`
+
+	// Linux is specific to Linux.
+	Linux *Linux `json:"linux,omitempty"`
+
+	// Annotations contains implementation-specific annotation strings,
+	// such as the implementation version, and third-party extensions.
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// Linux is specific to Linux.
+type Linux struct {
+	// Namespaces is the list of the recognized namespaces, e.g., "mount".
+	// Nil value means "unknown", not "no support for any namespace".
+	Namespaces []string `json:"namespaces,omitempty"`
+
+	// Capabilities is the list of the recognized capabilities , e.g., "CAP_SYS_ADMIN".
+	// Nil value means "unknown", not "no support for any capability".
+	Capabilities []string `json:"capabilities,omitempty"`
+
+	Cgroup   *Cgroup   `json:"cgroup,omitempty"`
+	Seccomp  *Seccomp  `json:"seccomp,omitempty"`
+	Apparmor *Apparmor `json:"apparmor,omitempty"`
+	Selinux  *Selinux  `json:"selinux,omitempty"`
+}
+
+// Seccomp represents the "seccomp" field.
+type Seccomp struct {
+	// Enabled is true if seccomp support is compiled in.
+	// Nil value means "unknown", not "false".
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// Actions is the list of the recognized actions, e.g., "SCMP_ACT_NOTIFY".
+	// Nil value means "unknown", not "no support for any action".
+	Actions []string `json:"actions,omitempty"`
+
+	// Operators is the list of the recognized actions, e.g., "SCMP_CMP_NE".
+	// Nil value means "unknown", not "no support for any operator".
+	Operators []string `json:"operators,omitempty"`
+
+	// Operators is the list of the recognized archs, e.g., "SCMP_ARCH_X86_64".
+	// Nil value means "unknown", not "no support for any arch".
+	Archs []string `json:"archs,omitempty"`
+}
+
+// Apparmor represents the "apparmor" field.
+type Apparmor struct {
+	// Enabled is true if AppArmor support is compiled in.
+	// Unrelated to whether the host supports AppArmor or not.
+	// Nil value means "unknown", not "false".
+	// Always true in the current version of runc.
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+// Selinux represents the "selinux" field.
+type Selinux struct {
+	// Enabled is true if SELinux support is compiled in.
+	// Unrelated to whether the host supports SELinux or not.
+	// Nil value means "unknown", not "false".
+	// Always true in the current version of runc.
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+// Cgroup represents the "cgroup" field.
+type Cgroup struct {
+	// V1 represents whether Cgroup v1 support is compiled in.
+	// Unrelated to whether the host uses cgroup v1 or not.
+	// Nil value means "unknown", not "false".
+	// Always true in the current version of runc.
+	V1 *bool `json:"v1,omitempty"`
+
+	// V2 represents whether Cgroup v2 support is compiled in.
+	// Unrelated to whether the host uses cgroup v2 or not.
+	// Nil value means "unknown", not "false".
+	// Always true in the current version of runc.
+	V2 *bool `json:"v2,omitempty"`
+
+	// Systemd represents whether systemd-cgroup support is compiled in.
+	// Unrelated to whether the host uses systemd or not.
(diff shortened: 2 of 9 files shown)


AkihiroSuda:
cc @giuseppe for consistency with crun.

This is expected to be merged in runc first and then crun.

giuseppe:
is it something that should be discussed in the runtime-specs so that other runtimes can benefit from it?

AkihiroSuda:
> is it something that should be discussed in the runtime-specs so that other runtimes can benefit from it?

I'm not sure, as the proposed JSON contains runc-specific features such as `criu`, and perhaps some CLI-specific stuffs in the future.

types/features/features.go
```diff
@@ -0,0 +1,102 @@
+// Package features provides the JSON structure that is printed in `runc features` (since runc v1.1.0)
```
kolyshkin, line 1: nit: `s/ in / by/`.

kolyshkin: ... and a missing period at EOL.

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+package features
+
+// Features is printed in `runc features`
```
kolyshkin, line 4: ditto

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+type Features struct {
+	// Capabilities is the list of the recognized capabilities , e.g., "CAP_SYS_ADMIN"
```
kolyshkin, line 6: nit: missing period at EOL.

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+	Criu     Criu     `json:"criu"`
+
+	// Annotations contains implementation-specific annotation strings,
+	// such as the implementation version, and third-party extensions.
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// Seccomp represents the "seccomp" field
+type Seccomp struct {
+	// Enabled is true if seccomp support is compiled in
```
kolyshkin, line 27: nit: missing period at EOL.

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+	Enabled bool `json:"enabled"`
+
+	// Actions is the list of the recognized actions, e.g., "SCMP_ACT_NOTIFY"
```
kolyshkin, line 30: ditto

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+	// Nil value means "unknown", not "no support for any action".
+	Actions []string `json:"actions,omitempty"`
+
+	// Operators is the list of the recognized actions, e.g., "SCMP_CMP_NE"
```
kolyshkin, line 34: ditto

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+	// Nil value means "unknown", not "no support for any operator".
+	Operators []string `json:"operators,omitempty"`
+
+	// Operators is the list of the recognized archs, e.g., "SCMP_ARCH_X86_64"
```
kolyshkin, line 38: ditto

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Apparmor represents the "apparmor" field
+type Apparmor struct {
+	// Enabled is true if AppArmor support is compiled in
+	// Unrelated to whether the host supports AppArmor or not.
+	// Always true in the current version of runc.
+	Enabled bool `json:"enabled"`
+}
+
+// Selinux represents the "selinux" field
```
kolyshkin, line 51: ditto

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+type Selinux struct {
+	// Enabled is true if SELinux support is compiled in
```
kolyshkin, line 53: ditto

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Criu represents the "criu" field
+type Criu struct {
+	// Enabled is true if CRIU support is compiled in
+	// Unrelated to whether the host supports CRIU or not.
+	// Always true in the current version of runc.
+	Enabled bool `json:"enabled"`
+}
+
+// Cgroup represents the "cgroup" field
```
kolyshkin, line 67: ditto

kolyshkin (commented):
This looks very good overall, thank you @AkihiroSuda!

My only comments are nits -- mostly, a few missing periods at end of sentences. 

Also, I would avoid using github markdown in docstrings /  comments -- they are not formatted by godoc (go.pkg.dev, etc) and thus look weird.

types/features/features.go
kolyshkin: Does it make sense to add libselinux version here?

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+type Selinux struct {
+	// Enabled is true if SELinux support is compiled in
+	// Unrelated to whether the host supports SELinux or not.
+	// Always true in the current version of runc.
+	Enabled bool `json:"enabled"`
+}
+
+// Criu represents the "criu" field
+type Criu struct {
+	// Enabled is true if CRIU support is compiled in
```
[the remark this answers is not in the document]
kolyshkin: Does it make sense to add criu version here?

types/features/features.go
AkihiroSuda: No, because not linked in runc.

types/features/features.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// SystemdUser represents whether user-scoped systemd-cgroup support is compiled in.
+	// Unrelated to whether the host uses systemd or not.
+	// Always true in the current version of runc.
+	SystemdUser bool `json:"systemdUser"`
+}
+
+const (
+	// AnnotationRuncVersion represents the version of runc, e.g., "1.2.3", or "1.2.3+dev".
+	// Third party implementations such as crun and runsc SHOULD NOT use this annotation.
```
AkihiroSuda, line 92: (On second thought, crun MAY use this  annotation to report the most compatible runc version’

types/features/features.go
kolyshkin: You mean, statically linked? Yet it is linked, just dynamically, and runc will show whatever version is found by ld.so.

Not sure if it's needed though, just thinking out loud.

AkihiroSuda: Not linked at all, AFAICS

```console
$ ldd runc
        linux-vdso.so.1 (0x00007fffaf3e6000)
        libseccomp.so.2 => /lib/x86_64-linux-gnu/libseccomp.so.2 (0x00007f1cab638000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1cab410000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1cac0fe000)
```

types/features/features.go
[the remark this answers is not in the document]
AkihiroSuda: CRIU is an external binary, so I'd prefer not to add its version here.

If it were a library we could put its version in `Annotations` field (as in libseccomp) though.

types/features/features.go
AkihiroSuda: Reworded: `Third party implementations such as crun and runsc MAY use this annotation to report the most compatible runc version, however, parsing this annotation value is discouraged.`

cyphar:
I'm a bit apprehensive about having this be something we define outside of the runtime-spec. We do have a lot of behaviour that is out-of-spec, but given that this feature appears to be needed to work around existing runtime-spec deficiencies (several aspects of the runtime-spec use strings to specify values which the kernel accepts as bitflags -- meaning that runtimes need to have the right string-to-bitflag mappings available), having this feature entirely out-of-spec seems less than ideal.

Maybe we can prototype it outside of the runtime-spec (but if we're doing that then this feature would need to be feature-gated or otherwise have a warning saying "this may change in the future"), but long-term this should be part of the runtime-spec because not doing so would widen the (already worryingly large) gap between the runtime-spec and what a runtime actually has to implement to be a drop-in replacement for runc.

AkihiroSuda:
> Maybe we can prototype it outside of the runtime-spec (but if we're doing that then this feature would need to be feature-gated or otherwise have a warning saying "this may change in the future"), but long-term this should be part of the runtime-spec because not doing so would widen the (already worryingly large) gap between the runtime-spec and what a runtime actually has to implement to be a drop-in replacement for runc.

SGTM.

The `Annotation` strings are runc-specific but other fields can be probably moved to Runtime Spec, eventually.

AkihiroSuda:
Moved the `criu` field to an annotation, as criu is out of the scope of the Runtime Spec.

types/features/features.go
kolyshkin: Ughm, I mixed selinux with seccomp 😵‍💫. Surely we either support selinux or not, and opencontainers/selinux (unlike seccomp/libseccomp-golang) does not use any external dependency.

kolyshkin (approved):
LGTM

I think we should note in the changelog that `runc features` command is considered experimental and the output format may change before runc 1.2 release. This way, if something is fundamentally wrong, we'll be able to fix it.

kolyshkin (approved):
still LGTM 🐰

kolyshkin:
@cyphar PTAL

kolyshkin:
Very much hope we can do an rc1 this week.

cyphar:
I'll do a quick pass through all the remaining PRs and set up the rc1 release PR.

features.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/urfave/cli"
+)
+
+var featuresCommand = cli.Command{
+	Name:      "features",
+	Usage:     "show the enabled features",
+	ArgsUsage: "",
+	Description: `Show the enabled features.
+   The result is parsable as a JSON.
+   See https://pkg.go.dev/github.com/opencontainers/runc/types/features for the type definition.
```
cyphar, line 22: It would be nice to mention here that the output is subject to change, which will be slightly more visible than the changelog. But we can add that later.

cyphar (approved):
LGTM, but as mentioned above I will specify in the changelog that this is subject to change. We should probably also mention this in the help output.

AkihiroSuda:
Runtime-spec PR: https://github.com/opencontainers/runtime-spec/pull/1130

PR #3297, opened 2021-11-29
[1.0] Remove sub cgroup when container exits

kolyshkin:
This is a backport of https://github.com/opencontainers/runc/pull/3226 to release-1.0 branch, fixing issue #3225.

Note it's not a clean backport as 1.0 lacks fsMgr field, and we do not want to instantiate one, so the commit was changed:
```diff
-       err := m.fsMgr.Destroy()
-       // fsMgr.Destroy has handled ErrNotExist
+       err := cgroups.RemovePath(m.path)
+       // cgroups.RemovePath has handled ErrNotExist
```

### Proposed changelog entry

```
Bugfixes:
* cgroup v2 systemd manager: delete sub-cgroups on cgroup delete
```

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -5,7 +5,6 @@ package systemd
 import (
 	"fmt"
 	"math"
-	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -307,9 +306,10 @@ func (m *unifiedManager) Destroy() error {
 		return err
 	}
 
-	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
-	if err != nil && !os.IsNotExist(err) {
+	// systemd 239 do not remove sub-cgroups.
+	err := cgroups.RemovePath(m.path)
+	// cgroups.RemovePath has handled ErrNotExist
+	if err != nil {
 		return err
 	}
 


AkihiroSuda (approved):

libcontainer/cgroups/systemd/v2.go
```diff
@@ -307,9 +306,10 @@ func (m *unifiedManager) Destroy() error {
 		return err
 	}
 
-	// XXX this is probably not needed, systemd should handle it
-	err := os.Remove(m.path)
-	if err != nil && !os.IsNotExist(err) {
+	// systemd 239 do not remove sub-cgroups.
```
thaJeztah, line 309:
```suggestion
	// systemd 239 does not remove sub-cgroups.
```

libcontainer/cgroups/systemd/v2.go
```diff
@@ -307,9 +306,10 @@ func (m *unifiedManager) Destroy() error {
...
+	err := cgroups.RemovePath(m.path)
+	// cgroups.RemovePath has handled ErrNotExist
+	if err != nil {
```
thaJeztah, line 312: Hm.. didn't notice this when reviewing https://github.com/opencontainers/runc/pull/3226 but, given that this is the last code of the function, looks like we could just `return cgroups.RemovePath(m.path)` here.


No need to change, as this more closely matches what's in `main` (but we could clean that up)

thaJeztah (approved):
LGTM

thaJeztah:
(I see you also requested a review from @cyphar, but feel free to merge if that was just "one of you both" 😅)

kolyshkin:
Both suggestions from @thaJeztah are valid, I guess it makes sense to fix them in main branch. Here, not so much, as this is a release branch that will be hopefully be closed some time after we release 1.1.

PR #3021, opened 2021-06-10
ci: add go1.17

kolyshkin:
Keeping this as draft until go1.17 is released. Will rebase periodically to test master with go 1.17.

* Jul 13: updated to rc1.
* Aug 3: updated to rc2.
* Aug 16: updated to GA.

PR #3299, opened 2021-11-30
[1.0]: ci: add go 1.17, drop go 1.15

kolyshkin:
_This is the same as https://github.com/opencontainers/runc/pull/3021 but for release-1.0._

Add Go 1.17, grop 1.15 from CI; since 1.17 release 1.15 is unsupported.
    
Keep 1.13 in 1.0 branch, since an older version of Docker/Moby might use it.

Keep 1.16 in Dockerfile, since this is a stable branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         # Docker/Moby still builds runc with Go 1.13, so we should still support Go 1.13.
-        go-version: [1.13.x, 1.15.x, 1.16.x]
+        go-version: [1.13.x, 1.16.x, 1.17.x]
         rootless: ["rootless", ""]
         race: ["-race", ""]
 


kolyshkin:
Weird; the CI failure should be fixed by commit fdee8658dfcfdf04ca6466c5e9e58a47efc3cffa. Ahh, somehow I managed to use very old release-1.0 branch. Should be fixed now.

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         # Docker/Moby still builds runc with Go 1.13, so we should still support Go 1.13.
```
AkihiroSuda, line 20: Now we can drop 1.13

kolyshkin: I know Moby moved on, but I'm not sure about everybody else.

This is a maintenance branch, perhaps it makes sense to keep 1.13 compatibility since the cost is zero (unless we'll want to backport something massive) and it might save someone a big headache.

kolyshkin:
I have removed go 1.15 tests as required for release-1.0 branch.

TODO: add go 1.17 tests as required once this is merged.

AkihiroSuda (dismissed):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.16
+ARG GO_VERSION=1.17
```
thaJeztah, line 1: Given that 1.16 is still supported, I'm slightly leaning towards sticking to that version (and not update the default for a patch release)

kolyshkin: CI only uses `Dockerfile` for making release binaries (see `release` job in `.github/workflows/validate.yml`), and I think using latest Go for building binaries is slightly better.

Also, I'm not sure whether @cyphar will build a release via Dockerfile or not, but in case he will be using Dockerfile, sticking to latest Go for release binaries is sort of a must.

thaJeztah: I think there's risk involved in switching go runtimes. We've hit odd regressions on various Golang updates in Moby / Docker Engine, so to reduce risk I'd rather do this for `v1.1.x`, unless we have a strong reason to move to 1.17

cyphar: I don't use Docker to build the releases so I use the latest Go release on my openSUSE Tumbleweed machine, though for the cross-builds I have to use Docker. At some point (once I sit down and finally hammer out the last bits of the `release.sh` improvements) I will probably use Docker for everything.

However, I agree with @thaJeztah. I remember being burned by Go releases many times in the past, and while things do appear more stable these days, Go 1.14 broke many Go programs thanks to their preemptive scheduling changes. (Though in practice distributions will build the binaries with whatever Go version they want -- in openSUSE we track upstream and only build with the Go version used by upstream but it's possible other distributions don't.)

kolyshkin: I do remember my share of woes switching to higher Go versions, both in moby/moby and here, but the last big one was EINTR and it's a few releases back. IOW things are getting more stable, or so it seems.

Nevertheless this is a stable branch, so I agree in this case it's better to not change horses in the middle of the stream.

Reverted.

.github/workflows/test.yml
thaJeztah: Yeah, the comment is no longer "correct", but it I guess it's fine to keep the `1.13` version in the test-matrix for now (it's just a single word).

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

PR #3301, opened 2021-11-30
libct/utils: ResolveRootfs: remove

kolyshkin:
Since commit 8850636eb3df747f (February 2015) this function is no longer
used (replaced by (*ConfigValidator).rootfs), so let's remove it,
together with its unit tests (which were added by commit 917c1f6d6 in
April 2016).

### DIFF
--- a/libcontainer/utils/utils.go
+++ b/libcontainer/utils/utils.go
@@ -33,16 +33,6 @@ func init() {
 	}
 }
 
-// ResolveRootfs ensures that the current working directory is
-// not a symlink and returns the absolute path to the rootfs
-func ResolveRootfs(uncleanRootfs string) (string, error) {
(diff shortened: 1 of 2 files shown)


kolyshkin:
A very interesting (yet unrelated) failure in centos-7:
```
=== RUN   TestWriteCgroupFileHandlesInterrupt
    file_test.go:38: Failed to write 1106947 on attempt 58371: failed to write "1106947": write /sys/fs/cgroup/memory/test-eint-67994606/memory.limit_in_bytes: no such device
--- FAIL: TestWriteCgroupFileHandlesInterrupt (0.58s)
```

I wonder who stole our cgroup...

AkihiroSuda (approved):

kolyshkin:
Let's add it to 1.1 since technically this is a public API change and the earlier we do it the better.

PR #3303, opened 2021-11-30
libcontainer: optimize utils.SearchLabels

kolyshkin:
* libct/utils: SearchLabels: optimize
    
    Using strings.Split generates temporary strings for GC to collect.
    Rewrite the function to not do that.
    
    Also, add a second return value, so that the caller can distinguish
    between an empty value found and no key found cases.
    
    Fix the test accordingly.

* libct: handleCriuConfigurationFile: use utils.SearchLabels
    
    The utils.Annotations was used here before only because it made it
    possible to distinguish between "key not found" and "empty value" cases.
    
    With the previous commit, utils.SearchLabels can do that, and so it
    makes sense to use it.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -162,7 +162,10 @@ func execProcess(context *cli.Context) (int, error) {
 	if err != nil {
 		return -1, err
 	}
-	bundle := utils.SearchLabels(state.Config.Labels, "bundle")
+	bundle, ok := utils.SearchLabels(state.Config.Labels, "bundle")
+	if !ok {
+		return -1, errors.New("bundle not found in labels")
+	}
 	p, err := getProcess(context, bundle)
 	if err != nil {
 		return -1, err
--- a/libcontainer/utils/utils.go
+++ b/libcontainer/utils/utils.go
@@ -132,19 +132,16 @@ func WithProcfd(root, unsafePath string, fn func(procfd string) error) error {
 	return fn(procfd)
 }
 
-// SearchLabels searches a list of key-value pairs for the provided key and
-// returns the corresponding value. The pairs must be separated with '='.
-func SearchLabels(labels []string, query string) string {
-	for _, l := range labels {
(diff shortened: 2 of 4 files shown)


libcontainer/utils/utils.go
```diff
@@ -144,17 +144,14 @@ func WithProcfd(root, unsafePath string, fn func(procfd string) error) error {
 
 // SearchLabels searches a list of key-value pairs for the provided key and
 // returns the corresponding value. The pairs must be separated with '='.
```
thaJeztah, line 146: Can you update the docs here to describe the purpose of the bool that's returned?

exec.go
```diff
@@ -162,7 +162,10 @@ func execProcess(context *cli.Context) (int, error) {
 	if err != nil {
 		return -1, err
 	}
-	bundle := utils.SearchLabels(state.Config.Labels, "bundle")
+	bundle, ok := utils.SearchLabels(state.Config.Labels, "bundle")
+	if !ok {
+		return -1, errors.New("bundle not found in labels")
+	}
 	p, err := getProcess(context, bundle)
```
thaJeztah, line 169: Are empty values allowed here for `bundle`? (Wondering if `""` is ok to pass, so if there's an actual difference between "not set" and "no value" here)

thaJeztah (changes requested):

exec.go
kolyshkin: It's a path to container root directory, and if empty, getProcess will return an error (from chdir). I do not think we need some additional checks here.

OTOH it make sense to move this code into getProcess as it is only needed if `--process` is not given.

kolyshkin: ... but I am going to address that separately.

PTAL @thaJeztah

kolyshkin:
@thaJeztah PTAL 🙏🏻

AkihiroSuda (approved):

PR #3304, opened 2021-12-01
ci: enable extra linters for new code

kolyshkin:
This adds a new GHA CI job which runs a few extra linters. This is only
done for pull requests, and should only warn about new code.

The justification is simple: we want more linters, but since this is not
a new project, adding a new linter meaning we have to fix all the
existing warnings. In some cases having all the warnings fixed is
difficult and takes time, plus it is usually a low priority task.

Therefore, we are stuck with inability to add new linters because we
can't fix all their warnings. Meanwhile, new pull requests add more
code which is not linted.

This is an attempt to break this vicious cycle. Let's enable godot
and revive for now and see how it is going.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -23,6 +23,26 @@ jobs:
           # must be specified without patch version
           version: v1.42
 
+  lint-extra:
+    # Extra linters, only checking new code from pull requests.
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-20.04
+    permissions:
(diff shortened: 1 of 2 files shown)


kolyshkin:
It works!
Screenshot from 2021-11-30 17-38-49

Removing the test commit, moving out of draft

AkihiroSuda (approved):

kolyshkin:
@thaJeztah PTAL

mrunalp (approved):

PR #3306, opened 2021-12-03
Faster Intel RDT init if the feature is unavailable

kolyshkin:
In a (quite common) case RDT is not supported by the kernel/hardware,
it does not make sense to parse /proc/cpuinfo and /proc/self/mountinfo,
and yet the current code does it (on every runc exec, for example).
    
Fortunately, there is a quick way to check whether RDT is available --
if so, kernel creates /sys/fs/resctrl directory. Check its existence,
and skip all the other initialization if it's not present.

(plus some cleanups)


kolyshkin:
@marquiz @xiaochenshen @Creatone PTAL (I don't even have hardware to test this)

libcontainer/intelrdt/intelrdt.go
```diff
@@ -276,35 +276,30 @@ func findIntelRdtMountpointDir(f io.Reader) (string, error) {
...
+		if err != nil {
+			intelRdtRootErr = err
+			return
+		}
+		root, err := findIntelRdtMountpointDir(f)
+		f.Close()
+		if err != nil {
+			intelRdtRootErr = err
+			return
+		}
```
marquiz: Do we have to take into account that `resctrl` might get mounted after the first check?

marquiz (commented):
Thanks for the PR @kolyshkin, good improvement. I can verify this patch on a system with RDT support

xiaochenshen:
@kolyshkin Thank you for working on this improvement.
Intel RDT kernel code checks hardware capabilities via CPUID, sets corresponding CPU flags in /proc/cpuinfo if features are supported by hardware, and then creates directory `/sys/fs/resctrl/` as the *default* mount point for resctrl filesystem. So checking existence of `/sys/fs/resctrl/` could fast return on platforms which don't support Intel RDT.

FYI - Actually on Intel RDT enabled platforms, we don't have to mount resctrl filesystem into directory `/sys/fs/resctrl/` every time. Here is a rare case to mount to another mount point (e.g., `/mnt/resctrl/`):
```
mount -t resctrl resctrl /mnt/resctrl/
cat /proc/self/mountinfo | grep resctrl
63 45 0:48 / /mnt/resctrl rw,relatime shared:40 - resctrl /mnt/resctrl rw
```
In this case, we have to call `findIntelRdtMountpointDir()` in `Root()` to get the real mount point `/mnt/resctrl/` instead of `/sys/fs/resctrl/`

So in my opinion, checking existence of `/sys/fs/resctrl/` is more like a capability check rather than a root directory check. We don't have to check it inside `Root()`. How about moving it out of `Root()` to the beginning of `featuresInit()`, prior to `parseCpuInfoFile()`? We could return even earlier on platforms which don't support Intel RDT.

kolyshkin:
> How about moving it out of `Root()` to the beginning of `featuresInit()`, prior to `parseCpuInfoFile()`?

Sorry, this was my intention indeed I switched to other work and then back, and published a WIP patch as a result.

Will fix and let you know.

libcontainer/intelrdt/intelrdt.go
kolyshkin: I don't think so. If we do, we have to re-read /proc/self/mountinfo over and over (assuming it's some long-lived daemon that imports libcontainer/intelrdt, such as kubelet; for runc itself it's not a issue, since it's a short-lived binary).

I suppose that resctrl is mounted during system init, and it should be mounted by the time we start runc or e.g. kubernetes. If not, it's a configuration issue.

kolyshkin:
OK this is how I intended it -- do `/sys/fs/resctrl` existence check in Root, and call Root() the first thing from initFeatures, in order to skip everything else, if directory is not there.

NOTE we could skip parsing mountinfo entirely by doing `statfs` on `/sys/fs/resctrl` and checking the filesystem type to have resctrl magic. Unfortunately this is impossible, since we also need to know mount flags, only available from mountinfo.

kolyshkin:
@marquiz @xiaochenshen @Creatone PTAL

libcontainer/intelrdt/intelrdt.go
marquiz: K, fair enough

marquiz (approved):
lgtm

xiaochenshen (approved):
LGTM

mrunalp (approved):

Creatone (approved):
LGTM

kolyshkin:
As this modifies `intelrdt.Root` which is used by e.g. cAdvisor, and the changes are pretty straightforward, perhaps we can squeeze this into 1.1.0? WDYT @AkihiroSuda @cyphar ?

kolyshkin:
@AkihiroSuda @cyphar @thaJeztah PTAL 🙏🏻

libcontainer/intelrdt/intelrdt_test.go
```diff
@@ -127,141 +125,3 @@ func TestApply(t *testing.T) {
...
-35 25 0:31 / /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime shared:17 - cgroup cgroup rw,cpuset
-36 25 0:32 / /sys/fs/cgroup/freezer rw,nosuid,nodev,noexec,relatime shared:18 - cgroup cgroup rw,freezer
-37 25 0:33 / /sys/fs/cgroup/net_cls,net_prio rw,nosuid,nodev,noexec,relatime shared:19 - cgroup cgroup rw,net_prio,net_cls
-38 18 0:34 / /sys/kernel/config rw,relatime shared:21 - configfs configfs rw
-40 0 253:0 / / rw,relatime shared:1 - ext4 /dev/mapper/vvrg-vvrg rw,data=ordered
-16 18 0:6 / /sys/kernel/debug rw,relatime shared:23 - debugfs debugfs rw
-41 18 0:16 / /sys/fs/resctrl rw,relatime shared:24 - resctrl resctrl rw,mba_MBps
-42 20 0:36 / /dev/hugepages rw,relatime shared:25 - hugetlbfs hugetlbfs rw
-43 19 0:37 / /proc/sys/fs/binfmt_misc rw,relatime shared:26 - autofs systemd-1 rw,fd=32,pgrp=1,timeout=0,minproto=5,maxproto=5,direct,pipe_ino=35492
-44 20 0:15 / /dev/mqueue rw,relatime shared:27 - mqueue mqueue rw
```
thaJeztah, line 193: Any test-case here that we're not covering yet in the https://github.com/moby/sys repository?

thaJeztah (approved):
LGTM

PR #3309, opened 2021-12-07
release: correctly handle binary signing for "make releaseall"

### LINKED ISSUE #3038: Include ARM64 static binary in the official release artifact
We should have `runc.arm64` as well as `runc.amd64` in the official release artifact: https://github.com/opencontainers/runc/releases

The binary can be cross-compiled with `CGO_ENABLED=1 GOARCH=arm64 CC=aarch64-linux-gnu-gcc make static`

@cyphar Is there anything we need to make this happen?

cyphar:
My GPG keys are not available inside the container, so it makes little
sense to try to sign the binaries inside the container's release.sh. The
solution is to split things into separate build and sign stages, with
signing ocurring after the in-Docker build.

Fixes #3038

### DIFF
--- a/script/release_sign.sh
+++ b/script/release_sign.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+# Copyright (C) 2017 SUSE LLC.
+# Copyright (C) 2017-2021 Open Containers Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+project="runc"
+root="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+
+# Print usage information.
+function usage() {
+	echo "usage: release_sign.sh [-S <gpg-key-id>] [-r <release-dir>]" >&2
+	exit 1
+}
+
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):

script/release_sign.sh
```diff
@@ -0,0 +1,107 @@
...
+# limitations under the License.
+
+set -e
+
+project="runc"
+root="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+
+# Print usage information.
+function usage() {
+	echo "usage: release_sign.sh [-S <gpg-key-id>] [-r <release-dir>]" >&2
```
kolyshkin, line 24: This lacks some options (-H, -v)

kolyshkin:
Hmm, can't we just do something like this instead (not tested)?

```diff
diff --git a/Makefile b/Makefile
index aeb62f8c..c4aff886 100644
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ releaseall: release
 release: runcimage
        $(CONTAINER_ENGINE) run $(CONTAINER_ENGINE_RUN_FLAGS) \
                --rm -v $(CURDIR):/go/src/$(PROJECT) \
+               -v ~/.gnupg:/root/.gnupg:ro \
                -e RELEASE_ARGS=$(RELEASE_ARGS) \
                $(RUNC_IMAGE) make localrelease
```

kolyshkin:
> Hmm, can't we just do something like this instead (not tested)?

Theoretically speaking, some malicious image can stole private keys. OTOH if we trust this image in building runc binary which we are releasing, it is to be trusted.

kolyshkin (approved):
LGTM (in case you want to merge this as is, which is fine by me)

cyphar:
I use a hardware token for my GPG keys, so passing the keydir won't work.

AkihiroSuda:
Two LGTMs, is this ready to merge?

cyphar:
Yup. The help thing is a minor issue, I can fix it some other time.

PR #3311, opened 2021-12-07
[backport] chown cgroup to process uid in container namespace

frasertweedale:
(backport of https://github.com/opencontainers/runc/pull/3057)

Delegating cgroups to the container enables more complex workloads,
including systemd-based workloads.  The OCI runtime-spec was
recently updated to explicitly admit such delegation, through
specification of cgroup ownership semantics:

  https://github.com/opencontainers/runtime-spec/pull/1123

Pursuant to the updated OCI runtime-spec, change the ownership of
the container's cgroup directory and particular files therein, when
using cgroups v2 and when the cgroupfs is to be mounted read/write.

As a result of this change, systemd workloads can run in isolated
user namespaces on OpenShift when the sandbox's cgroupfs is mounted
read/write.

It might be possible to implement this feature in other cgroup
managers, but that work is deferred.

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -3,8 +3,10 @@
 package systemd
 
 import (
+	"bufio"
 	"fmt"
 	"math"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


cyphar (changes requested):
We will be doing a 1.1 release soon, which will include this patch. I don't think we have plans for another 1.0.x unless it's really necessary, so this patch doesn't need to be backported. It also is implementing a new runtime-spec feature which makes it not really eligible for backporting in the first place.

As for the safety point, there have been a lot of changes in the cgroup code in `main` which haven't been backported to `release-1.0` which also makes me slightly nervous.

frasertweedale:
OK, wasn't aware of the timeline for 1.1.  I'll close this PR.

kolyshkin:
1.0.x is a maintenance branch -- i.e. this is mostly for security and bug fixes. Unless the lack of a feature is critical, we do not add new features.

PR #3313, opened 2021-12-09
libct/int: minor improvements

kolyshkin:
1. tests/int: ability to specify binary

    This can be used to specify a different runc binary, for example:
    
            sudo -E RUNC=$PWD/runc.mine tests/integration/cwd.bats
    
    A different (but compatible enough) runtime also works:
    
            sudo -E RUNC=/usr/local/bin/crun tests/integration/cwd.bats

2. tests/int: use update_config in hooks test
    
    Using "$@" instead of $1 in update_config() allows us to use it from
    hooks.bats, where jq is used with more options than usual.

### DIFF
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -74,7 +74,7 @@ function runc_spec() {
 
 # Helper function to reformat config.json file. Input uses jq syntax.
 function update_config() {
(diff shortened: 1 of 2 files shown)


cyphar (approved):
LGTM.

AkihiroSuda (approved):

PR #3314, opened 2021-12-09
release: update libseccomp to 2.5.3, minor script fixes

kolyshkin:
1. Update libseccomp to 2.5.3
    It was released about a month ago. I don't see anything major
    in the changelog but it makes sense to keep tracking upstream deps.

2. script/release_*.sh: fix usage
3. script/seccomp.sh: fix argc check


cyphar (dismissed):
LGTM.

cyphar:
Ah this has a conflict with #3309. Since you're touching `release.sh` can you also fix the `usage` issue in `release_sign.sh`?

cyphar:
Ah this has a conflict with #3309. Since you're touching `release.sh` can you also fix the `usage` issue in `release_sign.sh`? I need to write a changelog for this release anyway, so I will be setting up the release tomorrow so if we get this merged before then I'll include this in rc1 (if not, we can include it in the actual release).

kolyshkin:
> can you also fix the `usage` issue in `release_sign.sh`?

done, PTAL @cyphar

cyphar:
DCO is missing in commit two. ;)

kolyshkin:
> DCO is missing in commit two. ;)

yup. fixed

AkihiroSuda (approved):

cyphar (approved):
LGTM

PR #3315, opened 2021-12-09
[1.0] enable cross-compilation for make releaseall

kolyshkin:
I was not aware we're going to release 1.0 with cross-compiled binaries.

Just in case we'll ever do a release in 1.0 branch, this backports some PRs that enable building proper (static) cross binaries.

PRs backported:
* #3197 
* #3230
* #3309 
* #3314 

NOTE that I do not aim to merge this unless we'll have another 1.0.x release.


kolyshkin:
We are not going to make any new 1.0.z releases, so closing

PR #3316, opened 2021-12-09
Less interfaces

kolyshkin:
This refactors the code to remove some interfaces, methods and tests that are not needed, and in general make code more comprehensible.

Splitting this one into more digestible pieces
- [x] 1. https://github.com/opencontainers/runc/pull/3352
- [x] 2. https://github.com/opencontainers/runc/pull/3353
- [x] 3. https://github.com/opencontainers/runc/pull/3354
- [x] 4. https://github.com/opencontainers/runc/pull/3374
- [x] 5. https://github.com/opencontainers/runc/pull/3373
- [x] 6. https://github.com/opencontainers/runc/pull/3375
- [ ] 7. https://github.com/opencontainers/runc/pull/3385


kolyshkin:
No longer a draft; if others agree, this can be merged right after 1.1

libcontainer/configs/config.go
```diff
@@ -83,6 +83,8 @@ type Syscall struct {
 	Args     []*Arg `json:"args"`
 }
 
+type Capabilities = specs.LinuxCapabilities
```
thaJeztah, line 86: Should mark the local `Capabilities` as deprecated ("use `specs.LinuxCapabilities` instead")?

libcontainer/container_linux.go
```diff
@@ -695,13 +683,25 @@ func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
 	return cfg
 }
 
-func (c *linuxContainer) Destroy() error {
+// Destroys the container, if its in a valid state, after killing any
```
thaJeztah, line 687:
```suggestion
// Destroy destroys the container, if its in a valid state, after killing any
```

libcontainer/container_linux.go
```diff
@@ -695,13 +683,25 @@ func (c *linuxContainer) newInitConfig(process *Process) *initConfig {
...
+// Paused containers must first be resumed using Resume(..).
+func (c *Container) Destroy() error {
 	c.m.Lock()
 	defer c.m.Unlock()
 	return c.state.destroy()
 }
 
-func (c *linuxContainer) Pause() error {
+// If the Container state is RUNNING or CREATED, sets the Container
+// state to PAUSING and pauses the execution of any user processes.
```
thaJeztah, line 702:
```suggestion
// Pause sets the Container state to PAUSING and pauses the execution of any
// user processes if the Container state is RUNNING or CREATED.
```

libcontainer/container_linux.go
```diff
@@ -35,7 +35,7 @@ import (
 
 const stdioFdCount = 3
 
-type linuxContainer struct {
+type Container struct {
```
thaJeztah, line 38: Perhaps copy the previous docs here?

```suggestion
// Container is a libcontainer container object.
//
// Each container is thread-safe within the same process. Since a container can
// be destroyed by a separate process, any function may return that the container
// was not found.
type Container struct {
```

utils_linux.go
```diff
@@ -31,24 +31,26 @@ func loadFactory(context *cli.Context) (*libcontainer.Factory, error) {
...
+	}
 
 	// We resolve the paths for {newuidmap,newgidmap} from the context of runc,
 	// to avoid doing a path lookup in the nsexec context. TODO: The binary
 	// names are not currently configurable.
-	newuidmap, err := exec.LookPath("newuidmap")
-	if err != nil {
-		newuidmap = ""
+	if newuidmap, err := exec.LookPath("newuidmap"); err == nil {
+		f.NewuidmapPath = newuidmap
```
thaJeztah, line 43: TBH; not sure I like this direction; with these changes, it looks like `loadFactory()` becomes the actual constructor, and not `New()`. I feel like either `New()` should do the constructing, or we should remove `New()` altogether if it's not really a constructor

libcontainer/factory.go
```diff
@@ -223,12 +204,10 @@ func (l *Factory) Load(id string) (*Container, error) {
 		newuidmapPath:        l.NewuidmapPath,
 		newgidmapPath:        l.NewgidmapPath,
 		cgroupManager:        cm,
+		intelRdtManager:      intelrdt.NewManager(&state.Config, id, state.IntelRdtPath),
 		root:                 containerRoot,
 		created:              state.Created,
 	}
-	if l.NewIntelRdtManager != nil {
```
thaJeztah: Perhaps squash this commit with the previous one (as it's refactoring the same parts of the code)

libcontainer/configs/config.go
kolyshkin: Hmm; perhaps we can just remove the type altogether.

libcontainer/container_linux.go
kolyshkin: What this commit did is 

> Remove BaseContainer and Container interfaces, moving their methods
> documentation to linuxContainer.

I was not changing any comments.

Also, it seems that the doc is wrong and the async part and "pausing" state was removed long time ago (I was not able to find it being used in git history).

Let's address this separately (on top of this PR once it's merged). Opened https://github.com/opencontainers/runc/issues/3343 so we won't forget.

libcontainer/container_linux.go
kolyshkin: Frankly, I do not understand the part about thread-safeness, and the rest is kind of obvious.

utils_linux.go
kolyshkin: Yes, I guess I was going that way but fell short of the last step.

libcontainer/factory.go
kolyshkin: Done (I was afraid it's too much to digest so split between two logical parts -- removing the interface and removing its usage).

utils_linux.go
kolyshkin: OK, with a few more commits here and there I think I did it. PTAL @thaJeztah 

The only question is, should we call what's left of it a factory? I can certainly rename it but can't think of any good name -- suggestions welcome.

libcontainer/container_linux.go
kolyshkin: https://github.com/opencontainers/runc/pull/3316#discussion_r789072215

kolyshkin:
Rebased; PTAL @thaJeztah

kolyshkin:
This PR lacked needed documentation changes in libcontainer/README. Added now.

kolyshkin:
@adrianreber can you please review this single commit and tell us WDYT?

https://github.com/opencontainers/runc/pull/3316/commits/5ce43c44c21364f5708431b8bc106f85b61610e6

adrianreber:
> @adrianreber can you please review this single commit and tell us WDYT?
> 
> [5ce43c4](https://github.com/opencontainers/runc/commit/5ce43c44c21364f5708431b8bc106f85b61610e6)

I don't like it :wink: This is probably only an interface used by people adding new CRIU features to runc like me, but I use it regularly. I am also using it in my (hopefully) upcoming CRI-O checkpoint/restore implementation.

It is strange to have it at the main command level, it could also work as a checkpoint or restore subcommand option.

I am probably the only user of it but I like it. I will manage without that flag, but I would prefer to keep it.

libcontainer/configs/validate/validator.go
```diff
@@ -16,30 +16,19 @@ import (
...
-func New() Validator {
-	return &ConfigValidator{}
-}
-
-type ConfigValidator struct{}
-
 type check func(config *configs.Config) error
 
-func (v *ConfigValidator) Validate(config *configs.Config) error {
+func Validate(config *configs.Config) error {
```
thaJeztah, line 21: Thanks for doing this! I recall looking at this a couple of times (why the "Validator"??)

Perhaps as a follow-up, we should move `configs/validate` to `configs`. It's a bit weird to have it in a separate package, as it's validating the config. Moving it into configs, if could just be `validate(config)` (no need to export), or

```go
func (config *configs.Config) validate() error
```

libcontainer/container_linux.go
```diff
@@ -2164,11 +2162,16 @@ func (c *Container) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Namespa
...
-			if c.config.RootlessEUID && c.newuidmapPath != "" {
-				r.AddData(&Bytemsg{
-					Type:  UidmapPathAttr,
-					Value: []byte(c.newuidmapPath),
-				})
+			if c.config.RootlessEUID {
+				// We resolve the paths for new{u,g}idmap from
+				// the context of runc to avoid doing a path
+				// lookup in the nsexec context.
+				if path, err := exec.LookPath("newuidmap"); err == nil {
```
thaJeztah, line 2167: Silly question; I haven't fully checked which codepaths are executed, but I see that `bootstrapData()` is called multiple times; can this lead to `exec.LookPath()` being called multiple times (which will be iterating over all directories in `PATH`); is that a concern? https://github.com/golang/go/blob/f229e7031a6efb2f23241b5da000c3b3203081d6/src/os/exec/lp_unix.go#L47-L57

libcontainer/factory.go
```diff
@@ -27,11 +28,15 @@ var idRegex = regexp.MustCompile(`^[\w+-\.]+$`)
...
+	absRoot, err := filepath.Abs(root)
+	if err != nil {
+		return nil, err
+	}
 	if err := os.MkdirAll(root, 0o700); err != nil {
 		return nil, err
 	}
 	return &Factory{
-		Root: root,
+		Root: absRoot,
```
thaJeztah, line 39: Wondering why we made `Root` public; after all, its expected to only be set once (when doing `libcontainer.New()`), then add a `.Root()` function.

Once it's private, we should also validate (if needed) `root` here, so that we can remove redundant checks such as
https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L145-L147

and

https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L220-L222

And remove redundant code, such as https://github.com/opencontainers/runc/blob/6d35069b5ebeb5b5fd33a7778aab368fea7d5030/list.go#L118-L122

Also, not 100% sure about this utility; perhaps it's still needed though (but it's a bit of an odd one)
https://github.com/opencontainers/runc/blob/c9b8b4f35eae6a57f64b5d9b69bd49d133833c34/utils.go#L99-L101

thaJeztah (commented):
made another pass at reviewing; looks like it needs a rebase (sorry!)


overall looking good; left some suggestions (happy to discuss)

thaJeztah:
> ⚠️ somewhere in the middle of it hides the removal of runc --criu option -- please review carefully (or I can certainly move it out to a separate PR).

I think it would make sense to move it separate (for visibility)

libcontainer/configs/config.go
```diff
@@ -261,19 +261,6 @@ func KnownHookNames() []string {
...
-	Bounding []string
-	// Effective is the set of capabilities checked by the kernel.
-	Effective []string
-	// Inheritable is the capabilities preserved across execve.
-	Inheritable []string
-	// Permitted is the limiting superset for effective capabilities.
-	Permitted []string
-	// Ambient is the ambient set of capabilities that are kept.
-	Ambient []string
-}
```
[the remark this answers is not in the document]
kolyshkin: Well, this _might_ open the hell gates:

```console
[root@kir-rhat runc]# jq < /run/runc/xxx/state.json | grep -A10 capab
    "capabilities": {
      "Bounding": [
        "CAP_AUDIT_WRITE",
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE"
      ],
      "Effective": [
        "CAP_AUDIT_WRITE",
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE"
      ],
[root@kir-rhat runc]# jq < /run/runc/xx344/state.json | grep -A10 capab
    "capabilities": {
      "bounding": [
        "CAP_AUDIT_WRITE",
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE"
      ],
      "effective": [
        "CAP_AUDIT_WRITE",
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE"
      ],
```

From what I can see, this is not used anywhere, and yet it might become problematic.

I guess I'll revert this.

kolyshkin:
OK, I am going to split this one into parts (and will update description accordingly).

libcontainer/configs/config.go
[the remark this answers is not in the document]
thaJeztah: If these files are only used by runc and/or only used by Go code, then it shouldn't (AFAICS) be an issue. `json.Unmarshal()` is case insensitive for keys (for better or worse 😅)

https://go.dev/play/p/9bo-nSvEARD

<details>

```go
package main

import (
	"encoding/json"
	"fmt"
	"reflect"
)

type Capabilities struct {
	// Bounding is the set of capabilities checked by the kernel.
	Bounding []string
	// Effective is the set of capabilities checked by the kernel.
	Effective []string
	// Inheritable is the capabilities preserved across execve.
	Inheritable []string
	// Permitted is the limiting superset for effective capabilities.
	Permitted []string
	// Ambient is the ambient set of capabilities that are kept.
	Ambient []string
}

type LinuxCapabilities struct {
	// Bounding is the set of capabilities checked by the kernel.
	Bounding []string `json:"bounding,omitempty" platform:"linux"`
	// Effective is the set of capabilities checked by the kernel.
	Effective []string `json:"effective,omitempty" platform:"linux"`
	// Inheritable is the capabilities preserved across execve.
	Inheritable []string `json:"inheritable,omitempty" platform:"linux"`
	// Permitted is the limiting superset for effective capabilities.
	Permitted []string `json:"permitted,omitempty" platform:"linux"`
	// Ambient is the ambient set of capabilities that are kept.
	Ambient []string `json:"ambient,omitempty" platform:"linux"`
}

type runc struct {
	Capabilities *Capabilities `json:"capabilities"`
}

type oci struct {
	Capabilities *LinuxCapabilities `json:"capabilities"`
}

const jsonLower = `
{
  "capabilities": {
    "bounding": [
      "CAP_AUDIT_WRITE",
      "CAP_KILL",
      "CAP_NET_BIND_SERVICE"
    ],
    "effective": [
      "CAP_AUDIT_WRITE",
      "CAP_KILL",
      "CAP_NET_BIND_SERVICE"
    ]
  }
}`

const jsonUpper = `
{
  "capabilities": {
    "Bounding": [
      "CAP_AUDIT_WRITE",
      "CAP_KILL",
      "CAP_NET_BIND_SERVICE"
    ],
    "Effective": [
      "CAP_AUDIT_WRITE",
      "CAP_KILL",
      "CAP_NET_BIND_SERVICE"
    ]
  }
}`

func main() {
	var (
		runcLower runc
		runcUpper runc
		ociLower  oci
		ociUpper  oci
	)

	if err := json.Unmarshal([]byte(jsonLower), &runcLower); err != nil {
		fmt.Println(err)
	}
	if err := json.Unmarshal([]byte(jsonUpper), &runcUpper); err != nil {
		fmt.Println(err)
	}
	if err := json.Unmarshal([]byte(jsonLower), &ociLower); err != nil {
		fmt.Println(err)
	}
	if err := json.Unmarshal([]byte(jsonUpper), &ociUpper); err != nil {
		fmt.Println(err)
	}

	if reflect.DeepEqual(runcLower, runcUpper) {
		fmt.Printf("SAME: %v <--> %v\n", runcLower.Capabilities, runcUpper.Capabilities)
	}
	if reflect.DeepEqual(ociLower, ociUpper) {
		fmt.Printf("SAME: %v <--> %v\n", ociLower.Capabilities, runcUpper.Capabilities)
	}
}
```

</details>

libcontainer/container_linux.go
kolyshkin: bootstrapData is caller either for `runc start`/`run`, or for `runc init`

libcontainer/configs/validate/validator.go
kolyshkin: > Perhaps as a follow-up, we should move `configs/validate` to `configs`.

I gave it a try and got circular deps as a result. Will take a look later, but it seems this was the reason why it's in a separate package.

Oh, and we still have to export it because it's also used from libcontainer/specconv/spec_linux_test.go

libcontainer/factory.go
kolyshkin: Yes, I was going to clean this up.

utils_linux.go
kolyshkin: > It would almost make sense to just libcontainer.Create(), libcontainer.Get(), libcontainer.List() (which internally can do a new()), adding root as argument.

This makes great sense (and we can get rid of `factory` entirely, so we don't have to think about naming it!).

kolyshkin:
As per description, this was split into smaller PRs, of which all but #3385 were merged, so this can be closed now.

PR #3319, opened 2021-12-11
libct: Create: rm unneeded chown

kolyshkin:
Commit 7cfb107f2c5be3 started using unix.Geteuid(), unix.Getegid() as
uid/gid argument for chown. It seems that it should have removed chown
entirely, since, according to mkdir(2),

> The newly created directory will be owned by the effective user ID of
> the process. If the directory containing the file has the
> set-group-ID bit set, or if the filesystem is mounted with BSD
> group semantics (mount -o bsdgroups or, synonymously mount -o grpid),
> the new directory will inherit the group ownership from its parent;
> otherwise it will be > owned by the effective group ID of the process.

So, the only effect of the chown after mkdir is ignoring the sgid bit
on the parent directory (which is probably not the right thing to do).

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -198,9 +198,6 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := os.MkdirAll(containerRoot, 0o711); err != nil {
 		return nil, err
 	}
-	if err := os.Chown(containerRoot, unix.Geteuid(), unix.Getegid()); err != nil {
-		return nil, err
-	}
 	c := &linuxContainer{
 		id:            id,
 		root:          containerRoot,


AkihiroSuda (approved):

libcontainer/factory_linux.go
```diff
@@ -198,9 +198,6 @@ func (l *LinuxFactory) Create(id string, config *configs.Config) (Container, err
 	if err := os.MkdirAll(containerRoot, 0o711); err != nil {
 		return nil, err
 	}
-	if err := os.Chown(containerRoot, unix.Geteuid(), unix.Getegid()); err != nil {
-		return nil, err
-	}
```
thaJeztah, line 203: Could there be cases where `containerRoot` already exists? (in which case `os.MkdirAll()` is  a no-op, and the directory may not have the correct permissions?

kolyshkin: @thaJeztah see above (line 158) -- if the directory exists, this function errors out.

kolyshkin: IOW the last component of `containerRoot` is guaranteed to not exist before MkdirAll

thaJeztah: > see above (line 158) -- if the directory exists, this function errors out.

Ah, thanks! Looks like I didn't scroll up far enough 👍

thaJeztah (approved):
LGTM

PR #3212, opened 2021-09-13
[1.0] rootless+cgroup2+systemd: improve error message when dbus-user-session is not installed #3186

AkihiroSuda:
Cherry-pick https://github.com/opencontainers/runc/pull/3186 (clean)

PR #3320, opened 2021-12-11
CHANGELOG: add an in-repo changelog file

cyphar:
This will make releases much simpler. I've back-filled the changelog
with everything since runc 1.0.0 (there's not much point going further
back than that).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,228 @@
+# Changelog/
+This file documents all notable changes made to this project since runc 1.0.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Deprecated
+ * runc run/start now warns if a new container cgroup is non-empty or frozen;
+   this warning will become an error in runc 1.2. (#3132, #3223)
+ * runc can only be built with Go 1.16 or later from this release onwards.
+   (#3100, #3245)
+
+### Removed
+ * `cgroup.GetHugePageSizes` has been removed entirely, and been replaced with
+   `cgroup.HugePageSizes` which is more efficient. (#3234)
+ * `intelrdt.GetIntelRdtPath` has been removed. Users who were using this
+   function to get the intelrdt root should use the new `intelrdt.Root`
+   instead. (#2920, #3239)
+
+### Added
+ * Add support for RDMA cgroup added in Linux 4.11. (#2883)
+ * runc exec now produces exit code of 255 when the exec failed.
+   This may help in distinguishing between runc exec failures
+   (such as invalid options, non-running container or non-existent
+   binary etc.) and failures of the command being executed. (#3073)
+ * runc run: new `--keep` option to skip removal exited containers artefacts.
+   This might be useful to check the state (e.g. of cgroup controllers) after
+   the container has￼exited. (#2817, #2825)
+ * seccomp: add support for `SCMP_ACT_KILL_PROCESS` and `SCMP_ACT_KILL_THREAD`
+   (the latter is just an alias for `SCMP_ACT_KILL`). (#3204)
+ * seccomp: add support for `SCMP_ACT_NOTIFY` (seccomp actions). This allows
+   users to create sophisticated seccomp filters where syscalls can be
+   efficiently emulated by privileged processes on the host. (#2682)
+ * checkpoint/restore: add an option (`--lsm-mount-context`) to set
+   a different LSM mount context on restore. (#3068)
+ * runc releases are now cross-compiled for several architectures. Static
+   builds for said architectures will be available for all future releases.
+   (#3197)
+ * intelrdt: support ClosID parameter. (#2920)
+ * runc exec --cgroup: an option to specify a (non-top) in-container cgroup
+   to use for the process being executed. (#3040, #3059)
+ * cgroup v1 controllers now support hybrid hierarchy (i.e. when on a cgroup v1
+   machine a cgroup2 filesystem is mounted to /sys/fs/cgroup/unified, runc
+   run/exec now adds the container to the appropriate cgroup under it). (#2087,
+   #3059)
+ * sysctl: allow slashes in sysctl names, to better match `sysctl(8)`'s
+   behaviour. (#3254, #3257)
+ * mounts: add support for bind-mounts which are inaccessible after switching
+   the user namespace. Note that this does not permit the container any
+   additional access to the host filesystem, it simply allows containers to
+   have bind-mounts configured for paths the user can access but have
+   restrictive access control settings for other users. (#2576)
+ * Add support for recursive mount attributes using `mount_setattr(2)`. These
+   have the same names as the proposed `mount(8)` options -- just prepend `r`
+   to the option name (such as `rro`). (#3272)
+ * Add `runc features` subcommand to allow runc users to detect what features
+   runc has been built with. This includes critical information such as
+   supported mount flags, hook names, and so on. Note that the output of this
+   command is subject to change and will not be considered stable until runc
+   1.2 at the earliest. The runtime-spec specification for this feature is
+   being developed in [opencontainers/runtime-spec#1130]. (#3296)
+
+[opencontainers/runtime-spec#1130]: https://github.com/opencontainers/runtime-spec/pull/1130
+
+### Changed
+ * system: improve performance of `/proc/$pid/stat` parsing. (#2696)
+ * cgroup2: when `/sys/fs/cgroup` is configured as a read-write mount, change
+   the ownership of certain cgroup control files (as per
+   `/sys/kernel/cgroup/delegate`) to allow for proper deferral to the container
+   process. (#3057)
+ * docs: series of improvements to man pages to make them easier to read and
+   use. (#3032)
+
+#### libcontainer API
+ * internal api: remove internal error types and handling system, switch to Go
+   wrapped errors. (#3033)
+ * New configs.Cgroup structure fields (#3177):
+   * Systemd (whether to use systemd cgroup manager); and
+   * Rootless (whether to use rootless cgroups).
+ * New cgroups/manager package aiming to simplify cgroup manager instantiation.
+   (#3177)
+ * All cgroup managers' instantiation methods now initialize cgroup paths and
+   can return errors. This allows to use any cgroup manager method (e.g.
+   Exists, Destroy, Set, GetStats) right after instantiation, which was not
+   possible before (as paths were initialized in Apply only). (#3178)
+
+### Fixed
+ * nsenter: do not try to close already-closed fds during container setup and
+   bail on close(2) failures. (#3058)
+ * runc checkpoint/restore: fixed for containers with an external bind mount
+   which destination is a symlink. (#3047).
+ * cgroup: improve openat2 handling for cgroup directory handle hardening.
+   (#3030)
+ * `runc delete -f` now succeeds (rather than timing out) on a paused
+   container. (#3134)
+ * runc run/start/exec now refuses a frozen cgroup (paused container in case of
+   exec). Users can disable this using `--ignore-paused`. (#3132, #3223)
+ * config: do not permit null bytes in mount fields. (#3287)
+
+
+## [1.0.3] - 2021-12-06
+
+> If you were waiting for the opportune moment, that was it.
+
+### Security
+ * A potential vulnerability was discovered in runc (related to an internal
+   usage of netlink), however upon further investigation we discovered that
+   while this bug was exploitable on the master branch of runc, no released
+   version of runc could be exploited using this bug. The exploit required being
+   able to create a netlink attribute with a length that would overflow a uint16
+   but this was not possible in any released version of runc. For more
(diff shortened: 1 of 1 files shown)


cyphar:
The idea is that we will always keep the main branch CHANGELOG accurate (we can add `unreleased 1.0.z` sections for patches that have been backported) and people submitting patches to runc which are `impact/changelog` need to add their changelog entry to the file.

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
+# Changelog
+All notable changes to this project will be documented in this file.
```
kolyshkin, line 2: `s/will be/are/`?

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+   This might be useful to check the state (e.g. of cgroup controllers) after
+   the container has￼exited. (#2817 #2825)
+ * seccomp: add support for `SCMP_ACT_KILL_PROCESS` and `SCMP_ACT_KILL_THREAD`
+   (the latter is just an alias for `SCMP_ACT_KILL`). (#3204)
+ * seccomp: add support for `SCMP_ACT_NOTIFY` (seccomp actions). This allows
+   users to create sophisticated seccomp filters where syscalls can be
+   efficiently emulated by privileged processes on the host. (#2682)
+ * checkpoint/restore: add an option (`--lsm-mount-context`) to set
+   a different LSM mount context on restore. (#3068)
+ * runc releases now are cross-compiled for several architectures and static
```
kolyshkin, line 32: Feels like a missing comma before `and` (it's a compound sentence)

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Added
+ * Add support for RDMA cgroup added in Linux 4.11. (#2883)
+ * runc exec now produces exit code of 255 when the exec failed.
+   This may help in distinguishing between runc exec failures
+   (such as invalid options, non-running container or non-existent
+   binary etc.) and failures of the command being executed. (#3073)
+ * runc run: new `--keep` option to skip removal exited containers artefacts.
+   This might be useful to check the state (e.g. of cgroup controllers) after
+   the container has￼exited. (#2817 #2825)
```
kolyshkin, line 24: Earlier you used a comma to separate two PR IDs. I think it looks better with a comma, but in any case it's better to be consistent.

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+   runc has been built with. This includes critical information such as
+   supported mount flags, hook names, and so on. Note that the output of this
+   command is subject to change and will not be considered stable until runc
+   1.2 at the earliest. (#3296)
+
+### Changed
+ * system: improve performance of `/proc/$pid/stat` parsing. (#2696)
+ * cgroup2: when `/sys/fs/cgroup` is configured as a read-write mount, change
+   the ownership of certain cgroup control files (as per
+   `/sys/kernel/cgroup/delegate`) to allow for proper defferal to the container
```
kolyshkin, line 61: typo: `s/defferal/deferral/`

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+   possible before (as paths were initialized in Apply only). (#3178)
+
+### Fixed
+ * nsenter: do not try to close already-closed fds during container setup and
+   bail on close(2) failures. (#3058)
+ * runc checkpoint/restore: fixed for containers with an external bind mount
+   which destination is a symlink. (#3047).
+ * cgroup: improve openat2 handling for cgroup directory handle hardening.
+   (#3030)
+ * `runc delete -f` now succeeds (rather than timing out) on a paused
```
kolyshkin, line 84: `s/timing out/timeouts/`

Or "runc is now succeeding (rather than timing out)"...

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+   container. (#3134)
+ * runc run/start/exec now refuses a frozen cgroup (paused container in case of
+   exec). (#3132, #3223)
```
kolyshkin, line 87: Makes sense to also mention (in Added above) that runc exec adds `--ignore-paused` option to force running an exec in a paused container (frozen cgroup).

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Fixed
+ * Fixed inability to start a container with read-write bind mount of a
+   read-only fuse host mount. (#3283 #3292)
+ * Fixed inability to start when read-only /dev in set in spec (#3276 #3277)
+ * Fixed not removing sub-cgroups upon container delete, when rootless cgroup v2
+   is used with older systemd. (#3226 #3297)
+ * Fixed returning error from GetStats when hugetlb is unsupported (which causes
+   excessive logging for Kubernetes). (#3233 #3295)
+
```
kolyshkin, line 112: There's also a small improvement (clearer error message when dbus-user-session is not installed and
rootless + cgroup2 + systemd are used (#3212)

kolyshkin: Should go under "Changed" or "Added" I guess.

CHANGELOG.md
kolyshkin: I'd say

This file documents all notable changes to the project since runc 1.0.

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Deprecated
+ * runc run/start now warns if a new container cgroup is non-empty or frozen;
+   this warning will become an error in runc 1.2. (#3132, #3223)
+
```
kolyshkin, line 11: Should we add something like "go releases earlier than 1.16 can no longer be used to build this (#3100, #3245)"?

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+### Removed
+ * `cgroup.GetHugePageSizes` has been removed entirely, and been replaced with
+   `cgroup.HugePageSizes` which is more efficient. (#3234)
+
```
kolyshkin, line 15: * intelrdt: `GetIntelRdtPath` has been removed (#2920)

and below in Added

* intelrdt: added `Root`(#3239)

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+ * cgroup v1 controllers now support hybrid hierarchy (i.e. when on a cgroup v1
+   machine a cgroup2 filesystem is mounted to /sys/fs/cgroup/unified, runc
+   run/exec now adds the container to the appropriate cgroup under it). (#2087
+   #3059)
+ * sysctl: allow slashes in sysctl names, to better match `sysctl(8)`'s
+   behaviour. (#3254 #3257)
+ * mounts: when using user namespaces, open bind-mount sources from the host
+   namespace and use those file descriptors as the mount sources to allow runc
+   to create bind-mounts to paths that the container userns cannot access after
+   switching. (#2576)
```
kolyshkin, line 47: Not sure we have to dig deep into the implementation details. Can be simplified to

* Add support for bind mounts that are inaccessible from the container user namespace (#2576)

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+   supported mount flags, hook names, and so on. Note that the output of this
+   command is subject to change and will not be considered stable until runc
+   1.2 at the earliest. (#3296)
+
+### Changed
+ * system: improve performance of `/proc/$pid/stat` parsing. (#2696)
+ * cgroup2: when `/sys/fs/cgroup` is configured as a read-write mount, change
+   the ownership of certain cgroup control files (as per
+   `/sys/kernel/cgroup/delegate`) to allow for proper defferal to the container
+   process. (#3057)
```
kolyshkin, line 62: Also


* config: allow using `/` instead of `.` in systctl names (#3254, #3257)

and

* Better man pages (#3032)

kolyshkin (approved):
LGTM overall; added some nits that can either be addressed here, or post-merge if we want to release am rc faster.

CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
...
+   to create bind-mounts to paths that the container userns cannot access after
+   switching. (#2576)
+ * Add support for recursive mount attributes using `mount_setattr(2)`. These
+   have the same names as the proposed `mount(8)` options -- just prepend `r`
+   to the option name (such as `rro`). (#3272)
+ * Add `runc features` subcommand to allow runc users to detect what features
+   runc has been built with. This includes critical information such as
+   supported mount flags, hook names, and so on. Note that the output of this
+   command is subject to change and will not be considered stable until runc
+   1.2 at the earliest. (#3296)
```
AkihiroSuda, line 55: Could you link the runtime spec PR https://github.com/opencontainers/runtime-spec/pull/1130

CHANGELOG.md
cyphar: Timeouts doesn't sound right to me -- "timeouts" is a plural, it's not the same thing as "times out" -- that's why I changed it from the original changelog.

CHANGELOG.md
cyphar: I copied the header from https://keepachangelog.com/en/1.0.0/, but we can definitely customize it. I don't think it sounds wrong to use the future tense, since it's a living document, but either works (though the benefit of future tense is that it also acts as an instruction for contributors -- though that's not necessary).

CHANGELOG.md
kolyshkin: If you don't like "timeouts" as a verb, use "times out". In the current form, we're mixing present simple ("succeeds") and present continuous ("timing out"), and it also feels like "is" is missing.

So, I suggest either

* `runc delete -f` now succeeds (rather than times out) on a paused ...

or

* `runc delete -f` is now succeeding (rather than timing out) on a paused ...

cyphar: To be honest (as a native speaker) all of the options you've listed sound wrong as complete sentences. I think it's because of "rather than" being included in the mix -- if I'd written "succeeds or timing out" you would be entirely correct saying it's wrong (and the two choices you listed would be correct -- "succeeds or times out" or "succeeding or timing out" -- though in this context the first fits better) but that's not the sentence.

CHANGELOG.md
cyphar: Not sure I agree, but I reworded it to be two sentences. :smile:

CHANGELOG.md
cyphar: Yeah I'm not sure which looks better. Maybe it's better with commas.

CHANGELOG.md
cyphar: Good point, but I'll keep it in the same section.

CHANGELOG.md
cyphar: Ah, it wasn't in the 1.0.3 release notes. Will add.

cyphar: Oops, I missed the "enhancements" section of the release notes.

CHANGELOG.md
cyphar: I'll include them in one line since it's an API thing (as with the huge page changes) and we aren't generally advertising our API as something people should be using,.

CHANGELOG.md
cyphar: I'll reword it a bit, but given it's a fairly complicated change that might seem like a security concern I thought it warranted a bit more explanation than the headline description.

CHANGELOG.md
cyphar: The first one was already mentioned in the unreleased notes (it wasn't backported AFAICS). Will add the other one.

CHANGELOG.md
cyphar: I switch to:

>  * mounts: add support for bind-mounts which are inaccessible after switching
   the user namespace. Note that this does not permit the container any
   additional access to the host filesystem, it simply allows containers to
   have bind-mounts configured for paths the user can access but have
   restrictive access control settings for other users. (#2576)

AkihiroSuda (approved):

thaJeztah:
> and people submitting patches to runc which are impact/changelog need to add their changelog entry to the file.

In general, I applaud that effort; it's good to have the changelog entry written together with the pull request. That said, (of course depending somewhat on how many changes are relevant to the changelog), but I know a couple of repositories that take this approach, and it's often "merge conflict hell" (which would potentially happen _twice_; once for the main branch, and one for the backport).

Not sure what a good solution is though (also because "changelog-worthy" can be a bit of a grey area; a bug fix may initially seem "minor", but later turn out to be affecting many users, so worthy to mention).

GitHub's auto-generated release-notes may help (to some extend). Other approaches I've seen is for contributors to add a new file in a `changelogs/` directory (to prevent the conflicts), which then get included in a changelog, but that's a bit "quirky" / "noisy" as well.

thaJeztah (approved):
This change itself LGTM

(but perhaps we can find a solution to the above, but that can be resolved / discussed separately)

CHANGELOG.md
kolyshkin: Right, as of this morning I agree that "rather than" changes the structure; thanks for the explanation!

kolyshkin (approved):
LGTM

cyphar:
Yeah maybe changelog/ would be less prone to conflicts but I think we can revisit the issue if we end up with a lot of conflicts. Having something is better than nothing imho.

PR #3322, opened 2021-12-13
RFC: Set arbitrary uid gid in user namespace

AlexeyPerevalov:
Currently in case of rootless scenario which implies usernamespace creation. The only one user inside user namespace which is allowed - it's root user.

But for scenarios with higher level of security hardening more fine grained restriction could be required. For example none root in user namespace.

This PR removes hardcoded 0 uid in setresuid/setuid.
It uses config.process.user.[uid|gid] as uid|gid inside usernamesace.
But using process.user. is questionable.

Since it's RFC, unit tests arn't affected.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -90,10 +90,13 @@ struct nlconfig_t {
 	size_t uidmappath_len;
 	char *gidmappath;
 	size_t gidmappath_len;
+	uint32_t uid;
+	uint32_t gid;
 
 	/* Mount sources opened outside the container userns. */
 	char *mountsources;
 	size_t mountsources_len;
+
 };
 
 /*
@@ -127,6 +130,8 @@ static int loglevel = DEBUG;
 #define UIDMAPPATH_ATTR		27288
 #define GIDMAPPATH_ATTR		27289
 #define MOUNT_SOURCES_ATTR	27290
+#define UID_ATTR                27291
+#define GID_ATTR                27292
 
 /*
  * Use the raw syscall for versions of glibc which don't include a function for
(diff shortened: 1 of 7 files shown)


cyphar:
The reason we didn't support this from the outset is that some of the bits during container setup where you have to configure the namespaces won't really work if you `setresuid` to a non-root user -- doing that drops all of your capabilities. I suspect it should be possible still configure such a container, but I believe it'll take a bit more work than this PR.

I also suspect some kernel interfaces will struggle with a namespace that has no root user -- v3 file capabilities for instance (which embed the rootuid of the namespace and use that to check whether the capability should be applied). There are also a fair few "are you root" checks both in the kernel and userspace that would be confused by this too.

libcontainer/nsenter/nsexec.c
```diff
@@ -1210,7 +1227,10 @@ void nsexec(void)
 				}
 
 				/* Become root in the namespace proper. */
-				if (setresuid(0, 0, 0) < 0)
+				write_log(DEBUG, "Become root in the namespace proper.");
+				write_log(DEBUG, "uid: %d\n", config.uid);
+				// TODO get it from config
+				if (setresuid(config.uid, config.uid, config.uid) < 0)
```
cyphar, line 1233: In particular I'm not sure this will work -- and even if it did work you can't use `config.uid` here for containers that have a root user because in those cases you definitely wouldn't be able to configure the container properly. Maybe we don't need to setuid for user namespaces before we unshare the other namespaces? I'm not sure.

AlexeyPerevalov: You right about ambient capabilities. I decided to put it into separate PR.
https://github.com/AlexeyPerevalov/runc/commit/f1f54ad9905f22dcb912d914ccf48168aa405730
Not yet ready - since for me it looks hacky.
The main intention here was also be able to run runc as non root user.

For containers that have a root user - do you mean container where we're running process as root or where FS has a root in files ownership? If the latest one - yes, for testing I "repacked" images and rewrote perm's of the files, images which are just pulled from public domain - are not suitable, unless some shifting fs is not applied (personally I didn't try)

AlexeyPerevalov:
> The reason we didn't support this from the outset is that some of the bits during container setup where you have to configure the namespaces won't really work if you `setresuid` to a non-root user -- doing that drops all of your capabilities. I suspect it should be possible still configure such a container, but I believe it'll take a bit more work than this PR.
> 
> I also suspect some kernel interfaces will struggle with a namespace that has no root user -- v3 file capabilities for instance (which embed the rootuid of the namespace and use that to check whether the capability should be applied). There are also a fair few "are you root" checks both in the kernel and userspace that would be confused by this too.

Thank you very much for feedback.
The reason I did it
1) estimate amount of problems/posibilities in case  of supporting user namespace in k8s
2) ... for case when containerd/runc is running with non root user
Quit different directions.

PR #3325, opened 2021-12-14
libct/cg: rm go 1.15 compatibility

kolyshkin:
Since commit 12e99a0f8df0e6ac59 (PR #3245) we do require Go >= 1.16,
so this file is no longer needed.

Also, this actually ensures that go >= 1.16 is used (otherwise
libcontainer/cgroups/getallpids.go won't compile).

Tentatively including this into 1.1 milestone.

Found while vendoring runc 1.1rc into cri-o.


cyphar (approved):
LGTM.

AkihiroSuda (approved):

PR #3337, opened 2022-01-09
cgroups: Check that the given blkioThrottle.Rate isn't 0

skyshang1:
Following PR add verification to cgroup fs manager, avoid errors like `numerical result out of range` when writing cgroup file related to blkio throttle.

### DIFF
--- a/libcontainer/cgroups/fs/blkio.go
+++ b/libcontainer/cgroups/fs/blkio.go
@@ -50,23 +50,31 @@ func (s *BlkioGroup) Set(path string, r *configs.Resources) error {
 		}
 	}
 	for _, td := range r.BlkioThrottleReadBpsDevice {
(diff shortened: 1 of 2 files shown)


kolyshkin:
According to runtime-spec (see https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#block-io), both major/minor and rate are required. If major/minor is set, but rate is not, this is invalid configuration and should be rejected / cause an error. This is what runc currently does.

Why do you think this should be changed, @skyshang1?

kolyshkin:
No response from the reporter, and I consider this is not a bug, thus closing.

@skyshang1 let me know if you feel otherwise.

skyshang1:
> 
I have a question that should we do defensive programming？Or at least validate config.json to give clear error messages.

PR #3340, opened 2022-01-19
Minor cleanups in libcontainer and loadFactory

thaJeztah:
#### libct.IntelRdtFs(): simplify

- remove unneeded wrapping func, as intelrdt.NewManager already has the correct signature.
- reverse the if/else to be clearer that either "IsCATEnabled()" and/or  "IsMBAEnabled()" need to be enabled (and to allow skipping one of those checks in the condition).

#### loadFactory(): remove unneeded intermediate variable

### DIFF
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -48,16 +48,14 @@ func InitArgs(args ...string) func(*LinuxFactory) error {
 	}
 }
 
-// IntelRdtfs is an options func to configure a LinuxFactory to return
+// IntelRdtFs is an options func to configure a LinuxFactory to return
 // containers that use the Intel RDT "resource control" filesystem to
 // create and manage Intel RDT resources (e.g., L3 cache, memory bandwidth).
 func IntelRdtFs(l *LinuxFactory) error {
-	if !intelrdt.IsCATEnabled() && !intelrdt.IsMBAEnabled() {
-		l.NewIntelRdtManager = nil
+	if intelrdt.IsCATEnabled() || intelrdt.IsMBAEnabled() {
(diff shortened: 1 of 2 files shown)


libcontainer/factory_linux.go
```diff
@@ -48,16 +48,14 @@ func InitArgs(args ...string) func(*LinuxFactory) error {
...
 
-// IntelRdtfs is an options func to configure a LinuxFactory to return
+// IntelRdtFs is an options func to configure a LinuxFactory to return
 // containers that use the Intel RDT "resource control" filesystem to
 // create and manage Intel RDT resources (e.g., L3 cache, memory bandwidth).
 func IntelRdtFs(l *LinuxFactory) error {
-	if !intelrdt.IsCATEnabled() && !intelrdt.IsMBAEnabled() {
-		l.NewIntelRdtManager = nil
+	if intelrdt.IsCATEnabled() || intelrdt.IsMBAEnabled() {
+		l.NewIntelRdtManager = intelrdt.NewManager
```
thaJeztah, line 56: FWIW; wondering why `NewIntelRdtManager` (and other fields in `LinuxFactory` are exported; perhaps they should not be)

kolyshkin:
Can we please have this one after https://github.com/opencontainers/runc/pull/3316? That one is a major refactoring and is removing some stuff which is modified here.

thaJeztah:
Yes! Fine to do it after, if anything is left then (I didn't see those changes until after I created this PR ☺️)

kolyshkin:
It looks like there's nothing left to fix after #3354 (commit dbd990d5559733bd0f33e2bc23facd2e866cac53)

PR #3353, opened 2022-01-27
runc: remove --criu option

kolyshkin:
This was introduced in an initial commit, back in the day when criu was
a highly experimental thing. Today it's not; most users who need it have
it packaged by their distro vendor.

The usual way to run a binary is to look it up in directories listed in
$PATH. This is flexible enough and allows for multiple scenarios (custom
binaries, extra binaries, etc.). This is the way criu should be run.

Make `--criu` a hidden option (thus removing it from help). Remove the
option from man pages, integration tests, etc. Remove all traces of
CriuPath from data structures.

Add a warning that `--criu` is ignored and will be removed.

_This used to be part of #3316._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
+
+ * `runc` option `--criu` is now ignored (with a warning), and the option will
+   be removed entirely in a future release. Users who need a non-standard
+   `criu` binary should rely on the standard way of looking up binaries in
+   `$PATH`. (#3316)
+
 ### Changed
 
  * When Intel RDT feature is not available, its initialization is skipped,
--- a/contrib/completions/bash/runc
+++ b/contrib/completions/bash/runc
@@ -231,12 +231,11 @@ _runc_runc() {
 		--log
 		--log-format
 		--root
-		--criu
 		--rootless
 	"
 
 	case "$prev" in
-	--log | --root | --criu)
+	--log | --root)
 		case "$cur" in
 		*:*) ;; # TODO somehow do _filedir for stuff inside the image, if it's already specified (which is also somewhat difficult to determine)
 		'')
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -45,7 +45,6 @@ type linuxContainer struct {
 	initArgs             []string
 	initProcess          parentProcess
 	initProcessStartTime uint64
-	criuPath             string
 	newuidmapPath        string
 	newgidmapPath        string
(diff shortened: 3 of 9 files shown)


kolyshkin:
This should be done 5 years ago but I guess better late than never.

runc also uses `ps`, `busctl`, `systemctl`, `newuidmap`/`newgidmap` etc, but do not provide any options to configure those -- $PATH is sufficient and flexible enough.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
```
AkihiroSuda, line 9: Can be another PR, but maybe we should have `docs/deprecation.md` in addition to [`docs/experimental.md`](https://github.com/opencontainers/runc/blob/master/docs/experimental.md) ?

kolyshkin:
Now, an alternative would be to honor `--criu` but warn that it will be removed. I'd rather not do it, because:
 - most probably no one is using this option;
 - whoever uses it can employ a quick and easy fix:

```diff 
-runc --criu /some/custom/criu/dir/criu ...
+PATH=/some/custom/criu/dir:$PATH runc
```

The only question is, should using `--criu` result in an error  (rather than a warning).

CHANGELOG.md
kolyshkin: I don't know. It seems easy enough to search CHANGELOG.md for `eprecat`.

rhatdan:
LGTM

thaJeztah:
I guess this is fine, as the flag is still present (so things won't hard-fail).

I stumbled upon this option in containerd, which (I think) may be using this flag; https://github.com/containerd/containerd/blob/77d53d2d230c3bcd3f02e6f493019a72905c875b/cmd/containerd-shim/main_unix.go#L60

Originally added in https://github.com/containerd/containerd/commit/ab0cb4e756206158e32e1a5c7e6c00b6a73b30e8, and I can see the `--criu` flag being set in containerd/go-runc; https://github.com/containerd/go-runc/blob/1c99ade6c04ed41bc4c9a8ac12b81b5fd4e14bcc/runc.go#L730-L732

So that code probably should be removed, and the option deprecated on containerd's side as well, @AkihiroSuda

thaJeztah:
Opened https://github.com/containerd/go-runc/pull/82 and https://github.com/containerd/containerd/pull/6496

crosbymichael (approved):

thaJeztah (approved):
LGTM

thaJeztah:
@AkihiroSuda @cyphar good to go?

kolyshkin:
@adrianreber PTAL

AkihiroSuda (approved):

PR #3354, opened 2022-01-27
Factory cleanup

kolyshkin:
_This used to be part of #3316._

This set cleanups LinuxFactory, removing some interfaces, option functions etc. Lots of changes, but mostly removal.

This removes about 6K of code and 5K of data from the binary, reducing its overall size by 0.1% (not counting debuginfo, i.e. for a stripped binary) without any change in functionality.

More importantly, this improves readability a notch by simplifying some code and data structures, and removing about 200 lines of code and tests.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -41,7 +41,7 @@ type linuxContainer struct {
 	root                 string
 	config               *configs.Config
 	cgroupManager        cgroups.Manager
-	intelRdtManager      intelrdt.Manager
+	intelRdtManager      *intelrdt.Manager
 	initProcess          parentProcess
 	initProcessStartTime uint64
 	m                    sync.Mutex
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -41,8 +41,6 @@ type linuxContainer struct {
 	config               *configs.Config
 	cgroupManager        cgroups.Manager
 	intelRdtManager      intelrdt.Manager
-	initPath             string
-	initArgs             []string
 	initProcess          parentProcess
 	initProcessStartTime uint64
 	newuidmapPath        string
@@ -481,8 +479,8 @@ func (c *linuxContainer) newParentProcess(p *Process) (parentProcess, error) {
 }
 
 func (c *linuxContainer) commandTemplate(p *Process, childInitPipe *os.File, childLogPipe *os.File) *exec.Cmd {
-	cmd := exec.Command(c.initPath, c.initArgs[1:]...)
-	cmd.Args[0] = c.initArgs[0]
+	cmd := exec.Command("/proc/self/exe", "init")
+	cmd.Args[0] = os.Args[0]
 	cmd.Stdin = p.Stdin
 	cmd.Stdout = p.Stdout
 	cmd.Stderr = p.Stderr
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -43,8 +43,6 @@ type linuxContainer struct {
 	intelRdtManager      intelrdt.Manager
 	initProcess          parentProcess
 	initProcessStartTime uint64
-	newuidmapPath        string
-	newgidmapPath        string
 	m                    sync.Mutex
 	criuVersion          int
 	state                containerState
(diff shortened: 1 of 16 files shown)


libcontainer/container_linux.go
```diff
@@ -2148,11 +2146,16 @@ func (c *linuxContainer) bootstrapData(cloneFlags uintptr, nsMaps map[configs.Na
...
-			if c.config.RootlessEUID && c.newuidmapPath != "" {
-				r.AddData(&Bytemsg{
-					Type:  UidmapPathAttr,
-					Value: []byte(c.newuidmapPath),
-				})
+			if c.config.RootlessEUID {
+				// We resolve the paths for new{u,g}idmap from
+				// the context of runc to avoid doing a path
+				// lookup in the nsexec context.
+				if path, err := exec.LookPath("newuidmap"); err == nil {
```
kolyshkin, line 2153: Not really related to this PR, but it _may_ makes sense to bail out right here.

We are rootless, and have a UID mapping, so in the absence of newuidmap/newgidmap nsexec.c is doomed to fail anyway. Or could it be the case when writing to `/proc/<pid>/uid_map might succeed? I don't think so but I might be wrong.

WDYT @AkihiroSuda

kolyshkin:
Release job failed because of a network glitch:

> E: Failed to fetch https://provo-mirror.opensuse.org/repositories/devel:/tools:/criu/Debian_11/amd64/criu_3.16.1-3_amd64.deb  > Could not connect to provo-mirror.opensuse.org:443 (91.193.113.70). - connect (111: Connection refused)

Restarted.

kolyshkin:
@opencontainers/runc-maintainers PTAL (I have more changes pending that depend on these commits)

libcontainer/container_linux.go
thaJeztah: Should we use `golang.org/x/sys/execabs` for these? (https://github.com/golang/sys/tree/master/execabs)

See https://github.com/docker/cli/pull/2950 -  `golang.org/x/sys/execabs` is an updated version of `os/exec` that fixes the security issue that was found in Go 1.15.7; https://blog.golang.org/path-security

thaJeztah (commented):
One comment, but otherwise looks good

kolyshkin:
> Should we use `golang.org/x/sys/execabs` for these? (https://github.com/golang/sys/tree/master/execabs)

Interesting.

From my assessment, it seems we are fine. If the PATH contains `.` and `newuidmap`/`newgidmap` are present in the current directory, they will be used. Since they are executed from the same context as the user (we `cd $BUNDLE` before doing the lookup, and we're still in the same directory when we call `newuidmap`/`newgidmap`, so there is no issue; it works as intended.

Now, theoretically, if the code does `cd` after calling `ExecLookup` and before calling `newuidmap`/`newgidmap`, a different binary might be called, which is probably still OK, since the binary is still the host binary (i.e. it is not part of the container image downloaded from a random place).

Nevertheless, out of abundance of caution it might make sense to not allow a non-absolute path to `newuidmap`.

kolyshkin:
> Should we use golang.org/x/sys/execabs for these? (https://github.com/golang/sys/tree/master/execabs)

Added. I also took a look at other uses of exec.LookPath and exec.Command, and it seems that these should be left as is.

libcontainer/container_linux.go
kolyshkin: For some reason I was not able to reply to these comments from the main PR page (https://github.com/opencontainers/runc/pull/3354), so I replied "globally".

thaJeztah:
Thanks for looking at that; yes I was thinking along the same lines; in normal situations, `runc` and `newgidmap` would likely be living side-by-side already (which would be valid), but runc _could_ be executed from some non-standard path with a bad binary next to it. Looks like the extra caution doesn't hurt us, so 👍

thaJeztah (approved):
LGTM, nice, thanks!

kolyshkin:

kolyshkin:
@opencontainers/runc-maintainers PTAL 🙏🏻  (I have a few more cleanups on top of this)

hqhq:
Some of the changes might make code less scalable, but we can add them back when we really want to scale someday, it does make code cleaner, 👍

PR #3345, opened 2022-01-22
Fix working with read-only /dev

kolyshkin:
TL;DR: do not chown `/dev/std{in,out,err}` if there's no need; ignore EROFS when chown fails.

For the context, see https://github.com/containers/podman/pull/12954

Please review commit-by-commit.

PR #3355, opened 2022-01-27
[1.1] Fix working with read-only /dev

kolyshkin:
This is a backport of #3345 to 1.1 branch, needed for newer Podman (see https://github.com/containers/podman/pull/12954).

See original PR for details.


kolyshkin:
OK, it seems we have wrong branch protection rules for release-1.1 (that have `test (1.13.x` included, but no `1.17` ones).

Will fix.

kolyshkin:
I have created branch protection rules for release-1.1, and also enabled the following (as an experiment):

> * Require branches to be up to date before merging
> This ensures pull requests targeting a matching branch have been tested with the latest code.

Perhaps it makes sense to enable it for main as well, since I've seen a few times that independent merges result in broken code.

AkihiroSuda (approved):

mrunalp (approved):

PR #3356, opened 2022-01-28
libct/cg/sd: simplify DetectUserDbusSessionBusAddress

kolyshkin:
1. Properly escape the value of dbus address spec.

2. Remove using `systemctl --user --no-pager show-environment` because:
      - is does the same thing we do (get address from `DBUS_SESSION_BUS_ADDRESS` or try to construct it from `XDG_RUNTIME_DIR`);
      - it fails when both those variables are unset:

	$ echo $DBUS_SESSION_BUS_ADDRESS, $XDG_RUNTIME_DIR
	unix:path=/run/user/1000/bus, /run/user/1000
	$ systemctl --user --no-pager show-environment | grep DBUS_SESS
	DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
	$ unset DBUS_SESSION_BUS_ADDRESS
	$ systemctl --user --no-pager show-environment | grep DBUS_SESS
	DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
	$ unset XDG_RUNTIME_DIR
	$ systemctl --user --no-pager show-environment | grep DBUS_SESS
	Failed to connect to bus: $DBUS_SESSION_BUS_ADDRESS and $XDG_RUNTIME_DIR not defined (consider using --machine=<user>@.host --user to connect to bus of other user)

3. Since it does not make sense to suggest `systemctl --user start dbus`
if the proper environment is not set, remove that suggestion from the error
message text. Since `DBUS_SESSION_BUS_ADDRESS` environment variable
is set by dbus-run-session (or dbus-launch, or something similar that is
supposed to be run during the login process), add a suggestion to
re-login.

### DIFF
--- a/libcontainer/cgroups/systemd/user.go
+++ b/libcontainer/cgroups/systemd/user.go
@@ -77,9 +77,8 @@ func DetectUID() (int, error) {
 	return -1, errors.New("could not detect the OwnerUID")
 }
 
-// DetectUserDbusSessionBusAddress returns $DBUS_SESSION_BUS_ADDRESS if set.
-// Otherwise returns "unix:path=$XDG_RUNTIME_DIR/bus" if $XDG_RUNTIME_DIR/bus exists.
-// Otherwise parses the value from `systemctl --user show-environment` .
+// DetectUserDbusSessionBusAddress returns $DBUS_SESSION_BUS_ADDRESS, if set.
+// Otherwise it returns "unix:path=$XDG_RUNTIME_DIR/bus", if $XDG_RUNTIME_DIR/bus exists.
 func DetectUserDbusSessionBusAddress() (string, error) {
 	if env := os.Getenv("DBUS_SESSION_BUS_ADDRESS"); env != "" {
 		return env, nil
@@ -91,16 +90,5 @@ func DetectUserDbusSessionBusAddress() (string, error) {
 			return busAddress, nil
 		}
 	}
-	b, err := exec.Command("systemctl", "--user", "--no-pager", "show-environment").CombinedOutput()
-	if err != nil {
-		return "", fmt.Errorf("could not execute `systemctl --user --no-pager show-environment` (output=%q): %w", string(b), err)
-	}
-	scanner := bufio.NewScanner(bytes.NewReader(b))
-	for scanner.Scan() {
-		s := strings.TrimSpace(scanner.Text())
-		if strings.HasPrefix(s, "DBUS_SESSION_BUS_ADDRESS=") {
-			return strings.TrimPrefix(s, "DBUS_SESSION_BUS_ADDRESS="), nil
-		}
-	}
-	return "", errors.New("could not detect DBUS_SESSION_BUS_ADDRESS from `systemctl --user --no-pager show-environment`. Make sure you have installed the dbus-user-session or dbus-daemon package and then run: `systemctl --user start dbus`")
+	return "", errors.New("could not detect DBUS_SESSION_BUS_ADDRESS from the environment; make sure you have installed the dbus-user-session or dbus-daemon package; note you may need to re-login")
 }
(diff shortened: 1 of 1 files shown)


libcontainer/cgroups/systemd/user.go
```diff
@@ -91,16 +90,5 @@ func DetectUserDbusSessionBusAddress() (string, error) {
 			return busAddress, nil
 		}
```
AkihiroSuda, line 91: When `$DBUS_SESSION_BUS_ADDRESS` is unset but `$XDG_RUNTIME_DIR` is set, probably we can try `systemctl --user --no-pager show-environment`

AkihiroSuda (changes requested):
`systemctl --user show-environment` is for covering `$XDG_RUNTIME_DIR is set but $XDG_RUNTIME_DIR/bus does not exist` case, not for covering `$XDG_RUNTIME_DIR is unset` case

kolyshkin:
> `systemctl --user show-environment` is for covering `$XDG_RUNTIME_DIR is set but $XDG_RUNTIME_DIR/bus does not exist` case, not for covering `$XDG_RUNTIME_DIR is unset` case

Right, I missed that path.

However, current version of systemd does the same thing as we do -- it tries "unix:path=%s/bus", where `%s` is the value of `XDG_RUNTIME_DIR`. 

Here's the logic:
https://github.com/systemd/systemd/blob/be7148ebed5d73c4a76bc6089ebe2e82d8fa33e0/src/libsystemd/sd-bus/sd-bus.c#L1332-L1347

The only caveat is it uses `bus_address_escape` function:
https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/libsystemd/sd-bus/bus-internal.c#L294-L318

Will take a look into it.

kolyshkin:
> The only caveat is it uses bus_address_escape function:
https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/libsystemd/sd-bus/bus-internal.c#L294-L318
>
> Will take a look into it.

It seems that if we are re-creating DBUS_SESSION_BUS_ADDRESS from XDG_RUNTIME_DIR, we need to escape some characters, same as what `bus_address_escape` function in systemd does:
https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/libsystemd/sd-bus/bus-internal.c#L294-L318.

Once that is done, we can remove the fallback which this PR removes.

Changing to draft for now.

kolyshkin:
OK, I have added proper address value escaping, so now our fallback is the same as systemd's (see previous comments).

@AkihiroSuda PTAL

kolyshkin:
> > `systemctl --user show-environment` is for covering `$XDG_RUNTIME_DIR is set but $XDG_RUNTIME_DIR/bus does not exist` case, not for covering `$XDG_RUNTIME_DIR is unset` case
> 
> Right, I missed that path.
> 
> However, current version of systemd does the same thing as we do -- it tries "unix:path=%s/bus", where `%s` is the value of `XDG_RUNTIME_DIR`.

In other words,
 - if `XDG_RUNTIME_DIR` is set, and `$XDG_RUNTIME_DIR/bus` exists, our code will use it and won't fall back to calling `systemctl`;
 - if `XDG_RUNTIME_DIR` is not set, or `$XDG_RUNTIME_DIR/bus` does not exist (ot can't be connected to), `systemctl --user` can not be used at all, as it needs a dbus connection.
 
Therefore,
 - a fallback of using systemctl is useless, as it will fail anyway.
 - suggesting to try `systemctl --user start dbus` in error message is useless, too, as it will fail anyway.

AkihiroSuda (dismissed):

kolyshkin:
Removed the dbus bump commit as it is now obsoleted by #3398.

@AkihiroSuda @thaJeztah PTAL

AkihiroSuda (approved):

kolyshkin:
@thaJeztah @cyphar @mrunalp PTAL

thaJeztah (approved):
LGTM

PR #3358, opened 2022-01-28
Support fetching PSI stats for cgroupv2 containers

dqminh:
This supports fetching PSI stats for cgroupv2 containers. We read the PSI metrics if they are available from:
- cpu.pressure
- memory.pressure
- io.pressure

See more about PSI at https://www.kernel.org/doc/html/latest/accounting/psi.html

### DIFF
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"syscall"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
@@ -112,6 +113,19 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 	// cpu (since kernel 4.15)
 	// Note cpu.stat is available even if the controller is not enabled.
 	if err := statCpu(m.dirPath, st); err != nil && !os.IsNotExist(err) {
+		errs = append(errs, err)
+	}
+	// psi (since kernel 4.20)
+	// open *.pressure file returns
+	// - ErrNotExist when kernel < 4.20 or CONFIG_PSI is disabled
+	// - ENOTSUP when we requires psi=1 in kernel command line to enable PSI support
+	if err := statPSI(m.dirPath, "cpu.pressure", &st.CpuStats.PSI); err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, syscall.ENOTSUP) {
+		errs = append(errs, err)
+	}
+	if err := statPSI(m.dirPath, "memory.pressure", &st.MemoryStats.PSI); err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, syscall.ENOTSUP) {
+		errs = append(errs, err)
+	}
+	if err := statPSI(m.dirPath, "io.pressure", &st.BlkioStats.PSI); err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, syscall.ENOTSUP) {
 		errs = append(errs, err)
 	}
 	// hugetlb (since kernel 5.6)
--- a/libcontainer/cgroups/fs2/psi.go
+++ b/libcontainer/cgroups/fs2/psi.go
@@ -1,0 +1,78 @@
+package fs2
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+)
+
+func statPSI(dirPath string, file string, stats *cgroups.PSIStats) error {
+	f, err := cgroups.OpenFile(dirPath, file, os.O_RDONLY)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	sc := bufio.NewScanner(f)
+	for sc.Scan() {
+		parts := strings.Fields(sc.Text())
+		switch parts[0] {
+		case "some":
+			data, err := parsePSIData(parts[1:])
+			if err != nil {
+				return err
+			}
+			stats.Some = data
+		case "full":
+			data, err := parsePSIData(parts[1:])
+			if err != nil {
+				return err
+			}
+			stats.Full = data
+		}
+	}
+	if err := sc.Err(); err != nil {
+		return &parseError{Path: dirPath, File: file, Err: err}
+	}
+	return nil
+}
+
+func parsePSIData(psi []string) (data cgroups.PSIData, err error) {
+	for _, f := range psi {
+		kv := strings.SplitN(f, "=", 2)
+		if len(kv) != 2 {
+			return data, fmt.Errorf("invalid psi data: %q", f)
+		}
+		switch kv[0] {
+		case "avg10":
+			v, err := strconv.ParseFloat(kv[1], 64)
+			if err != nil {
+				return data, fmt.Errorf("invalid psi value: %q", f)
+			}
+			data.Avg10 = v
+		case "avg60":
+			v, err := strconv.ParseFloat(kv[1], 64)
+			if err != nil {
+				return data, fmt.Errorf("invalid psi value: %q", f)
+			}
+			data.Avg60 = v
+		case "avg300":
+			v, err := strconv.ParseFloat(kv[1], 64)
+			if err != nil {
+				return data, fmt.Errorf("invalid psi value: %q", f)
+			}
+			data.Avg300 = v
+		case "total":
+			v, err := strconv.ParseUint(kv[1], 10, 64)
+			if err != nil {
+				return data, fmt.Errorf("invalid psi value: %q", f)
+			}
+			data.Total = v
+		}
+	}
+	return data, nil
+}
--- a/libcontainer/cgroups/fs2/psi_test.go
+++ b/libcontainer/cgroups/fs2/psi_test.go
@@ -1,0 +1,47 @@
+package fs2
+
+import (
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+)
+
+const examplePSIData = `some avg10=1.71 avg60=2.36 avg300=2.57 total=230548833
+full avg10=1.00 avg60=1.01 avg300=1.00 total=157622356`
+
+func TestStatCPUPsi(t *testing.T) {
+	// We're using a fake cgroupfs.
+	cgroups.TestMode = true
+
+	fakeCgroupDir := t.TempDir()
+	statPath := filepath.Join(fakeCgroupDir, "cpu.pressure")
+
+	if err := os.WriteFile(statPath, []byte(examplePSIData), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	var psi cgroups.PSIStats
+	if err := statPSI(fakeCgroupDir, "cpu.pressure", &psi); err != nil {
+		t.Error(err)
+	}
+
+	if !reflect.DeepEqual(psi, cgroups.PSIStats{
+		Some: cgroups.PSIData{
+			Avg10:  1.71,
+			Avg60:  2.36,
+			Avg300: 2.57,
+			Total:  230548833,
+		},
+		Full: cgroups.PSIData{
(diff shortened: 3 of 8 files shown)


libcontainer/cgroups/stats.go
```diff
@@ -32,9 +32,22 @@ type CpuUsage struct {
 	UsageInUsermode uint64 `json:"usage_in_usermode"`
 }
 
+type PsiData struct {
+	Avg10  float64 `json:"avg10"`
+	Avg60  float64 `json:"avg60"`
+	Avg300 float64 `json:"avg300"`
```
AkihiroSuda, line 38: Maybe: `Avg map[int]float64` (key=10,60,300)

libcontainer/cgroups/fs2/psi_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+)
+
+const examplePsiData = `some avg10=1.71 avg60=2.36 avg300=2.57 total=230548833
+full avg10=1.00 avg60=1.01 avg300=1.00 total=157622356`
+
+func TestStatCPUPsi(t *testing.T) {
```
AkihiroSuda, line 15: Can we have integration tests too?

dqminh: added one for runc events

libcontainer/cgroups/stats.go
dqminh: i don't like that very much because the key avg10,60,300 is fixed, so invidual fields are much better for discovery.

kolyshkin: I was thinking about it, too. If this is fixed in the same way as, say, loadavg is fixed for 1, 5, 15 minutes, we can just use an array with 3 numbers.

OTOH maybe the best way is to have
```
Some, Full map [string]float64
```
where map keys are `avgXXX` and `total`.

In case we want `total` as an integer, maybe it's better to use whatever @AkihiroSuda is proposing, plus a separate Total field.

kolyshkin:
Can we please have `s/Psi/PSI/` since this is an acronym? I know we use things like `Json` before, but in the new code we can do better.

tests/integration/events.bats
```diff
@@ -26,6 +26,34 @@ function teardown() {
 	[[ "${lines[0]}" == *"data"* ]]
 }
 
+@test "events --stats with psi data" {
+	requires root cgroups_v2
+	if [ ! -f /sys/fs/cgroup/cpu.pressure ]; then
+		skip "requires PSI subsystem"
+	fi
```
kolyshkin, line 33: Can you move this to helpers.bash, function `requires`?

libcontainer/cgroups/stats.go
dqminh: the drawbacks of array or map is non-discoverability of keys and extra allocation where it's not necessary. I don't yet see anything that is bad with explicit keys.

We can make the same arguments with a lot of other statistics too such as io.stat. I think explicit struct are much better to use.

tests/integration/events.bats
dqminh: done, this is now `requires psi`.

kolyshkin (approved):
LGTM

tests/integration/helpers.bash
```diff
@@ -448,6 +448,13 @@ function requires() {
 				skip_me=1
 			fi
 			;;
+		psi)
+			# If PSI is not compiled in the kernel, the file will not exist.
+			# If PIS is compiled, but not enable, read will fail with ENOSUPP
+			if [[ ! $(cat /sys/fs/cgroup/cpu.pressure) ]]; then
```
kolyshkin, line 454: You don't need to `cat` the file, mere existence check should be sufficient.

IOW
```bash
if [ ! -f /sys/fs/cgroup/cpu.pressure ]; then
...
```

kolyshkin: Ah, I see, you *do* need to actually cat the file. I should have read better before writing comments 🤦🏻

tests/integration/helpers.bash
```diff
@@ -448,6 +448,13 @@ function requires() {
 				skip_me=1
 			fi
 			;;
+		psi)
+			# If PSI is not compiled in the kernel, the file will not exist.
+			# If PIS is compiled, but not enable, read will fail with ENOSUPP
```
kolyshkin, line 453: typo: `s/PIS/PSI/`.

kolyshkin: nit: period missing at EOL.

kolyshkin: typo: `s/enable/enabled/`.

kolyshkin: typo: `s/ENOSUPP/ENOTSUP/`.

kolyshkin (changes requested):
a few typos in helpers.bash, otherwise LGTM

tests/integration/helpers.bash
dqminh: Thanks, updated.

kolyshkin (approved):
LGTM

dqminh:

dqminh:
ping @opencontainers/runc-maintainers again

kolyshkin:
@dqminh can you please rebase? We have changed something in the tests and they need to be re-run.

dqminh:
@kolyshkin thanks, rebase is done. I had to add `# shellcheck disable=SC2030` for the integration test.

bobbypage:
ping @kolyshkin, this would be great to get in, especially for use downstream in cAdvisor / k8s

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -114,6 +114,16 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 	if err := statCpu(m.dirPath, st); err != nil && !os.IsNotExist(err) {
 		errs = append(errs, err)
 	}
+	// psi ( since kernel 4.20)
```
kolyshkin, line 117: nit: extra space after `(`

libcontainer/cgroups/fs2/psi.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+	if err := sc.Err(); err != nil {
+		return &parseError{Path: dirPath, File: file, Err: err}
+	}
+	return nil
+}
+
+func parsePSIData(psi []string) (data cgroups.PSIData, err error) {
+	for i := 0; i < len(psi); i++ {
```
kolyshkin, line 45: I think it'll be more readable to use
```go
          for _, f := range psi {
```
(and use `f` instead of `psi[i]` below).

libcontainer/cgroups/fs2/psi_test.go
kolyshkin: nit: capitalize `PSI`

libcontainer/cgroups/fs2/psi_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+)
+
+const examplePsiData = `some avg10=1.71 avg60=2.36 avg300=2.57 total=230548833
```
kolyshkin, line 12: nit: capitalize `PSI`

kolyshkin (commented):
A few nits, otherwise looks good.

dqminh:
@kolyshkin updated the nits, also added a fix to check for ENOTSUP when kernel compiled with PSI_DEFAULT_DISABLED, so reading *.pressure file return ENOTSUP as seen in https://cirrus-ci.com/task/4929935776153600

bobbypage:
ping @kolyshkin for another round of a reviews :)

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -114,6 +115,19 @@ func (m *manager) GetStats() (*cgroups.Stats, error) {
 	if err := statCpu(m.dirPath, st); err != nil && !os.IsNotExist(err) {
 		errs = append(errs, err)
 	}
+	// psi (since kernel 4.20)
+	// open *.pressure file returns
+	// - ErrNotExist when kernel < 4.20 or CONFIG_PSI is disabled
+	// - ENOTSUP when we requires psi=1 in kernel command line to enable PSI support
+	if err := statPSI(m.dirPath, "cpu.pressure", &st.CpuStats.PSI); err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, syscall.ENOTSUP) {
```
kolyshkin, line 122: It's probably better to use `unix.ENOTSUP` from `golang.org/x/sys/unix` here (as `syscall` is considered an obsoleted package).

tests/integration/helpers.bash
```diff
@@ -450,6 +450,13 @@ function requires() {
 				skip_me=1
 			fi
 			;;
+		psi)
+			# If PSI is not compiled in the kernel, the file will not exist.
+			# If PSI is compiled, but not enabled, read will fail with ENOTSUPP.
+			if [[ ! $(cat /sys/fs/cgroup/cpu.pressure) ]]; then
```
kolyshkin, line 456: No need to use `[[` here.

```bash
if ! cat /sys/fs/cgroup/cpu.pressure &>/dev/null; then
```

should be sufficient.

bobbypage:
gentle ping to @dqminh regarding the updates mentioned. would be great to get this in so we can consume PSI metrics downstream in cAdvisor / k8s

AkihiroSuda:
@kolyshkin Maybe we can address those issues after merging?

Furisto:
Hey @dqminh this is a great PR. There has been no activity for some time so I am wondering if you are still interested? Otherwise I would be willing to take over and make the requested changes.

Furisto:
@kolyshkin Given that this PR has stalled, how would you like to proceed? I can make the requested changes. Would be great if we could make PSI metrics available.

dqminh:
@thaJeztah i'm really sorry about the delay. We have been running with the patch for so long (!) that i forgot it hasn't been merged.

I left a few comments in the PR on my fork. Generally i think we can try to make the refactoring of existing practices separately from the current change.

szuecs:
@dqminh I read the comments and tried to fix everything and opened https://github.com/dqminh/runc/pull/37 . I have here issue I can't compile it and test wait for fetching containers since 15 minutes.
Dirty, because I did not commit and pushed changes when I tried to `make`.

```
% make
go build -trimpath "-buildmode=pie"  -tags "seccomp" -ldflags "-X main.gitCommit=v1.1.0-181-g6414629a-dirty -X main.version=1.1.0+dev " -o runc .
# pkg-config --cflags  -- libseccomp
Package libseccomp was not found in the pkg-config search path.
Perhaps you should add the directory containing `libseccomp.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libseccomp', required by 'virtual:world', not found
pkg-config: exit status 1
make: *** [Makefile:33: runc] Error 2
zsh: exit 2     make

..

[~/go/src/github.com/opencontainers/runc]% make test 
docker build  -t runc_dev:psi .
Sending build context to Docker daemon  33.68MB
Step 1/17 : ARG GO_VERSION=1.18
Step 2/17 : ARG BATS_VERSION=v1.3.0
Step 3/17 : ARG LIBSECCOMP_VERSION=2.5.3
Step 4/17 : FROM golang:${GO_VERSION}-bullseye
1.18-bullseye: Pulling from library/golang
17c9e6141fdb: Downloading [=======================>                           ]  26.28MB/55.05MB
de4a4c6caea8: Download complete
4edced8587e6: Download complete
a7969cffbf46: Downloading [=====================>                             ]  23.61MB/54.59MB
efa7548a174d: Downloading [=========>                                         ]  16.11MB/85.97MB
ca7db62c462d: Waiting
623d7734067d: Waiting

```
I hope this helps to get it merged. :)

h-vetinari:
Stray comment: As of linux 6.1 there's an [interface](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bd9a3dba185ce6701b41f0341470d3f53bbbbaed) called `cgroup.pressure` to switch PSI accounting on/off per cgroup. I suppose the code here will have to check if accounting is turned off.

szuecs:
FYI: I created https://github.com/opencontainers/runc/pull/3679 as follow up

PR #3360, opened 2022-01-29
libcontainer: remove "pausing" state

### LINKED ISSUE #3343: remove Pausing state, update docs
This needs to be done after merging #3316.

The "Pausing" state of container is not used, and the Pause() docs incorrectly says that it works asynchronously.

Need to
 - remove Pausing state entirely;
 - update the docs (make them conform to golang style)

Ref: https://github.com/opencontainers/runc/pull/3316#discussion_r787643478

danishprakash:
Upon running tests via `make test`, I encountered flaky failures, I'll look into those and see if they have something to do with my host because I noticed the same behavior on `main` too.

Addresses #3343

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -100,7 +100,7 @@ type Container interface {
 	// Restore restores the checkpointed container to a running state using the criu(8) utility.
 	Restore(process *Process, criuOpts *CriuOpts) error
 
-	// If the Container state is RUNNING or CREATED, sets the Container state to PAUSING and pauses
+	// If the Container state is RUNNING or CREATED, sets the Container state to PAUSED and pauses
 	// the execution of any user processes. Asynchronously, when the container finished being paused the
 	// state is changed to PAUSED.
 	// If the Container state is PAUSED, do nothing.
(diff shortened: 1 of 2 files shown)


libcontainer/container_linux.go
```diff
@@ -100,7 +100,7 @@ type Container interface {
 	// Restore restores the checkpointed container to a running state using the criu(8) utility.
 	Restore(process *Process, criuOpts *CriuOpts) error
 
-	// If the Container state is RUNNING or CREATED, sets the Container state to PAUSING and pauses
+	// If the Container state is RUNNING or CREATED, sets the Container state to PAUSED and pauses
```
danishprakash, line 103: Although the actual state that `Pause` internally sets is `FROZEN`, I hope this should be okay as an externally exposed func doc? cc/ @kolyshkin

kolyshkin: Yes, it's "frozen" in cgroup lingo, and "paused" in libcontainer lingo.

kolyshkin (approved):
LGTM

kolyshkin:
> Closes #3343

That issue also says to convert docs to golang style, so we can't close it just yet.

Surely this needs to be a separate PR.

So, replace `Closes:` with `Addresses:` or something.

danishprakash:
Yes, that can be a separate PR of its own. Updated.

kolyshkin:
@thaJeztah PTAL

thaJeztah (approved):
LGTM, thanks for contributing!

PR #3363, opened 2022-01-31
signals: fix signal name debug print

kolyshkin:
Here's how it looks now:

```console
$ runc --debug exec ctid sleep 1h
...
DEBU[0000]signals.go:102 main.(*signalHandler).forward() sending signal to process urgent I/O condition
DEBU[0000]signals.go:102 main.(*signalHandler).forward() sending signal to process urgent I/O condition
DEBU[0022]signals.go:102 main.(*signalHandler).forward() sending signal to process terminated
DEBU[0022]signals.go:102 main.(*signalHandler).forward() sending signal to process urgent I/O condition
```
This is obviously not very readable.

Use unix.SignalName, and add PID while we're at it.

With this commit:

```console
DEBU[0000]signals.go:103 main.(*signalHandler).forward() forwarding signal 23 (SIGURG) to 891345
DEBU[0020]signals.go:103 main.(*signalHandler).forward() forwarding signal 45 () to 891345
DEBU[0020]signals.go:103 main.(*signalHandler).forward() forwarding signal 23 (SIGURG) to 891345
DEBU[0020]signals.go:103 main.(*signalHandler).forward() forwarding signal 23 (SIGURG) to 891345
```

### DIFF
--- a/signals.go
+++ b/signals.go
@@ -99,8 +99,9 @@ func (h *signalHandler) forward(process *libcontainer.Process, tty *tty, detach 
 				}
 			}
 		default:
-			logrus.Debugf("sending signal to process %s", s)
-			if err := unix.Kill(pid1, s.(unix.Signal)); err != nil {
+			us := s.(unix.Signal)
+			logrus.Debugf("forwarding signal %d (%s) to %d", int(us), unix.SignalName(us), pid1)
+			if err := unix.Kill(pid1, us); err != nil {
 				logrus.Error(err)
 			}
 		}


signals.go
```diff
@@ -99,7 +99,7 @@ func (h *signalHandler) forward(process *libcontainer.Process, tty *tty, detach
 				}
 			}
 		default:
-			logrus.Debugf("sending signal to process %s", s)
+			logrus.Debugf("forwarding %s to %d", unix.SignalName(s.(unix.Signal)), pid1)
```
thaJeztah, line 102: Looks like `unix.SignalName()` can potentially return an empty string if it doesn't know the signal; https://github.com/golang/sys/blob/15123e1e1f71fb044f0ebc6b23b37d9744890bf9/unix/syscall_unix.go#L67-L75

(It's of course only a debug log), but that could mean that (e.g.) one of the `SIGRTMIN` signals won't be printed. I guess we could add a wrapper and in that case just print the numeric value?

kolyshkin: You're quite right, `unix` does not know about signals > 31, and will return an empty string.

Since this is a debug print, I don't want to write anything more complicated than the one liner. So, I have modified it to print both signal number and the name, if known.

Looks like this:
```
    DEBU[0000]signals.go:103 main.(*signalHandler).forward() forwarding signal 23 (SIGURG) to 891345
    DEBU[0020]signals.go:103 main.(*signalHandler).forward() forwarding signal 45 () to 891345
    DEBU[0020]signals.go:103 main.(*signalHandler).forward() forwarding signal 23 (SIGURG) to 891345
    DEBU[0020]signals.go:103 main.(*signalHandler).forward() forwarding signal 23 (SIGURG) to 891345
```

thaJeztah: Yes, that looks like a good middle-ground 👍  (I agree to not over-complicate it, as it's just for debugging currently)

thaJeztah (approved):
LGTM

signals.go
```diff
@@ -99,8 +99,9 @@ func (h *signalHandler) forward(process *libcontainer.Process, tty *tty, detach
 				}
 			}
 		default:
-			logrus.Debugf("sending signal to process %s", s)
-			if err := unix.Kill(pid1, s.(unix.Signal)); err != nil {
+			us := s.(unix.Signal)
+			logrus.Debugf("forwarding signal %d (%s) to %d", int(us), unix.SignalName(us), pid1)
```
thaJeztah, line 103: One more thought for consideration (I wanted to suggest so earlier, but now also see similar code is used above in this function).

We could add `pid`, `signal`, and `signal_name` (e.g.) as fields; something like;

```go
logrus.WithFields(logrus.Fields{
    "pid":    pid1,
    "signal": us,
    "signal_name": unix.SignalName(us),
}).Debug("forwarding signal")
```

mrunalp (approved):

signals.go
kolyshkin: I thought about that, too, but at this time I'd rather not add more structured logging, as I guess at some point we'll need to replace logrus...

thaJeztah: Yeah, but it's still there, and it would be producing somewhat more consistent logs.

Anyway, not a blocker

PR #3365, opened 2022-02-01
libct/specconv: checkPropertyName speedup

kolyshkin:
See commits for details.

Mostly done because the comment was misleading, not in line with what the code did.

Fix the code, so now the comment is fine.

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -538,8 +538,10 @@ func checkPropertyName(s string) error {
 	if len(s) < 3 {
 		return errors.New("too short")
 	}
-	// Check ASCII characters rather than Unicode runes.
-	for _, ch := range s {
+	// Check ASCII characters rather than Unicode runes,
+	// so we have to use indexes rather than range.
+	for i := 0; i < len(s); i++ {
+		ch := s[i]
 		if (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') {
 			continue
 		}
--- a/libcontainer/specconv/spec_linux_test.go
+++ b/libcontainer/specconv/spec_linux_test.go
@@ -762,7 +762,7 @@ func TestInitSystemdProps(t *testing.T) {
 	}
 }
 
-func TestIsValidName(t *testing.T) {
+func TestCheckPropertyName(t *testing.T) {
 	testCases := []struct {
 		in    string
 		valid bool
@@ -787,7 +787,7 @@ func TestIsValidName(t *testing.T) {
 	}
 }
 
-func BenchmarkIsValidName(b *testing.B) {
+func BenchmarkCheckPropertyName(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		for _, s := range []string{"", "xx", "xxx", "someValidName", "A name", "Кир", "მადლობა", "合い言葉"} {
 			_ = checkPropertyName(s)


libcontainer/specconv/spec_linux.go
```diff
@@ -538,8 +538,10 @@ func checkPropertyName(s string) error {
 	if len(s) < 3 {
 		return errors.New("too short")
 	}
-	// Check ASCII characters rather than Unicode runes.
-	for _, ch := range s {
+	// Check ASCII characters rather than Unicode runes,
+	// so we have to use indexes rather than range.
+	for i := 0; i < len(s); i++ {
+		ch := s[i]
 		if (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') {
```
thaJeztah, line 545: Could also use the `unicode` utilities for this - seems to be even (but marginally) faster 🤔 (but worth verifying on your machine if you see that as well)

```
BenchmarkCheckPropertyName
BenchmarkCheckPropertyName-8          	 1904448	       617.9 ns/op	     320 B/op	      12 allocs/op
BenchmarkCheckPropertyNameUnicode
BenchmarkCheckPropertyNameUnicode-8   	 1977055	       594.7 ns/op	     320 B/op	      12 allocs/op
```

```go
func checkPropertyName(s string) error {
	if len(s) < 3 {
		return fmt.Errorf("too short")
	}
	// Check ASCII characters rather than Unicode runes,
	// so we have to use indexes rather than range.
	for i := 0; i < len(s); i++ {
		ch := s[i]
		if ch > unicode.MaxASCII || !unicode.IsLetter(rune(ch)) {
			return fmt.Errorf("contains non-alphabetic character")
		}
	}
	return nil
}
```

thaJeztah (approved):
LGTM

libcontainer/specconv/spec_linux.go
thaJeztah: Actually, not sure if that makes sense (covering to a `rune()`); guess that would've been more accurate if we kept the `range` loop

kolyshkin: Not sure how you managed to get faster results, as obviously two calls and a conversion is more expensive than 4 byte checks. This is probably the effect of dynamic power management / CPU frequency scaling. I work around that by doing a few rounds, and checking that numbers are consistent (benchstat checks that, too), and I usually skip the first measurement (of course, the more correct approach would be to disable CPU freq scaling).

Here's what I see on my machine (tested each implementation 3 times, first round discarded, `benchstat` output):
```
name                 old time/op    new time/op    delta
CheckPropertyName-4     122ns ± 0%     152ns ± 1%   ~     (p=0.100 n=3+3)

name                 old alloc/op   new alloc/op   delta
CheckPropertyName-4     64.0B ± 0%     96.0B ± 0%   ~     (p=0.100 n=3+3)

name                 old allocs/op  new allocs/op  delta
CheckPropertyName-4      4.00 ± 0%      6.00 ± 0%   ~     (p=0.100 n=3+3)
```

Here's the diff:

```diff
diff --git a/libcontainer/specconv/spec_linux.go b/libcontainer/specconv/spec_linux.go
index fad7b802..151174e1 100644
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -11,6 +11,7 @@ import (
        "strings"
        "sync"
        "time"
+       "unicode"
 
        systemdDbus "github.com/coreos/go-systemd/v22/dbus"
        dbus "github.com/godbus/dbus/v5"
@@ -542,10 +543,9 @@ func checkPropertyName(s string) error {
        // so we have to use indexes rather than range.
        for i := 0; i < len(s); i++ {
                ch := s[i]
-               if (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') {
-                       continue
+               if ch > unicode.MaxASCII || !unicode.IsLetter(rune(ch)) {
+                       return errors.New("contains non-alphabetic character")
                }
-               return errors.New("contains non-alphabetic character")
        }
        return nil
 }
```

In any case, the difference is negligible -- and as I said, this PR is mostly to fix the comment (by making the code in line with what the comment says).

> Actually, not sure if that makes sense (covering to a `rune()`); guess that would've been more accurate if we kept the `range` loop

I'd do the same if I would use unicode functions, but it does make sense in the first place.

thaJeztah: > This is probably the effect of dynamic power management / CPU frequency scaling.

Yes, perhaps; I ran it a couple of times, and was somewhat scratching my head.

kolyshkin: I played a bit with the if statement (there's a way to optimize the checks) but it did not show any difference (probably because the compiled optimizes this anyway).

Here is the (unsuccessful) optimization attempt:

```diff
diff --git a/libcontainer/specconv/spec_linux.go b/libcontainer/specconv/spec_linux.go
index fad7b802..fbe1fc19 100644
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -541,11 +541,10 @@ func checkPropertyName(s string) error {
        // Check ASCII characters rather than Unicode runes,
        // so we have to use indexes rather than range.
        for i := 0; i < len(s); i++ {
-               ch := s[i]
-               if (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') {
-                       continue
+               c := s[i]
+               if c < 'A' || c > 'z' || c > 'Z' && c < 'a' {
+                       return errors.New("contains non-alphabetic character")
                }
-               return errors.New("contains non-alphabetic character")
        }
        return nil
 }
```

thaJeztah: LOL; I was playing with that line as well, but (unfortunately) `A-Z` and `a-z` are distinct ranges (numbers are in between I think?), so I don't think there's a very clean way to do it.

I went looking if there's some smart code in stdlib (which is how I came across the unicode thing), and found [some old code (too complicated)](https://pkg.go.dev/golang.org/x/exp/utf8string#String.IsASCII), some [personal ASCII package](https://pkg.go.dev/github.com/galsondor/go-ascii) (looked neat, but too much _just for this_), and some code in the `strings` package, that didn't do it any different https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/strings/strings.go;l=550-559

kolyshkin: Well, there's a very smart code in strings package, which uses some bit arithmetics, and that _might_ be faster, but again I don't thing this code is worth optimizing more
https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/strings/strings.go;l=794-818;drc=refs%2Ftags%2Fgo1.17.6

kolyshkin:
@opencontainers/runc-maintainers PTAL (this is very easy to review)

AkihiroSuda (approved):

PR #3367, opened 2022-02-07
tests: add `set -u`

kolyshkin:
Fix all existing cases where `set -u` can be problematic, and add it globally for all tests.

This is a way to prevent the code doing something really bad when a
variable it uses is not set. Good to have since it helps to catch some
logical errors etc.

This also fixes a bug of not running CRIU tests; see https://github.com/opencontainers/runc/pull/3367#discussion_r823305088 for details.

For brevity, this PR concentrates on enabling `set -u` only. Various cleanups on top of it are in #3369.

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -252,7 +252,7 @@ function simple_cr() {
 	# in time when the last page is lazily transferred to the destination.
 	# Killing the CRIU on the checkpoint side will let the container
 	# continue to run if the migration failed at some point.
-	[ -n "$RUNC_USE_SYSTEMD" ] && set_cgroups_path
+	[ -v RUNC_USE_SYSTEMD ] && set_cgroups_path
 	runc_restore_with_pipes ./image-dir test_busybox_restore --lazy-pages
 
 	wait $cpt_pid
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -54,7 +54,8 @@ function runc() {
 
 # Raw wrapper for runc.
 function __runc() {
-	"$RUNC" ${RUNC_USE_SYSTEMD+--systemd-cgroup} --root "$ROOT/state" "$@"
+	"$RUNC" ${RUNC_USE_SYSTEMD+--systemd-cgroup} \
+		${ROOT:+--root "$ROOT/state"} "$@"
 }
 
 # Wrapper for runc spec.
--- a/tests/integration/helpers.bash
+++ b/tests/integration/helpers.bash
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-# bats-core v1.2.1 defines BATS_RUN_TMPDIR
-if [ -z "$BATS_RUN_TMPDIR" ]; then
+# bats-core v1.2.1 defines BATS_RUN_TMPDIR.
+if [ ! -v BATS_RUN_TMPDIR ]; then
 	echo "bats >= v1.2.1 is required. Aborting." >&2
 	exit 1
 fi
@@ -18,6 +18,10 @@ unset IMAGES
 RECVTTY="${INTEGRATION_ROOT}/../../contrib/cmd/recvtty/recvtty"
 SD_HELPER="${INTEGRATION_ROOT}/../../contrib/cmd/sd-helper/sd-helper"
 SECCOMP_AGENT="${INTEGRATION_ROOT}/../../contrib/cmd/seccompagent/seccompagent"
+
+# Some variables may not always be set. Set those to empty value,
+# if unset, to avoid "unbound variable" error.
+: "${ROOTLESS_FEATURES:=}"
 
 # Test data path.
 # shellcheck disable=SC2034
@@ -66,7 +70,7 @@ function runc_spec() {
 	runc spec $rootless
 
 	# Always add additional mappings if we have idmaps.
-	if [[ "$ROOTLESS" -ne 0 ]] && [[ "$ROOTLESS_FEATURES" == *"idmap"* ]]; then
+	if [[ "$ROOTLESS" -ne 0 && "$ROOTLESS_FEATURES" == *"idmap"* ]]; then
 		runc_rootless_idmap
 	fi
 }
@@ -88,7 +92,7 @@ function runc_rootless_idmap() {
 
 # Returns systemd version as a number (-1 if systemd is not enabled/supported).
 function systemd_version() {
-	if [ -n "${RUNC_USE_SYSTEMD}" ]; then
+	if [ -v RUNC_USE_SYSTEMD ]; then
 		systemctl --version | awk '/^systemd / {print $2; exit}'
 		return
 	fi
@@ -98,7 +102,7 @@ function systemd_version() {
 
 function init_cgroup_paths() {
 	# init once
-	test -n "$CGROUP_UNIFIED" && return
+	[ -v CGROUP_UNIFIED ] && return
 
 	if stat -f -c %t /sys/fs/cgroup | grep -qFw 63677270; then
 		CGROUP_UNIFIED=yes
@@ -106,7 +110,7 @@ function init_cgroup_paths() {
 		# For rootless + systemd case, controllers delegation is required,
 		# so check the controllers that the current user has, not the top one.
 		# NOTE: delegation of cpuset requires systemd >= 244 (Fedora >= 32, Ubuntu >= 20.04).
-		if [[ "$ROOTLESS" -ne 0 && -n "$RUNC_USE_SYSTEMD" ]]; then
+		if [[ "$ROOTLESS" -ne 0 && -v RUNC_USE_SYSTEMD ]]; then
 			controllers="/sys/fs/cgroup/user.slice/user-$(id -u).slice/user@$(id -u).service/cgroup.controllers"
 		fi
 
@@ -141,11 +145,11 @@ function init_cgroup_paths() {
 }
 
 function create_parent() {
-	if [ -n "$RUNC_USE_SYSTEMD" ]; then
-		[ -z "$SD_PARENT_NAME" ] && return
+	if [ -v RUNC_USE_SYSTEMD ]; then
+		[ ! -v SD_PARENT_NAME ] && return
 		"$SD_HELPER" --parent machine.slice start "$SD_PARENT_NAME"
 	else
-		[ -z "$REL_PARENT_PATH" ] && return
+		[ ! -v REL_PARENT_PATH ] && return
 		if [ "$CGROUP_UNIFIED" == "yes" ]; then
 			mkdir "/sys/fs/cgroup$REL_PARENT_PATH"
 		else
@@ -161,11 +165,11 @@ function create_parent() {
 }
 
 function remove_parent() {
-	if [ -n "$RUNC_USE_SYSTEMD" ]; then
-		[ -z "$SD_PARENT_NAME" ] && return
+	if [ -v RUNC_USE_SYSTEMD ]; then
+		[ ! -v SD_PARENT_NAME ] && return
 		"$SD_HELPER" --parent machine.slice stop "$SD_PARENT_NAME"
 	else
-		[ -z "$REL_PARENT_PATH" ] && return
+		[ ! -v REL_PARENT_PATH ] && return
 		if [ "$CGROUP_UNIFIED" == "yes" ]; then
 			rmdir "/sys/fs/cgroup/$REL_PARENT_PATH"
 		else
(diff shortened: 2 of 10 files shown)


kolyshkin:
Aaand it's finally working as it should.

tests/integration/checkpoint.bats
```diff
@@ -252,7 +252,7 @@ function simple_cr() {
 	# in time when the last page is lazily transferred to the destination.
 	# Killing the CRIU on the checkpoint side will let the container
 	# continue to run if the migration failed at some point.
-	[ -n "$RUNC_USE_SYSTEMD" ] && set_cgroups_path
+	[ -v RUNC_USE_SYSTEMD ] && set_cgroups_path
```
thaJeztah, line 255: I'm not familiar with the `-v` option (tried searching the man page, but couldn't find it 😅); what does it do? Is it something Bash specific? (in which case perhaps this one should explicitly use `[[`) 

If it's to differentiate between "var not set" and "var set, but empty", I'm wondering if it's more error-prone 🤔 (maybe not, not sure, would have to think of various scenarios)

tests/integration/helpers.bash
```diff
@@ -59,12 +59,10 @@ function __runc() {
 
 # Wrapper for runc spec.
 function runc_spec() {
-	local args=()
-	if [ "$ROOTLESS" -ne 0 ]; then
-		args+=("--rootless")
-	fi
+	local rootless=""
+	[ "$ROOTLESS" -ne 0 ] && rootless="--rootless"
```
thaJeztah, line 70: I got confused here for a bit, and thought that `$ROOTLESS` was a boolean(isn) variable, but it looks to be the `uid` (from a couple of lines up)

```bash
# Check if we're in rootless mode.
ROOTLESS=$(id -u)
```

Do we need the actual `id` in the tests? If not, I'm wondering if we should change that line to set `$ROOTLESS` to a boolean(ish) value (empty or "1"); that would make the comment on it more matching reality, and perhaps make it easier (less error-prone) to use overall

tests/integration/checkpoint.bats
kolyshkin: help [
help [[

It tells if the variable is set.

It is bash specific, but we use bats, which is already bash-specific, and have /bin/bash shebang everywhere.

[ and [[ have slightly different rules and both support -v in bash.

The alternative, also bash-specific, is something like `[ -n "${VAR:-x}" ]`, which looks way more cryptic than -v.

tests/integration/helpers.bash
kolyshkin: I have a follow-up commit which removes ROOTLESS entirely and uses UID instead. The mere aim of this PR is to enable `-u` (which turned out to be quite challenging).

tests/integration/checkpoint.bats
thaJeztah: Interesting; neither `help [` / `help [[` nor `man [` / `man test` (`man [[` returns an error on Ubuntu) showed it (at least, couldn't find it).

thaJeztah: So my concern is a bit that `-v ENABLE_FOO` only checks if it's set, but can be confusing if (e.g.) `ENABLE_FOO=0`

```bash
(set -eu; [ -v ENABLE_FOO ] && echo "FOO is enabled"; echo $?)
1

(set -eu; ENABLE_FOO=0; [ -v ENABLE_FOO ] && echo "FOO is enabled"; echo $?)
FOO is enabled
0

(set -eu; ENABLE_FOO=1; [ -v ENABLE_FOO ] && echo "FOO is enabled"; echo $?)
FOO is enabled
0
```

From that perspective, `[ "${ENABLE_FOO}" = "1" ]` (e.g.) feels more transparent (and possibly less error prone?)
But if we really need `-u` everywhere, that would probably need some initialization code to be sure variables
are set to a default, like

```bash
: "${ENABLE_FOO:=0}"
```

(which we do for `$ROOTLESS_FEATURES` I see)

kolyshkin: Yes, what you propose is an alternative to using `-v`, and is sometimes used.

I guess what we need to do is document all those variables. For most of them, any value means its "on/true/1/yes/enabled", and the unset variable means "off/false/0/no/disabled".

The only exception is `CGROUP_UNIFIED`, which is either `yes` or `no` (or unset). When I think about it, it makes sense to change this to `CGROUP_V1` and `CGROUP_V2`, so we treat these the same as all the others (unset means no, set means yes).

kolyshkin: Since I am already doing some changes with `CGROUP_UNIFIED` in #3369, let's replace it with `CGROUP_V2` and `CGROUP_V1` in there, and keep this PR focused on one single thing -- enabling `set -u`.

kolyshkin: https://github.com/opencontainers/runc/pull/3369/commits/14445d5edf05fef2aeeb6bdeb649afdaf6629bf9

tests/integration/helpers.bash
kolyshkin: https://github.com/opencontainers/runc/pull/3369/commits/dba23d7b24cc21bbcec9d83c96f472f83c1facc7

kolyshkin:
@thaJeztah I think I have addressed all of your review comments; please take another look at this 🙏🏻

tests/integration/checkpoint.bats
kolyshkin: IOW this is addressed in #3369

tests/integration/helpers.bash
kolyshkin: IOW this is addressed in #3369

kolyshkin:
Now, this PR is not about "making our shell scripts nicer". It is a way to catch *very bad* errors.

For example, consider the following use case, where `setup_test` sets `TESTDIR`, and then its value is used:
```bash
setup_test
VARDIR="$TESTDIR/var"
mkdir -p "$VARDIR"
....
rm -rf "$VARDIR"
```

All is fine and dandy. Now, someone replicates this code, but makes one small mistake:

```bash
VARDIR="$TESTDIR/var"
setup_test
mkdir -p $"VARDIR"
....
rm -rf "$VARDIR
```

This error is not getting caught by bash itself or e.g. `shellcheck`, and as a result, host `/var` will be removed as a result of running this "test". This is a somewhat extreme but quite realistic example; in fact, this PR was motivated by a very similar issue found in cri-o.

PTAL @AkihiroSuda @thaJeztah @cyphar

kolyshkin:
@tianon I'd appreciate your review on that, if you can spare some time.

thaJeztah (approved):
SGTM to fix the issue, but always open to suggestions from @tianon :)

tests/integration/checkpoint.bats
tianon: I agree that `-v VAR` is very subtly different from `-n $VAR` -- when I make these conversions, I usually just use `:-` (or `-`) such that it doesn't trip `set -u` (unless it's used multiple times, then I use the `: "${VAR:=}"` as suggested by Sebastiaan):

```suggestion
	[ -n "${RUNC_USE_SYSTEMD:-}" ] && set_cgroups_path
```

It looks more cryptic, but it matches the old behavior (where both empty _and_ unset are considered "false" -- in many cases/places, it's trivial to set a variable to false and impossible to unset it, like Docker's `--env` for example).

tests/integration/helpers.bash
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
-# bats-core v1.2.1 defines BATS_RUN_TMPDIR
-if [ -z "$BATS_RUN_TMPDIR" ]; then
+set -u
```
tianon, line 3: One problem at a time and all that, but this probably ought to consider getting at least `-e` as well (if not `-o pipefail` too). :see_no_evil:

(Also a much less trivial change than it appears on the face, just like enabling `set -u` :sweat_smile:)

tests/integration/root.bats
```diff
@@ -11,7 +11,7 @@ function setup() {
 function teardown() {
 	if [ -n "$ALT_ROOT" ]; then
 		ROOT=$ALT_ROOT __runc delete -f test_dotbox
-		rm -rf "$ALT_ROOT"
+		unset ALT_ROOT
```
tianon, line 14: Was this `rm -rf` not doing something meaningful/useful?

tests/integration/root.bats
```diff
@@ -11,7 +11,7 @@ function setup() {
 function teardown() {
 	if [ -n "$ALT_ROOT" ]; then
```
tianon, line 12: Is it possible for `ALT_ROOT` to be unset before this function is called? :eyes:  (ie, if this function got called twice somehow?  from the name it sounds like that's terribly unlikely, but maybe worth thinking about :see_no_evil:)

tianon (commented):
I was scrolling through the comments here and had _almost_ written off reviewing this when I saw your explicit call-out -- I'm honored to review. :sweat_smile: :heart: :bow:

I agree with @thaJeztah re: `-v` vs `-n`/`-z`, but I've put a bit more detail in the appropriate conversation below.

(I think the next "goal" I'd suggest is consistent use of `set -e` and even `set -o pipefail`, but as noted below, that's probably an even bigger bridge :innocent:)

tests/integration/checkpoint.bats
kolyshkin: @tianon thanks for your input! 

I am very open to use '${VAR:-}` where appropriate, but to me it seems that `-v VAR` is a tad cleaner, because it is more readable, and readability is a big factor in my book.

Note that I only switch to `-v` in cases where the flag is essentially a binary 0/1, since set/unset fits it better than
- empty/non-empty (because we do not have to provide a kludge to satisfy `set -u`),
- "yes"/"no", 1/0 etc. (because string or numeric values are not standardized anywhere in the code, and we don't have to set the "no"/0 case).

I can certainly rework the whole set again to use `-n|-z "${VAR:-}"` instead of `[!] -v VAR`, if I clearly see its benefits.

You point out to two things
1. Matching the old behavior. I agree, but it is not an issue here since this is an isolated codebase which we fully control, so we can convert it all in one sitting (which is what this PR does) and won't add any new bugs (well, hopefully).
2. Inability to unset a variable when using Docker. Also true, but I don't see this is practically an issue (for the same reason -- this is an isolated set of tests which we run in some specific environments we fully control).

My third reason is sheer laziness -- it took a while to convert to `-v`, and I have another PR which is based on this one (brining in some more cleanups). I can't convince myself to spend a few hours to rework this (and that PR) one more time.

All it all, I'd like to leave this as is, unless there's a strong objection. I am listening and can definitely reconsider.

tests/integration/root.bats
kolyshkin: It used to be the case earlier that tests called teardown before setup. This must be a leftover from the era.

You're right that this does not work well with `set -u`. Let me remove it.

tests/integration/root.bats
kolyshkin: If you look at `setup()`, `$ALT_ROOT` is a subdirectory of `$ROOT`, which is removed by `teardown_bundle`, so this extra `rm -rf` is not really needed.

I'll rework this part.

tests/integration/checkpoint.bats
tianon: Totally fair 👍

tests/integration/root.bats
kolyshkin: Reworked ALT_ROOT handling in `teardown` is a separate commit; PTAL.

tests/integration/root.bats
kolyshkin: Reworked ALT_ROOT handling in `teardown` is a separate commit; PTAL.

tests/integration/helpers.bash
kolyshkin: In fact, `bats` sets `test -e` for test cases (i.e. anything that is called from `@test xxx {...}`), so most of the code is `set -e`-friendly already, it's only the code that is called from `setup/cleanup` that is not.

I do have that in mind, but thanks for reminding. Heck, let's open an issue about it. Here we go: https://github.com/opencontainers/runc/issues/3399 (oh, and feel free to tackle it, we need more eyes and hands in this repo)

kolyshkin:
@tianon can you please take another look? I think I have addressed all of your review comments.

tianon:
Mea culpa :bow:

It seems fine to me! :+1:

tests/integration/helpers.bash
```diff
@@ -349,7 +353,7 @@ function requires() {
 		local skip_me
 		case $var in
 		criu)
-			if [ -n "$HAVE_CRIU" ]; then
+			if [ ! -v HAVE_CRIU ]; then
```
kolyshkin, line 358: So, this fixes the bug of not running criu tests; the old check for `-n "$HAVE_CRIU`, introduced by commit 6e1d476aa, is wrong, it should have been `-z "$HAVE_CRIU"`. This is also fixed by #3380.

kolyshkin:
@thaJeztah @AkihiroSuda PTAL

kolyshkin:
@thaJeztah @AkihiroSuda PTAL (this is now blessed by @tianon, although I guess he forgot the formal "approve")

AkihiroSuda (approved):

mrunalp (approved):

PR #3370, opened 2022-02-09
ci: bump golangci-lint to v1.44, golangci-lint-action to v3

kolyshkin:
and format sources using gofumpt 0.2.1 (which is uncluded to that golangci-lint version).

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -7,6 +7,8 @@ on:
       - master
       - release-*
   pull_request:
+env:
+  GO_VERSION: 1.17.x
 
 jobs:
 
@@ -14,11 +16,14 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "${{ env.GO_VERSION }}"
       - name: install deps
         run: |
           sudo apt -q update
           sudo apt -q install libseccomp-dev
-      - uses: golangci/golangci-lint-action@v2
+      - uses: golangci/golangci-lint-action@v3
         with:
           version: v1.44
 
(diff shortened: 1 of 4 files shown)


thaJeztah (dismissed):
LGTM

AkihiroSuda:
gofumpt seems too much nitpicky.
Can we just disable it?

kolyshkin:
> gofumpt seems too much nitpicky.

I bet you haven't seen `wsl` linter yet 🤣

kolyshkin:
So, due to my bad memory I ended up reimplementing this today, to find out I did that once already.

I have cleaned up the last commit a bit, and added one more to prepare for https://github.com/golangci/golangci-lint-action/pull/403 (so that https://github.com/opencontainers/runc/pull/3397 will succeed, once rebased on top of this one).

Can we please have this merged? @opencontainers/runc-maintainers @AkihiroSuda PTAL

AkihiroSuda (approved):

.github/workflows/validate.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.x # Latest stable
```
thaJeztah, line 19: Perhaps cleaner to combine this commit with the actual PR that updates to v3, that way the change can be reviewed together with that 

Also wondering if `1.x` should better be tied to actual versions we're testing against; I've had cases where Go updates caused the linter to fail (or golangci-lint not yet being compatible with the latest version on "day 1"), and it's nice to have changes related to that together with the actual go update / addition in CI)

thaJeztah: Note that it's already possible (and a good idea) to specify the Go version to use for golangci-lint, as it's currently dependent on the version you install; see https://github.com/containerd/containerd/commit/534685f9558cc67125156ba2fbdd0e0ade74cc33, which did so for containerd

kolyshkin: So, basically, `golangci/golangci-lint-action@v3` makes `skip-go-installation: true` the only way.

I am using `go-version 1.x` to avoid updating go every time the new release is becoming "latest". Quoting the commit message:

> In our case, we need at least go 1.16, and can use 1.17 or 1.18 (when
> it is released), but we don't set it explicitly here to ease the
> maintenance burden (1.x should point to the latest stable).

You're right that we're risking something to be broken here when the go version is changed.

Updated to follow your recommendations; PTAL @thaJeztah

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

PR #3371, opened 2022-02-11
ci: shellcheck: update to 0.8.0, fix/suppress new warnings

kolyshkin:
1. This valid warning is reported by shellcheck v0.8.0:
```
	In tests/integration/helpers.bash line 38:
	KERNEL_MINOR="${KERNEL_VERSION#$KERNEL_MAJOR.}"
				       ^-----------^ SC2295 (info): Expansions inside ${..} need to be quoted separately, otherwise they match as patterns.

	Did you mean:
	KERNEL_MINOR="${KERNEL_VERSION#"$KERNEL_MAJOR".}"
```
Fix this.

2. These (invalid) warnings are also reported by the new version:
```
	In tests/integration/events.bats line 13:
	@test "events --stats" {
	^-- SC2030 (info): Modification of status is local (to subshell caused by @bats test).

	In tests/integration/events.bats line 41:
		[ "$status" -eq 0 ]
		   ^-----^ SC2031 (info): status was modified in a subshell. That change might be lost.
```
Basically, this is happening because shellcheck do not really track
the call tree and/or local variables. This is a known (and reported)
deficiency, and the alternative to disabling these warnings is moving
the code around, which is worse due to more changes in git history.

So we have to silence/disable these.

3. Update shellcheck to 0.8.0.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -96,9 +96,9 @@ jobs:
       - uses: actions/checkout@v2
       - name: vars
         run: |
-          echo 'VERSION=v0.7.2' >> $GITHUB_ENV
+          echo 'VERSION=v0.8.0' >> $GITHUB_ENV
           echo 'BASEURL=https://github.com/koalaman/shellcheck/releases/download' >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


kolyshkin:
Centos7 ci failure is some unrelated glitch; filed https://github.com/opencontainers/runc/issues/3372 to not forget.

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

PR #3373, opened 2022-02-11
Remove Factory and LinuxFactory

kolyshkin:
~~_Based on and currently includes #3374. Draft until that one is merged._~~

Continuing the spring cleaning started in #3354. This dismantles Factory and LinuxFactory, doing some cleanups along the way.

Loosely based on parts of #3316.

PR #3375, opened 2022-02-14
libct: rm BaseContainer and Container interfaces

kolyshkin:
~~_Based on and currently includes #3373. Draft until that one is merged._~~

The only implementation of `BaseContainer` and `Container` is `linuxContainer`. It does not make
sense to have an interface with a single implementation, and we do not
foresee other types of containers being added to runc.
    
 - Remove `BaseContainer` and`Container` interfaces, moving their methods
   documentation to `linuxContainer`.
 
- Rename `linuxContainer` to `Container`.
    
- Adopt users from using interface to using struct.

Continuing the spring cleaning started in #3354. Loosely based on parts of #3316.

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -13,7 +13,7 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func killContainer(container libcontainer.Container) error {
+func killContainer(container *libcontainer.Container) error {
 	_ = container.Signal(unix.SIGKILL, false)
 	for i := 0; i < 100; i++ {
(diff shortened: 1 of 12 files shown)


thaJeztah:
@kolyshkin this one can be rebased (was considering "we should remove those interfaces", then recalled: "we had a PR for that") 😂

AkihiroSuda (approved):

thaJeztah (approved):
LGTM!

PR #3377, opened 2022-02-16
libct/cg: support SCHED_IDLE for runc cgroupfs

wineway:
runtime-spec: https://github.com/opencontainers/runtime-spec/pull/1136

### DIFF
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -94,6 +94,14 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 			}
 		}
 	}
+
+	if r.CPUIdle != nil {
+		idle := strconv.FormatInt(*r.CPUIdle, 10)
+		if err := cgroups.WriteFile(path, "cpu.idle", idle); err != nil {
+			return err
+		}
+	}
+
 	return s.SetRtSched(path, r)
 }
 
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -84,6 +84,9 @@ type Resources struct {
 	// MEM to use
 	CpusetMems string `json:"cpuset_mems"`
 
+	// cgroup SCHED_IDLE
+	CPUIdle *int64 `json:"cpu_idle,omitempty"`
+
 	// Process limit; set <= `0' to disable limit.
 	PidsLimit int64 `json:"pids_limit"`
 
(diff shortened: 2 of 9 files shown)


wineway:
@kolyshkin PTAL

libcontainer/cgroups/fs/cpu.go
```diff
@@ -94,6 +94,14 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 			}
 		}
 	}
+
+	if r.CpuIdle != nil {
+		idle := strconv.FormatInt(*r.CpuIdle, 10)
+		if err := cgroups.WriteFile(path, "cpu.idle", idle); err != nil {
+			return err
```
kolyshkin, line 101: Maybe we should ignore the error `if idle == 0 and errors.Is(err, unix.ENOENT)`, since the default is 0 and ENOENT means we have an older kernel which essentially means the behavior is the same as if `0` is written.

libcontainer/configs/cgroup_linux.go
```diff
@@ -84,6 +84,10 @@ type Resources struct {
 	// MEM to use
 	CpusetMems string `json:"cpuset_mems"`
 
+	//nolint:revive
```
kolyshkin, line 87: Maybe instead of ignoring revive warning, call it `CPUIdle` (hopefully we'll rename the rest of `Cpu*` fields at some point).

libcontainer/specconv/spec_linux.go
```diff
@@ -733,6 +733,7 @@ func CreateCgroupConfig(opts *CreateOpts, defaultDevs []*devices.Device) (*confi
 				if r.CPU.Quota != nil {
 					c.Resources.CpuQuota = *r.CPU.Quota
 				}
+
```
kolyshkin, line 736: not needed.

kolyshkin (changes requested):
A few nits, plus can you please make the runtime-spec bump a separate commit (for clarity)?

kolyshkin:
Rebased to fix CI

libcontainer/cgroups/fs/cpu.go
wineway: It seems setting `cpu.idle` is illegal behavior in the older kernel which does not support it, should we throw an error explicitly to the user who wants to do this?

if we permit the user to set `cpu.idle=0` in this situation, I think this will make confused when the user want to modify `cpu.idle=1`

wineway:
@kolyshkin Can I ask you to review it again?

AkihiroSuda:
Looks good but needs rebase

AkihiroSuda:
@wineway Could you rebase?

wineway:
> @wineway Could you rebase?

sorry, I missed the last comment; rebased

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #3378, opened 2022-02-16
shellcheck/shfmt more files; run check-config in CI

kolyshkin:
This _almost_ fixes the TODO item for shellcheck.

Please review commit-by-commit; see individual commits for details.

### DIFF
--- a/man/md2man-all.sh
+++ b/man/md2man-all.sh
@@ -2,7 +2,7 @@
 set -e
 
 # get into this script's directory
-cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
+cd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
 
 [ "$1" = '-q' ] || {
 	set -x
@@ -18,7 +18,7 @@ for FILE in *.md; do
 	base="$(basename "$FILE")"
 	name="${base%.md}"
 	num="${name##*.}"
-	if [ -z "$num" -o "$name" = "$num" ]; then
+	if [ -z "$num" ] || [ "$name" = "$num" ]; then
 		# skip files that aren't of the format xxxx.N.md (like README.md)
 		continue
 	fi
--- a/script/check-config.sh
+++ b/script/check-config.sh
@@ -73,10 +73,19 @@ wrap_color() {
 }
 
 wrap_good() {
-	echo "$(wrap_color "$1" white): $(wrap_color "$2" green)"
+	local name="$1"
+	shift
+	local val="$1"
+	shift
+	echo "$(wrap_color "$name" white): $(wrap_color "$val" green)" "$@"
 }
+
 wrap_bad() {
-	echo "$(wrap_color "$1" bold): $(wrap_color "$2" bold red)"
+	local name="$1"
+	shift
+	local val="$1"
+	shift
+	echo "$(wrap_color "$name" bold): $(wrap_color "$val" bold red)" "$@"
 }
 wrap_warning() {
 	wrap_color >&2 "$*" red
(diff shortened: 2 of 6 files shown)


kolyshkin:
@tianon PTAL 🙏🏻

man/md2man-all.sh
```diff
@@ -18,7 +18,7 @@ for FILE in *.md; do
 	base="$(basename "$FILE")"
 	name="${base%.md}"
 	num="${name##*.}"
-	if [ -z "$num" -o "$name" = "$num" ]; then
+	if [[ -z "$num" || "$name" = "$num" ]]; then
```
tianon, line 21: This seems like a weird use of `[[` IMO -- shouldn't this just be:

```suggestion
	if [ -z "$num" ] || [ "$name" = "$num" ]; then
```

script/check-config.sh
```diff
@@ -73,10 +73,11 @@ wrap_color() {
 }
 
 wrap_good() {
-	echo "$(wrap_color "$1" white): $(wrap_color "$2" green)"
+	echo "$(wrap_color "$1" white): $(wrap_color "$2" green) ${3:-}"
```
tianon, line 76: If we use `shift`, we can use `"$@"` to get this extra functionality somewhat more safely:

```suggestion
	local name="$1"
	shift
	local value="$1"
	shift
	echo "$(wrap_color "$name" white): $(wrap_color "$value" green)" "$@"
```

script/check-config.sh
```diff
@@ -73,10 +73,11 @@ wrap_color() {
...
 }
+
 wrap_bad() {
-	echo "$(wrap_color "$1" bold): $(wrap_color "$2" bold red)"
+	echo "$(wrap_color "$1" bold): $(wrap_color "$2" bold red) ${3:-}"
```
tianon, line 80:
```suggestion
	local name="$1"
	shift
	local value="$1"
	shift
	echo "$(wrap_color "$name" bold): $(wrap_color "$value" bold red)" "$@"
```

script/check-config.sh
```diff
@@ -99,7 +100,8 @@ check_flags() {
 }
 
 check_distro_userns() {
-	source /etc/os-release 2>/dev/null || /bin/true
+	# shellcheck source=/dev/null
+	source /etc/os-release 2>/dev/null || return 0
```
tianon, line 104: Pursuant to https://github.com/moby/moby/pull/43230, I think this probably makes sense to also be a bit more explicitly protective of? :innocent:

```suggestion
	if [ ! -e /etc/os-release ]; then
		return 0
	fi
	# shellcheck source=/dev/null
	source /etc/os-release 2>/dev/null || return 0
```

(Although I see we're still `bash` here, hence `source` vs `.`, so grain of salt)

script/check-config.sh
```diff
@@ -118,8 +120,7 @@ is_config() {
 }
 
 search_config() {
-	local target_dir="$1"
-	[[ "$target_dir" ]] || target_dir=("${possibleConfigs[@]}")
+	local target_dir=("${1:-${possibleConfigs[@]}}")
```
tianon, line 123: I wasn't sure if the quoting here would have the right effect, but it appears that it does!

```console
$ arr=( 'foo bar' 'baz' ); foo() { local foo=("${1:-${arr[@]}}"); local bar; for bar in "${foo[@]}"; do echo "$bar"; done; }; foo bar; foo
bar
foo bar
baz
```

script/check-config.sh
```diff
@@ -212,11 +213,11 @@ flags=(
 )
 check_flags "${flags[@]}"
 
-if [ "$kernelMajor" -lt 5 ] || [ "$kernelMajor" -eq 5 -a "$kernelMinor" -le 1 ]; then
+if [ "$kernelMajor" -lt 5 ] || [[ "$kernelMajor" -eq 5 && "$kernelMinor" -le 1 ]]; then
```
tianon, line 216: I think you're probably going to like my `[[` avoidance even less here than usual :innocent: :sweat_smile:

```suggestion
if [ "$kernelMajor" -lt 5 ] || { [ "$kernelMajor" -eq 5 ] && [ "$kernelMinor" -le 1 ]; }; then
```

script/check-config.sh
```diff
@@ -212,11 +213,11 @@ flags=(
...
 	check_flags NF_NAT_IPV4
 fi
 
-if [ "$kernelMajor" -lt 5 ] || [ "$kernelMajor" -eq 5 -a "$kernelMinor" -le 2 ]; then
+if [ "$kernelMajor" -lt 5 ] || [[ "$kernelMajor" -eq 5 && "$kernelMinor" -le 2 ]]; then
```
tianon, line 220:
```suggestion
if [ "$kernelMajor" -lt 5 ] || { [ "$kernelMajor" -eq 5 ] && [ "$kernelMinor" -le 2 ]; }; then
```

(basically, these are probably only useful if you think it might be useful for this script to swap to `/bin/sh` someday, given what its purpose is)

script/check-config.sh
```diff
@@ -233,23 +234,23 @@ echo 'Optional Features:'
 
 	check_flags MEMCG_SWAP
 
-	if [ "$kernelMajor" -lt 5 ] || [ "$kernelMajor" -eq 5 -a "$kernelMinor" -le 8 ]; then
+	if [ "$kernelMajor" -lt 5 ] || [[ "$kernelMajor" -eq 5 && "$kernelMinor" -le 8 ]]; then
```
tianon, line 237:
```suggestion
	if [ "$kernelMajor" -lt 5 ] || { [ "$kernelMajor" -eq 5 ] && [ "$kernelMinor" -le 8 ]; }; then
```

script/check-config.sh
```diff
@@ -233,23 +234,23 @@ echo 'Optional Features:'
...
 		check_flags MEMCG_SWAP_ENABLED
 		if is_set MEMCG_SWAP && ! is_set MEMCG_SWAP_ENABLED; then
-			echo "    $(wrap_color '(note that cgroup swap accounting is not enabled in your kernel config, you can enable it by setting boot option "swapaccount=1")' bold black)"
+			wrap_color '    (note that cgroup swap accounting is not enabled in your kernel config, you can enable it by setting boot option "swapaccount=1")' bold black
 		fi
 	fi
 }
 
-if [ "$kernelMajor" -lt 4 ] || [ "$kernelMajor" -eq 4 -a "$kernelMinor" -le 5 ]; then
+if [ "$kernelMajor" -lt 4 ] || [[ "$kernelMajor" -eq 4 && "$kernelMinor" -le 5 ]]; then
```
tianon, line 245:
```suggestion
if [ "$kernelMajor" -lt 4 ] || { [ "$kernelMajor" -eq 4 ] && [ "$kernelMinor" -le 5 ]; }; then
```

script/check-config.sh
```diff
@@ -233,23 +234,23 @@ echo 'Optional Features:'
...
 	check_flags MEMCG_KMEM
 fi
 
-if [ "$kernelMajor" -lt 3 ] || [ "$kernelMajor" -eq 3 -a "$kernelMinor" -le 18 ]; then
+if [ "$kernelMajor" -lt 3 ] || [[ "$kernelMajor" -eq 3 && "$kernelMinor" -le 18 ]]; then
```
tianon, line 249:
```suggestion
if [ "$kernelMajor" -lt 3 ] || { [ "$kernelMajor" -eq 3 ] && [ "$kernelMinor" -le 18 ]; }; then
```

script/check-config.sh
```diff
@@ -233,23 +234,23 @@ echo 'Optional Features:'
...
 	check_flags RESOURCE_COUNTERS
 fi
 
-if [ "$kernelMajor" -lt 3 ] || [ "$kernelMajor" -eq 3 -a "$kernelMinor" -le 13 ]; then
+if [ "$kernelMajor" -lt 3 ] || [[ "$kernelMajor" -eq 3 && "$kernelMinor" -le 13 ]]; then
```
tianon, line 253:
```suggestion
if [ "$kernelMajor" -lt 3 ] || { [ "$kernelMajor" -eq 3 ] && [ "$kernelMinor" -le 13 ]; }; then
```

tianon (commented):
Hopefully my suggestions on these PRs aren't making you regret asking me for my thoughts on them :joy: :innocent:

Generally speaking, I try to avoid bash-isms unless they provide obvious value (or I know for certain the script will never want or need to become `/bin/sh`).  This makes sure that when I _do_ have to write `/bin/sh` (such as in `Dockerfile` `RUN` lines :sweat_smile:), I'm not constantly getting the syntax wrong because my muscle memory is already mostly right. :smile:

man/md2man-all.sh
kolyshkin: fixed

script/check-config.sh
kolyshkin: This new check better be `-r` not `-e` since if the file is present but not readable, shell may still error out.

Added.

script/check-config.sh
kolyshkin: done

kolyshkin:
@tianon addressed all your comments; PTAL

tianon:
Nice!  I love how much cleaner your `kernel_lt` function makes those checks :heart_eyes:

kolyshkin:
@thaJeztah @AkihiroSuda PTAL

kolyshkin:
@thaJeztah @AkihiroSuda PTAL (this was already blessed by @tianon)

thaJeztah (dismissed):
LGTM, nice!

kolyshkin:
Rebased. PTAL @thaJeztah @AkihiroSuda 🙏🏻

thaJeztah (approved):
still LGTM

kolyshkin:
Rebased; PTAL @AkihiroSuda

kolyshkin:
Do I still need to address anything in this PR? PTAL @thaJeztah @AkihiroSuda @opencontainers/runc-maintainers

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

PR #3379, opened 2022-02-16
ci: bump shfmt to 3.5.1, simplify CI setup

kolyshkin:
1. Bump shfmt to v3.5.1. Release notes:
    https://github.com/mvdan/sh/releases
    
2. Since shfmt v3.5.0, specifying `-l bash` (or `-l bats`) is no longer
     necessary. Therefore, we can use shfmt to find all the files.
     Add `.editorconfig` to ignore `vendor` subdirectory.
    
3. Use shfmt docker image, so that we don't have to install anything
    explicitly. This greatly simplifies the shfmt CI job. Add
    localshfmt target so developers can still use a local shfmt binary
    when necessary.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# This file is used by shfmt. See https://EditorConfig.org
+
+# This is a top-most EditorConfig file.
+root = true
+
+# Ignore the entire "vendor" directory.
+[vendor/**]
+ignore = true
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -67,22 +67,6 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
-    - name: vars
-      run: |
-        echo "VERSION=3.3.1" >> $GITHUB_ENV
-        echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
-    - name: cache go mod and $GOCACHE
-      uses: actions/cache@v3
-      with:
-        path: |
-          ~/go/pkg/mod
-          ~/.cache/go-build
-        key: ${{ runner.os }}-shfmt-${{ env.VERSION }}
-        restore-keys: ${{ runner.os }}-shfmt-
-    - name: install shfmt
-      run: |
-        command -v shfmt || \
-          (cd ~ && GO111MODULE=on time go get mvdan.cc/sh/v3/cmd/shfmt@v$VERSION)
     - name: shfmt
       run: make shfmt
 
--- a/Makefile
+++ b/Makefile
@@ -165,10 +165,12 @@ shellcheck:
 	# TODO: add shellcheck for more sh files (contrib/completions/bash/runc).
 
 shfmt:
-	shfmt -ln bats -d -w tests/integration/*.bats
-	shfmt -ln bash -d -w man/*.sh script/* \
-		tests/*.sh tests/integration/*.bash tests/fuzzing/*.sh \
-		contrib/completions/bash/runc contrib/cmd/seccompagent/*.sh
+	$(CONTAINER_ENGINE) run $(CONTAINER_ENGINE_RUN_FLAGS) \
+		--rm -v $(CURDIR):/src -w /src \
+		mvdan/shfmt:v3.5.1 -d -w .
+
+localshfmt:
+	shfmt -d -w .
 
 vendor:
 	$(GO) mod tidy
@@ -192,5 +194,5 @@ verify-dependencies: vendor
 	localrelease dbuild lint man runcimage \
 	test localtest unittest localunittest integration localintegration \
 	rootlessintegration localrootlessintegration shell install install-bash \
-	install-man clean cfmt shfmt shellcheck \
+	install-man clean cfmt shfmt localshfmt shellcheck \
 	vendor verify-changelog verify-dependencies


AkihiroSuda (dismissed):

kolyshkin:
Rebased; bumped shfmt to 3.4.3.

@AkihiroSuda @thaJeztah PTAL

kolyshkin:
CI is failing because of Azure Ubuntu mirror issues

> E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libc/libcap2/libcap-dev_2.32-1_amd64.deb  Connection failed [IP: 52.252.75.106 80]

will restart it later.

AkihiroSuda (approved):

.github/workflows/validate.yml
```diff
@@ -77,22 +77,9 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
-    - name: vars
-      run: |
-        echo "VERSION=3.3.1" >> $GITHUB_ENV
-        echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
-    - name: cache go mod and $GOCACHE
-      uses: actions/cache@v2
+    - uses:  mfinelli/setup-shfmt@v1
```
thaJeztah, line 80: I'm slightly on the fence on running a random action in CI;  seems like the "old" code was more complicated than it had to be? Would just a one-line `go install` work?;

```
GO111MODULE=on go inastall mvdan.cc/sh/v3/cmd/shfmt@v3.3.1
```

kolyshkin: The one-liner won't work because at the very least we need to add `GOPATH/bin` to `$PATH` (which the old code did).

For the newer shfmt, I guess, it won't work as ubuntu-20.04 used in this job comes with go 1.13.8, which might not be enough for shfmt 3.4.3.

Using a pre-released binary is better because:
 - no need to install golang
 - no need to download dependencies and build
 - the released binary is more stable/predictable than the one we build ourselves

If you don't like using a third-party action (I took a look at it and it seems short, focused, and being maintained), I can certainly rewrite this using wget (since it's Linux only we don't have to care about .exe so it should be simple enough).

Let me know how do you want this to be handled.

thaJeztah: It looks like Go 1.15 and up are on those runners (from the link in the workflow); https://github.com/actions/virtual-environments/blob/ubuntu20/20220227.1/images/linux/Ubuntu2004-Readme.md#go

Well, they're listed under "cached tools", so I _think_ they're there? (but with a custom path, so perhaps it expects the env-var to be se with go-setup); of course there's always the docker image we could run (and just 1 mb); https://hub.docker.com/r/mvdan/shfmt

kolyshkin: @thaJeztah PTAL ^^

kolyshkin: @thaJeztah PTAL ^^

dqminh: Fetching the prebuilt binary sounds good to me.

thaJeztah: If we stick with the action, at least we should pin it to a specific version.

I'm good with downloading the binary; that said; the docker image _is a single-binary image_, and docker should be present in GHA so would make it easier to run (and it's multi-arch, so no need to take different architectures into account)

Slightly orthogonal, but looking at the make target we currently use, it looks like we don't need to use two separate runs for `bats` and `bash` files; `shfmt` automatically detects what to use based on the file-extension, and once https://github.com/mvdan/sh/commit/a876f44daf649bc0184b0ad7685224a65fbce8f0 is in a release, even would detect a shebang in the file. Given that our `bats` files have a `.bats` extension, it looks like we don't need the shebang detection, so (using the docker image, but same for binary) so it looks like we can combine both runs (omitting the `-ln` option);

```bash
docker run --rm -v $(pwd):/src -w /src mvdan/shfmt -d -w man/*.sh script/* tests/*.sh tests/integration/*.bash tests/integration/*.bats
```

kolyshkin: > it looks like we don't need to use two separate runs for `bats` and `bash` files

Yes, I do have that in mind and even filed a few issues to shfmt for that to happen. Looks like it's fully working now so I added a commit that does that.

I don't like using docker from `Makefile` for shfmt; the main reason is I can use custom `shfmt` binary to check things out (and I did a number of times when helping in its development and testing).

Using docker from CI job file is all right I guess; the only problem is we have to repeat the `tests/integration/*.bats man/*.sh script/* tests/*.sh tests/integration/*.bash` line again, which creates a ever-so-slight maintenance burden (and a possiblity that CI job and the makefile target will stop being in sync).

A solution for that would be using `shfmt -f .`. I remember it was not working before for some reason (probably because we needed separate lists for bash and bats). Actually, it works now, so we no longer have to supply the fixed list of paths.

Switching this PR to use docker.

> If we stick with the action, at least we should pin it to a specific version.

Shouldn't we do that with docker image, too?

kolyshkin: Damn. shfmt 3.4.3 can not yet mix bats and bash in the same invocation (i.e. `-ln bash` and `-ln bats` are still required). So everything I did with Docker does not work.

Setting this on pause until shfmt 3.5.0 is released.

kolyshkin:
So, as I pointed out in https://github.com/opencontainers/runc/pull/3379#discussion_r833610510, we need shfmt 3.5.0 (not yet released) to further simplify things and use docker, so let's set this to pause for now.

kolyshkin:
OK, shfmt 3.5.0 is out as of 8 days ago, but the docker image is still not :(

kolyshkin:
> OK, shfmt 3.5.0 is out as of 8 days ago, but the docker image is still not :(

And the reason is https://github.com/mvdan/sh/issues/860

I'm going back to fetching the binary.

kolyshkin:
This only took a few months 🤣
@opencontainers/runc-maintainers PTAL (easy to review)

thaJeztah (dismissed):
LGTM 🎉

thaJeztah:
```
mvdan/shfmt:v3.5.1
-w cannot be used on standard input
make: *** [Makefile:168: shfmt] Error 1
Error: Process completed with exit code 2.

**Comment by kolyshkin** [comment]
> ```
> mvdan/shfmt:v3.5.1
> -w cannot be used on standard input
> make: *** [Makefile:168: shfmt] Error 1
> Error: Process completed with exit code 2.
> ```

Weird, I checked it locally (with podman) and it works.

Ah! We are using _local_ shfmt to generate the list of files, and the one inside the docker image to work on them.

Hmm, I am not quite sure how to fix this to use the in-the-image shfmt for both.

**Comment by kolyshkin** [comment]
Or don't use `shfmt -f`, replacing it with something else.

**Comment by kolyshkin** [comment]
OK I looked into the logic of `shfmt -l` and it is somewhat complicated:
 - it walks the directory tree;
 - prints all files which have `\.(sh|bash|mksh|bats|zsh)$` suffix;
 - for other files found, drop those that contain a dot `.` in their names;
 - for other files, check if their first line matches the `^#!\s?/(usr/)?bin/(env\s+)?(sh|bash|mksh|bats|zsh)(\s|$)` regex.

So, I am not going to recreate this functionality here.

Perhaps we can just go back to downloading a binary, wdyt @thaJeztah?

**Comment by kolyshkin** [comment]
> OK, `git ls-files | grep this | grep that` is good enough for now. Unlike shfmt -f, it does not look for shebangs, but the one file we have without the shell suffix is `contrib/completions/bash/runc`, and we can explicitly add it to the regexp.
> 
> I'd still love to use `shfmt -f` but what I have now will suffice.

Solved this one via `.editorconfig`.

PTAL @thaJeztah

File: .editorconfig [hunk 2]
```diff
@@ -0,0 +1,8 @@
+# See https://EditorConfig.org
```
AkihiroSuda, line 1: Could you add a comment to clarify that this file is for shfmt

AkihiroSuda (approved):

.editorconfig
kolyshkin: done

kolyshkin:
Rebased, addressed a nit from @AkihiroSuda 

@thaJeztah PTAL

AkihiroSuda:
Needs rebase

kolyshkin:
Rebased; PTAL @thaJeztah @AkihiroSuda

AkihiroSuda (approved):

mrunalp (approved):

kolyshkin:
CI failure is a flake, see https://github.com/opencontainers/runc/issues/3540; CI restarted.

PR #3381, opened 2022-02-18
ensure the path is a sub-cgroup path

lifubang:
As `runc exec --cgroup` means: run the process in an (existing) sub-cgroup(s), so I think we should ensure the given path in the param is a sub-cgroup path.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -632,7 +632,11 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, messageSockP
 			// cgroup v1: using the same path for all controllers.
 			// cgroup v2: the only possible way.
 			for k := range proc.cgroupPaths {
-				proc.cgroupPaths[k] = path.Join(proc.cgroupPaths[k], add)
+				subPath := path.Join(proc.cgroupPaths[k], add)
+				if !strings.HasPrefix(subPath, proc.cgroupPaths[k]) {
+					return nil, fmt.Errorf("%s is not a sub cgroup path", add)
+				}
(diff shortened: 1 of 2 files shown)


libcontainer/container_linux.go
```diff
@@ -632,7 +632,12 @@ func (c *linuxContainer) newSetnsProcess(p *Process, cmd *exec.Cmd, messageSockP
 			// cgroup v1: using the same path for all controllers.
 			// cgroup v2: the only possible way.
 			for k := range proc.cgroupPaths {
-				proc.cgroupPaths[k] = path.Join(proc.cgroupPaths[k], add)
+				subPath := path.Join(proc.cgroupPaths[k], add)
+				if strings.HasPrefix(subPath, proc.cgroupPaths[k]) {
+					proc.cgroupPaths[k] = subPath
+				} else {
+					return nil, fmt.Errorf("%s is not a sub cgroup path", add)
+				}
```
kolyshkin, line 640: This can be simplified to
```go
if !strings.HasPrefix(subPath, proc.cgroupPaths[k]) {
	return nil, fmt.Errorf("%s is not a sub cgroup path", add)
}
proc.cgroupPaths[k] = subPath
```

kolyshkin:
> We can do something like
> 
> ```go
> func checkPath(path) error {
>    if utils.CleanPath(path) != path {
>       return errors.New("bad sub-cgroup path: " + path)
>    }
> }
> ```

OTOH using `strings.HasPrefix` is fine, too.

Just one nit to avoid `else` in the implementation, otherwise LGTM.

kolyshkin (approved):
LGTM, thanks!

AkihiroSuda (approved):

PR #3385, opened 2022-02-19
init simplification

kolyshkin:
~~_Based on and currently includes #3375. Draft until that one is merged._~~
~~_Based on and currently includes #3854. Draft until that one is merged._~~

This  moves logging setup out of `func init()` and into `StartInitialization()` (where it should belong);

This concludes the spring cleaning started in #3354 (promise). Loosely based on parts of #3316.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -501,9 +501,10 @@ func (c *Container) commandTemplate(p *Process, childInitPipe *os.File, childLog
 
 	cmd.ExtraFiles = append(cmd.ExtraFiles, childLogPipe)
 	cmd.Env = append(cmd.Env,
-		"_LIBCONTAINER_LOGPIPE="+strconv.Itoa(stdioFdCount+len(cmd.ExtraFiles)-1),
-		"_LIBCONTAINER_LOGLEVEL="+p.LogLevel,
-	)
+		"_LIBCONTAINER_LOGPIPE="+strconv.Itoa(stdioFdCount+len(cmd.ExtraFiles)-1))
+	if p.LogLevel != "" {
+		cmd.Env = append(cmd.Env, "_LIBCONTAINER_LOGLEVEL="+p.LogLevel)
+	}
 
 	// NOTE: when running a container with no PID namespace and the parent process spawning the container is
 	// PID1 the pdeathsig is being delivered to the container's init process by the kernel for some reason
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -92,9 +92,7 @@ func StartInitialization() (retErr error) {
 	envInitPipe := os.Getenv("_LIBCONTAINER_INITPIPE")
 	pipefd, err := strconv.Atoi(envInitPipe)
 	if err != nil {
-		err = fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE: %w", err)
-		logrus.Error(err)
-		return err
+		return fmt.Errorf("unable to convert _LIBCONTAINER_INITPIPE: %w", err)
 	}
 	pipe := os.NewFile(uintptr(pipefd), "pipe")
 	defer pipe.Close()
@@ -111,6 +109,26 @@ func StartInitialization() (retErr error) {
 			return
 		}
 	}()
+
+	// Set up logging. This is used rarely, and mostly for init debugging.
+
+	// Passing log level is optional; currently libcontainer/integration does not do it.
+	if levelStr := os.Getenv("_LIBCONTAINER_LOGLEVEL"); levelStr != "" {
+		logLevel, err := strconv.Atoi(levelStr)
+		if err != nil {
+			return fmt.Errorf("unable to convert _LIBCONTAINER_LOGLEVEL: %w", err)
+		}
+		logrus.SetLevel(logrus.Level(logLevel))
+	}
+
+	logFD, err := strconv.Atoi(os.Getenv("_LIBCONTAINER_LOGPIPE"))
+	if err != nil {
+		return fmt.Errorf("unable to convert _LIBCONTAINER_LOGPIPE: %w", err)
+	}
+
+	logrus.SetOutput(os.NewFile(uintptr(logFD), "logpipe"))
+	logrus.SetFormatter(new(logrus.JSONFormatter))
+	logrus.Debug("child process in init()")
 
 	// Only init processes have FIFOFD.
 	fifofd := -1
@@ -131,12 +149,6 @@ func StartInitialization() (retErr error) {
 		}
 		consoleSocket = os.NewFile(uintptr(console), "console-socket")
 		defer consoleSocket.Close()
-	}
-
-	logPipeFdStr := os.Getenv("_LIBCONTAINER_LOGPIPE")
-	logPipeFd, err := strconv.Atoi(logPipeFdStr)
-	if err != nil {
-		return fmt.Errorf("unable to convert _LIBCONTAINER_LOGPIPE: %w", err)
 	}
 
 	// Get mount files (O_PATH).
@@ -166,7 +178,7 @@ func StartInitialization() (retErr error) {
 	}()
 
 	// If init succeeds, it will not return, hence none of the defers will be called.
-	return containerInit(it, pipe, consoleSocket, fifofd, logPipeFd, mountFds{sourceFds: mountSrcFds, idmapFds: idmapFds})
+	return containerInit(it, pipe, consoleSocket, fifofd, logFD, mountFds{sourceFds: mountSrcFds, idmapFds: idmapFds})
 }
 
 func containerInit(t initType, pipe *os.File, consoleSocket *os.File, fifoFd, logFd int, mountFds mountFds) error {
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"runtime"
 	"runtime/debug"
 	"strconv"
 	"strings"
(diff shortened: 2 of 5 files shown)


kolyshkin:
Rebased on top of recently rebased #3375

kolyshkin:
Rebased; no longer a draft! Also, this concludes what was started as #3316

libcontainer/init_linux.go
```diff
@@ -167,52 +167,45 @@ func StartInitialization() (err error) {
...
 			pipe:          pipe,
 			consoleSocket: consoleSocket,
 			parentPid:     unix.Getppid(),
 			config:        config,
 			fifoFd:        fifoFd,
 			logFd:         logFd,
 			mountFds:      mountFds,
-		}, nil
+		}
+		return i.Init()
```
kolyshkin, line 206: This looks a tad weird. Maybe we should have `standardInit()` and `setnsInit() functions with the appropriate arguments instead. Any suggestions are welcome.

kolyshkin: I took a look and it's better stay as is, at least for now.

kolyshkin:

kolyshkin:
Rebased; @opencontainers/runc-maintainers PTAL (this concludes a very long and heavy refactoring, the new code is easier to follow and understand).

kolyshkin:
@thaJeztah @AkihiroSuda @cyphar PTAL

thaJeztah:
The first commit still has this paragraph in the commit message, which no longer appears to be the case;

> Remove logrus setup and use from TestMain/init, in preparation for the
next commit.

libcontainer/factory_linux.go
```diff
@@ -160,6 +160,28 @@ func Load(root, id string) (*Container, error) {
 // to read the configuration and state. This is a low level implementation
 // detail of the reexec and should not be consumed externally.
 func StartInitialization() (err error) {
+	// Set up logging.
+	level := int(logrus.DebugLevel) // default to debug
+	// Passing log level is optional; currently libcontainer/integration does not do it.
+	if levelStr := os.Getenv("_LIBCONTAINER_LOGLEVEL"); levelStr != "" {
+		level, err = strconv.Atoi(levelStr)
+		if err != nil {
+			panic(fmt.Errorf("unable to convert _LIBCONTAINER_LOGLEVEL: %w", err))
```
thaJeztah: Can we change these to return error instead of panicking? Previously this code was in `init()` which required a panic, but here we can just return if I guess

libcontainer/factory_linux.go
```diff
@@ -160,6 +160,28 @@ func Load(root, id string) (*Container, error) {
...
+		}
+	}
+
+	logPipeFd, err := strconv.Atoi(os.Getenv("_LIBCONTAINER_LOGPIPE"))
+	if err != nil {
+		panic(fmt.Errorf("unable to convert _LIBCONTAINER_LOGPIPE: %w", err))
```
thaJeztah: same for this one

libcontainer/init_linux.go
```diff
@@ -71,6 +73,108 @@ type initConfig struct {
 	Cgroup2Path      string                `json:"cgroup2_path,omitempty"`
 }
 
+// StartInitialization loads a container by opening the pipe fd from the parent
+// to read the configuration and state. This is a low level implementation
+// detail of the reexec and should not be consumed externally.
+func StartInitialization() (err error) {
```
thaJeztah, line 79: not new, but can we rename the named error return? It's hard to see if it's potentially shadowed in various locations; (`newVar, err := something()`)

Or we should rewrite some code to explicitly declare variables to be a bit more defensive

libcontainer/init_linux.go
```diff
@@ -71,6 +73,108 @@ type initConfig struct {
...
+	if err != nil {
+		return err
+	}
+
+	// clear the current process's environment to clean any libcontainer
+	// specific env vars.
+	os.Clearenv()
+
+	defer func() {
+		if e := recover(); e != nil {
```
thaJeztah, line 160: if we rename the named return, we can probably rename `e` to a more common `err`

libcontainer/init_linux.go
```diff
@@ -166,52 +166,45 @@ func StartInitialization() (err error) {
...
 		}
 
-		return &linuxSetnsInit{
+		i := &linuxSetnsInit{
 			pipe:          pipe,
 			consoleSocket: consoleSocket,
 			config:        config,
 			logFd:         logFd,
-		}, nil
+		}
```
thaJeztah, line 193: Are you planning to tackle this one as well in a similar way?

kolyshkin:
> The first commit still has this paragraph in the commit message, which no longer appears to be the case;
> 
> > Remove logrus setup and use from TestMain/init, in preparation for the
> > next commit.

@thaJeztah why do you think so? The first commit removes the logrus setup from TestMain, and
it also removes logrus use from init() in the same file.

libcontainer/init_linux.go
kolyshkin: I'd rather leave this as is for now, this is only used on these few lines and `e` and `ee` seems fine.

kolyshkin: On the second thought, I take it back. Renamed; see the last commit.

libcontainer/init_linux.go
kolyshkin: Done, thanks for the suggestion -- I think this made the code a tad cleaner.

libcontainer/init_linux.go
kolyshkin: Sorry, I do not understand the question :-\

libcontainer/factory_linux.go
kolyshkin: Done

libcontainer/factory_linux.go
kolyshkin: done

kolyshkin:
Let me simplify this even further.

kolyshkin:
OK, I wanted to remove logging from the Go part of `runc init`, but decided to not do it. So, going back to the original plan.

In short, this is a series of small commits that untangle some logic around `runc init` (mostly the logging part), hopefully making it easier to grok. In the process, I am fixing a very minor and subtle bug of not reporting the `unable to convert _LIBCONTAINER_INITPIPE`.

libcontainer/init_linux.go
```diff
@@ -99,6 +97,26 @@ func StartInitialization() (retErr error) {
...
 
+	// Set up logging. This is used rarely, and mostly for init debugging.
+	logLevel := int(logrus.DebugLevel) // default to debug
+	// Passing log level is optional; currently libcontainer/integration does not do it.
+	if levelStr := os.Getenv("_LIBCONTAINER_LOGLEVEL"); levelStr != "" {
+		logLevel, err = strconv.Atoi(levelStr)
+		if err != nil {
+			return fmt.Errorf("unable to convert _LIBCONTAINER_LOGLEVEL: %w", err)
+		}
+	}
```
thaJeztah, line 114: It looks like now we're defaulting to "Debug" logs, which previously wasn't the case? Should we only set the log-level if any was passed?

```suggestion
	// Passing log level is optional; currently libcontainer/integration does not do it.
	if levelStr := os.Getenv("_LIBCONTAINER_LOGLEVEL"); levelStr != "" {
		logLevel, err := strconv.Atoi(levelStr)
		if err != nil {
			return fmt.Errorf("unable to convert _LIBCONTAINER_LOGLEVEL: %w", err)
		}
		logrus.SetLevel(logrus.Level(logLevel))
	}
```

libcontainer/container_linux.go
```diff
@@ -488,9 +488,10 @@ func (c *Container) commandTemplate(p *Process, childInitPipe *os.File, childLog
 
 	cmd.ExtraFiles = append(cmd.ExtraFiles, childLogPipe)
 	cmd.Env = append(cmd.Env,
-		"_LIBCONTAINER_LOGPIPE="+strconv.Itoa(stdioFdCount+len(cmd.ExtraFiles)-1),
-		"_LIBCONTAINER_LOGLEVEL="+p.LogLevel,
-	)
+		"_LIBCONTAINER_LOGPIPE="+strconv.Itoa(stdioFdCount+len(cmd.ExtraFiles)-1))
+	if p.LogLevel != "" {
```
thaJeztah, line 492: Not new, but `LogLevel` is weird; it's a string, but it's not storing a named level, but a string representation of the numeric value (this code is converting the `logrus.Level` to an integer -> string, instead of taking its name; https://github.com/opencontainers/runc/blob/5f5b35dad46455dddd007d95b12ae78081377a02/utils_linux.go#L217

libcontainer/integration/init_test.go
```diff
@@ -21,10 +22,14 @@ func init() {
 	runtime.GOMAXPROCS(1)
 	runtime.LockOSThread()
 	if err := libcontainer.StartInitialization(); err != nil {
-		// as the error is sent back to the parent there is no need to log
-		// or write it to stderr because the parent process will handle this
-		os.Exit(1)
```
thaJeztah: Looks like this is the code that was added in the first commit; perhaps it would make sense to move this commit (and the previous one) to the start to the list of commits.

kolyshkin: So, the first commit ("libct/int: remove logger from init") is a preparation for "init.go: move logger setup to StartInitialization". Logically, they belong together, but since test changes in libct/int are somewhat larger (they also involve a test case), I separate them out for easier review.

Now, the change in the first commit (""libct/int: remove logger from init") brings libcontainer/integration `func init()` in line with the main runc's `func init`. Later, we change `StartInitialization` to return error in some cases, so we remove the `os.Exit(1)`.

All I can do is to rearrange the commits so these two are following each other:
- libct/int: remove logger from init
- init.go: move logger setup to StartInitialization

kolyshkin: Now, the code added is correct (when it is added), but since "libct/StartInitialization: amend docs, do not error" it needs to be fixed. Thus the two changes. Hope that makes sense.

libcontainer/container_linux.go
kolyshkin: This was done deliberately in commit 6c4a3b13d132aa460e610cf2c972e9ed71e4a42d (part of #3201) as a preparation for f1b703fc45dd61429634e9d448206c9b75d87147 (which would be much more complicated otherwise).

I guess what's missing from that PR is a docstring saying this is a string representation of a numeric log level. I'll open a PR to add it.

kolyshkin: https://github.com/opencontainers/runc/pull/3852

libcontainer/init_linux.go
kolyshkin: done

kolyshkin:
It's hard to split changes to separate small commits, as they are all inter-dependent. But I did another try; PTAL.

kolyshkin:
OK, maybe so many commits are way too much to digest. I have separated some prep commits to #3854, and made this one a draft.

libcontainer/init_linux.go
```diff
@@ -149,7 +161,7 @@ func StartInitialization() (retErr error) {
 	}()
 
 	// If init succeeds, it will not return, hence none of the defers will be called.
-	return containerInit(it, pipe, consoleSocket, fifofd, logPipeFd, mountFds)
+	return containerInit(it, pipe, consoleSocket, fifofd, logFD, mountFds)
```
kolyshkin, line 187: OK, we now use 3 different ways to say "FD", but I'm pretty sure the last one is the best one.

kolyshkin:
@thaJeztah @AkihiroSuda PTAL

AkihiroSuda (approved):

AkihiroSuda:
Needs rebase

kolyshkin:
Rebased. @lifubang @thaJeztah PTAL

lifubang (approved):

libcontainer/init_linux.go
cyphar: I personally prefer Fd. I can write up a PR to unify this.

Issue #3387, opened 2022-02-21
v2 cgroup delegation fails: memory.oom.group: no such file or directory


dkadioglu:
After the update of runc from 1.0.3 to 1.1.0, one of my containers doesn't start anymore and errors out with the following message:
```Error response from daemon: Cannot restart container dkhm: failed to create shim: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: chown /sys/fs/cgroup/system.slice/docker-77a1ac22d8bfb1658e9bb8430d7dd8c75e8fad69b4e3b389f8d25a09fde26d21.scope/memory.oom.group: no such file or directory: unknown``` Downgrading to 1.0.3 fixes it and the container starts again.

My 16 other containers run as before.

My host:
- RPi 4B 8G
- Arch Linux ARM fully updated

I have no clue where to start searching for the culprit of all this. Is there any further info, which might help to investigate this?

kolyshkin:
Thanks for reporting this!

Looks like a regression from #3057. The workaround for you would be to NOT mount /sys/fs/cgroup read-write.

The proposed fix is https://github.com/opencontainers/runc/pull/3389. You can test it by using a pre-built binary for your arch (arm64 I guess) from the zip archive at the very bottom of this page: https://github.com/opencontainers/runc/actions/runs/1877860537

dkadioglu:
Thank you very much for looking into this! And for providing a fix so fast!
I rebuilt runc 1.1.0 including the mentioned patch and so far it seems to work again, i.e. the respective containers starts and runs now. Perfect!
Just to mention: I am still on armv7h.

PR #3389, opened 2022-02-21
libct/cg/sd/v2: fix ENOENT on cgroup delegation

kolyshkin:
Apparently, not all files listed in `/sys/kernel/cgroup/delegate` must
exist in every cgroup, so we should ignore ENOENT.

Dot not ignore ENOENT on the directory itself though.

Change `cgroupFilesToChown` to not return `.`, and refactor it to not do
any dynamic slice appending in case we're using the default built-in
list of files.

Fixes: #3387
Fixes: 35d20c4e0

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -2,6 +2,7 @@ package systemd
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"math"
 	"os"
@@ -292,6 +293,12 @@ func (m *unifiedManager) Apply(pid int) error {
 	}
 
 	if c.OwnerUID != nil {
+		// The directory itself must be chowned.
+		err := os.Chown(m.path, *c.OwnerUID, -1)
+		if err != nil {
+			return err
+		}
+
 		filesToChown, err := cgroupFilesToChown()
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


kolyshkin:
@frasertweedale PTAL

frasertweedale (approved):

frasertweedale:
/lgtm

Thanks for the fix @kolyshkin .  I'll also create a PR against runtime-spec to add a note about the possibility of missing files.

kolyshkin:

libcontainer/cgroups/systemd/v2.go
```diff
@@ -312,21 +320,23 @@ func (m *unifiedManager) Apply(pid int) error {
...
-		scanner := bufio.NewScanner(f)
-		for scanner.Scan() {
-			filesToChown = append(filesToChown, scanner.Text())
-		}
-		if err := scanner.Err(); err != nil {
-			return nil, fmt.Errorf("error reading %s: %w", cgroupDelegateFile, err)
-		}
-	} else {
-		filesToChown = append(filesToChown, "cgroup.procs", "cgroup.subtree_control", "cgroup.threads")
+	if err != nil {
```
AkihiroSuda, line 326: Can we logrus.Debugf the err?

AkihiroSuda (approved):

Issue #2430, opened 2020-05-25
support seccomp flags such as SECCOMP_FILTER_FLAG_SPEC_ALLOW (OCI Runtime Spec v1.0.2)


AkihiroSuda:
OCI Runtime Spec v1.0.2 supports specifying three seccomp flags: `SECCOMP_FILTER_FLAG_TSYNC`, `SECCOMP_FILTER_FLAG_LOG`, and `SECCOMP_FILTER_FLAG_SPEC_ALLOW` (https://github.com/opencontainers/runtime-spec/commit/d1ef109cd0b39239ff82c267df314f7ed2da576b).
However, these flags are currently unimplemented by runc (but implemented by crun).

Notably we should support `SECCOMP_FILTER_FLAG_SPEC_ALLOW`  ([Disable Speculative Store Bypass mitigation, since Linux 4.17](http://man7.org/linux/man-pages/man2/seccomp.2.html)).
The mitigation is enabled by default when a seccomp profile is specified and has serious performance impact on bytecode interpreters including Ruby and Python.

http://mamememo.blogspot.com/2020/05/cpu-intensive-rubypython-code-runs.html

> On the host:
> ```console
> $ ruby -ve 't = Time.now; i=0;while i<100_000_000;i+=1;end; puts "#{ Time.now - t } sec"'
> ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]
> 1.321703922 sec
> ```
>
> On a Docker container:
> ```console
> $ docker run -it --rm ruby:2.7 ruby -ve 't = Time.now; i=0;while i<100_000_000;i+=1;end; puts "#{ Time.now - t } sec"'
> ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]
> 2.452876383 sec
> ```
>
> If you specify an option `"--security-opt seccomp=unconfined"` for `docker run` command, it runs as fast as the host.
> ```console
> $ docker run --security-opt seccomp=unconfined -it --rm ruby:2.7 ruby -ve 't = Time.now; i=0;while i<100_000_000;i+=1;end; puts "#{ Time.now - t } sec"'
>  ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux]
>  1.333669449 sec
> ```

AkihiroSuda:
cc @KentaTada

KentaTada:
@AkihiroSuda 
[...]

PR #3390, opened 2022-02-22
seccomp: add support for flags

alban:
List of seccomp flags defined in runtime-spec:
* SECCOMP_FILTER_FLAG_TSYNC
* SECCOMP_FILTER_FLAG_LOG
* SECCOMP_FILTER_FLAG_SPEC_ALLOW

Note that runc does not apply SECCOMP_FILTER_FLAG_TSYNC. It does not
make sense to apply the seccomp filter on only one thread; other threads
will be terminated after exec anyway.

See similar commit in crun:
https://github.com/containers/crun/commit/fefabffa2816ea343068ed036a86944393db189a

Note that SECCOMP_FILTER_FLAG_WAIT_KILLABLE_RECV (introduced by
https://git.kernel.org/pub/scm/linux/kernel/git/kees/linux.git/commit/?id=c2aa2dfef243
in Linux 5.19-rc1) is not added yet because Linux 5.19 is not released
yet.

-----

Tests:
```
$ make localintegration
...
ok 103 runc run [seccomp] (SECCOMP_FILTER_FLAG_*)
```

cc @rata @saschagrunert

### DIFF
--- a/libcontainer/configs/config.go
+++ b/libcontainer/configs/config.go
@@ -31,12 +31,13 @@ type IDMap struct {
 // for syscalls. Additional architectures can be added by specifying them in
 // Architectures.
 type Seccomp struct {
-	DefaultAction    Action     `json:"default_action"`
-	Architectures    []string   `json:"architectures"`
-	Syscalls         []*Syscall `json:"syscalls"`
-	DefaultErrnoRet  *uint      `json:"default_errno_ret"`
-	ListenerPath     string     `json:"listener_path,omitempty"`
-	ListenerMetadata string     `json:"listener_metadata,omitempty"`
+	DefaultAction    Action                   `json:"default_action"`
+	Architectures    []string                 `json:"architectures"`
+	Flags            []specs.LinuxSeccompFlag `json:"flags"`
+	Syscalls         []*Syscall               `json:"syscalls"`
+	DefaultErrnoRet  *uint                    `json:"default_errno_ret"`
+	ListenerPath     string                   `json:"listener_path,omitempty"`
+	ListenerMetadata string                   `json:"listener_metadata,omitempty"`
 }
 
 // Action is taken upon rule match in Seccomp
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/seccomp/patchbpf"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
 var (
@@ -83,6 +84,29 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 		}
 		if err := filter.AddArch(scmpArch); err != nil {
 			return -1, fmt.Errorf("error adding architecture to seccomp filter: %w", err)
+		}
+	}
+
+	// Add extra flags
+	for _, flag := range config.Flags {
+		switch flag {
+		case "SECCOMP_FILTER_FLAG_TSYNC":
+			// libseccomp-golang always use filterAttrTsync when
+			// possible so all goroutines will receive the same
+			// rules, so there is nothing to do. It does not make
+			// sense to apply the seccomp filter on only one
+			// thread; other threads will be terminated after exec
+			// anyway.
+		case specs.LinuxSeccompFlagLog:
+			if err := filter.SetLogBit(true); err != nil {
+				return -1, fmt.Errorf("error adding log flag to seccomp filter: %w", err)
+			}
+		case specs.LinuxSeccompFlagSpecAllow:
+			if err := filter.SetSSB(true); err != nil {
+				return -1, fmt.Errorf("error adding SSB flag to seccomp filter: %w", err)
+			}
+		default:
+			return -1, fmt.Errorf("seccomp flags %q not yet supported by runc", flag)
 		}
 	}
 
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -1016,13 +1016,21 @@ func SetupSeccomp(config *specs.LinuxSeccomp) (*configs.Seccomp, error) {
 		return nil, nil
 	}
 
-	// We don't currently support seccomp flags.
-	if len(config.Flags) != 0 {
-		return nil, errors.New("seccomp flags are not yet supported by runc")
-	}
-
 	newConfig := new(configs.Seccomp)
 	newConfig.Syscalls = []*configs.Syscall{}
+
+	// The list of flags defined in runtime-spec is a subset of the flags
+	// in the seccomp() syscall
+	for _, flag := range config.Flags {
+		switch flag {
+		case "SECCOMP_FILTER_FLAG_TSYNC":
+			// Tsync can be silently ignored
+		case specs.LinuxSeccompFlagLog, specs.LinuxSeccompFlagSpecAllow:
+			newConfig.Flags = append(newConfig.Flags, flag)
+		default:
+			return nil, fmt.Errorf("seccomp flag %q not yet supported by runc", flag)
+		}
+	}
 
 	if len(config.Architectures) > 0 {
 		newConfig.Architectures = []string{}
--- a/tests/integration/seccomp.bats
+++ b/tests/integration/seccomp.bats
@@ -66,6 +66,38 @@ function teardown() {
 	[[ "$output" == *"Network is down"* ]]
 }
 
+@test "runc run [seccomp] (SECCOMP_FILTER_FLAG_*)" {
+	# Linux 4.14: SECCOMP_FILTER_FLAG_LOG
+	# Linux 4.17: SECCOMP_FILTER_FLAG_SPEC_ALLOW
+	requires_kernel 4.17
+	SECCOMP_FILTER_FLAGS=(
+		'' # no flag
+		'"SECCOMP_FILTER_FLAG_LOG"'
+		'"SECCOMP_FILTER_FLAG_SPEC_ALLOW"'
+		'"SECCOMP_FILTER_FLAG_TSYNC"'
+		'"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_SPEC_ALLOW"'
+		'"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_TSYNC"'
+		'"SECCOMP_FILTER_FLAG_SPEC_ALLOW","SECCOMP_FILTER_FLAG_TSYNC"'
+		'"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_SPEC_ALLOW","SECCOMP_FILTER_FLAG_TSYNC"'
+	)
+	for flags in "${SECCOMP_FILTER_FLAGS[@]}"; do
+		update_config '   .process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"]
+				| .process.noNewPrivileges = false
+				| .linux.seccomp = {
+					"defaultAction":"SCMP_ACT_ALLOW",
+					"architectures":["SCMP_ARCH_X86","SCMP_ARCH_X32","SCMP_ARCH_X86_64","SCMP_ARCH_AARCH64","SCMP_ARCH_ARM"],
+					"flags":['"${flags}"'],
+					"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO"}]
+				}'
+
+		# This test checks that the flags are accepted without errors but does
+		# not check they are effectively applied
+		runc run test_busybox
+		[ "$status" -ne 0 ]
+		[[ "$output" == *"mkdir:"*"/dev/shm/foo"*"Operation not permitted"* ]]
+	done
+}
+
 @test "runc run [seccomp] (SCMP_ACT_KILL)" {
 	update_config '  .process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"]
 			| .process.noNewPrivileges = false
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/moby/sys/mountinfo v0.6.2
 	github.com/mrunalp/fileutils v0.5.0
-	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
+	github.com/opencontainers/runtime-spec v1.0.3-0.20220718201635-a8106e99982b
 	github.com/opencontainers/selinux v1.10.1
 	github.com/seccomp/libseccomp-golang v0.10.0
 	github.com/sirupsen/logrus v1.8.1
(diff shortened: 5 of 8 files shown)


tests/integration/seccomp.bats
```diff
@@ -66,6 +66,22 @@ function teardown() {
...
+	update_config '   .process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"]
+			| .process.noNewPrivileges = false
+			| .linux.seccomp = {
+				"defaultAction":"SCMP_ACT_ALLOW",
+				"architectures":["SCMP_ARCH_X86","SCMP_ARCH_X32"],
+				"flags":["SECCOMP_FILTER_FLAG_LOG"],
+				"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO"}]
+			}'
+
+	runc run test_busybox
```
rata, line 80: I'd add a comment saying we just test that the flag is accepted and the container runs as expected. But we don't read the logs

rata (dismissed):
I'm no maintainer here, but this LGTM. Just a minor comment on the integration test

tests/integration/seccomp.bats
alban: Thanks. Updated.

kolyshkin (changes requested):
> I didn't add support for `SECCOMP_FILTER_FLAG_SPEC_ALLOW` (#2430) because libseccomp-golang does not support it yet.

It does; calling `filter.SetSSB(true)` will enable `SECCOMP_FILTER_FLAG_SPEC_ALLOW`.

libcontainer/specconv/spec_linux.go
```diff
@@ -1016,14 +1016,19 @@ func SetupSeccomp(config *specs.LinuxSeccomp) (*configs.Seccomp, error) {
...
-		return nil, errors.New("seccomp flags are not yet supported by runc")
-	}
-
 	newConfig := new(configs.Seccomp)
 	newConfig.Syscalls = []*configs.Syscall{}
 
+	// We don't currently support all seccomp flags.
+	for _, flag := range config.Flags {
+		switch flag {
+		case "SECCOMP_FILTER_FLAG_LOG":
```
kolyshkin, line 1025: It would be nice to not have this set of flags handled in two separate places. Perhaps we can add some common part to `libcontainer/seccomp`?

While at it, adding `SCMP_FLTATR_CTL_OPTIMIZE` would be awesome

saschagrunert: Right now we have four supported flags following the same naming convention supported by the SECCOMP_SET_MODE_FILTER operation.

SECCOMP_FILTER_FLAG_LOG
SECCOMP_FILTER_FLAG_NEW_LISTENER
SECCOMP_FILTER_FLAG_SPEC_ALLOW
SECCOMP_FILTER_FLAG_TSYNC

How would `SCMP_FLTATR_CTL_OPTIMIZE` map to that?

alban: > While at it, adding `SCMP_FLTATR_CTL_OPTIMIZE` would be awesome

The [set of flags listed in the runtime-spec](https://github.com/opencontainers/runtime-spec/blob/8958f93039ab90be53d803cd7e231a775f644451/config-linux.md) is a [subset](https://github.com/opencontainers/runtime-spec/pull/1096) of the flag supported by the Linux API in the seccomp syscall:

> flags (array of strings, OPTIONAL) - list of flags to use with seccomp(2).
>
> A valid list of constants is shown below.
>
> SECCOMP_FILTER_FLAG_TSYNC
> SECCOMP_FILTER_FLAG_LOG
> SECCOMP_FILTER_FLAG_SPEC_ALLOW

`SCMP_FLTATR_CTL_OPTIMIZE` is part of libseccomp API but not part of the Linux API. It is not in the runtime-spec because container runtimes don't have to use libseccomp. Are you suggesting runc should support it even if it's not in the spec?

alban: > It would be nice to not have this set of flags handled in two separate places. Perhaps we can add some common part to libcontainer/seccomp?

Now that I have added the 2 additional flags, the code looks different, and I don't see how to add some common part to libcontainer/seccomp... any suggestions?

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -86,6 +86,29 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 		}
 	}
 
+	// Add extra flags
+	for _, flag := range config.Flags {
+		switch flag {
+		case "SECCOMP_FILTER_FLAG_TSYNC":
```
saschagrunert, line 92: Note: I think we should have them as consts in the runtime spec. I proposed that change in https://github.com/opencontainers/runtime-spec/pull/1138

saschagrunert (dismissed):

tests/integration/seccomp.bats
```diff
@@ -66,6 +66,58 @@ function teardown() {
...
+				"flags":["SECCOMP_FILTER_FLAG_TSYNC"],
+				"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO"}]
+			}'
+
+	# This test checks that the tsync flag is accepted
+	runc run test_busybox
+	[ "$status" -ne 0 ]
+	[[ "$output" == *"mkdir:"*"/dev/shm/foo"*"Operation not permitted"* ]]
+}
+
```
kolyshkin, line 120: I don't like the same code being repeated 3 times. Can probably factor it out to a function accepting the list of flags. That way we can also test a combination of flags, not just individual ones.

libcontainer/specconv/spec_linux.go
```diff
@@ -1016,14 +1016,19 @@ func SetupSeccomp(config *specs.LinuxSeccomp) (*configs.Seccomp, error) {
...
-	if len(config.Flags) != 0 {
-		return nil, errors.New("seccomp flags are not yet supported by runc")
-	}
-
 	newConfig := new(configs.Seccomp)
 	newConfig.Syscalls = []*configs.Syscall{}
 
+	// We don't currently support all seccomp flags.
+	for _, flag := range config.Flags {
+		switch flag {
```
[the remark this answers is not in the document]
kolyshkin: `SECCOMP_FILTER_FLAG_TSYNC` is there already and it's on by default (in libseccomp-golang). How would it be supported? Do you mean "let's explicitly ignore it", as I see no other way.

libcontainer/specconv/spec_linux.go
kolyshkin: Re `SCMP_FLTATR_CTL_OPTIMIZE`.

Perhaps runc should just set it to `2` unconditionally. Also, this is orthogonal to this PR, so nevermind.

libcontainer/specconv/spec_linux.go
```diff
@@ -1016,14 +1016,20 @@ func SetupSeccomp(config *specs.LinuxSeccomp) (*configs.Seccomp, error) {
...
-	}
-
 	newConfig := new(configs.Seccomp)
 	newConfig.Syscalls = []*configs.Syscall{}
 
+	// The list of flags defined in runtime-spec is a subset of the flags
+	// in the seccomp() syscall
+	for _, flag := range config.Flags {
+		switch flag {
+		case "SECCOMP_FILTER_FLAG_TSYNC", "SECCOMP_FILTER_FLAG_LOG", "SECCOMP_FILTER_FLAG_SPEC_ALLOW":
```
kolyshkin, line 1026: `SECCOMP_FILTER_FLAG_TSYNC` can just be silently ignored, i.e. it does not make sense to pass it around.

libcontainer/configs/config.go
```diff
@@ -33,6 +33,7 @@ type IDMap struct {
 type Seccomp struct {
 	DefaultAction    Action     `json:"default_action"`
 	Architectures    []string   `json:"architectures"`
+	Flags            []string   `json:"flags"`
```
kolyshkin, line 36: So, this is our internal representation. Not sure why we use strings here at all. Could be
```go
const SecccompFilterFlags (
	SeccompTSync = iota
	SeccompFlagLog
	SeccompFlagSpecAllow
)
```
and then specconv can do the conversion.

Yet better, do that in libseccomp-golang can have these. I'll take a look.

libcontainer/seccomp/seccomp_linux.go
kolyshkin: Note that seccomp-golang sets this flag unconditionally, and we (runc/libcontainer/seccomp/patchbpf) do not set this at all (not sure if this was by design or just forgotten; @cyphar WDYT?)

From what I see, it seems that for runc it does not make sense to do so, as we execute runc init right after applying seccomp rules, and it does not matter much whether these rules were applied to all threads or not.

libcontainer/configs/config.go
alban: Did you have a chance to look at libseccomp-golang?

tests/integration/seccomp.bats
alban: Updated with a loop to test all set of flags possible

libcontainer/seccomp/seccomp_linux.go
cyphar: There was an issue with setting it in patchbpf, I can't remember the exact details.

tests/integration/seccomp.bats
```diff
@@ -66,6 +66,37 @@ function teardown() {
...
+		update_config '   .process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"]
+				| .process.noNewPrivileges = false
+				| .linux.seccomp = {
+					"defaultAction":"SCMP_ACT_ALLOW",
+					"architectures":["SCMP_ARCH_X86","SCMP_ARCH_X32"],
+					"flags":['"${flags}"'],
+					"syscalls":[{"names":["mkdir"], "action":"SCMP_ACT_ERRNO"}]
+				}'
+
+		# This test checks that the log flag is accepted but does not check
```
kolyshkin, line 93: This sentence is not finished.

alban: Fixed

vnsavage:
It seems like all requested changes are implemented on this diff? Is there anything else that needs doing before this gets merged?

tests/integration/seccomp.bats
```diff
@@ -66,6 +66,38 @@ function teardown() {
...
+		'"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_TSYNC"'
+		'"SECCOMP_FILTER_FLAG_SPEC_ALLOW","SECCOMP_FILTER_FLAG_TSYNC"'
+		'"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_SPEC_ALLOW","SECCOMP_FILTER_FLAG_TSYNC"'
+	)
+	for flags in "${SECCOMP_FILTER_FLAGS[@]}"; do
+		update_config '   .process.args = ["/bin/sh", "-c", "mkdir /dev/shm/foo"]
+				| .process.noNewPrivileges = false
+				| .linux.seccomp = {
+					"defaultAction":"SCMP_ACT_ALLOW",
+					"architectures":["SCMP_ARCH_X86","SCMP_ARCH_X32"],
```
AkihiroSuda, line 88: Needs "SCMP_ARCH_AARCH64","SCMP_ARCH_ARM"

AkihiroSuda (approved):

alban:
It seems the last changes reliably break `TestUsernsCheckpoint`. Any ideas why?

kolyshkin:
The error is EINVAL from move_mount(2) syscall. It can be returned if the kernel is too old, but it seems that
 - Linux v5.15 is used, which is sufficient;
 - criu does the check for the `MOVE_MOUNT_SET_GROUP` feature and finds it working.

There are other cases when the move mount set group can return EINVAL, but it does not seem likely.

Ah! It might be caused by the fact that criu ppa comes with criu 3.17-1ppa2.20.04 which is old (should be 3.17.1). I just saw that @rst0git updated that package 3 hours ago. So, let me restart the CI.

kolyshkin:
TestUsernsCheckpoint also fails on CentOS Stream 9, although with a different diagnostics (from https://cirrus-ci.com/task/5426803694108672?logs=unit_tests#L1368):

```
    checkpoint_test.go:163: (00.041591) pie: 1: restoring lsm profile (current) unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
    checkpoint_test.go:163: (00.041654) Error (criu/cr-restore.c:1510): 65454 stopped by signal 11: Segmentation fault
    checkpoint_test.go:163: (00.042114) mnt: Switching to new ns to clean ghosts
    checkpoint_test.go:163: (00.042270) uns: calling exit_usernsd (-1, 1)
    checkpoint_test.go:163: (00.042312) uns: daemon calls 0x47fe40 (65452, -1, 1)
    checkpoint_test.go:163: (00.042319) uns: `- daemon exits w/ 0
    checkpoint_test.go:163: (00.042824) uns: daemon stopped
    checkpoint_test.go:163: (00.042829) Error (criu/cr-restore.c:2536): Restoring FAILED.
    checkpoint_test.go:163: === END ===
```

kolyshkin:
@alban you also need to fix a linter warning

libcontainer/specconv/spec_linux.go
```diff
@@ -1016,14 +1016,22 @@ func SetupSeccomp(config *specs.LinuxSeccomp) (*configs.Seccomp, error) {
...
+	// The list of flags defined in runtime-spec is a subset of the flags
+	// in the seccomp() syscall
+	for _, flag := range config.Flags {
+		switch flag {
+		case "SECCOMP_FILTER_FLAG_TSYNC":
+			// Tsync can be silently ignored
+		case specs.LinuxSeccompFlagLog, specs.LinuxSeccompFlagSpecAllow:
+			newConfig.Flags = append(newConfig.Flags, flag)
+		default:
+			return nil, fmt.Errorf("seccomp flags %q not yet supported by runc", flag)
```
kolyshkin, line 1031: nit: `s/flags/flag/`

alban:
* PR updated and rebased.
* The linter warning is fixed.
* The TestUsernsCheckpoint issue is resolved in github-actions but there is still an issue in cirrus-ci with TestUsernsCheckpoint and TestCheckpoint.

rst0git:
> TestUsernsCheckpoint also fails on CentOS Stream 9, although with a different diagnostics (from https://cirrus-ci.com/task/5426803694108672?logs=unit_tests#L1368):

We recently updated the criu package in CentOS Stream (https://github.com/opencontainers/runc/issues/3532)

kolyshkin (approved):
LGTM (CI failure is unrelated and is being addressed in https://github.com/opencontainers/runc/issues/3532)

AkihiroSuda (approved):

AkihiroSuda:
> Required statuses must pass before merging

Can't hit the merge button :(

rata:
Again `--- FAIL: TestUsernsCheckpoint (0.43s)`. I guess @alban needs to push again (or maybe rebase) so the tests are happy? Or is the problem on the CI not fixed?

rata:
@alban a fix was just merged for the test (https://github.com/opencontainers/runc/pull/3539). If you rebase and push, the tests should re-start and work this time :)

alban:
@rata updated!

kolyshkin (approved):
LGTM

AkihiroSuda (approved):

Issue #3391, opened 2022-02-22
not ok 42 runc delete: cgroup not cleaned up correctly


kolyshkin:
This is an interesting failure I haven't seen before. Happened in centos-stream-8 CI here: https://cirrus-ci.com/task/6376444287778816?logs=integration_systemd#L49

```
not ok 42 runc delete
# (from function `fail' in file tests/integration/helpers.bash, line 338,
#  in test file tests/integration/delete.bats, line 30)
#   `[ "$output" = "" ] || fail "cgroup not cleaned up correctly: $output"' failed
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-68012/runc.IPOypI/tty/sock testbusyboxdelete (status=0):
# 
# runc state testbusyboxdelete (status=0):
# {
#   "ociVersion": "1.0.2-dev",
#   "id": "testbusyboxdelete",
#   "pid": 71979,
#   "status": "running",
#   "bundle": "/tmp/bats-run-68012/runc.IPOypI/bundle",
#   "rootfs": "/tmp/bats-run-68012/runc.IPOypI/bundle/rootfs",
#   "created": "2022-02-19T00:12:45.413050609Z",
#   "owner": ""
# }
# runc kill testbusyboxdelete KILL (status=0):
# 
# runc delete testbusyboxdelete (status=0):
# 
# runc state testbusyboxdelete (status=1):
# time="2022-02-19T00:12:45Z" level=error msg="container does not exist"
# cgroup not cleaned up correctly: /sys/fs/cgroup/pids/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
# /sys/fs/cgroup/cpu,cpuacct/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
# /sys/fs/cgroup/blkio/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
# /sys/fs/cgroup/devices/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
# /sys/fs/cgroup/memory/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
[...]

PR #3392, opened 2022-02-22
tests/int: runc delete: fix flake, enable for rootless

kolyshkin:
The following failure was observed in CI (on centos-stream-8 in
integration-cgroup suite):

	not ok 42 runc delete
	 (from function `fail' in file tests/integration/helpers.bash, line 338,
	  in test file tests/integration/delete.bats, line 30)
	   `[ "$output" = "" ] || fail "cgroup not cleaned up correctly: $output"' failed
	....
	cgroup not cleaned up correctly: /sys/fs/cgroup/pids/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
	/sys/fs/cgroup/cpu,cpuacct/system.slice/tmp-bats\x2drun\x2d68012-runc.IPOypI-state-testbusyboxdelete-runc.zriC8C.mount
	...

Apparently, this is a cgroup systemd creates for a mount unit which
appears then runc does internal /proc/self/exe bind-mount. The test
case should not take it into account.

Fix the find arguments to look for a specific cgroup name, and add
a check that these arguments are correct (i.e. the cgroup is found
when the container is running).

This added correctness check reveals another problem -- for rootless,
the test is not able to create a cgroup, so the test case is not checking anything.
The obvious fix would be to add `requires rootless_cgroups`. It is not the right
fix, because for rootless + fs cgroup driver runc does not actually create a cgroup
(it is done by tests/rootless.sh). So, require systemd (which allows to create
user cgroups), and do not test the fs driver.

Fixes: https://github.com/opencontainers/runc/issues/3391

### DIFF
--- a/tests/integration/delete.bats
+++ b/tests/integration/delete.bats
@@ -11,10 +11,22 @@ function teardown() {
 }
 
 @test "runc delete" {
+	# Need a permission to create a cgroup.
+	# XXX(@kolyshkin): currently this test does not handle rootless when
+	# fs cgroup driver is used, because in this case cgroup (with a
+	# predefined name) is created by tests/rootless.sh, not by runc.
+	[[ "$ROOTLESS" -ne 0 ]] && requires systemd
+	set_resources_limit
+
 	runc run -d --console-socket "$CONSOLE_SOCKET" testbusyboxdelete
 	[ "$status" -eq 0 ]
 
 	testcontainer testbusyboxdelete running
+	# Ensure the find statement used later is correct.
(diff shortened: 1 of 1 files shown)


kolyshkin:
No longer a draft; PTAL @AkihiroSuda

kolyshkin:
@opencontainers/runc-maintainers an we please have this merged? This fixes a [rare] flake in CI, and enables `runc delete` for rootless.

AkihiroSuda (approved):

hqhq (approved):

Issue #3401, opened 2022-03-03
Support `LinuxDeviceCgroup` updates


agners:
To be able to adjust device permissions at runtime, this issue seeks to implement support for Linux Device CGroup updates. It is a prerequisite to get such support in Moby/Docker, see also moby/moby#42255.

Currently supplying `devices` to `runc update --resources -` does not adjust cgroup(v2) device permissions for an existing container. Some issues with device updates have been outlined in #2366 and addressed in various PRs linked to that issue.

To implement support updates would need to pass `devices` to the internal `config.Cgroups.Resources.Devices` structure. Besides [update also sets `config.Cgroups.SkipDevices` to false explicitly](https://github.com/opencontainers/runc/blob/v1.1.0/update.go#L337), which probably would need changing.

I guess the semantics of update would need to be defined exactly. A simplistic approach would be to implement a behavior similar to what `CreateCgroupConfig` implements: Clear the current list of devices, append the list provided by the update command, then append the defaultDevs.

In PR #2951 @cyphar [added a comment](https://github.com/opencontainers/runc/pull/2951/files#diff-8a7713337aa452775ee332800f9726c3fa5b98919692cd1b6b2da47f2caf660bR655-R659) saying:

```
	# Unfortunately we can't update device rules directly with runc ("runc
	# update" doesn't support it, and adding support would require ironing out
	# some long-standing design issues with device configuration). So instead
	# we just run "runc update" many times, relying on the fact that runc will
	# re-apply devices cgroup rules on each runc update.
```

What are those design issues exactly?

utam0k:
FYI: `containerd` recently added support for it.
https://github.com/containerd/cgroups/pull/225

agners:
@utam0k hm, that is the containerd API basically allowing to update using a Resource object correct? I guess it still requires the runc changes to make the device part actually work.

utam0k:
@agners As far as I know, `containerd` has its own device emulate. I guess it use own devices emulate to update resources. If I am wrong, please let me 🙏

agners:
@utam0k from what I understand `containerd` uses `runc` as low-level runtime by default. I don't think it has a built-in low level runtime, it requires runc to work (see https://github.com/containerd/containerd/blob/main/BUILDING.md#runc). It seems to support alternative runtimes such as `crun`.

utam0k:
@agners That's right. But `containerd` has its own cgroup controller. 
https://github.com/containerd/cgroups

PR #3402, opened 2022-03-03
Add device resource update support

agners:
Implement support for updating device resources to allow updating CGroup v1/v2 device permissions on container update.

Fixes: #3401


kolyshkin:
This calls for test cases.

libcontainer/devices/device.go
```diff
@@ -24,6 +24,9 @@ type Device struct {
 
 	// Gid of the device.
 	Gid uint32 `json:"gid"`
+
+	// Is Default Device
+	IsDefault bool `json:"is_default"`
```
cyphar, line 29: This struct is stored on disk and so, on upgrade, any existing containers will have a JSON config that has ever device be "non-default". It seems like this is only used during specconv, so maybe this should be a hidden JSON field rather than being stored on-disk?

(As an aside, we really should have tests to make sure we never break this again -- Docker 17.06.1 was a complete nightmare. But in this particular case, nothing would break but the behaviour would still not be ideal.)

agners: > It seems that this flag is only used for an internal check (and the on-disk flag is never checked)

Afaik, the on-disk flag is used when updating. `CreateDefaultDevicesCgroups` is now called from `update.go` (see `updateCommand`).

From what I remember this was necessary to "clear" allowed devices. Otherwise, the user would have to pass the default devices to get back to default state.

> This struct is stored on disk and so, on upgrade, any existing containers will have a JSON config that has ever device be "non-default".

Hm, this is a problem, can we migrate the data?

agners: I've reviewed this again, the reason it is required is because default allowed devices get special treatment for via `CreateCgroupConfig`. To do this, we need to understand which devices have been added through spec, and which have been added from  `AllowedDevices`. This used to be done in a single function `createDevices`. With commit https://github.com/opencontainers/runc/pull/3402/commits/390ac2b10b5d1b46479d1bb1ca64a99d72cbd972 de-duplicates this logic so it can be used for updates.

There is no need to store this flag in JSON, so I think we can use ``json:"-"`` which we use in other places too.

agners: It turns out, the `is_default` needs to be stored in the container state to successfully "re-duplicate" on update.

agners:
Any thoughts or comments for this PR?

kolyshkin:
> Any thoughts or comments for this PR?

Can you please squash in the intermediate changes  so that it can be reviewed commit by commit. You had three commits originally, now it's 11.

agners:
I've rebased and suqashed the commits.  There are essentially the original three commits plus one for the integration tests.

agners:
Any thoughts on this PR? It is still a feature we are looking for.

agners:
@kolyshkin is there a chance to get this feature moving forward?

It is kinda road block for our users to adopt CGroupsV2 :see_no_evil: In CGroupsV1 we were able to update the device permission directly (which was hacky, granted, but it worked). However, this approach is much more sane and works for CGroupsV2 too, so we really like to have such support upstream.

atostivint:
Is there anything missing on this commit to be pushed? It's still a blocker for HomeAssistant Supervised https://github.com/home-assistant/supervised-installer/issues/372

PR #3404, opened 2022-03-08
[1.1] libct/cg/sd/v2: fix ENOENT on cgroup delegation

kolyshkin:
_Backport of #3389 to 1.1. Original description follows._

----

Apparently, not all files listed in `/sys/kernel/cgroup/delegate` must
exist in every cgroup, so we should ignore `ENOENT`.

Dot not ignore `ENOENT` on the directory itself though.

Change cgroupFilesToChown to not return `.`, and refactor it to not do
any dynamic slice appending in case we're using the default built-in
list of files.

Fixes: 35d20c4e0

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -2,6 +2,7 @@ package systemd
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"math"
 	"os"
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

kolyshkin:
Hmm, apparently the "update with rebase" is working, with all LGTMs left intact, so I have enabled this option for all PRs.

thaJeztah (approved):
LGTM, thanks!

PR #3419, opened 2022-03-15
Add / switch to Go 1.18

kolyshkin:
- switch to golangci-lint 1.45 (which is go1.18 compatible);
 - switch to Go 1.18 as the default Go version in CI;
 - remove 1.16 compatibility, and remove 1.16 from CI matrix.


kolyshkin:
Ah! golangci-lint is not working with go 1.18 yet (see https://github.com/golangci/golangci-lint/pull/2438).

kolyshkin:
`test (1.17.x, criu-dev)` is not shown in checks since it became `test (1.18.x, criu-dev)`.

.github/workflows/test.yml
```diff
@@ -21,13 +21,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.16.x, 1.17.x, 1.18.0-beta1]
+        go-version: [1.16.x, 1.17.x, 1.18.x]
```
AkihiroSuda, line 24: We can drop 1.16

AkihiroSuda (dismissed):

.github/workflows/validate.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v2
         with:
-          go-version: "${{ env.GO_VERSION }}"
+          go-version: 1.17.x # TODO switch to "${{ env.GO_VERSION }}"
```
thaJeztah, line 21: Looks like go 1.18 may be supported now (although it looks like they will skip some linters, based on some tickets linked to that one); not sure if we're running all of those though; https://github.com/golangci/golangci-lint/pull/2438

kolyshkin: Let me try again )

thaJeztah: Yeah, no promises; just noticed the linked ticket was closed, so perhaps we could have everything done in one go 👍 🤞

.github/workflows/test.yml
kolyshkin: Good; modified accordingly. The change to drop 1.16 is somewhat big so implemented in a separate commit.

kolyshkin:
> Looking better; one deprecation warning about `strings.Title()`;
> 
> ```
>   Error: SA1019: strings.Title is deprecated: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (staticcheck)
> ```
> 
> My recommendation would be to add a `nolint` for that one, as it's perfectly fine to use for ASCII(ish) characters; their recommendation is ridiculous, and brings in 20K lines of code (see [moby/moby#43201](https://github.com/moby/moby/pull/43201))

I ended up with own implementation of `title()` which is just enough for the purposes.

If a function is deprecated, it might be removed later, so I'd rather not use it.

thaJeztah:
> If a function is deprecated, it might be removed later, so I'd rather not use it.

It's in stdlib, so won't be removed until Go 2.0 at least, but it's silly they marked it as deprecated. The discussion that lead to that was because there was a "note" in the function's description that there were some corner cases where it didn't work, then someone suggested; let's then deprecate it!

Instead, they could've just documented its limitations. The same problem applies to most of stdlib (no full support for all unicode corner cases).

kolyshkin:
OK, Fedora 35 comes with Go 1.16 so we need to install golang if we drop Go 1.16 support.

dqminh (approved):

libcontainer/container_linux.go
```diff
@@ -919,8 +919,19 @@ func (c *linuxContainer) criuSupportsExtNS(t configs.NamespaceType) bool {
...
+		return s
+	}
+	s0 -= 'a' - 'A' // convert to uppercase
+
+	return string(s0) + s[1:]
+}
+
 func criuNsToKey(t configs.NamespaceType) string {
-	return "extRoot" + strings.Title(configs.NsName(t)) + "NS"
+	return "extRoot" + title(configs.NsName(t)) + "NS"
```
AkihiroSuda, line 934: Is this parsed by something?
Do we really need to titlize it?

AkihiroSuda (approved):

libcontainer/container_linux.go
```diff
@@ -919,8 +919,19 @@ func (c *linuxContainer) criuSupportsExtNS(t configs.NamespaceType) bool {
 	return c.checkCriuVersion(minVersion) == nil
 }
 
+// title converts the first character of ASCII string s to uppercase.
+func title(s string) string {
+	s0 := s[0]
```
thaJeztah, line 924: This will panic if string would ever be empty; is that something that could possibly happen? (also could we get non-ascii (multi-byte) characters, in which case we'd be doing the wrong thing?)

libcontainer/container_linux.go
kolyshkin: Quoting commit message

> The alternative is to not capitalize the namespace string; this will break
> restoring of a container checkpointed by earlier version of runc.

libcontainer/container_linux.go
kolyshkin: If the first character is non-ASCI (e.g. хлеб), it will be returned as-is (no capitalization).

Indeed it will panic with empty string as an argument, and it is theoretically possible to have it.

I guess I'll just go back to using strings.Title and suppressing the deprecation warning.

libcontainer/container_linux.go
kolyshkin: I moved back to using `strings.Title` instead, suppressing the deprecation warning as it seems easier.

kolyshkin:
I have switched back to using `strings.Title` (suppressing the deprecation warning as it was suggested by @thaJeztah earlier). Not other changes, please re-LGTM @thaJeztah @AkihiroSuda @dqminh.

I will merge this myself since this involves changing the branch protection rules for `main` (remove `test (1.16.x, ...)`, add `test(1.18.x, ...)`, change `test (1.17.x, criu-dev)` to 1.18.x).

libcontainer/container_linux.go
thaJeztah: Ah, yeah, perhaps best to do that.

I'm considering proposing to un-deprecate this function (I think it's more disruptive than useful)

^^ feel free to beat me to that though if you agree 😁

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

kolyshkin:
Branch protection rules updated. This also means all older PRs need to be re-tested...

PR #3420, opened 2022-03-17
libct/configs/validate: rootlessEUIDMount: speedup

kolyshkin:
1. Fix function docs. In particular, remove the part
   which is not true ("verifies that the user isn't trying to set up any
   mounts they don't have the rights to do"), and fix the part that
   says "that doesn't resolve to root" (which is no longer true since
   commit d8b669400adc).

2. Replace fmt.Sscanf (which is slow and does lots of allocations)
   with strings.TrimPrefix and strconv.Atoi.

3. Add a benchmark for rootlessEUIDMount. Comparing the old and the new
   implementations:
```
	name                 old time/op    new time/op    delta
	RootlessEUIDMount-4    1.01µs ± 2%    0.16µs ± 1%  -84.15%  (p=0.008 n=5+5)

	name                 old alloc/op   new alloc/op   delta
	RootlessEUIDMount-4      224B ± 0%       80B ± 0%  -64.29%  (p=0.008 n=5+5)

	name                 old allocs/op  new allocs/op  delta
	RootlessEUIDMount-4      7.00 ± 0%      1.00 ± 0%  -85.71%  (p=0.008 n=5+5)
```
Note this code is already tested (in rootless_test.go).

Fixes: d8b669400adc

### DIFF
--- a/libcontainer/configs/validate/rootless.go
+++ b/libcontainer/configs/validate/rootless.go
@@ -2,7 +2,7 @@ package validate
 
 import (
 	"errors"
-	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

libcontainer/configs/validate/rootless.go
```diff
@@ -51,9 +51,8 @@ func rootlessEUIDMappings(config *configs.Config) error {
 	return nil
 }
 
-// mount verifies that the user isn't trying to set up any mounts they don't have
-// the rights to do. In addition, it makes sure that no mount has a `uid=` or
-// `gid=` option that doesn't resolve to root.
+// rootlessEUIDMount verifies that all mounts have valid uid=/gid= options,
```
kolyshkin, line 54: This is slightly incorrect. Should have been something like

```suggestion
// rootlessEUIDMount verifies that all mounts' uid=/gid= options, if present, are valid,
```

PR #3421, opened 2022-03-18
libct/cg: add disable_cgroup_devices build tag

kolyshkin:
The idea is to use this tag for software which wants to use
`libcontainer/cgroups` packages, but does not require the functionality
to manage devices in cgroups, which brings some additional overhead,
especially in `fs2` where it exports `cilium/ebpf`. Using this  build tag
enables that software to save about **586 Kb** (text+data+bss), which
reduces runc binary size by about **6%** (using go 1.18.0):

           text    data     bss     dec     hex filename
        5800416 3745184  229768 9775368  952908 runc.dev
        5440984 3505336  228904 9175224  8c00b8 runc.no-dev

(The difference is more dramatic then using older versions of golang, e.g. 1.17.x)

Obviously, we do not want runc to be compiled with this tag, so add
a guard to `main.go` and libcontainer itself.

NOTE than the tag is used, the packages still create a directory under
devices cgroup root, and it is used by Exists/GetPids/GetAllPids,
but otherwise nothing is done with devices. The effect is similar to
that of `Resources.SkipDevices` and `Resources.SkipFreezeOnSet`
(for systemd cgroup v1) both set to `true`, except it is a compile-time thing.

Document the tag in README, and add a unit tests of libct/cg with this tag set.


AkihiroSuda (changes requested):
I don't think we should complicate tags.

Can we just move [`fs2.setDevices()`](https://github.com/opencontainers/runc/blob/c258ed0fc3cfdeaf696272cd222578c9248a5944/libcontainer/cgroups/fs2/devices.go#L56) to `factory_linux.go` so as to decouple `fs2` pkg from the `ebpf` pkg?

kolyshkin:
> Can we just move [`fs2.setDevices()`](https://github.com/opencontainers/runc/blob/c258ed0fc3cfdeaf696272cd222578c9248a5944/libcontainer/cgroups/fs2/devices.go#L56) to `factory_linux.go` so as to decouple `fs2` pkg from the `ebpf` pkg?

I looked at it, and it seems this will require more changes. Currently devices are an integral part of cgroup manager, and while I can see how it is decoupled, it seems it's going to look ugly.

kolyshkin:
> Can we just move [`fs2.setDevices()`](https://github.com/opencontainers/runc/blob/c258ed0fc3cfdeaf696272cd222578c9248a5944/libcontainer/cgroups/fs2/devices.go#L56) to `factory_linux.go` so as to decouple `fs2` pkg from the `ebpf` pkg?

I took a look, and it seems more complicated than that. Devices is the integral part of fs manager, and they can definitely be separated out, but that looks like a medium-sized refactoring, which I don't quite know how to do yet.

In the meantime I updated this PR to be more reviewable -- first commit merely moves the code around (no code or functionality changes), and the second adds the tags, tests, and docs.

cdoern:
@kolyshkin this seems to be the preferable option in podman with the tag as is

kolyshkin:
> Can we just move [`fs2.setDevices()`](https://github.com/opencontainers/runc/blob/c258ed0fc3cfdeaf696272cd222578c9248a5944/libcontainer/cgroups/fs2/devices.go#L56) to `factory_linux.go` so as to decouple `fs2` pkg from the `ebpf` pkg?

So, I took another look at this and it seems this is more complicated than that. Doing this properly requires removing devices management from the cgroup managers (basically, creating a separate "devices" manager which is now part of cgroups.

Ultimately it _might_ be the best thing to do (for one thing, we can remove all these ugly `SkipDevices` and `SkipFreezeOnSet` flags), but it is nevertheless a big change, whereas adding a build tag (this PR) seems not overly complicated.

I don't know. Should I try to separate devices out of cgroup manager instead of what this PR does, @opencontainers/runc-maintainers ?

kolyshkin:
> I don't know. Should I try to separate devices out of cgroup manager instead of what this PR does, @opencontainers/runc-maintainers ?

Implemented the separation in #3452, PTAL

PR #3332, opened 2022-01-05
CI/cirrus: add centos-stream-9

kolyshkin:
Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

PR #3426, opened 2022-03-22
CI/cirrus: add centos-stream-9

kolyshkin:
TODO


kolyshkin:
Apparently creating a new PR from a branch that was previously deleted (#3332) results in CI not being kicked.

Oh my.

PR #3429, opened 2022-03-23
Support Id mapped mounts for shared volumes

AlexeyPerevalov:
This pull request adds support for id map mount feature for shared volumes.
For rootfs this is already implemented in https://github.com/containerd/containerd/pull/5890.

Both commit has a code in common, which should be placed on the same place in the future.

There are two possibilities to use mount_setattr syscall with fd returned by open_tree or with fd returned by fsmount syscall. This PR uses open_tree, since this approach was used in examples, as well as in crun.

As a fallback scenario, mount syscall still used for bind mount. Open_tree syscall as well as mount_setattr are quite new syscall, to use it latest kernel version is required. In this patch set, kernel version is checked, but as an alternative approach we could just check syscall existence, any suggestions are welcome.
Also this PR includes specs-go which could be moved to separate PR.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	securejoin "github.com/cyphar/filepath-securejoin"
@@ -18,6 +20,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups/fs2"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/opencontainers/runc/libcontainer/devices"
+	"github.com/opencontainers/runc/libcontainer/user"
 	"github.com/opencontainers/runc/libcontainer/userns"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -437,9 +440,21 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 		if err := prepareBindMount(m, rootfs, mountFd); err != nil {
 			return err
 		}
-		if err := mountPropagate(m, rootfs, mountLabel, mountFd); err != nil {
+
+		treeFd, err := unix.OpenTree(-int(unix.EBADF), m.Source, uint(unix.OPEN_TREE_CLONE|unix.OPEN_TREE_CLOEXEC|unix.AT_EMPTY_PATH|unix.AT_RECURSIVE))
+		if err != nil || treeFd < 0 {
 			return err
 		}
+
+		err = unix.MoveMount(treeFd, "", unix.AT_FDCWD, dest, unix.MOVE_MOUNT_F_EMPTY_PATH)
+		if err != nil {
+			return err
+		}
+
+		if err := mountIDMapMapped(m, treeFd); err != nil {
+			return err
+		}
+
 		// bind mount won't change mount options, we need remount to make mount options effective.
 		// first check that we have non-default options required before attempting a remount
 		if m.Flags&^(unix.MS_REC|unix.MS_REMOUNT|unix.MS_BIND) != 0 {
@@ -472,7 +487,7 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 		}
 		return mountPropagate(m, rootfs, mountLabel, mountFd)
 	}
-	if err := setRecAttr(m, rootfs); err != nil {
+	if err := setMountAttr(m, rootfs); err != nil {
 		return err
 	}
 	return nil
@@ -1044,7 +1059,7 @@ func remount(m *configs.Mount, rootfs string, mountFd *int) error {
 
 	return utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
 		flags := uintptr(m.Flags | unix.MS_REMOUNT)
-		err := mount(source, m.Destination, procfd, m.Device, flags, "")
+		err := mount(source, m.Destination, procfd, m.Device, flags, m.Data)
 		if err == nil {
 			return nil
 		}
@@ -1058,7 +1073,8 @@ func remount(m *configs.Mount, rootfs string, mountFd *int) error {
 		}
 		// ... and retry the mount with ro flag set.
 		flags |= unix.MS_RDONLY
-		return mount(source, m.Destination, procfd, m.Device, flags, "")
+
+		return mount(source, m.Destination, procfd, m.Device, flags, m.Data)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -24,3 +25,5 @@ require (
 	golang.org/x/sys v0.0.0-20211116061358-0a5406a5449c
 	google.golang.org/protobuf v1.27.1
 )
+
+replace github.com/opencontainers/runtime-spec/specs-go => ./staging/src/github.com/opencontainers/runtime-spec/specs-go
```
AkihiroSuda, line 29: Please open PR for runtime-spec

libcontainer/init_linux.go
```diff
@@ -97,6 +97,7 @@ func newContainerInit(t initType, pipe *os.File, consoleSocket *os.File, fifoFd,
 			logFd:         logFd,
 		}, nil
 	case initStandard:
+		logrus.Infof("mountFds: %v", mountFds)
```
AkihiroSuda, line 100:
```suggestion
		logrus.Debugf("mountFds: %v", mountFds)
```

go.mod
AlexeyPerevalov: ok

go.mod
```diff
@@ -12,8 +12,9 @@ require (
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/moby/sys/mountinfo v0.6.0
 	github.com/mrunalp/fileutils v0.5.0
-	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
+	github.com/opencontainers/runtime-spec/specs-go v0.0.0
 	github.com/opencontainers/selinux v1.10.0
+	github.com/pkg/errors v0.9.1
```
AkihiroSuda, line 17: Deprecated

go.mod
AlexeyPerevalov: @AkihiroSuda https://github.com/opencontainers/runtime-spec/pull/1143

libcontainer/mount_linux.go
```diff
@@ -3,9 +3,19 @@ package libcontainer
...
+	"unsafe"
+
 	"golang.org/x/sys/unix"
 )
 
+const (
+	OPEN_TREE_CLONE         = 1
+	OPEN_TREE_CLOEXEC       = 02000000
+	MOVE_MOUNT_F_EMPTY_PATH = 0x00000004
+	AT_FDCWD                = -100
```
kolyshkin, line 16: Should these be defined in x/sys/unix? Please open a CL against it -- they are usually more than happy to accept such stuff.

libcontainer/rootfs_linux.go
```diff
@@ -68,6 +74,7 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig, mountFds []int) (err
 		cgroupns:        config.Namespaces.Contains(configs.NEWCGROUP),
 	}
 	setupDev := needsSetupDev(config)
+	//time.Sleep(30 * time.Second)
```
kolyshkin, line 77: Debug session leftover?

libcontainer/rootfs_linux.go
```diff
@@ -380,6 +387,9 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 		return err
 	}
 
+	if mountFd != nil {
+		logrus.Infof("mountFd: %d", *mountFd)
```
kolyshkin, line 437: either `logrus.Debugf` or remove it.

libcontainer/rootfs_linux.go
```diff
@@ -1041,7 +1063,9 @@ func remount(m *configs.Mount, rootfs string, mountFd *int) error {
 
 	return utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
 		flags := uintptr(m.Flags | unix.MS_REMOUNT)
-		err := mount(source, m.Destination, procfd, m.Device, flags, "")
+		logrus.Infof("m.Data: %v\n", m.Data)
+		logrus.Infof("flags: %v\n", flags)
```
kolyshkin, line 1119: No need for `\n` here. Also seems like a debug leftover.

kolyshkin:
OK, this should be a draft / DNM until changes to runtime-spec and x/sys/unix are landed.

libcontainer/mount_linux.go
AlexeyPerevalov: yeah, I wrote about it, in cover message, but definitely x/sys/unix is better candidate.

libcontainer/rootfs_linux.go
AlexeyPerevalov: yes, initial version, after moving helpers to x/sys/unix and runtime-spec, it would be clear.

libcontainer/mount_linux.go
AlexeyPerevalov: @kolyshkin I've seen your https://go-review.googlesource.com/c/sys/+/363444 I assume this change was made w/o code generation e.g. by sys/unix/mkall.sh ?

AlexeyPerevalov:
This PR depends on golang's x/sys library modifications:
https://go-review.googlesource.com/c/sys/+/397095
https://go-review.googlesource.com/c/sys/+/397094
and depends on https://github.com/opencontainers/runtime-spec/pull/1143.
Patches for x/sys are merged, just need to wait for appropriate version release of lib.
runtime-spec change is stuck on review, how to initiate it?

libcontainer/rootfs_linux.go
```diff
@@ -326,6 +330,53 @@ func mountCgroupV2(m *configs.Mount, c *mountConfig) error {
...
 }
 
+var (
+	syscallOnce sync.Once
+	syscallErr  error
+)
+
+func mountByMove(m *configs.Mount, dest string) error {
+	// Check once for all mount point that syscall exists
+	syscallOnce.Do(func() {
```
jiangliu, line 340: Could this be exported as an interface so other parts may query the syscall availability status? We face the same issue too and seems it could be reused.

rata:
@AlexeyPerevalov the runtime spec PR was already merged (https://github.com/opencontainers/runtime-spec/pull/1143). It would be great if you can finish this PR now :)

AlexeyPerevalov:
runtime-spec changes are not yet tagged, latest tag is v1.0.2
the same with x/sys golang lib, not yet published

rata:
@AlexeyPerevalov No need to wait for a runtime-spec tag, IMHO. We are not using a tag already: https://github.com/opencontainers/runc/blob/a776ec9c90d7a67a924211d6cb1b38620afa6f49/go.mod#L15

For the x/sys pkg, I think it makes sense to define ourselves the constants if that is not going to be released soon. Do you happen to know when they plan to release that with the constants we need? If there is no visibility into that, I don't know what others think, but I would guess it seems reasonable to use some tmp file with the constants defined here. We can switch whenever that becomes available.

kolyshkin:
> runtime-spec changes are not yet tagged, latest tag is v1.0.2

We are already using a sha for it (due to the lack of runtime-spec maintainers I guess).

> the same with x/sys golang lib, not yet published

This one is never tagged, i.e. all users are just pin it to a particular sha.

AlexeyPerevalov:
in centos-stream-9 two tests are failed:
TestUsernsCheckpoint
TestCheckpoint 

Tests related to checkpoint restore failed due to segfault in criu, I reproduced it on Centos Stream 8 (I don't have Stream 9), but It is not reproducible on Ubuntu 20.04, with working criu.

libcontainer/rootfs_linux.go
```diff
@@ -437,9 +488,13 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 		if err := prepareBindMount(m, rootfs, mountFd); err != nil {
 			return err
 		}
-		if err := mountPropagate(m, rootfs, mountLabel, mountFd); err != nil {
-			return err
+
+		if mountByMove(m, dest) != nil {
+			if err := mountPropagate(m, rootfs, mountLabel, mountFd); err != nil {
+				return err
```
rata, line 494: This means, if we fail to mount with idmap mounts, it just does a bind-mount?

I'm not sure what is the right behavior, because just mounting and ignoring the mapping for that mount will probably create issues to access the files (if the container running inside a userns).

I'm not sure what the right behavior should be. What do you think? And other maintainers?

rata:
@AlexeyPerevalov but it seems all test have passed... Was that maybe before the latest fix merged yesterday to fix the CI? (although IIRC it was only about one of those test, not both)

rata:
@AlexeyPerevalov friendly ping? It would be great to have this in for runc 1.2.

(otherwise, we will have to wait for another runc release, that takes a while, some more time for stable versions of high level container runtimes to include it, etc. :))

libcontainer/rootfs_linux.go
rata: Thinking a little bit how higher layers will use this, yes, I think we need to error out. At the k8s layer we want to start the container with an id mapped mount or not start it at all. If it is started without the id mapped mounts, the files created will have just random garbage as UID/GID, won't be able to read files... it will be a complete mess.

I think if we just return an error if an id mapped mount was requested and we failed to create it is the right path here.

Also, even if in the future we want to open this door (doubt about it), we can always open a door. But we won't be able to close it if we open it now.

@AlexeyPerevalov Can we make this error out if the id map mount fails?

libcontainer/rootfs_linux.go
```diff
@@ -437,9 +488,13 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 		if err := prepareBindMount(m, rootfs, mountFd); err != nil {
 			return err
 		}
-		if err := mountPropagate(m, rootfs, mountLabel, mountFd); err != nil {
-			return err
+
+		if mountByMove(m, dest) != nil {
```
rata, line 492: I'm testing this, and it seems this can't work. At this point we are inside the userns (I've added some `logrus.Debugf` to print the content of `/proc/self/uid_map` to verify this) and therefore we create an idmap mount. They only work from the initial userns.

Did you made this runc code work for you at some point? Or was this a c&p from working code in a standalone binary (that you tried from the host, so the initial userns) that never was tested in the runc binary (that IIUC this part runs inside the userns).

Am I missing something? Or do we need a big rework to make this work?

rata (commented):
@AlexeyPerevalov Hey, I was testing this and can't make it work.

It seems the code to create an idmap mount is fine (except probably the move mount should be last), but as this code is run already inside the userns, we always get EPERM for the mountSetAttr() syscall, because that can only be run from the initial userns to create a IDMAP mount.

Have you tested this with runc and it worked for you? It would be great if I'm missing something and there is a simple way to make this work :)

rata:
I really wish to include this functionality in the next runc release. This is required for our userns work in Kubernetes. I'll try to make this work.

If anyone has any thoughts on the approach to take, please let me know :)

AkihiroSuda:
> I really wish to include this functionality in the next runc release. This is required for our userns work in Kubernetes. I'll try to make this work.
> 
> If anyone has any thoughts on the approach to take, please let me know :)

Thanks @rata , could you open a new PR to carry this PR?

rata:
Opened this draft PR with an approach that works: https://github.com/opencontainers/runc/pull/3717

PR #3432, opened 2022-03-25
Move the cgroups v2 hybrid check from init()

mrunalp:
Other code that imports the libcontainer cgroups library
may be vulnerable to panic in the function due to insufficient
permissions.

### DIFF
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -32,14 +32,6 @@ var subsystems = []subsystem{
 
 var errSubsystemDoesNotExist = errors.New("cgroup: subsystem does not exist")
 
-func init() {
-	// If using cgroups-hybrid mode then add a "" controller indicating
-	// it should join the cgroups v2.
-	if cgroups.IsCgroup2HybridMode() {
-		subsystems = append(subsystems, &NameGroup{GroupName: "", Join: true})
-	}
-}
-
 type subsystem interface {
 	// Name returns the name of the subsystem.
 	Name() string
@@ -67,6 +59,12 @@ func NewManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Manager, e
 	}
 	if cg.Resources.Unified != nil {
 		return nil, cgroups.ErrV1NoUnified
+	}
+
+	// If using cgroups-hybrid mode then add a "" controller indicating
+	// it should join the cgroups v2.
+	if cgroups.IsCgroup2HybridMode() {
+		subsystems = append(subsystems, &NameGroup{GroupName: "", Join: true})
 	}
 
 	if paths == nil {
--- a/libcontainer/cgroups/fs/fs.go
+++ b/libcontainer/cgroups/fs/fs.go
@@ -13,23 +13,6 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
-var subsystems = []subsystem{
-	&CpusetGroup{},
-	&DevicesGroup{},
-	&MemoryGroup{},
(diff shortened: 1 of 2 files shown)


mrunalp:
We could also save the error from the function and propagate it up. `IsCgroup2UnifiedMode` also has a similar panic but isn't called from any init functions.

mrunalp:
@kolyshkin @thaJeztah @AkihiroSuda ptal

liggitt:
is a panic on a permissions error on that path really the correct response for all callers? returning a bool, error and letting the caller decide if it should be a fatal error would have been what I expected.

mrunalp:
> is a panic on a permissions error on that path really the correct response for all callers? returning a bool, error and letting the caller decide if it should be a fatal error would have been what I expected.

Yes, that's what I posed in my [comment](https://github.com/opencontainers/runc/pull/3432#issuecomment-1079485234) above to the other maintainers. This particular function is a relative easy fix with 2 callers. We also have the existing `IsCgroup2UnifiedMode` with many more callers. I think the risk of hitting this is very low based on the permissions of these directories unless there is some misconfiguration of cgroups setup on a distro. 

Our options are:
1. We could get this in and it will be a similar risk to what's vendored right now with the existing calls of `IsCgroup2UnifiedMode`.
2. Fix up the callers of `IsCgroup2HybridMode` only in this PR and fix the remaining later. 
3. Fix both of them. Maybe refactor so we detect the failure in a different call early and all callers don't have to be modified for `IsCgroup2UnifiedMode`. This could lead to API changes, so needs more time and thought.

AkihiroSuda (dismissed):

libcontainer/cgroups/fs/fs.go
```diff
@@ -69,6 +61,12 @@ func NewManager(cg *configs.Cgroup, paths map[string]string) (cgroups.Manager, e
 		return nil, cgroups.ErrV1NoUnified
 	}
 
+	// If using cgroups-hybrid mode then add a "" controller indicating
+	// it should join the cgroups v2.
+	if cgroups.IsCgroup2HybridMode() {
+		subsystems = append(subsystems, &NameGroup{GroupName: "", Join: true})
```
liggitt, line 67: This is mutating a package var in a constructor. Can that race with other readers of subsystems? If the constructor is called multiple times, this will append multiple times, right?

mrunalp: Ugh, yeah, from a quick look, we could move the subsystems under the manager instance.

mrunalp: Pushed a commit to test that out (will check later).

libcontainer/cgroups/fs/fs.go
```diff
@@ -46,9 +29,10 @@ type subsystem interface {
 }
 
 type manager struct {
-	mu      sync.Mutex
-	cgroups *configs.Cgroup
-	paths   map[string]string
+	mu         sync.Mutex
+	cgroups    *configs.Cgroup
+	paths      map[string]string
+	subsystems []subsystem
```
kolyshkin, line 35: It does not make sense to have a per-manager subsystem list.

Instead, I think, we should keep subsystems global, and wrap their initialization into `once.Do`.

kolyshkin:
I took a look at it seems it's better to just avoid panicking in `IsCgroup2HybridMode`. This is what  https://github.com/opencontainers/runc/pull/3433 does; PTAL.

mrunalp:
Closing in favor of #3433

mrunalp:
Yeah that’s what I was debating, too. I think your getting rid of the panic is probably best for now.

PR #3433, opened 2022-03-27
libct/cg: IsCgroup2HybridMode: don't panic

kolyshkin:
In case `statfs("/sys/fs/cgroup/unified")` fails with any error other
than `ENOENT`, current code panics. As `IsCgroup2HybridMode` is called from
`libcontainer/cgroups/fs`'s `init` function, this means that any user of
libcontainer may panic during initialization, which is ugly.

Avoid panicking; instead, do not enable hybrid hierarchy support and
report the error (under debug level, not to confuse anyone).

Basically, replace the panic with "turn off hybrid mode support"
(which makes total sense since we were unable to statfs its root).

This is an alternative to #3432. The bug is originally reported by @liggitt
in https://github.com/kubernetes/kubernetes/pull/109029#pullrequestreview-922146797

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -55,12 +55,12 @@ func IsCgroup2HybridMode() bool {
 		var st unix.Statfs_t
 		err := unix.Statfs(hybridMountpoint, &st)
 		if err != nil {
-			if os.IsNotExist(err) {
-				// ignore the "not found" error
-				isHybrid = false
-				return
+			isHybrid = false
+			if !os.IsNotExist(err) {
+				// Report unexpected errors.
+				logrus.WithError(err).Debugf("statfs(%q) failed", hybridMountpoint)
 			}
-			panic(fmt.Sprintf("cannot statfs cgroup root: %s", err))
+			return
 		}
 		isHybrid = st.Type == unix.CGROUP2_SUPER_MAGIC
 	})


kolyshkin:
@AkihiroSuda @mrunalp @thaJeztah PTAL

mrunalp:
As a library we should avoid logging but I think this case is probably fine.

mrunalp (dismissed):

libcontainer/cgroups/utils.go
```diff
@@ -55,12 +55,11 @@ func IsCgroup2HybridMode() bool {
...
-			if os.IsNotExist(err) {
-				// ignore the "not found" error
-				isHybrid = false
-				return
+			isHybrid = false
+			if !os.IsNotExist(err) {
+				// Report unexpected errors.
+				logrus.WithError(err).Debugf("statfs(%q) failed", hybridMountpoint)
 			}
-			panic(fmt.Sprintf("cannot statfs cgroup root: %s", err))
```
liggitt, line 63: still need to `return` rather than continue to `isHybrid = st.Type == unix.CGROUP2_SUPER_MAGIC` in the error case, right?

kolyshkin: Right. Updated; thanks for catching this!

I am also on the verge of removing `isHybrid = false` line since it does nothing (since isHybrid is false by default), but I guess it is better to leave it as is it for the sake of readability.

kolyshkin:
> As a library we should avoid logging but I think this case is probably fine.

Good point! There are a bunch of other places like this in libcontainer unfortunately. Which has to be solved globally (perhaps by having a user-specifiable logger which defaults to "no logging at all".

mrunalp (approved):

AkihiroSuda (approved):

kolyshkin:
1.1 backport: #3435

PR #3435, opened 2022-03-28
[release-1.1] libct/cg: IsCgroup2HybridMode: don't panic

mrunalp:
In case statfs("/sys/fs/cgroup/unified") fails with any error other
than ENOENT, current code panics. As IsCgroup2HybridMode is called from
libcontainer/cgroups/fs's init function, this means that any user of
libcontainer may panic during initialization, which is ugly.

Avoid panicking; instead, do not enable hybrid hierarchy support and
report the error (under debug level, not to confuse anyone).

Basically, replace the panic with "turn off hybrid mode support"
(which makes total sense since we were unable to statfs its root).

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -55,12 +55,12 @@ func IsCgroup2HybridMode() bool {
 		var st unix.Statfs_t
 		err := unix.Statfs(hybridMountpoint, &st)
 		if err != nil {
-			if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


mrunalp:
@AkihiroSuda @thaJeztah @kolyshkin I picked this into 1.1 so we could tag a release there to vendor into k8s.

cc: @liggitt

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

Issue #3437, opened 2022-03-28
Release 1.1.1


kolyshkin:
As of now, I think we have a critical mass to bake a release in 1.1.x branch.

Here's a list of PRs:
https://github.com/opencontainers/runc/pulls?q=is%3Apr+label%3Abackport%2F1.1-pr+

In particular, https://github.com/opencontainers/runc/pull/3435 is important, as it blocks bumping runc to 1.1 in kubernetes (see https://github.com/kubernetes/kubernetes/pull/109029).

kolyshkin:
Ah! We need release notes; let me prepare it.

kolyshkin:
OK, please proofread the relnotes here: https://github.com/opencontainers/runc/pull/3439

AkihiroSuda:
:tada: https://github.com/opencontainers/runc/releases/tag/v1.1.1

Foxboron:
Yo,

It seems like 1.1.1 is missing a signed tag, and the commit is not signed either. This is a change from previous releases. Is this intentional and should we only utilize the signed tarballs?

kolyshkin:
> It seems like 1.1.1 is missing a signed tag, and the commit is not signed either. This is a change from previous releases. Is this intentional and should we only utilize the signed tarballs?

My bad for not signing the commit; it was not intentional.

The tag is actually signed (by me), but it was not shown as signed because I did not upload my GPG key to github (which is now fixed).

Foxboron:
Ah, the UX for this isn't great and I was going off on the notes the other Arch package maintainer got me :)

We pushed runc with the signed tarball this release and there isn't any good reason for us to use the git remote directly I think. I thought it would just be a good idea to mention it.

cyphar:
I auto-sign all of my commits so when I do a release they'll always be signed. But yeah, part of our release process is generating signed tarballs and those are probably a better source to use (that's what I use in openSUSE for packaging as well).

PR #3406, opened 2022-03-08
[release-1.1] configs/validate: looser validation for RDT

marquiz:
_Backport of https://github.com/opencontainers/runc/pull/3382._

----

Don't require CAT or MBA because we don't detect those correctly (we
don't support L2 or L3DATA/L3CODE for example, and in the future
possibly even more). With plain "ClosId mode" we don't really care: we
assign the container to a pre-configured CLOS without trying to do
anything smarter.

Moreover, this was a duplicate/redundant check anyway, as for CAT and
MBA there is another specific sanity check that is done if L3 or MB
is specified in the config.

Signed-off-by: Markus Lehtonen <markus.lehtonen@intel.com>
(cherry picked from commit 1d5c331042acbce433e48a2f3a277f8e08fe998c)

PR #3439, opened 2022-03-28
Release runc 1.1.1

kolyshkin:
Changes (quoting CHANGELOG.md):

## [1.1.1] - 2022-03-28

> Violence is the last refuge of the incompetent.

### Added
 * CI is now also run on centos-stream-9. (#3436)

### Fixed
 * `runc run/start` can now run a container with read-only `/dev` in OCI spec,
   rather than error out. (#3355)
 * `runc exec` now ensures that `--cgroup` argument is a sub-cgroup. (#3403)
 * libcontainer systemd v2 manager no longer errors out if one of the files
   listed in `/sys/kernel/cgroup/delegate` do not exist in container's cgroup.
   (#3387, #3404)
 * Loose OCI spec validation to avoid bogus "Intel RDT is not supported" error.
   (#3406)
 * libcontainer/cgroups no longer panics in cgroup v1 managers if `stat`
   of `/sys/fs/cgroup/unified` returns an error other than ENOENT. (#3435)

Fixes: #3437


mrunalp (approved):

AkihiroSuda (approved):

CHANGELOG.md
```diff
@@ -233,7 +252,8 @@ implementation (libcontainer) is *not* covered by this policy.
    cgroups at all during `runc update`). (#2994)
 
 <!-- minor releases -->
-[Unreleased]: https://github.com/opencontainers/runc/compare/v1.1.0...HEAD
+[Unreleased]: https://github.com/opencontainers/runc/compare/v1.1.1...HEAD
+[1.1.1]: https://github.com/opencontainers/runc/compare/v1.1.0...v1.1.1
```
cyphar, line 256: NIT: I would've put this in a "`1.1.z` patch release" section (and split out the "unreleased 1.1.z" list) but that doesn't really matter. I'll fix it next time I touch the changelog.

cyphar (approved):
LGTM

CHANGELOG.md
cyphar: Also I usually put the changelog bump in the same commit as the version bump, but also this doesn't matter too much.

PR #3447, opened 2022-04-03
Add support for CRIU's 'ignore' manage-cgroup mode

adrianreber:
This adds an already existing CRIU cgroup mode to be selectable from runc: 'ignore'.

This is a fix for errors seen with Podman containers that have been restored multiple times into new cgroups. CRIU fails to checkpoint a container that has already been restored once in a cgroup named differently than the original container.

CRIU keeps information about the original cgroup and during restore in a differently named cgroup CRIU still tries to restore to original cgroup. This leads to a process with wrong cgroup settings and a second checkpoint of that container fails.

With 'ignore' CRIU will ignore any cgroups and leave the cgroup setting to runc.

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -405,3 +405,40 @@ function simple_cr() {
 	# busybox should be back up and running
 	testcontainer test_busybox running
 }
+
+@test "checkpoint and restore into a different cgroup" {
+	local pid name=test-diff-cgroup
+
+	set_cgroups_path
+	runc run -d --console-socket "$CONSOLE_SOCKET" --pid-file pid "$name"
+	[ "$status" -eq 0 ]
+
+	testcontainer "$name" running
+	# Check the cgroup.
+	pid="$(cat pid)"
+	original_cgroup="$OCI_CGROUPS_PATH"
+	echo "OCI_CGROUPS_PATH=$OCI_CGROUPS_PATH"
+	cat "/proc/$pid/cgroup"
+
+	runc checkpoint --manage-cgroups-mode ignore --work-path ./work-dir "$name"
+	grep -B 5 Error ./work-dir/dump.log || true
+	[ "$status" -eq 0 ]
+
+	testcontainer "$name" checkpointed
+
+	# Modify the cgroup path.
+	set_cgroups_path
+	runc restore --manage-cgroups-mode ignore --pid-file pid -d \
+		--work-path ./work-dir --console-socket "$CONSOLE_SOCKET" "$name"
+	grep -B 5 Error ./work-dir/restore.log || true
+	[ "$status" -eq 0 ]
+
+	testcontainer "$name" running
+
+	# Check the cgroup name has changed.
+	pid="$(cat pid)"
+	echo "OCI_CGROUPS_PATH=$OCI_CGROUPS_PATH"
+	cat "/proc/$pid/cgroup"
+	new_cgroup="$OCI_CGROUPS_PATH"
+	[ "$original_cgroup" != "$new_cgroup" ]
+}
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -38,7 +38,7 @@ checkpointed.`,
 		cli.StringFlag{Name: "page-server", Value: "", Usage: "ADDRESS:PORT of the page server"},
 		cli.BoolFlag{Name: "file-locks", Usage: "handle file locks, for safety"},
 		cli.BoolFlag{Name: "pre-dump", Usage: "dump container's memory information only, leave the container running after this"},
-		cli.StringFlag{Name: "manage-cgroups-mode", Value: "", Usage: "cgroups mode: 'soft' (default), 'full' and 'strict'"},
+		cli.StringFlag{Name: "manage-cgroups-mode", Value: "", Usage: "cgroups mode: 'soft' (default), 'ignore', 'full' and 'strict'"},
 		cli.StringSliceFlag{Name: "empty-ns", Usage: "create a namespace, but don't restore its properties"},
 		cli.BoolFlag{Name: "auto-dedup", Usage: "enable auto deduplication of memory images"},
 	},
@@ -147,6 +147,8 @@ func setManageCgroupsMode(context *cli.Context, options *libcontainer.CriuOpts) 
 			options.ManageCgroupsMode = criu.CriuCgMode_FULL
 		case "strict":
 			options.ManageCgroupsMode = criu.CriuCgMode_STRICT
+		case "ignore":
+			options.ManageCgroupsMode = criu.CriuCgMode_IGNORE
 		default:
 			return errors.New("Invalid manage cgroups mode")
 		}
--- a/man/runc-checkpoint.8.md
+++ b/man/runc-checkpoint.8.md
@@ -57,7 +57,7 @@ together with **criu lazy-pages**. See
 : Do a pre-dump, i.e. dump container's memory information only, leaving the
 container running. See [criu iterative migration](https://criu.org/Iterative_migration).
 
-**--manage-cgroups-mode** **soft**|**full**|**strict**.
+**--manage-cgroups-mode** **soft**|**ignore**|**full**|**strict**.
 : Cgroups mode. Default is **soft**. See
 [criu --manage-cgroups option](https://criu.org/CLI/opt/--manage-cgroups).
 
--- a/man/runc-restore.8.md
+++ b/man/runc-restore.8.md
@@ -37,7 +37,7 @@ image files directory.
 : Allow checkpoint/restore of file locks. See
 [criu --file-locks option](https://criu.org/CLI/opt/--file-locks).
 
-**--manage-cgroups-mode** **soft**|**full**|**strict**.
+**--manage-cgroups-mode** **soft**|**ignore**|**full**|**strict**.
 : Cgroups mode. Default is **soft**. See
 [criu --manage-cgroups option](https://criu.org/CLI/opt/--manage-cgroups).
 
--- a/restore.go
+++ b/restore.go
@@ -53,7 +53,7 @@ using the runc checkpoint command.`,
 		cli.StringFlag{
 			Name:  "manage-cgroups-mode",
 			Value: "",
-			Usage: "cgroups mode: 'soft' (default), 'full' and 'strict'",
+			Usage: "cgroups mode: 'soft' (default), 'ignore', 'full' and 'strict'",
 		},
 		cli.StringFlag{
 			Name:  "bundle, b",
@@ -114,6 +114,9 @@ using the runc checkpoint command.`,
 			return err
 		}
 		if err := setEmptyNsMask(context, options); err != nil {
+			return err
+		}
+		if err := setManageCgroupsMode(context, options); err != nil {
 			return err
 		}
 		status, err := startContainer(context, CT_ACT_RESTORE, options)


kolyshkin:
> > 1. Do you think it will help the likes of [[centos7] not ok 15 checkpoint --lazy-pages and restore #2760](https://github.com/opencontainers/runc/issues/2760) and [[fedora] not ok 16 checkpoint --lazy-pages and restore #2924](https://github.com/opencontainers/runc/issues/2924)?
> 
> No, I don't think this is related.

I am asking because (as I said in https://github.com/opencontainers/runc/issues/2924#issuecomment-865465526) I suspect that criu restores the container into the same cgroup, while systemd removes this cgroup (as the container was gone).

With `--manage-cgroup-mode ignore` CRIU will do nothing to restore cgroups, and runc will correctly create a proper cgroup (and set all the parameters), so the above issues should be gone.

That is just my understanding, I might be wrong.

adrianreber:
> 1. Do you think it will help the likes of [[centos7] not ok 15 checkpoint --lazy-pages and restore #2760](https://github.com/opencontainers/runc/issues/2760) and [[fedora] not ok 16 checkpoint --lazy-pages and restore #2924](https://github.com/opencontainers/runc/issues/2924)?

No, I don't think this is related.

>     2. Do we need a test case for this?

Not sure. Do you have any ideas how to create a container with a cgroup and check that it has another after restore. I have seen a couple of cgroup functions in the bats files, but if you have a quick idea what would work here, we can add it.

> > > 1. Do you think it will help the likes of [[centos7] not ok 15 checkpoint --lazy-pages and restore #2760](https://github.com/opencontainers/runc/issues/2760) and [[fedora] not ok 16 checkpoint --lazy-pages and restore #2924](https://github.com/opencontainers/runc/issues/2924)?
> > 
> > 
> > No, I don't think this is related.
> 
> I am asking because (as I said in [#2924 (comment)](https://github.com/opencontainers/runc/issues/2924#issuecomment-865465526)) I suspect that criu restores the container into the same cgroup, while systemd removes this cgroup (as the container was gone).

Ah, right. Then this could indeed help.

> With `--manage-cgroup-mode ignore` CRIU will do nothing to restore cgroups, and runc will correctly create a proper cgroup (and set all the parameters), so the above issues should be gone.

There is also this discussion where the kernel might have a problem with correct cgroup deletion: https://github.com/checkpoint-restore/criu/issues/1774#issuecomment-1086875628

But with your explanation it sounds like a correctly working `ignore` could help you.

kolyshkin (changes requested):
@adrianreber I think that `setManageCgroupsMode` is missing from `restore.go`, and so the option is ignored upon restore. Can you add it here?

adrianreber:
> @adrianreber I think that `setManageCgroupsMode` call is missing from `restore.go`, and so the option is ignored upon restore. Can you add it here?

Good point, seems like the parameter never had any effect on restore until now. Updated.

kolyshkin:
> Do you have any ideas how to create a container with a cgroup and check that it has another after restore. I have seen a couple of cgroup functions in the bats files, but if you have a quick idea what would work here, we can add it.

Well, let's do it, since without such tests we do not know if the fix really works or not

Changing the cgroup path in container config is pretty simple -- you call `set_cgroups_path` and it randomizes it (well, just the suffix).

Here is a draft for the test (you can append this to tests/integration/checkpoint.bats):

```bash
@test "checkpoint and restore into a different cgroup" {
        local pid name=test-diff-cgroup

        set_cgroups_path
        runc run -d --console-socket "$CONSOLE_SOCKET" --pid-file pid "$name"
        [ "$status" -eq 0 ]

        testcontainer "$name" running
        # Check the cgroup.
        pid="$(cat pid)"
        echo "OCI_CGROUPS_PATH=$OCI_CGROUPS_PATH"
        cat "/proc/$pid/cgroup"

        runc checkpoint --manage-cgroups-mode ignore --work-path ./work-dir "$name"
        grep -B 5 Error ./work-dir/dump.log || true
        [ "$status" -eq 0 ]

        testcontainer "$name" checkpointed

        # Modify the cgroup path.
        set_cgroups_path
        runc restore --manage-cgroups-mode ignore --pid-file pid -d \
                --work-path ./work-dir --console-socket "$CONSOLE_SOCKET" "$name"
        grep -B 5 Error ./work-dir/restore.log || true
        [ "$status" -eq 0 ]

        testcontainer "$name" running

        # Check the cgroup name has changed.
        pid="$(cat pid)"
        echo "OCI_CGROUPS_PATH=$OCI_CGROUPS_PATH"
        cat "/proc/$pid/cgroup"
        exit 1
}
```

This test deliberately fails so you can see all the debug output.

Here is how to run it, and here's what it shows on my machine:

```console
$ make
...
$ sudo bats -f different tests/integration/checkpoint.bats 
 ✗ checkpoint and restore into a different cgroup
   (in test file tests/integration/checkpoint.bats, line 441)
     `exit 1' failed
   runc spec (status=0):
   
   runc run -d --console-socket /tmp/bats-run-OlQVIP/runc.urXhFf/tty/sock --pid-file pid test-diff-cgroup (status=0):
   
   runc state test-diff-cgroup (status=0):
   {
     "ociVersion": "1.0.2-dev",
     "id": "test-diff-cgroup",
     "pid": 181707,
     "status": "running",
     "bundle": "/tmp/bats-run-OlQVIP/runc.urXhFf/bundle",
     "rootfs": "/tmp/bats-run-OlQVIP/runc.urXhFf/bundle/rootfs",
     "created": "2022-05-12T02:31:20.689876979Z",
     "owner": ""
   }
   OCI_CGROUPS_PATH=/runc-cgroups-integration-test/test-cgroup-1873
   0::/runc-cgroups-integration-test/test-cgroup-1873
   runc checkpoint --manage-cgroups-mode ignore --work-path ./work-dir test-diff-cgroup (status=0):
   
   runc state test-diff-cgroup (status=1):
   time="2022-05-11T19:31:20-07:00" level=error msg="container does not exist"
   runc restore --manage-cgroups-mode ignore --pid-file pid -d --work-path ./work-dir --console-socket /tmp/bats-run-OlQVIP/runc.urXhFf/tty/sock test-diff-cgroup (status=0):
   
   runc state test-diff-cgroup (status=0):
   {
     "ociVersion": "1.0.2-dev",
     "id": "test-diff-cgroup",
     "pid": 181780,
     "status": "running",
     "bundle": "/tmp/bats-run-OlQVIP/runc.urXhFf/bundle",
     "rootfs": "/tmp/bats-run-OlQVIP/runc.urXhFf/bundle/rootfs",
     "created": "2022-05-12T02:31:21.053776806Z",
     "owner": ""
   }
   OCI_CGROUPS_PATH=/runc-cgroups-integration-test/test-cgroup-21203
   0::/runc-cgroups-integration-test/test-cgroup-1873
   rmdir: failed to remove '/sys/fs/cgroup//runc-cgroups-integration-test': Device or resource busy

1 test, 1 failure
```

So, from what I see it restores the container into the old cgroup (which ends in `1873`, while the config is changed before the restore to have cgroupPath which ends in  `21203`).

This is probably happening because I am using criu 3.16.1:
```console
[kir@kir-rhat runc]$ rpm -qf $(which criu)
criu-3.16.1-2.fc35.x86_64
```

Let's try with 3.17 (I am still on Fedora 35 so have to compile it).

Well, this is the same with 3.17. I guess some more changes are needed for runc, so it can put the container into proper cgroup itself (not via criu). I have some code somewhere that did it...will take a look next week.

kolyshkin:
@adrianreber my old code is at https://github.com/kolyshkin/runc/commits/restore-cgroup (last 4 commits), maybe you will find something that's missing from this PR. In particular, the changes in `libcontainer/container_linux.go` are interesting.

adrianreber:
@kolyshkin Thanks for the test example. I used it and it works correctly for me. Let's see what CI thinks about it.

kolyshkin:
fedora CI failure seems to be related, can you PTAL @adrianreber ?

adrianreber:
> fedora CI failure seems to be related, can you PTAL @adrianreber ?

I see the following error in CI:
```
+ disable_cgroup
+ '[' -d /sys/fs/cgroup//runc-cgroups-integration-test ']'
+ rmdir /sys/fs/cgroup//runc-cgroups-integration-test
rmdir: failed to remove '/sys/fs/cgroup//runc-cgroups-integration-test': Device or resource busy
make: *** [Makefile:131: localrootlessintegration] Error 1
Connection to 192.168.121.237 closed.
```

With my added test the rootless test seems to fail. Not sure why. I do not see similar problems locally. Strange.

kolyshkin:
@adrianreber the test still fails, and it seems there's a reason for that.

I wrote a small test (you can append this to the end of checkpoint.bats file):
```bats
@test "my" {
        set_cgroups_path
        [ "$status" -eq 0 ]
        testcontainer test_busybox running
        echo "0 CGROUP_PATH=$CGROUP_PATH"
        ls -ld "$CGROUP_PATH" || true

        runc checkpoint --work-path ./work-dir --manage-cgroups-mode ignore test_busybox
        grep -B 5 Error ./work-dir/dump.log || true
        [ "$status" -eq 0 ]
        # after checkpoint busybox is no longer running
        testcontainer test_busybox checkpointed

        echo "1 CGROUP_PATH=$CGROUP_PATH"
        ls -ld "$CGROUP_PATH" || true
        orig_cgroup_path=$CGROUP_PATH

        # restore from checkpoint
        set_cgroups_path
        runc "$@" restore -d --work-path ./work-dir --manage-cgroups-mode ignore --console-socket "$CONSOLE_SOCKET" test_busybox
        grep -B 5 Error ./work-dir/restore.log || true
        [ "$status" -eq 0 ]
        # busybox should be back up and running
        testcontainer test_busybox running

        echo "2 CGROUP_PATH=$CGROUP_PATH"
        ls -ld "$CGROUP_PATH" || true

        echo "3 orig_cgroup_path=$orig_cgroup_path"
        ls -ld "$orig_cgroup_path" || true
        exit 1 # this is for the test to always fails and show logs
}
```

Next, make sure everything is compiled:
```console
$ make all
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=v1.1.0-254-g0e8b1e7e-dirty -X main.version=1.1.0+dev " -o runc .
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=v1.1.0-254-g0e8b1e7e-dirty -X main.version=1.1.0+dev " -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=v1.1.0-254-g0e8b1e7e-dirty -X main.version=1.1.0+dev " -o contrib/cmd/sd-helper/sd-helper ./contrib/cmd/sd-helper
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=v1.1.0-254-g0e8b1e7e-dirty -X main.version=1.1.0+dev " -o contrib/cmd/seccompagent/seccompagent ./contrib/cmd/seccompagent
```

Finally, run the above test:
```console
$  sudo bats -f "my" tests/integration/checkpoint.bats
checkpoint.bats
 ✗ my
   (in test file tests/integration/checkpoint.bats, line 481)
     `exit 1' failed
   runc spec (status=0):
   
   runc run -d --console-socket /tmp/bats-run-vahX7f/runc.3kYwVb/tty/sock test_busybox (status=0):
   
   runc state test_busybox (status=0):
   {
     "ociVersion": "1.0.2-dev",
     "id": "test_busybox",
     "pid": 1194856,
     "status": "running",
     "bundle": "/tmp/bats-run-vahX7f/runc.3kYwVb/bundle",
     "rootfs": "/tmp/bats-run-vahX7f/runc.3kYwVb/bundle/rootfs",
     "created": "2022-07-30T01:38:56.151243561Z",
     "owner": ""
   }
   0 CGROUP_PATH=/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-17881
   drwxr-xr-x. 2 root root 0 Jul 29 18:38 /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-17881
   runc checkpoint --work-path ./work-dir --manage-cgroups-mode ignore test_busybox (status=0):
   
   runc state test_busybox (status=1):
   time="2022-07-29T18:38:56-07:00" level=error msg="container does not exist"
   1 CGROUP_PATH=/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-17881
   ls: cannot access '/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-17881': No such file or directory
   runc restore -d --work-path ./work-dir --manage-cgroups-mode ignore --console-socket /tmp/bats-run-vahX7f/runc.3kYwVb/tty/sock test_busybox (status=0):
   
   runc state test_busybox (status=0):
   {
     "ociVersion": "1.0.2-dev",
     "id": "test_busybox",
     "pid": 1194928,
     "status": "running",
     "bundle": "/tmp/bats-run-vahX7f/runc.3kYwVb/bundle",
     "rootfs": "/tmp/bats-run-vahX7f/runc.3kYwVb/bundle/rootfs",
     "created": "2022-07-30T01:38:56.551376328Z",
     "owner": ""
   }
   2 CGROUP_PATH=/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-9654
   drwxr-xr-x. 2 root root 0 Jul 29 18:38 /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-9654
   3 orig_cgroup_path=/sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-17881
   drwxr-xr-x. 2 root root 0 Jul 29 18:38 /sys/fs/cgroup/runc-cgroups-integration-test/test-cgroup-17881
   rmdir: failed to remove '/sys/fs/cgroup//runc-cgroups-integration-test': Device or resource busy

1 test, 1 failure

```

As you can see, the old/original cgroup path (the one that ends with `17881`) is non-existent after checkpoint, but it reappears upon restore.

I am not sure why, but at least you have easy repro.

tests/integration/checkpoint.bats
```diff
@@ -405,3 +405,40 @@ function simple_cr() {
...
+	[ "$status" -eq 0 ]
+
+	testcontainer "$name" running
+
+	# Check the cgroup name has changed.
+	pid="$(cat pid)"
+	echo "OCI_CGROUPS_PATH=$OCI_CGROUPS_PATH"
+	cat "/proc/$pid/cgroup"
+	new_cgroup="$OCI_CGROUPS_PATH"
+	[ "$original_cgroup" != "$new_cgroup" ]
```
kolyshkin, line 443: Ughm, so here you basically check that `set_cgroups_path` changed the value of `OCI_CGROUPS_PATH`.

Well yes it did.

What should have done is checking that
 - the new cgroup path exists
 - the old cgroup path does not exist
 - the contents of `/proc/$pid/cgroup` differs from the old one

kolyshkin (changes requested):

kolyshkin:
Here's a high level overview of what is going on in a test from the prev. comment:

1. We set cgroups path and create a container. Container is running, cgroup path exists.
2. We checkpoint the container with `--manage-cgroups-mode ignore`. Container is stopped, cgroup path does not exist.
3. We change the cgroupsPath in container's config.json.
4. We restore the container.

Apparently, what happens is
 - runc creates new cgroup as per config.json
 - the container is restored to the old cgroup :-\

adrianreber:
@kolyshkin thanks for looking so closely at this. I probably will not have time to look into this for the next couple of weeks, but I will try to see why it fails then.

kolyshkin:
@adrianreber I have finally figured it out in #3546.

PR #3449, opened 2022-04-05
Use `fmt.Sprintf` less

kolyshkin:
Two more cases where removing `fmt.Sprintf` makes things faster.

### DIFF
--- a/libcontainer/container_linux.go
+++ b/libcontainer/container_linux.go
@@ -2100,7 +2100,10 @@ func (c *Container) orderNamespacePaths(namespaces map[configs.NamespaceType]str
 func encodeIDMapping(idMap []configs.IDMap) ([]byte, error) {
 	data := bytes.NewBuffer(nil)
 	for _, im := range idMap {
-		line := fmt.Sprintf("%d %d %d\n", im.ContainerID, im.HostID, im.Size)
+		line := strconv.Itoa(im.ContainerID) + " " +
+			strconv.Itoa(im.HostID) + " " +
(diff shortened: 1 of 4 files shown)


AkihiroSuda (changes requested):
Sorry, I don't see benefit

cyphar (commented):
Is `Sprintf` really that slow, and is it really a significant bottleneck (looking at the implementation, it does use reflection so it might be slow but I'm not sure how much of a bottleneck it is)? `Sprintf` is far more readable than the alternative, and so unless it's causing an actual performance issue, I don't really want to have to switch to open-coding it everywhere.

kolyshkin:
> Is `Sprintf` really that slow, and is it really a significant bottleneck (looking at the implementation, it does use reflection so it might be slow but I'm not sure how much of a bottleneck it is)? `Sprintf` is far more readable than the alternative, and so unless it's causing an actual performance issue, I don't really want to have to switch to open-coding it everywhere.

For some reason it is (maybe because of reflection used for arguments). For example, `CgroupString` becomes 4x faster when `Sprintf` is replaced with string concatenation.

As for readability, I guess that Sprintf is more readable to those of us who used to program in C; in other words, we got used to matching these %-whatever with their respective arguments. I suspect that to an eye untrained with printf style, concatenation looks simpler.

```go
return fmt.Sprintf("%c %s:%s %s", d.Type, major, minor, d.Permissions)
return string(d.Type) + " " + major + ":" + minor + " " + string(d.Permissions)
```

Yet I agree overall that this should rather be fixed in golang.

Issue #3417, opened 2022-03-15
Why Runc do bind mount config.rootfs to config.rootfs while it create container ?


Marshalzxy:
In rootfs_linux.go prepareRoot(), Runc bind mount config.Rootfs to config.Rootfs . Why this step is necessary?   In my centos 8.2, while graphdriver is overlayfs2, it seem that this step can be skipped without any consequences.Can we optimize this step?

func prepareRoot(config *configs.Config) error {
	flag := unix.MS_SLAVE | unix.MS_REC
	if config.RootPropagation != 0 {
		flag = config.RootPropagation
	}
	if err := unix.Mount("", "/", "", uintptr(flag), ""); err != nil {
		return err
	}

	// Make parent mount private to make sure following bind mount does
	// not propagate in other namespaces. Also it will help with kernel
	// check pass in pivot_root. (IS_SHARED(new_mnt->mnt_parent))
	if err := rootfsParentMountPrivate(config.Rootfs); err != nil {
		return err
	}

	**return unix.Mount(config.Rootfs, config.Rootfs, "bind", unix.MS_BIND|unix.MS_REC, "")**
}

kolyshkin:
I am not quite sure but _maybe_ this mount is not needed if `config.Rootfs` is already a mount point.

If the above is true, we can make it optional. Problem is, detection of whether the path is a mount point is a costly operation itself, for kernels that lack openat2(2) support (that is, Linux < 5.6) detecting a bind mount requires parsing /proc/mountinfo.

What is your use case, and what is the benefit from making this mount optional?

kolyshkin:
> Problem is, detection of whether the path is a mount point is a costly operation itself, for kernels that lack openat2(2) support (that is, Linux < 5.6) detecting a bind mount requires parsing /proc/mountinfo.

OTOH we already parse mountinfo anyway in `rootfsParentMountPrivate`...

kolyshkin:
The following patch can prevent the bind mount if it's already a mount point:
```diff
diff --git a/libcontainer/rootfs_linux.go b/libcontainer/rootfs_linux.go
index 5290a45e..c123547e 100644
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -738,15 +738,18 @@ func getParentMount(rootfs string) (string, string, error) {
 	return mi[idx].Mountpoint, mi[idx].Optional, nil
 }
 
-// Make parent mount private if it was shared
[...]

PR #3450, opened 2022-04-05
libct: prepareRoot: avoid extra bind mount

kolyshkin:
We need rootfs to be a mount point. To that effect, we bind mount it to
itself. In case the rootfs itself is a mount point, this is not needed.

Add the code to determine whether rootfs is a mount point (since we
already read mountinfo it does not incur any additional overhead),
and skip the bind mount in case rootfs is already a mount point.

Fixes: #3417

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -738,14 +738,17 @@ func getParentMount(rootfs string) (string, string, error) {
 	return mi[idx].Mountpoint, mi[idx].Optional, nil
 }
 
-// Make parent mount private if it was shared
-func rootfsParentMountPrivate(rootfs string) error {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
This seems to need careful security review.

>  In case the rootfs itself is a mount point, this is not needed.

What happens on race?

kolyshkin:
> This seems to need careful security review.
> 
> > In case the rootfs itself is a mount point, this is not needed.
> 
> What happens on race?

Do you mean a race between two runc instances, or a race between e.g. `runc run` and someone unmounting the rootfs?

In any case, as I said in #3417, a single bind mount is probably not worth optimising out, since the resulting code is more difficult to follow, and the benefit is not very clear.

PR #3451, opened 2022-04-07
Allow mounting of /proc/sys/kernel/ns_last_pid

dsouzai:
The `CAP_CHECKPOINT_RESTORE` linux capability provides the ability to update `/proc/sys/kernel/ns_last_pid`. However, because this "file" is under `/proc`, and by default both K8s and CRI-O specify that `/proc/sys` should be mounted as Read-Only, even with the capability specified, a process will not be able to write to `ns_last_pid`.

To get around this, a pod author can specify a volume mount and a host path to bind-mount `/proc/sys/kernel/ns_last_pid`. However, `runc` does not allow specifying mounts under `/proc`.

This PR adds `/proc/sys/kernel/ns_last_pid` to the `validProcMounts` string array to enable a pod author to mount `ns_last_pid` as read-write. The default remains unchanged; unless explicitly requested as a volume mount, `ns_last_pid` will remain read-only regardless of whether or not `CAP_CHECKPOINT_RESTORE` is specified.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -554,6 +554,7 @@ func checkProcMount(rootfs, dest, source string) error {
 		"/proc/loadavg",
 		"/proc/slabinfo",
 		"/proc/net/dev",
+		"/proc/sys/kernel/ns_last_pid",
 	}
 	for _, valid := range validProcMounts {
 		path, err := filepath.Rel(filepath.Join(rootfs, valid), dest)
--- a/libcontainer/rootfs_linux_test.go
+++ b/libcontainer/rootfs_linux_test.go
@@ -38,6 +38,14 @@ func TestCheckMountDestFalsePositive(t *testing.T) {
 	}
 }
 
+func TestCheckMountDestNsLastPid(t *testing.T) {
+	dest := "/rootfs/proc/sys/kernel/ns_last_pid"
+	err := checkProcMount("/rootfs", dest, "/proc")
+	if err != nil {
+		t.Fatal("/proc/sys/kernel/ns_last_pid should not return an error")
+	}
+}
+
 func TestNeedsSetupDev(t *testing.T) {
 	config := &configs.Config{
 		Mounts: []*configs.Mount{


dsouzai:
fyi @mrunalp @haircommander

kolyshkin:
@dsouzai looks like your idea is to use criu from inside the container. Do you think it is feasible to do so?

adrianreber:
@kolyshkin see https://github.com/cri-o/cri-o/pull/5776 for a discussion about possible use cases.

dsouzai:
> Do you think it is feasible to do so?

@kolyshkin Yeah, we have a prototype where we
1. Build CRIU with the change to [allow rootless checkpointing](https://github.com/checkpoint-restore/criu/pull/1155) as well as [other changes](https://github.com/ymanton/criu/commits/test-rootless).
2. Build the JVM that links the criu lib, and which [exposes an API](https://github.com/eclipse-openj9/openj9/pull/12969) to allow java applications to self dump. 
3. Build a container in which the criu binary has the following set to it:
```
setcap cap_checkpoint_restore,cap_net_admin,cap_sys_ptrace=eip /usr/sbin/criu
```
4. Start said container, in which the java application self checkpoints and the process ends. The container is then committed to create a restore image.
5. Start new container based on the restore image wherein a script launches `criu restore`.

In K8s, I'm able to successfully run the restore container with the runc change in this PR and the following pod spec:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: liberty-criu
  namespace: criu
spec:
  replicas: 1
  selector:
    matchLabels:
      name: liberty-criu
  template:
    metadata:
      labels:
        name: liberty-criu
    spec:
      serviceAccount: criusvcacct
      serviceAccountName: criusvcacct
      containers:
        - name: liberty-criu
          image: localhost/liberty-criu:latest
          imagePullPolicy: Always
          volumeMounts:
          - mountPath: /proc/sys/kernel/ns_last_pid
            name: ns-last-pid-mount
          securityContext:
            capabilities:
              add: [ "CHECKPOINT_RESTORE", "NET_ADMIN", "SYS_PTRACE" ]
      volumes:
      - name: ns-last-pid-mount
        hostPath:
          path: /proc/sys/kernel/ns_last_pid
          type: File
```

kolyshkin (approved):
LGTM

kolyshkin:

kolyshkin:
@AkihiroSuda PTAL

rst0git (approved):
LGTM

thaJeztah (approved):
LGTM

cyphar:
Someone sent a mail on oci-dev asking when this will be in a release -- given that the patch is very simple and a bugfix, maybe we should backport it to release-1.1?

PR #3452, opened 2022-04-08
Decouple setting cgroup device rules from cgroup manager

kolyshkin:
This is an alternative to https://github.com/opencontainers/runc/pull/3421, which tries to remove the functionality of setting cgroup device access rules from the cgroup managers.

### Motivation

This should help runc/libct/cg users who are
1. Worried that adding libct/cg as a dependency increases the binary size considerably (because of cilium/ebpf dependency).
2. Not interested in managing device rules for cgroups (such as kubelet, which only uses libct/cg to create parent/pod cgroups). Previously we have added two workarounds: `SkipDevices` ~~and `SkipFreezeOnSet`~~ for such users, and with this code `SkipDevices` can be omitted entirely in case the sources do not include `github.com/opencontainers/runc/libcontainer/cgroups/devices` package.

### Implementation

This
 * Moves the functionality to set cgroup device rules to a separate package, `libct/cg/devices`. Those are `setV1` and `setV2`. Same for generating systemd cgroup device properties (`systemdProperties`).
* If the above package is included, it autoconfigures cgroup managers to manage devices; if not, they will return an error if any devices rules are specified. So, a user can choose to have a cgroup manager with or without device management functionality (by adding or not adding an import statement).
* 
As a result:
* By default, cgroup managers are not able to set any device rules (and will error out if any devices rules are set in config)
* If device rules management is needed, one has to do something like
```diff
+import _ "github.com/opencontainers/runc/libcontainer/cgroups/devices"
```

### Caveats

* It seems impossible to get rid of `SkipFreezeOnSet` for systemd v1 cgroup manager since it can manage an existing systemd unit which has some device rules set, and any update of such unit results in rules being reloaded, so we have to freeze it before set.


kolyshkin:
@AkihiroSuda PTAL 🙏🏻 (if you like this more or less than #3421). It is not finished, but it is more or less clear where this is going.

Personally I think this is better than the build tag, because we can use the same library with and without device management (i.e. a runtime switch is better than the compile time)

kolyshkin:
> This is an alternative to #3421, which tries to remove the functionality of setting cgroup device access rules from the cgroup managers.
> 
> ~~Currently a semi-finished draft, ~~Aimed for review to try to access which way is better -- decoupling (this PR) or a build tag (#3421).
> 
> ### Motivation
> This should help runc/libct/cg users who are
> 
> 1. Not interested in managing device rules for cgroups (such as kubelet, which only uses libct/cg to create parent/pod cgroups). Previously we have added two workarounds: `SkipDevices` ~and `SkipFreezeOnSet`~ for such users, and with this code those workarounds can be safely removed.
> 2. Worried that adding libct/cg as a dependency increases the binary size considerably (because of cilium/ebpf dependency)
> 
> ### Implementation
> This
> 
> * Moves the functionality to set cgroup device rules to a separate package, `libct/cg/devices`. Those are `setV1` and `setV2`. Same for generating systemd cgroup device properties (`systemdProperties`).
> * Keeps a shallow devices cgroup implementation in all cgroup managers -- they are now only creating/removing the cgroup, not setting any device access rules.
> * Amends all cgroup device managers with a `setDev` field, which, if set, points to a function from devices pkg (`setV1` or `setV2`, described above).
> * Similarly, systemd managers are also amended with `devProps` field, which, if set, points to a function from devices pkg (`systemdProperties`, described above).
> 
> As a result:
> 
> * Using a cgroup manager per se results in ignoring all the device rules
> * If device rules management is needed, one has to do something like
> 
> ```diff
> +import "github.com/opencontainers/runc/libcontainer/cgroups/devices"
> ...
> cm, err := manager.New(config.Cgroups)
> if err != nil {
>    return err
> }
> +devices.Enable(cm)
> ```
> 
> * if the above is not done, but device rules are specified to `Set`, it errors out.
> 
> ### Caveats
> * Previously, v1 fs cgroup manager were not creating device cgroup directory, if `SkipDevices` is set. It always creates this directory now.
> * It seems impossible to get rid of `SkipFreezeOnSet` for systemd v1 cgroup manager since it can manage an existing systemd unit which has some device rules set, and any update of such unit results in rules being reloaded, so we have to freeze it before set.
> 
> ### TODO
> Finish this, in particular:
> 
> * ~fix systemd drivers~
> * ~remove SkipDevices workaround~
> * ~privatize some functions in libct/cg/devices~
> * general cleanup and documentation

libcontainer/cgroups/devices/devicefilter.go
```diff
@@ -23,14 +22,14 @@ const (
 	license = "Apache"
 )
 
-// DeviceFilter returns eBPF device filter program and its license string
-func DeviceFilter(rules []*devices.Rule) (asm.Instructions, string, error) {
+// deviceFilter returns eBPF device filter program and its license string.
+func deviceFilter(rules []*devices.Rule) (asm.Instructions, string, error) {
```
AkihiroSuda, line 26: Probably no need to unexpose `DeviceFilter`, `Emulator`, etc.
Or this could be another PR.

AkihiroSuda (approved):

libcontainer/cgroups/devices/devicefilter.go
kolyshkin: It's a separate commit so I can just drop it.

Or, if moving everything in a single package is not that good (let's see what @cyphar says), this can be undone, too.

kolyshkin:
@cyphar PTAL

kolyshkin:
@cyphar @opencontainers/runc-maintainers PTAL

libcontainer/cgroups/devices/devicefilter.go
cyphar: Moving it to a single package is fine. Honestly I probably should've kept it in this package in the first place (I wish Go had project-level internal symbols that didn't require you to restructure everything to use `internal/`).

AkihiroSuda (approved):

kolyshkin:
> @kolyshkin @AkihiroSuda @cyphar is there any chance of getting this merged this week so work on [containers/common#936](https://github.com/containers/common/pull/936) can resume?

Being the author of this, I can not really influence its acceptance (although I am a maintainer here, I can not accept my own PRs, and so I rely on other maintainers for this, the same way they rely on me).

You can still work on that using my repo fork; I guess that until a tagged runc release is made, you can't include this anyway so the fork is as good as the unreleased version.

kolyshkin:
OK, I have simplified the code a bit, removed all the new methods from cgroup managers, replacing those with 3 functions variables added to cgroups package, and added `init` to libct/cg/dev which sets those three variables to its own implementations. This should simplify the review as well. Still a big change, but mostly it's just moving the code around without changing it.

libcontainer/cgroups/cgroups.go
```diff
@@ -1,9 +1,24 @@
...
 import (
+	"errors"
+
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
+var (
+	// ErrNoDevRules is an error returned when a cgroup manager
+	// is not configured to set device rules.
+	ErrNoDevRules = errors.New("cgroup manager is not configured to set device rules")
```
cyphar, line 12: `ErrDevicesUnsupported` would probably be a clearer name for the error. `ErrNoDevRules` seems to be indicating the opposite of what the issue is (there are devices rules but no devices manager configured).

libcontainer/factory_linux.go
```diff
@@ -12,6 +12,8 @@ import (
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"golang.org/x/sys/unix"
 
+	//nolint:revive // Enable cgroup manager to manage devices
+	_ "github.com/opencontainers/runc/libcontainer/cgroups/devices"
```
cyphar, line 16: I am slightly concerned that if we drop this by accident, it wouldn't be a compile-time error, but I guess since we have integration tests that would almost all fail if we do forget to include this, I guess it's not that big of an issue.

libcontainer/cgroups/systemd/freeze_test.go
```diff
@@ -218,3 +219,138 @@ func requireV1(t *testing.T) {
...
+	// Check that we put the "container" into the "pod" cgroup.
+	if !strings.HasPrefix(cm.Path("freezer"), pm.Path("freezer")) {
+		t.Fatalf("expected container cgroup path %q to be under pod cgroup path %q",
+			cm.Path("freezer"), pm.Path("freezer"))
+	}
+	// Check the container is not reported as frozen despite the frozen parent.
+	cf, err := cm.GetFreezerState()
+	if err != nil {
+		t.Fatal(err)
+	}
```
cyphar, line 325: Maybe I'm missing something -- is this behaviour we want? I vaguely recall there was a patch which changed the freezer behaviour some time ago, is this just a consequence of that older change?

cyphar (changes requested):

libcontainer/factory_linux.go
kolyshkin: We always have some device rules, so the resulting runc won't be able to start a container -- that's a kind of error that should be caught by our tests.

libcontainer/cgroups/cgroups.go
kolyshkin: Right; renamed.

libcontainer/factory_linux.go
cyphar: Yeah, fair enough.

libcontainer/cgroups/systemd/freeze_test.go
cyphar: Ah right. Yeah that was the patch I was thinking of. Fair enough.

cyphar (approved):
LGTM.

AkihiroSuda (approved):

PR #3457, opened 2022-04-12
ci/gha: convert lint-extra from a job to a step

kolyshkin:
There is no need to parallelize lint and lint-extra jobs,
and they only differ with the arguments to golangci-lint.
Given that the longest time spent in these jobs is installing
libseccomp-dev, and that the second linter run can probably
benefit a lot from caching, it makes sense to merge them.

Move lint-extra from a separate job to a step in lint job.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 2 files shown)


kolyshkin:
OK it's working, except that golangci-lint is installed twice. Filed an issue about it: https://github.com/golangci/golangci-lint-action/issues/449

kolyshkin:
OK, I redid it and it now works as expected. PTAL @thaJeztah @mrunalp

AkihiroSuda (approved):

thaJeztah (approved):
LGTM, thanks!

Issue #676, opened 2016-03-22
Expand support for container id naming


mikebrow:
Currently we limit container ids through regex to \w ascii numbers and letters and a few additional selected chars.

The spec has no such limit.

Suggest changing the code in one of the following ways to add the ability to support unicode names:

1) do a touch cid to see if the OS supports it as a filename and report otherwise
2) figure out how to the fix the expression for the plethora of unicode chars that should also be valid 
3) hash the passed in cid, store the original in the container's state.. (slow)
4) @wrking suggestions: use percent-encoding or base64-encoding before using them as filenames 

https://tools.ietf.org/html/rfc3986#section-2.1
http://tools.ietf.org/html/rfc4648#section-54

Opening the issue for discussion... I offer to do a PR based on discussions.

mikebrow:
little interest shown

PR #675, opened 2016-03-22
Allow + in container ID

pankit:
Signed-off-by: pankit thapar <pankit@umich.edu>

PR #647, opened 2016-03-15
Fixing valid-id in regex

rajasec:
Fixes #635
Also removed _ ,as _ is part of \w
Signed-off-by: rajasec <rajasec79@gmail.com>

Issue #635, opened 2016-03-11
Valid IDs


ip1981:
Why are IDs restricted by this pattern?

```
idRegex = regexp.MustCompile(`^[\w_-]+$`)
```

https://github.com/opencontainers/runc/blob/master/libcontainer/factory_linux.go#L29

crosbymichael:
It is for valid filepath names because the container id needs to be written to disk in something like `/run` to store state.

ip1981:
File names can include a way more other characters. I fail to see reasons for existing restrictions. `foo.com` is a valid file name, but not a valid id. Why?

PR #3460, opened 2022-04-15
Do not use regexp

kolyshkin:
Don't get me wrong, I love regular expressions. But the regexp package is somewhat big and slow, and we can do just fine without.

Remove two last uses of regexp in runc code (modulo tests).

**This saves about 200K (2.3%) from the resulting binary, without losing any of the functionality.**

~~Currently a draft pending~~
 - [x] #3514 
 - [x] #3484 (which has https://github.com/urfave/cli/pull/1383)
 - [x] #3491 (which has https://github.com/cilium/ebpf/pull/642)
 - [x] #3543 (which has https://go-review.googlesource.com/c/protobuf/+/402774)
 - [x] #3585 (which has https://github.com/docker/go-units/pull/40)

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"math"
 	"os"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -231,18 +230,22 @@ func systemdVersion(cm *dbusConnManager) int {
 	return version
 }
 
-func systemdVersionAtoi(verStr string) (int, error) {
-	// verStr should be of the form:
-	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32" (without quotes).
-	// The result for all of the above should be 245.
-	// Thus, we unconditionally remove the "v" prefix
-	// and then match on the first integer we can grab.
-	re := regexp.MustCompile(`v?([0-9]+)`)
-	matches := re.FindStringSubmatch(verStr)
-	if len(matches) < 2 {
-		return 0, fmt.Errorf("can't parse version %s: incorrect number of matches %v", verStr, matches)
+// systemdVersionAtoi extracts a numeric systemd version from the argument.
+// The argument should be of the form: "v245.4-1.fc32", "245", "v245-1.fc32",
+// "245-1.fc32" (with or without quotes). The result for all of the above
+// should be 245.
+func systemdVersionAtoi(str string) (int, error) {
+	// Unconditionally remove the leading prefix ("v).
+	str = strings.TrimLeft(str, `"v`)
+	// Match on the first integer we can grab.
+	for i := 0; i < len(str); i++ {
+		if str[i] < '0' || str[i] > '9' {
+			// First non-digit: cut the tail.
+			str = str[:i]
+			break
+		}
 	}
-	ver, err := strconv.Atoi(matches[1])
+	ver, err := strconv.Atoi(str)
 	if err != nil {
 		return -1, fmt.Errorf("can't parse version: %w", err)
 	}
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"regexp"
 	"runtime/debug"
 	"strconv"
 
@@ -26,8 +25,6 @@ const (
 	stateFilename    = "state.json"
 	execFifoFilename = "exec.fifo"
 )
-
-var idRegex = regexp.MustCompile(`^[\w+-\.]+$`)
 
 // Create creates a new container with the given id inside a given state
 // directory (root), and returns a Container object.
@@ -260,8 +257,47 @@ func loadState(root string) (*State, error) {
 	return state, nil
 }
 
+// validateID checks if the supplied container ID is valid, returning
+// the ErrInvalidID in case it is not.
+//
+// The format of valid ID was never formally defined, instead the code
+// was modified to allow or disallow specific characters.
+//
+// Currently, a valid ID is a non-empty string consisting only of
+// the following characters:
+// - uppercase (A-Z) and lowercase (a-z) Latin letters;
+// - digits (0-9);
+// - underscore (_);
+// - plus sign (+);
+// - minus sign (-);
+// - period (.).
+//
+// In addition, IDs that can't be used to represent a file name
+// (such as . or ..) are rejected.
+
 func validateID(id string) error {
-	if !idRegex.MatchString(id) || string(os.PathSeparator)+id != utils.CleanPath(string(os.PathSeparator)+id) {
+	if len(id) < 1 {
+		return ErrInvalidID
+	}
+
+	// Allowed characters: 0-9 A-Z a-z _ + - .
+	for i := 0; i < len(id); i++ {
+		c := id[i]
+		switch {
+		case c >= 'a' && c <= 'z':
+		case c >= 'A' && c <= 'Z':
+		case c >= '0' && c <= '9':
+		case c == '_':
+		case c == '+':
+		case c == '-':
+		case c == '.':
+		default:
+			return ErrInvalidID
+		}
+
+	}
+
+	if string(os.PathSeparator)+id != utils.CleanPath(string(os.PathSeparator)+id) {
 		return ErrInvalidID
 	}
 
--- a/libcontainer/cgroups/systemd/systemd_test.go
+++ b/libcontainer/cgroups/systemd/systemd_test.go
@@ -35,8 +35,11 @@ func TestSystemdVersion(t *testing.T) {
 		{`"v245.4-1.fc32"`, 245, false},
 		{`"241-1"`, 241, false},
 		{`"v241-1"`, 241, false},
-		{"NaN", 0, true},
-		{"", 0, true},
+		{`333.45"`, 333, false},
+		{`v321-0`, 321, false},
+		{"NaN", -1, true},
+		{"", -1, true},
+		{"v", -1, true},
 	}
 	for _, sdTest := range systemdVersionTests {
 		ver, err := systemdVersionAtoi(sdTest.verStr)


kolyshkin:
Fedora CI failed with
```
[04] run rootless tests ... (idmap+cgroup)
tests/rootless.sh: line 44: /etc/subuid.tmp: Read-only file system
make: *** [Makefile:103: localrootlessintegration] Error 1
```
which is pretty weird. Perhaps it's the problem with the underlying fs and so the kernel remounted it read-only.

Restarted.

AkihiroSuda:
> slow

Yes, but not the bottleneck of runc AFAICS.
I don’t see necessity to remove regexp.

kolyshkin:
A few of runc dependencies use regexp as well.
 * github.com/docker/go-units: fixed by https://github.com/docker/go-units/pull/40
 * github.com/cilium/ebpf: fixed by https://github.com/cilium/ebpf/pull/642
 * github.com/urfave/cli (via github.com/russross/blackfriday): being fixed by TODO (backport of https://github.com/urfave/cli/pull/1375)
 * google.golang.org/protobuf: being fixed by https://go-review.googlesource.com/c/protobuf/+/402774
 
 Once (and if) all of these are merged, I am going to bring this out of draft, as it will save us 200K (130K stripped) from the runc binary size (which is about 5%, comparing the stripped binaries).

kolyshkin:
Currently saves 531K from the stripped binary (which is 5.6% of binary size). Still very far from crun footprint, yet it seems like a worthwhile saving.

kolyshkin:
ebpf bump moved to https://github.com/opencontainers/runc/pull/3491, will rebase once merged.

kolyshkin:
`urfave_cli` bump moved to https://github.com/opencontainers/runc/pull/3484, will rebase once merged.

kolyshkin:
Rebased, made a fresh comparison:

```console
[kir@kir-rhat runc]$ go version
go version go1.18.1 linux/amd64

[kir@kir-rhat runc]$ git describe HEAD
v1.1.0-239-g93ad6a85

[kir@kir-rhat runc]$ ls -l runc.main runc.no-regexp
-rwxrwxr-x. 1 kir kir 12999304 Jul 14 13:31 runc.main
-rwxrwxr-x. 1 kir kir 12708080 Jul 14 13:31 runc.no-regexp

[kir@kir-rhat runc]$ strip runc.main runc.no-regexp

[kir@kir-rhat runc]$ ls -l runc.main runc.no-regexp
-rwxrwxr-x. 1 kir kir 9310880 Jul 14 13:33 runc.main
-rwxrwxr-x. 1 kir kir 9100352 Jul 14 13:33 runc.no-regexp

[kir@kir-rhat runc]$ size runc.main runc.no-regexp
   text	   data	    bss	    dec	    hex	filename
5645422	3657744	 229832	9532998	 917646	runc.main
5503086	3587520	 229416	9320022	 8e3656	runc.no-regexp
```

So, the saving is about 200K for a stripped binary.

go.mod
```diff
@@ -8,7 +8,6 @@ require (
 	github.com/containerd/console v1.0.3
 	github.com/coreos/go-systemd/v22 v22.3.2
 	github.com/cyphar/filepath-securejoin v0.2.3
-	github.com/docker/go-units v0.4.0
```
thaJeztah, line 11: I pushed v0.5.0 so you could update to that instead if you want to

kolyshkin: Thanks, replace the docker/units removal with the bump commit; PTAL

kolyshkin: Just out of curiosity, I have re-applied the commit that replaces docker/units with the in-house implementation of RAMInBytes, and the difference between these two is negligible (meaning, I guess, that the linker is smart enough to remove the functions that are not used). Here are the raw numbers:

```console
[kir@kir-rhat runc]$ size runc-no-regex runc-no-units
   text	   data	    bss	    dec	    hex	filename
5522150	3598080	 229384	9349614	 8ea9ee	runc-no-regex
5520182	3597232	 229352	9346766	 8e9ece	runc-no-units
[kir@kir-rhat runc]$ ls -l  runc-no-regex runc-no-units
-rwxr-xr-x. 1 kir kir 9129056 Aug 31 18:26 runc-no-regex
-rwxr-xr-x. 1 kir kir 9128992 Aug 31 18:32 runc-no-units
```

libcontainer/factory_linux.go
```diff
@@ -261,7 +258,28 @@ func loadState(root string) (*State, error) {
 }
 
 func validateID(id string) error {
```
thaJeztah, line 260: Perhaps would be good to (despite it not being exported) add a GoDoc to describe accepted formats, and outline some of it.


Looking at this PR made me go down a bit of a trip in history; let me post below, but I should probably post elsewhere

---------

I noticed the underscore being added, which was not explicitly included in the
regexp, so I wondered if that was _accidentally_ included in the past (as it's
easy to forget it's part of `\w`). I also was curious if this ID was meant to
match other ID formats (e.g., used in Moby), some of which are required to start
with an alpha-numeric character, and `._-+` only allowed as separator.

So, I went on a short trip down history;

- Validation was added in https://github.com/opencontainers/runc/commit/66e6806fd23f6c793f9478374278ed221dea3347 / https://github.com/docker-archive/libcontainer/commit/8fcda56d79d0c3652d0eb4f0317c951cfddc5b18 (https://github.com/docker-archive/libcontainer/pull/249),
  which allowed for "word" (`\w`) characters and had the underscore added explicitly. Maximum length was limited to 1024 characters (no reason mentioned for this).
- The length constraint was moved from the regex to a separate check in https://github.com/docker-archive/libcontainer/commit/de57f78590cefdf08249118e2d710908cc81b16a
- In https://github.com/opencontainers/runc/commit/d95050084baa71eb46bfeb3844cc3b806497f49c (https://github.com/opencontainers/runc/pull/31), hyphens (`-`) were added to the list of allowed characters;
  > A directory with a hyphen currently generates an InvalidId error because
  > of the regex in libcontainer. I don't believe there is any reason a
  > hyphen should be disallowed.
- Periods (`.`) were added in https://github.com/opencontainers/runc/commit/d4be3405c7f9cd88cfa4ea8cf6f6cdd4ff8ddd53 (https://github.com/opencontainers/runc/pull/647), which also removed the explicit underscore, acknowledging that it's reduntant, and already part of `\w`. The issue linked to that mentions https://github.com/opencontainers/runc/issues/635#issuecomment-196467367;
  > It is for valid filepath names because the container id needs to be written
  > to disk in something like `/run` to store state.
- Finally, `+` symbols were added in https://github.com/opencontainers/runc/commit/4629512d895e5754b93556e5093f5f0b82e5bda5 (https://github.com/opencontainers/runc/pull/675). The motivation there was https://github.com/opencontainers/runc/pull/675#issuecomment-199924556
  > I added just the + because I have a use case where the foo+abc is used as a
  > dir name and I think we should allow the containers to be named following the
  > same rules as file/dir names.
  > Yes, I agree with you that we should allow more chars that fall into this category.


So, from the above;

- The original intent was for the ID to be something that can be used as part of a path, and limiting it to "portable" characters.
- **Up until https://github.com/opencontainers/runc/pull/675, this meant IDs followed the [POSIX Portable Filename Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282)**
 
Some what unfortunately, the runtime-spec does _NOT_ define a format (or restrictions) in any way, which means that currently runc is more restrictive than the OCI runtime spec (overall this is fine). I found multiple attempts to have this included;

- There was a discussion in https://github.com/opencontainers/runtime-spec/pull/87#discussion_r35828046 (and [on the mailing list](https://groups.google.com/a/opencontainers.org/g/dev/c/P9gZBYhiqDE)) about the ID format being under-specified, and I see a patch was created for clarification (but not merged); https://github.com/wking/opencontainer-runtime-spec/commit/2ec34c62b019431c0edebead50b56ebafe70a67c (posted at the end of this comment).
- Following https://github.com/opencontainers/runc/pull/675, @mikebrow opened a ticket in the runc repository to gather interest on formalizing the validation https://github.com/opencontainers/runc/issues/676, which did not get participation, so was closed.
- https://github.com/opencontainers/runc/pull/675#issuecomment-199915890 also mentions that the runtime spec removed the requirement for IDs to be usable as element in paths. 

To summarize my thoughts;

- It's probably good to document this function, and to outline the accepted characters (possibly outline (some of) the motivations for the chosen set of characters.
- I (personally) would be in favor of formalizing accepted characters, and would've been in favor of adopting the [POSIX Portable Filename Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282), but the `+` addition was a bit unfortunate for that.
- Perhaps it's worth giving it another attempt to update the OCI spec (at least with a recommendation if that's acceptable)


<details>
<summary>proposed PATCH for formalizing the format from 2015</summary>

```patch
From 2ec34c62b019431c0edebead50b56ebafe70a67c Mon Sep 17 00:00:00 2001
From: "W. Trevor King" <wking@tremily.us>
Date: Mon, 10 Aug 2015 12:07:20 -0700
Subject: [PATCH] runtime: Document container ID charset and uniqueness domain

Allow the runtime to use it's own scheme, but let the caller use UUIDs
if they want.  Jonathan asked for clarification as part of #87, but
didn't suggest a particular approach [1].  When we discussed it in the
2015-08-26 meeting [2], the consensus was to just allow everything.
With container IDs like 'a/b/c' leading to state entries like
'/var/oci/containers/a/b/c/state.json'.  But that could get ugly with
container IDs that contain '../' etc.  And perhaps there are some
filesystems out there that cannot represent non-ASCII characters
(actually, I'm not even sure what charset our JSON is in ;).  I'd
rather pick this minimal charset which can handle UUIDs, and make life
easy for runtime implementers and safe for bundle consumers at a
slight cost of flexibility for bundle-authors.

There was some confusion on the list about what "ASCII letters" meant
[3], so I've explicitly listed the allowed character ranges.  Here's a
Python 3 script that shows the associated Unicode logic:

  import unicodedata

  # http://www.unicode.org/reports/tr44/tr44-4.html#GC_Values_Table
  category = {
    'Ll': 'lowercase letter',
    'Lu': 'uppercase letter',
    'Nd': 'decimal number',
    'Pd': 'dash punctuation',
  }

  for i in range(1<<7):
      char = chr(i)
      abbr = unicodedata.category(char)
      if abbr[0] in ['L', 'N'] or abbr == 'Pd':
          cat = category[abbr]
          print('{:02x} {} {}'.format(i, char, cat))

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r35828046
[2]: https://github.com/opencontainers/specs/wiki/Meeting-Minutes-2015-08-26
[3]: https://groups.google.com/a/opencontainers.org/d/msg/dev/P9gZBYhiqDE/-ptpOcQ5FwAJ
     Message-Id: <7ec9cff6-c1a6-4beb-82de-16eb412bf2f8@opencontainers.org>

Reported-by: Jonathan Boulle <jonathanboulle@gmail.com>
Signed-off-by: W. Trevor King <wking@tremily.us>
---
 runtime.md | 5 +++++
 1 file changed, 5 insertions(+)

diff --git a/runtime.md b/runtime.md
index be2045b58..c9da6af9a 100644
--- a/runtime.md
+++ b/runtime.md
@@ -11,6 +11,9 @@ By providing a default location that container state is stored external applicat

 * **version** (string) Version of the OCI specification used when creating the container.
 * **id** (string) ID is the container's ID.
+  Only ASCII letters, numbers, and hyphens are valid (a–z, A–Z, 0–9, and ‘-’).
+  This value must be unique for a given host, but need not be universally unique.
+  Runtimes must allow the caller to set this ID, so that callers may choose, for example, to use [UUIDs][uuid] for universal uniqueness.
 * **pid** (int) Pid is the ID of the main process within the container.
 * **root** (string) Root is the path to the container's bundle directory.

@@ -94,3 +97,5 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin

 `path` is required for a hook.
 `args` and `env` are optional.
+
+[uuid]: https://tools.ietf.org/html/rfc4122
```

</details>

libcontainer/cgroups/systemd/common.go
```diff
@@ -231,18 +230,21 @@ func systemdVersion(cm *dbusConnManager) int {
 	return version
 }
 
-func systemdVersionAtoi(verStr string) (int, error) {
-	// verStr should be of the form:
-	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32" (without quotes).
+func systemdVersionAtoi(str string) (int, error) {
+	// str should be of the form:
```
thaJeztah, line 234: Perhaps this would be useful to be put as GoDoc for the function as well.

libcontainer/cgroups/systemd/common.go
```diff
@@ -231,18 +230,21 @@ func systemdVersion(cm *dbusConnManager) int {
...
+	// Thus, we unconditionally remove the "v prefix
 	// and then match on the first integer we can grab.
-	re := regexp.MustCompile(`v?([0-9]+)`)
-	matches := re.FindStringSubmatch(verStr)
-	if len(matches) < 2 {
-		return 0, fmt.Errorf("can't parse version %s: incorrect number of matches %v", verStr, matches)
+	str = strings.TrimLeft(str, `"v`)
+	for i := 0; i < len(str); i++ {
+		if str[i] < '0' || str[i] > '9' {
+			// First non-digit.
```
thaJeztah, line 242: Had my wires crossed initially, and thought this was stripping any _leading_ non-digit characters (so making `"vHELLO245.4-1.fc32"` valid), but because we strip the `"v` first, this look (should) start at `245.4-1.fc32`, and removes the _trailing_ non-digit parts (`.4-1.fc32"`).

Perhaps

```suggestion
			// Strip non-digit suffix and quote (e.g. `.4-1.fc32"`).
```

thaJeztah (approved):
code LGTM, but left some comments to improve some of the comments (the "GoDoc" suggestions aren't new, so would also be fine as a follow-up)

libcontainer/cgroups/systemd/common.go
```diff
@@ -231,18 +230,21 @@ func systemdVersion(cm *dbusConnManager) int {
...
+	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32" (with quotes).
 	// The result for all of the above should be 245.
-	// Thus, we unconditionally remove the "v" prefix
+	// Thus, we unconditionally remove the "v prefix
 	// and then match on the first integer we can grab.
-	re := regexp.MustCompile(`v?([0-9]+)`)
-	matches := re.FindStringSubmatch(verStr)
-	if len(matches) < 2 {
-		return 0, fmt.Errorf("can't parse version %s: incorrect number of matches %v", verStr, matches)
+	str = strings.TrimLeft(str, `"v`)
```
thaJeztah, line 239: OH! Actually, this is not correct; this should be `strings.TrimPrefix`. `strings.TrimLeft` takes a "cut set" of characters, so currently `"vvvvvv""v"v"v""v245.4-1.fc32"` would be accepted;

https://go.dev/play/p/-Z4D9U4-oYM

thaJeztah (dismissed):
see my comment about `TrimLeft` vs `TrimPrefix`

libcontainer/cgroups/systemd/common.go
kolyshkin: So, this comment is a logical continuation of the previous one ("match on the first integer we can grab"). While that comment above is, I think, sufficient, I am adding a little bit more here.

Changed to `First non-digit: cut the tail.`

libcontainer/cgroups/systemd/common.go
kolyshkin: This is correct since it allows us to Trim either `v` or `"v` (i.e. with or without the quote).

Sure, it will also cut `v"` or `vvvv"""vvvv` but we do not expect such input here (the input expectations are documented).

Within the realm of expected input,
```go
strings.TrimLeft(str, `"v`)
```
work the same as
```go
strings.TrimPrefix(str, `"`)
strings.TrimPrefix(str, `v`)
```
but is somewhat more compact.

kolyshkin: @thaJeztah PTAL ^^

libcontainer/cgroups/systemd/common.go
kolyshkin: moved to godoc

kolyshkin:
@thaJeztah addressed your comments, PTAL

libcontainer/factory_linux.go
```diff
@@ -261,7 +258,28 @@ func loadState(root string) (*State, error) {
...
+	for i := 0; i < len(id); i++ {
+		c := id[i]
+		switch {
+		case c >= 'a' && c <= 'z':
+		case c >= 'A' && c <= 'Z':
+		case c >= '0' && c <= '9':
+		case c == '_':
+		case c == '+':
+		case c == '-':
+		case c == '.':
```
mrunalp, line 275: The current regex also includes the comma :/
https://go.dev/play/p/n9ZiwZM8KMd

mrunalp: It looks like we missed `\` before `-` making it a range from `+` to `.` instead here - https://github.com/opencontainers/runc/pull/675/files

It should have been `^[\w+\-\.]+$`

kolyshkin: Oh wow.

I guess that the best way is to document that we're no longer allowing `,` in container name.

The alternative is to add `,` to the above `switch`.

libcontainer/factory_linux.go
kolyshkin: > * Finally, + symbols were added in 4629512 (#675). 

Most interestingly, as pointed out by @mrunalp [below](https://github.com/opencontainers/runc/pull/3460#discussion_r1028587711), this commit also implicitly adds comma (`,`) to the list of allowed characters, since `+-\.` is now a range from `+` (ASCII 43) to `.` (ASCII 45) and that includes ASCII 44 (`,`).

Now we need to decide whether to keep supporting `,` or not. I vote to drop `,` and deal with the consequences.

kolyshkin:
@thaJeztah @mrunalp I have documented the function, PTAL

mrunalp (approved):

thaJeztah (approved):
LGTM

w.r.t.;

> I guess that the best way is to document that we're no longer allowing , in container name.
>
> The alternative is to add , to the above switch.

I think it should be ok to not allow `,`, as (from the discussion) it looks like that was a bug, but perhaps good to make sure it's called out in the changelog / release-notes

PR #3465, opened 2022-05-03
vendor: bump seccomp/libseccomp-golang to f33da4d

crazy-max:
fixes clang issue https://github.com/seccomp/libseccomp-golang/pull/90.

@thaJeztah there are other changes while vendoring it https://github.com/seccomp/libseccomp-golang/compare/3879420cc921...f33da4d89646. not sure about the impact of them on runc.

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>


thaJeztah:
Looks like there's some uses of `libseccomp.ActKill` that need to be updated locally;

```
Error: SA1019: libseccomp.ActKill is deprecated: use ActKillThread  (staticcheck)
  Error: SA1019: libseccomp.ActKill is deprecated: use ActKillThread  (staticcheck)
  
  Error: issues found
```

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -113,8 +113,8 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 // Convert Libcontainer Action to Libseccomp ScmpAction
 func getAction(act configs.Action, errnoRet *uint) (libseccomp.ScmpAction, error) {
 	switch act {
-	case configs.Kill:
-		return libseccomp.ActKill, nil
+	case configs.Kill, configs.KillThread:
```
thaJeztah, line 116: Wondering if one of `configs.Kill` also should be deprecated on this side 🤔  (not for this PR, just thinking out loud?)

thaJeztah:
@crazy-max can you swap the order of the commits (assuming both `ActKillThread` and `ActKill` were present before updating; if not, then it's best to squash the commits to preserve git bisect. 😅

libcontainer/seccomp/seccomp_linux.go
kolyshkin: We have the same problem in libcontainer/config -- Kill and KillThread are two distinct constants. Guess it makes sense to deprecate the former.

thaJeztah: Yup; that was my thought as well

I'm ok doing that separately, but something we should probably do

kolyshkin: Oh well, it's a bit bigger than I thought; see https://github.com/opencontainers/runc/pull/3466

@crazy-max feel free to cherry-pick that commit to here.

I also think we should really deprecate ACT_KILL, but it should be done in the runtime-spec repo first.

crazy-max: @kolyshkin

> feel free to cherry-pick that commit to here.

done

kolyshkin (approved):
LGTM

kolyshkin:
@thaJeztah I see your LGTM, but you probably forgot to actually "Accept" the PR

thaJeztah (approved):
whoops! probably picked "comment" instead of "approve" 😂 

LGTM (for real this time)

crazy-max:
@thaJeztah @kolyshkin is there a milestone that includes this PR to fix the clang issue? cc @tonistiigi

PR #3468, opened 2022-05-04
Remove tun/tap from the default device rules

kolyshkin:
Looking through git blame, this was added by commit 9fac18329
aka "Initial commit of runc binary", most probably by mistake.

Obviously, a container should not have access to tun/tap device, unless
it is explicitly specified in configuration.

Now, removing this might create a compatibility issue, but I see no
other choice.

Aside from the obvious misconfiguration, this should also fix the
annoying

> Apr 26 03:46:56 foo.bar systemd[1]: Couldn't stat device /dev/char/10:200: No such file or directory

messages from systemd on every container start, when runc uses systemd
cgroup driver, and the system runs an old (< v240) version of systemd
(the message was presumably eliminated by [1]).

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1945929

[1] https://github.com/systemd/systemd/pull/10996/commits/d5aecba6e0b7c73657c4cf544ce57289115098e7

### DIFF
--- a/libcontainer/cgroups/ebpf/devicefilter/devicefilter_test.go
+++ b/libcontainer/cgroups/ebpf/devicefilter/devicefilter_test.go
@@ -120,21 +120,14 @@ block-8:
         51: Mov32Imm dst: r0 imm: 1
         52: Exit
 block-9:
-// tuntap (c, 10, 200, rwm, allow)
-        53: JNEImm dst: r2 off: -1 imm: 2 <block-10>
-        54: JNEImm dst: r4 off: -1 imm: 10 <block-10>
-        55: JNEImm dst: r5 off: -1 imm: 200 <block-10>
-        56: Mov32Imm dst: r0 imm: 1
-        57: Exit
-block-10:
 // /dev/pts (c, 136, wildcard, rwm, true)
-        58: JNEImm dst: r2 off: -1 imm: 2 <block-11>
-        59: JNEImm dst: r4 off: -1 imm: 136 <block-11>
-        60: Mov32Imm dst: r0 imm: 1
-        61: Exit
-block-11:
-        62: Mov32Imm dst: r0 imm: 0
-        63: Exit
+        53: JNEImm dst: r2 off: -1 imm: 2 <block-10>
+        54: JNEImm dst: r4 off: -1 imm: 136 <block-10>
+        55: Mov32Imm dst: r0 imm: 1
+        56: Exit
+block-10:
+        57: Mov32Imm dst: r0 imm: 0
+        58: Exit
 `
 	var devices []*devices.Rule
 	for _, device := range specconv.AllowedDevices {
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -302,16 +302,6 @@ var AllowedDevices = []*devices.Device{
 			Allow:       true,
 		},
 	},
-	// tuntap
-	{
-		Rule: devices.Rule{
-			Type:        devices.CharDevice,
-			Major:       10,
-			Minor:       200,
-			Permissions: "rwm",
-			Allow:       true,
-		},
-	},
 }
 
 type CreateOpts struct {


rhatdan:
LGTM

thaJeztah (approved):
SGTM

(I'm asking around a bit if others can see potential issues with the change)

DentonGentry:
>Got some feedback from some people; one person mentioned that this could affect users such as Tailscale, who may be using this. We have a joined Slack channel with them, so I put out a question to ask for feedback.

Tailscale's instructions and sample docker-compose.yml files say to explicitly include /dev/net/tun, anyone who referenced those shouldn't be impacted. 

```
  devices:
    - /dev/net/tun:/dev/net/tun
  cap_add:
    - NET_ADMIN
    - NET_RAW
```

support@tailscale.com can watch for symptoms that look like a container which had been relying on getting a /dev/net/tun device via the prior default setting, and advise adding it in the docker-compose.yml or a command line argument.

AkihiroSuda:
Cc @giuseppe Is this safe for rootless Podman?

thaJeztah:
@kolyshkin would it be possible for you to post some test-builds from this PR somewhere for people to download?

giuseppe:
> Cc @giuseppe Is this safe for rootless Podman?

from a quick test, it seems the slirp4netns network works fine

mrunalp:
LGTM

kolyshkin:
> @kolyshkin would it be possible for you to post some test-builds from this PR somewhere for people to download?

Those are actually available (for every PR), although not in a very consumable form. Click on "Checks" tab, when click "validate" from the left-side list, and on the bottom of that screen you'll see a big `release-xxxxxx` tarball under "Artifacts". It contains all the static binaries.

AkihiroSuda (approved):

thaJeztah:
And.. there I went looking, but it's not in the `Checks` tab, it's in the top-level `Actions` tab; this is the last run of this PR; https://github.com/opencontainers/runc/actions/runs/2272679304

And download link is https://github.com/opencontainers/runc/suites/6379868023/artifacts/231538496

(super confusing!)

cyphar (approved):
LGTM, but I don't think this was a mistake -- the initial set of devices added was based on what was needed to get Docker containers to work with most programs people were using.

When I was reworking the devices handling code a year or two ago, I also wondered whether we should drop this (and I dropped some other far more dangerous rules, but left this one because @crosbymichael couldn't remember what this was needed for and I didn't want to break anything in a security update if I could help it).

That being said, Docker has their own default devices rules that are set externally to runc (so keeping this default is not necessary) and I don't see why a standard container would need tun/tap devices (though since the container is in a network namespace, I suspect that it wouldn't be an issue that they can create such devices). However I believe Docker generates their rules using our library functions, so maybe we should do a test build of Docker with this change and test that before we do a release?

thaJeztah:
> That being said, Docker has their own default devices rules that are set externally to runc (so keeping this default is not necessary)

Yes, I was thinking about that later; perhaps this change wouldn't affect the defaults on Docker  (or Containerd for that matter).

> However I believe Docker generates their rules using our library functions, so maybe we should do a test build of Docker with this change and test that before we do a release?

Good point; perhaps won't hurt to do a draft PR in the docker repo to verify (at least CI).

thaJeztah:
opened https://github.com/moby/moby/pull/43569 to get a run of CI with these changes

kolyshkin:
> opened [moby/moby#43569](https://github.com/moby/moby/pull/43569) to get a run of CI with these changes

Seems passing. Seems that we have a consensus this can be merged. @opencontainers/runc-maintainers this is your last chance to say no :)

AkihiroSuda (approved):

thaJeztah:
Okay; found it; it looks like it changed in the containerd integration; https://github.com/moby/moby/pull/20662

Looks like this commit; https://github.com/moby/moby//commit/9c4570a958df42d1ad19364b1a8da55b891d850a#diff-6e3774954611ecbf441ac0aa5841f11926c1e618b391ff64c330cd64e8875756L126 removed the use of `configs.DefaultAllowedDevices` (from runc/libcontainer/configs)

```go
allowedDevices := mergeDevices(configs.DefaultAllowedDevices, userSpecifiedDevices)
```

and replaced it with a local list of defaults; https://github.com/moby/moby//commit/9c4570a958df42d1ad19364b1a8da55b891d850a#diff-0a4d29b610dc02821e36ed4b8ce4a2dffc64d09b1b554bfb277d735acac459d4R89-R208

thaJeztah:
Let me bring this one in 👍

zhsj:
Hi,

Is there any security concerns for using tun/tap in container? I think not.?

If not, I think it makes people more difficult to use tun/tap on Kubernetes. Now they need to use some tun/tap device plugins... (There's no convenient --device option like docker cli).

thaJeztah:
@zhsj what runtime are you using for kubernetes? As at least docker (moby) and containerd already seem to be using the same defaults;

> Okay; found it; it looks like it changed in the containerd integration; https://github.com/moby/moby/pull/20662
>
> Looks like this commit; https://github.com/moby/moby//commit/9c4570a958df42d1ad19364b1a8da55b891d850a#diff-6e3774954611ecbf441ac0aa5841f11926c1e618b391ff64c330cd64e8875756L126 removed the use of configs.DefaultAllowedDevices (from runc/libcontainer/configs)

zhsj:
> @zhsj what runtime are you using for kubernetes? As at least docker (moby) and containerd already seem to be using the same defaults;

Hmm, I'm using docker. After I re-read your comments on containerd's PR, I realized you said docker has already removed tun/tap long before. Probably I added `CAP_SYS_ADMIN` so that I can still use tun/tap without `--device` option, while cgroup won't deny it on this scenario .

cyphar:
@kroese The tun devices are not included in the default set of devices that we actually create, and the only supported way of specifying them (`--device` with Docker, or devices entries with `runc`) will add the device to the allow access list.

If you depended on the device being available, it would've been advisable to specify it in a supported way... Doing `mknod` of devices inside the container itself isn't really best practice because it doesn't work with user namespaces and it depends on (in this case) undocumented behaviour. But specifying devices does work in all cases and is fully supported. As mentioned above, compose supports this (and other containerised tools like tailwind do this correctly).

We might need to revert this if it's a very common issue, but imho such setups were already kind of questionable from a best practices perspective...

> To use `/dev/net/tun` you still need the NET_ADMIN capability, so there seems no big security risk in the first place.

Adding permissions to devices that we don't enable by default is needlessly adding attack surface. If a container is configured to use `CAP_NET_ADMIN` for some other reason other than configuring tun/tap, we don't want to allow more access than the user expected.

maggie44:
Biggest impact on the Kubernetes side for me so far, is I having to run pods with `privileged` instead of `NET_ADMIN` (https://github.com/tailscale/tailscale/pull/14262#issue-2713214388). Seems like a big jump up in permissions just to gain the functionality that could have been achieved before with `NET_ADMIN`.

nillestimothy:
> Biggest impact on the Kubernetes side for me so far, is having to run pods with `privileged` instead of `NET_ADMIN` ([tailscale/tailscale#14262 (comment)](https://github.com/tailscale/tailscale/pull/14262#issue-2713214388)). Seems like a big jump up in permissions just to gain the functionality that could have been achieved before with `NET_ADMIN`. I am creating `/dev/net/tun` in the container rather than host mounting, but have to grant elevated privileges to use it.

It appears that we are trading one aspect of security for another. While I may not have extensive experience, isn’t specifying NET_ADMIN to limit access to /dev/net/tun an example of implementing the principle of least privilege?

Although this change was introduced two years ago, the recent update to containerd, which now uses the runc binary version 1.2.1 instead of 1.1.14, is beginning to impact a broader range of users. This includes Docker Swarm users, who are particularly affected by the inability to specify device mappings like /dev/net/tun directly in the deploy section of a Compose file.

kolyshkin:
So, it looks like upper-level tools do not have ways to specify an additional device to be added to the list, and because of that users have to use privileged containers. This appears very wrong to me, so I'm open to reverting the change.

maggie44:
> So, it looks like upper-level tools do not have ways to specify an additional device to be added to the list, and because of that users have to use privileged containers. This appears very wrong to me, so I'm open to reverting the change.

It is certainly problematic having to grant privileged access, but that doesn’t change the fact that the original issue still stands, a container should not have access to tun/tap device, unless it is explicitly specified in configuration. I’m not sure the scope of NET_ADMIN, but mounting the device like with Docker is probably a better option overall than NET_ADMIN if it could be supported by upper level tools.

thaJeztah:
> So, it looks like upper-level tools do not have ways to specify an additional device to be added to the list,

For docker, it looks like it's possible for `docker run` (and docker compose) by adding it as device; https://github.com/opencontainers/runc/pull/3468#issuecomment-1118721925

I'd have to check on the swarm services side; I know swarm services (by design) have been more constrained on some parts to keep the service-spec portable between nodes in a cluster.

cyphar:
It also  seems to be impacting a fair number of users, so we might have to revert it (especially if there really isn't a practical workaround in higher-level orchestration tools)...

It would be really nice if we could give a warning to users to let them know they should stop depending on this mis-feature and specify device configurations properly, but unfortunately there isn't a way of knowing ahead of time if a given container is going to use `/dev/net/tun` (there are ways of getting that information by tracing the container in various ways, but that kind of tracing would not fit with the runtime model of `runc` -- this would be something that higher level runtimes would need to implement, and I suspect they'd have little interest in doing that).

ClashTheBunny:
Is there an issue we can follow that is tracking decisions on removal?

Is the only known way to get this working on kubernetes with 1.31 to elevate all of the internet accessable security critical containers to be fully privileged? k3s, for example, has hit 1.31 with the stable channel, so it's going to be going out widely, and once people put `privileged: true` into their manifests, you're not going to be getting that removed anytime soon, if ever.

SuzukiHonoka:
This breaks my application, I'll need to set privileged=true to use the tun device, which is very riscky than NET_ADMIN.

PR #3470, opened 2022-05-09
Add support for cpu.idle

kolyshkin:
Kernel 5.15 adds a way to set `SCHED_IDLE` on a per-cgroup basis.

* kernel support was added in https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee
* runtime-spec support was added in https://github.com/opencontainers/runtime-spec/pull/1136, amended by https://github.com/opencontainers/runtime-spec/pull/1145 and https://github.com/opencontainers/runtime-spec/pull/1146
* systemd support is not there yet (https://github.com/systemd/systemd/pull/23299)
* crun support: https://github.com/containers/crun/pull/915

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/moby/sys/mountinfo v0.6.1
 	github.com/mrunalp/fileutils v0.5.0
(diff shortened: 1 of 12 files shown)


wineway:
https://github.com/opencontainers/runc/pull/3377 I have pull a request about this🤔

kolyshkin:
Closing in favor of https://github.com/opencontainers/runc/pull/3377 (which has the test case picked up from there).

PR #3474, opened 2022-05-20
seccomp: enosys: always return -ENOSYS for setup(2)

cyphar:
On s390x, syscalls above 255 are multiplexed using the (now otherwise
unused) setup(2) syscall (syscall number 0). If the kernel supports the
syscall then it will correctly translate the syscall number such that
seccomp will correctly detect it -- however, for unknown syscalls the
syscall number remains unchanged. This can be verified by running the
following program under strace:

	int main(void)
	{
		scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_TRAP);
		seccomp_load(ctx);

		return syscall(439, AT_FDCWD, "asdf", X_OK, 0);
	}

Which will then die with the following signal (on pre-5.8 kernels):

	--- SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP,
	            si_call_addr=0x3ffb3006c22, si_syscall=__NR_setup,
	            si_arch=AUDIT_ARCH_S390X} ---

(Note that the si_syscall is __NR_setup, not __NR_faccessat2.)

As a result, the -ENOSYS handling we had previously did not work
completely correctly on s390x because any syscall not supported by the
kernel would be treated as syscall number 0 rather than the actual
syscall number.

Always returning -ENOSYS will not cause any issues because in all of the
cases where this multiplexing occurs, seccomp will see the remapped
syscall number -- and no userspace program will call setup(2)
intentionally (the syscall has not existed in Linux for decades and was
originally a hack used early in Linux init prior to spawning pid1 -- so
you will get -ENOSYS from the kernel anyway).

### DIFF
--- a/libcontainer/seccomp/patchbpf/enosys_linux.go
+++ b/libcontainer/seccomp/patchbpf/enosys_linux.go
@@ -80,6 +80,9 @@ import "C"
 
 var retErrnoEnosys = uint32(C.C_ACT_ERRNO_ENOSYS)
 
+// Assume sizeof(int) == 4 in the BPF program.
+const bpfSizeofInt = 4
+
 // This syscall is used for multiplexing "large" syscalls on s390(x). Unknown
 // syscalls will end up with this syscall number, so we need to explcitly
 // return -ENOSYS for this syscall on those architectures.
@@ -99,7 +102,6 @@ func isAllowAction(action configs.Action) bool {
 
 func parseProgram(rdr io.Reader) ([]bpf.RawInstruction, error) {
 	var program []bpf.RawInstruction
-loop:
 	for {
 		// Read the next instruction. We have to use NativeEndian because
 		// seccomp_export_bpf outputs the program in *host* endian-ness.
@@ -107,7 +109,7 @@ loop:
 		if err := binary.Read(rdr, utils.NativeEndian, &insn); err != nil {
 			if errors.Is(err, io.EOF) {
 				// Parsing complete.
-				break loop
+				break
 			}
 			if errors.Is(err, io.ErrUnexpectedEOF) {
 				// Parsing stopped mid-instruction.
@@ -321,7 +323,7 @@ func generateEnosysStub(lastSyscalls lastSyscallMap) ([]bpf.Instruction, error) 
 		// share this code between architecture branches.
 		section := []bpf.Instruction{
 			// load [0] (syscall number)
-			bpf.LoadAbsolute{Off: 0, Size: 4}, // NOTE: We assume sizeof(int) == 4.
+			bpf.LoadAbsolute{Off: 0, Size: bpfSizeofInt},
 		}
 
 		switch len(maxSyscalls) {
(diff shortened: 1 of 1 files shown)


libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -522,10 +522,30 @@ func generateEnosysStub(lastSyscalls lastSyscallMap) ([]bpf.Instruction, error)
...
+		// jne 0,1
+		bpf.JumpIf{
+			Cond:     bpf.JumpNotEqual,
+			Val:      0,
+			SkipTrue: 1,
+		},
+		// ret [ENOSYS]
+		bpf.RetConstant{Val: retErrnoEnosys},
+	}, programTail...)
+
```
AkihiroSuda, line 548: - This should be appended only for `GOARCH == s390x` ?
- Could you define a `const` for the syscall number of `setup(2)`?

libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -522,10 +522,30 @@ func generateEnosysStub(lastSyscalls lastSyscallMap) ([]bpf.Instruction, error)
...
+	// Return -ENOSYS for setup(2) -- on s390x (and possibly other
+	// platforms) this syscall is used for multiplexing "large syscall
+	// number" syscalls, but if the syscall number is not known to the
+	// kernel then the syscall number is left unchanged (and is thus
+	// sys_num=0). The actual setup(2) syscall is never used by userspace
+	// anymore (and hasn't existed for decades) outside of this multiplexing
+	// scheme so returning -ENOSYS is fine.
+	programTail = append([]bpf.Instruction{
+		// load [0] (syscall number)
+		bpf.LoadAbsolute{Off: 0, Size: 4}, // NOTE: We assume sizeof(int) == 4.
```
AkihiroSuda, line 538: Not new in this PR, but we should have a `const` for the `sizeof(int)`

libcontainer/seccomp/patchbpf/enosys_linux.go
cyphar: 1. That would be one way but I suspect we need the same hack for s390 as well (I'm not sure -- I don't have an s390 box).

I'm currently working on making this addition architecture-independent so that we add this rule for every architecture (since `setup(2)` isn't used by any user code intentionally we should be able to always return `-ENOSYS` even for architectures that don't do this weird multiplexing thing).

cyphar: libseccomp doesn't define setup for any architecture so I just hard-coded it (with a defined const) for s390 and s390x. The only other architecture which has a setup syscall is MIPS (and only on Linux-o32). So this should cover all of our bases.

AkihiroSuda (approved):

cyphar:
Sorry @AkihiroSuda I added a changelog entry, can you re-approve this?

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #3478, opened 2022-05-23
[1.1] seccomp: enosys: always return -ENOSYS for setup(2) on s390(x)

cyphar:
On s390x, syscalls above 255 are multiplexed using the (now otherwise
unused) setup(2) syscall (syscall number 0). If the kernel supports the
syscall then it will correctly translate the syscall number such that
seccomp will correctly detect it -- however, for unknown syscalls the
syscall number remains unchanged. This can be verified by running the
following program under strace:

	int main(void)
	{
		scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_TRAP);
		seccomp_load(ctx);

		return syscall(439, AT_FDCWD, "asdf", X_OK, 0);
	}

Which will then die with the following signal (on pre-5.8 kernels):

	--- SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP,
	            si_call_addr=0x3ffb3006c22, si_syscall=__NR_setup,
	            si_arch=AUDIT_ARCH_S390X} ---

(Note that the si_syscall is __NR_setup, not __NR_faccessat2.)

As a result, the -ENOSYS handling we had previously did not work
completely correctly on s390x because any syscall not supported by the
kernel would be treated as syscall number 0 rather than the actual
syscall number.

Always returning -ENOSYS will not cause any issues because in all of the
cases where this multiplexing occurs, seccomp will see the remapped
syscall number -- and no userspace program will call setup(2)
intentionally (the syscall has not existed in Linux for decades and was
originally a hack used early in Linux init prior to spawning pid1 -- so
you will get -ENOSYS from the kernel anyway).

Backport of #3474 

### DIFF
--- a/libcontainer/seccomp/patchbpf/enosys_linux.go
+++ b/libcontainer/seccomp/patchbpf/enosys_linux.go
@@ -72,6 +72,11 @@ import "C"
 
 var retErrnoEnosys = uint32(C.C_ACT_ERRNO_ENOSYS)
 
+// This syscall is used for multiplexing "large" syscalls on s390(x). Unknown
+// syscalls will end up with this syscall number, so we need to explcitly
+// return -ENOSYS for this syscall on those architectures.
+const s390xMultiplexSyscall libseccomp.ScmpSyscall = 0
+
 func isAllowAction(action configs.Action) bool {
 	switch action {
 	// Trace is considered an "allow" action because a good tracer should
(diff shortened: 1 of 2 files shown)


kolyshkin (approved):
LGTM

AkihiroSuda (approved):

PR #3484, opened 2022-05-24
vendor: bump urfave/cli, add urfave_cli_no_docs tag

kolyshkin:
This removes the runc dependency on cpuguy83/md2man and
russross/blackfriday, which saves more than 400 KB (more than 300 KB
once stripped) from the binary.


cyphar (approved):
lgtm

mrunalp (approved):

Issue #3181, opened 2021-08-25
runc exec is 5x slower than crun exec


kolyshkin:
```console
root@ubu2004:/home/kir/git/runc/tst# time for ((i=0; i<1000;i++)); do runc exec xx45 true; done

real	0m20.726s
user	0m11.635s
sys	0m14.094s
root@ubu2004:/home/kir/git/runc/tst# time for ((i=0; i<1000;i++)); do crun exec xx4 true; done

real	0m4.802s
user	0m2.634s
sys	0m2.283s
```

Need to
- figure out why the difference is so high
- eliminate some of the bottlenecks

kolyshkin:
One thing `runc exec` does and `crun exec` does not is parsing `/etc/passwd` and `/etc/group`. I do not understand why it's needed, since we pass uid:gid as numbers (stringified but still) so the only need for passwd/group lookup is to $HOME value ... which is not used, unless not available from environment.

This needs to be fixed.

AkihiroSuda:
Is this a recent performance regression?

eero-t:
Docker supports passing username/groupname, but it reads them from inside the container, not from host which IMHO makes that feature useless (only reason I can think of for using names is matching host IDs for things that correspond to host, like device files).

cyphar:
> One thing runc exec does and crun exec does not is parsing /etc/passwd and /etc/group. I do not understand why it's needed, since we pass uid:gid as numbers (stringified but still) so the only need for passwd/group lookup is to $HOME value ... which is not used, unless not available from environment.

We fetch the supplementary group IDs from the container even if the user ID specified is numeric, which Docker doesn't do for us IIRC. This is also something we do with the regular config so I'm not sure we can change the behaviour at this late stage. The whole stringification logic is a quirk from when libcontainer used by use directly by Docker, but we probably can't change that either because it's saved to the on-disk state format and changing that leads to very bad issues when upgrading runc.

I'm surprised that crun doesn't do this at all, because it's something AFAIK they need to do in order to have parity with what Docker expects of runc.

> Docker supports passing username/groupname, but it reads them from inside the container, not from host which IMHO makes that feature useless (only reason I can think of for using names is matching host IDs for things that correspond to host, like device files).

[...]

PR #3485, opened 2022-05-25
libct/intelrdt: skip reading /proc/cpuinfo

corhere:
- Relates to #3181 

Reading `/proc/cpuinfo` is a surprisingly expensive operation. [Since kernel version 4.12][1], opening `/proc/cpuinfo` on an x86 system can block for around 20 milliseconds while the kernel samples the current CPU frequency. There is [a very recent patch][2] which gets rid of the delay, but has yet to make it into the mainline kernel. Regardless, kernels for which opening `/proc/cpuinfo` takes 20ms will continue to be run in production for years to come. libcontainer only opens `/proc/cpuinfo` to read the processor feature flags so all the delays to get an accurate snapshot of the CPU frequency are just wasted time.

If we wanted to, we could interrogate the CPU features directly from userspace using the `CPUID` instruction. However, Intel and AMD CPUs have flags in different positions for their analogous sub-features and there are [CPU quirks][3] which would need to be accounted for. Some Haswell server CPUs support RDT/CAT but are missing the `CPUID` flags advertising their support; [the kernel checks for support on that processor family by probing the the hardware using privileged `RDMSR`/`WRMSR` instructions][4]. This sort of probing could not be implemented in userspace so it would not be possible to check for RDT feature support in userspace without false negatives on some hardware configurations.

It looks like libcontainer reads the CPU feature flags as a kind of optimization so that it can skip checking whether the kernel supports an RDT sub-feature if the hardware support is missing. As the kernel only exposes subtrees in the `resctrl` filesystem for RDT sub-features with hardware and kernel support, checking the CPU feature flags is redundant from a correctness point of view. Remove the `/proc/cpuinfo` check as it is an optimization which actually hurts performance.

<details>
<summary>
Benchmarks: <tt>runc run</tt> is 17 ms faster when <tt>"intelRdt"</tt> is enabled in config.json
</summary>

- Hardware: AWS EC2 c5n.metal
- OS: Ubuntu 20.04 LTS
- `/sys/fs/resctrl` mounted
- OCI config:
  - `"args": ["/bin/true"]`
  - `"linux": {"intelRdt": {"memBwSchema": "MB:0=20;1=70"}}`

<table>

<tr>
  <th></th>
  <th>3f0daac9082e422486fdac35b6445f75dd98eb64</th>
  <th>This PR</th>
</tr>

<tr><th>Run 1</th>
<td>

```
real	0m0.098s
user	0m0.029s
sys	0m0.043s
```

</td><td>

```
real	0m0.092s
user	0m0.023s
sys	0m0.048s
```

</td></tr>

<tr><th>Run 2</th>
<td>

```
real	0m0.110s
user	0m0.019s
sys	0m0.052s
```

</td><td>

```
real	0m0.074s
user	0m0.022s
sys	0m0.047s
```

</td></tr>

<tr><th>Run 3</th>
<td>

```
real	0m0.087s
user	0m0.019s
sys	0m0.051s
```

</td><td>

```
real	0m0.077s
user	0m0.025s
sys	0m0.046s
```

</td></tr></table>

</details>

[1]: https://unix.stackexchange.com/a/526679
[2]: https://lore.kernel.org/all/20220415161206.875029458@linutronix.de/
[3]: https://github.com/torvalds/linux/blob/7cf6a8a17f5b134b7e783c2d45c53298faef82a7/arch/x86/kernel/cpu/resctrl/core.c#L834-L851
[4]: https://github.com/torvalds/linux/blob/a6b450573b912316ad36262bfc70e7c3870c56d1/arch/x86/kernel/cpu/resctrl/core.c#L111-L153

### DIFF
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -479,7 +479,7 @@ func (m *Manager) Destroy() error {
 	// Don't remove resctrl group if closid has been explicitly specified. The
 	// group is likely externally managed, i.e. by some other entity than us.
 	// There are probably other containers/tasks sharing the same group.
-	if m.config.IntelRdt == nil || m.config.IntelRdt.ClosID == "" {
+	if m.config.IntelRdt != nil && m.config.IntelRdt.ClosID == "" {
 		m.mu.Lock()
 		defer m.mu.Unlock()
 		if err := os.RemoveAll(m.GetPath()); err != nil {
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -152,9 +152,13 @@ type Manager struct {
 	path   string
 }
 
-// NewManager returns a new instance of Manager, or nil, if the Intel RDT
-// functionality is not available from hardware or not enabled in the kernel.
+// NewManager returns a new instance of Manager, or nil if the Intel RDT
+// functionality is not specified in the config, available from hardware or
+// enabled in the kernel.
 func NewManager(config *configs.Config, id string, path string) *Manager {
+	if config.IntelRdt == nil {
+		return nil
+	}
 	if _, err := Root(); err != nil {
 		// Intel RDT is not available.
 		return nil
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -182,8 +182,6 @@ var (
 	catEnabled bool
 	// The flag to indicate if Intel RDT/MBA is enabled
 	mbaEnabled bool
-	// The flag to indicate if Intel RDT/MBA Software Controller is enabled
-	mbaScEnabled bool
 
 	// For Intel RDT initialization
 	initOnce sync.Once
@@ -216,12 +214,7 @@ func featuresInit() {
 				catEnabled = true
 			}
 		}
-		if mbaScEnabled {
-			// We confirm MBA Software Controller is enabled in step 2,
-			// MBA should be enabled because MBA Software Controller
-			// depends on MBA
-			mbaEnabled = true
-		} else if flagsSet.MBA {
+		if flagsSet.MBA {
 			if _, err := os.Stat(filepath.Join(root, "info", "MB")); err == nil {
 				mbaEnabled = true
 			}
@@ -258,11 +251,6 @@ func findIntelRdtMountpointDir() (string, error) {
 	}
 	if len(mi) < 1 {
 		return "", errNotFound
-	}
-
-	// Check if MBA Software Controller is enabled through mount option "-o mba_MBps"
-	if strings.Contains(","+mi[0].VFSOptions+",", ",mba_MBps,") {
-		mbaScEnabled = true
 	}
 
 	return mi[0].Mountpoint, nil
@@ -491,12 +479,6 @@ func IsCATEnabled() bool {
 func IsMBAEnabled() bool {
 	featuresInit()
 	return mbaEnabled
-}
-
-// Check if Intel RDT/MBA Software Controller is enabled
-func IsMBAScEnabled() bool {
-	featuresInit()
-	return mbaScEnabled
 }
 
 // Get the path of the clos group in "resource control" filesystem that the container belongs to
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -224,7 +224,7 @@ func featuresInit() {
 	})
 }
 
-// Return the mount point path of Intel RDT "resource control" filesystem.
+// findIntelRdtMountpointDir returns the mount point of the Intel RDT "resource control" filesystem.
 func findIntelRdtMountpointDir() (string, error) {
 	mi, err := mountinfo.GetMounts(func(m *mountinfo.Info) (bool, bool) {
 		// similar to mountinfo.FSTypeFilter but stops after the first match
@@ -250,24 +250,32 @@ var (
 	rootOnce        sync.Once
 )
 
+// The kernel creates this (empty) directory if resctrl is supported by the
+// hardware and kernel. The user is responsible for mounting the resctrl
+// filesystem, and they could mount it somewhere else if they wanted to.
+const defaultResctrlMountpoint = "/sys/fs/resctrl"
+
 // Root returns the Intel RDT "resource control" filesystem mount point.
 func Root() (string, error) {
 	rootOnce.Do(func() {
-		// If resctrl is available, kernel creates this directory.
-		if unix.Access("/sys/fs/resctrl", unix.F_OK) != nil {
-			intelRdtRootErr = errNotFound
-			return
-		}
-
-		// NB: ideally, we could just do statfs and RDTGROUP_SUPER_MAGIC check, but
-		// we have to parse mountinfo since we're also interested in mount options.
-		root, err := findIntelRdtMountpointDir()
-		if err != nil {
+		// Does this system support resctrl?
+		var statfs unix.Statfs_t
+		if err := unix.Statfs(defaultResctrlMountpoint, &statfs); err != nil {
+			if errors.Is(err, unix.ENOENT) {
+				err = errNotFound
+			}
 			intelRdtRootErr = err
 			return
 		}
 
-		intelRdtRoot = root
+		// Has the resctrl fs been mounted to the default mount point?
+		if statfs.Type == unix.RDTGROUP_SUPER_MAGIC {
+			intelRdtRoot = defaultResctrlMountpoint
+			return
+		}
+
+		// The resctrl fs could have been mounted somewhere nonstandard.
+		intelRdtRoot, intelRdtRootErr = findIntelRdtMountpointDir()
 	})
 
 	return intelRdtRoot, intelRdtRootErr
--- a/libcontainer/intelrdt/intelrdt.go
+++ b/libcontainer/intelrdt/intelrdt.go
@@ -1,7 +1,6 @@
 package intelrdt
 
 import (
-	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
@@ -199,40 +198,28 @@ func featuresInit() {
 			return
 		}
 
-		// 2. Check if hardware and kernel support Intel RDT sub-features.
-		flagsSet, err := parseCpuInfoFile("/proc/cpuinfo")
+		// 2. Check if Intel RDT sub-features are available in "resource
+		// control" filesystem. Intel RDT sub-features can be
+		// selectively disabled or enabled by kernel command line
+		// (e.g., rdt=!l3cat,mba) in 4.14 and newer kernel
+		if _, err := os.Stat(filepath.Join(root, "info", "L3")); err == nil {
+			catEnabled = true
+		}
+		if _, err := os.Stat(filepath.Join(root, "info", "MB")); err == nil {
+			mbaEnabled = true
+		}
+		if _, err := os.Stat(filepath.Join(root, "info", "L3_MON")); err != nil {
+			return
+		}
+		enabledMonFeatures, err = getMonFeatures(root)
 		if err != nil {
 			return
 		}
-
-		// 3. Double check if Intel RDT sub-features are available in
-		// "resource control" filesystem. Intel RDT sub-features can be
-		// selectively disabled or enabled by kernel command line
-		// (e.g., rdt=!l3cat,mba) in 4.14 and newer kernel
-		if flagsSet.CAT {
-			if _, err := os.Stat(filepath.Join(root, "info", "L3")); err == nil {
-				catEnabled = true
-			}
+		if enabledMonFeatures.mbmTotalBytes || enabledMonFeatures.mbmLocalBytes {
+			mbmEnabled = true
 		}
-		if flagsSet.MBA {
-			if _, err := os.Stat(filepath.Join(root, "info", "MB")); err == nil {
-				mbaEnabled = true
-			}
-		}
-		if flagsSet.MBMTotal || flagsSet.MBMLocal || flagsSet.CMT {
-			if _, err := os.Stat(filepath.Join(root, "info", "L3_MON")); err != nil {
-				return
-			}
-			enabledMonFeatures, err = getMonFeatures(root)
-			if err != nil {
-				return
-			}
-			if enabledMonFeatures.mbmTotalBytes || enabledMonFeatures.mbmLocalBytes {
-				mbmEnabled = true
-			}
-			if enabledMonFeatures.llcOccupancy {
-				cmtEnabled = true
-			}
+		if enabledMonFeatures.llcOccupancy {
+			cmtEnabled = true
 		}
 	})
 }
@@ -284,58 +271,6 @@ func Root() (string, error) {
 	})
 
 	return intelRdtRoot, intelRdtRootErr
-}
-
-type cpuInfoFlags struct {
-	CAT bool // Cache Allocation Technology
-	MBA bool // Memory Bandwidth Allocation
-
-	// Memory Bandwidth Monitoring related.
-	MBMTotal bool
-	MBMLocal bool
-
-	CMT bool // Cache Monitoring Technology
-}
-
-func parseCpuInfoFile(path string) (cpuInfoFlags, error) {
-	infoFlags := cpuInfoFlags{}
-
-	f, err := os.Open(path)
-	if err != nil {
-		return infoFlags, err
-	}
-	defer f.Close()
-
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		line := s.Text()
-
-		// Search "cat_l3" and "mba" flags in first "flags" line
-		if strings.HasPrefix(line, "flags") {
-			flags := strings.Split(line, " ")
-			// "cat_l3" flag for CAT and "mba" flag for MBA
-			for _, flag := range flags {
-				switch flag {
-				case "cat_l3":
-					infoFlags.CAT = true
-				case "mba":
-					infoFlags.MBA = true
-				case "cqm_mbm_total":
-					infoFlags.MBMTotal = true
-				case "cqm_mbm_local":
-					infoFlags.MBMLocal = true
-				case "cqm_occup_llc":
-					infoFlags.CMT = true
-				}
-			}
-			return infoFlags, nil
-		}
-	}
-	if err := s.Err(); err != nil {
-		return infoFlags, err
-	}
-
-	return infoFlags, nil
 }
 
 // Gets a single uint64 value from the specified file.


kolyshkin:
Since commit edeb3b376c5cc763d7afd4c6a63d6e70f8f9992c (#3306) we only read cpuinfo if intelrdt is available in hardware, yet you're right, it does not make sense to read it at all, since we can get the needed info with a few stat(2) calls.

Can we take one more step and ditch mountinfo parsing as well? This is only used to check for mbaScEnabled, which is apparently not used anywhere except to avoid stat of `$root/info/MB` file. Would be nice to remove this.

@xiaochenshen @marquiz PTAL

xiaochenshen:
@corhere Thank you very much for working on it.

> * Relates to [runc exec is 5x slower than crun exec #3181](https://github.com/opencontainers/runc/issues/3181)
> 
> Reading `/proc/cpuinfo` is a surprisingly expensive operation. [Since kernel version 4.12](https://unix.stackexchange.com/a/526679), opening `/proc/cpuinfo` on an x86 system can block for around 20 milliseconds while the kernel samples the current CPU frequency.

I am really surprised for the performance hurt from reading /proc/cpuinfo. This issue should be fixed. Thank you.

> There is [a very recent patch](https://lore.kernel.org/all/20220415161206.875029458@linutronix.de/) which gets rid of the delay, but has yet to make it into the mainline kernel. Regardless, kernels for which opening `/proc/cpuinfo` takes 20ms will continue to be run in production for years to come. libcontainer only opens `/proc/cpuinfo` to read the processor feature flags so all the delays to get an accurate snapshot of the CPU frequency are just wasted time.
> 

Yes, in most of the cases the production kernels or older upstream kernels are without this fix. Reading /proc/cpuinfo is too expensive.
 
> 
> It looks like libcontainer reads the CPU feature flags as a kind of optimization so that it can skip checking whether the kernel supports an RDT sub-feature if the hardware support is missing. As the kernel only exposes subtrees in the `resctrl` filesystem for RDT sub-features with hardware and kernel support, checking the CPU feature flags is redundant from a correctness point of view. Remove the `/proc/cpuinfo` check as it is an optimization which actually hurts performance.
> Benchmarks: runc run is 17 ms faster when "intelRdt" is enabled in config.json

I agree with you. We could remove the redundant check on /proc/cpuinfo.

xiaochenshen (dismissed):
LGTM. Thank you!

xiaochenshen:
@kolyshkin 

> Can we take one more step and ditch mountinfo parsing as well? This is only used to check for mbaScEnabled, which is apparently not used anywhere except to avoid stat of `$root/info/MB` file. Would be nice to remove this.

I don't think the mountinfo check could be removed so far. [MBA Software Controller](https://github.com/opencontainers/runc/pull/1919) is a "pure" software feature by kernel, the only way of exposing the capability to user is through mount option "mba_MBps" in current kernel.

An alternative hacking way is to check "MB:" line in default schemata file: without mount option "mba_MBps", the data is a percentage in range [0, 100%], but with mount option "mba_MBps", the data is a u32 value in range [0, ((u32)~0U)].
```
$ cat /sys/fs/resctrl/schemata | grep MB
(1) w/o "mba_MBps" mount option: "MB:0= 100;1= 100"
(2) w/  "mba_MBps" mount option: "MB:0=4294967295;1=4294967295"
```

corhere:
Given that Intel's own libraries also parse mountinfo to check whether the `mba_MBps` option is enabled, I don't think there are any alternatives to parsing mountinfo or @xiaochenshen's hacky check.
https://github.com/intel/goresctrl/blob/1b61a39df7b692eb89e41a75a27b7f786461012d/pkg/rdt/info.go#L243-L251
https://github.com/intel/intel-cmt-cat/blob/c76f7a34195fba05c96ea3d6450a24a790041ef1/lib/os_cap.c#L745-L746

xiaochenshen:
@corhere 

> I suspect that @xiaochenshen's hacky check might result in false negatives. The `/sys/fs/resctrl/schemata` file describes the configuration for the default resource group so a system with the resctrl filesystem mounted with the `mba_MBps` option could also have the default group's memory bandwidth set to a value <= 100 MBps.
> 

Correct. This hacking way is not reliable. 

> That all being said, it looks like the only other place the `mbaScEnabled` variable is referenced is in the function `intelrdt.IsMBAScEnabled()`, which is unused in libcontainer. So we could elide parsing mountinfo in `rootOnce` without breaking compatibility with external users by lazily parsing mountinfo the first time `IsMBAScEnabled` is called.

Sounds good to me.

> But mountinfo is not only parsed to check if mba_sc is enabled; it is also used to find the mount-point of the resctrl fs. Otherwise libcontainer would need to assume that the resctrl fs is always mounted to `/sys/fs/resctrl`, which is a pretty reasonable assumption IMO and aligns with [Intel's C library](https://github.com/intel/intel-cmt-cat/blob/c76f7a34195fba05c96ea3d6450a24a790041ef1/lib/resctrl.h#L50-L61) (but [not their Go library](https://github.com/intel/goresctrl/blob/1b61a39df7b692eb89e41a75a27b7f786461012d/pkg/rdt/info.go#L102-L106)).

Yes. Parsing mountinfo is necessary to find the mount point of resctrl fs. We could not assume resctrl fs is always mounted to the recommended default mount point - `/sys/fs/resctrl`.
Actually, resctrl fs could be mounted at any proper place instead of `/sys/fs/resctrl`.

kolyshkin:
> intelrdt.IsMBAScEnabled(), which is unused in libcontainer

... and elsewhere (I tried to search through some well known users of runc/libcontainer and haven't found any).

> > But mountinfo is not only parsed to check if mba_sc is enabled; it is also used to find the mount-point of the resctrl fs. Otherwise libcontainer would need to assume that the resctrl fs is always mounted to `/sys/fs/resctrl`, which is a pretty reasonable assumption IMO and aligns with [Intel's C library](https://github.com/intel/intel-cmt-cat/blob/c76f7a34195fba05c96ea3d6450a24a790041ef1/lib/resctrl.h#L50-L61) (but [not their Go library](https://github.com/intel/goresctrl/blob/1b61a39df7b692eb89e41a75a27b7f786461012d/pkg/rdt/info.go#L102-L106)).
> 
> Yes. Parsing mountinfo is necessary to find the mount point of resctrl fs. We could not assume resctrl fs is always mounted to the recommended default mount point - `/sys/fs/resctrl`. Actually, resctrl fs could be mounted at any proper place instead of `/sys/fs/resctrl`.

Well, this is easy -- first we do statfs(2) on default path (`/sys/fs/resctrl`), checking that it exists and comparing the returned fstype (it should be `RDTGROUP_SUPER_MAGIC` as defined in kernel headers as well as x/sys/unix). This is very fast and will work in 99.9% cases.

If not, we fallback to parsing /proc/mountinfo.

Currently we can't do the fast path because unfortunately there's no other way to read mount options (and I've actually added the comment explaining that sad situation in commit 5e201e7ce2c48ce76010d4ab056c7a9a0c9429e2).

If we can get rid of unconditional parsing of mountinfo, that would be awesome.

@corhere @xiaochenshen thanks for looking into it!

corhere:
@kolyshkin @xiaochenshen I have updated the PR with the discussed fast path, plus some more improvements such that the slow path is only executed if RDT is configured in the container's OCI spec. PTAL.

xiaochenshen (dismissed):
LGTM for the fast path updates. Thank you.

marquiz:
I like the idea in this PR very much 😄 I haven't closely reviewed the code yet, though

libcontainer/intelrdt/intelrdt.go
```diff
@@ -201,45 +200,33 @@ func featuresInit() {
...
-		// "resource control" filesystem. Intel RDT sub-features can be
+		// 2. Check if Intel RDT sub-features are available in "resource
+		// control" filesystem. Intel RDT sub-features can be
 		// selectively disabled or enabled by kernel command line
 		// (e.g., rdt=!l3cat,mba) in 4.14 and newer kernel
-		if flagsSet.CAT {
-			if _, err := os.Stat(filepath.Join(root, "info", "L3")); err == nil {
-				catEnabled = true
-			}
+		if _, err := os.Stat(filepath.Join(root, "info", "L3")); err == nil {
```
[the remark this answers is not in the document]
kolyshkin: Yes, it seems this can be addressed in a follow-up. @marquiz feel free to create it :)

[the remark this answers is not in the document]
marquiz: K, I'm fine with this

marquiz (approved):
LGTM

libcontainer/intelrdt/intelrdt.go
```diff
@@ -152,9 +152,13 @@ type Manager struct {
...
 
-// NewManager returns a new instance of Manager, or nil, if the Intel RDT
-// functionality is not available from hardware or not enabled in the kernel.
+// NewManager returns a new instance of Manager, or nil if the Intel RDT
+// functionality is not specified in the config, available from hardware or
+// enabled in the kernel.
 func NewManager(config *configs.Config, id string, path string) *Manager {
+	if config.IntelRdt == nil {
+		return nil
+	}
```
kolyshkin, line 161: This means that the further checks for `m.config.IntelRdt != nil` can be dropped.

Can be addressed as a follow up for sure.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -509,7 +509,7 @@ func (m *Manager) Destroy() error {
 	// Don't remove resctrl group if closid has been explicitly specified. The
 	// group is likely externally managed, i.e. by some other entity than us.
 	// There are probably other containers/tasks sharing the same group.
-	if m.config.IntelRdt == nil || m.config.IntelRdt.ClosID == "" {
+	if m.config.IntelRdt != nil && m.config.IntelRdt.ClosID == "" {
```
kolyshkin, line 516: In a view of the commit that follows this one, the code here can be simplified to
```go
m.config.IntelRdt.ClosID == "" {
```
(surely this also means this commit should be placed after)

kolyshkin:
Left a couple of nits wrt two last commits, but overall LGTM

libcontainer/intelrdt/intelrdt.go
corhere: Is that safe to do? `m.config` is a pointer, so `m.config.IntelRdt` could be potentially set to nil after the manager is constructed. Right now, with all dereferences of `m.config` guarded, doing so would not result in a nil-dereference when calling any manager methods. I worry removing the guards could be a breaking change for some non-runc consumers of `libcontainer/intelrdt`.

kolyshkin: Makes sense; since the config we set is not a copy but merely a pointer, it can be changed. Let's leave these as is.

kolyshkin (dismissed):
LGTM

kolyshkin:
@AkihiroSuda @thaJeztah @cyphar PTAL

kolyshkin:
@AkihiroSuda @thaJeztah @cyphar PTAL

libcontainer/intelrdt/intelrdt.go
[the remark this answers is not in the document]
thaJeztah: Opened https://github.com/opencontainers/runc/issues/3537 so that we don't forget 👍

libcontainer/intelrdt/intelrdt.go
```diff
@@ -495,8 +448,14 @@ func IsMBAEnabled() bool {
...
-	featuresInit()
-	return mbaScEnabled
+	// Neither libcontainer nor runc need to check, and parsing mountinfo is
+	// the only way to check it. Parse it lazily.
+	mi, err := findResctrlMountinfo()
+	if err != nil {
+		return false
+	}
+	// Check if MBA Software Controller is enabled through mount option "-o mba_MBps"
+	return strings.Contains(","+mi.VFSOptions+",", ",mba_MBps,")
```
thaJeztah, line 458: Had to look at this one a few times, as it looked a bit brittle at a glance, but `VFSOptions` comes from mountinfo, so I assume no whitespaces around `,` would exist here

^^ this whole comment would become irrelevant if we decide to remove this function altogether.

libcontainer/intelrdt/intelrdt.go
```diff
@@ -495,8 +448,14 @@ func IsMBAEnabled() bool {
 
 // Check if Intel RDT/MBA Software Controller is enabled
 func IsMBAScEnabled() bool {
```
thaJeztah, line 450: Looks like this function is unused; looks like in the original implementation it was only used in a test; https://github.com/opencontainers/runc/commit/95af9eff82157317817d3d550626f3e5db240d8a#diff-10a177e788e7be5a6826c74946ebbb7390ead3a61f1342a4776dba4d57143942R86

```go
func TestIntelRdtSetMemBwScSchema(t *testing.T) {
 	if !IsMbaScEnabled() {
 		return
 	}
```

But not outside of that (only the related `mbaScEnabled` variable was used)


That check was removed in https://github.com/opencontainers/runc/commit/1b4c30fd8b719834a5492a17893994a996c0a7ce (https://github.com/opencontainers/runc/pull/2920), so it looks unused indeed

Also checked uses outside of runc itself, but it looks like there's none; https://grep.app/search?q=IsMbaScEnabled

thaJeztah: Looking further at this; if we would remove this function, then `findResctrlMountinfo()` is only used in a single place, as part of `func Root()`; which _also_ wraps it in a `sync.Once`;

```
// Root returns the Intel RDT "resource control" filesystem mount point.
 func Root() (string, error) {
 	rootOnce.Do(func() {
```

So in that case, we can remove the `findMountinfoOnce.Do` and associated variables for `findResctrlMountinfo()`

libcontainer/intelrdt/intelrdt.go
```diff
@@ -201,71 +202,62 @@ func featuresInit() {
...
+// for findResctrlMountinfo() use only.
+var (
+	resctrlMountinfo  *mountinfo.Info
+	findMountinfoErr  error
+	findMountinfoOnce sync.Once
+)
+
+// findResctrlMountinfo returns the mount info for the mounted Intel RDT "resource control" filesystem.
+func findResctrlMountinfo() (*mountinfo.Info, error) {
+	findMountinfoOnce.Do(func() {
```
thaJeztah, line 240: See my other comment; we don't need these changes if we remove the unused `IsMbaScEnabled()` function

thaJeztah (commented):
Had a go at reviewing this, and found some unused code that (as a result) could simplify this patch.

libcontainer/intelrdt/intelrdt.go
corhere: First: I didn't write this line. I just moved it around.

I had to look at this one a few times as well. It has nothing to do with whitespace. It's a simple and elegant solution for testing if a comma-separated list contains a value without the false positives which would come with a naive substring match. The list is prepended and appended with commas to eliminate the edge cases of the first and last list items.

thaJeztah: > The list is prepended and appended with commas to eliminate the edge cases of the first and last list items.

Yup, got that one; was mostly wondering at first how strict `VFSOptions` formatting itself was, i.e. if

`","+mi.VFSOptions+","` could be `","+"some, options, mba_MBps , other_option"+","`, so if it could have whitespace in the middle.

corhere: Ah. The list of options is synthesized by the kernel](https://github.com/torvalds/linux/blob/efb52a7d9511df818391f1afa459507425833438/arch/x86/kernel/cpu/resctrl/rdtgroup.c#L3184-L3196) when the mountinfo file is opened so no whitespace is possible

thaJeztah: Ah! Thx.

In either case, if we don't need this function, we don't have to worry about it at all 😄 

@kolyshkin @marquiz any reason we'd want to keep it? It looks unused, and would save a lot of extra complexity that's not needed if it's gone.

libcontainer/intelrdt/intelrdt.go
corhere: The other reviewers also noted that `IsMBAScEnabled` is not used in searchable public source code. I am not convinced it is proof enough that the function can be deleted without breaking anybody's code. Intel RDT is very much an enterprise datacenter feature so it's entirely possible that some company's private code is consuming it. Could we defer the removal of this function to a follow-up PR?

libcontainer/intelrdt/intelrdt.go
kolyshkin: `IsMBAScEnabled` first appeared in commit 95af9eff82157317817d3d550626f3e5db240d8a (as `IsMbaScEnabled`) and, despite being a public function, was only used by intelrdt unit tests. This usage was dropped by commit 1b4c30fd8b719834a5492a17893994a996c0a7ce, and since then this function is unused.

Yes, @thaJeztah is right, it should be removed in the first commit, simplifying further changes.

@corhere please 🙏🏻

corhere: @thaJeztah @kolyshkin will do 👍

libcontainer/intelrdt/intelrdt.go
thaJeztah: I think the relevant point here is that `libcontainer/` is considered not to be for external consumption; there's some remaining uses of it outside of runc itself (most of that for historical reasons), but the intent is to better defined "externally consumable" packages (if any) and "internal" packages; see https://github.com/opencontainers/runc/issues/3028

libcontainer/intelrdt/intelrdt.go
```diff
@@ -250,24 +250,32 @@ var (
...
 		}
 
-		// NB: ideally, we could just do statfs and RDTGROUP_SUPER_MAGIC check, but
-		// we have to parse mountinfo since we're also interested in mount options.
-		root, err := findIntelRdtMountpointDir()
-		if err != nil {
-			intelRdtRootErr = err
+		// Has the resctrl fs been mounted to the default mount point?
+		if statfs.Type == unix.RDTGROUP_SUPER_MAGIC {
+			intelRdtRoot = defaultResctrlMountpoint
```
thaJeztah, line 277: heh (very!) random comment 😅; I recall some "lunch time" discussions I had with @kolyshkin about "where" comments should go for conditional statements.

My take was that putting the comment _before_ the condition is an easy pitfal to describe the _code_, not the _intent_ (and you may end up with describing the "if" statement, only in words instead of code), whereas putting it inside the branch "gently" forces one to write the branch itself (_why_ we're doing it).

```go
// check if foo is enabled
if foo {
    do stuff
}
```

```go
if foo {
  // foo is enabled; start the flux capacitor to go back to the future.
  do stuff
}
```

(Of course, it highly depends on the situation; in some cases you'd want to describe a more complex conditional block _as a whole_ to describe the whole intent.)

thaJeztah (approved):
LGTM, thanks!

libcontainer/intelrdt/intelrdt.go
corhere: Ahh, I couldn't remember if it was all of libcontainer or just certain subpackages

thaJeztah: Yes, I think the docs / readme should probably more clear on that once we settle "what bits" (some bits are a bit of a "grey" area though)

thaJeztah:
@kolyshkin PTAL

kolyshkin (approved):
LGTM, thanks!

PR #3486, opened 2022-05-26
ci: shellcheck job nits

kolyshkin:
Use env directive instead of adding to $GITHUB_ENV.

Use herefile to feed sha256sum instead of pipe to grep (inspired by https://github.com/opencontainers/runc/pull/3482#pullrequestreview-983894532)


.github/workflows/validate.yml
```diff
@@ -90,24 +90,23 @@ jobs:
...
         run: |
           mkdir ~/bin
           curl -sSfL --retry 5 $BASEURL/$VERSION/shellcheck-$VERSION.linux.x86_64.tar.xz |
             tar xfJ - -C ~/bin --strip 1 shellcheck-$VERSION/shellcheck
-          sha256sum ~/bin/shellcheck | grep -q $SHA256SUM
+          sha256sum --strict --check - <<<"$SHA256 *$HOME/bin/shellcheck"
           # make sure to remove the old version
           sudo rm -f /usr/bin/shellcheck
+          # Add ~/bin to $PATH.
+          echo ~/bin >> $GITHUB_PATH
```
tianon, line 106: Ironically, I bet `shellcheck` would prefer if this `$GITHUB_PATH` were quoted (but this is how GitHub themselves put it in examples, so I doubt they'll ever break that with something like a space :smile:).

kolyshkin: Right ) that's quite ironical :) and yet I'd rather let it stay as is, since this is  what they do in their examples.

kolyshkin:

.github/workflows/validate.yml
thaJeztah: We should update the examples! 🙃

thaJeztah (approved):
LGTM

PR #3498, opened 2022-06-02
cgroups: systemd: skip adding device paths that don't exist

cyphar:
systemd emits very loud warnings when the path specified doesn't exist
(which can be the case for some of our default rules). We don't need the
ruleset we give systemd to be completely accurate (we discard some kinds
of wildcard rules anyway) so we can safely skip adding these.


kolyshkin (approved):
LGTM

kolyshkin:
@thaJeztah @AkihiroSuda PTAL

thaJeztah (approved):
LGTM

PR #3501, opened 2022-06-05
libct/nsenter: switch to sane_kill()

jellor:
kill(stage2_pid, SIGKILL) maybe will kill all processes except the process itself and init when the stage2_pid equals -1.

Signed-off-by: Guodong Zhu <guodong9211@gmail.com>

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -1068,7 +1068,7 @@ void nsexec(void)
 
 					s = SYNC_MOUNTSOURCES_ACK;
 					if (write(syncfd, &s, sizeof(s)) != sizeof(s)) {
-						kill(stage1_pid, SIGKILL);
+						sane_kill(stage1_pid, SIGKILL);
 						bail("failed to sync with child: write(SYNC_MOUNTSOURCES_ACK)");
 					}
 					break;
@@ -1230,7 +1230,7 @@ void nsexec(void)
 			if (config.mountsources) {
 				s = SYNC_MOUNTSOURCES_PLS;
 				if (write(syncfd, &s, sizeof(s)) != sizeof(s)) {
-					kill(stage2_pid, SIGKILL);
+					sane_kill(stage2_pid, SIGKILL);
 					bail("failed to sync with parent: write(SYNC_MOUNTSOURCES_PLS)");
 				}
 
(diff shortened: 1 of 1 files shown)


cyphar:
We shouldn't reach those `kill` sections if the pid is -1, but even if that was something that happened the correct patch would check for `pid <= 0` rather than removing the kill calls. It is necessary to send a kill signal in these cases because we don't want the child process to stay alive if there is an issue during setup (we can't use `PDEATHSIG` since we use `CLONE_PARENT`).

cyphar (changes requested):
We shouldn't reach those `kill` sections if the pid is -1, but even if that was something that happened the correct patch would be to switch all of the `kill` calls to `sane_kill` (which checks for `pid >= 0`) rather than removing the `kill` calls. It is necessary to send a kill signal in these cases because we don't want the child process to stay alive if there is an issue during setup (we can't use `PDEATHSIG` since we use `CLONE_PARENT`).

jellor:
> @jellor do you have some kind of a repro?

@kolyshkin One case like using BCC script (https://gist.github.com/jellor/1f2fe1147ca63c031d729dfdd2ede065) to override `write(SYNC_MOUNTSOURCES_PLS)` return value, then it will reach `kill(stage2_pid = -1, SIGKILL)` call. NB: it can send SIGKILL to every process but except the process itself and init.

> We shouldn't reach those `kill` sections if the pid is -1, but even if that was something that happened the correct patch would be to switch all of the `kill` calls to `sane_kill` (which checks for `pid >= 0`) rather than removing the `kill` calls. It is necessary to send a kill signal in these cases because we don't want the child process to stay alive if there is an issue during setup (we can't use `PDEATHSIG` since we use `CLONE_PARENT`).

Yes, switch `kill` calls to `sane_kill`  (which checks for `pid > 0`) can resolve this issue.

kolyshkin:
Commit subject (and PR subject) should be modified accordingly (`s/remove kill/switch to sane_kill/` or something like that).

jellor:
@cyphar PTAL, thanks!

jellor:
@kolyshkin What should I need to do?

cyphar (approved):
LGTM.

cyphar:
@jellor  You create a PR against `release-1.1` with the patch cherry-picked (`git cherry-pick`).

cyphar:
> One case like using BCC script (https://gist.github.com/jellor/1f2fe1147ca63c031d729dfdd2ede065) to override write(SYNC_MOUNTSOURCES_PLS) return value, then it will reach kill(stage2_pid = -1, SIGKILL) call

While I'm fine with this patch (since `sane_kill` is the reasonable thing to do anyway), I highly doubt it's actually possible for the `write` to fail in practice because it's a small write on a pipe.

PR #3512, opened 2022-06-16
Refactor mountFd code

kolyshkin:
This is a followup to #3510, doing some refactoring of the code introduced by #2576.

This does the following:

1. Simplify `mount` call by removing the procfd argument, and use the new
   simplified mount where procfd is not used. Now, the mount arguments are
   the same as those of `unix.Mount`.

2. Introduce a new `mountWithFD` function, which is similar to the old
   mount(), except it can take procfd for both source and target.
   The new arguments are called `srcFD` and `dstFD`.

3. Modify the mount error to show both `srcFD` and `dstFD` so it's clear
   which one is used for which purpose. This fixes the issue of having
   a somewhat cryptic errors like this:

   > mount /proc/self/fd/11:/sys/fs/cgroup/systemd (via /proc/self/fd/12), flags: 0x20502f: operation not permitted
 
   (in which fd 11 is actually the source, and fd 12 is the target).

4. Fix the `mountWithFD` callers to use `dstFD` (rather than `procfd`) for the
   variable name.

5. Use `srcFD` where `mountFd` is set.

6. Remove fd field from the `mountConfig` (since it contains data not specific to
   a particular mount, while fd is a per mount entry attribute). Introduce `mountEntry`
   structure, which embeds `configs.Mount` and adds `srcFd` to replace the
   removed `mountConfig.fd`.

### DIFF
--- a/libcontainer/mount_linux.go
+++ b/libcontainer/mount_linux.go
@@ -10,8 +10,9 @@ import (
 type mountError struct {
 	op     string
 	source string
+	srcFD  string
 	target string
-	procfd string
+	dstFD  string
 	flags  uintptr
 	data   string
 	err    error
@@ -22,19 +23,21 @@ func (e *mountError) Error() string {
 	out := e.op + " "
 
 	if e.source != "" {
-		out += e.source + ":" + e.target
-	} else {
-		out += e.target
+		out += "src=" + e.source + ", "
+		if e.srcFD != "" {
+			out += "srcFD=" + e.srcFD + ", "
+		}
 	}
-	if e.procfd != "" {
-		out += " (via " + e.procfd + ")"
+	out += "dst=" + e.target
+	if e.dstFD != "" {
+		out += ", dstFD=" + e.dstFD
 	}
 
 	if e.flags != uintptr(0) {
-		out += ", flags: 0x" + strconv.FormatUint(uint64(e.flags), 16)
+		out += ", flags=0x" + strconv.FormatUint(uint64(e.flags), 16)
 	}
 	if e.data != "" {
-		out += ", data: " + e.data
+		out += ", data=" + e.data
 	}
 
 	out += ": " + e.err.Error()
@@ -47,19 +50,36 @@ func (e *mountError) Unwrap() error {
 	return e.err
 }
 
-// mount is a simple unix.Mount wrapper. If procfd is not empty, it is used
-// instead of target (and the target is only used to add context to an error).
-func mount(source, target, procfd, fstype string, flags uintptr, data string) error {
-	dst := target
-	if procfd != "" {
-		dst = procfd
+// mount is a simple unix.Mount wrapper, returning an error with more context
+// in case it failed.
+func mount(source, target, fstype string, flags uintptr, data string) error {
+	return mountViaFDs(source, "", target, "", fstype, flags, data)
+}
+
+// mountViaFDs is a unix.Mount wrapper which uses srcFD instead of source,
+// and dstFD instead of target, unless those are empty. The *FD arguments,
+// if non-empty, are expected to be in the form of a path to an opened file
+// descriptor on procfs (i.e. "/proc/self/fd/NN").
+//
+// If case an FD is used instead of a source or a target path, the
+// corresponding path is only used to add context to an error in case
+// the mount operation has failed.
+func mountViaFDs(source, srcFD, target, dstFD, fstype string, flags uintptr, data string) error {
+	src := source
+	if srcFD != "" {
+		src = srcFD
 	}
-	if err := unix.Mount(source, dst, fstype, flags, data); err != nil {
+	dst := target
+	if dstFD != "" {
+		dst = dstFD
+	}
+	if err := unix.Mount(src, dst, fstype, flags, data); err != nil {
 		return &mountError{
 			op:     "mount",
 			source: source,
+			srcFD:  srcFD,
 			target: target,
-			procfd: procfd,
+			dstFD:  dstFD,
 			flags:  flags,
 			data:   data,
 			err:    err,
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -28,13 +28,26 @@ import (
 
 const defaultMountFlags = unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NODEV
 
+// mountConfig contains mount data not specific to a mount point.
 type mountConfig struct {
 	root            string
 	label           string
 	cgroup2Path     string
 	rootlessCgroups bool
 	cgroupns        bool
-	fd              *int
+}
+
+// mountEntry contains mount data specific to a mount point.
+type mountEntry struct {
+	*configs.Mount
+	srcFD string
+}
+
+func (m *mountEntry) src() string {
+	if m.srcFD != "" {
+		return m.srcFD
+	}
+	return m.Source
 }
 
 // needsSetupDev returns true if /dev needs to be set up.
@@ -67,21 +80,20 @@ func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig, mountFds []int) (err
 		rootlessCgroups: iConfig.RootlessCgroups,
 		cgroupns:        config.Namespaces.Contains(configs.NEWCGROUP),
 	}
-	setupDev := needsSetupDev(config)
 	for i, m := range config.Mounts {
+		entry := mountEntry{Mount: m}
 		// Just before the loop we checked that if not empty, len(mountFds) == len(config.Mounts).
 		// Therefore, we can access mountFds[i] without any concerns.
 		if mountFds != nil && mountFds[i] != -1 {
-			mountConfig.fd = &mountFds[i]
-		} else {
-			mountConfig.fd = nil
+			entry.srcFD = "/proc/self/fd/" + strconv.Itoa(mountFds[i])
 		}
 
-		if err := mountToRootfs(m, mountConfig); err != nil {
+		if err := mountToRootfs(mountConfig, entry); err != nil {
 			return fmt.Errorf("error mounting %q to rootfs at %q: %w", m.Source, m.Destination, err)
 		}
 	}
 
+	setupDev := needsSetupDev(config)
 	if setupDev {
 		if err := createDevices(config); err != nil {
 			return fmt.Errorf("error creating device nodes: %w", err)
@@ -201,12 +213,8 @@ func cleanupTmp(tmpdir string) {
 	_ = os.RemoveAll(tmpdir)
 }
 
-func prepareBindMount(m *configs.Mount, rootfs string, mountFd *int) error {
-	source := m.Source
-	if mountFd != nil {
-		source = "/proc/self/fd/" + strconv.Itoa(*mountFd)
-	}
-
+func prepareBindMount(m mountEntry, rootfs string) error {
+	source := m.src()
 	stat, err := os.Stat(source)
 	if err != nil {
 		// error out if the source of a bind mount does not exist as we will be
@@ -252,7 +260,7 @@ func mountCgroupV1(m *configs.Mount, c *mountConfig) error {
 		PropagationFlags: m.PropagationFlags,
 	}
 
-	if err := mountToRootfs(tmpfs, c); err != nil {
+	if err := mountToRootfs(c, mountEntry{Mount: tmpfs}); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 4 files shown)


libcontainer/mount_linux.go
```diff
@@ -10,8 +10,9 @@ import (
 type mountError struct {
```
rata, line 10: Commenting here as I can't comment on github commit description. In the commit you showed how it looked like before, but can you show how it looks now?

I can't think of a simple way add a test for a clear error too that will not need a refactor every time we change this. Probably it is not worth it adding a test.

libcontainer/mount_linux.go
```diff
@@ -26,8 +27,8 @@ func (e *mountError) Error() string {
 	} else {
 		out += e.target
 	}
-	if e.procfd != "" {
-		out += " (via " + e.procfd + ")"
+	if e.srcFD != "" || e.dstFD != "" {
+		out += " (via " + e.srcFD + ":" + e.dstFD + ")"
```
rata, line 31: I'm definitely curious to see how this error is shown now.

I wonder how much better this is than a something prints `%+v`, like `fmt.Sprintf("Error: %v trying to: %+v", e.err.Error(), e)`.

I guess probably what we do here is better, though

rata, line 31: Isn't this line confusing if `e.srcFD` is empty?

libcontainer/mount_linux.go
```diff
@@ -47,19 +48,31 @@ func (e *mountError) Unwrap() error {
...
+// mount is a simple unix.Mount wrapper, returning an error with more context
+// in case it failed.
+func mount(source, target, fstype string, flags uintptr, data string) error {
+	return mountWithFD(source, "", target, "", fstype, flags, data)
+}
+
+// mountWithFD is a unix.Mount wrapper which uses srcFD instead of source, and
+// dstFD instead of target, unless those are empty.  In such cases, source and
+// target are only used to add context in case the mount failed.
+func mountWithFD(source, srcFD, target, dstFD, fstype string, flags uintptr, data string) error {
```
rata, line 60: I wonder if we want a `mountWithDstFD()` in addition, as we use that a lot too. No strong opinion

rata, line 60: Maybe this should be `mountWithFDs()`?

rata (commented):
@kolyshkin This looks like a nice improvement, thanks! Left some comments

libcontainer/mount_linux.go
kolyshkin: I thought about it and decided it's too much.

One other way to solve this would be to refactor this to use a structure instead of list of parameters, but again I ended up not doing it.

kolyshkin: This belongs to #3510.

libcontainer/mount_linux.go
kolyshkin: I think it's understandable enough (`via :/proc/self/fd/5`) yet simple.

Before:
```
mount /proc/self/fd/11:/sys/fs/cgroup/systemd (via /proc/self/fd/12), flags: 0x20502f: operation not permitted
```

After:
```
mount /proc/self/fd/11:/sys/fs/cgroup/systemd (via :/proc/self/fd/12), flags: 0x20502f: operation not permitted
```

Or would you prefer something like this (untested):
```go
if e.srcFD != "" || e.dstFD != "" {
	const none="<none>"
	s := e.srcFD
	if s == "" {
		s = none
	}
	d := e.dstFD
	if d = "" {
		d  = none
	}
	out += " (via " + s + ":" + d + ")"
```

kolyshkin: If you can think of a better way to represent the error, let me know, I tried a few alternatives and did not like them.

kolyshkin: This would look like

```
Error: no such file or directory trying to: {op:mount source:/proc/self/fd/11 srcFD: target:/sys/fs/cgroup/systemd dstFD:/proc/self/fd/12 flags:0x20502f data: err:2}
```
and using the code in this PR looks like
```
mount /proc/self/fd/11:/sys/fs/cgroup/systemd (via :/proc/self/fd/12), flags: 0x20502f: operation not permitted
```

kolyshkin:
Rebased, addressed @rata comments, PTAL

libcontainer/mount_linux.go
rata: Heh, yeah, not sure I like other alternatives I'm thinking either. What the PR currently does is good enough for me, thanks! :)

libcontainer/mount_linux.go
rata: SGTM. Thanks!

rata (approved):
LGTM, thanks @kolyshkin !

kolyshkin:

libcontainer/mount_linux.go
kolyshkin: Commit amended to show both before and after error message.

rata: Thanks!

libcontainer/mount_linux.go
kolyshkin: I took a fresh look and ended up implementing a different format of an error message, using `src=`, `sfd=`, `dst=`, and `dfd=`, similar to what @thaJeztah suggested.

It makes some error messages look more heavy-weight than they should be, for example:

> umount /some/dir, flags: 0x1: device busy

vs

> umount dst=/some/dir, flags=0x1: device busy

but, I guess, it's even more clear that way

kolyshkin: Redid it again, see https://github.com/opencontainers/runc/pull/3512#discussion_r921532761

kolyshkin:
@rata @thaJeztah PTAL

libcontainer/mount_linux.go
```diff
@@ -47,19 +50,31 @@ func (e *mountError) Unwrap() error {
...
-// instead of target (and the target is only used to add context to an error).
-func mount(source, target, procfd, fstype string, flags uintptr, data string) error {
+// mount is a simple unix.Mount wrapper, returning an error with more context
+// in case it failed.
+func mount(source, target, fstype string, flags uintptr, data string) error {
+	return mountViaFDs(source, "", target, "", fstype, flags, data)
+}
+
+// mountViaFDs is a unix.Mount wrapper which uses srcFD instead of source, and
+// dstFD instead of target, unless those are empty. In such cases, source and
```
rata, line 60: IMHO, it seems worth documenting that srcFD and dstFD are a string and expected to have the format `/proc/self/fd/<int>` or something like that.

That is not obvious for a user that an fd is of type string and should be in that format, I think. It can be deduced by looking at how the code is used and it is arguable if before it was clearer because the var name was "procfd". I do think, though, that a small documentation sentence can help :)

What do you think?

rata (commented):
@kolyshkin Thanks! I left a few nits (super nits, so feel free to ignore). But this LGTM, is a nice improvement :)

libcontainer/mount_linux.go
```diff
@@ -22,19 +23,21 @@ func (e *mountError) Error() string {
...
-	if e.procfd != "" {
-		out += " (via " + e.procfd + ")"
+	out += "dst=" + e.target
+	if e.dstFD != "" {
+		out += ", dfd=" + e.dstFD
 	}
 
 	if e.flags != uintptr(0) {
-		out += ", flags: 0x" + strconv.FormatUint(uint64(e.flags), 16)
+		out += ", flags=0x" + strconv.FormatUint(uint64(e.flags), 16)
```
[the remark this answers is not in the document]
kolyshkin: Yeah, thought about it, too, decided to not implement it, since it is only used by developers, and we can cope with numeric values.

libcontainer/mount_linux.go
```diff
@@ -22,19 +23,21 @@ func (e *mountError) Error() string {
 	out := e.op + " "
 
 	if e.source != "" {
-		out += e.source + ":" + e.target
-	} else {
-		out += e.target
+		out += "src=" + e.source + ", "
+		if e.srcFD != "" {
+			out += "sfd=" + e.srcFD + ", "
```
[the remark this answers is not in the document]
kolyshkin: What you propose will result it `mount , dst=...` in case src is empty, which looks a tad weird. This is why it's done that way now.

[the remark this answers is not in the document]
kolyshkin: Changed to `srcFD=` and `dstFD=`.

libcontainer/mount_linux.go
```diff
@@ -47,19 +50,31 @@ func (e *mountError) Unwrap() error {
...
-func mount(source, target, procfd, fstype string, flags uintptr, data string) error {
+// mount is a simple unix.Mount wrapper, returning an error with more context
+// in case it failed.
+func mount(source, target, fstype string, flags uintptr, data string) error {
+	return mountViaFDs(source, "", target, "", fstype, flags, data)
+}
+
+// mountViaFDs is a unix.Mount wrapper which uses srcFD instead of source, and
+// dstFD instead of target, unless those are empty. In such cases, source and
+// target are only used to add context in case the mount failed.
```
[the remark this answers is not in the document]
kolyshkin: Reworded to make it clearer.

libcontainer/mount_linux.go
kolyshkin: It's clear from the function proto that FDs are strings, but it makes sense to point out they are like "/proc/self/fd/N".

Added.

kolyshkin:
@rata addressed your comments, PTAL.

kolyshkin:
CI is broken until https://github.com/opencontainers/runc/pull/3539 is merged

libcontainer/mount_linux.go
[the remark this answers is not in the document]
rata: Yeap, I ended up doing the same :-D. Some day it will happen :)

rata (approved):
LGTM, thanks @kolyshkin!

cyphar (dismissed):
LGTM.

kolyshkin:
@AkihiroSuda PTAL

kolyshkin:
@cyphar PTAL. You've LGTMed this before; I don't remember what were the changes since your LGTM but I think those are either a rebase or some cosmetics (like a typo in the commit message).

kolyshkin:
@thaJeztah @cyphar PTAL

kolyshkin:
@opencontainers/runc-maintainers anything I can do to move this forward?

AkihiroSuda (dismissed):

kolyshkin:
@thaJeztah @cyphar PTAL

AkihiroSuda (approved):

kolyshkin:
@thaJeztah @cyphar @hqhq PTAL 🙏🏻

libcontainer/rootfs_linux.go
```diff
@@ -1037,28 +1037,28 @@ func writeSystemProperty(key, value string) error {
...
-		err := mount(source, m.Destination, procfd, m.Device, flags, "")
+		err := mountViaFDs(m.Source, srcFD, m.Destination, dstFD, m.Device, flags, "")
 		if err == nil {
 			return nil
 		}
 		// Check if the source has ro flag...
 		var s unix.Statfs_t
-		if err := unix.Statfs(source, &s); err != nil {
-			return &os.PathError{Op: "statfs", Path: source, Err: err}
+		if err := unix.Statfs(m.Source, &s); err != nil {
```
hqhq, line 1052: path argument of `unix.Statfs` used to be proc fd path if `mountFd != nil`, is this gonna be a problem?

kolyshkin: Good question. I think it might be. Let me fix this.

kolyshkin: OK I was not able to repro the error that might happen here (as this requires a very narrow scenario that uses a bind mount of a read-only fuse-sshfs mount), but I fixed the issue so the statfs() uses either source or srcFD.

PTAL @hqhq

hqhq (approved):
LGTM

rata (approved):
LGTM

It was so long that I had to review again. It was mostly unchaged, but I think the mountEntry.src() method is new and makes a lot of sense :)

kolyshkin:
Rebased; now I need to re-review it myself.

kolyshkin:
> Rebased; now I need to re-review it myself.

LGTM

kolyshkin:
@thaJeztah @AkihiroSuda PTAL

AkihiroSuda (approved):

PR #3514, opened 2022-06-16
Add go 1.19, require go 1.18, drop go 1.17

kolyshkin:
This makes go 1.18 a minimally required version, removing go 1.17 from and adding go 1.19 to CI.

Also:
 - [x] remove old-style `+build` tags
 - [x] fix the docstrings to be compatible with go 1.19 (see https://tip.golang.org/doc/comment)
 - [x] wait for golangci-lint release that fixes https://github.com/golangci/golangci-lint/issues/2922
 - [x] remove `test (1.17.x)` from the required set of tests
 - [ ] add `test (1.19.x)` to required list of tests (once this is merged)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.17.x, 1.18.x]
+        go-version: [1.18.x, 1.19.x]
         rootless: ["rootless", ""]
         race: ["-race", ""]
         criu: [""]
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -32,7 +32,7 @@ jobs:
           sudo apt -q install libseccomp-dev
       - uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.46
+          version: v1.48
       # Extra linters, only checking new code from a pull request.
       - name: lint-extra
         if: github.event_name == 'pull_request'
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -8,7 +8,7 @@ on:
       - release-*
   pull_request:
 env:
-  GO_VERSION: 1.18.x
+  GO_VERSION: 1.19.x
 permissions:
   contents: read
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ A third party security audit was performed by Cure53, you can see the full repor
 
 ## Building
 
-`runc` only supports Linux. It must be built with Go version 1.17 or higher.
+`runc` only supports Linux. It must be built with Go version 1.18 or higher.
 
 In order to enable seccomp support you will need to install `libseccomp` on your platform.
 > e.g. `libseccomp-devel` for CentOS, or `libseccomp-dev` for Ubuntu
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/opencontainers/runc
 
-go 1.17
+go 1.18
 
 require (
 	github.com/checkpoint-restore/go-criu/v5 v5.3.0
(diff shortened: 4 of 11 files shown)


kolyshkin:
OK, it seems to be working fine, except for golangci-lint (filed https://github.com/golangci/golangci-lint/issues/2922)

AkihiroSuda:
1.19 is GA now

kolyshkin:
Alas we have to wait for a new golangci-lint release that has https://github.com/golangci/golangci-lint/issues/2922 fixed. The fix is already merged but no release has been made yet (I guess it will be v1.48).

kolyshkin:
Since I've already marked 1.19.x CI tests as required, can we please merge this? @opencontainers/runc-maintainers PTAL

libcontainer/apparmor/apparmor_unsupported.go
```diff
@@ -1,5 +1,4 @@
 //go:build !linux
-// +build !linux
```
thaJeztah, line 2: I'm on the fence on already dropping these, as we know libcontainer is still used by external projects, and those may not yet have updated to go 1.18+ (I saw there was some discussion among containerd maintainers as well)

I guess it slightly depends on which release this will get into, and overall it seems ok to (start) dropping 1.17 in CI, but just on the fence on doing a hard _break_ for Go < 1.18

kolyshkin: The tags removal is done deliberately so the codebase won't work with older golang. This way we can start using new(ish) features of golang 1.18.

I was more conservative about it in the past, supporting three golang releases (including the latest unsupported version), but now when we have branches, we can always say "if you want older golang, use older runc/libcontainer). To me it seems a nice balance between using new golang features and supporting older releases.

So, this code is in main, meaning it's going to be 1.2.0, which we will hopefully release in September/October. From what I see, k8s (I guess this is the biggest user of runc) is switching to newest golang very aggressively.

kolyshkin: > overall it seems ok to (start) dropping 1.17 in CI, but just on the fence on doing a hard break for Go < 1.18

Doesn't work this way. Once we drop Go 1.17 and someone will add e.g. use of `strings.Cut`, we'll never know that the code is broken for go < 1.18 until someone will use it (after the release). So it's either we support it (and test it in CI), or we don't (and we make this deliberate). Having something in the middle doesn't seem like a good way to me.

kolyshkin: @thaJeztah PTAL ^^

kolyshkin: @AkihiroSuda what do you think?

thaJeztah: My main concern is that `<unknown>` project or distro may still be using older Go versions (I recall discussions about Debian etc not updating).

And while we no longer test on those versions, making things a fully breaking change (not even possible to compile) a bit harsh, and _possibly_ even would require a major version update (from a SemVer perspective); I know Golang's own packages, such as `golang.org/x/sys` continue to include both old and new-style build-tags; https://github.com/golang/sys/blob/1c4a2a72c664c42691e45fe3a31510ec5ba13cfb/unix/str.go#L5-L6

kolyshkin: Removed the last commit; PTAL @thaJeztah @AkihiroSuda

thaJeztah (approved):
LGTM

kolyshkin:
@AkihiroSuda PTAL 🙏🏻

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

thaJeztah:
@kolyshkin are you updating the repo settings? ("add test (1.19.x) to required list of tests (once this is merged)")

PR #3518, opened 2022-06-20
libct: Don't use stale mount fds

rata:
This is an alternative fix for the same bug that PR https://github.com/opencontainers/runc/pull/3510 is tackling. Both PRs should have the same effect (the .fd field is nil when it should).

I'm just opening this PR so people can chose the PR they prefer. Both are correct and easy to backport.

What I like about this is that similar bugs should not happen, as the mountConfig is cleared on every iteration, and should be very simple to backport too.

Please keep me in the loop in the future too, so I can help to fix the bugs that come around :)


rata:
I've also verified that the test passes with the fix and fails without (see the revert commit that I removed now)

kolyshkin:
I think that #3512 is a better way long term

rata:
Sure, closing then!

PR #3521, opened 2022-06-27
refactor: move main package into ./cmd/runc

junnplus:
Signed-off-by: Ye Sijun <junnplus@gmail.com>


junnplus:
@AkihiroSuda @cyphar @kolyshkin any ideas?

kolyshkin:
Not sure if it helps anything. We are not going to add more commands...

thaJeztah:
Yeah, from a go modules perspective, this wouldn't bring benefits, unless it was its own module (which is why we did something like this in "library" repositories, such as https://github.com/containerd/continuity/pull/200)

For runc, it's actually the reverse; the binary is the main project, and some packages (libcontainer) are effectively a "side product" that happen to be used by some external consumers.

Moving this to cmd/ will actually be a breaking change, as "go install github.com/opencontainers/runc@version" will no longer work.

junnplus:
OK, let's leave it as is.

Issue #3520, opened 2022-06-24
runc run failure logging when the binary is on noexec fs


kolyshkin:
When `runc run` is called by podman, podman gathers run output, and if it contains "permission denied", exits with code of 126.

As demonstrated:
```console
[kir@kir-rhat podman]$ ./bin/podman run --runtime=`pwd`/../../opencontainers/runc/runc busybox /etc
Error: /home/kir/go/src/github.com/containers/podman/../../opencontainers/runc/runc: runc create failed: unable to start container process: exec: "/etc": permission denied: OCI permission denied
[kir@kir-rhat podman]$ echo $?
126
```

Now, for some reason yet unknown to me, if the non-executable binary is on volume, runc fails to propagate the error back to podman.

Here's a repro with a slightly modified runc:

```diff
diff --git a/libcontainer/factory_linux.go b/libcontainer/factory_linux.go
index 17b05a55..2bed4846 100644
--- a/libcontainer/factory_linux.go
+++ b/libcontainer/factory_linux.go
@@ -175,10 +175,12 @@ func StartInitialization() (err error) {
                // We have an error during the initialization of the container's init,
                // send it back to the parent process in the form of an initError.
                if werr := writeSync(pipe, procError); werr != nil {
+                       fmt.Fprintln(os.Stderr, "init: write procError:", werr)
                        fmt.Fprintln(os.Stderr, err)
                        return
                }
                if werr := utils.WriteJSON(pipe, &initError{Message: err.Error()}); werr != nil {
+                       fmt.Fprintln(os.Stderr, "init: write initError:", werr)
                        fmt.Fprintln(os.Stderr, err)
                        return
                }
```

here is what happens normally:
```console
[kir@kir-rhat podman]$ mkdir vol
[kir@kir-rhat podman]$ cat << EOL > ./vol/myscript
> #!/bin/sh
> echo hey you
> EOL
[kir@kir-rhat podman]$ chmod a+x ./vol/myscript 
[kir@kir-rhat podman]$ ./vol/myscript 
hey you
[kir@kir-rhat podman]$ ./bin/podman run --runtime=`pwd`/../../opencontainers/runc/runc --volume `pwd`/vol:/vol busybox /vol/myscript
hey you
```

Now, if we add `noexec` flag to the bind mount:
```console
[...]

PR #3522, opened 2022-06-30
Fix error from runc run on noexec fs

kolyshkin:
When starting a new container, and the very last step of executing of a
user process fails (last lines of `(*linuxStandardInit).Init)`, it is too
late to print a proper error since both the log pipe and the init pipe
are closed.

This is partially mitigated by using exec.LookPath() which is supposed
to say whether we will be able to execute or not. Alas, it fails to do
so when the binary to be executed resides on a filesystem mounted with
noexec flag.

A workaround is to use access(2) with X_OK flag. Alas, it is not
working when runc itself is a setuid (or setgid) binary. In this case,
faccessat2(2) with AT_EACCESS can be used, but it is only available
since Linux v5.8.

So, use faccessat2(2) with AT_EACCESS if available. If not, fall back to
access(2) for non-setuid runc, and do nothing for setuid runc (as there
is nothing we can do). Note that this check if in addition to whatever
exec.LookPath does.

The proper fix is something like https://go-review.googlesource.com/c/go/+/414824 but it won't be available for a while.

Fixes https://github.com/opencontainers/runc/issues/3520

### DIFF
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -198,6 +198,13 @@ func (l *linuxStandardInit) Init() error {
 	if err != nil {
 		return err
 	}
+	// exec.LookPath might return no error for an executable residing on a
+	// file system mounted with noexec flag, so perform this extra check
+	// now while we can still return a proper error.
+	if err := system.Eaccess(name); err != nil {
+		return &os.PathError{Op: "exec", Path: name, Err: err}
+	}
+
 	// Set seccomp as close to execve as possible, so as few syscalls take
 	// place afterward (reducing the amount of syscalls that users need to
 	// enable in their seccomp profiles). However, this needs to be done
--- a/libcontainer/system/linux.go
+++ b/libcontainer/system/linux.go
@@ -31,6 +31,25 @@ func (p ParentDeathSignal) Set() error {
 	return SetParentDeathSignal(uintptr(p))
 }
 
+// Eaccess is similar to unix.Access except for setuid/setgid binaries
+// it checks against the effective (rather than real) uid and gid.
+func Eaccess(path string) error {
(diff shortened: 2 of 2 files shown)


kolyshkin:
@opencontainers/runc-maintainers do we ever use `runc` as a setuid or setgid binary? If yes, this commit will break it, since `access(2)` checks the real, not effective, uid/gid, and we need AT_EACCESS (which is actually only supported since linux v5.8 or so).

AkihiroSuda (changes requested):
Sorry NOTLGTM, as potentially somebody might be using runc with setuid bit

(I know colima and Rancher Desktop launch nerdctl with setuid bit. AFAICS they do not launch runc with setuid bit, but I assume there are some similar projects and some of them may launch runc with setuid bit)

kolyshkin:
> (I know colima and Rancher Desktop launch nerdctl with setuid bit. AFAICS they do not launch runc with setuid bit, but I assume there are some similar projects and some of them may launch runc with setuid bit)

Thanks! I explained above this is not a problem -- we just need a more complex check for this case.

Working on it.

kolyshkin:
@AkihiroSuda added suid runc support; PTAL

AkihiroSuda (approved):

thaJeztah (approved):
SGTM

do we need more eyes on this?

AkihiroSuda:
@cyphar LGTY?

cyphar (approved):
LGTM.

AkihiroSuda:
Can we revert this?

For
- https://github.com/opencontainers/runc/issues/3715

kolyshkin:
@AkihiroSuda from what I see, we need to add a check for CAP_DAC_OVERRIDE -- it it is set, than the lack of x bit can be ignored. I will look at it tomorrow.

PR #3526, opened 2022-07-04
tests: replace local hello world bundle with busybox bundle

zhsj:
Currently only amd64 and arm64v8 tarball have been checked in testdata,
while busybox bundle is downloaded on fly, and supports multiple architectures.

To enable integration tests for more architectures, the hello world
bundle is replaced by busybox one.

Signed-off-by: Shengjing Zhu <zhsj@debian.org>


zhsj:
I'm running integration tests for the runc Debian package. I finally make the tests [paas](https://ci.debian.net/user/zhsj/jobs?package=runc) on amd64 and arm64. Now I want to enable more architectures, which are armhf, armel, i386, ppc64el, s390x, riscv64.

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

I guess we don't really benefit from the fact that hello image is smaller and doesn't have to be downloaded, because in most cases we still use busybox.

Issue #3532, opened 2022-07-20
TestCheckpoint and TestUsernsCheckpoint are failing on centos-stream-9 (Segmentation fault)


AkihiroSuda:
https://cirrus-ci.com/task/4890404691640320?logs=unit_tests#L1358

```
    checkpoint_test.go:163: (00.023368) pie: 1: restoring lsm profile (current) unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
    checkpoint_test.go:163: (00.023421) Error (criu/cr-restore.c:1510): 65440 stopped by signal 11: Segmentation fault
    checkpoint_test.go:163: (00.023704) mnt: Switching to new ns to clean ghosts
    checkpoint_test.go:163: (00.023809) Error (criu/cr-restore.c:2536): Restoring FAILED.
    checkpoint_test.go:163: === END ===
    checkpoint_test.go:164: criu failed: type NOTIFY errno 0
        log file: /tmp/TestCheckpoint1076343382/003/criu/restore.log
--- FAIL: TestCheckpoint (0.33s)
```

adrianreber:
The latest glibc changed the restartable sequence interface and that requires a CRIU rebuild. Rebuild is already triggered: https://gitlab.com/redhat/centos-stream/rpms/criu/-/commit/b6cb718321341b624fc6f2a217d04da775ae32a1 but it might take a week or so to show up in the repositories.

@rst0git FYI

thaJeztah:
Thanks for the update!

kolyshkin:
This is being fixed in `criu-3.17-4.el9` (which is not in the repo yet).

Workaround: https://github.com/opencontainers/runc/pull/3539

thaJeztah:
@kolyshkin do you think we should reopen this ticket for tracking ("remove the workaround once we can"?)

PR #3539, opened 2022-07-27
CI: workaround CentOS Stream 9 criu issue

kolyshkin:
Older criu versions fails to work with userns due to changes in glibc's
rseq. Skip a test if an older criu version is found.

Fixes (well, works around): https://github.com/opencontainers/runc/issues/3532

### DIFF
--- a/libcontainer/integration/checkpoint_test.go
+++ b/libcontainer/integration/checkpoint_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

rata (approved):
LGTM :)

alban (approved):

libcontainer/integration/checkpoint_test.go
```diff
@@ -61,6 +62,12 @@ func testCheckpoint(t *testing.T, userns bool) {
 		t.Skipf("criu binary not found: %v", err)
 	}
 
+	// Workaround for https://github.com/opencontainers/runc/issues/3532.
+	out, err := exec.Command("rpm", "-q", "criu").CombinedOutput()
```
rst0git, line 66: This might not work in Debian-based distributions such as Ubuntu.

rata: I think that is on purpose. 

If rpm doesn't exist then the output won't match the regex and that is okay, ubuntu doesn't have this criu issue. We just need to skip it in very specific cases, this will work on those cases and not fail the test in the cases we don't need to skip it.

rata: That is probably why all the tests passed here too :)

cyphar (approved):
LGTM.

libcontainer/integration/checkpoint_test.go
kolyshkin: This is why the code doesn't do anything if there is an error from this command. Since we only need this workaround for  CentOS Stream 9 which we can safely assume has working rpm.

PR #3544, opened 2022-07-29
Added support nsswitch

denisoster:
relate [issue](https://github.com/containers/podman/issues/15114)


libcontainer/user/lookup_unix.go
```diff
@@ -17,88 +24,132 @@ const (
...
+			return nil, fmt.Errorf("getent unable to find entry %q in %s database", key, database)
+		case 3:
+			return nil, fmt.Errorf("getent database doesn't support enumeration")
+		default:
+			return nil, err
+		}
+
+	}
+	return bytes.NewReader(out), nil
+}
```
cyphar, line 129: To be honest I don't like this. The main reason is that we are now going to execute an attacker-controlled binary in a context where we haven't applied all of our security profiles. I don't see how we can support NSS without running attacker code but doing it this way is going to make it easy to exploit a host (though in fairness, the only "proper" way I can think of would be to do a proper exec into the container, but there are some issues with doing that too -- mainly performance).

I'm also not sure that this behaviour would be correct for all users of `libcontainer/user` -- from memory, Docker calls us from the host and I suspect (though I haven't checked) that this would lead to executing `getent` on the host to resolve passwd/group entries in the container.

cyphar (changes requested):

libcontainer/user/lookup_unix.go
denisoster: Implement via Varlink?
https://systemd.io/USER_GROUP_API/

PR #3546, opened 2022-08-02
checkpoint/restore: implement --manage-cgroups-mode ignore

kolyshkin:
This patchset fixes a few issues and adds support for `--manage-cgroups-mode ignore`. This option allows to restore a container into a different cgroup than the original one. A test case is added to check that it works as expected.

Loosely based on https://github.com/opencontainers/runc/pull/3447 and my earlier work from 2021.

Closes: https://github.com/opencontainers/runc/pull/3447

Also, this uses the new mode in lazy checkpoint/restore test, hopefully fixing this test flakiness.

Fixes: https://github.com/opencontainers/runc/issues/2760
Fixes: https://github.com/opencontainers/runc/issues/2924
Fixes: https://github.com/opencontainers/runc/issues/2475

PS I am thinking whether this (`--manage-cgroups-mode ignore` should become a default in 1.2.0, because otherwise changing the `cgroupsPath` in container's `config.json` before restore doesn't have any effect)

### DIFF
--- a/tests/integration/checkpoint.bats
+++ b/tests/integration/checkpoint.bats
@@ -224,7 +224,14 @@ function simple_cr() {
 	# TCP port for lazy migration
 	port=27277
 
-	__runc checkpoint --lazy-pages --page-server 0.0.0.0:${port} --status-fd ${lazy_w} --work-path ./work-dir --image-path ./image-dir test_busybox &
+	__runc checkpoint \
+		--lazy-pages \
+		--page-server 0.0.0.0:${port} \
+		--status-fd ${lazy_w} \
+		--manage-cgroups-mode=ignore \
+		--work-path ./work-dir \
+		--image-path ./image-dir \
+		test_busybox &
 	cpt_pid=$!
 
 	# wait for lazy page server to be ready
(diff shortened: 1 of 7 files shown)


kolyshkin:
@opencontainers/runc-maintainers PTAL (this, among the other things, fixes a few flaky tests we had for a few years)

kolyshkin:

kolyshkin:
@adrianreber PTAL

1235467Ad:
Keep me posted on the details.

1235467Ad:
Someone is pretending to be me

kolyshkin:
@opencontainers/runc-maintainers PTAL (this, among the other things, fixes a few flaky tests we had for a few years)

kolyshkin:
@adrianreber PTAL

adrianreber:
Looks good. Thanks!

kolyshkin:

kolyshkin:
This fixes a few real bugs on which I worked (on and off) for the last few years, and it's a pity to see this sitting 4 months without a review.

tests/integration/checkpoint.bats
```diff
@@ -405,3 +416,44 @@ function simple_cr() {
...
+	[ "$status" -eq 0 ]
+	testcontainer test_busybox running
+
+	local orig_path
+	orig_path=$(get_cgroup_path "pids")
+	# Check that the cgroup exists.
+	test -d "$orig_path"
+
+	runc checkpoint --work-path ./work-dir --manage-cgroups-mode ignore test_busybox
+	grep -B 5 Error ./work-dir/dump.log || true
```
AkihiroSuda, line 433: I don't think this is robust, but I don't have an alternative idea, so LGTM.

Eventually we need to have a robust error reporting system.

AkihiroSuda (approved):

tests/integration/checkpoint.bats
kolyshkin: This is just to print anything containing `Error` (with some context) in case the checkpointing failed. It's merely a way to debug criu failures, and does not affect the test itself (only its output in case of an error). The `|| true` part here is so that the test won't fail in case there's no error; IOW, we ignore grep exit code.

Alternatively, we could create an after-test artefact containing all the files, but this works OK so far.

Thinking about it, we might work on making runc do something like what grep does here in case of an error. Currently, this is just criu writing the log file.

kolyshkin:
Rebased to resolve a conflict caused by #3655.

@AkihiroSuda please re-LGTM

AkihiroSuda (approved):

mrunalp (approved):

tests/integration/checkpoint.bats
kolyshkin: Filed https://github.com/opencontainers/runc/issues/3711 to not forget about it

PR #3549, opened 2022-08-04
seccomp: allow ActNotify if write(2) is in allowed syscalls

saschagrunert:
We can notify on the default action if write is part of the allowed syscall list. This means that profiles are now able to specify `SCMP_ACT_NOTIFY` in the default profile action if `write(2)` is allowed.

cc @mauriciovasquezbernal @haircommander

### DIFF
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -42,7 +42,14 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 
 	// Ignore the error since pre-2.4 libseccomp is treated as API level 0.
 	apiLevel, _ := libseccomp.GetAPI()
+	const writeSyscall = "write"
+	writeAllowed := false
 	for _, call := range config.Syscalls {
+		// If write is in the allowed list of syscalls, then we can also notify on the defaultAction.
+		if call.Name == writeSyscall {
+			writeAllowed = syscallIsAlwaysAllowed(call, config.Architectures)
+		}
+
 		if call.Action == configs.Notify {
 			if apiLevel < 6 {
 				return -1, fmt.Errorf("seccomp notify unsupported: API level: got %d, want at least 6. Please try with libseccomp >= 2.5.0 and Linux >= 5.7", apiLevel)
@@ -60,14 +67,14 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 			// filter). We will be blocked on read()/close() inside syncParentSeccomp() but if the seccomp
 			// agent allows those syscalls to proceed, initialization works just fine and the agent can
 			// handle future read()/close() syscalls as it wanted.
-			if call.Name == "write" {
+			if call.Name == writeSyscall {
 				return -1, errors.New("SCMP_ACT_NOTIFY cannot be used for the write syscall")
 			}
 		}
 	}
 
-	// See comment on why write is not allowed. The same reason applies, as this can mean handling write too.
-	if defaultAction == libseccomp.ActNotify {
+	// See comment on why write is not allowed. The same reason may apply if writeAllowed is false, as this can mean handling write too.
+	if !writeAllowed && defaultAction == libseccomp.ActNotify {
 		return -1, errors.New("SCMP_ACT_NOTIFY cannot be used as default action")
 	}
 
@@ -132,6 +139,38 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 	}
 
 	return seccompFd, nil
+}
+
+func syscallIsAlwaysAllowed(syscall *configs.Syscall, architectures []string) bool {
+	// We cannot guarantee if the syscall is fully allowed when additional args
+	// are provided, therefore we exclude that case.
+	if syscall == nil || len(syscall.Args) > 0 {
+		return false
+	}
+
+	switch syscall.Action {
+	case configs.Allow, configs.Trace, configs.Log:
+		// Verify if the architecture matches
+		nativeArch, err := libseccomp.GetNativeArch()
+		if err != nil {
+			return false
+		}
+
+		for _, arch := range architectures {
+			scmpArch, err := libseccomp.GetArchFromString(arch)
+			if err != nil {
+				return false
+			}
+
+			if nativeArch == scmpArch {
+				return true
+			}
+		}
+
+		fallthrough
+	default:
+		return false
+	}
 }
 
 // Convert Libcontainer Action to Libseccomp ScmpAction


libcontainer/seccomp/seccomp_linux.go
```diff
@@ -42,7 +42,23 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
...
+	writeAllowed := false
 	for _, call := range config.Syscalls {
+		// If write is in the allowed list of syscalls, then we can also notify on the defaultAction.
+		if call.Name == writeSyscall {
+			switch call.Action {
+			case configs.Allow:
+				fallthrough
+			case configs.Trace:
+				fallthrough
+			case configs.Log:
```
haircommander, line 55: this could be shorter with `case configs.Allow, configs.Trace, configs.Log:`

saschagrunert: Yep, improved as suggested :+1:

kolyshkin:
@rata @alban PTAL

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -42,7 +42,19 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
...
 	for _, call := range config.Syscalls {
+		// If write is in the allowed list of syscalls, then we can also notify on the defaultAction.
+		if call.Name == writeSyscall {
+			switch call.Action {
+			case configs.Allow, configs.Trace, configs.Log:
+				writeAllowed = true
+			default:
+				writeAllowed = false
+			}
+		}
```
rata, line 56: I would move this to another function and then just say here (pseudo code and not sure this function names are good ones, just to clearly express my idea):

`if allowedSyscall("write") && defaultAction is not write ....`

WAIT, no. As I mentioned in the other comment, this is not good enough to know if write is allowed. More info in the other comment :)

But with the other comment and suggestions there in mind, this definitely makes sense to be a different function, maybe stress the fact that we check that the syscall is _always_ allowed (no conditions) and explain this problem in a comment there.

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -60,14 +72,14 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
...
+			if call.Name == writeSyscall {
 				return -1, errors.New("SCMP_ACT_NOTIFY cannot be used for the write syscall")
 			}
 		}
 	}
 
-	// See comment on why write is not allowed. The same reason applies, as this can mean handling write too.
-	if defaultAction == libseccomp.ActNotify {
+	// See comment on why write is not allowed. The same reason may apply if writeAllowed is false, as this can mean handling write too.
+	if !writeAllowed && defaultAction == libseccomp.ActNotify {
```
rata, line 82: I think this is buggy, the default action can still be a write and we should deny it. Just because we don't know what path of all the seccomp rules will be used when runc issues a write.

But leaving that technicality aside, there seems to be a deeper issue that is not easy to fix. The issue is: seccomp profiles allows to do complex rules to match when a syscall will be allowed and when they won't. I know most profiles are just a list of allow/deny syscalls (although moby profile is not trivial), but the profile allows _much_ more than that and there is where this becomes complex. With that flexibility we can't easily answer if "write will always be allowed" easily and reliably.

Let me elaborate.

The logic checking if write is allowed here is quite simple, and is not good enough to know if a write call will be _always_ be allowed. The seccomp profile let's you express lot of subtleties (like if the param to the syscall is this or the other), it can totally be that the write is allowed when the fd is 0 and that will make this container not start and hang in a very hard to debug way. This logic doesn't detect that.

Detecting this reliably is tricky, but I _think_ we can make heuristics. We can check if args is not set, and in that case if the write is unconditionally allowed it might work.

I would like to think and have an eye from others about this fact, though.

What do you think?

rata (changes requested):
I don't think this works :(

Knowing if a syscall will always be allowed is not that straight forward as what we do here to conclude that. I think this can easily break if write is allowed only with certain params, for example.

But more importantly, what is the use case? Did you face this limitation or is something you thought "let's remove just remove this" but don't have a use case?

libcontainer/seccomp/seccomp_linux.go
rata: Another option, rather than heuristics, would be to try to use libseccomp to know if a syscall will always be allowed. I don't know if they have such a thing in their API, but if they do we could solve the problem :)

rata: @kolyshkin is a maintainer, might now... ;)

saschagrunert:
> I don't think this works :(
> 
> Knowing if a syscall will always be allowed is not that straight forward as what we do here to conclude that. I think this can easily break if write is allowed only with certain params, for example.

Got it, hm… I'll revisit this PR after my vacation ends (2 weeks from now on).

> But more importantly, what is the use case? Did you face this limitation or is something you thought "let's remove just remove this" but don't have a use case?

I'm working on an experimental secocmp notifier implementation in CRI-O right now, which stops a container once we hit a erroneous syscall: https://github.com/cri-o/cri-o/pull/6120. The runtime/default profiles switched to the default action `SCMP_ACT_ERRNO` since a while: 

https://github.com/containers/common/blob/8c1847e01634d88f4488d5f35508d2c939a8d700/pkg/seccomp/seccomp.json#L2-L4

Means we leave a gap between what we can notify on, so my goal was to close this one. I see that it's hard to achieve, so we may consider closing this PR as-is.

rata:
> I'm working on an experimental secocmp notifier implementation in CRI-O right now, which right now stops a container once we hit a erroneous syscall (or we could throw metrics later on): [cri-o/cri-o#6120](https://github.com/cri-o/cri-o/pull/6120).

Oh, cool. We have implemented something like that too: https://github.com/kinvolk/seccompagent/pull/20 and https://github.com/kinvolk/seccompagent/pull/18. Maybe we can talk on slack to collaborate on something? :) I don't think any of us at Kinvolk is planning to work on this soon, but maybe we can find room to do some collaboration :)

> 
> The runtime/default profile switched to the default action `SCMP_ACT_ERRNO` since a while:
> 
> https://github.com/containers/common/blob/8c1847e01634d88f4488d5f35508d2c939a8d700/pkg/seccomp/seccomp.json#L2-L4

Oh, and that profile has write allowed and no args. So the heuristic I mentioned might work for you, maybe that is good enough and is very easy to implement too (we probably need to check for architecture and things like that, though. I see that profile uses non-oci constructs with includes/excludes, so we have to see what OCI seccomp profile it generates).

I think that can work :)

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -134,6 +141,21 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
...
+		return false
+	}
+
+	switch syscall.Action {
+	case configs.Allow, configs.Trace, configs.Log:
+		return true
+	default:
+		return false
+	}
+}
```
saschagrunert, line 157: @rata I changed the logic to include an additional cehck for the args. Do you think this is feasible?

rata: As I mentioned [here][link] we might need to check the architecture matches?

I mean, what we need to know if the write done by runc will be allowed or not. For that to happen, if args is nil is simpelr to see (no need to really parse the language) but the arch  needs to match too. I don't know what would be the right thing to do if arch SCMP_ARCH_X86 is used but this is a SCMP_ARCH_X86_64 host (probably write will not match, as probably runc will be running in 64 bits and therefore not match that rule, even if the container started later uses x86 binaries?).

I think with those restrictions, we can say it will be used and be safe. But I'd like others to comment, in case I'm missing something.

cc @alban @cyphar @kolyshkin 


[link]: https://github.com/opencontainers/runc/pull/3549#issuecomment-1206317116

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -42,7 +42,14 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 
 	// Ignore the error since pre-2.4 libseccomp is treated as API level 0.
 	apiLevel, _ := libseccomp.GetAPI()
+	const writeSyscall = "write"
+	writeAllowed := false
 	for _, call := range config.Syscalls {
+		// If write is in the allowed list of syscalls, then we can also notify on the defaultAction.
+		if call.Name == writeSyscall {
+			writeAllowed = syscallIsAlwaysAllowed(call, config.Architectures)
```
rata, line 50: Hmm, if for some reason the syscall entry is repeated for write, what should we do? I don't know what libseccomp will do, nor if that behavior is guaranteed.

AkihiroSuda:
Needs rebase

Zheaoli:
I think maybe I can open a new PR to rebase the main code and change some impl detail and ask review again

saschagrunert:
> I think maybe I can open a new PR to rebase the main code and change some impl detail and ask review again

Sounds good to me, thanks for jumping in!

PR #3553, opened 2022-08-08
ci/cirrus: enable some rootless tests on cs9

kolyshkin:
We were not running localrootlessintegration test on CentOS Stream 9
because of some failures.

Fix those, and enable rootless integration with both systemd and fs drivers.

Amends: #3427


kolyshkin:
Need to figure out the reason of failures, but currently don't have enough time.

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

Issue #3551, opened 2022-08-05
OCI runtime exec failed: exec failed: unable to start container process: open /dev/pts/0: operation not permitted: unknown


Azertooth:
As title, this error is generated by a breaking bug [found](https://github.com/containerd/containerd/issues/7219) on runc version 1.1.3

kolyshkin:
I wonder if this is related to https://github.com/opencontainers/runc/pull/3504?

kolyshkin:
> I wonder if this is related to #3504?

OK, looks like it is.

I've added the following debug patch to the tip of `release-1.1` branch:

```diff
diff --git a/libcontainer/cgroups/systemd/common.go b/libcontainer/cgroups/systemd/common.go
index 5a68a3cf..b153ec51 100644
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -295,6 +295,8 @@ func generateDeviceProperties(r *configs.Resources) ([]systemdDbus.Property, err
                // have a corresponding path.
                if _, err := os.Stat(entry.Path); err == nil {
                        deviceAllowList = append(deviceAllowList, entry)
+               } else {
+                       logrus.Warnf("Skipping device %+v: %w", entry, err)
                }
        }
 
```

and got the following output (from one of the tests -- doesn't matter):
```
   runc run -d --console-socket /tmp/bats-run-JcLxzP/runc.7sdPPv/tty/sock test_busybox (status=0):
   time="2022-08-10T16:47:33-07:00" level=warning msg="Skipping device {Path:block-* Perms:m}: %!w(*fs.PathError=&{stat block-* 2})"
   time="2022-08-10T16:47:33-07:00" level=warning msg="Skipping device {Path:char-* Perms:m}: %!w(*fs.PathError=&{stat char-* 2})"
   time="2022-08-10T16:47:33-07:00" level=warning msg="Skipping device {Path:char-pts Perms:rwm}: %!w(*fs.PathError=&{stat char-pts 2})"
```

So, what happens, I guess, is:
1. With  #3504 in place, we are not longer adding `char-pts rwm` rule to the container's `DeviceAllow` systemd unit property.
2. The fact that the rule is excluded doesn't matter, since we also apply the rules directly (for cgroup v1, by writing to `devices.allow`)
3. `systemctl daemon-reload` reapplies the `DeviceAllow` rules, thus removing the char-pts rule.

Fix is coming.

[...]

PR #3554, opened 2022-08-11
[1.1] Fix failed exec after systemctl daemon-reload (regression in 1.1.3)

kolyshkin:
A regression reported for runc v1.1.3 says that after systemctl
daemon-reload runc exec fails:

> exec failed: unable to start container process: open /dev/pts/0: operation not permitted: unknown

Apparently, with commit 7219387eb7db69b we are no longer adding
"DeviceAllow=char-pts rwm" rule (because os.Stat("char-pts") returns
ENOENT).

The bug can only be seen after "systemctl daemon-reload" because runc
also applies the same rules manually (by writing to devices.allow for
cgroup v1), and apparently reloading systemd leads to re-applying the
rules that systemd has (thus removing the char-pts access).

The fix is to do os.Stat only for "/dev" paths.

Also, emit a warning that the path was skipped. Since the original idea
was to emit less warnings, demote the level to debug.

Note this also fixes the issue of not adding "m" permission for block-*
and char-* devices.
    
A test case is added, which reliably fails before the fix
on both cgroup v1 and v2.

Fixes: https://github.com/opencontainers/runc/issues/3551
Fixes: 7219387eb7db69b4dae740c9d37d973d9a735801

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -288,14 +288,16 @@ func generateDeviceProperties(r *configs.Resources) ([]systemdDbus.Property, err
 			case devices.CharDevice:
 				entry.Path = fmt.Sprintf("/dev/char/%d:%d", rule.Major, rule.Minor)
 			}
+			// systemd will issue a warning if the path we give here doesn't exist.
+			// Since all of this logic is best-effort anyway (we manually set these
+			// rules separately to systemd) we can safely skip entries that don't
+			// have a corresponding path.
+			if _, err := os.Stat(entry.Path); err != nil {
+				logrus.Debugf("skipping device %s for systemd: %s", entry.Path, err)
+				continue
+			}
 		}
-		// systemd will issue a warning if the path we give here doesn't exist.
-		// Since all of this logic is best-effort anyway (we manually set these
-		// rules separately to systemd) we can safely skip entries that don't
-		// have a corresponding path.
-		if _, err := os.Stat(entry.Path); err == nil {
-			deviceAllowList = append(deviceAllowList, entry)
-		}
+		deviceAllowList = append(deviceAllowList, entry)
 	}
 
 	properties = append(properties, newProp("DeviceAllow", deviceAllowList))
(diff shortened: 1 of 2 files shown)


kolyshkin:
CI failure is to be fixed by https://github.com/opencontainers/runc/pull/3558

Still no regression test :(

thaJeztah:
Tried a build from this PR, and that fixes the issue reported in https://github.com/moby/moby/discussions/43960#discussioncomment-3409545

AkihiroSuda:
Thanks, but I don't see the main PR, isn't the main branch affected?

kolyshkin:
The fix is quite obvious, so perhaps we can merge it without a regression test? WDYT @cyphar @thaJeztah @AkihiroSuda

thaJeztah (dismissed):
LGTM

I'm ok with looking for a test as follow-up

thaJeztah:
> Thanks, but I don't see the main PR, isn't the main branch affected?

good point; PR for main is in https://github.com/opencontainers/runc/pull/3559, but indeed still in draft; do we want the one in main first?

kolyshkin:
Draft pending #3559 merge.

AkihiroSuda (approved):

kolyshkin:
Needs a rebase once #3562 is merged to fix CI.

AkihiroSuda (dismissed):

AkihiroSuda:
Please run git cherry-pick with -x for tracking the main commit

kolyshkin:
> Please run git cherry-pick with -x for tracking the main commit

That is what I usually do, although not in this case. The reasons are:
 - I initially created the fix in `release-1.1` branch and then forward-ported it to `main` (so this one is not a cherry-pick)
 - Automatic cherry-pick (either way) is not possible because the code we patch was moved to a different file.
 
 I can certainly add a reference to commit 58b1374f0ad98cc9390adc43dc22ddbb4f84d72e.

kolyshkin:
> I can certainly add a reference to commit [58b1374](https://github.com/opencontainers/runc/commit/58b1374f0ad98cc9390adc43dc22ddbb4f84d72e) to the commit message.

Done; PTAL @AkihiroSuda

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

thaJeztah:
@kolyshkin @AkihiroSuda @cyphar should we do a `1.1.4` release to fix the regression, or do we have more things to fix? current diff looks like; https://github.com/opencontainers/runc/compare/v1.1.3...release-1.1

PR #3555, opened 2022-08-11
[debug] Int test for #3551

kolyshkin:
Making sure the test fails before the fix

### DIFF
--- a/tests/integration/dev.bats
+++ b/tests/integration/dev.bats
@@ -128,3 +128,19 @@ function teardown() {
 	runc exec test_allow_block sh -c 'fdisk -l '"$device"''
 	[ "$status" -eq 0 ]
 }
+
+# https://github.com/opencontainers/runc/issues/3551
+@test "runc exec vs systemctl daemon-reload" {
+	requires systemd root
+
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_exec
+	[ "$status" -eq 0 ]
+
+	runc exec -t test_exec sh -c "ls -l /proc/self/fd/0; echo 123"
+	[ "$status" -eq 0 ]
+
+	systemctl daemon-reload
+
+	runc exec -t test_exec sh -c "ls -l /proc/self/fd/0; echo 123"
+	[ "$status" -eq 0 ]
+}


kolyshkin:
Hmm, I am having a problem reproducing this. Perhaps we need to have some special (non-default) device rules, such as those set in containerd. Could you some help with this

thaJeztah:
@kolyshkin not sure if relevant for the CI environment, but in https://github.com/moby/moby/discussions/43960 I noticed that the failures only occurred with SELinux in enforce mode.

kolyshkin:
Ah! Finally reproduced that locally:

```console
[kir@kir-rhat runc-tst]$ sudo ./runc --systemd-cgroup run -d 234
[kir@kir-rhat runc-tst]$ sudo ./runc --systemd-cgroup exec -t 234 sh -c 'ls -l /proc/self/fd/0; echo 123'
lrwx------    1 root     root            64 Aug 17 22:13 /proc/self/fd/0 -> /dev/pts/0
123
[kir@kir-rhat runc-tst]$ sudo ./runc --systemd-cgroup exec -t 234 sh -c 'ls -l /proc/self/fd/0; echo 123'
lrwx------    1 root     root            64 Aug 17 22:13 /proc/self/fd/0 -> /dev/pts/0
123
[kir@kir-rhat runc-tst]$ sudo systemctl daemon-reload
[kir@kir-rhat runc-tst]$ sudo ./runc --systemd-cgroup exec -t 234 sh -c 'ls -l /proc/self/fd/0; echo 123'
ERRO[0000] exec failed: unable to start container process: open /dev/pts/0: operation not permitted 
```

kolyshkin:
> Hmm, I am having a problem reproducing this.

Ah! It was missing `-t` from the `runc exec` 🤦🏻

kolyshkin:
OK, the test fails for `ci / test` jobs (recent Ubuntu) as well as for CentOS 8, 9, and Fedora 36.

For some reason it does not fail for CentOS 7 -- I guess the version of systemd in there does not reload device rules upon `systemctl daemon-reload`.

PR #3559, opened 2022-08-16
Fix failed exec after systemctl daemon-reload

kolyshkin:
This is a forward port of #3554 to the main branch. Original description follows.

----

A regression reported for runc v1.1.3 says that after systemctl
daemon-reload runc exec fails:

> exec failed: unable to start container process: open /dev/pts/0: operation not permitted: unknown

Apparently, with commit 7219387eb7db69b we are no longer adding
"DeviceAllow=char-pts rwm" rule (because os.Stat("char-pts") returns
ENOENT).

The bug can only be seen after "systemctl daemon-reload" because runc
also applies the same rules manually (by writing to devices.allow for
cgroup v1), and apparently reloading systemd leads to re-applying the
rules that systemd has (thus removing the char-pts access).

The fix is to do os.Stat only for "/dev" paths.

Also, emit a warning that the path was skipped. Since the original idea
was to emit less warnings, demote the level to debug.

Note this also fixes the issue of not adding "m" permission for block-*
and char-* devices.
    
A test case is added, which reliably fails before the fix
on both cgroup v1 and v2 (for a test run before the fix, see #3555).

Fixes: https://github.com/opencontainers/runc/issues/3551
Fixes: 7219387eb7db69b4dae740c9d37d973d9a735801

### DIFF
--- a/libcontainer/cgroups/devices/systemd.go
+++ b/libcontainer/cgroups/devices/systemd.go
@@ -128,14 +128,16 @@ func systemdProperties(r *configs.Resources) ([]systemdDbus.Property, error) {
 			case devices.CharDevice:
 				entry.Path = fmt.Sprintf("/dev/char/%d:%d", rule.Major, rule.Minor)
 			}
+			// systemd will issue a warning if the path we give here doesn't exist.
+			// Since all of this logic is best-effort anyway (we manually set these
+			// rules separately to systemd) we can safely skip entries that don't
+			// have a corresponding path.
+			if _, err := os.Stat(entry.Path); err != nil {
+				logrus.Debugf("skipping device %s for systemd: %s", entry.Path, err)
+				continue
+			}
 		}
-		// systemd will issue a warning if the path we give here doesn't exist.
-		// Since all of this logic is best-effort anyway (we manually set these
-		// rules separately to systemd) we can safely skip entries that don't
-		// have a corresponding path.
-		if _, err := os.Stat(entry.Path); err == nil {
-			deviceAllowList = append(deviceAllowList, entry)
-		}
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
In the commit message could you explain how to test this?
Does this need cgroup v1 or SELinux for testing?

AkihiroSuda:
Is this still draft? I think we want to merge this main PR first and then cherry-pick

thaJeztah:
Looks like it was @kolyshkin's intention to merge the 1.1 fix first, then forward-port to main;

> This is a forward port of https://github.com/opencontainers/runc/pull/3554 to the main branch.

Have we done this before? I think generally we go from `main` -> `release-branch` to prevent the risk of regressions in main if we forget to forward-port, or have we done the other way round before?

kolyshkin:
Got some progress with the test case in https://github.com/opencontainers/runc/pull/3555. Will add the test case here once it's working. Draft for now.

While originally this is a forward port (because the bug was reported against 1.1 and I initially fixed it in there), let's merge this one first, and then follow up with 1.1.

kolyshkin:
Draft pending test case addition.

kolyshkin:
Ready for review. The `validate / codespell` CI failure is unrelated and is being fixed by PR #3561

kolyshkin:
The test failure before the fix can be seen in PR #3555

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

do we want to ignore the code spell issue, or should we do a rebase? (probably that requires re-cherry-picking for the release branch though)

AkihiroSuda (approved):

mrunalp (approved):

PR #3563, opened 2022-08-19
libcontainer: LockOSThread around Umask calls

haircommander:
umask operates on a thread basis, and we risk leaking the umask to other routines
if the go runtime interrupts.

Signed-off-by: Peter Hunt~ <pehunt@redhat.com>

### DIFF
--- a/libcontainer/utils/utils_unix.go
+++ b/libcontainer/utils/utils_unix.go
@@ -6,6 +6,7 @@ package utils
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"strconv"
 
 	"golang.org/x/sys/unix"
@@ -66,4 +67,29 @@ func NewSockPair(name string) (parent *os.File, child *os.File, err error) {
 		return nil, nil, err
 	}
 	return os.NewFile(uintptr(fds[1]), name+"-p"), os.NewFile(uintptr(fds[0]), name+"-c"), nil
+}
+
+// RunWithUmask runs a function f with umask mask.
+// It does so while locking the OS thread, as umask
+// operates on threads, and this avoid potentially leaking umask(0000)
+// to other threads.
+func RunWithUmask(mask int, f func() error) error {
+	errC := make(chan error)
+	go func() {
+		runtime.LockOSThread()
+		// The umask is part of the "filesystem information" whose sharing is governed by the `CLONE_FS` flag.
+		// The Go runtime clones new threads with the `CLONE_FS` flag set, so any call to umask immediately affects
+		// all goroutines in the process.
+		if err := unix.Unshare(unix.CLONE_FS); err != nil {
+			errC <- err
+			runtime.UnlockOSThread()
+			return
+		}
+		// However, unshare(CLONE_FS) is irreversible. Thus, ensure the thread is terminated by
+		// not calling UnlockOSThread (which signals to the Go runtime to terminate or wedge the thread),
+		// so this OSThread is not reused in other parts of the program.
+		unix.Umask(mask)
+		errC <- f()
+	}()
+	return <-errC
 }
--- a/libcontainer/console_linux.go
+++ b/libcontainer/console_linux.go
@@ -3,22 +3,23 @@ package libcontainer
 import (
 	"os"
 
+	"github.com/opencontainers/runc/libcontainer/utils"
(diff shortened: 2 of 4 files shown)


libcontainer/utils/utils_unix.go
```diff
@@ -67,3 +68,16 @@ func NewSockPair(name string) (parent *os.File, child *os.File, err error) {
...
 }
+
+// RunWithUmask runs a function f with umask mask.
+// It does so while locking the OS thread, as umask
+// operates on threads, and this avoid potentially leaking umask(0000)
+// to other threads.
+func RunWithUmask(mask int, f func() error) error {
+	runtime.LockOSThread()
+	oldMask := unix.Umask(mask)
+	err := f()
```
cyphar, line 79: If we're being very pedantic, this won't handle panics -- you probably want to do the unlock and resetting of the umask with `defer`s (which are much faster in modern versions of Go than they were in the earlier days).

haircommander: pedantry works for me!

AkihiroSuda (approved):

libcontainer/utils/utils_unix.go
```diff
@@ -67,3 +68,15 @@ func NewSockPair(name string) (parent *os.File, child *os.File, err error) {
...
+// RunWithUmask runs a function f with umask mask.
+// It does so while locking the OS thread, as umask
+// operates on threads, and this avoid potentially leaking umask(0000)
+// to other threads.
+func RunWithUmask(mask int, f func() error) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+	oldMask := unix.Umask(mask)
+	defer unix.Umask(oldMask)
+	return f()
```
corhere, line 81: The umask is part of the "filesystem information" whose sharing is governed by the `CLONE_FS` flag. [The Go runtime clones new threads with the `CLONE_FS` flag set](https://github.com/golang/go/blob/895664482c0ebe5cec4a6935615a1e9610bbf1e3/src/runtime/os_linux.go#L153) so any call to `unix.Umask()` immediately affects all goroutines in the process. `runtime.LockOSThread()` alone is not enough to prevent the change from leaking to other threads.

Here's what would work:
```suggestion
	errC := make(chan error)
	go func() {
		runtime.LockOSThread()
		if err := unix.Unshare(unix.CLONE_FS); err != nil {
			errC <- err
			runtime.UnlockOSThread()
			return
		}
		unix.Umask(mask)
		errC <- f()
	}()
	return <-errC
```

`unix.Unshare(CLONE_FS)` is an irreversible operation; the calling thread will never again have its umask, cwd or root directory shared with other threads. The thread will no longer be indistinguishable from other threads which run goroutines. Returning from the goroutine while locked to the thread signals to the Go runtime to terminate the thread ([or wedge, if it's the main thread](https://github.com/golang/go/blob/895664482c0ebe5cec4a6935615a1e9610bbf1e3/src/runtime/proc.go#L1476-L1483)) for situations exactly like this one. ([This article has some good information as well.](https://www.weave.works/blog/linux-namespaces-golang-followup)) Just be aware that [terminating threads can have some undesirable interactions with `(syscall.SysProcAttr).Pdeathsig`](https://go.dev/issue/27505) which need to be accounted for. It's straightforward enough to deal with once you know how what's happening and why, e.g.: https://github.com/moby/moby/pull/44215.

corhere (changes requested):
I just so happen to be working on a very similar change over in moby/moby. Locking the goroutine to a thread is not sufficient to prevent umask changes from leaking into other threads.

libcontainer/utils/utils_unix.go
haircommander: thanks for the information! updated as suggested

corhere (approved):

cyphar (changes requested):
Sorry, I wrote a review some time ago but never submitted it. In short, I think that we should just remove all usage of `unix.Umask` in our codebase and just explicitly set the modes of everything we create. The `CLONE_NEWFS` issue is enough of a problem that we should just avoid over-complicating it (burning a thread each time we have to call `umask(2)` is kind of insane) and just do the simple thing.

haircommander:
closing in favor of https://github.com/opencontainers/runc/pull/4039

PR #3565, opened 2022-08-22
Add check for CONFIG_CGROUP_BPF in check-config.sh

### LINKED ISSUE #3547: Add check for CONFIG_CGROUP_BPF in check-config.sh
The kernel config option `CONFIG_CGROUP_BPF=y` is required to be able to inject the BPF program of type `BPF_PROG_TYPE_CGROUP_DEVICE`.

If this option is not set, no containers can be started with `runc` (BTW: the same holds for `crun`).

Currently, the `check-config.sh` script does not test for this option, leading to all-green results for kernels where runc fails to start a container.

dharmicksai:
For runc to start containers, the kernel config option CONFIG_CGROUP_BPF=y is required.
check-config.sh now checks if the CONFIG_CGROUP_BPF kernel config option is set.

Added flag `CGROUP_BPF` to the `flags` array. The `flags` array contains the list of all essential flags that will be checked on running the check-config.sh script 

Closes #3547

Signed-off-by: K Dharmick Sai <dharmicksaik@gmail.com>


kolyshkin (changes requested):
From what I understand,
1. CONFIG_CGROUP_BPF is only needed when cgroup v2 is used
2. It is only available since kernel 4.15 (commit ebc614f687369f9df99828572b1d85a7c2de3d92), so at least there should be a version guard.

dharmicksai:
I have added the version guard to make sure the kernel version is at least 4.15 before checking for CONFIG_CGROUP_BPF
The check is under Optional Features as it is required only when cgroup v2 is used.

kolyshkin:
> I have added the version guard to make sure the kernel version is at least 4.15 before checking for CONFIG_CGROUP_BPF

Thanks!

> The check is under Optional Features as it is required only when cgroup v2 is used.

To my mind, this setting is not at all needed when cgroup v1 is used, and is required when cgroup v2 is used.

So maybe it makes sense to do something like
```diff
diff --git a/script/check-config.sh b/script/check-config.sh
index ec8bc63a..bc5a1d61 100755
--- a/script/check-config.sh
+++ b/script/check-config.sh
@@ -177,15 +177,17 @@ wrap_color "info: reading kernel config from $CONFIG ..." white
 echo
 
 echo 'Generally Necessary:'
-
+cgroup=''
 echo -n '- '
 if [ "$(stat -f -c %t /sys/fs/cgroup 2>/dev/null)" = "63677270" ]; then
        wrap_good 'cgroup hierarchy' 'cgroupv2'
+       cgroup=v2
 else
        cgroupSubsystemDir="$(awk '/[, ](cpu|cpuacct|cpuset|devices|freezer|memory)[, ]/ && $3 == "cgroup" { print $2 }' /proc/mounts | head -n1)"
        cgroupDir="$(dirname "$cgroupSubsystemDir")"
        if [ -d "$cgroupDir/cpu" ] || [ -d "$cgroupDir/cpuacct" ] || [ -d "$cgroupDir/cpuset" ] || [ -d "$cgroupDir/devices" ] || [ -d "$cgroupDir/freezer" ] || [ -d "$cgroupDir/memory" ]; then
                wrap_good 'cgroup hierarchy' 'properly mounted' "[$cgroupDir]"
+               cgroup=v1
        else
                if [ "$cgroupSubsystemDir" ]; then
                        wrap_bad 'cgroup hierarchy' 'single mountpoint!' "[$cgroupSubsystemDir]"
```

and later, in "Generally necessary section", do
```bash
if [ "$cgroup" = "v2" ] && ! kernel_lt 4_15; then
  check_flags CONFIG_CGROUP_BPF
fi
```

(The above code is totally untested)

kolyshkin (approved):
LGTM

thaJeztah (approved):
LGTM, thanks!

Issue #3577, opened 2022-08-28
runc checkpoint: destroy container error 


gosoon:
I create a runc container `mycontainerid` with README [using-runc](https://github.com/gosoon/runc#using-runc) section.But exec `runc checkpoint mycontainerid` cmd failed with error when container is running or paused.As shown below：

```
# runc list
ID              PID         STATUS      BUNDLE         CREATED                          OWNER
mycontainerid   16670       running     /mycontainer   2022-08-28T11:49:44.809303123Z   root


# runc checkpoint mycontainerid
ERRO[0000] container still running
ERRO[0000] criu failed: type NOTIFY errno 0
log file: dump.log


# runc list
ID              PID         STATUS      BUNDLE         CREATED                          OWNER
mycontainerid   16670       paused      /mycontainer   2022-08-28T11:49:44.809303123Z   root

# runc checkpoint mycontainerid
ERRO[0000] container paused
ERRO[0000] criu failed: type NOTIFY errno 0
log file: dump.log


# runc -v
runc version 1.1.0+dev
commit: v1.1.0-272-g4a51b04
spec: 1.0.2-dev
go: go1.19
libseccomp: 2.3.1
```

kolyshkin:
So, for some reason, container checkpoint failed. It can happen, and in general this is not a bug per se.

Are you saying that the issue here is leaving the container in a paused state after a failed checkpointing?

gosoon:
I'm sorry, I may not have described it clearly.

The above are two cases that execute `runc checkpoint mycontainerid ` failed.

When container is `paused` or `running` status, execute  `runc checkpoint `  and the container need be destroy by default,not use `--leave-running` or `--pre-dump` flag.But it actually fails in the destroy container stage,the error message is `container still running` when container is `running` status,the error message is `container paused` when container is `paused` status.



I looked through the Runc code, the container need be destroy when execute  `runc checkpoint `  not use  `--leave-running` or `--pre-dump` flag.The code is shown below：

https://github.com/opencontainers/runc/blob/4a51b047036cf16d4f124548c2a7ff24b5640bad/checkpoint.go#L70-L73

The code is shown below and have an error `ErrRunning`   , if destroy `running` status container：

https://github.com/opencontainers/runc/blob/4a51b047036cf16d4f124548c2a7ff24b5640bad/libcontainer/state_linux.go#L131-L136

The code is shown below and have an error `ErrPaused`   , if destroy `paused` status container：

[...]

PR #3578, opened 2022-08-28
fix: checkpoint error when destroy container

gosoon:
fix [issue#3577](https://github.com/opencontainers/runc/issues/3577)


checkpoint.go
```diff
@@ -69,7 +69,11 @@ checkpointed.`,
 
 		if !(options.LeaveRunning || options.PreDump) {
 			// destroy container unless we tell CRIU to keep it
-			defer destroy(container)
+			defer func() {
+				if err := killContainer(container); err != nil {
```
kolyshkin, line 73: I don't understand this change. What destroy does is removing the container. What `killContainer` does is sending SIGKILL to the container. Replacing one with the other does not make sense.

I don't understand how it fixes https://github.com/opencontainers/runc/issues/3577 either.

kolyshkin (changes requested):
I don't understand the logic of this change, nor how it fixes #3577.

kolyshkin:
@gosoon PTAL ^^^

kolyshkin:
@gosoon I left a question above. Your patch comes without an explanation. Can you please describe/explain why this change is needed, what it does, etc.

checkpoint.go
kolyshkin: @gosoon ^^^^

kolyshkin:
OK, I took a look and it seems that destroy() should not be called at all if the checkpointing failed.

Here's a PR: https://github.com/opencontainers/runc/pull/3655

kolyshkin:
Superceded by #3655

Issue #3509, opened 2022-06-14
cgroup v1/v2 compatibility issue when setting memory below the current usage


kolyshkin:
With cgroup v1, when we set the memory limit to below the current usage, the kernel returns EBUSY and runc fails with a nice error message:
```
ERRO[0000] unable to set memory limit to 27033 (current usage: 270336, peak usage: 6082560) 
```

With cgroup v2, when do do this, kernel OOM killer just kill the container. This makes this behavior incompatible with cgroup v1.

One (imperfect) workaround is to add a flag to OCI spec that disallows to set memory limit to the value lower than the current usage. This is borderline ugly but at least in most cases we'll return an error instead of letting the container being OOM killed.

(the other, much less serious part of the problem is, when container is disappearing in the middle of runc update, we get all sorts of ugly messages)

giuseppe:
could the Kubelet use `memory.high` instead of `memory.max`?

danishprakash:
I don't have a complete understanding at this point but are we talking about cgroup memory limit applied at the time of container creation? And if that's the case, is the difference then the fact that in cgroupv2 the kernel isn't returning an EBUSY anymore?

> add a flag to OCI spec
And then have runc parse it and fail early instead of the container being OOMKilled?

mrunalp:
This is when we try to update the memory limit of an already running container to a value that is less than what it is currently using. In v1, we got EBUSY, but in v2, kernel applies the value and if it is low, the container is OOM Killed.

kolyshkin:
> could we use `memory.high` instead of `memory.max`?

From the vertical pod autoscaler POV -- yes. Meaning, it will still have to distinguish between v1 and v2. Meaning, it does not make sense to add a flag I have proposed in the description.

mrunalp:
> could we use memory.high instead of memory.max

I think that will have to be phase 2 with cgroups v2 in k8s. Phase 1 is just a direct mapping to v1.

utam0k:
Is it possible to get the current memory usage from `memory.current` and if it is lower than that, not update it and return an error? This may be too much help as OCI runtime...?

kamizjw:
Is there a similar problem with other configurations other than memory?

kolyshkin:
> Is there a similar problem with other configurations other than memory?

Not that I know of.

PR #3579, opened 2022-08-29
runc update: implement memory.checkBeforeUpdate

kolyshkin:
This is aimed at solving the problem of cgroup v2 memory controller
behavior which is not compatible with that of cgroup v1.

In cgroup v1, if the new memory limit being set is lower than the
current usage, setting the new limit fails.

In cgroup v2, same operation succeeds, and the container is OOM killed.

Introduce a new setting, memory.checkBeforeUpdate, and use it to mimic
cgroup v1 behavior.

Note that this is not 100% reliable because of TOCTOU, but this is the
best we can do.

Fixes: https://github.com/opencontainers/runc/issues/3509

~~Currently a draft, pending https://github.com/opencontainers/runtime-spec/pull/1158 merge.~~

### DIFF
--- a/libcontainer/cgroups/fs2/fs2.go
+++ b/libcontainer/cgroups/fs2/fs2.go
@@ -269,3 +269,35 @@ func (m *Manager) OOMKillCount() (uint64, error) {
 
 	return c, err
 }
+
+func CheckMemoryUsage(dirPath string, r *configs.Resources) error {
+	if !r.MemoryCheckBeforeUpdate {
+		return nil
+	}
+
+	if r.Memory <= 0 && r.MemorySwap <= 0 {
+		return nil
+	}
+
+	usage, err := fscommon.GetCgroupParamUint(dirPath, "memory.current")
+	if err != nil {
+		// This check is on best-effort basis, so if we can't read the
+		// current usage (cgroup not yet created, or any other error),
+		// we should not fail.
+		return nil
+	}
+
+	if r.MemorySwap > 0 {
+		if uint64(r.MemorySwap) <= usage {
+			return fmt.Errorf("rejecting memory+swap limit %d <= usage %d", r.MemorySwap, usage)
+		}
+	}
+
+	if r.Memory > 0 {
+		if uint64(r.Memory) <= usage {
+			return fmt.Errorf("rejecting memory limit %d <= usage %d", r.Memory, usage)
+		}
+	}
+
+	return nil
+}
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -174,7 +174,14 @@ func unifiedResToSystemdProps(cm *dbusConnManager, res map[string]string) (props
 	return props, nil
 }
 
-func genV2ResourcesProperties(r *configs.Resources, cm *dbusConnManager) ([]systemdDbus.Property, error) {
+func genV2ResourcesProperties(dirPath string, r *configs.Resources, cm *dbusConnManager) ([]systemdDbus.Property, error) {
+	// We need this check before setting systemd properties, otherwise
+	// the container is OOM-killed and the systemd unit is removed
+	// before we get to fsMgr.Set().
+	if err := fs2.CheckMemoryUsage(dirPath, r); err != nil {
+		return nil, err
+	}
+
 	var properties []systemdDbus.Property
 
 	// NOTE: This is of questionable correctness because we insert our own
@@ -437,7 +444,7 @@ func (m *UnifiedManager) Set(r *configs.Resources) error {
 	if r == nil {
 		return nil
 	}
-	properties, err := genV2ResourcesProperties(r, m.dbus)
+	properties, err := genV2ResourcesProperties(m.fsMgr.Path(""), r, m.dbus)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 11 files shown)


libcontainer/cgroups/fs2/fs2.go
```diff
@@ -269,3 +269,32 @@ func (m *manager) OOMKillCount() (uint64, error) {
...
+	if !r.MemoryCheckBeforeUpdate {
+		return nil
+	}
+
+	if r.Memory <= 0 && r.MemorySwap <= 0 {
+		return nil
+	}
+
+	usage, err := fscommon.GetCgroupParamUint(dirPath, "memory.current")
+	if err != nil {
```
kolyshkin, line 283: This check is on best-effort basis, so if we can't read the current usage (cgroup not yet created, or any other error), we should not fail.

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -269,3 +269,32 @@ func (m *manager) OOMKillCount() (uint64, error) {
...
+		return nil
+	}
+
+	if r.MemorySwap > 0 {
+		if uint64(r.MemorySwap) <= usage {
```
kolyshkin, line 288: This code assumes that memory.current value include used swap. I'm not 100% sure it's true.

libcontainer/cgroups/systemd/v2.go
```diff
@@ -174,7 +174,11 @@ func unifiedResToSystemdProps(cm *dbusConnManager, res map[string]string) (props
 	return props, nil
 }
 
-func genV2ResourcesProperties(r *configs.Resources, cm *dbusConnManager) ([]systemdDbus.Property, error) {
+func genV2ResourcesProperties(dirPath string, r *configs.Resources, cm *dbusConnManager) ([]systemdDbus.Property, error) {
+	if err := fs2.CheckMemoryUsage(dirPath, r); err != nil {
```
kolyshkin, line 178: We need this check before setting systemd properties, otherwise the container is OOM-killed and the systemd unit is removed before we get to `fsMgr.Set()`

kolyshkin:
The test case was failing:
```
not ok 167 update memory vs CheckBeforeUpdate
....
# runc update test_update --memory 1024 (status=1):
# time="2022-09-10T02:16:40Z" level=warning msg="Setting back cgroup configs failed due to error: unable to set unit properties: Unit runc-cgroups-integration-test-32127.scope not found., your state.json and actual configs might be inconsistent."
# time="2022-09-10T02:16:40Z" level=error msg="openat2 /sys/fs/cgroup/user.slice/user-2000.slice/user@2000.service/machine.slice/runc-cgroups-integration-test-32127.scope/memory.max: no such file or directory"
```

This is where we set memory limit too low without using `checkBeforeUpdate`. In this case, the container is OOM-killed (and its cgroup and systemd unit are removed) in the midst of `runc update`. Depending on timing, runc might succeed or fail, since it's a race between runc, systemd, and the kernel.

Modified the test case to not check the error code from this run.

kolyshkin:
Rebased; PTAL @opencontainers/runc-maintainers

libcontainer/cgroups/fs2/fs2.go
mrunalp: Worth adding this as a comment?

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -269,3 +269,32 @@ func (m *manager) OOMKillCount() (uint64, error) {
...
+			return fmt.Errorf("rejecting memory+swap limit %d > usage %d", r.MemorySwap, usage)
```
mrunalp, line 289: Shouldn't this be the other way? `<`

libcontainer/cgroups/fs2/fs2.go
```diff
@@ -269,3 +269,32 @@ func (m *manager) OOMKillCount() (uint64, error) {
...
+		}
+	}
+
+	if r.Memory > 0 {
+		if uint64(r.Memory) <= usage {
+			return fmt.Errorf("rejecting memory limit %d > usage %d", r.Memory, usage)
```
mrunalp, line 295: Same here `<`

libcontainer/cgroups/fs2/fs2.go
kolyshkin: Good catch, thanks!

mrunalp (approved):

tests/integration/update.bats
```diff
@@ -699,3 +699,43 @@ EOF
 	runc resume test_update
 	[ "$status" -eq 0 ]
 }
+
+@test "update memory vs CheckBeforeUpdate" {
+	requires cgroups_v2
```
AkihiroSuda, line 704: What should happen on cgroups_v1 ?

kolyshkin: This knob is for cgroup v2, it is ignored (not used) on v1. Since we're testing the knob here, and not the kernel cgroup behavior, the test requires cgroup v2.

I don't know how to write a test to ensure the knob is ignored. In general, runc ignores all the parameters it's not aware of, and cgroup v1 code is not aware of CheckBeforeUpdate.

AkihiroSuda (approved):

Issue #3540, opened 2022-07-28
TestFreezeBeforeSet on CentOS 7


kolyshkin:
Happened once in Cirrus CI (https://cirrus-ci.com/task/5418288451682304?logs=unit_tests#L549) while running CI for https://github.com/opencontainers/runc/pull/3512

```
=== RUN   TestFreezeBeforeSet
=== RUN   TestFreezeBeforeSet/slice,skip-devices
    freeze_test.go:171: failed to write "0\n": write /sys/fs/cgroup/cpuset/system.slice/system-runc_test_freeze_1.slice/cpuset.mems: no such device
=== RUN   TestFreezeBeforeSet/scope,skip-devices
=== RUN   TestFreezeBeforeSet/slice,will-freeze
=== RUN   TestFreezeBeforeSet/slice,pre-frozen,will-freeze
=== RUN   TestFreezeBeforeSet/scope,pre-frozen,skip-devices
=== RUN   TestFreezeBeforeSet/scope,pre-frozen
--- FAIL: TestFreezeBeforeSet (0.41s)
    --- FAIL: TestFreezeBeforeSet/slice,skip-devices (0.06s)
    --- PASS: TestFreezeBeforeSet/scope,skip-devices (0.07s)
    --- PASS: TestFreezeBeforeSet/slice,will-freeze (0.07s)
    --- PASS: TestFreezeBeforeSet/slice,pre-frozen,will-freeze (0.06s)
    --- PASS: TestFreezeBeforeSet/scope,pre-frozen,skip-devices (0.09s)
    --- PASS: TestFreezeBeforeSet/scope,pre-frozen (0.05s)
```

I've seen similar bugs but not this one. Filing just to ensure it's not lost.

kolyshkin:
In general we seem to have a problem with disappearing cgroups on CentOS 7, e.g. from https://cirrus-ci.com/task/5772431724904448 (while testing #3379):

```
=== RUN   TestPodSkipDevicesUpdate
    systemd_test.go:86: failed to write 6485: write /sys/fs/cgroup/cpuset/system.slice/system-runc_test_podTestPodSkipDevicesUpdate.slice/test-PodSkipDevicesUpdate.scope/cgroup.procs: no such device
--- FAIL: TestPodSkipDevicesUpdate (0.75s)
```

kolyshkin:
See also #3351

kolyshkin:
Similar but a different test:

```
=== RUN   TestSkipDevicesFalse
[...]

PR #3580, opened 2022-08-29
seccomp: do not ignore SPEC_ALLOW flag

kolyshkin:
Commit 58ea21daefea8e3447db added support for seccomp flags such as
SPEC_ALLOW, but it does not work, as since commit 7a8d7162f9d72f20d83ea
we do not use libseccomp-golang's Load(), but handle flags separately in patchbfp.

This fixes setting SPEC_ALLOW flag.

Add a comment to not forget to amend `filterFlags` when adding new flags.

Add a test case that the flags are set as expected (see the second commit for details).

Fixes: https://github.com/opencontainers/runc/issues/3582

### DIFF
--- a/libcontainer/seccomp/patchbpf/enosys_linux.go
+++ b/libcontainer/seccomp/patchbpf/enosys_linux.go
@@ -43,6 +43,11 @@ const uintptr_t C_SET_MODE_FILTER = SECCOMP_SET_MODE_FILTER;
 #endif
 const uintptr_t C_FILTER_FLAG_LOG = SECCOMP_FILTER_FLAG_LOG;
 
+#ifndef SECCOMP_FILTER_FLAG_SPEC_ALLOW
+#	define SECCOMP_FILTER_FLAG_SPEC_ALLOW (1UL << 2)
+#endif
+const uintptr_t C_FILTER_FLAG_SPEC_ALLOW = SECCOMP_FILTER_FLAG_SPEC_ALLOW;
+
 #ifndef SECCOMP_FILTER_FLAG_NEW_LISTENER
 #	define SECCOMP_FILTER_FLAG_NEW_LISTENER (1UL << 3)
 #endif
@@ -641,8 +646,13 @@ func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags 
 			flags |= uint(C.C_FILTER_FLAG_LOG)
 		}
 	}
-
-	// TODO: Support seccomp flags not yet added to libseccomp-golang...
+	if apiLevel >= 4 {
+		if ssb, err := filter.GetSSB(); err != nil {
+			return 0, false, fmt.Errorf("unable to fetch SECCOMP_FILTER_FLAG_SPEC_ALLOW bit: %w", err)
+		} else if ssb {
+			flags |= uint(C.C_FILTER_FLAG_SPEC_ALLOW)
+		}
+	}
 
 	for _, call := range config.Syscalls {
 		if call.Action == configs.Notify {
--- a/libcontainer/seccomp/patchbpf/enosys_linux.go
+++ b/libcontainer/seccomp/patchbpf/enosys_linux.go
@@ -665,6 +665,9 @@ func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags 
 }
 
 func sysSeccompSetFilter(flags uint, filter []unix.SockFilter) (fd int, err error) {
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
Would it be possible to have a test,?

libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -641,8 +646,13 @@ func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags
 			flags |= uint(C.C_FILTER_FLAG_LOG)
 		}
 	}
-
-	// TODO: Support seccomp flags not yet added to libseccomp-golang...
```
haircommander, line 645: is this the only flag not supported by libseccomp-golang? If not, is this comment still relevant?

kolyshkin: Currently, libseccomp-golang is on par with the runtime-spec, thus I removed this comment.

kolyshkin:
> Would it be possible to have a test?

I can't think of a way to check if SPEC_ALLOW (or any other flag) is set in a particular ebpf loaded into the kernel, thus no way to test this. If anyone knows such a way, please share.

kolyshkin:
> > Would it be possible to have a test?
> 
> I can't think of a way to check if SPEC_ALLOW (or any other flag) is set in a particular ebpf loaded into the kernel, thus no way to test this. If anyone knows such a way, please share.

Also, we already have a test, added in #3390. The problem is, it doesn't check whether the flags are actually set -- merely that they are accepted.

Hmm, we can actually use strace to check which flags are being set! Let me see...

kolyshkin:
It is still possible to use strace from the test, but I did something different -- modified the code to actually debug print the flags value, and amended the seccomp flags setting test to check if the flags value is as expected.

@AkihiroSuda PTAL

tests/integration/seccomp.bats
```diff
@@ -70,31 +70,47 @@ function teardown() {
...
+		*)
+			update_config ' .linux.seccomp.flags = [ '"${key}"' ]'
+			;;
+		esac
+
+		runc --debug run test_busybox
 		[ "$status" -ne 0 ]
 		[[ "$output" == *"mkdir:"*"/dev/shm/foo"*"Operation not permitted"* ]]
+
+		# Check the numeric flags value is as expected.
```
AkihiroSuda, line 110:
```suggestion
		# Check the numeric flags value is printed in the debug log as expected.
```

libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -655,6 +665,7 @@ func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags
 }
 
 func sysSeccompSetFilter(flags uint, filter []unix.SockFilter) (fd int, err error) {
+	logrus.Debugf("seccomp filter flags: %d", flags)
```
AkihiroSuda, line 668:
```suggestion
         // this debug print is validated in tests/integration/seccomp.bats
	logrus.Debugf("seccomp filter flags: %d", flags)
```

AkihiroSuda:
Thanks, LGTM but a couple of nits

kolyshkin:
> Thanks, LGTM but a couple of nits

Addressed.

AkihiroSuda (approved):

kolyshkin:
CentOS 7 failure is a flake (https://github.com/opencontainers/runc/issues/3540); restarted.

libcontainer/seccomp/patchbpf/enosys_linux.go
alban: Currently correct, but I'd like support for SECCOMP_FILTER_FLAG_WAIT_KILLABLE_RECV (Linux 5.19, https://github.com/torvalds/linux/commit/c2aa2dfef243).

https://github.com/opencontainers/runtime-spec/pull/1161

Not yet in libseccomp: https://github.com/seccomp/libseccomp/issues/387

alban:
I am not a maintainer and I didn't test it but the patch looks good to me.

libcontainer/seccomp/patchbpf/enosys_linux.go
kolyshkin: Right, but I guess we'll support it in libseccomp and libseccomp-golang first, when here in runc.

mrunalp (approved):

PR #3581, opened 2022-08-30
seccomp: set SPEC_ALLOW by default

kolyshkin:
If no seccomps flags are set in OCI runtime spec (not even the empty
set), set SPEC_ALLOW by default. Otherwise, use the flags set.
    
This mimics the crun behavior, and makes runc seccomp performance on par
with crun.

The test case added in #3580 checks all the cases:
 - no flags (this is when we set SPEC_ALLOW)
 - empty set of flags (this is when we don't set SPEC_ALLOW)
 - specific flags set

_This currently includes #3580, thus a draft pending its merge._

### DIFF
--- a/libcontainer/seccomp/patchbpf/enosys_linux.go
+++ b/libcontainer/seccomp/patchbpf/enosys_linux.go
@@ -43,6 +43,11 @@ const uintptr_t C_SET_MODE_FILTER = SECCOMP_SET_MODE_FILTER;
 #endif
 const uintptr_t C_FILTER_FLAG_LOG = SECCOMP_FILTER_FLAG_LOG;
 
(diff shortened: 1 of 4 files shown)


kolyshkin:
> # time="2022-08-31T02:39:05Z" level=error msg="runc run failed: unable to start container process: unable to init seccomp: error adding SSB flag to seccomp filter: SetSSB requires libseccomp >= 2.5.0 and API level >= 4 (current version: 2.3.1, API level: 0)"

OK, we should check if the flag is supported by the running kernel/libseccomp before adding it.

kolyshkin:
Closing in favor of #3588 (which does the same thing and more).

kolyshkin:
Did I say close?

PR #3588, opened 2022-09-02
seccomp: refactor flags support; add flags to features, set SPEC_ALLOW by default

kolyshkin:
~~_Currently a draft pending #3580 merge. When reviewing as a draft, please skip the first 2 commits (they are from #3580)._~~

This PR reworks seccomp flags support:
 - adds new functions `KnownFlags()`, `SupportedFlags()` and `FlagSupported()` to seccomp pkg;
 - adds showing of known and supported seccomp filter flags to `runc features`;
 - consolidates/deduplicates code that handles flags (as much as possible);
 - changes the default (when no flags are set) to have `SPEC_ALLOW`, if supported (crun does that).
 
~~TODO: fix the seccomp flags test to use runc features and not depend on kernel version (maybe in a separate PR).~~

### DIFF
--- a/libcontainer/seccomp/config.go
+++ b/libcontainer/seccomp/config.go
@@ -5,7 +5,12 @@ import (
 	"sort"
 
 	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
+
+// flagTsync is recognized but ignored by runc, and it is not defined
+// in the runtime-spec.
+const flagTsync = "SECCOMP_FILTER_FLAG_TSYNC"
 
 var operators = map[string]configs.Operator{
 	"SCMP_CMP_NE":        configs.NotEqualTo,
@@ -110,4 +115,36 @@ func ConvertStringToArch(in string) (string, error) {
 		return arch, nil
 	}
 	return "", fmt.Errorf("string %s is not a valid arch for seccomp", in)
+}
+
+// List of flags known to this version of runc.
+var flags = []string{
+	flagTsync,
+	string(specs.LinuxSeccompFlagSpecAllow),
+	string(specs.LinuxSeccompFlagLog),
+}
+
+// KnownFlags returns the list of the known filter flags.
+// Used by `runc features`.
+func KnownFlags() []string {
+	return flags
+}
+
+// SupportedFlags returns the list of the supported filter flags.
+// This list may be a subset of one returned by KnownFlags due to
+// some flags not supported by the current kernel and/or libseccomp.
+// Used by `runc features`.
+func SupportedFlags() []string {
+	if !Enabled {
+		return nil
+	}
+
+	var res []string
+	for _, flag := range flags {
+		if FlagSupported(specs.LinuxSeccompFlag(flag)) == nil {
+			res = append(res, flag)
+		}
+	}
+
+	return res
 }
--- a/libcontainer/seccomp/patchbpf/enosys_linux.go
+++ b/libcontainer/seccomp/patchbpf/enosys_linux.go
@@ -643,6 +643,7 @@ func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags 
 			flags |= uint(C.C_FILTER_FLAG_SPEC_ALLOW)
 		}
 	}
+	// XXX: add newly supported filter flags above this line.
 
 	for _, call := range config.Syscalls {
 		if call.Action == configs.Notify {
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -87,27 +87,10 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 		}
 	}
 
-	// Add extra flags
+	// Add extra flags.
 	for _, flag := range config.Flags {
-		switch flag {
-		case "SECCOMP_FILTER_FLAG_TSYNC":
-			// libseccomp-golang always use filterAttrTsync when
-			// possible so all goroutines will receive the same
-			// rules, so there is nothing to do. It does not make
-			// sense to apply the seccomp filter on only one
-			// thread; other threads will be terminated after exec
-			// anyway.
-		case specs.LinuxSeccompFlagLog:
-			if err := filter.SetLogBit(true); err != nil {
-				return -1, fmt.Errorf("error adding log flag to seccomp filter: %w", err)
-			}
-		case specs.LinuxSeccompFlagSpecAllow:
-			if err := filter.SetSSB(true); err != nil {
-				return -1, fmt.Errorf("error adding SSB flag to seccomp filter: %w", err)
-			}
-		// NOTE when adding more flags, make sure to also modify filterFlags in patchbpf.
-		default:
-			return -1, fmt.Errorf("seccomp flags %q not yet supported by runc", flag)
+		if err := setFlag(filter, flag); err != nil {
+			return -1, err
 		}
 	}
 
@@ -147,6 +130,67 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 	}
 
 	return seccompFd, nil
+}
+
+type unknownFlagError struct {
+	flag specs.LinuxSeccompFlag
+}
+
+func (e *unknownFlagError) Error() string {
+	return "seccomp flag " + string(e.flag) + " is not known to runc"
+}
+
+func setFlag(filter *libseccomp.ScmpFilter, flag specs.LinuxSeccompFlag) error {
+	switch flag {
+	case flagTsync:
+		// libseccomp-golang always use filterAttrTsync when
+		// possible so all goroutines will receive the same
+		// rules, so there is nothing to do. It does not make
+		// sense to apply the seccomp filter on only one
+		// thread; other threads will be terminated after exec
+		// anyway.
+		return nil
+	case specs.LinuxSeccompFlagLog:
+		if err := filter.SetLogBit(true); err != nil {
+			return fmt.Errorf("error adding log flag to seccomp filter: %w", err)
+		}
+		return nil
+	case specs.LinuxSeccompFlagSpecAllow:
+		if err := filter.SetSSB(true); err != nil {
+			return fmt.Errorf("error adding SSB flag to seccomp filter: %w", err)
+		}
+		return nil
+	}
+	// NOTE when adding more flags above, do not forget to also:
+	// - add new flags to `flags` slice in config.go;
+	// - add new flags to tests/integration/seccomp.bats flags test;
+	// - modify func filterFlags in patchbpf/ accordingly.
+
+	return &unknownFlagError{flag: flag}
+}
+
+// FlagSupported checks if the flag is known to runc and supported by
+// currently used libseccomp and kernel (i.e. it can be set).
+func FlagSupported(flag specs.LinuxSeccompFlag) error {
+	filter := &libseccomp.ScmpFilter{}
+	err := setFlag(filter, flag)
+
+	// For flags we don't know, setFlag returns unknownFlagError.
+	var uf *unknownFlagError
+	if errors.As(err, &uf) {
+		return err
+	}
+	// For flags that are known to runc and libseccomp-golang but can not
+	// be applied because either libseccomp or the kernel is too old,
+	// seccomp.VersionError is returned.
+	var verErr *libseccomp.VersionError
+	if errors.As(err, &verErr) {
+		// Not supported by libseccomp or the kernel.
+		return err
+	}
+
+	// All other flags are known and supported.
+	return nil
 }
 
 // Convert Libcontainer Action to Libseccomp ScmpAction
--- a/libcontainer/seccomp/seccomp_linux.go
+++ b/libcontainer/seccomp/seccomp_linux.go
@@ -163,7 +163,7 @@ func setFlag(filter *libseccomp.ScmpFilter, flag specs.LinuxSeccompFlag) error {
 	}
 	// NOTE when adding more flags above, do not forget to also:
 	// - add new flags to `flags` slice in config.go;
-	// - add new flags to tests/integration/seccomp.bats flags test;
+	// - add new flag values to flags_value() in tests/integration/seccomp.bats;
 	// - modify func filterFlags in patchbpf/ accordingly.
 
 	return &unknownFlagError{flag: flag}
--- a/tests/integration/seccomp.bats
+++ b/tests/integration/seccomp.bats
@@ -66,11 +66,32 @@ function teardown() {
 	[[ "$output" == *"Network is down"* ]]
 }
 
-@test "runc run [seccomp] (SECCOMP_FILTER_FLAG_*)" {
-	# Linux 4.14: SECCOMP_FILTER_FLAG_LOG
-	# Linux 4.17: SECCOMP_FILTER_FLAG_SPEC_ALLOW
-	requires_kernel 4.17
+# Prints the numeric value of provided seccomp flags combination.
+# The parameter is flags string, as supplied in OCI spec, for example
+# '"SECCOMP_FILTER_FLAG_TSYNC","SECCOMP_FILTER_FLAG_LOG"'.
+function flags_value() {
+	# Numeric values of seccomp flags.
+	declare -A values=(
+		['"SECCOMP_FILTER_FLAG_TSYNC"']=0 # Supported but ignored by runc, thus 0.
+		['"SECCOMP_FILTER_FLAG_LOG"']=2
+		['"SECCOMP_FILTER_FLAG_SPEC_ALLOW"']=4
+		# XXX: add new values above this line.
+	)
+	# Split the flags.
+	IFS=',' read -ra flags <<<"$1"
 
+	local flag v sum=0
+	for flag in "${flags[@]}"; do
+		# This will produce "values[$flag]: unbound variable"
+		# error for a new flag yet unknown to the test.
+		v=${values[$flag]}
+		((sum += v)) || true
+	done
+
+	echo $sum
+}
+
+@test "runc run [seccomp] (SECCOMP_FILTER_FLAG_*)" {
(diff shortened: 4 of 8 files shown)


kolyshkin:
@rata @alban I'll be much obliged if you can PTAL.

rata:
@kolyshkin thanks for the cc! I'll be on and off for a few weeks. Is this something we want for the next runc release and I should try to prioritize its review? Or can it wait a few weeks maybe?

alban:
> changes the default (when no flags are set) to have SPEC_ALLOW, if supported (crun does that).

It is done here:
https://github.com/containers/crun/blob/0b3086af0926146ac13ebdb8639bbbd3c93a2617/src/libcrun/seccomp.c#L198-L200

The man page says:

> SECCOMP_FILTER_FLAG_SPEC_ALLOW (since Linux 4.17)
> Disable Speculative Store Bypass mitigation.

Does it mean it is less secure with this flag?

kolyshkin:
> > changes the default (when no flags are set) to have SPEC_ALLOW, if supported (crun does that).
> 
> It is done here: https://github.com/containers/crun/blob/0b3086af0926146ac13ebdb8639bbbd3c93a2617/src/libcrun/seccomp.c#L198-L200
> 
> The man page says:
> 
> > SECCOMP_FILTER_FLAG_SPEC_ALLOW (since Linux 4.17)
> > Disable Speculative Store Bypass mitigation.
> 
> Does it mean it is less secure with this flag?

It sort of is -- meaning, if someone will find a way to use seccomp ebpf program to exploit the hardware SSB issue. Practically, this seems highly unlikely, since the seccomp ebpf code is generated by libseccomp. IOW a user do not have direct control over ebpf code being loaded into the kernel, they can only influence is via the seccomp rules, which makes it much harder to exploit the vulnerability.

kolyshkin:
Also, it seems that in kernel v5.16 this (disabling the mitigations for seccomp) is the default (according to https://git.kernel.org/pub/scm/linux/kernel/git/kees/linux.git/commit/?h=for-next/seccomp&id=2f46993d83ff4abb310ef7b4beced56ba96f0d9d, which made its way into v5.16). By the way, the commit mistakenly claims runc sets this flag by default -- it does not (until this PR is merged, that is).

libcontainer/seccomp/patchbpf/enosys_linux.go
```diff
@@ -665,6 +665,9 @@ func filterFlags(config *configs.Seccomp, filter *libseccomp.ScmpFilter) (flags
 }
 
 func sysSeccompSetFilter(flags uint, filter []unix.SockFilter) (fd int, err error) {
+	// This debug output is validated in tests/integration/seccomp.bats
+	// by the SECCOMP_FILTER_FLAG_* test.
+	logrus.Debugf("seccomp filter flags: %d", flags)
```
rata, line 670: Any reason to not use `%v`? I find it more clear to use %v unless you want something else. Isn't %v printing the same here?

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -135,6 +118,67 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
 	return seccompFd, nil
 }
 
+type unknownFlagError struct {
+	flag specs.LinuxSeccompFlag
+}
+
+func (e *unknownFlagError) Error() string {
+	return "seccomp flag " + string(e.flag) + " is not known to runc"
```
rata, line 126: not known? Not supported seems a more standard way to say this

tests/integration/seccomp.bats
```diff
@@ -70,31 +70,47 @@ function teardown() {
...
+	declare -A FLAGS=(
+		['REMOVE']=0 # No setting, use built-in default.
+		['EMPTY']=0  # Empty set of flags.
+		['"SECCOMP_FILTER_FLAG_LOG"']=2
+		['"SECCOMP_FILTER_FLAG_SPEC_ALLOW"']=4
+		['"SECCOMP_FILTER_FLAG_TSYNC"']=0 # tsync flag is ignored.
+		['"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_SPEC_ALLOW"']=6
+		['"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_TSYNC"']=2
+		['"SECCOMP_FILTER_FLAG_SPEC_ALLOW","SECCOMP_FILTER_FLAG_TSYNC"']=4
+		['"SECCOMP_FILTER_FLAG_LOG","SECCOMP_FILTER_FLAG_SPEC_ALLOW","SECCOMP_FILTER_FLAG_TSYNC"']=6
```
rata, line 91: A comment explaining where you got these values seems useful for future collaborators.

Also, the tsync ignored comment can either be in all instances of tsync being used or a more visible comment on the top, maybe?

rata (approved):
@kolyshkin Thanks, on a cursory look, it LGTM.

I have not tried hard to see if more code simplifications are possible. But the PR seems reasonable to me :)

libcontainer/seccomp/patchbpf/enosys_linux.go
kolyshkin: So, `%d` says "print argument as an integer", and `%v` means "print argument using the default formatting for its type". Since the type is integer, in this case %v will result in the same output as %d. The only difference (I guess but not entirely sure), is %v incurs slightly more runtime overhead, and the only benefit of using %v here would be that we won't have to change it if we change the `flags` type.

Since this is clearly an integer, I don't see why we shouldn't use `%d`. Personally, I find it more clear, as I can see we're printing an integer without looking at the flags type.

So I have to ask the same question to you -- why not use %d?

libcontainer/seccomp/seccomp_linux.go
kolyshkin: I think I copy-pasted that from the earlier code. And yes, I agree that "not supported" seems more common language, OTOH "more known" is more specific in this particular case.

"Not supported" is sort of generic, and can mean multiple things, such as the support for this feature or flags is
 - not yet added;
 - dropped;
 - not compiled it;
 - is impossible because of the current hardware or the kernel;
 and so on.
 
"Not known" is more specific, meaning this runc version is not aware that such flag exists (so, if it is exists, this is a clear hint that the runc version used is too old).

tests/integration/seccomp.bats
kolyshkin: I wanted to rework this test so that I won't have to add all the combinations manually. Guess this is what I'll do now :)

libcontainer/seccomp/patchbpf/enosys_linux.go
kolyshkin: OK, it seems that you have ignored the description and reviewed the first two commits, which are actually from https://github.com/opencontainers/runc/pull/3580 and were merged in there.

libcontainer/seccomp/seccomp_linux.go
kolyshkin: So, if you take a look at the code adding flags to `runc features`, you'll see there are "known" and "supported" set of flags (with supported being a subset of known). This is why I'm using the word "known" here.

tests/integration/seccomp.bats
kolyshkin: Reworked the test (see the last commit). Beware: hardcode bash.

libcontainer/seccomp/seccomp_linux.go
rata: :+1:

libcontainer/seccomp/patchbpf/enosys_linux.go
rata: Oh, yes, I missed the description. Sorry!

I see reasons to one or the other, I personally prefer the other and was wondering if this was on purpose. SGTM if you prefer it :)

tests/integration/seccomp.bats
```diff
@@ -79,18 +75,68 @@ function teardown() {
...
+	# Check that any new seccomp flags added are actually tested.
+	for f in "${sup_flags[@]}"; do
+		if ! [ -v 'flags[$f]' ]; then
+			echo "ERROR: seccomp flag $f not added to the test; fix the text!" 1>&2
+			exit 1
+		fi
+	done
+
+	# Filter out flags not supported by the current seccomp/kernel,
+	# so this test can be run on older systems.
```
rata, line 105: I'm sure I'm missing something, but why hardcode the vars in the `flags` array and then ignore the not supported ones? Wouldn't be the same to just test all the supported ones? (and we don't harcode any flag here)

I see we might get unlucky in the CI and maybe some flag is never tested, and that is interesting to know. But we are not detecting that here either, right? Maybe that can be a different test for that (well, not sure it is easy to detect that... and seems unlikely)?

rata (approved):
LGTM, one minor question

tests/integration/seccomp.bats
kolyshkin: Since this test is actually checking that the flags are set as expected (the check was added by commit 26dc55ef1a56ea02794 after we found out in #3580 that SPEC_ALLOW is just ignored), we need to know their numeric values.

Surely, we can just get the list of supported flags (which we should test) from `runc features`, but then again we still need their values.

Let me see if I can refactor the code to simplify it more (use supported flags as received from runc directly, and only use the hardcoded ones for numeric values).

rata: Oh, and wouldn't make sense for `runc features` to also print the numeric values (if we can easily parse it here, not worth the pain if this is hard)?

But the code is okay as it is now too from my POV :)

kolyshkin: > Oh, and wouldn't make sense for runc features to also print the numeric values (if we can easily parse it here, not worth the pain if this is hard)?

Well, no one ever needs the numeric values (except for this test), and `runc features` target audience is runc users, not our tests.

Thus we have to hardcode it (unfortunately there is no easy way to see if a seccomp flag is set or not, so the test have to rely on debug output and compare the numbers).

kolyshkin:

types/features/features.go
```diff
@@ -53,13 +53,21 @@ type Seccomp struct {
...
 	Operators []string `json:"operators,omitempty"`
 
-	// Operators is the list of the recognized archs, e.g., "SCMP_ARCH_X86_64".
+	// Archs is the list of the recognized archs, e.g., "SCMP_ARCH_X86_64".
 	// Nil value means "unknown", not "no support for any arch".
 	Archs []string `json:"archs,omitempty"`
+
+	// KnownFlags is the list of the recognized filter flags, e.g., "SECCOMP_FILTER_FLAG_LOG".
+	// Nil value means "unknown", not "no support for any flag".
+	KnownFlags []string `json:"known_flags,omitempty"`
```
AkihiroSuda, line 66:
```suggestion
	KnownFlags []string `json:"knownFlags,omitempty"`
```

types/features/features.go
```diff
@@ -53,13 +53,21 @@ type Seccomp struct {
...
+
+	// Flags is the list of the supported filter flags, e.g., "SECCOMP_FILTER_FLAG_LOG".
+	// Nil value means "unknown", not "no support for any flag".
+	SupportedFlags []string `json:"supported_flags,omitempty"`
```
AkihiroSuda, line 70:
```suggestion
	SupportedFlags []string `json:"supportedFlags,omitempty"`
```

types/features/features.go
```diff
@@ -53,13 +53,21 @@ type Seccomp struct {
...
-	// Operators is the list of the recognized archs, e.g., "SCMP_ARCH_X86_64".
+	// Archs is the list of the recognized archs, e.g., "SCMP_ARCH_X86_64".
 	// Nil value means "unknown", not "no support for any arch".
 	Archs []string `json:"archs,omitempty"`
+
+	// KnownFlags is the list of the recognized filter flags, e.g., "SECCOMP_FILTER_FLAG_LOG".
+	// Nil value means "unknown", not "no support for any flag".
+	KnownFlags []string `json:"known_flags,omitempty"`
+
+	// Flags is the list of the supported filter flags, e.g., "SECCOMP_FILTER_FLAG_LOG".
```
AkihiroSuda, line 68: Doesn't match the variable name

AkihiroSuda: How is "supported" different from "recognized"?

Having two concepts doesn't seem consistent with other properties in this JSON

kolyshkin: oops, fixed

types/features/features.go
kolyshkin: Fixed, thanks!

types/features/features.go
kolyshkin: fixed

types/features/features.go
kolyshkin: In this context "recognized" means known, i.e. runc is aware of the flag (the flag is known to runc).

OTOH "supported" means this flag is not merely known to (recognized by) runc, but is also supported by the currently used libseccomp and the running kernel.

If you don't like such distinction, we can only print one set -- choose which one makes more sense. To be, "supported" makes more sense than "known" as despite the flag being known you can't use it anyway. The downside is you don't know if more flags would be supported after e.g. a kernel upgrade.

tests/integration/seccomp.bats
rata: Ohh, ok. I thought it might have been present on some errors as hex or something maybe. But makes sense, in any case it is not so useful either

kolyshkin:

types/features/features.go
kolyshkin: @AkihiroSuda PTAL 🙏🏻

kolyshkin:

libcontainer/seccomp/config.go
```diff
@@ -111,3 +112,35 @@ func ConvertStringToArch(in string) (string, error) {
 	}
 	return "", fmt.Errorf("string %s is not a valid arch for seccomp", in)
 }
+
+// List of flags known to this version of runc.
+var flags = []string{
+	"SECCOMP_FILTER_FLAG_TSYNC",
```
mrunalp, line 118: Does this need to be added to the spec?

libcontainer/seccomp/seccomp_linux.go
```diff
@@ -135,6 +118,67 @@ func InitSeccomp(config *configs.Seccomp) (int, error) {
...
+}
+
+func setFlag(filter *libseccomp.ScmpFilter, flag specs.LinuxSeccompFlag) error {
+	switch flag {
+	case "SECCOMP_FILTER_FLAG_TSYNC":
```
mrunalp, line 131: Should we define a contant for this?

libcontainer/seccomp/config.go
kolyshkin: We recognize this flag, as it is a valid flag, but we ignore it, since in our context it does not make sense. This is why it is not added to the spec (since it does not do anything for the container).

libcontainer/seccomp/seccomp_linux.go
kolyshkin: done

kolyshkin:
@mrunalp addressed your review comments

mrunalp (dismissed):

kolyshkin:
@cyphar @AkihiroSuda @thaJeztah PTAL

kolyshkin:
@cyphar @AkihiroSuda @thaJeztah PTAL

mrunalp (dismissed):

types/features/features.go
kolyshkin: @AkihiroSuda let me add some more context here.

All the "features" that we have in `runc features` are "known features", meaning they are recognized by this runc build. Some of those (like mount flags, for example) can not be used because the kernel version is too old, some library version is old, or a binary is missing. So, the feature is known but not supported.

For the sake of the seccomp flags test case, I need to have seccomp flags that are supported by the current runtime (meaning runc + libraries + kernel), so I added a distinction between "known" and "supported" flags.

I guess that the use case for this (distinguishing between known and supported flags) can also be used by the upper level (container orchestration) to query the runtime platform capabilities. Say, some mount attributes (like "rro") might be known but not supported due to older kernel.

I agree that this distinction doesn't really fit into current output of `runc features`, since we print known (rather than supported) features and flags.

Do you have any idea how to add such a distinction (between known and supported), if we ever want to add it? Something like `runc features --supported` (and maybe `--known` for the sake of completeness?).

It is really hard to write a seccomp flags test (in bash, that is) without relying on `runc features` to print the list of supported seccomp flags, but I guess this is a separate issue. Perhaps I can rewrite the test in Go (and make it the first runc integration test written in Go).

AkihiroSuda (approved):
Thanks, sorry for delay in review.

LGTM but needs https://github.com/opencontainers/runc/pull/3700

AkihiroSuda:
Can't click the merge button until getting one more LGTM 😞 

Opened a proposal to uncheck the  `Dismiss stale pull request approvals when new commits are pushed` checkbox
- https://github.com/opencontainers/runc/issues/3706

dqminh (approved):

dqminh:
LGTM !

bobbypage:
Is there any thoughts around backporting this change to runc 1.1?

The change in https://github.com/opencontainers/runc/pull/3588/commits/ac04154f0b136ad3f8636acb8ed29431b7bddd83 is especially important in context of k8s graduating secomp to default in 1.27 (https://github.com/kubernetes/enhancements/pull/3718).

The change in https://github.com/opencontainers/runc/pull/3588/commits/ac04154f0b136ad3f8636acb8ed29431b7bddd83 is critical for performance as otherwise there is very significant [perf degradation with seccomp enabled](https://github.com/kubernetes/enhancements/pull/3718#issuecomment-1416543001)

rata:
@bobbypage I'll be fine with it. Although, I'd prefer if we can make a smaller change to just enable SPEC_ALLOW by default (that is what you need, right?) and have the rest of this patch land in runc 1.2. That can be potentially way simpler (i hope :)). I've not tried to see if it is easy to create a backport a simpler patch just for the SPEC_ALLOW by default case, though.

Any maintainers has thoughts on this?

PR #3590, opened 2022-09-06
execfifo maybe be blocked all the time if containerd is killed in abnomal scenario

kamizjw:
Openat will be blocked until the fifo on the other side in opened, but in some

abnomal scenario(e.g. containerd is killed), Openat maybe be blocked all the time.

Signed-off-by: zhongjiawei <zhongjiawei1@huawei.com>

### DIFF
--- a/libcontainer/standard_init_linux.go
+++ b/libcontainer/standard_init_linux.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"strconv"
+	"time"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux"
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
How can we test this?

libcontainer/standard_init_linux.go
```diff
@@ -233,13 +234,26 @@ func (l *linuxStandardInit) Init() error {
...
+		}
+		ch <- nil
+	}()
+
+	select {
+	case chErr := <-ch:
+			if chErr != nil {
+					return chErr
+			}
+	case <-time.After(120 * time.Second):
```
fuweid, line 255: To be honest, it is possible to call runC start after 2mins. And I don't think the runC-init process is leaky because contained got killed. 

It is possible if the user uses containerd-shim-v1, because the resource cleanup requires shim-v1 server is alive. But the community deprecates containerd-shim-v1 and suggest that user uses v2. In v2 design, the containerd can make sure the resource created by shim will be cleanup, including runC-init process.

IMO, the runC-init process should be cleanup by caller, instead of subside.

kolyshkin: I also think this is trying to fix the problem on the wrong side. By design, runc waits indefinitely, and if I remember correctly the way to clean it up is to call runc kill.

kamizjw:
> How can we test this?

libcontainer/standard_init_linux.go
kamizjw: shmi-v2 also doesn't fix the init leak if the containerd exits unexpectedly like a segmentfault

kolyshkin:
OK, so this is a breaking change.

Currently, `runc start` can be called at any time after `runc create`. We don't know if there are any users that rely on the current behavior.

Overall, I think, something like this should be done, but with a change to documentation and an ability to revert to an old behavior, or making the old behavior the default, and the new one optional (say add `--timeout` flag for runc create`).

WDYT @AkihiroSuda @cyphar ?

kamizjw:
> 

add timeout option is good idea  for any scenario

libcontainer/standard_init_linux.go
fuweid: @kamizjw could you mind to share how to reproduce this? If it is bug in containerd, I would like to fix it in containerd site.

fuweid:
The timeout can be option. But it might be hard to integrate with existing system. Currently, many scenarios has timeout during the RPC call. IMO, if we can make sure that there is always record (runC state), the containerd/CRI-O/other components can cleanup it. Just my two cents.

cyphar:
NACK. The semantics of `runc create` and `runc start` are defined by the spec. There is no mechanism for us to have a timeout (nor should there be one built-in IMHO -- it is trivial for the runc user to implement a timeout by calling `runc start` after the desired timeout period has elapsed). It seems like there is a containerd bug which this is attempting to fix -- I humbly suggest fixing the issue in containerd.

If containerd wants to avoid this entirely, they can use `runc run` but I suspect they're using the `create`+`start` mechanism the way it is intended (to do various operations on the container before the main process is properly started). As such, it is possible for some operation to take longer than expected, and a timeout would result in the main process starting before the container is ready.

PR #3593, opened 2022-09-06
update state earlier to avoid cgroup leak when process failed

kamizjw:
update state earlier to avoid cgroup leak when process failed
if process stuck in somewhere. upper caller like containerd may
have a timeout for process launching.

process will be killed after this timeout, and then call `runc
delete` to retrieve its resource like cgroup and perform poststop
hook.

if process got stuck right before updating state, and after cgroup
applied, like prestart-hook. In such case, `runc delete xxx` will
do nothing because state file is missing, runc is not aware of this
container. so process cgroup will stay and never get removed.

Signed-off-by: zhongjiawei <zhongjiawei1@huawei.com>

### DIFF
--- a/libcontainer/process_linux.go
+++ b/libcontainer/process_linux.go
@@ -416,6 +416,12 @@ func (p *initProcess) start() (retErr error) {
 			return fmt.Errorf("unable to apply Intel RDT configuration: %w", err)
 		}
 	}
+	// update state here, so we can retrieve process resource
+	// even it get killed by accident
+	if _, err := p.container.updateState(p); err != nil {
+		return err
+	}
+
 	if _, err := io.Copy(p.messageSockPair.parent, p.bootstrapData); err != nil {
 		return fmt.Errorf("can't copy bootstrap data to pipe: %w", err)
 	}


kolyshkin:
@kamizjw thank you for your PR. Do you think it's possible to create a test case for this issue?

kamizjw:
> @kamizjw thank you for your PR. Do you think it's possible to create a test case for this issue?

Does constructing a scenario verification that exits unexpectedly cause leakage?

kolyshkin:
> > @kamizjw thank you for your PR. Do you think it's possible to create a test case for this issue?
> 
> Does constructing a scenario verification that exits unexpectedly cause leakage?

Something like this. Maybe you can use prestart-hook to do that.

Having a test case as part of the PR would be very nice.

kamizjw:
> > > @kamizjw thank you for your PR. Do you think it's possible to create a test case for this issue?
> > 
> > 
> > Does constructing a scenario verification that exits unexpectedly cause leakage?
> 
> Something like this. Maybe you can use prestart-hook to do that.
> 
> Having a test case as part of the PR would be very nice.


I'm not good at writing go test cases, can you help me write one？thank you very much

kamizjw:
> > I'm not good at writing go test cases, can you help me write one？thank you very much
> 
> Write any kind of reproducer for the issue you're solving with this patch. This can be a script, or a text description of what steps to take. Once we have it, I'll help you to convert a reproducer to the test case.

1.Prepare a script to execute in the prestart-hook phase, which takes a while to execute
2.Configure the execution of this script under the prestart option of the configuration file
3.Make sure that the runc delete command is executed when runc runs the prestart-hook

kolyshkin:
> update state earlier to avoid cgroup leak when process failed
> if process stuck in somewhere. upper caller like containerd may
> have a timeout for process launching.

Perhaps the best approach to avoid this situation is to specify the hook timeout for the prestart hook, and make sure it is less than the internal containerd timeout.

kamizjw:
> 

hook timeout 

> > update state earlier to avoid cgroup leak when process failed
> > if process stuck in somewhere. upper caller like containerd may
> > have a timeout for process launching.
> 
> Perhaps the best approach to avoid this situation is to specify the hook timeout for the prestart hook, and make sure it is less than the internal containerd timeout.

this needs to consider different application scenarios, and updating state.json immediately is the most concise way

Issue #2532, opened 2020-07-29
too many mount/umount syscalls


axot:
I'm not sure this is the right place to discuss. We are observing high cpu usage by systemd init process[1]. After some digging, it is possible caused by many mount/umount syscalls. Is it because runc was be executed?[3] At same time there is no new pod be scheduled to this instance or be deleted. What is the purpose runc was executed in this case?

Thanks!

[1] high cpu usage init process.
![image](https://user-images.githubusercontent.com/611054/88808502-ed7bf680-d1ed-11ea-8ca4-eae5d9e896cc.png)

[2]
```
/# mountsnoop
COMM             PID     TID     MNT_NS      CALL
exe              1774545 1774545 4026531840  mount("/proc/self/exe", "/var/run/docker/runtime-runc/moby/910fdb2d258cb4e4c504d1414c9c635074d64aa4554671123d4cc652c007dd2e/runc.GC2atV", "", MS_BIND, "") = 0
exe              1774545 1774545 4026531840  mount("", "/var/run/docker/runtime-runc/moby/910fdb2d258cb4e4c504d1414c9c635074d64aa4554671123d4cc652c007dd2e/runc.GC2atV", "", MS_RDONLY|MS_REMOUNT|MS_BIND, "") = 0
exe              1774545 1774545 4026531840  umount("/var/run/docker/runtime-runc/moby/910fdb2d258cb4e4c504d1414c9c635074d64aa4554671123d4cc652c007dd2e/runc.GC2atV", MNT_DETACH) = 0
exe              1774598 1774598 4026531840  mount("/proc/self/exe", "/var/run/docker/runtime-runc/moby/41d0d870275682fd5c20ada27dec21ac2302cc2d2c3064e719094870934e30fb/runc.4yGzla", "", MS_BIND, "") = 0
exe              1774598 1774598 4026531840  mount("", "/var/run/docker/runtime-runc/moby/41d0d870275682fd5c20ada27dec21ac2302cc2d2c3064e719094870934e30fb/runc.4yGzla", "", MS_RDONLY|MS_REMOUNT|MS_BIND, "") = 0
exe              1774598 1774598 4026531840  umount("/var/run/docker/runtime-runc/moby/41d0d870275682fd5c20ada27dec21ac2302cc2d2c3064e719094870934e30fb/runc.4yGzla", MNT_DETACH) = 0
```

[3]
nsenter was import from https://github.com/opencontainers/runc/blob/master/init.go#L10
[...]

PR #3599, opened 2022-09-09
libct/nsenter: namespace the bindfd shuffle

corhere:
* closes https://github.com/opencontainers/runc/issues/2532
* relates to https://github.com/systemd/systemd/issues/8703

Processes can watch `/proc/self/mounts` or `/mountinfo`, and the kernel will notify them whenever the namespace's mount table is modified. The notified process still needs to read and parse the mountinfo to determine what changed once notified. Many such processes, including `udisksd` and systemd < v248, make no attempt to rate-limit their mountinfo notifications. This tends to not be a problem on many systems, where mount tables are small and mounting and unmounting is uncommon. Every runC exec which successfully uses the `try_bindfd` container-escape mitigation performs two `mount()`s and one `umount()` in the host's mount namespace, causing any mount-watching processes to wake up and parse the mountinfo file three times in a row. Consequently, using 'exec' health checks on containers has a larger-than-expected impact on system load when such mount-watching daemons are running. Furthermore, the size of the mount table in the host's mount namespace tends to be proportional to the number of OCI containers as a unique mount is required for the rootfs of each container. Therefore, on systems with mount-watching processes, the system load increases *quadratically* with the number of running containers which use health checks!

Prevent runC from incidentally modifying the host's mount namespace for container-escape mitigations by setting up the mitigation in a private mount namespace.

### DIFF
--- a/libcontainer/nsenter/cloned_binary.c
+++ b/libcontainer/nsenter/cloned_binary.c
@@ -49,6 +49,7 @@
 #include <fcntl.h>
 #include <errno.h>
 
+#include <sched.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/statfs.h>
@@ -56,7 +57,12 @@
 #include <sys/mman.h>
 #include <sys/mount.h>
 #include <sys/sendfile.h>
+#include <sys/socket.h>
 #include <sys/syscall.h>
+#include <sys/wait.h>
+
+#include "ipc.h"
+#include "log.h"
 
 /* Use our own wrapper for memfd_create. */
 #ifndef SYS_memfd_create
@@ -394,6 +400,123 @@ static int seal_execfd(int *fd, int fdtype)
 	return -1;
 }
 
+struct bindfd_child_args {
+	int sockfd;
+	const char *mount_target;
+};
+
+static int bindfd_in_subprocess(void *arg)
+{
+	/*
+	 * In the interests of efficiency (read: minimizing the syscall count)
+	 * and conciseness, no attempt is made to release resources which would
+	 * be cleaned up automatically on process exit, i.e. when this function
+	 * returns. This includes filesystem mounts, as this function is
+	 * executed in a dedicated mount namespace.
+	 */
+
+	/*
+	 * For obvious reasons this won't work in rootless mode because we
+	 * haven't created a userns -- but getting that to work will be a bit
+	 * complicated and it's only worth doing if someone actually needs it.
+	 */
+	if (mount("none", "/", NULL, MS_SLAVE | MS_REC, NULL) < 0)
+		return errno;
+	/*
+	 * The kernel resolves the magic symlink /proc/self/exe to the real file
+	 * _in the original mount namespace_. Cross-namespace bind mounts are
+	 * not allowed, so we must locate the file inside the current mount
+	 * namespace to be able to bind-mount it. (The mount(8) command resolves
+	 * symlinks, which is why it appears to work at first glance.)
+	 */
+	char linkbuf[PATH_MAX + 1] = { 0 };
+	ssize_t linkpathlen = readlink("/proc/self/exe", linkbuf, sizeof(linkbuf));
+	if (linkpathlen < 0)
+		return errno;
+	if (linkpathlen == sizeof(linkbuf)) {
+		/*
+		 * The link path is longer than PATH_MAX, and the contents of
+		 * linkbuf might have been truncated. A truncated path could
+		 * happen to be a valid path to a different file, which could
+		 * allow for local privilege escalation if we were to exec it.
+		 * The mount syscall doesn't accept paths longer than PATH_MAX,
+		 * anyway.
+		 */
+		return ENAMETOOLONG;
+	}
+
+	int srcfd = open(linkbuf, O_PATH | O_CLOEXEC);
+	if (srcfd < 0)
+		return errno;
+	/*
+	 * linkbuf holds the path to the binary which the parent process was
+	 * launched from. Someone could have moved a different file to that path
+	 * in the interim, in which case srcfd is not the file we want to
+	 * bind-mount. Guard against this situation by verifying srcfd is the
+	 * same file as /proc/self/exe.
+	 */
+	struct stat realexe = { 0 };
+	if (stat("/proc/self/exe", &realexe) < 0)
+		return errno;
+	struct stat resolved = { 0 };
+	if (fstat(srcfd, &resolved) < 0)
+		return errno;
+	if (resolved.st_dev != realexe.st_dev || resolved.st_ino != realexe.st_ino)
+		return ENOENT;
+	if (snprintf(linkbuf, sizeof(linkbuf), "/proc/self/fd/%d", srcfd) == sizeof(linkbuf))
+		return ENAMETOOLONG;
+
+	const struct bindfd_child_args *args = arg;
+	if (mount(linkbuf, args->mount_target, "", MS_BIND, "") < 0)
+		return errno;
+	if (mount("", args->mount_target, "", MS_REMOUNT | MS_BIND | MS_RDONLY, "") < 0)
+		return errno;
+
+	int fd = open(args->mount_target, O_PATH | O_CLOEXEC);
+	if (fd < 0)
+		return errno;
+
+	/*
+	 * Make sure the MNT_DETACH works, otherwise we could get remounted
+	 * read-write and that would be quite bad.
+	 */
+	if (umount2(args->mount_target, MNT_DETACH) < 0)
+		return errno;
+
+	if (send_fd(args->sockfd, fd) < 0)
+		return errno;
+	return 0;
+}
+
+static int spawn_bindfd_child(const struct bindfd_child_args *args) __attribute__((noinline));
+static int spawn_bindfd_child(const struct bindfd_child_args *args)
+{
+	/*
+	 * Carve out a chunk of our call stack for the child process to use as
+	 * we can be sure it is correctly mapped for use as stack. (Technically
+	 * only the libc clone() wrapper writes to this buffer. The child
+	 * process operates on a copy of the parent's virtual memory space and
+	 * so can safely overflow into the rest of the stack memory region
+	 * without consequence.)
+	 */
+	char stack[4 * 1024] __attribute__((aligned(16)));
+	int tid = clone(bindfd_in_subprocess,
+			/*
+			 * Assume stack grows down, as HP-PA, the only Linux
+			 * platform where stack grows up, is obsolete.
+			 */
+			stack + sizeof(stack),
+			/*
+			 * Suspend the parent process until the child has exited to
+			 * save an unnecessary context switch as we'd just be
+			 * waiting for the child process to exit anyway.
+			 */
+			CLONE_NEWNS | CLONE_VFORK, (void *)args);
+	if (tid < 0)
+		return -errno;
+	return tid;
+}
+
 static int try_bindfd(void)
 {
 	int fd, ret = -1;
@@ -415,32 +538,53 @@ static int try_bindfd(void)
 	close(fd);
 
 	/*
-	 * For obvious reasons this won't work in rootless mode because we haven't
-	 * created a userns+mntns -- but getting that to work will be a bit
-	 * complicated and it's only worth doing if someone actually needs it.
+	 * Daemons such as systemd and udisks2 watch /proc/self/mountinfo and
+	 * re-parse it on every change, which gets expensive when the mount table
+	 * is large and/or changes frequently. Perform the mount operations in a
+	 * new, private mount namespace so as not to wake up those processes
+	 * every time we nsexec into a container. We clone a child process into
+	 * a new mount namespace to do the dirty work so the side effects of
+	 * unsharing the mount namespace do not leak into the current process.
 	 */
-	ret = -EPERM;
-	if (mount("/proc/self/exe", template, "", MS_BIND, "") < 0)
-		goto out;
-	if (mount("", template, "", MS_REMOUNT | MS_BIND | MS_RDONLY, "") < 0)
-		goto out_umount;
-
-	/* Get read-only handle that we're sure can't be made read-write. */
-	ret = open(template, O_PATH | O_CLOEXEC);
-
-out_umount:
-	/*
-	 * Make sure the MNT_DETACH works, otherwise we could get remounted
-	 * read-write and that would be quite bad (the fd would be made read-write
-	 * too, invalidating the protection).
-	 */
-	if (umount2(template, MNT_DETACH) < 0) {
-		if (ret >= 0)
-			close(ret);
-		ret = -ENOTRECOVERABLE;
+	int sock[2];
+	if (socketpair(AF_LOCAL, SOCK_STREAM, 0, sock) < 0) {
+		ret = -errno;
+		goto cleanup_unlink;
 	}
 
-out:
+	struct bindfd_child_args args = {
+		.sockfd = sock[0],
+		.mount_target = template,
+	};
+	int cpid = spawn_bindfd_child(&args);
+	close(sock[0]);
+	if (cpid < 0) {
+		ret = cpid;
+		goto cleanup_socketpair;
+	}
+
+	int wstatus = 0;
+	if (waitpid(cpid, &wstatus, __WCLONE) < 0)
+		bail("error waiting for bindfd child process to exit");
+	if (WIFEXITED(wstatus)) {
+		if (WEXITSTATUS(wstatus)) {
+			ret = -WEXITSTATUS(wstatus);
+			goto cleanup_socketpair;
+		}
+	} else if (WIFSIGNALED(wstatus)) {
+		int sig = WTERMSIG(wstatus);
+		bail("bindfd child process terminated by signal %d (%s)", sig, strsignal(sig));
+	} else {
+		/* Should never happen... */
+		bail("unexpected waitpid() status for bindfd child process: 0x%x", wstatus);
+	}
+
+	ret = receive_fd(sock[1]);
+
+cleanup_socketpair:
+	close(sock[1]);
+
+cleanup_unlink:
 	/*
 	 * We don't care about unlink errors, the worst that happens is that
 	 * there's an empty file left around in STATEDIR.
--- a/libcontainer/nsenter/ipc.h
+++ b/libcontainer/nsenter/ipc.h
@@ -1,7 +1,12 @@
 #ifndef NSENTER_IPC_H
 #define NSENTER_IPC_H
 
-void receive_fd(int sockfd, int new_fd);
-void send_fd(int sockfd, int fd);
+int receive_fd(int sockfd);
+
+/*
+ * send_fd passes the open file descriptor fd to another process via the UNIX
+ * domain socket sockfd. The return value of the sendmsg(2) call is returned.
+ */
+int send_fd(int sockfd, int fd);
 
 #endif /* NSENTER_IPC_H */
--- a/libcontainer/nsenter/ipc.h
+++ b/libcontainer/nsenter/ipc.h
@@ -1,0 +1,7 @@
+#ifndef NSENTER_IPC_H
+#define NSENTER_IPC_H
+
+void receive_fd(int sockfd, int new_fd);
+void send_fd(int sockfd, int fd);
+
+#endif /* NSENTER_IPC_H */
--- a/libcontainer/nsenter/getenv.c
+++ b/libcontainer/nsenter/getenv.c
@@ -1,0 +1,27 @@
+#define _GNU_SOURCE
+#include <errno.h>
+#include <stdlib.h>
+#include "getenv.h"
+#include "log.h"
+
+int getenv_int(const char *name)
+{
+	char *val, *endptr;
+	int ret;
+
+	val = getenv(name);
+	/* Treat empty value as unset variable. */
+	if (val == NULL || *val == '\0')
+		return -ENOENT;
+
+	ret = strtol(val, &endptr, 10);
+	if (val == endptr || *endptr != '\0')
+		bail("unable to parse %s=%s", name, val);
+	/*
+	 * Sanity check: this must be a non-negative number.
+	 */
+	if (ret < 0)
+		bail("bad value for %s=%s (%d)", name, val, ret);
+
+	return ret;
+}
--- a/libcontainer/nsenter/getenv.h
+++ b/libcontainer/nsenter/getenv.h
@@ -1,0 +1,13 @@
+#ifndef NSENTER_GETENV_H
+#define NSENTER_GETENV_H
+
+/*
+ * Returns an environment variable value as a non-negative integer, or -ENOENT
+ * if the variable was not found or has an empty value.
+ *
+ * If the value can not be converted to an integer, or the result is out of
+ * range, the function bails out.
+ */
+int getenv_int(const char *name);
+
+#endif /* NSENTER_GETENV_H */
--- a/libcontainer/nsenter/ipc.c
+++ b/libcontainer/nsenter/ipc.c
@@ -30,7 +30,7 @@ int receive_fd(int sockfd)
 
 	memset(msg.msg_control, 0, msg.msg_controllen);
 
-	bytes_read = recvmsg(sockfd, &msg, 0);
+	bytes_read = recvmsg(sockfd, &msg, MSG_CMSG_CLOEXEC);
 	if (bytes_read != 1)
 		bail("failed to receive fd from unix socket %d", sockfd);
 	if (msg.msg_flags & MSG_CTRUNC)
--- a/libcontainer/nsenter/ipc.c
+++ b/libcontainer/nsenter/ipc.c
@@ -1,13 +1,12 @@
 #define _GNU_SOURCE
-#include <fcntl.h>
+#include <alloca.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/socket.h>
-#include <unistd.h>
 #include "ipc.h"
 #include "log.h"
 
-void receive_fd(int sockfd, int new_fd)
+int receive_fd(int sockfd)
 {
 	int bytes_read;
 	struct msghdr msg = { };
@@ -16,7 +15,6 @@ void receive_fd(int sockfd, int new_fd)
 	char null_byte = '\0';
 	int ret;
 	int fd_count;
-	int *fd_payload;
 
 	iov.iov_base = &null_byte;
 	iov.iov_len = 1;
@@ -52,21 +50,13 @@ void receive_fd(int sockfd, int new_fd)
 	if (fd_count != 1)
 		bail("received control message from unix socket %d with too many fds: %d", sockfd, fd_count);
 
-	fd_payload = (int *)CMSG_DATA(cmsg);
-	ret = dup3(*fd_payload, new_fd, O_CLOEXEC);
-	if (ret < 0)
-		bail("cannot dup3 fd %d to %d", *fd_payload, new_fd);
-
+	ret = *(int *)CMSG_DATA(cmsg);
 	free(msg.msg_control);
-
-	ret = close(*fd_payload);
-	if (ret < 0)
-		bail("cannot close fd %d", *fd_payload);
+	return ret;
 }
 
-void send_fd(int sockfd, int fd)
+int send_fd(int sockfd, int fd)
 {
-	int bytes_written;
 	struct msghdr msg = { };
 	struct cmsghdr *cmsg;
 	struct iovec iov[1] = { };
@@ -81,11 +71,7 @@ void send_fd(int sockfd, int fd)
 	/* We send only one fd as specified by cmsg->cmsg_len below, even
 	 * though msg.msg_controllen might have more space due to alignment. */
 	msg.msg_controllen = CMSG_SPACE(sizeof(int));
-	msg.msg_control = malloc(msg.msg_controllen);
-	if (msg.msg_control == NULL) {
-		bail("Can't allocate memory to send fd.");
-	}
-
+	msg.msg_control = alloca(msg.msg_controllen);
 	memset(msg.msg_control, 0, msg.msg_controllen);
 
 	cmsg = CMSG_FIRSTHDR(&msg);
@@ -94,10 +80,5 @@ void send_fd(int sockfd, int fd)
 	cmsg->cmsg_len = CMSG_LEN(sizeof(int));
 	memcpy(CMSG_DATA(cmsg), &fd, sizeof(int));
 
-	bytes_written = sendmsg(sockfd, &msg, 0);
-
-	free(msg.msg_control);
-
-	if (bytes_written != 1)
-		bail("failed to send fd %d via unix socket %d", fd, sockfd);
+	return sendmsg(sockfd, &msg, 0);
 }
--- a/libcontainer/nsenter/ipc.c
+++ b/libcontainer/nsenter/ipc.c
@@ -1,0 +1,103 @@
+#define _GNU_SOURCE
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <unistd.h>
+#include "ipc.h"
+#include "log.h"
+
+void receive_fd(int sockfd, int new_fd)
+{
+	int bytes_read;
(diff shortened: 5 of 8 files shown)


libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,29 +423,89 @@ static int try_bindfd(void)
...
+			 * linkbuf might have been truncated. A truncated path could
+			 * happen to be a valid path to a different file, which could
+			 * allow for local privilege escalation if we were to exec it.
+			 * The mount syscall doesn't accept paths longer than PATH_MAX,
+			 * anyway.
+			 */
+			ret = -ENAMETOOLONG;
+			goto out_child;
+		}
+		linkbuf[linkpathlen] = '\0';
```
kolyshkin, line 466: From the code it looks like this is redundant, since linkbuf is initialized with zeroes, and the `if` above guarantees that at least one zero remains in it. OTOH it doesn't hurt either.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,29 +423,89 @@ static int try_bindfd(void)
...
+		if ((ret = mount(linkbuf, template, "", MS_BIND, "")) < 0)
+			goto out_child;
+		if ((ret = mount("", template, "", MS_REMOUNT | MS_BIND | MS_RDONLY, "")) < 0)
+			goto out_child;
```
kolyshkin, line 470: This seems to leave a stray mount.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,29 +423,89 @@ static int try_bindfd(void)
...
+
+		/* Get read-only handle that we're sure can't be made read-write. */
+		fd = open(template, O_PATH | O_CLOEXEC);
+		if (fd < 0) {
+			ret = fd;
+			goto out_child;
```
kolyshkin, line 476: Same here.

libcontainer/nsenter/cloned_binary.c
kolyshkin: If that's by design (since the mntns is soon be gone anyway), maybe add a comment saying so.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,29 +423,89 @@ static int try_bindfd(void)
...
+	} else {
+		int cpid = clone_longjmp(&env, 1, CLONE_NEWNS | SIGCHLD);
+		if (cpid < 0) {
+			ret = -errno;
+			goto out;
+		}
+		int wstatus = 0;
+		if ((ret = waitpid(cpid, &wstatus, 0)) < 0)
+			goto out;
+		if ((ret = -WEXITSTATUS(wstatus)))
```
kolyshkin, line 503: Shouldn't WEXITSTATUS be called only if WIFEXITED(STATUS) is true?

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,29 +423,89 @@ static int try_bindfd(void)
...
+			ret = -ENOTRECOVERABLE;
+			goto out_child;
+		}
+		send_fd(pipe[1], fd);
+out_child:
+		if (fd >= 0)
+			close(fd);
+		close(pipe[1]);
+		exit(-ret);
+	} else {
```
kolyshkin, line 494: 1. nit: annotate with `// Parent.` or something

2. I'd add `close pipe[1]` here.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,29 +423,89 @@ static int try_bindfd(void)
...
+		}
+		int wstatus = 0;
+		if ((ret = waitpid(cpid, &wstatus, 0)) < 0)
+			goto out;
+		if ((ret = -WEXITSTATUS(wstatus)))
+			goto out;
+		fd = receive_fd(pipe[0]);
+		if ((ret = fcntl(ret, F_SETFD, O_PATH | O_CLOEXEC)) < 0)
+			goto out;
+		ret = fd;
```
kolyshkin, line 508: It looks like a mixup of `ret` and `fd` here. For one thing, passing `ret` to `fcntl` makes no sense; neither does `ret = fd` (or I'm missing something).

kolyshkin:
Looks good overall, and it's amazing that it works! Left some nits.

libcontainer/nsenter/clone.c
```diff
@@ -0,0 +1,39 @@
...
+
+/* Assume the stack grows down, so arguments should be above it. */
+struct clone_t {
+	/*
+	 * Reserve some space for clone() to locate arguments
+	 * and retcode in this place
+	 */
+	char stack[4096] __attribute__((aligned(16)));
+	char stack_ptr[0];
+
```
AkihiroSuda, line 14: Do we need some "canary" field here in case the stack overflows?

libcontainer/nsenter/clone.c
```diff
@@ -0,0 +1,39 @@
+#define _GNU_SOURCE
+#include <sched.h>
+#include <errno.h>
+#include "clone.h"
+
+/* Assume the stack grows down, so arguments should be above it. */
```
AkihiroSuda, line 6: This assumption seems untrue for HP PA (PA-RISC) https://man7.org/linux/man-pages/man2/clone.2.html

I don't think we will ever support PA-RISC, but probably it should be mentioned here

libcontainer/nsenter/clone.c
corhere: (This is existing code which I merely refactored.)
I don't think so, as I understand it. Only enough stack to execute the `longjmp()` is required as that call will restore the stack pointer, switching which stack is used for the remainder of the child process's execution.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -394,6 +400,98 @@ static int seal_execfd(int *fd, int fdtype)
...
+		      * Assume stack grows down, as HP-PA, the only Linux
+		      * platform where stack grows up, is obsolete.
+		      */
+		     stack + sizeof(stack),
+		     /*
+		      * Suspend the parent process until the child has exited to
+		      * save an unnecessary context switch as we'd just be
+		      * waiting for the child process to exit anyway.
+		      */
+		     CLONE_NEWNS | CLONE_VFORK, (void *)args);
```
corhere, line 492: It is possible to also set the `CLONE_VM` flag as I did all the prerequisite work: making `send_fd()` vfork-safe by hoisting the `bail()` calls out, so libc `exit()` wouldn't get called on any code path. I did a little bit of benchmarking with and without `CLONE_VM` but found no significant differences in `runc run` time so I went with the safer choice of copy-on-write.

AkihiroSuda (dismissed):

libcontainer/nsenter/cloned_binary.c
kolyshkin: Nice; that is a clever use of CLONE_VFORK :)

libcontainer/nsenter/cloned_binary.c
```diff
@@ -394,6 +400,98 @@ static int seal_execfd(int *fd, int fdtype)
...
+static int spawn_bindfd_child(const struct bindfd_child_args *args)
+{
+	/*
+	 * Carve out a chunk of our call stack for the child process to use as
+	 * we can be sure it is correctly mapped for use as stack. (Technically
+	 * only the libc clone() wrapper writes to it; the child process
+	 * operates on a copy of the parent's virtual memory space unless we use
+	 * the CLONE_VM flag.)
+	 */
+	char stack[128 * 1024] __attribute__((aligned(16)));
```
kolyshkin, line 480: It doesn't make any practical difference, but the other user of clone sets the stack to be just 1K, and here it's 128K. As this is only used by clone() libc wrapper, perhaps just set this to 1K here as well?

libcontainer/nsenter/cloned_binary.c
kolyshkin: Nit: casting to `(void *)` is redundant.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,32 +513,51 @@ static int try_bindfd(void)
 	close(fd);
 
 	/*
-	 * For obvious reasons this won't work in rootless mode because we haven't
-	 * created a userns+mntns -- but getting that to work will be a bit
-	 * complicated and it's only worth doing if someone actually needs it.
+	 * Daemons such as systemd and udisks2 watch /proc/self/mountinfo and
+	 * re-parse it on every change, which gets expensive when the mount table
+	 * is large and/or changes frequently. Set up the bind-mounts in a new,
```
kolyshkin, line 518: > the bind-mounts

Nit: we only have one single bind mount.

libcontainer/nsenter/ipc.h
```diff
@@ -2,6 +2,11 @@
 #define NSENTER_IPC_H
 
 int receive_fd(int sockfd);
-void send_fd(int sockfd, int fd);
```
kolyshkin: Ideally this should be a separate patch. Same for receive_fd which is converted in the same way in the commit claiming to only move the code around.

So maybe
 - move the code around in the first commit
 - do the changes like this in the second commit
 - introduce the new feature in the last commit.

libcontainer/nsenter/cloned_binary.c
corhere: The child process will use this stack segment for its entire lifetime. It doesn't switch stacks with `longjmp()` as the other user of clone does.

libcontainer/nsenter/cloned_binary.c
corhere: Right; that's poorly worded. There is only one bind mount, but three mutations to the mount table. The kernel notifies those daemons for each `mount` and `umount2` syscall occuring in the mount namespace.

libcontainer/nsenter/cloned_binary.c
corhere: The `const` qualifier needs to be casted away [to suppress a compiler warning.](https://godbolt.org/#g:!((g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:1,endLineNumber:6,positionColumn:1,positionLineNumber:6,selectionStartColumn:1,selectionStartLineNumber:6,startColumn:1,startLineNumber:6),source:'void+foo(void+*x)%3B%0A%0Avoid+bar(const+int+*x)+%7B%0A++++foo(x)%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cg122,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+12.2+(C,+Editor+%231,+Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+12.2+(Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

kolyshkin: Ah, OK, I missed the `const`.

libcontainer/nsenter/cloned_binary.c
kolyshkin: Perhaps something like `s/Set up the bind-mounts in/Perform the mounts in/`

kolyshkin (approved):
LGTM, and thanks!

kolyshkin:
@AkihiroSuda @cyphar PTAL

libcontainer/nsenter/cloned_binary.c
```diff
@@ -394,6 +400,99 @@ static int seal_execfd(int *fd, int fdtype)
...
+	 */
+	if (mount("none", "/", NULL, MS_SLAVE | MS_REC, NULL) < 0)
+		return errno;
+	/*
+	 * The kernel refuses to bind-mount from the magic symlink when the
+	 * process has been cloned (or unshared) into a new mount namespace,
+	 * at least on Linux 4.4.
+	 */
+	char linkbuf[PATH_MAX + 1] = { 0 };
+	ssize_t linkpathlen = readlink("/proc/self/exe", linkbuf, sizeof(linkbuf));
```
cyphar, line 431: If we cannot use `/proc/self/exe` (or `/proc/$runc_pid/exe`) then we can't really do this. It's not safe to be operating on random paths that you get from resolving `/proc/self/exe` (and more importantly, this won't be doable once we start work on porting runc to libpathrs). (Obviously this is running "on the host" but I just don't feel comfortable with resolving magiclinks because they aren't symlinks at all and any similarity to symlinks is a very leaky illusion.)

However, I'm not sure it's actually true that you cannot use the magiclink as a mount source -- based on a quick test with `unshare` (Linux 5.19.10) it works without issue and I'm not sure what would've caused it to fail in older kernel versions (`nd_jump_link` really doesn't care about what you're doing with the path and the mount infrastructure doesn't care about the source IIRC). You probably cannot use it as a mount *target* but that shouldn't be an issue. What problem did you run into when using it as a mount source?

libcontainer/nsenter/cloned_binary.c
```diff
@@ -394,6 +400,99 @@ static int seal_execfd(int *fd, int fdtype)
...
+		return ENAMETOOLONG;
+	}
+
+	const struct bindfd_child_args *args = arg;
+	if (mount(linkbuf, args->mount_target, "", MS_BIND, "") < 0)
+		return errno;
+	if (mount("", args->mount_target, "", MS_REMOUNT | MS_BIND | MS_RDONLY, "") < 0)
+		return errno;
+
+	/* Get read-only handle that we're sure can't be made read-write. */
```
cyphar, line 452: Reading this comment (I wrote) again, this probably should be dropped because it's not true (it's trivial to re-open `O_PATH` descriptors at the moment -- though I am working on fixing that in the kernel).

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,32 +514,51 @@ static int try_bindfd(void)
...
-	 * Make sure the MNT_DETACH works, otherwise we could get remounted
-	 * read-write and that would be quite bad (the fd would be made read-write
-	 * too, invalidating the protection).
-	 */
-	if (umount2(template, MNT_DETACH) < 0) {
-		if (ret >= 0)
-			close(ret);
-		ret = -ENOTRECOVERABLE;
+	ret = receive_fd(sock[1]);
+	if (fcntl(ret, F_SETFD, O_PATH | O_CLOEXEC) < 0) {
```
cyphar, line 553: Should be `FD_CLOEXEC` (which `!= O_CLOEXEC`) and `O_PATH` here doesn't make sense. Also I'm not sure it makes sense to do this anyway -- we know we set `O_CLOEXEC` when we sent it and in general setting `O_CLOEXEC` after the fact is not safe.

I'm a bit surprised you don't get an error from this...

libcontainer/nsenter/cloned_binary.c
```diff
@@ -415,32 +514,51 @@ static int try_bindfd(void)
...
+		ret = -1;
```
cyphar, line 554: I prefer `-errno` returns in general.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -394,6 +400,99 @@ static int seal_execfd(int *fd, int fdtype)
...
+
+static int bindfd_in_subprocess(void *arg)
+{
+	/*
+	 * In the interests of efficiency (read: minimizing the syscall count)
+	 * and conciseness, no attempt is made to release resources which would
+	 * be cleaned up automatically on process exit, i.e. when this function
+	 * returns. This includes filesystem mounts, as this function is
+	 * executed in a dedicated mount namespace.
+	 */
```
cyphar, line 416: It would be nice to mention that this function is run in a vfork, so folks need to keep in mind they can't use stuff that is forbidden in vfork.

cyphar (changes requested):

libcontainer/nsenter/cloned_binary.c
corhere: On both Linux 4.4.0-1122-aws and Linux 5.15.0-33-generic (Ubuntu) any attempts to bind-mount `/proc/self/exe` in an unshared mount namespace yield `EINVAL`.
```console
ubuntu@ubuntu2204:~$ sudo strace -f -e trace=mount ./runc run --bundle bundle foobar
strace: Process 62293 attached
strace: Process 62292 attached
strace: Process 62294 attached
strace: Process 62295 attached
strace: Process 62296 attached
strace: Process 62298 attached
[pid 62298] mount("none", "/", NULL, MS_REC|MS_SLAVE, NULL) = 0
[pid 62298] mount("/proc/self/exe", "/run/runc/foobar/runc.G52pod", 0x563c345027e2, MS_BIND, 0x563c345027e2) = -1 EINVAL (Invalid argument)
[pid 62298] +++ exited with 22 +++
[...]
```

`mount(8)` canonicalizes the source path with `realpath(3)` which is why it appeared to work in your test.
```console
ubuntu@ubuntu2204:~$ sudo unshare --mount
root@ubuntu2204:/home/ubuntu# strace -e trace=readlink,mount mount --bind /proc/self/exe selfexe
readlink("/proc", 0x7fff7a3b6420, 1023) = -1 EINVAL (Invalid argument)
readlink("/proc/self", "62403", 1023)   = 5
readlink("/proc/62403", 0x7fff7a3b6420, 1023) = -1 EINVAL (Invalid argument)
readlink("/proc/62403/exe", "/usr/bin/mount", 1023) = 14
readlink("/usr", 0x7fff7a3b6420, 1023)  = -1 EINVAL (Invalid argument)
readlink("/usr/bin", 0x7fff7a3b6420, 1023) = -1 EINVAL (Invalid argument)
readlink("/usr/bin/mount", 0x7fff7a3b6420, 1023) = -1 EINVAL (Invalid argument)
readlink("/home/ubuntu/selfexe", 0x7fff7a3b6420, 1023) = -1 EINVAL (Invalid argument)
mount("/usr/bin/mount", "/home/ubuntu/selfexe", 0x563f8a351840, MS_BIND, NULL) = 0
+++ exited with 0 +++
```

<details><summary>Minimal reproducer:</summary>

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <sched.h>
#include <sys/mount.h>

int main(int argc, char** argv) {
	if (argc > 1) {
		puts("Unsharing mount namespace...");
		if (unshare(CLONE_NEWNS) < 0) {
			perror("unshare");
			exit(EXIT_FAILURE);
		}
	}
	if (mount("/proc/self/exe", "selfexe", "", MS_BIND, "") < 0) {
		perror("mount");
		exit(EXIT_FAILURE);
	}
	int ret = system("/bin/sh");
	if (ret < 0) {
		perror("system");
		exit(EXIT_FAILURE);
	}
	return ret;
}
```

```console
$ gcc main.c && touch selfexe && sudo ./a.out unshare
Unsharing mount namespace...
mount: Invalid argument
$ sudo ./a.out
# exit
$ sudo umount selfexe
```

</details>

Indirecting the mount through `/proc/self/fd/<n>` also yields `EINVAL`, regardless of whether the file descriptor to `/proc/self/exe` was opened before or after the mount namespace was unshared, with or without `O_PATH`. My best guess of what's happening is that `/proc/self/exe` resolves to a `struct path` value which, being magic, has its `mnt->mnt_ns` set to the parent mount namespace, [which trips the `check_mnt()` validation here.](https://elixir.bootlin.com/linux/v6.0/source/fs/namespace.c#L2399) If my interpretation is correct, there is no way to successfully bind-mount a process' executable across namespaces without going through path resolution machinery in the process' current mount namespace.

Would you feel more comfortable if there was more assurance that the path resolved to the same file as `/proc/self/exe` @cyphar? Here's what I'm thinking:
1. `open(O_PATH)` both `/proc/self/exe` and `realpath("/proc/self/exe")`
2. Check that the file descriptors refer to the same file by `fstat`'ing both and testing whether they have the same `st_dev` and `st_ino` values
3. Bind-mount the `/proc/self/fd/<n>` for the realpath fd

libcontainer/nsenter/cloned_binary.c
kolyshkin: @corhere ^^^

libcontainer/nsenter/cloned_binary.c
kolyshkin: @cyphar ^^^

libcontainer/nsenter/cloned_binary.c
corhere: > Also I'm not sure it makes sense to do this anyway -- we know we set `O_CLOEXEC` when we sent it

I don't think the `O_CLOEXEC` flag of the sending file descriptor is relevant.

> [Semantically, [the SCM_RIGHTS] operation is equivalent to duplicating (`dup(2)`) a file descriptor into the file descriptor table of another process.](https://man7.org/linux/man-pages/man7/unix.7.html#:~:text=Semantically%2C%20this%20operation%20is%20equivalent%20to%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20duplicating%20(dup(2))%20a%20file%20descriptor%20into%20the%20file%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20descriptor%20table%20of%20another%20process.)

> The `dup()` system call allocates a new file descriptor that refers to the same open file description as the descriptor `oldfd`. [...] [The two file descriptors do not share file descriptor flags (the close-on-exec flag). The close-on-exec flag (`FD_CLOEXEC`; see `fcntl(2)`) for the duplicate descriptor is off.](https://man7.org/linux/man-pages/man2/dup.2.html#:~:text=The%20two%20file%20descriptors%20do%20not%20share%20file%20descriptor%20flags%20(the%0A%20%20%20%20%20%20%20close%2Don%2Dexec%20flag).%20%20The%20close%2Don%2Dexec%20flag%20(FD_CLOEXEC%3B%20see%0A%20%20%20%20%20%20%20fcntl(2))%20for%20the%20duplicate%20descriptor%20is%20off.)

That being said, `recvmsg(2)` takes a flag `MSG_CMSG_CLOEXEC` which atomically sets `FD_CLOEXEC` on all received file descriptors. I will set that flag and drop the `fcntl` call.

libcontainer/nsenter/cloned_binary.c
cyphar: Fair enough.

libcontainer/nsenter/cloned_binary.c
cyphar: Yeah, looking at [`scm_detach_fds`](https://elixir.bootlin.com/linux/v6.0/source/net/core/scm.c#L297) the receive is done in a context where it can't even know if the original file descriptor was CLOEXEC. But yes, we should be setting `MSG_CMG_CLOEXEC` in `msg_flags`.

kolyshkin (approved):
LGTM

@cyphar @AkihiroSuda PTAL

thaJeztah:
@cyphar @AkihiroSuda PTAL 🤗

thaJeztah:
@cyphar could you have a look at https://github.com/opencontainers/runc/pull/3599#discussion_r997394477 ?

(I chatted with @corhere and he'll rebase after that)

AkihiroSuda:
Needs rebase

corhere:
Rebased. @cyphar PTAL

xujihui1985:
a different approach is instead of doing mount in a separate mountns, just put kubelet and containerd into a separate  mountns.
Pros
- easy to implement
Cons
- it must be setup in a fresh machine

changweige:
> @cyphar ptal would be good to see this moving 🙏

+1
@cyphar I personally feel like this patch is a strong enhancement and benefits users with large-scale K8s deployment. So PTAL. Thanks

AkihiroSuda:
Needs rebase

adamqqqplay:
This issue is also affecting us, if it can be fixed we will be very grateful.

corhere:
This PR has languished for far too long [without any feedback from the maintainer who requested changes](https://github.com/opencontainers/runc/blob/ba58ee9c3b9550c3e32b94802b0fb29761955290/MAINTAINERS_GUIDE.md?plain=1#L26), even after I made changes which directly address the reason that changes were requested. @crosbymichael could you please take a look?

AkihiroSuda (approved):

cyphar (approved):
I think `try_bindfd` should die and be replaced with just the memfd code ([which eventually will no longer be needed when I finish working on the magic-link hardening work to block these re-opening attacks](https://lwn.net/Articles/934460/) -- though we will need to keep it around for a long while...).

However, looking at this PR again, it probably makes sense to carry this (even if we end up dropping bindfd in runc 1.3.0). The extra CPU cost of creating a new mount namespace for such a trivial task kind of sucks, but them's the breaks (you could even argue this makes it clearer that we should drop bindfd sooner).

Sadly the proposal I had a while ago for removing the cost of memfds doesn't work (I figured out this out a while ago but it's probably a good idea to write that down somewhere) -- you cannot bind-mount memfds at all because they are all in the `MNT_NS_INTERNAL` namespace.

libcontainer/nsenter/cloned_binary.c
cyphar: Yes, you were both quite right -- I had forgotten about the annoying way `mount(8)` handles paths. The future libpathrs issue is still a problem but I think we need to drop this bindfd code before then anyway, so we can cross that bridge if we ever get to it.

libcontainer/nsenter/cloned_binary.c
```diff
@@ -394,6 +400,123 @@ static int seal_execfd(int *fd, int fdtype)
...
+			 * Assume stack grows down, as HP-PA, the only Linux
+			 * platform where stack grows up, is obsolete.
+			 */
+			stack + sizeof(stack),
+			/*
+			 * Suspend the parent process until the child has exited to
+			 * save an unnecessary context switch as we'd just be
+			 * waiting for the child process to exit anyway.
+			 */
+			CLONE_NEWNS | CLONE_VFORK, (void *)args);
```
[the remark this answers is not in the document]
cyphar: Unfortunately the only alternative is to remove bindfd (which I would like to do!) but will probably have to wait until runc 1.3 since it could cause regressions.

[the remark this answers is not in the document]
xujihui1985: @cyphar what about instead of create/destroy the mountns everytime exec executed, create one long running mountpoint once when containerd started and paas the handle of mountns to runc when it was forked, be benefit of this approach is 
1. no need to create/destroy mountns everytime
2. the handle of mountns is optional, it's better for backward compatibility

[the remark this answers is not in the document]
113xiaoji: I encountered a similar issue, and this PR did not solve my problem; on the contrary, from a black-box perspective, it has exacerbated my problem. My scenario is to deploy 100 pods on a single node, each with a container.

Before the modification, due to the high CPU usage of systemd, a large number of pods failed to deploy. However, there is a chance that the pods can successfully restart. After the merger of this PR, the CPU usage of systemd indeed decreased. But the hang up phenomenon still exists.

A large number of pods still fail to deploy, and retries also fail. Not using try_bindfd but using memfd works very well, all pods are deployed successfully for the first time. Even deploying 170 pods at the same time will still succeed.

I think try_bindfd should die and be replaced with just the memfd code.

113xiaoji:
> try_bindfd



> I think `try_bindfd` should die and be replaced with just the memfd code ([which eventually will no longer be needed when I finish working on the magic-link hardening work to block these re-opening attacks](https://lwn.net/Articles/934460/) -- though we will need to keep it around for a long while...).
> 
> However, looking at this PR again, it probably makes sense to carry this (even if we end up dropping bindfd in runc 1.3.0). The extra CPU cost of creating a new mount namespace for such a trivial task kind of sucks, but them's the breaks (you could even argue this makes it clearer that we should drop bindfd sooner).
> 
> Sadly the proposal I had a while ago for removing the cost of memfds doesn't work (I figured out this out a while ago but it's probably a good idea to write that down somewhere) -- you cannot bind-mount memfds at all because they are all in the `MNT_NS_INTERNAL` namespace.

cannot agree more

libcontainer/nsenter/cloned_binary.c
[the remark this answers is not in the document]
cyphar: @xujihui1985 Given the experience we had with `--console-socket` and that I plan to remove `bindfd` entirely, exposing this messiness to our callers and requiring everyone to add support for this (and maintain support for it for a long time) is really unappealing -- while support for this would be optional, all of our callers would probably implement it and then would not be able to easily stop using it. There are alternatives of course (we could bind-mount the namespace on the host and then re-use it for all containers, but then we have to store where it is, and I would prefer to avoid global storage requirements for runc outside of the containers-specific storage for `/var/run/runc`).

@113xiaoji I would like to drop the bindfd code, but I'm confused why you're seeing even worse performance. Is the contention on `mount_lock` really that bad? Now that we are creating a new namespace it's possible we're creating some more contention on `namespace_sem`. If that's the issue then there really isn't an alternative (even `open_tree` cannot help because it creates a new namespace internally which means you'd hit all of the same lock contention issues) aside from having a long-lived namespace -- but if the core problem is `mount_lock` contention then there's really no other alternative because we need to create a read-only bind-mount for this protection to work.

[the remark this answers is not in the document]
113xiaoji: @cyphar 
I'm not sure how to capture the duration of each step, specifically where the performance bottleneck is. Could you teach me some methods? How can I be notified of any new changes in real time so that I can test my use case scenarios?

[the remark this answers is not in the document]
xujihui1985: @cyphar ya, you have your point, thanks for clarify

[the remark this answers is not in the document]
113xiaoji: > @cyphar ya, you have your point, thanks for clarify, maybe the best option at this moment is to avoid exec probe for massive pods
For services migrating from virtual machines to containers, changing the probe from exec mode to non-exec mode would be a substantial amount of work.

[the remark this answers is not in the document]
xujihui1985: @113xiaoji yes, we have the same problem here, I have tried another option to setup a new mountns and "move the containerd and kubelet into the mountns", you can refer to this article https://docs.openshift.com/container-platform/4.12/scalability_and_performance/optimization/optimizing-cpu-usage.html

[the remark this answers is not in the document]
113xiaoji: If we are in a rush and can't wait for the 1.3.0 version, what do you think about us creating our own version, where we remove the relevant logic of try_bindfd to meet our business needs? Besides memory overhead, do you foresee any other serious issues?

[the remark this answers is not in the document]
cyphar: It contributes to the container's memory accounting limit, but only while the runc binary is actually running -- the runc process is generally very short-lived (it only exists when it needs to do container operations, once a container process is spawned it disappears -- at least, this is how it works when you use it in the context of Docker/cri-o/containerd/Kubernetes). Each invocation of runc causes a ~10MB (the size of the runc binary) memory allocation in whatever container's cgroup is associated with that runc invocation.

The patch to remove `try_bindfd` is very trivial, you can see the patch in #3509 (I discovered some kernel bugs in `memfd_create` when I was working on this, I was going to send it this morning...)

The thing is, creating a container involves doing many mounts already, so I'm still a little surprised that this one mount is causing so many performance issues. I suspect that the the host bind-mount would be okay if it weren't for systemd ingesting every mount event on the system, but adding the short-lived namespace might be triggering a different contention issue.

PR #3609, opened 2022-09-21
WIP DNM support Path in OCI spec's LinuxDeviceCgroup

kolyshkin:
FIXME doc this


libcontainer/devices/device.go
```diff
@@ -13,9 +13,6 @@ const (
 type Device struct {
 	Rule
 
-	// Path to the device.
-	Path string `json:"path"`
-
```
cyphar, line 18: I suspect this will cause issues with upgrades.

kolyshkin: Well, this can't affect `runc exec`, `runc stop` etc, and for `runc update` we currently skip devices entirely. Having said that, this is a valid concern as we might change `runc update` to manage devices in the future.

What I wanted to achieve with this is to not have two fields named `Path` as this makes the code less readable. Seems that's inevitable...

cyphar: I meant more generally that I'm worried we'll have a repeat of the `initProcessTime` issue (which ended up with Docker having [some super dodgy patching code to work around the issue](https://github.com/crosbymichael/upgrade/tree/master/v17_06_1)) -- `state.json` is deserialised when loading container state, and if you upgrade runc such that the two JSON specs are no longer compatible you'll end up with errors after upgrading. Then again, I think that extra fields don't produce errors in `json.Decode`?

yangfeiyu20102011:
Hi, @kolyshkin , I want to know when this PR will be completed and merged to solve the problem.
Thanks!

kolyshkin:
OK, we have a better solution now, see https://github.com/opencontainers/runc/pull/3842

PR #3613, opened 2022-09-27
mount: add enhanced mount functionality to support run container in userns with host network

shidao1:


libcontainer/rootfs_linux.go
```diff
@@ -400,12 +400,25 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 			return err
 		}
 		// Selinux kernels do not support labeling of /proc or /sys
+		if m.IsMove() && *c.fd >= 0 {
```
kolyshkin, line 403: Nit: unlike in C, you don't have to dereference a pointer to a struct when accessing its members.

IOW `s/*c.fd/c.fd/`

libcontainer/container_linux.go
```diff
@@ -536,15 +536,46 @@ func (c *Container) newInitProcess(p *Process, cmd *exec.Cmd, messageSockPair, l
...
 	userNsMountFds := [3]int{-1, -1, -1}
 
+	// Enable open_tree()/move_mount() for special filesystems to support cross user namespace mounting.
+	if _, ok := nsList[configs.NEWUSER]; ok {
+		for idx, m := range c.config.Mounts {
+			if m.Device == "proc" && m.Source == "proc" && m.Destination == "/proc" &&
+				m.Flags == unix.MS_NODEV|unix.MS_NOEXEC|unix.MS_NOSUID {
+				// procfs depends on Pid namespace
+				if _, exist := nsList[configs.NEWPID]; exist {
+				}
```
kolyshkin, line 555: Is some code missing here?

kolyshkin:
Also, I'm afraid you'll have to redo this once #3599 is merged, which refactors some C code in nsenter.

jiangliu:
> Also, I'm afraid you'll have to redo this once #3599 is merged, which refactors some C code in nsenter.

Thanks for reminder, we also noticed PR #3599, so open this PR:)
We will wait for #3599 to settle down first.

Zheaoli:
I think this PR is not active for long time, may I take handle the rest work for making this PR ready to merge? cc @AkihiroSuda

cyphar:
@Zheaoli You'll need to base it on top of #3985, which reworks all of the mountfd logic. I'm not sure how easy it'll be to use the new Go-based setup to implement this though. I suspect you can do it by creating a locked goroutine that joins the container's non-userns namespaces, but the slight issue is that we cannot create a procfs mount that uses the containers pidns because procfs uses the _active_ pidns, not the `for_children` one (in fact, I'm not sure this PR handles procfs correctly).

Also, you don't want to use `open_tree(2)` like this -- a much better way is to use `fsopen` and `fsconfig` to configure the mount without touching the filesystem, and thus having an anonymous mountfd that you can then provide to the container. (To be fair, I'm not sure if the permissions work out okay with user namespaces in that case.)

PR #3614, opened 2022-09-27
Fix centos 7 ci

kolyshkin:
Turns out, CentOS 7 image used in Cirrus CI had EPEL repo configured, and now it's not. This, together with a "feature" of older yum NOT returning an error when some of the packages requested are not found, resulted in failures in tests that use `jq` and `fuse-sshfs`.

The fix is to add/enable EPEL repo.

1.1 backport: https://github.com/opencontainers/runc/pull/3618

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -72,6 +72,7 @@ task:
     CIRRUS_WORKING_DIR: /home/runc
     GO_VERSION: "1.18"
     BATS_VERSION: "v1.3.0"
+    RPMS: gcc git iptables jq glibc-static libseccomp-devel make criu fuse-sshfs
     # yamllint disable rule:key-duplicates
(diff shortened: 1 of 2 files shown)


kolyshkin:
OK, the issue seems that in CentOS 7 image used by Cirrus CI, EPEL is no longer enabled, despite the docs at https://cloud.google.com/compute/docs/images/os-details (under "Image configuration") saying

> For CentOS 7, EPEL is enabled.

I am pretty sure it used to be enabled, but apparently Google Cloud has changed something.

The other part of the story is `yum` from CentOS 7 does not fail (== return exit code of 0) when some requested packages are not found.

AkihiroSuda (approved):

mrunalp (approved):

PR #3615, opened 2022-09-28
Replace !IsNotExist with IsExist

thesw4rm:
Removed unnecessary boolean logic.

Signed-off-by: Yadunandan Pillai <ytpillai@pm.me>


kolyshkin (changes requested):
Please read the documentation about both `os.IsExist` and `os.IsNotExist`, and understand that `!os.IsNotExist` is not equivalent to `os.IsExist`.

thesw4rm:
> Please read the documentation about both `os.IsExist` and `os.IsNotExist`, and understand that `!os.IsNotExist` is not equivalent to `os.IsExist`.

Sorry! Was too hasty.

PR #3620, opened 2022-10-02
[1.1] Fixes inability to use /dev/null when inside a container

evanphx:
The original code depended on the origin filesystem to have /dev/{block,char} populated. This is done by udev normally and while is very common non-containerized systemd installs, it's very easy to start systemd in a container created by runc itself and not have /dev/{block,char} populated. When this occurs, the following error output is observed:

$ docker run hello-world
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error reopening /dev/null inside container: open /dev/null: operation not permitted: unknown.

/dev/null can't be opened because it was not added to the deviceAllowList, as there was no /dev/char directory. The change here utilizes the fact that when sysfs in in use, there is a /sys/dev/{block,char} that is kernel maintained that we can check.

I'd also recommend having a base allow list that includes /dev/null (1:3) since runc implicitly depends on it.

### DIFF
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -293,8 +293,18 @@ func generateDeviceProperties(r *configs.Resources) ([]systemdDbus.Property, err
 			// rules separately to systemd) we can safely skip entries that don't
 			// have a corresponding path.
 			if _, err := os.Stat(entry.Path); err != nil {
-				logrus.Debugf("skipping device %s for systemd: %s", entry.Path, err)
-				continue
+				// Also check /sys/dev so that we don't depend on /dev/{block,char}
+				// being populated. (/dev/{block,char} is populated by udev, which
+				// isn't strictly required for systemd). Ironically, this happens most
+				// easily when starting containerd within a runc created container
+				// itself.
+
+				// We don't bother with securejoin here because we create entry.Path
+				// right above here, so we know it's safe.
+				if _, err := os.Stat("/sys" + entry.Path); err != nil {
+					logrus.Warnf("skipping device %s for systemd: %s", entry.Path, err)
+					continue
+				}
 			}
 		}
 		deviceAllowList = append(deviceAllowList, entry)


AkihiroSuda:
What distro needs this fix?

libcontainer/cgroups/systemd/common.go
```diff
@@ -293,8 +293,14 @@ func generateDeviceProperties(r *configs.Resources) ([]systemdDbus.Property, err
 			// rules separately to systemd) we can safely skip entries that don't
 			// have a corresponding path.
 			if _, err := os.Stat(entry.Path); err != nil {
-				logrus.Debugf("skipping device %s for systemd: %s", entry.Path, err)
-				continue
+				// Also check /sys/dev so that we don't depend on /dev/{block,char} being populated.
+				// (/dev/{block,char} is populated by udev, which isn't strictly required for systemd).
+				// Ironically, this happens most easily when starting containerd within a runc created
+				// container itself.
+				if _, err := os.Stat("/sys" + entry.Path); err != nil {
```
AkihiroSuda, line 300: Please use securejoin

https://github.com/opencontainers/runc/search?q=securejoin

evanphx: Done. Wasn't sure we needed to do that since entry.Path is created in the lines right above this and thusly we know it's content is safe.

evanphx:
It's not a specific distro that needs the fix, it was more of a one-off. I was experimenting with running systemd inside a containerd managed container, with docker installed inside the that container (so, nested containerds). The installation of systemd didn't pull in udev (it was ubuntu) and thusly /dev/{block,char} were not created.

thaJeztah:
Looks like this PR was opened against the 1.1 branch; the standard flow is to open against main, and then cherry-pick backports; could you change the PR to tarter "main"?

evanphx:
@thaJeztah The code on main has been reorganized so the patch wouldn't be easily backported. I'm happy to open an additional PR with the fix for main.

thaJeztah:
Thanks! Just making sure that we have all fixes in `main` (otherwise we may forget porting, and end up with regressions if a release is cut from the main branch 😅)

kolyshkin:
I suggest we discuss this in #3623 first.

libcontainer/cgroups/systemd/common.go
kolyshkin: > Done. Wasn't sure we needed to do that since entry.Path is created in the lines right above this and thusly we know it's content is safe.

I concur.

kolyshkin:
#3623 is merged, please backport

evanphx:
@kolyshkin ready to go!

kolyshkin:
CI centos-stream-8 timed out; restarted.

kolyshkin (dismissed):
LGTM

AkihiroSuda (changes requested):
Please use `git cherry-pick -x` for tracking the main commit

evanphx:
@AkihiroSuda No problem, done!

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

mprasil:
> It's not a specific distro that needs the fix, it was more of a one-off.

Just a follow up on this. @evanphx's fix is actually useful for systems running on top of LXD. (possibly in certain configuration, although I [hit this](https://github.com/opencontainers/runc/discussions/3795) with very standard setup IMO) So this is probably more common than you'd expect.

I am certainly very grateful for this fix. 🙏

PR #3623, opened 2022-10-03
Fixes inability to use /dev/null when inside a container

evanphx:
This is a forward port of https://github.com/opencontainers/runc/pull/3620

The original code depended on the origin filesystem to have /dev/{block,char} populated. This is done by udev normally and while is very common non-containerized systemd installs, it's very easy to start systemd in a container created by runc itself and not have /dev/{block,char} populated. When this occurs, the following error output is observed:

$ docker run hello-world
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error reopening /dev/null inside container: open /dev/null: operation not permitted: unknown.

/dev/null can't be opened because it was not added to the deviceAllowList, as there was no /dev/char directory. The change here utilizes the fact that when sysfs in in use, there is a /sys/dev/{block,char} that is kernel maintained that we can check.

Signed-off-by: Evan Phoenix <evan@phx.io>

### DIFF
--- a/libcontainer/cgroups/devices/systemd.go
+++ b/libcontainer/cgroups/devices/systemd.go
@@ -133,8 +133,18 @@ func systemdProperties(r *configs.Resources) ([]systemdDbus.Property, error) {
 			// rules separately to systemd) we can safely skip entries that don't
 			// have a corresponding path.
 			if _, err := os.Stat(entry.Path); err != nil {
-				logrus.Debugf("skipping device %s for systemd: %s", entry.Path, err)
-				continue
+				// Also check /sys/dev so that we don't depend on /dev/{block,char}
+				// being populated. (/dev/{block,char} is populated by udev, which
+				// isn't strictly required for systemd). Ironically, this happens most
+				// easily when starting containerd within a runc created container
+				// itself.
+
+				// We don't bother with securejoin here because we create entry.Path
+				// right above here, so we know it's safe.
+				if _, err := os.Stat("/sys" + entry.Path); err != nil {
+					logrus.Warnf("skipping device %s for systemd: %s", entry.Path, err)
+					continue
+				}
 			}
 		}
 		deviceAllowList = append(deviceAllowList, entry)


libcontainer/cgroups/devices/systemd.go
```diff
@@ -133,8 +134,21 @@ func systemdProperties(r *configs.Resources) ([]systemdDbus.Property, error) {
...
-				continue
+				// Also check /sys/dev so that we don't depend on /dev/{block,char}
+				// being populated. (/dev/{block,char} is populated by udev, which
+				// isn't strictly required for systemd). Ironically, this happens most
+				// easily when starting containerd within a runc created container
+				// itself.
+				testPath, err := securejoin.SecureJoin("/sys", entry.Path)
+				if err != nil {
+					logrus.Errorf("error joining entry path: %s", err)
+					continue
```
kolyshkin, line 145: If that's an error, I guess we should not continue but return it.

kolyshkin: Or maybe we should not use securejoin here, since this is merely a check that such device exists, and we do not use `testPath` in any other way. Worst case scenario, a bad guy will trick the system into adding a line for a device that does not exist into systemd unit properties.

@cyphar WDYT?

thaJeztah: The original patch didn't have secure join, but I saw @AkihiroSuda preferred having it (not sure if it's _really_ needed here indeed); see https://github.com/opencontainers/runc/pull/3620#discussion_r985375647

kolyshkin: @AkihiroSuda WDYT?

AkihiroSuda: SGTM, but probably the reason not to use securejoin should be recorded in the comment lines in the source

evanphx: No problem, I removed securejoin and added a comment why it's not used.

cyphar (approved):
LGTM, though I do wonder if we could just always use `/sys/dev/...`.

libcontainer/cgroups/devices/systemd.go
```diff
@@ -133,8 +133,18 @@ func systemdProperties(r *configs.Resources) ([]systemdDbus.Property, error) {
...
-				continue
+				// Also check /sys/dev so that we don't depend on /dev/{block,char}
+				// being populated. (/dev/{block,char} is populated by udev, which
+				// isn't strictly required for systemd). Ironically, this happens most
+				// easily when starting containerd within a runc created container
+				// itself.
+
+				// We don't bother with securejoin here because we create entry.Path
+				// right above here, so we know it's safe.
+				if _, err := os.Stat("/sys" + entry.Path); err != nil {
```
cyphar, line 144: I just realised we probably want to do `entry.Path = "/sys" + entry.Path` here, don't we? Otherwise systemd will get a `/dev/...` path that doesn't exist?

evanphx: So I’m not entirely clear on what systemd does with these values but I can tell you that in my setup, those /dev/char entries don’t exist and systemd still does the right thing. Maybe it parses the path and uses the major:minor directly?

kolyshkin: Nah, the `/sys/dev/{block,char}/<major>:<minor>` is a directory, not a device file, so we can't use it for systemd.

Perhaps systemd has a set of built-in devices it can recognize. I'll take a look.

kolyshkin:
backport to 1.1 is done in #3620

PR #3627, opened 2022-10-06
Support fetching PSI stats for cgroupv2 containers [dnm / carry]

thaJeztah:
Testing my suggestions from https://github.com/opencontainers/runc/pull/3358 / https://github.com/dqminh/runc/pull/35

### DIFF
--- a/types/events.go
+++ b/types/events.go
@@ -78,7 +78,7 @@ type CpuUsage struct {
 type Cpu struct {
 	Usage      CpuUsage   `json:"usage,omitempty"`
 	Throttling Throttling `json:"throttling,omitempty"`
-	PSI        PSIStats   `json:"psi,omitempty"`
+	PSI        *PSIStats  `json:"psi,omitempty"`
 }
 
 type CPUSet struct {
@@ -109,7 +109,7 @@ type Memory struct {
 	Kernel    MemoryEntry       `json:"kernel,omitempty"`
 	KernelTCP MemoryEntry       `json:"kernelTCP,omitempty"`
 	Raw       map[string]uint64 `json:"raw,omitempty"`
-	PSI       PSIStats          `json:"psi,omitempty"`
+	PSI       *PSIStats         `json:"psi,omitempty"`
 }
 
 type L3CacheInfo struct {
--- a/types/events.go
+++ b/types/events.go
@@ -21,6 +21,18 @@ type Stats struct {
 	NetworkInterfaces []*NetworkInterface `json:"network_interfaces"`
 }
 
+type PSIData struct {
+	Avg10  float64 `json:"avg10"`
+	Avg60  float64 `json:"avg60"`
+	Avg300 float64 `json:"avg300"`
+	Total  uint64  `json:"total"`
+}
+
+type PSIStats struct {
+	Some PSIData `json:"some,omitempty"`
+	Full PSIData `json:"full,omitempty"`
+}
+
 type Hugetlb struct {
 	Usage   uint64 `json:"usage,omitempty"`
 	Max     uint64 `json:"max,omitempty"`
@@ -43,6 +55,7 @@ type Blkio struct {
 	IoMergedRecursive       []BlkioEntry `json:"ioMergedRecursive,omitempty"`
 	IoTimeRecursive         []BlkioEntry `json:"ioTimeRecursive,omitempty"`
 	SectorsRecursive        []BlkioEntry `json:"sectorsRecursive,omitempty"`
+	PSI                     PSIStats     `json:"psi,omitempty"`
 }
 
 type Pids struct {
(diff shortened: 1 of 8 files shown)


kolyshkin:
```
Retrieving https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
Preparing...                          ########################################
	package epel-release-7-14.noarch is already installed
```

Ah! they fixed it and now our CI is b0rked again.

kolyshkin:
> ```
> Retrieving https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
> Preparing...                          ########################################
> 	package epel-release-7-14.noarch is already installed
> ```
> 
> Ah! they fixed it and now our CI is b0rked again.

Fixed by https://github.com/opencontainers/runc/pull/3628

types/events.go
```diff
@@ -21,6 +24,11 @@ type Stats struct {
 	NetworkInterfaces []*NetworkInterface `json:"network_interfaces"`
 }
 
+type (
+	PSIData  = cgroups.PSIData
```
thaJeztah, line 28: FWIW, not sure if we need the aliases at all; I need to look into history to understand why we did so for the other ones (separate types from libcontainer); perhaps we don't need those either.

It _may_ have been to allow external consumer to use the "types" package without requiring libcontainer, but if that was the design, we already broke that.

If, on the other hand, nobody is expected to consume "types", we may not need all the duplicated definitions and we could just alias or use them directly

thaJeztah:
Failure on centos stream 9;
https://github.com/opencontainers/runc/pull/3627/checks?check_run_id=8779692013

```
# runc events --stats test_busybox (status=1):
# time="2022-10-08T10:57:34Z" level=error msg="unable to get container cgroup stats: error while statting cgroup v2: [unable to parse /sys/fs/cgroup/system.slice/runc-test_busybox.scope/cpu.pressure: read /sys/fs/cgroup/system.slice/runc-test_busybox.scope/cpu.pressure: operation not supported unable to parse /sys/fs/cgroup/system.slice/runc-test_busybox.scope/memory.pressure: read /sys/fs/cgroup/system.slice/runc-test_busybox.scope/memory.pressure: operation not supported unable to parse /sys/fs/cgroup/system.slice/runc-test_busybox.scope/io.pressure: read /sys/fs/cgroup/system.slice/runc-test_busybox.scope/io.pressure: operation not supported]"
ok 55 events --stats with psi data # skip test requires psi
```

thaJeztah:
Still not sure what's failing on Fedora (tests seem to pass, but then it marks it as failed (exit status 1))

types/events.go
dqminh: We expect `types` to be consistent i.e. not depending on `libcontainer` internals, i..e consumer of `runc events` exect the format to be consistent even though we may have change `libcontainer/cgroups` for some reasons.

So some of the duplication is intentional, and some differences are also intentional i.e. `camelCase` vs `snake_case`.

kolyshkin:
```
=== RUN   TestNilResources
--- FAIL: TestNilResources (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x59c2b6]
goroutine 6 [running]:
testing.tRunner.func1.2({0x5cc6e0, 0x75b020})
	/usr/lib/golang/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/usr/lib/golang/src/testing/testing.go:1392 +0x39f
panic({0x5cc6e0, 0x75b020})
	/usr/lib/golang/src/runtime/panic.go:838 +0x207
github.com/opencontainers/runc/libcontainer/cgroups/fs2.statPSI({0xc00001a180, 0x29}, {0x6009ca, 0xc}, 0x0)
	/vagrant/libcontainer/cgroups/fs2/psi.go:33 +0x1f6
github.com/opencontainers/runc/libcontainer/cgroups/fs2.(*manager).GetStats(0xc000054060)
	/vagrant/libcontainer/cgroups/fs2/fs2.go:122 +0x5fe
github.com/opencontainers/runc/libcontainer/cgroups/manager.TestNilResources(0xc00007b1e0)
	/vagrant/libcontainer/cgroups/manager/manager_test.go:40 +0x1ad
testing.tRunner(0xc00007b1e0, 0x60e8d0)
	/usr/lib/golang/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/usr/lib/golang/src/testing/testing.go:1486 +0x35f
FAIL	github.com/opencontainers/runc/libcontainer/cgroups/manager	0.014s
```

This is the test case I've added in commit 57edce4659725c7c4a9727e7cca5e9bc27898572to make sure we don't panic if resources were not provided.

kolyshkin:
>	if err := statPSI(m.dirPath, "cpu.pressure", st.CpuStats.PSI); err != nil {

Apparently the PSI pointer is never initialized so it's nil.

Perhaps it's better to revert the struct to pointer-to-struct conversion patch (and ditch the `omitempty` as it does not make sense).

PR #3628, opened 2022-10-07
cirrus-ci: install EPEL on CentOS 7 conditionally

kolyshkin:
This is a followup to https://github.com/opencontainers/runc/pull/3614. Apparently, GCE fixed the centos 7 image issue, which broke our CI for the second time.

Let's install EPEL conditionally, so it will work either way.


thaJeztah (approved):
LGTM, thanks!

Issue #3612, opened 2022-09-26
outdated and inacurrate comment


godsarmy:
i start a container with only pid namespace and i find when runc is killed, subprocesses are not always waited unless --no-subreaper is applied,

Therefore the comment of https://github.com/opencontainers/runc/blob/main/libcontainer/init_linux.go#L525 is outdated and inaccurate. If you agree, i can submit a PR to correct the comment.

PR #3631, opened 2022-10-09
Fix comment of signalAllProcesses for process wait due to sigkill

godsarmy:
ref #3612 the old comment is outdated

### DIFF
--- a/libcontainer/init_linux.go
+++ b/libcontainer/init_linux.go
@@ -522,7 +522,8 @@ func isWaitable(pid int) (bool, error) {
 
 // signalAllProcesses freezes then iterates over all the processes inside the
 // manager's cgroups sending the signal s to them.
-// If s is SIGKILL then it will wait for each process to exit.
+// If s is SIGKILL and subreaper is not enabled then it will wait for each
+// process to exit.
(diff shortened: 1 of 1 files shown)


kolyshkin (approved):
LGTM

kolyshkin:
@AkihiroSuda @thaJeztah PTAL

libcontainer/init_linux.go
```diff
@@ -522,7 +522,8 @@ func isWaitable(pid int) (bool, error) {
 
 // signalAllProcesses freezes then iterates over all the processes inside the
 // manager's cgroups sending the signal s to them.
-// If s is SIGKILL then it will wait for each process to exit.
+// If s is SIGKILL and subreaper is not enabled then it will wait for each
+// process to exit.
```
thaJeztah, line 526: nit (and not new); try to avoid future tense in tech-docs. This could be

```suggestion
// If s is SIGKILL and subreaper is not enabled then it waits for each
// process to exit.
```

thaJeztah (approved):
LGTM (one suggestion, but not a blocker for me if you want to merge as-is)

kolyshkin (approved):
still LGTM

PR #3632, opened 2022-10-12
Libcontainer README update to include factory for container creation

nravic:
Was a little confused getting started with libcontainer today, which the package [documentation](https://pkg.go.dev/github.com/opencontainers/runc/libcontainer#section-readme) helped with. 

Quick PR to get parity w/ the docs to prevent further confusion.

### DIFF
--- a/libcontainer/README.md
+++ b/libcontainer/README.md
@@ -32,7 +32,8 @@ func init() {
 	if len(os.Args) > 1 && os.Args[1] == "init" {
 		runtime.GOMAXPROCS(1)
 		runtime.LockOSThread()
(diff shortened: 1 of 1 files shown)


kolyshkin:
@nravic this does not make sense. Most probably you try to move back to whatever we had in runc 1.1.

Please see PR #3373.

nravic:
Ahh my mistake then! Then the godocs are incorrect - I can open a PR to correct those instead.

kolyshkin:
Godocs are probably fine, too, but they are shown from the last released
version (1.1.4), which doesn't have these new changes. This, I guess, is
the source of confusion.

Thanks for looking into it.

PR #3655, opened 2022-11-03
runc checkpoint: destroy only on success

kolyshkin:
If checkpointing has failed, the container is kept running. We do not want to, and we can't remove it in such case.

Do not try to remove the container if there's an error from checkpointing.

This avoids an unclear error message from destroy() saying "container still running" or "container paused".

While at it, avoid using defer since it does not make a lot of sense here.

Fixes: #3577

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -67,10 +67,6 @@ checkpointed.`,
 			return err
 		}
 
-		if !(options.LeaveRunning || options.PreDump) {
-			// destroy container unless we tell CRIU to keep it
-			defer destroy(container)
-		}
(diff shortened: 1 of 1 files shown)


kolyshkin:
@gosoon @adrianreber @rst0git PTAL

adrianreber (approved):
It feels like we had to deal with something like this a couple of years ago. Looks good to me.

AkihiroSuda (approved):

rst0git (approved):
LGTM

mrunalp (approved):

PR #3660, opened 2022-11-11
fix libcontainer example

yzxiu:
Signed-off-by: yaozhenxiu <946666800@qq.com>

fix libcontainer example

### DIFF
--- a/libcontainer/README.md
+++ b/libcontainer/README.md
@@ -45,7 +45,7 @@ struct describing how the container is to be created. A sample would look simila
 
 ```go
 defaultMountFlags := unix.MS_NOEXEC | unix.MS_NOSUID | unix.MS_NODEV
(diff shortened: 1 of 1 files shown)


kolyshkin:
Some history:
1. Commit 677baf22d2481 moved `DeviceRule` (et al) from `libcontainer/configs` to `libcontainers/devices` package (leaving the backward compatibility) 
2. Commit 4fc2de77e993f86d41f3 renamed "DeviceRule` to `Rule` to avoid stuttering
3. Commit 3f23a736cb28933 removed the backward compatibility earlier added in (1)

Neither of those touched libcontainer/README.md, and after (3) it is no longer valid.

yzxiu:
@kolyshkin Thanks for your reviews, that is real name and email.

AkihiroSuda (approved):

PR #3667, opened 2022-11-16
support to print stack

ningmingxiao:
Signed-off-by: ningmingxiao <ning.mingxiao@zte.com.cn>
sometimes runc is hanged ,this will show where hang

### DIFF
--- a/init.go
+++ b/init.go
@@ -1,16 +1,23 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"os/signal"
 	"runtime"
 	"strconv"
 
 	"github.com/opencontainers/runc/libcontainer"
 	_ "github.com/opencontainers/runc/libcontainer/nsenter"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 func init() {
+	// Support to print stack when receive signal SIGUSR1
+	sc := make(chan os.Signal, 1)
+	handleSignals(sc)
+	signal.Notify(sc, unix.SIGUSR1)
 	if len(os.Args) > 1 && os.Args[1] == "init" {
 		// This is the golang entry point for runc init, executed
 		// before main() but after libcontainer/nsenter's nsexec().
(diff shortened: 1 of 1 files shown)


init.go
```diff
@@ -1,16 +1,24 @@
...
 	_ "github.com/opencontainers/runc/libcontainer/nsenter"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 func init() {
+	// Support to print stack when receive signal SIGUSR1
+	signals := make(chan os.Signal, 2048)
+	handleSignals(signals)
+	signal.Notify(signals)
```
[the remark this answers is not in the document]
ningmingxiao: done

init.go
```diff
@@ -40,3 +48,44 @@ func init() {
...
+	if writeToFile {
+		// Also write to file to aid gathering diagnostics
+		name := filepath.Join(os.TempDir(), fmt.Sprintf("runc.%d.stacks.log", os.Getpid()))
+		f, err := os.Create(name)
+		if err != nil {
+			return
+		}
+		defer f.Close()
+		_, _ = f.WriteString(string(buf))
+	}
```
cyphar, line 90: There are several ways I can imagine this leading to a vulnerability (a container process would be able to trigger this while runc is joining -- and the non-dumpable bit doesn't help here -- thus causing a more privileged write to a predictable path which could be a symlink or magic-link pointing to who knows where).

I'd prefer we not have this "write to a random file" feature at all. This information should just be output to stderr.

ningmingxiao: If I print these messages to  stderr, how can I get this messages if I don't save these message to a file.

cyphar: If you use `--log` with runc you can pick a file to write to, and in Docker and containerd, runc logs are piped through to their logs so you should be able to access it with `--debug` (or something similar). Another huge benefit of logging (through `logrus` at least) is that if the runc process inside the container gets stuck you'll get the stack information in the host (rather than dumping it somewhere in the container).

ningmingxiao: I print stack into log.json

ningmingxiao:
but  after unix.Close(l.logFd) at standard_init_linux.go , I can't use logrus to print stack, this will miss some messages

kolyshkin:
> but after unix.Close(l.logFd) at standard_init_linux.go , I can't use logrus to print stack, this will miss some messages @kolyshkin

Yes, since the logs are being forwarded to the parent, the parent has to live until there are logs. Previously (before #2835) sometimes the parent was exiting prematurely, resulting in some logs being lost. Now, we do wait for the child to close the log fd, to make sure all the logs are forwarded.

All it means for this case is it's not possible to use logrus from runc init after closing the logs. I am not sure how to handle this, perhaps the bold approach would be to remove logs forwarding goroutine entirely, instead passing the logrus fd directly to init. This needs some thinking and experimenting though

kolyshkin:
> the bold approach would be to remove logs forwarding goroutine entirely, instead passing the logrus fd directly to init. 

This might work, but we need to teach the C code to work the same way as logrus -- meaning, produce either text or json, depending on settings and whether fd is a terminal.

ningmingxiao:
> the bold approach would be to remove logs forwarding goroutine entirely

I don't  understand "the bold approach would be to remove logs forwarding goroutine entirely" can you explain it ?

kolyshkin:
> I don't understand "the bold approach would be to remove logs forwarding goroutine entirely" can you explain it ?

For `runc init`, we use a goroutine that receives log entries from the C code in `libcontainer/nsenter` and sends them to logrus (see `ForwardLogs` in `libcontainer/logs`).

In theory, we can remove this log forwarding, instead passing the log file descriptor to C code, and log directly from there.

PR #3669, opened 2022-11-16
[debug] add a bad comment

kolyshkin:
Checking if godot is working.


kolyshkin:
Hmm, as I suspected, lint-extra job is not working. Investigating.

PR #3684, opened 2022-12-09
Check userns

g0dA:
Issue: https://github.com/opencontainers/runc/issues/3672

in user namespace, crun bind /sys not sysfs, so should do the same for runc like https://github.com/containers/crun/commit/6785cefbdf982c97a5552c9ce7017b0e8309c291

Signed-off-by: L_AnG [lyi4ng@gmail.com](mailto:lyi4ng@gmail.com)


libcontainer/utils/utils.go
```diff
@@ -19,6 +19,9 @@ const (
 	exitSignalOffset = 128
 )
 
+// RunInUserns is the user namespace state of the process.
+var RunInUserns = -1
```
AkihiroSuda, line 23: You can use this https://github.com/opencontainers/runc/blob/e7461c8c29ab387962255e12ea50d02fc29bb4ec/libcontainer/userns/userns.go#L5

AkihiroSuda (changes requested):

libcontainer/rootfs_linux.go
```diff
@@ -1089,7 +1089,21 @@ func mountPropagate(m *configs.Mount, rootfs string, mountLabel string, mountFd
 	if err := utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
 		return mount(source, m.Destination, procfd, m.Device, uintptr(flags), data)
 	}); err != nil {
-		return err
+		if m.Device == "sysfs" {
+			/* If we are running in a user namespace, just bind mount /sys if creating
+			   sysfs failed.  */
```
AkihiroSuda, line 1094: Any reason not to let the higher level runtime generate config.json with this bind mount ?

libcontainer/utils/utils.go
AkihiroSuda: Any why is this in mountPropagate()? Doesn’t seem relevant to mount propagation

libcontainer/rootfs_linux.go
```diff
@@ -1089,7 +1089,21 @@ func mountPropagate(m *configs.Mount, rootfs string, mountLabel string, mountFd
...
+			ret := utils.CheckRunningInUserNamespace()
+			if ret < 0 {
+				return err
+			}
+			if err = utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
+				return mount("/sys", m.Destination, procfd, "/sys", unix.MS_BIND|unix.MS_REC|unix.MS_SLAVE, data)
```
AkihiroSuda, line 1100: What if flags contain MS_RDONLY

libcontainer/utils/utils.go
g0dA: > You can use this
> 
> https://github.com/opencontainers/runc/blob/e7461c8c29ab387962255e12ea50d02fc29bb4ec/libcontainer/userns/userns.go#L5

Oh, thanks, i just know that there is `RunningInUserNS` in the code

> Any why is this in mountPropagate()? Doesn’t seem relevant to mount propagation

In fact, this logic should be placed in `mountToRootfs`, but this will make it a separate mount operation, so I will put it in `mountPropagate`.

If it is not suitable, it can still be put into `mountToRootfs`

libcontainer/rootfs_linux.go
g0dA: > What if flags contain MS_RDONLY

Yes, but not necessary, because this is in a rootless environment

libcontainer/rootfs_linux.go
g0dA: This is problematic because /sys/class/net entries are network-namespaced, if you use privileged mode, there may be escape problems, but rootless don't need to worry

libcontainer/rootfs_linux.go
AkihiroSuda: /sys main contain user-writable directories especially under /sys/fs/cgroup.
So it has to be mounted as recursively read-only (`mount_setattr(2)` with `AT_RECURSIVE`)

AkihiroSuda: And needs integration tests

g0dA: ok, i try

kolyshkin (changes requested):
I don't like the way this is handled in `mountPropagate`. This fallback should be done in `mountToRootfs` instead.

kolyshkin:
One other thing, specifying noexec, nodev, nosuid is useless for a bind mount, since the bind mount has the same mount options as the underlying mount.

kolyshkin:
I was wrong about noexec, nodev, nosuid flags (and removed the comment) -- those are per-mount-point flags.

libcontainer/rootfs_linux.go
```diff
@@ -400,7 +400,29 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
...
+		if err != nil && m.Device == "sysfs" {
+			/* If we are running in a user namespace, just bind mount /sys if creating
+			   sysfs failed.  */
+			if !userns.RunningInUserNS() {
+				return err
+			}
+			m.Source, m.Device = "/sys", "bind"
+			m.Flags = unix.MS_BIND | unix.MS_REC | unix.MS_NODEV | unix.MS_NOSUID | unix.MS_NOEXEC
+			if err = utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
+				return mount(m.Source, m.Destination, procfd, m.Device, uintptr(m.Flags), label.FormatMountLabel(m.Data, mountLabel))
```
kolyshkin, line 413: Why are you changing `m` here?

libcontainer/rootfs_linux.go
```diff
@@ -400,7 +400,29 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
...
+			}); err != nil {
+				return err
+			}
+			m.RecAttr = &unix.MountAttr{
+				Attr_set: unix.MOUNT_ATTR_RDONLY,
+			}
+			err = setRecAttr(m, rootfs)
```
kolyshkin, line 420: Ah, so you can use setRecAttr? I think it's easier not to change `m` but use `unix.MountSetattr` directly.

libcontainer/rootfs_linux.go
```diff
@@ -400,7 +400,29 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
...
+			}
+			m.Source, m.Device = "/sys", "bind"
+			m.Flags = unix.MS_BIND | unix.MS_REC | unix.MS_NODEV | unix.MS_NOSUID | unix.MS_NOEXEC
+			if err = utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
+				return mount(m.Source, m.Destination, procfd, m.Device, uintptr(m.Flags), label.FormatMountLabel(m.Data, mountLabel))
+			}); err != nil {
+				return err
+			}
+			m.RecAttr = &unix.MountAttr{
+				Attr_set: unix.MOUNT_ATTR_RDONLY,
```
kolyshkin, line 418: I think it should be set to RO only if the original flags had RDONLY flag set.

libcontainer/rootfs_linux.go
kolyshkin, line 413: For a bind mount, fstype and data arguments are ignored, no need to set or pass those.

libcontainer/rootfs_linux.go
g0dA: yes, i can use `setRecAttr`, because I hope to use `utils.WithProcfd` to avoid security issues

libcontainer/rootfs_linux.go
g0dA: Because I think this mount logic is equivalent to modifying `configs.Mount`, so I hope it can be expressed in the code

g0dA: yes, you are right, thanks

AkihiroSuda:
@kolyshkin @g0dA Please check your gmail boxes (if you haven't checked)

g0dA:
ok

Akihiro Suda ***@***.***> 于2023年1月5日周四 15:50写道：

libcontainer/rootfs_linux.go
g0dA: If `MOUNT_ATTR_RDONLY` is set only when `RDONLY` is originally set, there will be security issues when `/sys/fs/cgroup` is not set in config.json.

Maybe it can be like the modification of crun, only when `RDONLY` is originally set, then `sysfs` => `/sys`

AkihiroSuda:
The crun implementation is being updated in
- https://github.com/containers/crun/pull/1114

runc should follow the same convention

aviallon:
Has this been fixed yet? I still have the dreaded issue.

Issue #3696, opened 2023-01-11
libcontainer/configs: mounts is no longer OS agnostic


egernst:
libcontainer/configs build on Darwin now fails with the following two changes:

- https://github.com/opencontainers/runc/commit/9c444070ec7bb83995dbc0185da68284da71c554
- https://github.com/opencontainers/runc/commit/382eba4354d764aaffec82720c23ef429496acf9

We should make sure mount.go is split out for _linux and/or unsupported.

PR #3697, opened 2023-01-11
libcontainer: configs: ensure can build on darwin

egernst:
Addresses https://github.com/opencontainers/runc/issues/3696

Signed-off-by: Eric Ernst <eric_ernst@apple.com>


egernst:
AFAICT, these CI failures aren't triggered by my change

libcontainer/configs/mount_linux.go
```diff
@@ -0,0 +1,36 @@
+package configs
+
+import "golang.org/x/sys/unix"
+
+type Mount struct {
```
rata, line 5: IMHO The struct duplication seems not ideal. I see they are mostly identical, except the non linux doesn't have the unix bits.

I'm not sure we need two structs, but even if we do, can't we have one with the common parts and just embed it here?  I mean embed as shown in these examples: https://gobyexample.com/struct-embedding

libcontainer/configs/mount.go
```diff
@@ -1,42 +1,7 @@
 package configs
 
-import "golang.org/x/sys/unix"
```
rata, line 3: What is the issue with having this here? Mac is a unix, so this should work on mac, right? Why can't we share the same struct here?

Also, if Mac doesn't support bind mounts (do they really don't support it?), can't we just define that function in different platforms? Why do we need to change the struct too?

libcontainer/configs/mount_unsupported.go
```diff
@@ -0,0 +1,34 @@
+//go:build !linux
+// +build !linux
```
rata, line 2: Why unsupported instead of unix?

rata (commented):
It would be great if you can paste the compilation error you see (not all the lines, just the important ones if it is too verbose), so we can easily know how this fails and why this fixes it.

With the info you posted (here and in the issue) is hard to know if some changes are needed (posted those inline).

Also, why do you need this to compile on Mac? I mean, none of the cgroup stuff works there, right? Is it even a supported platform for runc?

I'm fine with the PR, but if we don't want to regress, we should probably add a test to compile on Mac, that I'm not sure what maintainers think about that.

libcontainer/configs/mount_unsupported.go
egernst: I'm following existing pattern in the package (ie, see cgroups_unsupported.go)

libcontainer/configs/mount_linux.go
egernst: > It would be great if you can paste the compilation error you see (not all the lines, just the important ones if it is too verbose), so we can easily know how this fails and why this fixes it.
>

Updated the PR description 
 

> Also, why do you need this to compile on Mac? I mean, none of the cgroup stuff works there, right? Is it even a supported platform for runc?
> 

Libcontainer configs is utilized in other projects, including Kata Containers, which is adding support for building and running on Darwin.

> I'm fine with the PR, but if we don't want to regress, we should probably add a test to compile on Mac, that I'm not sure what maintainers think about that.

Sure - you don't need to be on Mac. Just run "GOOS=darwin go build".

egernst:
> t would be great if you can paste the compilation error you see (not all the lines, just the important ones if it is too verbose), so we can easily know how this fails and why this fixes it.
Good call - updated the PR description

> Also, why do you need this to compile on Mac? I mean, none of the cgroup stuff works there, right? Is it even a supported platform for runc?

Libcontainer configs is utilized in other projects, including Kata Containers, which is adding support for building and running on Darwin.

> I'm fine with the PR, but if we don't want to regress, we should probably add a test to compile on Mac, that I'm not sure what maintainers think about that.

Sure - you don't need to be on Mac. Just run "GOOS=darwin go build".

libcontainer/configs/mount.go
egernst: See compilation issue,.  Indeed it doesn't support bind mounts, as well as MountAttr

libcontainer/configs/mount_linux.go
egernst: Agreed that it's sub-optimal.

libcontainer/configs/mount.go
rata: Thanks!

libcontainer/configs/mount_unsupported.go
rata: Oh, makes sense. Thanks!

rata:
> > Also, why do you need this to compile on Mac? I mean, none of the cgroup stuff works there, right? Is it even a supported platform for runc?
> 
> Libcontainer configs is utilized in other projects, including Kata Containers, which is adding support for building and running on Darwin.

Ohh, seems reasonable. Can you add that to the PR description and commit message?

> > I'm fine with the PR, but if we don't want to regress, we should probably add a test to compile on Mac, that I'm not sure what maintainers think about that.

> 
> Sure - you don't need to be on Mac. Just run "GOOS=darwin go build".

Right, thanks! :)

libcontainer/configs/mount_linux.go
egernst: I created a base struct which can be used to help dedup, PTAL -- thank for the suggestion!

egernst:
Updated the commit/PR to add more detail, and made the changes to dedup the struct based on your feedback.  Thanks Rodrigo!

libcontainer/configs/mount.go
```diff
@@ -1,14 +1,12 @@
...
-
 const (
 	// EXT_COPYUP is a directive to copy up the contents of a directory when
 	// a tmpfs is mounted over it.
 	EXT_COPYUP = 1 << iota //nolint:golint // ignore "don't use ALL_CAPS" warning
 )
 
-type Mount struct {
+type mount_base struct {
```
rata, line 9: s/_b/B?

rata (approved):
LGTM, thanks!

libcontainer/configs/mount.go
egernst: ugh I haven't Go'd in a while, my bad!

rata (approved):
LGTM

rata:
Ups, the linter seems very unhappy about this change. Can you fix it? :)

egernst:
Ouch, it seems like the golangci-linter still is not able to handle embedded structs (looking at https://github.com/golangci/golangci-lint/issues/1517). I'm open to feedback on how we should progress (moving away from embedded struct to appease the linting tool, or... ?).

egernst:
Ouch, this deals less with lint, and more with completion failure. My bad.

This issue is not being able to directly compose a struct that has an embedded struct. An example, we cannot directly initialize the embedded structs fields:
```golang
			m := &configs.Mount{Destination: node.Path, Source: node.Path}
```

This results in the `Field Destination doesn't exist`. We would need to do something like:
```golang
			m := &configs.Mount{}
			m.Destination = node.Path
			m.Source = node.Path
```

I don't love the making that change throughout the code and this the UX in Golang for this is quite lacking. Curious to hear feedback on how you think it'd make sense to proceed here.

libcontainer/configs/mount_unsupported.go
```diff
@@ -0,0 +1,34 @@
+//go:build !linux
```
dcantah, line 1: Should we just make this an empty struct like what was done with [cgroups](https://github.com/opencontainers/runc/blob/da548c1c1b0c43cd8718636e8acfdb536e010bbe/libcontainer/configs/cgroup_unsupported.go#L9) and [namespaces](https://github.com/opencontainers/runc/blob/da548c1c1b0c43cd8718636e8acfdb536e010bbe/libcontainer/configs/namespaces_unsupported.go#L8). We don't need anything from the structure right? It's just because libcontainer/configs is imported at all that's halting compilation

libcontainer/configs/mount_unsupported.go
```diff
@@ -0,0 +1,34 @@
...
+// +build !linux
+
+package configs
+
+type Mount struct {
```
dcantah, line 6: Should we just make this an empty struct like what was done with [cgroups](https://github.com/opencontainers/runc/blob/da548c1c1b0c43cd8718636e8acfdb536e010bbe/libcontainer/configs/cgroup_unsupported.go#L9) and [namespaces](https://github.com/opencontainers/runc/blob/da548c1c1b0c43cd8718636e8acfdb536e010bbe/libcontainer/configs/namespaces_unsupported.go#L8). We don't need anything from the structure right? It's just because libcontainer/configs is imported at all that's halting compilation

egernst: Yeah. Theoretically the other fields should be fine, but since I'm not directly consuming the struct I'm fine with this as well.

dcantah (approved):
LGTM

libcontainer/configs/mount_unsupported.go
rata: I really like this, given that we can't just embed it

rata (approved):
LGTM, thanks!

crosbymichael (approved):

AkihiroSuda (approved):

PR #3719, opened 2023-02-02
Disable clang-format

kolyshkin:
For the sake of developers who have LSP configured to auto-format the code upon save (that would me with my new nvim setup), let's not auto-format the C code when using clangd.

Initially I tried to write a set of rules for clang-format which is identical to what we use (indent with a handful of options invoked from cfmt target in Makefile), but it appears to be impossible.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,8 @@
+---
+# We use GNU indent from the Makefile to format C code in this project. Alas,
+# there is no way to map indent options to clang-format style options in a way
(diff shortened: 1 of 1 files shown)


austinvazquez (approved):

AkihiroSuda (approved):

kolyshkin:
@thaJeztah PTAL (low hanging fruit)

thaJeztah (approved):
LGTM, thanks!

PR #3707, opened 2023-01-20
libcontainer: skip chown of /dev/null caused by fd redirection

Dzejrou:
In #3355 the check whether the STDIO file descriptors point to /dev/null was removed which can cause /dev/null to change ownership e.g. when using docker exec on a running container:

$ ls -l /dev/null
crw-rw-rw- 1 root root 1, 3 Aug 1 14:12 /dev/null
$ docker exec -u test 0ad6d3064e9d ls
$ ls -l /dev/null
crw-rw-rw- 1 test root 1, 3 Aug 1 14:12 /dev/null

This PR reintroduces that check and fixes the issue for me on runc v1.1.{3,4} (I did not find any information about this change being intentional, if it was feel free to close this PR).

Fixes: #3674

PR #3720, opened 2023-02-02
[DNM] integration test for #3707

kolyshkin:
This is an integration test for the fix in #3707

### DIFF
--- a/tests/integration/exec.bats
+++ b/tests/integration/exec.bats
@@ -125,8 +125,23 @@ function teardown() {
 
 	runc exec --user 1000:1000 test_busybox id
 	[ "$status" -eq 0 ]
-
 	[[ "${output}" == "uid=1000 gid=1000"* ]]
+}
+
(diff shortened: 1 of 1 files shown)


kolyshkin:
I will now retry the same test with the fix removed

kolyshkin:
OK, I have tested that the added case fails without the fix (#3707) and passes with it, as it should.

```
not ok 69 runc exec --user vs /dev/null ownership
# (in test file tests/integration/exec.bats, line 144)
#   `[ "$UG" = "0:0" ]' failed
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-18871/runc.Cuk9gE/tty/sock test_busybox (status=0):
# 
# crw-rw-rw- 1 root root 1, 3 Feb  3 00:00 /dev/null
# uid=1000 gid=1000
# crw-rw-rw- 1 runneradmin root 1, 3 Feb  3 00:00 /dev/null
```

PR #3721, opened 2023-02-03
[1.1] nsexec: Check for errors in write_log()

rata:
This is a backport of #3712 for release 1.1 branch.

---

First, check if strdup() fails and error out.

While we are there, the else case was missing brackets, as we only need to check ret in the else case. Fix that too

Signed-off-by: Rodrigo Campos <rodrigoca@microsoft.com>
(cherry picked from commit 5ce511d6a65809be3fc58f8e2df585abb9c616d6)

PR #3701, opened 2023-01-12
Explicitly pin busybox and debian downloads

tianon:
Closes https://github.com/opencontainers/runc/issues/3699
Alternative to #3700

This pins to an _older_ `debian:buster-slim`, back before it moved to LTS and lost many of the supported architectures.

<details>
<summary>Gory details of how I generated these before munging by hand:</summary>

```console
$ bashbrew cat --format '{{ range .TagEntry.Architectures }}# {{ $.TagEntry.ArchGitRepo . | trimSuffixes ".git" }}/tree/{{ $.TagEntry.ArchGitFetch . | trimPrefixes "refs/heads/" }}{{ "\n" }}{{ . }}) commit={{ $.TagEntry.ArchGitCommit . }} ;;{{ "\n" }}{{ end }}' 'https://github.com/docker-library/official-images/raw/ce10f6b60289c0c0b5de6f785528b8725f225a58/library/debian:buster-slim'
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-amd64
amd64) commit=686d9f6eaada08a754bc7abf6f6184c65c5b378f ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm32v5
arm32v5) commit=155640b6e2e249dfaeee8795d5de539ef3e49417 ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm32v7
arm32v7) commit=60ff0c2c6ce9556e5d8a2758dd2b3f3731716a6f ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-arm64v8
arm64v8) commit=2f108af35e22064c848b8628a7cac56192246dba ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-i386
i386) commit=e4db8aa97f4366e6f27ddbdeaed0773fe0288d47 ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-mips64le
mips64le) commit=e28cbd76dcfba10446b1722aebb5a996121e3d27 ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-ppc64le
ppc64le) commit=3ba08903ca3fd48fe59ba92b02744a2f5d4d9d6f ;;
# https://github.com/debuerreotype/docker-debian-artifacts/tree/dist-s390x
s390x) commit=2fddbf8fe632fc5865b140341b68a1358586fff2 ;;

[...]

PR #3700, opened 2023-01-11
tests/integration/get-images.sh: fix busybox.tar.xz URL

AkihiroSuda:
Fix #3699

PR #3723, opened 2023-02-06
[1.1] Fix CI

rata:
When I opened PR #3721 the CI fails due to downloading the busybox image. This PR should fix that.

I'm backporting #3700 #3701 and #3526 for this. The fist two are fixes for this (and is simpler to backport both than to bakport one and fix the conflicts). The last one (#3526) I've backported it too, so I can apply the first two without conflicts. It only changes tests, so it seems safe to backport.

I think backporting all of these makes it simpler also to backport any more change needed in the future (as these files are in sync between 1.1 branch and main), and as they only change the tests it seems fine IMHO. Let me know if you prefer more chirurgical changes and to deal with conflicts when backporting.


thaJeztah (dismissed):
LGTM

(assuming the CI failure is related; looks like a hiccup, so needs to be triggered again later)

rata:
I'm not sure if the vagrant failure is valid (but not caused by this). Seems weird it doesn't happen on main, though. Will repush

rata:
@kolyshkin @thaJeztah Do you want to carry forward this PR and fix vagrant (vagrant failure seems legit, the other fails seems like azure downtime or some more fixes needed :-/)

kolyshkin:
Re: vagrant, it seems that the F35 image was removed, we need to switch to a newer version. This means backporting #3442 and maybe #3670.

kolyshkin:
I have added the commits from here to https://github.com/opencontainers/runc/pull/3729, so closing this one. Thanks @rata!

PR #3724, opened 2023-02-06
nsexec: Remove bogus kill to stage_2_pid

rata:
stage_2_pid is not yet assigned, so this kills the PID -1 (that has a special meaning). Just remove these calls to kill stage_2_pid before it is cloned/assigned.

I've checked by executing the error paths that no binary is left by mistake.

Signed-off-by: Rodrigo Campos <rodrigoca@microsoft.com>


kolyshkin (dismissed):
There was a recent PR #3501 that touched this code; in particular, sane_kill is used which ignores `-1`, so there is no issue to fix.

OTOH indeed the stage2 process is being spawned after dealing with mount sources, so there is little sense to have this code.

LGTM

kolyshkin:
@cyphar PTAL

libcontainer/nsenter/nsexec.c
```diff
@@ -1231,23 +1231,17 @@ void nsexec(void)
 			/* Ask our parent to send the mount sources fds. */
 			if (config.mountsources) {
 				s = SYNC_MOUNTSOURCES_PLS;
-				if (write(syncfd, &s, sizeof(s)) != sizeof(s)) {
-					sane_kill(stage2_pid, SIGKILL);
```
fuweid, line 1235: https://github.com/opencontainers/runc/blob/99968fc0874642300f688ce4d617a1e42ca7ce29/libcontainer/nsenter/nsexec.c#L1034-L1038

Do we need to remove 👆 line? It is unlikely, but that kill the wrong pid is bad.

kolyshkin: I have also noticed this one while looking into it. Obviously, we can't trust the `stage2_pid` being correct in this case.

Yes, we do (as a separate commit I guess).

rata: @fuweid wanna do that PR? :)

fuweid: Just point it out. 😂
I think you can handle it in this pr. 😄

kolyshkin:
This needs a rebase to fix CI

libcontainer/nsenter/nsexec.c
rata: @fuweid sure, added. Thanks!

rata:
Ok, now CI is failing in some other library when the stdout pipe breaks in centos-9:

```
=== RUN   TestPodSkipDevicesUpdate
    systemd_test.go:114: container stderr not empty: error writing "stdout": broken pipe
            while executing
        "puts stdout {test 0 = 1;}"
            (procedure "renderFalse" line 17)
            invoked from within
        "renderFalse"
            invoked from within
        "if {[catch {
           # parse all command-line arguments before doing any action, no output is
           # made during argument parse to wait for potential paging ..."
            (file "/usr/share/Modules/libexec/modulecmd.tcl" line 14687)
--- FAIL: TestPodSkipDevicesUpdate (0.13s)
```

I'll re-push, but I guess @kolyshkin might be interested to know this happened? :)

rata:
Created an issue for the flaky test

rata:
All green now!

fuweid (approved):

AkihiroSuda (approved):

rata:
Friendly ping?

kolyshkin (approved):
LGTM

Issue #3727, opened 2023-02-07
Runc process crashes when parsing an invalid JSON specification file


AITsygunka:
During the fuzzing process of function ``func loadSpec(cPath string) (spec *specs.Spec, err error)`` was detected a bug. Runc process crushes when parsing invalid ``config.json`` specification file (example [config.json.txt](https://github.com/opencontainers/runc/files/10672389/config.json.txt)):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x55a5f5207b55]
 
goroutine 1 [running]:
main.loadSpec({0x55a5f5217bdd, 0xb})
    github.com/opencontainers/runc/spec.go:129 +0x2b5
main.setupSpec(0xc0000d8580?)
    github.com/opencontainers/runc/utils.go:78 +0x4c
main.startContainer(0xc0000d8580, 0x2, 0x0)
    github.com/opencontainers/runc/utils_linux.go:343 +0x5b
main.glob..func13(0xc0000d8580?)
    github.com/opencontainers/runc/run.go:72 +0x3f
github.com/urfave/cli.HandleAction({0x55a5f5319460?, 0x55a5f53b0c40?}, 0x3?)
    github.com/urfave/cli@v1.22.9/app.go:524 +0x50
github.com/urfave/cli.Command.Run({{0x55a5f5214372, 0x3}, {0x0, 0x0}, {0x0, 0x0, 0x0}, {0x55a5f52203a7, 0x1a}, {0x0, ...}, ...}, ...)
    github.com/urfave/cli@v1.22.9/command.go:173 +0x65b
github.com/urfave/cli.(*App).Run(0xc0000fc380, {0xc0000c4000, 0x7, 0x7})
    github.com/urfave/cli@v1.22.9/app.go:277 +0x8a7
main.main()
    github.com/opencontainers/runc/main.go:165 +0x1208
``` 
The reason is that call ``json.NewDecoder(cf).Decode(&spec)`` https://github.com/opencontainers/runc/blob/32d741358f8fc843b267608b2b684711a8dcaa2a/spec.go#L126 for some cases doesn't return error and returns null pointer to 'spec'.

PR #3728, opened 2023-02-07
Fix runc crushes when parsing invalid JSON specification file

AITsygunka:
Fixes [#3727](https://github.com/opencontainers/runc/issues/3727#issue-1573797885)

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 
@@ -125,6 +126,9 @@ func loadSpec(cPath string) (spec *specs.Spec, err error) {
 
 	if err = json.NewDecoder(cf).Decode(&spec); err != nil {
 		return nil, err
+	}
+	if spec == nil {
+		return nil, errors.New("config cannot be null")
 	}
 	return spec, validateProcessSpec(spec.Process)
 }


spec.go
```diff
@@ -123,8 +123,12 @@ func loadSpec(cPath string) (spec *specs.Spec, err error) {
...
 
-	if err = json.NewDecoder(cf).Decode(&spec); err != nil {
-		return nil, err
+	if err = json.NewDecoder(cf).Decode(&spec); err != nil || spec == nil {
+		if err == nil {
+			return nil, fmt.Errorf("Invalid JSON specification file %s format", cPath)
+		} else {
+			return nil, err
+		}
 	}
```
cyphar, line 132:
```suggestion
	if err = json.NewDecoder(cf).Decode(&spec); err != nil || spec == nil {
		return nil, err
	} else if spec == nil {
		return nil, fmt.Errorf("config cannot be null")
	}
```

It should be noted the issue here is not with invalid JSON (that would return an error from `json.Decode`), but instead with a `config.json` that is set to `null` (which is valid JSON). I feel like another option would be to have `spec` be `specs.Spec` which would stop it from being able to be `nil`.

cyphar:

###### **NOTE**: This is a saved reply. Sorry if it reads as a cookie-cutter response, it was written so that newcomers understand what a "DCO" is and make the process for contributing a little less scary. ######

spec.go
```diff
@@ -125,6 +125,8 @@ func loadSpec(cPath string) (spec *specs.Spec, err error) {
 
 	if err = json.NewDecoder(cf).Decode(&spec); err != nil {
 		return nil, err
+	} else if spec == nil {
```
kolyshkin, line 128: `else` is redundant here because of the return statement above it.

spec.go
```diff
@@ -125,6 +125,8 @@ func loadSpec(cPath string) (spec *specs.Spec, err error) {
...
+		return nil, fmt.Errorf("config cannot be null")
```
kolyshkin, line 129: Nit: hen there are no %-style arguments, `errors.New` is better.

kolyshkin (changes requested):

kolyshkin (dismissed):
LGTM

kolyshkin:
CI failure is unrelated; it was recently reported as #3742

kolyshkin:
@AITsygunka please rebase to fix CI

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #3735, opened 2023-02-10
libct/int: make TestFdLeaks more robust

kolyshkin:
The purpose of this test is to check that there are no extra file
descriptors left open after repeated calls to runContainer. In fact,
the first call to runContainer leaves a few file descriptors opened,
and this is by design.
    
Previously, this test relied on two things:
    1. some other tests were run before it (and thus all such opened-once
       file descriptors are already opened);
    2.  explicitly excluding fd opened to /sys/fs/cgroup.
    
Now, if we run this test separately, it will fail (because of 1 above).
The same may happen if the tests are run in a random order.
    
To fix this, add a container run before collection the initial fd list,
so those fds that are opened once are included and won't be reported.

(plus some minor refactoring commits)


kolyshkin:
Now, this should also hopefully fix the flakiness observed in https://github.com/opencontainers/runc/pull/3721#issuecomment-1424117142

austinvazquez (approved):
Looks good!

AkihiroSuda (approved):

mrunalp (approved):

PR #3739, opened 2023-02-10
specconv: avoid mapping "acl" to MS_POSIXACL

### LINKED ISSUE #3738: mount: `acl` implementation is probably wrong
https://github.com/opencontainers/runc/blob/32d741358f8fc843b267608b2b684711a8dcaa2a/libcontainer/specconv/spec_linux.go#L69

Currently, runc maps the `acl` mount option strings to `MS_POSIXACL`, but this option doesn't seem used by `util-linux`: https://github.com/util-linux/util-linux/search?q=MS_POSIXACL

Surprisingly, `MS_POSIXACL` isn't even referred by the kernel,  except in `security/apparmor/mount.c:audit_mnt_flags()`: https://elixir.bootlin.com/linux/v6.1.11/C/ident/MS_POSIXACL

Probably the `acl` string should be just passed to `mount(2)` via `void *data`.

AkihiroSuda:
From https://github.com/torvalds/linux/commit/caaef1ba8c9ee7a54b53dd8bf4bb7e8658185583 :

> In fact SB_POSIXACL is an internal flag, and accepting MS_POSIXACL on
> the mount(2) interface is possibly a bug.

Fix #3738

### DIFF
--- a/libcontainer/specconv/spec_linux.go
+++ b/libcontainer/specconv/spec_linux.go
@@ -66,7 +66,7 @@ func initMaps() {
 			clear bool
 			flag  int
 		}{
-			"acl":           {false, unix.MS_POSIXACL},
+			// "acl" cannot be mapped to MS_POSIXACL: https://github.com/opencontainers/runc/issues/3738
 			"async":         {true, unix.MS_SYNCHRONOUS},
 			"atime":         {true, unix.MS_NOATIME},
 			"bind":          {false, unix.MS_BIND},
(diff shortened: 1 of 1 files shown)


kolyshkin (approved):
LGTM.

Maybe it makes sense to add a comment saying that we do not map "acl/noacl" to `unix.MS_POSIXACL` as Linux kernel does not support it (or just link to #3738). This way, the flag/option won't be re-added by someone.

AkihiroSuda:
> Maybe it makes sense to add a comment saying that we do not map "acl/noacl" to `unix.MS_POSIXACL` as Linux kernel does not support it (or just link to #3738). This way, the flag/option won't be re-added by someone.

done

kolyshkin (approved):
LGTM

kolyshkin:

PR #3745, opened 2023-02-14
libct/cg: IsCgroup2UnifiedMode: don't panic

shoenig:
This the same as https://github.com/opencontainers/runc/commit/8290c4cf58fbc82b6a940dc1a091a6e2f3cbc919 but for the `IsCgroup2UnifiedMode` function.

Related [bug report](https://github.com/hashicorp/nomad/issues/16179) from real world (Nomad)

plagiarized description below

> In case statfs("/sys/fs/cgroup/unified") fails with any error other than ENOENT, current code panics. As IsCgroup2UnifiedMode is called from libcontainer/cgroups/fs's init function, this means that any user of libcontainer may panic during initialization, which is ugly.
> 
> Avoid panicking; instead, do not enable unified hierarchy support and report the error (under debug level, not to confuse anyone).
> 
> Basically, replace the panic with "turn off unified mode support" (which makes total sense since we were unable to statfs its root).

### DIFF
--- a/libcontainer/cgroups/utils.go
+++ b/libcontainer/cgroups/utils.go
@@ -36,14 +36,15 @@ func IsCgroup2UnifiedMode() bool {
 		var st unix.Statfs_t
 		err := unix.Statfs(unifiedMountpoint, &st)
 		if err != nil {
-			if os.IsNotExist(err) && userns.RunningInUserNS() {
+			isErrNotExist := errors.Is(err, os.ErrNotExist)
+			if isErrNotExist && userns.RunningInUserNS() {
 				// ignore the "not found" error if running in userns
 				logrus.WithError(err).Debugf("%s missing, assuming cgroup v1", unifiedMountpoint)
 				isUnified = false
 				return
 			}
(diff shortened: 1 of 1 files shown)


kolyshkin:
The code added by this PR is over-complicated.

> This the same as 8290c4c

Not really. The above commit is slightly different, because we do not always expect `/sys/fs/cgroup/unified` to be present (i.e. having no such directory is totally fine).

In this case, we always expect `/sys/fs/cgroup` to be present, and any error (even `ENOENT`) here is unexpected, including ENOENT. Currently, we make an exception when we're in a userns AND the error is ENOENT.

If you need the code to not panic at all, just do something like

```go
if err != nil {
     logrus.Warnf("statfs(%s) failed: %w; assuming cgroup v1", path, err)
     return
}
```

This is still ugly (since the rest of the code assumes that either cgroup v1 or v2 is available), but slightly less ugly when panicking.

kolyshkin:
@shoenig you still want to work on that?

PR #3749, opened 2023-02-23
[Carry #3205] libct/cg: add CFS bandwidth burst for CPU

Zheaoli:
Carry #3205

### DIFF
--- a/libcontainer/cgroups/fs/cpu.go
+++ b/libcontainer/cgroups/fs/cpu.go
@@ -84,6 +84,28 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 			period = ""
 		}
 	}
+
+	var burst string
+	if r.CpuBurst != nil {
+		burst = strconv.FormatUint(*r.CpuBurst, 10)
+		if err := cgroups.WriteFile(path, "cpu.cfs_burst_us", burst); err != nil {
+			// this is a special trick for burst feature, the current systemd and low version of kernel will not support it.
+			// So, an `no such file or directory` error would be raised, and we can ignore it .
+			if !errors.Is(err, unix.ENOENT) {
+				// Sometimes when the burst to be set is larger
+				// than the current one, it is rejected by the kernel
+				// (EINVAL) as old_quota/new_burst exceeds the parent
+				// cgroup quota limit. If this happens and the quota is
+				// going to be set, ignore the error for now and retry
+				// after setting the quota.
+				if !errors.Is(err, unix.EINVAL) || r.CpuQuota == 0 {
+					return err
+				}
+			}
+		} else {
+			burst = ""
+		}
+	}
 	if r.CpuQuota != 0 {
 		if err := cgroups.WriteFile(path, "cpu.cfs_quota_us", strconv.FormatInt(r.CpuQuota, 10)); err != nil {
 			return err
@@ -91,6 +113,13 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 		if period != "" {
 			if err := cgroups.WriteFile(path, "cpu.cfs_period_us", period); err != nil {
 				return err
+			}
+		}
+		if burst != "" {
+			if err := cgroups.WriteFile(path, "cpu.cfs_burst_us", burst); err != nil {
+				if !errors.Is(err, unix.ENOENT) {
+					return err
+				}
 			}
 		}
 	}
--- a/libcontainer/cgroups/fs2/cpu.go
+++ b/libcontainer/cgroups/fs2/cpu.go
@@ -2,8 +2,11 @@ package fs2
 
 import (
 	"bufio"
+	"errors"
 	"os"
 	"strconv"
+
+	"golang.org/x/sys/unix"
 
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runc/libcontainer/cgroups/fscommon"
@@ -11,7 +14,7 @@ import (
 )
 
 func isCpuSet(r *configs.Resources) bool {
-	return r.CpuWeight != 0 || r.CpuQuota != 0 || r.CpuPeriod != 0 || r.CPUIdle != nil
+	return r.CpuWeight != 0 || r.CpuQuota != 0 || r.CpuPeriod != 0 || r.CPUIdle != nil || r.CpuBurst != nil
 }
 
 func setCpu(dirPath string, r *configs.Resources) error {
@@ -32,6 +35,23 @@ func setCpu(dirPath string, r *configs.Resources) error {
 		}
 	}
 
+	var burst string
+	if r.CpuBurst != nil {
+		burst = strconv.FormatUint(*r.CpuBurst, 10)
+		if err := cgroups.WriteFile(dirPath, "cpu.max.burst", burst); err != nil {
+			// Sometimes when the burst to be set is larger
+			// than the current one, it is rejected by the kernel
+			// (EINVAL) as old_quota/new_burst exceeds the parent
+			// cgroup quota limit. If this happens and the quota is
+			// going to be set, ignore the error for now and retry
+			// after setting the quota.
+			if !errors.Is(err, unix.EINVAL) || r.CpuQuota == 0 {
+				return err
+			}
+		} else {
+			burst = ""
+		}
+	}
 	if r.CpuQuota != 0 || r.CpuPeriod != 0 {
 		str := "max"
 		if r.CpuQuota > 0 {
@@ -46,6 +66,11 @@ func setCpu(dirPath string, r *configs.Resources) error {
 		str += " " + strconv.FormatUint(period, 10)
 		if err := cgroups.WriteFile(dirPath, "cpu.max", str); err != nil {
 			return err
+		}
+		if burst != "" {
+			if err := cgroups.WriteFile(dirPath, "cpu.max.burst", burst); err != nil {
+				return err
+			}
 		}
 	}
 
--- a/libcontainer/configs/cgroup_linux.go
+++ b/libcontainer/configs/cgroup_linux.go
@@ -69,6 +69,9 @@ type Resources struct {
 	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
 	CpuQuota int64 `json:"cpu_quota"`
 
+	// CPU hardcap burst limit (in usecs). Allowed accumulated cpu time additionally for burst in a given period.
+	CpuBurst *uint64 `json:"cpu_burst"` //nolint:revive
+
 	// CPU period to be used for hardcapping (in usecs). 0 to use system default.
 	CpuPeriod uint64 `json:"cpu_period"`
 
--- a/update.go
+++ b/update.go
@@ -44,6 +44,7 @@ The accepted format is as follow (unchanged values can be omitted):
   "cpu": {
     "shares": 0,
     "quota": 0,
+    "burst": 0,
     "period": 0,
     "realtimeRuntime": 0,
     "realtimePeriod": 0,
@@ -72,6 +73,10 @@ other options are ignored.
 		cli.StringFlag{
 			Name:  "cpu-quota",
 			Usage: "CPU CFS hardcap limit (in usecs). Allowed cpu time in a given period",
+		},
+		cli.StringFlag{
+			Name:  "cpu-burst",
+			Usage: "CPU CFS hardcap burst limit (in usecs). Allowed accumulated cpu time additionally for burst a given period",
 		},
 		cli.StringFlag{
 			Name:  "cpu-share",
@@ -153,6 +158,7 @@ other options are ignored.
 			CPU: &specs.LinuxCPU{
 				Shares:          u64Ptr(0),
 				Quota:           i64Ptr(0),
+				Burst:           u64Ptr(0),
 				Period:          u64Ptr(0),
 				RealtimeRuntime: i64Ptr(0),
 				RealtimePeriod:  u64Ptr(0),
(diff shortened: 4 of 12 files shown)


Zheaoli:
Seems the systemd is not support the CPU burst feature yet. https://github.com/systemd/systemd/issues/26658

So should we keep pushing this PR or hold on it utill the feature has been supported in systemd @AkihiroSuda

AkihiroSuda:
> Seems the systemd is not support the CPU burst feature yet. [systemd/systemd#26658](https://github.com/systemd/systemd/issues/26658)
> 
> So should we keep pushing this PR or hold on it utill the feature has been supported in systemd @AkihiroSuda

You can just let fs2 handle it
https://github.com/opencontainers/runc/blob/a187c84e42015402263167a2c1007b5c62d494f3/libcontainer/cgroups/systemd/v2.go#L168-L170

Zheaoli:
> > Seems the systemd is not support the CPU burst feature yet. [systemd/systemd#26658](https://github.com/systemd/systemd/issues/26658)
> > So should we keep pushing this PR or hold on it utill the feature has been supported in systemd @AkihiroSuda
> 
> You can just let fs2 handle it
> 
> https://github.com/opencontainers/runc/blob/a187c84e42015402263167a2c1007b5c62d494f3/libcontainer/cgroups/systemd/v2.go#L168-L170

Got it

Zheaoli:
> > Seems the systemd is not support the CPU burst feature yet. [systemd/systemd#26658](https://github.com/systemd/systemd/issues/26658)
> > So should we keep pushing this PR or hold on it utill the feature has been supported in systemd @AkihiroSuda
> 
> You can just let fs2 handle it
> 
> https://github.com/opencontainers/runc/blob/a187c84e42015402263167a2c1007b5c62d494f3/libcontainer/cgroups/systemd/v2.go#L168-L170

But the burst feature supports both cv1 and cv2. hmmm I think maybe I can limit it should be only being used in non-systemd driver?

Zheaoli:
@AkihiroSuda Hi, I have written the test, but the rootless test failed. I have no idea what happened. would you mind take a look at for me ?

Zheaoli:
https://github.com/opencontainers/runc/actions/runs/4462135582/jobs/7836459757?pr=3749
https://github.com/opencontainers/runc/actions/runs/4462008272/jobs/7836249615?pr=3749

It seems strange why `no directory specified for cpu.cfs_burst_us` has been raised but not for other values like period and quota(both of them shared the same path

Zheaoli:
fine, I figure out the reason, all ci has been passed

Zheaoli:
https://cirrus-ci.com/task/5913443228712960
https://cirrus-ci.com/task/4887318499164160
https://cirrus-ci.com/task/4605843522453504

Should I skip the systemd test for burst? cc @AkihiroSuda

Zheaoli:
@AkihiroSuda ping(would you mind to give me some advise about test failed in Centos?

AkihiroSuda:
CI failing

Zheaoli:
@AkihiroSuda It seems like there may be some compatibility issues between CPU Burst and Centos7 and Centos8, which could be causing the CI to fail.  Would your mind to give me some suggestion about how to pass all the CI?

AkihiroSuda:
> It seems like there may be some compatibility issues between CPU Burst and Centos7 and Centos8, which could be causing the CI to fail. Would your mind to give me some suggestion about how to pass all the CI?

I guess you can just check existence of `cpu.cfs_burst_us` and skip the test when it does not exist?

Zheaoli:
> > It seems like there may be some compatibility issues between CPU Burst and Centos7 and Centos8, which could be causing the CI to fail. Would your mind to give me some suggestion about how to pass all the CI?
> 
> I guess you can just check existence of `cpu.cfs_burst_us` and skip the test when it does not exist?

Seems good, I'll make a try! Thanks a lot!

Zheaoli:
@AkihiroSuda Seems all CI has been passed except lint-extra test.

Some warning message has been raised from this step:" var-naming: struct field CpuBurst should be CPUBurst". But I think `CpuBurst` is better than `CPUBurst` because we should keep consistency with other fields?

libcontainer/configs/cgroup_linux.go
```diff
@@ -69,6 +69,9 @@ type Resources struct {
 	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
 	CpuQuota int64 `json:"cpu_quota"`
 
+	// CPU hardcap burst limit (in usecs). Allowed accumulated cpu time additionally for burst in a given period.
+	CpuBurst uint64 `json:"cpu_burst"`
```
AkihiroSuda, line 73:
```suggestion
	CpuBurst uint64 `json:"cpu_burst"` //nolint:revive
```

Zheaoli: emmm, it's weird; it seems not need to add an extra flag to other filed. = =

AkihiroSuda (approved):

Zheaoli:
@kolyshkin Hello. Would you mind looking at this PR if you have time? Thanks!

AkihiroSuda:
Fedora CI is failing, rebasing may solve the issue.

AkihiroSuda:
Please update https://github.com/opencontainers/runc/blob/main/docs/spec-conformance.md too

libcontainer/configs/cgroup_linux.go
kolyshkin: This is because lint-extra is a job which runs more strict linters on new code only. But yes, I agree that if other fields are named `Cpu*`, we should use `Cpu`, too.

Zheaoli:
@AkihiroSuda Thanks for your tips, and sorry for replying late. 

I have merged the newest code from the main branch and updated the docs. But there is still some issues:

1. the CI on Centos-Stream-9 has failed. The message is `Command "grep -q oom events.log" failed 30 times. Output:`. I think this issue is not relevant to this PR. Would you mind rerunning this step for me?
2. The spell check has failed. I'm sure if I can skip the check or not.

AkihiroSuda (approved):

Zheaoli:
@kolyshkin ping~

lifubang:
@Zheaoli I think you should use `git rebase main` instead of `git merge main`.

Zheaoli:
> @Zheaoli I think you should use `git rebase main` instead of `git merge main`.

copy that, I will squash commit later

libcontainer/cgroups/fs/cpu.go
```diff
@@ -84,6 +84,28 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
...
+		if err := cgroups.WriteFile(path, "cpu.cfs_burst_us", burst); err != nil {
+			// Sometimes when the burst to be set is larger
+			// than the current one, it is rejected by the kernel
+			// (EINVAL) as old_quota/new_burst exceeds the parent
+			// cgroup quota limit. If this happens and the quota is
+			// going to be set, ignore the error for now and retry
+			// after setting the quota.
+			if !errors.Is(err, unix.ENOENT) {
+				// this is a special trick for burst feature, the current systemd and low version of kernel will not support it.
+				// So, an `no such file or directory` error would be raised, and we can ignore it .
```
lifubang, line 100: I think these two lines comments is for `unix.ENOENT`, you should put the comments in the correct place, so please exchange the comments of `unix.ENOENT` and `unix.EINVAL`.

libcontainer/cgroups/fs/cpu.go
```diff
@@ -93,6 +115,11 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 				return err
 			}
 		}
+		if burst != "" {
+			if err := cgroups.WriteFile(path, "cpu.cfs_burst_us", burst); err != nil {
+				return err
```
lifubang, line 120: Should we ignore the error of `unix.ENOENT` here too?

libcontainer/cgroups/fs2/cpu.go
```diff
@@ -32,6 +35,23 @@ func setCpu(dirPath string, r *configs.Resources) error {
...
+	if r.CpuBurst != 0 {
+		burst = strconv.FormatUint(r.CpuBurst, 10)
+		if err := cgroups.WriteFile(dirPath, "cpu.max.burst", burst); err != nil {
+			// Sometimes when the burst to be set is larger
+			// than the current one, it is rejected by the kernel
+			// (EINVAL) as old_quota/new_burst exceeds the parent
+			// cgroup quota limit. If this happens and the quota is
+			// going to be set, ignore the error for now and retry
+			// after setting the quota.
+			if !errors.Is(err, unix.EINVAL) || r.CpuQuota == 0 {
```
lifubang, line 48: I don't know whether there is a case of `unix.ENOENT` in `cgroupv2` or not? I think there is a high probability that there is no this type case. But maybe I'm wrong.

lifubang (changes requested):

libcontainer/cgroups/fs2/cpu.go
Zheaoli: I think it's good keep the same code style between cgroupv1 and v2

lifubang: So we need add the same logic for  'unix.ENOENT' like v1?

Zheaoli: Yep, I think this would be a good idea

lifubang:
👍 Please use ‘git rebase main’ to catch up with the latest change, we should not have a commit named ‘merge main’. Thanks.

Zheaoli:
@lifubang Thanks for the review, all things done

update.go
```diff
@@ -298,6 +305,12 @@ other options are ignored.
 			}
 		}
 
+		b := *r.CPU.Burst
+		if config.Cgroups.Resources.CpuPeriod == 0 && b != 0 {
```
lifubang, line 309: @Zheaoli Sorry for delayed so much time. Here I understand you mean that `cpu-burst` depends on `cpu-period`? But I think may be we should depends on `cpu-quota`, or maybe we should remove this dependent check? Let it check by the kernel?

Zheaoli: To be honest, I think it would be great to validate some base rule at the entry of the CLI like https://github.com/opencontainers/runc/blob/main/update.go#L318-L325

If we put the validation process in the kernel, I think the error message will confuse people.

lifubang: > I think the error message will confuse people.

But there is no such check when running a container. There would be different results for ‘run’ and ‘update’ if we only provide ’cpu-burst’.

libcontainer/cgroups/fs/cpu.go
```diff
@@ -84,6 +84,28 @@ func (s *CpuGroup) Set(path string, r *configs.Resources) error {
 			period = ""
 		}
 	}
+
+	var burst string
+	if r.CpuBurst != 0 {
```
lifubang, line 89: If we want to clear `cpu-burst`, the value should be 0. So we should change `r.CpuBurst` to a pointer and to check whether this is nil or not.

libcontainer/configs/cgroup_linux.go
```diff
@@ -69,6 +69,9 @@ type Resources struct {
 	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
 	CpuQuota int64 `json:"cpu_quota"`
 
+	// CPU hardcap burst limit (in usecs). Allowed accumulated cpu time additionally for burst in a given period.
+	CpuBurst uint64 `json:"cpu_burst"` //nolint:revive
```
lifubang, line 73: Please change to `CpuBurst *unit64`

lifubang (changes requested):

lifubang:
I'm so sorry to let you spend more time on this. We need to cover as many test cases as possible. ☕

libcontainer/configs/cgroup_linux.go
Zheaoli: I think the filed type should be as same as other fields like `CpuQuota`

Zheaoli:
> We need to cover as many test cases as possible. ☕

I think we already have enough test cases here. May I ask what's kind of the corner case in your thought not covered by current test code?

libcontainer/configs/cgroup_linux.go
lifubang: No, For `CpuQuota`, `-` means `no limit`, but for `CpuBurst`, `0` means `no config`.

libcontainer/cgroups/fs/cpu.go
Zheaoli: I think splitting the nil check and zero-value check process is unnecessary here.

In my opinion, the nil value(the user doesn't input the burst value) and the zero-value(the user input burst as zero) means the same thing for us which is we should not write the burst value to the cgroup file

lifubang:
> May I ask what's kind of the corner case in your thought not covered by current test code?

For example, how to clear `cpu-burst` value in `update` after the container started with a `cpu-burst` value?

libcontainer/cgroups/fs/cpu.go
lifubang: For example, the container `test` started with the config:
```
cat config.json | jq .linux.resources.cpu
{
  "burst": 1000
}
```

After some time, we want to clear burst value, we may run:
```
runc update --cpu-burst 0 test
```

libcontainer/configs/cgroup_linux.go
lifubang: I think `0` is a legal value which should be allowed to write to cgroup.

Zheaoli: SGTM

lifubang (approved):
LGTM

lifubang:
Left some further works:
1. No unit test for cpu controller in cgroupv2;
2. There are some different results for cgroup v1 and v2, for example:
```
runc update --cpu-burst 30001 --cpu-quota 30000 test
```
v1: fail, v2: succ
Because for v2, the kernel write seconds to `cpu.max.burst`, but not milliseconds.
3. There is no integration test for `cpu-burst` when create/run a container.

We can do these works in further PRs.

lifubang:
@AkihiroSuda PTAL. I think this could be merged.

AkihiroSuda (approved):

PR #3753, opened 2023-02-28
Fix runc run "permission denied" when rootless

### LINKED ISSUE #3715: Runc 1.1.4 container fails with `permission denied: unknown` when relying on capabilities
After upgrading from runc 1.1.3 to runc 1.1.4 we started to see a particular container fail with an error like:
```
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec /entrypoint.sh: permission denied: unknown
```

The same container runs successfully with runc 1.1.3. 

Upon inspection of the container, we discoverd that the entrypoint's permissions were set to 744, but it was owned by root instead of the container's user. We were able to fix the issue by correctly `chown`ing the entrypoint script, but the upgrade caused an outage, so I'm reporting it here as a bug.

__Root Cause__
I traced the cause down to this PR https://github.com/opencontainers/runc/pull/3522

In that PR, a new check was introduced to exit early if the entrypoint isn't executable. It does so by
1) Attempting to use `faccessat2(2)` to check the effective permissions if available (requires kernel 5.8+ and libseccomp 2.4.0+)
2) Falling back to `access(2)` to check the real users permission if `faccessat2(2)` isn't available.

We were hitting case 2. The problem is that this check is more restrictive than `execve(2)` because `access(2)` removes capabilities when checking for permission. 

Docker and containerd have a set of default capabilities that they add to the runtime spec. Both include `CAP_DAC_OVERRIDE` in this set. See [containerd's list](https://github.com/containerd/containerd/blob/60443133dab961cc87da83b7837e13f22d1d0475/oci/spec.go#L100), [docker's list](https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/oci/caps/defaults.go#L7).
The capabilities man page says this about `CAP_DAC_OVERRIDE`
> Bypass file read, write, and execute permission checks.  (DAC is an abbreviation of "discretionary access control".)

So in runc 1.1.3, `execve` worked because runc has the `CAP_DAC_OVERRIDE` capability to ignore the fact that the user doesn't actually have execute permission. In runc 1.1.4, the addtional `access(2)` check striped capabilities when checking permissions and caused runc to exit with `permission denied: unknown`.

__Conditions for this bug__
1) kernel < 5.8 or libseccomp < 2.4.0
2) An entry point script where the owner has execute permission, but the container user has readonly permission
3) `CAP_DAC_OVERRIDE` or similar capabilities set (which is the default for both containerd and docker)

__Reproduction__

entrypoint.sh
```
#!/bin/sh

echo "Hello, World!"
```
Dockerfile
```
FROM alpine:latest

COPY ./entrypoint.sh /entrypoint.sh
RUN  chown root:root /entrypoint.sh\
      && chmod 744 /entrypoint.sh

RUN adduser --disabled-password testuser
USER testuser
ENTRYPOINT ["/entrypoint.sh"]

```

When run with runc 1.1.3:
```
$ sudo nerdctl run --rm -it --net=none runc114error:latest
Hello, World!
```

When run with runc 1.1.4:
```
$ sudo nerdctl run --rm -it --net=none runc114error:latest
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec /entrypoint.sh: permission denied: unknown
```

kolyshkin:
Fixes: #3715

Since PR #3522 (commit 957d97bcf43f41beef96) was made to fix issue #3520,
a few things happened:
    
 - a similar functionality appeared in go 1.20 [1], so the issue
   mentioned in the comment (being removed) is no longer true;
  - a bug in runc was found [2], which also affects go [3];
  - the bug was fixed in go 1.21 [4] and 1.20.2 [5];
  - a similar fix was made to x/sys/unix.Faccessat [6].
    
The essense of the bug #3715 is, even if a (non-root) user that the container is
run as does not have execute permission bit set for the executable, it
should still work in case runc has the CAP_DAC_OVERRIDE capability set.
    
To fix #3715 without reintroducing the older bug [7]:
 - drop own Eaccess implementation;
 - use the one from x/sys/unix for Go 1.19 (depends on [6]);
 - do not use anything when Go 1.20+ is used.
 
Add a test case to make sure we won't regress.
    
**NOTE it is virtually impossible to fix the bug [2] when Go 1.20 or Go
1.20.1 is used because of [3].**
       
[1] https://go-review.googlesource.com/c/go/+/414824
[2] https://github.com/opencontainers/runc/issues/3715
[3] https://go.dev/issue/58552
[4] https://go-review.googlesource.com/c/go/+/468735
[5] https://go-review.googlesource.com/c/go/+/469956
[6] https://go-review.googlesource.com/c/sys/+/468877
[7] https://github.com/opencontainers/runc/issues/3520


AkihiroSuda:
Go 1.20.2 seems released

kolyshkin:
Reworked to not re-introduced #3520. Description updated.

kolyshkin:
@opencontainers/runc-maintainers PTAL (this needs to be merged then backported to 1.1, and since it is a regression in 1.1.4 I want to fix it in time for 1.1.5).

mrunalp (approved):
LGTM

libcontainer/eaccess_stub.go
```diff
@@ -0,0 +1,7 @@
+//go:build go1.20
+
+package libcontainer
+
+func eaccess(path string) error {
```
AkihiroSuda, line 5: Could you add comment lines to explain why this is NOP

tests/integration/start_hello.bats
```diff
@@ -37,6 +37,36 @@ function teardown() {
 	[[ "${output}" == *"Hello"* ]]
 }
 
+# https://github.com/opencontainers/runc/issues/3715, https://go.dev/issue/58552
+@test "runc run as user with no exec bit but CAP_DAC_OVERRIDE set" {
+	requires root # Can't chown/chmod otherwise.
+	case "$(__runc --version | grep '^go:')" in
+	"go: go1.20" | "go: go1.20.1")
+		skip "go 1.20[.1] is broken; see https://go.dev/issue/58552"
```
thaJeztah, line 45: Do we still need to take go 1.20 - 1.20.1 into account here? Or could we just assume; our CI will be on 1.20.2 (and if others don't, then it's their problem?)

.cirrus.yml
```diff
@@ -70,7 +70,7 @@ task:
   env:
     HOME: /root
     CIRRUS_WORKING_DIR: /home/runc
-    GO_VERSION: "1.19"
+    GO_VERSION: "1.20.2"
```
thaJeztah, line 73: Given that this PR has a go1.19 variant, perhaps it would be good to move the go1.20 update separate. 

That way, this PR will be testing the go1.19 branch (which would also be the case if we backport this to 1.1.x (?)). We already had a run of CI with 1.20 on this PR, and once the 1.20 update (separate PR) is merged, `main` will be testing that again.

tests/integration/start_hello.bats
kolyshkin: It took me some time to write this :) so I am hesitant as to removing this.

But you are right -- it is better to assume that Go 1.20 >= 1.20.2 is used, and we should fail otherwise.

I am replacing this one with a mere comment.

.cirrus.yml
kolyshkin: Sure. Done.

kolyshkin:
@AkihiroSuda @thaJeztah addressed your review comments, PTAAL.

AkihiroSuda (approved):

mrunalp (approved):

kolyshkin:
1.1 backport: #3817

PR #3755, opened 2023-03-05
enable stack protection with CGO crosscompile.

xuegege5290:
https://github.com/moby/moby/issues/45045
It appears that docker-tools binaries which import c code are built without Stack Protection.

Some local vulnerability scanners attempt to detect binaries that are compiled without stack cookies (e.g. -fstack-protector vs. -fno-stack-protector with GCC), and flag that as a vulnerability. such as checksec http://www.trapkit.de/tools/checksec.html
What should Go's stance be for this sort of thing?
Do docker-tools will also add stack protection?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,11 @@ GPG_KEYID ?= asarai@suse.de
 # Some targets need cgo, which is disabled by default when cross compiling.
 # Enable cgo explicitly for those.
 # Both runc and libcontainer/integration need libcontainer/nsenter.
-runc static localunittest: export CGO_ENABLED=1
+#enable stack protection with CGO compile.
+runc static localunittest: export CGO_ENABLED=1 && CGO_CFLAGS='-fstack-protector-strong'
 # seccompagent needs libseccomp (when seccomp build tag is set).
 ifneq (,$(filter $(BUILDTAGS),seccomp))
-seccompagent: export CGO_ENABLED=1
+seccompagent: export CGO_ENABLED=1 && export CGO_CFLAGS='-fstack-protector-strong'
 endif
 
 .DEFAULT: runc


xuegege5290:
@evanphx @bernerdschaefer @calavera @proppy

evanphx:
@xuegege5290 For the last time, PLEASE do not tag random people. It only makes people upset at you.

xuegege5290:
@evanphx I feel upset at you!When i try to make a contribution,and wanna someone to check, and give some advice.I use @ ,it show your name and others,but you say such  words???I really feel upset at you .And you should upset at the  recommendation systems.Fine ,i will never @ you again。

Makefile
```diff
@@ -49,10 +49,11 @@ GPG_KEYID ?= asarai@suse.de
 # Some targets need cgo, which is disabled by default when cross compiling.
 # Enable cgo explicitly for those.
 # Both runc and libcontainer/integration need libcontainer/nsenter.
-runc static localunittest: export CGO_ENABLED=1
+#enable stack protection with CGO compile.
+runc static localunittest: export CGO_ENABLED=1 && CGO_CFLAGS='-fstack-protector-strong'
```
kolyshkin, line 53: 1. What if someone sets CGO_CFLAGS? Maybe use `+=` instead?
2. Frankly, I don't understand the use of `&&` here. Note this is GNU make, not shell.

This all this, seems like it should be
```make
export CGO_ENABLED=1 CGO_CFLAGS+='-fstack-protector-strong'
```

Makefile
```diff
@@ -49,10 +49,11 @@ GPG_KEYID ?= asarai@suse.de
...
 # seccompagent needs libseccomp (when seccomp build tag is set).
 ifneq (,$(filter $(BUILDTAGS),seccomp))
-seccompagent: export CGO_ENABLED=1
+seccompagent: export CGO_ENABLED=1 && export CGO_CFLAGS='-fstack-protector-strong'
```
kolyshkin, line 56: seccompagent is a helper binary used for tests only, it is not shipped by any distro, and so I don't think we need to make it more secure.

Makefile
kolyshkin: One other thing, what if a different C backend is used (not gcc) and it could not understand `-fstack-protector-strong`? Just thinking out loud here.

kolyshkin:
Also please see https://github.com/golang/go/issues/21871#issuecomment-329330371

kolyshkin:
> Also please see [golang/go#21871 (comment)](https://github.com/golang/go/issues/21871#issuecomment-329330371)

In view of the above, and taking into account that there's not much C code in this project, I guess we should not proceed with this change and rely on the defaults provided by golang.

PR #3756, opened 2023-03-06
rootfs: prohibit symlinks that conflicts with `readonlyPaths` and/or `maskedPaths` to prevent CVE-2023-27561

### LINKED ISSUE #3751: CVE-2019-19921 re-introduction/regression
Hi,

I'm part of the Debian Long Term Support (LTS) team, and I'm currently working on an update for package [runc](https://security-tracker.debian.org/tracker/source-package/runc).

As explained in https://github.com/opencontainers/runc/issues/2197#issuecomment-1437617334 , while working on fixing CVE-2019-19921, I noticed the fix was apparently broken by the one for CVE-2021-30465 (0ca91f44f1664da834bc61115a849b56d22f595f).

I can reproduce the issue with branch `main`, using the [original reproducer](https://gist.github.com/LiveOverflow/c937820b688922eb127fb760ce06dab9) from #2197 (host's `proc/sys/kernel/core_pattern` overwritten from `container-2` after a few tries).

Because various GNU/Linux distributions have incorporated the initial fix (or upgraded runc), and marked the security issue "fixed", I would recommend registering a new CVE to indicate that a follow-up fix is needed.
Do you confirm?

Thanks and best regards.

AkihiroSuda:
"/proc" and "/sys" in the container rootfs can no longer be symlinks, as they could be tricked to obtain the host root.

Fix #3751 ("CVE-2019-19921 re-introduction/regression")
Fix CVE-2023-27561

The CVE has been already public, so I'm just going to open this PR publicly.

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -52,6 +52,10 @@ func needsSetupDev(config *configs.Config) bool {
 // finalizeRootfs after this function to finish setting up the rootfs.
 func prepareRootfs(pipe io.ReadWriter, iConfig *initConfig, mountFds []int) (err error) {
 	config := iConfig.Config
+	if err := validateRoot(config); err != nil {
+		return fmt.Errorf("error validating rootfs: %w", err)
+	}
+
 	if err := prepareRoot(config); err != nil {
 		return fmt.Errorf("error preparing rootfs: %w", err)
 	}
@@ -1114,4 +1118,38 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
 	return utils.WithProcfd(rootfs, m.Destination, func(procfd string) error {
 		return unix.MountSetattr(-1, procfd, unix.AT_RECURSIVE, m.RecAttr)
 	})
+}
+
+// validateRoot detects symlinks in config.Rootfs that may break config.ReadonlyPaths
+// and config.MaskPaths.
+//
+// For preventing CVE-2023-27561 (CVE-2019-19921 re-introduction/regression).
+// https://github.com/opencontainers/runc/issues/3751
+func validateRoot(config *configs.Config) error {
+	mustNotBeSymlink := make(map[string]struct{})
+	for _, f := range append(config.ReadonlyPaths, config.MaskPaths...) {
+		if !filepath.IsAbs(f) || f != filepath.Clean(f) {
+			return fmt.Errorf("bad path: %q", f)
+		}
+		for f != "/" {
+			mustNotBeSymlink[f] = struct{}{}
+			f = filepath.Dir(f)
+		}
+	}
+	// mustNotBeSymlink typically contains {"/sys", "/proc"} in addition to readonlyPaths and maskedPaths.
+	rootFD, err := unix.Open(config.Rootfs, unix.O_DIRECTORY|unix.O_RDONLY, 0o600)
+	if err != nil {
+		return &os.PathError{Op: "open", Path: config.Rootfs, Err: err}
+	}
+	defer unix.Close(rootFD)
+	for f := range mustNotBeSymlink {
+		// fstatat(2): "If pathname is absolute, then dirfd is ignored."
+		fRel := "./" + f
+		var st unix.Stat_t
+		err := unix.Fstatat(rootFD, fRel, &st, unix.AT_SYMLINK_NOFOLLOW)
+		if err == nil && st.Mode&unix.S_IFMT == unix.S_IFLNK {
+			return fmt.Errorf("must not be a symlink (conflicts with readonlyPaths and/or maskedPaths): %q", f)
+		}
+	}
+	return nil
 }
--- a/tests/integration/mask.bats
+++ b/tests/integration/mask.bats
@@ -6,11 +6,12 @@ function setup() {
 	setup_busybox
 
 	# Create fake rootfs.
-	mkdir rootfs/testdir
+	mkdir rootfs/testdir rootfs/testdir2
 	echo "Forbidden information!" >rootfs/testfile
+	echo "Forbidden information! (in a nested dir)" >rootfs/testdir2/file2
 
 	# add extra masked paths
-	update_config '(.. | select(.maskedPaths? != null)) .maskedPaths += ["/testdir", "/testfile"]'
+	update_config '(.. | select(.maskedPaths? != null)) .maskedPaths += ["/testdir", "/testfile", "/testdir2/testfile2"]'
 }
 
 function teardown() {
@@ -55,4 +56,42 @@ function teardown() {
 	runc exec test_busybox umount /testdir
 	[ "$status" -eq 1 ]
 	[[ "${output}" == *"Operation not permitted"* ]]
+}
+
+@test "mask paths [prohibit symlink /proc]" {
+	ln -s /symlink rootfs/proc
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *"must not be a symlink"* ]]
+}
+
+@test "mask paths [prohibit symlink /sys]" {
+	ln -s /symlink rootfs/sys
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *"must not be a symlink"* ]]
+}
+
+@test "mask paths [prohibit symlink /testdir]" {
+	rmdir rootfs/testdir
+	ln -s /symlink rootfs/testdir
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *"must not be a symlink"* ]]
+}
+
+@test "mask paths [prohibit symlink /testfile]" {
+	rm -f rootfs/testfile
+	ln -s /symlink rootfs/testfile
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *"must not be a symlink"* ]]
+}
+
+@test "mask paths [prohibit symlink /testdir2 (parent of /testdir2/testfile2)]" {
+	rm -rf rootfs/testdir2
+	ln -s /symlink rootfs/testdir2
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *"must not be a symlink"* ]]
 }


AkihiroSuda:
(We should also try to eliminate the TOCTOU issue, but just preventing the symlinks is probably enough as a quick fix for the CVE)

tests/integration/mask.bats
```diff
@@ -56,3 +57,41 @@ function teardown() {
...
+
+@test "mask paths [prohibit symlink /testfile]" {
+	rm -f rootfs/testfile
+	ln -s /symlink rootfs/testfile
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
+	[ "$status" -eq 1 ]
+	[[ "${output}" == *"must not be a symlink"* ]]
+}
+
+@test "mask paths [prohibit symlink /testdir2 (parent of /testdir2/testfile2)]" {
```
AkihiroSuda, line 91: RFC: Should we prohibit these symlinks, or should we just prohibit symlinking /proc and /sys ?

The latter might be better for compatibility, but might be less secure.

libcontainer/rootfs_linux.go
```diff
@@ -1115,3 +1120,51 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
...
+			mustNotBeSymlink[g] = struct{}{}
+		}
+	}
+	// mustNotBeSymlink typically contains {"/sys", "/proc"} in addition to readonlyPaths and maskedPaths.
+	for f := range mustNotBeSymlink {
+		if f != filepath.Clean(f) {
+			return fmt.Errorf("dirty path: %q", f)
+		}
+		// securejoin can't be used here, as it follows symlinks.
+		absMustNotBeSymlink := filepath.Join(config.Rootfs, f)
```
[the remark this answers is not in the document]
cyphar: IMHO `config.Rootfs` being a symlink or not isn't really important since it's something controlled by the container manager. I don't think we want to block having `config.Rootfs` contain symlink path components.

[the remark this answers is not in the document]
thaJeztah: Ah, yes, makes sense. Guess I was wondering if such symlinks would also affect the `mask paths` matching (I only had a cursory look ☺️)

libcontainer/rootfs_linux.go
```diff
@@ -1115,3 +1120,51 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to compute the ancestor paths of %q: %w", f, err)
+		}
+		for _, g := range a {
+			mustNotBeSymlink[g] = struct{}{}
+		}
+	}
+	// mustNotBeSymlink typically contains {"/sys", "/proc"} in addition to readonlyPaths and maskedPaths.
+	for f := range mustNotBeSymlink {
+		if f != filepath.Clean(f) {
```
kolyshkin, line 1143: It seems that  this check was already performed in `ancestorPaths` for all the paths. What's the point of repeating it?

AkihiroSuda: `mustNotBeSymlink[f] = struct{}{}` in L1132 is not covered by that check

libcontainer/rootfs_linux.go
```diff
@@ -1115,3 +1120,51 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
...
+	mustNotBeSymlink := make(map[string]struct{})
+	for _, f := range append(config.ReadonlyPaths, config.MaskPaths...) {
+		mustNotBeSymlink[f] = struct{}{}
+		a, err := ancestorPaths(f)
+		if err != nil {
+			return fmt.Errorf("failed to compute the ancestor paths of %q: %w", f, err)
+		}
+		for _, g := range a {
+			mustNotBeSymlink[g] = struct{}{}
+		}
```
kolyshkin, line 1139: I think this can be much simpler. Something like:
```go
if !filepath.IsAbs(f) || f != filepath.Clean(f) {
	return fmt.Errorf("bad path: %q", f)
}
for f != "/" {
	mustNotBeSymlink[f] = struct{}{}
	f = filepath.Dir(f)
}
```

libcontainer/rootfs_linux.go
kolyshkin: In fact it is covered, since `f` is passed on to `ancestorPaths` and the first thing it is doing is checking if the path is clean

libcontainer/rootfs_linux.go
```diff
@@ -1115,3 +1120,51 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
...
+	}
+	// mustNotBeSymlink typically contains {"/sys", "/proc"} in addition to readonlyPaths and maskedPaths.
+	for f := range mustNotBeSymlink {
+		if f != filepath.Clean(f) {
+			return fmt.Errorf("dirty path: %q", f)
+		}
+		// securejoin can't be used here, as it follows symlinks.
```
kolyshkin, line 1146: This is very minor, but I'd open rootfs directory and used fstatat(2) with dirfd here instead of joining the paths.

Faster, more secure, less garbage to collect.

Warning -- untested code:

```go
rootfd, err := os.OpenDir(config.Rootfs)
if err != nil {
	return err
}
defer rootfd.Close()
for f := range mustNotBeSymlink {
	var st unix.Stat_t
	err := unix.Fstatat(rootfd.Fd(), f, &st, unix.AT_SYMLINK_NOFOLLOW)
	if err == nil && stat.Mode&unix.S_IFMT == unix.S_IFLNK {
		return fmt.Errorf("must not be a symlink (conflicts with readonlyPaths and/or maskedPaths): %q", absMustNotBeSymlink)
	}
}

**Review by kolyshkin — CHANGES_REQUESTED** [review]
LGTM overall, left a couple of comments about how to simplify the code.

File: libcontainer/rootfs_linux.go [hunk 6]
```diff
@@ -1115,3 +1120,51 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
 		return unix.MountSetattr(-1, procfd, unix.AT_RECURSIVE, m.RecAttr)
 	})
 }
+
+// validateRoot detects symlinks in config.Rootfs that may break config.ReadonlyPaths
+// and config.MaskPaths.
+//
+// For preventing CVE-2023-27561 (CVE-2019-19921 re-introduction/regression).
+// https://github.com/opencontainers/runc/issues/3751
+func validateRoot(config *configs.Config) error {
```
cyphar, line 1129: Given that this is still subject to races, I wonder if a better solution might be to make `utils.WithProcfd` take an argument (or have a `utils.WithProcfdNoSymlinks` variant) which blocks symlink traversal so that we don't allow mounting on symlinks for paths where we don't want indirection (such as for masked and read-only paths as well as `procfs` and `sysfs`)? With `openat2` this can be done with `RESOLVE_NO_SYMLINKS` -- though I'll need to see what to do about path components...

To be honest, I read the issue and I don't understand why this appears to be a race condition problem -- isn't the issue that the check we added for the original CVE is completely ignored now because `utils.WithProcfd` blindly resolves the symlink?

AkihiroSuda: > openat2

Too new (kernel 5.6)

AkihiroSuda: Looks like the problem is just that `securejoin` follows symlinks?

AkihiroSuda: Looks like we don't use `WithProcfd` for `maskPath()`
https://github.com/opencontainers/runc/blob/214c16fd9a3bd9fe56a2476933e31749a7c0576b/libcontainer/rootfs_linux.go#L1017-L1030

libcontainer/rootfs_linux.go
AkihiroSuda: That doesn't seem to cover `/foo` for `/foo/bar/baz`

libcontainer/rootfs_linux.go
AkihiroSuda: Applied

libcontainer/rootfs_linux.go
kolyshkin: > That doesn't seem to cover `/foo` for `/foo/bar/baz`

I don't quite understand, but `/foo` will be added to the map.

Here's a simplified repro without the map, using examples from your test case: https://go.dev/play/p/_YmpzjwAJNz

tests/integration/mask.bats
kolyshkin: I think that the former is the right approach (parents should not be symlinks).

libcontainer/rootfs_linux.go
```diff
@@ -1115,3 +1119,53 @@ func setRecAttr(m *configs.Mount, rootfs string) error {
...
+	}
+	// mustNotBeSymlink typically contains {"/sys", "/proc"} in addition to readonlyPaths and maskedPaths.
+	rootFD, err := unix.Open(config.Rootfs, unix.O_DIRECTORY|unix.O_RDONLY, 0o600)
+	if err != nil {
+		return err
+	}
+	defer unix.Close(rootFD)
+	for f := range mustNotBeSymlink {
+		// fstatat(2): "If pathname is absolute, then dirfd is ignored."
+		fRel := strings.TrimPrefix(f, string(os.PathSeparator))
```
kolyshkin, line 1148: As the path is absolute and cleaned, it would be safe to use `f[1:]` in the call to `Fstatat`.

AkihiroSuda: For readability I don't prefer to use`f[1:]`

kolyshkin: The alternative is using `"."+f` in call to Fstatat but I don't like it either.

libcontainer/rootfs_linux.go
AkihiroSuda: Sorry, `for EXP` looked like `if EXP` to me 😓 

Applied your suggestion and added `Co-authored-by` with your name, thanks

libcontainer/rootfs_linux.go
cyphar: I'm not sure I understand how the issue got reintroduced then... I'm taking a look...

AkihiroSuda:
@opencontainers/runc-maintainers What should we do with this?

libcontainer/rootfs_linux.go
kolyshkin: @cyphar the issue was reintroduced because the checking code added by https://github.com/opencontainers/runc/pull/2207/commits/3291d66b98445bd7f7d02eac7f2bca2ac2c56942:

https://github.com/opencontainers/runc/blob/206008ab9c8f08db3a55fb8d4e3fe216ffbbb59d/libcontainer/rootfs_linux.go#L392

depends on `dest` being the actual path provided by a user. It stopped working as it is supposed to after commit https://github.com/opencontainers/runc/commit/0ca91f44f1664da834bc61115a849b56d22f595f#diff-e0932af06cb46065b2ef5fc89bc2bd9b880c9578e01b89ef4ecaeffb78a8bef0L331 changed from `dest = filepath.Join (rootfs, dest)` to `dest = securejoin.SecureJoin(rootfs, dest)`.

Now, `SecureJoin` resolves all symlinks under `dest` and thus checking is now done for link destination not the link itself.

kolyshkin: ... thus the check "if it's a symlink" is no longer working.

kolyshkin:
OTOH here's an alternative that makes the original fix from #2207 working again: https://github.com/opencontainers/runc/pull/3773

kolyshkin:
Yet another alternative, just for the sake of completeness, is to make SecureJoin not follow symlinks. Here's what I ended up with: 

```go
  // noSymVFS is a VFS interface for securejoin which disallows symlinks in the unsafePath.
  type noSymVFS struct{}  
   
  func (o noSymVFS) Lstat(name string) (fi os.FileInfo, err error) {
          fi, err = os.Lstat(name)
          if err == nil && fi.Mode()&os.ModeSymlink == os.ModeSymlink {
                  return nil, fmt.Errorf("%q is a symlink", name)
          }
          return
  }
   
  func (o noSymVFS) Readlink(name string) (string, error) { return os.Readlink(name) }
   
  func secureJoinNoSymlinks(root, unsafePath string) (string, error) {
          return securejoin.SecureJoinVFS(root, unsafePath, &noSymVFS{})
  }
   
  func mountToRootfs(m *configs.Mount, c *mountConfig) error {
  ...          
          secJoin := securejoin.SecureJoin
          if m.Device == "proc" || m.Device == "sysfs" {
                  // If the destination already exists and is not a directory, we bail
                  // out. This is to avoid mounting through a symlink or similar -- which
                  // has been a "fun" attack scenario in the past.
                  // TODO: This won't be necessary once we switch to libpathrs and we can
                  //       stop all of these symlink-exchange attacks.
                  secJoin = secureJoinNoSymlinks
          }
          dest, err := secJoin(rootfs, m.Destination)
          
```

This has not been tested but I definitely don't like this code. The approach used in #3773 (or here) seems better.

libcontainer/rootfs_linux.go
cyphar: Ah, you're right. I kept missing than when I was looking over it. I personally prefer the approach in #3773. Doing a pre-flight "scan" of the rootfs seems less than ideal (neither approach protects against races but the pre-flight one is more obviously incapable of handling races -- the approach in #3773 seems closer to what libpathrs would do and actually fixes the core bug in the CVE fix I wrote).

AkihiroSuda:
Replaced by
- https://github.com/opencontainers/runc/pull/3773

PR #3762, opened 2023-03-08
ci/cirrus: fix failures on Centos 9

kolyshkin:
...debugging...


kolyshkin:
Oh well. This is more complicated than I thought. Updating the systemd rpms helped, but I guess it happened only because systemd was restarted during the update.

Looking.

AkihiroSuda (approved):
Thanks

kolyshkin:
Alas, it's not working :(

PR #3772, opened 2023-03-16
nsexec: retry unshare on EINVAL

kolyshkin:
Older kernels may return EINVAL on unshare when a process is reading runc's /proc/$PID/status or /proc/$PID/maps. This was fixed by kernel commit 12c641ab8270f ("unshare: Unsharing a thread does not require unsharing a vm") in Linux v4.3.

For CentOS 7, the fix was backported to CentOS 7.7 (kernel 3.10.0-1062).

To work around this kernel bug, let's retry on EINVAL a few times.

Not adding a test case as the fix is short and obvious, and the repro is somewhat complicated (see #3705).

Closes: #3705

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -833,6 +833,25 @@ void send_mountsources(int sockfd, pid_t child, char *mountsources, size_t mount
 		bail("failed to close container mount namespace fd %d", container_mntns_fd);
 }
 
+void try_unshare(int flags, const char *msg)
+{
+	write_log(DEBUG, "unshare %s", msg);
+	/*
+	 * Kernels prior to v4.3 may return EINVAL on unshare when another process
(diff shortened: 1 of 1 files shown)


cyphar (approved):
lgtm

AkihiroSuda (approved):

AkihiroSuda:
> Older kernels may return EINVAL on unshare when a process is reading runc's /proc/$PID/status or /proc/$PID/maps. This was fixed by kernel commit 12c641ab8270f ("unshare: Unsharing a thread does not require unsharing a vm") in Linux v4.3.
>
> For CentOS 7, the fix was backported to CentOS 7.7 (kernel 3.10.0-1062).

Could you add these lines to the comment li?

rata (approved):
LGTM

kolyshkin:
> Could you add these lines to the comment lines in the code?

I had it in there but later decided to be more concise in the code and move all the details into the commit message.

PR #3773, opened 2023-03-16
Prohibit /proc and /sys to be symlinks

kolyshkin:
Commit 3291d66b9844 introduced a check for /proc and /sys, making sure the destination (dest) is a directory (and not e.g. a symlink).

Later, a hunk from commit 0ca91f44f switched from using `filepath.Join` to `SecureJoin` for `dest`. As `SecureJoin` follows and resolves symlinks, the check whether dest is a symlink no longer works.

To fix, do the check without/before using SecureJoin.

Add integration tests (stolen from #3756) to make sure we won't regress.

_This is an alternative to #3756._

Fixes: #3751
Fixes: CVE-2023-27561

### DIFF
--- a/libcontainer/rootfs_linux.go
+++ b/libcontainer/rootfs_linux.go
@@ -375,6 +375,35 @@ func doTmpfsCopyUp(m *configs.Mount, rootfs, mountLabel string) (Err error) {
 
 func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 	rootfs := c.root
+
+	// procfs and sysfs are special because we need to ensure they are actually
+	// mounted on a specific path in a container without any funny business.
+	switch m.Device {
+	case "proc", "sysfs":
+		// If the destination already exists and is not a directory, we bail
+		// out. This is to avoid mounting through a symlink or similar -- which
+		// has been a "fun" attack scenario in the past.
+		// TODO: This won't be necessary once we switch to libpathrs and we can
+		//       stop all of these symlink-exchange attacks.
+		dest := filepath.Clean(m.Destination)
+		if !strings.HasPrefix(dest, rootfs) {
+			// Do not use securejoin as it resolves symlinks.
+			dest = filepath.Join(rootfs, dest)
+		}
+		if fi, err := os.Lstat(dest); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+		} else if !fi.IsDir() {
+			return fmt.Errorf("filesystem %q must be mounted on ordinary directory", m.Device)
+		}
+		if err := os.MkdirAll(dest, 0o755); err != nil {
+			return err
+		}
+		// Selinux kernels do not support labeling of /proc or /sys.
+		return mountPropagate(m, rootfs, "", nil)
+	}
+
 	mountLabel := c.label
 	mountFd := c.fd
 	dest, err := securejoin.SecureJoin(rootfs, m.Destination)
@@ -383,24 +412,6 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
 	}
 
 	switch m.Device {
-	case "proc", "sysfs":
-		// If the destination already exists and is not a directory, we bail
-		// out This is to avoid mounting through a symlink or similar -- which
-		// has been a "fun" attack scenario in the past.
-		// TODO: This won't be necessary once we switch to libpathrs and we can
-		//       stop all of these symlink-exchange attacks.
-		if fi, err := os.Lstat(dest); err != nil {
-			if !os.IsNotExist(err) {
-				return err
-			}
-		} else if fi.Mode()&os.ModeDir == 0 {
-			return fmt.Errorf("filesystem %q must be mounted on ordinary directory", m.Device)
-		}
-		if err := os.MkdirAll(dest, 0o755); err != nil {
-			return err
-		}
-		// Selinux kernels do not support labeling of /proc or /sys
-		return mountPropagate(m, rootfs, "", nil)
 	case "mqueue":
 		if err := os.MkdirAll(dest, 0o755); err != nil {
 			return err
--- a/tests/integration/mask.bats
+++ b/tests/integration/mask.bats
@@ -56,3 +56,22 @@ function teardown() {
 	[ "$status" -eq 1 ]
 	[[ "${output}" == *"Operation not permitted"* ]]
 }
+
+@test "mask paths [prohibit symlink /proc]" {
+	ln -s /symlink rootfs/proc
+	runc run -d --console-socket "$CONSOLE_SOCKET" test_busybox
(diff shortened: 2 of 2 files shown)


libcontainer/rootfs_linux.go
```diff
@@ -377,25 +377,35 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
...
 		}
+	}
+
+	dest, err := securejoin.SecureJoin(rootfs, m.Destination)
+	if err != nil {
+		return err
+	}
+
+	switch m.Device {
+	case "proc", "sysfs":
```
cyphar, line 408: This can be simplified by just having the proc/sysfs case be a standalone `switch` which returns and then have everything afterwards. Something like:

```patch
index 2a98372b561d..b698718da85d 100644
@@ -377,13 +377,18 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
        rootfs := c.root
        mountLabel := c.label
        mountFd := c.fd
-       dest, err := securejoin.SecureJoin(rootfs, m.Destination)
-       if err != nil {
-               return err
-       }

+       // procfs and sysfs are special because we need to ensure they are actually
+       // mounted on a specific path in a container without any funny business.
        switch m.Device {
        case "proc", "sysfs":
+               // Scope the path to the rootfs. We cannot use SecureJoin because it
+               // resolves symlinks which means we cannot check if the target is a
+               // symlink.
+               dest := m.Destination
+               if !strings.HasPrefix(dest, rootfs) {
+                       dest = filepath.Join(rootfs, dest)
+               }
                // If the destination already exists and is not a directory, we bail
                // out This is to avoid mounting through a symlink or similar -- which
                // has been a "fun" attack scenario in the past.
@@ -393,7 +398,7 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
                        if !os.IsNotExist(err) {
                                return err
                        }
-               } else if fi.Mode()&os.ModeDir == 0 {
+               } else if !fi.IsDir() {
                        return fmt.Errorf("filesystem %q must be mounted on ordinary directory", m.Device)
                }
                if err := os.MkdirAll(dest, 0o755); err != nil {
@@ -401,6 +406,14 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
                }
                // Selinux kernels do not support labeling of /proc or /sys
                return mountPropagate(m, rootfs, "", nil)
+       }
+
+       dest, err := securejoin.SecureJoin(rootfs, m.Destination)
+       if err != nil {
+               return err
+       }
+
+       switch m.Device {
        case "mqueue":
                if err := os.MkdirAll(dest, 0o755); err != nil {
                        return err
```

cyphar:
With the above comment in mind, I prefer this patch to #3756.

libcontainer/rootfs_linux.go
```diff
@@ -377,25 +377,35 @@ func mountToRootfs(m *configs.Mount, c *mountConfig) error {
...
 
 	switch m.Device {
 	case "proc", "sysfs":
 		// If the destination already exists and is not a directory, we bail
-		// out This is to avoid mounting through a symlink or similar -- which
+		// out. This is to avoid mounting through a symlink or similar -- which
 		// has been a "fun" attack scenario in the past.
 		// TODO: This won't be necessary once we switch to libpathrs and we can
 		//       stop all of these symlink-exchange attacks.
+		dest := m.Destination
```
thaJeztah, line 388: Do we need to `filepath.Clean()` before checking the prefix?

cyphar: Probably `dest := filepath.Clean("/"+m.Destination)` makes the most sense.

kolyshkin: In the light of that, I guess, we need to finish what we started in commits 1f1e91b1a and b31a9340f. Before doing that, I guess, we could not use `filepath.Clean("/"+m.Destination)`.

libcontainer/rootfs_linux.go
kolyshkin: updated

libcontainer/rootfs_linux.go
kolyshkin: And once we'll finish it, it won't be necessary to prepend `"/"`. So, I guess, bare `filepath.Clean` is enough.

kolyshkin:
Thinking about one more alternative, which is:

1. ensure dest is cleaned
2. open an fd at roofs+dest with O_PATH|O_NOFOLLOW|O_DIRECTORY
3. In case of ENOENT, mkdir and retry once; bail out for other errors.
4. pass the fd to mountPropagate
 
 This way we'll remove the TOCTOU race and we don't even need openat2.
 
 WDYT @cyphar ?

cyphar:
@kolyshkin I don't really like that idea. I would prefer if we just use this solution now and we can solve this (and the many other issues we have) with libpathrs in the future.

You can have container configs that mount sysfs and procfs in places other than `/sys` and `/proc` and I'd be far too concerned with that resulting in another awful symlink-related container attack.

cyphar (approved):
LGTM.

thaJeztah:
Probably not for this PR, but I noticed that we're inconsistent in handling "where to return". It looks like _all switch cases_ have a `return`, except for one, and that one is put outside of the switch; https://github.com/opencontainers/runc/blob/efad7a3b80f4585065f93aad3e91989da5012d63/libcontainer/rootfs_linux.go#L475-L478

Perhaps we should move that one inside the `case "bind"` branch to make it clearer what condition it relates to. Doing so may also help linters detect that we're missing a return statement if we would add a new branch to any of the switches.
https://github.com/opencontainers/runc/blob/efad7a3b80f4585065f93aad3e91989da5012d63/libcontainer/rootfs_linux.go#L436

thaJeztah (approved):
LGTM

@AkihiroSuda PTAL

AkihiroSuda (approved):

thaJeztah:
I guess this one needs to be cherry-picked into the release-1.1 branch?

thaJeztah:
opened https://github.com/opencontainers/runc/pull/3785

kolyshkin:
> Probably not for this PR, but I noticed that we're inconsistent in handling "where to return". It looks like _all switch cases_ have a `return`, except for one, and that one is put outside of the switch;
> 
> https://github.com/opencontainers/runc/blob/efad7a3b80f4585065f93aad3e91989da5012d63/libcontainer/rootfs_linux.go#L475-L478
> 
> Perhaps we should move that one inside the `case "bind"` branch to make it clearer what condition it relates to. Doing so may also help linters detect that we're missing a return statement if we would add a new branch to any of the switches.
> 
> https://github.com/opencontainers/runc/blob/efad7a3b80f4585065f93aad3e91989da5012d63/libcontainer/rootfs_linux.go#L436

Makes sense, opened https://github.com/opencontainers/runc/pull/3787

Issue #3780, opened 2023-03-23
runc systemd cgroup driver logic is wrong


kolyshkin:
### Description

While investigating issue #3760 I found the following.

## Issue A: leftover failed scope unit.

1. By default, systemd removes transient units (e.g. scopes) that have ended (are no longer active), except those that failed. The failed units are not being removed, they stay.
2. RHEL9's systemd, for some reason (I guess a RHEL9-specific patch I can't find yet), works differently than in does in older CentOS, or Fedora 37 (or earlier Fedora versions, I guess). The difference is, if the container is OOM-killed (which happen in "event oom" test), its scope state becomes "failed".

This above is the first issue. The container ended, but its scope stays. Here is a repro (works on CentOS Stream 9 only):

```
[root@localhost runc]# systemctl status runc-test_busybox.scope
Unit runc-test_busybox.scope could not be found.
[root@localhost runc]# bats tests/integration/events.bats 
1..5
ok 1 events --stats
ok 2 events --interval default
ok 3 events --interval 1s
ok 4 events --interval 100ms
ok 5 events oom
[root@localhost runc]# systemctl status runc-test_busybox.scope
Warning: The unit file, source configuration file or drop-ins of runc-test_busybox.scope changed on disk. Run 'systemctl daemon->
× runc-test_busybox.scope - libcontainer container test_busybox
     Loaded: loaded (/run/systemd/transient/runc-test_busybox.scope; transient)
  Transient: yes
    Drop-In: /run/systemd/transient/runc-test_busybox.scope.d
             └─50-DevicePolicy.conf, 50-DeviceAllow.conf, 50-MemoryMax.conf
     Active: failed (Result: oom-kill) since Thu 2023-03-23 01:58:22 UTC; 2s ago
   Duration: 5.729s
        CPU: 710ms

Mar 23 01:58:16 localhost systemd[1]: Started libcontainer container test_busybox.
Mar 23 01:58:22 localhost systemd[1]: runc-test_busybox.scope: A process of this unit has been killed by the OOM killer.
Mar 23 01:58:22 localhost systemd[1]: runc-test_busybox.scope: Killing process 248957 (sh) with signal SIGKILL.
Mar 23 01:58:22 localhost systemd[1]: runc-test_busybox.scope: Killing process 249001 (sh) with signal SIGKILL.
Mar 23 01:58:22 localhost systemd[1]: runc-test_busybox.scope: Killing process 249007 (dd) with signal SIGKILL.
[...]

Issue #3760, opened 2023-03-08
`centos-stream-9` CI is failing for the main branch


AkihiroSuda:
https://cirrus-ci.com/task/5246167815028736 (https://github.com/opencontainers/runc/commit/6d0261c1ac38ff84b4aea73f02383d6d42ebfdfd)

```
ssh -tt localhost "make -C /home/runc localintegration RUNC_USE_SYSTEMD=yes"
make: Entering directory '/home/runc'
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=6d0261c1 -X main.version=1.1.0+dev " -o runc .
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=6d0261c1 -X main.version=1.1.0+dev " -o contrib/cmd/recvtty/recvtty ./contrib/cmd/recvtty
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=6d0261c1 -X main.version=1.1.0+dev " -o contrib/cmd/sd-helper/sd-helper ./contrib/cmd/sd-helper
go build -trimpath "-buildmode=pie"  -tags "seccomp urfave_cli_no_docs" -ldflags "-X main.gitCommit=6d0261c1 -X main.version=1.1.0+dev " -o contrib/cmd/seccompagent/seccompagent ./contrib/cmd/seccompagent
bats -t tests/integration
1..181
ok 1 runc run no capability
ok 2 runc run with unknown capability
...
ok 97 runc pause and resume
ok 98 runc pause and resume with nonexist container
not ok 99 ps
# (in test file tests/integration/ps.bats, line 27)
#   `[[ "${lines[1]}" == *"$(id -un 2>/dev/null)"*[0-9]* ]]' failed
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-78374/runc.VKglS3/tty/sock test_busybox (status=0):
# 
# runc state test_busybox (status=0):
# {
#   "ociVersion": "1.1.0-rc.1",
#   "id": "test_busybox",
#   "pid": 90743,
#   "status": "running",
#   "bundle": "/tmp/bats-run-78374/runc.VKglS3/bundle",
#   "rootfs": "/tmp/bats-run-78374/runc.VKglS3/bundle/rootfs",
[...]

PR #3782, opened 2023-03-23
Fix systemd cgroup driver's Apply

kolyshkin:
This fixes the wrong logic of systemd cgroup manager's Apply method. For more details, see commit messages and #3780.

Fixes: #3780
Fixes: #3760

### DIFF
--- a/tests/integration/cgroups.bats
+++ b/tests/integration/cgroups.bats
@@ -356,7 +356,7 @@ function setup() {
 	[ "$output" = "ok" ]
 }
 
-@test "runc run/create should warn about a non-empty cgroup" {
+@test "runc run/create should error for a non-empty cgroup" {
 	[ $EUID -ne 0 ] && requires rootless_cgroup
 
 	set_cgroups_path
@@ -366,12 +366,12 @@ function setup() {
 
 	# Run a second container sharing the cgroup with the first one.
 	runc --debug run -d --console-socket "$CONSOLE_SOCKET" ct2
-	[ "$status" -eq 0 ]
+	[ "$status" -ne 0 ]
 	[[ "$output" == *"container's cgroup is not empty"* ]]
 
 	# Same but using runc create.
 	runc create --console-socket "$CONSOLE_SOCKET" ct3
-	[ "$status" -eq 0 ]
+	[ "$status" -ne 0 ]
 	[[ "$output" == *"container's cgroup is not empty"* ]]
 }
 
--- a/libcontainer/cgroups/systemd/common.go
+++ b/libcontainer/cgroups/systemd/common.go
@@ -124,7 +124,7 @@ func isUnitExists(err error) bool {
 	return isDbusError(err, "org.freedesktop.systemd1.UnitExists")
 }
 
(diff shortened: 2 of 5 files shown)


kolyshkin:
The failure with "runc ps" on CentOS 9 is gone as expected.

This failure:

> not ok 168 update cgroup cpu.idle

is a separate issue not addressed by this PR.

tests/integration/cgroups.bats
```diff
@@ -356,7 +356,7 @@ function setup() {
 	[ "$output" = "ok" ]
 }
 
-@test "runc run/create should warn about a non-empty cgroup" {
+@test "runc run/create should error for a non-empty cgroup" {
```
AkihiroSuda, line 359: Is this a breaking change?

kolyshkin: To begin with, it never worked anyway, because if the systemd unit exists, the error is ignored and the new container is never added to the unit (and to the cgroup). To repeat, it never worked anyway.

We can try making it work in 1.1 though. The fix would be very different from this one, something like "if `startUnit` returned `UnitExists` error, call `setUnitProperties` with properties of `PIDs=[new pid]`".

I am not sure that this will work (maybe, maybe not -- it's complicated. I am also unsure if we want to go that route at all -- I mean trying to fix something that never worked anyway.

In this version (1.2.x), I think this is the way it should be done.

In 1.1, we can discuss it later (for the 1.1.6 I guess).

kolyshkin: To reiterate -- let's concentrate on how can we fix it in main branch for now, and think about 1.1 backport later.

AkihiroSuda (dismissed):

tests/integration/cgroups.bats
kolyshkin: Thought about this a bit -- I think we can still allow shared cgroup when fs cgroup driver is used, keeping the warning, and error out in case of systemd cgroup driver. This is sort of a breaking change, but since

1. the functionality never worked correctly (UnitExists error was ignored, and container was not placed into the proper systemd unit and/or cgroup), and
2. will be deprecated in runc 1.2,
3. implementing such a feature (adding a container to an existing systemd unit) is not very easy,

it makes little sense in trying to do that.

In particular, this test can be changed to look for a warning in case of an fs cgroup driver, and error in case of systemd cgroup driver.

kolyshkin:
Rebased. The only CI failure to expect here is centos-stream-9 one from #3786 (`not ok NN update cgroup cpu.idle`), being fixed/mitigated by #3788.

kolyshkin:
Rebased. The only CI failure to expect here is centos-stream-9 one from #3786 (not ok NN update cgroup cpu.idle), being fixed/mitigated by #3788.

tests/integration/cgroups.bats
mrunalp: @kolyshkin That sounds good!

kolyshkin:
1.1 backport, more conservative: #3806.

tests/integration/cgroups.bats
kolyshkin: Implemented as described above in https://github.com/opencontainers/runc/pull/3806

PR #3787, opened 2023-03-27
mountToRootfs: minor refactor

kolyshkin:
The setRecAttr is only called for "bind" case, as cases end with a return statement. Indeed, recursive mount attributes only make sense for bind mounts.

Move the code to under case "bind" to improve readability. No change in logic.

Fixes: 382eba4354d764aaff
Reported-by: Sebastiaan van Stijn <github@gone.nl>


kolyshkin:
@thaJeztah PTAL

AkihiroSuda (approved):

Issue #3786, opened 2023-03-27
centos9+systemd: setting cpu-period resets cpu.idle


kolyshkin:
### Description

One of our integration tests is failing on CentOS Stream 9 when using systemd cgroup driver.

What happens is `cpu.idle` setting is getting reset from 1 to 0 when setting cpu-period.

### Steps to reproduce the issue

Minimal repro in bats:

```bats
#!/usr/bin/env bats

load helpers

function teardown() {
        teardown_bundle
}

function setup() {
        setup_busybox
        set_cgroups_path
}

@test "update cgroup cpu.idle" {
        requires cgroups_cpu_idle
        [ $EUID -ne 0 ] && requires rootless_cgroup

        runc run -d --console-socket "$CONSOLE_SOCKET" test_update
        [ "$status" -eq 0 ]

        check_cgroup_value "cpu.idle" "0"
        runc update --cpu-idle 1 test_update
        check_cgroup_value "cpu.idle" "1"

        # test update other option won't impact on cpu.idle
        runc update --cpu-period 10000 test_update
        [ "$status" -eq 0 ]
        check_cgroup_value "cpu.idle" "1"
}
```

### Describe the results you received and expected
```
[root@localhost runc]# export RUNC_USE_SYSTEMD=yes
[root@localhost runc]# bats  tests/integration/u2.bats 
1..1
not ok 1 update cgroup cpu.idle
# (from function `check_cgroup_value' in file tests/integration/helpers.bash, line 265,
#  in test file tests/integration/u2.bats, line 28)
#   `check_cgroup_value "cpu.idle" "1"' failed
# runc spec (status=0):
# 
# runc run -d --console-socket /tmp/bats-run-102847/runc.ooH08F/tty/sock test_update (status=0):
# 
# current 0 !? 0
# runc update --cpu-idle 1 test_update (status=0):
# 
# current 1 !? 1
# runc update --cpu-period 10000 test_update (status=0):
# 
# current 0 !? 1
```
### What version of runc are you using?
```
[root@localhost runc]# ./runc --version
runc version 1.1.0+dev
commit: e6cc9a67
spec: 1.1.0-rc.1
go: go1.19
libseccomp: 2.5.2
```
### Host OS information
```
[root@localhost runc]# cat /etc/os-release 
NAME="CentOS Stream"
VERSION="9"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="9"
PLATFORM_ID="platform:el9"
[...]

PR #3788, opened 2023-03-27
cg/sd: support CPUWeight=idle (aka cpu.idle=1)

kolyshkin:
Systemd v252 added support for setting cgroup v2 cpu.idle in https://github.com/systemd/systemd/pull/23299

The way it works is
 - if CPUWeight == 0, cpu.idle is set to 1;
 - if CPUWeight != 0, cpu.idle is set to 0.

This PR implements support for setting `cpu.idle` via systemd cgroup
v2 driver. In case CPUIdle is set to non-zero value, the driver sets
CPUWeight=0 property, which will result in systemd setting cpu.idle
to 1.
    
Unfortunately, there's no way to set cpu.idle to 0 without also changing
the CPUWeight value, so the driver doesn't do anything if CPUIdle is
explicitly set to 0. This case is handled by the fs driver which is
always used as a followup to setting systemd unit properties.
    
Also, handle cpu.idle set via unified map. In case it is set to non-zero
value, add CPUWeight=0 property, and ignore cpu.weight (otherwise we'll
get two different CPUWeight properties set).
    
Add a unit test for new values in unified map, and an integration test case.
Modify docs to show the newly supported property.

Fixes: #3786

### DIFF
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -64,8 +64,7 @@ func unifiedResToSystemdProps(cm *dbusConnManager, res map[string]string) (props
 		if strings.Contains(k, "/") {
 			return nil, fmt.Errorf("unified resource %q must be a file name (no slashes)", k)
 		}
-		sk := strings.SplitN(k, ".", 2)
-		if len(sk) != 2 {
+		if strings.IndexByte(k, '.') <= 0 {
 			return nil, fmt.Errorf("unified resource %q must be in the form CONTROLLER.PARAMETER", k)
 		}
 		// Kernel is quite forgiving to extra whitespace
--- a/libcontainer/cgroups/systemd/v2.go
+++ b/libcontainer/cgroups/systemd/v2.go
@@ -20,6 +20,10 @@ import (
 	"github.com/opencontainers/runc/libcontainer/configs"
 )
 
+const (
+	cpuIdleSupportedVersion = 252
+)
+
 type UnifiedManager struct {
 	mu      sync.Mutex
 	cgroups *configs.Cgroup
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
> currently only CentOS 9

And Fedora?

kolyshkin:
> > currently only CentOS 9
> 
> And Fedora?

F37 on my laptop has systemd v231

AkihiroSuda:
> > > currently only CentOS 9
> > 
> > 
> > And Fedora?
> 
> F37 on my laptop has systemd v231

👀 

I was expecting CentOS 9 Stream to be more conservative than Fedora.
Was there any specific reason that CentOS 9 Stream adopted the new systemd version ahead of Fedora?

AkihiroSuda (approved):

kolyshkin:
It is really hard to say how to handle this properly. The kernel allows to set `cpu.weight` and `cpu.idle` at the same time (and if `cpu.idle` is enabled, `cpu.weight` is ignored). Systemd does not allow that, overwriting `cpu.idle` in many scenarios (including, but apparently not limited to setting cpu-shares which is translated to CPUWeight property).

That means, for example, that running `runc update --cpu-shares 200 --cpu.idle 1` will cause systemd to set `cpu.idle` to `0`, and then runc will set it to `1` (because we back up everything that systemd does by the cgroup fs driver). Some other update, even unrelated, may trigger systemd to set cpu.idle back to 0 again.

For example, this is what we see when changing cpu-period (see #3786), which appears to be unrelated to cpu weight and cpu idle.

So, maybe we should make systemd cgroup v2 driver  to deny setting some CPU-related parameters if cpu.idle was/is set to 1.

kolyshkin:
Oh my, finally the CI is green again. If we merge this PR this week, we will be able to brag that it only took us a month to fix the CI (in main branch, that is).

libcontainer/cgroups/systemd/v2.go
```diff
@@ -64,8 +64,7 @@ func unifiedResToSystemdProps(cm *dbusConnManager, res map[string]string) (props
 		if strings.Contains(k, "/") {
 			return nil, fmt.Errorf("unified resource %q must be a file name (no slashes)", k)
 		}
-		sk := strings.SplitN(k, ".", 2)
-		if len(sk) != 2 {
+		if strings.IndexByte(k, '.') < 2 {
```
mrunalp, line 79: why less than 2? `< 1` should work, right? (Assuming there is a hypothetical one character controller in the future..)

kolyshkin: You're right, and I have a correct description in the commit message, but not in the code :)

Fixed

AkihiroSuda (approved):

PR #3790, opened 2023-03-28
Fix undefined behavior of setjmp() macro.

nanasi880:
Accepting the return value of a C setjmp macro as a variable is undefined behavior and should generally not be considered safe.

reference: https://pubs.opengroup.org/onlinepubs/009604599/functions/setjmp.html
```
An application shall ensure that an invocation of setjmp() appears in one of the following contexts only:
    - The entire controlling expression of a selection or iteration statement
    - One operand of a relational or equality operator with the other operand an integral constant expression, with the resulting expression being the entire controlling expression of a selection or iteration statement
    - The operand of a unary '!' operator with the resulting expression being the entire controlling expression of a selection or iteration
    - The entire expression of an expression statement (possibly cast to void)

If the invocation appears in any other context, the behavior is undefined.
```

It was introduced for debugging purposes in the following commit, and earlier implementations did not have this problem.
https://github.com/opencontainers/runc/commit/64bb59f5920b15d886cb2be52aede641fd4a047b

The implementation has been modified to conform to C standards while maintaining this debugging information where possible.
In the default case, we have given up on displaying `current_stage`.

### DIFF
--- a/libcontainer/nsenter/nsexec.c
+++ b/libcontainer/nsenter/nsexec.c
@@ -980,8 +980,7 @@ void nsexec(void)
 	 * -- Aleksa "what has my life come to?" Sarai
 	 */
 
-	current_stage = setjmp(env);
-	switch (current_stage) {
+	switch (setjmp(env)) {
 		/*
 		 * Stage 0: We're in the parent. Our job is just to create a new child
 		 *          (stage 1: STAGE_CHILD) process and write its uid_map and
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, but please sign the commit for DCO
https://github.com/apps/dco

(run `git commit -a -s --amend`, and make sure that the `Signed-off-by: NAME <EMAIL>` line with your real name is included in the commit message)

nanasi880:
Completed sign the commit for DCO.

AkihiroSuda (approved):

kolyshkin:
Rebased for the sake of green CI

cyphar (approved):
lgtm

PR #3791, opened 2023-03-28
[1.1] refresh ci

### LINKED ISSUE #3715: Runc 1.1.4 container fails with `permission denied: unknown` when relying on capabilities

kolyshkin:
This is a prerequisite to fix #3715 in 1.1.6. Once merged, I will backport the fix (https://github.com/opencontainers/runc/pull/3753).

This PR removes Go 1.16 support, and adds Go 1.19 and Go 1.20 support. Mostly touches CI and formatting.

 - add Go 1.19.x and Go 1.20.x to CI matrix
 - update some dependencies (x/sys/unix, needed for #3715)
 - drop Go 1.16.x support (needed for the above bump)
 - backport some other bumps and fixes so that we can use go 1.19 and 1.20 (for CI and releases)
 - make use of Go 1.20 for release job (bumps Go to 1.20 in Dockerfile)

The reason we have to drop Go 1.16 support is to fix #3715. It looks like we can still keep Go 1.17 support.

This also backports
 - #3378 and #3379, because `go get` just stopped working in Ubuntu 20.04 gha image
 - #3735 to fix a flaky test case

~~The CS9 CI failure is unrelated and is being fixed by #3806.~~

Once merged, I need to modify the required CI jobs list:
 - remove test (1.16.x) jobs;
 - remove lint-extra job (it's now part of lint);
 - add 1.19.x and 1.20.x jobs.


kolyshkin:
Two questions for @opencontainers/runc-maintainers (cc @AkihiroSuda @thaJeztah @cyphar)
1. Can we afford losing Go 1.16.x compatibility in a stable branch? (it's either that or not fixing #3715)?
2. More generally, do we need to support older (unsupported) Go releases in a stable branch?
3. Should we support latest Go releases in a stable branch?

My answer to q 3 is **yes**, and that alone might force us to say **no** to q 2.

kolyshkin:
The CI is all green now. @opencontainers/runc-maintainers PTAL -- merging this will help to unblock other 1.1 PRs.
Cc @thaJeztah @AkihiroSuda

haircommander:
Not a maintainer, but I am invested in 1.1.6, so I've decided to review and throw my 2c into the mix

> More generally, do we need to support older (unsupported) Go releases in a stable branch?
> Should we support latest Go releases in a stable branch?

an approach I've seen work answers these questions as yes and no respectively.  I would personally advocate for keeping the lowest possible golang version supported on this branch. If distributions want the stability of patch releases, they likely are staying on lower versions of golang.

that said, runc and more specifically libcontainer's latest release is in the 1.1 cycle, and other projects are likely building with newer golang versions, but still require a cut version. That would require minor releases of runc be cut more frequently to provide a matrix of support/compatiblity with newer golangs.

If I were to get a vote, I would say the purpose of this PR should focus on bumping to golang 1.17 (and probably also 1.18 because runc generally supports two golang versions), but drop the pieces focusing on 1.19/1.20, and direct that energy towards cutting 1.2.0, which would have support for newer golangs. That would simplify this PR, and reduce the maintenance burden of keeping 1.1 working with newer libraries, while also fixing the issue with golang 1.16 and fixing the runc 1.1 CI generally

kolyshkin:
> Not a maintainer, but I am invested in 1.1.6, so I've decided to review and throw my 2c into the mix
> 
> > More generally, do we need to support older (unsupported) Go releases in a stable branch?
> > Should we support latest Go releases in a stable branch?
> 
> an approach I've seen work answers these questions as yes and no respectively. I would personally advocate for keeping the lowest possible golang version supported on this branch. If distributions want the stability of patch releases, they likely are staying on lower versions of golang.
> 
> that said, runc and more specifically libcontainer's latest release is in the 1.1 cycle, and other projects are likely building with newer golang versions, but still require a cut version. That would require minor releases of runc be cut more frequently to provide a matrix of support/compatiblity with newer golangs.
> 
> If I were to get a vote, I would say the purpose of this PR should focus on bumping to golang 1.17 (and probably also 1.18 because runc generally supports two golang versions), but drop the pieces focusing on 1.19/1.20, and direct that energy towards cutting 1.2.0, which would have support for newer golangs. That would simplify this PR, and reduce the maintenance burden of keeping 1.1 working with newer libraries, while also fixing the issue with golang 1.16 and fixing the runc 1.1 CI generally

My train of thoughts is this: in 1.1.x we either stick to the same Go releases as in 1.1.0, or we don't. It worked just fine with sticking until recently. It's different now -- sticking to Go 1.16 stands in a way of fixing #3715, so we have to choose one.

Personally, I choose to fix (maybe because it's my bug), meaning we have to drop Go 1.16, meaning we do not stick to the same releases as in runc 1.1.0. This allows us to make one more step forward and actually support latest and non-obsoleted Go releases. For one thing, this makes sense as kubernetes, one of the biggest runc/libcontainer users, always uses latest Go.

At the same time, I see no meaning in trying to add support for Go 1.18 as it is obsoleted.

As to the complexity of this PR, believe me, I feel the pain. It's not as ugly as it looks though -- if you check the actual changes in .go files, those are minimal. The changes is mostly CI, and tests, and linters, that sort of thing.

haircommander:
> My train of thoughts is this: in 1.1.x we either stick to the same Go releases as in 1.1.0, or we don't. It worked just fine with sticking until recently. It's different now -- sticking to Go 1.16 stands in a way of fixing https://github.com/opencontainers/runc/issues/3715, so we have to choose one.

I understand the position of being able to rip off the bandaid of supporting old releases if we're dropping support for 1.16 anyway. However, a distribution that maintains legacy versions of packages is more likely to have support for 1.17 than 1.19 at this point, so the point of supporting an older release is not entirely moot.

> It's not as ugly as it looks though

FWIW: I did go through each commit, and they all collectively LGTM if this is the approach taken

thaJeztah:
My quick takes;

1. Yes, we should try to get to 1.2.0
2. Go recommendations are to support "latest" and "latest - 1". Those are the only maintained releases by upstream.
3. Our _recommendation_ is aligned with the above; we cannot commit ourselves to maintaining an LTS fork of Go, so we recommend latest (or latest - 1), and (as a project) cannot provide any guarantees for EOL, unsupported Go versions.
4. But: continue to try to run CI on "version that was used when the release branch was cut (or "first release from the branch")
    - call out that this is on a **best effort base**.
    - Do _NOT_ raise the minimum _required_ version, unless there is a warranted reason to (i.e., don't require go1.19, because we could make things a bit more DRY with "generic")
    - 👉 In this case, we have a warranted reason for bumping the minimum version (go1.16 -> go1.17)

As to distros providing LTS releases, which may include a custom LTS version of (EOL) Go versions;

- These distros provide the LTS versions as a value-add / service. With providing an LTS comes the responsibility.
- ^^ this project is open-source, and if maintaining such an LTS release requires them to maintain a fork (or apply patches), they are free to do so.
- ^^ we will continue to attempt not to make life hard for them (for no reason), i.e., not the aforementioned "let's do generics in a patch release, because they're cool"
- ^^ and (within reason) will accept contributions that could ease the pain of maintaining a fork.

From the above, the only part we could consider in this PR is to (if possible) keep the `_go116.go` and `_go117.go` files (while _untested_) to reduce the code-churn and (potentially) make life easier for those maintaining a fork. But if that's not possible, then removing those seems perfectly reasonable to me.

mrunalp (approved):
I made a pass commit by commit and the changes look good.

thaJeztah (approved):
LGTM

(unless we want to consider keeping the `_go116.go` and `_go117.go` files (as I described above) but not a blocker for me

kolyshkin:
> the only part we could consider in this PR is to (if possible) keep the `_go116.go` and `_go117.go` files (while _untested_) to reduce the code-churn and (potentially) make life easier for those maintaining a fork. But if that's not possible, then removing those seems perfectly reasonable to me.

Thank you for your input, I concur we should support the non-obsoleted releases, plus maybe some of the obsoleted ones, on a best-effort basis.

Regarding _go116/7.go files, the only one being removed is for libcontainer/devices unit tests. Since this package uses s/sys/unix, which, in the bumped version, uses `unsafe.Slice`, which requires Go 1.17, it does not make sense to have _go116.go leftovers as they can't be used anyway.

kolyshkin:
I'll go ahead and merge this, to unblock the 1.1.6 release work.

PR #3797, opened 2023-03-29
libct/cg: rm EnterPid

kolyshkin:
Since commit 39914db679b008d74 this function is not used by runc (see that commit to learn why this function is not that good).

I was not able to find any external users either.

Since it's not a good function, with no users, and it is rather trivial, let's remove it right away (rather than mark as deprecated).


kolyshkin:
This is the query I used to look for potential users (besides grepping kubernetes repos):

https://sourcegraph.com/search?q=context:global+EnterPid+lang:go+-file:vendor/+-file:Godeps/&patternType=standard&sm=1&groupBy=repo

AkihiroSuda (approved):

cyphar (approved):
lgtm

Issue #3792, opened 2023-03-29
CI: `mounts.bats` began to fail for rootless on Ubuntu (But still passing on Fedora)


AkihiroSuda:
>  CI began to fail  for Ubuntu 🤦 
> https://github.com/opencontainers/runc/commit/0d62b950e60f6980b54fe3bafd9a9c608dc1df17
> https://github.com/opencontainers/runc/actions/runs/4550888689/jobs/8024349935
>  ```
>  not ok 94 runc run [ro /sys/fs/cgroup mount]
>  # (in test file tests/integration/mounts.bats, line 74)
>  #   `for line in "${lines[@]}"; do [[ "${line}" == *'ro,'* ]]; done' failed
>  # runc spec --rootless (status=0):
>  # 
>  # runc run test_busybox (status=0):
>  # cgroup /sys/fs/cgroup/rdma cgroup ro,nosuid,nodev,noexec,relatime,rdma 0 0
>  # cgroup2 /sys/fs/cgroup/unified cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate 0 0
>  # cgroup /sys/fs/cgroup/systemd cgroup ro,nosuid,nodev,noexec,relatime,xattr,name=systemd 0 0
>  # cgroup /sys/fs/cgroup/net_cls,net_prio cgroup ro,nosuid,nodev,noexec,relatime,net_cls,net_prio 0 0
>  # cgroup /sys/fs/cgroup/devices cgroup ro,nosuid,nodev,noexec,relatime,devices 0 0
>  # cgroup /sys/fs/cgroup/memory cgroup ro,nosuid,nodev,noexec,relatime,memory 0 0
>  # cgroup /sys/fs/cgroup/freezer cgroup ro,nosuid,nodev,noexec,relatime,freezer 0 0
>  # cgroup /sys/fs/cgroup/hugetlb cgroup ro,nosuid,nodev,noexec,relatime,hugetlb 0 0
>  # cgroup /sys/fs/cgroup/cpu,cpuacct cgroup ro,nosuid,nodev,noexec,relatime,cpu,cpuacct 0 0
>  # cgroup /sys/fs/cgroup/misc cgroup ro,nosuid,nodev,noexec,relatime,misc 0 0
>  # cgroup /sys/fs/cgroup/cpuset cgroup ro,nosuid,nodev,noexec,relatime,cpuset 0 0
[...]

PR #3798, opened 2023-03-30
Fix mount test

kolyshkin:
This fixes a few issues with the recently added test case, including #3792. Please see individual commits for details.

Fixes:  #3792

### DIFF
--- a/tests/integration/mounts.bats
+++ b/tests/integration/mounts.bats
@@ -64,17 +64,26 @@ function teardown() {
 	[ "$status" -eq 0 ]
 }
 
-# https://github.com/opencontainers/runc/security/advisories/GHSA-m8cg-xc2p-r3fc
-@test "runc run [ro /sys/fs/cgroup mount]" {
-	# With cgroup namespace
-	update_config '.process.args |= ["sh", "-euc", "for f in `grep /sys/fs/cgroup /proc/mounts | awk \"{print \\\\$2}\"| uniq`; do grep -w $f /proc/mounts | tail -n1; done"]'
-	runc run test_busybox
-	[ "$status" -eq 0 ]
-	[ "${#lines[@]}" -ne 0 ]
-	for line in "${lines[@]}"; do [[ "${line}" == *'ro,'* ]]; done
-
-	# Without cgroup namespace
+@test "runc run [ro /sys/fs/cgroup mounts]" {
+	# Without cgroup namespace.
 	update_config '.linux.namespaces -= [{"type": "cgroup"}]'
+	test_ro_cgroup_mount
+}
+
+# shellcheck disable=SC2030
+@test "runc run [ro /sys/fs/cgroup mounts + cgroupns]" {
+	requires cgroupns
+	# With cgroup namespace.
+	update_config '.linux.namespaces |= if index({"type": "cgroup"}) then . else . + [{"type": "cgroup"}] end'
+	test_ro_cgroup_mount
+}
+
+# https://github.com/opencontainers/runc/security/advisories/GHSA-m8cg-xc2p-r3fc
+# shellcheck disable=SC2031
+function test_ro_cgroup_mount() {
+	local lines status
+	# shellcheck disable=SC2016
+	update_config '.process.args |= ["sh", "-euc", "for f in `grep /sys/fs/cgroup /proc/mounts | awk \"{print \\\\$2}\"| uniq`; do grep -w $f /proc/mounts | tail -n1; done"]'
 	runc run test_busybox
 	[ "$status" -eq 0 ]
 	[ "${#lines[@]}" -ne 0 ]
--- a/tests/integration/mounts.bats
+++ b/tests/integration/mounts.bats
@@ -83,7 +83,7 @@ function teardown() {
 function test_ro_cgroup_mount() {
 	local lines status
 	# shellcheck disable=SC2016
-	update_config '.process.args |= ["sh", "-euc", "for f in `grep /sys/fs/cgroup /proc/mounts | awk \"{print \\\\$2}\"| uniq`; do grep -w $f /proc/mounts | tail -n1; done"]'
+	update_config '.process.args |= ["sh", "-euc", "for f in `grep /sys/fs/cgroup /proc/mounts | awk \"{print \\\\$2}\"| uniq`; do test -e $f && grep -w $f /proc/mounts | tail -n1; done"]'
 	runc run test_busybox
 	[ "$status" -eq 0 ]
 	[ "${#lines[@]}" -ne 0 ]


tests/integration/mounts.bats
```diff
@@ -69,7 +69,7 @@ function teardown() {
 	# Without cgroup namespace.
 	# shellcheck disable=SC2016
 	update_config '	  .linux.namespaces -= [{"type": "cgroup"}]
-			| .process.args |= ["sh", "-euc", "for f in `grep /sys/fs/cgroup /proc/mounts | awk \"{print \\\\$2}\"| uniq`; do grep -w $f /proc/mounts | tail -n1; done"]'
+			| .process.args |= ["sh", "-euc", "for f in `grep /sys/fs/cgroup /proc/mounts | awk \"{print \\\\$2}\"| uniq`; do test -e $f && grep -w $f /proc/mounts | tail -n1; done"]'
```
kolyshkin, line 72: The alternative fix would be to actually mount `/sys/fs/cgroup/unified` into the container (and then remount it read-only). I'd rather not do that.

kolyshkin:
> Wondering why they are not removed (potentially insecure?)

I was hoping you can explain that to me 😆.

My limited understanding is, once we do pivot_root and unmount everything under the old root:
https://github.com/opencontainers/runc/blob/0d62b950e60f6980b54fe3bafd9a9c608dc1df17/libcontainer/rootfs_linux.go#L888
the mountinfo should be clear. It looks like this is what happens when running as root, but not when rootless.

I don't think this is insecure though, unless this is a real mount (shadowed by /sys/fs/cgroup tmpfs). I'll experiment a bit more.

kolyshkin:
> Wondering why they are not removed (potentially insecure?)

OK, what happens here, I assume, is

1. `runc spec --rootless` generates `rbind,ro` mount for /sys. This means that all the mounts under host `/sys` are propagated to the container's `/sys`.
2. Next, `mountCgroupV1` mounts `/sys/fs/cgroup` as tmpfs, shadowing everything that's under it. This is why, by the way, we see two sets of cgroup mounts in a rootless container:

```console
# + grep cgroup /proc/self/mountinfo
# 830 828 0:30 / /sys/fs/cgroup ro,nosuid,nodev,noexec master:9 - tmpfs tmpfs ro,mode=755
# 831 830 0:31 / /sys/fs/cgroup/unified rw,nosuid,nodev,noexec,relatime master:10 - cgroup2 cgroup2 rw,nsdelegate
# 832 830 0:32 / /sys/fs/cgroup/systemd rw,nosuid,nodev,noexec,relatime master:11 - cgroup cgroup rw,xattr,name=systemd
# 833 830 0:36 / /sys/fs/cgroup/rdma rw,nosuid,nodev,noexec,relatime master:16 - cgroup cgroup rw,rdma
# 834 830 0:37 / /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime master:17 - cgroup cgroup rw,cpuset
# 835 830 0:38 / /sys/fs/cgroup/perf_event rw,nosuid,nodev,noexec,relatime master:18 - cgroup cgroup rw,perf_event
# 836 830 0:39 / /sys/fs/cgroup/freezer rw,nosuid,nodev,noexec,relatime master:19 - cgroup cgroup rw,freezer
# 837 830 0:40 / /sys/fs/cgroup/memory rw,nosuid,nodev,noexec,relatime master:20 - cgroup cgroup rw,memory
# 838 830 0:41 / /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime master:21 - cgroup cgroup rw,devices
# 839 830 0:42 / /sys/fs/cgroup/hugetlb rw,nosuid,nodev,noexec,relatime master:22 - cgroup cgroup rw,hugetlb
# 840 830 0:43 / /sys/fs/cgroup/blkio rw,nosuid,nodev,noexec,relatime master:23 - cgroup cgroup rw,blkio
# 841 830 0:44 / /sys/fs/cgroup/cpu,cpuacct rw,nosuid,nodev,noexec,relatime master:24 - cgroup cgroup rw,cpu,cpuacct
# 842 830 0:45 / /sys/fs/cgroup/net_cls,net_prio rw,nosuid,nodev,noexec,relatime master:25 - cgroup cgroup rw,net_cls,net_prio
# 843 830 0:46 / /sys/fs/cgroup/pids rw,nosuid,nodev,noexec,relatime master:26 - cgroup cgroup rw,pids
# 851 830 0:60 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime - tmpfs tmpfs rw,mode=755,uid=2000,gid=2000
# 852 851 0:32 /user.slice/user-1000.slice/session-1.scope /sys/fs/cgroup/systemd ro,nosuid,nodev,noexec,relatime master:11 - cgroup cgroup rw,xattr,name=systemd
# 853 851 0:36 / /sys/fs/cgroup/rdma ro,nosuid,nodev,noexec,relatime master:16 - cgroup cgroup rw,rdma
# 854 851 0:37 / /sys/fs/cgroup/cpuset ro,nosuid,nodev,noexec,relatime master:17 - cgroup cgroup rw,cpuset
# 855 851 0:38 / /sys/fs/cgroup/perf_event ro,nosuid,nodev,noexec,relatime master:18 - cgroup cgroup rw,perf_event
# 856 851 0:39 / /sys/fs/cgroup/freezer ro,nosuid,nodev,noexec,relatime master:19 - cgroup cgroup rw,freezer
# 857 851 0:40 /user.slice/user-1000.slice/session-1.scope /sys/fs/cgroup/memory ro,nosuid,nodev,noexec,relatime master:20 - cgroup cgroup rw,memory
# 858 851 0:41 /user.slice /sys/fs/cgroup/devices ro,nosuid,nodev,noexec,relatime master:21 - cgroup cgroup rw,devices
# 859 851 0:42 / /sys/fs/cgroup/hugetlb ro,nosuid,nodev,noexec,relatime master:22 - cgroup cgroup rw,hugetlb
# 860 851 0:43 /user.slice /sys/fs/cgroup/blkio ro,nosuid,nodev,noexec,relatime master:23 - cgroup cgroup rw,blkio
# 861 851 0:44 /user.slice /sys/fs/cgroup/cpu,cpuacct ro,nosuid,nodev,noexec,relatime master:24 - cgroup cgroup rw,cpu,cpuacct
# 862 851 0:45 / /sys/fs/cgroup/net_cls,net_prio ro,nosuid,nodev,noexec,relatime master:25 - cgroup cgroup rw,net_cls,net_prio
# 863 851 0:46 /user.slice/user-1000.slice/session-1.scope /sys/fs/cgroup/pids ro,nosuid,nodev,noexec,relatime master:26 - cgroup cgroup rw,pids
```

This is also why you had to do `tail -n1` in your test case.

3. On top of that tmpfs, we bind mount host /sys/fs/cgroup/$NAME to container's /sys/fs/cgroup/$NAME. This is done for all cgroup mounts except `/sys/fs/cgroup/unified` since `cgroups.GetCgroupMounts` does not return it in case of cgroup v1.

kolyshkin:
I have no idea of how to unmount those "extra/no-longer-needed" mounts though, as either of these syscalls return EINVAL. I tried:
```go
unmount(path.Join(c.root, m.Destination) // m.Destination == "/sys/fs/cgroup"
```
and
```go
unmount(path.Join(c.root, "/sys/fs/cgroup/unified")
```
and they both return EINVAL.

I did that right before doing the tmpfs mount for /sys/fs/cgroup, as after that mount has happened, the directory is shadowed and can't be accessed by umount(2).

My guess, it might work later, in pivotRoot, but at that time the /sys/fs/cgroup directory is shadowed already.

So, unless someone has any new ideas, those mounts has to stay.

kolyshkin:
Ah OK, so maybe commit ff692f289b60e19b3079cbbbaa14c426225dc20f inadvertently introduced this double-set-of-mounts. Before it, we had no cgroup mount for `/sys/fs/cgroup` in the rootless spec, meaning those mounts propagated from the `/sys` rbind were used.

On the positive side, before that we had the whole hierarchy of /sys/fs/cgroup inside every rootless container, which is not good either I guess.

@AkihiroSuda @cyphar any thoughts?

AkihiroSuda (dismissed):

tests/integration/mounts.bats
```diff
@@ -66,15 +66,18 @@ function teardown() {
...
 	runc run test_busybox
 	[ "$status" -eq 0 ]
 	[ "${#lines[@]}" -ne 0 ]
 	for line in "${lines[@]}"; do [[ "${line}" == *'ro,'* ]]; done
 
-	# Without cgroup namespace
-	update_config '.linux.namespaces -= [{"type": "cgroup"}]'
+	# With cgroup namespace, if available.
+	[ ! -e "/proc/self/ns/cgroup" ] && return
```
cyphar, line 79: Seems like this test should be split, with one test having `requires cgroupns`. Otherwise we'll have a test where pass might mean only half of the test was actually run.

cyphar:
In a rootless container I'm pretty sure you can't unmount the cgroupfs mount no matter what you try (the kernel considers it as a revealing mount IIRC). Having some dummy information in `/proc/self/mountinfo` isn't the end of the world -- IIRC there was some time in the past when you had even more host information visible in `mountinfo`.

tests/integration/mounts.bats
kolyshkin: This is fine with me, because there's absolutely no way to run the other half of the test anyway. The only difference is we won't know if the second part was run or not.

Nevertheless, let me split this.

kolyshkin: done; alas in the process I was not able to have such a fine grained ("one fix -- one commit") patch series.

AkihiroSuda (approved):

kolyshkin:
1.1. backport: https://github.com/opencontainers/runc/pull/3799

cyphar (approved):
lgtm

Issue #3803, opened 2023-04-03
Still got an occasional "permission denied" error  while cgroup frozen failed.


jiusanzhou:
### Description

Currenty we have frozen the container to avoid the container get
an occasional "permission denied" error, while the systemd's application of device
rules is done disruptively.

But when the processes in the container can not be frozen over 2 seconds, we still update the cgroup which resulting the container get
an occasional "permission denied" error in some cases.

The code in `libcontainer/cgroups/systemd/v1.go`
```go
func (m *LegacyManager) Set(r *configs.Resources) error {
	if needsFreeze {
		if err := m.doFreeze(configs.Frozen); err != nil {
			// If freezer cgroup isn't supported, we just warn about it.
			logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
		}
	}
	setErr := setUnitProperties(m.dbus, unitName, properties...)
}
```

btw, 2 seconds set in `libcontainer/cgroups/fs/freezer.go` 
```go
func (s *FreezerGroup) Set(path string, r *configs.Resources) (Err error) {
	for i := 0; i < 1000; i++ {
		time.Sleep(10 * time.Microsecond)
	}
}
```

### Steps to reproduce the issue

1. Kubelet and containerd, which with systemd v1 as cgroup driver;
2. Start a container which start a process that reads the urandom device 1 time per second;
3. Start some process with D state in the container.

### Describe the results you received and expected

Avoid an occasional "permission denied" error.

### What version of runc are you using?

1.1.2

jiusanzhou:
Sorry I gave a wrong version, it's actually 1.1.2
According to the latest code, I think this issue still exists.

PR #3804, opened 2023-04-03
libct/cg/sd/v1: do not update non-frozen cgroup after frozen failed.

jiusanzhou:
In code we have frozen the cgroup to avoid the processes get
an occasional "permission denied" error, while the systemd's application of device
rules is done disruptively. When the processes in the container can not
be frozen over 2 seconds (which defined in fs/freezer.go),
we still update the cgroup which resulting the container get an occasional
"permission denied" error in some cases.

This PR make sure to return error directly without updating cgroup, when freeze fails.

Fixes: #3803

### DIFF
--- a/libcontainer/cgroups/systemd/v1.go
+++ b/libcontainer/cgroups/systemd/v1.go
@@ -350,6 +350,15 @@ func (m *LegacyManager) Set(r *configs.Resources) error {
 		if err := m.doFreeze(configs.Frozen); err != nil {
 			// If freezer cgroup isn't supported, we just warn about it.
 			logrus.Infof("freeze container before SetUnitProperties failed: %v", err)
+			// skip update the cgroup while frozen failed. #3803
+			if !errors.Is(err, errSubsystemDoesNotExist) {
+				if needsThaw {
+					if thawErr := m.doFreeze(configs.Thawed); thawErr != nil {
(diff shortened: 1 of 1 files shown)


kolyshkin (commented):
Overall the fix makes sense to me -- we should not update a non-frozen container.

Will take a deeper look later.

kolyshkin:
@jiusanzhou you need to rebase and fix linter warnings (the ones from `validate` and `commit` CI jobs look legit)

kolyshkin:
For the commit subject, I suggest something like

`libct/cg/sd/v1: do not update non-frozen cgroup`

For the description, you can get some text from #3803 to describe what happens and then go on to say how the commit is fixing this.

jiusanzhou:
@kolyshkin thank you for your advice, CI checking is green now. Please take another look.

jiusanzhou:
@cyphar PTAL

zvier:
any update？

kolyshkin:
@thaJeztah @AkihiroSuda @cyphar PTAL

kolyshkin:

AkihiroSuda:
Can we have a test?

kolyshkin:
> Can we have a test?

I'm not sure we can. To test the branch of code being added, we need cgroup freeze to fail, and I don't know how to do that reliably.

jiusanzhou:
1.1 backport: #3921

